PR #2780, opened 2023-05-25
feat: formatter updated

nityanandagohain:
Adds int32 and in64


pkg/query-service/utils/format.go
```diff
@@ -144,7 +144,7 @@ func ValidateAndCastValue(v interface{}, dataType v3.AttributeKeyDataType) (inte
 // ClickHouseFormattedValue formats the value to be used in clickhouse query
 func ClickHouseFormattedValue(v interface{}) string {
 	switch x := v.(type) {
-	case int:
+	case int, int32, int64:
```
srikanthccv, line 147: Add the same below for array `x[0].(type)` type cases. Also, It doesn't hurt to keep all integer types here.

nityanandagohain: Added them

srikanthccv (approved):

makeavish (approved):

Issue #1988, opened 2023-01-06
Caching the results of `query_range` apis for overlapping period


ankitnayan:
We can speed up loading of dashboards when there is an overlap of time period of `query_range` APIs querying the same data.

ankitnayan:
We should log how much of cache is hit for a query.

I am guessing there will be some randomness in query response time depending on which hit the cache. The above will help us debug better

srikanthccv:
Yes, we log the cache retrieve status https://github.com/SigNoz/signoz/blob/9826ab04b3faa5d991e46be8e9ea8630f48f68ed/pkg/query-service/app/querier/helper.go#L163C18-L163C33. The log level should be changed.

PR #2782, opened 2023-05-26
feat: add querier interface and initial implementation

srikanthccv:
Part of #1988 

This introduces a querier interface which will be responsible for running the queries for query range requests with optional caching. The current implementation doesn't do any caching for traces and logs query range requests and doesn't support ClickHouse query since it can be very dynamic in nature. Otherwise, the following is the flow.

- Create a cache key for the builder query
- See if there is cached data from the previous run
- Find the missing time range by comparing the current start, and end with the cached start and end
- Query data for the missing time ranges
- Prepare the new merged series from the cache data and query data
- Store this for the potential next

There are some configurable options, 1. Flux interval - querying and caching the more recent data could give incorrect results since some data might not arrive yet. This setting prevents the caching of very recent results that might still be in flux. Default 5 minutes. If you believe this is low, we can bump it up. 2. NoCache - Option for the client to disable the caching entirely.


pkg/query-service/app/querier/querier.go
```diff
@@ -0,0 +1,429 @@
...
+		err = fmt.Errorf("error in builder queries")
+	}
+	return seriesList, err, errQueriesByName
+}
+
+func (q *querier) runPromQueries(ctx context.Context, params *v3.QueryRangeParamsV3) ([]*v3.Series, error, map[string]string) {
+	seriesList := make([]*v3.Series, 0)
+	errQueriesByName := make(map[string]string)
+	var err error
+	for queryName := range params.CompositeQuery.PromQueries {
```
nityanandagohain, line 332: same here

pkg/query-service/app/querier/querier.go
```diff
@@ -0,0 +1,429 @@
...
+		}
+		if err != nil {
+			errQueriesByName[queryName] = err.Error()
+			continue
+		}
+		misses := q.findMissingTimeRanges(params.Start, params.End, params.Step, cachedData)
+		missedSeries := make([]*v3.Series, 0)
+		cachedSeries := make([]*v3.Series, 0)
+		for _, miss := range misses {
+			query := metricsV3.BuildPromQuery(
```
nityanandagohain, line 349: same we can push these to goruotines

pkg/query-service/app/querier/querier_test.go
```diff
@@ -0,0 +1,511 @@
...
+			},
+		},
+	}
+	q := NewQuerier(opts)
+	expectedTimeRangeInQueryString := []string{
+		fmt.Sprintf("timestamp_ms >= %d AND timestamp_ms <= %d", 1675115596722, 1675115596722+120*60*1000),
+		fmt.Sprintf("timestamp_ms >= %d AND timestamp_ms <= %d", 1675115596722+120*60*1000+1, 1675115596722+180*60*1000),
+	}
+
+	fmt.Println("queries executed", q.QueriesExecuted())
```
nityanandagohain, line 496: remove

nityanandagohain (commented):
Looks good overall. Moving some of the db calls to goroutines can be taken care in this PR or you can create a new one, will leave it to you.

nityanandagohain (approved):

srikanthccv:
Merging this, it has been around two weeks.

PR #2784, opened 2023-05-26
fix: null value on select of chip

techchintan:


frontend/src/container/QueryBuilder/filters/QueryBuilderSearch/index.tsx
```diff
@@ -41,10 +45,42 @@ function QueryBuilderSearch({
...
+				  !!query.aggregateAttribute.dataType
+				: true,
+		[
+			query.aggregateAttribute.dataType,
+			query.aggregateOperator,
+			query.dataSource,
+		],
+	);
+
+	const { data } = useQuery(
```
palashgdev, line 64: please make the hook for this one as we are already using the same

palashgdev (changes requested):

PR #2790, opened 2023-05-29
feat: add zap otel logger support

makeavish:


srikanthccv:
IIRC, there was a requirement to send only those logs with the request time > x, When does it happen?

makeavish:
> IIRC, there was a requirement to send only those logs with the request time > x, When does it happen?

No, I remember we decided to send all logs and create alerts based on errors on some critical endpoints.

srikanthccv:
@nityanandagohain is the zap-otlp part ready to use? Or is there anything pending before we put it in production?

ee/query-service/main.go
```diff
@@ -3,25 +3,66 @@ package main
...
+	core := zapcore.NewTee(
+		zapcore.NewCore(consoleEncoder, os.Stdout, defaultLogLevel),
+	)
+
+	conn, err := grpc.DialContext(ctx, constants.OTLPTarget, grpc.WithBlock(), grpc.WithInsecure(), grpc.WithTimeout(time.Second*10))
+	if err != nil {
+		log.Println("failed to connect to otlp collector to export query service logs with error:", err)
+	} else {
+		ws := zapcore.AddSync(zapotlpsync.NewOtlpSyncer(conn, zapotlpsync.Options{
+			BatchSize:      2,
```
nityanandagohain, line 55: Increase batchSize/use from env

nityanandagohain (commented):
This seems to be updating the global logger, which might case some issue if by default we enable otlp logging. 

I.e  it is a self hosted instance ex:- docker deployment then the logs will appear twice one from otlp and other from filelog.

I think otlp should only be enabled when data is to be sent to our collector so that we can check errors of our users.

ee/query-service/main.go
makeavish: Done

nityanandagohain (dismissed):

nityanandagohain:
Ahh missed it, for the usecase we are solving don't think anything else is required.

Also since this won't be enabled by default we can enable it for our own deployments monitor it for few days and then enable it for our sass users one one. cc @makeavish

srikanthccv:
I thought there was more work after the initial commit to the zap-otlp repo. I would suggest we do some testing to make sure this shouldn't affect the query-service in any condition.

nityanandagohain:
Do you suggest deploying this build in one of our own instance ?

makeavish:
@srikanthccv Added an env variable `ENABLE_QUERY_SERVICE_LOG_OTLP_EXPORT`

nityanandagohain:
@makeavish please add command line flag instead of env as discussed

nityanandagohain:
Command line argument will be more explicit call to enable it, as discussed in the meeting.

makeavish:
Added command line argument for to enable log export.
@srikanthccv @nityanandagohain

ee/query-service/main.go
```diff
@@ -3,25 +3,73 @@ package main
...
-func initZapLog() *zap.Logger {
+func initZapLog(enableQueryServiceLogOTLPExport bool) *zap.Logger {
 	config := zap.NewDevelopmentConfig()
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
+	defer stop()
+
+	config.EncoderConfig.EncodeDuration = zapcore.StringDurationEncoder
+	otlpEncoder := zapotlpencoder.NewOTLPEncoder(config.EncoderConfig)
+	consoleEncoder := zapcore.NewConsoleEncoder(config.EncoderConfig)
+	defaultLogLevel := zapcore.DebugLevel
```
srikanthccv, line 37: info level

makeavish: I don't think we should change log level in this PR, I have seen that we are logging at Debug level at some places where level should be info or error.

makeavish: Created an issue for same: https://github.com/SigNoz/signoz/issues/2921

srikanthccv (approved):

nityanandagohain (approved):

Issue #2687, opened 2023-05-12
[Logs] Refactor metadata enrichment in QB


nityanandagohain:
As of now metadata enrichment is spread over in the parser we can refractor it to be done at an initial step and the parser functions can assume that metadata will be present.

Issue #2374, opened 2023-02-24
(logs): limit, offset and pagesize


nityanandagohain:
- Pagination in list view without aggregation. - 0.5D
    - Id for timestamp
    - limit offset for others
- Pagination with aggregation    - 0.5D
    - ID for timestamp
    - limit offset for others

PR #2791, opened 2023-05-29
Fixes for pagination in listview

nityanandagohain:
it has some small fixes which make pagination correct fixes https://github.com/SigNoz/signoz/issues/2374
also fixes https://github.com/SigNoz/signoz/issues/2687

* Check if metadata enrichment is needed
* If needed get the metadata from db and enrich the query
* Orderby logic updated for list view.


srikanthccv (commented):
Some questions; still need to look at the order by.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -4199,15 +4199,15 @@ func (r *ClickHouseReader) GetListResultV3(ctx context.Context, query string) ([
...
-		vars[i] = reflect.New(columnTypes[i].ScanType()).Interface()
-	}
 
 	var rowList []*v3.Row
 
 	for rows.Next() {
+		var vars = make([]interface{}, len(columnTypes))
+		for i := range columnTypes {
+			vars[i] = reflect.New(columnTypes[i].ScanType()).Interface()
+		}
```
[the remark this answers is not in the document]
nityanandagohain: So if we don't move this inside the loop all of the data get set to the last value read by the vars variable.

pkg/query-service/app/logs/v3/enrich_query.go
```diff
@@ -0,0 +1,141 @@
...
+	return false
+}
+
+func isEnriched(field v3.AttributeKey) bool {
+	// if it is timestamp/id dont check
+	if field.Key == "timestamp" || field.Key == "id" {
+		return true
+	}
+
+	if field.Type == "" || field.DataType == "" {
```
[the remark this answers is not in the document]
nityanandagohain: Makes sense, added check for `isColumn` along with a test for it.

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -376,33 +377,37 @@ func reduceQuery(query string, reduceTo v3.ReduceToOperator, aggregateOperator v
...
-		query = addOffsetToQuery(query, mq.Offset)
+	if panelType == v3.PanelTypeList || panelType == v3.PanelTypeTable {
+		if mq.Limit == 0 {
+			// 100 is the default max limit as well as page limit
+			mq.Limit = 100
+		}
+		if mq.PageSize > 0 {
+			if (mq.Offset + mq.PageSize) > mq.Limit {
+				return "", fmt.Errorf("max limit exceeded")
+			}
```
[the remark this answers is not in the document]
nityanandagohain: Didn't get you on this properly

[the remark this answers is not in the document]
srikanthccv: Should you return an error for Offset : 800, PageSize: 25, Limit: 812?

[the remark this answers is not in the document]
nityanandagohain: Ahh makes sense

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -99,8 +72,12 @@ func getClickhouseLogsColumnDataType(columnDataType v3.AttributeKeyDataType) str
 }
 
 // getClickhouseColumnName returns the corresponding clickhouse column name for the given attribute/resource key
-func getClickhouseColumnName(key v3.AttributeKey, fields map[string]v3.AttributeKey) (string, error) {
+func getClickhouseColumnName(key v3.AttributeKey) (string, error) {
 	clickhouseColumn := key.Key
+	if key.Key == "timestamp" || key.Key == "id" {
+		return key.Key, nil
+	}
```
srikanthccv, line 79: Do you need this conditional block here? Since both are top-level columns, `return clickhouseColumn, nil` below does the same?

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -330,9 +320,20 @@ func orderBy(items []v3.OrderBy, tags []string) string {
 	for _, item := range items {
 		if item.ColumnName == constants.SigNozOrderByValue {
 			orderBy = append(orderBy, fmt.Sprintf("value %s", item.Order))
+			addedToGroupBy[item.ColumnName] = true
```
srikanthccv, line 323: Can you add a test for this? I think the above loop also adds a `#SIGNOZ_VALUE ASC/DESC` to the order by which will not exist in the select labels as `SIGNOZ_VALUE`.

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -330,9 +320,20 @@ func orderBy(items []v3.OrderBy, tags []string) string {
...
+		}
+	}
+
+	// add the remaining items
+	for _, item := range items {
+		// since these are not present in tags we will have to select them correctly
+		if !addedToGroupBy[item.ColumnName] {
+			attr := v3.AttributeKey{Key: item.ColumnName, DataType: item.DataType, Type: item.Type, IsColumn: item.IsColumn}
+			name, _ := getClickhouseColumnName(attr)
```
srikanthccv, line 332: You don't seem to return an error from `getClickhouseColumnName`. You can change the function signature instead of ignoring with `_` here.

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -330,9 +320,20 @@ func orderBy(items []v3.OrderBy, tags []string) string {
 	for _, item := range items {
 		if item.ColumnName == constants.SigNozOrderByValue {
 			orderBy = append(orderBy, fmt.Sprintf("value %s", item.Order))
+			addedToGroupBy[item.ColumnName] = true
+		}
+	}
+
+	// add the remaining items
+	for _, item := range items {
+		// since these are not present in tags we will have to select them correctly
```
srikanthccv, line 329: You need to handle this based on aggregation type and panel type. Say the aggregation is `SUM` and group by `service_name`. The OrderBy can't be anything other than `value` or `service_name`. But this does seem to add everything.

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -330,9 +320,20 @@ func orderBy(items []v3.OrderBy, tags []string) string {
...
+		if !addedToGroupBy[item.ColumnName] {
+			attr := v3.AttributeKey{Key: item.ColumnName, DataType: item.DataType, Type: item.Type, IsColumn: item.IsColumn}
+			name, _ := getClickhouseColumnName(attr)
+			orderBy = append(orderBy, fmt.Sprintf("%s %s", name, item.Order))
```
srikanthccv, line 333: >since these are not present in tags we will have to select them correctly

Are these part of the select labels? (i.e a label need to be in SELECT ... before we can use them in orderby)

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -285,8 +265,12 @@ func buildLogsQuery(start, end, step int64, mq *v3.BuilderQuery, fields map[stri
 		query := fmt.Sprintf(queryTmpl, step, op, filterSubQuery, groupBy, having, orderBy)
 		return query, nil
 	case v3.AggregateOperatorNoOp:
-		queryTmpl := constants.LogsSQLSelect + "from signoz_logs.distributed_logs where %s %s"
-		query := fmt.Sprintf(queryTmpl, timeFilter, filterSubQuery)
+		queryTmpl := constants.LogsSQLSelect + "from signoz_logs.distributed_logs where %s %sorder by %s"
+		if orderBy == "" {
+			orderBy = "timestamp"
```
srikanthccv, line 270: You wanted to use `id` here?

srikanthccv (commented):
Some more comments about the orderby

pkg/query-service/app/logs/v3/query_builder.go
nityanandagohain: As we have decided to return all fields instead of user selecting some fields in the api response. So orderby doesn't need to check select.

pkg/query-service/app/logs/v3/query_builder.go
nityanandagohain: The reason for this is that in the `attributes` api we don't return `timestamp` and `id` . So the frontend won't have any context 

ex:-
```
"A": {
                "queryName": "A",
                "dataSource": "logs",
                "aggregateOperator": "noop",
                "expression": "A",
                "disabled": false,
                "pageSize": 10,
                "filters": {
                    "items": [
                        {
                            "key": {
                                "key": "method",
                                "type": "tag",
                                "dataType": "string"
                            },
                            "op": "=",
                            "value": "GET"
                        }
                    ],
                    "op": "AND"
                },
                "orderBy": [
                    {
                        "columnName": "timestamp",
                        "order": "desc"
                    }
                ]
            }
   ```
   
   Here we don't need to enrich.

File: pkg/query-service/app/logs/v3/query_builder.go (hunk 6 above)
Comment (nityanandagohain) [reply pkg/query-service/app/logs/v3/query_builder.go]: Yeah thanks, have updated the function signature.

File: pkg/query-service/app/logs/v3/query_builder.go (hunk 7 above)
Comment (nityanandagohain) [reply pkg/query-service/app/logs/v3/query_builder.go:316]: I think jusing keeping this for panel type `list` will sove it.

Comment (nityanandagohain) [reply pkg/query-service/app/logs/v3/query_builder.go]: Have added the changes with tests.

File: pkg/query-service/app/logs/v3/query_builder.go (hunk 9 above)
Comment (nityanandagohain) [reply pkg/query-service/app/logs/v3/query_builder.go:257]: I can see where you are coming from but default is fine and the rows will appear as they were ingested.

File: pkg/query-service/app/logs/v3/query_builder.go (hunk 5 above)
Comment (nityanandagohain) [reply pkg/query-service/app/logs/v3/query_builder.go:310]: Didn't get you properly on this

File: pkg/query-service/app/clickhouseReader/reader.go (hunk 1 above)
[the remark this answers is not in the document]
Comment (nityanandagohain) [reply pkg/query-service/app/clickhouseReader/reader.go:4210]: So in `GetTimeSeriesResultV3` you copy them to a new variable using the `readRow` function

File: pkg/query-service/app/logs/v3/query_builder.go [hunk 10]
```diff
@@ -330,17 +307,30 @@ func orderBy(items []v3.OrderBy, tags []string) string {
...
+			// since these are not present in tags we will have to select them correctly
+			if !addedToGroupBy[item.ColumnName] {
+				attr := v3.AttributeKey{Key: item.ColumnName, DataType: item.DataType, Type: item.Type, IsColumn: item.IsColumn}
+				name := getClickhouseColumnName(attr)
+				orderBy = append(orderBy, fmt.Sprintf("%s %s", name, item.Order))
+			}
 		}
 	}
-	return strings.Join(orderBy, ",")
+	return strings.Trim(strings.Join(orderBy, ","), ",")
```
srikanthccv, line 325: nit: this `Trim` is redundant, Join doesn't add a separator at the last https://go.dev/play/p/0UGbGI6Rz7Z

pkg/query-service/app/logs/v3/query_builder.go
srikanthccv: I think I misunderstood. Disregard this comment.

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -330,17 +307,30 @@ func orderBy(items []v3.OrderBy, tags []string) string {
...
 			orderBy = append(orderBy, fmt.Sprintf("value %s", item.Order))
+			addedToGroupBy[item.ColumnName] = true
+		}
+	}
+
+	// add the remaining items
+	if panelType == v3.PanelTypeList {
+		for _, item := range items {
+			// since these are not present in tags we will have to select them correctly
+			if !addedToGroupBy[item.ColumnName] {
```
srikanthccv, line 318: The list type won't have the groupby since we show the raw data. But I don't see that as a problem here since `addedToGroupBy` will be empty. You can may be write some comments for the reader who goes through this code.

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -225,18 +189,22 @@ func buildLogsQuery(start, end, step int64, mq *v3.BuilderQuery, fields map[stri
...
 
 	groupBy := groupByAttributeKeyTags(mq.GroupBy...)
-	orderBy := orderByAttributeKeyTags(mq.OrderBy, mq.GroupBy)
+	orderBy := orderByAttributeKeyTags(panelType, mq.OrderBy, mq.GroupBy)
+
+	if mq.AggregateOperator != v3.AggregateOperatorNoOp {
+		if !strings.Contains(orderBy, "ts") {
+			orderBy = orderBy + "ts"
+		}
+		orderBy = strings.Replace(orderBy, "timestamp", "ts", 1)
```
srikanthccv, line 201: I would avoid the `.Contains` and `.Replace`. It's easy to have a case where this will break. Will this still work if there is a different order by could that contains substring `ts`, for instance, `requests_order` or similar? Will this still work if a column appears at the beginning of orderBy, which contains `timestamp` as a substring?

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -376,33 +366,37 @@ func reduceQuery(query string, reduceTo v3.ReduceToOperator, aggregateOperator v
...
 	}
 	if panelType == v3.PanelTypeValue {
 		query, err = reduceQuery(query, mq.ReduceTo, mq.AggregateOperator)
 	}
 
-	query = addLimitToQuery(query, mq.Limit, panelType)
-
-	if mq.Offset != 0 {
-		query = addOffsetToQuery(query, mq.Offset)
+	if panelType == v3.PanelTypeList || panelType == v3.PanelTypeTable {
```
srikanthccv, line 386: Read this https://signoz-team.slack.com/archives/C02B85R4R0A/p1685603948437199. You will have to update this for table view.

srikanthccv (commented):
Some questions, mainly the string manipulation. Otherwise mostly LGTM.

pkg/query-service/app/logs/v3/query_builder.go
nityanandagohain: makes sense

pkg/query-service/app/logs/v3/query_builder.go
nityanandagohain: So from the convesation it seems that the limit for table will be similar to timeseries and implementation has to be decided right ?

Let me remove paneltype from the above condition

pkg/query-service/app/logs/v3/query_builder.go
nityanandagohain: Makes sense, have removed this logic

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -376,33 +369,37 @@ func reduceQuery(query string, reduceTo v3.ReduceToOperator, aggregateOperator v
...
+		if mq.Limit == 0 {
+			// 100 is the default max limit as well as page limit
+			mq.Limit = 100
+		}
+		if mq.PageSize > 0 {
+			if mq.Offset > mq.Limit {
+				return "", fmt.Errorf("max limit exceeded")
+			}
+			query = addLimitToQuery(query, mq.PageSize)
+			query = addOffsetToQuery(query, mq.Offset)
```
srikanthccv, line 399: Please help me with this. I want to browse several pages until I find what I want. The limit is not set. I set the page size to 50. After three pages, will this give the results or throw an error?

srikanthccv (approved):
One last question.

pkg/query-service/app/logs/v3/query_builder.go
nityanandagohain: Yeah, this case you mentioned will result in error after three pages. From my side I don't think that should be an issue.

What are your thoughts ?

srikanthccv: I believe that is unexpected from an end-user POV. When a limit is not applied, it shouldn't give an error when I go to the third/fourth page. Can you share why you want to raise `max limit exceeded`? Like what was the reasoning to introduce this error?

nityanandagohain: In in table view(groupby), the UI will have some set default limit which the user can change. Now during pagination if the limit is hit then it should be stopped, thats is POV for adding this error.

srikanthccv: It sets the limit as null by default. It has to be changed by the user. Should you be setting the limit (to 100) when no limit is applied on the frontend? It is fair to throw an error if they _set_ the limit and request for data beyond that, but if they don't set any limit and ask for data, we should return data.

nityanandagohain: Yeah don't see an issue with that let me add it.

srikanthccv (dismissed):

nityanandagohain:
@srikanthccv  please approve again, as I had to update the querier interface with these changes

srikanthccv (approved):

Issue #1015, opened 2022-04-24
The install script should allow no input to proceed when asking for email


JLHasson:
## Bug description

At the end of the installer you are prompted to enter your email "to receive support & updates". Many users will not want to enter their email at this prompt. Currently pressing "enter" (i.e. no input) repeats the prompt indefinitely. It is impossible to know as the user of this script if the script has completed or if there is some further code that will run after entering your email. The script should accept no input & proceed to exit.

## Expected behavior

You should be able to press enter to proceed with the script without entering an email.

Workaround: just enter some other random text
```
📨 Please share your email to receive support & updates about SigNoz!
Email:
Email:
Email: no

🙏 Thank you!
```

## How to reproduce

1. run `./install.sh`
2. when you get to the prompt
```
📨 Please share your email to receive support & updates about SigNoz!
Email:
```
press enter
3. see that it does not accept this as valid input

## Version information 
* **Signoz version**: 
* **Browser version**:
* **Your OS and version**: ubuntu 20.04
* **Your CPU Architecture**(ARM/Intel):

pranay01:
Thanks for opening this issue @JLHasson Looks like a simple improvement we can make :)

rharyanp:
great

Nageshbansal:
Hey, have you tried https://github.com/signoz/foundry? It removes the necessity of using the `install.sh` process.

Nageshbansal:
Hey, closing this with the `install.sh` deprecation (#10924). We're replacing the script with [Foundry](https://github.com/SigNoz/foundry), which doesn't collect any PII; its usage analytics are anonymous and opt-out, documented in the [ledger doc](https://github.com/SigNoz/foundry/blob/main/docs/ledger.md).

Setup with Foundry: https://signoz.io/docs/install/docker/. Thanks for raising it.

PR #2794, opened 2023-05-29
Bugfix: installer blocks for user input in non-interactive sessions.

andrewjanssen:
Per title. Without this patch, my Ansible script blocks when invoking `deploy/installer.sh` because that script blocks for user input. Using `deploy/installer.sh` in automations would be very valuable. See #1015.

Thank you SigNoz maintainers for this awesome project!


prashant-shahi:
@andrewjanssen It seem to just bypass the email flow even in terminal of local machine and VM since it is detecting it to be _non-interactive_.

Perhaps you could use command argument instead to pass email id: `./install.sh test@domain.com`
It could look something like:

```bash
# Checking OS and assigning package manager
desired_os=0
os=""
email="$1"

...

if [[ $email == "" ]]; then
        echo -e "\n📨 Please share your email to receive support & updates about SigNoz!"
        read -rp 'Email: ' email

        while [[ $email == "" ]]
        do
            read -rp 'Email: ' email
        done
fi
```

PR #2810, opened 2023-06-02
fix: org name is fixed

### LINKED ISSUE #2820: Org name missing and unable to signup using invite link
New users who signup with a valid org name don't see org name after refresh:

image

Invited users can't signup as org name is missing:

image

palashgdev:
Fixes #2820


prashant-shahi (approved):
LGTM.

I see that multiple API calls are happening for updating both user and org meta.
We could improve on that in another PR after internal discussion.

image

Issue #2823, opened 2023-06-05
Show all the operations in `Key Operations` table


srikanthccv:
The list of operations we currently show are top 10 by P99. Top P99 doesn't always mean they are interested in it because it could be that by nature, the P99 for the service is high, which is not very useful.

prashant-shahi:
I had faced this issue of not seeing desired operation under Key Operations, which is why the custom dashboards for showing all operations was introduced: https://github.com/SigNoz/dashboards/tree/main/key-operations

@srikanthccv I have a doubt though. Wouldn't it be cluttered with too many operations? Will there be an option to choose desired number of operations? It could be dropdown list to select 10,20,50,100, etc.

srikanthccv:
It will be paginated, and users can reorder the data by asc/desc of any column.

PR #2824, opened 2023-06-05
fix: remove the top 10 limit on key operations table

srikanthccv:
Fixes https://github.com/SigNoz/signoz/issues/2823

### DIFF
--- a/pkg/query-service/app/clickhouseReader/reader.go
+++ b/pkg/query-service/app/clickhouseReader/reader.go
@@ -1870,14 +1870,18 @@ func (r *ClickHouseReader) GetTopOperations(ctx context.Context, queryParams *mo
 	if errStatus != nil {
 		return nil, errStatus
 	}
-	query += " GROUP BY name ORDER BY p99 DESC LIMIT 10"
+	query += " GROUP BY name ORDER BY p99 DESC"
+	if queryParams.Limit > 0 {
+		query += " LIMIT @limit"
+		args = append(args, clickhouse.Named("limit", queryParams.Limit))
+	}
 	err := r.db.Select(ctx, &topOperationsItems, query, args...)
 
 	zap.S().Debug(query)
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


makeavish (commented):
We also need to add offset param for pagination.

srikanthccv:
That is out of the scope of this issue. The Client never sends any limit or offset or searchText here.

makeavish:
> That is out of the scope of this issue. The Client never sends any limit or offset or searchText here.

Client has to send that as without limit and offset pagination won't work. Without pagination there can performance issues.

srikanthccv:
I am saying that is out of the scope. The Client never sends the offset and limit since the beginning. The pagination that will work is the default pagination by antd table.

srikanthccv:
I created a separate issue here https://github.com/SigNoz/signoz/issues/2833

makeavish:
> I am saying that is out of the scope. The Client never sends the offset and limit since the beginning. The pagination that will work is the default pagination by antd table.

Limit has been added as a param here, so I think offset should also be added. Shipping this without offset might cause issues as operations can be in thousands even for small systems.
If https://github.com/SigNoz/signoz/issues/2833 will be shipped along with this PR then we can merge this PR.

srikanthccv:
I honestly don't know why I add the Limit param.

>as operations can be in thousands even for small systems

I haven't seen a small system with thousands of operations. Even the most powerful user we know today, wombo has few hundred operations combined for all of their services. It is usually limited because the number of APIs is small, and third-party lib calls such as HTTP or DB also produce a small number of operations. This has been the experience with other users also. If this breaks here, this breaks in other pages also because we don't query with limit and offset or searchText. My broader point is that this is a generic problem across the system. It is better to address them holistically or even discuss when we choose local antd pagination and search vs backend powered pagination and search.

I don't mind changing it if you feel strongly about this.

makeavish (approved):

Issue #2840, opened 2023-06-06
[FE-Logs]: Add option to download logs data


techchintan:
Add Option to download logs data as a file

1. CSV
2. Excel(xlsx)

PR #2841, opened 2023-06-06
feat: added option to download logs in csv & excel format

techchintan:
Close: https://github.com/SigNoz/signoz/issues/2840


frontend/src/container/LogControls/index.tsx
```diff
@@ -72,12 +75,61 @@ function LogControls(): JSX.Element | null {
...
 		});
 	};
+
 	const handleNavigateNext = (): void => {
 		dispatch({
 			type: GET_NEXT_LOG_LINES,
 		});
 	};
 
+	const flattenObject = useCallback(
```
palashgdev, line 85: name of the variable should also convey we are stringifying the nested objects

frontend/src/container/LogControls/index.tsx
```diff
@@ -72,12 +75,61 @@ function LogControls(): JSX.Element | null {
...
+	const downloadExcelFile = useCallback((): void => {
+		console.log('Write a code here for excel download');
+	}, []);
+
+	const downloadCsvFile = useCallback((): void => {
+		const flattenLogData = logs.map((entry) => flattenObject(entry));
+		const csv = Papa.unparse(flattenLogData);
+		const csvBlob = new Blob([csv], { type: 'text/csv;charset=utf-8;' });
+		const csvUrl = URL.createObjectURL(csvBlob);
+		const downloadLink = document.createElement('a');
```
palashgdev, line 109: this process can block the thread add loader and see if can use the already created element or remove the node from DOM after the completion

palashgdev (changes requested):

srikanthccv:
This seems good to me from the functionality perspective (tested with filters, pagination and page sizes, etc..) but I would let others to give it a test also.

palashgdev (approved):

PR #2847, opened 2023-06-07
feat: trace view and list view for traces

makeavish:
Closes(https://github.com/SigNoz/signoz/issues/2395 & https://github.com/SigNoz/signoz/issues/2848)


pkg/query-service/constants/constants.go
```diff
@@ -236,6 +236,14 @@ const (
...
 		"CAST((attributes_float64_key, attributes_float64_value), 'Map(String, Float64)') as  attributes_float64," +
 		"CAST((resources_string_key, resources_string_value), 'Map(String, String)') as resources_string "
+	TracesExplorerViewSQLSelectWithSubQuery = "WITH subQuery AS ( " +
+		"SELECT distinct on (traceID) traceID, durationNano, serviceName, name " +
+		"FROM " + SIGNOZ_TRACE_DBNAME + "." + SIGNOZ_SPAN_INDEX_TABLENAME +
+		" WHERE " + "%s " + "%s " + "ORDER BY durationNano DESC "
+	TracesExplorerViewSQLSelectQuery = "SELECT subQuery.serviceName, subQuery.name, count() AS span_count, subQuery.durationNano, " +
+		"traceID FROM " + SIGNOZ_TRACE_DBNAME + "." + SIGNOZ_SPAN_INDEX_TABLENAME + " INNER JOIN subQuery ON " + SIGNOZ_SPAN_INDEX_TABLENAME +
+		".traceID = subQuery.traceID GROUP BY traceID, subQuery.durationNano, subQuery.name, subQuery.serviceName " +
+		"ORDER BY subQuery.durationNano desc;"
```
srikanthccv, line 246: Format with backticks, it will be clean.

pkg/query-service/app/traces/v3/query_builder.go
```diff
@@ -404,17 +425,19 @@ func addOffsetToQuery(query string, offset uint64) string {
...
-	query, err := buildTracesQuery(start, end, mq.StepInterval, mq, constants.SIGNOZ_SPAN_INDEX_TABLENAME, keys)
+	query, err := buildTracesQuery(start, end, mq.StepInterval, mq, constants.SIGNOZ_SPAN_INDEX_TABLENAME, keys, panelType)
 	if err != nil {
 		return "", err
 	}
 	if panelType == v3.PanelTypeValue {
 		query, err = reduceToQuery(query, mq.ReduceTo, mq.AggregateOperator)
 	}
-	query = addLimitToQuery(query, mq.Limit, panelType)
+	if panelType != v3.PanelTypeTrace {
```
srikanthccv, line 435: The condition here is `if panelType != v3.PanelTypeTrace ` but there is a change in `addLimitToQuery` with  `if ... || panelType == v3.PanelTypeTrace` which is the opposite. I understand what is happening here, but it will be hard for future readers to make sense of what is happening. I would consider refactoring.

pkg/query-service/app/traces/v3/query_builder.go
```diff
@@ -297,12 +303,27 @@ func buildTracesQuery(start, end, step int64, mq *v3.BuilderQuery, tableName str
...
+			if mq.Offset != 0 {
+				withSubQuery = addOffsetToQuery(withSubQuery, mq.Offset)
+			}
+			query = withSubQuery + ") " + constants.TracesExplorerViewSQLSelectQuery
+		} else if panelType == v3.PanelTypeList {
+			if len(mq.SelectColumns) == 0 {
+				return "", fmt.Errorf("select columns cannot be empty for panelType %s", panelType)
+			}
+			selectColumns := getSelectColumns(mq.SelectColumns, keys)
+			queryNoOpTmpl := fmt.Sprintf("SELECT timestamp as timestamp_datetime, spanID, traceID, "+"%s ", selectColumns) + "from " + constants.SIGNOZ_TRACE_DBNAME + "." + constants.SIGNOZ_SPAN_INDEX_TABLENAME + " where %s %s"
```
srikanthccv, line 319: Why alias `timestamp` as `timestamp_datetime` and add an additional branch here https://github.com/SigNoz/signoz/pull/2847/files#diff-a8c15149b48aee5261176aa62fa630bb7bde1c4362583ff95871ba24cec75bf9R4226

pkg/query-service/app/traces/v3/query_builder.go
```diff
@@ -297,12 +303,27 @@ func buildTracesQuery(start, end, step int64, mq *v3.BuilderQuery, tableName str
...
+			query = fmt.Sprintf(queryNoOpTmpl, spanIndexTableTimeFilter, filterSubQuery)
```
srikanthccv, line 320: Where is the order by?

pkg/query-service/app/traces/v3/query_builder.go
```diff
@@ -297,12 +303,27 @@ func buildTracesQuery(start, end, step int64, mq *v3.BuilderQuery, tableName str
 		query := fmt.Sprintf(queryTmpl, step, op, filterSubQuery, groupBy, having, orderBy)
 		return query, nil
 	case v3.AggregateOperatorNoOp:
-		// queryTmpl := constants.TracesSQLSelect + "from " + constants.SIGNOZ_TRACE_DBNAME + "." + constants.SIGNOZ_SPAN_INDEX_TABLENAME + " where %s %s"
-		// query := fmt.Sprintf(queryTmpl, spanIndexTableTimeFilter, filterSubQuery)
-		// return query, nil
-		return "", fmt.Errorf("not implemented, part of traces page")
+		var query string
+		if panelType == v3.PanelTypeTrace {
+			withSubQuery := fmt.Sprintf(constants.TracesExplorerViewSQLSelectWithSubQuery, spanIndexTableTimeFilter, filterSubQuery)
```
srikanthccv, line 308: I don't see anywhere `parentSpanID = ''`, which should always be added in the backend regardless of the user filters. We only should consider root span duration for the trace view.

pkg/query-service/app/traces/v3/query_builder.go
makeavish: This is because timestamp is in unix int in logs and in traces it's in time.Time

pkg/query-service/app/traces/v3/query_builder.go
makeavish: Added it

pkg/query-service/constants/constants.go
makeavish: I made it cleaner by using formatting. Using backticks makes tests fail maybe due to multiline char.

pkg/query-service/app/traces/v3/query_builder.go
makeavish: Added

pkg/query-service/app/traces/v3/query_builder.go
```diff
@@ -322,41 +362,75 @@ func groupByAttributeKeyTags(keys map[string]v3.AttributeKey, tags ...v3.Attribu
...
-	if len(str) > 0 {
-		str = str + ","
+	orderByArray := orderBy(panelType, items, groupTags, keys)
+
+	found := false
+	for i := 0; i < len(orderByArray); i++ {
+		if strings.Compare(orderByArray[i], constants.TIMESTAMP) == 0 {
+			orderByArray[i] = "ts"
+			break
+		}
```
srikanthccv, line 423: What is happening here? 
- The result `orderByArray` will have order asc/desc, so it will never match the `constants.TIMESTAMP`?
- Why does it replace with simple `ts`? Is this the correct alias for all cases?

pkg/query-service/app/traces/v3/query_builder.go
```diff
@@ -322,41 +362,75 @@ func groupByAttributeKeyTags(keys map[string]v3.AttributeKey, tags ...v3.Attribu
...
 	}
+	if !found {
+		if aggregatorOperator == v3.AggregateOperatorNoOp {
+			orderByArray = append(orderByArray, constants.TIMESTAMP)
+		} else {
+			orderByArray = append(orderByArray, "ts")
+		}
```
srikanthccv, line 430: - the `found` is never set?
- default order is not specified?

srikanthccv (commented):
couple of questions about order by

pkg/query-service/app/traces/v3/query_builder.go
makeavish: Handled all issues related to timestamp orderBy

pkg/query-service/app/traces/v3/query_builder.go
```diff
@@ -297,13 +304,46 @@ func buildTracesQuery(start, end, step int64, mq *v3.BuilderQuery, tableName str
...
-		return "", fmt.Errorf("unsupported aggregate operator")
+		return "", fmt.Errorf("unsupported aggregate operator %s", mq.AggregateOperator)
+	}
+}
+
+func enrichOrderBy(items []v3.OrderBy, keys map[string]v3.AttributeKey) []v3.OrderBy {
+	enrichedItems := []v3.OrderBy{}
+	for i := 0; i < len(items); i++ {
+		attributeKey := enrichKeyWithMetadata(v3.AttributeKey{
+			Key: items[i].Key,
```
srikanthccv, line 335: I just noticed this. You need to use `ColumnName` here; otherwise, it doesn't work. See the struct definition of `OrderBy`. The `Key` has a `-` tag because it's a placeholder, and the client doesn't send data now. Try writing a test for list type with only `ColumnName` and `Order` (the payload clients send as of now) and check it yourself.

pkg/query-service/app/traces/v3/query_builder_test.go
```diff
@@ -835,32 +922,59 @@ var testBuildTracesQueryData = []struct {
...
+			},
+			QueryName:         "A",
+			AggregateOperator: v3.AggregateOperatorNoOp,
+			Expression:        "A",
+			Filters:           &v3.FilterSet{Operator: "AND", Items: []v3.FilterItem{}},
+		},
+		ExpectedQuery: "SELECT timestamp as timestamp_datetime, spanID, traceID," +
+			" name as `name`  from signoz_traces.distributed_signoz_index_v2 where " +
+			"(timestamp >= '1680066360726210000' AND timestamp <= '1680066458000000000') ",
+		PanelType: v3.PanelTypeList,
```
srikanthccv, line 944: Add a couple more tests.

pkg/query-service/app/traces/v3/query_builder_test.go
```diff
@@ -334,15 +376,38 @@ var testOrderBy = []struct {
...
+			{
+				ColumnName: "response_time",
+				Order:      "desc",
+				Key:        "response_time",
+				Type:       v3.AttributeKeyTypeTag,
+				DataType:   v3.AttributeKeyDataTypeString,
+			},
+		},
+		Tags:   []string{"name", "bytes"},
+		Result: []string{"name asc", "bytes asc", "stringTagMap['response_time'] desc"},
```
srikanthccv, line 402: Add some tests that don't have `Tags` but only `Items`.

srikanthccv (commented):
One last comment. Otherwise mostly looks good to me.

srikanthccv (approved):

Issue #2032, opened 2023-01-12
remove loadash from fe dependency


palashgdev:
## Is your feature request related to a problem?

*Please describe.*

## Describe the solution you'd like
use native js features for loadash method

Vipul2504:
Hello @palashgdev  can I contribute to this?

palashgdev:
Sure please go ahead @Vipul2504

Vipul2504:
hey can you guide me on how to setup in my local

fakoredeDamilola:
Hello, is this still open. Can I work on this

sakshigangwani:
Hi! I see PR #11695 partially addresses this by replacing noop, isNull, isNumber, and isFunction. I'd like to continue this work and replace isEmpty usages with native equivalents across the frontend. Would a scoped PR for just isEmpty be welcome?

GermaVinsmoke:
Just a heads up, replying to this after a long time, but some times it's better to use library functions rather than "reinventing the wheel all the time". Those are tested with many edge cases for a long time, rewriting it on our own and introducing bugs just for saving few kilobytes? Don't think it's a good trade off, rest is upto the repo maintainers.

TheJhyeFactor:
Hey, I had a look at this and the current open PR. PR #11695 already covers `noop`, `isNull`, `isNumber`, and `isFunction`, so I don’t want to duplicate that work.

I’d like to take a separate small batch of the remaining Lodash usages where the native replacement is behaviour-equivalent. I’ll keep the change focused and add tests for any edge cases instead of trying to remove the whole dependency in one PR.

Is that scoped follow-up useful, and is this issue still available for another contributor?

PR #2886, opened 2023-06-10
fix: create and store label when colon is pressed

### LINKED ISSUE #2692: Alert labels are not stored
## Bug description

When creating a new alert, labels are not stored.

For example, creating an alert with label abc: xyz

image

When alert is saved, there are no labels for the alert:

image

## Expected behavior

Labels should be saved.

## Version information 
* **Signoz version**: v0.18.3
* **Browser version**: Chrome 113.0.5672.92 
* **Your OS and version**: MacOS Ventura 13.1

Damans227:
Fixes #2692

When creating an Alert, the key value pair (label) field allows the user to include a colon in either key or value. It can cause confusion and lead users to believe the key and value must be included as one string in order to account for a label. This is exactly what caused #2692.

I would recommend that in addition to using `ENTER` as a way to end a key or value string, we also allow `COLON (:)` to do the same. In this way, the user will have a better experience. [This screen recording](https://www.loom.com/share/4eaa4272413d44f1813ad1c98e3be44e) shows how it works.

### DIFF
--- a/frontend/src/container/FormAlertRules/labels/index.tsx
+++ b/frontend/src/container/FormAlertRules/labels/index.tsx
@@ -5,7 +5,7 @@ import {
 import { useMachine } from '@xstate/react';
 import { Button, Input, message, Modal } from 'antd';
 import { useIsDarkMode } from 'hooks/useDarkMode';
-import { map, replace } from 'lodash-es';
+import { map } from 'lodash-es';
 import { ChangeEvent, useCallback, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Labels } from 'types/api/alerts/def';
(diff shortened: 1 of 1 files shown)


frontend/src/container/FormAlertRules/labels/index.tsx
```diff
@@ -84,8 +84,9 @@ function LabelSelect({
 		handleBlur();
 	}, [handleBlur]);
 
-	const handleChange = (e: ChangeEvent<HTMLInputElement>): void => {
-		setCurrentVal(e.target?.value);
+	const handleLabelChange = (e: ChangeEvent<HTMLInputElement>): void => {
+		const processedCurrentVal = replace(e.target?.value, ':', '');
```
palashgdev, line 88: can we use native replace

palashgdev (changes requested):

palashgdev:
Hi @Damans227 
Thanks for raising the PR can you please the small suggestion though we in process to remove [lodash](https://github.com/SigNoz/signoz/issues/2032)

frontend/src/container/FormAlertRules/labels/index.tsx
Damans227: Sure. That sounds good.

palashgdev (approved):

PR #2911, opened 2023-06-15
chore: use the local table for inner subquery

srikanthccv:
Part of https://github.com/SigNoz/engineering-pod/issues/903

https://clickhouse.com/docs/en/engines/table-engines/special/distributed#distributed-writing-data
>Queries are used that require joining data (IN or JOIN) by a specific key. If data is sharded by this key, you can use local IN or JOIN instead of GLOBAL IN or GLOBAL JOIN, which is much more efficient.


srikanthccv:
@ankitnayan This is the query service part for this change https://github.com/SigNoz/signoz-otel-collector/pull/142.

makeavish (approved):

Issue #2333, opened 2023-02-22
List view tab in the traces explorer


srikanthccv:
<img width="847" alt="Screenshot 2023-02-22 at 8 37 05 PM" src="https://user-images.githubusercontent.com/22846633/220663072-0bf8e335-6d26-434f-b38d-98cc07a54aae.png">

The API will return raw data in the response in a paginated manner for the list view. The data will contain a timestamp field and the number of key-value pairs indicating the column and its value. The page will show X number of rows initially, and the user can move to the next/prev pages. There is no count API involved for the total pages. The client will send the offset and the pageSize to query service. Where no data is returned, the next/prev buttons will be disabled.

PR #2947, opened 2023-06-21
feat: add the list view for the traces explorer 

dnazarenkoo:
Close - [2333](https://github.com/SigNoz/signoz/issues/2333)


makeavish:
In list view PR, when we switch from list view to time series view then two query range API calls are made.
Can you please check that?
@dnazarenkosignoz

dnazarenkoo:
> In list view PR, when we switch from list view to time series view then two query range API calls are made. Can you please check that? @dnazarenkosignoz

@makeavish fixed

makeavish:
In list view there should also be option for 10/page, and set it as default
Screenshot 2023-06-27 at 20 35 39

makeavish:
The list view row should be clickable and link to trace details page.
@dnazarenkoo 
cc: @palashgdev

makeavish:
@dnazarenkoo When there's an error from list view backend API then the frontend is stuck on loading.

makeavish:
@dnazarenkoo Boolean columns are not handled.
Example: hasError when false should be showing false instead of N/A, similarly should show true when true instead of empty value which is being shown currently.
Screenshot 2023-07-03 at 17 05 10

makeavish:
When number is 0 (zero) then show zero instead of N/A
@dnazarenkoo

makeavish:
When a user selects spanID then show `spanID`, similarly for `parentSpanID`, `traceID`.
Screenshot 2023-07-03 at 17 11 16

makeavish:
@dnazarenkoo Found another issue. 
https://share.vidyard.com/watch/rHKWAW2MNCzBVtxeF8BQfd?

For this you can create a separate PR and merge this PR after conflicts are resolved.

makeavish:
Date column is shown wrong and we can't expand the view as well

Screenshot 2023-07-04 at 12 24 37

makeavish:
Columns which are already added to column shouldn't be shown in add a column list.

Screenshot 2023-07-04 at 12 27 38

makeavish:
Aggregate Attribute should be disabled in list view. 
Currently user is able to write text in the field and API is also called.

Screenshot 2023-07-04 at 12 38 01

makeavish:
Hide legend in list and traces view.

Screenshot 2023-07-04 at 12 39 34

makeavish:
Hide aggregate every in list view.

Screenshot 2023-07-04 at 12 40 35

makeavish (dismissed):
LGTM, apart from the issues mentioned.
Pending issues can be solved in separate PRs.

makeavish:
Sort by `timestamp desc` by default, similar to logs list view.

makeavish (approved):

palashgdev (approved):

Issue #2744, opened 2023-05-22
Sort tooltip on values descending order


daniel-t4e:
## Is your feature request related to a problem?

I wish the names of traces would be sorted in the tooltip on graphs on the dashboard.
![image](https://github.com/SigNoz/signoz/assets/130993189/61f4ea01-0da9-4bfe-99a9-71ad4c4f3ba7)
As you can see it currently is not 

## Describe the solution you'd like

A way where the values would be sorted in descending order on the tooltip.

PR #2948, opened 2023-06-21
feat: sorting for tooltip in graph view

Rajat-Dabade:
Close https://github.com/SigNoz/signoz/issues/2744


frontend/src/components/Graph/index.tsx
```diff
@@ -181,6 +181,9 @@ function Graph({
 							},
 						},
 						position: 'custom',
+						itemSort(a, b) {
```
palashgdev, line 184: can we have a better name here a,b does not highlight much

Rajat-Dabade: Sure

makeavish (approved):

palashgdev (approved):

Issue #2797, opened 2023-05-30
Don't call query range APIs when user updates layout on dashboard 


makeavish:
Currently whenever user updates layout on dashboard, then query range APIs are called for all dashboard panels. This is unnecessary.

makeavish:
This is fixed in latest version

PR #2958, opened 2023-06-22
fix: no query_range api call on layout change

Rajat-Dabade:
Close https://github.com/SigNoz/signoz/issues/2797


frontend/src/container/GridGraphLayout/index.tsx
```diff
@@ -217,28 +217,28 @@ function GridGraph(props: Props): JSX.Element {
...
-								yAxisUnit={currentWidget?.yAxisUnit}
-								layout={layout}
-								setLayout={setLayout}
-								onDragSelect={onDragSelect}
-							/>
-						),
-					};
-				}),
-			);
+			for (let i = 0; i < layout.length; i += 1) {
```
palashgdev, line 220:
```suggestion
const setLayoutFunction = useCallback(
	(layout: Layout[]) => {
		layout.forEach((item, i) => {
			const currentWidget =
				widgets?.find((widget) => widget.id === item.i) || ({} as Widgets);
			layouts[i] = {
				...item,
				Component: (): JSX.Element => (
					<Graph
						name={currentWidget.id}
						widget={currentWidget}
						yAxisUnit={currentWidget?.yAxisUnit}
						layout={layout}
						setLayout={setLayout}
						onDragSelect={onDragSelect}
					/>
				),
			};
		});

		setLayout((layouts) => layouts);
	},
	[widgets, onDragSelect, layouts],
);
```

palashgdev (changes requested):

makeavish:
@Rajat-Dabade Is it required to call query range API, when we save the layout?
IMO it's not required as layout change is not related to query range API.
cc: @palashgdev

Rajat-Dabade:
@makeavish, It should not be required, should we do it as a part of this PR, or should create a new issue for that?
Thoughts @palashgdev

makeavish:
> @makeavish, It should not be required, should we do it as a part of this PR, or should create a new issue for that? Thoughts @palashgdev

This is happening after these PR changes and is related to current changes only. On staging we don't make query range API on saving layout. So I think it should be part of this PR.

Rajat-Dabade:
@makeavish, Okay I got it. Looking into it.

makeavish:
I think this PR is obsolete and can be closed, reopen if needed.

PR #2960, opened 2023-06-22
fix(FE): variables with multiple values are merged into one string

### LINKED ISSUE #2526: Variables with multiple values are merged into one string
## Bug description

Variables with multiple values are injected incorrectly into the clickhouse query builder. The value that is injected isn't the correct array with the selected values. It's like this: `['A,B,C','D']` instead of `['A','B','C','D']`. The values A, B and C are merged into one string somehow.

It looks like this:
image

## Expected behavior

image
(You can see the workaround I'm using in the query field)

## How to reproduce

1. Create a variable in a dashboard of type custom and call it example
2. Add some values separated by comma (for example A, B, C, D)
3. Enable multiple values to be checked
4. Save
5. Select more than 2 of them (for example A, B, C)
6. Add a panel of type value and add this clickhouse query: `select "{{.example}}"` (it's incorrect, but it shows the value that is injected)

image

The image shows that multiple values are injected incorrectly.

## Version information 
* **Signoz version**: v0.17.0
* **Browser version**: Firefox 111.0.1
* **Your OS and version**: Ubuntu 22.04.2 LTS

## Additional context

I'm working with the dashboards in signoz and want to add a panel that shows our warnings grouped by the service. Also I want to select which services should be included. For example to only show 3 of them (to compare them). This should work with variables with multiple values. What I expect is that I can select the services in the drop down list of the variable and then the selected services are included into the array of the variable.

Bergiu:
Fixes #2526


palashgdev:
@Bergiu can you please check the tsc error though it is breaking

frontend/src/container/NewDashboard/DashboardVariablesSelection/VariableItem.tsx
```diff
@@ -141,9 +141,26 @@ function VariableItem({
 			}
 	};
 
-	const selectValue = variableData.allSelected
-		? 'ALL'
-		: variableData.selectedValue?.toString() || '';
+	const getSelectValue = (): string | number | (number | string)[] => {
```
palashgdev, line 144:
```suggestion
	const getSelectValue = (): IDashboardVariable['selectedValue'] => {
```

palashgdev (changes requested):

frontend/src/container/NewDashboard/DashboardVariablesSelection/VariableItem.tsx
Bergiu: `IDashboardVariable['selectedValue']` contains `boolean` (and `boolean[]`), which is not allowed for `<Select value=`, as you can see here: https://ant.design/components/select#api.

That's the reason why the method converts booleans to strings.

This would result in a type mismatch.

Bergiu:
Is the Select component from [ant](https://ant.design/components/select#api) or from [rc-select](https://www.npmjs.com/package/rc-select)?

It's imported from 'antd' but the error message looks like it's using the Select component from rc-select.

frontend/src/container/NewDashboard/DashboardVariablesSelection/VariableItem.tsx
Bergiu: This would result in a type mismatch, because `IDashboardVariable['selectedValue']` contains `boolean` (and depending on which package is imported also `number`) which is not allowed in the select component for the value attribute.

Bergiu:
Well, for the two failing checks, I don't know what I need to do.

srikanthccv:
You don't need to worry about them.

Bergiu:
Is there any problem left or is the issue ready to be merged?

srikanthccv:
Sorry we haven't had a chance to review/test this. Will get back to you on this soon.

srikanthccv:
I tested this. This seems to have fixed the original issue. I will be testing it for other types, such as queries.

srikanthccv (dismissed):
Thank you

Bergiu:
Thanks for approving, but the merge is still blocked because of the update to the main branch. Could you merge it directly after the merge+approve?

frontend/src/container/NewDashboard/DashboardVariablesSelection/VariableItem.tsx
```diff
@@ -141,9 +141,15 @@ function VariableItem({
...
-		? 'ALL'
-		: variableData.selectedValue?.toString() || '';
+	const getSelectValue = (): string | string[] => {
+		if (Array.isArray(variableData.selectedValue)) {
+			return variableData.selectedValue.map((item) => item.toString());
+		}
+		return variableData.selectedValue?.toString() || '';
+	};
+
+	const selectValue = variableData.allSelected ? 'ALL' : getSelectValue();
```
palashgdev, line 151: only the concern is that when the component is re-rendered function context would be re-created so can we have a memo and a single utility to get the selectedValue

palashgdev:
@Bergiu small changes after LGTM :-D to merge

frontend/src/container/NewDashboard/DashboardVariablesSelection/VariableItem.tsx
Bergiu: okay i've moved the function above and added a memo

srikanthccv (approved):

srikanthccv:
@palashgdev please take a look

palashgdev (approved):

palashgdev:
thanks @Bergiu for the PR :-D

Issue #2535, opened 2023-04-03
Noticing a slight flickering of charts


ankit01-oss:
Noticing a slight flickering of charts while revisiting a tab.

https://www.loom.com/share/12c28e1d5085448ab9cee77b8727eb3b

PR #2965, opened 2023-06-22
fix: the flickering issue on service onclick

Rajat-Dabade:
Close https://github.com/SigNoz/signoz/issues/2535


palashgdev:
not needed as this reducer would be removed

PR #2970, opened 2023-06-23
feat: add support for multiquery in ts with limit

### LINKED ISSUE #2845: Add support for limit on Time series
Currently limit on time series is not supported. 
Limit on time series limits on number of series after ordering by Aggregation(key).

nityanandagohain:
Fixes #2845 

* two queries are created `pre` and `main`
* during execution if main is found then pre is executed and the result is added to main query and then main query is executed.

test curl

```
curl --location 'localhost:8080/api/v3/query_range' \
--data '{
    "start": 1680712088000000000,
    "end": 1690771611000000000,
    "step": 60,
    "dataSource": "logs",
    "compositeQuery": {
        "queryType": "builder",
        "panelType": "graph",
        "limit": 1,
        "builderQueries": {
            "A": {
                "stepInterval": 60,
                "queryName": "A",
                "dataSource": "logs",
                "aggregateOperator": "count",
                "filters": {},
                "limit": 3,
                "groupBy": [
                    {
                        "key": "method",
                        "isColumn": true
                    }
                ],
                "expression": "A",
                "disabled": false
            }
        }
    }
}'
```


nityanandagohain:
Have added support for multiple group by

pkg/query-service/app/http_handler.go
```diff
@@ -2464,13 +2464,58 @@ func (aH *APIHandler) execClickHouseGraphQueries(ctx context.Context, queries ma
...
+					put | error
+					put | debug
+				*/
+				// filter : ( method = 'put' AND level = 'error' ) OR ( method = 'put' AND level = 'debug' )
+				filterString := ""
+				l := len(list)
+				for i, row := range list {
+					tFilter := ""
+					for name, val := range row.Data {
+						tFilter = tFilter + fmt.Sprintf("%s = '%s' AND ", name, *val.(*string))
```
srikanthccv, line 2493: > `*val.(*string) `

The value can't always be a string? Please use `ClickHouseFormattedValue`. I would `AND` and then trim right below. Instead, I would prepare a list of conditionals and use `strings.Join` which is cleaner. And finally, join those intermediary independent condition statements with `OR`. I find it easier to follow as a reader.

nityanandagohain: Thanks have updated it.

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -371,8 +402,28 @@ func addOffsetToQuery(query string, offset uint64) string {
...
+		query = addLimitToQuery(query, mq.Limit)
+
+		return query, nil
+	} else if graphLimitQtype == constants.SecondQueryGraphLimit {
+		query, err := buildLogsQuery(panelType, start, end, mq.StepInterval, mq, graphLimitQtype)
+		if err != nil {
+			return "", err
+		}
+		return query, nil
+	}
```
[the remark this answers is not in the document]
nityanandagohain: Okay, but would you suggest doing it in this PR ? I beleive this will add a good amount of changes.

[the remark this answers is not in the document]
makeavish: I agree on updating query builder for panel type. But that will be a complete rewrite of query builder and that can be taken up separately.

pkg/query-service/app/http_handler.go
```diff
@@ -2464,13 +2465,55 @@ func (aH *APIHandler) execClickHouseGraphQueries(ctx context.Context, queries ma
...
+				}
+
+				// create the filterset from the first query result
+				/*
+					put | error
+					put | debug
+				*/
+				// filter : ( method = 'put' AND level = 'error' ) OR ( method = 'put' AND level = 'debug' )
+				filterarr := []string{}
+				for _, row := range list {
```
makeavish, line 2490: Add a function and test case of the same.

pkg/query-service/app/logs/v3/query_builder.go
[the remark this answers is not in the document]
srikanthccv: I won't be able to provide any good feedback on this as is since I can't wrap my head around it. Speaking from experience, the _taken up separately_ doesn't get picked up separately as people move on to different things. I didn't mean to say it in this PR. I would first create refactor PR and then rebase this when that gets merged.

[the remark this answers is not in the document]
srikanthccv: I also don't mind approving this, but I believe that's pointless and can't give any meaningful reviews for future changes if I don't understand things.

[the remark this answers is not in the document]
makeavish: These changes have to be taken up separately **seriously** as in the future it's going to be difficult making changes and introducing new panels without separation of panels in codebase.

Yes, PR can't be approved if we don't understand things. Maybe we can get on a call with @nityanandagohain for this PR to understand this PR and clear doubts.

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -89,17 +89,29 @@ func getClickhouseColumnName(key v3.AttributeKey) string {
...
+func getSelectKeys(aggregatorOperator v3.AggregateOperator, groupBy []v3.AttributeKey) string {
+	var selectLabels string
+	if aggregatorOperator == v3.AggregateOperatorNoOp {
+		selectLabels = ""
+	} else {
+		for _, tag := range groupBy {
+			selectLabels += fmt.Sprintf("%s,", tag.Key)
+		}
+	}
+	return selectLabels
```
srikanthccv, line 114: I see `TrimRight` below, I would suggest using `strings.Join` here again to avoid trim by the caller.

pkg/query-service/app/queryBuilder/query_builder.go
```diff
@@ -152,11 +153,28 @@ func (qb *QueryBuilder) PrepareQueries(params *v3.QueryRangeParamsV3, args ...in
 					}
 					queries[queryName] = queryString
 				case v3.DataSourceLogs:
-					queryString, err := qb.options.BuildLogQuery(params.Start, params.End, compositeQuery.QueryType, compositeQuery.PanelType, query)
-					if err != nil {
-						return nil, err
+					// if panel type is timeseries/ table, handle limit differently using two queries
+					if (compositeQuery.PanelType == v3.PanelTypeGraph || compositeQuery.PanelType == v3.PanelTypeTable) && query.Limit > 0 && len(query.GroupBy) > 0 {
```
srikanthccv, line 157: I don't think we need to handle the table view separately like we do for the time series since there is going to be only one for each unique label set.

pkg/query-service/app/queryBuilder/query_builder.go
```diff
@@ -152,11 +153,28 @@ func (qb *QueryBuilder) PrepareQueries(params *v3.QueryRangeParamsV3, args ...in
...
+						if err != nil {
+							return nil, err
+						}
+
+						placeholderQuery, err := qb.options.BuildLogQuery(params.Start, params.End, compositeQuery.QueryType, compositeQuery.PanelType, query, constants.SecondQueryGraphLimit)
+						if err != nil {
+							return nil, err
+						}
+						queries[queryName+constants.FirstQuerySuffix] = limitQuery
+						queries[queryName+constants.SecondQuerySuffix] = placeholderQuery
```
srikanthccv, line 169: Since you are using the suffixes, will it work for the formula such as `A/2` or similar? The formula will look for `A` whereas the `queries` map contains `A-#first` and `A-#second`?

pkg/query-service/app/queryBuilder/query_builder.go
```diff
@@ -185,6 +203,9 @@ func (qb *QueryBuilder) PrepareQueries(params *v3.QueryRangeParamsV3, args ...in
 
 	// filter out disabled queries
 	for queryName := range queries {
+		if strings.HasSuffix(queryName, constants.FirstQuerySuffix) || strings.HasSuffix(queryName, constants.SecondQuerySuffix) {
+			continue
```
srikanthccv, line 207: A query with a limit can be disabled too.

pkg/query-service/app/queryBuilder/query_builder.go
nityanandagohain: Ahh with the new changes we have removed timestamp from the result so we don't need it in table view. Is that what you are trying to say ?

cc @makeavish

pkg/query-service/app/queryBuilder/query_builder.go
nityanandagohain: Ahh it won't work as of now, thinking of ways we can solve this.

Any suggestions on ways to implement this as we need the first query to be complete so that that the second query can be formed which can be used in the expression evaluation.

pkg/query-service/app/queryBuilder/query_builder.go
srikanthccv: Yes

pkg/query-service/app/queryBuilder/query_builder.go
srikanthccv: Can you run all of the pre queries first and save the result and then prepare the final queries to be executed?

srikanthccv:
https://clickhouse.com/docs/en/sql-reference/operators/in

> The left side of the operator is either a single column or a tuple.

My suggestion for the growing complexity of this is to use a subquery to get the tuple of columns you want to filter on and then use that in the outer query. You will get the same benefit as the two-query approach but without the complexity. You don't have to worry about the subquery affecting performance because it will only be executed once. And don't need to worry about the distributed setup (GLOBAL IN) either because IN doesn't involve a separate table; instead, it is a tuple of columns. If anything, it should be better than the two-query because we eliminate the query-service from the equation.

The one query equivalent of the two-query approach is:

```sql
SELECT
    toStartOfInterval(fromUnixTimestamp64Nano(timestamp), toIntervalSecond(60)) AS ts,
    method AS method,
    severity_number AS severity_number,
    toFloat64(count(*)) AS value
FROM signoz_logs.distributed_logs
WHERE timestamp >= 1680712088000000000 AND timestamp <= 1690771611000000000 AND (method, severity_number) IN (
    SELECT
        method,
        severity_number
    FROM
    (
        SELECT
            method AS method,
            severity_number AS severity_number,
            toFloat64(count(*)) AS value
        FROM signoz_logs.distributed_logs
        WHERE (timestamp >= 1680712088000000000) AND (timestamp <= 1690771611000000000)
        GROUP BY
            method,
            severity_number
        ORDER BY value ASC
    )
    LIMIT 1
)
GROUP BY
    method,
    severity_number,
    ts
ORDER BY
    method ASC,
    severity_number ASC,
    ts ASC
```

srikanthccv:
You will still use the part of work you worked on already; instead of running queries and the preparing addition part of where you can replace it with a subquery.

pkg/query-service/utils/format.go
```diff
@@ -143,15 +143,19 @@ func ValidateAndCastValue(v interface{}, dataType v3.AttributeKeyDataType) (inte
 
 // ClickHouseFormattedValue formats the value to be used in clickhouse query
 func ClickHouseFormattedValue(v interface{}) string {
+	// if it's pointer convert it to a value
+	v = getPointerValue(v)
```
nityanandagohain, line 147: This functionality is not being used as of now with the new changes ? Not sure if we are going to use it in the future.

Should we keep it or remove ?  @srikanthccv  @makeavish

srikanthccv: It doesn't hurt to have it

pkg/query-service/app/logs/v3/query_builder_test.go
```diff
@@ -936,3 +938,78 @@ func TestOrderBy(t *testing.T) {
...
+			Expression:         "A",
+			Filters: &v3.FilterSet{Operator: "AND", Items: []v3.FilterItem{
+				{Key: v3.AttributeKey{Key: "method", DataType: v3.AttributeKeyDataTypeString, Type: v3.AttributeKeyTypeTag}, Value: "GET", Operator: "="},
+			},
+			},
+			Limit:   10,
+			GroupBy: []v3.AttributeKey{{Key: "method", DataType: v3.AttributeKeyDataTypeString, Type: v3.AttributeKeyTypeTag}},
+		},
+		TableName:     "logs",
+		ExpectedQuery: "SELECT method from (SELECT attributes_string_value[indexOf(attributes_string_key, 'method')] as method, toFloat64(count(distinct(attributes_string_value[indexOf(attributes_string_key, 'name')]))) as value from signoz_logs.distributed_logs where (timestamp >= 1680066360726210000 AND timestamp <= 1680066458000000000) AND attributes_string_value[indexOf(attributes_string_key, 'method')] = 'GET' AND indexOf(attributes_string_key, 'method') > 0 group by method order by method ASC) LIMIT 10",
```
srikanthccv, line 979: Shouldn't this be `value asc/desc`? We want to fetch the x series based on the value, the order of other tags is relevant in the final query not the limit query?

nityanandagohain: Okay so do you mean to remove the default order by that we generate from `group by` and keep default as `value`.

If user enters a custom order by then `value` will be replaced by that key ?

nityanandagohain: Have added the changes, please review

nityanandagohain:
Have added new changes where if there is no orderBy provided
* For table and TS -> add `value DESC` as orderBy
* For listview -> add `timestamp DESC` as orderBy

If an order by key is provided.
* For table, ts and listview -> use key provided for ordering.

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -273,82 +291,56 @@ func buildLogsQuery(panelType v3.PanelType, start, end, step int64, mq *v3.Build
 
 // groupBy returns a string of comma separated tags for group by clause
 // `ts` is always added to the group by clause
-func groupBy(panelType v3.PanelType, tags ...string) string {
-	if panelType == v3.PanelTypeGraph || panelType == v3.PanelTypeValue {
+func groupBy(panelType v3.PanelType, graphLimitQtype string, tags ...string) string {
```
srikanthccv, line 294: If it helps, you can have separate functions just for this first query kind.

srikanthccv (approved):

pkg/query-service/app/logs/v3/query_builder.go
nityanandagohain: Yeah, anyway I have to clean it up in the refractoring processs. Keeping it as it is for now.

srikanthccv: Sure

PR #2978, opened 2023-06-27
feat: ability to configure noisy top level operations to discard

srikanthccv:
Adds a config option `--skip-top-level-ops` to skip certain top-level operations for each service.

The sample config looks like the following.

```
services:
  - name: redis
    operations:
      - GetDriver
```

We need service wise config because what might be incorrect top-level operation options could be a correct top-level operation in another service. These are discarded from the call `GetTopLevelOperations` so that it reflects everywhere.


pkg/query-service/model/config.go
```diff
@@ -0,0 +1,57 @@
...
+type SkipConfig struct {
+	Services []ServiceSkipConfig `yaml:"services"`
+}
+
+type ServiceSkipConfig struct {
+	Name       string   `yaml:"name"`
+	Operations []string `yaml:"operations"`
+}
+
+func (s *SkipConfig) ShouldSkip(serviceName, name string) bool {
```
ankitnayan, line 18: does it make more sense to use array vs map in reading from `yaml` config? Otherwise map would have been more efficient. It scares me when I see 2 for loops in a function. The good thing is we are not expecting perf hits as arrays would be very small. From DS perspective, 2 for loops are bad.

ankitnayan (approved):

pkg/query-service/model/config.go
srikanthccv: Yes, it doesn't change anything significant whether we use maps or slices when the number of items is very small.

PR #2992, opened 2023-06-28
Feat/list infinity scroll

yeshev:


makeavish:
@yeshev One UI issue
https://share.vidyard.com/watch/NVgD1EHoX1mf7JmGnGTEJJ?

makeavish:
Timestamp invalid

Screenshot 2023-07-03 at 19 07 05

makeavish:
View details button not working
Screenshot 2023-07-03 at 19 07 05

nityanandagohain:
image

Selecting pageSize should be removed -> frontend should directly send the pagisize value in the API and keep the value as 200  in constants.

nityanandagohain (changes requested):
* Option to select pagesize should be removed -> Frontend should store 100 as the value of pageSize in contants and use it in the API.

* Id filter used for pagination should be hidden from the UI. Now when we paginate by scrolling the `id` filter appears in the filter box

* Extra scrollbar should be removed from the logs view, when we paginate an extra scrollbar appears on the left which should not be there, only the scrollbar from the browser should be present

nityanandagohain (changes requested):
* offset should start from 0 and not 1, when using different orderby

nityanandagohain (changes requested):
* Infinite scrolling has stopped working, new api calls doesn't seems to be triggered on reaching the end of the page

> offset should start from 0 and not 1, when using different orderby other than timestamp

Please add this change as well

yeshev:
> * offset should start from 0 and not 1, when using different orderby

@nityanandagohain I've just fixed it also

yeshev:
> * Infinite scrolling has stopped working, new api calls doesn't seems to be triggered on reaching the end of the page
> 
> > offset should start from 0 and not 1, when using different orderby other than timestamp
> 
> Please add this change as well

@nityanandagohain Also fixed it in last commit

nityanandagohain:
* Set limit to 100000 only for list View
* Timeseries and table keep limit 100
* Use correct operator in timestamp asc and timestamp desc

nityanandagohain (commented):
the changes looks good, please resolve the conflicts and they are good to merge

nityanandagohain (dismissed):

nityanandagohain (approved):

makeavish:
Some feedbacks for this PR:
@yeshev You can create a separate PR for these issues.

1. https://github.com/SigNoz/signoz/issues/3049
2. https://github.com/SigNoz/signoz/issues/3050
3. https://github.com/SigNoz/signoz/issues/3051
4. https://github.com/SigNoz/signoz/issues/3052

makeavish (approved):

PR #3004, opened 2023-06-29
fix: formula for table

yeshev:


makeavish:
@yeshev I see an extra column which is empty. Please fix this.
Screenshot 2023-07-03 at 16 11 47

makeavish:
I don't see formula values, instead I am seeing formula name
Screenshot 2023-07-03 at 22 21 35

makeavish:
In formula column, formula values should be inserted.

Screenshot 2023-07-04 at 16 21 52

makeavish (approved):

palashgdev (approved):

PR #3007, opened 2023-06-30
feat: add log explorer detail

yeshev:


nityanandagohain (changes requested):
* If no data present in the selected time range then loading image should be removed.
* When we change the order by then the `id` filter used for pagination should be removed.

nityanandagohain:
While adding new order by should the old ones appear as well ? @makeavish  @srikanthccv 
image

in this case level and method should not be present right ?

nityanandagohain:
@yeshev can you resolve the conflicts

yeshev:
> @yeshev can you resolve the conflicts
@nityanandagohain i've resolved

nityanandagohain (approved):

PR #3009, opened 2023-06-30
feat: add group by in the logs chart

yeshev:
When at least one of queries has the groupBy then render all queries in the chart bar otherwise only first query


nityanandagohain:
Can you please add a description of what this PR does?

From what I can guess it renders the chart in the explorer page, but when I run this branch no query for redering chart is sent

nityanandagohain (approved):

PR #3010, opened 2023-07-02
fix: alert evaluation params and query

srikanthccv:
1. Update the rate query to not generate intermediary +inf value when the denominator is zero
2. Adjust the start and end time to incorporate data in movement
3. Round the start and end to minute
4. Add log to find the exact query that triggered alert for troubleshooting ;

### DIFF
--- a/pkg/query-service/rules/thresholdRule.go
+++ b/pkg/query-service/rules/thresholdRule.go
@@ -22,7 +22,6 @@ import (
 	querytemplate "go.signoz.io/signoz/pkg/query-service/utils/queryTemplate"
 	"go.signoz.io/signoz/pkg/query-service/utils/times"
 	"go.signoz.io/signoz/pkg/query-service/utils/timestamp"
-	"go.signoz.io/signoz/pkg/query-service/utils/value"
 
 	logsv3 "go.signoz.io/signoz/pkg/query-service/app/logs/v3"
 	metricsv3 "go.signoz.io/signoz/pkg/query-service/app/metrics/v3"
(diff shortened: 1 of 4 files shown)


pkg/query-service/rules/thresholdRule.go
```diff
@@ -355,21 +354,37 @@ func (r *ThresholdRule) CheckCondition(v float64) bool {
 func (r *ThresholdRule) prepareQueryRange(ts time.Time) *v3.QueryRangeParamsV3 {
 	// todo(amol): add 30 seconds to evalWindow for rate calc
 
+	// todo(srikanthccv): make this configurable
+	// 2 minutes is reasonable time to wait for data to be available
+	// 60 seconds (SDK) + 10 seconds (batch) + rest for n/w + serialization + write to disk etc..
+	start := ts.Add(-time.Duration(r.evalWindow)).UnixMilli() - 2*60*1000
```
mindhash, line 360: Not sure about the intent here.  [is this to solve rate calculation problem?]

wouldn't this add 2 min delay in raising alerts?  The alert manager also groups alerts in 5(?) mins window.  So the actual notification would get pushed out a further. 

If this is for a specific use case, we could introduce a command line parameter similar to --groupInterval  or --sendDelay etc. I see a todo there but may be it can be prioritised in case things dont work out, customers can rollback.

srikanthccv: This is not a specific use case or rate but a problem in general. The duration we consider for the alerts evaluation is adjusted so that the data is still in motion is considered. Assume the alert evaluator sets the end time as 3:19:01 pm; in practice, the data for this interval is yet to arrive or be available in ClickHouse. Currently, when we create a query with `timestamp between start and end`; it triggers false alerts because of incomplete data, which is more annoying to users.

srikanthccv: Discussed in Slack

srikanthccv:
@mindhash @ankitnayan leave an approval if there are no concerns.

ankitnayan (approved):

ankitnayan:
> @mindhash @ankitnayan leave an approval if there are no concerns.

@srikanthccv the `build-query-service` build has failed. Have a look

srikanthccv:
Fixed a failing test. PTAL

ankitnayan (approved):

PR #3022, opened 2023-07-04
feat: minor fixes to logs QB

nityanandagohain:
* Was returning wrong variable in enrich funtion
* Timestamp order by should not be added when a custom order by is selected in list panel 
* Previous logic of order by was added under the impression that timestamp can come in order by for aggregation -> but it won’t come since timestamp won’t be there in group by

### DIFF
--- a/pkg/query-service/app/logs/v3/query_builder.go
+++ b/pkg/query-service/app/logs/v3/query_builder.go
@@ -321,22 +321,13 @@ func orderByAttributeKeyTags(panelType v3.PanelType, aggregatorOperator v3.Aggre
 	}
 	orderByArray := orderBy(panelType, items, groupTags)
 
-	found := false
-	for i := 0; i < len(orderByArray); i++ {
-		if strings.Compare(orderByArray[i], constants.TIMESTAMP) == 0 {
-			orderByArray[i] = "ts"
-			break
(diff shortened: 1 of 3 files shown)


nityanandagohain:
Please don't review right now

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -321,19 +321,13 @@ func orderByAttributeKeyTags(panelType v3.PanelType, aggregatorOperator v3.Aggre
...
-	for i := 0; i < len(orderByArray); i++ {
-		if strings.Compare(orderByArray[i], constants.TIMESTAMP) == 0 {
-			orderByArray[i] = "ts"
-			break
-		}
-	}
-	if !found {
-		if aggregatorOperator == v3.AggregateOperatorNoOp {
+	if panelType == v3.PanelTypeList {
+		if len(orderByArray) == 0 {
```
makeavish, line 325: I think you can update this in single line:
`	if panelType == v3.PanelTypeList && len(orderByArray) == 0 {
`

nityanandagohain: No since, if panel type is list but len is not 0 then the next line will execute

makeavish (approved):

Issue #3019, opened 2023-07-04
Query range API failure


makeavish:
Query range API fails on below query:
<img width="1240" alt="Screenshot 2023-07-04 at 12 33 43" src="https://github.com/SigNoz/signoz/assets/9512100/ceff0eea-5a45-4a12-b044-ee50e263470e">

<details open>
<summary>cURL</summary>

```
curl 'https://testingapp.signoz.io/api/v3/query_range' \
  -H 'Accept: application/json, text/plain, */*' \
  -H 'Accept-Language: en-GB,en-US;q=0.9,en;q=0.8' \
  -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJlbWFpbCI6ImhlbGxvQHNpZ25vei5pbyIsImV4cCI6MTY4ODQ1NTg0MCwiZ2lkIjoiY2UwMTNkOTUtOGE0NS00ZDU4LWI3YzctOTBkYzVmMDU5Nzk1IiwiaWQiOiJjYmE1NWM2YS1jOWNkLTRhZjAtOWU2OS01Mjk0YzNjOWFjOGQifQ.SfIrMOG9Ia3jRKPHBUIXRNPLp34M7FnkDQqh9OP2YNQ' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: application/json' \
  -H 'Origin: http://localhost:3301' \
  -H 'Referer: http://localhost:3301/' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: cross-site' \
  -H 'Sec-GPC: 1' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36' \
  -H 'sec-ch-ua: "Not.A/Brand";v="8", "Chromium";v="114", "Brave";v="114"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
[...]

PR #3023, opened 2023-07-04
Fix/handle hypen attributes

makeavish:
Query range API are failing on attributes with `-` in name: https://github.com/SigNoz/signoz/issues/3019
So updated order by and group by to use backticks.


pkg/query-service/app/traces/v3/query_builder_test.go
```diff
@@ -419,7 +419,7 @@ var testOrderBy = []struct {
 			},
 		},
 		Tags:   []string{},
-		Result: []string{"name asc", "bytes asc", "stringTagMap['response_time'] desc"},
+		Result: []string{"`name` asc", "`bytes` asc", "`stringTagMap['response_time']` desc"},
```
srikanthccv, line 422: This query will not run. The "\`stringTagMap['response_time']\`" will be interpreted as a literal column name instead of the `stringTagMap['response_time']`, but the column name doesn't exist

makeavish: Thanks for pointing out!
Have fixed this by only using backticks on columns in orderBy

srikanthccv (dismissed):

srikanthccv (approved):

Issue #2861, opened 2023-06-08
(traces): add support for table panel type


srikanthccv:
traces backend issue for #2842

PR #3047, opened 2023-07-06
feat: add support for table view in traces

makeavish:
closes(https://github.com/SigNoz/signoz/issues/2861)


pkg/query-service/app/traces/v3/query_builder.go
```diff
@@ -473,7 +473,16 @@ func addOffsetToQuery(query string, offset uint64) string {
 }
 
 func PrepareTracesQuery(start, end int64, queryType v3.QueryType, panelType v3.PanelType, mq *v3.BuilderQuery, keys map[string]v3.AttributeKey) (string, error) {
-	query, err := buildTracesQuery(start, end, mq.StepInterval, mq, constants.SIGNOZ_SPAN_INDEX_TABLENAME, keys, panelType)
+
+	stepInterval := mq.StepInterval
+	// update step interval for table view
+	if panelType == v3.PanelTypeTable {
+		stepInterval = (end - start)/1000
```
srikanthccv, line 480: This breaks with `toStartOfInterval`. Assume the start if 12:45 and the end is 1:15. When we use `toStartOfInterval(..., toIntervalSecond(...))`, you will have two points at 12:30 and 1:00 with data from 12:45-12:59:59 and 1:00-1:15 respectively when we should only return one value for the whole aggregation interval.

pkg/query-service/app/traces/v3/query_builder.go
```diff
@@ -235,22 +235,37 @@ func buildTracesQuery(start, end, step int64, mq *v3.BuilderQuery, tableName str
...
+
+	if panelType == v3.PanelTypeTable {
+		queryTmpl =
+			"SELECT" + selectLabels +
+				" %s as value " +
+				"from " + constants.SIGNOZ_TRACE_DBNAME + "." + constants.SIGNOZ_SPAN_INDEX_TABLENAME +
+				" where " + spanIndexTableTimeFilter + "%s" +
+				"%s%s " +
+				"order by %s"
+	} else if panelType == v3.PanelTypeGraph {
```
nityanandagohain, line 248: add valueType as well.

makeavish: Done

nityanandagohain (approved):

pkg/query-service/app/traces/v3/query_builder.go
```diff
@@ -235,22 +235,37 @@ func buildTracesQuery(start, end, step int64, mq *v3.BuilderQuery, tableName str
...
+	var queryTmpl string
+
+	if panelType == v3.PanelTypeTable {
+		queryTmpl =
+			"SELECT" + selectLabels +
+				" %s as value " +
+				"from " + constants.SIGNOZ_TRACE_DBNAME + "." + constants.SIGNOZ_SPAN_INDEX_TABLENAME +
+				" where " + spanIndexTableTimeFilter + "%s" +
+				"%s%s " +
+				"order by %s"
```
srikanthccv, line 247: this will produce an invalid query when orderBy is empty.

makeavish: Good catch, fixed.
Thanks

srikanthccv:
Address the points related to this from this issue https://github.com/SigNoz/signoz/issues/3095 or anything else that is not covered.

makeavish:
> Address the points related to this from this issue #3095 or anything else that is not covered.

Points related to issue #3095 will be addressed separately.

srikanthccv:
What is the reasoning? That doesn't make sense to me because this PR is adding support for the table view in traces and it is incomplete knowing something is not working. You could address it separately for other panel types.

pkg/query-service/app/traces/v3/query_builder.go
```diff
@@ -350,17 +367,19 @@ func enrichOrderBy(items []v3.OrderBy, keys map[string]v3.AttributeKey) []v3.Ord
...
-	tags = append(tags, "ts")
+func groupBy(panelType v3.PanelType, tags ...string) string {
+	if panelType == v3.PanelTypeGraph || panelType == v3.PanelTypeValue {
+		tags = append(tags, "ts")
+	}
 	return strings.Join(tags, ",")
 }
 
-func groupByAttributeKeyTags(keys map[string]v3.AttributeKey, tags ...v3.AttributeKey) string {
+func groupByAttributeKeyTags(keys map[string]v3.AttributeKey, panelType v3.PanelType, tags ...v3.AttributeKey) string {
```
nityanandagohain, line 377: We can remove  `keys` variable

makeavish: Done.

srikanthccv:
Please check this with the formula. I suspect it will break because the formula needs a join key. One way I workaround this in metrics is by adding `now()` for the timestamp so the `ts` is there for the sake of JOIN and still produces the correct result.

makeavish:
> What is the reasoning? That doesn't make sense to me because this PR is adding support for the table view in traces and it is incomplete knowing something is not working. You could address it separately for other panel types.

Those changes are not related to table view changes. The issue existed previously in the function common for time-series and table view.

srikanthccv:
If it was mentioned that the issue exists within the reader.go then it would have been easier to see why it made sense to think about addressing separately. Otherwise, it's not clear where the issue exists and why it should be addressed separately.

makeavish:
> Please check this with the formula. I suspect it will break because the formula needs a join key. One way I workaround this in metrics is by adding `now()` for the timestamp so the `ts` is there for the sake of JOIN and still produces the correct result.

Addressed.

makeavish:
> Please check this with the formula. I suspect it will break because the formula needs a join key. One way I workaround this in metrics is by adding `now()` for the timestamp so the `ts` is there for the sake of JOIN and still produces the correct result.

Addressed with `now() as ts`.
@srikanthccv

nityanandagohain (commented):
Please update the tests

makeavish:
> Please update the tests

Updated.

srikanthccv (approved):

nityanandagohain (approved):

PR #3070, opened 2023-07-07
fix: disable step interval for all views

yeshev:


makeavish:
We also need to disable aggregation interval in query builder of dashboards.
Screenshot 2023-07-07 at 17 45 30

@yeshev

yeshev:
> We also need to disable aggregation interval in query builder of dashboards. Screenshot 2023-07-07 at 17 45 30
> 
> @yeshev

@makeavish Fixed

makeavish (approved):

PR #3103, opened 2023-07-11
feat: update where clause based on log field

yeshev:
update where clause based on log detail field


palashgdev (approved):

palashgdev:
there are multiple calls while updating the where clause data and opening the page

PR #3104, opened 2023-07-11
fix: autcomplete top level keys

### LINKED ISSUE #3078: [Logs] Autocomplete suggestions not working for top level keys
Autocomplete suggestions not working for top level  keys like trace_id, severity_text etc.

nityanandagohain:
We directly query the column for top level keys, but it was not working since there was a wrong codition, have fixed it.

Fixes #3078

### DIFF
--- a/pkg/query-service/app/clickhouseReader/reader.go
+++ b/pkg/query-service/app/clickhouseReader/reader.go
@@ -3958,7 +3958,10 @@ func (r *ClickHouseReader) GetLogAttributeValues(ctx context.Context, req *v3.Fi
 
 	// if dataType or tagType is not present return empty response
 	if len(req.FilterAttributeKeyDataType) == 0 || len(req.TagType) == 0 || req.FilterAttributeKey == "body" {
-		return &v3.FilterAttributeValueResponse{}, nil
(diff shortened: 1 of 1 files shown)


pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3958,7 +3958,10 @@ func (r *ClickHouseReader) GetLogAttributeValues(ctx context.Context, req *v3.Fi
 
 	// if dataType or tagType is not present return empty response
 	if len(req.FilterAttributeKeyDataType) == 0 || len(req.TagType) == 0 || req.FilterAttributeKey == "body" {
-		return &v3.FilterAttributeValueResponse{}, nil
+		// also check if it is not a top level key
+		if _, ok := constants.StaticFieldsLogsV3[req.FilterAttributeKey]; !ok {
+			return &v3.FilterAttributeValueResponse{}, nil
+		}
```
srikanthccv, line 3964: This LGTM, but I would expect the static fields to have the dataType and the tagType set. Isn't that the case?

srikanthccv (approved):

PR #3114, opened 2023-07-12
chore: update query range telemetry

makeavish:


pkg/query-service/telemetry/telemetry.go
```diff
@@ -73,28 +74,41 @@ func (a *Telemetry) IsSampled() bool {
...
+			}
 		}
-	}
-
-	if compositeQuery.PromQueries != nil && len(compositeQuery.PromQueries) > 0 {
-		if !strings.Contains(compositeQuery.PromQueries["A"].Query, "signoz_") && len(compositeQuery.PromQueries["A"].Query) > 0 {
-			signozMetricsNotFound = true
+	} else if postData.CompositeQuery.QueryType == v3.QueryTypePromQL {
+		for _, query := range postData.CompositeQuery.PromQueries {
+			if strings.Contains(query.Query, "signoz_") && len(query.Query) > 0 {
```
ankitnayan, line 94: @makeavish this `signoz_` check is to identify APM metrics being used at serviceOverview pages. This check should be included for `QueryTypeBuilder` also.

ankitnayan:
@makeavish the `/api/v2/metrics/query_range` api was used only in dashboards and alerts. With the v3 APIs being used at explorer pages also, this would be quite noisy. We should check again the need to capture this with the API being present at many places going forward.

pkg/query-service/telemetry/telemetry.go
```diff
@@ -73,28 +74,41 @@ func (a *Telemetry) IsSampled() bool {
...
+	signozMetricsUsed := false
+
+	if postData.CompositeQuery.QueryType == v3.QueryTypeBuilder {
+		for _, query := range postData.CompositeQuery.BuilderQueries {
+			if query.DataSource == v3.DataSourceLogs {
+				signozLogsUsed = true
+			} else if query.DataSource == v3.DataSourceTraces {
+				signozTracesUsed = true
+			} else if query.DataSource == v3.DataSourceMetrics {
+				signozMetricsUsed = true
```
ankitnayan, line 89: empty panels will make the state of `signozMetricsUsed` to be `true`

PR #3117, opened 2023-07-12
fix: filter change value

yeshev:
Fix for getting value from current autocomplete response instead of all options.


makeavish (changes requested):
Still able to reproduce the bug. Though only on first time.

makeavish (approved):

palashgdev (approved):

Issue #3136, opened 2023-07-14
(metrics): result should respect the user order by selection 


srikanthccv:
## Bug description

Parts of the implementation disregard the user orderby and decides on the final series based on the aggregation value—however, the final result honours the order for labels. We should change the logic to use the user order by to decide which series gets selected also.

PR #3138, opened 2023-07-14
fix: order by selection decides the result series

srikanthccv:
Fixes https://github.com/SigNoz/signoz/issues/3136. Added two tests

1. Order by `operation`, which will ignore the value and return a series based on `operation` (in this case, the series has a lower value)
2. Order by with multiple columns `priority asc`, `status desc`. Which doesn't look at the value again and returns result based on `priority` [two P0] and `status` [200, 400]. 

Partly believe this is going lead to more confusion for most users (I hope I am wrong).


nityanandagohain (commented):
Looks good, can you add one more test where nothing is specified in order by. It should be good to go after that

nityanandagohain (approved):

Issue #3113, opened 2023-07-12
FE: [LOGS] Details page inconsistent


nityanandagohain:
In explorer page when I am in raw view `attributes` prefix is added to attributes in details
![image](https://github.com/SigNoz/signoz/assets/26831659/02cc6437-6b34-4a5c-8082-9d316d849f6d)

But in table view it is not added
![image](https://github.com/SigNoz/signoz/assets/26831659/35dc67e0-fe7a-42ce-8f5f-4aa19aa3c688)

PR #3153, opened 2023-07-17
fix: table view on click is now taking raw logs

palashgdev:
Close https://github.com/SigNoz/signoz/issues/3113


nityanandagohain (commented):
This changes looks good, @palashgdev  we also discussed regarding how the contents are displayed i.e in the old details page `timestamp` is always at the top then `id` etc. But it is not present in the new detail page. Do you want to address it in this same PR ?

palashgdev:
> This changes looks good, @palashgdev we also discussed regarding how the contents are displayed i.e in the old details page `timestamp` is always at the top then `id` etc. But it is not present in the new detail page. Do you want to address it in this same PR ?

different PR please make a issue for same

nityanandagohain (approved):

Issue #3095, opened 2023-07-11
Order By doesn't work consistently in timeseries/table view


makeavish:
This is true for logs and traces.

Scenario: Order by any key in graph/table panel in traces/logs. The result is not sorted by the key.

nityanandagohain:
This will be true for metrics as well right ?

makeavish:
> This will be true for metrics as well right ?

Yes, but we don't have ORDER BY in graph view for metrics. And table view is not ready yet for metrics so couldn't verify this.
@srikanthccv Can you please verify if ORDER BY is working fine in table view of metrics.

srikanthccv:
Table view implementation is yet to be completed.

> Order by any key in graph/table panel in traces/logs. The result is not sorted by the key.


What does it mean to sort results by the key in graph view? In general, order by doesn't have any effect in graph/chart view unless a limit is applied which I think is part of the open PRs.

PR #3163, opened 2023-07-17
fix: ordering of ts and table panel

makeavish:
closes(https://github.com/SigNoz/signoz/issues/3095)

Order of result was not ensured as we were iterating over a map. 
Changed the map to slice and also removed duplicates from the slice.


pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -4182,12 +4182,24 @@ func readRowsForTimeSeriesResult(rows driver.Rows, vars []interface{}, columnNam
...
+	// remove duplicates from keys array
+	encountered := map[string]bool{}
+	uniqKeys := []string{}
+
+	for _, element := range keys {
+		if encountered[element] == false {
+			encountered[element] = true
+			uniqKeys = append(uniqKeys, element)
+		}
+	}
```
srikanthccv, line 4199: You could use the map from above `seriesToAttrs` for a unique set of keys in the `rows.Next()` loop itself.

makeavish: Good idea, updated.

srikanthccv (approved):
Thanks, LGTM

PR #3182, opened 2023-07-20
Failing query range request due to top level operation dependency

Rajat-Dabade:
This PR fixes the query range bad request issue for the service overview latency graph.


frontend/src/container/MetricsApplication/Tabs/Overview/ServiceOverview.tsx
```diff
@@ -49,6 +49,8 @@ function ServiceOverview({
 		[servicename, tagFilterItems, isSpanMetricEnable, topLevelOperationsRoute],
 	);
 
+	const isQueryEnabled = topLevelOperationsRoute.length > 0;
```
palashgdev, line 52: i like it thanks @Rajat-Dabade 👍

frontend/src/container/GridGraphLayout/Graph/index.tsx
```diff
@@ -287,7 +288,10 @@ function GridCardGraph({
 		);
 	}
 
-	if (prevChartDataSetRef?.labels === undefined && queryResponse.isLoading) {
+	if (
+		prevChartDataSetRef?.labels === undefined &&
```
palashgdev, line 292: can you confirm do we need `prevChartDataSetRef` check also ?

Rajat-Dabade: Got it, we don't need this one as it is handled down below.

Rajat-Dabade: I will remove it.

palashgdev (approved):

Issue #2591, opened 2023-04-19
[Logs] Handle top level columns in AggregateOpeatorCount


nityanandagohain:
Top-level columns of the logs model is not handled in AggregateOpeatorCount

nityanandagohain:
Whether it is a top-level column or a materialized column, we will have to check if the value is present in the row of the column or not, the implementation can be finalised once we solve this issue https://github.com/SigNoz/signoz/issues/2594

nityanandagohain:
Should we handle this issue in the same way https://github.com/SigNoz/signoz/pull/3183#issuecomment-1663738241 ?

@srikanthccv

srikanthccv:
I didn't really understand the question

nityanandagohain:
So I mean, count on top level columns should we ingore the default values like '0' or '' or should we consider them. If we consider them then the count of them will remain same as normal count

nityanandagohain:
Closing this issue, as we don't want to add something from our side.

srikanthccv:
@nityanandagohain question: Is this handled for aggregate operators such as min, avg, etc... on columns that are either materialised or not materialised? For example, MIN of attribute_int64_items where `items` is key from the log attributes. I have checked the tests, but I hope the key exists condition is added for them?

nityanandagohain:
@srikanthccv  do you mean min(key), with/without any groupby. Then we should also add key_exists=true right ?

nityanandagohain:
@srikanthccv  thanks for pointing it out, yes that is missing. Let me create a new PR for that.

srikanthccv:
@makeavish you may want to check this as well https://github.com/SigNoz/signoz/issues/2591#issuecomment-1693120878.

PR #3183, opened 2023-07-20
fix: default values for materialized columns added

### LINKED ISSUE #2594: Check if can set a default value if some value is not present in materialized column
Right now if key is not present the default value is set by clickhouse

### LINKED ISSUE #2599: [LOGS] Handle zero values in groupBy
> When using a map from ClickHouse, it (i.e stringTagMap['...']) returns zero values for the rows that don't have the key. The result shouldn't include such values.
Good point!
This is happening because the map values are not nullable. Adding a where clause to check if the key exists can solve this.
Eg: Add has(stringTagMap, 'key') to where clause

> @nityanandagohain This problem might also be there in logs, please check

https://github.com/SigNoz/signoz/pull/2571#discussion_r1168576181

nityanandagohain:
Since the changes are very related added them in same PR.
* Defined Max values for data types, Fixes #2594 
* If there is some key in groupBy call `getExistsFilter` and apply the filter, Fixes #2599
* If there is something in aggregate operator call `getExistsFilter` and apply the filter, Fixes https://github.com/SigNoz/signoz/issues/2591

### DIFF
--- a/pkg/query-service/app/clickhouseReader/reader.go
+++ b/pkg/query-service/app/clickhouseReader/reader.go
@@ -3443,9 +3443,21 @@ func (r *ClickHouseReader) UpdateLogField(ctx context.Context, field *model.Upda
 	if field.Selected {
 		// if the type is attribute or resource, create the materialized column first
 		if field.Type == constants.Attributes || field.Type == constants.Resources {
-			// create materialized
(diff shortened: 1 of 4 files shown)


nityanandagohain:
Going with a different approach!

Issue #1477, opened 2022-08-05
APDEX score support 


pranay01:
## Is your feature request related to a problem?

Many New Relic users are used to seeing an APDEX score for the health of their services. We have had feedback from some users asking for supporting it

## Describe the solution you'd like

Support APDEX scores in SigNoz or some similar measure. We have to understand what extra information does an APDEX score provide which is not included in current SigNoz dashboards.

https://docs.newrelic.com/docs/apm/new-relic-apm/apdex/apdex-measure-user-satisfaction/

Also, not sure if the word 'APDEX' can be used as this may be a NewRelic trademark word

srikanthccv:
We have chosen 500ms as a default threshold. A user with editor/admin access can update it to a sensible value for each service, which will persist in DB. An endpoint will be added to fetch and set the setting for the threshold.

Following the same approach as other graphs, the implementation will allow users to switch between traces or span metrics for the score calculation and uses the `v3/query_range`. The content of the payload is dependent on the data source and settings. It is certainly possible that the user sets a threshold `t`, which doesn't have the exact bucket boundary `le` for the satisfied count or the bucket boundary `le` = `4t` for tolerating count. We will use the nearest upper boundary `le` in such cases.

srikanthccv:
Postman collection here https://signoz.postman.co/workspace/Team-Workspace~cfa5c7b5-124f-4bbc-88d9-2351af8dd95b/collection/21397560-42ea3640-244b-4399-93d7-ba2933c290eb?action=share&creator=21397560

srikanthccv:
@Rajat-Dabade, This is very unlikely but when the upper bound for the threshold doesn't exist in the `le` array, use `+Inf`.

PR #3186, opened 2023-07-21
feat: add support for apdex settings

srikanthccv:
Part of https://github.com/SigNoz/signoz/issues/1477

https://en.wikipedia.org/wiki/Apdex

The default apdex threshold for a service is 500ms, and users can (and ideally should) update the threshold with the appropriate value for the service. They can optionally specify a comma-separated string of status codes to ignore when deciding the error calls. This adds support for persisting and retrieving the settings.

We use the bucket boundaries when this score is calculated from metrics since we don't have raw samples. For example, if the threshold is 500ms, the boundaries used are `le=500` and `le=2000`. See this https://stagingapp.signoz.io/dashboard/4254c4e1-3a5f-4d82-a1cc-a5c084abcb6a for example. The `/api/v1/metric_meta` is added for the `signoz_latency_bucket` meta. This endpoint provides the metric temporality and bucket boundaries. The temporality is important because the final formula for delta and metrics differ.


pkg/query-service/model/db.go
```diff
@@ -36,6 +36,12 @@ type User struct {
 	GroupId            string `json:"groupId,omitempty" db:"group_id"`
 }
 
+type ApdexSettings struct {
+	ServiceName        string  `json:"serviceName" db:"service_name"`
+	Threshold          float64 `json:"threshold" db:"threshold"`
+	ExcludeStatusCodes string  `json:"excludeStatusCodes" db:"exclude_status_codes"` // sqlite doesn't support array type
```
makeavish, line 42: Why exclude status codes is required?

makeavish (approved):

PR #3192, opened 2023-07-24
feat: use global in instead of in for second graph query

### LINKED ISSUE #3160: IN/JOIN in distributed setup
For all our queries which use IN/JOIN on distributed clickhouse setup, we might see this error: https://github.com/ClickHouse/ClickHouse/issues/15888
We need to use global IN/JOIN for individual query. 
Here's a doc for more info: https://clickhouse.com/docs/en/sql-reference/operators/in#distributed-subqueries

nityanandagohain:
Fixes #3160


srikanthccv (approved):
This doesn't change anything about the correctness of the queries or data since you are already using the distributed table in sub-query.

PR #3188, opened 2023-07-23
Key Operation to Metrics using USE_SPAN_METRIC feature flag

Rajat-Dabade:
This PR will shift key operation API calls from traces to metrics using the `USE_SPAN_METRIC` feature flag.

PR #2998, opened 2023-06-29
refactor: remove the dependency of services using redux

Rajat-Dabade:
- Removed the redux dependency to get the list of services.

PR #3196, opened 2023-07-25
Service layer to metrics using `USE_SNAP_METRIC` feature flag

Rajat-Dabade:
The PR added the support to use metric data for the service layer (Service List view). 

Note: this PR is stacked on https://github.com/SigNoz/signoz/pull/3188 and https://github.com/SigNoz/signoz/pull/2998 PRs. So before merging this PR, these two PRs should be merged.


srikanthccv:
If one request fails for any reason, the entire table has no data, even though other requests succeed. Please handle that.

srikanthccv:
@palashgdev, please review; I would like to see this getting in for Wed regular release.

frontend/src/container/MetricsApplication/Tabs/Overview/TopOperationMetrics.tsx
```diff
@@ -87,34 +77,19 @@ function TopOperationMetrics(): JSX.Element {
...
 	if (errorMessage) {
 		return <div>{errorMessage}</div>;
 	}
 
 	return (
 		<QueryTable
-			title={title}
 			query={updatedQuery}
 			queryTableData={queryTableData}
 			loading={isLoading}
```
palashgdev, line 92: if query table takes loading do we still need Spinner ?

frontend/src/container/ServiceApplication/ServiceMetrics/ServiceMetricTable.tsx
```diff
@@ -0,0 +1,62 @@
...
+		queryKey: [
+			`GetMetricsQueryRange-${queryRangeRequestData[0].selectedTime}-${globalSelectedInterval}`,
+			maxTime,
+			minTime,
+			globalSelectedInterval,
+		],
+		keepPreviousData: true,
+		enabled: true,
+		refetchOnMount: false,
+		onError: (error) => {
```
palashgdev, line 33: if we are not using error let's remove this console log

frontend/src/container/ServiceApplication/types.ts
```diff
@@ -0,0 +1,38 @@
...
+import { MetricRangePayloadProps } from 'types/api/metrics/getQueryRange';
+import { ServicesList } from 'types/api/metrics/getService';
+
+export default interface ServiceTableProps {
+	services: ServicesList[];
+	loading: boolean;
+}
+
+export interface ServiceApplicationProps {
+	servicename: string;
```
palashgdev, line 15: IService prop is already present

frontend/src/container/ServiceApplication/ServiceMetrics/ServiceMetricTable.tsx
Rajat-Dabade: Okay

frontend/src/container/ServiceApplication/Columns/ColumnContants.ts
```diff
@@ -0,0 +1,26 @@
+export enum ColumnKey {
+	Application = 'serviceName',
+	P99 = 'p99',
+	ErrorRate = 'errorRate',
+	Operations = 'callRate',
+}
+
+export const ColumnTitle: {
```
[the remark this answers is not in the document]
Rajat-Dabade: Okay

frontend/src/container/ServiceApplication/ServiceMetrics/ServiceMetricTable.tsx
Rajat-Dabade: Handled it with notifications.

frontend/src/container/MetricsApplication/Tabs/Overview/TopOperationMetrics.tsx
Rajat-Dabade: No need removing it.

frontend/src/container/ServiceApplication/types.ts
Rajat-Dabade: Remove it as it is not used anywhere.

palashgdev (approved):

srikanthccv:
I will be doing some more testing.

srikanthccv (approved):
Approving the Service page changes. 

Along with the remaining point from last comment, following in the Key operations table need to be addressed in separate PR.

- The order of columns should be same as the traces (it in random now). 
- The column has so sort asc/desc

palashgdev:
> * The order of columns should be same as the traces (it in random now).

this is from the backend can we ensure the order is from the backend itself or does this need to be handled by the client let me... though there is no logic as of now

srikanthccv:
It needs to be handled on the client side. Please use the ordering of the `queryName`s `A`, `B`, `C`, `F1`, and `F2` for ordering.

palashgdev:
made the issue https://github.com/SigNoz/signoz/issues/3249 to address the enchantment

PR #3197, opened 2023-07-26
feat: introduce PreferRPM feature

makeavish:
Completed in logs and traces but tests of metrics are not updated and querier is also not updated.


makeavish:
@srikanthccv, please review the metrics changes for converting to rpm based on feature flag.
I haven’t updated metrics tests as I wasn’t confident about metrics changes.

srikanthccv:
I would avoid calling checkFeature everywhere, making it hard to follow. If it is enabled, it will be enabled entirely, so there is no need to invoke it in all builder functions. I would also keep the logic rps as is and wrap it with `value*60` when rpm is enabled.

srikanthccv:
Use the `PrepareMetricQuery` to return the rpm query using the prepared rps query. We don't need to change anything for quantiles in metrics for this. Just update those that apply to rpm.

srikanthccv:
Just sharing where I am up to on metrics change - the `value * 60` will vary depending on the panel type and temporality so looking into it.

pkg/query-service/app/traces/v3/query_builder.go
```diff
@@ -302,7 +307,13 @@ func buildTracesQuery(start, end, step int64, mq *v3.BuilderQuery, tableName str
 		v3.AggregateOperatorRateAvg,
 		v3.AggregateOperatorRateMin,
 		v3.AggregateOperatorRate:
-		op := fmt.Sprintf("%s(%s)/%d", aggregateOperatorToSQLFunc[mq.AggregateOperator], aggregationKey, step)
+
+		rate := float64(step)
+		if options.PreferRPM {
```
srikanthccv, line 312: missing in the count operator?

pkg/query-service/constants/constants.go
```diff
@@ -55,6 +56,8 @@ var DurationSortFeature = GetOrDefaultEnv("DURATION_SORT_FEATURE", "true")
 
 var TimestampSortFeature = GetOrDefaultEnv("TIMESTAMP_SORT_FEATURE", "true")
 
+var PreferRPMFeature = GetOrDefaultEnv("PREFER_RPM_FEATURE", "false")
```
srikanthccv, line 59: Can you move this to flag in main.go?

srikanthccv:
Also fix the failing tests

pkg/query-service/app/traces/v3/query_builder.go
makeavish: Count is different from rate. It is expected that count is commutatively calculated over a time interval.
Like count of time period 11:00-11:20 will always be less than or equal to 10:00-12:00.

makeavish:
> Also fix the failing tests

Updated a few tests which were failing.
Also I am not sure about querier changes needed. Can you please check that?
@srikanthccv

pkg/query-service/constants/constants.go
makeavish: Why? 
We are using env variables for feature flags.
Eg: https://signoz.io/docs/operate/feature-flags/#adding-configs-to-query-service

pkg/query-service/app/traces/v3/query_builder.go
srikanthccv: I didn't mean/say they are the same. Frontend has cps, cpm units https://github.com/SigNoz/signoz/blob/2bf534b56f6ca85c75ec6f85dd9e37a4210a89f0/frontend/src/container/NewWidget/RightContainer/alertFomatCategories.ts#L91-L97 so I was asking if the count should bother about this flag.

pkg/query-service/constants/constants.go
srikanthccv: The feature "flag" usually goes in the CLI flags. This is a pretty common practice, and you can see it used (the closest example of what we rely on is a [collector](https://github.com/SigNoz/signoz/blob/develop/deploy/docker/clickhouse-setup/docker-compose.yaml#L218)) in many executable go applications (some argparse underlying libraries support the env version of the flag with `_`, but it's the CLI flag that makes it possible). We may have used env in the past, but I believe we shouldn't do that going forward for feature flags.

srikanthccv (commented):
The querier is not used anywhere and is not functional so the changes are fine.

pkg/query-service/app/traces/v3/query_builder.go
makeavish: counts/sec, ops/sec, requests/sec are all rate. So they should be covered with rate changes.

pkg/query-service/constants/constants.go
makeavish: It might be a common practice, can you list the reasons behind it? How will it be easier for SigNoz users or us to use CLI flags rather than env variables?
I am not against CLI flags or have a hard opinion but I was thinking that it will bloat the APIhandler code. But that should be a much of a problem.

srikanthccv: It's easier to discover and manage. When we use the env for both feature flag and general purpose, they are all over the place with mixed usage with no distinction. What does the feature flag do and what type values does it accept? What are the default values and does it return error when incorrect type given or does it silently skip? and answer to these questions is usually the flags. There maybe other reasons which I am not aware.

srikanthccv (approved):
We would like to use a similar approach by other oss projects in enabling the features flag, a comma-separated list of flags. And avoid the fragmentation for some flags vs some env. 

- https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
- https://prometheus.io/docs/prometheus/latest/feature_flags/
- https://github.com/open-telemetry/opentelemetry-collector/blob/main/featuregate/README.md#controlling-gates

nityanandagohain (approved):

PR #3203, opened 2023-07-26
fix: aggregate options is not intuitive

### LINKED ISSUE #3151: Aggregate Options only shown after selecting `Group By` filter is not intuitive
In the new logs and trace explorer pages, the aggregate option shows `NOOP` until and unless a Group By filter is applied. 

We can show the list of aggregate options and have them disabled instead of locking the entire dropdown. Upon hovering, we can show a pop-up message saying, "You need to apply a `Group By` filter to use this aggregate option.

gitstart:
Fixes #3151 

Loom: https://www.loom.com/share/2372b3a5830c45229bfe2b678c9188fe

---
This code was written and reviewed by GitStart Community. Growing great engineers, one PR at a time.


gitstart:
The fact that it is disabled is related to the selected panel and not if there is any option selected in the groupBy field, even if the fields are related, the Aggregate Options field can be enabled even if there is no selection in the groupBy, so we modified the context of the phrase in the tooltip in addition to modifying the order of the fields.

YounixM:
Stale / Inactive PR. Closing it for now.

Issue #3204, opened 2023-07-26
Query builder breaks with percentile operator in where clause when used with limit


makeavish:
<img width="1237" alt="Screenshot 2023-07-26 at 15 55 52" src="https://github.com/SigNoz/signoz/assets/9512100/4fe2e160-2b85-48ed-ba44-fffe43064e1d">

srikanthccv:
No longer an issue

PR #3205, opened 2023-07-26
fix: percentile operator bug in limit queries

makeavish:
close(https://github.com/SigNoz/signoz/issues/3204)


pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -228,7 +228,7 @@ func buildLogsQuery(panelType v3.PanelType, start, end, step int64, mq *v3.Build
 	}
 
 	if graphLimitQtype == constants.SecondQueryGraphLimit {
-		filterSubQuery = filterSubQuery + " AND " + fmt.Sprintf("(%s) GLOBAL IN (", getSelectKeys(mq.AggregateOperator, mq.GroupBy)) + "%s)"
+		filterSubQuery = filterSubQuery + " AND " + fmt.Sprintf("(%s) GLOBAL IN (", getSelectKeys(mq.AggregateOperator, mq.GroupBy)) + "{{.FilterQuery}})"
```
nityanandagohain, line 231: FilterSubquery

pkg/query-service/app/queryBuilder/query_builder.go
```diff
@@ -201,7 +211,15 @@ func (qb *QueryBuilder) PrepareQueries(params *v3.QueryRangeParamsV3, args ...in
...
 							return nil, err
 						}
-						query := fmt.Sprintf(placeholderQuery, limitQuery)
+						data := map[string]string{
+							"FilterQuery": limitQuery,
+						}
+						t := template.Must(template.New("query").Parse(placeholderQuery))
+						builder := &strings.Builder{}
+						if err := t.Execute(builder, data); err != nil {
+							panic(err)
```
nityanandagohain, line 220: return err

pkg/query-service/app/queryBuilder/query_builder.go
```diff
@@ -181,7 +182,16 @@ func (qb *QueryBuilder) PrepareQueries(params *v3.QueryRangeParamsV3, args ...in
...
```
nityanandagohain, line 191: return err

nityanandagohain (approved):

srikanthccv:
Please help me understand why templating is needed here

makeavish:
> Please help me understand why templating is needed here

If user uses `%` in the filter query then `fmt.Sprintf` was failing as `%` was treated as variable insertion operator. We could have escaped the characters but that could have more code changes. So used templates.

srikanthccv:
I do not think templating should be used because we need to make more code changes. What more changes are we talking about? There is one `ClickHouseFormattedValue` which should take of it.

makeavish:
> I do not think templating should be used because we need to make more code changes. What more changes are we talking about? There is one `ClickHouseFormattedValue` which should take of it.

Apart from filter `%` can occur at multiple places, what if `%` is present in a key? So the cases keeps on increasing. 
Using Template was safest bet for future changes or such scenarios.

srikanthccv:
>Apart from filter % can occur at multiple places, what if % is present in a key? So the cases keeps on increasing

This is very unlikely. I don't think that justifies it. Assume such a key exists for the sake of argument. Does this fix solve all the issues? This is simply checking specific limit cases.

makeavish:
> > Apart from filter % can occur at multiple places, what if % is present in a key? So the cases keeps on increasing
> 
> This is very unlikely. I don't think that justifies it. Assume such a key exists for the sake of argument. Does this fix solve all the issues? This is simply checking specific limit cases.

In cases where we are using limit, it won't be a reproducible but template doesn't fix other such possible cases. We can use template everywhere to avoid the issues.

srikanthccv:
My point is that issue arose because of a special character, and it should be escaped. The idea of templatizing the whole query builder is a new effort from scratch; having seen the builder code, I think it will turn out to be a net negative and (most likely) will not fix the special character problems. I do not think we should introduce a templating here, and "mode code changes" is not a justifiable reason. I can think of many cases where not escaping is the source of bugs (try `level = 'info` in logs or something similar in traces or metrics). I do not want this PR to fix all potential such issues but escaping is the way to go, and it should start with this. I am open here others' thoughts on this.

// @nityanandagohain @rkssisodiya @ankitnayan

ankitnayan:
There must be many golang projects that deal with the same issue. It would be wise to look into a few of them and get inspiration. @makeavish please post your findings

makeavish:
Other projects uses DB library query parameters methods instead of manual escaping. We should also use clickhouse named parameters. We are already using clickhouse named parameters for `getFilteredSpan` function in `reader.go`.

makeavish:
Closing this as we will go ahead with using clickhouse named operators: https://github.com/SigNoz/signoz/issues/3345

PR #3207, opened 2023-07-26
fix: update how limit is added to table queries

### LINKED ISSUE #3206: Table view limit doesn't work in logs explorer
https://www.loom.com/share/408d618fdb854fdfa627ba37c8f834f5?sid=5e379295-4f5d-4aad-81c0-e0cd283acf6e

nityanandagohain:
Fixes #3206

### DIFF
--- a/pkg/query-service/app/logs/v3/query_builder.go
+++ b/pkg/query-service/app/logs/v3/query_builder.go
@@ -442,7 +442,7 @@ func PrepareLogsQuery(start, end int64, queryType v3.QueryType, panelType v3.Pan
 		query, err = reduceQuery(query, mq.ReduceTo, mq.AggregateOperator)
 	}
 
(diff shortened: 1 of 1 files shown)


makeavish (approved):

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -452,8 +452,9 @@ func PrepareLogsQuery(start, end int64, queryType v3.QueryType, panelType v3.Pan
 		} else {
 			query = addLimitToQuery(query, mq.Limit)
 		}
+	} else if panelType == v3.PanelTypeTable {
+		query = addLimitToQuery(query, mq.Limit)
```
srikanthccv, line 456: How does this fix the issue? There is an else clause after `if mq.PageSize > 0`, which is doing the same thing. Is pageSize non-zero for table type?

srikanthccv (approved):

PR #3211, opened 2023-07-27
feat: usage exporter logic updated

nityanandagohain:
Fixes https://github.com/SigNoz/engineering-pod/issues/1005

* Use cron instead of interval and upload at 00:00
* Upload usage when starting up QS
* Upload usage before shutdown of QS


srikanthccv:
My familiarity with this code is virtually nil; I can still take a look at it but consider the review of someone who knows it better.

nityanandagohain:
@ankitnayan  should we merge it ?

PR #3221, opened 2023-07-28
feat: hotfix check if limit is not zero

nityanandagohain:
Add limit if it is non zero for panel type table

### DIFF
--- a/pkg/query-service/app/logs/v3/query_builder.go
+++ b/pkg/query-service/app/logs/v3/query_builder.go
@@ -452,7 +452,7 @@ func PrepareLogsQuery(start, end int64, queryType v3.QueryType, panelType v3.Pan
 		} else {
 			query = addLimitToQuery(query, mq.Limit)
 		}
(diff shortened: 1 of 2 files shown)


pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -452,7 +452,7 @@ func PrepareLogsQuery(start, end int64, queryType v3.QueryType, panelType v3.Pan
 		} else {
 			query = addLimitToQuery(query, mq.Limit)
 		}
-	} else if panelType == v3.PanelTypeTable {
+	} else if panelType == v3.PanelTypeTable && mq.Limit > 0 {
```
makeavish, line 455: I think Limit>0 condition can be added to `addLimitToQuery` function itself. This avoids repeated checks for limit>0 outside the function.

makeavish (approved):

srikanthccv (approved):

PR #3229, opened 2023-07-29
feat: show page name in document title using React Helmet

### LINKED ISSUE #3201: Show page name instead of default heading in SigNoz app
Screenshot 2023-07-26 at 12 56 51

mustajab-ikram:
I have completed the task related to Issue #3201. I have added the Page Title functionality using React Helmet as discussed in our previous chats. 
Please take a look at the PR when you get a chance. Let me know if there are any further changes or improvements you would like to see.
image
close #3201


frontend/src/constants/routeTitleMap.ts
```diff
@@ -0,0 +1,38 @@
+const routeTitleMap: { [key: string]: string } = {
```
palashgdev, line 1: Record<keyof typeof ROUTES, string>

frontend/src/constants/routeTitleMap.ts
```diff
@@ -0,0 +1,38 @@
...
+	SIGN_UP: 'SigNoz | Sign Up',
```
palashgdev, line 2: would be good if we could incorporate translation also

palashgdev:
thanks, @mustajab-ikram for the PR 

just small chances would be good to have

mustajab-ikram:
fix: Add translation support for page titles

- Incorporate translation in the Head component to dynamically set page titles based on the current route.
- Add translation files for 'en' and 'en-GB' locales for the page titles.

palashgdev (approved):
thanks @mustajab-ikram for raising the pr 

small, enhancement which I did updated the return type of the function and we already have AppLayout so move the logic there

And thanks for sharing the pr :-D working seamlessly

PR #3243, opened 2023-08-01
feat: limit offset logic updated for logs list view

### LINKED ISSUE #3242: LOGS: Limit offset issue in logs list view
Limit doesn't work when order by is ts

nityanandagohain:
Fixes #3242

### DIFF
--- a/pkg/query-service/app/logs/v3/query_builder.go
+++ b/pkg/query-service/app/logs/v3/query_builder.go
@@ -413,6 +413,13 @@ type Options struct {
 	IsLivetailQuery bool
 }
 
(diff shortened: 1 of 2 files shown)


srikanthccv:
I couldn't understand from the issue/PR description why it doesn't work. Please help me understand the issue.

nityanandagohain:
So we didn't handle how limit should work when the order by is timestamp, this PR handles this. So now the frontend has to send limit and offset as well for order by timestamp.


Also it handles the edge cases were limit is less that pagesize or the last page where actual values returned to should be offset-limit instead of pagesize

@srikanthccv

srikanthccv (approved):

PR #3244, opened 2023-08-01
fix: alerts work when "equals to" or "not equals to" used with "all t…

### LINKED ISSUE #2953: alerts with "all the time" triggers not working as expected
**What did you do?**

Created an alert to know when some endpoints are down using synthetic checks.

**What did you expect to see?**

Alert trigger only when those endpoints are down "all the time" for 5 minutes.

**What did you see instead? Under which circumstances?**

Alerts trigger even when they are down for few seconds and recovered.

**Environment**

- testbed (internal)

**Alertmanager version:**

	version=0.23.0
	branch=release/v0.23.0-0.2
	revision=f3e6c2166de499b7a9b8f53de278ddd56f3ded07

/cc @ankitnayan @srikanthccv

srikanthccv:
…he times" or "at least once"

Fixes #2953 

The logic here doesn't take care of the compare operators `=` and `!=` https://github.com/SigNoz/signoz/blob/ae3d4fece8523f2fd4da65e95415eeb1da367a05/pkg/query-service/rules/thresholdRule.go#L508-L524

As a result, the initial value mapped here will be used for evaluation, which is incorrect.

https://github.com/SigNoz/signoz/blob/ae3d4fece8523f2fd4da65e95415eeb1da367a05/pkg/query-service/rules/thresholdRule.go#L540

To correctly evaluate, we need all values for the series (or let me know if there is a better way that works for all cases). This commit fixes the issue by preparing `Vs` and then running the compare operator logic.


srikanthccv:
Please review/test

prashant-shahi (approved):
:shipit: LGTM 

Created an alert with "all the time" within "last 5 minutes".

**Scene 1: set threshold to test for false positive case => no alert**

Screenshot 1

**Scene 2: set threshold to test positive case => alert after the 5 minute grouping time**

Screenshot 2

Issue #1463, opened 2022-08-02
Alerts query-builder support for `absent`


prashant-shahi:
Not able to replicate equivalent of the following:

```
absent(up{host_name="my-machine-hostname",job="otel-collector"})
```

The above query helps to figure out if any of the agents are down:
- `1` when the agent is up
- no values are returned when the agent is down

cc @mindhash

srikanthccv:
@pranay01 One way we could do this is to give a checkbox option to select whether to send an absent alert when data is missing. Please let us know what you think.

pranay01:
@srikanthccv How do products like Grafana handle this issue

PR #3245, opened 2023-08-01
feat: add support for alerting on absent metric

srikanthccv:
Part of https://github.com/SigNoz/signoz/issues/1463


prashant-shahi (dismissed):
Requesting changes below:
- The metric absent section should be below the threshold section.
- Duration of metrics absent should not accept a negative number.

_Screenshot:_

Screenshot

srikanthccv:
I will fix them. Let me know if there are any other functional issues.

pkg/query-service/rules/alerting.go
```diff
@@ -142,6 +142,8 @@ type RuleCondition struct {
 	CompositeQuery *v3.CompositeQuery `json:"compositeQuery,omitempty" yaml:"compositeQuery,omitempty"`
 	CompareOp      CompareOp          `yaml:"op,omitempty" json:"op,omitempty"`
 	Target         *float64           `yaml:"target,omitempty" json:"target,omitempty"`
+	AlertOnAbsent  bool               `yaml:"alertOnAbsent,omitempty" json:"alertOnAbsent,omitempty"`
+	For            time.Duration      `yaml:"for,omitempty" json:"for,omitempty"`
```
ankitnayan, line 146: can this`For` be used for things other than `Absent`? Else we should keep `For` in a struct or use `AbsentFor`

pkg/query-service/rules/thresholdRule.go
```diff
@@ -603,6 +606,18 @@ func (r *ThresholdRule) runChQuery(ctx context.Context, db clickhouse.Conn, quer
 
 	zap.S().Debugf("ruleid:", r.ID(), "\t resultmap(potential alerts):", len(resultMap))
 
+	// if the data is missing for `For` duration then we should send alert
+	if r.ruleCondition.AlertOnAbsent && r.lastTimestampWithDatapoints.Add(r.Condition().For).Before(time.Now()) {
+		zap.S().Debugf("ruleid:", r.ID(), "\t msg: no data found for rule condition")
+		lbls := labels.NewBuilder(labels.Labels{})
+		lbls.Set("lastSeen", r.lastTimestampWithDatapoints.Format("2006-01-02 15:04:05"))
```
ankitnayan, line 613: We should define the format somewhere else

pkg/query-service/rules/resultTypes.go
```diff
@@ -15,7 +15,8 @@ type Vector []Sample
 type Sample struct {
 	Point
 
-	Metric labels.Labels
+	Metric    labels.Labels
+	IsMissing bool
```
ankitnayan, line 19: Where is `IsMissing` referenced apart from being populated? Maybe not a good idea to change `Sample` struct to handle `absent` case?

pkg/query-service/rules/thresholdRule.go
```diff
@@ -447,6 +449,7 @@ func (r *ThresholdRule) runChQuery(ctx context.Context, db clickhouse.Conn, quer
 		if err := rows.Scan(vars...); err != nil {
 			return nil, err
 		}
+		r.lastTimestampWithDatapoints = time.Now()
```
ankitnayan, line 452: what if the QS restarts? If the new eval period does not have any instance of the fingerprint, `rows` will be empty. Should we use CH query to get the `lastTimestampWithDatapoints`?

srikanthccv: It would still trigger an alert because the default timestamp would be before but the `lastSeen` label we added below would be incorrect. I don't think there is a need to query the ClickHouse again. If it didn't fire when the query service was running but becomes true for new eval after the restart we know that past period data existed and can use that information for lastSeen.

srikanthccv:
While this is helpful, it has certain limitations. It won't be useful when there is partial data for metrics. For instance, take host metrics, a user wants to get alerted when any host stops sending data. They would have to create one alert for each host which is not practical. The common way to achieve this is to use two queries one with last for 10 minutes back and one with a current window and alert. That would be possible when we have support for timeshift function.

pkg/query-service/rules/alerting.go
srikanthccv: Updated to `AbsentFor`. And also prom queries now support range queries.

pkg/query-service/rules/resultTypes.go
srikanthccv: I remembered why I added this. When the sample is missing the description and summary should be updated.

ankitnayan (approved):

ankitnayan:
please add tests with some edge/different cases

YounixM (approved):

PR #3250, opened 2023-08-02
Refactor: Created utility for generating columns for `ResizeTable`

Rajat-Dabade:


palashgdev:
shared the pointers offline ... this is not i am expecting 

as there should be a base static array and dynamic props would be added based on the key

palashgdev (approved):

PR #3275, opened 2023-08-07
feat: get and update log fields updated to support new materialized c…

### LINKED ISSUE #2594: Check if can set a default value if some value is not present in materialized column

nityanandagohain:
…olumns and index

Fixes #2594


srikanthccv:
@nityanandagohain are there any changes yet to be done following yesterday discussion or does it have all changes now?

nityanandagohain:
No changes are left @srikanthccv , have made the change where we were changing the format of naming the materialized column

srikanthccv:
Thanks for the confirmation. I will review.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3427,36 +3427,88 @@ func (r *ClickHouseReader) GetLogFields(ctx context.Context) (*model.GetFieldsRe
...
 
 func (r *ClickHouseReader) UpdateLogField(ctx context.Context, field *model.UpdateField) *model.ApiError {
+	// don't allow updating static fields
+	if field.Type == constants.Static {
+		err := errors.New("cannot update static fields")
+		return &model.ApiError{Err: err, Typ: model.ErrorBadData}
+	}
+
+	prefix := field.Type[:len(field.Type)-1]
+	// columns name is <type>_<name>_<datatype>
```
srikanthccv, line 3455:
```suggestion
	// columns name is <type>_<datatype>_<name>
```

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3427,36 +3427,88 @@ func (r *ClickHouseReader) GetLogFields(ctx context.Context) (*model.GetFieldsRe
...
 	if field.Selected {
-		// if the type is attribute or resource, create the materialized column first
-		if field.Type == constants.Attributes || field.Type == constants.Resources {
-			// create materialized
-			query := fmt.Sprintf("ALTER TABLE %s.%s ON CLUSTER %s ADD COLUMN IF NOT EXISTS %s %s MATERIALIZED %s_%s_value[indexOf(%s_%s_key, '%s')] CODEC(LZ4)", r.logsDB, r.logsLocalTable, cluster, field.Name, field.DataType, field.Type, strings.ToLower(field.DataType), field.Type, strings.ToLower(field.DataType), field.Name)
+		keyColName := fmt.Sprintf("%s_%s_key", field.Type, strings.ToLower(field.DataType))
+		valueColName := fmt.Sprintf("%s_%s_value", field.Type, strings.ToLower(field.DataType))
+
+		// create materialized column
+		query := fmt.Sprintf("ALTER TABLE %s.%s ON CLUSTER %s ADD COLUMN IF NOT EXISTS %s %s MATERIALIZED %s[indexOf(%s, '%s')] CODEC(LZ4)",
```
srikanthccv, line 3464: >In ClickHouse Cloud, we utilize the [ZSTD compression algorithm](https://en.wikipedia.org/wiki/Zstd) (with a [default](https://clickhouse.com/docs/en/operations/server-configuration-parameters/settings#server-settings-compression) value of 1) by default. While compression speeds can vary for this algorithm, depending on the compression level (higher = slower), it has the advantage of being consistently fast on decompression (around [20% variance](https://engineering.fb.com/2016/08/31/core-data/smaller-and-faster-data-compression-with-zstandard/)) and also benefiting from the ability to be parallelized. Our historical tests also suggest that this algorithm is often sufficiently effective and can even outperform LZ4 combined with a codec. It is effective on most data types, and information distributions and thus is a sensible general-purpose default and why our initial earlier compression is already excellent. In our results below, this value is represented as “DEFAULT(ZSTD)”.


>ZSTD all the way - ZSTD with no codec often outperforms other options concerning compression or is at the very least competitive: especially for floating points. This is thus our default compression in ClickHouse Cloud.


>LZ4 over ZSTD if possible - if you get comparable compression between LZ4 and ZSTD, favor the former since it offers faster decompression and needs less CPU. However, ZSTD will outperform LZ4 by a significant margin in most cases. Some of these codecs may work faster in combination with LZ4 while providing similar compression compared to ZSTD without a codec. This will be data specific, however, and requires testing.


https://clickhouse.com/blog/optimize-clickhouse-codecs-compression-schema

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3427,36 +3427,88 @@ func (r *ClickHouseReader) GetLogFields(ctx context.Context) (*model.GetFieldsRe
...
+			valueColName,
+			keyColName,
+			field.Name,
+		)
 
-			err := r.db.Exec(ctx, query)
-			if err != nil {
-				return &model.ApiError{Err: err, Typ: model.ErrorInternal}
-			}
+		// create an exists column
```
srikanthccv, line 3473: this isn't an `_exists` column?

```suggestion
		// create a column
```

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3427,36 +3427,88 @@ func (r *ClickHouseReader) GetLogFields(ctx context.Context) (*model.GetFieldsRe
...
+			cluster,
+			colname, field.DataType,
+		)
+		err = r.db.Exec(ctx, query)
+		if err != nil {
+			return &model.ApiError{Err: err, Typ: model.ErrorInternal}
+		}
+
+		// create exists column
+		query = fmt.Sprintf("ALTER TABLE %s.%s ON CLUSTER %s ADD COLUMN IF NOT EXISTS %s_exists bool MATERIALIZED if(indexOf(%s, '%s') != 0, true, false) CODEC(LZ4)",
```
srikanthccv, line 3490: Same as above ZSTD

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3466,20 +3518,52 @@ func (r *ClickHouseReader) UpdateLogField(ctx context.Context, field *model.Upda
...
 
 	} else {
-		// remove index
-		query := fmt.Sprintf("ALTER TABLE %s.%s ON CLUSTER %s DROP INDEX IF EXISTS %s_idx", r.logsDB, r.logsLocalTable, cluster, field.Name)
+		// Delete the index first
+		query := fmt.Sprintf("ALTER TABLE %s.%s ON CLUSTER %s DROP INDEX IF EXISTS %s_idx", r.logsDB, r.logsLocalTable, cluster, colname)
 		err := r.db.Exec(ctx, query)
-		// we are ignoring errors with code 341 as it is an error with updating old part https://github.com/SigNoz/engineering-pod/issues/919#issuecomment-1366344346
-		if err != nil && !strings.HasPrefix(err.Error(), "code: 341") {
+		if err != nil {
```
srikanthccv, line 3538: You have been getting non-nil errors here since we do not use replicated merge trees. For replicated tables, all ALTER queries are async. Recalling our last conv during standup, we decided to show an error when the log field materializes operation fails; I think this should handle async also. It might be fine for now, but you should keep this in mind because we are going to move to replicated at some time. Please create a separate issue. This is applicable for all ALTER statements in this function.

pkg/query-service/app/logs/parser.go
```diff
@@ -304,6 +304,13 @@ func replaceFieldInToken(queryToken string, selectedFieldsLookup map[string]mode
 			} else if strings.Compare(strings.ToLower(*col), "fulltext") != 0 && field.Type != constants.Static {
 				return "", fmt.Errorf("field not found for filtering")
 			}
+		} else {
+			field := selectedFieldsLookup[sqlColName]
+			if field.Type != constants.Static {
+				prefix := field.Type[:len(field.Type)-1]
+				// columns name is <type>_<name>_<datatype>
+				sqlColName = fmt.Sprintf("%s_%s_%s", strings.ToLower(prefix), strings.ToLower(field.DataType), sqlColName)
+			}
```
srikanthccv, line 313: Couldn't follow this. Please help me understand this.

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -84,7 +84,21 @@ func getClickhouseColumnName(key v3.AttributeKey) string {
 		columnType := getClickhouseLogsColumnType(key.Type)
 		columnDataType := getClickhouseLogsColumnDataType(key.DataType)
 		clickhouseColumn = fmt.Sprintf("%s_%s_value[indexOf(%s_%s_key, '%s')]", columnType, columnDataType, columnType, columnDataType, key.Key)
+		return clickhouseColumn
 	}
+
+	// check if it is a static field
+	if key.Type == v3.AttributeKeyTypeUnspecified {
```
srikanthccv, line 91: What do you think about adding a new `AttributeKeyTypeStatic` type here?

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3427,36 +3427,88 @@ func (r *ClickHouseReader) GetLogFields(ctx context.Context) (*model.GetFieldsRe
...
 }
 
-func isSelectedField(tableStatement, field string) bool {
-	return strings.Contains(tableStatement, fmt.Sprintf("INDEX %s_idx", field))
+func isSelectedField(tableStatement string, field model.LogField) bool {
+	// in case of attributes and resources, if there is a materialized column present then it is selected
+	// TODO: handle partial change complete eg:- index is removed but materialized column is still present
+	prefix := field.Type[:len(field.Type)-1]
+	name := fmt.Sprintf("%s_%s_%s", strings.ToLower(prefix), strings.ToLower(field.DataType), field.Name)
+	return strings.Contains(tableStatement, fmt.Sprintf("`%s`", name))
```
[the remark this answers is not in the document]
nityanandagohain: Yeah since previously we didn't remove the column when a field is converted back to interesting, because of which we just checked the index, but in this case we are checking the column.

pkg/query-service/app/logs/parser.go
nityanandagohain: So in the previous logs parser if we send a query as `method gt 200`

the backend first extracts the name `method` and check if it is an selected field, if not then it converts to the correct attribute name i.e `attributes_int64_vale[indexOf(attributes_in64_key, 'method'')]` for SQL query

But if it is a selected field then it add `method > 200` in the SQL query which is changed with recent changes where it should be `attributes_int64_method > 200`

pkg/query-service/app/clickhouseReader/reader.go
nityanandagohain: Thanks have created an issue here https://github.com/SigNoz/signoz/issues/3324

pkg/query-service/app/logs/v3/query_builder.go
nityanandagohain: Definately it will be good addition, but can be scoped as different PR.

srikanthccv:
Thanks for addressing comments. I think this is mostly good, but I will retake a look.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3434,36 +3434,84 @@ func (r *ClickHouseReader) GetLogFields(ctx context.Context) (*model.GetFieldsRe
...
+		err := r.db.Exec(ctx, query)
+		if err != nil {
+			return &model.ApiError{Err: err, Typ: model.ErrorInternal}
+		}
 
-			err := r.db.Exec(ctx, query)
-			if err != nil {
-				return &model.ApiError{Err: err, Typ: model.ErrorInternal}
-			}
+		query = fmt.Sprintf("ALTER TABLE %s.%s ON CLUSTER %s ADD COLUMN IF NOT EXISTS %s %s MATERIALIZED -1",
```
srikanthccv, line 3484: Why is this `MATERIALIZED -1`?

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3434,36 +3434,84 @@ func (r *ClickHouseReader) GetLogFields(ctx context.Context) (*model.GetFieldsRe
...
+			colname,
+			keyColName,
+			field.Name,
+		)
+		err = r.db.Exec(ctx, query)
+		if err != nil {
+			return &model.ApiError{Err: err, Typ: model.ErrorInternal}
+		}
+
+		query = fmt.Sprintf("ALTER TABLE %s.%s ON CLUSTER %s ADD COLUMN IF NOT EXISTS %s_exists bool MATERIALIZED false",
```
srikanthccv, line 3507: Same question; this is just a view table for underlying local tables; why does this need `MATERIALIZED false`?

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3883,8 +3963,15 @@ func (r *ClickHouseReader) GetLatencyMetricMetadata(ctx context.Context, metricN
...
 
-func isColumn(tableStatement, field string) bool {
-	return strings.Contains(tableStatement, fmt.Sprintf("`%s` ", field))
+func isColumn(tableStatement, attrType, field, datType string) bool {
+	// value of attrType will be `resource` or `tag`, if `tag` change it to `attribute`
+	prefix := attrType
+	if attrType == string(v3.AttributeKeyTypeTag) {
+		prefix = constants.Attributes[:len(constants.Attributes)-1]
+	}
+	name := fmt.Sprintf("%s_%s_%s", strings.ToLower(prefix), field, strings.ToLower(datType))
```
srikanthccv, line 3972: Shouldn't the `strings.ToLower(datType)` come in the middle? Can you use this `isColumn` and the new `prepareColumn`? elsewhere in this PR to avoid potentially missing updating in multiple places?

pkg/query-service/app/clickhouseReader/reader.go
nityanandagohain: So if you don't put something here, the insertion doesn't work on the distributed table.

Eg:- if I just add column as `attribute_int64_x int64` . Inserting to this table will give error that `attribute_int64_x` is materialized.

We used this from previous implementation where I found this solution somewhere.

pkg/query-service/app/clickhouseReader/reader.go
nityanandagohain: Same as above 

Eg:- if I just add column as `attribute_int64_x_exists bool` . Inserting to this table will give error that `attribute_int64_x_exists` is materialized. But for bool columns -1 doesn't work so had to add false here.

pkg/query-service/app/clickhouseReader/reader.go
nityanandagohain: Yeah makes sense.

nityanandagohain: Have made the changes

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3434,36 +3434,81 @@ func (r *ClickHouseReader) GetLogFields(ctx context.Context) (*model.GetFieldsRe
 func extractSelectedAndInterestingFields(tableStatement string, fieldType string, fields *[]model.LogField, response *model.GetFieldsResponse) {
 	for _, field := range *fields {
 		field.Type = fieldType
-		if isSelectedField(tableStatement, field.Name) {
+		// all static fields are assumed to be selected as we don't allow changing them
```
srikanthccv, line 3437: nit: You probably want to update the references, such as "interesting" here https://github.com/SigNoz/signoz/blob/ec7c99dd26566b5b161a1dedab104e23485dff36/pkg/query-service/constants/constants.go#L200-L226

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3434,36 +3434,81 @@ func (r *ClickHouseReader) GetLogFields(ctx context.Context) (*model.GetFieldsRe
...
-	return strings.Contains(tableStatement, fmt.Sprintf("INDEX %s_idx", field))
+func isSelectedField(tableStatement string, field model.LogField) bool {
+	// in case of attributes and resources, if there is a materialized column present then it is selected
+	// TODO: handle partial change complete eg:- index is removed but materialized column is still present
+	name := utils.GetClickhouseColumnName(field.Type, field.DataType, field.Name)
+	return strings.Contains(tableStatement, fmt.Sprintf("`%s`", name))
 }
 
 func (r *ClickHouseReader) UpdateLogField(ctx context.Context, field *model.UpdateField) *model.ApiError {
+	// don't allow updating static fields
```
srikanthccv, line 3454: question: I assume from the frontend side static fields won't be shown in settings since it doesn't make sense to keep them if they can't be modified?

pkg/query-service/utils/format.go
```diff
@@ -221,3 +222,16 @@ func getPointerValue(v interface{}) interface{} {
...
 	}
 }
+
+func GetClickhouseColumnName(typeName string, dataType, field string) string {
+	if typeName == string(v3.AttributeKeyTypeTag) {
+		typeName = constants.Attributes
+	}
+
+	if typeName != string(v3.AttributeKeyTypeResource) {
+		typeName = typeName[:len(typeName)-1]
```
srikanthccv, line 232: This can result in panic when typeName is empty. I assume that doesn't happen as of now, but to that law, Anything that can go wrong will go wrong at some point later. You can address this separately.

srikanthccv (approved):
non blocking comments

pkg/query-service/utils/format.go
nityanandagohain: Sure

pkg/query-service/app/clickhouseReader/reader.go
nityanandagohain: So new page for indexing is not designed here we will make the decision there, but yes showing them won't make sense.

pkg/query-service/app/clickhouseReader/reader.go
nityanandagohain: Have made the change

srikanthccv (dismissed):

srikanthccv (approved):
Reviewed this latest commit https://github.com/SigNoz/signoz/pull/3275/commits/062f69572bda7cbc30d70d24b83802b81707476b. LGTM

deploy/docker/clickhouse-setup/docker-compose.yaml
```diff
@@ -3,7 +3,8 @@ version: "2.4"
 x-clickhouse-defaults:
   &clickhouse-defaults
   restart: on-failure
-  image: clickhouse/clickhouse-server:22.8.8-alpine
+  # addding non LTS version due to this fix https://github.com/ClickHouse/ClickHouse/commit/32caf8716352f45c1b617274c7508c86b7d1afab
+  image: clickhouse/clickhouse-server:23.7.3-alpine
```
prashant-shahi, line 7: update the same for Core, Swarm and Query-Service Tests directories

prashant-shahi (approved):

srikanthccv (approved):

PR #3280, opened 2023-08-07
fix: key operation table title for metrics

Rajat-Dabade:


palashgdev:
we are breaking a principle of dry (already there is a function which returns the same string can we re-use that one) here 

also, function returns the string there is no use of memo

palashgdev (approved):

PR #3282, opened 2023-08-07
fix: getting first staged query

yeshev:


palashgdev:
probably we should seek for next available query as first is disabled we need to check for next available query

yeshev:
> probably we should seek for next available query as first is disabled we need to check for next available query

@palashgdev Fixed

palashgdev (approved):

PR #3284, opened 2023-08-07
fix: pipeline change history: update config deployment status without having to reload

rkssisodiya:
pipeline-status-auto-update-demo

### DIFF
--- a/frontend/src/container/PipelinePage/Layouts/ChangeHistory/index.tsx
+++ b/frontend/src/container/PipelinePage/Layouts/ChangeHistory/index.tsx
@@ -5,12 +5,13 @@ import { changeHistoryColumns } from '../../PipelineListsView/config';
 import { HistoryTableWrapper } from '../../styles';
 import { historyPagination } from '../config';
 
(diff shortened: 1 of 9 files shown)


frontend/src/container/PipelinePage/Layouts/ChangeHistory/index.tsx
```diff
@@ -1,24 +1,48 @@
...
 import { changeHistoryColumns } from '../../PipelineListsView/config';
 import { HistoryTableWrapper } from '../../styles';
 import { historyPagination } from '../config';
 
-function ChangeHistory({ piplineData }: ChangeHistoryProps): JSX.Element {
+function ChangeHistory({
+	refetchPipelineLists,
+	pipelineData,
+}: ChangeHistoryProps): JSX.Element {
+	useEffect(() => {
```
palashgdev, line 13: would recommend to avoid use of useEffect and make use of `refetchInterval` from useQuery

palashgdev: i guess we can write custom logic in the refetchInterval also

palashgdev (approved):

PR #3285, opened 2023-08-07
feat: feature flag rpm is being taken care of

palashgdev:


frontend/src/container/ServiceApplication/Columns/ServiceColumn.ts
```diff
@@ -56,3 +60,9 @@ export const getColumns = (
 		dynamicColumnOption,
 	});
 };
+
+interface ITableColumnProps {
```
Rajat-Dabade, line 64: Can we move this to `src/container/ServiceApplication/types.ts`

Rajat-Dabade (changes requested):
Just a small minor refactoring change, everything else LGTM 👍

Rajat-Dabade:
Also, the build pipeline is broken.

Rajat-Dabade (approved):

makeavish:
In graph use ops/min instead of op/s when PreferRPM feature flag is true
Screenshot 2023-08-10 at 06 47 45

PR #3288, opened 2023-08-08
feat: menu header is controlled from menuList rather than boolean for each menu

palashgdev:


Rajat-Dabade (approved):
LGTM 👍 Awesome work @palashgdev 🚀

Rajat-Dabade (approved):

PR #3290, opened 2023-08-08
fix: empty filters field

### LINKED ISSUE #3274: Broken where clause in logs explorer
Where clause in logs explorer doesn't work when multiple queries are added.

The query is not built correctly at frontend, filter param is empty: 
image

yeshev:
Close #3274


palashgdev:
> I only verified that payload is sent. @palashgdev please review the code with focus on not introducing any regression.

yeah I have tested major scenarios and code looks good to me :-D

makeavish (approved):

palashgdev (approved):

nityanandagohain (approved):

Issue #3210, opened 2023-07-27
Operator `IN` is case sensetive in query.


Rajat-Dabade:
## Bug description
- While trying to build the query for the latency graph, notices that the operator `IN` was case sensitive in `filter.items` for traces/metrics.
- The query_range API call was returning with a `404` status code. 

![Screenshot 2023-07-19 at 7 53 35 AM](https://github.com/SigNoz/signoz/assets/26198062/960abfa4-3fd3-4ad1-9a7d-8efad71e04ab)
![Screenshot 2023-07-19 at 7 53 53 AM](https://github.com/SigNoz/signoz/assets/26198062/fce4df55-9a8f-4573-8db4-643297af9578)

## Expected behavior
- The query_range API should work regardless of the operator `IN` case sensitivity. 

## Version information 
* **Signoz version**: v0.24.0-16-g

palashgdev:
@Rajat-Dabade we need to convert from FE side also

PR #3310, opened 2023-08-10
fix: lowercase filter operator and increase default conn settings

srikanthccv:
This commit allows configuring the ClickHouse client connection options `MaxOpenConns`, `MaxIdleConns`, and `DialTimeout`. The default values from the client library do not work well when tens of alerts are configured, and when several team members in org access the UI, it is easy to have ten running connections busy and the remaining request timeout quickly.

And also fixes https://github.com/SigNoz/signoz/issues/3210

### DIFF
--- a/ee/query-service/app/api/api.go
+++ b/ee/query-service/app/api/api.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"net/http"
+	"time"
 
 	"github.com/gorilla/mux"
 	"go.signoz.io/signoz/ee/query-service/dao"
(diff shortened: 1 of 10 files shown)


srikanthccv:
I believe logs operator is already taken care here https://github.com/SigNoz/signoz/blob/c35f2ec0aaa8790310d8b0514c91ba3c32cd94e6/pkg/query-service/app/logs/v3/query_builder.go#L122

ankitnayan (approved):

srikanthccv:
@makeavish I'd appreciate your review. The latency chart breaks when (cloud) users use filtering


Screenshot 2023-08-10 at 4 18 16 PM
.

makeavish (approved):

makeavish:
> @makeavish I'd appreciate your review. The latency chart breaks when (cloud) users use filtering
> 
> Screenshot 2023-08-10 at 4 18 16 PM .

Wasn't aware that it was causing issues on staging. According to discussions I understood that this was not affecting users currently as it is being handled on the frontend temporarily.

Issue #1958, opened 2023-01-03
Enable `Saved Views` with specific filters in logs search


pranay01:
## Is your feature request related to a problem?

Users may want to save a particular filter which they commonly apply when they explore logs. Currently there is no way to save a filter or view in logs

## Describe the solution you'd like
Provide provision to create `Saved Views` which has specific filters defined so that users can filter the results very easily, especially for queries which they use often

## Describe alternatives you've considered
Typing the filter query in search bar everytime

## Additional context

Intial Design on how this could look like

![views_list](https://user-images.githubusercontent.com/504541/210306074-c96f9bc5-3c12-49a0-b40b-5a2b2ac309f3.png)
![save-view](https://user-images.githubusercontent.com/504541/210306076-fbb46734-f682-40a7-8853-fc4d4e294958.png)

#### *Thank you* for your feature request – we love each and every one!

makeavish:
@pranay01 Please share final design along with requirements.
cc: @palashgdev

pranay01:
@kaushikmacharla07 will share designs on this. Please ignore the earlier designs shared

srikanthccv:
Reopening since the search is yet to be done.

srikanthccv:
@Rajat-Dabade please add the search and address pending comments from the review https://github.com/SigNoz/signoz/pull/3404#pullrequestreview-1602891946

PR #3314, opened 2023-08-10
chore: update saved views endpoints

srikanthccv:
Part of #1958 

- ExplorerQuery -> SavedView
- Add who created/updated view and when is it created/updated.
- Remove the RBAC. We don't have any RBAC for now i.e everyone in org can create/view/update everything.


srikanthccv:
Please review

nityanandagohain (approved):
Looks good to me, but I see the need of creating a new `tags` table in future. If we allow things like filtering by tags etc.

PR #3318, opened 2023-08-11
fix: null issue when switching from clickhouse to Query builder

Rajat-Dabade:


frontend/src/container/QueryBuilder/filters/QueryBuilderSearch/index.tsx
```diff
@@ -125,7 +125,7 @@ function QueryBuilderSearch({
 
 	useEffect(() => {
 		const initialTagFilters: TagFilter = { items: [], op: 'AND' };
-		const initialSourceKeys = query.filters.items.map(
+		const initialSourceKeys = query.filters.items?.map(
```
palashgdev, line 128: this will not solve the issue as items is an empty array

palashgdev (approved):

PR #3319, opened 2023-08-11
fix: resource attribute is fixed trace data source

palashgdev:


srikanthccv (approved):
Tested following cases

- Span metrics enabled
   - With resource attributes
   - Without resource attributes
- Span metrics disabled
  - With resource attributes
  - Without resource attributes

PR #3335, opened 2023-08-15
fix(FE): fix table panel bottom border appearance

### LINKED ISSUE #3238: The bottom border in table panel doesn't look good
The bottom border in table panel doesn't look good, can we hide it, reduce the width of it to a thin line ?- <https://www.loom.com/share/8c4dce3f1fcb4754a6f91251ae333385?sid=55c1a1b3-6786-43ea-aec1-2fdf899786c8>

[Slack Message](https://signoz-team.slack.com/archives/C02B99JKTMY/p1690870230854359)

mustajab-ikram:
Resolve the issue where the bottom border of the table panel doesn't look good. This commit applies the necessary CSS rule to hide the border to improve the visual presentation.

Closes #3238
image


palashgdev:
@mustajab-ikram adding hidden would not work as there might be a case that important info is not visible to the user

Rajat-Dabade:
Hey @mustajab-ikram, I can still reproduce this issue, the step I follow are
 - Try to horizontally scroll the table
 - Hover over the scrollbar.
 - See the issue. 

Screenshot 2023-08-16 at 9 10 04 PM

So here we have 2 scrollbars, one is default HTML and one from Antd Table. We want to remove the default one as it does not go with the theme. Can you please look into it? Thank you.

mustajab-ikram:
> Hey @mustajab-ikram, I can still reproduce this issue, the step I follow are
> 
> * Try to horizontally scroll the table
> * Hover over the scrollbar.
> * See the issue.
> 
> Screenshot 2023-08-16 at 9 10 04 PM
> So here we have 2 scrollbars, one is default HTML and one from Antd Table. We want to remove the default one as it does not go with the theme. Can you please look into it? Thank you.

hey @Rajat-Dabade,
I have removed the scroll bar for now, since this doesn't matches our theme.
image

But I think it would be better if we show a right arrow indicating that table is scroll-able, something like:
image

Should we do this @palashgdev @Rajat-Dabade @pranay01 ?

Rajat-Dabade:
@mustajab-ikram, After making `overflow: hidden` I am not able to see the scroll bar now. This looks good to me 🚀 .

Regarding the right arrow, I don't think that will be a good idea as the Antd table already has the white highlight indicator on either side when there is a scroll in the table.

Rajat-Dabade:
@pranay01, can you please confirm that you are not able to see the scroll in the table now?

Rajat-Dabade (approved):

Issue #3133, opened 2023-07-14
FE: Live view in logs explorer


makeavish:
Subtask of https://github.com/SigNoz/signoz/issues/2822

PR #3337, opened 2023-08-15
feat: get live data

yeshev:
part of #3133


frontend/src/constants/liveTail.ts
```diff
@@ -0,0 +1,3 @@
+export const LIVE_TAIL_HEARTBEAT_TIMEOUT = 10 * 60 * 1000;
```
YounixM, line 1: Comment the logic around LIVE_TAIL_HEARTBEAT_TIMEOUT .

Can directly use `600000` in place of `10 * 60 * 1000`.

frontend/src/container/LiveLogs/LiveLogsListChart/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+		PANEL_TYPES.TIME_SERIES,
+
+		{
+			enabled: isConnectionOpen,
+			refetchInterval: LIVE_TAIL_GRAPH_INTERVAL,
+		},
+		{ dataSource: DataSource.LOGS },
+	);
+
+	const chartData: QueryData[] = useMemo(() => {
```
YounixM, line 47: I think useMemo isn't required here. We aren't computing anything

frontend/src/container/LiveLogs/LiveLogsListChart/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+		if (!data) return [];
+
+		return data.payload.data.result;
+	}, [data]);
+
+	const isLoading = useMemo(
```
YounixM, line 53: Same comment as above

frontend/src/container/LiveLogsTopNav/index.tsx
```diff
@@ -1,30 +1,51 @@
...
+	onOpenConnection,
+}: LiveLogsTopNavProps): JSX.Element {
+	const {
 		isConnectionOpen,
 		isConnectionLoading,
+		initialLoading,
+		handleCloseConnection,
+	} = useEventSource();
+
+	const isPlaying = useMemo(
```
YounixM, line 19: Same comment as above

frontend/src/container/LogsExplorerChart/index.tsx
```diff
@@ -2,25 +2,29 @@ import Graph from 'components/Graph';
...
-	) => ({
-		label: allLabels[index],
-		data: element,
-		backgroundColor: colors[index % colors.length] || 'red',
-		borderColor: colors[index % colors.length] || 'red',
-	});
+	const handleCreateDatasets: Required<GetChartDataProps>['createDataset'] = useCallback(
+		(element, index, allLabels) => ({
+			data: element,
+			backgroundColor: colors[index % colors.length] || 'red',
```
YounixM, line 18: In place of `red`, we should be using the hex from our figma color 

@kaushikmacharla07  :  Please help with the red color hex we want to use across the product.

frontend/src/container/LiveLogs/LiveLogsListChart/index.tsx
yeshev: @YounixM We use it like props with memo component. Also, we have context which causes rerender, so useMemo is a good solution here and below examples. If we don't have computing, it doesn't mean we shouldn't use useMemo.

frontend/src/container/LiveLogs/LiveLogsContainer/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+				{initialLoading ? (
+					<Col span={24}>
+						<Spinner style={{ height: 'auto' }} />
+					</Col>
+				) : (
+					<>
+						<Col span={24}>
+							<LiveLogsListChart />
+						</Col>
+						test
```
YounixM, line 151: Remove `test`

palashgdev (approved):

PR #3338, opened 2023-08-15
fix: Adjust query-service Dockerfiles to make /root/query-service executable for non-root users

rybnico:
Currently, the docker images for the query-service (non-ee and ee) are not usable in environments where running a container as root is not permitted (e.g. hardened Kubernetes, OpenShift).

This PR changes the file permissions of the `/root` directory and the `/root/query-service` binary to `755`, to make the directory and binary readable and executable by everyone. Since it is currently only possible to run this container as `root`, this shouldn't have any security implications.

A more comprehensive fix would include adjusting the Helm chart, docker-compose files etc..


prashant-shahi (approved):
> Currently, the docker images for the query-service (non-ee and ee) are not usable in environments where running a container as root is not permitted (e.g. hardened Kubernetes, OpenShift).

PR change itself looks good to me.

However, I believe we would also require to move out of root paths in query-service and other SigNoz components to non-root paths for being compatible in rootless mode (non-root container user).

rybnico:
Yes, indeed, there are some things to change to make the helm chart and the images work without root out of the box.

I use the Signoz Helm chart with Kustomize and patch some fields to make it work. It would also work if you wrote custom deployments, but it can't work without a non-root executable `query-service' binary.

This is just a workaround that would save me (and possibly others - I've found older non-root images of the query service in the wild) from having to build custom images until the full non-root support is ready.

srikanthccv (approved):

PR #3350, opened 2023-08-16
fix: log details action not it

yeshev:


palashgdev:
@yeshev can you please check why the type of the column is null after selecting for not in

palashgdev (approved):

Issue #2255, opened 2023-02-14
Turn more easy create Alerts for default charts on three panes


paulohmorais:
## Is your feature request related to a problem?

Not easy to create alerts from the most important charts that we have by default on services.

## Describe the solution you'd like

On the default services page, we have these charts on three panes:
Overview metrics: Latency, Rate (ops/s), Error Percentage.
Database Calls: Database Calls RPS, Database Calls Avg Duration
External Calls: External Call Error PErcentage, External Call duration. External Call RPS (By Address), External Call duration(by Address)

But if I want to create alerts for one or more, I need to create alerts by my hand. But on the new relic, we just give a Threshold for these default charts and the alert are created. In my environment just alerts for these defaults charts we have is enough.

For example:
5% of an error on Error Percentage on Overview Metrics of service.
Rate(ops/s) is bellow to 5op/s on Rate(ops/s) on Overview Metrics of service.
Database Calls RPS below to 5op/s on Database Calls RPS on Database Calls of Service.
External Call Error Percentage more than 5% on External Call error percentage.

All default charts we have on three panes from default on signoz are amazing and all we need. But creating alerts from then is not so easy, I need to create them on my own. So maybe a button on the chart to create an alert will turn very useful and easy for signoz users.

Thanks.

#### *Thank you* for your feature request – we love each and every one!

PR #3358, opened 2023-08-17
feat: create menu items is added in the service application widgets

palashgdev:
added the option of creating alerts in the metrics application widgets

https://github.com/SigNoz/signoz/assets/88981777/12304b28-5afe-4402-a2e4-786ff0dc0b6b

part of https://github.com/SigNoz/signoz/issues/2255


YounixM (dismissed):

Rajat-Dabade (dismissed):
LGTM 🚀, unable to test it as the backend is not merged on staging.

srikanthccv:
@palashgdev, there is one issue that should be fixed. When there is more than one enabled query/formula, we need to ask the user to select one of them to create an alert. For instance, the latency chart has p50, p95, and p99; the user should be asked to select one and then redirected to the alert page with the selected data. This is important because an alert can't be created for pXX, and pYY at the same time.

srikanthccv:
This is not shippable @palashgdev. We put it on hold because of a top-level operations issue. We only need to allow the creation of alerts from dashboards for now.

PR #3362, opened 2023-08-17
feat: regex operator is added in the string type for all data source

palashgdev:


srikanthccv:
This is working now

srikanthccv (approved):

makeavish:
@srikanthccv @palashgdev Regex is not supported on trace backend yet.
Also we are not showing actual error here:
Screenshot 2023-08-18 at 18 50 11

palashgdev:
> @srikanthccv @palashgdev Regex is not supported on trace backend yet. Also we are not showing actual error here: Screenshot 2023-08-18 at 18 50 11

do we have a design for same ?

Issue #3361, opened 2023-08-17
FE: Optimise Dashboard Performance 


palashgdev:
Jittery Experience when we load a large number of data points and the browser becomes unresponsive for some amount of time

recht:
"Some amount of time" can be forever, which means that it's pretty easy to get into a state where too many points are returned, and then the whole UI freezes until the browser gives up. If you managed to save a dashboard in that state then there's no way to recover that since editing it will cause the same to happen :-(

srikanthccv:
Please web profile and find the underlying issues first. I don't think we have enough clarity on the bottleneck pieces (or if we do, we should target them mainly). Without that, any improvements will not make any difference. I was trying to see something in the existing dashboard and the page froze. This isn't raw data or anything like that, but a request rate by operation. The current dashboards is unusable.

AshwinBhatkal:
This is fixed now

PR #3371, opened 2023-08-17
fix: chart data is more optimised

palashgdev:
part of https://github.com/SigNoz/signoz/issues/3361


frontend/src/lib/getChartData.ts
```diff
@@ -10,79 +10,73 @@ const getChartData = ({
...
 	const uniqueTimeLabels = new Set<number>();
-	queryData.forEach((data) => {
-		data.queryData.forEach((query) => {
-			query.values.forEach((value) => {
-				uniqueTimeLabels.add(value[0]);
-			});
-		});
-	});
-	const labels = Array.from(uniqueTimeLabels).sort((a, b) => a - b);
+	const allLabels: string[] = [];
```
CKanishka, line 13: I think it will be helpful for anyone referring to this function in future, if a sample can be added on the transformation that this function is performimng, like a sample input and output.

palashgdev:
dropping the chart js in favour of uplot

Issue #3374, opened 2023-08-17
Enable automatic provisioning of users using SSO as login mechanism


ankitnayan:
After enabling SSO at SigNoz, the Org needs to create invites for all the users and send them manually to enable them to login to SigNoz.

We should enable better ways to reduce this friction. An option is to sync the users from the IDP platform like Okta to SigNoz automatically.

PR #3394, opened 2023-08-21
feat: Just-in-time provisioning of SSO users 

ankitnayan:
Fixes: https://github.com/SigNoz/signoz/issues/3374

This PR involves:
- don't send `isUser: false` field when the user doesn't exist. Rather keep it true for SSO users
- removing the check if the user exists in the db in `PrepareSamlRequest` method
- create a user if it does not exist


ee/query-service/dao/sqlite/auth.go
```diff
@@ -24,7 +69,20 @@ func (m *modelDao) PrepareSsoRedirect(ctx context.Context, redirectUri, email st
...
 
-	tokenStore, err := baseauth.GenerateJWTForUser(&userPayload.User)
+	user := &basemodel.User{}
+
+	if userPayload == nil {
+		newUser, apiErr := m.createUserForSAMLRequest(ctx, email)
+		user = newUser
+		if apiErr != nil {
+			zap.S().Errorf(" failed to get user with email received from auth provider", apierr.Error())
+			return "", model.BadRequestStr("invalid user email received from the auth provider")
```
srikanthccv, line 79:
```suggestion
			zap.S().Errorf("failed to create user with email received from auth provider: %v", apierr.Error())
			return "", apiErr
```

ee/query-service/dao/sqlite/auth.go
```diff
@@ -67,15 +125,15 @@ func (m *modelDao) PrecheckLogin(ctx context.Context, email, sourceUrl string) (
...
+	// if baseApiErr != nil {
+	// 	return resp, baseApiErr
+	// }
 
-	if userPayload == nil {
-		resp.IsUser = false
-	}
+	// if userPayload == nil {
+	// 	resp.IsUser = false
+	// }
```
srikanthccv, line 136: Desription says

> don't send isUser: false field when the user doesn't exist. Rather keep it true for SSO users

You are removing it entirely. Did you mean to do it only for SSO users? Otherwise, This will make the non-SSO users' experience confusing.  With this check removed, users will no longer see the _get invite link from admin_ message.

ankitnayan: right. It should not be removed entirely. `IsUser: true` should be set exceptionally when `SSO` feature is enabled. Sending a fix.

ankitnayan: I have tested the fix in the new commit for non-SSO admin register, non-SSO invite register and SSO users

srikanthccv (dismissed):

srikanthccv (approved):

Issue #3372, opened 2023-08-17
[authn] support IDP initiated session


ankitnayan:
Most of the users in an Org use the Okta browser plugin to quickly get to their work apps.

Currently, a user won't be able to login to SigNoz from Okta but the user will have to login via SigNoz UI.

pavelpp-topia:
hoping this will get fixed soon, as it stopped working after the recent SSO changes.

therealpandey:
This is supported starting release `v0.104.0`.

PR #3396, opened 2023-08-21
fix: signoz login from IDP

ankitnayan:
Fixes: https://github.com/SigNoz/signoz/issues/3372
This PR is stacked on changes of https://github.com/SigNoz/signoz/pull/3394.

The current approach used `domainId` being passed as part of `relayState` to fetch the Org details. `domainId` can be handled easily when a user is trying to login from SigNoz's login page. This does not work for login initiated from IDP as we cannot set `domainId` in the relay state of IDP config.

The PR enables an alternate way to fetch Org details using `domainName` passed in the `relayState` which can be configured at the IDP config. An example of `relayState` would be `https://testbed-in.signoz.io?domainName=signoz.io`

### DIFF
--- a/ee/query-service/dao/sqlite/domain.go
+++ b/ee/query-service/dao/sqlite/domain.go
@@ -4,8 +4,8 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
-	"net/url"
 	"fmt"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


ee/query-service/dao/sqlite/domain.go
```diff
@@ -28,29 +28,58 @@ type StoredDomain struct {
...
+		}
 	}
 
-	domainId, err := uuid.Parse(domainIdStr)
-	if err != nil {
-		zap.S().Errorf("failed to parse domain id from relay state", err)
-		return nil, fmt.Errorf("failed to parse response from IdP response")
+	if domainIdStr != "" {
+		domainId, _ := uuid.Parse(domainIdStr)
+		domain, _ = m.GetDomain(ctx, domainId)
```
srikanthccv, line 50: I am assuming using `domainId` on non-nil uuid.Parse errors is still safe (no panics etc..c), otherwise, let's handle it.

ee/query-service/dao/sqlite/domain.go
```diff
@@ -28,29 +28,58 @@ type StoredDomain struct {
...
+	}
+	if domain != nil {
+		return domain, nil
 	}
 
+	return nil, fmt.Errorf("failed to find domain received in IdP response")
+}
+
+// GetDomain returns org domain for a given domain id
+func (m *modelDao) GetDomainByName(ctx context.Context, name string) (*model.OrgDomain, basemodel.BaseApiError) {
```
srikanthccv, line 67:
```suggestion
// GetDomainByName returns org domain for a given domain name
func (m *modelDao) GetDomainByName(ctx context.Context, name string) (*model.OrgDomain, basemodel.BaseApiError) {
```

ee/query-service/dao/sqlite/domain.go
```diff
@@ -28,29 +28,58 @@ type StoredDomain struct {
...
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return nil, model.BadRequest(fmt.Errorf("invalid domain name"))
+		}
+		return nil, model.InternalError(err)
+	}
+
+	domain := &model.OrgDomain{Id: stored.Id, Name: stored.Name, OrgId: stored.OrgId}
+	if err := domain.LoadConfig(stored.Data); err != nil {
+		return domain, model.InternalError(err)
```
srikanthccv, line 81: Not sure what the `Data` part contains, does it make sense to return both domain and error at the same time? And the error is ignored by the caller `domain, _ = m.GetDomainByName(ctx, domainNameStr)`.

ankitnayan: I got this from `GetDomainByEmail` function already present. I checked `stored.Data`, it contains the SAML config stored i.e., ACS Url, EntityId and X.509 certificate. I think, it makes sense to send back the domain for further checks if there is error in loading the SAML config.

srikanthccv: The `SamlConfig` entries/fieds are used to prepare the request https://github.com/SigNoz/signoz/blob/ee6b290a0ca2300412427824f44487462794d1d8/ee/query-service/model/domain.go#L114-L136

Since the config is the main part of request, I think we shouldn't have the both domain and error. Or if it is fine to send them both we need to have the caller to act based on both the returned values. In this change the caller ignores the error because it checks if domain != nil. This may not be an issue in the happy flow when the jsondata is correct but we should avoid preparing invalid request.

ee/query-service/dao/sqlite/domain.go
ankitnayan: IDK why I removed them. Added them back.

ee/query-service/dao/sqlite/domain.go
srikanthccv: Sure

ankitnayan: I am not sure what you are referring in the above context. What I checked is that if the caller function immediately returns when an error is observed without making any use of `domain` then it's useless to return. I checked and none of the caller functions are using `domain` returned but just return when an error is observed.

ee/query-service/dao/sqlite/domain.go
```diff
@@ -28,29 +28,70 @@ type StoredDomain struct {
...
+			return nil, fmt.Errorf("invalid credentials")
+		}
+	}
+
+	if domainNameStr != "" {
+
+		domainFromDB, err := m.GetDomainByName(ctx, domainNameStr)
+		domain = domainFromDB
+		if (err != nil) || domain == nil {
+			zap.S().Errorf("failed to find domain from domainName received in IdP response", err.Error())
```
ankitnayan, line 67: @srikanthccv Eg here, `domain` is not used if the `err != nil`

ee/query-service/dao/sqlite/domain.go
ankitnayan: We can discuss more on https://github.com/SigNoz/signoz/pull/3421

srikanthccv: > I checked and none of the caller functions are using domain returned but just return when an error is observed.


This was not the case when PR was initially opened. This discussion doesn't make sense after this commit https://github.com/SigNoz/signoz/pull/3396/commits/502540921cfff9bcbbb34d42d53157fc4ed1db2d.

PR #3417, opened 2023-08-23
fix: use query-service API to fetch triggered alerts

srikanthccv:
Fixes https://github.com/SigNoz/engineering-pod/issues/528

This change allows the frontend to fetch triggered alerts from the query service, eliminating the need for alertmanager accessibility from the frontend. The `/api/v1/alerts` API of the query-service uses the `/alerts` of alertmanager to fetch the alerts with given criteria.


ankitnayan (approved):

prashant-shahi (approved):
LGTM :shipit: 

_Screenshot:_

Screenshot 1

Issue #2869, opened 2023-06-08
Flickering of screen when a new chart is created


ankit01-oss:
Please try the steps to notice the flicker. Sometimes, it's not getting captured on the video. It happens when you create a new chart in `Dashboards`.

https://www.loom.com/share/fac497c345f8418ca7141105e6441173

makeavish:
This is fixed in latest version

PR #3427, opened 2023-08-24
fix: flicker issue after adding panel in dashboard

Rajat-Dabade:
close https://github.com/SigNoz/signoz/issues/2869


palashgdev:
reducer would be eventually removed closing for same

PR #3434, opened 2023-08-25
fix: exists, nexists support for top level logs columns

### LINKED ISSUE #2592: [Logs] Handle top level columns in exists and notexists
As of now the struct is only 
```
	v3.FilterOperatorExists:          "has(%s_%s_key, '%s')",
	v3.FilterOperatorNotExists:       "not has(%s_%s_key, '%s')",
```

nityanandagohain:
Fixes #2592 
Fixes https://github.com/SigNoz/signoz/issues/3438

* This PR corrects how exists, nexists operator is handeled.
* Add exists check to aggregate attribute

### DIFF
--- a/pkg/query-service/app/logs/v3/query_builder.go
+++ b/pkg/query-service/app/logs/v3/query_builder.go
@@ -164,6 +164,13 @@ func buildLogsTimeSeriesFilterQuery(fs *v3.FilterSet, groupBy []v3.AttributeKey)
 							}
 						}
 
+					} else if item.Key.IsColumn {
+						name := getClickhouseColumnName(item.Key)
+						val := true
+						if op == v3.FilterOperatorNotExists {
+							val = false
+						}
+						conditions = append(conditions, fmt.Sprintf("%s_exists=%v", name, val))
 					} else {
 						columnType := getClickhouseLogsColumnType(item.Key.Type)
 						columnDataType := getClickhouseLogsColumnDataType(item.Key.DataType)
--- a/pkg/query-service/app/logs/v3/query_builder.go
+++ b/pkg/query-service/app/logs/v3/query_builder.go
@@ -165,12 +165,11 @@ func buildLogsTimeSeriesFilterQuery(fs *v3.FilterSet, groupBy []v3.AttributeKey)
 						}
 
 					} else if item.Key.IsColumn {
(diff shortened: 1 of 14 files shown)


pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -129,17 +129,46 @@ func buildLogsTimeSeriesFilterQuery(fs *v3.FilterSet, groupBy []v3.AttributeKey)
...
+							} else {
+								// we just have two types, number and string
+								conditions = append(conditions, fmt.Sprintf("%s %s 0", columnName, top))
+							}
+						}
+
+					} else {
+						columnType := getClickhouseLogsColumnType(item.Key.Type)
+						columnDataType := getClickhouseLogsColumnDataType(item.Key.DataType)
+						conditions = append(conditions, fmt.Sprintf(logsOp, columnType, columnDataType, item.Key.Key))
```
srikanthccv, line 170: Can this be changed to use the `_exists` column? I am OK if you want to address it separately.

srikanthccv (approved):

pkg/query-service/app/logs/v3/query_builder.go
nityanandagohain: Right, let me add the change

nityanandagohain: Have added the change

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -123,23 +123,59 @@ func getSelectKeys(aggregatorOperator v3.AggregateOperator, groupBy []v3.Attribu
...
+	if item.Key.Type == v3.AttributeKeyTypeUnspecified {
+		top := "!="
+		if op == v3.FilterOperatorNotExists {
+			top = "="
+		}
+		if val, ok := constants.StaticFieldsLogsV3[item.Key.Key]; ok {
+			// skip for timestamp and id
+			if val.Key == "" {
+				return ""
+			}
```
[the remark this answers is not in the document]
nityanandagohain: No we don't have any case where timestart will show up in suggestion as it will break the start and end keys. Also don't expect it from the client.

srikanthccv (approved):

PR #3453, opened 2023-08-29
fix: live tail query formatting fix

nityanandagohain:
Fixes formatting issue which was breaking contains


pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -484,6 +481,7 @@ func PrepareLogsQuery(start, end int64, queryType v3.QueryType, panelType v3.Pan
 		if err != nil {
 			return "", err
 		}
+		fmt.Println("asass1  : " + query)
```
srikanthccv, line 484:
```suggestion
```

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -4694,13 +4694,12 @@ func (r *ClickHouseReader) LiveTailLogsV3(ctx context.Context, query string, tim
...
 			// get the new 100 logs as anything more older won't make sense
-			tmpQuery := fmt.Sprintf("timestamp >='%d'", timestampStart)
+			tmpQuery := fmt.Sprintf("AND timestamp >='%d'", timestampStart)
 			if idStart != "" {
 				tmpQuery = fmt.Sprintf("%s AND id > '%s'", tmpQuery, idStart)
 			}
-			tmpQuery = fmt.Sprintf(query, tmpQuery)
 			// the reason we are doing desc is that we need the latest logs first
-			tmpQuery = fmt.Sprintf("%s order by timestamp desc, id desc limit 100", tmpQuery)
+			tmpQuery = query + tmpQuery + " order by timestamp desc, id desc limit 100"
```
srikanthccv, line 4702: Does this produce a valid query if the filter is empty? The `buildLogsLiveTailQuery` returns query with empty where, and the `tmpQuery` starts with `AND`. Can the query be like `... WHERE AND timestamp >= ...`

srikanthccv (approved):

PR #3486, opened 2023-09-06
chore(FE): reduce main.js chunk size

### LINKED ISSUE #3485: [perf] FE: Reduce main.js chunk size
## Bug description

The main.js chunk that gets loaded is very large now ~4 m.b. It needs to be reduced down.

image

CKanishka:
Fixes #3485 

| New State | Prev State |
|----------|----------|
| image | image |

Reduction of > **50%**

### DIFF
--- a/frontend/src/AppRoutes/routes.ts
+++ b/frontend/src/AppRoutes/routes.ts
@@ -1,5 +1,4 @@
 import ROUTES from 'constants/routes';
-import DashboardWidget from 'pages/DashboardWidget';
 import { RouteProps } from 'react-router-dom';
 
 import {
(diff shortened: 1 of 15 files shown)


Rajat-Dabade:
@CKanishka, this is really good. As I can see in the repo there are roughly 14 files that use the `import *`, it would be great if we change that as well which will bring down the main.js size as a part of this PR.

Screenshot 2023-09-06 at 1 31 12 PM

CKanishka:
> @CKanishka, this is really good. As I can see in the repo there are roughly 14 files that use the `import *`, it would be great if we change that as well which will bring down the main.js size as a part of this PR.
> 
> Screenshot 2023-09-06 at 1 31 12 PM
> Please ignore some of the files, where you have already made the changes.

Sure I will update, wherever its feasible

palashgdev:
@CKanishka can you check the pipeline

palashgdev (approved):

PR #3500, opened 2023-09-06
fix: avoid pipeline builder panic when first op is disabled

raj-k-singh:


pkg/query-service/app/logparsingpipeline/pipelineBuilder.go
```diff
@@ -58,7 +58,7 @@ func getOperators(ops []model.PipelineOperator) []model.PipelineOperator {
 	filteredOp := []model.PipelineOperator{}
 	for i, operator := range ops {
 		if operator.Enabled {
-			if i > 0 {
+			if len(filteredOp) > 0 {
```
srikanthccv, line 61: It's hard to follow this logic. Can you help me understand what `getOperators` is supposed to do?

nityanandagohain: Yeah, it's just filtering out the disabled operators and stitching them correctly.

nityanandagohain (approved):

pkg/query-service/app/logparsingpipeline/pipelineBuilder.go
srikanthccv: >and stitching them correctly

Can you elaborate this part?

nityanandagohain: eg:-
```
id: op1
enabled: true
output: op2

id: op2
enabled: false
output: op3

id: op3
enabled: true
```


the final config should be
```
id: op1
enabled: true
output: op3

id: op3
enabled: true
```

So here we had to updated the values of `ouput`

srikanthccv (approved):

PR #3505, opened 2023-09-07
chore: allow {{.top_level_operations}} variable in alerts and dashboard

srikanthccv:
Part of https://github.com/SigNoz/signoz/issues/2255

### DIFF
--- a/pkg/query-service/app/http_handler.go
+++ b/pkg/query-service/app/http_handler.go
@@ -3011,7 +3011,12 @@ func (aH *APIHandler) queryRangeV3(ctx context.Context, queryRangeParams *v3.Que
 }
 
 func (aH *APIHandler) QueryRangeV3(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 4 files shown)


pkg/query-service/app/http_handler.go
```diff
@@ -3011,7 +3011,12 @@ func (aH *APIHandler) queryRangeV3(ctx context.Context, queryRangeParams *v3.Que
 }
 
 func (aH *APIHandler) QueryRangeV3(w http.ResponseWriter, r *http.Request) {
-	queryRangeParams, apiErrorObj := ParseQueryRangeParams(r)
+	topLevelOps, apiErr := aH.reader.GetTopLevelOperations(r.Context(), aH.skipConfig)
+	if apiErr != nil {
+		RespondError(w, apiErr, nil)
```
palashgdev, line 3016: do we need to add return here IMO

PR #3516, opened 2023-09-08
fix: login/precheck api in non-ee variant

### LINKED ISSUE #3367: FE: Fix login after building without `ee` folder
## Bug description

After removing the `ee` directory and building Signoz, we are able to log in and explore the UI. 

But after logout, the login doesn't work because the frontend sends an API call to `/api/v1/loginPrecheck` which is part of the [ee query-service](https://github.com/SigNoz/signoz/blob/5ac105beca19e6b43887388386950ef669ad52a7/ee/query-service/app/api/api.go#L105).

Screenshot 2023-08-17 at 4 31 45 PM

## Expected behavior

Login should work without any issues after compiling without `ee` folder

## How to reproduce

1. Import `model` from `pkg` library in the [dynamic_config.go](https://github.com/SigNoz/signoz/blob/5ac105beca19e6b43887388386950ef669ad52a7/pkg/query-service/integrations/signozio/dynamic_config.go#L9) file
2. Delete the `ee` folder
3. Build the frontend and query-service
4. Start frontend service and also start query-service by passing `--skip-top-level-ops` argument
5. Login into Signoz and create a user. Logout and then try logging in again.

## Version information 
* **Signoz version**: v0.24.0
* **Browser version**: Chrome 115.0.5790.114
* **Your OS and version**:  macOS Ventura & 13.2

vishalsharma1800:
Fixes #3367


pkg/query-service/app/http_handler.go
```diff
@@ -1864,6 +1865,19 @@ func (aH *APIHandler) registerUser(w http.ResponseWriter, r *http.Request) {
...
 }
 
+func (aH *APIHandler) precheckLogin(w http.ResponseWriter, r *http.Request) {
+
+	email := r.URL.Query().Get("email")
+	sourceUrl := r.URL.Query().Get("ref")
+
+	resp, apierr := aH.appDao.PrecheckLogin(context.Background(), email, sourceUrl)
+	if apierr != nil {
+		RespondError(w, apierr, resp)
```
srikanthccv, line 1875: You need to add the return statement here.

srikanthccv (commented):
Overall LGTM, I will test and review this again.

pkg/query-service/app/http_handler.go
vishalsharma1800: Added the return statement

srikanthccv (approved):

PR #3544, opened 2023-09-13
feat: json filter bool support

### LINKED ISSUE #3529: LOGS [BE] Support for querying JSON in body
Support for querying JSON data without parsing.

nityanandagohain:
Fixes #3529 

So
`JSONExtract(JSON_VALUE(body, '$.bool'), 'Bool')` doesn't work, so we are treating it as a string.

The other solution could have been to use `JSONExtractBool` , but keeping the same structure.

### DIFF
--- a/frontend/src/AppRoutes/index.tsx
+++ b/frontend/src/AppRoutes/index.tsx
@@ -32,6 +32,8 @@ function App(): JSX.Element {
 
 	const dispatch = useDispatch<Dispatch<AppActions>>();
 
+	const { hostname } = window.location;
+
 	const featureResponse = useGetFeatureFlag((allFlags) => {
 		const isOnboardingEnabled =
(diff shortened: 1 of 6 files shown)


srikanthccv:
>JSONExtract(JSON_VALUE(body, '$.bool'), 'Bool') doesn't work, so we are treating it as a string.


```
SELECT JSONExtract(JSON_VALUE('{"bool": true}', '$.bool'), 'Bool')

Query id: 1012fc4f-97f0-4262-9c26-1a7e903d965a

┌─JSONExtract(JSON_VALUE('{"bool": true}', '$.bool'), 'Bool')─┐
│ true                                                        │
└─────────────────────────────────────────────────────────────┘
```

It does seem to work. What's the issue?

nityanandagohain:
I think I was doing something wrong, it seems to be working. I have updated the changes.

Issue #3299, opened 2023-08-09
Logs filter U/X: Improve how attributes are shown in logs filter UI.


nityanandagohain:
* When an attribute is selected we show it as `<type>_<name>` where  `<type>` can be `tag/resource`   and `<name>` is the name of the attrbute.
![image](https://github.com/SigNoz/signoz/assets/26831659/44483617-c46b-4130-89e8-90e3e3f0716a)

* Once the attribute is selected it becomes just `<name>`
![image](https://github.com/SigNoz/signoz/assets/26831659/e66174d1-48e6-47f2-a45d-39fd945cee13)

* We need a better way to convery the user what is the `type` and `dataType` of the tag instead of the aboe

pranay01:
`dataType` could be `int64` or `string`

pranay01:
Figma Design - https://www.figma.com/file/i2CsZFl27dhmER6gnx08wj/Integ.-QueryBuilder%2B3SigNalCorrelation?type=design&node-id=4211%3A20265&mode=design&t=JlXlGD1o8XhNN9OL-1

PR #3555, opened 2023-09-14
[Refactor]: Added tag and datatype in options for query builder

Rajat-Dabade:
Close https://github.com/SigNoz/signoz/issues/3299
Loom: https://www.loom.com/share/aa61fb1cd97c41eba18b18e095f67588?sid=59117602-790e-47d6-aea6-cbed3fb5d5a5


nityanandagohain (approved):
LGTM, but check with @pranay01  if we can shift the details to the left. I think that will be better.

Rajat-Dabade:
@pranay01 this is the possible solution where to show the type and datatype in the option for the select

Option 1: 
Screenshot 2023-09-15 at 2 09 54 PM

Option 2: 

Screenshot 2023-09-15 at 2 08 18 PM

Option 3: 
@suggested by @YounixM 
Screenshot 2023-09-15 at 2 11 36 PM

palashgdev:
✅ 3

pranay01:
@Rajat-Dabade Lets go with Option 1. I had a discussion with Kaushik earlier on this, and from a UX perspective Option 1 made more sense - option 3 confused user with details on data type when his first objective is to select the right variable

srikanthccv:
Looking at the code, this is done for the where clause only. There are other filters where we created the `tag_`/`resource_` prefixes. Shouldn't we do the same for aggregate attributes, group by and wherever this was done?

ankitnayan:
> Looking at the code, this is done for the where clause only. There are other filters where we created the `tag_`/`resource_` prefixes. Shouldn't we do the same for aggregate attributes, group by and wherever this was done?

this is already done in the logs detail view. @nityanandagohain what other places needs to be taken care of?

nityanandagohain:
Yeah it's done in log detail view, I think the groupby and aggreagte attribute field is only left.

nityanandagohain:
Bumping this, can the changes be aded to groupBY and aggreagte attribute ?

srikanthccv:
@Rajat-Dabade please add the same to the group by clause everywhere and aggregate attributes for logs and traces only, because the metric name doesn't have any associated type and the data type is always a number.

srikanthccv:
@nityanandagohain PTAL

nityanandagohain (changes requested):
After selecting, the `tag` prefix should be removed

image

Rajat-Dabade:
> After selecting, the `tag` prefix should be removed
> 
> image

@nityanandagohain Done. Please review it once.

nityanandagohain (approved):
LGTM! But there is another issue where removing field from groupby is slow sometimes, please create a separate for that as it is an existing issue in staging as well.

srikanthccv:
@YounixM @palashgdev please review this

palashgdev (approved):

frontend/src/container/QueryBuilder/filters/QueryBuilderSearch/style.ts
```diff
@@ -15,3 +15,27 @@ export const TypographyText = styled(Typography.Text)<{
 export const StyledCheckOutlined = styled(CheckOutlined)`
 	float: right;
 `;
+
+export const SelectOptionContainer = styled.div`
```
YounixM, line 19: We could have created a component style file and added new styles there. When working on any piece of code, please try and improve it according to the latest standards. This will reduce rework.

frontend/src/container/QueryBuilder/filters/QueryBuilderSearch/style.ts
```diff
@@ -15,3 +15,27 @@ export const TypographyText = styled(Typography.Text)<{
...
+	display: flex;
+	justify-content: space-between;
+	align-items: center;
+`;
+
+export const TagContainer = styled(Tag)`
+	&&& {
+		border-radius: 0.25rem;
```
YounixM, line 27: Did we get these styles from design files?

frontend/src/container/QueryBuilder/filters/QueryBuilderSearch/OptionRenderer.tsx
```diff
@@ -0,0 +1,45 @@
...
+	dataType,
+}: OptionRendererProps): JSX.Element {
+	const optionType = getOptionType(label);
+
+	return (
+		<span>
+			{optionType ? (
+				<SelectOptionContainer>
+					<div>{value}</div>
+					<div>
```
YounixM, line 21: Can we fragment.

Rajat-Dabade:
@YounixM Will take up these changes in subsequent PR.

Issue #3506, opened 2023-09-07
[Logs] Support for using QueryBuilder filterset as log pipeline filter


raj-k-singh:
Right now pipeline filters are writter in expr language, which is new and unfamiliar to users.
QueryBuilder filters will be familiar and consistent with rest of the product

PR #3560, opened 2023-09-15
Add support in query service for querybuilder filterset based log pipelines

raj-k-singh:
Contributes to https://github.com/SigNoz/signoz/issues/3506

### DIFF
--- a/pkg/query-service/model/v3/v3.go
+++ b/pkg/query-service/model/v3/v3.go
@@ -1,6 +1,7 @@
 package v3
 
 import (
+	"database/sql/driver"
 	"encoding/json"
 	"fmt"
 	"sort"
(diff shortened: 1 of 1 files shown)


pkg/query-service/app/logparsingpipeline/pipelineBuilder.go
```diff
@@ -47,13 +62,28 @@ func PreparePipelineProcessor(pipelines []model.Pipeline) (map[string]interface{
...
 	}
 	return processors, names, nil
 }
 
+func PipelineFilterExpr(serializedFilter string) (string, error) {
+	var filterset v3.FilterSet
+	err := json.Unmarshal([]byte(serializedFilter), &filterset)
+	if err != nil {
+		return "", errors.Wrap(err, "could not unmarshal pipeline filterset json")
+	}
```
srikanthccv, line 77: Is there a reason to send the JSON String and unmarshal here rather than sending the filter data structure in the payload directly?

srikanthccv (approved):
Thanks, that was a neat change.

Issue #3389, opened 2023-08-19
Limit is missing in metrics query builder


srikanthccv:
## Bug description

There is no Limit textbox for the time series chart for metrics. It makes it hard to see the top K items in the panel.

ankitnayan:
Do we have a provision to apply limit on Formula? I think we can keep default limit for formula from the backend

PR #3566, opened 2023-09-15
feat: limit & order by is enabled in the metrics query builder

palashgdev:
Close https://github.com/SigNoz/signoz/issues/3389


srikanthccv:
@palashgdev we need order by also.

palashgdev:
> @palashgdev we need order by also.
but the issue says limit only ... but let me update orderBy also

palashgdev:
@srikanthccv updated

srikanthccv:
I tested this and it's working for time series. I need to update some backend logic for table panel type.

srikanthccv (approved):

PR #3605, opened 2023-09-21
[Fix]: copy to clipboard without quotes

Rajat-Dabade:
Fix the copy to the clipboard without quotes.


palashgdev:
can we handle this one inside the HOC ?

makeavish (approved):

Rajat-Dabade:
@palashgdev In certain scenarios, there may arise a need to include double quotation marks when copying content. Handling this within a Higher-Order Component (HOC) might limit our ability to achieve this requirement.

palashgdev (approved):

PR #3613, opened 2023-09-23
fix: alert eval for "="/"!=" combination with "at least once"/"all th…

### LINKED ISSUE #3578: "all the time" & "at least once" with "equal to" & "not equal to" triggers do not work as expected
## Bug description

The alert condition when the comparison operator `=`/`!=` does not seem to work as expected. 

https://github.com/SigNoz/signoz/blob/416a058eab4cf2c6e282319491b9a3f139b35d85/pkg/query-service/rules/thresholdRule.go#L576-L602

srikanthccv:
…e times"

Fixes #3578 

Assuming that prepared queries are correct, the tests mock the required ClickHouse conn in `runChQuery` and `rows.Next` returns the rows as if this was an actual database connection and verifies the number of alerts from evaluation.

https://github.com/SigNoz/signoz/blob/043e5ca880a9a16e8461ab12f00bcafba98263c6/pkg/query-service/rules/thresholdRule.go#L420-L627


srikanthccv:
Please review. I would like to get this merged and make it part of release.

prashant-shahi (approved):

PR #3625, opened 2023-09-26
Add support for freehand json query

### LINKED ISSUE #3592: LOGS: Support for free hand query in json
Add support for free hand query in json i.e json query by writing it in the filter box

nityanandagohain:
Fixes #3592


pkg/query-service/app/logs/v3/enrich_query.go
```diff
@@ -149,3 +151,69 @@ func enrichFieldWithMetadata(field v3.AttributeKey, fields map[string]v3.Attribu
...
+				value = valueStr
+			} else {
+				valueType = "float64"
+			}
+		} else {
+			valueType = "int64"
+		}
+	} else {
+		valueType = "bool"
+	}
```
srikanthccv, line 216: The deep nested conditions was hard to read. Might be better to refactor something like this.
```golang
if parsedValue, err := strconv.ParseBool(valueStr); err == nil {
    value = parsedValue
    valueType = "bool"
} else if parsedValue, err := strconv.ParseInt(valueStr, 10, 64); err == nil {
    value = parsedValue
    valueType = "int64"
} else if parsedValue, err := strconv.ParseFloat(valueStr, 64); err == nil {
    value = parsedValue
    valueType = "float64"
} else {
    value = valueStr
    valueType = "string"
}
```

pkg/query-service/app/logs/v3/enrich_query_test.go
```diff
@@ -289,3 +289,154 @@ func TestEnrichParams(t *testing.T) {
...
+	{
+		Name: "int64",
+		Filter: v3.FilterItem{
+			Key: v3.AttributeKey{
+				Key:      "body.intx",
+				DataType: v3.AttributeKeyDataTypeUnspecified,
+				Type:     v3.AttributeKeyTypeUnspecified,
+			},
+			Operator: "=",
+			Value:    10,
```
srikanthccv, line 329: Most values you get from the frontend will have string type.

```suggestion
			Value:    "10",
```

pkg/query-service/app/logs/v3/enrich_query.go
```diff
@@ -149,3 +151,69 @@ func enrichFieldWithMetadata(field v3.AttributeKey, fields map[string]v3.Attribu
...
+	case bool:
+		valueType = "bool"
+	}
+
+	// check if it is array
+	if strings.HasSuffix(filter.Key.Key, "[*]") {
+		valueType = fmt.Sprintf("array(%s)", valueType)
+	}
+
+	filter.Key.DataType = v3.AttributeKeyDataType(valueType)
```
srikanthccv, line 179: I believe only handwritten filters wouldn't have a data type, but the ones created from UI already have it. Does it affect anything if you always prefer to infer logic regardless of the type sent by the frontend?

pkg/query-service/app/logs/v3/enrich_query_test.go
nityanandagohain: yeah test for that is present in  `TestParseStrValue` . These tests were added before I considered the string to value conversion thing. So kept them.

pkg/query-service/app/logs/v3/enrich_query.go
nityanandagohain: So if dataType of a field is not present then it will be enriched from clickhouse. 

The above logic will only run for keys having prefix `body.` and it will only happen for json query so this won't affect anything.

pkg/query-service/app/logs/v3/enrich_query.go
nityanandagohain: Done

srikanthccv (approved):

Issue #3666, opened 2023-10-04
Save and cancel button overflow in view mode 


Calm-Rock:
## Bug description

The save and cancel buttons in view mode overflow when there is a scrollbar in the labels section
![save-cancel-overflow](https://github.com/SigNoz/signoz/assets/31571545/3fe0ce63-1e53-402b-994d-333cf8d2c1a0)

## How to reproduce

1. Check any dashboard chart in demo site with >4 labels

## Version information 
* **Signoz version**: Cloud
* **Browser version**:
* **Your OS and version**: MacOs; Ventura
* **Your CPU Architecture**(ARM/Intel): ARM

PR #3671, opened 2023-10-05
[Refactor]: graph manager to scss and fix the height issue

Rajat-Dabade:
Close https://github.com/SigNoz/signoz/issues/3666


frontend/src/container/GridGraphLayout/Graph/FullView/GraphManager.styles.scss
```diff
@@ -0,0 +1,21 @@
+.filter-save-container {
```
palashgdev, line 1: I am not sure but these classes would be available in global can we wrap them under one class name so it does not collide with global CSS or mess with same class name

Rajat-Dabade: Sure.

palashgdev (approved):

palashgdev (approved):

PR #3680, opened 2023-10-06
feat: json filter -> attributes/materialized column and dot support

### LINKED ISSUE #3553: [LOGS] Use materialized column if present for JSON search.
If user uses json key for search check if an attribute/materialized column is already present and then query based on that.

nityanandagohain:
Fixes #3553


srikanthccv:
The general rule of thumb (I believe) is that the existing tests are invalid the moment your data characteristics change. They may still be helpful, but they don't guarantee or assert your new use cases work. It would be great if you could add new tests with dot support for the corresponding old ones.

pranay01:
relevant issue wrt dot support - https://github.com/ClickHouse/ClickHouse/issues/55452#issuecomment-1755688552

ankitnayan:
> relevant issue wrt dot support - [ClickHouse/ClickHouse#55452 (comment)](https://github.com/ClickHouse/ClickHouse/issues/55452#issuecomment-1755688552)

yes Pranay, after this we are not planning to move ahead with DOT naming conventions at SigNoz

nityanandagohain:
we are not going ahead with `.`

Issue #3509, opened 2023-09-07
[Logs] "Test Pipeline" button and pipeline preview experience in UI


raj-k-singh:
See details in [PRD](https://www.notion.so/signoz/Pipeline-Preview-PRD-07186e602e014628baeaa047e8f8e2d8#f007cf0ca67c477c90ad74f0e9095fb2)

By default, preview on most recent logs that match the pipeline filter.
Also allow choosing specific time periods (not too large) to look for sample logs that match.

PR #3694, opened 2023-10-09
QS: logparsingpipeline previews

raj-k-singh:
Contributes to https://github.com/SigNoz/signoz/issues/3509

### DIFF
--- a/pkg/query-service/app/logparsingpipeline/preview.go
+++ b/pkg/query-service/app/logparsingpipeline/preview.go
@@ -21,22 +21,24 @@ func SimulatePipelinesProcessing(
 	ctx context.Context,
 	pipelines []Pipeline,
 	logs []model.SignozLog,
-) ([]model.SignozLog, *model.ApiError) {
+) (
+	[]model.SignozLog, *model.ApiError,
+) {
+
 	if len(pipelines) < 1 {
 		return logs, nil
 	}
 
 	// Collector simulation does not guarantee that logs will come
-	// out in the same order as that in the input.
+	// out in the same order as in the input.
 	//
-	// Add a temp attribute for maintaining same order of logs
-	// in simulation output as that in input
-	previewOrderAttribKey := "__signoz_preview_log_idx__"
+	// Add a temp attribute for sorting logs in simulation output
+	inputOrderAttribute := "__signoz_input_idx__"
 	for i := 0; i < len(logs); i++ {
 		if logs[i].Attributes_int64 == nil {
 			logs[i].Attributes_int64 = map[string]int64{}
 		}
-		logs[i].Attributes_int64[previewOrderAttribKey] = int64(i)
+		logs[i].Attributes_int64[inputOrderAttribute] = int64(i)
 	}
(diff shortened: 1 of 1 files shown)


pkg/query-service/app/logparsingpipeline/preview.go
```diff
@@ -0,0 +1,254 @@
...
+	"go.opentelemetry.io/collector/processor"
+	"go.signoz.io/signoz/pkg/query-service/collectorsimulator"
+	"go.signoz.io/signoz/pkg/query-service/model"
+	"gopkg.in/yaml.v3"
+)
+
+func SimulatePipelinesProcessing(
+	ctx context.Context,
+	pipelines []Pipeline,
+	logs []model.SignozLog,
```
srikanthccv, line 23: Will the input be a structured log or a bunch of log lines in an unstructured string form?

pkg/query-service/app/logparsingpipeline/preview.go
```diff
@@ -0,0 +1,254 @@
+package logparsingpipeline
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/logstransformprocessor"
```
srikanthccv, line 10: Is it just the logtransform processor we support?

pkg/query-service/app/logparsingpipeline/preview.go
```diff
@@ -0,0 +1,254 @@
...
+) {
+
+	if len(pipelines) < 1 {
+		return logs, nil
+	}
+
+	// Collector simulation does not guarantee that logs will come
+	// out in the same order as in the input.
+	//
+	// Add a temp attribute for sorting logs in simulation output
```
srikanthccv, line 35: The input and output logs would have timestamps, Could we use that instead? Or in the absence of that, we could also use the observed timestamp on record instead of this new input order attribute?

pkg/query-service/app/logparsingpipeline/preview.go
```diff
@@ -0,0 +1,254 @@
...
+			Config: procConfRaw,
+		})
+	}
+
+	return processorConfigs, nil
+}
+
+// plog doesn't contain an ID field.
+// SignozLog.ID is stored as a log attribute in plogs for processing
+// and gets hydrated back later.
```
srikanthccv, line 131: The `ID` is generated by us dynamically for each record. It looks like this will be missing? Does the frontend work without ID?

srikanthccv (commented):
The tests have a structured input logs and give the structured output log. Can we have some test that takes the plain string as input and gives the structured log as output? I think that is more often and first thing users do.

pkg/query-service/app/logparsingpipeline/preview.go
srikanthccv: I think this is fine. I had a different assumption. I thought the user would simply paste the logs from their application instead of UI using the logs from the query range.

srikanthccv (approved):
Thanks for answering my questions. I need to see the UI in action once to get good idea.

Issue #3682, opened 2023-10-06
FE: supporting update in operators.


palashgdev:
Where clause support for updating the operator

PR #3695, opened 2023-10-09
fix: where clause is updated

palashgdev:
where clause fixes for an operator like exist,nin,like,nlike
close https://github.com/SigNoz/signoz/issues/3682


nityanandagohain (changes requested):
Upper case is not working now

* If I type `container_Id CONTAINS` then conten it still shows me operators instead of suggesting values
image

nityanandagohain (commented):
Looks good to me.

@palashgdev  please create the issue where updating a old filter for has/nhas doesn't work

@srikanthccv can you also give this a look once ?

nityanandagohain (changes requested):
It's breaking in dashboards page @palashgdev 

https://github.com/SigNoz/signoz/assets/26831659/d6a9f344-95bd-4195-9c21-c610be2d0b7a

palashgdev:
@nityanandagohain https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded/50387233#50387233 we can ignore this error

nityanandagohain (dismissed):

srikanthccv:
superseded by https://github.com/SigNoz/signoz/pull/4379

PR #3678, opened 2023-10-05
feat: add getIngestionCred API

makeavish:

PR #3699, opened 2023-10-09
feat: ingestion key management

makeavish:
Add ingestion key management in query service.
Integration with otel auth is still pending.
Improved implementation of https://github.com/SigNoz/signoz/pull/3678


srikanthccv (approved):

pkg/query-service/app/http_handler.go
```diff
@@ -377,6 +377,8 @@ func (aH *APIHandler) RegisterRoutes(router *mux.Router, am *AuthMiddleware) {
 	router.HandleFunc("/api/v1/settings/ttl", am.ViewAccess(aH.getTTL)).Methods(http.MethodGet)
 	router.HandleFunc("/api/v1/settings/apdex", am.AdminAccess(aH.setApdexSettings)).Methods(http.MethodPost)
 	router.HandleFunc("/api/v1/settings/apdex", am.ViewAccess(aH.getApdexSettings)).Methods(http.MethodGet)
+	router.HandleFunc("/api/v1/settings/ingestion_key", am.AdminAccess(aH.insertIngestionKey)).Methods(http.MethodPost)
+	router.HandleFunc("/api/v1/settings/ingestion_key", am.ViewAccess(aH.getIngestionKeys)).Methods(http.MethodGet)
```
prashant-shahi, line 381: shouldn't even the `GET` call be only accessible to Admin level users?

makeavish: No we want even viewers to be able to send data which requires ingestion key.

prashant-shahi: got it

PR #3704, opened 2023-10-10
chore: add billing api resources

srikanthccv:


srikanthccv:
I would like to get this merged. Please take a look.

ee/query-service/app/api/license.go
```diff
@@ -38,3 +72,150 @@ func (ah *APIHandler) applyLicense(w http.ResponseWriter, r *http.Request) {
 
 	ah.Respond(w, license)
 }
+
+func (ah *APIHandler) checkout(w http.ResponseWriter, r *http.Request) {
+
+	type checkoutResponse struct {
```
nityanandagohain, line 78: can move this out ?

srikanthccv: Did you mean out of this function or out of this file to under `/model`? This is a response struct from subscriptions API so I just kept it here since it will not be used anywhere else.

nityanandagohain: Cool, we can move it if in future we use it somewhere. Apart from that things look good to me.

nityanandagohain (approved):

Issue #3724, opened 2023-10-12
LOGS [FE]: Render escape characters in detai page.


nityanandagohain:
Render the escape characters propely in detail page 
![image](https://github.com/SigNoz/signoz/assets/26831659/ffec7610-982b-4bb8-beae-e63a45269ad2)

PR #3726, opened 2023-10-12
[Refactor]: removed escape character from the string

Rajat-Dabade:
Close https://github.com/SigNoz/signoz/issues/3724


palashgdev:
these would be appearing in another view also?

Rajat-Dabade:
@palashgdev, I have checked with other views, it's not there, The reason is that here we are stringifying the value when rendering it in the table view

```javascript
const dataSource =
		flattenLogData !== null &&
		Object.keys(flattenLogData)
			.filter((field) => fieldSearchFilter(field, fieldSearchInput))
			.map((key) => ({
				key,
				field: key,
				value: removeEscapeCharacters(JSON.stringify(flattenLogData[key])),
			}));
```

and in other views, we are directly using the `flattenLogData[key]` without stringifying it.

palashgdev (approved):

PR #3735, opened 2023-10-12
chore: update query service telemetry

makeavish:
And add new distination for saas telemetry


ankitnayan:
Since we are removing `/api/v2/metrics/query_range` altogether rather than adding `/api/v3/query_range`, we would be losing analytics for those.

pkg/query-service/telemetry/telemetry.go
```diff
@@ -74,28 +75,32 @@ func (a *Telemetry) IsSampled() bool {
...
+				signozMetricsUsed = true
+			}
 		}
-	}
-	if compositeQuery.ClickHouseQueries != nil && len(compositeQuery.ClickHouseQueries) > 0 {
-		if !strings.Contains(compositeQuery.ClickHouseQueries["A"].Query, "signoz_") && len(compositeQuery.ClickHouseQueries["A"].Query) > 0 {
-			signozMetricsNotFound = true
+	} else if postData.CompositeQuery.QueryType == v3.QueryTypeClickHouseSQL {
+		for _, query := range postData.CompositeQuery.ClickHouseQueries {
+			if !strings.Contains(query.Query, "signoz_metrics") && len(query.Query) > 0 {
```
ankitnayan, line 98: Should it be `signoz_` instead of `signoz_metrics`?

ee/query-service/app/server.go
```diff
@@ -191,6 +193,7 @@ func NewServer(serverOptions *ServerOptions) (*Server, error) {
 	}
 
 	telemetry.GetInstance().SetReader(reader)
+	telemetry.GetInstance().SetSaasOperator(constants.SaasSegmentKey)
```
ankitnayan, line 196: Did you miss this at `pkg/query-service/app/server.go` ?

ankitnayan:
Since `IdentifyUser` is not called on `query-service` restarts, the `SetUserEmail` and `SetCompanyDomain` won't be called there. `SetCompanyDomain` was hence called separately at `InitializeOrgpreferences` at https://github.com/SigNoz/signoz/blob/cb155a11728984c49cfc0a33070bd3398ca7df59/pkg/query-service/dao/sqlite/connection.go#L145

So, restart of the query-service will cause no events to be sent any further due to check at https://github.com/SigNoz/signoz/pull/3735/files#diff-8a30fc5d1386f1d6463d9edd9b054e6396e8544640566ff7f3174d4058127d83R377

makeavish:
> Since we are removing `/api/v2/metrics/query_range` altogether rather than adding `/api/v3/query_range`, we would be losing analytics for those.

No we are not using `/api/v2/metrics/query_range` in SigNoz at all so we won't be losing any analytics.

ee/query-service/app/server.go
makeavish: No, `pkg/query-service/app/server.go` is only used in non-ee builds. We don't want to track non-ee builds so didn't add there.

pkg/query-service/telemetry/telemetry.go
makeavish: We are not using Clickhouse query in default dashboard so we can actually consider any clickhouse query which uses `signoz_metrics` table as query written by user.
The condition should be updated to `strings.Contains(query.Query, "signoz_metrics")` instead of `!strings.Contains....`

ankitnayan: We used to reject events caused by APM metrics which contain `signoz_` in metrics name. How is that handled here?

makeavish: That is handled here: https://github.com/SigNoz/signoz/blob/f36e16084c9acb8f22466585d513ad97d8ec06fc/pkg/query-service/telemetry/telemetry.go#L86

pkg/query-service/telemetry/telemetry.go
```diff
@@ -249,10 +256,20 @@ func getOutboundIP() string {
 }
 
 func (a *Telemetry) IdentifyUser(user *model.User) {
+	if user.Email == "admin@admin.com" || user.Email == "admin@signoz.cloud" {
+		return
+	}
 	a.SetCompanyDomain(user.Email)
+	a.SetUserEmail(user.Email)
```
ankitnayan, line 263: I don't think user events will be correct due to this. The funnels created have to be based on companyDomain and any user funnel would be broken IMO. What do you say?

ankitnayan: How can we sort the user metrics?

pkg/query-service/telemetry/telemetry.go
ankitnayan: shouldn't this check be added to ClickhouseBuilder too? Or you want to consider folks users creating CH queries for APM to be real metrics users. It might be okay to make that assumption.

makeavish: Yeah we are making that assumption as we don't have any examples to write clickhouse query on metrics in our docs and it's very unlikely a novice user will be able to perform this action.

ankitnayan:
This causes metrics datasource to change to active
https://github.com/SigNoz/signoz/issues/3753

pkg/query-service/telemetry/telemetry.go
makeavish: Yes, users will add variables only if data is already there. So we can consider user as active user of metrics.

pkg/query-service/telemetry/telemetry.go
makeavish: Still looking into this.

pkg/query-service/telemetry/telemetry.go
```diff
@@ -74,28 +75,32 @@ func (a *Telemetry) IsSampled() bool {
...
+				signozMetricsUsed = true
+			}
 		}
-	}
-	if compositeQuery.ClickHouseQueries != nil && len(compositeQuery.ClickHouseQueries) > 0 {
-		if !strings.Contains(compositeQuery.ClickHouseQueries["A"].Query, "signoz_") && len(compositeQuery.ClickHouseQueries["A"].Query) > 0 {
-			signozMetricsNotFound = true
+	} else if postData.CompositeQuery.QueryType == v3.QueryTypeClickHouseSQL {
+		for _, query := range postData.CompositeQuery.ClickHouseQueries {
+			if strings.Contains(query.Query, "signoz_metrics") && len(query.Query) > 0 {
```
ankitnayan, line 98: It would be a good idea to add traces and logs check here too as dashboards can have CH queries on logs and traces as datasource too

makeavish: For traces active user we consider number of services > 0 and for logs we consider if user has used any filter.
Primary objective of this PR is track active user.
Tracking CH query can be added later as it's unlikely that a user writes CH query but don't send traces data or don't use logs filter.

makeavish:
> trying to create an empty panel makes it `active` metrics which is incorrect. You should check if `metric_name` is present or not for it to be marked as active in metrics

Updated.

pkg/query-service/telemetry/telemetry.go
makeavish: Added group call to identify companies.
Now funnels can be created based on company domain.

pkg/query-service/telemetry/telemetry.go
makeavish: We should have different events to track clickhouse queries.

ankitnayan (approved):

Issue #3721, opened 2023-10-12
Show time range dropdown option in alert creation page 


srikanthccv:
## Bug description

It's unclear to users what the time range of the preview chart is on the alert creation page. The global time range dropdown is present on every page except alerts. This is confusing and unclear for users.

ankitnayan:
Just enabling the global time range dropdown should be fine? 
The `during the last` param in the alerts configuration can be treated independently?

srikanthccv:
Yes, we should enable the global time range and treat the query over the last X minutes independently.

PR #3751, opened 2023-10-16
refactor: global time range downdown for alert edit page

Rajat-Dabade:
Close https://github.com/SigNoz/signoz/issues/3721


srikanthccv:
The chart currently uses the time range selected in _... during the last ..._ part of the __Step 2 - Define Alert Conditions__. With this change, it should use the global time range.

srikanthccv:
@Rajat-Dabade, let me know if my last comment was not clear. Just enabling the global time range is not enough. It should be used for the graph we show.

srikanthccv:
Discussed over a call

- The charts should be based on the global time selection. 
- The timestamps should be part of the URL, and when a user lands on this page with a shared URL, the timestamps from URL params will be honoured.
- The range from alert definition configuration is independent.

Rajat-Dabade:
@srikanthccv All the points mentioned above are fixed. Please review it once. Thank you.

srikanthccv:
Thanks, I will review it.

srikanthccv:
Where is the global time range filter? How do you expect the user to change the time range to see the data? The issue says we should show the time range dropdown. Or did I miss something?

Screenshot 2023-10-31 at 9 16 14 AM

Rajat-Dabade:
@srikanthccv, for me it is showing

Screenshot 2023-10-31 at 10 18 07 AM

Have to take the latest pull of the branch?

srikanthccv (dismissed):

Rajat-Dabade:
https://www.loom.com/share/c40667132c7e45b0a3c0239ebfac5d4d?sid=b2544154-f7b2-4b84-8195-d5bca43d8582
cc: @YounixM

srikanthccv (approved):

PR #3759, opened 2023-10-17
fix: update version check login and other minor UI fixes

YounixM:


palashgdev (approved):

frontend/src/container/BillingContainer/BillingContainer.tsx
```diff
@@ -202,18 +161,54 @@ export default function BillingContainer(): JSX.Element {
...
 		[licensesData?.payload?.onTrial],
 	);
 
+	const { isLoading, data: usageData } = useQuery(
+		[REACT_QUERY_KEY.GET_BILLING_USAGE, user?.userId],
+		{
+			queryFn: () => getUsage(activeLicense?.key || ''),
+			onError: handleError,
+			enabled: activeLicense !== null,
+			onSuccess: processUsageData,
```
Rajat-Dabade, line 186: `useEffect` on line 212 can be remove as it is handle here.

frontend/src/utils/app.ts
```diff
@@ -17,3 +17,16 @@ export const isCloudUser = (): boolean => {
 
 	return hostname?.endsWith('signoz.cloud');
 };
+
+export const checkVersionState = (
```
Rajat-Dabade, line 21:
```suggestion
export const checkVersionState = (
export const checkVersionState = (currentVersion: string, latestVersion: string): boolean => {
  const versionCore = currentVersion?.split('-')[0];
  return versionCore === latestVersion;
};

```

We can remove the unnecessary if statement as `split` function will return an array containing the original string as its single element if it doesn't find the specified separator in the string

Rajat-Dabade (changes requested):
Some optimisation changes, but everything else looks good to me.

Rajat-Dabade (approved):
LGTM 🚀

PR #3776, opened 2023-10-19
chore: add signozspanmetrics delta temporality pipeline

srikanthccv:
This adds a parallel pipeline with delta temporality to the existing pipeline so that when we switch to delta as default the data is backfilled and doesn't disrupt. I removed the otel collector metrics from the dependencies as it doesn't make much sense in docker environment.


srikanthccv:
@prashant-shahi PTAL

prashant-shahi:
@srikanthccv Can we comment the otel-collector-metrics instead of removing?
Or perhaps include docs for how to add another collector with prometheus config for Docker.

That could help users to migrate any existing prometheus scraper or healthcheck receivers.

srikanthccv:
Let me share my reasoning for removing instead of commenting in docker. The idea behind keeping the otel-collector metrics in k8s makes sense because the powerful nature of k8s allows them to create replicas with the breeze and we don't want to duplicate the scraped metrics in each replica of the main collector.

However, the same is not true for docker env. One can't have multiple services listening on the same port in swarm/compose mode, or unix in general. It's very unlikely someone is running multiple instances of otel collector in docker with some additional proxy set up to route to 431{x/y/z...} of OTLP. So many users are better off adding the scrape config to the same collector config instead of spinning another container. To those who have setup the additional proxy approach mentioned earlier I can write docs for them.

ankitnayan:
Where is the port conflict? 
otel-collector listens on 4317/8 and otel-collector-metrics does not need to expose any port to scrape and store metrics.

Also, how would this work out for folks already using otel-collector-metrics to scrape and store prometheus metrics?

srikanthccv:
Any scraping can happen in the `otel-collector` itself. Why is there a need for `otel-collector-metrics` anymore? It made sense in the k8s because you can run multiple replicas of `otel-collector` and don't want to add redundant data so you would move to `otel-collector-metrics`. However, that's doesn't make sense in docker env.

> how would this work out for folks already using otel-collector-metrics to scrape and store prometheus metrics?

They can simply move the config to otel-collector

ankitnayan:
you can have replicas of otel-collector in `docker-compose` and `swarm`

srikanthccv:
Technically you can run multiple collector containers but each would have to use a different port number (swarm is not an exception) and need another proxy set up to route the data to these replicas which is very unlikely IMO. Even if someone is out there with such a setup I will write a guide for them. But for a large set of users, there is no need to have separate `otel-collector-metrics` and we should keep things simple.

ankitnayan:
ok. We are making it a non-necessary component and available for those who run multiple-replicas of otel-collector via a guide, right? This would be impacting the current OSS users also. Probably they would have added some scraping config and would need to resolve conflicts before pulling the latest. We should add things in the release then

srikanthccv:
Correct, there will be a migration guide for existing users and a general guide to scrape metrics.

ankitnayan (approved):

srikanthccv:
@prashant-shahi, if you don't have any other concerns, I will merge this.

prashant-shahi:
@srikanthccv Can you work on the migration guide and required changes ready for both Docker and K8s before merging this?

Otherwise, these changes will be out with the upcoming release and the docs for migration not ready for public.

srikanthccv:
@prashant-shahi, We haven't removed the `otel-collector-metrics` in k8s but just here is docker env. I added the draft guide here https://github.com/SigNoz/signoz.io/pull/1434, please review.

prashant-shahi:
@srikanthccv Looks good for Docker.

For K8s, it will be difficult for users to go about it - we could suggest OSS users to use otel-deployment (of k8s-infra chart) - while for cloud users we could introduce optional component which is enabled by a flag.

prashant-shahi:
Sounds good then.

Will do the following to better handle it for K8s:
- `otelCollectorMetrics.enabled` flag for condition to enable/disable the component
- Introduce generic scraper config job with empty targets - most likely only after moving the `signoz.io/scrape` scraper out of otelcollector-metrics

prashant-shahi (approved):

prashant-shahi:
@srikanthccv In case of Docker Swarm, we have Global mode set for OtelCollector - which creates multiple containers in each swarm node. It would be best to include an info note about that for Swarm.

PR #3784, opened 2023-10-23
fix: promql alert expr formatting for composite queries with join/unless

srikanthccv:
For expressions such as 

```
count(process_cpu_time{process_command_line=~"/usr/local/cqonline", host_name=~"cq"} offset 5m) by (host_name) unless count(process_cpu_time{process_command_line=~"/usr/local/cqonline", host_name=~"cq"}) by (host_name)
```

The current comparison isn't done against the whole query result but the right side of it which is incorrect.


ankitnayan (approved):

prashant-shahi (approved):

srikanthccv:
Please include this in the release along with the line chart issue.

Issue #3169, opened 2023-07-18
incorrect/missing ID for widget should not make page go blank white screen


srikanthccv:
It is not clear under what circumstance this issue occurs. Whenever
there is an item in the `layout` array of the dashboard json has an `i` with
some id value with no corresponding widget for that ID in the `widgets`
array, the page goes blank white screen. This has occurred a couple of times
but doesn't happen regularly. The entire page shouldn't be breaking for
one missing widget. Two things to fix
1. Handle missing ids gracefully.
2. Troubleshoot the underlying issue and fix it

srikanthccv:
At the minimum, the entire page shouldn't go blank.

sharpSteff:
any workarround or recovery know to this?

with v0.23.1 this issue happend regulary to me, adding/editing a panel and then close the browser tab without saving

PR #3830, opened 2023-10-29
fix: empty widget is handled

palashgdev:
Close https://github.com/SigNoz/signoz/issues/3169

if widget is not present in the layout
Screenshot 2023-10-29 at 3 21 18 PM

### DIFF
--- a/frontend/src/container/ListOfDashboard/ImportJSON/index.tsx
+++ b/frontend/src/container/ListOfDashboard/ImportJSON/index.tsx
@@ -66,8 +66,10 @@ function ImportJSON({
 			setDashboardCreating(true);
 			const dashboardData = JSON.parse(editorValue) as DashboardData;
 
-			if (dashboardData.layout) {
+			if (dashboardData?.layout) {
 				dashboardData.layout = getUpdatedLayout(dashboardData.layout);
+			} else {
+				dashboardData.layout = [];
 			}
 
 			const response = await createDashboard({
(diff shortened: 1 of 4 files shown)


frontend/src/container/ListOfDashboard/ImportJSON/index.tsx
```diff
@@ -65,6 +66,10 @@ function ImportJSON({
 			setDashboardCreating(true);
 			const dashboardData = JSON.parse(editorValue) as DashboardData;
 
+			if (dashboardData.layout) {
```
YounixM, line 69: Safety check

YounixM: Add else condition handling

frontend/src/lib/dashboard/getUpdatedLayout.ts
```diff
@@ -0,0 +1,23 @@
...
+	widgetLayout = widgetLayout?.filter((i) => {
+		const duplicate = seen.has(i.i);
+		seen.add(i.i);
+		return !duplicate;
+	});
+
+	// filter EMPTY_WIDGET from i due to previous version of signoz
+	widgetLayout = widgetLayout?.filter((i) => i.i !== PANEL_TYPES.EMPTY_WIDGET);
+
+	return widgetLayout || [];
```
srikanthccv, line 22: I think this kind of code is wiping out the layout array. Assume the dashboard GET request failed for some reason. What would be the layout array on the client side? And if the backend works again after some time will there be PUT with the empty layout array?

palashgdev: discussed offline need to find the case where layout is getting empty which is not related to this PR

palashgdev: cc: @srikanthccv @YounixM resolving this thread

YounixM (approved):

Issue #3833, opened 2023-10-30
Enable `export as csv` in top operations section in the service overview page


ankitnayan:
Mainly 2 usecases:
- download csv and send it to clients to show p99 weekly latency
- download csv to internally compare the perf of endpoints as compared to last week or after a release or perf test

There is already a component to download logs as csv. The same component can be used at frontend

Issue #3832, opened 2023-10-30
Enable text search without suggestions in the `top operations` section of APM


ankitnayan:
It becomes difficult to find and API endpoint as there are a lot of other operations emitted by the service. Moreover viewing a group of APIs like `/v1/group/xxxx` 

- No API call needed as filtering can be done in the frontend only
- No suggestions are needed as suggestions are useful to find a specific endpoint

PR #3848, opened 2023-10-31
[Feat]: Download as CSV and Execl and Search option for key-operation table

Rajat-Dabade:
Close: https://github.com/SigNoz/signoz/issues/3832
Close: https://github.com/SigNoz/signoz/issues/3833

### DIFF
--- a/frontend/src/container/QueryTable/utils.ts
+++ b/frontend/src/container/QueryTable/utils.ts
@@ -5,10 +5,10 @@ export function createDownloadableData(
 ): Record<string, string>[] {
 	return inputData.map((row) => ({
 		Name: row.operation.toString(),
-		P50: row.A.toString(),
-		P90: row.B.toString(),
-		P99: row.C.toString(),
+		'P50 (in ns)': row.A.toString(),
+		'P90 (in ns)': row.B.toString(),
+		'P99 (in ns)': row.C.toString(),
 		'Number Of Calls': row.F.toString(),
-		'Error Rate (%)': row.F1 !== 'N/A' ? row.F1.toString() : '0',
+		'Error Rate (%)': row.F1 && row.F1 !== 'N/A' ? row.F1.toString() : '0',
 	}));
 }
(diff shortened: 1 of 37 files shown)


palashgdev:
@Rajat-Dabade this method does not take unit into account please fix this one

example: 
Screenshot 2023-10-31 at 3 48 55 PM

palashgdev:
ideally this downloading of csv should be handled from the backend itself

ankitnayan:
> ideally this downloading of csv should be handled from the backend itself taking the account of pagination in future it will be hard for the user/ client to aggregate the data

yeah..when pagination comes to play or there is huge data the search should also query the backend. But this data is not supposed to grow big so we should be good

ankitnayan:
Screenshot 2023-10-31 at 9 21 53 PM

1. Name should be the 1st column
2. P50 should be `P50 (in ms)` and the same for the other P95 and P99
3. P50 (in ms) value should be in `ms`. The current values are in nano. 
4. Change styling after UX consultation in a separate PR @GeekBoySupreme

ankitnayan:
- It's `Error (%)` and not `Error Count` and the value is also incorrect
- the values of `P50`, `P95` and `P99` are duplicated and incorrect. Please verify data sanity before pushing any changes
- append name of the service to filename? `top-operations-customer.xls/csv`

ankitnayan (dismissed):

ankitnayan:
@Rajat-Dabade can you also check why the numbers are treated as text in `xls` but as numbers in `csv`. We should treat them as numbers. Selecting multiple values in the column `P50` should show Sum for numbers.

palashgdev (approved):

frontend/src/container/QueryTable/utils.ts
```diff
@@ -0,0 +1,14 @@
+import { RowData } from 'lib/query/createTableColumnsFromQuery';
+
+export function createDownloadableData(
+	inputData: RowData[],
+): Record<string, string>[] {
+	return inputData.map((row) => ({
```
YounixM, line 6: Add safety checks wherever required. We should handle safety checks everywhere by default when not strictly typed.

Rajat-Dabade: Added the safety check.

YounixM (approved):

PR #3856, opened 2023-11-01
chore: dashboard locking ee query-service

srikanthccv:


srikanthccv:
We don't export the locked field in the JSON. Import JSON creates a new dashboard without any locking. If one wants to lock they need to explicitly lock from UI. The same applies to other top-level fields such as "created{At, By}", and "updated{At, By}". They are not exported or imported in JSON.

srikanthccv:
Only this `Data` is exported/imported https://github.com/SigNoz/signoz/blob/092d164d55dd4784036f47f11cb28d7811b2a27f/pkg/query-service/app/dashboards/model.go#L143

pkg/query-service/app/dashboards/model.go
```diff
@@ -141,6 +148,7 @@ type Dashboard struct {
 	UpdateBy  *string   `json:"updated_by" db:"updated_by"`
 	Title     string    `json:"-" db:"-"`
 	Data      Data      `json:"data" db:"data"`
+	Locked    *int      `json:"locked" db:"locked"`
```
ankitnayan, line 151: doesn't this get exported?

srikanthccv: No, only the `Data` part gets exported or imported. The JSON from the `data` field is the only part we export or import (Look at the API response of dashboard calls for a better idea).

ankitnayan:
the `Lock` UI component should be visible to the OSS users. If they try to lock, there has to be an implementation in the OSS api too saying it's available in the `ee` plan something like with SSO?

Or the OSS users should not see this button? @pranay01

pkg/query-service/app/dashboards/model.go
ankitnayan: ok

srikanthccv:
Assuming by OSS you mean an entirely EE license-free build, we took a call not to upsell to them. They will simply not see the button. However, we upsell to EE community edition like a limit on panel count. Did you intend to say that this should be only available to ENTERPRISE_LICENSE users but not for the community edition?

ankitnayan:
> Assuming by OSS you mean an entirely EE license-free build, we took a call not to upsell to them. They will simply not see the button. However, we upsell to EE community edition like a limit on panel count. Did you intend to say that this should be only available to ENTERPRISE_LICENSE users but not for the community edition?

Ok, if we decided not to upsell to community edition, then:
- The default build is enterprise build and hence should show this button.
- The community build should not see this in the UI. We can enable/disable this using the upsell flag. I am not sure if this is already provisioned and implemented for SSO

srikanthccv:
Just to make sure we are using the same names to not confuse

1. OSS Users - The non-EE code build who use `-oss` docker builds - We don't upsell this audience. 
2. EE with Free License aka Community edition - The default build published by and used by many users - We upsell to this audience. 
3. EE with Enterprise license - They don't have an upsell point in the product (for now).

ankitnayan:
> Just to make sure we are using the same names to not confuse
> 
> 1. OSS Users - The non-EE code build who use `-oss` docker builds - We don't upsell this audience.
> 2. EE with Free License aka Community edition - The default build published by and used by many users - We upsell to this audience.
> 3. EE with Enterprise license - They don't have an upsell point in the product (for now).

Got it. I wanted to point out that users in 1 should be taken care. Is this enabled in the product using the upsell flag and checks at frontend? Or is it yet to be taken care of?

srikanthccv:
Yes, It is taken care of by using the `OSS` & `DISABLE_UPSELL` feature flags. https://github.com/SigNoz/signoz/blob/7603e0ebe04fdced5ad8408fa18f4038c05a5308/pkg/query-service/model/featureSet.go#L26-L28 I haven't tested it lately so I am not sure if they are still working.

ankitnayan:
> Yes, It is taken care of by using the `OSS` & `DISABLE_UPSELL` feature flags.
> 
> https://github.com/SigNoz/signoz/blob/7603e0ebe04fdced5ad8408fa18f4038c05a5308/pkg/query-service/model/featureSet.go#L26-L28
> 
> I haven't tested it lately so I am not sure if they are still working.

@YounixM you should be keeping this in mind. Also, give the thread a read.

srikanthccv:
@ankitnayan please review. I would like to get it merged so that https://github.com/SigNoz/signoz/pull/3880 can be tested against staging.

ankitnayan (approved):

PR #3873, opened 2023-11-02
fix(tags): tag modification in triggered alerts page

joemiltonm:
Hi SigNoz team, 

I have modified the tags column in the triggered alerts page by reusing the LableColumn component, which is used by alerts list page. 

below is the screenshot before modification
Screenshot from 2023-11-02 16-21-03

here is the page after modification
Screenshot from 2023-11-02 16-19-15


palashgdev:
thanks @joemiltonm seems good :-D

palashgdev (approved):
@YounixM please take a look

PR #3894, opened 2023-11-03
fix: Tooltip values is now fixed

palashgdev:
https://github.com/SigNoz/signoz/assets/88981777/dc9c1a85-d3eb-4658-8b3c-e324f7f4d9ef

### DIFF
--- a/frontend/src/lib/uPlotLib/getUplotChartData.ts
+++ b/frontend/src/lib/uPlotLib/getUplotChartData.ts
@@ -115,7 +115,7 @@ const createDivsFromArray = (
 				);
 
 				div.textContent = formattedDate;
-			} else if (item.show) {
+			} else if (item.show && data[index][idx]) {
 				const color = colors[(index - 1) % colors.length];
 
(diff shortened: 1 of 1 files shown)


frontend/src/lib/uPlotLib/getUplotChartData.ts
```diff
@@ -118,33 +118,31 @@ const createDivsFromArray = (
 			} else {
 				const color = colors[(index - 1) % colors.length];
 
-				if (data[idx]) {
-					const squareBox = document.createElement('div');
-					squareBox.classList.add('pointSquare');
+				const squareBox = document.createElement('div');
```
YounixM, line 121: We should have this check to avoid creating unwanted div s when data isn't present

palashgdev: this is already controlled via

```tsx
if (item.show) {
					div.appendChild(squareBox);
					div.appendChild(text);
				}
```

YounixM (approved):

Issue #3912, opened 2023-11-07
Creating alerts directly from Dashboards


pranay01:
Users should be able to create Alerts from Dashboards for:
- [ ] TimeSeries panel
- [ ] Values panel

srikanthccv:
This is done.

PR #3924, opened 2023-11-08
chore: dashboards to alerts creation support in query-service

srikanthccv:
Part of https://github.com/SigNoz/signoz/issues/3912

Users will be redirected from the dashboard page to the alerts page with all the queries in a widget since there can be more than one query (for example P90, P95, P99), we ask users to select the query (it can't be multiple) that should be used against the threshold. This change add the support with `selectedQueryName` that falls back to old logic when it doesn't exist.

### DIFF
--- a/deploy/docker-swarm/clickhouse-setup/otel-collector-config.yaml
+++ b/deploy/docker-swarm/clickhouse-setup/otel-collector-config.yaml
@@ -61,35 +61,6 @@ receivers:
               job_name: otel-collector
 
 processors:
-  logstransform/internal:
-    operators:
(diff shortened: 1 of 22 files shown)


nityanandagohain:
Don't have much knowledge on alerts, I am probably not the best presion to review thest.

srikanthccv:
Honestly, nobody else has any context either and that's why I tried to add as many comments as needed to help understand the change. I will approval merge it so I asked both of you.

nityanandagohain (approved):
From what I understand this looks good to me. But hope the happy flow is tested properly.

PR #3925, opened 2023-11-08
feat: replace time series graph in logs explorer and trace explorer w…

YounixM:
…ith uplot

### DIFF
--- a/frontend/src/container/TimeSeriesView/TimeSeriesView.tsx
+++ b/frontend/src/container/TimeSeriesView/TimeSeriesView.tsx
@@ -1,7 +1,9 @@
-import Graph from 'components/Graph';
 import Spinner from 'components/Spinner';
-import getChartData from 'lib/getChartData';
-import { useMemo } from 'react';
+import Uplot from 'components/Uplot';
+import { useIsDarkMode } from 'hooks/useDarkMode';
+import { getUPlotChartOptions } from 'lib/uPlotLib/getUplotChartData';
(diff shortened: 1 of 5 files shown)


palashgdev (approved):
LGTM

frontend/src/container/TimeSeriesView/TimeSeriesView.tsx
```diff
@@ -13,31 +15,45 @@ function TimeSeriesView({
...
+
+	const chartOptions = getUPlotChartOptions({
+		yAxisUnit: yAxisUnit || '',
+		apiResponse: data?.payload,
+		dimensions: {
+			width,
+			height,
+		},
+		isDarkMode,
+	});
```
Rajat-Dabade, line 42: This is the same code we are using in `FullView` [here](https://github.com/SigNoz/signoz/blob/2d11aa7fc6f97b54490c0125c25a63d58f13e0c1/frontend/src/container/GridCardLayout/GridCard/FullView/index.tsx#L97C5-L97C5) as well, Can we create a Custom hook for which can take ref as input and return the chart options?

Rajat-Dabade (changes requested):

frontend/src/container/TimeSeriesView/TimeSeriesView.tsx
YounixM: It's just return height and width, the fallback height and width can be set up case by case.

Rajat-Dabade (approved):

PR #3928, opened 2023-11-08
feat: added services page playwright tests

Vikrant2520:
Test Case added :- 

No Services Test -> 

- Login into the application 
- render the services page with empty data
- expect the no data text to be present on the screen

With Services Data -> 

- Login into the application 
- Render the base service page
- Check for the presence of correct breadcrumbs 
- Check if the table is rendered with the correct columns
- Check if the table is rendered with the correct service data

Screenshot 2023-11-09 at 1 24 08 AM

### DIFF
--- a/frontend/tests/service/servicesLanding.spec.ts
+++ b/frontend/tests/service/servicesLanding.spec.ts
@@ -1,0 +1,82 @@
+import { expect, Page, test } from '@playwright/test';
+import ROUTES from 'constants/routes';
+
+import servicesSuccessResponse from '../fixtures/api/services/200.json';
+import { loginApi } from '../fixtures/common';
+
+let page: Page;
+
+test.describe('Service Page', () => {
+	test.beforeEach(async ({ baseURL, browser }) => {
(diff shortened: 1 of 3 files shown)


Vikrant2520:
this is base setup this test would be extended to cover happy / edge case flows

YounixM (approved):

frontend/tests/service/servicesLanding.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+		const p99Latency = await page
+			.locator(
+				`th[aria-label*="this column's title is P99 latency (in ms)"] .ant-table-column-title`,
+			)
+			.textContent();
+
+		await expect(p99Latency).toEqual('P99 latency (in ms)');
+		const errorRate = await page
+			.locator(
+				`th[aria-label*="this column's title is Error Rate (% of total)"] .ant-table-column-title`,
```
palashgdev, line 67: would recommend to add data test id / constant hard code thing would not scale

Vikrant2520: yes will add data-test-id's when completing the flow

PR #3933, opened 2023-11-09
chore: :wrench: disable internal metrics collection of OtelCollectors

prashant-shahi:
Signed-off-by: Prashant Shahi <prashant@signoz.io>

### DIFF
--- a/deploy/docker-swarm/clickhouse-setup/otel-collector-config.yaml
+++ b/deploy/docker-swarm/clickhouse-setup/otel-collector-config.yaml
@@ -56,7 +56,8 @@ receivers:
         - job_name: otel-collector
           static_configs:
           - targets:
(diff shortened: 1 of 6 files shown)


srikanthccv:
There is no need to remove it here and self-hosted users want to their ingestion and internal metrics. You should update the docs on signoz.io Cloud instructions instead https://signoz.io/docs/tutorial/opentelemetry-binary-usage-in-virtual-machine/ and other places where Cloud users are mistakingly sending metrics that they don't use.

srikanthccv:
Please just update the Cloud user docs

prashant-shahi:
@srikanthccv Doesn't this affect the self-host users with Enterprise license the same way for that of Cloud users?

PR #3943, opened 2023-11-10
fix: adding 2 pipeline processors with same name should not break the UI

raj-k-singh:
Fixes https://github.com/SigNoz/signoz/issues/3922

Before:
name-collision-in-ui-before

After:
name-collision-in-ui-after

### DIFF
--- a/frontend/src/container/PipelinePage/PipelineListsView/AddNewProcessor/index.tsx
+++ b/frontend/src/container/PipelinePage/PipelineListsView/AddNewProcessor/index.tsx
@@ -7,6 +7,7 @@ import {
 	PipelineData,
 	ProcessorData,
 } from 'types/api/pipeline/def';
+import { v4 } from 'uuid';
 
 import { ModalButtonWrapper, ModalTitle } from '../styles';
 import { getEditedDataSource, getRecordIndex } from '../utils';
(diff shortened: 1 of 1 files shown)


frontend/src/container/PipelinePage/PipelineListsView/AddNewProcessor/index.tsx
```diff
@@ -58,8 +58,12 @@ function AddNewProcessor({
 	const onFinish = (values: { name: string }): void => {
 		const totalDataLength = expandedPipelineData?.config?.length || 0;
 
+		const nameWithoutSpaces = (values?.name || '').replace(/\s/g, '');
```
palashgdev, line 61: would trim help here ?

frontend/src/container/PipelinePage/PipelineListsView/AddNewProcessor/index.tsx
```diff
@@ -58,8 +58,12 @@ function AddNewProcessor({
...
+		const randomIdSuffix = Math.random().toString(36).slice(2, 6);
```
palashgdev, line 62: would uuid here also

palashgdev (approved):

PR #3944, opened 2023-11-10
Fix/app dex aligment

palashgdev:


palashgdev:
small styles update @YounixM merging now

Issue #3679, opened 2023-10-06
Restriction and limits on potentially dangerous queires


srikanthccv:
## Bug description

1. Some queries can attempt to process a lot of raw data that is practically useless; these are slower, memory intensive and generally resource-draining for the result that is little to no use. 
2. Some queries can return many rows for aggregation queries that contain a lot of values. These values without limits don't make sense. We should either avoid running such queries or cancel them if the result turns out to be large.
3. Limit the amount of data we try to aggregate and make user reduce/refine their search.

srikanthccv:
After spending some time on this, I realized it's premature to attempt to build any robust solution with separate users, quotas, limits etc... at this point as we are still in the process of understanding query patterns, and schemas performance https://github.com/SigNoz/signoz/issues/6503, consideration of various strategies for addressing the genuine user requirements https://www.notion.so/signoz/CH-query-resources-management-144fcc6bcd1980f8b634d4735f7c9b12 (I was trying to attribute the amount of CPU/Memory/IO by sources of the query and data source to use it as datapoint for potential reserved resources but that would not be relevant based our actions from the linked doc). In this phase, I think we should enforce generous global limits that shouldn't be a problem for most users, if someone runs into it, we will understand the problem and actively help and learn to generalize the solution. 

@ankitnayan let me know what you think about this.

PR #3956, opened 2023-11-14
feat: add support for max timeseries per query limit

srikanthccv:
Part of https://github.com/SigNoz/signoz/issues/3679

Content of the research work [doc](https://www.notion.so/signoz/ClickHouse-limit-resource-usage-676546168e4242798b2fef3b9141aa56) for this so that everyone can see it.

# ClickHouse limit resource usage:

ClickHouse limit resource usage:

There are primarily two ways to apply limits in ClickHouse:

1. Quotas
2. Query wise limits

## Quotas

- Quotas are used to limit resource usage per user or per user/host combination.
- Quotas are applied over a period of time.
- Should be configured in the `users.xml`.

## Query wise limits

- Applied to a single query.
- Can return partial results or throw an exception when the limit is reached.

### Read limits

- Limits the number of rows/bytes read from the table. In the case of distributed queries, the limit is applied to each shard.
- `max_{rows/bytes}_to_read[_leaf]`
- `max_{rows/bytes}_in_distinct`

In general, It’s hard to come up with a reasonable limit for these settings. We can’t put restrictions on the number of rows alone, because a simple COUNT query may read billions of rows but complete in a few milliseconds. However, for time series data, we can limit the number of distinct fingerprints read from the table. We can throw an exception when the limit is reached and ask the user to narrow down the query by adding more filters. This limit should be configurable from the admin API.

Example:

```sql
SELECT fingerprint
FROM signoz_metrics.distributed_time_series_v2
WHERE metric_name = 'signoz_calls_total'
SETTINGS max_rows_to_read = 300000

Query id: b4c64e1f-02a0-4704-8557-d59deac52a7d

Elapsed: 0.019 sec.

Received exception from server (version 23.11.1):
Code: 158. DB::Exception: Received from localhost:9000. DB::Exception: Limit for rows or bytes to read exceeded, max rows: 300.00 thousand, current rows: 567.84 thousand: While executing Remote. (TOO_MANY_ROWS)
```

```sql
SELECT fingerprint
FROM signoz_metrics.distributed_time_series_v2
WHERE metric_name = 'signoz_calls_total'
SETTINGS max_rows_to_read_leaf = 300000

Query id: 1e3905e3-e26c-4884-a95f-1f0daed9c506

┌─count()─┐
│   92062 │
└─────────┘

1 row in set. Elapsed: 0.086 sec. Processed 567.84 thousand rows, 567.84 KB (6.63 million rows/s., 6.63 MB/s.)
Peak memory usage: 678.16 KiB.
```

### Limits during the GROUP BY clause

- `max_rows_to_group_by`, `group_by_overflow_mode`

This is not the number of rows read from the table, but the number of rows in the result of the GROUP BY clause Limits the number of unique keys in the GROUP BY clause. The more unique keys there are, the more memory is required and the slower the query is. This is useful to prevent the server from running out of memory when the number of unique keys is too large.

The overflow mode `any` can return partially correct results.

Example:

```sql
SELECT
    count(),
    serviceName
GROUP BY serviceName
SETTINGS max_rows_to_group_by = 10

Query id: bfd21191-8762-452b-adfe-a4f7003a25bc

Elapsed: 0.098 sec.

Received exception from server (version 23.11.1):
Code: 158. DB::Exception: Received from localhost:9000. DB::Exception: Limit for rows to GROUP BY exceeded: has 34 rows, maximum: 10: While executing AggregatingTransform. (TOO_MANY_ROWS)
```

```sql
SELECT
    count(),
    serviceName
GROUP BY serviceName
SETTINGS max_rows_to_group_by = 40
34 rows in set. Elapsed: 0.656 sec. Processed 209.99 million rows, 209.99 MB (320.12 million rows/s., 320.12 MB/s.)
Peak memory usage: 1.29 MiB.
```

If a high cardinality column is used in the GROUP BY clause, the query will be slow and may run out of memory. We can prevent such queries from running by setting a limit. In the builder UI, we have a `HAVING` clause which can be used to filter out the results which might conflict with the limit setting.

### Limits during the ORDER BY clause

- `max_{rows/bytes}_to_sort`

Limits the number of rows/bytes to sort. The query will fail if the limit is exceeded. Allows to prevent the server from running out of memory when sorting large amounts of data. Any ORDER BY that doesn’t use a sorting key will require a full sort of the data. This is useful to prevent the server from running out of memory when the number of rows to sort is too large.

Example:

```sql
SELECT body
FROM signoz_logs.logs
ORDER BY span_id ASC
LIMIT 1
SETTINGS max_rows_to_sort = 500000

Query id: 01dce124-cb9a-4a7c-b86d-5aeb3b29830c

Elapsed: 0.685 sec. Processed 7.17 million rows, 1.83 GB (10.47 million rows/s., 2.68 GB/s.)
Peak memory usage: 2.66 GiB.

Received exception from server (version 23.11.1):
Code: 396. DB::Exception: Received from localhost:9000. DB::Exception: Limit for result exceeded, max rows: 500.00 thousand, current rows: 501.20 thousand. (TOO_MANY_ROWS_OR_BYTES)
```

This mostly applies to traces/logs list queries. We should disallow sorting by columns that do full sorting.

### Limits during the results phase

- `max_result_{rows/bytes}`

Limits the number of rows/bytes in the result.

We know the number of rows we expect to return for each panel type. We can set the limit based on that. For example, for a line chart, we can set the limit to (200 series * 300 points) = 60,000 rows. For a list query, we can set the limit to 1000 rows.

### Limits during the execution phase

- `max_execution_time`
- `{max/min}_execution_speed[_bytes]`

The query is cancelled if the projected exec time is greater than the limit. This operates based on interpolation relative to the current query execution speed.

Some queries can become extremely slow due to the nature of the query. It’s better to cancel such queries instead of letting them run for a long time. We can set a limit based on the expected execution time for each panel type.

### Limits during the JOIN/IN sections

- `max_{rows/bytes}_in_join`
- `max_{rows/bytes}_to_transfer`

Limits the number of rows/bytes in the result of the JOIN/IN sections. The query will fail if the limit is exceeded.

Might not be very useful for us. We expect the left table (samples) to be large.

# This PR

In every metrics query, there is a subquery that is responsible for fetching the time series based on the filters which are then joined with the samples table. When the number of time series to join against is high, the query takes the high resources. This PR makes use of the `max_rows_in_distinct` setting to limit the number of result rows in the time series subquery. If the distinct number of fingerprints exceeds the limit set then the query is cancelled with an error and we let the user know the reason for query failure and notify them to add more filters. Any zero value limit is same as no limit.


ee/query-service/app/api/metrics.go
```diff
@@ -234,3 +237,52 @@ func (ah *APIHandler) queryRangeMetricsV2(w http.ResponseWriter, r *http.Request
...
+
+	if len(queryRangeParams.CompositeQuery.BuilderQueries) > 0 {
+		for idx := range queryRangeParams.CompositeQuery.BuilderQueries {
+			queryRangeParams.CompositeQuery.BuilderQueries[idx].QueryLimits = v3.QueryLimits{
+				MaxTimeSeries: timeSeriesLimt,
+			}
+			// TODO(srikanthccv): should apply to explore page also
+			if !strings.Contains(queryRangeParams.SourcePage, "dashboard") {
+				queryRangeParams.CompositeQuery.BuilderQueries[idx].QueryLimits.MaxTimeSeries = 0
+			}
```
srikanthccv, line 274: When you are setting up an alert, there is a high chance the query crosses the threshold limit pretty easily. For example, one would want to apply a threshold query for all possible combinations of (service_name, endpoint) which could potentially mean as much data as a whole metric. Should there be limits for alert previews as well?

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -4491,6 +4491,14 @@ func readRowsForTimeSeriesResult(rows driver.Rows, vars []interface{}, columnNam
 		seriesToPoints[key] = append(seriesToPoints[key], metricPoint)
 	}
 
+	if rows.Err() != nil {
+		zap.S().Errorf("error while reading time series result %v", rows.Err())
+		if strings.Contains(rows.Err().Error(), "code: 191") {
+			return nil, fmt.Errorf(`Exceeded max number of series per query. Please reduce the number of series by applying filters.`)
```
srikanthccv, line 4497: I don't like the error string checking but there is no other option AFAIK.

ee/query-service/app/api/metrics.go
srikanthccv: The chances of APM metrics going beyond the limit are low because we control what labels go in the `signoz_` metrics. But there is a chance some users would put timestamps/url params in the span names which can create an issue. Should there be limit for APM metrics but little higher than the configured limit.

shivanshuraj1333:
This needs some rework, can sync on this to understand the PR better.

PR #3972, opened 2023-11-15
fix: improve user telemetry

makeavish:
- move GetEmailFromJwt to common function
- update AttachJwtToContext() to use standard way of attaching value to context
- update userEmail in every possible sendEvent call
- send groupId in sendEvent call for SaaS operator analytics


pkg/query-service/telemetry/telemetry.go
```diff
@@ -334,8 +334,16 @@ func (a *Telemetry) checkEvents(event string) bool {
 	return sendEvent
 }
 
-func (a *Telemetry) SendEvent(event string, data map[string]interface{}, opts ...bool) {
+func (a *Telemetry) SendEvent(event string, data map[string]interface{}, userEmail string, opts ...bool) {
 
+	// ignore telemetry for default user
+	if userEmail == "admin@signoz.cloud" {
```
nityanandagohain, line 340: move this value to constants.go

pkg/query-service/app/http_handler.go
```diff
@@ -1397,8 +1397,11 @@ func (aH *APIHandler) submitFeedback(w http.ResponseWriter, r *http.Request) {
 		"email":   email,
 		"message": message,
 	}
-	telemetry.GetInstance().SendEvent(telemetry.TELEMETRY_EVENT_INPRODUCT_FEEDBACK, data)
-
+	ctx := auth.AttachJwtToContext(r.Context(), r)
+	userEmail, err := auth.GetEmailFromJwt(ctx)
```
nityanandagohain, line 1401: can make a one liner for this, since every where two lines are getting replicated ?

makeavish: These two lines are not replicated everywhere and these two lines perform separate functions. So creating a function doesn't makes sense.

pkg/query-service/telemetry/telemetry.go
makeavish: Done

ee/query-service/app/api/license.go
```diff
@@ -64,7 +64,7 @@ func (ah *APIHandler) applyLicense(w http.ResponseWriter, r *http.Request) {
 		RespondError(w, model.BadRequest(fmt.Errorf("license key is required")), nil)
 		return
 	}
-
+	ctx := auth.AttachJwtToContext(r.Context(), r)
```
srikanthccv, line 67: Why is this done here?

ee/query-service/app/server.go
```diff
@@ -437,7 +438,11 @@ func extractQueryRangeV3Data(path string, r *http.Request) (map[string]interface
 			telemetry.GetInstance().AddActiveLogsUser()
 		}
 		data["dataSources"] = dataSources
-		telemetry.GetInstance().SendEvent(telemetry.TELEMETRY_EVENT_QUERY_RANGE_V3, data, true)
+		ctx := auth.AttachJwtToContext(r.Context(), r)
```
srikanthccv, line 441: I would instead attach this at the beginning of the request.

srikanthccv (commented):
The user object is already attached to every request context. You can use `common.GetUserFromContext(ctx)`. There is no need to call `auth.AttachJwtToContext` under every route that you need email.

ee/query-service/app/server.go
makeavish: user is not available with this function

ee/query-service/app/api/license.go
makeavish: It's used [here](https://github.com/SigNoz/signoz/pull/3972/files#diff-571d493907b4315fc1ac269efcf7668db417a5a591c6cd62e582046ed24b1043R263)

ee/query-service/app/server.go
srikanthccv: If it doesn't exist because this middleware is running much earlier, add a middleware that attaches the user/jwt to the context instead of doing it in each endpoint.

makeavish: Yeah it makes sense to add a new middleware, Let me try that.

pkg/query-service/app/http_handler.go
makeavish: Moved AttachJwtToContext to middleware

srikanthccv (dismissed):

srikanthccv (approved):

PR #3989, opened 2023-11-17
feat: added dashboard list and create tests

Vikrant2520:
Test takes care of the following use case - 

- navigate to the dashboards list page 
- check the dashboards listing with no data
- create a new dashboard 
- navigate to the settings drawer 
- edit the name and the description of the dashboard
- save the config
- validate the updated config
- add a timeseries widget with a specific title 
- validate the presence of the widget in the landing page

### DIFF
--- a/frontend/src/container/GridCardLayout/GridCardLayout.tsx
+++ b/frontend/src/container/GridCardLayout/GridCardLayout.tsx
@@ -118,7 +118,11 @@ function GraphLayout({ onAddPanelHandler }: GraphLayoutProps): JSX.Element {
 					)}
 
 					{addPanelPermission && (
-						<Button onClick={onAddPanelHandler} icon={<PlusOutlined />}>
+						<Button
+							onClick={onAddPanelHandler}
+							icon={<PlusOutlined />}
+							data-testid="add-panel"
+						>
 							{t('dashboard:add_panel')}
 						</Button>
 					)}
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -33,9 +33,7 @@ if (container) {
 						<Provider store={store}>
 							<AppRoutes />
 						</Provider>
-						{process.env.NODE_ENV === 'development' && (
-							<ReactQueryDevtools initialIsOpen />
-						)}
+						{process.env.NODE_ENV === 'development' && <ReactQueryDevtools />}
 					</QueryClientProvider>
 				</ThemeProvider>
 			</HelmetProvider>
--- a/frontend/tests/dashboards/utils.ts
+++ b/frontend/tests/dashboards/utils.ts
@@ -6,6 +6,8 @@ export const getDashboardsListEndpoint = 'v1/dashboards';
 
 export const getIndividualDashboardsEndpoint = 'v1/dashboards/**';
 
+export const queryRangeApiEndpoint = 'query_range';
+
 // element's data-testid's
 export const newDashboardBtnID = 'create-new-dashboard';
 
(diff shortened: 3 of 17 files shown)


frontend/src/index.tsx
```diff
@@ -33,9 +33,7 @@ if (container) {
 						<Provider store={store}>
 							<AppRoutes />
 						</Provider>
-						{process.env.NODE_ENV === 'development' && (
-							<ReactQueryDevtools initialIsOpen />
-						)}
+						{process.env.NODE_ENV === 'development' && <ReactQueryDevtools />}
```
Vikrant2520, line 36: we do not need the react-query tools to be open by default

palashgdev (approved):

frontend/src/container/GridCardLayout/GridCardLayout.tsx
```diff
@@ -118,7 +118,11 @@ function GraphLayout({ onAddPanelHandler }: GraphLayoutProps): JSX.Element {
 					)}
 
 					{addPanelPermission && (
-						<Button onClick={onAddPanelHandler} icon={<PlusOutlined />}>
+						<Button
+							onClick={onAddPanelHandler}
+							icon={<PlusOutlined />}
+							data-testid="add-panel"
```
Rajat-Dabade, line 124: Can we do this for all the `data-testid` in the PR so that there will be only a single source of truth?
```suggestion
							data-testid={addPanelID}
```

frontend/tests/dashboards/utils.ts
```diff
@@ -0,0 +1,180 @@
+import { Page } from '@playwright/test';
+import { JsonApplicationType } from '../fixtures/constant';
+
+// API endpoints
+export const getDashboardsListEndpoint = 'v1/dashboards';
```
Rajat-Dabade, line 5: Please create `constants.ts` and shift all the constants there.

Rajat-Dabade (commented):
LGTM 🚀 , some minor changes. Nice work @Vikrant2520

frontend/src/container/GridCardLayout/GridCardLayout.tsx
Vikrant2520: @Rajat-Dabade ideally one data-testid should be linked to only one element , so basically there is always only one source of truth.

frontend/tests/dashboards/utils.ts
Vikrant2520: will take care of this in follow-up PR for all three pages

frontend/src/container/GridCardLayout/GridCardLayout.tsx
Rajat-Dabade: I was concerned about variables not with respect to `data-testid`. 

For example, the test case will fail if I change the string "add-panel" to something else. If I create a variable and use it at both places and change it at any time, the test case will still pass as the `data-testid` for the component and the test case will be the same. 

Hope it is clear, let me know.

frontend/tests/dashboards/utils.ts
Rajat-Dabade: Works 👍

frontend/src/container/GridCardLayout/GridCardLayout.tsx
Vikrant2520: https://signoz-team.slack.com/archives/C062T9FF9C3/p1700202040468499?thread_ts=1700165038.498609&cid=C062T9FF9C3

Rajat-Dabade (approved):

YounixM (approved):

Issue #3574, opened 2023-09-15
After saving a panel the user is taken to the top of the dashboard


ankitnayan:
Consider the user behaviour:
- user scrolls down the panels in a dashboard page
- user edits a panel
- user is taken to the top of the dashboards page after hitting save
- user will have to scroll down and find the panel again to check it

Ask: When a user edits a panel and saves it, he should be at the position of the dashboards page where he can see the panel

PR #4046, opened 2023-11-23
FE: while saving widget scrolled into view

palashgdev:
Close https://github.com/SigNoz/signoz/issues/3574

after:

https://github.com/SigNoz/signoz/assets/88981777/913fafd3-6c1f-471f-a467-0cbf3e31ba98

before

https://github.com/SigNoz/signoz/assets/88981777/8226a739-b4df-4aab-b0ea-301a66c1fb4a

### DIFF
--- a/frontend/src/container/GridCardLayout/GridCard/index.tsx
+++ b/frontend/src/container/GridCardLayout/GridCard/index.tsx
@@ -9,7 +9,8 @@ import { getUPlotChartOptions } from 'lib/uPlotLib/getUplotChartOptions';
 import { getUPlotChartData } from 'lib/uPlotLib/utils/getUplotChartData';
 import isEmpty from 'lodash-es/isEmpty';
 import _noop from 'lodash-es/noop';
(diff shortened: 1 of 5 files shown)


Rajat-Dabade (approved):

Rajat-Dabade (dismissed):
LGTM 🚀

srikanthccv:
I don't this is the right fix. It makes the experience worse because by moving it to the first position you are disturbing their panel organization. Their order contains some meaning to it. The panels are organized in a way that they can see similar panels in the same window view. We shouldn't move the panel to the first position. I think what @ankitnayan is asking is to scroll them to the panel's original position after it is saved so they see the edited panel in the current window view. 

Here is a video to give an idea: https://www.loom.com/share/b9219ac09ff340dda2fd4df602f53d96

srikanthccv:
In general, I think the reviews of frontend changes should be reviewed with user experience in mind.

palashgdev:
blocked over https://github.com/SigNoz/signoz/issues/4081

YounixM (approved):

PR #4059, opened 2023-11-24
fix: sort tooltip value based on value and highlight on hover

YounixM:


ankitnayan (dismissed):

palashgdev:
before and after screenshot is missing there is no issue which we can look into

palashgdev (approved):

Rajat-Dabade (approved):

PR #4068, opened 2023-11-25
fix: use window function lagInFrame for rate calculation

srikanthccv:
Fixes https://github.com/SigNoz/engineering-pod/issues/1084


srikanthccv:
I verified by comparing the side-by-side version of this window function vs the running difference in production. The query timing is mostly the same. The window function is guaranteed to produce correct results all the time and customers won't see something like this https://signoz-team.slack.com/archives/C060B97HFRN/p1700744154708009

srikanthccv:
I would like to get this merged and released.

srikanthccv:
Please review

ankitnayan (approved):

Issue #4032, opened 2023-11-23
Threshold in Alerts UI doesn't work


srikanthccv:
## Bug description

Please see https://www.loom.com/share/16e652d6fe5845429c366344aef77884

PR #4074, opened 2023-11-27
[Fix]: threshold in alerts

Rajat-Dabade:
Close https://github.com/SigNoz/signoz/issues/4032

Loom: https://www.loom.com/share/3c483f18b6e948df9a7f686336ef2a8a

### DIFF
--- a/frontend/src/container/FormAlertRules/ChartPreview/index.tsx
+++ b/frontend/src/container/FormAlertRules/ChartPreview/index.tsx
@@ -124,10 +124,11 @@ function ChartPreview({
 						keyIndex: 0,
 						moveThreshold: (): void => {},
 						selectedGraph: PANEL_TYPES.TIME_SERIES, // no impact
-						thresholdValue,
+						thresholdValue: threshold,
 						thresholdLabel: `${t(
 							'preview_chart_threshold_label',
 						)} (y=${thresholdValue} ${query?.unit || ''})`,
+						thresholdUnit: alertDef?.condition.targetUnit,
 					},
 				],
 			}),
@@ -136,8 +137,10 @@ function ChartPreview({
 			queryResponse?.data?.payload,
 			containerDimensions,
 			isDarkMode,
+			threshold,
 			t,
 			thresholdValue,
+			alertDef?.condition.targetUnit,
 		],
 	);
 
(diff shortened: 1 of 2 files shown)


frontend/src/lib/getConvertedValue.ts
```diff
@@ -268,7 +270,19 @@ export function convertValue(
...
 	targetUnit: string,
 ): number | null {
-	if (targetUnit === 'none') {
+	if (
+		targetUnit === 'none' ||
+		!currentUnit ||
+		!targetUnit ||
+		currentUnit === targetUnit
+	) {
+		if (
```
YounixM, line 279: Are we handling `percentage`? What about other units ?

Rajat-Dabade: The only problem was for `percentage` unit, every other unit is working fine.

srikanthccv: I think it wasn't working for other units as well. Let me check and share video.

srikanthccv:
These are the main issues I could find

- Can the value be short form? For example, here 4GB instead of 4000000000 bytes
   
Screenshot 2023-11-28 at 3 15 20 PM

- The threshold line doesn't appear if the threshold value is more than the max value in the chart i.e. The following chart has around ~ 9 GB as max but If I set the threshold as 15GB the threshold doesn't appear. 
Screenshot 2023-11-28 at 3 17 51 PM
Screenshot 2023-11-28 at 3 18 01 PM


- The threshold line does exist in edit mode for percent unit but works for other units 
   
Screenshot 2023-11-28 at 3 21 13 PM

No threshold line even though there is threshold

Screenshot 2023-11-28 at 3 21 44 PM

srikanthccv (dismissed):

srikanthccv:
Tested

- The dashboard widgets work when there are NaN in the result
- Services tab charts works with regular data and also NaN
- The three points mentioned here https://github.com/SigNoz/signoz/pull/4074#issuecomment-1829467279

frontend/src/container/FormAlertRules/ChartPreview/index.tsx
```diff
@@ -110,6 +105,9 @@ function ChartPreview({
 
 	const isDarkMode = useIsDarkMode();
 
+	const optionName =
+		getFormatNameByOptionId(alertDef?.condition.targetUnit || '') || '';
```
YounixM, line 109: Should we be calling the function only if the targetUnit exists?

YounixM (approved):

Issue #4037, opened 2023-11-23
Alert edit data doesn't reflect unless page refreshed


srikanthccv:
## Bug description

See https://www.loom.com/share/5ba8ee6734a449fab07ca92acc1bf819

PR #4076, opened 2023-11-27
fix: updated the form value on mount

palashgdev:
Close https://github.com/SigNoz/signoz/issues/4037

before:

https://github.com/SigNoz/signoz/assets/88981777/519cc444-1f0d-447e-be78-c151b03eb005

after: 

https://github.com/SigNoz/signoz/assets/88981777/96af09ce-5481-4829-8b1f-31d1c187e8a8

### DIFF
--- a/frontend/src/container/EditAlertChannels/index.tsx
+++ b/frontend/src/container/EditAlertChannels/index.tsx
@@ -22,7 +22,7 @@ import {
 import FormAlertChannels from 'container/FormAlertChannels';
 import { useNotifications } from 'hooks/useNotifications';
 import history from 'lib/history';
-import { useCallback, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router-dom';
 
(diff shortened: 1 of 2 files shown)


srikanthccv (dismissed):

Rajat-Dabade:
@palashgdev, In the video demonstration, an observation wherein, upon clicking the "save" button, the displayed value momentarily reverts to its previous state. Similarly, when the "edit" button is clicked, there is a brief period during which the value appears as it was before. The cause of this behaviour is uncertain, I suspect it is because of the `useEffect` not very sure.

palashgdev:
> @palashgdev, In the video demonstration, an observation wherein, upon clicking the "save" button, the displayed value momentarily reverts to its previous state. Similarly, when the "edit" button is clicked, there is a brief period during which the value appears as it was before. The cause of this behaviour is uncertain, I suspect it is because of the `useEffect` not very sure.

thanks component is mounted though isLoading is replaced with isFetching which fixed the issue

YounixM (approved):

Rajat-Dabade (approved):
LGTM 🚀

Issue #4081, opened 2023-11-27
FE: body overflow issue


palashgdev:
if we remove the body overflow hidden from the body

white background is added which should not be there would recommend to remove the white background but good to have overflow hidden in the body

reference: https://www.loom.com/share/d7541d1a7b7d43cb9bcec368fe544a32

PR #4092, opened 2023-11-27
fix: [GH-4081]: no whitespace should appear when we remove hidden from body styles

Vikrant2520:
- The presence of `,` in the app file was rendering it at the end hence causing the scroll 

https://github.com/SigNoz/signoz/assets/54737045/684f8624-e625-44b1-995e-1b606ccf4037

This PR solves #4081


palashgdev (approved):
Thanks @Vikrant2520

Issue #4109, opened 2023-11-29
[FE] Where clause buggy on same value as suggestions


makeavish:
https://www.loom.com/share/3e9ad809b19d4dd8ac1c62f110bda026?sid=e144b734-3a04-48fc-af16-9ca4727833cc

srikanthccv:
Fixed with new query builder.

PR #4120, opened 2023-11-30
Fix: lettercase bug in query-builder

s4nju:
### Summary
This PR solves the issue mentioned in #4109


makeavish (approved):

makeavish:
This PR misses edge cases when keys has Upper case values

s4nju:
@makeavish can you please provide example for which this is not working?

makeavish:
Please check case when we select `dbName` key in filter, we can't select suggestions

Screenshot 2023-12-19 at 16 59 25

s4nju:
Got it. Should I raise another PR for this? or make changes in the previous one? 😕

makeavish:
> Got it. Should I raise another PR for this? or make changes in the previous one? 😕

Create another PR, if that works then we won't merge the revert PR https://github.com/SigNoz/signoz/pull/4255

Issue #3681, opened 2023-10-06
LOGS: [FE] Dot support in frontend


nityanandagohain:
* if attribute name is `test.key` i.e it contains dot then only `test` is displayed in detail page
* filtering from the detail page also doesn't work for these attributes.

REF https://github.com/SigNoz/signoz/issues/3553

nityanandagohain:
@YounixM  can we prioritise this cc @ankitnayan

YounixM:
Assgined to @raj-k-singh as he will be picking this.

PR #4121, opened 2023-11-30
fix: dot support in attribute name

### LINKED ISSUE #3553: [LOGS] Use materialized column if present for JSON search.

nityanandagohain:
### Summary

This PR adds the following support
* Proper formatting of attributes names with `.` in order by and group by
* In logs lines it doesn't return the `_` attributes.
* When a materialized column is created for an attribute with `.` in name it is replaced with `$$`. Eg:- `request.method` becomes `attribute_string_request$$method`
* Doesn't return the duplicate underscore attributes in the getFields API
* For checking exists we used indexOf(attr, key) > 0. Replaced it with has(attr, key) to make it consistant.
* For update fields endpoint use `default` instead of materialzied

#### Related Issues / PR's

Fixes #3553 

please look at https://github.com/SigNoz/signoz-otel-collector/pull/233 first

clickhouse migration : https://github.com/SigNoz/signoz-db-migrations/pull/8

Frontend changes required for this release https://github.com/SigNoz/signoz/issues/3681

### Tested
* For both index and non indexed values with and without dot in name
  * filtering
  * group by 
  * order by is working
  * Formulas are working
* Enrichment of queries are working fine
* json queries are working as expected with no interference
* limit queries

### Blast radius
* Alerts and Dhashboards and saved filters
* If in any of the above the attribute name was actually containing `.` it must be updated within a month else it will start failing.


srikanthccv:
Test the formulas once.

nityanandagohain:
Have tested for formulas.

Also update the code to change the update fields fuctions to use Default instead of materialized.

srikanthccv:
Can you add a test for formula query?

nityanandagohain:
Thanks, @srikanthccv , yeah I tested only filters along with formulas, when testing in a group by it broke. Have fixed and added a test as well.

srikanthccv:
LGTM now; I can't of think any other case where it can fail. I will take another look and approve.

nityanandagohain:
Thanks, appreciate it

srikanthccv (approved):
Please check time series chart dot attribute names with a limit applied once. Otherwise LGTM

nityanandagohain:
Yeah, right that got left out , have fixed it

srikanthccv (approved):

pranay01:
Merged based on Nitya's request

PR #4131, opened 2023-11-30
fix: update logic for handling data for uplot charts

YounixM:
### Summary

Issues: 
1. Data missing in the plotted graph
2. The tooltip shows 0 values but the graph doesn't plot 0 points in the graph.

Before:
image (2)

After:
#### Screenshots
image

RCA: 
The fillMissingTimestamps function was not handling the subsetArray values correctly and introduced null values even when the response had data in few cases. (Mostly in cases when there is deviation amongst the timestamps across different series)

Rewrote the logic to handle data for Uplot as below.

1. Computing values for x-axis points 
    *  Loop through all the series data to create a unique set of timestamps. (previously we were relying on a single series to 
        define our x-axis points)
 2. Filling missing timestamps and corresponding values for all series.
     * Loop through the x-axis timestamps array and series [timestamp, value] array and check if all x-axis timestamps exist in the series timestamps array. If the timestamp doesn't exist in the array, add the timestamp to the series with the value to be `0` or `null` based on the fillSpans value.
3. Maintain a counter for tooltip values and show the tooltip only if a value exists. 

`We need to figure out how to handle 0 values in tooltips as 0 values are not plotted in the graph`

### DIFF
--- a/frontend/src/container/NewWidget/RightContainer/index.tsx
+++ b/frontend/src/container/NewWidget/RightContainer/index.tsx
@@ -107,48 +107,8 @@ function RightContainer({
 				}
 			/>
 
-			{/* <TextContainer>
-				<Typography>Stacked Graphs :</Typography>
-				<Switch
-					checked={stacked}
-					onChange={(): void => {
-						setStacked((value) => !value);
-					}}
-				/>
-			</TextContainer> */}
-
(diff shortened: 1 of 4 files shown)


palashgdev (approved):

palashgdev:
@Rajat-Dabade can you also verify this one

Rajat-Dabade (approved):
LGTM 🚀

PR #4166, opened 2023-12-06
chore: add prepare query for cumulative/unspecified timeseries

srikanthccv:
### Summary

Part 2 of #4016 

The following example illustrates how the sub-query is used to calculate the sume of values for each
time series in a 15 seconds interval:

### Sum, Avg, Min, Max etc.. 

```
timestamp    01.00  01.05  01.10  01.15  01.20  01.25  01.30  01.35  01.40
            +------+------+------+------+------+------+------+------+------+
            |      |      |      |      |      |      |      |      |      |
            |  v1  |  v2  |  v3  |  v4  |  v5  |  v6  | v7   |  v8  |  v9  |
            |      |      |      |      |      |      |      |      |      |
            +------+------+------+------+------+------+------+------+------+
               |      |      |      |      |      |       |     |      |
                 |    |    |          |    |    |           |   |    |
                      |                    |                    |
                   +------+             +------+            +------+
                   | v1+  |             |  v4+ |            |  v7+ |
                   | v2+  |             |  v5+ |            |  v8+ |
                   | v3   |             |  v6  |            |  v9  |
                   +------+             +------+            +------+
                    01.00                01.15               01.30
```

Calculating the rate/increase involves an additional step. We first calculate the maximum value for each time series
in a 15 seconds interval. Then, we calculate the difference between the current maximum value and the previous
maximum value

### Rate, Increase
The following example illustrates how the sub-query is used to calculate the rate of change for each time series
in a 15 seconds interval:

```
timestamp    01.00  01.05  01.10  01.15  01.20  01.25  01.30  01.35  01.40
            +------+------+------+------+------+------+------+------+------+
            |      |      |      |      |      |      |      |      |      |
            |  v1  |  v2  |  v3  |  v4  |  v5  |  v6  | v7   |  v8  |  v9  |
            |      |      |      |      |      |      |      |      |      |
            +------+------+------+------+------+------+------+------+------+
               |      |      |      |      |      |       |     |      |
                 |    |    |          |    |    |           |   |    |
                      |                    |                    |
                   +------+             +------+            +------+
               max(| v1,  |         max(|  v4, |        max(|  v7, |
                   | v2,  |             |  v5, |            |  v8, |
                   | v3   |)            |  v6  |)           |  v9  |)
                   +------+             +------+            +------+
                    01.00                01.15               01.30

                             +-------+             +--------+
                             | V6-V2 |             |  V9-V6 |
                             |       |             |        |
                             |       |             |        |
                             +------+              +--------+
                               01.00                  01.15
```

The rate of change is calculated as (Vy - Vx) / (Ty - Tx) where Vx and Vy are the values at time Tx and Ty respectively.
In an ideal scenario, the last value of each interval could be used to calculate the rate of change. Instead, we use
the maximum value of each interval to calculate the rate of change. This is because any process restart can cause the
value to be reset to 0. This will produce an inaccurate result. The max is the best approximation we can get.
We don't expect the process to restart very often, so this should be a good approximation.

### ClickHouse Functions to understand

See https://clickhouse.com/docs/en/sql-reference/window-functions for more details on `lagInFrame` function

### Explanation for `rateWithoutNegative` and `increaseWithoutNegative` consts

Calculating the rate of change of a metric is a common use case.
Requests and errors are two examples of metrics that are often expressed as a rate of change.
The rate of change is the difference between the current value and the previous value divided by
the time difference between the current and previous values (i.e. the time interval).

The value of a cumulative counter always increases. However, the rate of change can be negative
if the value decreases between two samples. This can happen if the counter is reset when the
application restarts or if the counter is reset manually. In this case, the rate of change is
not meaningful and should be ignored.

The condition `(per_series_value - lagInFrame(per_series_value, 1, 0) OVER rate_window) < 0`
checks if the rate of change is negative. If it is negative, the value is replaced with `nan`.

The condition `ts - lagInFrame(ts, 1, toDate('1970-01-01')) OVER rate_window) >= 86400` checks
if the time difference between the current and previous values is greater than or equal to 1 day.
The first sample of a metric is always `nan` because there is no previous value to compare it to.
When the first sample is encountered, the previous value for the time is set to default i.e `1970-01-01`.
Since any difference between the first sample timestamp and the previous value timestamp will be
greater than or equal to 1 day, the rate of change for the first sample will be `nan`.

If neither of the above conditions is true, the rate of change is calculated as
`(per_series_value - lagInFrame(per_series_value, 1, 0) OVER rate_window) / (ts - lagInFrame(ts, 1, toDate('1970-01-01')) OVER rate_window)`
where `rate_window` is a window function that partitions the data by fingerprint and orders it by timestamp.
We want to calculate the rate of change for each time series, so we partition the data by fingerprint.

The `increase` function is similar to the `rate` function, except that it does not divide by the time interval.


pkg/query-service/app/metrics/v4/cumulative/helper.go
```diff
@@ -0,0 +1,57 @@
...
+	return strings.Join(groupTags, ", ")
+}
+
+// orderBy returns a string of comma separated tags for order by clause
+// if the order is not specified, it defaults to ASC
+func orderByAttributeKeyTags(items []v3.OrderBy, tags []v3.AttributeKey) string {
+	var orderBy []string
+	for _, tag := range tags {
+		found := false
+		for _, item := range items {
```
nityanandagohain, line 42: if the value of len(tags) * len(items) is higher we can use a lookup.

srikanthccv: They are usually lists with a single digit length and should not exceed a few hundred operations. I prefer to keep the dumb version here since it's not performance sensitive.

dhawal1248 (approved):

PR #4168, opened 2023-12-06
chore: add prepare query for delta/unspecified table

srikanthccv:
### Summary

Part 4 of #4016 

The `table` panel type is almost the same as the time series panel but the line chart values in the timeframe are reduced to represent one value for each row. The reduced aggregation is usually (and by default) will be `avg`. In the new metrics improvements, We do not want to apply the reduced aggregation in the ClickHouse query itself as it is not cache friendly. We fetch the values for each result series and aggregate them in query-service (to come in follow-up PRs) before sending the response.


pkg/query-service/app/metrics/v4/delta/timeseries.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 // prepareTimeAggregationSubQueryTimeSeries builds the sub-query to be used for temporal aggregation
-func prepareTimeAggregationSubQueryTimeSeries(start, end, step int64, mq *v3.BuilderQuery) (string, error) {
+func prepareTimeAggregationSubQuery(start, end, step int64, mq *v3.BuilderQuery) (string, error) {
```
ankitnayan, line 13: this change should be at [cumulative/timeseries.go](https://github.com/SigNoz/signoz/tree/7d960b79dd34ae8413b2dab0a10224eec64b34a1/pkg/query-service/app/metrics/v4/cumulative) too?

srikanthccv: It will be done in upcoming https://github.com/SigNoz/signoz/pull/4169

srikanthccv:
Yes, your understanding it correct.

ankitnayan (approved):

PR #4170, opened 2023-12-06
chore: add querier v2

srikanthccv:
### Summary

Part 6 of #4016 

#### Screenshots

NA

### DIFF
--- a/pkg/query-service/app/metrics/v4/query_builder.go
+++ b/pkg/query-service/app/metrics/v4/query_builder.go
@@ -3,8 +3,11 @@ package v4
 import (
 	"fmt"
 	"strings"
+	"time"
 
+	metricsV3 "go.signoz.io/signoz/pkg/query-service/app/metrics/v3"
 	"go.signoz.io/signoz/pkg/query-service/constants"
+	"go.signoz.io/signoz/pkg/query-service/model"
 	v3 "go.signoz.io/signoz/pkg/query-service/model/v3"
 	"go.signoz.io/signoz/pkg/query-service/utils"
 )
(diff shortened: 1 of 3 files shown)


pkg/query-service/app/metrics/v4/query_builder.go
```diff
@@ -3,8 +3,11 @@ package v4
 import (
 	"fmt"
 	"strings"
+	"time"
 
+	metricsV3 "go.signoz.io/signoz/pkg/query-service/app/metrics/v3"
```
ankitnayan, line 8: Should we import just the `options` from `metrics/v3` or create a separate options for `v4`?

pkg/query-service/app/querier/v2/helper.go
```diff
@@ -0,0 +1,306 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	logsV3 "go.signoz.io/signoz/pkg/query-service/app/logs/v3"
```
ankitnayan, line 11: will logs and traces continue to use `v3` even in `querier/v2`?

ankitnayan (approved):

pkg/query-service/app/querier/v2/helper.go
```diff
@@ -0,0 +1,306 @@
...
+	metricsV3 "go.signoz.io/signoz/pkg/query-service/app/metrics/v3"
+	metricsV4 "go.signoz.io/signoz/pkg/query-service/app/metrics/v4"
+	tracesV3 "go.signoz.io/signoz/pkg/query-service/app/traces/v3"
+	"go.signoz.io/signoz/pkg/query-service/cache/status"
+	"go.signoz.io/signoz/pkg/query-service/constants"
+	v3 "go.signoz.io/signoz/pkg/query-service/model/v3"
```
ankitnayan, line 17: Is it recommended to use create a new model per version?

pkg/query-service/app/metrics/v4/query_builder.go
srikanthccv: This was required to make both v3 and v4 co-exist.

pkg/query-service/app/querier/v2/helper.go
srikanthccv: Yes, they will continue to use v3.

pkg/query-service/app/querier/v2/helper.go
srikanthccv: Yes, that would be an ideal approach. We are using the same model for traces and logs as well. If we were to add a new version it would just apply for metrics but no change in the traces & logs. I was aiming for co-existence that doesn't confuse or require any changes from the traces/logs side.

ankitnayan: ok

pkg/query-service/app/metrics/v4/query_builder.go
ankitnayan: ok

Issue #1855, opened 2022-12-12
Support for secure ClickHouse server


prashant-shahi:
This would make sure of secure connection between the services and ClickHouse.
Also, we would support ClickHouse Cloud, which provides secure ClickHouse server.

PR #4178, opened 2023-12-06
Add basic support for secure clickhouse connections

features-not-bugs:
### Summary

Add basic support for secure clickhouse connections using DSN queries

#### Related Issues / PR's

https://github.com/SigNoz/signoz/issues/1855

#### Screenshots

NA

Query service package will need to be tested to ensure three additional connection combinations work as intended, connecting with TLS Enabled and with either TLS verification enabled or disabled, although no real logic has been written here, it's simply adding to clickhouse-go's configuration struct

## Summary by CodeRabbit

- **Refactor**
	- Improved the handling of ClickHouse connection options for enhanced flexibility and clarity.
	- Simplified the initialization of the `APIHandler` by removing certain fields, making the service configuration more straightforward.
	- Updated default values and descriptions for connection parameters in the main service configuration to reflect the new handling approach.


pkg/query-service/app/clickhouseReader/options.go
```diff
@@ -108,6 +110,11 @@ func defaultConnector(cfg *namespaceConfig) (clickhouse.Conn, error) {
 		}
 		options.Auth = auth
 	}
+	if strings.ToLower(dsnURL.Query().Get("secure")) == "true" {
```
SpencerTorres, line 113: I suggest using [`strconv.ParseBool`](https://pkg.go.dev/strconv#ParseBool) for this since it handles the casing and matches more than just `true`

For reference, [here is how these options are parsed](https://github.com/ClickHouse/clickhouse-go/blob/a6e001eb87398dec2062dfbe02f27b9168c51fd3/clickhouse_options.go#L242) in the `fromDSN` function from `clickhouse-go`

srikanthccv:
Please use ParseDSN from clickhouse https://github.com/ClickHouse/clickhouse-go/blob/5e9073dcd69a1082d59b5ee49b43534d82b5fb34/clickhouse_options.go#L110-L116

features-not-bugs:
> Please use ParseDSN from clickhouse https://github.com/ClickHouse/clickhouse-go/blob/5e9073dcd69a1082d59b5ee49b43534d82b5fb34/clickhouse_options.go#L110-L116

This makes much more sense, I've made the changes as requested and removed the redundant fields from configuration objects and cmd flags that we're being propagated down to the clickhouseReader.

features-not-bugs:
Merged the latest commits from develop in, project builds successfully

akalitenya:
@srikanthccv Is it planned to be completed (merged) for next release?

srikanthccv:
Sorry I missed the last update. This will be a breaking change. Let me think about this and get back soon but can't promise if this will be merged in next release.

pkg/query-service/main.go
```diff
@@ -39,18 +38,11 @@ func main() {
...
-	var dialTimeout time.Duration
-
 	flag.StringVar(&promConfigPath, "config", "./config/prometheus.yml", "(prometheus config to read metrics)")
 	flag.StringVar(&skipTopLvlOpsPath, "skip-top-level-ops", "", "(config file to skip top level operations)")
 	flag.BoolVar(&disableRules, "rules.disable", false, "(disable rule evaluation)")
 	flag.BoolVar(&preferDelta, "prefer-delta", false, "(prefer delta over cumulative metrics)")
 	flag.BoolVar(&preferSpanMetrics, "prefer-span-metrics", false, "(prefer span metrics for service level metrics)")
-	flag.IntVar(&maxIdleConns, "max-idle-conns", 50, "(number of connections to maintain in the pool.)")
-	flag.IntVar(&maxOpenConns, "max-open-conns", 100, "(max connections for use at any time.)")
-	flag.DurationVar(&dialTimeout, "dial-timeout", 5*time.Second, "(the maximum time to establish a connection.)")
```
srikanthccv, line 53: The default values used here should be used in `defaultConnector` parse logic if they are not set.

features-not-bugs: There doesn't seem to be any defined "Defaults" in the clickhouse library, although there is some logic checks in the setDefaults function on the clickhouse_options.go file that do basic out of range logic checks.

I've followed this semantic by setting the flag defaults to 0 of which will trigger the clickhouse defaults to be set.

srikanthccv: No, we don't want the ClickHouse-go defaults. They are really small and throw error for non-trival dashboards. And when the number of alerts is high. We have found the current settings to be effective and they should be the defaults.

features-not-bugs: Ok,

I've added in logic to use the Flags only if the DSN does not explicitly set them. This will ensure the DSN configuration will be respected first and then the flags thereafter.

srikanthccv:
thanks, please resolve the conflicts

features-not-bugs:
> thanks, please resolve the conflicts

Done

srikanthccv (approved):

Issue #3645, opened 2023-09-28
QB: [FE] Support for operators in smallcase.


nityanandagohain:
We should support operators in small case, example `CONTAINS` and `contains` . Both should work

srikanthccv:
You created this last week https://github.com/SigNoz/signoz/issues/3585

nityanandagohain:
Both are different @srikanthccv

srikanthccv:
Help me understand how they are different. There won't be an attribute in any data source containing the white space (this was even an assumption written in query builder PRD, but the frontend didn't handle this), so it is generic. The mentioned expectation in the linked issue is when a user enters `eamil ex`, the suggestions should show `email exists`, which is based on the `ex` matches with lowercase `EXISTS`, and it should be handled. If the linked issue gets fixed, this would be addressed as well. It may have the logs label, but this issue on the frontend is not separate. I am writing this detail so they won't be treated as separate (unless I missed something).

nityanandagohain:
Yeah makes sense 👍

palashgdev:
@nityanandagohain what will happen if the key contains the operator 

```
eg: exin
```

here key would ex and it would be the operator I don't think this would scale well

nityanandagohain:
`exin` is the key only only after the first space we can check for operators

srikanthccv:
> only after the first space we can check for operators

This is not mandatory, and shouldn't be because users write operators right after the key sometimes and that should be supported, for example `trace_id=r4fv...`.

srikanthccv:
We don't support it for operators that are text (LIKE/CONTAINS) but support symbols such as `=/!=/><`, which makes sense since the symbols won't be part of the attributes key. However, the operator text can appear in the attribute key itself, so we expect the user to use the separator there.

nityanandagohain:
Agree, for symbols it should work without space.

nityanandagohain:
Reopening this, since it wasted a lot of time and had to be reverted. Better to work on this with basic unit tests as the other thing was breaking as one thing changed.

PR #4179, opened 2023-12-06
feat: [GH-3645]: case insensitive operators for query builder

vikrantgupta25:
### Summary

- updated the tag regex to match the `case-insensitive` operators and convert them in upper case as the entire codebase handling depends on the same. 
- this takes care of suggestions and query

#### Related Issues / PR's

This solves #3645 

#### Screenshots

NA


palashgdev:
suggested options are still in upper case can you please confirm ?

Vikrant2520:
@palashgdev the recommendations for operators should be uni-case as that is the standard format , as soon as we enter case insensitive operator then the further recommendations work as expected.  CC :- @nityanandagohain

palashgdev:
ok then can you please confirm this one 

https://github.com/SigNoz/signoz/assets/88981777/8f7dfa95-8b14-469c-8975-d1d7f49c1166

nityanandagohain (changes requested):
* Whenver we select an  operator like `=` from the UI the value suggestion is not working, screenshot attached.
* Found the same issue as palash reported, we can maybe suggest values for not symbolic operators only after space is added.
image

Vikrant2520:
@nityanandagohain i have addressed the first issue you posted, discussing about the second in the channel

PR #4190, opened 2023-12-07
fix: incorrect alert description and summary for prom rules

srikanthccv:
### Summary

The current logic concatenates the threshold value (0.4) and target unit text (%). This is fine for units such as data (10 gbytes) and time but the percentages need conversion. The fix is to use a formatter that cares about such details. 

Fixes https://github.com/SigNoz/engineering-pod/issues/1132.

### DIFF
--- a/pkg/query-service/rules/promRule.go
+++ b/pkg/query-service/rules/promRule.go
@@ -3,7 +3,6 @@ package rules
 import (
 	"context"
 	"fmt"
-	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


srikanthccv:
@ankitnayan easy one. please review.

srikanthccv:
Yes, this is the correct usage. When the issue was originally reported we only fixed it for the builder in this PR https://github.com/SigNoz/signoz/pull/4047. The same issue existed in PromQL queries till now. It should be working now for all types of queries.

PR #4191, opened 2023-12-08
chore: add variable templating endpoint

srikanthccv:
### Summary

Part of https://github.com/SigNoz/engineering-pod/issues/1119


srikanthccv:
@palashgdev, this PR adds an endpoint to get the query data back with formatted/replaced templated variables. You need to use this before redirecting the user from the panel to alert using the URL. Please see https://github.com/SigNoz/engineering-pod/issues/1119 for the issue with the current experience.

palashgdev (approved):

Issue #4202, opened 2023-12-11
[Issue]: Time series chart starting from 0


Rajat-Dabade:
Time series chart starting from 0 even if there is no point in the chart with value 0. 

![Screenshot 2023-12-09 at 09 25 18](https://github.com/SigNoz/signoz/assets/26198062/8941337d-71f5-4548-9c2a-a4133f8bf358)

Issue #3561, opened 2023-09-15
[Feat]: Set Min and Max range in charts


Rajat-Dabade:
Default chart behaviour configures the y-axis based on data point ranges corresponding to x-axis timestamps. However, some users require the ability to set the y-axis to begin at zero. Similar to Grafana's functionality, addressing this issue can be achieved by defining minimum and maximum limits for the y-axis.

Loom: https://www.loom.com/share/f664097cf0974f199d53385df83319c2?sid=4394c217-2dd5-4c82-88bc-d076ca348e54

PR #4203, opened 2023-12-11
[Fix]: the min data point as 0 in time series

Rajat-Dabade:
Before: 

Screenshot 2023-12-11 at 2 43 57 PM

Before with threshold: 
Screenshot 2023-12-11 at 2 44 18 PM

After:

Screenshot 2023-12-11 at 2 44 50 PM

After with threshold: 
Screenshot 2023-12-11 at 2 45 49 PM

Fixes: https://github.com/SigNoz/signoz/issues/4202

Docs: https://www.notion.so/signoz/Logic-for-range-in-uplot-95dbc85dd53d43769fdc0e126b717b32?pvs=4


palashgdev:
@Rajat-Dabade it will be good if we can control this one with dashboard json in future?

srikanthccv:
Yes, that would be a good addition. What we are _fixing_ in this is the desired behaviour for some users https://signoz-team.slack.com/archives/C05AXQL92BT/p1694622771188419. If it's easier to make this configurable let's do it. If it is non-trivial let address it in separate PR.

Rajat-Dabade:
@srikanthccv  I believe storing the minimum and maximum range in the dashboard JSON is not advisable since the values of series data can change dynamically (unless we are computing the min and max from the backend). On the based series data and threshold data which are more likely to change every time, we have to compute the min and max for the chart every time. 

Instead, the softmin and softmax, which will be introduced shortly, should be stored in the dashboard JSON where the user can specify the min and max value they want to chart to be. 

cc: @palashgdev

palashgdev:
> @srikanthccv I believe storing the minimum and maximum range in the dashboard JSON is not advisable since the values of series data can change dynamically (unless we are computing the min and max from the backend). On the based series data and threshold data which are more likely to change every time, we have to compute the min and max for the chart every time.
> 
> Instead, the softmin and softmax, which will be introduced shortly, should be stored in the dashboard JSON where the user can specify the min and max value they want to chart to be.
> 
> cc: @palashgdev

I am not recommending storing min, and max in the dashboard JSON all we need to store is whether we want to go ahead with min zero or not it's more kind of a flag

srikanthccv:
Please add your points related to min and max here instead https://github.com/SigNoz/signoz/issues/3561. Let's not hold this PR for additional changes.

YounixM (approved):

PR #4205, opened 2023-12-11
fix: added the validation of query params in validateQueryRangeParams V3 for start,step,end

palashgdev:
Added the validation for step,start,end


palashgdev:
@srikanthccv can you roughly check this pr

srikanthccv:
Seems fine mostly but we don't make this mandatory for ClickHouse queries i.e queryType=clickhouse_sql because they are not needed there.

palashgdev:
> Seems fine mostly but we don't make this mandatory for ClickHouse queries i.e queryType=clickhouse_sql because they are not needed there.

i see if we make step=0 or we skip sending then validateQueryRangeParamsV3 is breaking

PR #4237, opened 2023-12-15
fix: do not use grouping sets when there is no tags for group by

srikanthccv:
The recent upgrade to ClickHouse started behaving incorrectly for specific conditions. There is no need to use GROUPING SETS when there is no group by attributes.


srikanthccv:
@ankitnayan The recent clickhouse version returns incorrect results occasionally for grouping sets when there is no group by key other than the timestamp https://signoz-team.slack.com/archives/C05GQNXL3FG/p1702573925564949

ankitnayan (approved):

PR #4240, opened 2023-12-16
fix: enrichment for json in alerts and disable timestamp adjustment

nityanandagohain:
* disable timestamp roundoff for logs entirely
* json enrichment for alerts.


srikanthccv (approved):

srikanthccv:
@nityanandagohain fix the tests

Issue #4180, opened 2023-12-07
[Log Parsing Pipelines] Processor parse from field path containing `$` breaks collector config.


raj-k-singh:
```
[...]

PR #4242, opened 2023-12-16
Fix: query service: escape dollars in collector config generated for pipelines.

raj-k-singh:
### Summary
Right now, using pipeline field values containing $ (eg: `attributes.$data`) breaks collector config.
This happens because the collector attempts to expand tokens starting with $ as env variables and interpolates them to empty string.

The other issue was that pipeline previews were working just fine.
If the preview had surfaced this error, the user wouldn't have had to find out after saving and deploying the pipeline.

This PR includes the following changes

- Updates collectorsimulator (pipeline preview) to use the same config provider as the actual collector (file provider with env var substitution extension). This will ensure such issues will not go uncaught by the preview.
- Unfies pipeline config generation logic across collector simulation and actual config recommendations, and escapes `$` by replacing it with `$$` in that logic.

#### Related Issues / PR's

Fixes https://github.com/SigNoz/signoz/issues/4180

### DIFF
--- a/frontend/src/hooks/queryBuilder/useOptions.ts
+++ b/frontend/src/hooks/queryBuilder/useOptions.ts
@@ -4,6 +4,7 @@ import {
 } from 'container/QueryBuilder/filters/QueryBuilderSearch/utils';
 import { Option } from 'container/QueryBuilder/type';
 import { transformStringWithPrefix } from 'lib/query/transformStringWithPrefix';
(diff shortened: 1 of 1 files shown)


srikanthccv (approved):

srikanthccv:
Are there any other special symbols that we need to escape?

ankitnayan (approved):

PR #4252, opened 2023-12-19
FE: Log Detail View: Support for dots in log attributes

raj-k-singh:
### Summary

The log detail view doesn't display fields containing dot in their name correctly.
It is also not possible to filter by such fields from the log detail view.

#### Related Issues / PR's
Fixes https://github.com/SigNoz/signoz/issues/3681

#### Screenshots
Display of fields containing dot in name
Before:
Screenshot from 2023-12-19 10-35-31

After:
Screenshot from 2023-12-19 10-32-30

Filtering by field containing dot by clicking on field name
Before:
dot-support-in-click-to-filter-log-details-before

After:
dot-support-in-click-to-filter-log-details-after

Filtering by field containing dot in name from action menu
Before:
dot-support-in-filter-from-log-details-before

After:
dot-support-in-filter-from-log-details-after

Log detail view. Tested manually

### DIFF
--- a/frontend/src/container/LogDetailedView/TableView.tsx
+++ b/frontend/src/container/LogDetailedView/TableView.tsx
@@ -22,8 +22,8 @@ import { ILog } from 'types/api/logs/log';
 import ActionItem, { ActionItemProps } from './ActionItem';
 import FieldRenderer from './FieldRenderer';
 import {
+	filterKeyForField,
 	flattenObject,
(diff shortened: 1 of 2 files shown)


nityanandagohain (changes requested):
Displaying and filter in/out is fine, but when you click on the key in detail page then it is not correct in the filter.

YounixM (approved):

Issue #4261, opened 2023-12-20
Unable to plot the query for SigNozDown alert.


Rajat-Dabade:
https://signoz-team.slack.com/archives/C02TJ466H8U/p1702991462753209

<img width="959" alt="Screenshot 2023-12-20 at 5 27 11 PM" src="https://github.com/SigNoz/signoz/assets/26198062/546ab18e-bd28-4d46-b35e-de96318b4fd0">

![Screenshot 2023-12-19 at 6 51 18 PM](https://github.com/SigNoz/signoz/assets/26198062/5e97b62f-d53f-43db-bb14-bede09bfa93f)

PR #4262, opened 2023-12-20
[Fix]: range issue in uplot charts

Rajat-Dabade:
Fixes: https://github.com/SigNoz/signoz/issues/4261

Screenshot 2023-12-20 at 5 07 25 PM

### DIFF
--- a/frontend/src/lib/uPlotLib/utils/getYAxisScale.ts
+++ b/frontend/src/lib/uPlotLib/utils/getYAxisScale.ts
@@ -9,7 +9,7 @@ function findMinMaxValues(data: QueryDataV3[]): [number, number] {
 		entry.series?.forEach((series) => {
 			series.values.forEach((valueObj) => {
 				const value = parseFloat(valueObj.value);
-				if (!value) return;
+				if (Number.isNaN(value)) return;
 				min = Math.min(min, value);
 				max = Math.max(max, value);
 			});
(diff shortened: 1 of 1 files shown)


frontend/src/lib/uPlotLib/utils/getYAxisScale.ts
```diff
@@ -9,7 +9,7 @@ function findMinMaxValues(data: QueryDataV3[]): [number, number] {
 		entry.series?.forEach((series) => {
 			series.values.forEach((valueObj) => {
 				const value = parseFloat(valueObj.value);
-				if (!value) return;
+				if (Number.isNaN(value)) return;
```
YounixM, line 12: @Rajat-Dabade : Let compute min / max only if the value is of type number

Rajat-Dabade: That's what we are doing here. Series can have NaN along with numbers, so numbers should be compared and NaN's should be skipped.

YounixM: It is possible that the value can be of any type, making sure the value is of type number guarantees that there are no edge case breaks.

Rajat-Dabade: Sure.

YounixM (approved):

Issue #1327, opened 2022-06-28
Date for 2-3 minute at the edges is lost when plotting time series graphs using Query Builder/ClickHouse query


pranay01:
## Bug description
2-3 minutes lost at the edges when plotting time series graphs using Query Builder/ClickHouse query
[Loom Video](https://www.loom.com/share/ccde7ac0c809488b8329f59051810367)

## Expected behavior
The graph should plot the exact time range as selected by user

## How to reproduce

1. Go to Dashboards
2. Plot any query via Query Builder or ClickHouse query
3. Check the time stamps in graph vs expected time range

## Version information 
* **Signoz version**: v0.9.0-rc2
* **Browser version**: Chrome 100
* **Your OS and version**: MacOS 12.3.1
* **Your CPU Architecture**(ARM/Intel):  ARM

Issue #4116, opened 2023-11-30
Query Range, roundoff shouldn't discard 33 mins of data when step interval is 33 mins.


nityanandagohain:
As of now the query range api discard 33 mins of data when the step interval is 33 mins.

nityanandagohain:
For logs we have removed roundoff

srikanthccv:
@makeavish I see the same logic is used in traces, so only traces in pending now. Please take a look.

PR #4263, opened 2023-12-21
fix: adjust the start and end more accurately

srikanthccv:
Fixes https://github.com/SigNoz/signoz/issues/1327
Partially address https://github.com/SigNoz/signoz/issues/4116

- For rate queries, the first point gets discarded since the previous value is unknown. This PR moves the start time by one step interval backwards to accommodate it.
- The end timestamp is adjusted to the nearest minute or step interval whichever is less. We don't want to move it by step interval because for larger time ranges the step can be as big as 33 minutes and ignoring 33 minutes data is not acceptable.


pkg/query-service/app/metrics/v3/query_builder.go
```diff
@@ -447,10 +448,14 @@ func reduceQuery(query string, reduceTo v3.ReduceToOperator, aggregateOperator v
...
-	// adjust the start and end time to be aligned with the step interval
 	start = start - (start % (mq.StepInterval * 1000))
-	end = end - (end % (mq.StepInterval * 1000))
+	// if the query is a rate query, we adjust the start time by one more step
+	// so that we can calculate the rate for the first data point
+	if mq.AggregateOperator.IsRateOperator() && mq.Temporality != v3.Delta {
+		start -= mq.StepInterval * 1000
+	}
+	adjustStep := int64(math.Min(float64(mq.StepInterval), 60))
+	end = end - (end % (adjustStep * 1000))
```
ankitnayan, line 458: why do we need to adjust the end timestamp? What unexpected will happen if we don't?

ankitnayan (approved):

pkg/query-service/app/metrics/v3/query_builder.go
srikanthccv: The value drops suddenly at the last point because of the partial data. For instance, if we have an aggregation interval of 60 seconds and the end time is 12(h):59(m):16(s), the ideal way to aggregate is [59:16-58:16], [58:16, 57:16], ..... [29:16-28:16], but we use the `toStartOfInterval` which creates the [59:16-59:00], [59:00-58:00] .... [29:00-28:00]. The `[59:16-59:00]` period is incomplete and produces a small value for the partial duration that makes the line drop suddenly at the last point. This is why we round it to the nearest full interval. Let me know if this is not clear.

ankitnayan: So a user has to wait for an interval to be complete to see the value, right? And they cannot see the value as it is. I am trying to compare with candlesticks of finance exchanges where they show partial data too. But okay for now as experience is poor when users see a sudden drop. May be we could color the last datapoint differently so that users know it's being filled right now

ankitnayan: I am sensing the above logic for `end` still might cause seeing huge drop in the last interval data as you would aggregate to the last minute say in an step of 33 mins

srikanthccv: > So a user has to wait for an interval to be complete to see the value

A minute in the worst case. 

> I am sensing the above logic for end still might cause seeing huge drop in the last interval data as you would aggregate to the last minute say in an step of 33 mins.

I didn't understand the point. 

> Can you test this out once?

I tested this already. Here is a quick example https://go.dev/play/p/eSnl6SiBYXo

ankitnayan: >	// 1703754689000 - 28 December 2023 14:41:29
	fmt.Println(old(1703754689000, 1980)) // 28 December 2023 14:36:00
	fmt.Println(new(1703754689000, 1980)) // 28 December 2023 14:41:00

the last bucket will have 5 mins of data and will still show huge drop as compared to the second last result point?

srikanthccv: Yes, the second last point contains 33 minutes of data and the last point contains 5 minutes of data. Our earlier thinking was when people look at a large range i.e. 1 week they will probably be fine with not having the data for last incomplete interval. But that assumption can't be true for everyone so we are making it better adjusted end time.

ankitnayan: So, the adjusted end time does not solve for the sudden drop, right? Then should it be removed altogether?

ankitnayan (approved):

PR #4266, opened 2023-12-21
feat: show rate limit warning in services page when total RPS > 100

YounixM:
### Summary

Show rate limit warning on services page when total RPS > 100

Screenshot 2023-12-21 165527


srikanthccv:
This shouldn't be shown to everyone but only those who are on trial and haven't paid for a subscription. Also, I believe the number should be configurable. We run a configs service to enable things like this without the need to build and deploy if we want to change it. What's the use of configs service if it isn't used for things like this?

Rajat-Dabade (approved):
LGTM 🚀

YounixM:
Agreed @srikanthccv, this we wanted to push along with this release. So started off with this.

Rajat-Dabade (approved):

makeavish (approved):

Issue #4218, opened 2023-12-13
Show checkbox options for dashboard variables


srikanthccv:
## Is your feature request related to a problem?

>It'd be nice if the drop down for variables had a checkbox. This would be useful for when I want to select all but one metric. I'd select All, then unselect the 1 I don't want selected. Currently with Signoz today, I have to go through and select every single variable I need.

YounixM:
@Rajat-Dabade: What's the status of this task?

SagarRajput-7:
Done under this PR - https://github.com/SigNoz/signoz/pull/5191

pending for review and merge

srikanthccv:
@SagarRajput-7 The bug I mentioned here https://signoz-team.slack.com/archives/C02TJ466H8U/p1718194817205719?thread_ts=1718175690.454299&cid=C02TJ466H8U is still there and breaks the dashboard with a mix of single select and multi-select chained variables. It's best if you verify the changes yourself to avoid the back and forth. If you prefer communication on the GitHub issue/PR, I will be happy to do the same.

SagarRajput-7:
Thanks, @srikanthccv - I will check and fix this.

PR #4275, opened 2023-12-21
[Refactor]: added checkbox to select and deslect the variables

Rajat-Dabade:
Fixes: https://github.com/SigNoz/signoz/issues/4218

Loom: https://www.loom.com/share/33b9caaabd6741be84985eff1ab08b74?sid=80f1803a-f8c4-431f-b63f-f9964210ad32


YounixM:
@Rajat-Dabade: The opening and closing of the dropdown on select is not consistent. 

[] ALL - Spacing issue

Clicking on the name of the item should only select that item and deselect all other items.

srikanthccv:
@Rajat-Dabade in the video you explain how one could deselect one item and get the values updated. However, the common use case often involves selecting one or two items as opposed to deselecting. For instance, there are 30+ hosts - By default all hosts are selected and now I want to look at data for host-6. How do I do that? I was expecting clicking on `ALL` would toggle the selected status and then I can search and select the interested host but that didn't work.


> Clicking on the name of the item should only select that item and deselect all other items.

@YounixM Why should it deselect all other items? What would happen If I select one item and move one to select another item, it shouldn't deselect earlier item(s).

PR #4283, opened 2023-12-23
chore: fix the query range cache gaps

srikanthccv:
### Summary

Fixes https://github.com/SigNoz/signoz/issues/1988


pkg/query-service/app/querier/querier.go
```diff
@@ -160,11 +160,15 @@ func findMissingTimeRanges(start, end int64, seriesList []*v3.Series, fluxInterv
 		}
 	}
 
+	endMillis := time.Now().UnixMilli()
```
ankitnayan, line 163: can this be extracted into a function as I see multiple occurrences of this adjustment logic? If we need to change this again later, we should need not look for different places or miss it at some places in hurry

ankitnayan (approved):

pkg/query-service/app/querier/querier.go
srikanthccv: I can do that

Issue #4208, opened 2023-12-12
[Log Parsing Pipelines] Using attributes paths like `attributes["http.status.code"]` doesn't work


raj-k-singh:
When using dot separated names for attributes, it should be possible to use paths like attributes.http.status.code and attributes["http.status.code"]

PR #4284, opened 2023-12-24
Fix: QS: Log Pipelines: generate correct nil checks for operators referencing fields like attributes["http.status.code"]

raj-k-singh:
### Summary
When using dot separated names for attributes, it should be possible to use paths like `attributes["http.status.code"]`

This PR includes changes for generating nil checks for field refs with membership op like `attributes["http.status.code"]`
Before this change, the nil check generation logic did not work for such paths and would generate `attributes["http?.status?.code"]`

#### Related Issues / PR's

Fixes: https://github.com/SigNoz/signoz/issues/4208

Pipelines. Tested via the test suite and manually e2e

### DIFF
--- a/frontend/src/container/GridTableComponent/styles.ts
+++ b/frontend/src/container/GridTableComponent/styles.ts
@@ -19,5 +19,13 @@ export const WrapperStyled = styled.div`
 	& .ant-table {
 		flex: 1;
 		overflow: auto;
+
+		> .ant-table-container {
+			> .ant-table-content {
+				> table {
+					min-width: 99% !important;
+				}
+			}
+		}
 	}
 `;
--- a/frontend/src/lib/uPlotLib/getUplotChartOptions.ts
+++ b/frontend/src/lib/uPlotLib/getUplotChartOptions.ts
@@ -136,7 +136,18 @@ export const getUPlotChartOptions = ({
 							if (threshold.thresholdLabel) {
 								const text = threshold.thresholdLabel;
 								const textX = plotRight - ctx.measureText(text).width - 20;
-								const textY = yPos - 15;
+
+								const canvasHeight = ctx.canvas.height;
+								const yposHeight = canvasHeight - yPos;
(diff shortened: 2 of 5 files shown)


nityanandagohain:
For a key named "log.file.name" are we supporting attributes["log.file.name"] as well as attributes.log.file.name ? because that would be wrong and might create more confusion


https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/29696#issuecomment-1851056298

nityanandagohain (approved):

pkg/query-service/app/logparsingpipeline/pipelineBuilder.go
```diff
@@ -169,3 +214,151 @@ func cleanTraceParser(operator *PipelineOperator) {
...
+		return strings.ReplaceAll(
+			strings.ReplaceAll(path, "?.", "."), ".", "?.",
+		)
+	}
+
+	// Optional chaining before membership ops is not supported by expr.
+	// Eg: The field `attributes.test["a.b"].value["c.d"].e` can't be checked using
+	// the nil check `attributes.test?.["a.b"]?.value?.["c.d"]?.e != nil`
+	// This needs to be worked around by checking that the target of membership op is not nil first.
+	// Eg: attributes.test != nil && attributes.test["a.b"]?.value != nil && attributes.test["a.b"].value["c.d"]?.e != nil
```
antonmedv, line 237: I'm working on adding support for `foo?.["bar"]` in Expr right now. Will be done in the next version. ;)

antonmedv: As well as on mode in which all `.` operations will be set to `?.`.

antonmedv: Strange. I'm getting BinaryNode:
Screenshot 2024-01-16 at 16 47 22

PR #4293, opened 2023-12-27
fix: ignore timestamp from the subquery result for value panel

srikanthccv:
### Summary

We currently use the timestamp from the subquery for the _reduced_ value of select aggregation (last, sum, min, max). This works fine for single query and formulas if all the queries involved in the expression have data for the same time range. For certain cases the most recent data may not exist for all queries involved in formula expression. And join clause on different timestamps returns No Data https://signoz-team.slack.com/archives/C05AXQL92BT/p1703672801446009. 

This change fixes the issue by using `now()` as the common timestamp. The value of timestamp in the value type panel is irrelevant so this doesn't make any difference to correctness.

### DIFF
--- a/pkg/query-service/app/metrics/v3/query_builder.go
+++ b/pkg/query-service/app/metrics/v3/query_builder.go
@@ -427,15 +427,15 @@ func reduceQuery(query string, reduceTo v3.ReduceToOperator, aggregateOperator v
 	// chart with just the query value. For the quer
 	switch reduceTo {
 	case v3.ReduceToOperatorLast:
-		query = fmt.Sprintf("SELECT *, timestamp AS ts FROM (SELECT anyLastIf(value, toUnixTimestamp(ts) != 0) as value, anyIf(ts, toUnixTimestamp(ts) != 0) AS timestamp %s FROM (%s) %s)", selectLabels, query, groupBy)
+		query = fmt.Sprintf("SELECT *, now() AS ts FROM (SELECT anyLastIf(value, toUnixTimestamp(ts) != 0) as value, anyIf(ts, toUnixTimestamp(ts) != 0) AS timestamp %s FROM (%s) %s)", selectLabels, query, groupBy)
 	case v3.ReduceToOperatorSum:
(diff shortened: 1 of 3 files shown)


pkg/query-service/app/metrics/v3/query_builder.go
```diff
@@ -427,15 +427,15 @@ func reduceQuery(query string, reduceTo v3.ReduceToOperator, aggregateOperator v
 	// chart with just the query value. For the quer
 	switch reduceTo {
 	case v3.ReduceToOperatorLast:
-		query = fmt.Sprintf("SELECT *, timestamp AS ts FROM (SELECT anyLastIf(value, toUnixTimestamp(ts) != 0) as value, anyIf(ts, toUnixTimestamp(ts) != 0) AS timestamp %s FROM (%s) %s)", selectLabels, query, groupBy)
+		query = fmt.Sprintf("SELECT *, now() AS ts FROM (SELECT anyLastIf(value, toUnixTimestamp(ts) != 0) as value, anyIf(ts, toUnixTimestamp(ts) != 0) AS timestamp %s FROM (%s) %s)", selectLabels, query, groupBy)
```
ankitnayan, line 430: I guess `anyIf(ts, toUnixTimestamp(ts) != 0) AS timestamp` won't be needed too?

ankitnayan (approved):

ankitnayan:
Irrelevant to the PR:
 
 `now()` is completely useless? Why do we need it at all?

pkg/query-service/app/metrics/v3/query_builder.go
srikanthccv: Right, it's not needed.

srikanthccv:
> Why do we need it at all?

We need at least one column for the join key for a valid ClickHouse query and we use this ts as that key. If we move the formula expression logic outside of ClickHouse then there is no need to have `ts` column.

srikanthccv:
The way we evaluate the value type panel can be improved a lot. It's also not cache friendly. I think that cleanup deserves its own issue and separate PR. Hence merging.

Issue #4330, opened 2024-01-04
[FE]: Date selector, custom option hides due to height issue


Rajat-Dabade:
<img width="240" alt="Screenshot 2024-01-04 at 6 29 47 PM" src="https://github.com/SigNoz/signoz/assets/26198062/acf5a030-5684-4521-89b3-d0ab911bc762">

PR #4331, opened 2024-01-04
refactor: updated height for date selector

Rajat-Dabade:
Fixes: https://github.com/SigNoz/signoz/issues/4330

After: 

Screenshot 2024-01-04 at 6 35 10 PM

### DIFF
--- a/frontend/src/container/TopNav/DateTimeSelection/index.tsx
+++ b/frontend/src/container/TopNav/DateTimeSelection/index.tsx
@@ -295,6 +295,7 @@ function DateTimeSelection({
 						style={{
 							minWidth: 120,
 						}}
+						listHeight={400}
 					>
 						{options.map(({ value, label }) => (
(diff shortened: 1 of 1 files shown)


Vikrant2520 (approved):

frontend/src/container/TopNav/DateTimeSelection/index.tsx
```diff
@@ -295,6 +295,7 @@ function DateTimeSelection({
 						style={{
 							minWidth: 120,
 						}}
+						listHeight={400}
```
YounixM, line 298: How was the height working before this? Why do we need the explicit height ?

Rajat-Dabade: Screenshot 2024-01-05 at 11 40 39 AM

It was using the default height before.

YounixM (approved):

PR #4332, opened 2024-01-04
fix: throw error when response for value panel returns more than one …

### LINKED ISSUE #3772: Throw error when the `value` panel API responds with multiple timeseries?
If the value panel API call return multiple timeseries or multiple datapoints in a timeseries, we should show that the query returned multiple datapoints and that you should check your query?

srikanthccv:
…series

Fixes #3772


ankitnayan:
Can you think of better words for below? Users might have hard time understanding these and we can expect them to ask about these in intercom if we do not handle it better here
- `active query`
- `only one result series`

srikanthccv:
I can change `active query` to `enabled query`. Essentially we want users to have only one active query when they work with multiple queries, usually because of the formula.

I am not sure how to simplify the `only one result series` part.

srikanthccv:
I think the first message is clear enough and people should be able to understand that there can only be one active query. For the second case, I am not sure how to simplify it further and also to note we don't allow group by in so this only happens in promql/clickhouse_sql.

Issue #4320, opened 2024-01-04
[FE] Add column: true for attributes which are column


makeavish:
![image](https://github.com/SigNoz/signoz/assets/9512100/45226b74-b237-4504-b963-8ceeda3c9fbc)

makeavish:
https://github.com/SigNoz/signoz/pull/4334#issuecomment-1899035438

srikanthccv:
Not relevant anymore

PR #4334, opened 2024-01-05
[Feat]: added iscolumn in option rendering

Rajat-Dabade:
Fixes: https://github.com/SigNoz/signoz/issues/4320

Screenshot 2024-01-05 at 3 27 06 PM


Vikrant2520 (approved):

Vikrant2520 (approved):

YounixM (approved):

srikanthccv:
It's already enough packed and nobody customer asked for this. I don't like this yet another field that's not really useful. Please check with @pranay01 and @GeekBoySupreme once.
Screenshot 2024-01-19 at 12 22 00 AM

pranay01:
Agree with @srikanthccv This makes the options too cluttered. 

I don't think this is the right UX to solve the issue

PR #4346, opened 2024-01-08
fix: empty string search doesn't return all exceptions

YounixM:
### Summary

1. Search is not working as expected in the Exceptions tab, empty results showing after removing the search word
-->Search any keyword (ex: prod) in the Application under the Exceptions tab
-->Remove the above keyword then search

### DIFF
--- a/frontend/src/container/AllError/index.tsx
+++ b/frontend/src/container/AllError/index.tsx
@@ -162,16 +162,27 @@ function AllErrors(): JSX.Element {
 				filterKey,
 				filterValue || '',
 			);
-			history.replace(
-				`${pathname}?${createQueryParams({
-					order: updatedOrder,
-					offset: getUpdatedOffset,
-					orderParam: getUpdatedParams,
-					pageSize: getUpdatedPageSize,
-					exceptionType: exceptionFilterValue,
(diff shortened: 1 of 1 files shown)


makeavish:
When the server is not responding then we just see the loading screen, we should see the server error page as you described above.

Screenshot 2024-01-09 at 10 29 23

YounixM:
@vishalsharma1800 : Did we get a 5xx response in the above case ?

makeavish:
> @vishalsharma1800 : Did we get a 5xx response in the above case ?

Wrong vishal BTW.

We get `ERR_NAME_NOT_RESOLVED` without any response code.

Rajat-Dabade (commented):
Added some non-blocking comments.

Test:
Search in exception - Application - works fine. 
Server Error flow - discuss with @makeavish and @YounixM

Rajat-Dabade (dismissed):
LGTM 👍

Vikrant2520 (dismissed):

Rajat-Dabade (approved):

Vikrant2520 (approved):

makeavish (approved):

PR #4361, opened 2024-01-10
chore: add /v4/query_range endpoint

srikanthccv:
### Summary

Part 7 of https://github.com/SigNoz/signoz/issues/4016

This PR

- Uses the queriev2 introduced in https://github.com/SigNoz/signoz/pull/4170 (which uses metricsV4.PrepareMetricQuery) and add a new endpoint `/v4/query_range`
- Some refactoring by moving the redundant helper code into the helpers package
- Update the query preparation for delta to skip the per series aggregation because we can combine all the results. Example time aggregation - sum, space aggregation - sum, here we can sum up all the values without the need for time/space aggregation (see tests, the group by fingerprint is not required)
- `pkg/query-service/app/metrics/v4/query_builder_test.go` demonstrates the `rate`, `quantile` and working with gauge metric `system_cpu_usage`.

Some post-processing work such as limit, having etc.. will be added in follow-up PR.


pkg/query-service/app/http_handler.go
```diff
@@ -542,7 +559,7 @@ func (aH *APIHandler) addTemporality(ctx context.Context, qp *v3.QueryRangeParam
 	if qp.CompositeQuery != nil && len(qp.CompositeQuery.BuilderQueries) > 0 {
 		for name := range qp.CompositeQuery.BuilderQueries {
 			query := qp.CompositeQuery.BuilderQueries[name]
-			if query.DataSource == v3.DataSourceMetrics {
+			if query.DataSource == v3.DataSourceMetrics && query.Temporality == "" {
```
ankitnayan, line 562: you are trying to fill only empty temporality values in the query payload? Why would frontend send empty Temporality for only a few of the metric names for which you need to query DB to populate it.

pkg/query-service/app/http_handler.go
```diff
@@ -3241,3 +3258,67 @@ func (aH *APIHandler) liveTailLogs(w http.ResponseWriter, r *http.Request) {
...
+
+		spanKeys, err = aH.getSpanKeysV3(ctx, queryRangeParams)
+		if err != nil {
+			apiErrObj := &model.ApiError{Typ: model.ErrorInternal, Err: err}
+			RespondError(w, apiErrObj, errQuriesByName)
+			return
+		}
+	}
+
+	result, err, errQuriesByName = aH.querierV2.QueryRange(ctx, queryRangeParams, spanKeys)
```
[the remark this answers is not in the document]
ankitnayan: I see these 2 missing:
1. https://github.com/SigNoz/signoz/blob/c28f367f46f92292f85884e5a0bc195882efd8f5/pkg/query-service/app/http_handler.go#L3053
2. https://github.com/SigNoz/signoz/blob/c28f367f46f92292f85884e5a0bc195882efd8f5/pkg/query-service/app/http_handler.go#L3059-L3067

pkg/query-service/app/http_handler.go
```diff
@@ -3241,3 +3258,67 @@ func (aH *APIHandler) liveTailLogs(w http.ResponseWriter, r *http.Request) {
...
+
+	resp := v3.QueryRangeResponse{
+		Result: result,
+	}
+
+	aH.Respond(w, resp)
+}
+
+func (aH *APIHandler) QueryRangeV4(w http.ResponseWriter, r *http.Request) {
+	queryRangeParams, apiErrorObj := ParseQueryRangeParams(r)
```
ankitnayan, line 3306: ` ParseQueryRangeParams` is expected to version change later as per the needs of enhanced QB?

pkg/query-service/app/metrics/v4/delta/timeseries.go
```diff
@@ -118,3 +160,37 @@ func prepareMetricQueryDeltaTimeSeries(start, end, step int64, mq *v3.BuilderQue
 
 	return query, nil
 }
+
+// canShortCircuit returns true if we can use the optimized query
+// for the given query
+// This is used to avoid the group by fingerprint thus improving the performance
+// for certain queries
+// cases where we can short circuit:
+// 1. time aggregation = (rate|increase) and space aggregation = sum
```
ankitnayan, line 169: This is `sum_rate` which one of the most used aggregations? What is the perf gained?

pkg/query-service/app/metrics/v4/delta/timeseries.go
```diff
@@ -118,3 +160,37 @@ func prepareMetricQueryDeltaTimeSeries(start, end, step int64, mq *v3.BuilderQue
...
+// 2. time aggregation = sum and space aggregation = sum
+//   - sum of sums is same as sum of all values
+//
+// 3. time aggregation = min and space aggregation = min
+//   - min of mins is same as min of all values
+//
+// 4. time aggregation = max and space aggregation = max
+//   - max of maxs is same as max of all values
+//
+// all of this is true only for delta metrics
```
ankitnayan, line 181: Have we considered using cumulativeToDelta processor? If there is a perf gain, we should

pkg/query-service/app/http_handler.go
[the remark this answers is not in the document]
srikanthccv: I mentioned in the PR description that post-processing logic such as limit is to follow in separate PR. I am aware of the 2nd context finish not being present. I wanted to read about HTTP semantics and handle it better otherwise I will just copy paste the same block.

pkg/query-service/app/http_handler.go
srikanthccv: Yes, if the situation demands we will do that.

pkg/query-service/app/metrics/v4/delta/timeseries.go
srikanthccv: The last time I measured it makes queries run ~2x faster.

pkg/query-service/app/metrics/v4/delta/timeseries.go
srikanthccv: We can't use it in our collector instance because it is a stateful processor and the same data point has to arrive at the same collector for it to work correctly which is not guaranteed so we are not going to use it.

I have mentioned this elsewhere in the delta EPIC issue, when we roll out the delta and make it the default we will update our docs to send cumulativetodelta on client collector instances and SDK instructions via [ENV](https://github.com/open-telemetry/opentelemetry-specification/blob/12cd1b1a4750a287282c4bfcd713de38d6716700/specification/metrics/sdk_exporters/otlp.md#additional-configuration) to prefer delta temporality.

pkg/query-service/app/http_handler.go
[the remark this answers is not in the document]
srikanthccv: By HTTP semantics I meant something like this https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206

pkg/query-service/app/metrics/v4/delta/timeseries.go
ankitnayan: nice

pkg/query-service/model/v3/v3.go
```diff
@@ -500,6 +509,7 @@ type BuilderQuery struct {
 	SelectColumns      []AttributeKey    `json:"selectColumns,omitempty"`
 	TimeAggregation    TimeAggregation   `json:"timeAggregation,omitempty"`
 	SpaceAggregation   SpaceAggregation  `json:"spaceAggregation,omitempty"`
+	Quantile           float64           `json:"quantile,omitempty"`
```
ankitnayan, line 512: I am not sure if it's wise to keep this field at toplevel. It seems like it should belong to AggreagtorOperator but I understand that would mean changing `AggreagtorOperator` to struct. I see you planning to use this field as a special case in your calculations too. Is that right?

pkg/query-service/app/http_handler.go
ankitnayan: had a chat in slack

pkg/query-service/app/http_handler.go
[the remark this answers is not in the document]
ankitnayan: ok

pkg/query-service/app/metrics/v4/delta/timeseries.go
ankitnayan: ok..so we are good to receive delta but not convert cumulative to delta

pkg/query-service/model/v3/v3.go
ankitnayan: >How would you model this differently?

```type AggreagtorOperator  struct {
 type: string, // sum, mix, max, percentile
 value: null // expect for percentile
}
```
The advantage is scoping percentile to AggregateOperator. Any validation, restriction, etc has to be done inside AggregateOperator for percentiles too

ankitnayan: Using separate `TimeAggregation` and `SpaceAggregation` for metrics

ankitnayan (approved):

srikanthccv:
Will be moving the `Quantile` to `SpaceAggregation` in next PR.

PR #4370, opened 2024-01-12
feat: integrate date time selector across app

Vikrant2520:
### Summary

- integrate the new date time selector across application 

#### Related Issues / PR's

Solves :- #SIG-511

#### Screenshots

https://github.com/SigNoz/signoz/assets/54737045/865eb4b2-73a1-4c61-9b09-425789327712

All modules that consume date-time modal

### DIFF
--- a/frontend/src/components/ExplorerCard/ExplorerCard.tsx
+++ b/frontend/src/components/ExplorerCard/ExplorerCard.tsx
@@ -244,7 +244,7 @@ function ExplorerCard({
 				</ExplorerCardHeadContainer>
 			)}
 
-			{children}
+			<div style={{ marginTop: '1rem' }}>{children}</div>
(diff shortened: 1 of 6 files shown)


Rajat-Dabade (approved):
LGTM 🚀, my only concern is the Panel Time Preference in Dashboard > Panel > Edit.

Screenshot 2024-01-12 at 2 49 30 PM


Do you think we should replace this one as well? Users may want the particular panel to use the 45min or 10d timestamp which is part of a global timestamp but not in the dropdown.

Vikrant2520:
@Rajat-Dabade the panel dropdown currently matches the left side of new date time selector, the right side is just some helper times for the user instead of custom timestamps everytime, but sure we can discuss the same in standup

Issue #3932, opened 2023-11-09
Do not retry responses with 400 Bad Request status


srikanthccv:
## Bug description

The charts retry logic attempts to retry failed requests. The "failed" request should be those where server has some error. The retry mechanism retries all responses regardless of status codes and makes unnecessary requests.

PR #4376, opened 2024-01-16
fix: [GH-3932]: do not retry API's in case of 4XX status code

Vikrant2520:
### Summary

Three scenarios:- 

- The API has a `try` / `catch` block generally with `ErrorResponseHandler`, in that case, the API Error is not propagated to the `QueryClient` so the retries do not trigger, even if we pass `retry: true`, So no change here
- The second case is when we do not catch the error in the API getter, In that case, the error hits the QueryClient and the retry mechanism kicks in.
- The third case where we catch the API error and manually throw a different error which might / might not contain the API response status, so here as well the retry logic won't kick in,So while handling such cases when adding a new API we should add status code to such manually error handling.

#### Related Issues / PR's

#3932

### DIFF
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -16,6 +16,15 @@ const queryClient = new QueryClient({
 	defaultOptions: {
 		queries: {
 			refetchOnWindowFocus: false,
+			retry(failureCount, error): boolean {
+				if (
+					error instanceof Error &&
+					error.message.includes('API responded with 400')
+				) {
+					return false;
+				}
+				return failureCount < 2;
+			},
 		},
 	},
 });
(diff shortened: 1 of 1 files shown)


frontend/src/index.tsx
```diff
@@ -15,6 +15,15 @@ const queryClient = new QueryClient({
 	defaultOptions: {
 		queries: {
 			refetchOnWindowFocus: false,
+			retry(failureCount, error): boolean {
+				if (
+					error instanceof Error &&
+					error.message.includes('API responded with 400')
```
YounixM, line 21: Error doesn't have Status Code ?

YounixM (approved):

PR #4378, opened 2024-01-16
fix: virtuoso scroll refresh issue

Vikrant2520:
### Summary

- used `isLoading` instead of `isFetching` as it unmounts the component and resets the scroll position 
- Loading state should be present in for first load itself which is taken care by `isLoading` prop

> isLoading returns true when you're fetching data for the first time. On the other hand, isFetching will be true anytime there's an ongoing data fetching process, not just for the first time

#### Screenshots

https://github.com/SigNoz/signoz/assets/54737045/d6091925-c405-43e3-8b50-3398a79c404a

All the three views Raw / Default / Column View


Rajat-Dabade (approved):
LGTM 👍

PR #4381, opened 2024-01-17
chore: add functions support

srikanthccv:
### Summary

Part 7 of #4016 

This change adds support for applying certain functions/transformations on the time series result from the query. Here is the list of things we post-process.

1. Applying the `Having` filter on the result - in v3 this is done using ClickHouse query. This doesn't play nicely with caching - a change in `having` filter requires running the query again even though the underlying query didn't change. With this post-processing, we effectively make use of the cached results.
2. Applying limit for the metric result - this is described below in detail why we do this. tldr; It's faster to do it this way than to run a subquery.
3. Applying the `ReduceTo` on the result - in the table panel, there should be one result per row and in the value panel, there should just be one result overall. We moved this to query service for the same reason we moved `Having`.
4. Apply functions - Having control over the functions we support and how they work is important. Not all of the functions can be supported in ClickHouse. Even if does, it makes query preparation complicated because different function expects different input data. It's much easier and more flexible to support these functions in query service instead.

The diff might seem but most of it is tests and those tests are not even exhaustive. In separate PR I will be refactoring bunch of these tests by reading the json data into structs instead of defining them in the code.


srikanthccv:
@ankitnayan please review

pkg/query-service/app/queryBuilder/functions.go
```diff
@@ -0,0 +1,242 @@
...
+	return result
+}
+
+// funcEWMA3 returns the exponential weighted moving average for each point in a series
+func funcEWMA3(result *v3.Result, alpha float64) *v3.Result {
+	for _, series := range result.Series {
+		var prev float64
+		for _, point := range series.Points {
+			if math.IsNaN(prev) {
+				point.Value = point.Value
```
ankitnayan, line 113: Do you think cases when `point.Value` can be NaN as functions can be chained. Eg say `funcEWMA3 (clampMax())`?

ankitnayan:
I see `point.Value = math.NaN()` assignments unhandled in chained functions.

pkg/query-service/app/limit_test.go
```diff
@@ -0,0 +1,624 @@
...
+								Timestamp: 0,
+								Value:     340,
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			// ["GET /api/v1/health", "DELETE /api/v1/health"] so result should be ["DELETE /api/v1/health"] although it has lower value
```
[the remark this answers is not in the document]
ankitnayan: yeah..I got that

pkg/query-service/app/queryBuilder/functions.go
```diff
@@ -0,0 +1,242 @@
...
+	case "log2":
+		return funcLog2(result)
+	case "log10":
+		return funcLog10(result)
+	case "cumSum":
+		return funcCumSum(result)
+	case "ewma3", "ewma5", "ewma7":
+		alpha, ok := fn.Args[0].(float64)
+		if !ok {
+			alpha = 0.25
```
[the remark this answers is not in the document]
srikanthccv: Yes, I read anything from 0.2-0.3 makes a decent default value.

pkg/query-service/app/limit_test.go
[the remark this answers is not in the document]
ankitnayan: maybe if you had made `DELETE /api/v1/health` with higher GroupingSetsPoint value, it would have been clearer

pkg/query-service/app/queryBuilder/functions.go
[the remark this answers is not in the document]
ankitnayan: maybe declare it somewhere?

pkg/query-service/app/queryBuilder/functions.go
```diff
@@ -0,0 +1,242 @@
...
+	for _, series := range result.Series {
+		for i := 3; i < len(series.Points)-3; i++ {
+			point := series.Points[i]
+			point.Value = (series.Points[i-3].Value + series.Points[i-2].Value + series.Points[i-1].Value + series.Points[i].Value + series.Points[i+1].Value + series.Points[i+2].Value + series.Points[i+3].Value) / 7
+		}
+	}
+	return result
+}
+
+func ApplyFunction(fn v3.Function, result *v3.Result) *v3.Result {
```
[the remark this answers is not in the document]
srikanthccv: Yes, it makes sense to having validation what functions are supported.

pkg/query-service/app/queryBuilder/functions.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	return result
+}
+
+// funcMedian5 returns the median of 5 points for each point in a series
+func funcMedian5(result *v3.Result) *v3.Result {
+	for _, series := range result.Series {
+		for i := 2; i < len(series.Points)-2; i++ {
+			point := series.Points[i]
+			point.Value = (series.Points[i-2].Value + series.Points[i-1].Value + series.Points[i].Value + series.Points[i+1].Value + series.Points[i+2].Value) / 5
```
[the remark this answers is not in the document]
srikanthccv: Correct, this must be fixed.

pkg/query-service/app/queryBuilder/functions.go
srikanthccv: Yes, it's possible but it's a good point. Let me think about if the same should be considered in other functions such as cumulative sum.

pkg/query-service/app/queryBuilder/functions.go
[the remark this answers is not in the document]
srikanthccv: The better approach is to use alpha = 2 / (N + 1) (from pandas) - updated it.

pkg/query-service/app/queryBuilder/functions.go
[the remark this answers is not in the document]
srikanthccv: Added validation

pkg/query-service/app/queryBuilder/functions.go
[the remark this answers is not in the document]
srikanthccv: Updated to return median instead of mean

pkg/query-service/app/queryBuilder/functions.go
srikanthccv: Handled NaN in all relevenat functions

pkg/query-service/app/queryBuilder/functions.go
```diff
@@ -0,0 +1,286 @@
...
+
+		// Set the median7 values for the series
+		for i := 3; i < len(series.Points)-3; i++ {
+			series.Points[i].Value = median7[i-3]
+		}
+	}
+	return result
+}
+
+func median(values []float64) float64 {
```
ankitnayan, line 230: this is true assuming `values` are sorted. Where is this asserted?

srikanthccv: No, there is a sort `sort.Float64s(values)` just below this line

ankitnayan: ah..my bad

pkg/query-service/app/queryBuilder/functions.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	return result
+}
+
+// funcCumSum returns the cumulative sum for each point in a series
+func funcCumSum(result *v3.Result) *v3.Result {
+	for _, series := range result.Series {
+		var sum float64
+		for idx, point := range series.Points {
+			if !math.IsNaN(point.Value) {
```
ankitnayan, line 100: should this check be part of all functions? Eg `funcClampMax`?

pkg/query-service/app/queryBuilder/functions.go
```diff
@@ -184,58 +199,87 @@ func funcMedian5(result *v3.Result) *v3.Result {
...
-	case "ewma3", "ewma5", "ewma7":
+	case v3.FunctionNameEWMA3, v3.FunctionNameEWMA5, v3.FunctionNameEWMA7:
 		alpha, ok := fn.Args[0].(float64)
 		if !ok {
-			alpha = 0.25
+			// alpha = 2 / (n + 1) where n is the window size
+			if fn.Name == v3.FunctionNameEWMA3 {
+				alpha = 0.5
+			} else if fn.Name == v3.FunctionNameEWMA5 {
+				alpha = 0.34
```
ankitnayan, line 272: it would be better to take approximation of `0.33` for the value of `0.3333...`

pkg/query-service/app/queryBuilder/functions.go
srikanthccv: No, it's only needed where the other values would do arithmetic and produce NaN result. The threshold functions will not need this as any comparison with NaN returns and false and operation is skipped https://go.dev/play/p/nb6aUSZBbUY

pkg/query-service/app/queryBuilder/functions.go
```diff
@@ -0,0 +1,286 @@
...
+			if len(values) == 0 {
+				median7 = append(median7, math.NaN())
+				continue
+			}
+
+			median7 = append(median7, median(values))
+		}
+
+		// Set the median7 values for the series
+		for i := 3; i < len(series.Points)-3; i++ {
```
ankitnayan, line 223: the first 2 and last 2 points will not be medians. Similar loops containing window size usually has rough edges in the start and end. I am not sure how to handle these edges. Try giving this a read. In image processing, we used to run similar window functions over the image pixel by pixel and the 2 pixel boundary in a window size of 5 used to be weird. So, we used to remove those 2 pixel border from the image itself.

pkg/query-service/app/queryBuilder/functions.go
ankitnayan: so `sum += NaN` will evaluate to NaN?

srikanthccv: Yes, we skip NaNs in functions where there is some arithmetic.

pkg/query-service/app/queryBuilder/functions.go
srikanthccv: In an ideal scenario, the edges would be replaced with nil/NaN. The same is true for arrays with smaller lengths. As an end user, I would prefer to see the values instead.

ankitnayan: ok

pkg/query-service/app/queryBuilder/functions.go
```diff
@@ -0,0 +1,286 @@
...
+		}
+	}
+	return result
+}
+
+// funcClampMin cuts off values below the threshold and replaces them with the threshold
+func funcClampMin(result *v3.Result, threshold float64) *v3.Result {
+	for _, series := range result.Series {
+		for idx, point := range series.Points {
+			if point.Value < threshold {
```
ankitnayan, line 40: if `point.Value=NaN`, this evaluates to false and hence the point is not used for clamping?

pkg/query-service/app/queryBuilder/functions.go
ankitnayan: got it

ankitnayan (approved):

PR #4382, opened 2024-01-17
chore: support p9{9,5,0},75,50 for space aggregation

srikanthccv:
### Summary

Part of #4016 

Follow-up from https://github.com/SigNoz/signoz/pull/4361#issuecomment-1893493293


pkg/query-service/model/v3/v3.go
```diff
@@ -569,7 +604,7 @@ func (b *BuilderQuery) Validate() error {
 			return fmt.Errorf("data source is invalid: %w", err)
 		}
 		if b.DataSource == DataSourceMetrics {
-			if b.TimeAggregation == TimeAggregationUnspecified && b.Quantile == 0 {
+			if b.TimeAggregation == TimeAggregationUnspecified {
```
ankitnayan, line 607: why is the Quantile check not needed anymore?

ankitnayan: yeah..what is not clear is why you needed to validate Quantile when it was a separate field in the struct vs when moved to SpaceAggregation

srikanthccv: Let me clarify what is happening. The validation is written in a way that either it has to follow either v4 or v3 query_range format. If there are no v4 fields we do the validation on old `aggregateOperator` which is what happening in following like `if err := b.AggregateOperator.Validate(); err != nil { `.

Prior to this PR spaceAggregation, there are two ways to send the query to v4, use `timeAggregation` & `spaceAggregation` for non-histogram metric types and use `quantile` for histograms. So for it to be v3 query range format, both `timeAggregation` and `quantile` must not have been set. Going forward there is no separate treatment for histograms. It must be part of the `spaceAggregation`.

ankitnayan: I see. This logic is not simple. Let's discuss later how to simplify this validation or write it down for other's to understand

ankitnayan (approved):

PR #4384, opened 2024-01-17
fix: set max 6 months for user entered time

YounixM:
image

### DIFF
--- a/frontend/src/components/CustomTimePicker/CustomTimePicker.styles.scss
+++ b/frontend/src/components/CustomTimePicker/CustomTimePicker.styles.scss
@@ -48,7 +48,7 @@
 
 .valid-format-error {
 	margin-top: 4px;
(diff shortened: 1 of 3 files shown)


frontend/src/components/CustomTimePicker/CustomTimePicker.styles.scss
```diff
@@ -43,7 +48,9 @@
 
 .valid-format-error {
 	margin-top: 4px;
-	color: var(--bg-cherry-400, #ea6d71);
+	color: var(--bg-cherry-400, #ea6d71) !important;
+	font-size: 13px !important;
+	font-weight: 400 !important;
```
Rajat-Dabade, line 53: `--font-weight-normal` can be used here

Vikrant2520 (approved):

Rajat-Dabade (approved):
One minor comment, every thing else LGTM ✅

Issue #4326, opened 2024-01-04
Automatically save layout when panels are rearranged


srikanthccv:
The rearranged layout is lost If the user forgets to save it. It's painful to rearrange things again and again. Why do we make the user explicitly click on the save layout? Can we save the layout when there is any rearrangement of the panel position automatically?

PR #4385, opened 2024-01-17
fix: autosave layout to layout change

YounixM:
Fixes: https://github.com/SigNoz/signoz/issues/4326


Vikrant2520 (dismissed):

frontend/src/container/GridCardLayout/GridCardLayout.tsx
```diff
@@ -108,6 +110,27 @@ function GraphLayout({ onAddPanelHandler }: GraphLayoutProps): JSX.Element {
...
 		: [...ViewMenuAction];
 
+	const handleLayoutChange = (layout: Layout[]): void => {
+		if (!isEqual(layout, dashboardLayout)) {
+			setDashboardLayout(layout);
+		}
+	};
+
+	useEffect(() => {
+		if (
```
Rajat-Dabade, line 120: We should move this piece of code inside the `handleLayoutChange` function so that the `useEffect` will be removed. 

This `useEffect` will be executed every time the component mounts and will trigger an unnecessary update dashboard API call.

YounixM: It is dependent on handleLayout change event itself. however, as we dispatch event to save layout, there will be a raise condition and it doesn't update the layout with new layout

Rajat-Dabade: https://www.loom.com/share/12bc0fd7da054455b5863e1ff6161cf1?sid=eb9b306a-5083-4dcd-a609-71d56148aaf5

Rajat-Dabade (approved):

Rajat-Dabade (approved):

Vikrant2520 (approved):

Issue #4383, opened 2024-01-17
Services overview page breaks if there is a `/` in the name of the service


srikanthccv:
## Bug description

Context https://signoz-community.slack.com/archives/C01HWQ1R0BC/p1705426248031489

## Expected behavior

Clicking on the service name (which contains `/`) should show the Overview metrics

## How to reproduce

1. Send trace data from the application with the `service.name` attribute that has `/` in it. Example `ms/cart` 
2. Click on the service name

PR #4388, opened 2024-01-18
fix: [GH-4383]: handle special characters in the services name

Vikrant2520:
### Summary

- handle special characters in the service name 
- `encode` the service name while updating the URL and decode the same while referencing in other components

#### Related Issues / PR's

#4383 

#### Screenshots

https://github.com/SigNoz/signoz/assets/54737045/73362cf0-550e-4369-b777-555d32c2dd61

- services page and nested routes tested

### DIFF
--- a/frontend/src/container/MetricsApplication/Tabs/DBCall.tsx
+++ b/frontend/src/container/MetricsApplication/Tabs/DBCall.tsx
@@ -30,7 +30,6 @@ import {
 
 function DBCall(): JSX.Element {
 	const { servicename } = useParams<IServiceName>();
-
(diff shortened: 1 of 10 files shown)


YounixM (dismissed):

YounixM (approved):

Rajat-Dabade (approved):

srikanthccv:
@Vikrant2520 This is buggy and not tested correctly. When you send a request to the backend for data, you should not send the encoded service name but the original service name. This page doesn't load https://stagingapp.signoz.io/services/hello%2Fworld. Even in the video, it can be seen that page didn't load.

PR #4393, opened 2024-01-18
fix: allow workspace blocked users to extend trial

YounixM:
Screenshot 2024-01-18 215138

image


Vikrant2520 (approved):

makeavish (approved):

srikanthccv:
Why aren't _SigNoz URL_ and _Admin Email_ filled? We know only the admin sees this option and we would have their tenant URL. Why not just fill in the details for them?

YounixM:
My bad. I merged it before seeing the comment

YounixM:
@srikanthccv: We want the user to add details to make sure the user has the actual intent to extend and is not just playing around.

srikanthccv:
I think we should generally assume positive intent when it comes to things like this. But it may be fine as is.

PR #4399, opened 2024-01-19
chore: add tenantId and orgName in usage

nityanandagohain:
Fixes https://github.com/SigNoz/engineering-pod/issues/1187

### DIFF
--- a/ee/query-service/usage/manager.go
+++ b/ee/query-service/usage/manager.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
+	"regexp"
 	"strings"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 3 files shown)


ee/query-service/usage/manager.go
```diff
@@ -123,6 +139,18 @@ func (lm *Manager) UploadUsage() {
...
+	// Try to get the org name
+	orgName := ""
+	orgNames, err := lm.modelDao.GetOrgs(ctx)
+	if err != nil {
+		zap.S().Errorf("failed to get org data: %v", zap.Error(err))
+	}
+	if len(orgNames) != 1 {
+		zap.S().Errorf("expected one org but got %d orgs", len(orgNames))
+	} else {
+		orgName = orgNames[0].Name
```
srikanthccv, line 151: Will there be non-zero `orgNames` all the time, even if `err` is non-nil?

nityanandagohain: Thanks for noticing, have fixed.

srikanthccv (approved):

PR #4402, opened 2024-01-22
chore: add formula eval in query-service

srikanthccv:
### Summary

Part 9 of #4016; makes the `v4/query_range` to the same level as the `v3/query_range`. See this collection with examples https://signoz.postman.co/workspace/Team-Workspace~cfa5c7b5-124f-4bbc-88d9-2351af8dd95b/collection/21397560-d1dd16af-8de8-438d-8a3d-cd96e05c6440?action=share&creator=21397560.

## What do we do today for formula evaluation? 

We join the result tables using the `INNER JOIN` and get the result. Example

```sql
SELECT B.value/A.value FROM A INNER JOIN B ON A.tag1 = B.tag1 AND A.tag2 = B.tag2
```

WHERE A and B are complex queries on their own such as sum rate queries.

## What are the cons of this approach?

- The cached results can't be effectively used as the inner queries must exist for the above query to work.
- The JOIN evaluates the right side of the query first and then join the result with the first query result. This becomes really slow. We could instead run the queries in parallel and combine result in memory better.

## How do eval formula in query-service?

We make use of the library https://github.com/SigNoz/govaluate to run the arithmetic expressions. The steps to decide how do we join the result series.

1. Find all unique label sets
2. For each unique label set, find a series that matches the label set from each query result
3. Join the series on timestamp and calculate the new values
4. Return the new series

## Summary by CodeRabbit

- **New Features**
    - Introduced advanced query processing functionality allowing for unique label set identification, series joining by timestamps, and value calculation through mathematical expressions.
    - Added support for evaluating expressions and defining mathematical functions within queries.

- **Bug Fixes**
    - Enhanced error handling in query result processing to ensure robustness.

- **Tests**
    - Implemented comprehensive tests for new query processing and evaluation functionalities, covering a wide range of scenarios including division operations and expression evaluations.


ankitnayan:
this PR is difficult to review as in current state. I spent quite sometime in understanding an example and pass it through different stages as mentioned. 
1. Find all unique label sets
2. For each unique label set, find a series that matches the label set from each query result
3. Join the series on timestamp and calculate the new values
4. Return the new series

 I felt 2 things missing to help review:
 1. An example step by step output of each stage
 2. What is the role of each step in the bigger picture. Maybe this gets solved when I am able to follow an example

srikanthccv:
Here is an explanation with example


- A series can be joined with another series if they have the same label set or one is a subset of the other.

Example:

The series in the `{}` represents the result label set of the query.
```
A = {"host"="server1", "region"="us-east"}, {"host"="server1", "region"="us-west"}
B = {"host"="server2"}
C = {"host"="server1"}
D = {"host"="server2, "region"="us-east"}
E = {"host"="server1", "region"="us-east"}
F = {"host"="server1", "region"="us-west"}
G = {"host"="server3"}
```

`A` can be joined with `C`, but not with `B` or `D`. `C` can be joined with `E` and `F`

- The `Result` struct contains the output series for each query. This can be one or more.

Example:
The result of the C query contains two output series `C = {"host"="server1", "region"="us-east"}, {"host"="server1", "region"="us-west"}`

- Series'es within the "Result" struct can't be joined with each other as they belong to the same query.

Example:

The `{"host"="server1", "region"="us-east"}` can't be joined with `{"host"="server1", "region"="us-west"}` as they belong to the same query.

The work `processResults` does:

1. The `findUniqueLabelSets` attempts to find the unique label sets. The formula expression result label set will be a subset of this. The role of this step in the overall process is to prepare the potential label sets that could be used to join the series. Note: We can't rely on the first series label set as it might not contain all the relevant sets.

Example:

The `findUniqueLabelSets` will return `{"host"="server1"}, {"host"="server2"}, {"host"="server1", "region"="us-east"}, {"host"="server1", "region"="us-west"}, {"host"="server2, "region"="us-east"}`

2. For each series in the `findUniqueLabelSets`, the `joinAndCalculate` will attempt to find the series that can be joined with it in the `Result`s. If we could find a series for each query the appeared in the formula expression, we will calculate the formula expression.

Example:

If we take the expression `A/(C+G)`.

The output of the `findUniqueLabelSets` will be `{"host"="server1", "region"="us-east"}, {"host"="server1", "region"="us-west"}, {"host"="server3"}` as the `{"host"="server1"}` is a subset of `{"host"="server1", "region"="us-east"}` and `{"host"="server1", "region"="us-west"}`.

Now we try to join the series

```
for item in [{"host"="server1", "region"="us-east"}, {"host"="server1", "region"="us-west"}, {"host"="server3"}]:
    currentSeries = item
    seriesToJoin = []
    for result in results:
        for series in result:
            if series.labelSet is a subset of currentSeries:
                seriesToJoin.append(series)

    if seriesToJoin contains a series for each query:
        resultSeries = calculate the formula expression for the currentSeries
```

3. Join the series on timestamp and calculate the new values

Example:

Use the intermediary `resultSeries` to join the series on timestamp and calculate the new values.

ankitnayan (approved):

srikanthccv:
I will hold the merging till I add more tests and further verification against the formulas our users use.

srikanthccv:
I have added more tests and some validations since the last approval. Please take another look.

srikanthccv:
@ankitnayan ICYMI ^

pkg/query-service/app/parser_test.go
```diff
@@ -862,3 +860,353 @@ func TestParseQueryRangeParamsPromQLVars(t *testing.T) {
...
+						QueryName:          "C",
+						DataSource:         v3.DataSourceMetrics,
+						AggregateOperator:  v3.AggregateOperatorSum,
+						AggregateAttribute: v3.AttributeKey{Key: "signoz_calls_total"},
+						GroupBy:            []v3.AttributeKey{{Key: "service_name"}, {Key: "operation_name"}, {Key: "status_code"}},
+						Expression:         "C",
+					},
+					"F1": {
+						QueryName:  "F1",
+						Expression: "C/(A + B)",
```
ankitnayan, line 1174: A+B doesn't make sense in this example?

ankitnayan: I feel A+B always should match groupBy keys. Any example where they are different?

srikanthccv: I agree the arithmetic doesn't make sense. We are not interested in that. This is testing whether it is possible to join them. The `expectErr: true` cases include the ones where they are different.

ankitnayan: I see that but not able to make sense why this test case has `expectErr: false` but not for https://github.com/SigNoz/signoz/pull/4402/commits/535eaa07f35fc15eb60066fefad2c9c9f81fb48e#diff-c92468e920f66a6614fa0b6c5b37f368faec8c9514790ae182f2807adcbc8bd1R1110

srikanthccv: The logic is simple. Either fully match or subset. `[operation_name status_code]` is not a subset of `[service_name operation_name]`

srikanthccv: If these were tables, You wouldn't be able to join the tables because the `status_code` is missing from the left table `A` and `service_name` is missing from the right table `B`

ankitnayan: It's easier to see it this way

ankitnayan (approved):

PR #4408, opened 2024-01-22
fix: logs page crash when special chars present in the value of query

Vikrant2520:
### Summary

Page crash for logs page on query is happening when there are special characters present in the Value string of the query as we create a `RegExp`  based on the partial value which fails because those char's have different meaning in regEx context.

Solution :- Escape the special chars in the Partial Value

#### Screenshots

https://github.com/SigNoz/signoz/assets/54737045/e74c8cb4-ef7b-4486-8905-52861cf53951


prashant-shahi (approved):
:shipit: LGTM

Tested with the following metrics in staging and testing envs.

```
container_name IN some-dummy-container-name-
```

YounixM (approved):

PR #4445, opened 2024-01-28
chore: add time shift function and some refactoring

srikanthccv:
### Summary

Part of #4016 

The number of files it touches is big but the changes are scoped. Instead of creating one PR for each I thought to include them all here.

- Use v4 tables wherever possible in the new builder and update tests
- Add metric type so that frontend can have options relevant to each type.
- Add endpoint for metric metadata which includes the `unit`, and `description` from the instrument. This can be used to autofill the y-axis unit.
- Update temporality logic to fetch only if we don't know.
- Adjust the queried to not run formula anymore and replace it with query-serivce
- Fix some gaps noticed like double `le`
- Add some more validation
- Add support for timeshift.

## Summary by CodeRabbit

- **New Features**
	- Introduced enhanced metrics retrieval and metadata fetching capabilities.
	- Added time-shifting functionality for metrics data.
	- Implemented more efficient and accurate query processing for metrics.
- **Improvements**
	- Improved handling of query temporality to reduce redundant operations.
	- Refined error handling and validation for more robust query parsing.
- **Bug Fixes**
	- Fixed inconsistencies in time filtering across various metrics retrieval processes.
- **Refactor**
	- Updated and optimized SQL queries and logic for handling metrics data.
	- Enhanced query builder to support new functionalities and improve efficiency.


ankitnayan:
>Add endpoint for metric metadata which includes the unit, and description from the instrument. This can be used to autofill the y-axis unit.

This can also be used to suggest/restrict the aggregations that can be used for the metric name? So, should be added to the api that fetches/suggests metric name too? If this is incremental, we can pick this up when needed

srikanthccv:
Yes, I added the metric type here https://github.com/SigNoz/signoz/pull/4445/files#diff-a8c15149b48aee5261176aa62fa630bb7bde1c4362583ff95871ba24cec75bf9R3976, which helps frontend to show the appropriate options. I discussed this with @YounixM already. The unit, description is good to have that can't be added in the `v3.AttributeKey` so those two can be added to the response api when we update the apis structure to v4.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -4111,6 +4104,68 @@ func (r *ClickHouseReader) GetLatencyMetricMetadata(ctx context.Context, metricN
 	}, nil
 }
 
+func (r *ClickHouseReader) GetMetricMetadata(ctx context.Context, metricName, serviceName string) (*v3.MetricMetadataResponse, error) {
```
ankitnayan, line 4107: this method is expected to work only for the APM data as it contains `serviceName`? What's the planned scope of using metric metadata?

ankitnayan:
> Yes, I added the metric type here https://github.com/SigNoz/signoz/pull/4445/files#diff-a8c15149b48aee5261176aa62fa630bb7bde1c4362583ff95871ba24cec75bf9R3976, which helps frontend to show the appropriate options. I discussed this with @YounixM already. The unit, description is good to have that can't be added in the `v3.AttributeKey` so those two can be added to the response api when we update the apis structure to v4.

So the frontend is still planned to work with v3 APIs and the backend will be prepared for v4 structure?

pkg/query-service/app/metrics/v4/cumulative/timeseries.go
```diff
@@ -107,19 +107,19 @@ const (
...
 	}
 
-	samplesTableFilter := fmt.Sprintf("metric_name = %s AND timestamp_ms >= %d AND timestamp_ms <= %d", utils.ClickHouseFormattedValue(mq.AggregateAttribute.Key), start, end)
+	samplesTableFilter := fmt.Sprintf("metric_name = %s AND unix_milli >= %d AND unix_milli < %d", utils.ClickHouseFormattedValue(mq.AggregateAttribute.Key), start, end)
 
 	// Select the aggregate value for interval
 	queryTmpl :=
 		"SELECT fingerprint, %s" +
-			" toStartOfInterval(toDateTime(intDiv(timestamp_ms, 1000)), INTERVAL %d SECOND) as ts," +
+			" toStartOfInterval(toDateTime(intDiv(unix_milli, 1000)), INTERVAL %d SECOND) as ts," +
```
ankitnayan, line 120: I see column `timestamp_ms`  being renamed to `unix_milli` in the collector repo. Any significance?

pkg/query-service/app/metrics/v4/helpers/clauses.go
```diff
@@ -37,6 +37,17 @@ func GroupByAttributeKeyTags(tags ...v3.AttributeKey) string {
 	return strings.Join(groupTags, ", ")
 }
 
+func GroupByAttributeKeyTagsWithoutLe(tags ...v3.AttributeKey) string {
```
ankitnayan, line 40: where is this referenced?

srikanthccv:
Yes, the work I did is under the v4 endpoint, but the request/response payload is structurally v3 (added time, space aggregation fields) and the frontend will work with that. V4 request/response structure will be a breaking change with each signal sharing the common parts and extending to accommodate their signal-specific details.

Since we are discussing these details, let me share the migration plan we want to take as well. We will start maintaining the version for dashboards from now on and make upgrades gracefully from the UI side when possible. The path we discussed for metrics is, to use the dashboard version to decide whether to show time and space aggregation options or show the existing version. We will provide the instructions and scripts to move away but there is no instant migration for all users since we can't automatically migrate some aggregate operators without knowing the context of the metric.

pkg/query-service/app/clickhouseReader/reader.go
srikanthccv: The current scope is APM, since it uses the delta/cumulative to prepare an appropriate formula for apdex.

pkg/query-service/app/metrics/v4/cumulative/timeseries.go
srikanthccv: No strong reason other than making it more explicit that the format is epoch milli.

pkg/query-service/app/metrics/v4/helpers/clauses.go
srikanthccv: This is not needed anymore. I will remove it.

pkg/query-service/app/metrics/v4/helpers/sub_query.go
```diff
@@ -3,21 +3,54 @@ package helpers
...
+// start and end are in milliseconds
+func which(start, end int64) (int64, int64, string) {
+	// If time range is less than 6 hours, we need to use the `time_series_v4` table
+	// else if time range is less than 1 day and greater than 6 hours, we need to use the `time_series_v4_6hrs` table
+	// else we need to use the `time_series_v4_1day` table
+	var tableName string
+	if end-start <= sixHoursInMilliseconds {
+		// adjust the start time to nearest 1 hour
+		start = start - (start % (time.Hour.Milliseconds() * 1))
+		tableName = constants.SIGNOZ_TIMESERIES_v4_LOCAL_TABLENAME
```
ankitnayan, line 27: Does this table have 1hr replacement logic?

srikanthccv: Yes, https://github.com/SigNoz/signoz-otel-collector/blob/0485ef8e0482f4fd7c188c8dd5ecf7b92280e683/exporter/clickhousemetricsexporter/clickhouse.go#L453

pkg/query-service/app/parser.go
```diff
@@ -1046,6 +1045,25 @@ func ParseQueryRangeParams(r *http.Request) (*v3.QueryRangeParamsV3, *model.ApiE
 					}
 				}
 			}
+
+			var timeShiftBy int64
+			if len(query.Functions) > 0 {
+				for idx := range query.Functions {
+					function := &query.Functions[idx]
+					if function.Name == v3.FunctionNameTimeShift {
+						// move the function to the beginning of the list
```
ankitnayan, line 1054: why can't a user take the moving average of a time-shifted series?

pkg/query-service/app/queryBuilder/functions.go
```diff
@@ -281,6 +281,21 @@ func ApplyFunction(fn v3.Function, result *v3.Result) *v3.Result {
...
+		shift, ok := fn.Args[0].(float64)
+		if !ok {
+			return result
+		}
+		return funcTimeShift(result, shift)
+	}
+	return result
+}
+
+func funcTimeShift(result *v3.Result, shift float64) *v3.Result {
```
ankitnayan, line 294: this function cannot be used to compare to a previous interval right? Eg, I want to compare today's error rate with yesterday's error rate.

pkg/query-service/app/parser.go
srikanthccv: They can do it, it's just that the position of timeshift doesn't matter. Whether the data manipulation function comes after or before the timeshift produces the same value.

pkg/query-service/app/metrics/v4/helpers/sub_query.go
ankitnayan: right. thanks

pkg/query-service/app/clickhouseReader/reader.go
ankitnayan: ok

pkg/query-service/app/queryBuilder/functions.go
srikanthccv: No, it can be used and that is the goal of this function. Here https://github.com/SigNoz/signoz/pull/4445/files#diff-998df71bdc12a1398abbbdabd8dbab8fce344d7d59e1653dd2f14b9ac4e8911fR24-R27 we shift the time by x amount to get the data for past using the query. Since the result time will be in past data but the UI chart will have the current window, we adjust the timeshifted query to the current time window.

ankitnayan: right. Got it

pkg/query-service/app/parser.go
ankitnayan: and timeshift must be the 1st function to act in code as it changes the end and start and hence the input for the other functions to work on

ankitnayan (approved):

srikanthccv:
We will have the delta data backfilled in a week for span metrics and the tables used are v4. Here is the script I used to compare the delta and cumulative side by side https://gist.github.com/srikanthccv/882027fe4229cef8fa403f28fe8babd0. We will most likely shift to the delta + v4 tables for span metrics soon unless something strange comes up in further testing.

ankitnayan:
> We will have the delta data backfilled in a week for span metrics and the tables used are v4. Here is the script I used to compare the delta and cumulative side by side https://gist.github.com/srikanthccv/882027fe4229cef8fa403f28fe8babd0. We will most likely shift to the delta + v4 tables for span metrics soon unless something strange comes up in further testing.

But using delta is not just for quantiles right? The script compares quantiles

srikanthccv:
Right, but quantiles use the sum rate internally. If the quantiles are correct, we can assume the sum rate (by bucket) is correct as well.

Issue #2765, opened 2023-05-24
Ability to go from alert notification to explorer page


srikanthccv:
To be built after explorer pages. With both Explorer and Alert with the builder in place, when an alert notification is sent, the user should be able to navigate to the Explorer page with all the query context taken along with the time it fired.

ankitnayan:
@srikanthccv this is complete?

srikanthccv:
I left this open since what we have is signal-specific and doesn't include the cross-signal explorer pages. But there will be a lot of groundwork needed to make sure attributes follow the same convention (dot, underscore, snake case etc...) across signals to make that happen. Closing this for now.

PR #4446, opened 2024-01-28
feat: add alerts to explorer link in notification

srikanthccv:
### Summary

Part of #2765

#### Screenshots

NA

## Summary by CodeRabbit

- **New Features**
	- Enhanced the default selection of columns in the Traces List View.
	- Introduced new query parameter types for improved query management.
	- Added new functionalities in the threshold rules to support dynamic linking to logs and traces based on specific conditions.
- **Tests**
	- Implemented new tests for verifying the correct preparation of links to logs and traces.


pkg/query-service/rules/thresholdRule.go
```diff
@@ -625,6 +629,207 @@ func (r *ThresholdRule) prepareBuilderQueries(ts time.Time) (map[string]string,
...
+	compositeQuery := url.QueryEscape(url.QueryEscape(string(data)))
+
+	optionsData, _ := json.Marshal(options)
+	urlEncodedOptions := url.QueryEscape(string(optionsData))
+
+	return fmt.Sprintf("compositeQuery=%s&timeRange=%s&startTime=%d&endTime=%d&options=%s", compositeQuery, urlEncodedTimeRange, tr.Start, tr.End, urlEncodedOptions)
+}
+
+func (r *ThresholdRule) hostFromSource() string {
+	parsedUrl, err := url.Parse(r.source)
```
srikanthccv, line 826: The `source` is populated by the frontend, with the alert page. We are relying on it to provide the link to instance. In an ideal case, it should be from the orgs table where we know the tenant URL but that's not available in the query-service at the moment. If the source is missing (possibly from API created alert), then the link won't be clickable.

pkg/query-service/rules/thresholdRule.go
```diff
@@ -625,6 +629,207 @@ func (r *ThresholdRule) prepareBuilderQueries(ts time.Time) (map[string]string,
...
+	}
+
+	// Logs list view expects time in milliseconds
+	tr := timeRange{
+		Start:    ts.Add(-time.Duration(r.evalWindow)).UnixMilli(),
+		End:      ts.UnixMilli(),
+		PageSize: 100,
+	}
+
+	options := Options{
```
nityanandagohain, line 713: Can you describe a bit more on how this will be used by the frontend ?

srikanthccv: From what I understood
- `maxLines` refers to max lines per row i.e UI shows max 2 lines in each row for log record and then fills the rest with `...`
- `format` is list view
- SelectColumns is by default empty for logs in UI and we don't have any context in query-service either.

nityanandagohain: got it thanks 👍

nityanandagohain (approved):
LGTM! on the frontend we can add a popup or something saying that when formula queries are not supported.

pkg/query-service/rules/thresholdRule.go
ankitnayan: this will be a problem. We should store tenantID then in the query-service too

srikanthccv: I agree and it's not just for the related logs or the related traces but the main link to the alert page as well. However, storing the tenant info in the query-service should be addressed separately. We skip setting related logs/traces when the link is not usable.

pkg/query-service/rules/thresholdRule.go
```diff
@@ -625,6 +629,210 @@ func (r *ThresholdRule) prepareBuilderQueries(ts time.Time) (map[string]string,
...
+		}
+	}
+
+	return filterItems
+}
+
+func (r *ThresholdRule) prepareLinksToLogs(ts time.Time, lbls labels.Labels) string {
+	selectedQuery := r.GetSelectedQuery()
+
+	// TODO(srikanthccv): handle formula queries
```
ankitnayan, line 701: This can be a prominent usecase? Eg, error percentage?

srikanthccv: It's very rare to see formulas in the logs data source. It's usually alerting on counts. I looked at the customer who asked for this feature and you can see they are all count(env) with some filters. I don't think formulas is common use-case int logs/traces.

ankitnayan: ok

pkg/query-service/rules/thresholdRule.go
```diff
@@ -625,6 +629,207 @@ func (r *ThresholdRule) prepareBuilderQueries(ts time.Time) (map[string]string,
...
+// take us to the traces page with the filter serviceName=A
+// So for all the missing labels in the where clause, we add them as key = value
+//
+// Example for case 1:
+// Severity text IN (WARN, ERROR)
+// In this case, the Severity text will appear in the `lbls` if it were part of the group
+// by clause, in which case we replace it with the actual value for the notification
+// i.e Severity text = WARN
+// If the Severity text is not part of the group by clause, then we add it as it is
+func (r *ThresholdRule) fetchFilters(selectedQuery string, lbls labels.Labels) []v3.FilterItem {
```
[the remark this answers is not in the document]
ankitnayan: I guess this is the most crucial piece which is responsible for creating a correct where clause in the QB for the explorer pages?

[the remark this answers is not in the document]
ankitnayan: Can we say that it does not matter if the labels found in the alert sample exists in the selectedQuery or not?

pkg/query-service/rules/thresholdRule.go
```diff
@@ -625,6 +629,210 @@ func (r *ThresholdRule) prepareBuilderQueries(ts time.Time) (map[string]string,
...
+
+	if r.ruleCondition.CompositeQuery.QueryType == v3.QueryTypeBuilder &&
+		r.ruleCondition.CompositeQuery.BuilderQueries[selectedQuery] != nil &&
+		r.ruleCondition.CompositeQuery.BuilderQueries[selectedQuery].Filters != nil {
+
+		for _, item := range r.ruleCondition.CompositeQuery.BuilderQueries[selectedQuery].Filters.Items {
+			exists := false
+			for _, label := range lbls {
+				if item.Key.Key == label.Name {
+					// if the label is present in the where clause, replace it with key = value
```
ankitnayan, line 665: why do we need to check `if the label is present in the where clause`?

pkg/query-service/rules/thresholdRule.go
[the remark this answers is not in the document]
srikanthccv: >I guess this is the most crucial piece which is responsible for creating a correct where clause in the QB for the explorer pages?

Yes

>Can we say that it does not matter if the labels found in the alert sample exists in the selectedQuery or not?

No, it is relevant. Because the notification link should be more specific to the problem.

pkg/query-service/rules/thresholdRule.go
srikanthccv: If the same label exists in the WHERE clause we replace the label value we found in the alert sample. For instance, assume you had a label `serviceName IN ['frontend', 'redis', 'route']` and the result label contains `serviceName` with a value `route`. In this case, we replace the `serviceName IN ['frontend', 'redis', 'route']` with `serviceName=route`. We do this so the data you are looking at is a more drilled-down version rather than a plain alert filter.

pkg/query-service/rules/thresholdRule.go
ankitnayan: ok

pkg/query-service/rules/thresholdRule.go
ankitnayan: right

pkg/query-service/rules/thresholdRule.go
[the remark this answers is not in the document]
ankitnayan: got it at https://github.com/SigNoz/signoz/pull/4446#discussion_r1475947204

ankitnayan (approved):

PR #4457, opened 2024-01-29
fix: switching between logs display tabs

Vikrant2520:
### Summary

- the switching between the view tabs of logs explorer rendered the older data for a moment and then switched the tab 
- so rather than manually updating the state we rely on QB state to take care of the same after it triggers the API call

#### Screenshots

https://github.com/SigNoz/signoz/assets/54737045/b23ba95d-3880-4ee5-9cfc-625f3bddfd45


YounixM:
@Vikrant2520: How would be handling the error in this case ?

Rajat-Dabade (approved):

YounixM (approved):

Issue #4038, opened 2023-11-23
PromQL based alerts should support all the times


srikanthccv:
The way it works currently is an instant query is run and it fires an alert if the value crosses the threshold. The current design of the PromQL alert eval can't support all the time. For this reason, we don't give all the times option for this query type. It should work as `For` in alertmanager.

Issue #3917, opened 2023-11-07
Lookback delta discrepancy b/w RangeQuery and InstantQuery for promql


srikanthccv:
## Bug description

Metrics that are reported with some delay (common on AWS) require an offset in the query to work. For range queries, there is an additional look-back delta of 5 minutes which doesn't exist for alerts (since it uses Instant query).

ankitnayan:
> For range queries, there is an additional look-back delta of 5 minutes

This is not clear. Please explain more

srikanthccv:
That description is not fully correct. Today PromQL alerts run an instant query for each evaluation interval. The instant selector uses the value from the most recent sample(s). The maximum lookback period to fetch this recent sample is (now-5min, now); there can be more but it will use the latest one. The range query is an instant query at each step interval. 

The main issue here is, that metrics from AWS arrive late, commonly 10m late. Since the query service uses the instant query for alerts and AWS metrics arrive late there will never be an alert because of this. This is where the major issue is with what they see in UI and what happens in query service. When they visit the alert page, which uses a range query, they may see some data which makes them think the alert should trigger. The linked PR updates alerts to use the range query for eval window.

ankitnayan:
thanks. This is much clear now

PR #4461, opened 2024-01-30
fix: update prom rule to use range query

srikanthccv:
### Summary

Fixes https://github.com/SigNoz/signoz/issues/4038
Fixes https://github.com/SigNoz/signoz/issues/3917

### What do we support today:

Instant query with only `At least once` comparison condition.

Screenshot 2024-02-01 at 9 07 30 AM

### Issues with the current promql alerts

- `On average`, `In total` and `All the times` are not supported
- Evaluation of the range is not supported.
- Inconsistent exp with the other two modes of alert creation.

What users see in the chart UI when working with promql comes from the Range query API. They expect the same when creating alerts. For instance, there is no way to set an alert if the latency is greater than 800ms for the last 10 minutes. It's common to have occasional spikes and triggering alerts with `At least once` can be noisy.

This PR moves to RangeQuery instead of InstantQuery. 

## Summary by CodeRabbit

- **New Features**
	- Introduced Prometheus-specific evaluation windows for alert rules.
	- Enhanced alert query functionality to support time range and interval parameters.
- **Refactor**
	- Improved clarity and extensibility of alert rule options rendering.
	- Updated alert rule logic to handle different match types and comparison operations more effectively.
- **Tests**
	- Added new tests for evaluating Prometheus rules based on various scenarios.
- **Documentation**
	- Updated comments for better clarity in code documentation.


ankitnayan:
>On average, In total and All the times are supported

You mean `not supported` in the issue description?

srikanthccv:
Yes

ankitnayan (approved):

YounixM (approved):

frontend/src/container/FormAlertRules/RuleOptions.tsx
```diff
@@ -98,38 +98,42 @@ function RuleOptions({
 		</InlineSelect>
 	);
 
-	const renderPromMatchOpts = (): JSX.Element => (
+	const onChangeEvalWindow = (value: string | unknown): void => {
```
YounixM, line 101: We can omit the unknown type.

frontend/src/container/FormAlertRules/RuleOptions.tsx
```diff
@@ -98,38 +98,42 @@ function RuleOptions({
...
-			style={{ minWidth: '130px' }}
-			value={alertDef.condition?.matchType}
-			onChange={(value: string | unknown): void => handleMatchOptChange(value)}
+			defaultValue={defaultEvalWindow}
+			style={{ minWidth: '120px' }}
+			value={alertDef.evalWindow}
+			onChange={onChangeEvalWindow}
 		>
-			<Select.Option value="1">{t('option_atleastonce')}</Select.Option>
+			<Select.Option value="5m0s">{t('option_5min')}</Select.Option>
```
YounixM, line 117: Can use a const array with options and loop through

frontend/src/container/FormAlertRules/RuleOptions.tsx
```diff
@@ -98,38 +98,42 @@ function RuleOptions({
...
-			onChange={(value: string | unknown): void => {
-				const ew = (value as string) || alertDef.evalWindow;
-				setAlertDef({
-					...alertDef,
-					evalWindow: ew,
-				});
-			}}
+			onChange={onChangeEvalWindow}
 		>
 			<Select.Option value="5m0s">{t('option_5min')}</Select.Option>
```
YounixM, line 134: Same as above

srikanthccv:
The comments are non-blocking. I will merge this for now and address them in follow up PR.

PR #4473, opened 2024-01-31
Revert "fix: Logs UI: querybuildersearch: avoid emptying out query on sourceKeys update if tags are yet to be populated "

YounixM:
Reverts SigNoz/signoz#4355

## Summary by CodeRabbit

- **Bug Fixes**
	- Fixed an issue where queries were not updating under specific conditions related to tag length.


YounixM:
It's breaking clearing the current tags. We have to find another way to handle this.

Vikrant2520 (approved):

YounixM:
Merging this as this is a regression. This will revert the fix related to the above PR. We can take it up next.

PR #4492, opened 2024-02-05
chore: bump clickhouse-server to 24.1.2-alpine

srikanthccv:
#### Screenshots

NA


nityanandagohain (approved):
After going through the changelog don't see something backward incompatabile that will  break. But should be in staging server for sometime before a public release.

prashant-shahi (approved):

PR #4496, opened 2024-02-05
fix: don't throw error while fetching orgname

nityanandagohain:
if there are no rows it doesn't throw an error but since the err variable was initialized in the same block again with orgNames it caused the error. Have fixed and updated the description.

## Summary by CodeRabbit

- **Refactor**
	- Simplified the logic for retrieving organization names in the usage upload process, enhancing efficiency by reducing error handling complexity.

### DIFF
--- a/ee/query-service/usage/manager.go
+++ b/ee/query-service/usage/manager.go
@@ -139,17 +139,11 @@ func (lm *Manager) UploadUsage() {
 
 	zap.S().Info("uploading usage data")
 
-	// Try to get the org name
 	orgName := ""
-	orgNames, err := lm.modelDao.GetOrgs(ctx)
-	if err != nil {
-		zap.S().Errorf("failed to get org data: %v", zap.Error(err))
-	} else {
-		if len(orgNames) != 1 {
-			zap.S().Errorf("expected one org but got %d orgs", len(orgNames))
-		} else {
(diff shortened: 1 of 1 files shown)


ee/query-service/usage/manager.go
```diff
@@ -139,17 +139,11 @@ func (lm *Manager) UploadUsage() {
...
-	if err != nil {
-		zap.S().Errorf("failed to get org data: %v", zap.Error(err))
-	} else {
-		if len(orgNames) != 1 {
-			zap.S().Errorf("expected one org but got %d orgs", len(orgNames))
-		} else {
-			orgName = orgNames[0].Name
-		}
+	// fetching orgname will result in error only when org is not set, so ignoring it.
+	orgNames, _ := lm.modelDao.GetOrgs(ctx)
```
srikanthccv, line 144: It must be returning something similar to `sql.NoRows` or something? I would check for that specific case to ignore instead of ignoring every kind of error.

nityanandagohain: Thanks, have corrected it.

So basically if there are no rows it doesn't throw an error but since the err variable was initialized in the same block again with orgNames it caused the error. Have fixed and updated the description.

srikanthccv (approved):

PR #4516, opened 2024-02-08
feat: support cache in logs

### LINKED ISSUE #4513: LOGS[BE]: Add cache support
Add cache support in logs

nityanandagohain:
Fixes #4513 

## Summary by CodeRabbit

- **New Features**
	- Introduced caching for log queries to enhance performance.
	- Added functionality to handle log queries with various conditions and limits.
- **Refactor**
	- Improved logic for preparing log queries with caching.
	- Enhanced key generation for log queries to include more detailed information.


pkg/query-service/app/queryBuilder/query_builder.go
```diff
@@ -320,7 +320,46 @@ func (c *cacheKeyGenerator) GenerateKeys(params *v3.QueryRangeParamsV3) map[stri
...
 	for queryName, query := range params.CompositeQuery.BuilderQueries {
-		if query.Expression == queryName && query.DataSource == v3.DataSourceMetrics {
+		if query.Expression == queryName && query.DataSource == v3.DataSourceLogs {
+			var parts []string
+
+			// We need to build uniqe cache query for BuilderQuery
+			parts = append(parts, fmt.Sprintf("source=%s", query.DataSource))
+			parts = append(parts, fmt.Sprintf("step=%d", query.StepInterval))
+			parts = append(parts, fmt.Sprintf("aggregate=%s", query.AggregateOperator))
+			parts = append(parts, fmt.Sprintf("limit=%d", query.Limit))
```
srikanthccv, line 330: Ideally, the limit and order by etc... shouldn't be part of the cache key because they don't change the main query but the end result processing. Please create an issue to move this post-processing logic to query-service. We already have things in place to do this today but you address it separately.

pkg/query-service/app/queryBuilder/query_builder.go
```diff
@@ -320,7 +320,46 @@ func (c *cacheKeyGenerator) GenerateKeys(params *v3.QueryRangeParamsV3) map[stri
...
+				for idx, orderBy := range query.OrderBy {
+					parts = append(parts, fmt.Sprintf("orderBy-%d=%s", idx, orderBy.CacheKey()))
+				}
+			}
+
+			if len(query.Having) > 0 {
+				for idx, having := range query.Having {
+					parts = append(parts, fmt.Sprintf("having-%d=%s", idx, having.CacheKey()))
+				}
+			}
```
srikanthccv, line 358: I guess the same reasoning as above applies here? that this makes the sub-optimal use of caching as having come under post-processing logic.

pkg/query-service/app/queryBuilder/query_builder.go
```diff
@@ -320,7 +320,46 @@ func (c *cacheKeyGenerator) GenerateKeys(params *v3.QueryRangeParamsV3) map[stri
...
+
+			key := strings.Join(parts, "&")
+			keys[queryName] = key
```
srikanthccv, line 361: Below there is a `isMetricExpression` the skips the cache key general formula that involves any other source, you should be updating that as well? Please add some tests

pkg/query-service/app/querier/helper.go
```diff
@@ -284,18 +384,28 @@ func (q *querier) runBuilderExpression(
...
 
+	var mergedSeriesData []byte
+	missedSeriesLen := len(missedSeries)
+	if missedSeriesLen > 0 && !params.NoCache && q.cache != nil {
+		// caching the data
+		var err error
+		mergedSeriesData, err = json.Marshal(mergedSeries)
+		if err != nil {
+			zap.S().Error("error marshalling merged series", zap.Error(err))
+			return
```
srikanthccv, line 395: You don't want to return here because there was some issue with marshalling the series. The return here is not the same as the return after `ch <- channelResult`. Just log the error and proceed. And handle the `mergedSeriesData` before `q.cache.Store`

pkg/query-service/app/querier/helper.go
```diff
@@ -208,14 +292,30 @@ func (q *querier) runBuilderQuery(
...
-		err = q.cache.Store(cacheKey, mergedSeriesData, time.Hour)
+	}
+
+	// response doesn't need everything
+	filterCachedPoints(mergedSeries, params.Start, params.End)
+	ch <- channelResult{
+		Err:    nil,
+		Name:   queryName,
+		Series: mergedSeries,
+	}
```
srikanthccv, line 314: There is additional `ch <- channelResult ` that is not removed above in this function, why?

pkg/query-service/app/querier/helper.go
```diff
@@ -208,14 +292,30 @@ func (q *querier) runBuilderQuery(
...
+
+	var mergedSeriesData []byte
+	missedSeriesLen := len(missedSeries)
+	if missedSeriesLen > 0 && !params.NoCache && q.cache != nil {
+		// caching the data
+		var err error
+		mergedSeriesData, err = json.Marshal(mergedSeries)
 		if err != nil {
 			zap.S().Error("error marshalling merged series", zap.Error(err))
 			return
```
srikanthccv, line 304: Same, don't return if there are errors related to cache stuff. If you see other parts of the such as this https://github.com/SigNoz/signoz/blob/ab4f6adb198363c4efa7bf00d67342b80f38d74f/pkg/query-service/app/querier/helper.go#L282-L284, it doesn't return prematurely. The context when return is used is important.

pkg/query-service/app/queryBuilder/query_builder.go
nityanandagohain: we need it for logs

ex:- order by route asc with limit 1 and order by route desc with limit 1 will result in different response.

For logs we generate different queries for the both above cases.

pkg/query-service/app/querier/helper.go
nityanandagohain: make sense, thanks

pkg/query-service/app/queryBuilder/query_builder.go
nityanandagohain: created a different function altogether for logs.

pkg/query-service/app/queryBuilder/query_builder.go
nityanandagohain: Yeah this can come under post processing, creating an issue for this.

pkg/query-service/app/queryBuilder/query_builder.go
nityanandagohain: added to https://github.com/SigNoz/signoz/issues/4527

pkg/query-service/app/querier/helper.go
```diff
@@ -35,59 +90,88 @@ func (q *querier) runBuilderQuery(
...
-				params.Start,
-				params.End,
-				params.CompositeQuery.QueryType,
-				params.CompositeQuery.PanelType,
-				builderQuery,
-				logsV3.Options{GraphLimitQtype: constants.SecondQueryGraphLimit, PreferRPM: preferRPM},
-			)
+			series, err := q.execClickHouseQuery(ctx, query)
+			ch <- channelResult{Err: err, Name: queryName, Query: query, Series: series}
+			return
```
srikanthccv, line 104: With the return here, will the result ever get cached? There is no cache key to begin with there will never be one because we return.

nityanandagohain: This logic is for queries is where caching is not required.

nityanandagohain: cacheKeys are generated by the GenerateKeys, so it will always be there for queries that needs to use cache.

srikanthccv (dismissed):
Please test various queries once. Are regular full text search also cached or only aggregation queries. I think it should only include aggregation queries.

nityanandagohain:
Yes it's only for aggreagte queries.

PR #4523, opened 2024-02-10
[Refactor]: added extra param to check isdepend on query build

Rajat-Dabade:
Loom: https://www.loom.com/share/02c2606d055e43d0bc812e671fb8632e

Root Cause Analysis (RCA):

Issue:
The isEnableQuery flag depends on the URL composite query. However, in the onboarding flow, where no query is present in the URL, the isEnableQuery flag is always false. To address this limitation and enable the query regardless of isEnableQuery, an additional prop isDependentOnQB has been introduced.

Root Causes:

Dependency on URL composite query: The isEnableQuery flag's functionality is tied to the presence of a query in the URL. This design assumes that queries are always in the URL, which is not the case during the onboarding flow.

Solution:
To address the issue and ensure that queries can be enabled regardless of the presence of URL queries, an additional prop isDependentOnQB has been introduced. This prop allows explicit control over the query's enablement, bypassing the dependency on the URL composite query.

## Summary by CodeRabbit

- **New Features**
	- Enhanced the control flow in the Onboarding process for a smoother user experience.
	- Improved query builder functionality with additional customization options.


Vikrant2520 (dismissed):

frontend/src/hooks/queryBuilder/useGetExplorerQueryRange.ts
```diff
@@ -32,11 +33,11 @@ export const useGetExplorerQueryRange = (
 	const isEnabled = useMemo(() => {
 		if (!options) return isEnabledQuery;
 		if (typeof options.enabled === 'boolean') {
-			return isEnabledQuery && options.enabled;
+			return !isDependentOnQB || (isEnabledQuery && options.enabled);
```
YounixM, line 36: This would fail if options.enabled is conditionally false.

YounixM (changes requested):
This would fail if options.enabled is conditionally false.

frontend/src/hooks/queryBuilder/useGetExplorerQueryRange.ts
Rajat-Dabade: Updated condition

frontend/src/hooks/queryBuilder/useGetExplorerQueryRange.ts
```diff
@@ -32,11 +33,13 @@ export const useGetExplorerQueryRange = (
 	const isEnabled = useMemo(() => {
 		if (!options) return isEnabledQuery;
 		if (typeof options.enabled === 'boolean') {
-			return isEnabledQuery && options.enabled;
+			return (
+				(!isDependentOnQB && options.enabled) || (isEnabledQuery && options.enabled)
```
YounixM, line 37: option 1:

if(isDependentOnQB) {
	return isEnabledQuery && options.enabled
} else {
	return options.enabled
}

option 2:

options.enabled && (!isDependentOnQB || isEnabledQuery)

Rajat-Dabade: Updated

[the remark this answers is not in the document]
Rajat-Dabade: Done

YounixM (approved):

PR #4600, opened 2024-02-26
feat: route to the onboarding flow when clicking the sending logs in case of no logs

Vikrant2520:
### Summary

- refactor the onboarding flow logic to append `module` in the URL params when selecting it.
- handle the route from logs / traces to get started in the case of no data yet

#### Related Issues / PR's

https://linear.app/signoz-io/issue/SIG-534/empty-logs-text-update

#### Screenshots

NA

## Summary by CodeRabbit

- **New Features**
	- Introduced new routes for application monitoring, logs management, and infrastructure monitoring, enhancing the user's navigation experience.
	- Expanded full-screen mode to support additional monitoring and management routes.
	- Enhanced the onboarding process with improved navigation and module selection based on user choices.
	- Updated the top navigation bar to exclude new monitoring and management routes, streamlining user interface.
- **Bug Fixes**
	- Fixed an issue where users could not dynamically navigate to different monitoring or management sections based on the data source selected.


srikanthccv:
The simple re-route logic will be annoying because it could be possible that the user sets up the logging by following instructions but stops sending data for some reason or it could be that their ingest pattern is not continuous logging. At this point, they shouldn't be redirected because they already completed the onboarding workflow and sent some logs. The only time the user should be redirected is when they are on the platform for the first time and didn't send any logs/traces in the past.

Vikrant2520:
> The simple re-route logic will be annoying because it could be possible that the user sets up the logging by following instructions but stops sending data for some reason or it could be that their ingest pattern is not continuous logging. At this point, they shouldn't be redirected because they already completed the onboarding workflow and sent some logs. The only time the user should be redirected is when they are on the platform for the first time and didn't send any logs/traces in the past.

@srikanthccv we are not re-routing the user to the onboarding page , we render the logs screen itself and on click of `Start Sending Logs` we route them to onboarding instead of `docs page` , Just wondering if there was a way we can tell if the user has sent logs before and they are empty in the current time frame ? And if yes then we might need to think of better empty state in case of no logs (with logs being sent previously )

srikanthccv:
The specifics of re-route are not important. It is annoying if you ask to send logs instead of showing no logs when they already completed the onboarding workflow and sent some logs.

srikanthccv:
I don't know the details about the onboarding related things but we shouldn't show that screen if the already sent some logs in the past and completed the onboarding flow.

Rajat-Dabade (approved):

YounixM (approved):

PR #4607, opened 2024-02-27
feat: support for timeshift in logs

### LINKED ISSUE #4583: LOGS[BE]: Support for timeshift in logs v3
Support for timeshift in logs v3

nityanandagohain:
Fixes #4583 

* The runQueryBuilder checks if there is sifftBy and then modifies the start and end timestamp so that correct query is generated and caching is also properly used.
* The applyFunction function makes sure that even though the results has shifted timestamp it  is moved back to the actual timestamp so that it can be fitted in the panel.


pkg/query-service/app/metrics/v4/query_builder.go
```diff
@@ -21,11 +21,6 @@ func PrepareMetricQuery(start, end int64, queryType v3.QueryType, panelType v3.P
 
 	start, end = common.AdjustedMetricTimeRange(start, end, mq.StepInterval, mq.TimeAggregation)
 
-	if mq.ShiftBy != 0 {
```
nityanandagohain, line 24: removed this from here as it is taken care in the `runBuilderQuery` function which takes care of caching as well

pkg/query-service/app/http_handler.go
```diff
@@ -3403,7 +3409,7 @@ func applyFunctions(results []*v3.Result, queryRangeParams *v3.QueryRangeParamsV
 	for idx, result := range results {
 		builderQueries := queryRangeParams.CompositeQuery.BuilderQueries
 
-		if builderQueries != nil && (builderQueries[result.QueryName].DataSource == v3.DataSourceMetrics) {
+		if builderQueries != nil && (builderQueries[result.QueryName].DataSource == v3.DataSourceMetrics || builderQueries[result.QueryName].DataSource == v3.DataSourceLogs) {
```
srikanthccv, line 3412: Let's just remove this data source condition and add functions for all signals

srikanthccv (commented):
Please add couple of tests

srikanthccv (approved):

srikanthccv:
Please get the frontend capability added as well.

PR #4618, opened 2024-02-28
feat: support drag select in chart - alerts page

YounixM:


Vikrant2520 (changes requested):
We should handle back navigation here as well

PR #4657, opened 2024-03-06
[Refactor]: Height for log explorer list panel

Rajat-Dabade:
Fix the height issue for logs.

https://github.com/SigNoz/signoz/assets/26198062/30ccedb0-010a-4dbb-8680-d0d4aaa5a70d

### DIFF
--- a/frontend/src/components/Logs/TableView/config.ts
+++ b/frontend/src/components/Logs/TableView/config.ts
@@ -14,12 +14,12 @@ export function getDefaultCellStyle(isDarkMode?: boolean): CSSProperties {
 		lineHeight: '18px',
 		letterSpacing: '-0.07px',
 		marginBottom: '0px',
+		minWidth: '10rem',
 	};
 }
 
 export const defaultTableStyle: CSSProperties = {
 	minWidth: '40rem',
-	maxWidth: '40rem',
 };
 
 export const defaultListViewPanelStyle: CSSProperties = {
--- a/frontend/src/container/LogsExplorerList/LogsExplorerList.style.scss
+++ b/frontend/src/container/LogsExplorerList/LogsExplorerList.style.scss
@@ -1,4 +1,7 @@
 .logs-list-view-container {
+	flex: 1;
+	display: flex;
+	flex-direction: column;
 	font-family: 'Space Mono', monospace;
 	font-size: 14px;
 	font-weight: 400;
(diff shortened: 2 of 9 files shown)


Vikrant2520 (approved):

Vikrant2520 (approved):

frontend/src/components/Logs/TableView/config.ts
```diff
@@ -14,12 +14,12 @@ export function getDefaultCellStyle(isDarkMode?: boolean): CSSProperties {
 		lineHeight: '18px',
 		letterSpacing: '-0.07px',
 		marginBottom: '0px',
+		minWidth: '10rem',
```
YounixM, line 17: ??

YounixM:
@Rajat-Dabade: Please add an RCA around. It will help understanding the internal functioning and requirement for change.

frontend/src/components/Logs/TableView/config.ts
Rajat-Dabade: This is not super important, but it looks cleaner with minWidth and also we have a horizontal scroll, so there will be no problem. 

With MinWidth: 

https://github.com/SigNoz/signoz/assets/26198062/e8b1715c-df65-4de0-a0fb-4d03b9b23b99

Without minWidth:

https://github.com/SigNoz/signoz/assets/26198062/a11b8376-0edc-49f1-8430-a7aeecac779f

Rajat-Dabade:
Root Cause Analysis (RCA):

Issue Description:
The last `<div>` element within a flex container does not take up the full remaining height, despite setting `display: flex;` and `flex-direction: column;` on the container.

As you can see in the image the last `div` is not taking the full height available. 

Screenshot 2024-03-11 at 3 03 43 PM




Root Cause:

The reason for the last `<div>` not taking the full remaining height within the flex container is due to the default behaviour of flex items in a flex container. By default, flex items do not automatically stretch to fill the remaining space along the cross-axis (in this case, the vertical axis since the flex-direction is set to column).

Analysis:
Flexbox Behavior: In a flex container with `flex-direction: column;` , the main axis runs vertically from top to bottom, while the cross-axis runs horizontally. By default, flex items only occupy the height required by their content along the cross-axis.

Lack of Flex Grow Property: Without explicitly specifying the `flex: 1` property on the last `<div>` element, it does not have instructions to expand and fill the remaining space within the flex container along the cross-axis.

Solution:

To address the issue and ensure that the last `<div>` takes up the full remaining height within the flex container, the `flex: 1` property needs to be applied to the last `<div>` element. Setting `flex: 1;` set `flex-grow: 1`  on the last <div> instructs it to expand and fill the remaining space along the cross-axis, thereby resolving the issue.

YounixM (approved):

Issue #4654, opened 2024-03-05
Handle high cardinality span names


makeavish:
According to Otel spec span names MUST not contain high cardinal IDs. But some users accidentally make instrumentation mistakes which causes high cardinal span names.
It breaks service list on SigNoz home page. There should in product handling of such errors and guide user to fix their instrumentation.

ankitnayan:
@srikanthccv did we plan for anything more when we shipped the last solution?

srikanthccv:
No. The next remaining step is updating SDKs to send the data for https://github.com/SigNoz/signoz-otel-collector/pull/296

PR #4666, opened 2024-03-08
chore: limit number of top level operations in services list

srikanthccv:
### Summary

Part of https://github.com/SigNoz/signoz/issues/4654

This is the schema of the `top_level_operations`

```sql
CREATE TABLE signoz_traces.top_level_operations
(
    `name` LowCardinality(String) CODEC(ZSTD(1)),
    `serviceName` LowCardinality(String) CODEC(ZSTD(1)),
    `time` DateTime DEFAULT now() CODEC(ZSTD(1))
)
ENGINE = ReplacingMergeTree
ORDER BY (serviceName, name)
TTL time + toIntervalMonth(1)
SETTINGS index_granularity = 8192
```

The table contains the most recent instance of the tuple (serviceName, name). And the time value is `now()`. We can use the `time` to discard the old operations that were created before the start time. 

This change

- Discard the old operation
- Use a maximum of 1500 operations for the `IN` query (this should be enough?) 
- Send all service ops to the client so we can show them to the user.


srikanthccv:
We considered showing the users some of the selected top-level operations along with the warning text

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -810,6 +824,22 @@ func (r *ClickHouseReader) GetServices(ctx context.Context, queryParams *model.G
 			defer func() { <-sem }()
 			var serviceItem model.ServiceItem
 			var numErrors uint64
+
+			// Even if the total number of operations within the time range is less and the all
+			// the top level operations are high, we want to warn to let user know the issue
+			// with the instrumentation
+			serviceItem.DataWarning = model.DataWarning{
+				TopLevelOps: (*allTopLevelOps)[svc],
```
ankitnayan, line 832: sending all the values of `allTopLevelOps` in field `DataWarning`? This is a lot of unwanted data?

srikanthccv: It's not unwanted. It is possible that In the last 30 minutes, the unique top-level ops may be well within the limit, however, overall customers sending operations that are cardinal enough to cause issues. We will show warning message even if the API works without an issue for the selected time but has underlying data problems.

ankitnayan: >but has underlying data problems
where is this logic?

PR #4677, opened 2024-03-11
fix: update the md file for integrations config

Vikrant2520:
### Summary

- update the `.md` file for the integrations config.
- the markdown styling are dependent on the `bash` script being present in the code blocks 

#### Screenshots

NA


YounixM (approved):

Vikrant2520:
> LGTM 👍
> Do all code blocks need to have bash annotation or should it be content specific? Eg: Should yaml blocks have yaml etc?

We have yaml as well, we can check this once we start finalising the config, will share the details with you soon

PR #4681, opened 2024-03-12
chore: add LogCommentEnricher middleware

srikanthccv:
#### Screenshots

NA

### DIFF
--- a/pkg/query-service/app/server.go
+++ b/pkg/query-service/app/server.go
@@ -9,7 +9,9 @@ import (
 	"net"
 	"net/http"
 	_ "net/http/pprof" // http profiler
+	"net/url"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/gorilla/handlers"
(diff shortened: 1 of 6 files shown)


pkg/query-service/app/server.go
```diff
@@ -305,6 +308,65 @@ func loggingMiddleware(next http.Handler) http.Handler {
...
+				viewName = referrerURL.Query().Get("viewName")
+			} else if strings.Contains(path, "/services") {
+				page = "services"
+				tab = referrerURL.Query().Get("tab")
+				if tab == "" {
+					tab = "OVER_METRICS"
+				}
+			}
+		} else {
+			client = "api"
```
ankitnayan, line 352: `api` client calls would not have any log information?

ankitnayan:
A better solution would be to pass the product info in the api call to backend from the frontend and use that to add product info in the logs. We can get started with this and later switch to that.

pkg/query-service/app/server.go
srikanthccv: It's just a query range API call from user code so no other info can be added.

srikanthccv:
That is true but @YounixM was reluctant to that idea for now and we have sufficient information using the referrer and don't require changes in the frontend.

srikanthccv:
And his reservations are valid because it's generally not a good idea to put the things that don't need to be in the request payload. All such product info for monitoring sent to the backend would still be in some form of the header.

PR #4682, opened 2024-03-12
fix: use same time window as the logs/traces query

srikanthccv:
#### Screenshots

NA

### DIFF
--- a/pkg/query-service/rules/thresholdRule.go
+++ b/pkg/query-service/rules/thresholdRule.go
@@ -823,10 +823,11 @@ func (r *ThresholdRule) prepareLinksToLogs(ts time.Time, lbls labels.Labels) str
 		return ""
 	}
 
+	q := r.prepareQueryRange(ts)
 	// Logs list view expects time in milliseconds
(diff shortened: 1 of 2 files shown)


ankitnayan (approved):

pkg/query-service/rules/thresholdRule.go
```diff
@@ -823,10 +823,11 @@ func (r *ThresholdRule) prepareLinksToLogs(ts time.Time, lbls labels.Labels) str
 		return ""
 	}
 
+	q := r.prepareQueryRange(ts)
 	// Logs list view expects time in milliseconds
 	tr := timeRange{
-		Start:    ts.Add(-time.Duration(r.evalWindow)).UnixMilli(),
-		End:      ts.UnixMilli(),
+		Start:    q.Start,
```
ankitnayan, line 829: do we need to convert to nanoSeconds as logs work in nanoSeconds precision?

srikanthccv: The logs list view expects them in millis.

Issue #4693, opened 2024-03-13
.NET notation of SEVERITY_TEXT_TYPE not implemented 


DanikRaikhlin:
## Bug description
With [this PR](https://github.com/SigNoz/signoz/pull/4643) nice color indication for log levels is implemented however the .NET notation of Trace, Debug, Information, Warning, Error and Critical was not implemented and thus colors within logs are not working.

<img width="1890" alt="image" src="https://github.com/SigNoz/signoz/assets/2235048/cfada407-41ae-47b9-92af-c4aa133a4bb2">

## Expected behavior

Colors should be woring 

## How to reproduce

Use .NET to produce logs.

## Version information 
* **Signoz version**: 0.41.0
* **Browser version**:
* **Your OS and version**:
* **Your CPU Architecture**(ARM/Intel): 

## Additional context

The PR to fix this is very minor so i've taken the liberty to make it and will link this issue to it.

PR #4694, opened 2024-03-13
fix: use severity_number to determine severity_text_type

DanikRaikhlin:
### Summary

Add .NET notations for SEVERITY_TEXT_TYPE and use them to fix log level coloring.

#### Related Issues / PR's

Issue : #4693

#### Screenshots

image


srikanthccv:
Hello, I would solve this issue differently. I would use the `severity_number` to show the relevant colour schema for logs. This will work regardless of the severity text which can be an arbitrary string and varies from one system to another.

DanikRaikhlin:
@srikanthccv I fully agree that would be a better solution, I'll adjust the PR.

DanikRaikhlin:
PR updated to use severity_number

srikanthccv:
LGTM, i'll let the owners of the code review it.

Issue #3568, opened 2023-09-15
Applying legends should not need to run `Stage and Run Query` again


ankitnayan:
Legend is purely frontend filtering and should not need to do api call again to apply.

palashgdev:
@YounixM how we are thinking to solve this one ? though right now we are taking legend from composite query itself

PR #4720, opened 2024-03-19
fix: [SIG-567]: prevented stage-&-run API on legend change

SagarRajput-7:
### Summary

Prevented stage-&-run API on legend change:
- we were updating legends based on api-response earlier hence this required us to run `stage-&-run` query API every time  we want to update the `legend format`.
- Added other option to update `legend` format from current query coming from `useQueryBuilder` as this was getting updated on every `onChange` event of the legend format input.
- Have safety check so that when the current query is not provided then the updation will happen as previously.

#### Related Issues / PR's

#3568 

https://github.com/SigNoz/signoz/assets/162284829/5c9caba8-e4ee-438b-9e5c-5fac3d06eb78

#### Screenshots

NA


Vikrant2520 (approved):

Rajat-Dabade (approved):

YounixM:
Can you please add the details around the change for this?

YounixM (approved):

Issue #1101, opened 2022-05-05
Error message is not informative when incorrect current password is entered


pranay01:
## Bug description

Error message is not informative when incorrect current password is entered

*If this affects the front-end, screenshots would be of great help.*  

https://www.loom.com/share/7554baa8a38b4b3aac5c066f3985fe59

## Expected behavior

The error message should say something like `invalid current password` 

## How to reproduce

1. Go to my account from navbar pop up
2. Enter incorrect current password
3. Enter a different new password and try to change it. The error message displayed only says ` Something went wrong` which is not very informative

## Version information 
* **Signoz version**: v0.8.0-rc5
* **Browser version**: Chome
* **Your OS and version**: macOs 12.3
* **Your CPU Architecture**(ARM/Intel):  ARM

palashgdev:
if the status code is 500 frontend will show Something went wrong

gxwilkerson415:
@pranay01 can i be assigned to this issue?

YounixM:
Fixed by https://github.com/SigNoz/signoz/pull/4733

Closing the issue.

PR #4733, opened 2024-03-21
(fix) send 403 on wrong password entry during change password operation

grandwizard28:
### Summary

Incorrect status code is being sent when a user enters wrong password while trying to update their password!

#### Related Issues / PR's

https://github.com/SigNoz/signoz/issues/1101

#### Screenshots

NA

- Change Password Screen


pkg/query-service/app/http_handler.go
```diff
@@ -2364,7 +2364,7 @@ func (aH *APIHandler) changePassword(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := auth.ChangePassword(context.Background(), req); err != nil {
-		if aH.HandleError(w, err, http.StatusInternalServerError) {
+		if aH.HandleError(w, err, http.StatusForbidden) {
```
makeavish, line 2367: There can be other cases as well.
Only return status forbidden only when password is invalid https://github.com/SigNoz/signoz/blob/0df86454ce42796ec9b1015aa0fadd6141f254a1/pkg/query-service/auth/auth.go#L245

grandwizard28:
This is how are we handling `/login` as well
```
resp, err := auth.Login(context.Background(), req)
	if aH.HandleError(w, err, http.StatusUnauthorized) {
		return
	}
```

There were multiple cases to consider but we just focused on the `StatusUnauthorized` case!

Is it safe to presume we need to handle multiple cases in login as well?

srikanthccv (approved):

makeavish (approved):

PR #4736, opened 2024-03-21
feat: [SIG-571]: added support for has and nhas operator for json filter

### LINKED ISSUE #3629: LOGS[FE]: JSON filter operator suggestion
With the addition of support for writing JSON query the has and nhas operators should be suggsted when there is a key which ends with `[*]`

Vikrant2520:
### Summary

- added support for `has` and `nhas` operators when the key ends with `[*]` and starts with `body.`

#### Related Issues / PR's

Fixes #3629 

#### Screenshots

https://github.com/SigNoz/signoz/assets/54737045/850e2bdf-977f-4038-9929-3b705413b415

### DIFF
--- a/frontend/src/hooks/queryBuilder/useOperators.ts
+++ b/frontend/src/hooks/queryBuilder/useOperators.ts
@@ -1,4 +1,7 @@
-import { QUERY_BUILDER_OPERATORS_BY_TYPES } from 'constants/queryBuilder';
+import {
+	OPERATORS,
+	QUERY_BUILDER_OPERATORS_BY_TYPES,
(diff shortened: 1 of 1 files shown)


Rajat-Dabade (dismissed):

nityanandagohain (changes requested):
* for check add prefix `body.` as well along with `[*]` .
* Once the check is passed, don't suggest other operators other than has and nhas.

This is because has and nhas works with only json full text search as of now.

nityanandagohain (approved):

YounixM (approved):

PR #4758, opened 2024-03-28
[Refactor]: Table Grid Formula issue. 

Rajat-Dabade:
Fixes: https://github.com/SigNoz/engineering-pod/issues/1227

### DIFF
--- a/frontend/src/lib/query/createTableColumnsFromQuery.ts
+++ b/frontend/src/lib/query/createTableColumnsFromQuery.ts
@@ -7,7 +7,7 @@ import {
 import { FORMULA_REGEXP } from 'constants/regExp';
 import { QUERY_TABLE_CONFIG } from 'container/QueryTable/config';
 import { QueryTableProps } from 'container/QueryTable/QueryTable.intefaces';
-import { isObject } from 'lodash-es';
+import { isEqual, isObject } from 'lodash-es';
 import { ReactNode } from 'react';
 import {
 	IBuilderFormula,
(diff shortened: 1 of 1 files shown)


frontend/src/lib/query/createTableColumnsFromQuery.ts
```diff
@@ -306,8 +301,6 @@ const fillRestAggregationData = (
 		if (!isEqual) {
 			equalQueriesByLabels.push(column.field);
 		}
-
-		column.data.push(parseFloat(targetSeria.values[0].value).toFixed(2));
```
Rajat-Dabade, line 310: This code is executed only when there are multiple formulas. It checks if there are similar labels present in other formulas and, if found, adds them to the corresponding column data in the table.

However, since we're already looping through the series, which is responsible for adding data to its columns, adding it here may be redundant and may cause duplicated data to be added in the column for formulas (i.e. if the series contains 28 data point's then it will be 56).

The line 307 i.e. `equalQueriesByLabels.push(column.field);` is crucial. It ensures that no additional rows are added to the table for similar labels across all formulas here is how this check is applied: 
https://github.com/SigNoz/signoz/blob/e6e377beff550b5505a708cedd8688055f15b0cf/frontend/src/lib/query/createTableColumnsFromQuery.ts#L370

Vikrant2520: @Rajat-Dabade we can add this as inline comment? PR might get lost in future during some refactor but inline comment would help

Rajat-Dabade: Sure, I will add it.

Vikrant2520 (approved):

PR #4759, opened 2024-03-28
feat: support of changing panel type in dashboards

### LINKED ISSUE #2410: ability to change the panel type in the dashboards
The list of supported panel types are following:
- Value type
- Time series
- Table view
- List view

We need to support the changing panel type and that means we have 12 possible combinations of the current panel type and the target panel type. Following is the expected behaviour.

## Time series -> value type

The value type requires the query to result to be a signle number. There must be some aggregation function and group by cannot be there as it would result in multiple values. The time series panel would alredy have the aggregation function so the remaining part to be taken care is the group by part. Upon going to value type panel the `groupBy` option will disappear and the `reduceTo` will show with the default option.

#### Example

```
    AVG - (span.duration) - WHERE operation like '%GET%' - GROUP BY service_name

    will be changed to

    AVG - (span.duration) - WHERE operation like '%GET%' - REDUCE TO LAST
```

## Time series -> Table view

In our case, the table view is just an tabular representation of the time series. The table view is a list of rows, where each row is a timestamp and a value and set of attributes. The timestamp is the time at which the value was recorded. The value is the value of the aggregation result at that time. Nothing in the query changes.

## Time series -> List view

The list view is result of the query without any aggregation. And it doesn't make sense to have the `groupBy` and `having` clauses here so they will not be shown.

#### Example
```
    AVG - (span.duration) - WHERE operation like '%GET%' - GROUP BY service_name

    will be changed to

    WHERE operation like '%GET%'
```

## Value type -> Time series

The reduce to field is removed. The query is changed to have a group by in the query. What does the `groupBy` contain? If they directly landed on this panel type nothing empty by default. However, if the have come here from some panel type which has some `groupBy` selected then that state of `groupBy` needs to be preserved. 

#### Example 

```
    AVG - (span.duration) - WHERE operation like '%GET%' - REDUCE TO LAST

    will be changed to

    AVG - (span.duration) - WHERE operation like '%GET%' - GROUP BY (user defined)
```

#### Example of two changes (time series -> value type -> time series)

1. Time sereis -> value type
```
    AVG - (span.duration) - WHERE operation like '%GET%' - GROUP BY service_name

    will be changed to

    AVG - (span.duration) - WHERE operation like '%GET%' - REDUCE TO LAST
```
2. Value type -> Time series

```
    AVG - (span.duration) - WHERE operation like '%GET%' - REDUCE TO LAST

    will be changed to

    AVG - (span.duration) - WHERE operation like '%GET%' - GROUP BY service_name

```

In the above example notice the removed `groupBy` in the value type is restored when the type is changed to time sereis type. We need to maintain some global common state and show only the fields that are relevant to the current query.

## Value type -> Table view

Same as the Value type -> Time series except the data is rendered in a table

## Value type -> List view

The list view is result of the query without any aggregation. And it doesn't make sense to have the `groupBy` and `having` clauses here so they will not be shown.

#### Example
```
    AVG - (span.duration) - WHERE operation like '%GET%' - REDUCE TO LAST

    will be changed to

    WHERE operation like '%GET%'
```

The state presevation applies everywhere. For instance if we go from value type -> list view -> time series, the relevant state needs to be applied. 

#### Example of two changes

1. value type -> list view

```
    AVG - (span.duration) - WHERE operation like '%GET%' - REDUCE TO LAST

    will be changed to

    WHERE operation like '%GET%'
```

2. list view -> time series

```
    WHERE operation like '%GET%'

    will be changed to

    AVG - (span.duration) - WHERE operation like '%GET%'
```

## Table view -> Time series

No change in the query. The result of the query is rendered as a time series using graph

## Table view -> Value type

The group by is removed. The query is changed to have a reduce to field. The result of the query is rendered as a single value.

#### Example
```
    AVG - (span.duration) - WHERE operation like '%GET%' - GROUP BY service_name

    will be changed to

    AVG - (span.duration) - WHERE operation like '%GET%' - REDUCE TO LAST
```

#### Example of two changes (table view -> value type -> time series)

1. Table view -> value type
```
    AVG - (span.duration) - WHERE operation like '%GET%' - GROUP BY service_name

    will be changed to

    AVG - (span.duration) - WHERE operation like '%GET%' - REDUCE TO LAST
```

2. Value type -> time series

```
    AVG - (span.duration) - WHERE operation like '%GET%' - REDUCE TO LAST

    will be change to 

    AVG - (span.duration) - WHERE operation like '%GET%' - GROUP BY service_name

```

## Table view -> List view

The list view is a result of a query with no aggregation. When there is no aggregation it doesn't make sense to have `groupBy` and `having` so they are removed. 

#### Example
```
    COUNT - (span.error) - WHERE operation like '%GET%' - GROUP BY service_name

    will be changed to

    WHERE operation like '%GET%'
```

## List view -> Time series

Since the list view is a result of query without aggregation on any attribute, the default aggregation that works here is `COUNT`. The default aggregation `COUNT` is added. The query is changed to have a group by in the query. The result of the query is rendered as a time series using graph.

#### Example

```
    WHERE operation like '%GET%'

    will be changed to

    COUNT - (span.duration)(optional) - WHERE operation like '%GET%' - GROUP BY (user defined)
```

## List view -> Value type

The default aggregation COUNT is added. The query is changed to have a reduce to field. The result of the query is rendered as a single value.

#### Example

```
    WHERE operation like '%GET%'

    will be changed to

    COUNT - (span.duration)(optional) - WHERE operation like '%GET%' - REDUCE TO LAST
```

## List view -> Table view

The default aggregation COUNT is added. The query is changed to have a group by in the query. The result of the query is rendered as a table.

#### Example 

```
    WHERE operation like '%GET%'

    will be changed to

    COUNT - (span.duration)(optional) - WHERE operation like '%GET%' - GROUP BY (user defined)
```

One of main points to note here is persistence of state going from one panel ot other via multiple steps, in that case the relevant state from the previous steps needs to be persisted and show in the UI.

Vikrant2520:
### Summary

- Add support for changing panel type in dashboard widgets
- remove the list view type from the panel list when the query contains `metrics` datasource

#### Related Issues / PR's

Fixes #2410

#### Screenshots

NA


srikanthccv (approved):
Tested
- v3 and v4 compatibility
- changing panels for metrics, logs and traces

YounixM (approved):

PR #4762, opened 2024-03-28
(feature) multiple ingestion keys

grandwizard28:
### Summary

- Add support for multiple ingestion keys

#### Screenshots

NA

- Add `unit` tests for gateway interface


deploy/docker/clickhouse-setup/docker-compose.yaml
```diff
@@ -169,6 +169,7 @@ services:
     command:
       [
         "-config=/root/config/prometheus.yml",
+        "-gateway-url=https://api.staging.signoz.cloud"
```
srikanthccv, line 172: Remove it from the general docker-compose (our OSS users use the docker deployment) and pass it wherever you need to enable the gateway setup.

ee/query-service/license/db.go
```diff
@@ -49,15 +50,18 @@ func (r *Repo) GetLicenses(ctx context.Context) ([]model.License, error) {
...
 	licenses := []model.License{}
 
 	query := "SELECT key, activationId, planDetails, validationMessage FROM licenses"
 
 	err = r.db.Select(&licenses, query)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get active licenses from db: %v", err)
+		if err == sql.ErrNoRows {
+			return nil, basemodel.NotFoundError(errors.New("no licenses found"))
+		}
```
srikanthccv, line 63: The `Select` never returns the `sql.ErrNoRows` so this change doesn't add anything. Some call paths expect a nil error when there is no license instead of a NotFound error otherwise, the server won't start https://github.com/SigNoz/signoz/blob/93df4759699025aaf328f18e00e8e0d758a06553/ee/query-service/main.go#L154. Either we should handle the no license error wherever GetActiveLicense is involved or return nil error when there is no active licence.

grandwizard28: - Why won't Select return a NoRows error? There can be cases where no license is there right?
 - Not sure whether the above is happening or not. (testing had no license, server started). If an active license is not found, the default is set to BASIC as per my understanding. 
 
 Ref line 123 in `[ee/query-service/license/manager.go](https://github.com/SigNoz/signoz/pull/4762/files#diff-571d493907b4315fc1ac269efcf7668db417a5a591c6cd62e582046ed24b1043)`:
 ```
if active != nil {
		lm.SetActive(active)
	} else {
		zap.L().Info("No active license found, defaulting to basic plan")
		// if no active license is found, we default to basic(free) plan with all default features
		lm.activeFeatures = model.BasicPlan
		setDefaultFeatures(lm)
```

Comment (srikanthccv) [reply ee/query-service/license/db.go:63]: >Why won't Select return a NoRows error? There can be cases where no license is there, right?

The [`Get`](https://pkg.go.dev/github.com/jmoiron/sqlx@v1.3.4#DB.Get) returns the `NoRows` error but [`Select`](https://pkg.go.dev/github.com/jmoiron/sqlx@v1.3.4#Select) doesn't.

> If an active license is not found, the default is set to BASIC as per my understanding.

If you see the same function, it returns the error on line 121 if it's non-nil. The default to the basic plan happens only when the err is nil and the active plan is nil. There is no issue right now because `Select` doesn't return NoError but I think we should handle it?

**Comment by grandwizard28** [comment]
Implemented a very rudimentary feature flag here! @srikanthccv

File: ee/query-service/app/api/gateway.go [hunk 3]
```diff
@@ -0,0 +1,34 @@
...
+	var licenseKey string
+	if license != nil {
+		licenseKey = license.Key
+	}
+
+	req.Header.Set("X-Signoz-Cloud-Api-Key", licenseKey)
+	req.Header.Set("X-Consumer-Username", "lid:00000000-0000-0000-0000-000000000000")
+	req.Header.Set("X-Consumer-Groups", "ns:default")
+
+	ah.Gateway().ServeHTTP(rw, req)
```
srikanthccv, line 33: Does this panic with nil pointer dereference if the gateway URL is not provided? I see the feature will be inactive but seems like it would crash if the gateway URL is invoked.

srikanthccv (commented):
Overall LGTM. Just one comment.

ee/query-service/app/api/gateway.go
grandwizard28: Yeah it will panic. Not ideal I know. Do you want me to workaround it?

I'll have to create an interface in noop to override the serveHTTP behaviour.

srikanthccv: I do not expect users/frontend to hit this URL when the gateway is not enabled but it's a good idea to not panic because of nil pointer dereference. You can address it later if you perfer.

srikanthccv (approved):

ee/query-service/app/api/gateway.go
grandwizard28: Got it!

prashant-shahi (approved):

grandwizard28:
Screenshot 2024-06-04 at 5 38 08 PM
Corresponding Frontend PR: https://github.com/SigNoz/signoz/pull/5105

PR #4787, opened 2024-04-01
fix: use proper indexes for full text search

### LINKED ISSUE #4259: [LOGS] Index on body is not working.
Index on body is not working.

nityanandagohain:
Fixes #4259 

changes
* body searches use lower(body) now.
* ~~attributes search with operator  `contains` will also use `lower(attribute)` now but there is no change for them.~~
* json search add's new filters i.e where clauses.


pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -192,10 +192,17 @@ func buildLogsTimeSeriesFilterQuery(fs *v3.FilterSet, groupBy []v3.AttributeKey,
...
 					columnName := getClickhouseColumnName(item.Key)
-					conditions = append(conditions, fmt.Sprintf("%s %s '%%%s%%'", columnName, logsOp, item.Value))
+					conditions = append(conditions, fmt.Sprintf("lower(%s) %s lower('%%%s%%')", columnName, logsOp, item.Value))
 				default:
 					columnName := getClickhouseColumnName(item.Key)
 					fmtVal := utils.ClickHouseFormattedValue(value)
+
+					// for use lower for like and ilike
+					if op == v3.FilterOperatorLike || op == v3.FilterOperatorNotLike {
+						columnName = fmt.Sprintf("lower(%s)", columnName)
```
srikanthccv, line 202: Can you share some examples with and without `lower`? The way I imagined this would work is that we need to make the index lower and convert the search value to lower but do not use `lower` function. Just to double check we will not run into this kind of problem https://stackoverflow.com/questions/62629579/mysql-why-using-where-on-date-column-does-not-use-index

nityanandagohain: This is tested , using `lower(body)` in the where clause uses the corresponding `lower(body) index`
https://github.com/SigNoz/signoz/issues/4259#issuecomment-1970424329

srikanthccv: Can I play with this somewhere?

srikanthccv (approved):

pkg/query-service/app/logs/v3/query_builder.go
nityanandagohain: please check slack

pkg/query-service/app/logs/v3/json_filter.go
```diff
@@ -106,6 +107,29 @@ func getJSONFilterKey(key v3.AttributeKey, op v3.FilterOperator, isArray bool) (
 	return keyname, nil
 }
 
+// takes the path and the values and generates where clauses for better usage of index
+func getPathIndexFilter(path string) string {
+	filters := []string{}
+	keyArr := strings.Split(path, ".")
+	if len(keyArr) < 2 {
+		return ""
+	}
```
srikanthccv, line 116: This `< 2` always becomes a puzzle. It seems like this is for JSON querying. Can we somehow make use of the `IsJSON` info instead of this split and length comparison?

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -193,10 +193,17 @@ func buildLogsTimeSeriesFilterQuery(fs *v3.FilterSet, groupBy []v3.AttributeKey,
 				case v3.FilterOperatorContains, v3.FilterOperatorNotContains:
 					columnName := getClickhouseColumnName(item.Key)
 					val := utils.QuoteEscapedString(fmt.Sprintf("%v", item.Value))
-					conditions = append(conditions, fmt.Sprintf("%s %s '%%%s%%'", columnName, logsOp, val))
+					conditions = append(conditions, fmt.Sprintf("lower(%s) %s lower('%%%s%%')", columnName, logsOp, val))
```
srikanthccv, line 196: Does the same `lower` mentioned for body work for regular attributes also?

nityanandagohain: There are two points to this
* Previously we were utilising `ilike` for attributes as well. The index was not at all getting used.
* Now it will not work as well since we are using lower(attribute_name) but we will have to address this by making our string indexes to lower.


but here I am finding an issue that if we make it lowercase than the '=' ,  'in' should also be chaned to lowercase which might cause issue with 'Appname' and 'appname'

srikanthccv: Do you want to address this independently?

nityanandagohain: Yeah can do that, will remove the changes for attributes then for now.

pkg/query-service/app/logs/v3/json_filter.go
nityanandagohain: So we already have that check, this is just an extra independent check for this function.

https://github.com/SigNoz/signoz/blob/e639d79ba9864547217c45531d9f08569d461f7f/pkg/query-service/app/logs/v3/query_builder.go#L166

pkg/query-service/app/logs/v3/query_builder.go
nityanandagohain: have made the changes, now it will only apply for body

pkg/query-service/app/logs/v3/json_filter.go
srikanthccv: The problem with this kind of code is that you understand things in your head but it's hard to reason about for everyone else. What do you think?

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -193,10 +195,24 @@ func buildLogsTimeSeriesFilterQuery(fs *v3.FilterSet, groupBy []v3.AttributeKey,
...
 					columnName := getClickhouseColumnName(item.Key)
 					fmtVal := utils.ClickHouseFormattedValue(value)
+
+					// for use lower for like and ilike
+					if op == v3.FilterOperatorLike || op == v3.FilterOperatorNotLike {
+						if columnName == BODY {
+							columnName = fmt.Sprintf("lower(%s)", columnName)
+							fmtVal = fmt.Sprintf("lower(%s)", fmtVal)
+						}
+						fmtVal = strings.ToLower(fmtVal)
```
srikanthccv, line 213: What is this line doing?

pkg/query-service/app/logs/v3/json_filter.go
nityanandagohain: yeah aggree, I think I should just remove this then as it is just creating confusion and the check is already done before

nityanandagohain: I don't think I should remove this, instead I added a comment which explain this to future reviewers, hope it helps.

pkg/query-service/app/logs/v3/query_builder.go
nityanandagohain: https://github.com/SigNoz/signoz/blob/dfc7f827d8d3917d0d9b62a6809f3c4e58018ac2/pkg/query-service/app/logs/v3/json_filter.go#L76

pkg/query-service/app/logs/v3/query_builder.go
nityanandagohain: ahh it was a bug, have added a test case now to cover this. Thanks for catching.

nityanandagohain:
One other thing that this PR changes is that it uses like in favour of ilike, which means that attribute queries for contains and like won't be case independent now. I am not sure if we should remove it for this PR. cc @srikanthccv

srikanthccv:
Yes, that should have a separate issue for itself.

nityanandagohain:
Have made the changes, now it will only apply for body .

Attributes and resource attributes won't be touched.

pkg/query-service/app/logs/v3/json_filter.go
```diff
@@ -154,11 +179,28 @@ func GetJSONFilter(item v3.FilterItem) (string, error) {
...
+	pathFilter := getPathIndexFilter(item.Key.Key)
+	if pathFilter != "" {
+		filters = append(filters, pathFilter)
+	}
+	if op == v3.FilterOperatorContains ||
+		op == v3.FilterOperatorEqual ||
+		op == v3.FilterOperatorHas {
+		val, ok := item.Value.(string)
+		if ok && len(val) >= NGRAM_SIZE {
+			filters = append(filters, fmt.Sprintf("lower(body) like lower('%%%s%%')", strings.ToLower(val)))
```
srikanthccv, line 193: The value should be escaped otherwise it will produce an invalid query.

pkg/query-service/app/logs/v3/json_filter_test.go
```diff
@@ -311,7 +311,7 @@ var testGetJSONFilterData = []struct {
 			Operator: "contains",
 			Value:    "hello 'world'",
 		},
-		Filter: "JSON_EXISTS(body, '$.\"message\"') AND JSON_VALUE(body, '$.\"message\"') ILIKE '%hello \\'world\\'%'",
+		Filter: "lower(body) like lower('%message%') AND lower(body) like lower('%hello 'world'%') AND JSON_EXISTS(body, '$.\"message\"') AND JSON_VALUE(body, '$.\"message\"') ILIKE '%hello \\'world\\'%'",
```
srikanthccv, line 314: the example of invalid query `lower(body) like lower('%hello 'world'%')`

pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -193,10 +195,24 @@ func buildLogsTimeSeriesFilterQuery(fs *v3.FilterSet, groupBy []v3.AttributeKey,
 				case v3.FilterOperatorContains, v3.FilterOperatorNotContains:
 					columnName := getClickhouseColumnName(item.Key)
 					val := utils.QuoteEscapedString(fmt.Sprintf("%v", item.Value))
-					conditions = append(conditions, fmt.Sprintf("%s %s '%%%s%%'", columnName, logsOp, val))
+					if columnName == BODY {
+						logsOp = strings.Replace(logsOp, "ILIKE", "LIKE", 1) // removing i from ilike and not ilike
+						conditions = append(conditions, fmt.Sprintf("lower(%s) %s lower('%%%s%%')", columnName, logsOp, val))
```
srikanthccv, line 200: I would use some function on the `logsOp` itself like `logsOp.<func>()`  rather than the `strings.Replace`.

pkg/query-service/app/logs/v3/query_builder_test.go
```diff
@@ -287,6 +295,22 @@ var timeSeriesFilterQueryData = []struct {
...
+	{
+		Name: "Test for body contains and ncontains",
+		FilterSet: &v3.FilterSet{Operator: "AND", Items: []v3.FilterItem{
+			{Key: v3.AttributeKey{Key: "body", DataType: v3.AttributeKeyDataTypeString, IsColumn: true}, Operator: "contains", Value: "test"},
+			{Key: v3.AttributeKey{Key: "body", DataType: v3.AttributeKeyDataTypeString, IsColumn: true}, Operator: "ncontains", Value: "test1"},
+		}},
+		ExpectedFilter: "lower(body) LIKE lower('%test%') AND lower(body) NOT LIKE lower('%test1%')",
+	},
+	{
+		Name: "Test for body contains and ncontains",
```
srikanthccv, line 307:
```suggestion
		Name: "Test for body like and ilike",
```

pkg/query-service/app/logs/v3/query_builder.go
nityanandagohain: but the implementation will remain the same right ? or you mean somethig else ?

srikanthccv: We support case-insensitive operations but this change doesn't account for that. If you want to account for that you can either add a statement here and below or do all of the logic separately that explicitly highlights why we do it and use that function here.

srikanthccv (approved):

pkg/query-service/app/logs/v3/json_filter.go
```diff
@@ -72,6 +73,7 @@ func getPath(keyArr []string) string {
 
 func getJSONFilterKey(key v3.AttributeKey, op v3.FilterOperator, isArray bool) (string, error) {
 	keyArr := strings.Split(key.Key, ".")
+	// i.e it should be at least body.name, and not something like body
 	if len(keyArr) < 2 {
```
vikrantgupta25, line 77: does this account for `body. ` where there is just space after the `.` ?

nityanandagohain: keys dont contain space, it will be a invalid query

vikrantgupta25: it will be returned before reaching this point ?

PR #4808, opened 2024-04-03
chore: update events

makeavish:
### Summary

Update Query Range API event and send heart beat and dashboard events for all registered users

### DIFF
--- a/ee/query-service/app/server.go
+++ b/ee/query-service/app/server.go
@@ -401,7 +401,7 @@ func extractQueryRangeData(path string, r *http.Request) (map[string]interface{}
 	data := map[string]interface{}{}
 	var postData *v3.QueryRangeParamsV3
 
-	if (path == pathToExtractBodyFromV3 && (r.Method == "POST")) || (path == pathToExtractBodyFromV4 && (r.Method == "POST")) {
+	if (r.Method == "POST") && ((path == pathToExtractBodyFromV3) || (path == pathToExtractBodyFromV4)) {
 		if r.Body != nil {
 			bodyBytes, err := io.ReadAll(r.Body)
 			if err != nil {
--- a/ee/query-service/app/server.go
+++ b/ee/query-service/app/server.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	_ "net/http/pprof" // http profiler
 	"os"
+	"regexp"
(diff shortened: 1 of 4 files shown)


ee/query-service/app/server.go
```diff
@@ -393,13 +394,14 @@ func (lrw *loggingResponseWriter) Flush() {
...
-	pathToExtractBodyFrom := "/api/v3/query_range"
+func extractQueryRangeData(path string, r *http.Request) (map[string]interface{}, bool) {
+	pathToExtractBodyFromV3 := "/api/v3/query_range"
+	pathToExtractBodyFromV4 := "/api/v4/query_range"
 
 	data := map[string]interface{}{}
 	var postData *v3.QueryRangeParamsV3
 
-	if path == pathToExtractBodyFrom && (r.Method == "POST") {
+	if (path == pathToExtractBodyFromV3 && (r.Method == "POST")) || (path == pathToExtractBodyFromV4 && (r.Method == "POST")) {
```
prashant-shahi, line 404:
```suggestion
	if (r.Method == "POST") && ((path == pathToExtractBodyFromV3) || (path == pathToExtractBodyFromV4)) {
```

ee/query-service/app/server.go
```diff
@@ -445,6 +466,20 @@ func extractQueryRangeV3Data(path string, r *http.Request) (map[string]interface
...
+				data["screen"] = "alert"
+			case logsExplorerMatched:
+				data["screen"] = "logs-explorer"
+			case traceExplorerMatched:
+				data["screen"] = "traces-explorer"
+			default:
+				data["screen"] = "unknown"
+				return data, true
+			}
 			telemetry.GetInstance().SendEvent(telemetry.TELEMETRY_EVENT_QUERY_RANGE_API, data, userEmail, true, false)
```
prashant-shahi, line 483: In case of the `default` fallback, event won't be sent.
Is that expected behaviour?

makeavish: Yes we don't want to send events when users open dashboards and all the other scenarios.
This will reduce events.

prashant-shahi: alright, LGTM 👍

prashant-shahi (approved):

PR #4825, opened 2024-04-06
feat: clickhouse optimize S3

makeavish:
### Summary

Clickhouse S3 optimizer runs OPTIMIZE TABLE query to reduce excessive PUT calls on S3

---------

Co-authored-by: Prashant Shahi <prashant@signoz.io>


pkg/query-service/app/clickhouseOptimizeS3/optimize.go
```diff
@@ -0,0 +1,143 @@
...
+		zap.L().Error("failed to check if S3 is enabled", zap.Error(err))
+		return err
+	} else if !s3Enabled {
+		zap.L().Info("S3 is not enabled, skipping clickhouse s3 optimization")
+		return nil
+	}
+
+	err = runCronJobAsync(chConn)
+	if err != nil {
+		zap.L().Error(fmt.Sprintf(ERROR_RUNNING_CRON_JOB), zap.Error(err))
```
srikanthccv, line 47: nit: there is no format specifier here for Sprintf

srikanthccv (commented):
My main concern with this is it's going to put system under so much pressure whenever it is triggered https://clickhouse.com/docs/en/optimize/avoidoptimizefinal. This can make the customer env the pod running effect everyone adversely. How big of the data we tested this on? Could this process be controlled with an upper limit on resources?

makeavish:
> My main concern with this is it's going to put system under so much pressure whenever it is triggered https://clickhouse.com/docs/en/optimize/avoidoptimizefinal. This can make the customer env the pod running effect everyone adversely. How big of the data we tested this on? Could this process be controlled with an upper limit on resources?

We use `optimize_skip_merged_partitions=1` which ensures that already merged partitions are not merged. Using upper limit on resources can be done but depends on load. Limiting resources too much can slow down merges eventually resulting in slowdown of queries. I would suggest we release this and then manually adjust resource limits based on load. After some feedbacks we can also automate resource limits.

pkg/query-service/app/clickhouseOptimizeS3/optimize.go
```diff
@@ -0,0 +1,143 @@
...
+		"signoz_metrics.time_series_v2",
+		"signoz_traces.usage_explorer",
+		"signoz_traces.span_attributes",
+		"signoz_traces.dependency_graph_minutes",
+		"signoz_traces.dependency_graph_minutes_v2",
+		"signoz_traces.signoz_error_index_v2",
+		"signoz_traces.signoz_index_v2",
+		"signoz_traces.signoz_spans",
+		"signoz_traces.durationSort",
+	}
```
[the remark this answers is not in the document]
makeavish: I see we are not allowing user to change ttl or move to s3 for `signoz_metrics.time_series_v4` or other time_series tables. Any reason for that?

srikanthccv:
IMO this is not one of those features where we get the initial working piece merged and think about it more later. It has the potential to bring down the entire system from ingestion, alerting and product usage. We know the mutations are the worst thing in ClickHouse. Even on the table with not more than a few tens of million rows, the CPU usage went near 100% which delayed ingestion leading to false alerts and the product was unusable during the whole time https://signoz-team.slack.com/archives/C06C5U3TUDP/p1706604434441239. Once it got into a bad state there was no way to stop it immediately (The kill wouldn't work). We had to wait till it was completed on its own. 

Did we test this on the system which had a non-trivial amount of data? What was system the of resource usage change observed? How did it affect the rest of the product usage? My main point about the resource limits is that there should be sensible limits when we initiate an unscheduled merge. I didn't say we should limit resources too much.

pkg/query-service/app/clickhouseOptimizeS3/optimize.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/ClickHouse/clickhouse-go/v2"
+	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
+	"github.com/go-co-op/gocron"
+	"go.uber.org/zap"
+)
+
+// General
+const (
+	CH_OPTIMIZE_INTERVAL_IN_HOURS = 24
+	CH_TIMEOUT_WAIT_IN_MINUTES    = 30
```
ankitnayan, line 19: this is not respected in case of slowdown

PR #4846, opened 2024-04-10
chore: use last 1day data for apdex latency metric meta

srikanthccv:
#### Screenshots

NA


srikanthccv:
This PR 

- Updates the frontend to use the new endpoint to fetch the metric meta
- Sort the buckets before returning the value
- Use the correct formula expression

query-service changes that are tested
- Delete the old code
- Update the query to only read data from last one day and not the whole table.

YounixM (approved):

PR #4851, opened 2024-04-10
fix: functions rendering wrong data for table dashboard type

Vikrant2520:
### Summary

Detailed Explanation Here - https://www.notion.so/signoz/Table-Graph-Type-Calculation-Analysis-aa72d70f5e544c1c9003c95193d78774

#### Screenshots

NA

### DIFF
--- a/frontend/src/lib/query/createTableColumnsFromQuery.ts
+++ b/frontend/src/lib/query/createTableColumnsFromQuery.ts
@@ -287,6 +287,7 @@ const fillRestAggregationData = (
 	queryTableData: QueryDataV3[],
 	seria: SeriesItem,
 	equalQueriesByLabels: string[],
+	unusedColumnsKeys: Set<keyof RowData>,
 ): void => {
 	const nextQueryData =
 		queryTableData.find((q) => q.queryName === column.field) || null;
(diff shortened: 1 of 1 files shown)


Rajat-Dabade (dismissed):

srikanthccv:
Is this a regression? Wasn't this already fixed?

Vikrant2520:
No @srikanthccv , not sure about before , was working on the performance of the same when came across this issue

srikanthccv:
This PR https://github.com/SigNoz/signoz/pull/4758 removed the line as a part of the bug fix and you are adding it back.

srikanthccv:
I don't know which is the correct fix but it certainly shows why there is a strong need for minimal tests.

Vikrant2520:
This is required as when in case of functions we try to fill the function columns if the column match happens as you can see we do not push to column data but in case it doesn't match we push `N/A` which is correct but we do need to push to the rows in case of value exist. I am writing a detailed document around the logic going on for table processing. will attach it here for more clarity.

Vikrant2520:
@srikanthccv please find the detailed calculation document attached in the description

srikanthccv:
I also see code changes in the last commit. For someone unfamiliar with the codebase (I also can't provide an objective review for the same reason), this gives less confidence in the change.

Vikrant2520:
Totally agreed to that, that's why wanted to streamline this calculation and explained what happens where in the document so that for someone next visiting this code has fair good understanding of what happens and why it happens

srikanthccv:
OK, please get reviews from whoever knows this part of the code. While your intent is noble, my experience is that people do not usually read separate docs. The correctness should be maintained by various means such as strong reviews and existing tests. At this point, I am just repeating the same things but given we time we spend on fixing the bugs could be avoided with a few things in place and I hope to see us reach there soon. Because these kinds of bugs are not visual and only customers tell us the issue. If you want to test it against https://github.com/SigNoz/engineering-pod/issues/1227, please ping me on Slack for creds and dashboard.

Rajat-Dabade (approved):

PR #4865, opened 2024-04-15
 feat: allow using the consistent naming with the signoz collector

git-hulk:
### Summary

Allow using the consistent ClickHouse cluster name in the collector and query service.

#### Related Issues / PR's

This resolves https://github.com/SigNoz/signoz-otel-collector/issues/308


git-hulk:
@srikanthccv I didn't remove the old flag `--cluster` to avoid breaking the old users, maybe we can remove this in the future after changing all dependencies like the helm charts, and so on.

frontend/src/container/OnboardingContainer/Modules/AwsMonitoring/ECSExternal/md-docs/ecsExternal-createDaemonService.md
```diff
@@ -41,8 +41,7 @@ aws cloudformation create-stack --stack-name AOCECS-daemon-${CLUSTER_NAME}-${REG
 To verify that the daemon service is running, you can run the following command, which should output the task ARN of the Daemon service as the output.
 
 ```bash
-aws ecs list-tasks --cluster ${CLUSTER_NAME} --region ${REGION}
-
+aws ecs list-tasks --cluster-name ${CLUSTER_NAME} --region ${REGION}
```
srikanthccv, line 44: This is unrelated and should be reverted.

pkg/query-service/main.go
```diff
@@ -52,6 +52,8 @@ func main() {
 	flag.StringVar(&cacheConfigPath, "experimental.cache-config", "", "(cache config to use)")
 	flag.StringVar(&fluxInterval, "flux-interval", "5m", "(cache config to use)")
 	flag.StringVar(&cluster, "cluster", "cluster", "(cluster name - defaults to 'cluster')")
+	// Allow using the consistent naming with the signoz collector
+	flag.StringVar(&cluster, "cluster-name", "cluster", "(cluster name - defaults to 'cluster')")
```
srikanthccv, line 56: Wouldn't this override the `--cluster` always i.e even if someone provided `--cluster=my-cluster` the value would be default `cluster` from `--cluster-name` arg?

git-hulk: @srikanthccv No, it won't be overridden unless the `--cluster-name` is provided since the flag parse was only invoked while the flag name was passed. That said, if users specify:

1. `--cluster ABC`, then the cluster will be ABC
2. `--cluster-name DEF`, then the cluster will be DEF
3. `--cluster ABC --cluster-name DEF`, then the cluster will be DEF
4. `--cluster-name DEF --cluster ABC`, then the cluster will be ABC

git-hulk: @srikanthccv Any comment on this?

srikanthccv: I just verified. This looks good to me.

srikanthccv (approved):

pkg/query-service/main.go
git-hulk: @srikanthccv Thank you!

PR #4876, opened 2024-04-17
fix: alert threshold form is resetting to default query option on stage & run

SagarRajput-7:
#### Screenshots

NA

### DIFF
--- a/frontend/src/container/FormAlertRules/index.tsx
+++ b/frontend/src/container/FormAlertRules/index.tsx
@@ -138,15 +138,17 @@ function FormAlertRules({
 
 	useEffect(() => {
 		// Set selectedQueryName based on the length of queryOptions
-		setAlertDef((def) => ({
-			...def,
-			condition: {
-				...def.condition,
-				selectedQueryName:
-					queryOptions.length > 0 ? String(queryOptions[0].value) : undefined,
-			},
-		}));
-	}, [currentQuery?.queryType, queryOptions]);
+		if (!alertDef?.condition?.selectedQueryName) {
+			setAlertDef((def) => ({
+				...def,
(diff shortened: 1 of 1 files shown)


frontend/src/container/FormAlertRules/index.tsx
```diff
@@ -138,15 +138,17 @@ function FormAlertRules({
...
-				...def.condition,
-				selectedQueryName:
-					queryOptions.length > 0 ? String(queryOptions[0].value) : undefined,
-			},
-		}));
-	}, [currentQuery?.queryType, queryOptions]);
+		if (!alertDef?.condition?.selectedQueryName) {
+			setAlertDef((def) => ({
+				...def,
+				condition: {
```
Vikrant2520, line 144: let's say we have 2 queries initially `A` and `B` and the value is set to `B`, if query B is deleted then it resets to A? Please verify the above case once for before and after

SagarRajput-7: **Added safety check **

- Before - as earlier every time the value is getting set to the first index queryOption i.e. A hence it doesn't matter there if any queryOption is deleted or not as the value was not retaining

https://github.com/SigNoz/signoz/assets/162284829/4a0c45af-5350-41a9-98f7-dd74e4aadc4f

- After - added safety check to check if the queryOption selected is present in the updated list or not if not then it will reset to the 0th index of queryOptions


https://github.com/SigNoz/signoz/assets/162284829/804558a7-0e6c-4d99-b491-451d499f72ae


Thanks!

Vikrant2520 (approved):
lgtm

YounixM (approved):

PR #4888, opened 2024-04-19
feat: [SIG-584]: moved facing issue btn tracking from trackEvent to logEvent

SagarRajput-7:
### Summary

Moved facing issue btn tracking from trackEvent to logEvent

#### Screenshots

https://github.com/SigNoz/signoz/assets/162284829/9d56b41d-0a00-47b2-9434-5038b9f4f28c


makeavish (dismissed):

frontend/src/hooks/analytics/useAnalytics.tsx
```diff
@@ -1,3 +1,4 @@
+import logEvent from 'api/common/logEvent';
```
Vikrant2520, line 1: why are we importing `logEvent` in the file and exporting it from the hook ? can we directly use it ?

frontend/src/container/OnboardingContainer/common/ModuleStepsContainer/ModuleStepsContainer.tsx
```diff
@@ -381,11 +381,12 @@ export default function ModuleStepsContainer({
 	};
 
 	const handleFacingIssuesClick = (): void => {
-		trackEvent('Onboarding V2: Facing Issues Sending Data to SigNoz', {
+		logEvent('Onboarding V2: Facing Issues Sending Data to SigNoz', {
```
Vikrant2520, line 384: NIT: we do not need any error / success handling here ? as in we are not concerned with API returns whatever response here ?

SagarRajput-7: yes, here we are not concerned about the response and also dont want to block the process due to this call.

frontend/src/hooks/analytics/useAnalytics.tsx
SagarRajput-7: Majorly done just to categorize this logEvent under useAnalytics, as its a similar option as trackEvent

frontend/src/container/OnboardingContainer/common/ModuleStepsContainer/ModuleStepsContainer.tsx
Vikrant2520: so if somebody wants to use the `logEvent` they will have to process the entire hook code. We should keep it separate.

frontend/src/hooks/analytics/useAnalytics.tsx
Vikrant2520: so if somebody wants to use the logEvent they will have to process the entire hook code. We should keep it separate.

SagarRajput-7: Sure, will use logEvent directly here

Vikrant2520 (approved):

YounixM (approved):

PR #4898, opened 2024-04-22
feat: show error message received from the server in logs explorer views

YounixM:
Screenshot 2024-04-23 at 12 54 54 AM

### DIFF
--- a/frontend/src/container/LogsError/LogsError.tsx
+++ b/frontend/src/container/LogsError/LogsError.tsx
@@ -7,7 +7,15 @@ import history from 'lib/history';
 import { ArrowRight } from 'lucide-react';
 import { isCloudUser } from 'utils/app';
 
-export default function LogsError(): JSX.Element {
+export interface QueryRangeErrorCauseType {
+	error: string;
+}
+
+export default function LogsError({
+	errorData,
+}: {
+	errorData?: Error | null;
+}): JSX.Element {
 	const handleContactSupport = (): void => {
 		if (isCloudUser()) {
 			history.push('/support');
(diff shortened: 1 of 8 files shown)


nityanandagohain (approved):

frontend/src/container/LogsError/LogsError.tsx
```diff
@@ -7,14 +7,29 @@ import history from 'lib/history';
...
 import { isCloudUser } from 'utils/app';
 
-export default function LogsError(): JSX.Element {
+export interface QueryRangeErrorCauseType {
+	error: string;
+}
+
+export default function LogsError({
+	errorData,
+}: {
```
palashgdev, line 16: ideally we should avoid using inline defining of interface

frontend/src/container/LogsError/LogsError.tsx
```diff
@@ -7,14 +7,29 @@ import history from 'lib/history';
...
 			history.push('/support');
 		} else {
 			window.open('https://signoz.io/slack', '_blank');
 		}
 	};
+
+	let serverErrorMessage = null;
+
+	if (errorData && errorData.cause) {
+		serverErrorMessage = (errorData.cause as QueryRangeErrorCauseType).error;
```
palashgdev, line 30: there should not be any type of assertion bad practice 

ideally, this should be handled from central place

YounixM: Error type has a standard type in case of New Error() and the cause property type is set to unknown and we are setting cause to be an object with the error property. Hence this.

Vikrant2520 (approved):

SagarRajput-7 (approved):

makeavish:
Why is this PR pending?

PR #4903, opened 2024-04-23
fix: handle-large-traces

makeavish:
### Summary

1. Fix `span not found` error when opening large traces from Traces tab
2. Fix query service crashing with OOM errors due to loading huge traces in memory by introducing max span in trace limit with variable: `MAX_SPANS_IN_TRACE`, set to 250000 by default
3. Reduce `SPAN_RENDER_LIMIT` to 2500 by default. Allow changing `SPAN_RENDER_LIMIT` with `spanRenderLimit` param in trace detail URL - Will need frontend change, should not be required for now as `SPAN_RENDER_LIMIT` is just an env variable. Will update it later

**Screenshots**

Screenshot 2024-05-24 at 17 56 40

Screenshot 2024-05-24 at 17 56 14

Screenshot 2024-05-24 at 21 29 24


srikanthccv (approved):

frontend/src/container/TraceDetail/index.tsx
```diff
@@ -142,9 +143,10 @@ function TraceDetail({ response }: TraceDetailProps): JSX.Element {
 							Trace Details
 						</StyledTypography.Title>
 						<StyledTypography.Text styledclass={[styles.removeMargin]}>
-							{traceMetaData.totalSpans} Span
+							{traceMetaData.totalSpans} Spans
 						</StyledTypography.Text>
 						{hasMissingSpans && <MissingSpansMessage />}
+						{response[0].isSubTree && <SubTreeMessage />}
```
YounixM, line 149: add safety check here

makeavish: Done

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -2014,15 +2014,47 @@ func (r *ClickHouseReader) GetUsage(ctx context.Context, queryParams *model.GetU
...
+	if countSpans > uint64(params.MaxSpansInTrace) {
+		zap.L().Error("Max spans allowed in a trace limit reached", zap.Int("MaxSpansInTrace", params.MaxSpansInTrace),
+			zap.Uint64("Count", countSpans))
+		userEmail, err := auth.GetEmailFromJwt(ctx)
+		if err == nil {
+			data := map[string]interface{}{
+				"traceSize":            countSpans,
+				"maxSpansInTraceLimit": params.MaxSpansInTrace,
+			}
+			telemetry.GetInstance().SendEvent(telemetry.TELEMETRY_EVENT_MAX_SPANS_ALLOWED_LIMIT_REACHED, data, userEmail, true, false)
```
srikanthccv, line 2038: nit: Is it possible to remove this event and just send the outer event `TELEMETRY_EVENT_TRACE_DETAIL_API ` with `maxSpansInTraceLimit` and filter at the analytics tool?

srikanthccv (approved):

YounixM (approved):

pkg/query-service/app/clickhouseReader/reader.go
makeavish: Yeah I thought so but it's more easier to run automation and big query over separate event as many tools we use don't provide filtering based triggers (customer.io)

PR #4916, opened 2024-04-25
fix: rate in table panel

makeavish:
### Summary

Step or aggregate interval is whole time period in case of table panel so updated that


nityanandagohain (approved):
These is done for the rate function ?

If yes please add some test cases as well.

nityanandagohain (changes requested):
These is done for the rate function ?

If yes please add some test cases as well.

nityanandagohain (approved):

PR #4934, opened 2024-04-29
fix: value panel restriction should be on enabled queries

### LINKED ISSUE #4855: Error in query in Value Panel
## Bug description
Creating a value panel from a function that combines two queries fails with an error of "there can be only one active query for value type panel".
*Please describe.*  
image

## Expected behavior
Should display the latest value from the function.

## How to reproduce

1. Create Value Panel
2. Add query A, and hide
3. Add query B, and hide
4. Add a function
5. Run query

## Version information 
* **Signoz version**: 
* query-service: 0.42.0
* otel-collector: 0.88.19
* **Browser version**:
* **Your OS and version**:

srikanthccv:
### Summary

Fixes #4855


nityanandagohain (approved):
lgtm!

PR #4941, opened 2024-04-30
fix: ch queries sending builder as query type in query range api for exceptions alerts

Vikrant2520:
### Summary

- the logic in the alerts was based on `datasource` from query hence the exceptions alerts were not working as expected as it always routed to trace based alerts.
- refactored the logic to include the `alertType` as query params to maintain the sanity 
- if the query param `alertType` is not present we will extract the same from staged query like we are doing currently to support backwards compatibility 

#### Screenshots

Logs -> Alerts 
Dashboard (Traces / Logs / Metrics ) -> Alerts

https://github.com/SigNoz/signoz/assets/54737045/5a16f352-093f-48cd-8f16-516f9ab675b4

Alerts UI navigation 

https://github.com/SigNoz/signoz/assets/54737045/53d61444-98bd-4e81-ad82-ced11aa3c2c0

### DIFF
--- a/frontend/src/container/CreateAlertRule/defaults.ts
+++ b/frontend/src/container/CreateAlertRule/defaults.ts
@@ -130,7 +130,7 @@ export const exceptionAlertDefaults: AlertDef = {
 					disabled: false,
 				},
 			},
-			queryType: EQueryType.QUERY_BUILDER,
+			queryType: EQueryType.CLICKHOUSE,
 			panelType: PANEL_TYPES.TIME_SERIES,
 			unit: undefined,
 		},
--- a/frontend/src/container/CreateAlertRule/defaults.ts
+++ b/frontend/src/container/CreateAlertRule/defaults.ts
@@ -130,7 +130,7 @@ export const exceptionAlertDefaults: AlertDef = {
 					disabled: false,
 				},
 			},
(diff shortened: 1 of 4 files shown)


frontend/src/container/CreateAlertRule/defaults.ts
```diff
@@ -130,7 +130,7 @@ export const exceptionAlertDefaults: AlertDef = {
 					disabled: false,
 				},
 			},
-			queryType: EQueryType.QUERY_BUILDER,
+			queryType: EQueryType.CLICKHOUSE,
```
SagarRajput-7, line 133: is this expected to change default to Clickhouse, other places have queryBuilder only i guess

SagarRajput-7 (approved):

frontend/src/container/CreateAlertRule/defaults.ts
Vikrant2520: yes this is expected

Vikrant2520: for exceptions the default is clickhouse

SagarRajput-7 (approved):

YounixM (approved):

SagarRajput-7 (approved):

Issue #4845, opened 2024-04-09
Migrate all possible v3 alerts and dashboards to v4


srikanthccv:
## Bug description

The percentage of alerts that can be migrated to v4 without any issues should be big. Write a migration script for alerts and dashboards.

srikanthccv:
This is done now.

PR #5010, opened 2024-05-15
chore: migrate alerts to v4 for supported operators

srikanthccv:
### Summary

Part of https://github.com/SigNoz/signoz/issues/4845

The migrations have been moved to query-service to address the collaborative/operational issues of running data migrations and less about the technical challenges such as db locking, and conflict resolution, etc... These changes allow authors to not worry about the mismatched types for reading data, pushing code to https://github.com/SigNoz/signoz-db-migrations, building images and pushing and finally working with Prashant to deploy. 

As and when we move out of SQLite and support multiple replicas of query service, several things will require re-evaluation and this migration is one of them. Unlike the schemas, there is a "down" for the data migrations. There is only forward migration mainly because supporting such a thing would mean db backed up and restored (this is easy for sqlite but not practical for relational dbs). The onus is on the author of the migration to make sure the migration doesn't corrupt the database and runs gracefully.

### DIFF
--- a/ee/query-service/main.go
+++ b/ee/query-service/main.go
@@ -14,7 +14,9 @@ import (
 	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
 	"go.signoz.io/signoz/ee/query-service/app"
 	"go.signoz.io/signoz/pkg/query-service/auth"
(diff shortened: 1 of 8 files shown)


nityanandagohain (approved):
lgtm

PR #5015, opened 2024-05-16
fix: multiple widgets getting created and hence blocking the delete

Vikrant2520:
### Summary

Fixes https://github.com/SigNoz/engineering-pod/issues/1288

- the pre and after widgets didn't handle the `-1` index case hence they were adding multiple widgets with same `ids`
- now since a few dashboards might have been corrupted so to allow users to delete the existing panels added support for users to multiple deletes when same widget id is removed multiple times

#### Screenshots

NA

- UI changes with latest backend changes
- backend changes with postman


srikanthccv (approved):
LGTM

pkg/query-service/app/dashboards/model.go
```diff
@@ -326,7 +326,15 @@ func UpdateDashboard(ctx context.Context, uuid string, data map[string]interface
 	if existingTotal > newTotal && existingTotal-newTotal > 1 {
 		// if the total count of panels has reduced by more than 1,
 		// return error
-		return nil, model.BadRequest(fmt.Errorf("deleting more than one panel is not supported"))
+		existingIds := getWidgetIds(dashboard.Data)
+		newIds := getWidgetIds(data)
+
+		differenceIds := getIdDifference(existingIds, newIds)
+
+		if len(differenceIds) > 1 {
```
ankitnayan, line 334: such checks are a hack and not a solution

srikanthccv (approved):

YounixM (approved):

pkg/query-service/app/dashboards/model.go
srikanthccv: We had a case where a client call would wipe out the whole dashboard. We restricted the update call since then. This loosens the restriction to delete any instances of widget with same id.

Vikrant2520: Currently we were checking if the difference in `existingCount` and `newCount` is more than 1 then we block the delete. But now since we want to support deleting multiple widgets with same `id` we maintain a map of id's that are passed to be deleted and check the unique id's should be 1 that should be deleted. 

can you please help with any suggestions on how can we improve this?

Vikrant2520: this could also happen while importing the dashboards as well because there is no check on that as well

srikanthccv:
Merging this. Please continue the discussion if it's not answered.

PR #5028, opened 2024-05-20
chore: use an equivalent fixed column for tag attributes

makeavish:


srikanthccv:
How about updating the logic to only show the original attributes and updating the query builder logic to use the column in the query for those attributes which have a column. This leads to an experience that doesn't require any explanation for what is different b/w dbSystem and `db.system` to the user and the same experience as logs.

makeavish:
> How about updating the logic to only show the original attributes and updating the query builder logic to use the column in the query for those attributes which have a column. This leads to an experience that doesn't require any explanation for what is different b/w dbSystem and `db.system` to the user and the same experience as logs.

But we want to show users which column is indexed and which is not.

srikanthccv:
You can still show that they indexed like logs do with the same original attribute names.

pkg/query-service/app/http_handler.go
```diff
@@ -3106,6 +3106,148 @@ func (aH *APIHandler) getSpanKeysV3(ctx context.Context, queryRangeParams *v3.Qu
...
+				Type:     v3.AttributeKeyTypeTag,
+				IsColumn: true,
+				DataType: v3.AttributeKeyDataTypeString,
+			}
+			spanKeys["client.address"] = v3.AttributeKey{
+				Key:      "httpHost",
+				Type:     v3.AttributeKeyTypeTag,
+				IsColumn: true,
+				DataType: v3.AttributeKeyDataTypeString,
+			}
```
srikanthccv, line 3204: Since the equivalent change in collect is recent and not released yet, we either have to hold this PR for enough data to backfill or just comment out these new additions so users don't see incorrect results for longer time ranges.

pkg/query-service/app/traces/v3/query_builder.go
```diff
@@ -85,18 +85,17 @@ func getClickhouseTracesColumnDataTypeAndType(key v3.AttributeKey) (v3.Attribute
 }
 
 func enrichKeyWithMetadata(key v3.AttributeKey, keys map[string]v3.AttributeKey) v3.AttributeKey {
-	if key.Type == "" || key.DataType == "" {
```
srikanthccv, line 88: I guess what breaks with removing this is when there is an attribute different type or datatype, irrespective of what user explicitly selects their preferred type and datatype, we override? I am not sure how much of a real problem this is. I have come across any customer who had the same attribute name with a different type or data type.

srikanthccv (commented):
Minor comments

pkg/query-service/app/traces/v3/query_builder.go
makeavish: That might be an issue but only when user is using custom attribute key with same name as otel key. I haven't encountered any users with the same attribute name and different types. 
In logs too we are also overriding them.

srikanthccv: > That might be an issue but only when user is using custom attribute key with same name as otel key. I haven't encountered any users with the same attribute name and different types. In logs too we are also overriding them.

There are a few users with the same key but different type/data type and they are not necessarily because of the custom attribute collision with otel key. See https://github.com/SigNoz/signoz/issues/5150 & https://github.com/SigNoz/signoz/issues/5359; @nityanandagohain created the issue after a bug request from a customer and then oss user also added that they are impacted by it.

makeavish: > > That might be an issue but only when user is using custom attribute key with same name as otel key. I haven't encountered any users with the same attribute name and different types. In logs too we are also overriding them.
> 
> There are a few users with the same key but different type/data type and they are not necessarily because of the custom attribute collision with otel key. See #5150 & #5359; @nityanandagohain created the issue after a bug request from a customer and then oss user also added that they are impacted by it.

Yeah, I saw community user faced the issue.

Let's block this until we find a fix

srikanthccv: They are independent problems, aren't they? Could we modify this pull request to still honor the user's selection (assuming the request payload is accurate), but for the attributes we know have fixed columns, use those instead?

Issue #3829, opened 2023-10-29
Option to interpret No Data as zero data


srikanthccv:
## Bug description

We show "No Data", where the query returns nothing, which usually happens when no samples are received for the duration. Users have use cases where the absence of data is the same as zero for them. The desired output should be configurable by user case by case.

makeavish:
Is this issue related to https://github.com/SigNoz/signoz/issues/1463?

srikanthccv:
Loosely related but primarily independent.

srikanthccv:
This is very common request: https://signoz-team.slack.com/archives/C05GQNXL3FG/p1699298851398549?thread_ts=1699298774.138579&cid=C05GQNXL3FG

srikanthccv:
This is done

PR #5031, opened 2024-05-20
fix: add zero value for missing timestamps in alert eval

srikanthccv:
### Summary

Part of https://github.com/SigNoz/signoz/issues/3829


srikanthccv:
The main issue with the formula should be addressed at https://github.com/SigNoz/signoz/pull/5045. The ability to configure this for single query is a separate enhancement.

PR #5045, opened 2024-05-21
chore: use querier in alerts

srikanthccv:
### Summary

Last part of https://github.com/SigNoz/signoz/issues/4845


srikanthccv:
This syncs the alerts evaluation to use the same process that query_range API does and keeps what users see and alerts see the same. The formula evaluation is also in sync i.e for v4 a formula if one of the queries doesn't have data then it is interpreted as 0.

pkg/query-service/rules/thresholdRule.go
```diff
@@ -1062,56 +703,91 @@ func (r *ThresholdRule) GetSelectedQuery() string {
...
+	}
 
-		if err != nil {
-			zap.L().Error("failed to prepare metric queries", zap.String("ruleid", r.ID()), zap.Error(err))
-			return nil, fmt.Errorf("failed to prepare metric queries")
+	if params.CompositeQuery.QueryType == v3.QueryTypeBuilder {
+		// check if any enrichment is required for logs if yes then enrich them
+		if logsv3.EnrichmentRequired(params) {
+			// Note: Sending empty fields key because enrichment is only needed for json
+			// TODO: Add support for attribute enrichment later
```
nityanandagohain, line 725: please create an issue for this if not already created.

reason:- 

* sometimes alert is created but the key is indexed later, but not enriching it uses the non indexed key

nityanandagohain (approved):
LGTM

PR #5074, opened 2024-05-24
fix: step interval not working for logs alerts

Vikrant2520:
#### Screenshots

NA


srikanthccv:
I shared the feedback here https://signoz-team.slack.com/archives/C02TJ466H8U/p1717012138360279?thread_ts=1716785112.844369&cid=C02TJ466H8U.

vikrantgupta25:
let me check the feedback and get this merged ✅

srikanthccv:
Here is what I suggest - don't populate the `stepInterval` on the client side by default or automatically when the time range changes. The backend will always select the most reasonable value even when nothing is sent by the client. Use the Aggregate Every for user input and never override it. This way you can always keep the user input value as the source.

vikrantgupta25:
> Here is what I suggest - don't populate the `stepInterval` on the client side by default or automatically when the time range changes. The backend will always select the most reasonable value even when nothing is sent by the client. Use the Aggregate Every for user input and never override it. This way you can always keep the user input value as the source.

done, applied the same logic in code as well.

srikanthccv:
The frontend changes are fine but some backend changes should be made before we merge this.

srikanthccv:
@vikrantgupta25 this change is working for alerts. please enable this across the product wherever query builder is used.

vikrantgupta25:
sure @srikanthccv on it...

vikrantgupta25:
> @vikrantgupta25 this change is working for alerts. please enable this across the product wherever query builder is used.

@srikanthccv it is done for all places where QB is present ✅

srikanthccv (approved):
Functionally LGTM. Tested
- Dashboard landing old and new dashboard
- Panel view and edit
- Alert from panel
- Alerts standalone

YounixM (approved):

PR #5083, opened 2024-05-27
fix: escape string for contains and ncontains

### LINKED ISSUE #4142: values with special chars are not escaped for the operators CONTAINS/NOT_CONTAINS
## Bug description

Use query builder with chip `key CONTAINS "string' with single quote"` and run. It should not fail with 400 bad request. The single quotes should be escaped. Was there a reason why we do this for other operators but not for the `CONTAINS`/`NOT_CONTAINS`

https://github.com/SigNoz/signoz/blob/0b991331d7afea3cb7a3b2a0520a38e4347ff9cd/pkg/query-service/app/logs/v3/query_builder.go#L189-L199

More context: https://signoz-team.slack.com/archives/C05GQNXL3FG/p1701444740388639

nityanandagohain:
escape string for contains and ncontains

fixes #4142


srikanthccv:
Should this done everywhere we support (n)contains https://github.com/search?q=repo%3ASigNoz%2Fsignoz%20v3.FilterOperatorContains%2C%20v3.FilterOperatorNotContains&type=code

srikanthccv (approved):

PR #5106, opened 2024-05-29
chore(query-service): update `GoValuateFuncs` definitions

prashant-shahi:
### Summary

- Fixed `now` function
- Remove `toUnixTimestamp` query-builder function which was breaking - we shall bring this back if the need arises for any customers for computing timestamp query with static date string values.

If you had any queries using `toUnixTimestamp(now())`, you can use `now()` instead.

#### Related Issues / PR's

Related to #4877

Any query using `toUnixTimestamp` or `now`.

Signed-off-by: Prashant Shahi <prashant@signoz.io>


srikanthccv (approved):

srikanthccv:
Since `A - now()` is not working. I don't think this will fix it because the error says mismatched data type.

prashant-shahi:
yeah, this will not fix that issue.

srikanthccv:
Please update the funtions `GoValuateFuncs["now"] ` and `GoValuateFuncs["toUnixTimestamp"]` to return a float64 value and check.

prashant-shahi:
> Please update the funtions GoValuateFuncs["now"]  and GoValuateFuncs["toUnixTimestamp"] to return a float64 value and check.

I have done the exact same thing locally. Testing it now. 😄

srikanthccv (dismissed):

srikanthccv (approved):

prashant-shahi:
> 	GoValuateFuncs["toUnixTimestamp"] = func(args ...interface{}) (interface{}, error) {

@srikanthccv If the returned value always needs to `float64` for queries to work, can we update it to explicitly return the same?

PR #5108, opened 2024-05-29
fix: remove the need for unnecessary dummy time column

srikanthccv:
### Summary

Fixes https://github.com/SigNoz/engineering-pod/issues/1284


pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -4433,7 +4433,7 @@ func (r *ClickHouseReader) GetLogAttributeValues(ctx context.Context, req *v3.Fi
 
 }
 
-func readRow(vars []interface{}, columnNames []string) ([]string, map[string]string, []map[string]string, v3.Point) {
+func readRow(vars []interface{}, columnNames []string, countOfNumberCols int) ([]string, map[string]string, []map[string]string, v3.Point) {
```
srikanthccv, line 4436: If the number of `Number` cols is just 1 in the result, then there is no ambiguity so it will be interpreted as the `value` column.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -4574,20 +4574,7 @@ func readRowsForTimeSeriesResult(rows driver.Rows, vars []interface{}, columnNam
...
-		var groupingSetsPoint *v3.Point
-		for idx, point := range points {
-			if point.Timestamp <= 0 {
-				groupingSetsPoint = &point
-				// remove the grouping sets point from the list of points
-				points = append(points[:idx], points[idx+1:]...)
-				break
-			}
-		}
-		series := v3.Series{Labels: seriesToAttrs[key], Points: points, GroupingSetsPoint: groupingSetsPoint, LabelsArray: labelsArray[key]}
```
srikanthccv, line 4590: The groping sets was primarily used for `Limit` clause in the metrics. However, we no longer rely on it and instead use the [ApplyMetricLimit](https://github.com/SigNoz/signoz/blob/592073a56426c757aa5c3660a376e4ce5453dff3/pkg/query-service/postprocess/limit.go#L13)

pkg/query-service/app/querier/querier.go
```diff
@@ -100,7 +100,7 @@ func (q *querier) execClickHouseQuery(ctx context.Context, query string) ([]*v3.
 		points := make([]v3.Point, 0)
 		for pointIdx := range series.Points {
 			point := series.Points[pointIdx]
-			if point.Timestamp > 0 {
+			if point.Timestamp >= 0 {
```
srikanthccv, line 103: `Timestamp` will be zero where it no timestamp column. We don't want to discard that point.

nityanandagohain (approved):

PR #5112, opened 2024-05-30
feat: base setup for table columns sorting

vikrantgupta25:
#### Screenshots

NA

### DIFF
--- a/frontend/src/container/GridTableComponent/index.tsx
+++ b/frontend/src/container/GridTableComponent/index.tsx
@@ -20,8 +20,10 @@ function GridTableComponent({
 	query,
 	thresholds,
 	tableProcessedDataRef,
+	sortColumns,
 	...props
 }: GridTableComponentProps): JSX.Element {
+	console.log(sortColumns);
 	const { t } = useTranslation(['valueGraph']);
 	const { columns, dataSource } = useMemo(
 		() =>
--- a/frontend/src/container/GridTableComponent/types.ts
+++ b/frontend/src/container/GridTableComponent/types.ts
@@ -5,12 +5,14 @@ import {
 	ThresholdProps,
 } from 'container/NewWidget/RightContainer/Threshold/types';
 import { RowData } from 'lib/query/createTableColumnsFromQuery';
+import { SORT_TYPES } from 'types/api/dashboard/getAll';
 import { Query } from 'types/api/queryBuilder/queryBuilderData';
 
 export type GridTableComponentProps = {
 	query: Query;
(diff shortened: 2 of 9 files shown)


ankitnayan:
will the sorting get reflected in the QB?

srikanthccv:
I do not think the sorting of the columns means ordering by their name but rather sorting by values. Users typically have n number of columns and several pages of values. Being able to sort by values helps in data analysis.

vikrantgupta25:
> I do not think the sorting of the columns means ordering by their name but rather sorting by values. Users typically have n number of columns and several pages of values. Being able to sort by values helps in data analysis.

they can already do that right ? in the orderBy clause

ankitnayan:
> I do not think the sorting of the columns means ordering by their name but rather sorting by values. Users typically have n number of columns and several pages of values. Being able to sort by values helps in data analysis.

this is right. Resizing and arranging columns is based on column names but sorting is on values of 1 column which should get reflexed in the QB. 

@srikanthccv we can leave sorting of values in clickhouse queries? Mostly for QB, sorting on clicking table names is just a shortcut to applying order by in the QB right?

ankitnayan:
> > I do not think the sorting of the columns means ordering by their name but rather sorting by values. Users typically have n number of columns and several pages of values. Being able to sort by values helps in data analysis.
> 
> they can already do that right ? in the orderBy clause

asked more https://signoz-team.slack.com/archives/C05AXQL92BT/p1717066664751189?thread_ts=1715348260.465959&cid=C05AXQL92BT

vikrantgupta25:
> we can leave sorting of values in clickhouse queries? Mostly for QB, sorting on clicking table names is just a shortcut to applying order by in the QB right

We can have permanent ordering by adding it in the query itself and the on the fly ordering from the table headers itself. If they are interested in one kind of ordering ( long term ) they can add it in the `orderBy` clause and some custom sorts which is needed on the fly to debug something can be temporarily added by table headers.

vikrantgupta25:
Having this temporary solution helps us to wipe up the distinction between QB / Clickhouse as we are not manipulating the query itself

ankitnayan:
okay to me. So, no changes to QB but just a sorting of values using table headers?

vikrantgupta25:
> okay to me. So, no changes to QB but just a sorting of values using table headers?
yes attaching the video for reference

https://github.com/SigNoz/signoz/assets/54737045/adc35931-8901-4941-94ae-2521a659dc37

vikrantgupta25:
https://github.com/SigNoz/signoz/pull/5114 PR according to the discussion

PR #5116, opened 2024-05-30
fix: random sizing issue with alerts query builder

vikrantgupta25:
### Summary

- global styles getting overriden for `options` class 
- make ant-picker component and popover to be displayed in modals and drawers. ( z-index )

#### Related Issues / PR's

fixes https://github.com/SigNoz/engineering-pod/issues/1322

#### Screenshots

NA

### DIFF
--- a/frontend/src/styles.scss
+++ b/frontend/src/styles.scss
@@ -247,28 +247,6 @@ body {
 	padding-right: 20px;
 }
 
-.ant-drawer {
-	// do not modify this as this is required to hide intercom in modals
-	z-index: 1000000000000;
-}
-
-.ant-modal-root {
-	.ant-modal-mask {
-		z-index: 2000000000000 !important;
-	}
-	.ant-modal-wrap {
-		z-index: 2000000000000 !important;
-	}
-}
-
-.ant-select-dropdown {
-	z-index: 3000000000000 !important;
-}
-
-.ant-picker-dropdown {
-	z-index: 3000000000000 !important;
-}
-
-.ant-popover {
-	z-index: 3000000000000 !important;
+.intercom-lightweight-app {
+	z-index: 980 !important;
 }
--- a/frontend/src/styles.scss
+++ b/frontend/src/styles.scss
@@ -247,6 +247,22 @@ body {
 	padding-right: 20px;
 }
 
+// we want the z-index to be less than 1000 (z-index for antd drawers) as we do not want intercom to interfare when drawers are open
 .intercom-lightweight-app {
 	z-index: 980 !important;
 }
+
+/**
+
+z-index chart for components across application
+
+intercom        -  980
+tooltip			-  1070
+popover 		-  1030
+drawer  		-  1000
+time picker 	-  1050
+dropdown 		-  1050
+modal 			-  1000
+toast			-  2050
+
+*/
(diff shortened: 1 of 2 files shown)


SagarRajput-7 (dismissed):

frontend/src/styles.scss
```diff
@@ -264,3 +264,11 @@ body {
 .ant-select-dropdown {
 	z-index: 3000000000000 !important;
 }
+
+.ant-picker-dropdown {
+	z-index: 3000000000000 !important;
```
YounixM, line 269: Why are we setting such high numbers for z-index? They should be incremental.

We should have a place where we can look up for the layout hierarchy

vikrantgupta25: Because the default numbers are very high already for intercom and we needed to exceed that

YounixM: We are increasing them randomly without consideration. We should fix all these places.  The next similar PR will have an even higher number.

YounixM: Check if you can update the old values in this PR or let's create a PR to fix the old values and then raise this PR on top on that.

vikrantgupta25: Not increasing them randomly. The drawer is set to be 1*(base limit) . the modals need to be rendered above drawer which is (2*base limit ). then modals and drawers can have dropdowns and popovers which is 3*(base limit )

YounixM: drawer z-index is 1000

vikrantgupta25:
```
.ant-drawer {
	// do not modify this as this is required to hide intercom in modals
	z-index: 1000000000000;
}

.ant-modal-root {
	.ant-modal-mask {
		z-index: 2000000000000 !important;
	}
	.ant-modal-wrap {
		z-index: 2000000000000 !important;
	}
}

.ant-select-dropdown {
	z-index: 3000000000000 !important;
}```

Comment (vikrantgupta25) [reply frontend/src/styles.scss:269]: image
This is the z-index for intercom

Comment (YounixM) [reply frontend/src/styles.scss:269]: > ```
> .ant-drawer {
> 	// do not modify this as this is required to hide intercom in modals
> 	z-index: 1000000000000;
> }
> 
> .ant-modal-root {
> 	.ant-modal-mask {
> 		z-index: 2000000000000 !important;
> 	}
> 	.ant-modal-wrap {
> 		z-index: 2000000000000 !important;
> 	}
> }
> 
> .ant-select-dropdown {
> 	z-index: 3000000000000 !important;
> }```
> ```

This was written by us.

File: frontend/src/styles.scss [hunk 2]
```diff
@@ -247,20 +247,6 @@ body {
...
-	}
-	.ant-modal-wrap {
-		z-index: 2000000000000 !important;
-	}
-}
-
-.ant-select-dropdown {
-	z-index: 3000000000000 !important;
+.intercom-lightweight-app {
+	z-index: 980 !important;
```
YounixM, line 251: Add reasoning on why we want it to be 980 and also list component that are affecting stacking context in our base css file for reference.

YounixM:
Elements that need z-index adjustment

1. Tooltip
2. Popover
3. Modal
4. Drawer
5. Dropdown
6. TimePicker

frontend/src/styles.scss
vikrantgupta25: done ✅

PR #5120, opened 2024-05-31
feat: add spanKind and status in span response

### LINKED ISSUE #4925: Span status message should be stored and shown in UI
## Bug description

Only the status code is captured and stored https://github.com/SigNoz/signoz-otel-collector/blob/main/exporter/clickhousetracesexporter/clickhouse_exporter.go#L373. Just knowing span has an error status doesn't help. The main part of the status is the message which adds more detail about the error.

nityanandagohain:
Fixes #4925 

add spanKind and statusCodeString and statusMessage


makeavish (approved):

makeavish (approved):

srikanthccv:
One question: do these show up in attribute suggestions? If not, I believe they should be included as well. It doesn't have to be in this PR if you prefer that way.

nityanandagohain:
@srikanthccv  yes they show up , as they are driven by tag attributes
image

image

image

srikanthccv:
Should we wait for the data to be backfilled for a week? and only show the string version kind i.e. `spanKind` and remove the number `kind` from suggestions?

nityanandagohain:
Yeah we can wait for backfill and then release this, not sure if we should remove kind from suggestion.

Meanwhile we will also need a frontend PR to show this values in the trace detail page.

srikanthccv (approved):
It's hard to remember the kind number mapping so I think it should be faded away.

Issue #5121, opened 2024-05-31
Incorrect variable values when dashboard is imported for the first time


srikanthccv:
## Bug description

See https://www.loom.com/share/2082bd7ebd1848289be88701969ea117

PR #5123, opened 2024-05-31
fix: dashboard variables should properly load for imported dashboards

vikrantgupta25:
### Summary

- imported dashboards variables not working and showing empty data
- now since we moved to sequential approach we check for `variablesToBeUpdated` queue and the variable should be present in the queue to get updated. 
- now even if we add the variables on the first load of dashboard it won't work as it will update all the existing values nested structure as well. 
- so we only update if either present in queue or the `selectedValue` is not something that came in the new options.  

#### Related Issues / PR's

fixes https://github.com/SigNoz/signoz/issues/5121

#### Screenshots

https://github.com/SigNoz/signoz/assets/54737045/a3221564-1951-40c3-a0f6-77c24058648b


srikanthccv:
This one is tricky. I will take a look.

srikanthccv (approved):
LGTM. One a separate note, It would be great to have tests for small changes so that the test suite grows over the time.

YounixM (approved):

PR #5161, opened 2024-06-06
fix: clone panel randomly jumping around in the middle causing layout issues

vikrantgupta25:
### Summary

- first issue :- the cloned panel was added in the bottom of the layout which caused the least priority in render. hence it was being randomly placed across the screen. `react-grid-layout` starts layouting from top

- second enhancement added here being that the cloned panel should have similar dimensions to the original panel from which it is cloned.

#### Screenshots

https://github.com/SigNoz/signoz/assets/54737045/79db3591-2cc5-49dd-9e43-2aa9fe56db8e

- cloned panels without rows
- cloned panel with rows 
- multiple layout positions.

### DIFF
--- a/frontend/src/container/GridCardLayout/GridCard/WidgetGraphComponent.tsx
+++ b/frontend/src/container/GridCardLayout/GridCard/WidgetGraphComponent.tsx
@@ -131,15 +131,22 @@ function WidgetGraphComponent({
 
 		const uuid = v4();
 
+		// this is added to make sure the cloned panel is of the same dimensions as the original one
+		const originalPanelLayout = selectedDashboard.data.layout?.find(
+			(l) => l.i === widget.id,
+		);
+
+		// added the cloned panel on the top as it is given most priority when arranging
+		// in the layout. React_grid_layout assigns priority from top, hence no random position for cloned panel
 		const layout = [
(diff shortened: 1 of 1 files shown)


frontend/src/container/GridCardLayout/GridCard/WidgetGraphComponent.tsx
```diff
@@ -131,15 +131,22 @@ function WidgetGraphComponent({
 
 		const uuid = v4();
 
+		// this is added to make sure the cloned panel is of the same dimensions as the original one
+		const originalPanelLayout = selectedDashboard.data.layout?.find(
+			(l) => l.i === widget.id,
```
SagarRajput-7, line 136: check if we need any memoisation here.

SagarRajput-7 (approved):

frontend/src/container/GridCardLayout/GridCard/WidgetGraphComponent.tsx
vikrantgupta25: memoization inside a user-triggered action? this will be called on click of `clonehandler` and should always fetch the latest results during that time

YounixM (approved):

PR #5162, opened 2024-06-06
fix: theme analytics not getting reported as user is being fetched delayed

vikrantgupta25:
### Summary

- the user properties being fetched by API call were delayed hence the event was not getting fired 

#### Screenshots

image


frontend/src/AppRoutes/index.tsx
```diff
@@ -178,23 +178,25 @@ function App(): JSX.Element {
...
-				LOCALSTORAGE.THEME_ANALYTICS,
-			);
-			if (!isThemeAnalyticsSent) {
-				trackEvent('Theme Analytics', {
-					theme: isDarkMode ? THEME_MODE.DARK : THEME_MODE.LIGHT,
-					user: pick(user, ['email', 'userId', 'name']),
-					org,
-				});
-				setLocalStorageApi(LOCALSTORAGE.THEME_ANALYTICS, 'true');
+		if (user && user?.email && user?.userId && user?.name) {
```
vikrantgupta25, line 181: added the check to send the event only after the user is fetched. and set the local state to not send further events

frontend/src/constants/localStorage.ts
```diff
@@ -18,5 +18,5 @@ export enum LOCALSTORAGE {
 	DASHBOARD_VARIABLES = 'DASHBOARD_VARIABLES',
 	SHOW_EXPLORER_TOOLBAR = 'SHOW_EXPLORER_TOOLBAR',
 	PINNED_ATTRIBUTES = 'PINNED_ATTRIBUTES',
-	THEME_ANALYTICS = 'THEME_ANALYTICS',
+	THEME_ANALYTICS_V1 = 'THEME_ANALYTICS_V1',
```
vikrantgupta25, line 21: renamed the local storage key as older keys would have already been set for the user.

YounixM (approved):

frontend/src/AppRoutes/index.tsx
```diff
@@ -178,23 +178,25 @@ function App(): JSX.Element {
...
+			try {
+				const isThemeAnalyticsSent = getLocalStorageApi(
+					LOCALSTORAGE.THEME_ANALYTICS_V1,
+				);
+				if (!isThemeAnalyticsSent) {
+					trackEvent('Theme Analytics', {
+						theme: isDarkMode ? THEME_MODE.DARK : THEME_MODE.LIGHT,
```
SagarRajput-7, line 188: We wanted to move away from trackEvent and use logEvent 

cc: @makeavish

SagarRajput-7: This is a NIT for now, approved the PR

SagarRajput-7 (approved):

frontend/src/AppRoutes/index.tsx
vikrantgupta25: we will anyways remove this event in a while so should be okay for now

Issue #5156, opened 2024-06-05
`result` field in table panel of dashboard always named as `count`


ankitnayan:
and if there is no `result` column, the table automatically adds a column called `result` filled with 0

srikanthccv:
@vikrantgupta25 The legend format for the aggregation column is not respected. Please try this query and check.
```
SELECT now() as ts, 
stringTagMap['service.name'] AS `service.name`, 
toFloat64(avg(durationNano))/1000/1000 AS value
FROM signoz_traces.distributed_signoz_index_v2  
WHERE timestamp BETWEEN {{.start_datetime}} AND {{.end_datetime}}
GROUP BY (`service.name`, ts) order by (`service.name`, ts) ASC;
```

> and if there is no result column, the table automatically adds a column called result filled with 0

@ankitnayan, the table column is usually viz for aggregation so it expects at least one aggregation result column. If there is no aggregation but just a plain list of values to read from somewhere, it should go in `list` panel type.

ankitnayan:
changing one of them to `result` still leads to incorrect columns and labels

vikrantgupta25:
merged the frontend changes for the above issue

PR #5164, opened 2024-06-06
fix: clickhouse and promQL queries table column headers not handled

vikrantgupta25:
### Summary

- the handling for adding the correct table columns based on legend or the operators was not present 
- fixes the column units respecting clickhouse and promQL queries as well 

#### Related Issues / PR's

fixes https://github.com/SigNoz/signoz/issues/5156

#### Screenshots

https://github.com/SigNoz/signoz/assets/54737045/bdf932f6-5cb0-4b89-ad9f-dde3729f1322


srikanthccv:
Let's hold this for some time. We are going to change the way ClickHouse queries are handled so we don't have to worry about the column name details.

vikrantgupta25:
> Let's hold this for some time. We are going to change the way ClickHouse queries are handled so we don't have to worry about the column name details.

can you point me to what change are we expecting ? we can get this out if that change might take some time ? because currently there is no handling for clickhouse and promQL so either the default `count` appears or whatever values present in the QB appears.

YounixM (approved):

srikanthccv:
> can you point me to what change are we expecting ?

There won't be any legend formatting for ClickHouse queries.

srikanthccv:
We can get this merged independent of the other changes that follow.

Issue #5153, opened 2024-06-05
Enable plotting multiple timeseries in a chart using clickhouse query


ankitnayan:
The current implementation restricts labels to 1 column with name `result`. This should support more labels and add docs about how to do this

srikanthccv:
Metabase: https://www.metabase.com/learn/visualization/line-charts

> Metabase can’t read minds (yet), so it doesn’t know which columns from the Orders table to use for the x and y axes. To create a line chart, you’ll need to pick a metric for Metabase to plot over time. For example, you could show order totals over time by setting the x axis to created_at and the y axis to total. Metabase will automatically plot the line chart:

ClickHouse data source grafana:

![Screenshot 2024-06-06 at 6 11 39 PM](https://github.com/SigNoz/signoz/assets/22846633/9250cfd5-6baf-4029-a2ab-a833e869d8ae)

For the query
```
SELECT
    type,
    client_version_major,
    avg(query_duration_ms) as avg_duration,
    toStartOfInterval(event_time, INTERVAL 1 MINUTE) AS ts
FROM system.query_log
WHERE $__timeFilter(event_time)
GROUP BY ts, type, client_version_major
HAVING avg_duration > 0 AND max_duration > 0
ORDER by ts, type, client_version_major
```

It considers every number as a time series result. As a result, the `client_version_major` 2 is the line chart instead of label.

ankitnayan:
I see, so that's a bug there. Okay to use `result_1`, `result_2` ? But the UI shouldn't should `result_1` and `result_2` as labels. Labels should be more meaningful.

srikanthccv:
Or how about asking users to cast the number type labels to string in the final result?

ankitnayan:
that is even better IMO as we can treat labels as string

srikanthccv:
The new expected format could break some existing dashboards and alerts since we are changing how the results are interpreted. I will collect queries from the existing users and find a plan for shipping this new change.

srikanthccv:
This is now possible but need to write docs for the expected format of the result. Will create a separate doc issue and get it published.

PR #5174, opened 2024-06-09
feat: reading multi-series result for ClickHouse queries

srikanthccv:
### Summary

Fixes https://github.com/SigNoz/signoz/issues/5153
Partially address https://github.com/SigNoz/signoz/issues/5156

The initial query range proposal included only one aggregation per-query. The logic for reading the data from the ClickHouse result was written with this assumption. It serves mostly fine. However, when there are more than one aggregation that uses the same set of filters, it is inefficient to run multiple queries. For instance, One can write a ClickHouse query to get the P95, P99 in the same query but they are forced to write two queries because of the limitation from our side. This change enables reading multi series from the ClickHouse. 

In this, we assume any number result is a value result and everything else should be string. For example, if you write a ClickHouse query with group by as http.status_code, you should make sure to cast it to string otherwise the status number will be interpreted as result. Please see the issue for the discussion to understand why we chose this path.


nityanandagohain (approved):
LGTM

prashant-shahi (approved):

shivanshuraj1333:
@srikanthccv shall we merge this?

srikanthccv:
This is a backwards incompatible change. I am going to hold it off and re-visit.

PR #5195, opened 2024-06-11
feat: added an option to create channel when no Channels are available in alert config

SagarRajput-7:
#### Related Issues / PR's

SigNoz/engineering-pod#1366 

#### Screenshots

Screenshot 2024-06-12 at 3 10 29 AM

https://github.com/SigNoz/signoz/assets/162284829/2b24d4b6-4234-4d40-a532-6b20578dd141

### DIFF
--- a/frontend/src/container/FormAlertRules/BasicInfo.tsx
+++ b/frontend/src/container/FormAlertRules/BasicInfo.tsx
@@ -1,11 +1,17 @@
+import './FormAlertRules.styles.scss';
+
 import { PlusOutlined } from '@ant-design/icons';
 import { Button, Form, Select, Switch, Tooltip } from 'antd';
 import getChannels from 'api/channels/getAll';
 import ROUTES from 'constants/routes';
+import useComponentPermission from 'hooks/useComponentPermission';
 import useFetch from 'hooks/useFetch';
 import { useCallback, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
+import { useSelector } from 'react-redux';
+import { AppState } from 'store/reducers';
 import { AlertDef, Labels } from 'types/api/alerts/def';
+import AppReducer from 'types/reducer/app';
 import { requireErrorMessage } from 'utils/form/requireErrorMessage';
 import { popupContainer } from 'utils/selectPopupContainer';
 
@@ -36,6 +42,11 @@ function BasicInfo({
 	const { t } = useTranslation('alerts');
 
 	const channels = useFetch(getChannels);
+	const { role } = useSelector<AppState, AppReducer>((state) => state.app);
+	const [addNewChannelPermission] = useComponentPermission(
+		['add_new_channel'],
+		role,
+	);
 
 	const [
 		shouldBroadCastToAllChannels,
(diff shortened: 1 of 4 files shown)


frontend/src/container/FormAlertRules/BasicInfo.tsx
```diff
@@ -54,6 +60,11 @@ function BasicInfo({
 		});
 	};
 
+	const noChannels = !channels.payload?.length;
```
vikrantgupta25, line 63: if the API fails then this would still evaluate to `true` and the tooltip would appear. ( Create a notification channel to configure alert ) , is this expected ?

frontend/src/container/FormAlertRules/BasicInfo.tsx
```diff
@@ -137,32 +148,58 @@ function BasicInfo({
...
+							noChannels
+								? 'Create a notification channel to configure alert'
+								: undefined
+						}
+						placement="right"
+					>
+						<Switch
+							checked={shouldBroadCastToAllChannels}
+							onChange={handleBroadcastToAllChannels}
+							disabled={noChannels}
```
vikrantgupta25, line 162: should be disabled on loading as well ?

vikrantgupta25:
image

Seems like the button has a box shadow, should remove that..

srikanthccv:
I don't see any role handling here. An _Editor_ can create alerts but they can't create an alert channel, which only the admin can do. Showing the button for everyone and then failing with a forbidden error is a bad user experience.

frontend/src/container/FormAlertRules/BasicInfo.tsx
SagarRajput-7: Done

frontend/src/container/FormAlertRules/BasicInfo.tsx
SagarRajput-7: Done

frontend/src/container/FormAlertRules/BasicInfo.tsx
```diff
@@ -137,32 +159,64 @@ function BasicInfo({
...
+				)}
+
+				{noChannels && addNewChannelPermission && (
+					<Button
+						onClick={handleCreateNewChannels}
+						icon={<PlusOutlined />}
+						className="create-notification-btn"
+					>
+						Create a notification channel
+					</Button>
```
srikanthccv, line 219: When there are no channels and the user is an editor, I think it makes sense to add a note that only the admin can create a new channel and nudge them to get the channel created. See this conversation https://app.intercom.com/a/inbox/wmkurvb7/inbox/conversation/3561

SagarRajput-7: Yeah, rather than hiding we can do some improvements here to better communicate with users to let them know what is happening here.

the current implementation is what we had on the alert channels page, I think we should improve it there also.

srikanthccv: I highly suggest we do it because we get support questions asking the same. I will leave it to you whether to address it in this PR or a separate PR.

SagarRajput-7: This looks good, made the change to both this page and alert channels page

image

image

image

makeavish (approved):

srikanthccv (approved):

YounixM (approved):

PR #5199, opened 2024-06-12
fix: default alerts yaxis unit doesn't reset on page change

vikrantgupta25:
### Summary

- the last value from current query was not getting rest hence the older value was being picked 
- hence the below dropdown was not getting values.
- selecting a default value without knowing the data type doesn't makes sense. It is a user action.

#### Screenshots

https://github.com/SigNoz/signoz/assets/54737045/543a1d5e-07a5-47bd-8a47-80185fce721e

- creation of alerts


frontend/src/container/FormAlertRules/index.tsx
```diff
@@ -102,6 +103,13 @@ function FormAlertRules({
 	const [alertDef, setAlertDef] = useState<AlertDef>(initialValue);
 	const [yAxisUnit, setYAxisUnit] = useState<string>(currentQuery.unit || '');
 
+	useEffect(() => {
+		if (!isEqual(currentQuery.unit, yAxisUnit)) {
```
vikrantgupta25, line 107: didn't add conditional check here for currentQuery.unit as the default for that is undefined hence we need `''` in that case

YounixM (approved):

Issue #5103, opened 2024-05-29
OrderBy and Limit does not work in Formula in dashboard


ankitnayan:

YounixM:
@srikanthccv: As discussed, assigning this to you.

PR #5201, opened 2024-06-12
fix: apply having and limit clause on formulas

srikanthccv:
### Summary

Fixes https://github.com/SigNoz/signoz/issues/5103

### DIFF
--- a/pkg/query-service/app/http_handler.go
+++ b/pkg/query-service/app/http_handler.go
@@ -3065,6 +3065,10 @@ func (aH *APIHandler) queryRangeV3(ctx context.Context, queryRangeParams *v3.Que
 		postprocess.ApplyFunctions(result, queryRangeParams)
 	}
 
+	if queryRangeParams.CompositeQuery.FillGaps {
+		postprocess.FillGaps(result, queryRangeParams)
+	}
+
 	resp := v3.QueryRangeResponse{
 		Result: result,
 	}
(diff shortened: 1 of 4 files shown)


nityanandagohain (approved):
changes looks good to me, but this also means that we should remove having from the SQL queries for builder queries right ?


I had one query though if there are two queries, A & B, and A has some having clause.

Now I have a formula F(A,B) with some having clause, should the having clause of A be taken into account, yes right ?


Also this changes mean that we can remove having for SQL queries to clickhouse right as it will be done in postprocess ?

srikanthccv:
> but this also means that we should remove having from the SQL queries for builder queries right ?

For v3, it doesn't make any difference because applying having again gives the same result. In v4, the formula query is evaluated using postprocess so there is nothing to remove. 

>Now I have a formula F(A,B) with some having clause, should the having clause of A be taken into account, yes right ?

Yes

>Also this changes mean that we can remove having for SQL queries to clickhouse right as it will be done in postprocess ?

Technically yes, and it's better because it will be cache effective. You can remove this https://github.com/SigNoz/signoz/blob/1f4a8b98340d83ff9b265395bc63c73ce64e0dfd/pkg/query-service/app/queryBuilder/query_builder.go#L371-L375. The having posts aggregation filter. For example, if there are two same queries but different having. Today logs use the two different cache entries but there is no need to do that. Just use the one cache item and run the having part in query service.

nityanandagohain:
Sure creating an issue for that.

PR #5202, opened 2024-06-12
fix: column sorting logic when units are present

vikrantgupta25:
### Summary

- fix the column sorting when units are being added. Extract the number from the table value columns and sort on the basis of that

#### Screenshots

NA

- table sorting with and without units


srikanthccv:
No, this is not correct. It should be done on the original raw value from the response. Imagine a case where there is `1GB` and `200MB`.

srikanthccv:
Ok. I just saw the description and wanted to share why it fails.

srikanthccv:
Fixed https://github.com/SigNoz/signoz/pull/5249

PR #5219, opened 2024-06-14
fix: add bar panel type support

makeavish:
### Summary

Add bar panel type. Bar panel type is exactly same as graph panel type. In case of traces and logs we are doing special handling for graph panel type so just added bar panel type there.
For metrics @srikanthccv mentioned that graph has no special handling so bar panel type should work.

WHY bar panel type?

Frontend was saving bar panel alerts and dashboards as panel type but querying as graph panel type. Since alert rule was being saved as bar panel type so when the alert where evaluated then those were evaluated as bar panel type and alerts based on logs and traces were failing.


pkg/query-service/app/logs/v3/query_builder.go
```diff
@@ -255,7 +255,7 @@ func buildLogsQuery(panelType v3.PanelType, start, end, step int64, mq *v3.Build
 			"SELECT now() as ts,"
 		// step or aggregate interval is whole time period in case of table panel
 		step = (utils.GetEpochNanoSecs(end) - utils.GetEpochNanoSecs(start)) / 1000000000
-	} else if panelType == v3.PanelTypeGraph || panelType == v3.PanelTypeValue {
+	} else if panelType == v3.PanelTypeGraph || panelType == v3.PanelTypeBar || panelType == v3.PanelTypeValue {
```
srikanthccv, line 258: Can this be `else` instead of `else if` that works for the rest of the panel types too?

makeavish: Don't want to change this as it might have unintended changes. For now let's keep it as it is.

srikanthccv: Please help me understand unintended things could happen. What worked will continue to work because nothing is being removed and every other panel type is an aggregation type so I want to understand what will be the issue.

srikanthccv: I would suggest taking time to fix this so that we don't get invalid query alerts. Meanwhile, I will fix the panel type in query builder alerts because rn we only support time series and there is no issue in fixing it in alerts.

makeavish: > Please help me understand unintended things could happen. What worked will continue to work because nothing is being removed and every other panel type is an aggregation type so I want to understand what will be the issue.

I was worried about list panel type case, I just checked and verified we are overwriting that so this case can be made.

makeavish: > I would suggest taking time to fix this so that we don't get invalid query alerts. Meanwhile, I will fix the panel type in query builder alerts because rn we only support time series and there is no issue in fixing it in alerts.

We have already addressed issue on frontend now. So creation of new alerts via UI should only be possible for time series. We can make changes on backend as well to prevent via API.

srikanthccv:
This pull request (PR) initially aimed to fix incorrect queries in alerts. However, we addressed that issue by making fixes in the frontend https://github.com/SigNoz/signoz/pull/5251 and query service https://github.com/SigNoz/signoz/pull/5284. Now, the main point I wanted to convey is that we should take the time to properly fix the panel types and their query builder because we should think about the API users who don't have the same insider knowledge about which panel type to use in the request payload. This applies to all signals equally Ideally, the APIs should work correctly regardless of the user's level of expertise.

Currently, there are two main categories of data results:

- Time series data, where we send multiple points per series for a given time range
- Single-point data, where we send one point per series for a given time range

For example, the "pie" panel in the frontend uses a "table" type, which isn't obvious to end-users. We should update the API and query builders to support such things properly, so we can avoid running into the same issue in the future.

makeavish:
Closing this branch and created an issue for handling these new panel types in better way.
https://github.com/SigNoz/signoz/issues/5379

PR #5247, opened 2024-06-17
chore: remove having caluse from key generation and sql query

### LINKED ISSUE #5218: Cache: Remove having caluse
Remove having clause from https://github.com/SigNoz/signoz/blob/1f4a8b98340d83ff9b265395bc63c73ce64e0dfd/pkg/query-service/app/queryBuilder/query_builder.go#L371-L375

as having is now done is post processing.

nityanandagohain:
Fixes #5218 

having clause is not required in cache key generation and SQL query as we are processing it in postprocess. This also enables better key generation for cache.


srikanthccv:
Can you update this https://github.com/SigNoz/signoz/blob/41f91db622d05346c130bac6b35c2ae0fcf6855e/pkg/query-service/app/queryBuilder/query_builder.go#L142-L149 and traces also?

nityanandagohain:
The changes for traces are made, but we have hit a roadbloack

In query range v3
* Formula/expression creates it's own sql query, if we remove having clause from the sql query there is no way to know what is the having clause to apply in the formula result as the formula consist of multiple queries having different having clause.

Issue #5228, opened 2024-06-14
Table row data doesn't align with the response


srikanthccv:
## Bug description

[...]

PR #5248, opened 2024-06-17
fix: table row data doesn't align with the response 

vikrantgupta25:
### Summary

- the calculation for filling of values in the respective columns didn't account for response of the format 
- hence sorting the series order internally based on label columns concatenated on values to maintain the same order across.
- passed the `isEqualQuery` from the parent to the child function as when the series with equal labels is traversed we should not fill `N/A` as it has already been filled by the parent row query

```
series A -> [ label1, label2] 

series B -> [label2, label1]
```

#### Related Issues / PR's

fixes https://github.com/SigNoz/signoz/issues/5228

#### Screenshots

before 
image

after 
image

Verified here http://stagingapp.signoz.io/dashboard/a08716a3-5af6-407f-b81f-7d2f89e33d4e?relativeTime=3d&expandedWidgetId=1f3881cb-1ae1-4d13-abba-64af5feffbb0

### DIFF
--- a/frontend/src/lib/query/createTableColumnsFromQuery.ts
+++ b/frontend/src/lib/query/createTableColumnsFromQuery.ts
@@ -365,6 +365,7 @@ const fillRestAggregationData = (
 	queryTableData: QueryDataV3[],
 	seria: SeriesItem,
 	equalQueriesByLabels: string[],
+	isEqualQuery: boolean,
 ): void => {
 	const nextQueryData =
 		queryTableData.find((q) => q.queryName === column.field) || null;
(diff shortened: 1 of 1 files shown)


srikanthccv (approved):

frontend/src/lib/query/createTableColumnsFromQuery.ts
```diff
@@ -570,6 +572,29 @@ export const createTableColumnsFromQuery: CreateTableDataFromQuery = ({
 		a.queryName < b.queryName ? -1 : 1,
 	);
 
+	// the reason we need this is because the filling of values in rows doesn't account for mismatch enteries
+	// in the response. Example : Series A -> [label1, label2] and Series B -> [label2,label1] this isn't accounted for
+	sortedQueryTableData.forEach((q) => {
```
YounixM, line 577: Isn't the data returned in response already sorted ?

vikrantgupta25: no @YounixM it has a custom sorting based on user entry which won't work for us

YounixM (approved):

PR #5249, opened 2024-06-17
fix: table sorting when units are present

vikrantgupta25:
### Summary

- when adding units to any column add another data column with label `label_without_unit` which helps us in sorting
- and when sorting if there is a column present with `without_unit` use that else use the normal column

#### Screenshots

https://github.com/SigNoz/signoz/assets/54737045/9a71831e-81ce-4c71-b31d-934874e6fac1

- table panel type with and without units

### DIFF
--- a/frontend/src/container/GridTableComponent/index.tsx
+++ b/frontend/src/container/GridTableComponent/index.tsx
@@ -62,10 +62,11 @@ function GridTableComponent({
 
 			mutateDataSource = mutateDataSource.map(
 				(val): RowData => {
-					const newValue = val;
+					const newValue = { ...val };
 					Object.keys(val).forEach((k) => {
 						if (columnUnits[k]) {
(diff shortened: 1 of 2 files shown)


frontend/src/container/GridTableComponent/index.tsx
```diff
@@ -62,10 +62,11 @@ function GridTableComponent({
 
 			mutateDataSource = mutateDataSource.map(
 				(val): RowData => {
-					const newValue = val;
+					const newValue = { ...val };
 					Object.keys(val).forEach((k) => {
 						if (columnUnits[k]) {
 							newValue[k] = getYAxisFormattedValue(String(val[k]), columnUnits[k]);
+							newValue[`${k}_without_unit`] = val[k];
```
vikrantgupta25, line 69: long term fix would be have the logic for proper units in house and have a reverse for the same as well

srikanthccv (approved):

YounixM (approved):

Issue #5172, opened 2024-06-07
Trace Explorer - Support column sorting for timestamp, duration


YounixM:
Trace Explorer - Support column sorting for timestamp, duration

Issue #4814, opened 2024-04-04
Allow sorting of list view by clicking on columns in list view


makeavish:
In older traces filter page it was easy to sort by timestamp or duration, by just clicking on column. 
Need similar functionality in New Traces Explorer

PR #5252, opened 2024-06-17
fix: added support column sorting for timestamp, duration

SagarRajput-7:
#### Related Issues / PR's

#5172 
#4814 

#### Screenshots

https://github.com/user-attachments/assets/a07204cf-afc6-45b9-9fa2-adaf356d701f

### DIFF
--- a/frontend/src/container/TracesExplorer/ListView/utils.tsx
+++ b/frontend/src/container/TracesExplorer/ListView/utils.tsx
@@ -62,7 +62,6 @@ export const getListColumns = (
 			},
 			sorter: true,
 			sortOrder: (orderByFromQuery || [])?.map((order) => {
-				console.log(order);
 				if (order.columnName === 'timestamp') {
 					return order.order === 'asc' ? 'ascend' : 'descend';
 				}
@@ -116,7 +115,6 @@ export const getListColumns = (
 			sortOrder:
 				key === 'durationNano'
 					? (orderByFromQuery || [])?.map((order) => {
-							console.log(order);
 							if (order.columnName === 'durationNano') {
 								return order.order === 'asc' ? 'ascend' : 'descend';
 							}
--- a/frontend/src/container/TracesExplorer/ListView/utils.tsx
+++ b/frontend/src/container/TracesExplorer/ListView/utils.tsx
@@ -58,6 +58,7 @@ export const getListColumns = (
 					</BlockLink>
 				);
 			},
+			sorter: true,
 		},
 	];
 
@@ -101,6 +102,10 @@ export const getListColumns = (
 				);
 			},
 			responsive: ['md'],
+			sorter:
+				key === 'durationNano'
+					? (a, b): number => Number(a.durationNano) - Number(b.durationNano)
+					: undefined,
 		})) || [];
 
 	return [...initialColumns, ...columns];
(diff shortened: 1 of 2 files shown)


vikrantgupta25 (dismissed):

YounixM (dismissed):

srikanthccv:
@SagarRajput-7 We shouldn't implement local sorting for this table since it's paginated, and the data is fetched through API requests. Local sorting only makes sense when the client already has the entire dataset. To illustrate, imagine a service has 100k spans in the last 15 minutes. Since we can't display them all, we use pagination through the list API. If users want to see the slowest span, they're looking for the slowest span in the entire last 15 minutes, not just the slowest span among the 200 spans on the current page. This user perspective helps in ensuring you avoid spending time on doing something that doesn't satisfy what users wants.

SagarRajput-7:
> @SagarRajput-7 We shouldn't implement local sorting for this table since it's paginated, and the data is fetched through API requests. Local sorting only makes sense when the client already has the entire dataset. To illustrate, imagine a service has 100k spans in the last 15 minutes. Since we can't display them all, we use pagination through the list API. If users want to see the slowest span, they're looking for the slowest span in the entire last 15 minutes, not just the slowest span among the 200 spans on the current page. This user perspective helps in ensuring you avoid spending time on doing something that doesn't satisfy what users wants.

done - 

https://github.com/SigNoz/signoz/assets/162284829/862f815f-445b-445e-b5af-eb6ff794fb4e

frontend/src/container/TracesExplorer/ListView/utils.tsx
```diff
@@ -38,10 +40,19 @@ export const getListColumns = (
 						: dayjs(item / 1e6).format('YYYY-MM-DD HH:mm:ss.SSS');
 				return <Typography.Text>{date}</Typography.Text>;
 			},
+			sorter: true,
+			sortOrder: (orderByFromQuery || [])?.map((order) => {
+				console.log(order);
```
vikrantgupta25, line 45: console.log

frontend/src/container/TracesExplorer/ListView/utils.tsx
```diff
@@ -67,6 +78,17 @@ export const getListColumns = (
 				return <Typography data-testid={key}>{value}</Typography>;
 			},
 			responsive: ['md'],
+			sorter: key === 'durationNano' ? true : undefined,
+			sortOrder:
+				key === 'durationNano'
+					? (orderByFromQuery || [])?.map((order) => {
+							console.log(order);
```
vikrantgupta25, line 85: console.log

vikrantgupta25:
https://github.com/SigNoz/signoz/assets/54737045/0a990653-3b28-455f-8e32-7b4b1fbe702b

when sorted by `timestamp asc` and then applying `duration desc` it should appear after the timestamp as the intent is to sort duration after sorting by timestamps.and we should respect the order of the same while updating as well not move everything to the front

vikrantgupta25:
also when doing multiple selections the state of the sorter is not maintained

SagarRajput-7:
@srikanthccv - as per our discussion I removed the multi-column sorting, we will take it separately, Thanks!


https://github.com/user-attachments/assets/4b2bbb71-f32c-441e-93fe-a574e9f00d45

frontend/src/container/TracesExplorer/ListView/utils.tsx
SagarRajput-7: done

frontend/src/container/TracesExplorer/ListView/utils.tsx
SagarRajput-7: done

vikrantgupta25:
adding `timestamp asc` and `duration nano asc` in the UI manually from the order by bar and then clicking on any of the sorters in the table clears both and adds one is this expected? Read the above comment regarding the multiple orderBy's but feel we should update whatever we can in this case rather than removing all of them.

vikrantgupta25:
if there are other order by present they get removed as well. this doesn't seem ideal. IMO just manipulate whatever key you are adding / changing, and leave the remaining as is. and the placement would be wherever it is present or if not then at the end.

PR #5274, opened 2024-06-18
fix: handled value as string from queryParams for trace filters

SagarRajput-7:
### Summary

Here the serviceName value coming as a string from queryParam and as spreading and includes works on string also, which was splitting the string into individual chars and indirectly making the value as an array of string with chars in it. Also, this was breaking the page due to missing null check.

#### Screenshots

https://github.com/SigNoz/signoz/assets/162284829/880f2cef-a93e-4b50-a8c8-890f6758339c

### DIFF
--- a/frontend/src/pages/TracesExplorer/Filter/SectionContent.tsx
+++ b/frontend/src/pages/TracesExplorer/Filter/SectionContent.tsx
@@ -4,7 +4,7 @@ import { Button, Card, Checkbox, Input, Tooltip } from 'antd';
 import { CheckboxChangeEvent } from 'antd/es/checkbox';
 import { ParaGraph } from 'container/Trace/Filters/Panel/PanelBody/Common/styles';
 import useDebouncedFn from 'hooks/useDebouncedFunction';
-import { isEmpty } from 'lodash-es';
+import { isArray, isEmpty } from 'lodash-es';
 import {
 	ChangeEvent,
 	Dispatch,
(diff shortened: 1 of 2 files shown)


frontend/src/pages/TracesExplorer/Filter/SectionContent.tsx
```diff
@@ -92,17 +92,21 @@ export function SectionBody(props: SectionBodyProps): JSX.Element {
 		if (checked) {
 			addFilter(type, newValue, setSelectedFilters, keys);
 			setCheckedItems((prev) => {
-				if (!prev.includes(newValue)) {
-					prev.push(newValue);
+				const arr = prev || [];
+				if (!arr.includes(newValue)) {
```
vikrantgupta25, line 96: string can still go inside the if block ?

vikrantgupta25:
>  isArray logic treats the string also as an array

This is not correct. https://lodash.com/docs/4.17.15#isArray

frontend/src/pages/TracesExplorer/Filter/SectionContent.tsx
SagarRajput-7: done, added isArray check

SagarRajput-7:
> > isArray logic treats the string also as an array
> 
> This is not correct. https://lodash.com/docs/4.17.15#isArray

yeah.
Here - as spreading and includes works on string also, which was splitting the string into individual chars and indirectly making the value as an array of string with chars in it

vikrantgupta25 (approved):

YounixM (approved):

PR #5276, opened 2024-06-18
feat: pass fill gaps to query range api

YounixM:
### Summary

Fixes: [#1266](https://github.com/SigNoz/engineering-pod/issues/1266)

Send `fillGaps: true` in the compositeQuery object. 

https://www.loom.com/share/3f774d65a8f145bdb960c843308d3761?sid=237ac498-4cbe-40e7-aa78-0ad0c11ca82e

### DIFF
--- a/frontend/src/container/GridCardLayout/GridCard/index.tsx
+++ b/frontend/src/container/GridCardLayout/GridCard/index.tsx
@@ -108,6 +108,7 @@ function GridCardGraph({
 				query: updatedQuery,
 				globalSelectedInterval,
 				variables: getDashboardVariables(variables),
+				fillGaps: widget.fillSpans,
 			};
 		}
 		updatedQuery.builder.queryData[0].pageSize = 10;
@@ -122,6 +123,7 @@ function GridCardGraph({
 					limit: updatedQuery.builder.queryData[0].limit || 0,
 				},
 			},
+			fillGaps: widget.fillSpans,
 		};
 	});
 
@@ -134,6 +136,8 @@ function GridCardGraph({
 		}
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [updatedQuery]);
+
+	console.log('requestData', requestData);
 
 	const queryResponse = useGetQueryRange(
 		{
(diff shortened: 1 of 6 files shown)


frontend/src/container/GridCardLayout/GridCard/index.tsx
```diff
@@ -135,6 +137,8 @@ function GridCardGraph({
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [updatedQuery]);
 
+	console.log('requestData', requestData);
```
vikrantgupta25, line 140: console.log

vikrantgupta25:
we should remove the frontend logic for fill gaps here as well ?

srikanthccv:
LGTM, please address the comments.

YounixM:
The existing fillSpans change wouldn't have any effect as the data would be as expected. It works as a fallback if the data is missing.

YounixM:
Screenshot 2024-06-20 at 11 23 46 AM

This is the only handling we have in UI for fillGaps. As there wouldn't be any missing timestamps, unless there is error from BE, this will not kick in. This will act as an fallback if that happens.

vikrantgupta25 (dismissed):

SagarRajput-7 (dismissed):

srikanthccv:
> As there wouldn't be any missing timestamps, unless there is error from BE, this will not kick in.

If there are regular queries and formula queries combined with different aggregation intervals (we support this now), then this logic would probably fail because the timestamp from the formula would be different from the simple query. This is why we moved the logic to query-service. We should remove the logic on the frontend to avoid such an issue.

vikrantgupta25:
>  timestamp from the formula would be different from the simple query.

just curious what calculation we do in this case ? can you point me to the change ?

srikanthccv:
We thought of using the "ideal" solution but then changed back to LCM of all the queries involved in formula expression https://github.com/SigNoz/signoz/blob/64e06ab3f9e675fbc130bac4edc0af447ee5f659/pkg/query-service/postprocess/gaps.go

YounixM:
Addressed. Removed the logic of fillGaps in frontend.

srikanthccv (approved):

Issue #5135, opened 2024-06-03
LOGS[FE]: Going back to list view doesn't remove group by


nityanandagohain:
Going back to list view doesn't remove group by

https://github.com/SigNoz/signoz/assets/26831659/f4400132-d45d-450b-a7ba-9cc419e552e4

srikanthccv:
This should be fixed. I ran into this and it was not clear what the issue is unless I open dev tools. Even with dev tools, one won't be able to figure out without the knowledge of aggregate operators in the query builder. We should fix this bug because seeing the count and switching to a list is a common task.

YounixM:
@nityanandagohain @srikanthccv : As discussed, assigning it back to you as it will be handled in BE.

PR #5277, opened 2024-06-18
fix: remove group by when user switches from qb view to search view

YounixM:
### Summary

Fixes: #5135


vikrantgupta25:
we should remove the frontend logic for fill gaps here as well ?

PR #5279, opened 2024-06-18
fix: update from typography link to react router dom link component t…

YounixM:
…o maintain global state

### Summary

Current Behaviour 

Trace Explorer -> Choose Last 1 Week -> Trace View -> Click on some old trace ID -> Select Span -> Go to related Logs -> Time frame resets to last 15 mins and the results are not return

As we were using AntD Typography Link component to render Link and handle the navigation, the page would re-render and the global state defaults would be set, hence the time frame would reset to last 15 mins when the user clicks on the go to related logs

Using Link from `react-router-dom` handles navigation without render and as a result, the select time frame is maintain across the navigation.

### DIFF
--- a/frontend/src/container/TraceDetail/index.tsx
+++ b/frontend/src/container/TraceDetail/index.tsx
@@ -246,13 +246,14 @@ function TraceDetail({ response }: TraceDetailProps): JSX.Element {
 
 			<Sider
 				className={cx('span-details-sider', isDarkMode ? 'dark' : 'light')}
-				style={{ background: isDarkMode ? '#000' : '#fff' }}
+				style={{ background: isDarkMode ? '#0b0c0e' : '#fff' }}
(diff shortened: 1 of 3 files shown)


vikrantgupta25 (approved):

frontend/src/container/TraceDetail/index.tsx
```diff
@@ -246,13 +246,14 @@ function TraceDetail({ response }: TraceDetailProps): JSX.Element {
 
 			<Sider
 				className={cx('span-details-sider', isDarkMode ? 'dark' : 'light')}
-				style={{ background: isDarkMode ? '#000' : '#fff' }}
+				style={{ background: isDarkMode ? '#0b0c0e' : '#fff' }}
```
SagarRajput-7, line 249: NIT: can use --bg-ink-500 instead

SagarRajput-7 (approved):

Issue #5267, opened 2024-06-18
Dependant variables do not get updated correctly.


srikanthccv:
## Bug description

Try playing around with this dashboard https://stagingapp.signoz.io/dashboard/9c1fa385-dea8-4880-878b-39412d4ae466?relativeTime=30m. I have mentioned this behaviour here https://signoz-team.slack.com/archives/C02TJ466H8U/p1718178568819059?thread_ts=1718175690.454299&cid=C02TJ466H8U

PR #5283, opened 2024-06-19
fix: dependent variable panel not updating

vikrantgupta25:
### Summary

- dependent variables were updating but the queue was not getting cleared because of similar prev and next options.
- cleared the variables queue when the normal update flow isn't triggered

#### Related Issues / PR's

fixes https://github.com/SigNoz/signoz/issues/5267

#### Screenshots

NA

Tested here - https://stagingapp.signoz.io/dashboard/9c1fa385-dea8-4880-878b-39412d4ae466?relativeTime=30m


srikanthccv (approved):
Can we tests for this? I think this like 5th time since we fix something related to dependant variable since the beginning of dashboard variables support in product. One has to manually verify if things are working fine. This will remain the important feature of dashboard so we would save a lot of time if this can covered with tests.

SagarRajput-7 (approved):

YounixM (approved):

PR #5286, opened 2024-06-19
Feat: Instrumentation scope support

### LINKED ISSUE #2705: [Logs] Add instrumention scope to logs model
instrumentation scope is missing from logs model.

nityanandagohain:
Fixes #2705 

* attribute suggestion
* attribute querying
* version and name suggestion and querying
----
No support for indexing these attributes.


srikanthccv:
Does it require any frontend changes?

nityanandagohain:
yes @srikanthccv , right now the attributes are not displayed correctly on the UI

nityanandagohain:
@srikanthccv please review this ? once reviewed with merge the otel collector PR, and then will ask the frontend changes to be done

srikanthccv:
I was assuming instrumentation scope collector PR gets merged first and then we will release this later after enough back filling. I will review it.

srikanthccv:
Can you add a test that shows how to query just by the scope name? I can see how it happens for the attributes but it's not clear about the name and version.

Can we rename `instrumentation_scope_attributes` to `scope`? This will make it easier to write ClickHouse queries and doesn't make the queries really long.

nityanandagohain:
Yeah can do that, I guess it will make it a bit more cleaner 

so you are suggesting to keep
`instrumentation_scope`, `instrumentation_scope_version`, `scope`

where `scope` will be the attributes. Does this look good ?

nityanandagohain:
Have made the changes @srikanthccv  ,  related collector PR https://github.com/SigNoz/signoz-otel-collector/pull/343

pkg/query-service/model/v3/v3.go
```diff
@@ -229,13 +229,14 @@ type AggregateAttributeRequest struct {
 type TagType string
 
 const (
-	TagTypeTag      TagType = "tag"
-	TagTypeResource TagType = "resource"
+	TagTypeTag                  TagType = "tag"
+	TagTypeResource             TagType = "resource"
+	TagTypeInstrumentationScope TagType = "instrumentation_scope"
```
srikanthccv, line 234: nit: this is not used anywhere.

srikanthccv:
Thanks for the test. Two questions, 1. What does this support mean for log pipelines 2. Is there no enrichment required here https://github.com/SigNoz/signoz/blob/1b0ec8ac4332b49d12d20099fd986946293f58c6/pkg/query-service/app/http_handler.go#L2940-L2972?

pkg/query-service/model/v3/v3.go
nityanandagohain: Ahh my bad it should have been `scope`, have updated it

it will allow these queries
```
{
    "id": "3d0fc1fc-08ba-4cb9-ab98-c38cb1dfb460",
    "key": {
        "key": "scope_attr",
        "dataType": "string",
        "type": "scope",
        "isColumn": false,
        "isJSON": false
    },
    "op": "in",
    "value": "attr_val"
}
```

nityanandagohain:
* Pipelines will have to add support for this, since we have changed the naming convention we will have to take care of it in pipelines, we can address it separetely.
* Enrichment is not required for instrumentation scope attributes because we anyway don't have support for `materialized` column so we are skipping them https://github.com/SigNoz/signoz/blob/9ab3a03bfe0073c0ae4162df59993aa08a73b2cb/pkg/query-service/app/logs/v3/enrich_query.go#L66 . For scope_name and scope_version it is done through static check https://github.com/SigNoz/signoz/blob/9ab3a03bfe0073c0ae4162df59993aa08a73b2cb/pkg/query-service/app/logs/v3/enrich_query.go#L142

nityanandagohain:
Closing in favour of https://github.com/SigNoz/signoz/pull/6339

PR #5295, opened 2024-06-20
feat: add support for multiple aggregations in clickhouse queries

vikrantgupta25:
#### Related Issues / PR's

fixes https://github.com/SigNoz/signoz/issues/5153

#### Screenshots

NA


vikrantgupta25:
As discussed this will also be returned in the updated response from the backend.

srikanthccv:
@vikrantgupta25 I was going to test this but realized this is closed. The format we discussed was for table panel type, right? Please share the format that works best for you for supporting multi-aggregations. I will make the appropriate backend changes.

vikrantgupta25:
@srikanthccv yes i closed this in favour of the table panel type changes being moved to backend. for the time series panel type, i will create a separate PR we can move that incrementally

Issue #5303, opened 2024-06-21
respect orderBy for clickhouse queries


vikrantgupta25:
- orderBy clause not respected while ordering the table panel type for clickhouse queries

PR #5308, opened 2024-06-21
fix: table order by with builder queries

srikanthccv:
### Summary

Fixes https://github.com/SigNoz/signoz/issues/5303


vikrantgupta25:
@srikanthccv  Looks good to me 👍  , Just a minor question :- 

```
-- will this only work in case of `queryType` === `QueryTypeBuilder` ? 

-- are we not respecting the labels in the backend ? 
```

srikanthccv:
> will this only work in case of `queryType` === `QueryTypeBuilder` ?

Yes, I want to limit the scope of this to query builder. For ClickHouse queries, the result shouldn't need to be manipulated. It can be rendered as is. Would it make it easy for you if the response it same for all query types?

> are we not respecting the legends in the backend ?  ( makes sense to not do it as can be done on frontend, just want to verify once so can take care at frontend )

As you know, we only support one aggregation in ClickHouse queries as of now. So the legend on the front end needs to be respected.

vikrantgupta25:
> Would it make it easy for you if the response it same for all query types?

Yes it would be better to have standard response set. else we ll end up in custom handlings again based on queryTypes

pkg/query-service/postprocess/table.go
```diff
@@ -0,0 +1,299 @@
...
+		return []*v3.Result{}
+	}
+
+	// Sort results by QueryName
+	sort.Slice(results, func(i, j int) bool {
+		return results[i].QueryName < results[j].QueryName
+	})
+
+	// Create a map to store all unique labels
+	seen := make(map[string]struct{})
```
vikrantgupta25, line 39: why not `map[string]bool`

pkg/query-service/postprocess/table.go
```diff
@@ -0,0 +1,299 @@
...
+	// Why don't we have a column for each query name?
+	// Because we don't know if the query is an aggregation query or a non-aggregation query
+	// So we create a column for each query name that has at least one point
+	columns := make([]*v3.TableColumn, 0)
+	for _, key := range labelKeys {
+		columns = append(columns, &v3.TableColumn{Name: key})
+	}
+	for _, result := range results {
+		if len(result.Series) > 0 && len(result.Series[0].Points) > 0 {
+			columns = append(columns, &v3.TableColumn{Name: result.QueryName})
```
vikrantgupta25, line 255: if there are multiple result columns for the same query then we would be adding duplicates here ?

pkg/query-service/postprocess/table.go
srikanthccv: https://medium.easyread.co/golang-series-empty-struct-ed317e6d8600

pkg/query-service/postprocess/table.go
srikanthccv: It's not possible to have multiple result columns in the same query as of now. There can only be one result column. Although this will be changed soon but it will be separately addressed.

pkg/query-service/postprocess/table.go
vikrantgupta25: nice read 👍 , can we add the same as an inline comment or the reason for doing the same inline as might be confusing for someone else seeing this later

vikrantgupta25 (approved):
LGTM

Issue #5273, opened 2024-06-18
Send alert description with missing data notification


srikanthccv:
The current missing data alert doesn't send anything except the summary "No data".

PR #5315, opened 2024-06-22
fix: send alert default annotations for missing data alert

srikanthccv:
### Summary

Fixes https://github.com/SigNoz/signoz/issues/5273

We currently replace the description label with the "Missing data" message. This is not desired because people put runbooks in the description. The runbook is intended for both missing data and actual alerts. This change removes the description replacement when there is no data.


pkg/query-service/rules/threshold_rule.go
```diff
@@ -942,13 +965,11 @@ func (r *ThresholdRule) Eval(ctx context.Context, ts time.Time, queriers *Querie
...
-			if smpl.IsMissing {
-				if a.Name == labels.AlertDescriptionLabel || a.Name == labels.AlertSummaryLabel {
-					a.Value = labels.AlertMissingData
-				}
-			}
 			annotations = append(annotations, labels.Label{Name: normalizeLabelName(a.Name), Value: expand(a.Value)})
 		}
+		if smpl.IsMissing {
+			lb.Set(labels.AlertNameLabel, "[No data] "+r.Name())
+		}
```
srikanthccv, line 972: This is the main change. The logic that substitutes the description/summary with missing data is removed and add `[No data]` to notification.

nityanandagohain (approved):

Issue #5049, opened 2024-05-22
When you click on "Related logs", the time filter always seems to be the last 15 minutes


YounixM:
Issue: When you click on "Related logs", the time filter always seems to be the last 15 minutes

YounixM:
This is working as expected.

YounixM:
Update: Got on call with user and was able to reproduce the issue in traces view.

YounixM:
We now respect the previously selected time range from the explorer page.

Need to improve the UX around this.  The start time and end time should be dependent on the trace details.

Will improve once we start getting these details from the API.

https://github.com/SigNoz/engineering-pod/issues/1406

srikanthccv:
reopening the issue since the issue is not completely fixed as the previously selected time range could still be incorrect.

srikanthccv:
@YounixM since Vishal has bandwidth constraints I picked this up and added the start and end time of the trace in miliis in response. Please use that information when redirecting users.

YounixM:
Thanks @srikanthccv

PR #5321, opened 2024-06-24
chore: add start/end millis for trace details response

srikanthccv:
### Summary

Part of https://github.com/SigNoz/signoz/issues/5049

Since the `traceID` is part of the path var, I assume it will always be there and the result `searchSpansResult` will contain exactly one item as initialized here https://github.com/SigNoz/signoz/blob/a7a160df76509ee1bd0ccac5fe598a9a7772a9cc/pkg/query-service/app/clickhouseReader/reader.go#L1943 and here https://github.com/SigNoz/signoz/blob/a7a160df76509ee1bd0ccac5fe598a9a7772a9cc/ee/query-service/app/db/trace.go#L120

We go through all the items scanned into `searchScanResponses` and send the `min(timestamp) - durationNano` as start and `max(timestamp) + durationNano` as end. The start and end are not exact but they serve the purpose.

### DIFF
--- a/pkg/query-service/app/clickhouseReader/reader.go
+++ b/pkg/query-service/app/clickhouseReader/reader.go
@@ -1924,6 +1924,7 @@ func (r *ClickHouseReader) SearchTraces(ctx context.Context, params *model.Searc
 		telemetry.GetInstance().SendEvent(telemetry.TELEMETRY_EVENT_TRACE_DETAIL_API, data, userEmail, true, false)
 	}
 
+	var startTime, endTime, durationNano uint64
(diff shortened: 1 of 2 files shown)


pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -1983,6 +1993,9 @@ func (r *ClickHouseReader) SearchTraces(ctx context.Context, params *model.Searc
 		}
 	}
 
+	searchSpansResult[0].StartTimestampMillis = startTime - durationNano
```
makeavish, line 1996: @srikanthccv Here we should have converted durationNano to milliseconds before subtracting or adding

PR #5326, opened 2024-06-24
feat: go to traces should use start time and end time from trace details

YounixM:
### Summary

feat: go to traces should use start time and end time from trace details

### DIFF
--- a/frontend/src/container/TraceDetail/SelectedSpanDetails/index.tsx
+++ b/frontend/src/container/TraceDetail/SelectedSpanDetails/index.tsx
@@ -19,11 +19,16 @@ import { CardContainer, CustomSubText, styles } from './styles';
 import Tags from './Tags';
 
 function SelectedSpanDetails(props: SelectedSpanDetailsProps): JSX.Element {
-	const { tree, firstSpanStartTime } = props;
-
 	const { maxTime, minTime } = useSelector<AppState, GlobalReducer>(
 		(state) => state.globalTime,
 	);
+
+	const {
+		tree,
+		firstSpanStartTime,
+		traceStartTime = minTime,
+		traceEndTime = maxTime,
+	} = props;
 
 	const { id: traceId } = useParams<Params>();
 
@@ -74,7 +79,7 @@ function SelectedSpanDetails(props: SelectedSpanDetailsProps): JSX.Element {
 	];
 
 	const onLogsHandler = (): void => {
-		const query = getTraceToLogsQuery(traceId, minTime, maxTime);
+		const query = getTraceToLogsQuery(traceId, traceStartTime, traceEndTime);
 
 		history.push(
 			`${ROUTES.LOGS_EXPLORER}?${createQueryParams({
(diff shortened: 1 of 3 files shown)


frontend/src/container/TraceDetail/index.tsx
```diff
@@ -131,6 +137,8 @@ function TraceDetail({ response }: TraceDetailProps): JSX.Element {
 
 	const isDarkMode = useIsDarkMode();
 
+	console.log('response', response);
```
makeavish, line 140: Do we need to log response?

makeavish (approved):

frontend/src/container/TraceDetail/SelectedSpanDetails/index.tsx
```diff
@@ -19,12 +19,17 @@ import { CardContainer, CustomSubText, styles } from './styles';
...
-	const { tree, firstSpanStartTime } = props;
-
 	const { maxTime, minTime } = useSelector<AppState, GlobalReducer>(
 		(state) => state.globalTime,
 	);
 
+	const {
+		tree,
+		firstSpanStartTime,
+		traceStartTime = minTime,
```
vikrantgupta25, line 29: minTime and maxTime from the global time selector will be overridden here.

frontend/src/container/TraceDetail/index.tsx
```diff
@@ -48,6 +48,12 @@ function TraceDetail({ response }: TraceDetailProps): JSX.Element {
 		[response],
 	);
 
+	const traceStartTime = useMemo(() => response[0].startTimestampMillis, [
```
vikrantgupta25, line 51: will `startTimestampMillis` be always present in the response ?

makeavish: Yes

frontend/src/container/TraceDetail/index.tsx
```diff
@@ -48,6 +48,12 @@ function TraceDetail({ response }: TraceDetailProps): JSX.Element {
...
+		response,
+	]);
+
+	const traceEndTime = useMemo(() => response[0].endTimestampMillis, [response]);
```
vikrantgupta25, line 55: same here ? do we need null checks for the same ?

frontend/src/container/TraceDetail/index.tsx
vikrantgupta25: and can the value of `startTimestampMillis` be ever undefined ?

YounixM: We have the default values if these values don't exist.

frontend/src/container/TraceDetail/index.tsx
YounixM: We have the default values (minTime, maxTime) if these values don't exist.

frontend/src/container/TraceDetail/index.tsx
vikrantgupta25: yes if it can never be `undefined` we should make these mandatory props.

vikrantgupta25: gives us tighter type safety than optional props. can be easy to miss when reusing this component somewhere else.

makeavish: No, it's not possible unless backend returns 500.

YounixM: It's fine. For some reason if this fails, we have a fallback to old behavior

vikrantgupta25: we can remove the optional params and it will still fallback to the older behaviour

PR #5331, opened 2024-06-24
fix: make the license key check case insensitive

vikrantgupta25:
#### Screenshots

NA

### DIFF
--- a/frontend/src/container/Header/index.tsx
+++ b/frontend/src/container/Header/index.tsx
@@ -27,6 +27,7 @@ import {
 import { useSelector } from 'react-redux';
 import { NavLink } from 'react-router-dom';
 import { AppState } from 'store/reducers';
+import { License } from 'types/api/licenses/def';
 import AppReducer from 'types/reducer/app';
 import { getFormattedDate, getRemainingDays } from 'utils/timeUtils';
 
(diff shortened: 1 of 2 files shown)


frontend/src/container/Header/index.tsx
```diff
@@ -109,9 +110,13 @@ function HeaderContainer(): JSX.Element {
 
 	const { data: licenseData, isFetching, status: licenseStatus } = useLicense();
 
+	const licensesStatus: string =
+		licenseData?.payload?.licenses?.find((e: License) => e.isCurrent)?.status ||
+		'';
+
 	const isLicenseActive =
-		licenseData?.payload?.licenses?.find((e) => e.isCurrent)?.status ===
-		LICENSE_PLAN_STATUS.VALID;
```
grandwizard28, line 114: Can we do something like:

`status.toLocalLowerCase() === LICENSE_PLAN_STATUS.VALID.toLocaleLowerCase()` just to ensure that a case insensitive check is happening?

grandwizard28 (changes requested):

frontend/src/container/Header/index.tsx
vikrantgupta25: we are doing that in the next line ?

grandwizard28: That was my point, my apologies, I didn't read it correctly.

grandwizard28 (approved):

SagarRajput-7 (approved):

srikanthccv:
Next time for new API response payloads do not accept such values where you would have to derive something. Demand the API itself to return the derived result.

vikrantgupta25:
yes did this as a one time thing as there was an incident involved cc :- @grandwizard28

YounixM (approved):

PR #5353, opened 2024-06-26
fix: search not working in services table

vikrantgupta25:
### Summary

- fix the search not working in services table , the return for the `onFilter` method was missing 

#### Screenshots

NA


YounixM (approved):

srikanthccv:
What would it take to write a test for this?

vikrantgupta25:
> What would it take to write a test for this?

let me add the same in an follow up PR, currently we do not have very lean wrappers which make these quick. @SagarRajput-7 let's get the wrapper out soon ?

PR #5365, opened 2024-06-27
fix: added right margin to facing issues btn on dashboad detail page

SagarRajput-7:
#### Screenshots

Before: 
--------
image

After: 
--------
Screenshot 2024-06-27 at 12 40 20 PM

### DIFF
--- a/frontend/src/container/NewDashboard/DashboardDescription/Description.styles.scss
+++ b/frontend/src/container/NewDashboard/DashboardDescription/Description.styles.scss
@@ -58,6 +58,7 @@
 		display: flex;
 		justify-content: space-between;
 		align-items: center;
+		padding-right: 16px;
 
 		.dashboard-breadcrumbs {
 			height: 48px;
--- a/frontend/src/container/NewDashboard/DashboardDescription/Description.styles.scss
+++ b/frontend/src/container/NewDashboard/DashboardDescription/Description.styles.scss
@@ -58,7 +58,7 @@
 		display: flex;
 		justify-content: space-between;
 		align-items: center;
-		padding-right: 16px;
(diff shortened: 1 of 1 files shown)


frontend/src/container/NewDashboard/DashboardDescription/Description.styles.scss
```diff
@@ -58,6 +58,7 @@
 		display: flex;
 		justify-content: space-between;
 		align-items: center;
+		padding-right: 16px;
```
vikrantgupta25, line 61: should we add padding or margin?

SagarRajput-7: I used padding as I didn't see the use of pushing the borders of the header inwards using margin, with padding the header's border will remain there and we will add space inside it.
also, the content below follows the same.

image

image

vikrantgupta25: didn't get you how is this causing any issues ? 
image

SagarRajput-7: Neither of them won't cause any issues, but I got your point about using margin to properly position the containers inside the containers.

vikrantgupta25 (approved):

frontend/src/container/NewDashboard/DashboardDescription/Description.styles.scss
vikrantgupta25: https://mailchimp.com/resources/padding-vs-margin/ for more details

YounixM (approved):

PR #5371, opened 2024-06-27
fix: added logic to keep sections uncollapsed for all filtered items

SagarRajput-7:
#### Related Issues / PR's

https://github.com/SigNoz/engineering-pod/issues/1422

#### Screenshots

https://github.com/SigNoz/signoz/assets/162284829/aae4b882-0531-4945-b8ec-19a71ddad86c

### DIFF
--- a/frontend/src/pages/TracesExplorer/Filter/Section.tsx
+++ b/frontend/src/pages/TracesExplorer/Filter/Section.tsx
@@ -1,7 +1,14 @@
 import './Filter.styles.scss';
 
 import { Button, Collapse, Divider } from 'antd';
-import { Dispatch, MouseEvent, SetStateAction } from 'react';
+import {
+	Dispatch,
+	MouseEvent,
+	SetStateAction,
+	useEffect,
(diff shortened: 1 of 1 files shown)


SagarRajput-7:
Added test case for this here - https://github.com/SigNoz/signoz/pull/5451

specifically - `should render the trace filter with the given query`

YounixM:
@SagarRajput-7 : Where are we stored the opened sections?

SagarRajput-7:
> @SagarRajput-7 : Where are we stored the opened sections?

We don't store the opened sections (except the default one `'hasError', 'durationNano', 'serviceName'`).  We are in sync with compositeQuery and rely on it for the section's state. So if a filter is added from a particular section then for better usability we will be keeping that section open by default.

YounixM:
Does it maintain the state on hard refresh?

SagarRajput-7:
> Does it maintain the state on hard refresh?

Yes

YounixM (approved):

Issue #5395, opened 2024-07-01
Resolve the cyclic dependency issues caused by telemetry


srikanthccv:
The current set of import dependencies makes it hard to write tests.

srikanthccv:
Should no longer be an issue with stats reporter.

PR #5396, opened 2024-07-01
chore: remove rules dependency in CH reader

srikanthccv:
### Summary

Part of https://github.com/SigNoz/signoz/issues/5395


srikanthccv:
The plan is to move the `/rules`, `/dashboards`, and `/explorer` from the reader. It it's current state the reader can't be instantiated in rules (for its own use or tests) because that would lead to cyclic dependency.

nityanandagohain (approved):

Issue #5323, opened 2024-06-24
improve user login experience when access updated


prashant-shahi:
## Summary

When access is updated for a user in the project, the updated user has to logout and log back in to reflect the change in access.

## How to reproduce

1. Login with Viewer-level user
2. In another tab login with admin user, and update the access of viewer-level user to Editor/Admin
3. Switch back to viewer-level user tab, and use the UI

## Additional context

NA

KJ1010G:
Can I work on the backend part of this issue ?

KJ1010G:
I went through the codebase regarding this issue and realized that this is not just a user experience issue, its also a security issue.
Please assign me this issue, i want to work on this.
@prashant-shahi ?

prashant-shahi:
> I went through the codebase regarding this issue and realized that this is not just a user experience issue, its also a security issue. Please assign me this issue, i want to work on this. @prashant-shahi ?

cc @YounixM

prashant-shahi:
@KJ1010G Assigned

KJ1010G:
Thanks!
I opened one pull request which I believe solves part of the problem.
More work needs to be done but that work relies on the availability of cache in signoz deployments.

`var c cache.Cache
if serverOptions.CacheConfigPath != "" {
cacheOpts, err := cache.LoadFromYAMLCacheConfigFile(serverOptions.CacheConfigPath)
if err != nil {
	return nil, err
}
c = cache.NewCache(cacheOpts)
}`
Above is code from queryservice/app/server.go

Is this CacheConfigPath always set in deployments ?

KJ1010G:
The way I am thinking of solving this issue is...

1. Cross check refresh token based logins (PR above does that)
2. On edit role and delete user api calls, create a cache entry with key being userID and value being the new group, or 'deleted'. The TTL of this cache entry being the 30mins (access token validity time). In case an admin is deleted or demoted, also revoke all their PATs if any.
3. On every request, when we fetch user details from access tokens, we check the cache if the key of that ID is present. If present, and the group from access token does not match cache value, we send the user ErrorTokenExpired response so the frontend tries to login again using refresh tokens and gets updated tokens.
4. On the login response add a new field called "oldRole". If it not empty, it means access has been recently updated. ( Code for this is also included in PR above). Frontend can use this to inform user of the change and do whatever else necessary to make for a good user experience.

[...]

PR #5402, opened 2024-07-02
fix: cross check refresh token's group id before generating new tokens

KJ1010G:
### Summary

Previously the groupID of refresh token was used to create new access tokens. Which means even if some user gets demoted from a higher privilege, they can continue to operate at that same privilage forever by just refreshing tokens using the refresh tokens.
Also a deleted user could still use old refresh tokens for continued access.

This commit fixes that. Now login using refresh token also cross checks database.

#### Related Issues / PR's

https://github.com/SigNoz/signoz/issues/5323

#### Screenshots

NA

NA


KJ1010G:
cleaned up commit, this one now just includes cross checking of refresh token

PR #5193, opened 2024-06-11
feat: show warning when the top level operations count is more than 1500

YounixM:
Screenshot 2024-06-12 at 1 11 11 AM

PageSize is set to 100 in code.

PR #5404, opened 2024-07-02
chore: only fetch top level operation from the selected time window

srikanthccv:
### Summary

Improvement on top of https://github.com/SigNoz/signoz/pull/5193


pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -706,47 +706,43 @@ func (r *ClickHouseReader) GetServicesList(ctx context.Context) (*[]string, erro
...
 		if err := rows.Scan(&name, &serviceName, &t); err != nil {
-			return nil, nil, &model.ApiError{Typ: model.ErrorInternal, Err: fmt.Errorf("error in reading data")}
+			return nil, &model.ApiError{Typ: model.ErrorInternal, Err: fmt.Errorf("error in reading data")}
 		}
 		if _, ok := operations[serviceName]; !ok {
-			operations[serviceName] = []string{}
-		}
-		if _, ok := allOperations[serviceName]; !ok {
-			allOperations[serviceName] = []string{}
+			operations[serviceName] = []string{"overflow_operation"}
```
vikrantgupta25, line 738: just want to understand what does this signify ?

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -706,47 +706,43 @@ func (r *ClickHouseReader) GetServicesList(ctx context.Context) (*[]string, erro
...
 
 	start = start.In(time.UTC)
 
 	// The `top_level_operations` that have `time` >= start
 	operations := map[string][]string{}
-	// All top level operations for a service
-	allOperations := map[string][]string{}
-	query := fmt.Sprintf(`SELECT DISTINCT name, serviceName, time FROM %s.%s`, r.TraceDB, r.topLevelOperationsTable)
+	// We can't use the `end` because the `top_level_operations` table has the most recent instances of the operations
+	// We can only use the `start` time to filter the operations
```
vikrantgupta25, line 716: any reason to accept and pass the `end` here if we are not utilising the same ?

vikrantgupta25 (approved):

pkg/query-service/app/clickhouseReader/reader.go
srikanthccv: We replace the regular operation with this here https://github.com/SigNoz/signoz-otel-collector/blob/d3ed24f34f1e321b8dbd2defcb2311af65467fed/processor/signozspanmetricsprocessor/processor.go#L58 when there are more than 2048 operations to track. This is a special operation that doesn't exist in DB so adding here.

pkg/query-service/app/clickhouseReader/reader.go
srikanthccv: to keep the interface clean with both start, end and change the table eventually if possible.

Issue #4700, opened 2024-03-14
Add Telegram Bot to the list of alert channels


AliArtukov:
Add Telegram Bot to the list of alert channels.
Use only Bot API Token and Chat ID for configuring channel.
Grafana used telegram bot for sending notifications (https://grafana.com/blog/2023/12/28/how-to-integrate-grafana-alerting-and-telegram/).

Ellipse0934:
Hi, I would like to work on this.

I've strolled through the codebase here and seen relevant PR's like the implementation of opsgenie and sns. I have also tested the telegram API a bit and know what to do. However, the signoz fork of alertmanager predates that of the prometheus telegram implementation. So maybe there is potential for some improvement if that's alright.

An idea for a better workflow is to make a group/supergroup and make the bot admin there. Hence, users can join and leave relevant groups for alerts requiring less hands-on management with each `chat_id`. This comes at basically zero cost since both(individual and group) are `chat_id`. It's also more practical if multiple people are in a group so they can coordinate actions via chat there itself. I'm not sure who uses Telegram for devops but wouldn't be surprised if there are teams in eastern Europe who use it seriously in such situations.

YounixM:
cc @srikanthccv

DES-Destry:
I see, that alert manager is already supports Telegram. But it still is not available from Signoz Notification Channel creation. I would really like this feature

sedyh:
Hello team. I was wondering if there might be any updates or plans regarding this issue? Appreciate any info - thank you!

I could try to make a PR from the Go part if it helps.

oitan:
Any news on this? Happy to implement this feature.

PR #5411, opened 2024-07-02
feat(alerts): add telegram alert channel

Ellipse0934:
### Summary

Add telegram alert channel. PR once completed must be merged with https://github.com/SigNoz/alertmanager/pull/28 

#### Related Issues / PR's

https://github.com/SigNoz/signoz/issues/4700


srikanthccv:
@Ellipse0934 great progress 🎉

Ellipse0934:
On the one hand its "almost" done, I can successfully add a telegram channel and receive messages. Currently no message templating is exposed to the user but its a good feature to have regardless. So I think it's better to have a template defined with the deceiver so the user can set it. We can possibly defer this for another PR if you want or we can start here.

I did accidentally break something, It was working. But now I only get a notification when I change/save the config. Subsequent alerts are sent from query-service to alertmanager but something is failing and the failure is silent. In a way its good because it may likely point to a fragile part and save someone trouble in the future.

There is also the issue of HTML escape. It's not being done correctly. I need to fix this on the alertmanager PR.

Previews: 
image
image

Ellipse0934:
1. I was confused about how alerting worked and thought that alert was fired continuously. There was no issue so ignore the previous comment.
2. Bug with parsemode is fixed. Now the telegram message is correctly sent and rendered as per telegram styling entities.

Let me know if you need any help reviewing. You'll need to create a telegram bot. Instructions for that can be found here: https://core.telegram.org/bots/tutorial

developerisnow:
guys can't wait hearings from you @YounixM , @srikanthccv ! Telegram integration doesn't work for newcomer as me

srikanthccv:
This missed my radar. @Ellipse0934 do you want to continue the work?

Ellipse0934:
No. This was working when I marked it for review. I had pinged on the community slack too but received no response.

Quite likely that this will work if you resolve the conflicts.

srikanthccv:
@developerisnow if you would like to see Telegram integration, you are free to take over this PR work and create a separate PR.

Issue #5394, opened 2024-07-01
URL state should not be transferred between pages


srikanthccv:
It is a hard to reproduce error but we have seen it happen several times. The user often doesn't know how to solve this unless someone guides them to reset the URL parameters.

- somehow `state` from alerts appears in the dashboard' sort `columnKey` which results in an empty https://signoz-team.slack.com/archives/C07446MCZ5X/p1718851203534629
- the `null` order from the alerts edit page gets transferred to log-explorer pages and breaks the query https://signoz-team.slack.com/archives/C04RPAKMUBZ/p1719826387048489

Our team members also faced this issue but they know how to workaround but users don't.

PR #5417, opened 2024-07-03
fix: url params should not propagate across pages

vikrantgupta25:
### Summary

Issue 1: 
- dashboards list url params gets initialised with the wrong params since the `DashboardsProvider` is wrapped at the very top of the application.
- moved the initialisation to pick values from the url only when present on the Dashboards List page.
- added extra checks in place to default to `updatedAt` when the key is mismatched and order to `descend`

Issue 2: 
- when on alerts list page if the order is set to `undefined` then after that navigating to the old logs explorer the order is retained. 
- why? because it is being initialised in the reducer on load of the app. so moved the condition to only pick from the url on load if the user is present at the old logs explorer page else default it to `desc` as happening before. 
- added extra checks in place to default to `desc` when the key is mismatched

Issue 3: 
- the performance of rendering dashboard list page was very poor as the component was re-rendering around a `100` times.
- refactored the component for better performance.
- the search string was not getting updated in the URL.
- because of this re-renders we were unable to write proper unit test for the same.

*Started adding test cases for the dashboards list page*

#### Related Issues / PR's

Fixes :- https://github.com/SigNoz/signoz/issues/5394 

Fixes :- https://github.com/SigNoz/signoz/issues/5448 

#### Screenshots

NA

- tested dashboard list page ( various use cases / set different sorts and goto details page and navigate back etc.)
- tested the url isolation issue with logs and dashboards


frontend/src/container/ListOfDashboard/DashboardsList.tsx
```diff
@@ -311,12 +301,15 @@ function DashboardsList(): JSX.Element {
 
 	const handleSearch = (event: ChangeEvent<HTMLInputElement>): void => {
 		setIsFilteringDashboards(true);
-		setSearchValue(event.target.value);
 		const searchText = (event as React.BaseSyntheticEvent)?.target?.value || '';
-		const filteredDashboards = filterDashboard(searchText, dashboardListResponse);
+		const filteredDashboards = filterDashboard(
+			searchText,
+			dashboardListResponse || [],
```
SagarRajput-7, line 307: You removed this `data: dashboardListResponse = [],` on top and using `[]` for occurences, any specific reason?

SagarRajput-7 (approved):

frontend/src/container/ListOfDashboard/DashboardsList.tsx
vikrantgupta25: everytime the component re-renders there is always a new reference for the `dashboardListResponse`. hence causing all the useEffects to trigger again. We do not want that hence never initialise the default values here!

ahmadshaheer (dismissed):

SagarRajput-7 (approved):

ahmadshaheer (approved):

YounixM (approved):

PR #5425, opened 2024-07-04
chore: better logging for duplicate keyboard shortcuts

vikrantgupta25:
#### Screenshots

NA


YounixM (approved):

YounixM:
@vikrantgupta25 : Shortcut errors can be silent errors. It should not break the UI or show up in UI. Please check if that's the case.

vikrantgupta25:
@YounixM added this as a breaking error so that while developing we never duplicate the shortcut while in dev since silent errors can be ignored quite easily. the issue of this occurring in prod should never be present. maybe some edge case hence added better logging. let me know if you think otherwise.

vikrantgupta25:
> We can keep it as a breaking error in all places except production mode

@YounixM done 👍

SagarRajput-7 (approved):

YounixM (approved):

PR #5449, opened 2024-07-08
Send usage data day wise for last 72 hours.

nityanandagohain:
Fixes https://github.com/SigNoz/engineering-pod/issues/1472

Previously if usage data of day-1 was not sent then on current day.
* If there are no restarts then yesterdays usage is added to todays usage.

Now if usage data of day-1 was not sent then on current day.
* If there are no restarts then yesterdays usage is sent  along with todays usage.

This is done for the last 72 hours as the TTL of the usage table is 72 hours.

### DIFF
--- a/ee/query-service/usage/manager.go
+++ b/ee/query-service/usage/manager.go
@@ -121,7 +121,7 @@ func (lm *Manager) UploadUsage() {
 
 	for _, db := range dbs {
 		dbusages := []model.UsageDB{}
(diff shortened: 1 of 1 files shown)


srikanthccv:
I am not familiar with this part of code so need to understand it first. I will take a look when I get some time.

nityanandagohain:
Sure @srikanthccv

Issue #5431, opened 2024-07-04
Creating alerts from dashboard should not change the default aggregation interval


srikanthccv:
- Open a dashboard
- Create a panel
- Select a time range of 1 day
- Check the default aggregation interval
- It should show 60
- Try creating an alert from here

What is the current behaviour

- It redirects with the modified step interval

What is the expected behaviour

- It should be using the default step interval

PR #5455, opened 2024-07-09
fix: retain the step interval while creating alerts from the dashboard panel

vikrantgupta25:
### Summary

[Issue]: while creating alerts from the dashboards panel we use `v3/query_range/format` API call which takes care of replacing the dashboard variables with the actual values. but the API internally intelligently updates the `stepInterval` as well based on the time range. since now we have the `stepInterval` to be user configurable we do not want that. 

[Solution]: we only rely on the `format` API response for the filters  values in case of builder queries. we use the dashboard widget query as the base query.

[Test cases]: added test cases for the utils responsible for the calculations of merging the backend response and the widget query. 

case 1 - without variables response and widget query merging should not update the step interval 

case 2 - the variables in the filters and expressions should get properly replaces without touching the step interval

case 3 - the default case without widget query should update everything like before.

#### Related Issues / PR's

fixes https://github.com/SigNoz/signoz/issues/5431

#### Screenshots

NA

- test with dashboard variables and then the alerts creation  [ ✅ ]

https://github.com/SigNoz/signoz/assets/54737045/ad82aeda-99f7-4cd0-a9c9-3d465aea0ea0

- test without dashboard variables [ ✅ ]

https://github.com/SigNoz/signoz/assets/54737045/6754ce7f-e3d7-477e-8df7-b85d5875b06f

- test other areas where similar functions have been used [✅ ]


SagarRajput-7 (approved):

ahmadshaheer (dismissed):

frontend/src/lib/newQueryBuilder/transformQueryBuilderDataModel.ts
```diff
@@ -12,17 +13,47 @@ import { QueryBuilderData } from 'types/common/queryBuilder';
...
+			}
 		} else {
-			const query = value as IBuilderQuery;
-			queryData.push({ ...initialQueryBuilderFormValuesMap.metrics, ...query });
+			const queryFromData = value as IBuilderQuery;
+			const baseQuery = query?.queryData?.find(
+				(q) => q.queryName === queryFromData.queryName,
+			);
+
+			if (!isUndefined(baseQuery)) {
```
srikanthccv, line 42: Please help me understand this with examples.

frontend/src/lib/newQueryBuilder/queryBuilderMappers/__tests__/mapQueryDataFromApiInputs.ts
```diff
@@ -0,0 +1,495 @@
...
+				orderBy: [],
+				pageSize: 0,
+				queryName: 'A',
+				reduceTo: 'avg',
+				spaceAggregation: 'sum',
+				stepInterval: 240,
+				timeAggregation: 'rate',
+			},
+		],
+		queryFormulas: [],
```
srikanthccv, line 488: It is possible that the user can go to the alerts page from the dashboard and then make the changes on the alerts query builder. What happens when the user goes to the alerts page with `A/B` and then on the alerts page deletes `F1` and `B`, Will the step interval remain unchanged?

vikrantgupta25: once the user navigates from the dashboards page to the alerts page our task is to take the query and move the same as is to alerts with the variable values filled. post that user can do whatever changes they feel necessary in the alerts page. we do not block any operation there.

frontend/src/lib/newQueryBuilder/transformQueryBuilderDataModel.ts
vikrantgupta25: here what we are doing is taking the base query of the widget as supplied in the dashboard and using the `filters` value from the format API call which fills the values for the variables and only updating the same in the base query.

frontend/src/lib/newQueryBuilder/queryBuilderMappers/__tests__/mapQueryDataFromApiInputs.ts
vikrantgupta25: added test cases for multiple queries with functions as well

srikanthccv (approved):

YounixM (approved):

Issue #4566, opened 2024-02-17
Removal of materialized column should inspect `system.distribution_queue` before removing local table column


srikanthccv:
## Bug description

We should make sure there are no pending items in the `distribution_queue` with the old schema before removing the column from the local table.

Why is this necessary?

Imagine a case where the `distribution_queue` has some pending items with a mat column `service$$name` but before they are sent to local shards, the column is removed from the local tables. Now, the pending items are incompatible with the current schema leading to continuous failures until either the items are removed or the schema is fixed.

Related Part of https://github.com/SigNoz/engineering-pod/issues/1204

srikanthccv:
Temporary work around for those who encounter `No such column` errors. DETAH and re ATTACH the distributed table.

```
DETACH TABLE signoz_logs.distributed_logs ON CLUSTER cluster
ATTACH TABLE signoz_logs.distributed_logs ON CLUSTER cluster
```

srikanthccv:
It's generally a good idea to not let mutation/mutation-adjacent operation so easily.

nityanandagohain:
> It's generally a good idea to not let mutation/mutation-adjacent operation so easily.


Yeah we should consider this in the new exporer page.

srikanthccv:
[...]

Issue #5150, opened 2024-06-05
LOGS[FE]: Key DataType not respected and DataType not shown for materialised columns


nityanandagohain:
* Respect the data type of the key
* respect the tagType of the Key (updated)
* Show the type and dataType even if it's materialized.

https://github.com/SigNoz/signoz/assets/26831659/69517cb5-097b-48ad-aadf-8e3846c5e1ca

atmask:
Just to add on to this. I have also been looking into this as it's been impacting me. The `Tag` value is also not respected when there is a `tag` and `resource` with the same name.

YounixM:
@vikrantgupta25 : As discussed, assigning this to you.

vikrantgupta25:
we also need a way to differentiate between the different kinds of attributes. let's say we type `customAttribute = X` without using the suggested options, we wouldn't know if the customAttribute is a tag / resource etc.

vikrantgupta25:
another issue with the same is no difference and wrong data being shown in the selected columns 
<img width="1280" alt="image" src="https://github.com/user-attachments/assets/f5e9112c-e7cd-42f3-a380-7fd9962a45e9">

and no visible difference in the option to add more columns

vikrantgupta25:
https://github.com/SigNoz/signoz/issues/5645

vikrantgupta25:
will be taken care here - https://github.com/SigNoz/signoz/pull/5659

vikrantgupta25:
https://github.com/SigNoz/signoz/issues/5359

PR #5457, opened 2024-07-09
fix: disable removing a selected field

nityanandagohain:
This will temporarily resolve the issue #4566 since it won't allow users to remove a selected field.

The message on the UI will look like this
image

How it affects the old explorer ?
* Since selected fields are connect to what fields are displayed in the old UI, users wont be able to remove that once added.

How it affects the new explorer ?
* As of now we don't display the `type` and `dataType` of selected fields. But it will be resolved once this issue is fixed, nothing apart from that. https://github.com/SigNoz/signoz/issues/5150


ankitnayan:
How displaying the `type` and `dataType` of a selected field is linked to this PR?

srikanthccv (commented):
Mistakenly committed local db. Please remove it.

srikanthccv (approved):

PR #5497, opened 2024-07-15
fix: incorrect response for promql value type panels

srikanthccv:
### Summary

We don't currently support caching for value panel types https://github.com/SigNoz/signoz/blob/3ecb2e35ef3d0755bf56e2b378f5e8e4d2951c1b/pkg/query-service/app/queryBuilder/query_builder.go#L325-L328 (with v4, the cache can be re-used for metrics but that's for separate issut). The func returns an empty cache keys map for non-graph types and the caller is responsible for making sure the cache key exists before using it. This is already done in builder queries but no such check exists for promql queries. This means any panel type that is not time series with promql queries uses the same empty cache key and returns incorrect results. This change fixes it.

### DIFF
--- a/pkg/query-service/app/querier/helper.go
+++ b/pkg/query-service/app/querier/helper.go
@@ -18,7 +18,7 @@ import (
 	"go.uber.org/zap"
 )
 
-func prepareLogsQuery(ctx context.Context,
+func prepareLogsQuery(_ context.Context,
 	start,
 	end int64,
 	builderQuery *v3.BuilderQuery,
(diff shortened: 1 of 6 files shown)


pkg/query-service/app/querier/helper.go
```diff
@@ -18,7 +18,7 @@ import (
 	"go.uber.org/zap"
 )
 
-func prepareLogsQuery(ctx context.Context,
+func prepareLogsQuery(_ context.Context,
```
nityanandagohain, line 21: why this change ? Is it only because we are not using it?

nityanandagohain (approved):

pkg/query-service/app/querier/helper.go
srikanthccv: because linter complained about it; build would have failed if it was used anywhere.

Issue #4796, opened 2024-04-02
User Preference API


YounixM:
User Preference API to help store user preferences across user sessions.

This will also help reduce storing a lot of data in local storage.

Use case - > https://signoz-team.slack.com/archives/C05AXQL92BT/p1716600256934909?thread_ts=1716589743.951289&cid=C05AXQL92BT

YounixM:
@vikrantgupta25 : Add TDD, Estimates around this.

vikrantgupta25:
will start working on the TDD today, we already have a base TDD/PRD ready but want to revamp and revisit that again.. Will be able to provide proper estimates once the TDD is ready.

vikrantgupta25:
PRD/TDD for the same :- https://www.notion.so/signoz/Preference-Framework-55b6ab60b1ad4ea3a6584814e16be031

PR #5527, opened 2024-07-19
feat: preference framework qs changes 

vikrantgupta25:
### Summary

- query service changes for preference framework

#### Related Issues / PR's

fixes https://github.com/SigNoz/signoz/issues/4796

#### Screenshots

NA


ee/query-service/app/server.go
```diff
@@ -110,6 +111,10 @@ func NewServer(serverOptions *ServerOptions) (*Server, error) {
 
 	baseexplorer.InitWithDSN(baseconst.RELATIONAL_DATASOURCE_PATH)
 
+	if err := preferences.InitDB(baseconst.RELATIONAL_DATASOURCE_PATH); err != nil {
```
grandwizard28, line 114: The schema should be bootstrapped with dao as in line 107.

Is there any specific reason this is being bootstrapped separately?

pkg/query-service/app/http_handler.go
```diff
@@ -2192,6 +2193,130 @@ func (aH *APIHandler) WriteJSON(w http.ResponseWriter, r *http.Request, response
 	w.Write(resp)
 }
 
+// Preferences
+func (ah *APIHandler) RegisterPreferenceRoutes(router *mux.Router, am *AuthMiddleware) {
```
grandwizard28, line 2197: These routes seem restfully wrong.

An organisation/user has preferences. The path to get preferences should be
- user/preferences for all preferences of a user
- user/preferences/{preferenceId} for an individual preference

Similarly for org.

pkg/query-service/app/preferences/model.go
```diff
@@ -0,0 +1,518 @@
...
+	PreferenceValueTypeString  string = "string"
+	PreferenceValueTypeBoolean string = "boolean"
+)
+
+const (
+	OrgAllowedScope  string = "org"
+	UserAllowedScope string = "user"
+)
+
+func checkIfInAllowedValues(preferenceValue interface{}, preference *Preference) bool {
```
grandwizard28, line 49: This can be a function of the preference object!

pkg/query-service/app/preferences/model.go
```diff
@@ -0,0 +1,518 @@
...
+	case PreferenceValueTypeInteger:
+		val, ok := preferenceValue.(int64)
+		if !ok {
+			floatVal, ok := preferenceValue.(float64)
+			if !ok || floatVal != float64(int64(floatVal)) {
+				return p.ErrorValueTypeMismatch()
+			}
+			val = int64(floatVal)
+		}
+		if !p.IsDescreteValues {
```
grandwizard28, line 70:
```suggestion
		if !p.IsDiscreteValues {
```

pkg/query-service/app/preferences/model.go
```diff
@@ -0,0 +1,518 @@
...
+			return true
+		} else {
+			return false
+		}
+	default:
+		return preferenceValue
+	}
+}
+
+type AllPreferences struct {
```
grandwizard28, line 129: What is the difference between this struct and the preference struct?

grandwizard28 (changes requested):

ee/query-service/app/server.go
vikrantgupta25: it helps keep the module initialization separate. like we do for dashboards and alerts in the follow up line

pkg/query-service/app/http_handler.go
vikrantgupta25: updated

pkg/query-service/app/preferences/model.go
vikrantgupta25: done

pkg/query-service/app/preferences/model.go
vikrantgupta25: the value is of the preference is a part of this struct where as the preference entity doesn't have the value attribute

grandwizard28: Can you use struct inheritance to make the code DRY?

pkg/query-service/app/http_handler.go
```diff
@@ -2192,6 +2193,130 @@ func (aH *APIHandler) WriteJSON(w http.ResponseWriter, r *http.Request, response
...
+	subRouter.HandleFunc("/org/preferences/{preferenceId}", am.AdminAccess(ah.getOrgPreference)).Methods(http.MethodGet)
+
+	subRouter.HandleFunc("/org/preferences/{preferenceId}", am.AdminAccess(ah.updateOrgPreference)).Methods(http.MethodPost)
+}
+
+func (ah *APIHandler) getUserPreference(
+	w http.ResponseWriter, r *http.Request,
+) {
+	preferenceId := mux.Vars(r)["preferenceId"]
+	orgId := r.URL.Query().Get("orgId")
```
grandwizard28, line 2219: The route for this is /user/preferences. It makes more sense to change the route to /org/{orgId}/user/preferences if you always need the orgId to be there.

Otherwise you will have to validate whether the mandatory orgId is present or not here.

pkg/query-service/app/http_handler.go
```diff
@@ -2192,6 +2193,130 @@ func (aH *APIHandler) WriteJSON(w http.ResponseWriter, r *http.Request, response
...
+	}
+
+	ah.Respond(w, preference)
+}
+
+func (ah *APIHandler) getOrgPreference(
+	w http.ResponseWriter, r *http.Request,
+) {
+	preferenceId := mux.Vars(r)["preferenceId"]
+	orgId := r.URL.Query().Get("orgId")
```
grandwizard28, line 2271: So does that mean the url is /org/preferences?orgId=<id>?

If yes, again it makes more idomatic sense to move it to /org/{orgId}/preferences!

grandwizard28 (changes requested):

pkg/query-service/app/http_handler.go
vikrantgupta25: this code will be changed, we will be relying on the orgID from the JWT token. so orgId will not be the part of the request altogether.

pkg/query-service/app/http_handler.go
vikrantgupta25: no it won't be the same. replied above

pkg/query-service/app/preferences/model.go
vikrantgupta25: done

pkg/query-service/app/preferences/model.go
```diff
@@ -0,0 +1,510 @@
...
+	}
+
+	preferenceMap = map[string]Preference{}
+	for _, preference := range preferences {
+		_, seen := preferenceMap[preference.Key]
+		if seen {
+			return fmt.Errorf("duplicate preference key in the preferences: %s", preference.Key)
+		}
+		preferenceMap[preference.Key] = preference
+	}
```
grandwizard28, line 212: What's the thought process here? Are we loading some default preferences?

pkg/query-service/app/preferences/model.go
```diff
@@ -0,0 +1,510 @@
...
+
+	// if the value doesn't exist in db then return the default value
+	if err == sql.ErrNoRows {
+		return &PreferenceKV{
+			PreferenceId:    preferenceId,
+			PreferenceValue: preference.DefaultValue,
+		}, nil
+	} else if err != nil {
+		return nil, &model.ApiError{Typ: model.ErrorExec, Err: fmt.Errorf("error in fetching the org preference: %s", err.Error())}
+	}
```
grandwizard28, line 244: Always try doing the nil check first! It cultivates a good habit to check with pointers and is a little more idiomatic

pkg/query-service/app/preferences/model.go
grandwizard28: If yes, the go binary will ultimately not have the json file and this will throw an error.
You can create a file `map.go` in preferences package and rewrite the json as a `map[string]interface{}`

pkg/query-service/app/preferences/model.go
```diff
@@ -0,0 +1,510 @@
...
+		}
+	}
+	return &allOrgPreferences, nil
+}
+
+// user preference functions
+func GetUserPreference(ctx context.Context, preferenceId string, orgId string) (*PreferenceKV, *model.ApiError) {
+
+	user := common.GetUserFromContext(ctx)
+	// check if the preference key exists
```
grandwizard28, line 343: This is an interesting choice here. The user is derived from the request context and is specific to the http entrypoint. The derivation of user in such a way makes it a part of the http entrypoint and as such the user should be passed in the GetUserPreference function as an argument.

By deriving the user from the context here, you are essentially saying, "Hey, this function can only be called if it comes from an http request."

grandwizard28 (changes requested):

pkg/query-service/app/preferences/model.go
vikrantgupta25: this is a nice thought here. makes sense to pass the user reference from the consumer here allowing this function to be reusable somewhere else as well when required.

pkg/query-service/app/preferences/model.go
vikrantgupta25: we are not reading the default preferences here but the overall preference entity that is defined across tenants. let me check on loading the JSON part.

vikrantgupta25: if i use the `go: embed` directive for this I should be able to load the file right ?

grandwizard28: You should be able to but the question is why would you want to read a json file, unmarshal it into a struct, then iterate and put into your map when you call initialize the map in the first place?

pkg/query-service/app/preferences/model.go
vikrantgupta25: done

pkg/query-service/app/preferences/model.go
vikrantgupta25: done

pkg/query-service/app/preferences/model.go
vikrantgupta25: done

grandwizard28 (dismissed):

pkg/query-service/app/preferences/model.go
```diff
@@ -0,0 +1,479 @@
...
+			}
+		}
+	}
+	return isPreferenceEnabledForGivenScope
+}
+
+func (p *Preference) SanitizeValue(preferenceValue interface{}) interface{} {
+	switch p.ValueType {
+	case PreferenceValueTypeBoolean:
+		if preferenceValue == "1" {
```
srikanthccv, line 115: if you are feeling more permissive.
```suggestion
		if preferenceValue == "1" || preferenceValue == "true" {
```

pkg/query-service/app/preferences/model.go
```diff
@@ -0,0 +1,479 @@
...
+
+	return nil
+}
+
+// org preference functions
+func GetOrgPreference(ctx context.Context, preferenceId string, orgId string) (*PreferenceKV, *model.ApiError) {
+	// check if the preference key exists or not
+	preference, seen := preferenceMap[preferenceId]
+	if !seen {
+		return nil, &model.ApiError{Typ: model.ErrorBadData, Err: fmt.Errorf("no such preferenceId exists: %s", preferenceId)}
```
srikanthccv, line 195: ErrorNotFound is more appropriate?

```suggestion
		return nil, &model.ApiError{Typ: model.ErrorNotFound, Err: fmt.Errorf("no such preferenceId exists: %s", preferenceId)}
```

pkg/query-service/app/preferences/model.go
```diff
@@ -0,0 +1,479 @@
...
+	}
+
+	// check if the preference is enabled for org scope or not
+	isPreferenceEnabled := preference.IsEnabledForScope(OrgAllowedScope)
+	if !isPreferenceEnabled {
+		return nil, &model.ApiError{Typ: model.ErrorForbidden, Err: fmt.Errorf("preference is not enabled at org scope: %s", preferenceId)}
```
srikanthccv, line 201: Can I say preference not enabled for org is a preference that does not exist in the org context? I am trying to understand why is `ErrorForbidden` used here.

pkg/query-service/app/preferences/model.go
```diff
@@ -0,0 +1,479 @@
...
+			preferenceWithValue.AllowedScopes = preference.AllowedScopes
+			preferenceWithValue.AllowedValues = preference.AllowedValues
+			preferenceWithValue.DefaultValue = preference.DefaultValue
+			preferenceWithValue.Range = preference.Range
+			preferenceWithValue.ValueType = preference.ValueType
+			preferenceWithValue.IsDiscreteValues = preference.IsDiscreteValues
+			ok, seen := preferenceValueMap[preference.Key]
+
+			if seen {
+				preferenceWithValue.Value = ok
```
srikanthccv, line 301:
```suggestion
			value, seen := preferenceValueMap[preference.Key]

			if seen {
				preferenceWithValue.Value = value
```

pkg/query-service/app/preferences/model.go
```diff
@@ -0,0 +1,479 @@
...
+	err := preference.IsValidValue(preferenceValue)
+	if err != nil {
+		return nil, err
+	}
+	// update the user preference values
+	query := `INSERT INTO user_preference(preference_id,preference_value,user_id) VALUES($1,$2,$3)
+	ON CONFLICT(preference_id,user_id) DO
+	UPDATE SET preference_value= $2 WHERE preference_id=$1 AND user_id=$3;`
+
+	_, dberrr := db.Exec(query, preferenceId, preferenceValue, userId)
```
srikanthccv, line 394: I see `preference_value` is TEXT in db. Do you have some specific serialization format? For example, how to list type `preferenceValue` get stored in DB? You may not have that case right now but I want to understand how does the ser/deser work

pkg/query-service/app/preferences/map.go
```diff
@@ -0,0 +1,37 @@
...
+		Description:      "",
+		ValueType:        "integer",
+		DefaultValue:     10,
+		IsDiscreteValues: false,
+		Range: Range{
+			Min: 0,
+			Max: 100,
+		},
+		AllowedScopes: []string{"user", "org"},
+	},
```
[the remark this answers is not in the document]
vikrantgupta25: no these are just temporary will add the preferences in the follow up PR as and when we decide to onboard them in the UI

pkg/query-service/app/preferences/model.go
vikrantgupta25: yes this makes more sense, will update this error

pkg/query-service/app/preferences/model.go
vikrantgupta25: the values stored in the DB are stored as TEXT. as in this file we have `SanitizeValue` function which takes care of deserialising the values. similar to this when the need comes we will have the serialising function as well where we will perform the serialisations

pkg/query-service/app/preferences/model.go
```diff
@@ -0,0 +1,479 @@
...
+				return &model.ApiError{Typ: model.ErrorBadData, Err: fmt.Errorf("the preference value is not in the range specified, min: %v , max:%v", p.Range.Min, p.Range.Max)}
+			}
+		}
+	case PreferenceValueTypeString:
+		_, ok := preferenceValue.(string)
+		if !ok {
+			return p.ErrorValueTypeMismatch()
+		}
+	case PreferenceValueTypeFloat:
+		_, ok := preferenceValue.(float64)
```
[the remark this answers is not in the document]
vikrantgupta25: the integers number are also valid float numbers. `10.0` is also acceptable floating point and since we decode the request for the preference value all the numbers are by default parsed as `float64` , so we wouldn't know if the user passed 10 or 10.0

pkg/query-service/app/preferences/model.go
```diff
@@ -0,0 +1,479 @@
...
+	OrgAllowedScope  string = "org"
+	UserAllowedScope string = "user"
+)
+
+func (p *Preference) checkIfInAllowedValues(preferenceValue interface{}) bool {
+	isInAllowedValues := false
+	for _, value := range p.AllowedValues {
+		if value == preferenceValue {
+			isInAllowedValues = true
+		}
```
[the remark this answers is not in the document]
vikrantgupta25: this function is being called from the `isValidValue` check where the necessary type conversions have been made based on the `ValueType` of the preference

pkg/query-service/app/preferences/model.go
```diff
@@ -0,0 +1,479 @@
...
+	}
+	return nil
+}
+
+func (p *Preference) IsEnabledForScope(scope string) bool {
+	isPreferenceEnabledForGivenScope := false
+	if p.AllowedScopes != nil {
+		for _, allowedScope := range p.AllowedScopes {
+			if allowedScope == scope {
```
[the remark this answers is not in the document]
vikrantgupta25: done

pkg/query-service/app/preferences/model.go
vikrantgupta25: done , and the preferenceValue is either being picked from db in which case it will be "1" or from the default value where it would be `true`

pkg/query-service/app/preferences/model.go
vikrantgupta25: done

pkg/query-service/app/preferences/model.go
vikrantgupta25: done

pkg/query-service/app/preferences/model.go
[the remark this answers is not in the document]
srikanthccv: While that may be true now, eventually you either forget that the passed value should be a comparable interface or someone who doesn't know the codebase touches this part and won't be familiar with the detail. We want to avoid that. You either make sure not to panic or explicitly call out that this expects comparable interfaces only.

pkg/query-service/app/preferences/model.go
srikanthccv: `TEXT` is the data type, not a serialization format. Are you saying we will write a custom ser/deser for preference values?

vikrantgupta25: yes when writing values for non standard types like list or objects we will serialise them to strings ( `stringify`) and write to DB and when the `sanitizeValue` function will take care of non serialising when returning the values

pkg/query-service/app/preferences/model.go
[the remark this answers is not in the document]
vikrantgupta25: added checks in the `IsInAllowedValues` function to first check if the preference value is of the correct ValueType and then on each value of the allowedValues as well should be of the correct ValueType and then compare the same. if at any point something doesn't match with values return the error

srikanthccv (dismissed):

pkg/query-service/app/preferences/model.go
vikrantgupta25: @grandwizard28 the benefits we get out of using JSON is we can make changes to the config file independent of the go compilation and hence whenever we change the config file we do not have to rebuild the go binary,

grandwizard28: go:embed will rebuild it!

vikrantgupta25: we will not be using `go:embed` for the same rather pass it as a config so that whenever there is a change of addition or deletion for the preferences we do need to rebuild everything.

vikrantgupta25: made the changes please check @grandwizard28

grandwizard28: This is a lot of added complexity for handling a simple thing like constants. Please put it in the code itself.
There is integrity in the preference struct you create and initialize as a constant.

Otherwise, you need to unmarshal from json, put explicit checks for empty fields, correct fields etc.

deploy/docker/clickhouse-setup/docker-compose-local.yaml
```diff
@@ -14,13 +14,15 @@ services:
       - ClickHouseUrl=tcp://clickhouse:9000
       - ALERTMANAGER_API_PREFIX=http://alertmanager:9093/api/
       - SIGNOZ_LOCAL_DB_PATH=/var/lib/signoz/signoz.db
+      - PREFERENCES_PATH=/root/config/preferences.json
```
prashant-shahi, line 17: include the same changes for docker swarm and query-service testing-deploy docker-compose files

vikrantgupta25: done

grandwizard28 (approved):

srikanthccv (approved):

pkg/query-service/auth/jwt.go
```diff
@@ -35,6 +35,13 @@ func ParseJWT(jwtStr string) (jwt.MapClaims, error) {
 	if !ok || !token.Valid {
 		return nil, errors.Errorf("Not a valid jwt claim")
 	}
+
+	// the function validateUser panics to convert claims["orgId"] to string as the jwt.Parse doesn't check for claims. 
+	// the function jwt.ParseWithClaims does check for custom claims. 
+	// TODO[@vikrantgupta25] : to update this to the claims check function for better integrity of JWT
+	if claims["orgId"] == nil {
```
vikrantgupta25, line 42: reference - https://pkg.go.dev/github.com/golang-jwt/jwt/v5#Parser.ParseWithClaims

pkg/query-service/auth/jwt.go
```diff
@@ -47,11 +50,18 @@ func validateUser(tok string) (*model.UserPayload, error) {
 	if !claims.VerifyExpiresAt(now, true) {
 		return nil, model.ErrorTokenExpired
 	}
```
vikrantgupta25, line 52: orgId claim will not be present in the older jwt tokens hence the extra handling

pkg/query-service/app/http_handler.go
```diff
@@ -2192,6 +2210,136 @@ func (aH *APIHandler) WriteJSON(w http.ResponseWriter, r *http.Request, response
...
 }
 
+// Preferences
+
+func (ah *APIHandler) getUserPreference(
+	w http.ResponseWriter, r *http.Request,
+) {
+	preferenceId := mux.Vars(r)["preferenceId"]
+	user := common.GetUserFromContext(r.Context())
+
```
vikrantgupta25, line 2220: if the orgId is not present that means the user is consuming the older JWT token hence we return 401 from here indicating the frontend to explicit logout from here

srikanthccv (approved):

PR #5528, opened 2024-07-19
Feat: Logs new schema support

nityanandagohain:
PR for new logs schema

Fixes https://github.com/SigNoz/signoz/issues/5555

current limitations
* Attribute search is case sensitive, while resources and body search is case insensitive. 
* Minmax Index on numbers don't work (clickhouse limitation)

How to go about this PR

**Section 1:- Logs v3 and v4 builder**
* V3 is the existing builder and has no changes
* V4 is the new builder that supports the new schema, the test cases are similar to v3 with the changes related to the schema
* Resource filters are only added to the resource table and not to the logs_v2 table 

**Section 2:- The materialize api's and attribute api's in reader.go**
* These APIs just houses changes to support new schema through the feature flags

**Section 3:- ` runLogsListQuery` in querier.go**
We have updated the logs on how list data is fetched for logs, first we check the last 30 mins data to find the 100 logs and then increase the window size until 100 logs are found.

**Section 4: - Use of the useLogsNewSchema flag.**
Changes to different files are added just so that useLogsNewSchema can be passed around.


nityanandagohain:
I have added a description of how to go about the PR, let me know if a walkthrough is required.

srikanthccv:
Even with a walkthrough, I won't be able to review this (I made a couple of attempts to skim through but it's too big to comprehend). It is 4k+ additions. Please send them in multiple stacked PRs.

nityanandagohain:
Sure, I will try to break it down.

PR #5535, opened 2024-07-22
chore: added unit tests for help and support section

rahulkeswani101:
### Summary

- Test cases covered
   - Renders the support page with all support channels
   - should trigger correct handler function when channel button is clicked
   - Opens the Intercom chat widget when the Chat channel is clicked
   - Handling window.Intercom being undefined or null
   -  Handle missing or undefined `history.location.state`
   - Handle missing or undefined `channel.url`
   - Handle missing or undefined `channel.name`

### DIFF
--- a/frontend/src/pages/Support/Support.test.tsx
+++ b/frontend/src/pages/Support/Support.test.tsx
@@ -1,0 +1,135 @@
+import { Calendar } from 'antd';
+import { Book, Cable, Github, MessageSquare, Slack } from 'lucide-react';
+import { fireEvent, render } from 'tests/test-utils';
+
+import Support from './Support';
+
+const launchChat = 'Launch chat';
+const useAnalyticsMock = jest.fn();
+const useHistoryMock = jest.fn();
+
+jest.mock('hooks/analytics/useAnalytics', () => ({
+	__esModule: true,
+	default: jest.fn(() => ({ trackEvent: useAnalyticsMock })),
+}));
+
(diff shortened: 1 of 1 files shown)


frontend/src/pages/Support/Support.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+	default: jest.fn(() => ({ trackEvent: useAnalyticsMock })),
+}));
+
+jest.mock('react-router-dom', () => ({
+	useHistory: useHistoryMock,
+}));
+
+const supportChannels = [
+	{
+		key: 'documentation',
```
vikrantgupta25, line 22: we already have the support channels in the code right ? please export it from there and use here. this increases the maintainability of the code and helps keep the code DRY

frontend/src/pages/Support/Support.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+	});
+});
+
+describe('Handle channels with null or undefined properties', () => {
+	it('should handle window.Intercom being undefined or null', () => {
+		window.Intercom = null;
+		const { getByText } = render(<Support />);
+		const button = getByText(launchChat);
+		fireEvent.click(button);
+		expect(window.Intercom).toBeNull();
```
vikrantgupta25, line 112: what is the purpose of this test? at line 108 you declared the variable to be null. now even if you skip line 109-111 it will always be null.

PR #5539, opened 2024-07-22
fix: handle the super set query reset state when changing widgets

vikrantgupta25:
### Summary

[Issue]: the state of the superset query was retained even on changing the widgets due to which the values appeared in the other panels from the older accessed panels.

[Solution]: reset the superset query on load of a dashboard panel to the current query at that instant. this helps to clear any old hanging state from the same 

#### Related Issues / PR's

fixes https://github.com/SigNoz/engineering-pod/issues/1557

#### Screenshots

NA


YounixM (approved):

srikanthccv:
@vikrantgupta25 https://github.com/SigNoz/signoz/issues/5591

PR #5542, opened 2024-07-24
messaging queue, consumer lag APIs

shivanshuraj1333:
fixes https://github.com/SigNoz/engineering-pod/issues/1589 and https://github.com/SigNoz/engineering-pod/issues/1586
epic: https://github.com/SigNoz/engineering-pod/issues/1584
adds producer, and consumer API


pkg/query-service/app/integrations/messagingQueues/kafka/translator.go
```diff
@@ -0,0 +1,101 @@
...
+	if !ok {
+		return nil, fmt.Errorf("invalid type for Topic")
+	}
+
+	partition, ok := messagingQueue.Variables["partition"]
+
+	if !ok {
+		return nil, fmt.Errorf("invalid type for Partition")
+	}
+	query := generateProducerSQL(start, end, topic, partition)
```
ankitnayan, line 60: Is this prone to sql injection? I have seen folks use prepared statements for CH query

ankitnayan:
The queries miss filtering on `msg.system=kafka` else the query will run on any messaging queue

ankitnayan:
1) Consumer Lag Graph and 4) Network latency tab are pending

pkg/query-service/app/integrations/messagingQueues/kafka/translator.go
shivanshuraj1333: No, it's not prone to SQL injection as the query itself is not exposed in the API endpoint, only a few parameters are exposed as request body parameters.

Also, thinking more, I think all our `existing` APIs are prone to SQL injection since users know the payload contains the composite query, and the code is open. They can inject SQL based composite query IMO.
I have created an issue [here](https://github.com/SigNoz/signoz/issues/5566) for query range APIs.

shivanshuraj1333:
> The queries miss filtering on msg.system=kafka else the query will run on any messaging queue

The API endpoint justifies that, for now let's use `/api/v1/messaging-queues/kafka/*` where `messaging-queues/<queue_name>` the common interface pattern is not there today to handle different queues.

Once `Kafka` is done and the functions are stabilised and we want to expand it to other queues, the code will be translated to a common interface and then to reduce the number of APIs we can introduce `msg.system=kafka` and reduce the API endpoints to `/api/v1/messaging-queues/*` + `msg.system=kafka` in request body. Right now for v1 it's not needed, later on it'd be easier to extend and modify the APIs with minimal changes.

shivanshuraj1333:
> Consumer Lag Graph and 4) Network latency tab are pending

I'll be having a separate PR for that, let's merge this first, all the features in one PR would make it difficult to review.

ankitnayan:
> > The queries miss filtering on msg.system=kafka else the query will run on any messaging queue
> 
> The API endpoint justifies that, for now let's use /api/v1/messaging-queues/kafka/* where messaging-queues/<queue_name> the common interface pattern is not there today to handle different queues.
> 
> Once Kafka is done and the functions are stabilised and we want to expand it to other queues, the code will be translated to a common interface and then to reduce the number of APIs we can introduce msg.system=kafka and reduce the API endpoints to /api/v1/messaging-queues/* + msg.system=kafka in request body. Right now for v1 it's not needed, later on it'd be easier to extend and modify the APIs with minimal changes.

Sorry for the confusion, I meant to add `msg.system=kafka` in the sql query. Eg at https://github.com/SigNoz/signoz/pull/5542/files#diff-480b722b377e24e0d7eff30cee9640595400fd21ffbc67283b67ff7e78385815R21

ankitnayan:
> > Consumer Lag Graph and 4) Network latency tab are pending
> 
> I'll be having a separate PR for that, let's merge this first, all the features in one PR would make it difficult to review.

Let' wait for some more reviews. You can create a PR with this PR as base

shivanshuraj1333:
> Sorry for the confusion, I meant to add msg.system=kafka in the sql query. Eg at https://github.com/SigNoz/signoz/pull/5542/files#diff-480b722b377e24e0d7eff30cee9640595400fd21ffbc67283b67ff7e78385815R21

I see, this is for `topic` there's another attribute `msgSystem`, and it's there for other SQLs, good catch need to add it here as well.
Thanks!

pkg/query-service/app/integrations/messagingQueues/kafka/translator.go
```diff
@@ -0,0 +1,104 @@
...
+	topic, ok := messagingQueue.Variables["topic"]
+	if !ok {
+		return nil, fmt.Errorf("invalid type for Topic")
+	}
+
+	partition, ok := messagingQueue.Variables["partition"]
+
+	if !ok {
+		return nil, fmt.Errorf("invalid type for Partition")
+	}
```
grandwizard28, line 62: The extraction of what you need from the messagingQueue struct for your clickhouse query should be moved to a common function.

Right now, it's duplicated in `buildProducerClickHouseQuery` and  `buildConsumerClickHouseQuery`. The more functions you add, the more difficult it starts to get to maintain.

shivanshuraj1333: True, created in haste, jumbling with many things, let me fix this

shivanshuraj1333: fixed

pkg/query-service/app/http_handler.go
```diff
@@ -2192,70 +2194,174 @@ func (aH *APIHandler) WriteJSON(w http.ResponseWriter, r *http.Request, response
...
+	result = postprocess.TransformToTableForClickHouseQueries(result)
+
+	resp := v3.QueryRangeResponse{
+		Result: result,
+	}
+	aH.Respond(w, resp)
+}
+
+// ParseMessagingQueueParams parse for messaging queue params
+func ParseMessagingQueueParams(r *http.Request) (*mq.MessagingQueue, *model.ApiError) {
```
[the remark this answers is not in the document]
shivanshuraj1333: done

pkg/query-service/app/integrations/messagingQueues/kafka/consumerLag.md
```diff
@@ -0,0 +1,197 @@
...
+                    "columns": [
+                        {
+                            "name": "service_name",
+                            "queryName": "",
+                            "isValueColumn": false
+                        },
+                        {
+                            "name": "p99",
+                            "queryName": "",
+                            "isValueColumn": false
```
srikanthccv, line 46: @vikrantgupta25 I don't recall what is `isValueColumn` used for. What would break if we send a value column with `isValueColumn` false?

pkg/query-service/app/integrations/messagingQueues/kafka/consumerLag.md
```diff
@@ -0,0 +1,197 @@
...
+```
+
+```json
+{
+	"start": 1720685296000000000,
+	"end": 1721290096000000000,
+	"variables": {
+        "partition": "0",
+		"topic": "topic1",
+        "consumer_group": "cg1"
```
srikanthccv, line 23: How does the frontend fetch the values for `partition`, `topic` and `consumer_group`? Using `autocomplete/attribute_values`?

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
```diff
@@ -0,0 +1,171 @@
...
+
+func generateConsumerSQL(start, end int64, topic, partition, consumerGroup, queueType string) string {
+	query := fmt.Sprintf(`
+WITH 
+-- Sub query for p99 calculation
+p99_query AS (
+    SELECT
+        serviceName,
+        quantile(0.99)(durationNano) / 1000000 as p99
+    FROM signoz_traces.signoz_index_v2
```
srikanthccv, line 15: You should be using `distributed_signoz_index_v2` everywhere otherwise it will only work with the local shard data query-service is connected to.

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
```diff
@@ -0,0 +1,171 @@
...
+      AND stringTagMap['messaging.destination.partition.id'] = '%s'
+	  AND stringTagMap['messaging.kafka.consumer.group'] = '%s'
+    GROUP BY serviceName
+),
+
+-- Sub query for RPS calculation
+rps_query AS (
+    SELECT
+        serviceName,
+        count(*) / ((%d - %d) / 1000000000) AS rps  -- Convert nanoseconds to seconds
```
srikanthccv, line 31: I hope you made sure there are no overflow issue with `(%d - %d)` part

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
```diff
@@ -0,0 +1,171 @@
...
+        AND kind = 5
+		AND msgSystem = '%s' 
+        AND stringTagMap['messaging.destination.name'] = '%s'
+        AND stringTagMap['messaging.destination.partition.id'] = '%s'
+		AND stringTagMap['messaging.kafka.consumer.group'] = '%s'
+    GROUP BY serviceName
+),
+
+-- Sub query for error rate calculation
+error_rate_query AS (
```
srikanthccv, line 45: do we want to show the rate as in `errors/s` or `error %` = `errors/total`?

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
```diff
@@ -0,0 +1,171 @@
...
+        avg(numberTagMap['messaging.message.body.size']) AS avg_msg_size
+    FROM signoz_traces.signoz_index_v2
+    WHERE
+        timestamp >= '%d'
+        AND timestamp <= '%d'
+        AND kind = 5
+        AND msgSystem = '%s' 
+        AND stringTagMap['messaging.destination.name'] = '%s'
+        AND stringTagMap['messaging.destination.partition.id'] = '%s'
+		AND stringTagMap['messaging.kafka.consumer.group'] = '%s'
```
srikanthccv, line 75: Will there ever be a case where `messaging.destination.*` exist but `messaging.message.body.size` doesn't?

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
```diff
@@ -0,0 +1,171 @@
...
+        count(*) / ((%d - %d) / 1000000000) AS error_rate  -- Convert nanoseconds to seconds
+    FROM signoz_traces.signoz_index_v2
+    WHERE
+        timestamp >= '%d'
+        AND timestamp <= '%d'
+        AND statusCode = 2
+        AND kind = 5
+        AND msgSystem = '%s' 
+        AND stringTagMap['messaging.destination.name'] = '%s'
+        AND stringTagMap['messaging.destination.partition.id'] = '%s'
```
srikanthccv, line 57: Some of these subqueries should be combined to reduce the amount of data we read otherwise each query will process the same amount of bytes multiple times. For example, `error_rate_query` and `avg_msg_size_query` can be combined. I would suggest we reduce the bytes we scan as much as possible.

1.
```
WITH
    avg_msg_size_query AS
    (
        SELECT
            serviceName,
            avg(numberTagMap['messaging.message.body.size']) AS avg_msg_size
        WHERE ((stringTagMap['messaging.destination.name']) != '') AND (timestamp >= (now() - toIntervalHour(1)))
        GROUP BY serviceName
    ),
    p99_query AS
    (
        SELECT
            serviceName,
            quantile(0.99)(durationNano) / 1000000 AS p99
        WHERE ((stringTagMap['messaging.destination.name']) != '') AND (timestamp >= (now() - toIntervalHour(1)))
        GROUP BY serviceName
    )
SELECT *
FROM avg_msg_size_query
INNER JOIN p99_query ON avg_msg_size_query.serviceName = p99_query.serviceName

Query id: 898c1643-6d1e-4a7b-829d-f60e7032a94a

Ok.

0 rows in set. Elapsed: 0.141 sec. Processed 360.84 thousand rows, 138.43 MB (2.56 million rows/s., 982.57 MB/s.)
Peak memory usage: 19.95 MiB.
```

2.

```
WITH avg_msg_size_query AS
    (
        SELECT
            serviceName,
            avg(numberTagMap['messaging.message.body.size']) AS avg_msg_size,
            quantile(0.99)(durationNano) / 1000000 AS p99
        WHERE ((stringTagMap['messaging.destination.name']) != '') AND (timestamp >= (now() - toIntervalHour(1)))
        GROUP BY serviceName
    )
SELECT *
FROM avg_msg_size_query

Query id: 7110f972-5008-4abd-abd1-624399c90754

Ok.

0 rows in set. Elapsed: 0.094 sec. Processed 181.00 thousand rows, 72.20 MB (1.92 million rows/s., 767.35 MB/s.)
Peak memory usage: 17.15 MiB.
```

See the `138.43 MB` and `72.20 MB` difference.

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
```diff
@@ -0,0 +1,171 @@
...
+		partition, consumerGroup, end, start, start, end, queueType, topic, partition,
+		consumerGroup, end, start, queueType, topic, partition, consumerGroup)
+	return query
+}
+
+func generateProducerSQL(start, end int64, topic, partition, queueType string) string {
+	query := fmt.Sprintf(`
+
+-- producer
+WITH
```
srikanthccv, line 103: Most of the consumer comments apply here as well.

srikanthccv (commented):
Some questions/comments

pkg/query-service/app/integrations/messagingQueues/kafka/consumerLag.md
vikrantgupta25: @srikanthccv this was used to apply legends to the queries. let's say i check the latency of the services and `group_by` service name then we get 2 columns service_name and the value column A. and if we have some custom legend for the query then we check if it is a value column and then apply the same based on query name passed as well

pkg/query-service/app/integrations/messagingQueues/kafka/consumerLag.md
shivanshuraj1333: Yes, basically in the design there's a dropdown for each of them, the values of which can be fetched via autocomplete APIs, once selected, those can be passed in this and the other messaging queue APIs.

shivanshuraj1333: I had played around with `autocomplete/attribute_values` and it works for the partition, topic, and consumer_group
eg
the get API 
```bash
api/v3/autocomplete/attribute_values?aggregateOperator=avg&dataSource=traces&aggregateAttribute=kafka_consumer_group_lag&attributeKey=group&searchText=&filterAttributeKeyDataType=string&tagType=tag
```
would return all the consumer groups like this
```json
{
    "status": "success",
    "data": {
        "stringAttributeValues": [
            "cg4",
            "cg1",
            "cg2",
            "cg3"
        ],
        "numberAttributeValues": null,
        "boolAttributeValues": null
    }
}
```

pkg/query-service/app/integrations/messagingQueues/kafka/consumerLag.md
shivanshuraj1333: Should I mark this resolved? or do I need to sanitise the existing query_range `tabular` response for messaging queues?

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
shivanshuraj1333: I'm not handling the cases in query, but given the time ranges will be short for the Messaging queue, handling them as `errored response` would be enough.
Though, a check can be added before running the query (in the code itself, not in the query), WDYT? Should I add a check?
If yes, then the max limit is not known (the user would usually search for a week's data, I can add a check for a negative value, but I don't think UI would be doing that (logically from the time selector). So not sure what should be a reasonable range check here, before running the query...

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
shivanshuraj1333: The Idea is to support users who setup following the guidelines, and in the guideline, we are making sure that the users install equal or higher version of the agents, if they do so, the fields are available by default.

pkg/query-service/app/integrations/messagingQueues/kafka/consumerLag.md
srikanthccv: This can be resolved because there is no user input needed.

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
srikanthccv: can you maybe pass the denominator directly?

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
srikanthccv: Personally, error% makes more sense to me. For example, if there were 20 errors in a minute but the processed records are 100k, the error as a percentage of the total processed tells a different story than errors/s.

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
srikanthccv: There could be an issue with average value calculations, specifically related to the `messaging.message.body.size` data.

Currently, if `messaging.message.body.size` does not exist in our numberTagMap, it returns zero. This can lead to incorrect averages because our query includes all rows, even when `messaging.message.body.size` doesn't exist. To fix this, we could add a check using `has(numberTagMap, 'messaging.message.body.size')`. This would ensure we only use rows where this data actually exists. You already have filters for `messaging.destination.*` and `messaging.kafka.consumer.group`. If these filters indirectly guarantee that `messaging.message.body.size` always exists, we might not need the extra check.

To be safe, though, it might be best to always include the has(...) check for `messaging.message.body.size`. This way, we're sure we're only using valid data in our calculations.

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
shivanshuraj1333: SGTM

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
shivanshuraj1333: done

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
shivanshuraj1333: done

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
shivanshuraj1333: done

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
shivanshuraj1333: done

srikanthccv (approved):

kienzx203 (commented):
ghhhh

PR #5545, opened 2024-07-24
fix: relative time param from the url not respected

vikrantgupta25:
### Summary

- the `relativeTime` param was not being read and respected in the time range logic, only the search start time and search end time params were being respected. 
- added the preferences correctly to respect the `relativeTime` preference , then `startTime` and `endTime` preference, post that local storage values and then the defaults for the page.

#### Related Issues / PR's

fixes - https://github.com/SigNoz/engineering-pod/issues/1486 

#### Screenshots

https://github.com/user-attachments/assets/9ca92696-bb1a-4cf9-a3bd-39a950bfb4f1

### DIFF
--- a/frontend/src/container/TopNav/DateTimeSelectionV2/index.tsx
+++ b/frontend/src/container/TopNav/DateTimeSelectionV2/index.tsx
@@ -23,7 +23,7 @@ import NewExplorerCTA from 'container/NewExplorerCTA';
 import dayjs, { Dayjs } from 'dayjs';
 import { useQueryBuilder } from 'hooks/queryBuilder/useQueryBuilder';
 import useUrlQuery from 'hooks/useUrlQuery';
-import GetMinMax from 'lib/getMinMax';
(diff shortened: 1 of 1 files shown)


YounixM (dismissed):

frontend/src/container/TopNav/DateTimeSelectionV2/index.tsx
```diff
@@ -404,16 +405,25 @@ function DateTimeSelection({
 		time: Time,
 		currentRoute: string,
 	): Time | CustomTimeType => {
+		// if the relativeTime param is present in the url give top most preference to the same
+		if (relativeTimeFromUrl != null) {
+			return relativeTimeFromUrl as Time;
```
YounixM, line 410: Let's check how it will behave if we get corrupted relative time. It should fallback to the next priority value

vikrantgupta25: added validity checks here

YounixM (approved):

ahmadshaheer (approved):

Issue #5426, opened 2024-07-04
Send selected service and time range for `top_level_operations` request in overview page


srikanthccv:
We currently fetch all the top-level operations for all the services which is bad performance and user experience when there are more top-level operations. Add the selected service name and time range (in millis) to the request.

PR #5552, opened 2024-07-24
chore: added service name and time params for top level operations

vikrantgupta25:
### Summary

- added serviceName , minTime and maxTime for the `top_level_operations` API call to fetch the relevant data rather than fetching for all the services. 

#### Related Issues / PR's

fixes https://github.com/SigNoz/signoz/issues/5426

#### Screenshots

NA


frontend/src/hooks/useGetTopLevelOperations.ts
```diff
@@ -10,7 +10,7 @@ type UseGetTopLevelOperations = (
 const useGetTopLevelOperations: UseGetTopLevelOperations = (queryKey) =>
 	useQuery<ServiceDataProps>({
 		queryKey,
-		queryFn: getTopLevelOperations,
+		queryFn: (): Promise<ServiceDataProps> => getTopLevelOperations({}),
```
vikrantgupta25, line 13: couldn't find any reference of this code being used, hence not passing the params here.

srikanthccv: This one https://github.com/SigNoz/signoz/blob/1281330c52877ca893f3f87b6e7cf488875e0f98/frontend/src/container/ServiceApplication/ServiceMetrics/index.tsx#L5-L35

vikrantgupta25: follow the caller chain and you ll end up in a component which is not being utilised anywhere

srikanthccv: If you look at the commit that introduced the change, it can be noticed that it is built to support a feature flag. It might not be used but it will be used. Ideally, `ServicesUsingMetrics` should have been used somewhere with the feature flag `USE_SPAN_METRIC`.

vikrantgupta25: let me check this again and see what was the idea behind the same

frontend/src/api/metrics/getTopLevelOperations.ts
```diff
@@ -1,7 +1,19 @@
 import axios from 'api';
 
-const getTopLevelOperations = async (): Promise<ServiceDataProps> => {
-	const response = await axios.post(`/service/top_level_operations`);
+interface GetTopLevelOperationsProps {
+	service?: string;
+	start?: number;
+	end?: number;
```
srikanthccv, line 6: Can they ever be undefined? If so, it going to be problematic

srikanthccv (approved):
One question, LGTM. Tested against the tenant which has issue.

frontend/src/api/metrics/getTopLevelOperations.ts
vikrantgupta25: no they cannot be, just added safety checks here as in the second case we didn't have the service so handled such cases. but they will not be undefined

frontend/src/api/metrics/getTopLevelOperations.ts
```diff
@@ -1,7 +1,19 @@
...
+}
+
+const getTopLevelOperations = async (
+	props: GetTopLevelOperationsProps,
+): Promise<ServiceDataProps> => {
+	const response = await axios.post(`/service/top_level_operations`, {
+		start: `${props.start}`,
```
YounixM, line 13: What is the fallback in case of params being undefined?

frontend/src/api/metrics/getTopLevelOperations.ts
```diff
@@ -1,7 +1,20 @@
...
+	service?: string;
+	start?: number;
+	end?: number;
+}
+
+const getTopLevelOperations = async (
+	props: GetTopLevelOperationsProps,
+): Promise<ServiceDataProps> => {
+	const response = await axios.post(`/service/top_level_operations`, {
+		start: !isNil(props.start) ? `${props.start}` : undefined,
```
vikrantgupta25, line 14: this undefined will make sure the param itself is not being sent in the api call

YounixM (approved):

srikanthccv (approved):

PR #5585, opened 2024-07-28
Schedule maintainence release changes

SagarRajput-7:
### Summary

Fixes done -
* Repeats every feels like text box with cursor blinking. And doesn't close on hover out. Overall bad UX.
* Search for alert doesn't work
* Duration should not accept negative values
* Write a note that says "Leave empty to silence all alerts". UI should accept and send empty array as request payload to silence all alerts
* In recurrence schedule, the start and end are incorrectly saved & displayed without timezone considerations. They appear to not use the "recurrence" object values. Also the start and end are different in the "schedule" object and "schedule.recurrence" object
* Mark which fields are mandatory
     - Name, start and timezone are always mandatory
     - End is mandatory if the schedule is fixed
     - If the schedule is recurring
     - end is optional
     - duration is mandatory
* "1h30m" should be "90m" instead of "1h"
* Doesn't send repeat type for recurrence schedule edit.
* Delete window says "view" instead of "schedule"
* Timezone consideration for starttime and endtime

 Tested P0s - 

> fixed: 
> - New
>     - changed timezone
> - Edit
>     - changed timezone
> 
> Recuring:
> - New
>     - changed timezone
>     - daily
>     - weekly
>     - monthly
>     - with end time
>     - without end time
> - Edit
>     - changed timezone
>     - daily
>     - weekly
>     - monthly
>     - with end time
>     - without end time

#### Screenshots

https://github.com/user-attachments/assets/adf4d423-6a73-4fef-8d8d-3eaea053320f


frontend/src/container/PlannedDowntime/PlannedDowntime.tsx
```diff
@@ -128,17 +134,19 @@ export function PlannedDowntime(): JSX.Element {
...
-					initialValues={initialValues}
-					isError={isError}
-					isLoading={isLoading}
-					isOpen={isOpen}
-					setIsOpen={setIsOpen}
-					refetchAllSchedules={refetchAllSchedules}
-					isEditMode={isEditMode}
-					form={form}
-				/>
+				{isOpen && (
```
vikrantgupta25, line 137: why is this needed ?

frontend/src/container/PlannedDowntime/PlannedDowntimeForm.tsx
```diff
@@ -309,62 +305,65 @@ export function PlannedDowntimeForm(
 				</Form.Item>
 				<Form.Item
 					label="Repeats every"
-					name="recurrenceSelect"
-					required={false}
+					name={['recurrence', 'repeatType']}
```
vikrantgupta25, line 308: can you help me understand why these two names here ?

SagarRajput-7: This is required for nested approach, like this would be treated as `recurrence.repeatType` as this is format supported by payload and list call also

frontend/src/container/PlannedDowntime/PlannedDowntime.tsx
SagarRajput-7: As the component and logic added under this is dependent on this flag only, and few state settings and logic I wanted to segregate all those under a separate file.

srikanthccv:
1. For recurring schedules, Timeframe and Repeats show different time values. This is because the payload is incorrect for the recurring schedule. Example payload frontend is sending for schedule from 29th July 2:40 to 31st July 2:40

```json
{
  "alertIds": [
    "66"
  ],
  "name": "daily recurrence with end time",
  "schedule": {
    "startTime": "2024-07-29T14:40:00Z",
    "timezone": "Asia/Kolkata",
    "endTime": "2024-07-31T14:40:00Z",
    "recurrence": {
      "duration": "2h",
      "endTime": "2024-07-31T09:10:00.572Z",
      "startTime": "2024-07-29T09:10:00.221Z",
      "repeatType": "daily"
    }
  }
}
```

Screenshot 2024-07-29 at 2 25 22 PM

2. For recurring schedules with end time defined, in edit mode, Repeats every shows Does not repeat and saving it makes the schedule fixed one and erases the recurring data.

Screenshot 2024-07-29 at 2 45 16 PM

3. To be consistent with the rest of the product, order the list items by updated time.

SagarRajput-7:
FIxed all the above issues

srikanthccv (approved):
LGTM, All the issues mentioned above and the cases shared are working

vikrantgupta25 (approved):

YounixM (approved):

PR #5599, opened 2024-07-29
feat: update logEvent to silently handle errors

YounixM:


frontend/src/api/common/logEvent.ts
```diff
@@ -1,4 +1,4 @@
-import axios from 'api';
+import { ApiBaseInstance as axios } from 'api';
```
vikrantgupta25, line 1: as axios shouldn't be required here ?

vikrantgupta25 (approved):
LGTM

PR #5607, opened 2024-07-30
feat: faster traceID based filtering

makeavish:
### Summary

Faster traceID based filtering for traces by decreasing search space of query range API when traceID filter is found along with Equal or IN operators 

### Cases handled
- Multiple queries with/without traceIDs
- Limit on groupBy
- IN or equal operator

### DIFF
--- a/pkg/query-service/app/traces/v3/utils.go
+++ b/pkg/query-service/app/traces/v3/utils.go
@@ -35,6 +35,7 @@ func TraceIdFilterUsedWithEqual(params *v3.QueryRangeParamsV3) (bool, []string) 
 					val := item.Value
 					val, err = utils.ValidateAndCastValue(val, item.Key.DataType)
 					if err != nil {
+						zap.L().Error("invalid value for key", zap.String("key", item.Key.Key), zap.Error(err))
 						return false, []string{}
 					}
 					if val != nil {
--- a/pkg/query-service/app/traces/v3/utils.go
+++ b/pkg/query-service/app/traces/v3/utils.go
@@ -1,0 +1,182 @@
+package v3
+
+import (
+	"strconv"
+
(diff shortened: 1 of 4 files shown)


pkg/query-service/app/traces/v3/utils.go
```diff
@@ -0,0 +1,182 @@
...
+			for _, item := range query.Filters.Items {
+
+				if item.Key.Key == "traceID" && (item.Operator == v3.FilterOperatorIn ||
+					item.Operator == v3.FilterOperatorEqual) {
+					traceIdFilterUsed = true
+					// validate value
+					var err error
+					val := item.Value
+					val, err = utils.ValidateAndCastValue(val, item.Key.DataType)
+					if err != nil {
```
srikanthccv, line 37: May be log an error here?

srikanthccv (approved):
We're using the earliest and latest timestamps across all traceIDs for each query, rather than storing separate timestamps for every single traceID. This might not be ideal in every situation, but I assume it's rare to need multiple queries with different traceIDs. Most of the time, people just want to list queries for one specific trace?

And this probably doesn't matter for alerts? Setting an alert based on a trace ID doesn't really make sense.

Looks good to me!

makeavish:
> We're using the earliest and latest timestamps across all traceIDs for each query, rather than storing separate timestamps for every single query. This might not be ideal in every situation, but I assume it's rare to need multiple queries with different traceIDs. Most of the time, people just want to list queries for one specific trace?
> 
> And this probably doesn't matter for alerts? Setting an alert based on a trace ID doesn't really make sense.
> 
> Looks good to me!

Yes, it's intentional to use multiple traceIDs across queries as adding too many traceIDs is not common. Yes doesn't makes sense to have alerts on TraceID as it's unique ID for short time period which can't occur in future.

Thanks for review!

Issue #4978, opened 2024-05-10
Improve error message shown in the UI


ankitnayan:
Need better and relevant error messages to be shown to the users

<img width="1173" alt="Screenshot 2024-05-10 at 10 34 52 PM" src="https://github.com/SigNoz/signoz/assets/12460410/dbd4cc0c-6f47-448e-8369-2953dcd09081">

YounixM:
@ahmadshaheer : Please sync with @srikanthccv if you have any queries post going through the discussion threads.

ahmadshaheer:
@YounixM, Reached out to @srikanthccv regarding this issue. he mentioned that this ticket needs backend changes. I and Srikanth are going to have a meeting on Monday (5th of Aug) to further discuss it.

makeavish:
@srikanthccv / @ahmadshaheer Any update on this?
Feedback by another ICP user:

![Image](https://github.com/user-attachments/assets/1549ccdc-a077-4537-b651-966aee816c7a)

ahmadshaheer:
> @srikanthccv / @ahmadshaheer Any update on this? Feedback by another ICP user:
> 
[...]

PR #5628, opened 2024-08-02
chore: improve error message readability

srikanthccv:
### Summary

Part of https://github.com/SigNoz/signoz/issues/4978


pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -4889,7 +4889,7 @@ func (r *ClickHouseReader) GetTimeSeriesResultV3(ctx context.Context, query stri
 
 	if err != nil {
 		zap.L().Error("error while reading time series result", zap.Error(err))
-		return nil, err
+		return nil, errors.New(err.Error())
```
nityanandagohain, line 4892: confused here, apart from that it looks good.

nityanandagohain (approved):

pkg/query-service/app/clickhouseReader/reader.go
nityanandagohain: got it, this is to remove stacktrace.

nityanandagohain (approved):

Issue #5644, opened 2024-08-05
Aggregate interval changes automatically when changing panel types


ankitnayan:
Aggregation interval of a query changes back from 3600 to 60 seconds when you change panel types from
timeseries -> bar -> timeseries

Issue #5575, opened 2024-07-26
Issues with changing panel types


srikanthccv:
- [ ] Changing the panel type from time series to table doesn't change the underlying operators; See https://app.intercom.com/a/inbox/wmkurvb7/inbox/admin/6745296/conversation/4337#part_id=comment-4337-9442
- [ ] https://www.loom.com/share/1b46ce272d5e41bca2846e3b5dd977ae

Expectation https://github.com/SigNoz/signoz/issues/2410

PR #5648, opened 2024-08-06
fix: clean out the panel type change attribute dependency

vikrantgupta25:
### Summary

- clean out the `attribute to datasource to panel type map` and added the missing attributes 
- clean out the `handleQueryChange`  function as well 
- details here - https://www.notion.so/signoz/updated-panel-types-attributes-19feddcce1c545c489932546709c4dab#56fcc2f2c2784e7abdde25613a572752

#### Related Issues / PR's

fixes https://github.com/SigNoz/engineering-pod/issues/1620 
fixes https://github.com/SigNoz/signoz/issues/5644
fixes https://github.com/SigNoz/signoz/issues/5575

#### Screenshots

NA


frontend/src/container/NewWidget/utils.ts
```diff
@@ -419,7 +445,6 @@ export function handleQueryChange(
 					set(tempQuery, 'offset', 0);
 					set(tempQuery, 'pageSize', 10);
 				} else if (tempQuery.aggregateOperator === 'noop') {
-					set(tempQuery, 'aggregateOperator', 'count');
```
srikanthccv, line 422: Why was this added and does it break anything if we remove it now?

vikrantgupta25: when shifting from any panel other than list type we modify the operator to `noop` for list panel to work, we do not update the superset query here. so when moving back to any other panel from list panel the older operator should be selected by default we should not update it back to `count` here as it could have been anything.

vikrantgupta25:
- i found one issue with moving from list panel to other panel types when list panel was the first one to be opened. fixing this.

vikrantgupta25:
> i found one issue with moving from list panel to other panel types when list panel was the first one to be opened. fixing this.

this is fixed.

srikanthccv (approved):

YounixM (approved):

PR #5651, opened 2024-08-06
chore: enable caching for all panel types in metrics v4

srikanthccv:
#### Screenshots

NA

### DIFF
--- a/pkg/query-service/app/queryBuilder/query_builder.go
+++ b/pkg/query-service/app/queryBuilder/query_builder.go
@@ -324,6 +324,10 @@ func (c *cacheKeyGenerator) GenerateKeys(params *v3.QueryRangeParamsV3) map[stri
 
 	// Use query as the cache key for PromQL queries
 	if params.CompositeQuery.QueryType == v3.QueryTypePromQL {
+		if params.CompositeQuery.PanelType != v3.PanelTypeGraph {
+			return keys
+		}
+
 		for name, query := range params.CompositeQuery.PromQueries {
 			keys[name] = query.Query
 		}
--- a/pkg/query-service/app/queryBuilder/query_builder.go
+++ b/pkg/query-service/app/queryBuilder/query_builder.go
@@ -387,6 +387,18 @@ func (c *cacheKeyGenerator) GenerateKeys(params *v3.QueryRangeParamsV3) map[stri
 		} else if query.Expression == queryName && query.DataSource == v3.DataSourceMetrics {
 			var parts []string
 
+			// what is this condition checking?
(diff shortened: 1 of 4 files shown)


pkg/query-service/app/queryBuilder/query_builder.go
```diff
@@ -322,13 +322,12 @@ func isLogExpression(expression *govaluate.EvaluableExpression, params *v3.Query
 func (c *cacheKeyGenerator) GenerateKeys(params *v3.QueryRangeParamsV3) map[string]string {
 	keys := make(map[string]string)
 
-	// For non-graph panels, we don't support caching
-	if params.CompositeQuery.PanelType != v3.PanelTypeGraph {
-		return keys
-	}
-
```
srikanthccv, line 329: This is the main change. We used to return empty keys for every non-graph panel type. We still do it, but only when

1. query type is promql
2. The data source is logs

Now when it comes to metrics, things change

1. v3 - same as logs, only graph queries can be cached
2. v4 - everything can be cached

srikanthccv: We can't enable caching for other panels for logs and traces due to how their queries work. Unlike metrics, log queries don't return time series data that can be easily aggregated in query service. Instead, they fetch a single aggregated value for the entire time window. This makes it impossible to cache results for future use, like p95, avg, etc. We can support caching for simpler operations like sum or count, but we'd need to implement that first before enabling it here.

srikanthccv: Also there's no v4 for logs or traces - they work the same way.

ankitnayan: ok

pkg/query-service/app/queryBuilder/query_builder.go
```diff
@@ -383,6 +387,10 @@ func (c *cacheKeyGenerator) GenerateKeys(params *v3.QueryRangeParamsV3) map[stri
 		} else if query.Expression == queryName && query.DataSource == v3.DataSourceMetrics {
 			var parts []string
 
+			if params.Version != "v4" && params.CompositeQuery.PanelType != v3.PanelTypeGraph {
```
nityanandagohain, line 390: please add a code comment here stating that, for metring if the version is v3 then we only support caching for Graph.

Will be easier to look back

srikanthccv: Added in https://github.com/SigNoz/signoz/pull/5651/commits/6d845923899135c4f7df6a161164deaf2ddb717b

nityanandagohain (approved):

PR #5670, opened 2024-08-11
chore: standardize env for query service

### LINKED ISSUE #5266: Standardize ENV for query service
Right now, there's no standard way to configure environment variables for query service, for helm installation, the below ENVs are populated but similarly to running locally the ENVs are not well defined, it should match with the docker counter part.

from a container
```yaml
         env:
         - name: CLICKHOUSE_HOST
           value: my-release-clickhouse
         - name: CLICKHOUSE_PORT
           value: "9000"
         - name: CLICKHOUSE_HTTP_PORT
           value: "8123"
         - name: CLICKHOUSE_CLUSTER
           value: cluster
         - name: CLICKHOUSE_USER
           value: admin
         - name: CLICKHOUSE_PASSWORD
           value: 27ff0399-0d3a-4bd8-919d-17c2181e6fb9
         - name: CLICKHOUSE_SECURE
           value: "false"
         - name: STORAGE
           value: clickhouse
         - name: ClickHouseUrl
           value: tcp://my-release-clickhouse:9000/?username=admin&password=27ff0399-0d3a-4bd8-919d-17c2181e6fb9
         - name: ALERTMANAGER_API_PREFIX
           value: http://my-release-signoz-alertmanager:9093/api/
```

but the same is missing for locally running the query service and the parameters are hardcoded in strings 
eg
https://github.com/SigNoz/signoz/blob/develop/pkg/query-service/app/server.go#L113
https://github.com/SigNoz/signoz/blob/develop/pkg/query-service/app/clickhouseReader/reader.go#L146

slayer321:
### Summary

Standardize ENV variable for query service

#### Related Issues / PR's
fixes #5266

### DIFF
--- a/ee/query-service/app/server.go
+++ b/ee/query-service/app/server.go
@@ -10,7 +10,6 @@ import (
 	"net"
 	"net/http"
 	_ "net/http/pprof" // http profiler
-	"os"
 	"regexp"
 	"time"
 
@@ -141,7 +140,7 @@ func NewServer(serverOptions *ServerOptions) (*Server, error) {
 	readerReady := make(chan bool)
 
 	var reader interfaces.DataConnector
-	storage := os.Getenv("STORAGE")
+	storage := baseconst.StorageType
 	if storage == "clickhouse" {
 		zap.L().Info("Using ClickHouse as datastore ...")
 		qb := db.NewDataConnector(
--- a/ee/query-service/app/server.go
+++ b/ee/query-service/app/server.go
@@ -27,6 +27,7 @@ import (
 	"go.signoz.io/signoz/ee/query-service/integrations/gateway"
 	"go.signoz.io/signoz/ee/query-service/interfaces"
 	baseauth "go.signoz.io/signoz/pkg/query-service/auth"
+	"go.signoz.io/signoz/pkg/query-service/config"
 	"go.signoz.io/signoz/pkg/query-service/migrate"
 	"go.signoz.io/signoz/pkg/query-service/model"
 	v3 "go.signoz.io/signoz/pkg/query-service/model/v3"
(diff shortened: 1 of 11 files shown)


slayer321:
@shivanshuraj1333 Can you please review it.

slayer321:
Ok, make sense. 
About  handle errors  part do we want to have something like below ?

```
 // validateConfig checks if the configuration is valid
func validateConfig(config Config) error {
	if config.SignozJwtSecret == "" {
		return errors.New("missing SIGNOZ_JWT_SECRET")
	}
	if config.StorageType == "" {
		return errors.New("missing STORAGE_TYPE")
	}
	if config.ClickHouseUrl == "" {
		return errors.New("missing CLICKHOUSE_URL")
	}
	if config.SmtpHost == "" {
		return errors.New("missing SMTP_HOST")
	}
	if config.SmtpPort == 0 {
		return errors.New("invalid SMTP_PORT")
	}
	if config.SmtpUsername == "" {
		return errors.New("missing SMTP_USERNAME")
	}
	if config.SmtpPassword == "" {
		return errors.New("missing SMTP_PASSWORD")
	}
	if config.SmtpFrom == "" {
		return errors.New("missing SMTP_FROM")
	}
	return nil
}
```

shivanshuraj1333:
No,
before accessing the flags in the code, handle the error there using existing error handling (mostly log the error).

Aeolun:
Shouldn't this check and alert all errors on load? I don't think the system is much use without a `ClickhouseServer` variable defined.

grandwizard28:
Hi @slayer321,
Thank you for the PR! We are however in the middle of a major codebase revamp where we will also be standardising config as per https://github.com/SigNoz/signoz/issues/6805.

As part of the config being created, we are also going to validate and fail fast. That should take care of @Aeolun's point!

PR #5679, opened 2024-08-12
chore: update query builder to use 5min/30min aggregation tables

srikanthccv:
### Summary

Uses changes from https://github.com/SigNoz/signoz-otel-collector/pull/361

1. Table Selection and Aggregation Strategy

   We have two pre-aggregated tables:
   a) 5-minute aggregation
   b) 30-minute aggregation

   The choice of table depends on the query's time range:

   - Range < 1 day: No pre-aggregation table used (step interval < 5 minutes)
   - 1 day < Range < 1 week: Use 5-minute aggregation table (5 minutes < step interval < 33 minutes)
   - Range > 1 week: Use 30-minute aggregation table (step interval > 30 minutes)

2. Column Selection

   The pre-aggregated tables contain columns for min, max, sum, and count of values from raw samples.

   This PR changes the column selection logic:
   - Instead of always using the `value` column
   - We now select the appropriate column (`min`, `max`, `sum`, or `count`) based on the aggregation needs

These changes optimize query performance by using pre-aggregated data when possible and selecting the most appropriate columns for aggregation.


pkg/query-service/app/metrics/v4/helpers/sub_query.go
```diff
@@ -13,31 +13,129 @@ import (
...
+				aggregationColumn = "sum(value)"
+			}
+		case constants.SIGNOZ_SAMPLES_V4_AGG_5M_TABLENAME, constants.SIGNOZ_SAMPLES_V4_AGG_30M_TABLENAME:
+			switch mq.TimeAggregation {
+			case v3.TimeAggregationAnyLast:
+				aggregationColumn = "anyLast(last)"
+			case v3.TimeAggregationSum:
+				aggregationColumn = "sum(sum)"
+			case v3.TimeAggregationAvg:
+				aggregationColumn = "sum(sum) / sum(count)"
```
ankitnayan, line 124: more efficient to store than to do twice the work in query?

ankitnayan:
TTL also needs to be updated?

pkg/query-service/app/metrics/v4/helpers/sub_query.go
ankitnayan: This would be row level operation, so should be good I guess

srikanthccv: The other way would be storing the `avgState` which internally does the [same](https://github.com/ClickHouse/ClickHouse/blob/ae5223854f71361a4cafdd6bafad3e2cc770432b/src/AggregateFunctions/AggregateFunctionAvg.h#L104-L128) and `-Merge` at the query time. There won't be any noticeable difference.

srikanthccv:
For those customers with a different TTL than default, it will be updated.

ankitnayan (approved):

shivanshuraj1333:
I have one concern wrt user experience. Some edge cases are missing with this.
Say there are memory spikes for certain requests which are visible in 5 min window, but they are Averaged out in 30 min window. AFAIU the logic, users won't be able to see such spikes for data older than 5 mins.

srikanthccv:
> AFAIU the logic, users won't be able to see such spikes for data older data.

@shivanshuraj1333 we maintain mix, max, and count other than average. They can choose to use `max` aggregation if they are interested in seeing the spike.

srikanthccv:
The default time aggregation for gauge metrics is avergage, but we also allow them to change.

shivanshuraj1333 (approved):

PR #5684, opened 2024-08-13
feat: added blur event to having input in query section

rahulkeswani101:
#### Related Issues / PR's

https://github.com/SigNoz/engineering-pod/issues/1364

https://github.com/user-attachments/assets/b188f85f-7546-48fc-80ce-aca92771da2c


frontend/src/container/QueryBuilder/filters/HavingFilter/HavingFilter.tsx
```diff
@@ -198,6 +198,18 @@ export function HavingFilter({
 		resetChanges();
 	};
 
+	const handleBlur = useCallback((): void => {
+		if (searchText) {
+			const isValidSearch = isValidHavingValue(searchText);
```
SagarRajput-7, line 203: Just confirm once that the only case when we want this blur event to trigger is when we have searchText right? Rest lgtm

SagarRajput-7 (approved):

YounixM (dismissed):

srikanthccv:
@rahulkeswani101 we should only make the chip only if the entered input is complete having clause. For example, if I hover out before selecting any operator it creates this which is invalid.

Screenshot 2024-09-04 at 1 57 57 PM

rahulkeswani101:
> @rahulkeswani101 we should only make the chip only if the entered input is complete having clause. For example, if I hover out before selecting any operator it creates this which is invalid.
> 
> Screenshot 2024-09-04 at 1 57 57 PM


@srikanthccv Can you Please check this and let me know if any other changes are required?

YounixM (approved):

Issue #5609, opened 2024-07-31
Logs: Query Builder: New Suggestions API


raj-k-singh:
For next version of Logs QB suggestions

PR #5703, opened 2024-08-15
Chore: qs filter suggestions: example queries for multiple top attributes

raj-k-singh:
### Summary

Updates filter suggestion API to return example queries for multiple suggested attributes.
v0 returned an example query for only the first suggested attribute

#### Related Issues / PR's

Contributes to https://github.com/SigNoz/signoz/issues/5609

#### Screenshots

Before
Screenshot from 2024-08-15 20-03-07

After
Screenshot from 2024-08-15 20-00-35

Logs Query Builder


srikanthccv (approved):

nityanandagohain (approved):
while not a blocker/urgent thing, but can use cache in future since it's already available. With ttl of around half an hour for getValuesForLogAttributes

PR #5715, opened 2024-08-18
fix: do not use relative URLs for ws connections

vikrantgupta25:
### Summary

the websockets api recently started respecting the domain and support for relative URLs so we can't rely on every user being on the updated version hence added the extra handling for the same 

#### Screenshots

NA

### DIFF
--- a/frontend/src/api/common/getQueryStats.ts
+++ b/frontend/src/api/common/getQueryStats.ts
@@ -18,7 +18,13 @@ function getURL(baseURL: string, queryId: string): URL | string {
 		return `${baseURL}/ws/query_progress?q=${queryId}`;
 	}
 	const url = new URL(`/ws/query_progress?q=${queryId}`, window.location.href);
-	url.protocol = 'wss';
+
+	if (window.location.protocol === 'http:') {
+		url.protocol = 'ws';
+	} else {
+		url.protocol = 'wss';
+	}
+
 	return url;
 }
 
--- a/frontend/src/api/common/getQueryStats.ts
+++ b/frontend/src/api/common/getQueryStats.ts
@@ -16,10 +16,16 @@ export function getQueryStats(props: GetQueryStatsProps): void {
 	const { queryId, setData } = props;
 
 	const token = getLocalStorageApi(LOCALSTORAGE.AUTH_TOKEN) || '';
-	const socket = new WebSocket(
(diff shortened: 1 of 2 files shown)


frontend/src/api/common/getQueryStats.ts
```diff
@@ -12,14 +13,27 @@ interface GetQueryStatsProps {
...
 export function getQueryStats(props: GetQueryStatsProps): void {
 	const { queryId, setData } = props;
 
 	const token = getLocalStorageApi(LOCALSTORAGE.AUTH_TOKEN) || '';
-	const socket = new WebSocket(
-		`${ENVIRONMENT.wsURL}/api/v3/query_progress?q=${queryId}`,
-		token,
-	);
+
+	// https://github.com/whatwg/websockets/issues/20 reason for not using the relative URLs
```
YounixM, line 33: We can remove this comment now.

YounixM (dismissed):

frontend/src/api/common/getQueryStats.ts
vikrantgupta25: any reason for the same ? this states the reason for not using the relative URL, the initial code is logically correct but since the ws constructor doesn't take care of it in older browsers we are using this brute approach. so would be easy for someone else reading the code to follow.

YounixM: nvm. Additional info is always good 👍🏼

frontend/src/api/common/getQueryStats.ts
```diff
@@ -12,14 +13,27 @@ interface GetQueryStatsProps {
...
 
+function getURL(baseURL: string, queryId: string): URL | string {
+	if (baseURL && !isEmpty(baseURL)) {
+		return `${baseURL}/api/v3/query_progress?q=${queryId}`;
+	}
+	const url = new URL(
+		`/api/v3/query_progress?q=${queryId}`,
+		window.location.href,
+	);
+	url.protocol = 'wss';
```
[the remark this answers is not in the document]
vikrantgupta25: the local envs setup will be returned from the first `if` block , it won't reach here

grandwizard28:
Works on old chrome versions, approving @YounixM

grandwizard28 (approved):

YounixM (approved):

PR #5710, opened 2024-08-16
feat(binary): combine components into a single binary

grandwizard28:
#### Related Issues / PR's
- [ ] https://github.com/SigNoz/signoz/issues/3001
- [ ] https://github.com/SigNoz/signoz/issues/392
- [ ] https://github.com/SigNoz/signoz/issues/357
- [ ] https://github.com/SigNoz/signoz/issues/1236
- [ ] https://github.com/SigNoz/signoz/issues/241 
- [ ] https://github.com/SigNoz/signoz/issues/657
- [ ] https://github.com/SigNoz/signoz/issues/221
- [ ] https://github.com/SigNoz/signoz/issues/220
- [ ] https://github.com/SigNoz/signoz/issues/284
- [ ] https://github.com/SigNoz/signoz/issues/464
- [ ] https://github.com/SigNoz/signoz/issues/649
- [ ] https://github.com/SigNoz/signoz/issues/1015
- [ ] https://github.com/SigNoz/signoz/issues/1498
- [ ] https://github.com/SigNoz/signoz/issues/2028
- [ ] https://github.com/SigNoz/signoz/issues/2499
- [ ] https://github.com/SigNoz/signoz/issues/5246
#### Screenshots

NA

PR #5727, opened 2024-08-20
feat(packages): add first dedicated confmap, config, version and instrumentation packages

grandwizard28:
### Summary

A config package based on https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/confmap.go for signoz.

#### Related Issues / PR's

This is a part of https://github.com/SigNoz/signoz/pull/5710


pkg/confmap/config.go
```diff
@@ -0,0 +1,9 @@
+package confmap
```
grandwizard28, line 1: @srikanthccv can we find a better place for this interface?

pkg/instrumentation/build.go
```diff
@@ -0,0 +1,9 @@
+package instrumentation
```
grandwizard28, line 1: @srikanthccv can we find a better place for these settings?

srikanthccv: Let's move the build info to `pkg/version/build`?

pkg/instrumentation/instrumentation.go
```diff
@@ -0,0 +1,96 @@
...
+	// Create a new resource with default detectors.
+	// The upstream contrib repository is not taking detectors into account.
+	// We are, therefore, using some sensible defaults here.
+	resource, err := sdkresource.New(
+		ctx,
+		sdkresource.WithContainer(),
+		sdkresource.WithFromEnv(),
+		sdkresource.WithProcess(),
+		sdkresource.WithTelemetrySDK(),
+		sdkresource.WithHost(),
```
srikanthccv, line 44: I suggest we remove process and telemetrysdk and host. We can always add them later if we need them because they are rarely used but consume a non-trivial amount of space. The container and env should be enough.

pkg/confmap/provider/signozenvprovider/provider.go
```diff
@@ -0,0 +1,94 @@
...
+	return confmap.NewProviderFactory(newProvider)
+}
+
+func newProvider(settings confmap.ProviderSettings) confmap.Provider {
+	return &provider{
+		logger: settings.Logger,
+	}
+}
+
+func (provider *provider) Retrieve(_ context.Context, uri string, _ confmap.WatcherFunc) (*confmap.Retrieved, error) {
```
srikanthccv, line 45: Why does it only support env? I think it is easier to use and manage with the file provider. Not everything can be configured with env, for instance, how can I configure different readers for the meter provider?

pkg/confmap/config.go
srikanthccv: Maybe move to `pkg/config/config`?

srikanthccv (dismissed):

pkg/instrumentation/build.go
grandwizard28: Moved this to `pkg/version`

pkg/confmap/config.go
grandwizard28: Some circular import issue might come up! Letting it be for now

pkg/instrumentation/instrumentation.go
grandwizard28: Removed process and sdk. I guess host would help in debugging?

srikanthccv (approved):

therealpandey:
Part of https://github.com/SigNoz/platform-pod/issues/442

PR #5728, opened 2024-08-20
feat: consume the new search bar

vikrantgupta25:
### Summary

this pr adds the consumption for the new search bar v2 in the logs explorer page 

#### Related Issues / PR's

contributes to https://github.com/SigNoz/signoz/issues/5150

#### Screenshots

NA


frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/QueryBuilderSearchV2.tsx
```diff
@@ -324,6 +328,21 @@ function QueryBuilderSearchV2(
 
 				if (isMulti) {
 					const { tagKey, tagOperator, tagValue } = getTagToken(searchValue);
+					console.log(tagValue, value);
```
nityanandagohain, line 331: remove

nityanandagohain (changes requested):
Please add test cases to alteast start in that direction, else it will be treated as a secondary citizen and these manual testing will get monotonous.

Checks
- [ ]  static field data type is not shown eg:- severity_text, severity_number 
- [ ] Not able to update the value of operator after running a query ( replaces with old value) ( change the value wait for the suggestion to update then it works but rapid typing doesn't)
- [x] Contains and other filters working properly
- [ ] Filtering from the detail page by clicking on attribute name not working
- [ ] Filter in , filter out from detail page doesn't update the search bar. (refreshing the window makes them appear)
- [ ] There is a flicker after selecting a key value pair.
- [ ] The filter doesn't get cleared if I got to some other page like alerts and come back

Questions ?
* if attr named `x` is of both float and string, then if I type x then is string type selected by default ( that what it seems to be the case)?

Suggestions
* the tag/resource infor feels too far to the right
* the suggestions is scrollable ( the user won't know unless they look extreme right)

frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/QueryBuilderSearchV2.tsx
vikrantgupta25: done

vikrantgupta25:
@nityanandagohain fixed the 2nd point as discussed, remaining will move to separate issues

nityanandagohain (dismissed):
lgtm, please create issue for test and keep it in high priority so that we don't need to go through these tests again for one of the next pr's

vikrantgupta25:
@nityanandagohain also addressed the same name different data type issue. if no operator is entered until we show both and do not tokenise and once the operator comes we then pick on priority.

vikrantgupta25:
created one P0 task on my name for the test cases - https://github.com/SigNoz/engineering-pod/issues/1724 cc @nityanandagohain

YounixM (approved):

nityanandagohain (approved):
LGTM, @srikanthccv do skim through the feature flag part once if somethng looks suspecious/missing, I have less context on it.

PR #5743, opened 2024-08-21
feat(web): add web package

grandwizard28:
### Summary

Add a web package for serving frontend

#### Related Issues / PR's

https://github.com/SigNoz/signoz/pull/5710


pkg/web/web.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	return &Web{
+		logger: logger.Named("go.signoz.io/pkg/web"),
+		cfg:    cfg,
+	}, nil
+}
+
+func (web *Web) AddToRouter(router *mux.Router) error {
+	cache := middleware.NewCache(7 * 24 * time.Hour)
```
srikanthccv, line 57: The `NewCache` reads as if we are maintaining the cache rather than the middleware that controls the cache headers. Can every web asset be cached?

srikanthccv (approved):

pkg/web/web.go
grandwizard28: Yeah every web asset can be cached. If we were doing SSR, it would not have been possible. A static build does not change once built.

I hope I understood the question correctly!

therealpandey:
Part of https://github.com/SigNoz/platform-pod/issues/442

PR #5752, opened 2024-08-22
fix: handled defaultTraceSelected for traces list view

SagarRajput-7:
### Summary

Issue - The user added a column with a column name duplicate to what SigNoz provide by default due to which the Column config detail got overridden with the user details, and we are getting `N/A` as we are getting `isColumn` as false, which for our default cases should always be true as we have those default column defined on the DB and hence they exist.

Solution - Have identified & overridden the column config detail for the default cases that SigNoz provide like 'durationNano', 'name' etc.

#### Related Issues / PR's

https://github.com/SigNoz/engineering-pod/issues/1697

#### Screenshots

Before:
-------------
https://github.com/user-attachments/assets/f2359fc6-2f56-408c-9773-709821f9da70

After:
-------------
https://github.com/user-attachments/assets/a82867ae-1106-4b53-b131-3fbb167e3745


YounixM:
@SagarRajput-7 : please add the change details.

vikrantgupta25 (approved):
LGTM

makeavish (approved):

YounixM (approved):

PR #5760, opened 2024-08-23
feat: enable macro

nityanandagohain:
P.R to enable macro, This will be needed w.r.t new schema


srikanthccv (approved):
Please make it clear it in the next release notes that people should enable macros in their docker setups. Some existing OSS users don't use our `deploy` rather they maintain their own and copy changes when required.

nityanandagohain:
Yeah will add it to the release notes . cc @prashant-shahi

PR #5770, opened 2024-08-26
fix: copy to clipboard not copying complete value in case of numbers

YounixM:
### Summary

https://drive.google.com/file/d/1l3I1aFpcLu9zFsRkBVJiveaVn5BSNzhl

### DIFF
--- a/frontend/src/components/Logs/CopyClipboardHOC.tsx
+++ b/frontend/src/components/Logs/CopyClipboardHOC.tsx
@@ -4,6 +4,7 @@ import { ReactNode, useCallback, useEffect } from 'react';
 import { useCopyToClipboard } from 'react-use';
 
 function CopyClipboardHOC({
+	entityKey,
 	textToCopy,
 	children,
 }: CopyClipboardHOCProps): JSX.Element {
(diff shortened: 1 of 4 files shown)


frontend/src/components/Logs/CopyClipboardHOC.tsx
```diff
@@ -4,18 +4,23 @@ import { ReactNode, useCallback, useEffect } from 'react';
...
 	textToCopy,
 	children,
 }: CopyClipboardHOCProps): JSX.Element {
 	const [value, setCopy] = useCopyToClipboard();
 	const { notifications } = useNotifications();
 	useEffect(() => {
 		if (value.value) {
+			const key = entityKey || '';
+
+			const notificationMessage = `${key} copied to clipboard`;
```
vikrantgupta25, line 17: in case the key is not present the c of copied should be in Caps ?

frontend/src/container/LogDetailedView/FieldRenderer.styles.scss
```diff
@@ -18,6 +18,7 @@
 
 	.tags {
 		display: flex;
-		gap: 8;
+		flex-wrap: wrap;
```
vikrantgupta25, line 21: we should ideally avoid `flex-wrap` as there can be \n and other formatters in the log lines but anyways we currently do not take this into account for raw and other views as well

vikrantgupta25 (approved):
minor NIT comments

frontend/src/container/LogDetailedView/FieldRenderer.styles.scss
YounixM: This is only for the tags in attributes key columns @vikrantgupta25

PR #4573, opened 2024-02-19
fix: address gaps in alert to notification link

srikanthccv:
### Summary

This PR addresses some gaps found in the alert notification to the explorer link.

- The double encoding is no longer needed (implementation changed with new design changes). @YounixM I hope this remains unchanged going forward.
- Keep the original label format for where clause preparation
- Move the links from labels to annotations. The main difference is labels is used for grouping and we want to group alerts with the same label set, but different timestamps, together
- Fix the source link (messaging tools truncate the lengthy contents).

## Summary by CodeRabbit

- **New Features**
	- Enhanced the "Create Alert Channels" interface to include related logs and traces information.
	- Improved alert notification content by including original metric labels for creating related links.
- **Refactor**
	- Optimized URL construction in alerting rules to exclude unnecessary query parameters, ensuring concise generator URLs.
	- Adjusted handling of labels in alert rules to support both original and normalized labels for enhanced compatibility and link generation in notifications.

PR #5777, opened 2024-08-27
fix: double encode composite query for explorer links

srikanthccv:
### Summary

In this PR https://github.com/SigNoz/signoz/pull/4573/files#diff-ff903ea33b2803f758b4114f1e70d0e32ac645e679fae55332301bcf80db2b89R767-R830, we removed the double encoding because it seemed like encoding once works. But it fails in some specific cases. Add the double escaping back.


vikrantgupta25 (approved):
lgtm

PR #5793, opened 2024-08-29
feat: added meta tag to prevent page indexing

rahulkeswani101:
### Summary

Added a new meta tag to prevent page indexing.

#### Related Issues / PR's

https://github.com/SigNoz/engineering-pod/issues/1430

#### Screenshots

NA


frontend/public/robots.txt
```diff
@@ -1,3 +1,3 @@
 # https://www.robotstxt.org/robotstxt.html
-User-agent: *
-Disallow:
+User-agent: DatadogSynthetics
```
vikrantgupta25, line 2: this is intended ?

rahulkeswani101: Yes, We don't want to block all crawlers from crawling. So, I thought to disallow  DatadogSynthetics only.

vikrantgupta25: any specific reason to not stop all the crawlers ? and only DD crawler ?

vikrantgupta25: i am not much aware of the context here.. @YounixM for your eyes

SagarRajput-7 (approved):

YounixM (approved):

Issue #5555, opened 2024-07-25
[EPIC]: Logs new schema


nityanandagohain:
- [x] Otel collector changes: https://github.com/SigNoz/signoz-otel-collector/pull/348
  - [x] Write to new logs_v2 table and logs_v2_resource_bucket table
  - [x] Use maps for attributes and resource attributes, and added token_bf index 
  - [x] use `number` instead of int64 and float64
  - [x] use ngram index for body
  - [ ] ~~Update tags to use number instead of int64 and float64, breaking change~~
- [x] Query service changes: https://github.com/SigNoz/signoz/pull/5528
  - [x] Body search case insensitive and uses ngram
  - [x] even though tags are int64 and float64, they are converted to number.
  - [x] Any different orderby other than timestamp should use the new table.
  - [x] ~~Any different orderby (resource) should be added as a where clause in resource filter. (OR)~~ (since coverd by groupBy)
  - [x] Any different groupby (resource) should be added as a where clause in resource filter. (OR)
  - [x] Use proper indexes in resource filter 
  - [x] Proper indexes for attributes/resource attributes for the table ( case insensitive)
  - [ ] ~~Number support in tag attribute API's~~
- [x] Frontend Changes
  - [ ] ~~Support for number type~~
---
Decisions to be taken
- [ ] How should we handle group by, as of now if there are two fields in group by , we add exists filter with `AND` while filtering for it in resources, we can use `OR` here, if we don't use then no filtering will be done. ref:- https://github.com/SigNoz/signoz/pull/5420
- [x] As of now the otel collector writes to two table, the logs_v2 and logs. logs is the old table with order by timestamp, id. Should we keep it the same or create a materialized view for it ?

    We have moved on from logs and will be just using logs_v2
- [x] Do we even require materialized columns ?

     Yes only for group by 
- [x] Should filtering on attributes be case independent? decision = yes.

nityanandagohain:
Few changes
- [ ] Since `number` type on tag_attrbutes will require frontend changes, keeping it as float64 and then in qs we are mapping it to number.
- [ ] Add lower case indexing to the resource fingerprint labels.

nityanandagohain:
Decision w.r.t index

* Go ahead with case sensitive for attributes
* Remove minmax index and only keep bloom filter.

Create issues on clickhouse.

nityanandagohain:
- [ ] remove resource filtering from main table.
[...]

PR #5796, opened 2024-08-29
feat: v4 logs query builder

nityanandagohain:
This PR is the first breakdown PR for https://github.com/SigNoz/signoz/issues/5555

Major changes in v4 with respect to v3 QB are
* Resource filters are only applied to the resource table.
* Support for `__attrs` attribute to search in all attributes
* resource filters are optimised to used the n-gram index.
* Attribute search is case sensitive, while resources and body search is case insensitive.


srikanthccv:
Can you update the `pkg/query-service/app/logs/v4/query_builder.go` and `.../json_filter` to include only the changes that are specific to v4 and use the functions like `orderBy`, `groupByAttrs`, `reduce`, `having` (and everything else that doesn't change in v4) from v3 ?

nityanandagohain:
Have made the changes, orderBy was not possible as it was dependent on the name of the column which changed.

nityanandagohain:
closing in favour of this https://github.com/SigNoz/signoz/pull/5908

PR #5802, opened 2024-08-29
chore: remove old data migrations

srikanthccv:
### Summary

These migrations are not needed anymore. It will help me with make rule manger work with both ee and oss.


srikanthccv:
No new users need to run them. For older users, it can be good to have but not a must because the metrics v3 is going to be supported forever. The only drawback is they will see v3 UI for metrics alerts v3 alerts instead of getting migrated wherever possible.

nityanandagohain:
Got it

nityanandagohain (approved):

PR #5805, opened 2024-08-29
chore: change returned error from list to error

srikanthccv:
### Summary

The returned type is unnecessarily complicated. It is updated to just return the error. When there is more than one error from the validation step, they are combined and returned as one.


pkg/query-service/rules/manager.go
```diff
@@ -160,32 +160,20 @@ func (m *Manager) initiate() error {
...
 
-				if parsedRule == nil {
+				if err != nil {
 					zap.L().Error("failed to parse and initialize yaml rule", zap.String("name", taskName), zap.Error(err))
 					// just one rule is being parsed so expect just one error
-					loadErrors = append(loadErrors, errs[0])
+					loadErrors = append(loadErrors, err)
 					continue
-				} else {
-					// rule stored in yaml, so migrate it to json
```
srikanthccv, line 178: No record in DB should be yaml at this point

Issue #5469, opened 2024-07-11
Alerts based on deviations from baseline


ankitnayan:
Implement alerting from deviations from mean => ( Mean + 2/3/4* sigma )

ankitnayan:
Nice... @YounixM will you have some bandwidth to pick this up next week? If so, @srikanthccv can schedule a meet or if notion docs are enough to give an overview and get the design work started after a go-ahead on the plan

ankitnayan:
@srikanthccv the next question I see users asking why was this alert marked anomalous! Can we have a way to send this with the API response as maybe description and also print out to logs of the QS?

srikanthccv:
> Can we have a way to send this with the API response as maybe description

Yes, this API would return back all the context

 > also print out to logs of the QS
 
By this, I am assuming we should have the exact details of all the values printed in logs, if so, this will be needed even for us to understand. I have added some logs here https://github.com/SigNoz/signoz/compare/develop...anomaly-detection#diff-8cb0c66d798294eee45016205f530d321345767ea305d7dfc7ea3019cc1050a4R1180-R1204 but will be adding more.

YounixM:
Yes @ankitnayan : I can pick this up next week.

pranay01:
@YounixM is this shipped?

srikanthccv:
This is done

pranay01:
@srikanthccv have we also published docs for this?

srikanthccv:
@pranay01 yes, https://signoz.io/docs/alerts-management/anomaly-based-alerts/

PR #5806, opened 2024-08-29
chore: make prepare task configurable

srikanthccv:
### Summary

To be able to support anomaly alerts only in ee https://github.com/SigNoz/signoz/issues/5469#issuecomment-2316655664, we need to run a separate manager in ee. The `prepareTask` function on `Manager` is responsible for creating different kinds of rules depending on type. We will introduce a new rule type called anomaly and set a custom `prepareTask` that includes the `RuleTypeAnomaly` other than `RuleTypeProm`, `RuleTypeThreshold`.


nityanandagohain (approved):
LGTM.

One question though is that, even though a lock was added previously in prepareTask but was not used, any reason that we add that and didn't use ?

srikanthccv:
Only amol has knowledge about it. Because it is never used (by setting acquireLock=true) I removed it.

Issue #5808, opened 2024-08-30
CACHE: Don't merge missedSeres if there is no overlap with the cached series


nityanandagohain:
Don't merge missedSeres if there is no overlap with the cached series

PR #5811, opened 2024-08-30
fix: fix logic for cache

nityanandagohain:
Fixes https://github.com/SigNoz/signoz/issues/5808

The main issue was that when the misses were disjointed of the cached data we merged them regardless.

So for case 5 of generating misses i.e 
// Case 5: Cached time range is a disjoint of the requested time range
we just replace the cache with the new data.


srikanthccv:
I will need time to think more about the change and then review it.

nityanandagohain:
No, worries lets take the time, even I will think more

pkg/query-service/app/querier/querier.go
```diff
@@ -149,7 +149,8 @@ func (q *querier) execPromQuery(ctx context.Context, params *model.QueryRangePar
 //
 // The [End - fluxInterval, End] is always added to the list of misses, because
 // the data might still be in flux and not yet available in the database.
-func findMissingTimeRanges(start, end, step int64, seriesList []*v3.Series, fluxInterval time.Duration) (misses []missInterval) {
+func findMissingTimeRanges(start, end, step int64, seriesList []*v3.Series, fluxInterval time.Duration) (misses []missInterval, replaceCacheData bool) {
```
srikanthccv, line 152: LGTM. Add a todo item to remove the `replaceCacheData` and instead use the list.

nityanandagohain: > Add a todo item to remove the replaceCacheData and instead use the list.

Didn't get you on this

srikanthccv: I meant replacing is not a good idea and it should be removed eventually with better logic so please leave a TODO item in the code comments.

srikanthccv (approved):

Issue #4115, opened 2023-11-30
Default logs view should be close to default logs reading from stdout/console/terminal


srikanthccv:
## Bug description

I never used the log explorer for ClickHouse or any other pod logs because the default view shows me _incomplete_ log line in _boxes_. I always want to see the full log line in by default. When I search for errors I want to see full error details by default without having to click _View Details_ for each one. I think the default mode doesn't really do job of the default log viewer. I want to hear what other think about this.

vikrantgupta25:
@srikanthccv can you add some more details here on what you would like to see in the raw view apart from the entire message at one shot ?

cc - @GeekBoySupreme

PR #5816, opened 2024-08-31
chore: make the raw view the default view for logs

vikrantgupta25:
### Summary

- make the raw view as the default view for logs explorer page.

- skipping one assertion for now as there is some formatting issue with jest dom. will fix this later. not blocking the PR for now. 

#### Related Issues / PR's

contributes to - https://github.com/SigNoz/signoz/issues/4115 

#### Screenshots

NA

### DIFF
--- a/frontend/src/container/OptionsMenu/constants.ts
+++ b/frontend/src/container/OptionsMenu/constants.ts
@@ -7,7 +7,7 @@ export const URL_OPTIONS = 'options';
 export const defaultOptionsQuery: OptionsQuery = {
 	selectColumns: [],
 	maxLines: 2,
-	format: 'list',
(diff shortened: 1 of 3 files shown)


SagarRajput-7 (dismissed):

ankitnayan:
what about those who have already set `Column` view as their default? They should still see Column view when they open the logs explorer page

vikrantgupta25:
> what about those who have already set Column view as their default? They should still see Column view when they open the logs explorer page

yes this is just the default for whoever hasn't set it already in the local storage. No existing set preferences will be affected. @ankitnayan

SagarRajput-7 (approved):

YounixM (approved):

PR #5834, opened 2024-09-02
feat: onboarding v2 base setup

YounixM:
### Summary

https://signoz-team.slack.com/archives/C07KC2DD3CN/p1730277289255069?thread_ts=1730127377.586999&cid=C07KC2DD3CN


frontend/src/AppRoutes/Private.tsx
```diff
@@ -31,6 +35,14 @@ function PrivateRoute({ children }: PrivateRouteProps): JSX.Element {
 	const location = useLocation();
 	const { pathname } = location;
 
+	const { org, orgPreferences } = useSelector<AppState, AppReducer>(
+		(state) => state.app,
+	);
+
+	const [isOnboardingComplete, setIsOnboardingComplete] = useState<
+		boolean | null
```
vikrantgupta25, line 43: why `null` here ? it should be either true or false ?

frontend/src/AppRoutes/Private.tsx
```diff
@@ -44,6 +56,18 @@ function PrivateRoute({ children }: PrivateRouteProps): JSX.Element {
 		[pathname],
 	);
 
+	useEffect(() => {
+		if (orgPreferences && !isEmpty(orgPreferences)) {
+			const onboardingPreference = orgPreferences?.find(
+				(preference: Record<string, any>) => preference.key === 'ORG_ONBOARDING',
```
vikrantgupta25, line 62: maintain proper keys for preferences in the client.

frontend/src/AppRoutes/Private.tsx
```diff
@@ -81,6 +109,42 @@ function PrivateRoute({ children }: PrivateRouteProps): JSX.Element {
...
+			(user) => user.email !== 'admin@signoz.cloud',
+		);
+
+		return remainingUsers.length === 1;
+	};
+
+	// Check if the onboarding should be shown based on the org users and onboarding completion status, wait for org users and preferences to load
+	const shouldShowOnboarding = (): boolean => {
+		// Only run this effect if the org users and preferences are loaded
+		if (!isLoadingOrgUsers) {
```
vikrantgupta25, line 138: according to the above comment `org users and preferences are loaded` we should have waited for `isFetchingOrgPreferences` as well here ?

frontend/src/container/OnboardingQuestionaire/InviteTeamMembers/InviteTeamMembers.tsx
```diff
@@ -0,0 +1,434 @@
...
+	};
+
+	// Validation function to check all users
+	const validateAllUsers = (): boolean => {
+		let isValid = true;
+
+		const updatedValidity: Record<string, boolean> = {};
+
+		teamMembersToInvite?.forEach((member) => {
+			const emailValid = /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(member.email);
```
vikrantgupta25, line 111: extract this to a constant

vikrantgupta25:
1. cleanup the preferences / other moving parts from redux store and move to their corresponding contexts - https://github.com/SigNoz/engineering-pod/issues/1957#event-14999343047
2. maintain proper keys for preferences in the client - 
```
const onboardingPreference = orgPreferences?.find(
				(preference: Record<string, any>) => preference.key === 'ORG_ONBOARDING',
			);
```
3. The `Private.tsx` file needs to be cleaned up / redone properly, there are multiple areas which can currently open to a bunch of race conditions and flickers.

frontend/src/AppRoutes/index.tsx
```diff
@@ -182,6 +212,16 @@ function App(): JSX.Element {
 	}, [isLoggedInState, isOnBasicPlan, user]);
 
 	useEffect(() => {
+		if (pathname === ROUTES.ONBOARDING) {
+			window.Intercom('update', {
+				hide_default_launcher: true,
```
vikrantgupta25, line 217: should we enable this ? was wondering since we do not have skip option on the page and if user is facing some issue here then there is no way they can communicate...

vikrantgupta25 (approved):
non blocker comments can be taken up later

frontend/src/AppRoutes/index.tsx
YounixM: As this flow doesn't require alot of help, we choose to hide it for this to limit flows.

frontend/src/AppRoutes/Private.tsx
```diff
@@ -160,6 +234,35 @@ function PrivateRoute({ children }: PrivateRouteProps): JSX.Element {
...
+			history.push(ROUTES.ONBOARDING);
+		} else {
+			history.push(ROUTES.APPLICATION);
+		}
+	};
+
+	useEffect(() => {
+		// Only run this effect if the org users and preferences are loaded
+		if (!isLoadingOrgUsers && isOnboardingComplete !== null) {
+			const isFirstUser = checkFirstTimeUser();
```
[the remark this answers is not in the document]
YounixM: The flows are different on when both of them are triggered. We need to clean up this file to handling routing better.

ankit01-oss (approved):

PR #5844, opened 2024-09-03
fix: use better value for threshold value in alert description

srikanthccv:
### Summary

The alert sends a description, _the threshold is set to X, and the observed metric value is Y_

If the condition match type is

- `InTotal`: the observed metric value computed as `sum(values...)` ✅ 
- `OnAverage`: the observed metric value computed as `avg(values...)` ✅ 
- `AtleastOnce`: the observed metric value computed is whichever is the first value that meets the condition
  Example:
  - `ValueIsAbove`: any value (usually the first ordered by ts desc) in series > threshold
  - `ValueIsBelow`: any value (usually the first ordered by ts desc) in series < threshold
  - `ValueIsEq`: any value (usually the first ordered by ts desc) in series = threshold
  - `ValueIsNotEq`: any value (usually the first ordered by ts desc) in series != threshold

However, the current logic for `AllTheTimes` uses the `targetVal` which doesn't give a good idea of the observed metric value https://github.com/SigNoz/signoz/blob/709c2860869a58be97ade3220b8b47952b359678/pkg/query-service/rules/threshold_rule.go#L1200

This change updates the value to use in the description.

When:
- `ValueIsAbove`: since we can't show multiple values in the description, we use the min among the series values
- `ValueIsBelow`: since we can't show multiple values in the description, we use the max among the series values
- `ValueIsEq`: nothing needs to change
- `ValueIsNotEq`: any non-inf or nan value from the series

### DIFF
--- a/pkg/query-service/rules/prom_rule.go
+++ b/pkg/query-service/rules/prom_rule.go
@@ -593,12 +593,30 @@ func (r *PromRule) shouldAlert(series pql.Series) (pql.Sample, bool) {
 					break
 				}
 			}
+			if shouldAlert {
+				var maxValue float64 = math.Inf(-1)
+				for _, smpl := range series.Floats {
+					if smpl.F > maxValue {
+						maxValue = smpl.F
+					}
+				}
+				alertSmpl = pql.Sample{F: maxValue, Metric: series.Metric}
+			}
 		} else if r.compareOp() == ValueIsBelow {
 			for _, smpl := range series.Floats {
 				if smpl.F >= r.targetVal() {
 					shouldAlert = false
 					break
 				}
+			}
+			if shouldAlert {
+				var minValue float64 = math.Inf(1)
+				for _, smpl := range series.Floats {
+					if smpl.F < minValue {
+						minValue = smpl.F
+					}
+				}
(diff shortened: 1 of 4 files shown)


pkg/query-service/rules/prom_rule.go
```diff
@@ -593,13 +593,31 @@ func (r *PromRule) shouldAlert(series pql.Series) (pql.Sample, bool) {
 					break
 				}
 			}
+			if shouldAlert {
```
nityanandagohain, line 596: should be min value ?

pkg/query-service/rules/prom_rule.go
```diff
@@ -593,13 +593,31 @@ func (r *PromRule) shouldAlert(series pql.Series) (pql.Sample, bool) {
...
+			}
 		} else if r.compareOp() == ValueIsBelow {
 			for _, smpl := range series.Floats {
 				if smpl.F >= r.targetVal() {
 					shouldAlert = false
 					break
 				}
 			}
+			if shouldAlert {
+				var minValue float64 = math.Inf(1)
```
nityanandagohain, line 613: should be max value ?

nityanandagohain (changes requested):

nityanandagohain (approved):

PR #5845, opened 2024-09-03
chore: alert state change and overall status

srikanthccv:
### Summary

This PR

- changes the table name to `_v0` to be explicit in the version
- removes the random use of "firing" and "normal" string values in code to type from the model
- fix the issue with overall status where there are no entries or last entry doesn't have it's corresponding complement state, for example, the alert is firing but doesn't have resolved state yet so we need to add the "firing" entry at last
- 

Some of the code is the same as the threshold rule and prom rule. I plan to address this separately.


pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -5374,13 +5390,87 @@ ORDER BY firing_time ASC;`
...
+				End:   params.End,
+			})
+		} else {
+			// fetch the most recent firing event from the table in the given time range
+			var firingTime int64
+			firingQuery := fmt.Sprintf(`
+			SELECT
+				unix_milli
+			FROM %s.%s
+			WHERE rule_id = '%s' AND overall_state_changed = true AND overall_state = 'firing' AND unix_milli <= %d
```
nityanandagohain, line 5456: nit: use `model.StateFiring.String() ` ?

nityanandagohain (dismissed):

pkg/query-service/rules/prom_rule.go
```diff
@@ -338,6 +341,100 @@ func (r *PromRule) compareOp() CompareOp {
...
+
+		newState := model.StateInactive
+		for _, item := range revisedItemsToAdd {
+			if item.State == model.StateFiring || item.State == model.StateNoData {
+				newState = model.StateFiring
+				break
+			}
+		}
+		zap.L().Debug("newState", zap.String("ruleid", r.ID()), zap.Any("newState", newState))
+		if lastSavedState[0].State != newState {
```
nityanandagohain, line 406: please add a example comment for this as well after making the changes

pkg/query-service/rules/prom_rule.go
```diff
@@ -338,6 +341,100 @@ func (r *PromRule) compareOp() CompareOp {
 	return r.ruleCondition.CompareOp
 }
 
+// TODO(srikanthccv): implement base rule and use for all types of rules
```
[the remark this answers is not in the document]
srikanthccv: no this logic will remain the same independent of the alert type so will move it to the base rule that every other rule can use.

nityanandagohain (approved):

PR #5853, opened 2024-09-04
fix: invalidate alert rule details query on switching the alert disabled state

ahmadshaheer:
### Summary
invalidate alert rule details query on switching the alert disabled state

#### Screenshots

NA


ahmadshaheer:
close in favor of https://github.com/SigNoz/signoz/pull/5854

PR #5870, opened 2024-09-05
chore: calculate user count dynamically and set user role in identity…

makeavish:
… event

#### Screenshots

NA


pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3152,6 +3152,22 @@ func (r *ClickHouseReader) GetTotalSpans(ctx context.Context) (uint64, error) {
...
+	return role.Name, nil
+}
+
+func (r *ClickHouseReader) GetUserCount(ctx context.Context) (int, error) {
+	users, err := dao.DB().GetUsers(ctx)
+	if err != nil {
+		return 0, err
+	}
+	return len(users), nil
+}
```
srikanthccv, line 3169: Let's not add these to the ClickHouseReader interface. Please use dao in the telemetry instance to access the info. We are in the process of removing non-ClickHouse things from the ClickHouse reader interface.

makeavish: There will be cyclic dependency if we try to add dao

srikanthccv: We should fix the cyclic dependency.

makeavish: Yeah, that would involve creating an interface for dao, like we have for reader.
It would be a significant change, I think we are already working on improving query service structure in parallel so let's not block this PR.

srikanthccv: since your bandwidth is prioritized elsewhere I took a pull and moved them to callbacks. this will get count and role details without adding new functions to the CH reader interface.

srikanthccv (approved):

PR #5872, opened 2024-09-05
feat: logsV4 resource table query builder

nityanandagohain:
This PR just focuses just the query builder for the resource table.

it's a breakdown of this PR https://github.com/SigNoz/signoz/pull/5796

First PR for https://github.com/SigNoz/signoz/issues/5555


pkg/query-service/app/logs/v4/resource_query_builder.go
```diff
@@ -0,0 +1,204 @@
...
+	case v3.FilterOperatorNotExists:
+		return fmt.Sprintf("not simpleJSONHas(lower(labels), '%s')", key)
+	case v3.FilterOperatorRegex, v3.FilterOperatorNotRegex:
+		return fmt.Sprintf(logsOp, searchKey, chFmtVal)
+	case v3.FilterOperatorContains, v3.FilterOperatorNotContains:
+		// this is required as clickhouseFormattedValue add's quotes to the string
+		lowerEscapedStringValue := utils.QuoteEscapedString(strings.ToLower(fmt.Sprintf("%s", value)))
+		return fmt.Sprintf("%s %s '%%%s%%'", searchKey, logsOp, lowerEscapedStringValue)
+	default:
+		chFmtValLower := strings.ToLower(chFmtVal)
```
srikanthccv, line 30: On the main table also do we follow the case-insensitive search?

pkg/query-service/app/logs/v4/resource_query_builder.go
```diff
@@ -0,0 +1,204 @@
...
+func buildResourceIndexFilter(key string, op v3.FilterOperator, value interface{}) string {
+	// not using clickhouseFormattedValue as we don't wan't the quotes
+	formattedValueEscapedLower := utils.QuoteEscapedString(strings.ToLower(fmt.Sprintf("%s", value)))
+
+	// add index filters
+	switch op {
+	case v3.FilterOperatorContains, v3.FilterOperatorEqual, v3.FilterOperatorLike:
+		return fmt.Sprintf("lower(labels) like '%%%s%%%s%%'", key, formattedValueEscapedLower)
+	case v3.FilterOperatorNotContains, v3.FilterOperatorNotEqual, v3.FilterOperatorNotLike:
+		return fmt.Sprintf("lower(labels) not like '%%%s%%%s%%'", key, formattedValueEscapedLower)
```
srikanthccv, line 85: How do these work when value has `%` or `_` in them?

pkg/query-service/app/logs/v4/resource_query_builder.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	if len(conditions) == 0 {
+		return "", nil
+	}
+	conditionStr := strings.Join(conditions, " AND ")
+
+	// BUILD THE FINAL QUERY
+	query := fmt.Sprintf("(SELECT fingerprint FROM signoz_logs.%s WHERE (seen_at_ts_bucket_start >= %d) AND (seen_at_ts_bucket_start <= %d) AND ", DISTRIBUTED_LOGS_V2_RESOURCE, bucketStart, bucketEnd)
+
+	query = query + conditionStr + ")"
```
srikanthccv, line 201: Please try to avoid this. Why does this only add `)` at the end? Where did the `(` come from?

pkg/query-service/app/logs/v4/resource_query_builder.go
```diff
@@ -0,0 +1,204 @@
...
+	"fmt"
+	"strings"
+
+	v3 "go.signoz.io/signoz/pkg/query-service/model/v3"
+	"go.signoz.io/signoz/pkg/query-service/utils"
+)
+
+// buildResourceFilter builds a clickhouse filter string for resource labels
+func buildResourceFilter(logsOp string, key string, op v3.FilterOperator, value interface{}) string {
+	// we are using lower(labels) as we want case insensitive filtering
```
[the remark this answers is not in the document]
nityanandagohain: Apart from regex, everything else is case-insensitive.

pkg/query-service/app/logs/v4/resource_query_builder.go
```diff
@@ -0,0 +1,204 @@
...
+
+	chFmtVal := utils.ClickHouseFormattedValue(value)
+
+	switch op {
+	case v3.FilterOperatorExists:
+		return fmt.Sprintf("simpleJSONHas(lower(labels), '%s')", key)
+	case v3.FilterOperatorNotExists:
+		return fmt.Sprintf("not simpleJSONHas(lower(labels), '%s')", key)
+	case v3.FilterOperatorRegex, v3.FilterOperatorNotRegex:
+		return fmt.Sprintf(logsOp, searchKey, chFmtVal)
```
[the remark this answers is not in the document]
nityanandagohain: the existing order is correct, here is the example
logOp = match(%s, %s)
searchKey = simpleJSONExtractString(lower(labels), 'service_name')
chFmtVal = ".*'

finally it will be match(simpleJSONExtractString(lower(labels), 'service_name'), '.*')

pkg/query-service/app/logs/v4/resource_query_builder.go
```diff
@@ -0,0 +1,204 @@
...
+	"strings"
+
+	v3 "go.signoz.io/signoz/pkg/query-service/model/v3"
+	"go.signoz.io/signoz/pkg/query-service/utils"
+)
+
+// buildResourceFilter builds a clickhouse filter string for resource labels
+func buildResourceFilter(logsOp string, key string, op v3.FilterOperator, value interface{}) string {
+	// we are using lower(labels) as we want case insensitive filtering
+	searchKey := fmt.Sprintf("simpleJSONExtractString(lower(labels), '%s')", key)
```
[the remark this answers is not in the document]
nityanandagohain: All resource attributes are stored as strings in the main table, so if there is a resource named

`service.name = []{x,y}` then it will be stored as a string, but I got your point, let me add an example test case if possible.

pkg/query-service/app/logs/v4/resource_query_builder.go
nityanandagohain: For tag attributes, it's still case sensitive  due to limitations on map indexes.

All resources filtering goes to this table so all resource attributes search is case insensitive.

pkg/query-service/app/logs/v4/resource_query_builder.go
nityanandagohain: cool, "(" was added in the previous line, but let me clean this.

pkg/query-service/app/logs/v4/resource_query_builder.go
nityanandagohain: It should keep them as it is, 

i.e `x contains 'app%_test`

Should result in
`lower(labels) ilike  '%x%app%_test%'`

Adding a test case for this

pkg/query-service/app/logs/v4/resource_query_builder.go
```diff
@@ -0,0 +1,204 @@
...
+				continue
+			}
+			values = append(values, strV)
+		}
+	}
+
+	// if there are no values to filter on, return an empty string
+	if len(values) > 0 {
+		for _, v := range values {
+			conditions = append(conditions, fmt.Sprintf("lower(labels) %s '%%\"%s\":\"%s\"%%'", sqlOp, key, strings.ToLower(v)))
```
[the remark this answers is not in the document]
nityanandagohain: right let me add that

pkg/query-service/app/logs/v4/resource_query_builder.go
srikanthccv: I don't this is a good idea overall i.e search on resource attributes is case insensitive but a search on regular attributes is case sensitive. The filter by resource is common so the user forms an impression that a search on any attributes is case-insensitive but then when they do search on attributes with the same expectation it will fail.

pkg/query-service/app/logs/v4/resource_query_builder.go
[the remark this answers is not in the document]
nityanandagohain: thanks, updated the test case

pkg/query-service/app/logs/v4/resource_query_builder.go
nityanandagohain: have added a test case for this `test contains with % and _`

pkg/query-service/app/logs/v4/resource_query_builder.go
nityanandagohain: updated

pkg/query-service/app/logs/v4/resource_query_builder.go
nityanandagohain: So we want to get to a place we it's insensitive for attributes as well, but the limitation is restricting us. We eventually want to pick it up.

pkg/query-service/app/logs/v4/resource_query_builder.go
srikanthccv: `contains` is regular string matching. It should not keep them as is. If I am searching for `service.name` contains `my_service` but the `like` or `not like` will not treat `_`, `%` as literals.

nityanandagohain: got your point, even though the result will be there, better to keep expectations correct.

pkg/query-service/app/logs/v4/resource_query_builder.go
srikanthccv: When is that eventually thing going to happen? It wouldn't be a problem if the feature were to be rarely used. but working with attributes happens all the time and people shouldn't learn it hard way that resource attributes and regular attributes work differently.

nityanandagohain: > When is that eventually thing going to happen
I don't have time that I can give you.

I agree that there will be a different curve for this, but @ankitnayan  suggested going ahead with this as a starting point.

srikanthccv (approved):

PR #5898, opened 2024-09-09
fix: fix incorrect alert history state

ahmadshaheer:
### Summary
fixed the incorrect alert state in enable/disable switch. the issue was caused by old disabled state (from another alert). fixed by removing the old state on unmount.

#### Screenshots
Before:
Screenshot 2024-09-09 at 1 57 37 PM

Screenshot 2024-09-09 at 1 57 47 PM

After:
https://www.loom.com/share/9b15bc0edbf2439a90b5349e19562e6e

NA

### DIFF
--- a/frontend/src/providers/Alert.tsx
+++ b/frontend/src/providers/Alert.tsx
@@ -1,10 +1,8 @@
 import React, { createContext, useContext, useState } from 'react';
 
 interface AlertRuleContextType {
-	isAlertRuleDisabled: boolean | undefined;
-	setIsAlertRuleDisabled: React.Dispatch<
-		React.SetStateAction<boolean | undefined>
-	>;
+	alertRuleState: string | undefined;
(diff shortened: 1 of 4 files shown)


frontend/src/providers/Alert.tsx
```diff
@@ -16,15 +14,14 @@ function AlertRuleProvider({
...
 }): JSX.Element {
-	const [isAlertRuleDisabled, setIsAlertRuleDisabled] = useState<
-		boolean | undefined
-	>(undefined);
-
-	const value = React.useMemo(
-		() => ({ isAlertRuleDisabled, setIsAlertRuleDisabled }),
-		[isAlertRuleDisabled],
+	const [alertRuleState, setAlertRuleState] = useState<string | undefined>(
+		undefined,
```
rahulkeswani101, line 18: Can you please remove undefined which we are passing explicitly as the initial state?

YounixM (approved):

rahulkeswani101 (approved):

PR #5908, opened 2024-09-10
logs v4 qb refactor

nityanandagohain:
For https://github.com/SigNoz/signoz/issues/5555 , second PR . Follow up of https://github.com/SigNoz/signoz/pull/5872

Changes made code wise 
* In v3 folder `json_filter.go` , `json_filter_test.go`, `query_builder.go`
   Some functions and constants are made public in these files and no other change.
   
* In v4 folder `json_filter.go` and `json_filter_test.go`
  The only change for these two files is that the map `jsonLogOperators` uses `LIKE` instead of `ILIKE`. So the function `GetJSONFilter` is a local one and no change apart from that.
  

* In v4 folder `query_builder.go` and `query_builder_test.go`
  * It adds the resource filter caluse
  * new select clauses for map support.
  * refactoring the code as we did for v4 resource qb
  * Only body contains search is case insensitive


srikanthccv:
>The only change for these two files is that the map jsonLogOperators uses LIKE instead of ILIKE. So the function GetJSONFilter is a local one and no change apart from that.


Can we get rid of v3 in logs after v4 is fully rolled out?

nityanandagohain:
Yeah once it is fully rolled out, we will stop ingestion for the old table and the entire v3 code can be removed, some functions will have to be moved to the new folder that we are reusing.

pkg/query-service/app/logs/v4/query_builder.go
```diff
@@ -29,3 +35,492 @@ const (
...
+		return "bool"
+	}
+	return "string"
+}
+
+func getClickhouseKey(key v3.AttributeKey) string {
+	// check if it is a top level static field
+	if _, ok := constants.StaticFieldsLogsV3[key.Key]; ok && key.Type == v3.AttributeKeyTypeUnspecified {
+		return key.Key
+	}
```
srikanthccv, line 53: Is this needed _key.Type == v3.AttributeKeyTypeUnspecified_? Can there be an attribute which has same name as top level static field?

pkg/query-service/app/logs/v4/query_builder.go
```diff
@@ -29,3 +35,492 @@ const (
...
+	//if the key is present in the topLevelColumn then it will be only searched in those columns,
+	//regardless if it is indexed/present again in resource or column attribute
+	if !key.IsColumn {
+		columnType := logsV3.GetClickhouseLogsColumnType(key.Type)
+		columnDataType := getClickhouseLogsColumnDataType(key.DataType)
+		return fmt.Sprintf("%s_%s['%s']", columnType, columnDataType, key.Key)
+	}
+
+	// materialized column created from query
+	return "`" + utils.GetClickhouseColumnNameV2(string(key.Type), string(key.DataType), key.Key) + "`"
```
srikanthccv, line 64: materialized columns won't have any dots in them. Why do you need "``" around the column name?

pkg/query-service/app/logs/v4/query_builder.go
```diff
@@ -29,3 +35,492 @@ const (
...
+	if err != nil {
+		return "", err
+	}
+
+	var queryTmplPrefix string
+	if graphLimitQtype == constants.FirstQueryGraphLimit {
+		queryTmplPrefix = "SELECT"
+	} else if panelType == v3.PanelTypeTable {
+		queryTmplPrefix =
+			"SELECT now() as ts,"
```
srikanthccv, line 420: This `SELECT now() as ts` is no longer needed. The get time series result function on reader works without ts column.

pkg/query-service/app/logs/v4/query_builder.go
```diff
@@ -29,3 +35,492 @@ const (
...
+	aggregationKey := ""
+	if mq.AggregateAttribute.Key != "" {
+		aggregationKey = getClickhouseKey(mq.AggregateAttribute)
+	}
+
+	// for limit queries, there are two queries formed
+	// in the second query we need to add the placeholder so that first query can be placed
+	if graphLimitQtype == constants.SecondQueryGraphLimit {
+		filterSubQuery = filterSubQuery + " AND " + fmt.Sprintf("(%s) GLOBAL IN (", logsV3.GetSelectKeys(mq.AggregateOperator, mq.GroupBy)) + "#LIMIT_PLACEHOLDER)"
+	}
```
srikanthccv, line 408: Does it make any difference now to use two queries for limit queries now that we are using the resource filter?

pkg/query-service/app/logs/v4/query_builder.go
```diff
@@ -29,3 +35,492 @@ const (
...
+	if err != nil {
+		return "", err
+	}
+
+	switch mq.AggregateOperator {
+	case v3.AggregateOperatorNoOp:
+		query := constants.LogsSQLSelectV2 + "from signoz_logs." + DISTRIBUTED_LOGS_V2 + " where "
+		if len(filterSubQuery) > 0 {
+			query = query + filterSubQuery + " AND "
+		}
```
srikanthccv, line 451: Does live tail not use the resource filter?

srikanthccv (commented):
Some comments

pkg/query-service/app/logs/v4/query_builder.go
nityanandagohain: Yes, there can be an attribute named `body`, `trace_id` etc. During enrichment, the type will be updated.

pkg/query-service/app/logs/v4/query_builder.go
nityanandagohain: So this is the reason why we added '`' to all the materialized columns while referencing them. Let me add it as a comment.

nityanandagohain: https://github.com/SigNoz/signoz/pull/4775

pkg/query-service/app/logs/v4/query_builder.go
nityanandagohain: Sure.

pkg/query-service/app/logs/v4/query_builder.go
nityanandagohain: Yeah, logic with limit is different from what we are doing now, it will remain the same. I have updated the tets for better understanding.

pkg/query-service/app/logs/v4/query_builder.go
nityanandagohain: I have use the previous logic and update the code to support it. 

See commit https://github.com/SigNoz/signoz/pull/5908/commits/34c2c7b236db8a3b7e5cc8188b5419756877601d

This is how it will be used by the reader https://github.com/SigNoz/signoz/pull/5914/commits/fc08bd93d043ab021a834452420fd111040db0bf (different PR)

srikanthccv:
Mostly LGTM. @raj-k-singh please review

srikanthccv (approved):

PR #5912, opened 2024-09-10
feat: logs list API, logic update for better perf

nityanandagohain:
Improved logic in list api for better perf

Follow up of https://github.com/SigNoz/signoz/pull/5908

For issue https://github.com/SigNoz/signoz/issues/5555

Logic:
* Look for 100 logs in the last one hour **X**.
* If not found look for log in **X-3** to **X-1** hour.
* Keep looking back until sufficient data is found or it reaches timestamp start.

Note the same logic and test is repeated in querierV2


pkg/query-service/app/querier/querier.go
```diff
@@ -469,7 +470,115 @@ func (q *querier) runClickHouseQueries(ctx context.Context, params *v3.QueryRang
...
+func (q *querier) runLogsListQuery(ctx context.Context, params *v3.QueryRangeParamsV3, keys map[string]v3.AttributeKey, tsRanges []logsListTsRange) ([]*v3.Result, map[string]error, error) {
+	res := make([]*v3.Result, 0)
+	qName := ""
+	pageSize := uint64(0)
+
+	// se we are considering only one query
+	for name, v := range params.CompositeQuery.BuilderQueries {
+		qName = name
+		pageSize = v.PageSize
+	}
```
srikanthccv, line 519: Let's be clear about which query. Should this be the last query? Do we consider enabled/disabled?

pkg/query-service/app/querier/querier.go
```diff
@@ -469,7 +470,115 @@ func (q *querier) runClickHouseQueries(ctx context.Context, params *v3.QueryRang
...
+	})
+	return res, nil, nil
+}
+
 func (q *querier) runBuilderListQueries(ctx context.Context, params *v3.QueryRangeParamsV3, keys map[string]v3.AttributeKey) ([]*v3.Result, map[string]error, error) {
+	// List query has support for only one query.
+	if q.UseLogsNewSchema && params.CompositeQuery != nil {
+		for _, v := range params.CompositeQuery.BuilderQueries {
+			// only allow of logs queries with timestamp ordering desc
+			if v.DataSource == v3.DataSourceLogs && len(v.OrderBy) == 1 && v.OrderBy[0].ColumnName == "timestamp" && v.OrderBy[0].Order == "desc" {
```
srikanthccv, line 574: What happens in other cases?

pkg/query-service/app/querier/v2/querier.go
```diff
@@ -478,7 +479,115 @@ func (q *querier) runClickHouseQueries(ctx context.Context, params *v3.QueryRang
...
 }
 
+type logsListTsRange struct {
+	Start int64
+	End   int64
+}
+
+const HOUR_NANO = int64(3600000000000)
+
+func getLogsListTsRanges(start, end int64) []logsListTsRange {
```
srikanthccv, line 489: It seems like this doesn't need to exist here again. On a side note, I am finding you are forced to copy a lot of code to querier v2 even though it's not needed. Let's create a separate issue to address this.

srikanthccv (commented):
How does logs context work going forward?

pkg/query-service/app/querier/querier.go
nityanandagohain: Makes sense, I have restricted it to run only for queries where there is a single list query.  https://github.com/SigNoz/signoz/pull/5912/commits/803d480596ddb916d88146fc09c9b8997a1d0d8b .


We can come back if we want to support his feature for multiple list queries.

pkg/query-service/app/querier/querier.go
nityanandagohain: In other cases, the entire time range will be used. this logic will not be applied.

pkg/query-service/app/querier/v2/querier.go
nityanandagohain: Do you think utils will be a right pace for this function ?

for the second part https://github.com/SigNoz/signoz/issues/new

nityanandagohain:
No changes on the backend w.r.t logs context. On the frontend as of now order by is timestamp, we will add id as well in the order by. It's a non blocker.

pkg/query-service/app/querier/querier.go
srikanthccv: When the order is "asc" why can't we apply the same logic?

nityanandagohain: let me solve that separately in a different issue, as this will be the majority of users use-case.

https://github.com/SigNoz/signoz/issues/5943

pkg/query-service/app/querier/querier.go
```diff
@@ -469,7 +470,115 @@ func (q *querier) runClickHouseQueries(ctx context.Context, params *v3.QueryRang
...
+			params.CompositeQuery.BuilderQueries[qName].Filters.Items = append(params.CompositeQuery.BuilderQueries[qName].Filters.Items, v3.FilterItem{
+				Key: v3.AttributeKey{
+					Key:      "id",
+					IsColumn: true,
+					DataType: "string",
+				},
+				Operator: v3.FilterOperatorLessThan,
+				Value:    data[len(data)-1].Data["id"],
+			})
+		}
```
srikanthccv, line 556: Can you share examples of queries prepared when this is applied or data is paginated? Would there be multiple id filter items?

pkg/query-service/app/querier/v2/querier.go
srikanthccv: Yes, there is no specific reason why this should live here then it can go in utils.

srikanthccv (commented):
One main question around the `id`

srikanthccv (approved):

PR #5916, opened 2024-09-10
feat: anomaly detection UI

YounixM:
Feature: Anomal Detection 

Users should be able to setup a anomaly based alerts for metrics.

**Alert Selection Change**
Screenshot 2024-10-13 at 5 35 00 PM

**Switch between Threshold based alert and Anomaly based alert**
Screenshot 2024-10-13 at 5 35 11 PM

**Define Alert Section Change:**
Screenshot 2024-10-13 at 5 35 21 PM

**Evaluation view rendering**
Screenshot 2024-10-13 at 5 35 52 PM

Screenshot 2024-10-13 at 5 36 26 PM

Screenshot 2024-10-13 at 5 36 47 PM

**Anomaly Function for all queries**
Screenshot 2024-10-13 at 5 37 05 PM


frontend/src/container/AnomalyAlertEvaluationView/AnomalyAlertEvaluationView.tsx
```diff
@@ -0,0 +1,280 @@
...
+}: {
+	data: any;
+	yAxisUnit: string;
+}): JSX.Element {
+	const { spline } = uPlot.paths;
+	// eslint-disable-next-line @typescript-eslint/naming-convention
+	const _spline = spline ? spline() : undefined;
+	const chartRef = useRef<HTMLDivElement>(null);
+	const isDarkMode = useIsDarkMode();
+	const [seriesData, setSeriesData] = useState<any>({});
```
vikrantgupta25, line 63: let's not use any

frontend/src/container/AnomalyAlertEvaluationView/AnomalyAlertEvaluationView.tsx
```diff
@@ -0,0 +1,280 @@
...
+	};
+
+	const bandsPlugin = {
+		hooks: {
+			draw: [
+				(u: any): void => {
+					if (!selectedSeries) return;
+
+					const { ctx } = u;
+					const upperBandIdx = 3;
```
vikrantgupta25, line 94: let's add comments for what they represent

vikrantgupta25 (approved):

YounixM:
@vikrantgupta25 : Will take these up in the subsequent PR's.

srikanthccv:
- Let's add a beta tag for this
- This is an EE feature so we should only show when `ANOMALY_DETECTION` is enabled
- Can we make the graph experience the same as the rest of the product? The mouse movement on the graph has a completely different experience.
- We should add a simple search for legend selection since the list can be too long
- The lower band and upper band have straight lines even when the data has variation, the real and predicted lines are fine.
- There can only be one active query on the graph, which should be the same as the selected query condition. The data for other queries should not be shown
- We need to figure out a few things related to the formula (since there are no functions in the formula, nothing is shown)

YounixM:
@srikanthccv 


> Can we make the graph experience the same as the rest of the product? The mouse movement on the graph has a completely different experience.
Will make the change in the next PR.


Can you please provide more details on these points

> There can only be one active query on the graph, which should be the same as the selected query condition. The data for other queries should not be shown
> We need to figure out a few things related to the formula (since there are no functions in the formula, nothing is shown)

YounixM:
@srikanthccv : Will the API return only the non anomaly based alerts if the FF is disabled ?

srikanthccv:
Discussed disabling the anomaly detection by default for everyone and only enable for specific tenants (like private beta), starting with saasmonitor.

YounixM:
As discussed, merging the base PR. will work on the above points in next PR / subsequent PR's

YounixM:
1. Let's add a beta tag for this
2. This is an EE feature so we should only show when ANOMALY_DETECTION is enabled


Merged ✅

YounixM:
4. We should add a simple search for legend selection since the list can be too long - PR raised

YounixM:
4. We should add a simple search for legend selection since the list can be too long - PR merged ✅ 

3. Added tooltip to show series data. - PR raised

PR #5917, opened 2024-09-10
chore: move traces builder query attributes enrichment before query prep

srikanthccv:
### Summary

Currently, the attribute erichment is happening at the time of query preparation. This means we should pass keys all the way down which makes the whole call require passing keys. This commit removes it from params and enriches the attribute prior like we do for logs.


makeavish:
@CodiumAI-Agent /review

srikanthccv:
@makeavish the linked issue won't be solved by this yet. The keys are still loaded every time for query range request. This refactor streamlines things so that alerts and query range perform attribute enrichment upfront.

srikanthccv:
The follow up built on top of this https://github.com/SigNoz/signoz/pull/5925

srikanthccv:
@makeavish please review

makeavish (approved):

PR #5932, opened 2024-09-12
chore: move some structs out of v3

srikanthccv:
### Summary

More clean up after https://github.com/SigNoz/signoz/pull/5917. This moves non query range structs out of the v3. This is done to avoid the circular dependency. While doing so, I removed some dead code or code that doesn't work.


nityanandagohain (dismissed):
I have added logLivetailclientV2 as well , you want to move it ?

nityanandagohain (approved):

Issue #5964, opened 2024-09-13
Process get panic when all queries in alert rule are disabled


kobecal:
## Bug description
panic logs are as follow
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x1881f2c]

goroutine 183 [running]:
go.signoz.io/signoz/pkg/query-service/rules.(*ThresholdRule).buildAndRunQuery(0xc000f2a6e0, {0x26acf28, 0xc0010bcff0}, {0x32e?, 0xc000565180?, 0x0?}, {0x26bd7d0, 0xc0006ccc60})
	/home/runner/work/signoz/signoz/pkg/query-service/rules/thresholdRule.go:826 +0xcec
go.signoz.io/signoz/pkg/query-service/rules.(*ThresholdRule).Eval(0xc000f2a6e0, {0x26acf28?, 0xc0010bcff0}, {0x2692a10?, 0x1ca04e0?, 0x0?}, 0xc000014168)
	/home/runner/work/signoz/signoz/pkg/query-service/rules/thresholdRule.go:855 +0x105
go.signoz.io/signoz/pkg/query-service/rules.(*RuleTask).Eval.func1({0x26acf28?, 0xc0010bcf30?}, {0x1724f14c?, 0xc0010b4980?, 0x0?}, 0xc0001dea50, 0x0?, {0x26c8fb0, 0xc000f2a6e0})
	/home/runner/work/signoz/signoz/pkg/query-service/rules/ruleTask.go:351 +0x363
go.signoz.io/signoz/pkg/query-service/rules.(*RuleTask).Eval(0xc0001dea50, {0x26acf28, 0xc0010bcf30}, {0xc0000c9d80?, 0x470e93?, 0x0?})
	/home/runner/work/signoz/signoz/pkg/query-service/rules/ruleTask.go:368 +0x631
go.signoz.io/signoz/pkg/query-service/rules.(*RuleTask).Run.func1()
	/home/runner/work/signoz/signoz/pkg/query-service/rules/ruleTask.go:116 +0x7d
go.signoz.io/signoz/pkg/query-service/rules.(*RuleTask).Run(0xc0001dea50, {0x26acef0, 0x37ed020})
	/home/runner/work/signoz/signoz/pkg/query-service/rules/ruleTask.go:129 +0x488
go.signoz.io/signoz/pkg/query-service/rules.(*Manager).addTask.func1()
[...]

PR #5966, opened 2024-09-13
fix: check alert rule queries are all disabled if at least one query is set

kobecal:
### Summary
A alert rule is created and it's all queries are disabled. Then alert manager will get panic because of no matched query result to trigger alert.

#### Related Issues / PR's

https://github.com/SigNoz/signoz/issues/5964

#### Screenshots

NA

create, update alert rule


pkg/query-service/rules/api_params.go
```diff
@@ -168,18 +168,49 @@ func isValidLabelValue(v string) bool {
...
+}
+
 func (r *PostableRule) Validate() error {
 
 	var errs []error
 
 	if r.RuleCondition == nil {
 		errs = append(errs, errors.Errorf("rule condition is required"))
+		// will get panic if we try to access CompositeQuery, so return here
+		return multierr.Combine(errs...)
```
srikanthccv, line 203: Let's just return `errors.Errorf("rule condition is required")`

pkg/query-service/rules/api_params.go
```diff
@@ -168,18 +168,49 @@ func isValidLabelValue(v string) bool {
...
 }
 
+func isAllQueriesDisabled(compositeQuery *v3.CompositeQuery) bool {
+	if compositeQuery == nil {
+		return false
+	}
+	if compositeQuery.BuilderQueries == nil && compositeQuery.PromQueries == nil && compositeQuery.ClickHouseQueries == nil {
+		return false
+	}
+	for _, query := range compositeQuery.BuilderQueries {
```
srikanthccv, line 178: Check for the `QueryType` and use the corresponding query map.

srikanthccv (approved):

PR #5968, opened 2024-09-13
feat: send last log line time stamp for timestamp order-by desc

vikrantgupta25:
### Summary

- send last log line timestamp for list panel type when the orderBy is `timestamp desc` and is on the first position of orderBy 

#### Screenshots

NA


nityanandagohain (approved):
please create a pr for handing the relative timestamp issue where start also changes.

YounixM (approved):

PR #6000, opened 2024-09-18
fix: enrich attributes regardless if it is materialized

nityanandagohain:
Changes
* If a column is selected and later it was removed, or the dashboard is copied to an instance where it is not selected one then we should enrich regardless so that the query doesn't fail.
* Take care if the same key name is present in resource and tag or have multiple dataType.

Fixes https://github.com/SigNoz/engineering-pod/issues/1774


pkg/query-service/app/logs/v3/enrich_query_test.go
```diff
@@ -393,6 +394,71 @@ var testEnrichParamsData = []struct {
...
+			},
+		},
+		Fields: map[string]v3.AttributeKey{
+			"mat_resource": {
+				Key:      "mat_resource",
+				Type:     v3.AttributeKeyTypeResource,
+				DataType: v3.AttributeKeyDataTypeInt64,
+				IsColumn: false,
+			},
+			"mat_attr": {
```
srikanthccv, line 427: fields being map structure means only one item with a key can exist. what happens if there are conflicting types, the user selection should be preferred?

nityanandagohain: This condition takes care tat if the type doesn't match it won't update https://github.com/SigNoz/signoz/blob/0711c8855ea5c795202ea68fde9a75cf05119b1b/pkg/query-service/app/logs/v3/enrich_query.go#L150 .

but let me fix this in a better way

nityanandagohain: have made some changes please, have a look I have also updated the description.

pkg/query-service/app/logs/v3/enrich_query.go
```diff
@@ -144,13 +172,9 @@ func enrichFieldWithMetadata(field v3.AttributeKey, fields map[string]v3.Attribu
...
-	if existingField, ok := fields[field.Key]; ok {
-		// don't update if type is not the same
-		if (field.Type == "" && field.DataType == "") ||
-			(field.Type == existingField.Type && field.DataType == existingField.DataType) ||
-			(field.Type == "" && field.DataType == existingField.DataType) ||
-			(field.DataType == "" && field.Type == existingField.Type) {
-			return existingField
+	for _, key := range generateEnrichmentKeys(field) {
+		if val, ok := fields[key]; ok {
+			return val
```
srikanthccv, line 177:
```
	if field.Type != v3.AttributeKeyTypeUnspecified && field.DataType != v3.AttributeKeyDataTypeUnspecified {
		names = append(names, field.Key+"##"+field.Type.String()+"##"+field.DataType.String())
		return names
	}
```

Assume a user has requested with `type: tag` and `datatype: string` for a key `http.status_code` as an example. However, we only have a number type for it in DB. With this logic, it will continue to search for strings because `http.status_code##tag##string` doesn't exist in  `generateEnrichmentKeys` and the fallback is to use whatever the user provided. What is the sensible thing to do here? Reject the request saying there is no `http.status_code` with string?

srikanthccv: Since we are touching the enrichment part, I am trying to highlight the problems/cases so we can have a definitive solution to enrichment and never worry about it again.

nityanandagohain: We should run the query as it is in that case and let the query fail.

So we can say that enrich query
* If type and dataType is provided for an attribute, it will enrich if it's materialized or dematerialized.
* If one of type and dataType or both is not present, it will try to get whatever possible from the existing fields.
   * works correctly if there are no duplicate in attribute/resources or in dataType ( might be incorrect if there are duplicates)


So with this we can say that if user has provided, type, dataType and attribute in that case we should respect the user and just see if there is a corresponding materialized column. We would also want to do this for an attribute to arrive in the future.

srikanthccv: > We would also want to do this for an attribute to arrive in the future.

There is no contention about this case.

> We should run the query as it is in that case and let the query fail. ( return empty results)

I don't agree especially because when we know it's going to fail and alternative metadata would prevent it from happening we shouldn't knowingly fail. The main problem with these kinds of issues is they are silent and can be very frustrating to troubleshoot. Would you accept this if you were at the receiving end of the result? 

You can choose not to solve it right now by making a case that the likelihood of this happening is low and that you would rather get this change unblocked, but it's an entirely different thing to say we should let the query fail.

pkg/query-service/model/logs.go
```diff
@@ -49,7 +49,8 @@ func GetLogFieldsV3(ctx context.Context, queryRangeParams *v3.QueryRangeParamsV3
 				if pass {
 					continue
 				}
-				data[selectedField.Name] = v3.AttributeKey{
+				name := selectedField.Name + "##" + fieldType.String() + "##" + selectedField.DataType
+				data[name] = v3.AttributeKey{
```
srikanthccv, line 53: The change in the key format needs to be handled wherever `GetLogFieldsV3` is used. A quick search says it's used to create links. Please check.

pkg/query-service/app/logs/v3/enrich_query.go
nityanandagohain: > very frustrating to troubleshoot.
I don't think that should be the case, as suggestions will help the user choose the correct attribute for querying. In case of API's user might actually search for that attribute and changing it when they are actually providing something they want to use might not be a good idea according to me. What do you think ?

pkg/query-service/model/logs.go
nityanandagohain: oh right, while all others are using it for enrichment, `metaForLinks` is using it differently.

pkg/query-service/app/logs/v3/enrich_query.go
srikanthccv: >I don't think that should be the case, as suggestions will help the user choose the correct attribute for querying

The focus shouldn't be solely on the search bar as the source of this issue. The problem can manifest in various scenarios. Consider a situation where a user has a saved view, but the associated metadata no longer aligns with what's currently in the database. This metadata mismatch can occur due to factors beyond the user's control, such as third-party instrumentation changes or a colleague modifying data types. Why has a previously functional query stopped yielding results? How can users identify what has changed between the time when the query produced results and now when it returns nothing? 

> In case of API's user might actually search for that attribute and changing it when they are actually providing something they want to use might not be a good idea according to me.

Not sure if it came across that way, but I didn't mean to suggest we change it automatically without their knowledge. I was asking what would be the sensible thing to do in such cases here: https://github.com/SigNoz/signoz/pull/6000#discussion_r1764863495

nityanandagohain: Got your point.

So if user has provided a type and datatype and it doesn't exist, fallback to whatever is possible from the fields api is it ?

srikanthccv: I don't have a solution. I am just thinking out loud and highlighting where the enrichment logic breaks and leads to poor user experience.

nityanandagohain: Agree,

Also on a side note as of now these fields that we store are permanent and not bound to time. Which means let say a user sends an attribute x with type a. It will forever remain in the db that there is an attribute x with type a, even if the type changes in future. Unless we clear the db.

So in some cases we manually go ahead and truncate the attribute_keys table so that user doesn't see old keys.

pkg/query-service/model/logs.go
nityanandagohain: Thanks have made the changes for this. I didn't past the alertType as passing it results in cyclic dependency so went ahead with a boolean flag.

pkg/query-service/app/logs/v3/enrich_query.go
nityanandagohain: W.r.t respect to the above logic I the current implementation should be fine, since once a attribute is ingested it will remain in context forever. What do you think ?

nityanandagohain:
@srikanthccv  please leave your thoughts here if we should proceed with this and keep the concerns in an issue.

srikanthccv:
Please re-request a review after making the changes; otherwise, it doesn't appear on my "awaiting review from you" tab.

pkg/query-service/contextlinks/links.go
```diff
@@ -146,7 +147,7 @@ func PrepareLinksToLogs(start, end time.Time, filterItems []v3.FilterItem) strin
 // by clause, in which case we replace it with the actual value for the notification
 // i.e Severity text = WARN
 // If the Severity text is not part of the group by clause, then we add it as it is
-func PrepareFilters(labels map[string]string, whereClauseItems []v3.FilterItem, groupByItems []v3.AttributeKey, keys map[string]v3.AttributeKey) []v3.FilterItem {
+func PrepareFilters(alertTypeLogs bool, labels map[string]string, whereClauseItems []v3.FilterItem, groupByItems []v3.AttributeKey, keys map[string]v3.AttributeKey) []v3.FilterItem {
```
srikanthccv, line 150: Please remove the logs specific logic from here. The intent of this is to be generic. Treat the `labels` as if they are user given filters and enrich them appropriately.

pkg/query-service/app/logs/v3/enrich_query.go
srikanthccv: I had to take some time to recollect the points of discussion. Let's handle this in a separate issue.

srikanthccv (approved):

Issue #6002, opened 2024-09-18
Support caching multiple time ranges for same cache key


srikanthccv:
The current caching mechanism supports only caching one full-time range of data to cache. Here is the case where this is sub-optimal.

1. for a query, the user requests time series data for the last day, we fetch and cache it
2. next, users request data for the last 30 minutes now because we only maintain one entry in the cache, the last day cached data is now replicated with the most recent 30-minute data.

This is clearly the sub-optimal solution to replace the existing cache. Implement a solution that supports caching for multiple time ranges.

PR #6008, opened 2024-09-18
chore: add support for caching multiple time ranges for cache key

srikanthccv:
### Summary

Fixes https://github.com/SigNoz/signoz/issues/6002


nityanandagohain (approved):
Looks good to me, though do some sanity testing on actual prod data.

PR #6009, opened 2024-09-18
fix: added time range key for query and local storage handling

vikrantgupta25:
### Summary

- https://github.com/SigNoz/signoz/pull/5714 in this PR we started respecting the time range in the variables query. so when time range changes the options returned changes. 
- also the time parameters were not present in query key so the API was not getting triggered as well. 
- now since options changes and current selected is `ALL` the local storage values needs to be updated.

#### Screenshots

NA

### DIFF
--- a/frontend/src/container/GridCardLayout/GridCard/index.tsx
+++ b/frontend/src/container/GridCardLayout/GridCard/index.tsx
@@ -11,6 +11,7 @@ import { isEqual } from 'lodash-es';
 import isEmpty from 'lodash-es/isEmpty';
 import { useDashboard } from 'providers/Dashboard/Dashboard';
 import { memo, useEffect, useRef, useState } from 'react';
+import { useQueryClient } from 'react-query';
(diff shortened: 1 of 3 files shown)


YounixM (dismissed):

SagarRajput-7 (dismissed):

YounixM (dismissed):

srikanthccv:
In the result query, I could see two kinds of issues: 1. the list of values selected for the variable is different, and 2. the value for multi-select is sent as a single string instead of a list of strings with one item. Does this change fix both of them?

vikrantgupta25:
> Does this change fix both of them?

yes both of them have been fixed. let me know when you have tested the same. thank you 🙌

srikanthccv:
LGTM.

Things tested

- the multi-select single selection
- global time change and the variables change
- the panel final `query_range` call with correct variable values.

YounixM (approved):

PR #6034, opened 2024-09-19
fix: tlemetry for dashboard/alerts/views using contains on attributes

nityanandagohain:
* for alerts, dashboards and saved views checking how many are using `like` with attributes.
* update existing telemetry to check for clickhouse query only if the query type is clickhouse_sql.


pkg/query-service/app/dashboards/model.go
```diff
@@ -532,11 +531,35 @@ func extractDashboardName(data map[string]interface{}) string {
...
+func checkLogPanelAttrContains(data map[string]interface{}) int {
+	var logsPanelsWithAttrContains int
+	filters, ok := data["filters"].(map[string]interface{})
+	if ok && filters["items"] != nil {
+		items, ok := filters["items"].([]interface{})
+		if ok {
+			for _, item := range items {
+				itemMap, ok := item.(map[string]interface{})
+				if ok {
+					if slices.Contains([]string{"contains", "ncontains", "like", "nlike"}, itemMap["op"].(string)) {
```
srikanthccv, line 543: add assertion for `itemMap["op"].(string)` otherwise it will panic.

srikanthccv, line 543: why do we want to count `like` and `ilike` here because they do not get converted to `ilike` right?

pkg/query-service/app/dashboards/model.go
```diff
@@ -559,20 +582,33 @@ func countPanelsInDashboard(data map[string]interface{}) model.DashboardsInfo {
...
 					}
 				}
 			}
 		}
 	}
+
+	logChQueryCount := 0
+	if logChQuery {
+		logChQueryCount = 1
+	}
```
srikanthccv, line 605: What is happening here?

pkg/query-service/app/dashboards/model.go
nityanandagohain: all of them are mapped to like and ilike

pkg/query-service/app/dashboards/model.go
nityanandagohain: this is just telling that this dashboard has atleast one panel with clickhouse query.

pkg/query-service/app/dashboards/model.go
nityanandagohain: thanks 👍

srikanthccv (approved):

PR #6037, opened 2024-09-19
fix(install-script): deprecate external docker-compose

prashant-shahi:
### Summary

This PR deprecates external docker-compose from install script.
Instead it uses the bundled `docker compose`.


srikanthccv (approved):

prashant-shahi:
We will need to update the install script to include docker compose plugin - which is an add-on that may not be bundled for all Linux distro.

Will work on this when some bandwidth is freed.

PR #6045, opened 2024-09-20
feat: support for case insensitive for contains and like queries

nityanandagohain:
This PR adds case insensitive support to like/ilike/contains/ncontains

## Attributes
* We are taking the tradeoff that contains/ncontains/like/ilike won't use the index.

## Resources
* As of now we have two index for resources one is lowecase and the other one is uppercase.
* Since we want case insensitivity for  like/ilike/contains/ncontains , we are using the lowercase index.

For body and resource we are making sure case insensitive search works by using like with lower(column).
 

For https://github.com/SigNoz/signoz/issues/5555


srikanthccv (approved):

nityanandagohain:
Have made a small change with respect to index filter to build it with lowercase which was missed, please have a look at the last commit.

srikanthccv (approved):

PR #6051, opened 2024-09-23
fix: telemetry nil pointer error fix

nityanandagohain:
Fixes nil pointer error for heartbeat func


prashant-shahi (approved):

prashant-shahi:
FYI, there are other instances of the missing error handling in the same block.

nityanandagohain:
those are fine, we are not looking inside them.
if they are nil they will be sent as nil.

PR #6059, opened 2024-09-24
fix: added support for `body contains X` tag on pressing enter after selecting attribute key

vikrantgupta25:
### Summary

- added support for `body contains X` tag on pressing enter after selecting attribute key
- added an empty bar in the operator options which allows users to add `body contains x` tags quickly 
- added the support to quickly convert the space separated keys to `body contains` quickly 
- added a extra key on the top matching the searched value to help select the non dropdown values quickly 

#### Related Issues / PR's

fixes https://github.com/SigNoz/engineering-pod/issues/1817

#### Screenshots

https://github.com/user-attachments/assets/bcd3a76b-8931-43a5-883e-b7c669b2594b


YounixM (dismissed):

SagarRajput-7 (dismissed):

nityanandagohain (changes requested):
These quereis not work
* serv abcd def ( matches somethig from the attribute).
* abcd def

YounixM (approved):

Issue #5975, opened 2024-09-14
EPIC: support dots in metrics names and labels


srikanthccv:
- [ ] update alertmanager to use the latest prometheus/common and use utf-8 mode
- [ ] introduce clickhousemetriwritev2 that writes to current `samples_v4` and `time_series_v4` tables without normalizing
- [ ] separate the resource attributes and metric data point attributes
- [ ] update query service to hide (from querying/ui) the dot-supported metrics/labels till the data is backfilled.
- [ ] update signoz/prometheus to the latest upstream and set the utf-8 mode
- [ ] update the formulas to work with dotted labels in clickhouse
- [ ] automatically upgrade the query-builder queries to use resource attributes/data point attributes with dots
- [ ] offer an automated but human verifiable transition with --dry-run for promql and clickhouseql queries
- [ ] deprecate and remove the normalization of metrics

Ruppsn:
Hello, i just wanted to update my dahsboards (https://signoz.io/guides/metrics-migration/) to the new dot metrics but it seems to me that signoz is not presenting them to me. i have updated all the configs with the signozclickhousemetrics exporter as in deploy/docker-swarm/otel-collector-config.yaml. The collector is running with the new configuration for 2 weeks. 

![Image](https://github.com/user-attachments/assets/e754827a-8b11-4856-9376-464e4de25585)

If i understand the changelogs than both metric names should be usable/findable right now?!

srikanthccv:
Hi @Ruppsn, no currently only normalised metrics are visible (aka underscore metrics). There is one pre-requisite for enabling the dot metrics, that is, run the sqlite migration script to update the existing dashboards and alerts to use the new non-normalised metrics. We haven't published the docs for this yet. We plan to publish the steps for migration tentatively by next week.

srikanthccv:
@Ruppsn upgrade instructions https://signoz.io/docs/operate/migration/upgrade-0.88

srikanthccv:
This is complete now.

PR #6062, opened 2024-09-24
feat: add dot support for alerts

srikanthccv:
### Summary


nityanandagohain (approved):
The changes here looks good according to what you have done that  is previously we stored both the original labels and the normalized ones, now you have added only the original ones and for backward compatibility also considered the old normalized ones.

While this looks good, I don't know what extra edge cases might come up if any. But the label creation looks correct.

PR #6087, opened 2024-09-27
fix: fixed variable dropdown closing on select

SagarRajput-7:
#### Screenshots

https://github.com/user-attachments/assets/4f3889b6-c9d0-42dd-aacd-ddb28334e70e


srikanthccv:
@SagarRajput-7, it doesn't prevent it from closing, it seems like it closes and then opens again. Why can't we make it not close in the first place and avoid jerky behaviour on each selection, which is not a good UX?

srikanthccv:
Looked at the code just out of curiosity. I do not think this is a good solution to make it open by click event.

SagarRajput-7:
> Looked at the code just out of curiosity. I do not think this is a good solution to make it open by click event.

@srikanthccv - yes, that is why this PR is in work in progress, i am working on the UX and behaviour only, I just mentioned this PR there, to say that this issue is also in motion.

srikanthccv:
ok, it would have been best to keep it in the draft in that case. As you attached the video and requested for review I thought this was ready for review and the wip was about testing etc...

PR #6095, opened 2024-09-30
chore: bump SigNoz/prometheus

srikanthccv:
### Summary

Fixes https://github.com/SigNoz/signoz/issues/6068


srikanthccv:
In this commit https://github.com/SigNoz/prometheus/commit/6a1881c34a2f0d6f6e9c4512d385f2f060f06f24, I update our fork with upstream changes. The latest upstream support using dot in promql queries with quotes for example `signoz_calls_total{"host.name"="localhost"}`. However, to enable that we need to set the `NameValidationScheme` to utf-8 instead of legacy. You can read a bit about it here https://github.com/prometheus/common/blob/dae848db5327d2a4e2e06cbe883093a71b4226d7/model/metric.go#L27-L44. This needs to be done at the startup so we set the name validation using `init`.

shivanshuraj1333 (approved):

go.mod
```diff
@@ -1,9 +1,11 @@
 module go.signoz.io/signoz
 
-go 1.21.3
+go 1.22.0
+
+toolchain go1.22.7
```
nityanandagohain, line 5: this is required for the changes present in this PR ?

nityanandagohain (approved):

go.mod
srikanthccv: Yes

PR #6096, opened 2024-09-30
chore: skip showing metrics with dot in name

srikanthccv:
### Summary

Fixes https://github.com/SigNoz/signoz/issues/6067

We will write to the storage but don't want to show the metrics with dots in the suggestions till we ensure everything works. The existing data write path is not distributed, i.e. what worked earlier would continue to work

It could be done at the query itself, but the number of metrics names is not going to be much and importantly I want to filp the flag and allow dots in upcoming so I don't want to change query back later.

----

> [!IMPORTANT]
> Skip metrics with dots in their names from suggestions by adding a `skipDotNames` parameter to `GetMetricAggregateAttributes`.
> 
>   - **Behavior**:
>     - Skip metrics with dots in their names in suggestions by adding `skipDotNames` parameter to `GetMetricAggregateAttributes` in `reader.go`.
>     - Update `autocompleteAggregateAttributes` in `http_handler.go` to pass `true` for `skipDotNames`.
>   - **Interfaces**:
>     - Add `skipDotNames` parameter to `GetMetricAggregateAttributes` in `interface.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 63b21929b932715ab0f6eae8379d6af806475cd5. It will automatically update as commits are pushed.</sup>


shivanshuraj1333 (approved):

pkg/query-service/app/http_handler.go
```diff
@@ -3783,7 +3783,7 @@ func (aH *APIHandler) autocompleteAggregateAttributes(w http.ResponseWriter, r *
 
 	switch req.DataSource {
 	case v3.DataSourceMetrics:
-		response, err = aH.reader.GetMetricAggregateAttributes(r.Context(), req)
+		response, err = aH.reader.GetMetricAggregateAttributes(r.Context(), req, true)
```
nityanandagohain, line 3786: should be a flag  instead of hardcoding ?

srikanthccv: No, we don't want to show any metrics with a dot in its name till we make sure working with dots works in all kinds of queries.

nityanandagohain (approved):

PR #6097, opened 2024-09-30
Add onboarding APIs to check the attributes for Messaging Queues feature

shivanshuraj1333:
Fixes: https://github.com/SigNoz/engineering-pod/issues/1834
Epic: https://github.com/SigNoz/engineering-pod/issues/1584

These APIs would help identify the attributes that need to be there for correlation, it also helps in health checks to verify that the user is sending correct data.

For producers, consumers and kafka, it returns the corresponding checks and returns attributes which are not present.
Screenshot 2024-10-04 at 1 31 25 AM

Example of the UI https://www.notion.so/signoz/Engineering-Design-Doc-for-Kafka-d9db4d601519404c9e93d5f509efd037?pvs=4#d21d7ac0d983455f9ce824c9f6355d0b

### DIFF
--- a/pkg/query-service/app/http_handler.go
+++ b/pkg/query-service/app/http_handler.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"regexp"
 	"slices"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


srikanthccv:
1. I don't think you should use the `QueryRange` for these data quality checks. You can instead introduce new methods to do this job
2. is it possible to have more than one condition unsatisfied at any point? Ex `messaging.destination` and `messaging.message.body.size` do not exist. How is that handled?

srikanthccv (approved):

Issue #5470, opened 2024-07-12
Consider `materialize_ttl_after_modify=0` for TTL modification 


srikanthccv:
The `materialize_ttl_after_modify` by default is set to 1 and triggers the TTL materialisation process. This can be a heavy operation with a lot of I/O. We can disable this by setting `materialize_ttl_after_modify=0`. It will leave the old data to have old TTL rules and new data with modified TTL. I believe this is a great option for many and whoever runs the setup as it avoids the IO and allows the TTL to be updated easily. Users should be fine with this change most of the time, however, if they desire to apply the TTL for old parts, then they can override the setting.

srikanthccv:
Example CPU usage when TTL is being materialized.
![Screenshot 2024-08-13 at 12 51 53 AM](https://github.com/user-attachments/assets/cf5395e9-896e-48aa-8407-23a59efa7f6a)

PR #6106, opened 2024-10-02
chore: do not materialize ttl after modify

srikanthccv:
### Summary

Fixes https://github.com/SigNoz/signoz/issues/5470

### DIFF
--- a/pkg/query-service/app/clickhouseReader/reader.go
+++ b/pkg/query-service/app/clickhouseReader/reader.go
@@ -2222,7 +2222,7 @@ func (r *ClickHouseReader) SetTTL(ctx context.Context,
 				return
 			}
 			timeColumn := "timestamp_ms"
-			if strings.Contains(tableName, "v4") {
+			if strings.Contains(tableName, "v4") || strings.Contains(tableName, "exp_hist") {
 				timeColumn = "unix_milli"
 			}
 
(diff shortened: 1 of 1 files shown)


prashant-shahi (dismissed):
can you include the following table for Logs TTL updation as well?

- `signoz_logs.logs_v2`
- `signoz_logs.logs_v2_resource`

srikanthccv:
They are already covered in v2 handling. Please take a few minutes to read the code!

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -2218,7 +2247,7 @@ func (r *ClickHouseReader) SetTTL(ctx context.Context,
 				}
 				return
 			}
-			req += " SETTINGS distributed_ddl_task_timeout = -1"
```
makeavish, line 2221: @srikanthccv Why are we removing this setting?

srikanthccv: Why did we add it in the first place?

srikanthccv: It doesn't take any time (because it would just be a metadata update) if we do not materialize(the change this PR proposes).

makeavish: Great then it's not required, I hope you have tested these changes on a large DB.

makeavish (approved):

prashant-shahi:
> They are already covered in v2 handling.

Right, I remember now it was updated by Nitya in another PR.

I mentioned it since I didn't see it over the file changes of the PR.

Issue #5373, opened 2024-06-27
Infra monitoring (Out of the box module)


pranay01:
Today you can import dashboards for monitoring infrastrcuture like VMs, Kubernetes clusters in SigNoz, but there is no dedicated infra monitoring module ( may be as a tab in the left side bar) where users can go and out of the box get status of their infrastructure. Building such a module would enable users to get understanding of the infra out of the box

It should have out of box metrics for 

- [ ] VMs
- [ ] K8s Clusters
- [ ] ECS clusters
- [ ] Fargate nodes 
etc

ankitnayan:
+ show APM, spans and logs from the selected host

pranay01:
PRD - https://www.notion.so/signoz/PRD-Show-infrastructure-lists-more-easily-Infra-Monitoring-module-9a4bdc889dcc4f0cb06b9377d10a9570

pranay01:
- Initial focus is on completing the correlation story with infra module
- Lists of Hosts & K8s clusters

PR #6123, opened 2024-10-04
chore: add hosts list support

srikanthccv:
### Summary

Part of https://github.com/SigNoz/signoz/issues/5373

The v0 planned items here https://www.notion.so/signoz/V0-infra-monitoring-items-a3e94f6e80a94b33b42874d2f414edeb

This change adds the v0.1 item mentioned in https://www.notion.so/signoz/Infra-monitoring-epic-iterations-plan-05407ddb3d6a4ea7a0f59093cda93fc8, i.e. the ability to see the list of hosts with their status, important metrics.


shivanshuraj1333 (dismissed):
LGTM

pkg/query-service/app/http_handler.go
```diff
@@ -344,6 +350,13 @@ func (aH *APIHandler) RegisterQueryRangeV3Routes(router *mux.Router, am *AuthMid
 	subRouter.HandleFunc("/logs/livetail", am.ViewAccess(aH.liveTailLogs)).Methods(http.MethodGet)
 }
 
+func (aH *APIHandler) RegisterInfraMetricsRoutes(router *mux.Router, am *AuthMiddleware) {
+	subRouter := router.PathPrefix("/api/v1/hosts").Subrouter()
+	subRouter.HandleFunc("/attribute_keys", am.ViewAccess(aH.getHostAttributeKeys)).Methods(http.MethodPost)
+	subRouter.HandleFunc("/attribute_values", am.ViewAccess(aH.getHostAttributeValues)).Methods(http.MethodPost)
```
nityanandagohain, line 356: Why all of these are post request ? it can be easily catered by GET method.

pkg/query-service/app/inframetrics/hosts.go
```diff
@@ -0,0 +1,506 @@
...
+		Temporality: v3.Unspecified,
+		GroupBy:     req.GroupBy,
+	}
+	query, err := helpers.PrepareTimeseriesFilterQuery(req.Start, req.End, &mq)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(srikanthccv): remove this
+	query = strings.Replace(query, ".time_series_v4", ".distributed_time_series_v4", 1)
```
nityanandagohain, line 153: Please add a comment here stating that. 

query can contain 
`time_series_v4, time_series_v4_6hrs, time_series_v4_1day` and you are adding distributed as a prefix to them, otherwise requires a lot of looking around

pkg/query-service/app/inframetrics/hosts.go
```diff
@@ -0,0 +1,506 @@
...
+						Key:      "system_cpu_load_average_15m",
+						DataType: v3.AttributeKeyDataTypeFloat64,
+					},
+					Temporality:      v3.Unspecified,
+					Filters:          req.Filters,
+					GroupBy:          req.GroupBy,
+					Expression:       "A",
+					TimeAggregation:  v3.TimeAggregationAvg,
+					SpaceAggregation: v3.SpaceAggregationAvg,
+					Disabled:         false,
```
nityanandagohain, line 218: we should a limit for group by or is it fine like this ?

pkg/query-service/app/inframetrics/table_list_query.go
```diff
@@ -0,0 +1,337 @@
...
+						Key:      "os_type",
+						DataType: v3.AttributeKeyDataTypeString,
+						Type:     v3.AttributeKeyTypeResource,
+					},
+				},
+				Expression:       "A",
+				ReduceTo:         v3.ReduceToOperatorAvg,
+				TimeAggregation:  v3.TimeAggregationRate,
+				SpaceAggregation: v3.SpaceAggregationSum,
+				Disabled:         true,
```
nityanandagohain, line 55: here also a limit is required or not to prevent a blast ?

pkg/query-service/app/querier/v2/querier.go
```diff
@@ -171,6 +173,7 @@ func (q *querier) runBuilderQueries(ctx context.Context, params *v3.QueryRangePa
 
 	wg.Wait()
 	close(ch)
+	zap.L().Info("time taken to run builder queries", zap.Duration("duration", time.Since(now)))
```
nityanandagohain, line 176: we already have duration for elapsed time, rename it to sometime else ? like querier duration it will be easier to distinguish while filtering from the ui ?

pkg/query-service/app/inframetrics/hosts.go
srikanthccv: The limit here won't do anything for metrics. The restriction on metric aggregation should be controlled using query limits, which is a separate exercise. But generally, we expect the number of hosts to be several thousand, so it should not be an issue. However, there are some metrics, such as disk, and network which will add noisy devices and cause issues. We address it by 1. query limits (separate PR https://github.com/SigNoz/signoz/pull/3956) 2. not collecting such datahttps://github.com/SigNoz/charts/pull/516

nityanandagohain (approved):

pkg/query-service/app/http_handler.go
```diff
@@ -344,6 +350,13 @@ func (aH *APIHandler) RegisterQueryRangeV3Routes(router *mux.Router, am *AuthMid
 	subRouter.HandleFunc("/logs/livetail", am.ViewAccess(aH.liveTailLogs)).Methods(http.MethodGet)
 }
 
+func (aH *APIHandler) RegisterInfraMetricsRoutes(router *mux.Router, am *AuthMiddleware) {
+	subRouter := router.PathPrefix("/api/v1/hosts").Subrouter()
+	subRouter.HandleFunc("/attribute_keys", am.ViewAccess(aH.getHostAttributeKeys)).Methods(http.MethodGet)
+	subRouter.HandleFunc("/attribute_values", am.ViewAccess(aH.getHostAttributeValues)).Methods(http.MethodGet)
+	subRouter.HandleFunc("/list", am.ViewAccess(aH.getHostList)).Methods(http.MethodPost)
```
nityanandagohain, line 357: this can be get as well no ?

srikanthccv: No, for the same reason why query_range is not a get request (technically it could be but our JSON-based filter, and orderby etc.. make it complicated to put into URL query params).

nityanandagohain: ok

PR #6125, opened 2024-10-05
chore: add process list

srikanthccv:
### Summary

Part of https://github.com/SigNoz/signoz/issues/5373

The v0 planned items here https://www.notion.so/signoz/V0-infra-monitoring-items-a3e94f6e80a94b33b42874d2f414edeb

This change adds the v0.1-0.3 item mentioned in https://www.notion.so/signoz/Infra-monitoring-epic-iterations-plan-05407ddb3d6a4ea7a0f59093cda93fc8, i.e. the ability to see the list of processes with name, cpu and memory metrics.


shivanshuraj1333:
@srikanthccv this needs some resolution, overall LGTM

pkg/query-service/app/inframetrics/processes.go
```diff
@@ -0,0 +1,334 @@
...
+		groupKey += fmt.Sprintf("%s=%s,", key.Key, record.Meta[key.Key])
+	}
+	return groupKey
+}
+
+func (p *ProcessesRepo) getMetadataAttributes(ctx context.Context,
+	req model.ProcessListRequest) (map[string]map[string]string, error) {
+	processAttrs := map[string]map[string]string{}
+
+	keysToAdd := []string{"process_pid", "process_executable_name", "process_command", "process_command_line"}
```
shivanshuraj1333, line 79: same here, as it'd be easier for you to change it in the code later on

shivanshuraj1333 (approved):

shivanshuraj1333:
Some of the abstraction can be done if you a data model for the metrics, however, this structuring of the code might require some time. For now maybe having a common file, something like constant.go or metrics_names.go can be used to put the const metrics name. But that'd wouldn't be a blocker, if you want to merge the PR, go ahead.

srikanthccv:
@shivanshuraj1333 Good points about the constant metric names. I will add them in follow-up PR as I will be doing similar stuff in upcoming PRs. Can you elaborate on the abstraction part? Are you referring to the part where the hosts and process list look mostly the same? If so, I agree with that but personally I let the codebase evolve over some time and then refactor with better abstractions.

PR #6138, opened 2024-10-08
feat(log explorer): button style fixes

sharpshooter90:
- switch button for old/new UI style and text update for better space utilization and CTA priority
- improvised the query run button (no jerking and consistant pattern)

#### Related Issues / PR's
https://github.com/SigNoz/engineering-pod/issues/1871

#### Screenshots
image
image
image
image

image

----

> [!IMPORTANT]
> Improves button text and styles in log explorer and query builder components for better UI consistency.
> 
>   - **Button Text Updates**:
>     - Simplified button text in `config.ts` for UI switching: 'Old UI' and 'New UI'.
>   - **Button Style Changes**:
>     - Removed `danger` and `type` attributes from `Button` in `index.tsx`.
>     - Added `periscope-btn` class to `Button` in `index.tsx`.
>     - Set `minWidth` for buttons in `RightToolbarActions.tsx` for consistent size.
>   - **Style Cleanup**:
>     - Removed unused styles for `.loading-btn` and `.cancel-run` in `ToolbarActions.styles.scss`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for cc0694ae76c2eb9c418b7c9c00b0861ccc849b86. It will automatically update as commits are pushed.</sup>


GeekBoySupreme:
@sharpshooter90 I'd recommend against using this pattern for the cancel run button. When a button reflects an in progress or running state, I'd expect it to be disabled. Cancel run is an explicit action and we shouldn't conflagrate it with the loading state reflecting the running query.

For the other button, for now replace the UI in the button copy with Explorer and change the icon from the compass used for explorers to something like this. This is definitely better than the older red button, but we had plans to move it altogether to the top bar on the right.
cc @vikrantgupta25

sharpshooter90:
@Vikrant, for the user, the term “UI” will be a good trigger. The majority of users are already in the new UI; this is to provide support for the old users. The intention is to migrate users to the new. We are all on the same page that the new feature has all the new tech buzz going on.

The old text “Switch to Old Logs Explorer” is also not conveying much about the tech innovation.

The goal here is to reduce the priority to secondary with styling and have a minimal UX copy pattern that is good enough to trigger but not taking up too much space and attention. Also, please think about the impact that is going to happen (major users are on the new, so the use case won’t be affecting major users).

@Vikrant @Shuvam, let me know if you guys have any future thoughts. We are going to remove it; this is the first step to remove but make it minimal, test with users. In the intermediate stage, we will warn users when they switch back to the old and say we are dropping this view in x days. Feel free to get in touch with the team for any more queries.

Final action: remove it completely.

Issue #6148, opened 2024-10-09
Sorting Alert Trigger List by Table Header Unexpectedly Duplicates Rows


sharpshooter90:
When attempting to sort the Alert Trigger List by any table header, the rows unexpectedly increase due to the duplication of the same row. This issue occurs across all browsers and is reproducible in the latest version.

**Steps to Reproduce**

	1.	Navigate to the Alert Trigger List view.
	2.	Click on any table header to sort by that column.
	3.	Observe that the number of the rows increases and the contents appear duplicated.

https://www.loom.com/share/4d7b47d8277b4191b976b8351a68ae72?sid=ea62503c-5277-4e21-87f2-f51fa376ea43

sharpshooter90:
@vikrantgupta25 cc

PR #6154, opened 2024-10-10
fix: updated row key in triggered alert list table

rahulkeswani101:
### Summary

The key for each row was the same, duplicating the row if we did the sorting in the triggered alert list. To solve this issue we have to update the key to be unique.

#### Related Issues / PR's
#6148 

#### Screenshots

https://github.com/user-attachments/assets/9d2aaefd-785f-443c-b801-071929c29f4e

- Tested the triggered alert list view and checked if the current implementation doesn't affect other functionalities on this page.
-  Tested that duplicate rows are not created.

----

> [!IMPORTANT]
> Update `rowKey` in `NoFilterTable` to ensure unique keys and prevent duplicate rows when sorting.
> 
>   - **Behavior**:
>     - Update `rowKey` in `NoFilterTable` to `${record.startsAt}-${record.fingerprint}` to ensure unique keys and prevent duplicate rows when sorting.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 875c7968069a6766a3882f7c7daba11e5f168348. It will automatically update as commits are pushed.</sup>


YounixM (approved):

ahmadshaheer (approved):

frontend/src/container/TriggeredAlerts/NoFilterTable.tsx
```diff
@@ -98,7 +98,7 @@ function NoFilterTable({
 	return (
 		<ResizeTable
 			columns={columns}
-			rowKey="startsAt"
+			rowKey={(record): string => `${record.startsAt}-${record.fingerprint}`}
```
vikrantgupta25, line 101: these attributes will always be present ?

vikrantgupta25: just checked it does. gtg

vikrantgupta25 (approved):

PR #6174, opened 2024-10-13
Fix: log pipelines generates bad config if first op is disabled

raj-k-singh:
fixes https://github.com/SigNoz/engineering-pod/issues/1889


nityanandagohain:
@raj-k-singh  can you provide some description about how it fixes the issue ?

nityanandagohain (approved):

PR #6198, opened 2024-10-16
fix: flicker in logs explorer page

vikrantgupta25:
### Summary

- fix the flicker being seen in the logs explorer page 

#### Related Issues / PR's

fixes part of - https://github.com/SigNoz/engineering-pod/issues/1885 

#### Screenshots

NA

### DIFF
--- a/frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/QueryBuilderSearchV2.tsx
+++ b/frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/QueryBuilderSearchV2.tsx
@@ -788,12 +788,6 @@ function QueryBuilderSearchV2(
 
 		if (!isEqual(query.filters, filterTags)) {
 			onChange(filterTags);
-			setTags(
-				filterTags.items.map((tag) => ({
-					...tag,
-					op: getOperatorFromValue(tag.op),
-				})) as ITag[],
-			);
 		}
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [tags]);


frontend/src/pages/EditRules/EditRules.styles.scss
```diff
@@ -11,20 +11,19 @@
...
-	font-style: normal;
-	font-weight: 300;
-	font-size: 18px;
-	line-height: 20px;
-	display: flex;
-	align-items: center;
-	justify-content: center;
-	text-align: center;
-	margin: 0;
+	.content {
```
sharpshooter90, line 15: @vikrantgupta25, this won’t completely fix the issue. The `.content ` styles from other places will still apply, and using nesting will only increase the specificity, overriding it. What we need to do is rename `.content` to something else, like `edit-rules-card__content` using the BEM naming convention. This will resolve the scoping issue since the new name is unique, whereas `.content` is widely used elsewhere in our code and in other libraries as well.

vikrantgupta25: We follow a principle of scoping styles to the parent container which increases the specificity of our styles. 
`.content` will be applied to any other un-intended area if and only if that follows a similar hierarchy. 

> .content is widely used elsewhere in our code and in other libraries as well.

As long as it is scoped properly to the parent container class name we won't face any issues. Adding global styles always causes issues so we should avoid doing that, which is being fixed in this PR. I haven't renamed the styles as didn't want the blast radius of this PR to be around alerts module as well. @ahmadshaheer  has been made aware of this and if needed we will fix that separately.

srikanthccv (approved):

YounixM (approved):

PR #6200, opened 2024-10-16
fix: issue while switching between views

### LINKED ISSUE #6187: Selecting view in `explorer-options` does not switch to view
## Bug description

When using the view drop down in the explorer it does not switch between views.
Tested with logs but might affect other environments too.
Clearing the view does work, and reloading post selecting does bring up the right view, but click the view does not select it.
If a view is already selected then clicking a different view doesn't switch the drop down either.

https://github.com/user-attachments/assets/7cc9d473-8063-45c4-a4aa-b34b436518be

## Expected behavior

The view picker should work as it did in 0.55.0 where clicking a view switches to it.

## How to reproduce

1. Create at least 2 views in logs
2. Try to switch between those views

## Version information 
* **Signoz version**: 0.56.0
* **Browser version**: Chrome 129.0.6668.101 
* **Your OS and version**: macOS 14.6.1

## Additional context

This bug was introduced with 0.56.0, in 0.55.0 the explorer did not exhibit this issue.

ahmadshaheer:
### Summary
the root cause of #6187 was, using `useUrlQuery` to get the current URL's search  params, we used to get the stale `viewName`, therefore, it would update the selectColumns with the old `viewName` being selected. thus it would keep the old view selected.

by creating `URLSearchParams` inside the `redirectWithQuery`, we ensure that we're working with the most up-to-date URL state.

#### Related Issues / PR's
close https://github.com/SigNoz/signoz/issues/6187
close https://github.com/SigNoz/engineering-pod/issues/1913

#### Screenshots

NA

Affecter areas:
- Logs
- Traces

Tested:
- switching between saved views in logs explorer
- switching between saved views in traces explorer
- creating a new view and switching to another view in logs explorer
- creating a new view and switching to another view in traces explorer

----

> [!IMPORTANT]
> Fixes URL query handling in `useUrlQueryData` by using `URLSearchParams` for up-to-date query state and corrects import path for `useUrlQuery`.
> 
>   - **Behavior**:
>     - Fixes URL query handling in `useUrlQueryData` by using `URLSearchParams` to ensure up-to-date query state.
>     - Updates `redirectWithQuery` to construct URLs with current pathname and updated query string.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 88c3fa8a64aafb91e1b37d027ca382076d5f7dae. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/src/hooks/useUrlQueryData.ts
+++ b/frontend/src/hooks/useUrlQueryData.ts
@@ -22,14 +22,16 @@ const useUrlQueryData = <T>(
 		(newQueryData: T): void => {
 			const newQuery = JSON.stringify(newQueryData);
 
-			// Create a new URLSearchParams to get the latest URL state
+			// Create a new URLSearchParams object with the current URL's search params
+			// This ensures we're working with the most up-to-date URL state
 			const currentUrlQuery = new URLSearchParams(window.location.search);
 
-			// Update the query parameter
+			// Update or add the specified query parameter with the new serialized data
 			currentUrlQuery.set(queryKey, newQuery);
 
-			// Generate the new URL with updated parameters
+			// Construct the new URL by combining the current pathname with the updated query string
 			const generatedUrl = `${location.pathname}?${currentUrlQuery.toString()}`;
+
 			history.replace(generatedUrl);
 		},
 		[history, location.pathname, queryKey],
--- a/frontend/src/hooks/useUrlQueryData.ts
+++ b/frontend/src/hooks/useUrlQueryData.ts
@@ -1,6 +1,7 @@
-import useUrlQuery from 'hooks/useUrlQuery';
 import { useCallback, useMemo } from 'react';
 import { useHistory, useLocation } from 'react-router-dom';
+
+import useUrlQuery from './useUrlQuery';
 
 const useUrlQueryData = <T>(
 	queryKey: string,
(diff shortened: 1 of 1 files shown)


vikrantgupta25 (approved):

frontend/src/hooks/useUrlQueryData.ts
```diff
@@ -21,11 +22,19 @@ const useUrlQueryData = <T>(
...
+			// Update or add the specified query parameter with the new serialized data
+			currentUrlQuery.set(queryKey, newQuery);
+
+			// Construct the new URL by combining the current pathname with the updated query string
+			const generatedUrl = `${location.pathname}?${currentUrlQuery.toString()}`;
+
 			history.replace(generatedUrl);
 		},
-		[history, location, urlQuery, queryKey],
+		[history, location.pathname, queryKey],
```
vikrantgupta25, line 37: curious that even the presence if urlQuery in the dependency of `useCallback` isn't letting it to get updated value ?

SagarRajput-7 (approved):

YounixM (approved):

frontend/src/hooks/useUrlQueryData.ts
ahmadshaheer: Yeah, it wasn't working even if urlQuery is in the dependencies of useCallback.

The reason is that the `urlQuery` object is created using `useMemo` in `useUrlQuery`, which means it retains the same reference unless the search value changes. Including `urlQuery` in the dependency array of `useCallback` does not trigger a re-evaluation of the callback unless the reference itself changes. By directly using new `URLSearchParams(window.location.search)` in the `redirectWithQuery` function, we ensure that we always work with the latest URL state, avoiding any stale closures.

PR #6227, opened 2024-10-21
chore: update hosts list to use pre-aggregated data table dynamically

srikanthccv:
### Summary

Builds on top of https://github.com/SigNoz/signoz/pull/5679

We use the pre-aggregated table to fetch top X items based on specified order by column. The order by can be `cpu`, `memory`, `iowait` or `load15`. We then fetch the time series data for the X hosts like regular data. I also incorporated some suggestions from the previous review.

----

> [!IMPORTANT]
> Update hosts list retrieval to dynamically use pre-aggregated data tables for improved query efficiency.
> 
>   - **Behavior**:
>     - Use pre-aggregated tables for fetching top X hosts based on `cpu`, `memory`, `iowait`, or `load15` in `hosts.go`.
>     - Introduce `getParamsForTopItems()` in `common.go` to determine aggregation step and table names based on time range.
>     - Implement `getTopHosts()` in `hosts.go` to fetch and sort top hosts dynamically.
>   - **Query Logic**:
>     - Modify `whichTSTableToUse()` and `WhichSamplesTableToUse()` in `sub_query.go` to support table hints and adjust start times.
>     - Add `MetricTableHints` to `BuilderQuery` in `v3.go` for overriding default table selection.
>   - **Misc**:
>     - Remove redundant metadata fetching logic in `hosts.go`.
>     - Update `OrderBy` to use `Direction` type in `v3.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 8a69e8848d4f23f59457d804d1e02b877f618a6e. It will automatically update as commits are pushed.</sup>


pkg/query-service/app/inframetrics/hosts.go
```diff
@@ -21,23 +19,45 @@ type HostsRepo struct {
...
+		"type",
+		"os_type",
+		"process_cgroup",
+		"process_command",
+		"process_command_line",
+		"process_executable_name",
+		"process_executable_path",
+		"process_owner",
+		"process_parent_pid",
+		"process_pid",
```
shivanshuraj1333, line 44: do you want to keep these as constants to be reused somewhere else (if there's a need)

shivanshuraj1333 (approved):
Did a semantic review, looks good to me.

pkg/query-service/app/inframetrics/hosts.go
srikanthccv: I am going to remove them soon (when we have a way to easily distinguish metric data point attributes and resource attributes) so I don't see the reusability of aspect of it.

Issue #3174, opened 2023-07-19
Filter traces with span count > x


makeavish:
Allow user to filter traces based on span count of trace

ankitnayan:
We need to evaluate if this is easily doable with good perf. @eKuG does DD or HC allow filtering traces by span count?

I am not really sure about the usecases too. Or just a `span_count` order by desc is also fine if users just want to see the traces with highest span_count

PR #6374, opened 2024-11-05
feat: added support for filterBy in trace tab such as filterBy spanCount

eKuG:
### Summary

Solves #3174 

#### Screenshots

NA

----

> [!IMPORTANT]
> Add support for filtering trace queries by `spanCount` using a new `filterBy` parameter in query functions.
> 
>   - **Behavior**:
>     - Added `filterBy` parameter to `runBuilderQuery()` in `helper.go` and `v2/helper.go` to support filtering by `spanCount`.
>     - Updated `PrepareTracesQuery()` in `query_builder.go` and `traces/v3/query_builder.go` to accept `filterBy` and incorporate it into query construction.
>   - **Models**:
>     - Added `FilterBy` field to `QueryRangeParamsV3` in `v3.go` to store filter criteria.
>   - **Functions**:
>     - Modified `buildTracesQuery()` in `traces/v3/query_builder.go` to handle `filterBy` parameter.
>   - **Misc**:
>     - Updated function signatures for trace query preparation to include `filterBy` in `query_builder.go` and `traces/v3/query_builder.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for fd3160526db3772a3d5df04ae2d7621d3df26656. It will automatically update as commits are pushed.</sup>


srikanthccv:
@eKuG I see two issues with this change.

1. `... GROUP BY traceID  ...`

This is not going to work at scale. We can't do a group by a very high cardinal column at query. The number of traces can go from a few tens of millions to several hundred million and billions easily. Adding such a column in a group by is not an option because there won't be enough memory. Did you test this on a few billion traces or several hundred million traces at least? 

2. FilterBy string in the request payload

I would assume the capability should support the ability to specify the span count in the request, for example, give me traces with span count > 1000 and < 2000. I don't see how that's supported.

eKuG:
Hey @srikanthccv Lets talk about the second point first.
So "filterBy" is an attribute used to filter traces by either span_count or trace_duration. It has no other purpose. (In hindsight, I think I should rename thiss column to something like "trace_filter"). Let me know what you think.
If we need to finf the traces where 1000<spanCount<2000, that will go in the where clause. 

My final query was looking something like this. So the span_count > x will be present in the where clause with the timestamp.

```
SELECT
    traceID,
    COUNT() AS span_count,
    max(toInt64(timestamp) + durationNano) - min(toInt64(timestamp)) AS total_duration_nano
WHERE (timestamp >= '1730162616000000000') AND (timestamp <= '1730268878000000000')
GROUP BY traceID
ORDER BY span_count DESC
LIMIT 10
```

For second point, do let me know what you think should be the best approach for the same? I tested this query on approximately 200M traces and it perfromed as expected. Let me see, if we can optimise this query somehow.

srikanthccv:
>If we need to finf the traces where 1000<spanCount<2000, that will go in the where clause.

Please share how this is achieved in the code.

> I tested this query on approximately 200M traces and it perfromed as expected

Please help me understand the performed as expected part. Have you folks discussed some pre-requisite in terms of resources needed and how you would have billions of traceIDs? From my experience, anything beyond 10 million rows in the group by questionable and beyond 25 mil is where the line should be drawn in most cases. Unless the enormous amount of memory availability, the 200M should have crashed with memory error.

srikanthccv:
I don't think grouping by traceID at the query time is an option for us. I believe @nityanandagohain has created some summary tables as a part of traces v3, which helps specifically with count; please check with him once.

eKuG:
```
SELECT
    traceID,
    COUNT() AS span_count,
    max(toInt64(timestamp) + durationNano) - min(toInt64(timestamp)) AS total_duration_nano
WHERE (timestamp >= '1730162616000000000') AND (timestamp <= '1730268878000000000')
GROUP BY traceID
HAVING COUNT() > 5
ORDER BY span_count DESC
LIMIT 10
```

@srikanthccv According to the document I have written on notion, it should have been having instead of where for Span_count. 
Having said that, I understand your concern with group by clause with larger dataset. Let me start a conversation on slack with @nityanandagohain when he can add span_count MV on trace v3

eKuG:
> I don't think grouping by traceID at the query time is an option for us. I believe @nityanandagohain has created some summary tables as a part of traces v3, which helps specifically with count; please check with him once.

PR #6409, opened 2024-11-08
chore: add k8s statefulset resources

srikanthccv:
### Summary

Part of #5373 

----

> [!IMPORTANT]
> Add support for Kubernetes StatefulSets in the query service, including new API endpoints, repository methods, and model definitions.
> 
>   - **Behavior**:
>     - Add support for Kubernetes StatefulSets in `http_handler.go` and `infra.go` with new API endpoints for attribute keys, values, and list.
>     - Introduce `StatefulSetsRepo` in `statefulsets.go` to handle data retrieval for StatefulSets.
>   - **Models**:
>     - Add `StatefulSetListRequest`, `StatefulSetListResponse`, and `StatefulSetListRecord` to `infra.go`.
>   - **Misc**:
>     - Add `getParamsForTopStatefulSets()` in `common.go` for query parameter handling.
>     - Add builder queries and query names for StatefulSets in `statefulsets.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for c3d32c6c8c4ce233aceef98fd5ec346110fcd9bc. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/pkg/query-service/app/http_handler.go
+++ b/pkg/query-service/app/http_handler.go
@@ -119,8 +119,9 @@ type APIHandler struct {
 	namespacesRepo *inframetrics.NamespacesRepo
 	clustersRepo   *inframetrics.ClustersRepo
 	// workloads
-	deploymentsRepo *inframetrics.DeploymentsRepo
-	daemonsetsRepo  *inframetrics.DaemonSetsRepo
+	deploymentsRepo  *inframetrics.DeploymentsRepo
+	daemonsetsRepo   *inframetrics.DaemonSetsRepo
+	statefulsetsRepo *inframetrics.StatefulSetsRepo
 }
 
 type APIHandlerOpts struct {
@@ -198,6 +199,7 @@ func NewAPIHandler(opts APIHandlerOpts) (*APIHandler, error) {
 	clustersRepo := inframetrics.NewClustersRepo(opts.Reader, querierv2)
 	deploymentsRepo := inframetrics.NewDeploymentsRepo(opts.Reader, querierv2)
 	daemonsetsRepo := inframetrics.NewDaemonSetsRepo(opts.Reader, querierv2)
+	statefulsetsRepo := inframetrics.NewStatefulSetsRepo(opts.Reader, querierv2)
 
 	aH := &APIHandler{
 		reader:                        opts.Reader,
@@ -224,6 +226,7 @@ func NewAPIHandler(opts APIHandlerOpts) (*APIHandler, error) {
 		clustersRepo:                  clustersRepo,
 		deploymentsRepo:               deploymentsRepo,
 		daemonsetsRepo:                daemonsetsRepo,
+		statefulsetsRepo:              statefulsetsRepo,
 	}
 
 	logsQueryBuilder := logsv3.PrepareLogsQuery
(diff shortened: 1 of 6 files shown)


pkg/query-service/app/http_handler.go
```diff
@@ -412,6 +415,11 @@ func (aH *APIHandler) RegisterInfraMetricsRoutes(router *mux.Router, am *AuthMid
 	daemonsetsSubRouter.HandleFunc("/attribute_keys", am.ViewAccess(aH.getDaemonSetAttributeKeys)).Methods(http.MethodGet)
 	daemonsetsSubRouter.HandleFunc("/attribute_values", am.ViewAccess(aH.getDaemonSetAttributeValues)).Methods(http.MethodGet)
 	daemonsetsSubRouter.HandleFunc("/list", am.ViewAccess(aH.getDaemonSetList)).Methods(http.MethodPost)
+
+	statefulsetsSubRouter := router.PathPrefix("/api/v1/statefulsets").Subrouter()
+	statefulsetsSubRouter.HandleFunc("/attribute_keys", am.ViewAccess(aH.getStatefulSetAttributeKeys)).Methods(http.MethodGet)
```
shivanshuraj1333, line 423: having "_" to concatenate strings for API is not a good pattern, use "-" instead

pkg/query-service/app/inframetrics/statefulsets.go
```diff
@@ -0,0 +1,444 @@
...
+		"cpu":            {"A"},
+		"cpu_request":    {"B", "A"},
+		"cpu_limit":      {"C", "A"},
+		"memory":         {"D"},
+		"memory_request": {"E", "D"},
+		"memory_limit":   {"F", "D"},
+		"restarts":       {"G", "A"},
+		"desired_pods":   {"H"},
+		"available_pods": {"I"},
+	}
```
shivanshuraj1333, line 42: same comment as previous PR

pkg/query-service/app/http_handler.go
srikanthccv: I have generally seen `_` commonly. I don't know why it's not a good pattern. Stripe API example https://docs.stripe.com/api/customer_sessions?lang=curl

pkg/query-service/app/inframetrics/statefulsets.go
```diff
@@ -0,0 +1,444 @@
...
+	attributeValuesResponse, err := d.reader.GetMetricAttributeValues(ctx, &req)
+	if err != nil {
+		return nil, err
+	}
+
+	return attributeValuesResponse, nil
+}
+
+func (d *StatefulSetsRepo) getMetadataAttributes(ctx context.Context, req model.StatefulSetListRequest) (map[string]map[string]string, error) {
+	statefulSetAttrs := map[string]map[string]string{}
```
shivanshuraj1333, line 141: see if composite keys can work here in the logic.
eg
data["key1:key2"] = "value"

shivanshuraj1333 (approved):

pkg/query-service/app/inframetrics/statefulsets.go
srikanthccv: I needed to use them so it is easier to main the separate map.

pkg/query-service/app/http_handler.go
shivanshuraj1333: this is from past experience, it's okay if stripe is using it, but if we can avoid it, it'd be better.
this is a nit comment so it's fine to me either ways

```
The statement that using underscores (_) for concatenating strings in APIs is not a good pattern, and that dashes (-) should be used instead, is context-dependent and can be both correct and incorrect based on various factors such as conventions, language-specific practices, or industry standards.

Here are some general guidelines:

1. REST API and URL Conventions:
In the context of REST APIs, the use of dashes (-) is generally preferred for separating words in URL paths. This is primarily for readability and SEO (Search Engine Optimization). For example:
Good: /get-user-details/
Bad: /get_user_details/
Dashes are often seen as more human-readable than underscores, especially in URLs, and are more consistent with how web addresses are written.

2. Snake Case vs. Kebab Case:
Snake case (snake_case): This is commonly used in programming languages like Python and in certain API parameter names.

Kebab case (kebab-case): This is more commonly used in URLs and some JavaScript-related APIs, as mentioned above.

Snake_case (underscore) is typically used in API parameters or JSON keys, whereas kebab-case (dash) is more common in URLs.

3. Consistency with Standards:
Some popular web frameworks or platforms, such as GraphQL or RESTful APIs, have different conventions. For example, RESTful paths generally use kebab-case (/user-profile/), while JSON payloads might use snake_case (user_profile).
If you're working within a specific platform, it's always a good idea to follow their established conventions.

4. Language and Framework Preferences:
Python, for example, tends to prefer snake_case in code, so API parameters might often use underscores (_).
JavaScript and Ruby on Rails often use kebab-case (-) for URLs and routes.
Summary:
Dashes (-) are preferred for separating words in URLs (REST API paths).
Underscores (_) are more commonly used in programming for variable names (snake_case), but are less favored in URLs.
The choice depends on the context (URLs vs. parameters vs. variable names) and the established conventions for your specific project or language ecosystem.
If you're building an API with clean and standard URL paths, dashes are usually better for readability and consistency with web standards.
```

srikanthccv: Can you share the source for this?

PR #6419, opened 2024-11-11
fix: ignore ts for panel type table

nityanandagohain:
Since we don't have ts in table panel, it should not be there in expressions.

----

> [!IMPORTANT]
> Excludes `ts` from group tags in queries for table panel types in `query_builder.go`, with corresponding test updates in `query_builder_test.go`.
> 
>   - **Behavior**:
>     - In `query_builder.go`, `expressionToQuery()` now excludes `ts` from `groupTags` if `PanelType` is `v3.PanelTypeTable`.
>   - **Tests**:
>     - Updated `query_builder_test.go` to reflect changes in expected queries for table panel types, ensuring `ts` is not included.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 24860596d365a06e069218da213e379fb515a9c0. It will automatically update as commits are pushed.</sup>


nityanandagohain:
@srikanthccv  it's because of this https://github.com/SigNoz/signoz/pull/5908#discussion_r1751801692 . We have removed ts from panelTypeTable. Because of which expressions are failing in table panel types as it's trying to select timestamp from the subqueries and ts is not present in those subqueries.

srikanthccv (approved):

Issue #5713, opened 2024-08-18
[EPIC] Trace Performance Improvements


ankitnayan:
- [ ] Group by traceID. Sample queries at https://signoz.io/docs/tutorial/writing-clickhouse-queries-in-dashboard/
- [x] How to support searching by columns like `msgSystem` or `messaging.system` together? Today users get confused whether to use either of these in filtering and groupby. Also, if we decide to use custom column names to support semantic convention of attributes like `messaging.system`, should we use alias to still make `msgSystem` work to help backward compatibility?
- [x] Flatten Resource attributes? Also, remove from map to save storage
- [x] Search by messaging, http, db and other group attributes. Flatten these or introduce grouping?
- [x] SigNoz/engineering-pod#1886
- [ ] #6160
- [x] How to make search by resource attributes faster?
- [x] Add `env` in sorting key
- [x] We only need `signoz_spans` table to view trace detail page. If searching by traceID is fast enough in the `signoz_index_v2` table we can remove the `signoz_spans` table
- [x] Will we still need `signoz_spans` table for group by `traceID` features like trace funnels, time between producer and consumer span?
- [ ] Which schema will be best for support trace analytics like tree, map etc?
- [ ] Which schema will be best for searching in parent and child spans? @eKuG https://github.com/SigNoz/signoz/issues/3085 and https://github.com/SigNoz/signoz/issues/6013

nityanandagohain:
Tasks

- [ ] https://github.com/SigNoz/signoz-otel-collector/issues/420

nityanandagohain:
Things remaining

- [ ] script for migration of OLD data.
- [ ] OSS release docs for migration
- [ ] Docs for writing custom clickhouse queries.
- [ ] Migration of queries.

nityanandagohain:
New needs
- [ ] Creation of temporary signoz_spans table with ordery by timestamp for TTL data migration
- [ ] Adding support for smartTraceAlgo in the new schema as right now that is not implemented in the new schema.

nityanandagohain:
Closing this, there is one issue attached to this but it's not really realted ro perf

PR #6440, opened 2024-11-14
feat: support for window based pagination in new trace v4

nityanandagohain:

Adds support for window based pagination for traces.

### DIFF
--- a/pkg/query-service/app/querier/querier.go
+++ b/pkg/query-service/app/querier/querier.go
@@ -52,7 +52,8 @@ type querier struct {
 	returnedSeries []*v3.Series
 	returnedErr    error
 
-	UseLogsNewSchema bool
+	UseLogsNewSchema  bool
+	UseTraceNewSchema bool
 }
 
 type QuerierOptions struct {
@@ -308,7 +309,7 @@ func (q *querier) runClickHouseQueries(ctx context.Context, params *v3.QueryRang
 	return results, errQueriesByName, err
 }
 
-func (q *querier) runLogsListQuery(ctx context.Context, params *v3.QueryRangeParamsV3, tsRanges []utils.LogsListTsRange) ([]*v3.Result, map[string]error, error) {
+func (q *querier) runWindowBasedListQuery(ctx context.Context, params *v3.QueryRangeParamsV3, tsRanges []utils.LogsListTsRange) ([]*v3.Result, map[string]error, error) {
 	res := make([]*v3.Result, 0)
 	qName := ""
 	pageSize := uint64(0)
@@ -345,15 +346,27 @@ func (q *querier) runLogsListQuery(ctx context.Context, params *v3.QueryRangePar
 
 		// append a filter to the params
 		if len(data) > 0 {
-			params.CompositeQuery.BuilderQueries[qName].Filters.Items = append(params.CompositeQuery.BuilderQueries[qName].Filters.Items, v3.FilterItem{
-				Key: v3.AttributeKey{
-					Key:      "id",
-					IsColumn: true,
-					DataType: "string",
-				},
-				Operator: v3.FilterOperatorLessThan,
-				Value:    data[len(data)-1].Data["id"],
-			})
+			if params.CompositeQuery.BuilderQueries[qName].DataSource == v3.DataSourceLogs {
+				params.CompositeQuery.BuilderQueries[qName].Filters.Items = append(params.CompositeQuery.BuilderQueries[qName].Filters.Items, v3.FilterItem{
+					Key: v3.AttributeKey{
+						Key:      "id",
+						IsColumn: true,
+						DataType: "string",
+					},
+					Operator: v3.FilterOperatorLessThan,
+					Value:    data[len(data)-1].Data["id"],
+				})
+			} else {
+				// for traces setting offset = 0 works
+				// eg -
+				// 1)--- searching 100 logs in between t1, t10, t100 with offset 0
+				// if 100 logs are there in t1 to t10 then 100 will return immediately.
+				// if 10 logs are there in t1 to t10 then we get 10, set offset to 0 and search in the next timerange of t10 to t100.
+				// 1)--- searching 100 logs in between t1, t10, t100 with offset 100
+				// It will have offset = 0 till 100 logs are found in one of the timerange tx to tx-1
+				// If it finds <100 in tx to tx-1 then it will set offset = 0 and search in the next timerange of tx-1 to tx-2
+				params.CompositeQuery.BuilderQueries[qName].Offset = 0
+			}
 		}
 
 		if uint64(len(data)) >= pageSize {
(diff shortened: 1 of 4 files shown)


pkg/query-service/app/querier/querier.go
```diff
@@ -308,7 +309,7 @@ func (q *querier) runClickHouseQueries(ctx context.Context, params *v3.QueryRang
 	return results, errQueriesByName, err
 }
 
-func (q *querier) runLogsListQuery(ctx context.Context, params *v3.QueryRangeParamsV3, tsRanges []utils.LogsListTsRange) ([]*v3.Result, map[string]error, error) {
+func (q *querier) runWindowBasedListQuery(ctx context.Context, params *v3.QueryRangeParamsV3, tsRanges []utils.LogsListTsRange) ([]*v3.Result, map[string]error, error) {
```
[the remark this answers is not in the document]
nityanandagohain: it is known

pkg/query-service/app/querier/v2/querier.go
```diff
@@ -345,15 +346,27 @@ func (q *querier) runLogsListQuery(ctx context.Context, params *v3.QueryRangePar
...
+				})
+			} else {
+				// for traces setting offset = 0 works
+				// eg -
+				// 1)--- searching 100 logs in between t1, t10, t100 with offset 0
+				// if 100 logs are there in t1 to t10 then 100 will return immediately.
+				// if 10 logs are there in t1 to t10 then we get 10, set offset to 0 and search in the next timerange of t10 to t100.
+				// 1)--- searching 100 logs in between t1, t10, t100 with offset 100
+				// It will have offset = 0 till 100 logs are found in one of the timerange tx to tx-1
+				// If it finds <100 in tx to tx-1 then it will set offset = 0 and search in the next timerange of tx-1 to tx-2
```
srikanthccv, line 367: I don't follow. It says offset = 0 works for logs and then talks about logs.

pkg/query-service/app/querier/v2/querier.go
```diff
@@ -345,15 +346,27 @@ func (q *querier) runLogsListQuery(ctx context.Context, params *v3.QueryRangePar
...
+				params.CompositeQuery.BuilderQueries[qName].Offset = 0
```
srikanthccv, line 368: Why is offset hard coded to 0? When frontend makes a request `offset=100,limit=50` what happens?

srikanthccv: > it will have to set offset=0, and limit=20

Still not going to be correct. Please take a couple of examples of the client moving pages and show how it is going to return the correct result.

nityanandagohain: My bad it should be been, offset=0, and limit=30.

I have updated the example with proper code and example.

```
// 1)offset = 0, limit = 100, tsRanges = [t1, t10], [t10, 20], [t20, t30]
//
// if 100 traces are there in [t1, t10] then 100 will return immediately.
// if 10 traces are there in [t1, t10] then we get 10, set offset to 0 and limit to 90, search in the next timerange of [t10, t20]
// if we don't find any trace in [t1, t10], then we search in [t10, t20] with offset=0, limit=100

//
// 2) offset = 50, limit = 100, tsRanges = [t1, t10], [t10, r20], [t20, t30]
//
// If we find 100 traces in [t1, t10] then we return immediately
// If we find 50 in [t1, t10] then it will set offset = 0 and limit = 50 and search in the next timerange of [t10, r20]
// if we don't find any trace in [t1, t10], then we search in [t10, r20] with offset=50, limit=100
			
```

pkg/query-service/app/querier/v2/querier.go
```diff
@@ -344,20 +347,45 @@ func (q *querier) runLogsListQuery(ctx context.Context, params *v3.QueryRangePar
...
+
+			//
+			// 2) offset = 50, limit = 100, tsRanges = [t1, t10], [t10, 20], [t20, t30]
+			//
+			// If we find 100 traces in [t1, t10] then we return immediately
+			// If we find 50 in [t1, t10] then it will set offset = 0 and limit = 50 and search in the next timerange of [t10, 20]
+			// if we don't find any trace in [t1, t10], then we search in [t10, 20] with offset=50, limit=100
+			if len(data) > 0 {
+				params.CompositeQuery.BuilderQueries[qName].Offset = 0
+				params.CompositeQuery.BuilderQueries[qName].Limit = limit - uint64(len(data))
```
srikanthccv, line 383: You are over-subtracting by using the len of `data`? It should be subtracting the data fetched in the last recent request, not the whole `data`? Example, [t1-t2]:20, [t2-t3]:10, [t3-t4]:0, [t4:t5]:0, [t5:t6]:10, with offset = 0, limit = 40. after first iteration, the limit = 20, second iteration, the limit = 20 - (30) = -10.

Please add tests (use the clickhouse mocker to return rows).

srikanthccv: disregard, I see `limit` is a global value that's not changed on each iteration. I would still encourage adding tests.

nityanandagohain: Have added testes, please have a look again @srikanthccv

srikanthccv:
Here is a case (i have been trying to come up with countercases)

offset=100, limit =60

[t1-t2] = 50
[t2-t3] = 40
[t3-t4] = 30
[t4-t5] = 10
[t5-t6] = 20

srikanthccv:
[t1, t2]: 30 traces
[t2, t3]: 40 traces
[t3, t4]: 50 traces

Offset: 50, Limit: 100

nityanandagohain:
Yeah makes sense. To solve this I have a solution where if there is an offset and the result is empty then we do a temporary query with the same filters where limit = offset, and offset=0 to get the count data that is skipped by offset. 

and then for the next run, we can say offset = offset - count of data skipped from prev query.

@srikanthccv  let me know if this sounds good

srikanthccv:
can we do something to reduce the number of roundtrips to db?

nityanandagohain:
One way I can think of is to not set offset at all, i.e we get the data and skip manually. Which should be fine unless the offset is very huge.

srikanthccv:
I am OK with the product restriction that says _For any list query, you can only consume X spans at the maximum; beyond that, you need to refine the search_. IMO, 10k spans at maximum (all pages considered) in a list query is enough for all practical purposes (except when all 10k belong to the same trace, then it's not very useful because all of them open the same trace detail). Please bring it up in #product-pod and see what people think.

Alternatively, to handle large offsets, one option could be to perform a binary search on the tsRanges buckets. This would help identify which set of buckets effectively negate the offset, allowing you to start from the appropriate bucket with the offset reset to zero.

nityanandagohain:
@srikanthccv  you mean that when paginating we go a max of offset 10k and then stop right ?

srikanthccv:
Yes, assuming everyone is fine with such limitation.

nityanandagohain:
I have added it to this PR only @srikanthccv . Sorry for making this PR a pain to review.

srikanthccv (approved):

PR #6448, opened 2024-11-15
feat: removed dashboard uuid is all cases be it duplicate, empty or somevalid, while import json

SagarRajput-7:
### Summary

Here we want to remove `uuid` for people in case when they are empty string or duplicate, but for identifying a duplicate we would need an api call which is expensive, hence we decided to rather remove `uuid` (exisiting) in all cases, and later once the importing is done, we from behind the scene will provide each dashboard with a unique `uuid`.

With this we wont block user for import json jsut on the uuid parameters

#### Screenshots

https://github.com/user-attachments/assets/bb9d3ab3-e011-4fc7-b8bf-b1a08eeeaf10

- Tested import json flows with - 
   - uuid present duplicate or valid
   - uuid empty string
   - uuid not present

----

> [!IMPORTANT]
> Remove `uuid` from `dashboardData` during JSON import in `ImportJSON` component, regardless of its value.
> 
>   - **Behavior**:
>     - In `ImportJSON` component, `uuid` is removed from `dashboardData` during JSON import, regardless of its value (duplicate, empty, or valid).
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 8778934f9a7d401ab415df5c972f1e658f7a41ff. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/src/container/ListOfDashboard/ImportJSON/index.tsx
+++ b/frontend/src/container/ListOfDashboard/ImportJSON/index.tsx
@@ -83,7 +83,7 @@ function ImportJSON({
 			const dashboardData = JSON.parse(editorValue) as DashboardData;
 
 			// Add validation for uuid
-			if (dashboardData.uuid !== undefined && dashboardData.uuid.trim() === '') {
+			if (dashboardData.uuid !== undefined) {
 				// silently remove uuid if it is empty
 				delete dashboardData.uuid;
 			}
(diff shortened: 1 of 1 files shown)


vikrantgupta25 (dismissed):
Explain above what is the validation of the uuid, the comments is non intuitive

SagarRajput-7:
> Explain above what is the validation of the uuid, the comments is non intuitive

yes was about to update them.

YounixM:
@SagarRajput-7 : I am not sure about the functionality of this PR and why this is required. Can you add details around it please.

makeavish (approved):

makeavish (approved):

YounixM (approved):

PR #6468, opened 2024-11-18
chore: use the license v2 key to fill licenses v3 on startup

vikrantgupta25:
### Summary

- make the licenses v3 call on startup to avoid any explicit migration for new structure. 

#### Screenshots

NA

### DIFF
--- a/ee/query-service/license/manager.go
+++ b/ee/query-service/license/manager.go
@@ -67,6 +67,26 @@ func StartManager(dbType string, db *sqlx.DB, useLicensesV3 bool, features ...ba
 		repo: &repo,
 	}
 
+	if useLicensesV3 {
+		// get active license from the db
+		active, err := m.repo.GetActiveLicense(context.Background())
+		if err != nil {
+			return m, err
+		}
+
+		// fetch the new license structure from control plane
+		licenseV3, apiError := validate.ValidateLicenseV3(active.Key)
+		if apiError != nil {
+			return m, apiError
+		}
+
+		// insert the licenseV3 in sqlite db
+		apiError = m.repo.InsertLicenseV3(context.Background(), licenseV3)
+		if apiError != nil {
+			return m, apiError
+		}
+	}
+
 	if err := m.start(useLicensesV3, features...); err != nil {
 		return m, err
 	}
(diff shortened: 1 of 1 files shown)


vikrantgupta25:
contributes to - https://github.com/SigNoz/engineering-pod/issues/1035

ee/query-service/license/manager.go
```diff
@@ -67,6 +67,29 @@ func StartManager(dbType string, db *sqlx.DB, useLicensesV3 bool, features ...ba
...
+			licenseV3, apiError := validate.ValidateLicenseV3(active.Key)
+			if apiError != nil {
+				return m, apiError
+			}
+
+			// insert the licenseV3 in sqlite db
+			apiError = m.repo.InsertLicenseV3(context.Background(), licenseV3)
+			if apiError != nil {
+				return m, apiError
+			}
```
grandwizard28, line 89: Check whether license has already been inserted. If yes, do not do anything!

grandwizard28 (changes requested):

ee/query-service/license/manager.go
vikrantgupta25: donee

ee/query-service/license/manager.go
```diff
@@ -67,6 +67,30 @@ func StartManager(dbType string, db *sqlx.DB, useLicensesV3 bool, features ...ba
 		repo: &repo,
 	}
 
+	if useLicensesV3 {
+		// get active license from the db
+		active, err := m.repo.GetActiveLicense(context.Background())
```
grandwizard28, line 72: Are we sure `GetActiveLicense` will work when we haven't called `LoadActiveLicense`?

grandwizard28 (changes requested):

ee/query-service/license/manager.go
vikrantgupta25: why do we need `LoadActiveLicense` to call `GetActiveLicense` ? when the query service boots up we fetch the entire details from the sqlite DB and `LoadActiveLicense` also relies on the same `GetActiveLicense` to get the active licenses from DB to set as active in license manager and proceed with validation

grandwizard28 (approved):

Issue #2767, opened 2023-05-24
Ability to query event data from traces


srikanthccv:
Users should be able to query their event data using the query range.

ankitnayan:
https://www.reddit.com/r/devops/comments/1aue09o/comment/l6fxst5/

nityanandagohain:
Do you mean searching from the trace explorer page with other filters like traceId, service.name etc ?

srikanthccv:
This is not complete yet; it still remains the actual span event data is not queryable.

tushar-signoz:
https://opentelemetry.io/blog/2026/deprecating-span-events/
OTEL is moving away from span events, hence doesn't make sense to put efforts in this direction

PR #6469, opened 2024-11-18
Trace static fields and structs for response.

nityanandagohain:
* Has new static fields of traces  (used for suggestions)
* Has deprecated static fields of traces ( use to support old names)
* Static field traces is a union of the above
* The response structs will be used in trace detail page.

ref for static fields https://github.com/SigNoz/signoz-otel-collector/blob/main/cmd/signozschemamigrator/schema_migrator/traces_migrations.go

----

> [!IMPORTANT]
> Adds new static fields for traces, deprecates old ones, and introduces new response structs for trace details.
> 
>   - **Static Fields**:
>     - Introduces `NewStaticFieldsTraces` and `DeprecatedStaticFieldsTraces` in `constants.go`.
>     - Combines both into `StaticFieldsTraces` using `init()` function.
>   - **Response Structs**:
>     - Adds `SpanItemV2` and `TraceSummary` structs in `trace.go` for trace detail responses.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for ce7b509ddbd9fb8ba39dbca434cb0c103f58a368. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/pkg/query-service/model/response.go
+++ b/pkg/query-service/model/response.go
@@ -271,12 +271,12 @@ type SearchSpanResponseItem struct {
 
 type SearchSpanResponseItemV2 struct {
 	TimeUnixNano      time.Time          `json:"timestamp" ch:"timestamp"`
-	DurationNano      uint64             `json:"durationNano" ch:"durationNano"`
+	DurationNano      uint64             `json:"durationNano" ch:"duration_nano"`
 	SpanID            string             `json:"spanId" ch:"span_id"`
 	TraceID           string             `json:"traceId" ch:"trace_id"`
 	HasError          bool               `json:"hasError" ch:"has_error"`
 	Kind              int8               `json:"kind" ch:"kind"`
-	ServiceName       string             `json:"serviceName" ch:"resource_string$$service.name"`
+	ServiceName       string             `json:"serviceName" ch:"resource_string_service$$name"`
 	Name              string             `json:"name" ch:"name"`
 	References        string             `json:"references,omitempty" ch:"references"`
 	Attributes_string map[string]string  `json:"attributes_string" ch:"attributes_string"`
--- a/pkg/query-service/model/response.go
+++ b/pkg/query-service/model/response.go
@@ -269,32 +269,6 @@ type SearchSpanResponseItem struct {
 	SpanKind         string            `json:"spanKind"`
 }
 
-type SearchSpanResponseItemV2 struct {
-	TimeUnixNano      time.Time          `json:"timestamp" ch:"timestamp"`
-	DurationNano      uint64             `json:"durationNano" ch:"duration_nano"`
-	SpanID            string             `json:"spanId" ch:"span_id"`
-	TraceID           string             `json:"traceId" ch:"trace_id"`
-	HasError          bool               `json:"hasError" ch:"has_error"`
-	Kind              int8               `json:"kind" ch:"kind"`
-	ServiceName       string             `json:"serviceName" ch:"resource_string_service$$name"`
-	Name              string             `json:"name" ch:"name"`
-	References        string             `json:"references,omitempty" ch:"references"`
-	Attributes_string map[string]string  `json:"attributes_string" ch:"attributes_string"`
-	Attributes_number map[string]float64 `json:"attributes_number" ch:"attributes_number"`
-	Attributes_bool   map[string]bool    `json:"attributes_bool" ch:"attributes_bool"`
-	Events            []string           `json:"event" ch:"events"`
-	StatusMessage     string             `json:"statusMessage" ch:"status_message"`
-	StatusCodeString  string             `json:"statusCodeString" ch:"status_code_string"`
-	SpanKind          string             `json:"spanKind" ch:"kind_string"`
-}
-
-type TraceSummary struct {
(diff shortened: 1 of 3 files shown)


eKuG:
So are we moving from camelCase to snakeCase? Just wanted to know for context. I see some changes in the previous PR as well @nityanandagohain

pkg/query-service/model/response.go
```diff
@@ -269,6 +269,32 @@ type SearchSpanResponseItem struct {
...
+	SpanID            string             `json:"spanId" ch:"span_id"`
+	TraceID           string             `json:"traceId" ch:"trace_id"`
+	HasError          bool               `json:"hasError" ch:"has_error"`
+	Kind              int8               `json:"kind" ch:"kind"`
+	ServiceName       string             `json:"serviceName" ch:"resource_string_service$$name"`
+	Name              string             `json:"name" ch:"name"`
+	References        string             `json:"references,omitempty" ch:"references"`
+	Attributes_string map[string]string  `json:"attributes_string" ch:"attributes_string"`
+	Attributes_number map[string]float64 `json:"attributes_number" ch:"attributes_number"`
+	Attributes_bool   map[string]bool    `json:"attributes_bool" ch:"attributes_bool"`
```
srikanthccv, line 284: I do not think attributes should be separated by datatype in this API response. Why not return the original data with map[string]any?

pkg/query-service/model/response.go
```diff
@@ -269,6 +269,32 @@ type SearchSpanResponseItem struct {
...
+	Events            []string           `json:"event" ch:"events"`
```
srikanthccv, line 285: Same as above. The event should be easily consumed without having to client (API or frontend to json deserialize it). It may have been string till now but it doesn't make sense to maintain status quo with new traces work.

pkg/query-service/model/response.go
nityanandagohain: Correct, I am doing that in the the function to keep the actual response same https://github.com/SigNoz/signoz/blob/4f89f7e6fc26985566c8271355a45df8f376883b/pkg/query-service/app/clickhouseReader/reader.go#L1167

pkg/query-service/model/response.go
```diff
@@ -269,6 +269,32 @@ type SearchSpanResponseItem struct {
...
+type SearchSpanResponseItemV2 struct {
+	TimeUnixNano      time.Time          `json:"timestamp" ch:"timestamp"`
+	DurationNano      uint64             `json:"durationNano" ch:"duration_nano"`
+	SpanID            string             `json:"spanId" ch:"span_id"`
+	TraceID           string             `json:"traceId" ch:"trace_id"`
+	HasError          bool               `json:"hasError" ch:"has_error"`
+	Kind              int8               `json:"kind" ch:"kind"`
+	ServiceName       string             `json:"serviceName" ch:"resource_string_service$$name"`
+	Name              string             `json:"name" ch:"name"`
+	References        string             `json:"references,omitempty" ch:"references"`
```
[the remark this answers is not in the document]
nityanandagohain: We store references a  string only right now.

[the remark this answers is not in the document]
srikanthccv: Why does how you store it in DB dictate how it's sent in response? Just run the API once and see the output yourself.

pkg/query-service/model/response.go
srikanthccv: Span event has a strictly typed structure in the protocol. What is the reason for returning a JSON string in the response?

pkg/query-service/model/response.go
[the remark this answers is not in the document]
nityanandagohain: This is how references were returned previously
```
[
    "{TraceId=000000000000000065edb2f6f0d99672, SpanId=, RefType=CHILD_OF}"
],
```

And this is how we are returning it now
```
[
    "{TraceId=000000000000000020f6fff780d13b0c, SpanId=4e1a07fd6fd54b69, RefType=CHILD_OF}"
],
```

I am not sure what you are trying to say. I wanted to avoid frontend changes so I am going with keeping the API response same.

pkg/query-service/model/response.go
nityanandagohain: To keep the API and FE working as it as. Also you mentioned that span evens are a seperate piece in itself which itself need another set of work

https://github.com/SigNoz/signoz/issues/2767

srikanthccv: The ability to query span events has no relation with how they are sent in response for trace detail

pkg/query-service/model/response.go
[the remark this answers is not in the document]
nityanandagohain: Sorry for the confusion, it's not the actual response item, I have removed the json tags and moved it to a separete file. Thanks for pointing out.

pkg/query-service/model/response.go
nityanandagohain: Sorry for the confusion, it's not the actual response item, I have removed the json tags and moved it to a separete file. Thanks for pointing out.

srikanthccv (approved):

PR #6479, opened 2024-11-19
feat: update clickhouse reader to support new trace schema

nityanandagohain:
These PR has all the clickhouse reader changes to support the new v2 table.

The next PR's will be
* integration of the flag
* TTL

----

> [!IMPORTANT]
> Update ClickHouse reader to support new trace schema v3 with schema-specific functions and configurations.
> 
>   - **Schema Support**:
>     - Add support for trace schema v3 in `options.go` and `reader.go`.
>     - Introduce constants for v3 tables: `defaultTraceIndexTableV3`, `defaultTraceLocalTableName`, `defaultTraceResourceTableV3`, `defaultTraceSummaryTable`.
>     - Add `useTraceNewSchema` flag in `ClickHouseReader` to toggle schemas.
>   - **Functionality**:
>     - Modify functions like `GetServicesList`, `GetServices`, `GetServiceOverview`, `GetTopOperations`, `GetUsage`, `SearchTraces`, `GetTotalSpans`, `GetSpansInLastHeartBeatInterval`, `GetTagsInfoInLastHeartBeatInterval` to support v3 schema.
>     - Add v3-specific functions: `SearchTracesV2`, `GetTraceAggregateAttributesV2`, `GetTraceAttributeKeysV2`, `GetTraceAttributeValuesV2`, `GetSpanAttributeKeysV2`.
>   - **Miscellaneous**:
>     - Update `NewOptions` and `NewReaderFromClickhouseConnection` for v3 schema configurations.
>     - Add comments for future integration of `useTraceNewSchema` flag.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for bc5ab603c281a66ba0b9435a3fa8afce03d79ed4. It will automatically update as commits are pushed.</sup>


pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -704,7 +742,7 @@ func (r *ClickHouseReader) GetServiceOverview(ctx context.Context, queryParams *
 			count(*) as numErrors
 		FROM %s.%s
 		WHERE serviceName = @serviceName AND name In @names AND timestamp>= @start AND timestamp<= @end AND statusCode=2`,
-		r.TraceDB, r.indexTable,
+		r.TraceDB, r.traceTableName,
```
srikanthccv, line 745: Shouldn't there be a bucket filter for `GetServiceOverview ` queries as well?

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -1054,10 +1092,137 @@ func (r *ClickHouseReader) GetUsage(ctx context.Context, queryParams *model.GetU
...
+		data := map[string]interface{}{
+			"traceSize": traceSummary.NumSpans,
+		}
+		telemetry.GetInstance().SendEvent(telemetry.TELEMETRY_EVENT_TRACE_DETAIL_API, data, userEmail, true, false)
+	}
+
+	var startTime, endTime, durationNano uint64
+	var searchScanResponses []model.SpanItemV2
+
+	query := fmt.Sprintf("SELECT timestamp, duration_nano, span_id, trace_id, has_error, kind, resource_string_service$$name, name, references, attributes_string, events, status_message, status_code_string, kind_string FROM %s.%s WHERE trace_id=$1 and ts_bucket_start>=$2 and ts_bucket_start<=$3", r.TraceDB, r.traceTableName)
```
srikanthccv, line 1140: Why are we not reading `attibuted_number`, `attribtues_bool`, and `resource_string` and sending in response?

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -1054,10 +1092,137 @@ func (r *ClickHouseReader) GetUsage(ctx context.Context, queryParams *model.GetU
...
+			return nil, err
+		}
+
+		// merge attributes_number and attributes_bool to attributes_string
+		for k, v := range item.Attributes_bool {
+			item.Attributes_string[k] = fmt.Sprintf("%v", v)
+		}
+		for k, v := range item.Attributes_number {
+			item.Attributes_string[k] = fmt.Sprintf("%v", v)
+		}
```
srikanthccv, line 1173: Related to the above question, where are we getting data for `Attributes_bool` and  `Attributes_number`? In the v2 since resource attributes were part of string tag map, they were working but in v3, we must separately read them and send in response?

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3727,7 +3900,102 @@ func (r *ClickHouseReader) CheckClickHouse(ctx context.Context) error {
...
+		key := v3.AttributeKey{
+			Key:      tagKey,
+			DataType: v3.AttributeKeyDataType(dataType),
+			Type:     v3.AttributeKeyType(tagType),
+			IsColumn: isColumn(true, statements[0].Statement, tagType, tagKey, dataType),
+		}
+
+		if _, ok := constants.DeprecatedStaticFieldsTraces[tagKey]; !ok {
+			response.AttributeKeys = append(response.AttributeKeys, key)
+		}
```
srikanthccv, line 3979: Do we want to include the deprecated fields in the suggestions?

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3797,7 +4063,69 @@ func (r *ClickHouseReader) GetTraceAggregateAttributes(ctx context.Context, req
...
+			Type:     v3.AttributeKeyType(tagType),
+			IsColumn: isColumn(true, statements[0].Statement, tagType, tagKey, dataType),
+		}
+
+		// don't send deprecated static fields
+		// this is added so that once the old tenants are moved to new schema,
+		// they old attributes are not sent to the frontend autocomplete
+		if _, ok := constants.DeprecatedStaticFieldsTraces[tagKey]; !ok {
+			response.AttributeKeys = append(response.AttributeKeys, key)
+		}
```
srikanthccv, line 4109: It's not clear. The comment says don't send but the append adds old keys. Can you help me understand what is happening here?

pkg/query-service/app/clickhouseReader/reader.go
nityanandagohain: have added the filter, but do you know from where `GetServiceOverview` is called. The services page and the detail of a signle service doesn't call this API

pkg/query-service/app/clickhouseReader/reader.go
nityanandagohain: nice catch, thanks

pkg/query-service/app/clickhouseReader/reader.go
nityanandagohain: I have updated the query above to get the value. Yeah we have to combine them as the trace detail page works that way.

pkg/query-service/app/clickhouseReader/reader.go
nityanandagohain: It's same as above i.e if the keys are not presnt in `DeprecatedStaticFieldsTraces` then only send. Which means the top level keys like `statusCode` which is present in the table for existing users won't be returned

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3838,21 +4164,95 @@ func (r *ClickHouseReader) GetTraceAttributeKeys(ctx context.Context, req *v3.Fi
...
+	if _, ok := constants.StaticFieldsTraces[req.FilterAttributeKey]; ok {
+		// query the column for the last 48 hours
+		filterValueColumnWhere := req.FilterAttributeKey
+		selectKey := req.FilterAttributeKey
+		if req.FilterAttributeKeyDataType != v3.AttributeKeyDataTypeString {
+			filterValueColumnWhere = fmt.Sprintf("toString(%s)", req.FilterAttributeKey)
+			selectKey = fmt.Sprintf("toInt64(%s)", req.FilterAttributeKey)
+		}
+
+		query = fmt.Sprintf("select distinct %s from %s.%s where ts_bucket_start >= toUInt64(toUnixTimestamp(now() - INTERVAL 48 HOUR)) AND timestamp >= toDateTime64(now() - INTERVAL 48 HOUR, 9) and %s ILIKE $1 limit $2", selectKey, r.TraceDB, r.traceTableName, filterValueColumnWhere)
```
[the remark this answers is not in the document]
nityanandagohain: The reason is that the old exporter writes the values for top level keys in the span_attributes table. In the new exporter we don't write them to `span_attributes` and `span_attributes_keys` tables.

pkg/query-service/app/clickhouseReader/reader.go
srikanthccv: I am not sure. I would suggest talk to someone who knows what APIs are used on frontend and remove the ones that are no longer used.

pkg/query-service/app/clickhouseReader/reader.go
srikanthccv: The `resources_string` is not added?

pkg/query-service/app/clickhouseReader/reader.go
srikanthccv: btw, you are reading from the `span_attributes` table here not the `span_attributes_keys`.

pkg/query-service/app/clickhouseReader/reader.go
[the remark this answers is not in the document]
srikanthccv: Can you add a todo to remove the fixed 48 hour?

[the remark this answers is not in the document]
nityanandagohain: Sure, let me add.

pkg/query-service/app/clickhouseReader/reader.go
nityanandagohain: Remoing this in another PR.

srikanthccv (approved):

Issue #6477, opened 2024-11-19
No such column attrs in table signoz_metrics.distributed_time_series_v4


Ruppsn:
## Bug description

When updating a 0.57.0 to 0.58.2 Version i get the following log with the new introduced _clickhousemetricswritev2_ exporter.

otelcol:
```
{"level":"info","ts":1732012204.8112547,"caller":"internal/retry_sender.go:118","msg":"Exporting failed. Will retry the request after interval.","kind":"exporter","data_type":"metrics","name":"clickhousemetricswritev2","error":"code: 16, message: No such column attrs in table signoz_metrics.distributed_time_series_v4 (a541c577-bed4-4a95-84f7-315bc9cfbe0f)","interval":"6.398095484s"}
{"level":"info","ts":1732012206.004339,"caller":"internal/retry_sender.go:118","msg":"Exporting failed. Will retry the request after interval.","kind":"exporter","data_type":"metrics","name":"clickhousemetricswritev2","error":"code: 16, message: No such column attrs in table signoz_metrics.distributed_time_series_v4 (a541c577-bed4-4a95-84f7-315bc9cfbe0f)","interval":"5.565095083s"}
{"level":"info","ts":1732012206.6314,"caller":"internal/retry_sender.go:118","msg":"Exporting failed. Will retry the request after interval.","kind":"exporter","data_type":"metrics","name":"clickhousemetricswritev2","error":"code: 16, message: No such column attrs in table signoz_metrics.distributed_time_series_v4 (a541c577-bed4-4a95-84f7-315bc9cfbe0f)","interval":"7.117505655s"}
{"level":"info","ts":1732012209.7228727,"caller":"internal/retry_sender.go:118","msg":"Exporting failed. Will retry the request after interval.","kind":"exporter","data_type":"metrics","name":"clickhousemetricswritev2","error":"code: 16, message: No such column attrs in table signoz_metrics.distributed_time_series_v4 (a541c577-bed4-4a95-84f7-315bc9cfbe0f)","interval":"3.23480418s"}
{"level":"info","ts":1732012211.3184712,"caller":"internal/retry_sender.go:118","msg":"Exporting failed. Will retry the request after interval.","kind":"exporter","data_type":"metrics","name":"clickhousemetricswritev2","error":"code: 16, message: No such column attrs in table signoz_metrics.distributed_time_series_v4 (a541c577-bed4-4a95-84f7-315bc9cfbe0f)","interval":"8.474944602s"}
```

[...]

PR #6489, opened 2024-11-20
Fix for running signoz-schema-migrator under docker-swarm

Ruppsn:
Fixes: https://github.com/SigNoz/signoz/issues/6477

----

> [!IMPORTANT]
> Add `sync` and `--up=` arguments to `otel-collector-migrator` in `docker-compose.yaml` for Docker Swarm compatibility.
> 
>   - **Behavior**:
>     - Add `sync` and `--up=` arguments to `otel-collector-migrator` service in `docker-compose.yaml` to ensure proper execution under Docker Swarm.
>   - **Misc**:
>     - No changes to other services or files.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 0cc22df1db668e07909b4757468114abc79aa6cf. It will automatically update as commits are pushed.</sup>


prashant-shahi:
Both `sync` and `async` schema migrators are needed.
And the `async` must run after the `sync` is successfully executed.

In Docker Compose, we achieved this using `depends_on.<service-name>.condition`:

```yaml
depends_on:
      otel-collector-migrator-sync:
        condition: service_completed_successfully
```

However, this exact behaviour will not supported in Docker Swarm.
We will need to compromise for some other solution which achieves this behavior.

Ref: https://github.com/docker/cli/issues/3880

Ruppsn:
Does this mean that meanwhile docker swarm will not be runnable with signoz?

dieterblancke:
Wouldn't it then be better to change the migrator to run both (first sync, then async) migrations instead of spawning different containers for it?

prashant-shahi:
> Does this mean that meanwhile docker swarm will not be runnable with signoz?

With your PR changes, it should work fine for now as we do not have any async migrations yet.

Though after we include async migrations, we may need to spawn async migrations post install manually for swarm.

dieterblancke:
> > Wouldn't it then be better to change the migrator to run both (first sync, then async) migrations instead of spawning different containers for it?
> 
> Technically, we could do that as well. But it's best to separate these two for the ease of running only either of them when needed.
> 
> @srikanthccv any thoughts here?

Could still do it as it works now and accept both "sync" and "async" commands, and if both are present make sure it executes in the order of "first sync, then async".

I'm not sure as to how easy/hard that would be to achieve as I haven't gotten the time to look into how the migrator service works but in my head it sounds like it should be possible.

srikanthccv (approved):
Will address async in seperate PR

PR #6496, opened 2024-11-21
fix: fixed back navigation issue for dashboard details coming from dashboard list table

SagarRajput-7:
### Summary

Here inside the `div` of whole row, we have more click events trigger like on - name, view in popover action etc. so clicking those the click event was getting bubbled up to the divs click handler and hence we had in a way 2 url, which obstructed with the back navigation flow.

Added - `stopPropogation` to stop this behaviour

#### Screenshots

**Before:**
----------
https://github.com/user-attachments/assets/abd49ad5-72b8-48a2-87d5-d32f632a7c93

**After:** 
-----------
https://github.com/user-attachments/assets/4b0488ea-0e2a-4b88-8c70-0171fcbe971a

NA

- Test all relative click event trigger point, before and after the change.
- Checked for correct url and its params and configs

----

> [!IMPORTANT]
> Fixes back navigation issue in `DashboardsList.tsx` by adding `stopPropagation()` to prevent event bubbling on dashboard link clicks.
> 
>   - **Behavior**:
>     - Fixes back navigation issue in `DashboardsList.tsx` by adding `event.stopPropagation()` in `onClickHandler` and `Link` `onClick` to prevent event bubbling.
>     - Ensures dashboard links open correctly in new tabs or same tab based on user action (metaKey/ctrlKey).
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for f3dc92f4d411068080db37ca49064f20c9cdf16f. It will automatically update as commits are pushed.</sup>


frontend/src/container/ListOfDashboard/DashboardsList.tsx
```diff
@@ -438,6 +438,7 @@ function DashboardsList(): JSX.Element {
 				const getLink = (): string => `${ROUTES.ALL_DASHBOARD}/${dashboard.id}`;
 
 				const onClickHandler = (event: React.MouseEvent<HTMLElement>): void => {
+					event.stopPropagation();
```
vikrantgupta25, line 441: why it is required here ?

SagarRajput-7: this click handler is also used directly in the action popover - for the `view` option

vikrantgupta25: got it thanks!

vikrantgupta25 (approved):

YounixM (approved):

PR #6507, opened 2024-11-22
fix: decrease the pad angle and remove the empty legend entries

vikrantgupta25:
### Summary

- the presence of `padAngle` adds a padding between adjacent values which appears as a chunk on pie chart.
- since there were `0` entries also present they were consuming the padding still. removed the 0 entries. 

#### Screenshots

NA


vikrantgupta25:
contributes to - https://github.com/SigNoz/engineering-pod/issues/2026

YounixM (approved):

PR #6548, opened 2024-11-27
feat: handle keyboard navigations for column selection in logs explor…

YounixM:
https://github.com/user-attachments/assets/903838f8-5a0d-4e51-9a4f-691507537559

----

> [!IMPORTANT]
> Adds keyboard navigation for column selection in `LogsFormatOptionsMenu` with updated styles and state management.
> 
>   - **Behavior**:
>     - Adds keyboard navigation for column selection in `LogsFormatOptionsMenu.tsx` using `ArrowUp`, `ArrowDown`, and `Enter` keys.
>     - Updates `handleKeyDown` to manage selection and scrolling of column options.
>     - Introduces `selectedValue` state to track the currently selected column.
>   - **UI Components**:
>     - Adds `add-new-column-container` and related styles in `LogsFormatOptionsMenu.styles.scss`.
>     - Adjusts styles for `.back-btn`, `.text`, and `.column-name` for better UI consistency.
>   - **Misc**:
>     - Removes unused `Divider` import from `LogsFormatOptionsMenu.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 7ed2e78b71b2aafc081bf7aabfe2c8ce3974a737. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/src/components/LogsFormatOptionsMenu/LogsFormatOptionsMenu.styles.scss
+++ b/frontend/src/components/LogsFormatOptionsMenu/LogsFormatOptionsMenu.styles.scss
@@ -3,7 +3,7 @@
 	position: absolute;
 	right: -2px;
 	margin: 6px 0;
-	width: 160px;
+	width: 240px;
 
 	border-radius: 4px;
 
@@ -24,7 +24,7 @@
 		.back-btn {
 			display: flex;
 			align-items: center;
-			gap: 6px;
+			gap: 4px;
 			padding: 12px;
 			border: none !important;
 			box-shadow: none !important;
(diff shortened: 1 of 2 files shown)


SagarRajput-7:
@YounixM - 

the logic looks good but the UI interaction changed from the current, specifically talking about the below behaviour - 

On adding a column we used to see the selected values right away in a separate section just above in the same dropdown but now its 2 different component.
With the new one, after selection, it goes to the columns list outside, so if I want to immediately deselect, I need to close the add-column dropdown and go to the column to deselect.

SagarRajput-7 (dismissed):

GeekBoySupreme:
Fair point. And I think it's okay to have that as a trade-off for this first version here.
The ideal way would have been to have a list of all columns with a checkmark for the ones already selected, clicking them again would have deselected them.

We can do that even now instead of removing the column name from the list altogether - when I click on a column name, there's a checkmark against it. The entry does not get removed from the visible list, and you can click it again to deselect.
However, I really don't want 2 separate sections for added and not-added columns here because that increases noise and confusion.

vikrantgupta25 (approved):

SagarRajput-7 (approved):

PR #6570, opened 2024-12-02
feat: add timezone support to ingestion settings list page and alerts page when groupe filter is applied

ahmadshaheer:
### Summary
- [x] Add timezone support to ingestion settings list page
- [x] Add timezone support to triggered alerts page when groupe filter is applied

#### Related Issues / PR's
Part of https://github.com/SigNoz/engineering-pod/issues/2005

##### Ingestion settings
- Before
Screenshot from 2024-12-02 14-11-37

- After
Screenshot from 2024-12-02 14-11-08

##### Triggered alerts
- Before
Screenshot from 2024-12-02 14-12-31

- After
Screenshot from 2024-12-02 14-12-41

### DIFF
--- a/frontend/src/container/IngestionSettings/MultiIngestionSettings.tsx
+++ b/frontend/src/container/IngestionSettings/MultiIngestionSettings.tsx
@@ -31,7 +31,7 @@ import { AxiosError } from 'axios';
 import { getYAxisFormattedValue } from 'components/Graph/yAxisConfig';
 import Tags from 'components/Tags/Tags';
 import { SOMETHING_WENT_WRONG } from 'constants/api';
-import dayjs, { Dayjs } from 'dayjs';
+import dayjs from 'dayjs';
 import { useGetAllIngestionsKeys } from 'hooks/IngestionKeys/useGetAllIngestionKeys';
 import useDebouncedFn from 'hooks/useDebouncedFunction';
 import { useNotifications } from 'hooks/useNotifications';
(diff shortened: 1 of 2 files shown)


frontend/src/container/IngestionSettings/MultiIngestionSettings.tsx
```diff
@@ -70,7 +71,10 @@ const { Option } = Select;
 
 const BYTES = 1073741824;
 
-export const disabledDate = (current: Dayjs): boolean =>
+// Using any type here because antd's DatePicker expects its own internal Dayjs type
+// which conflicts with our project's Dayjs type that has additional plugins (tz, utc etc).
+// eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
+export const disabledDate = (current: any): boolean =>
```
SagarRajput-7, line 77: If you have a list of types that we need then plus antds' then we can make a interface and maybe pass here, as Partial also if required.

SagarRajput-7 (approved):

frontend/src/container/IngestionSettings/MultiIngestionSettings.tsx
ahmadshaheer: @SagarRajput-7, @YounixM, The issue here is that we have configured additional plugins for Dayjs(e.g. tz, utc) that are necessary for timezone. but the ant design's Dayjs type doesn't recognize the additional properties that are introduced by these plugins. 

I tried 

```typescript
type TimezoneAgnosticDate = Omit<
	Dayjs,
	'tz' | 'offsetName' | 'utc' | 'local' | 'isUTC'
>;
```

but it didn't work. 

The last resort was to use `any`.

A similar issue has been reported here as well: https://github.com/iamkun/dayjs/issues/942

PR #6579, opened 2024-12-03
feat: add byte rate to producer API

shivanshuraj1333:
fixes: https://github.com/SigNoz/engineering-pod/issues/2056


shivanshuraj1333:
the code that you highlighted is not used in the `api/v1/messaging-queues/kafka/topic-throughput/producer` API.
This API uses trace and metrics queries which under the hood uses table type table only.

For the highlighted one, it makes sense to use list, and it's powering another API `api/v1/messaging-queues/kafka/span/evaluation`

srikanthccv (approved):

Issue #6111, opened 2024-10-03
Clicking on the graph in one panel on service details page shows spinner at the top on other panels as well.


rahulkeswani101:
https://github.com/user-attachments/assets/daefa358-63b2-45b7-a00f-844f35af4050

yashrepal0508:
@rahulkeswani101 Please confirm, If I can work on this?

PR #4675, opened 2024-03-10
Dashboard Clean up and list view improvement. 

Rajat-Dabade:
Current Flow

CurrentFlow

New Flow

NewFlow

PR #6603, opened 2024-12-07
fix: resolve unnecessary refetching of graphs on service details page

amlannandy:
### Summary

If we were clicking on 1 graph, it was causing rerenders of the other graphs as well. That was happening because of `operationPerSecWidget`, `errorPercentageWidget` and `latencyWidget` widgets. Inside their functions, a new `uuid()` call was being made every time, due to which the props being passed to the child components were changing everytime, hence causing rerenders.
Fixed that by wrapping these functions with `useMemo`.

#### Related Issues / PR's

#6111 

#### Screenshots

https://github.com/user-attachments/assets/420e4437-13ea-430e-b249-d879e1950b74

Service Details page

----

> [!IMPORTANT]
> Optimize widget rendering on service details page using `useMemo` to prevent unnecessary re-renders.
> 
>   - **Optimization**:
>     - Use `useMemo` for `operationPerSecWidget`, `errorPercentageWidget` in `Overview.tsx` to prevent unnecessary re-renders.
>     - Use `useMemo` for `latencyWidget` in `ServiceOverview.tsx` to prevent unnecessary re-renders.
>   - **Behavior**:
>     - Prevents all graphs from re-rendering when one is clicked by stabilizing widget props.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for d0d6eb98078284259de0692a6d0de58970f67f04. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/src/container/MetricsApplication/Tabs/Overview.tsx
+++ b/frontend/src/container/MetricsApplication/Tabs/Overview.tsx
@@ -145,41 +145,49 @@ function Application(): JSX.Element {
 		[servicename, topLevelOperations],
 	);
 
-	const operationPerSecWidget = getWidgetQueryBuilder({
-		query: {
-			queryType: EQueryType.QUERY_BUILDER,
-			promql: [],
-			builder: operationPerSec({
-				servicename,
-				tagFilterItems,
-				topLevelOperations: topLevelOperationsRoute,
+	const operationPerSecWidget = useMemo(
+		() =>
+			getWidgetQueryBuilder({
+				query: {
+					queryType: EQueryType.QUERY_BUILDER,
+					promql: [],
+					builder: operationPerSec({
+						servicename,
+						tagFilterItems,
+						topLevelOperations: topLevelOperationsRoute,
+					}),
+					clickhouse_sql: [],
+					id: uuid(),
+				},
+				title: GraphTitle.RATE_PER_OPS,
+				panelTypes: PANEL_TYPES.TIME_SERIES,
+				yAxisUnit: 'ops',
+				id: SERVICE_CHART_ID.rps,
 			}),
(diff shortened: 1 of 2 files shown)


srikanthccv:
@amlannandy please git blame and share which PR introduced this. I want to understand if there was any reason to do this.

amlannandy:
@srikanthccv This change was introduced in this PR - https://github.com/SigNoz/signoz/pull/4675
Earlier it was using useMemo as well, but we removed that in this PR. The PR doesn't mention why though.

I did test this flow after making the change, it seems to be working fine for me. But maybe someone else should test this change as well, as I might be missing any details to be considered while using this flow.

Overall, I think fixing this behaviour makes sense. As all the graphs in the page shouldn't be getting re-rendered just because we're interacting with one of them.

@YounixM @vikrantgupta25 As you guys reviewed the previous PR, I was wondering if you might have any context on why this change was made?

srikanthccv:
I agree it shouldn't be re-rendered, I was mainly interested in what that ID is used for and whether we are breaking anything.

amlannandy:
> I agree it shouldn't be re-rendered, I was mainly interested in what that ID is used for and whether we are breaking anything.

The id is required while building the query builder widget. I did look into its internals and not updating the ID on every render should not break anything. The query props only require a unique identifier as prop for ID which we are satisfying here and won't matter if it stays the same between re-renders.

The variables that should trigger a re-rerender are `servicename, tagFilterItems, topLevelOperationsRoute` which are already added in the dependency array of useMemo.

srikanthccv (approved):

PR #6605, opened 2024-12-09
feat: changed start and end time logic for consumer lag details

SagarRajput-7:
### Summary

Updated the startime and endtime logic for the APIs under consumer lag details

```
----- new logic ----

we have a t - that is the point on graph we clicked
so now the start and end time would be

Start: T-5
End T+5
```

Reasons for this is that we get related producer span prior to the point in time also, so here we are creating a better timeframe

- Also changed order of tab for consumer lag and partition view

#### Screenshots

https://github.com/user-attachments/assets/2cae0cb1-f0cf-42d7-8e17-aa8f2f003902

Tested all the related APIs' effects on the demo.in env

----

> [!IMPORTANT]
> Update start/end time logic and default tab for consumer lag details in messaging queues.
> 
>   - **Behavior**:
>     - Update `getStartAndEndTimesInMilliseconds()` in `MessagingQueuesUtils.ts` to calculate start and end times around a point in time, converting timestamp to milliseconds.
>     - Change default tab in `MessagingQueuesDetails` in `MQDetails.tsx` from `ConsumerDetails` to `ProducerDetails`.
>   - **API Payload**:
>     - Modify `getMetaDataAndAPIPerView()` in `MessagingQueuesUtils.ts` to use microseconds for `start` and `end` in `tableApiPayload` for `consumerLag` view.
>   - **Misc**:
>     - Reorder `MessagingQueueServiceDetailType` in `MQServiceDetailTypePerView` in `MQDetails.tsx` for `consumerLag` and `partitionLatency` views.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 72252cdf7ab871bd4dcce90cb8ea7efa416a3c37. It will automatically update as commits are pushed.</sup>


vikrantgupta25:
When clicking on any point in the graph, the graph essentially shows results for the current aggregation interval.
Example if the aggregation interval is 10 minutes then the a single graph point refers to the data in `T-10 -> T` , So ideally the logic should be based on aggregation interval and not hardcoded `T-5 -> T+5`. Please correct me if i misunderstood the intention here. 

cc @shivanshuraj1333

shivanshuraj1333:
yes this is a valid point I was trying to think in that direction but I don't have much context on aggregation intervals.

vikrantgupta25:
yes so again choosing based on points wouldn't be the correct thing to do. the aggregation interval is calculated on the query service end and even frontend can do the same calculation if required. so any dot corresponds to previous data not the future data. So the logic should be `T- aggregationInterval` to `T`. We do the same thing when moving from APM to logs or Traces page. We can use the same calculation here as well.

vikrantgupta25:
some reference links - https://signoz.io/docs/metrics-management/types-and-aggregation/#step-1-find-the-aggregation-interval

shivanshuraj1333:
but these metrics aggregation windows doesn't correlates to spans, if the aggregation window is very small we may not see any spans, in those cases hardcoded values makes more sense.

SagarRajput-7:
@shivanshuraj1333 - should I change the logic or not here, according to the above discussion?

vikrantgupta25:
> but these metrics aggregation windows doesn't correlates to spans, if the aggregation window is very small we may not see any spans, in those cases hardcoded values makes more sense.

are these metrics derived from the spans or exported directly ?

shivanshuraj1333:
exported directly ?

> Metrics are collected separately from kafka brokers directly (code ref: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/kafkametricsreceiver/consumer_scraper.go#L152), spans are from SDK/Agents from the client side. There's no native correlation between them (wrt timestamps).

vikrantgupta25:
IMO since there is no exact co-relation between them, is it correct to co-relate them on the basis of timestamp as not all the spans would be contributors to those peak metrics right ?

shivanshuraj1333:
eg
```
GROUP               TOPIC           PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG             CONSUMER-ID                                                         HOST            CLIENT-ID
high-variable-group topic1          0          0               0               0               consumer-high-variable-group-1-4e811c9f-e504-4c0b-a6c6-473e4eace95d /172.18.0.10    consumer-high-variable-group-1
high-variable-group topic1          1          0               0               0               consumer-high-variable-group-1-4e811c9f-e504-4c0b-a6c6-473e4eace95d /172.18.0.10    consumer-high-variable-group-1
high-variable-group topic1          2          14193           30155           15962           consumer-high-variable-group-1-4e811c9f-e504-4c0b-a6c6-473e4eace95d /172.18.0.10    consumer-high-variable-group-1
```

shivanshuraj1333:
@SagarRajput-7 let's get this one merged

SagarRajput-7:
> @SagarRajput-7 let's get this one merged

Sure, @shivanshuraj1333 

@vikrantgupta25 please review

vikrantgupta25 (approved):

ahmadshaheer (approved):

YounixM (approved):

PR #6626, opened 2024-12-12
feat(licenses): deprecate licenses v2

grandwizard28:
### Summary

deprecate licenses v2

### DIFF
--- a/ee/query-service/app/api/license.go
+++ b/ee/query-service/app/api/license.go
@@ -102,7 +102,7 @@ func (ah *APIHandler) applyLicense(w http.ResponseWriter, r *http.Request) {
 		RespondError(w, model.BadRequest(fmt.Errorf("license key is required")), nil)
 		return
 	}
-	license, apiError := ah.LM().Activate(r.Context(), l.Key)
+	license, apiError := ah.LM().ActivateV3(r.Context(), l.Key)
 	if apiError != nil {
 		RespondError(w, apiError, nil)
 		return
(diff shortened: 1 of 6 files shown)


ee/query-service/app/api/license.go
```diff
@@ -102,7 +102,7 @@ func (ah *APIHandler) applyLicense(w http.ResponseWriter, r *http.Request) {
 		RespondError(w, model.BadRequest(fmt.Errorf("license key is required")), nil)
 		return
 	}
-	license, apiError := ah.LM().Activate(r.Context(), l.Key)
```
vikrantgupta25, line 105: remove the `UseLicensesV3` flag from line 87 as well ?

ee/query-service/main.go
```diff
@@ -148,7 +146,6 @@ func main() {
 		GatewayUrl:        gatewayUrl,
 		UseLogsNewSchema:  useLogsNewSchema,
 		UseTraceNewSchema: useTraceNewSchema,
-		UseLicensesV3:     useLicensesV3,
```
vikrantgupta25, line 151: we should remove all occurrences of this flag

vikrantgupta25 (changes requested):
we should also remove the unused `Validate` and `Activate` methods from integrations as well

ee/query-service/app/api/license.go
grandwizard28: Yes missed that

ee/query-service/main.go
grandwizard28: Yes missed this as well.

vikrantgupta25 (approved):
LGTM

PR #6645, opened 2024-12-16
fix: unable to remove query tags from the beginning when similar tags present 

vikrantgupta25:
### Summary

- when there were duplicate filters present the `Select` component behaves inconsistently, while the state updates well but the component doesn't update. 
- added `queryTags` as the key for the `Select` component but since the component re-renders the focus state is lost used auto focus prop for the same

### DIFF
--- a/frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/QueryBuilderSearchV2.tsx
+++ b/frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/QueryBuilderSearchV2.tsx
@@ -841,10 +841,10 @@ function QueryBuilderSearchV2(
 	useEffect(() => {
 		// this is required because on change of query tags the Select component re-renders which looses focus.
 		// this was added because Select component was acting inconsistently for similar tags
-		if (selectRef.current && queryTags.length > 0) {
+		if (selectRef.current && queryTags.length > 0 && isOpen) {
 			selectRef.current?.focus();
 		}
-	}, [queryTags]);
+	}, [queryTags, isOpen]);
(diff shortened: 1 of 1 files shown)


vikrantgupta25:
contributes to - https://github.com/SigNoz/engineering-pod/issues/1994

frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/QueryBuilderSearchV2.tsx
```diff
@@ -838,6 +838,14 @@ function QueryBuilderSearchV2(
 		[tags],
 	);
 
+	useEffect(() => {
+		// this is required because on change of query tags the Select component re-renders which looses focus.
+		// this was added because Select component was acting inconsistently for similar tags
+		if (selectRef.current && queryTags.length > 0 && isOpen) {
+			selectRef.current?.focus();
+		}
+	}, [queryTags, isOpen]);
```
SagarRajput-7, line 847: using a prop on the component for focus wouldn't help here.

SagarRajput-7 (approved):

frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/QueryBuilderSearchV2.tsx
vikrantgupta25: yes used `autoFocus` prop thanks!

SagarRajput-7 (approved):

YounixM (approved):

PR #6661, opened 2024-12-18
enable kafka_span_eval, enabling the feature to avoid 404 in the messaging queue tab

shivanshuraj1333:
Enables the feature by default to avoid 404

----

> [!IMPORTANT]
> Enables `KafkaSpanEval` by default in `constants.go` to prevent 404 errors in the messaging queue tab.
> 
>   - **Behavior**:
>     - Enables `KafkaSpanEval` by default in `constants.go` to prevent 404 errors in the messaging queue tab.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for cb87806835e5bbe2301ea4df20027fe48ea96a45. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/pkg/query-service/constants/constants.go
+++ b/pkg/query-service/constants/constants.go
@@ -90,7 +90,7 @@ func EnableHostsInfraMonitoring() bool {
 	return GetOrDefaultEnv("ENABLE_INFRA_METRICS", "true") == "true"
 }
 
-var KafkaSpanEval = GetOrDefaultEnv("KAFKA_SPAN_EVAL", "false")
+var KafkaSpanEval = GetOrDefaultEnv("KAFKA_SPAN_EVAL", "true")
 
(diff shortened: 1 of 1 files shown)


shivanshuraj1333:
@prashant-shahi either let's merge this PR or have this flag enabled in all the environments used for demo.

shivanshuraj1333:
@srikanthccv shall we merge this PR?
I don't see any issues so far with the inner join with ~1 month of data in demo environments, let's see how it plays out.

srikanthccv:
No, I am strongly against this idea of joining the trace table with itself enabled everywhere by default. It's going to wreck the whole paid pool users node because it causes high CPU and OOMkill. How much data do you have in demo and what is the resource usage when the query is run in demo.

PR #6663, opened 2024-12-18
Add support for trace_v3 schema in messaging queues

shivanshuraj1333:
Supports trace_v3 now

----

> [!IMPORTANT]
> Update SQL queries to support `trace_v3` schema and enable Kafka span evaluation by default.
> 
>   - **Behavior**:
>     - Update SQL queries in `sql.go` to use `signoz_traces.distributed_signoz_index_v3` instead of `v2`.
>     - Replace `serviceName` with `resource_string_service$$name` and `msgSystem` with `attribute_string_messaging$$system` in SQL queries.
>   - **Constants**:
>     - Change `KafkaSpanEval` default value to `true` in `constants.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 7460888db2f5d4b8722681cd67afdf9766a5107d. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/pkg/query-service/constants/constants.go
+++ b/pkg/query-service/constants/constants.go
@@ -90,7 +90,7 @@ func EnableHostsInfraMonitoring() bool {
 	return GetOrDefaultEnv("ENABLE_INFRA_METRICS", "true") == "true"
 }
 
-var KafkaSpanEval = GetOrDefaultEnv("KAFKA_SPAN_EVAL", "false")
+var KafkaSpanEval = GetOrDefaultEnv("KAFKA_SPAN_EVAL", "true")
 
 func IsDurationSortFeatureEnabled() bool {
 	isDurationSortFeatureEnabledStr := DurationSortFeature
--- a/pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
+++ b/pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
@@ -9,25 +9,25 @@ func generateConsumerSQL(start, end int64, topic, partition, consumerGroup, queu
 	query := fmt.Sprintf(`
 WITH consumer_query AS (
     SELECT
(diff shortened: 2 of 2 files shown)


pkg/query-service/constants/constants.go
```diff
@@ -90,7 +90,7 @@ func EnableHostsInfraMonitoring() bool {
 	return GetOrDefaultEnv("ENABLE_INFRA_METRICS", "true") == "true"
 }
 
-var KafkaSpanEval = GetOrDefaultEnv("KAFKA_SPAN_EVAL", "false")
+var KafkaSpanEval = GetOrDefaultEnv("KAFKA_SPAN_EVAL", "true")
```
nityanandagohain, line 93: intended change ?

nityanandagohain (approved):
overall changes looks good, please test all the changes before merging

pkg/query-service/constants/constants.go
shivanshuraj1333: yes

shivanshuraj1333: yes

srikanthccv: I am re-iterating my position on this as mentioned here https://github.com/SigNoz/signoz/pull/6661#issuecomment-2550840426, this change should not be enabled by default. Please revert it.

srikanthccv:
With the trace v3, the bucket filter must exist. Otherwise, they can be really slow (it would have been better if it was suggested during review) https://github.com/SigNoz/signoz/issues/6666, @shivanshuraj1333, please sync with @nityanandagohain  if you need to understand the bucketing.

PR #6667, opened 2024-12-18
fix: use placehold in limit and use proper exists

nityanandagohain:
Fixes https://github.com/SigNoz/engineering-pod/issues/2098

* Uses placeholder intead of string formatting for limit queries
* use proper column name for exists 

----
exist explaination

previously

* key http.route  => `http.route != ''`
* key http_url => `http_url != ''`
Here since we renamed `http.route` to `attribute_string_http$$route` in v3 . There is no column with `http.route` and the query failed as it was not use the correct name for the key

Now
* key http.route  => `attribute_string_http$$route != ''`
* key http_url => `http_url != ''`
Now we use context from the static fields to decide

----

> [!IMPORTANT]
> Improve query handling by using placeholders for limit queries and correcting column names for existence checks.
> 
>   - **Behavior**:
>     - Use `strings.Replace` instead of `fmt.Sprintf` for limit queries in `helper.go` and `v2/helper.go`.
>     - Correct column name handling for `exists` and `not exists` operators in `query_builder.go`.
>   - **Functions**:
>     - Modify `existsSubQueryForFixedColumn()` in `query_builder.go` to handle string type columns correctly.
>   - **Tests**:
>     - Update `TestPrepareTracesQuery` in `query_builder_test.go` to reflect changes in limit query handling.
>     - Add test cases for `exists` and `not exists` operators in `Test_buildTracesFilterQuery` in `query_builder_test.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for a6d819e3781f636fbd457b82190c67eabcc8b575. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/pkg/query-service/app/traces/v4/query_builder.go
+++ b/pkg/query-service/app/traces/v4/query_builder.go
@@ -74,6 +74,19 @@ func getSelectLabels(groupBy []v3.AttributeKey) string {
 	return strings.Join(labels, ",")
 }
 
+// TODO(nitya): use the _exists columns as well in the future similar to logs
+func existsSubQueryForFixedColumn(key v3.AttributeKey, op v3.FilterOperator) (string, error) {
+	if key.DataType == v3.AttributeKeyDataTypeString {
+		if op == v3.FilterOperatorExists {
+			return fmt.Sprintf("%s %s ''", getColumnName(key), tracesOperatorMappingV3[v3.FilterOperatorNotEqual]), nil
+		} else {
+			return fmt.Sprintf("%s %s ''", getColumnName(key), tracesOperatorMappingV3[v3.FilterOperatorEqual]), nil
+		}
+	} else {
+		return "", fmt.Errorf("unsupported operation, exists and not exists can only be applied on custom attributes or string type columns")
+	}
+}
+
 func buildTracesFilterQuery(fs *v3.FilterSet) (string, error) {
 	var conditions []string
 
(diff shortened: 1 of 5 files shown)


srikanthccv:
I couldn't follow what change regarding the exists, can you please share more detail?

nityanandagohain:
@srikanthccv  I have updated the description

pkg/query-service/app/traces/v4/query_builder.go
```diff
@@ -74,6 +74,19 @@ func getSelectLabels(groupBy []v3.AttributeKey) string {
...
 
+// TODO(nitya): use the _exists columns as well in the future similar to logs
+func existsSubQueryForFixedColumn(key v3.AttributeKey, op v3.FilterOperator) (string, error) {
+	if key.DataType == v3.AttributeKeyDataTypeString {
+		if op == v3.FilterOperatorExists {
+			return fmt.Sprintf("%s %s ''", getColumnName(key), tracesOperatorMappingV3[v3.FilterOperatorNotEqual]), nil
+		} else {
+			return fmt.Sprintf("%s %s ''", getColumnName(key), tracesOperatorMappingV3[v3.FilterOperatorEqual]), nil
+		}
+	} else {
```
srikanthccv, line 85: Why can't the user add `http.status_code EXISTS`. I don't know why this exists, but it seems like it was already there. Please create a separate issue as to why we don't support anything other than string.

srikanthccv (approved):

srikanthccv:
please hold releasing till other messaging changes are merged.

pkg/query-service/app/traces/v4/query_builder.go
nityanandagohain: https://github.com/SigNoz/signoz/issues/6670

Issue #6484, opened 2024-11-19
Formula not working with panel in dashboard


svsk417:
## Bug description
If I run any formula to be shown in a panel within a dashboard the result yielded is "No data". 
If I run the single queries on which the formula is based on, the correct value is shown.

## Expected behavior
The sum of the two values is shown.

![A](https://github.com/user-attachments/assets/17394ad6-39fa-4c67-b4f3-cfb9e40dcbbe)
![B](https://github.com/user-attachments/assets/9a6fd2a4-3427-4009-a9da-26ca04da33d2)
![AplusB](https://github.com/user-attachments/assets/ae8e3413-8bda-4dc7-b225-d40af07042b4)

## How to reproduce

Create a panel as shown in the screenshots.

## Version information 
* **Signoz version**: 0.58.2
* **Browser version**: 
* **Your OS and version**:
* **Your CPU Architecture**(ARM/Intel):

pranay01:
@svsk417 Is this only happening for this specific query and formula? Or none of the formulae are working?

svsk417:
Hi @pranay01, this is happening to all my formulas. 
Even if I just devide the CPU utilization by the number of CPU cores I get no result.

aniketio-ctrl:
please assign this to me

PR #6720, opened 2024-12-26
fix: removed caching for all other panel type for expression except time series

aniketio-ctrl:
…om graph in v3 | 6484

### Summary

Issue: Wrong caching for metric expressions apart from panel type graph in v3
According to current implementation , we are not caching queries, we are only caching expressions, we are caching if the panel type is graph, so if we are not caching single value panel type queries, we should not also be caching the result of the expression.

for eg- 
If the expression is 'A+B', caching key would be '+' that dosen't consider the fact that wether A or B were same or not.

So while creating caching keys for expression , added a check if the version is not v4 then don't proceed unless its panelType: Graph

#### Related Issues / PR's

Fixes https://github.com/SigNoz/signoz/issues/6484

#### Screenshots

We are getting cached result with actual result which are vague

Screenshot 2024-12-26 at 1 10 25 PM

Manually tested Aread ->
1.) Api -> /api/v3/query_range, tested with different panel type and checking wether we were getting correct result or not

----

> [!IMPORTANT]
> Fix caching behavior in `GenerateKeys` of `query_builder.go` to skip non-graph panel types in non-v4 versions.
> 
>   - **Behavior**:
>     - In `GenerateKeys` of `query_builder.go`, skip caching for expressions if version is not v4 and panel type is not `Graph`.
>   - **Caching**:
>     - Ensures caching only for `Graph` panel type in non-v4 versions to prevent incorrect cache results for other panel types.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 9fb71606445d351b0bca6914d744da7541dcbcf7. It will automatically update as commits are pushed.</sup>


srikanthccv:
@aniketio-ctrl, please add test https://github.com/SigNoz/signoz/blob/6e27df9dcb23aaebd904ac33754cd4236a6516a6/pkg/query-service/app/queryBuilder/query_builder_test.go#L800

aniketio-ctrl:
> @aniketio-ctrl, please add test and sign cla
> 
> https://github.com/SigNoz/signoz/blob/6e27df9dcb23aaebd904ac33754cd4236a6516a6/pkg/query-service/app/queryBuilder/query_builder_test.go#L800

Added the testcase please check

srikanthccv (approved):

PR #6728, opened 2024-12-27
feat: base branch for trace details

vikrantgupta25:
#### Screenshots

NA


vikrantgupta25:
do not need this will be merging everything to main as these are all new components and APIs so usage can be controlled easily. base branch will just make one huge PR.

Issue #6132, opened 2024-10-08
Ability to load 10K spans. Improve frontend perf


pranay01:

vikrantgupta25:
This is done!

pranay01:
Nice, now we should create good docs and demo on this

PR #6733, opened 2024-12-30
feat: added new cache package for query service

vikrantgupta25:
### Summary

- implement new cache package for query service caching 
- the inmemory cache now stores the value as is rather than ser/deser the values 
- the cacheable entity provides interfaces to add custom `MarshalBinary` and `UnmarshalBinary` to handle binary encodings for `redis`
- updated makefile for running all pkg tests
- added new `signoz` package for setting up the dependencies 

#### Related Issues / PR's

contributes to - https://github.com/SigNoz/signoz/issues/6132 

#### Screenshots

NA

----

> [!IMPORTANT]
> Introduces a new caching package for the query service with in-memory and Redis support, updates configuration, and adds tests.
> 
>   - **Caching**:
>     - Introduces `pkg/cache` package for query service caching.
>     - Supports in-memory and Redis caching strategies in `pkg/cache/strategy/memory` and `pkg/cache/strategy/redis`.
>     - `CacheableEntity` interface for custom binary encoding.
>     - Updates `server.go` to integrate caching with `ServerOptions`.
>   - **Configuration**:
>     - Adds cache configuration to `conf/defaults.yaml` and `pkg/config/config.go`.
>     - Updates `pkg/web/config.go` to include `skip_web_frontend` option.
>   - **Testing**:
>     - Adds tests for memory and Redis caching strategies in `memory_test.go` and `redis_test.go`.
>   - **Misc**:
>     - Updates `Makefile` to run all package tests.
>     - Adds `pkg/signoz` package for dependency setup.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 05b73e0d8c4d41c23f904c8e028eb6ff91de70c3. It will automatically update as commits are pushed.</sup>


pkg/cache/cache.go
```diff
@@ -0,0 +1,59 @@
+package cache
```
grandwizard28, line 1: There is a new config package that we utilising for this. Let's use that here as well

pkg/cache/status/status.go
```diff
@@ -0,0 +1,33 @@
+package status
```
grandwizard28, line 1: This seems like an unnecessary package. We can move this to `pkg/cache/status.go`. It will be semantically better to see `cache.RetrieveStatus` also.

pkg/cache/entity/entity.go
```diff
@@ -0,0 +1,25 @@
+package entity
```
grandwizard28, line 1: This seems like an unnecessary package. We can move this to `pkg/cache/cacheable.go`. `cache.Cacheable` seems like a nice reference also.

grandwizard28 (changes requested):

pkg/cache/status/status.go
vikrantgupta25: it will lead to cyclic dependency, the cache pkg imports redis and memory package which needs status package

pkg/cache/entity/entity.go
vikrantgupta25: this will also again lead to cyclic dependency. `memory` and `redis` require the `entity` package which are then imported in the `cache` package

ee/query-service/main.go
```diff
@@ -148,12 +148,13 @@ func main() {
 		zap.L().Fatal("Failed to create config", zap.Error(err))
 	}
 
-	web, err := signozweb.New(zap.L(), config.Web)
-	if err != nil && !skipWebFrontend {
-		zap.L().Fatal("Failed to create web", zap.Error(err))
```
grandwizard28, line 153: @vikrantgupta25 we overrode the skipWebFrontend change here.

We can pass this flag in signoz.New(config, skipWebFrontend) and have an action item on me to incorporate this better for now!

grandwizard28 (changes requested):

ee/query-service/main.go
vikrantgupta25: Added it to the web config itself to keep things centralised in place and to avoid messing up the design pattern

grandwizard28 (approved):

Issue #6755, opened 2025-01-05
[Messaging-Queues-Integration] Add Overview APIs for Celery and Kafka (extensible to other queues) in Messaging Queues integration


shivanshuraj1333:
Part of epic: https://github.com/SigNoz/engineering-pod/issues/2009

Implement Overview APIs for Messaging Queues.

Sample data:
<img width="1211" alt="Image" src="https://github.com/user-attachments/assets/8dba904a-16d9-45c9-8166-149df7a70bfe" />

PR #6756, opened 2025-01-05
[feat] Add overview APIs for Celery and Kafka in messaging queues integration

shivanshuraj1333:
Related to: https://github.com/SigNoz/engineering-pod/issues/2009

Fixes: https://github.com/SigNoz/signoz/issues/6755

----

> [!IMPORTANT]
> Add new API routes and SQL query generation for Celery and Kafka messaging queues, with corresponding tests.
> 
>   - **API Routes**:
>     - Added `/api/v1/messaging-queues/queue-overview` for general queue overview.
>     - Added `/api/v1/messaging-queues/celery` routes for Celery overview, tasks, and performance.
>     - Refactored Kafka routes under `/api/v1/messaging-queues/kafka`.
>   - **Query Generation**:
>     - Added `generateOverviewSQL()` in `sql.go` for queue overview queries.
>     - Added `generateProducerConsumerEvalSQL()` in `sql.go` for producer-consumer evaluation.
>   - **Models**:
>     - Added `QueueFilters` and `CeleryTask` structs in `model.go`.
>   - **Tests**:
>     - Added tests for new SQL functions in `sql_test.go`.
>     - Added tests for new models in `model_test.go`.
>     - Added tests for query parameter building in `translator_test.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for a4cb48276966b6d04bf005444f0fafae44f90e6e. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/pkg/query-service/app/integrations/messagingQueues/kafka/model.go
+++ b/pkg/query-service/app/integrations/messagingQueues/kafka/model.go
@@ -26,11 +26,11 @@ type OnboardingResponse struct {
 // QueueFilters
 // ToDo: add capability of dynamic filtering based on any of the filters
 type QueueFilters struct {
-	serviceName string
-	spanName    string
-	queue       string
-	destination string
-	kind        string
+	ServiceName []string
+	SpanName    []string
+	Queue       []string
+	Destination []string
+	Kind        []string
 }
 
 type CeleryTask struct {
--- a/pkg/query-service/app/integrations/messagingQueues/kafka/model.go
+++ b/pkg/query-service/app/integrations/messagingQueues/kafka/model.go
@@ -22,3 +22,37 @@ type OnboardingResponse struct {
 	Message   string `json:"error_message"`
 	Status    string `json:"status"`
 }
+
+// QueueFilters
+// ToDo: add capability of dynamic filtering based on any of the filters
+type QueueFilters struct {
+	serviceName string
+	spanName    string
+	queue       string
+	destination string
+	kind        string
+}
+
+type CeleryTask struct {
+	kind   string
+	status string
+}
+
+type CeleryTasks interface {
+	GetKind() string
+	GetStatus() string
+	Set(string, string)
+}
+
+func (r *CeleryTask) GetKind() string {
+	return r.kind
+}
+
+func (r *CeleryTask) GetStatus() string {
+	return r.status
+}
+
+func (r *CeleryTask) Set(kind, status string) {
+	r.kind = kind
+	r.status = status
+}
--- a/pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
+++ b/pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
@@ -318,6 +318,97 @@ GROUP BY
 	return query
 }
 
+func generateOverviewSQL(start, end int64) string {
+	// Convert from nanoseconds to float seconds in Go
+	// to avoid decimal overflow in ClickHouse
+	startSeconds := float64(start) / 1e9
+	endSeconds := float64(end) / 1e9
+
+	// For timeRange, if you want the difference in the DB:
+	timeRangeSecs := endSeconds - startSeconds // float64 difference in Go
+
+	// If you really need tsBucketStart, do the math in Go as well
+	tsBucketStart := startSeconds - 1800 // example
+	tsBucketEnd := endSeconds
+
+	query := fmt.Sprintf(`
+WITH
+    timeRange AS (
+        SELECT %f AS seconds
+    ),
+    processed_traces AS (
+        SELECT
+            resource_string_service$$name AS service_name,
+            name AS span_name,
+            CASE
+                WHEN attribute_string_messaging$$system = 'kafka' THEN 'kafka'
+                WHEN (has(attributes_string, 'celery.action') OR has(attributes_string, 'celery.task_name')) THEN 'celery'
+                ELSE 'other'
+            END AS messaging_system,
+            kind_string AS kind,
+            COALESCE(
+                NULLIF(attributes_string['messaging.destination.name'], ''),
+                NULLIF(attributes_string['messaging.destination'], '')
+            ) AS destination,
+            durationNano,
+            status_code
+        FROM signoz_traces.distributed_signoz_index_v3
+        WHERE
+            timestamp >= toDateTime64(%f, 9)
+            AND timestamp <= toDateTime64(%f, 9)
+            -- Only if your schema has ts_bucket_start (Float64 or DateTime64) 
+            AND ts_bucket_start >= toDateTime64(%f, 9)
+            AND ts_bucket_start <= toDateTime64(%f, 9)
+            AND (
+                attribute_string_messaging$$system = 'kafka'
+                OR has(attributes_string, 'celery.action')
+                OR has(attributes_string, 'celery.task_name')
+            )
+    ),
+    aggregated_metrics AS (
+        SELECT
+            service_name,
+            span_name,
+            messaging_system,
+            destination,
+            kind,
+            count(*) AS total_count,
+            sumIf(1, status_code = 2) AS error_count,
+            quantile(0.95)(durationNano) / 1000000 AS p95_latency -- Convert to ms
+        FROM
+            processed_traces
+        GROUP BY
+            service_name,
+            span_name,
+            messaging_system,
+            destination,
+            kind
+    )
+SELECT
+    service_name,
+    span_name,
+    messaging_system,
+    destination,
+    kind,
+    COALESCE(total_count / timeRange.seconds, 0) AS throughput,
+    COALESCE((error_count * 100.0) / total_count, 0) AS error_percentage,
+    p95_latency
+FROM
+    aggregated_metrics,
+    timeRange
+WHERE
+    messaging_system IN ('kafka', 'celery')
+ORDER BY
+    service_name,
+    span_name;`,
+		timeRangeSecs, // timeRange AS (SELECT %f AS seconds)
+		startSeconds, endSeconds,
+		tsBucketStart, tsBucketEnd,
+	)
+
+	return query
+}
+
 func generateProducerSQL(start, end int64, topic, partition, queueType string) string {
 	timeRange := (end - start) / 1000000000
 	tsBucketStart := (start / 1000000000) - 1800
--- a/pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
+++ b/pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
@@ -2,6 +2,7 @@ package kafka
 
 import (
 	"fmt"
+	"strings"
 )
 
 func generateConsumerSQL(start, end int64, topic, partition, consumerGroup, queueType string) string {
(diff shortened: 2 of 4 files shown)


pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
```diff
@@ -318,6 +318,97 @@ GROUP BY
...
+    ),
+    processed_traces AS (
+        SELECT
+            resource_string_service$$name AS service_name,
+            name AS span_name,
+            CASE
+                WHEN attribute_string_messaging$$system = 'kafka' THEN 'kafka'
+                WHEN (has(attributes_string, 'celery.action') OR has(attributes_string, 'celery.task_name')) THEN 'celery'
+                ELSE 'other'
+            END AS messaging_system,
```
srikanthccv, line 347: Can this be moved to write at collector and users can just rely on `attribute_string_messaging$$system` always?

pkg/query-service/app/integrations/messagingQueues/kafka/model.go
```diff
@@ -22,3 +22,37 @@ type OnboardingResponse struct {
...
+
+// QueueFilters
+// ToDo: add capability of dynamic filtering based on any of the filters
+type QueueFilters struct {
+	serviceName string
+	spanName    string
+	queue       string
+	destination string
+	kind        string
+}
```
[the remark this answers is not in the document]
shivanshuraj1333: for two reasons, I want to use it in the SQL query in this PR
And, there's a attribute which contains json body, I want to extend this to do filtering based on the key in the json body. This going to be part of v2 of this implementation.

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
shivanshuraj1333: I am keeping it here since ideally upstream should implement this in the SDK, once it's is implemented correctly in upstream, we don't need this check, there's an issue open in upstream for that https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2913

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
```diff
@@ -318,6 +318,97 @@ GROUP BY
...
+            kind_string AS kind,
+            COALESCE(
+                NULLIF(attributes_string['messaging.destination.name'], ''),
+                NULLIF(attributes_string['messaging.destination'], '')
+            ) AS destination,
```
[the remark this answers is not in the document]
shivanshuraj1333: My understanding is that, we would want to make our collector follow the semantic conventions, not the current conventions followed by agents since that is not stable, and all agents would implement the semantic conventions.

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
```diff
@@ -318,6 +318,97 @@ GROUP BY
...
+    service_name,
+    span_name,
+    messaging_system,
+    destination,
+    kind,
+    COALESCE(total_count / timeRange.seconds, 0) AS throughput,
+    COALESCE((error_count * 100.0) / total_count, 0) AS error_percentage,
+    p95_latency
+FROM
+    aggregated_metrics,
```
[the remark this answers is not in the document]
shivanshuraj1333: I have added a comparison here for CTE and subqueries, here https://www.notion.so/signoz/Eng-details-023f3c25cd974fa8b1b782c299169663?pvs=4#173fcc6bcd198099a84aea267ef70f52

shivanshuraj1333:
@srikanthccv I want to open up some more PRs, but those are making some slight changes, it would be a good idea to either get it merged, or I add code changes to this one (and make a big PR).

let me know WDYT

pkg/query-service/app/integrations/messagingQueues/kafka/model.go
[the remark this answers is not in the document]
srikanthccv: You can do all of the with `Filters`. If that's not the case, please be more specific and share an example.

srikanthccv (commented):
I think it's a good idea to keep the base changes setup and their reasoning in this PR. In the next PRs, then I don't have to think about them again

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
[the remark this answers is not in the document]
ankitnayan: `messaging.destination` seems a frequently used for any `messaging.system` related queries. I recommend materializing this to a column

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
```diff
@@ -318,6 +319,141 @@ GROUP BY
...
+    timeRange AS (
+        SELECT %f AS seconds
+    ),
+    processed_traces AS (
+        SELECT
+            resource_string_service$$name AS service_name,
+            name AS span_name,
+            CASE
+                WHEN attribute_string_messaging$$system = 'kafka' THEN 'kafka'
+                WHEN (has(attributes_string, 'celery.action') OR has(attributes_string, 'celery.task_name')) THEN 'celery'
```
ankitnayan, line 377: Using `celery` as value for messaging.system can be incorrect? Should we use `redis/rabbitmq` instead? Also, we should not auto-populate messaging.system on our own else this will not correlate to traces if `messaging.system` is filled at query time.

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
```diff
@@ -318,6 +319,141 @@ GROUP BY
...
+	query := fmt.Sprintf(`
+WITH
+    timeRange AS (
+        SELECT %f AS seconds
+    ),
+    processed_traces AS (
+        SELECT
+            resource_string_service$$name AS service_name,
+            name AS span_name,
+            CASE
```
ankitnayan, line 375: Why is there a need for this case? The feature should work for any `messaging.system` and not just kafka and celery

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
shivanshuraj1333: > Using celery as value for messaging.system can be incorrect? Should we use redis/rabbitmq instead?

Until upstream supports backend information (which is not straighforward) https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2913, this should be okay, there's no direct way to add backend information, since today, all the attributes are extracted from context https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/utils.py#L35-L54 which doesn't have a way to extract backend information. Some other type of instrumentation is needed to populate backend info.

Any suggestions for getting the information from celery and add that here?

If we don't add celery here, it would cause us to leave it blank, which is not a good UX.
Also, the broker in context of celery is like a middleman which stores the intermediary results, the actual processing is done by celery workers, so showing celery here makes sense.

> Also, we should not auto-populate messaging.system on our own else this will not correlate to traces if messaging.system is filled at query time.

If you're taking about clicking the row here, and then showing the traces, in case of celery, today we'd not propagate the messaging.system attribute (since it is blank).

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
shivanshuraj1333: This would work for other `messaging.system`, yes if the semantic conventions are followed.

But, I'm not sure what all instrumentations are supported for queues, this check here ensures that it works for celery and kafka without exception.

We can extend this by having a default case for any messaging queue, but probably not now, unless we have checked upstream instrumentations for different queues.

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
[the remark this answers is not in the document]
shivanshuraj1333: Makes sense, have created this, and can be picked up after initial release? https://github.com/SigNoz/signoz/issues/6764

Or should we do it first?

Currently semantically, `messaging.destination.name` is being used for kafka, but for celery it is `messaging.destination`

So, I suggest we first make the change in upstream to make it `messaging.destination.name` in celery as well and also create a materialized column for `messaging.destination.name`.

And then update the queries, makes sense?

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
shivanshuraj1333: this is how I am doing the filtering, https://github.com/SigNoz/signoz/pull/6756/files#diff-480b722b377e24e0d7eff30cee9640595400fd21ffbc67283b67ff7e78385815R324

pkg/query-service/app/integrations/messagingQueues/kafka/model.go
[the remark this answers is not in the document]
shivanshuraj1333: This is how I am doing the filtering https://github.com/SigNoz/signoz/pull/6756/files#diff-480b722b377e24e0d7eff30cee9640595400fd21ffbc67283b67ff7e78385815R324

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
ankitnayan: The PRD was designed considering all messaging queues. Please remove this restriction. There are many other messaging systems which will start showing up automatically in the UI and will help users

ankitnayan: the table only needs `messaging.system` and `messaging.destination` which are there for many other systems

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
shivanshuraj1333: makes sense, updated it with
`WHEN attribute_string_messaging$$system != '' THEN attribute_string_messaging$$system` for celery let's keep it same. We can show celery, the broker is intermediary and users would be using mostly a same broker.

When broker is supported (either we add it in upstream, or it gets added), we will start supporting that. (tracked in the above issue).

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
shivanshuraj1333: makes sense, updated it with
`WHEN attribute_string_messaging$$system != '' THEN attribute_string_messaging$$system` for celery let's keep it same.

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
ankitnayan: ok

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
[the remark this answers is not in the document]
ankitnayan: yeah..please create issues to look into after the first release

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
```diff
@@ -318,6 +319,141 @@ GROUP BY
...
+    aggregated_metrics.span_name,
+    aggregated_metrics.messaging_system,
+    aggregated_metrics.destination,
+    aggregated_metrics.kind_string,
+    COALESCE(aggregated_metrics.total_count / timeRange.seconds, 0) AS throughput,
+    COALESCE((aggregated_metrics.error_count * 100.0) / aggregated_metrics.total_count, 0) AS error_percentage,
+    aggregated_metrics.p95_latency
+FROM
+    aggregated_metrics
+CROSS JOIN timeRange
```
srikanthccv, line 429: Is there really a reason for this `timeRange` join and another layer? Can the `aggregated_metrics` query return all required result.

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
```diff
@@ -318,6 +319,141 @@ GROUP BY
...
+func inClause(fieldName string, values []string) string {
+	// Quote and escape each value for safety
+	var quoted []string
+	for _, v := range values {
+		// Simple escape: replace any single quotes in v
+		safeVal := strings.ReplaceAll(v, "'", "''")
+		quoted = append(quoted, fmt.Sprintf("'%s'", safeVal))
+	}
+	return fmt.Sprintf("%s IN (%s)", fieldName, strings.Join(quoted, ","))
+}
```
srikanthccv, line 455: Another reason why filters from v3 should be used, then `buildTracesFilterQuery` https://github.com/SigNoz/signoz/blob/ecd50f72327db889b7665ec0701bf5ed727d7a5f/pkg/query-service/app/traces/v4/query_builder.go#L90 can help prepare the query.

pkg/query-service/app/integrations/messagingQueues/kafka/translator.go
```diff
@@ -304,22 +321,60 @@ func BuildQRParamsWithCache(messagingQueue *MessagingQueue, queryContext string,
...
+	parts := strings.Split(val, ",")
+	var out []string
+	for _, p := range parts {
+		trimmed := strings.TrimSpace(p)
+		if trimmed != "" {
+			out = append(out, trimmed)
+		}
+	}
+	return out
+}
```
srikanthccv, line 350: Same, let's avoid feature specific parsing, and follow the query payload.

pkg/query-service/app/integrations/messagingQueues/kafka/model.go
```diff
@@ -22,3 +22,37 @@ type OnboardingResponse struct {
...
+
+// QueueFilters
+// ToDo: add capability of dynamic filtering based on any of the filters
+type QueueFilters struct {
+	ServiceName []string
+	SpanName    []string
+	Queue       []string
+	Destination []string
+	Kind        []string
+}
```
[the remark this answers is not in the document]
srikanthccv: And adding additional filters will not require any work from API because client can add them as they see fit. For instance, a user says they need a filter on `deployment.environment`, with this mechanism client has to wait for api to be ready to make the change.

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
srikanthccv: Maybe create a PR against this branch with the changes? Since this is the first PR for this feature, I suggest we get the filter changes in the same PR.

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
shivanshuraj1333: marking this as resolved

pkg/query-service/app/integrations/messagingQueues/kafka/sql.go
```diff
@@ -318,6 +319,141 @@ GROUP BY
...
+	}
+	if len(filters.Destination) > 0 {
+		whereClauses = append(whereClauses, inClause("aggregated_metrics.destination", filters.Destination))
+	}
+	if len(filters.Kind) > 0 {
+		whereClauses = append(whereClauses, inClause("aggregated_metrics.kind_string", filters.Kind))
+	}
+
+	// Combine all WHERE clauses with AND
+	whereSQL := strings.Join(whereClauses, "\n    AND ")
```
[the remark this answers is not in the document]
shivanshuraj1333: done

shivanshuraj1333:
@srikanthccv addressed the other changes, apart from filter is there anything else?
Would want to keep this PR going, filtering refactoring I need to do across all the branches (and open PRs).
So, it's better I do it in next iteration, and make the code APIs ready for UI.

The filtering refactoring will be done via this post first merge https://github.com/SigNoz/signoz/issues/6768

srikanthccv (approved):
All the custom parsing and query building means there could be number of issue with how they are done. I am not reviewing them since you want to address them later

PR #6770, opened 2025-01-08
fix(portal): use window.location.origin instead of the window.location.href

vikrantgupta25:
### Summary

- the portal session fails when the `return_url` is too large, instead use the origin only

#### Related Issues / PR's

contributes to - https://github.com/SigNoz/engineering-pod/issues/2129

### DIFF
--- a/frontend/src/container/AppLayout/index.tsx
+++ b/frontend/src/container/AppLayout/index.tsx
@@ -272,7 +272,7 @@ function AppLayout(props: AppLayoutProps): JSX.Element {
 	const handleFailedPayment = (): void => {
 		manageCreditCard({
 			licenseKey: activeLicenseV3?.key || '',
-			successURL: window.location.href,
+			successURL: window.location.origin,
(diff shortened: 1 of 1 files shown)


frontend/src/container/AppLayout/index.tsx
```diff
@@ -272,7 +272,7 @@ function AppLayout(props: AppLayoutProps): JSX.Element {
 	const handleFailedPayment = (): void => {
 		manageCreditCard({
 			licenseKey: activeLicenseV3?.key || '',
-			successURL: window.location.href,
+			successURL: window.location.origin,
 			cancelURL: window.location.href,
```
YounixM, line 276: Do we want to update the cancelURL too ?

vikrantgupta25: no we aren't using that in the API so didn't update the same

YounixM (approved):

Issue #6197, opened 2024-10-15
[authentication] trusted header authentication


dakriy:
## Is your feature request related to a problem?

This is related to #1188, but the solution provided in the enterprise plan is much more enterprise oriented and does not meet some of the needs that I have or that others voiced in that ticket. It would be nice to support simple trusted header auth (like what [Authelia](https://www.authelia.com/integration/trusted-header-sso/introduction/) provides) without having to go to the SigNoz login page. I implemented all of the code already and don't want to maintain my own separate fork and thought other self hosters that don't have the extra cash to spare on an enterprise plan like myself could benefit. This functionality is not currently in any enterprise plan and it would serves a hobbyist need rather than the more organizational focused needs provided in the enterprise plan.

## Describe the solution you'd like

Allow for simple trusted header login without showing the login page configured with environment variables (no GUI needed). This solution is already mostly implemented and working on my local dev machine.

## Describe alternatives you've considered

- Paying for an enterprise plan (I'm poor and doesn't meet my needs).
- Maintaining my own fork :(
- Opening up a self-hosted hobbyist SigNoz instance to the world wide web. I don't like having many different username/passwords on many different logins across my services. Also I like to protect all my services behind a more security focused trustworthy gateway as I have seen multiple 0-days occur with services I have run where having them under one single roof has prevented my home lab from getting hacked. Additionally, it is easier for people who don't keep up with movements in the security world and limited time to be pedantic about keeping one important auth service up to date rather than everything.
- Having 2 login pages. Really annoying

Thank you guys for this project! I am loving it so far!

dblundell:
@dakriy are you happy to share the pull request for this feature? I also would greatly prefer to have a dedicated security focused auth gateway in front of all services.  I'm happy to help maintain it on a public repo if it doesn't get merged.

mgilham:
@dblundell @dakriy I opened https://github.com/SigNoz/signoz/pull/6825 as a starting point.

scolastico:
In the mean time, here is my workaround: https://github.com/scolastico-dev/s.Containers/blob/main/src/signoz-auth-proxy/README.md

dakriy:
@dblundell sorry I didn't see your reply until now. I don't check my GitHub notifications often...

Here is the diff for what I did, it may help @mgilham  if he wants to integrate any of it into his PR

[...]

PR #6825, opened 2025-01-15
feat: header auth

mgilham:
### Summary

This is a first cut of supporting "trusted header authentication", as referenced in [this issue](https://github.com/SigNoz/signoz/issues/6197), and also inspired by Kubernetes' support for [authenticating proxies](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy) and [impersonation](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation). It is working in my cluster, but I expect will need more polish before merge. I wanted to open the PR to provide a concrete basis for discussion and alignment on the approach and design.

The feature is an alternative means of authenticating users, in addition to the existing email+password login and the ee-only SAML login. In this scenario:
- The SigNoz API is deployed behind a reverse proxy.
- The reverse proxy authenticates the user via some outside mechanism, e.g.
  - external JWT
  - mutual TLS
  - Kubernetes token
  - HTTP Basic
  - etc.
- The reverse proxy forwards the authenticated user identity to the SigNoz API via HTTP headers.
- The SigNoz API is secured such that it only responds to requests from the reverse proxy (or at least, only trusts the custom HTTP headers iff they only come from the reverse proxy).

The implementation in this PR expects that the query service is locked down at the network level to only permit access through the reverse proxy. This can be achieved via running the proxy as a container in the same pod and only listening on localhost, via network policy, or other means. Future enhancements could include:
- having the query service authenticate the reverse proxy via mutual TLS or JWT (like the Kubernetes service account token)
- an optional allowlist of IPs that are permitted to call the query service

#### Related Issues / PR's

- https://github.com/SigNoz/signoz/issues/6197

#### Screenshots

NA

The feature is behind a feature flag command-line arg that preserves the prior logic when it is not enabled. When it is enabled, the front-end will call the login API with only the email specified, and the query service will use the header injected by the proxy to authenticate the user, instead of password.

This implementation will create the specified user and org in the sqllite DB, if they do not already exist. This may merit discussion to ensure it aligns with the long-term intentions for SigNoz multi-tenancy. It doesn't appear that the current persisted user and org data are used for much, but I'm sure there are plans to build on them. Happy to discuss this on Slack or elsewhere.

Manually tested login, logout, email+password login when auth headers are absent.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -31,3 +31,6 @@ indent_size = 2
 [{package.json,.travis.yml}]
 indent_style = space
 indent_size = 2
+
+[*.{ts,tsx}]
+indent_style = tab
--- a/ee/query-service/app/api/auth.go
+++ b/ee/query-service/app/api/auth.go
@@ -50,7 +50,7 @@ func (ah *APIHandler) loginUser(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// if all looks good, call auth
-	resp, err := baseauth.Login(ctx, &req)
+	resp, err := baseauth.Login(ctx, &req, nil)
 	if ah.HandleError(w, err, http.StatusUnauthorized) {
 		return
 	}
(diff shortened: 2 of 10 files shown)


therealpandey:
Hi @mgilham,
Thank you for this PR. The codebase is finally in a shape that a trustedheaderauth Identification (IdentN) provider can be implemented cleanly. Would you be willing to contribute?

PR #6830, opened 2025-01-16
fix: logs window based pagination to pageSize offset instead of using…

### LINKED ISSUE #6829: [LOGS][BE]: API changes for pagination i.e remove id from filters.
[LOGS][BE]: API changes for pagination i.e remove id from filters and use limit offset

https://github.com/SigNoz/signoz/issues/6365

nityanandagohain:
Fixes #6829 

changes:- 

* update window based pagination logic to support limit, pageSize, offset instead of id based filtering for logs.
* the logic is now similar to traces where offset is done locally.
* the tests are same as traces, but what changed is used of pageSize instead of limit as limit is the max amount of data to be paginated using offset and pageSize for logs.
* There is no max value of offset right now.
* changes in querier/querier v2 is same .

@srikanthccv  requesting your review here as you have context on this part.

### DIFF
--- a/pkg/query-service/app/logs/v4/query_builder.go
+++ b/pkg/query-service/app/logs/v4/query_builder.go
@@ -534,6 +534,7 @@ func PrepareLogsQuery(start, end int64, queryType v3.QueryType, panelType v3.Pan
 			return "", fmt.Errorf("max limit exceeded")
 		}
 
+		// when pageSize is provided, we need to fetch the logs in chunks
 		if mq.PageSize > 0 {
 			if mq.Limit > 0 && mq.Offset+mq.PageSize > mq.Limit {
 				query = logsV3.AddLimitToQuery(query, mq.Limit-mq.Offset)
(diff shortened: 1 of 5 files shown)


srikanthccv:
Please share the plan. I see this is still not merged https://github.com/SigNoz/signoz-otel-collector/pull/489

nityanandagohain:
okay so here is the release plan.
* We will release the frontend and query service changes first. This is because this logic will work with existing id as well.
* Once done then we will release the collector changes in the next release.

nityanandagohain:
1. Why do we use different pagination mechanisms for traces and logs?
- So traces was built just with limit, offset. In logs we use limit, offset and pageSize. So for pagination limit, pageSize is used and limit is used to define the max limit pagination can be done. The traces UI was already using the limit, offset. But logs we decided to use limit, PageSize since we were using id pagination at that time.
2. Yeah correct, I will update it to have id as well for logs.

srikanthccv:
```
				v.OrderBy[0].ColumnName == "timestamp" &&
				v.OrderBy[0].Order == "desc" &&
				v.OrderBy[1].ColumnName == "id" &&
				v.OrderBy[1].Order == "desc" {

```

Is there a reason why only desc uses the window search now? Why can't the same be done for asc?

nityanandagohain:
So in this PR, wanted to focus on fixing the pagination just for the default page. So kept support for pagination asc aside.

srikanthccv:
Ok, if there is no technical limitation to make the asc follow the same window-based search, it seemed like a small change so I brought it up.

nityanandagohain:
Sure, let me make the change. Make sense to do it now as might be left out.

nityanandagohain:
@srikanthccv I have updated the code to support order by asc both for logs and traces. added tests as well, please have a look

srikanthccv:
LGTM, let me know if you need approval now or later when it is verified with frontend integration.

srikanthccv (approved):

PR #6901, opened 2025-01-22
fix: move analytics middleware

nityanandagohain:
Part of https://github.com/SigNoz/platform-pod/issues/406

this just moves the analystics middleware to common package , no change in logic. We will look into auth package that we are importing in this middleware later

----

> [!IMPORTANT]
> Moved analytics middleware to a common package without changing its logic.
> 
>   - **Middleware Move**:
>     - Moved analytics middleware logic from `ee/query-service/app/server.go` and `pkg/query-service/app/server.go` to `pkg/http/middleware/analytics.go`.
>     - Replaced `s.analyticsMiddleware` with `middleware.NewAnalytics(zap.L()).Wrap` in `createPrivateServer` and `createPublicServer` functions in `server.go` files.
>   - **No Logic Change**:
>     - The logic of the analytics middleware remains unchanged.
>   - **Future Work**:
>     - The `auth` package import in the middleware will be reviewed later.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for dc8cd055459b44f6e030be8de760ff845318dbd5. It will automatically update as commits are pushed.</sup>


pkg/http/middleware/analytics.go
```diff
@@ -0,0 +1,161 @@
...
+	// TODO(remove): Remove auth packages
+	"go.signoz.io/signoz/pkg/query-service/auth"
+
+	"github.com/gorilla/mux"
+	v3 "go.signoz.io/signoz/pkg/query-service/model/v3"
+	"go.signoz.io/signoz/pkg/query-service/telemetry"
+	"go.uber.org/zap"
+)
+
+type AnalyticsMiddleware struct {
```
grandwizard28, line 19: We dont want to expose this outside the package
```suggestion
type analytics struct {
```

pkg/http/middleware/analytics.go
```diff
@@ -0,0 +1,161 @@
...
+	logger *zap.Logger
+}
+
+func NewAnalyticsMiddleware(logger *zap.Logger) *AnalyticsMiddleware {
```
grandwizard28, line 23:
```suggestion
func NewAnalytics(logger *zap.Logger) *analytics {
```
Omit the middleware so that its more in line with other implementations in the package.

grandwizard28 (changes requested):

pkg/http/middleware/analytics.go
nityanandagohain: thanks, have made the changes

grandwizard28 (approved):

PR #6917, opened 2025-01-24
fix: logging middleware add excluded routes

nityanandagohain:
Part of https://github.com/SigNoz/platform-pod/issues/406

Logging middleware not takes an array of routes to exclude, and apiserver configs excepts a logging config.

----

> [!IMPORTANT]
> Add support for excluding specific routes from logging in middleware by updating configuration and server setup.
> 
>   - **Behavior**:
>     - `NewLogging` in `logging.go` now accepts `excludedRoutes` to skip logging for specified paths.
>     - Updated `createPrivateServer` and `createPublicServer` in `server.go` to use the new logging configuration.
>   - **Configuration**:
>     - Added `Logging` struct in `config.go` with `ExcludedRoutes`.
>     - Updated `example.yaml` to include `logging.excluded_routes`.
>   - **Tests**:
>     - Updated `TestNewWithEnvProvider` in `config_test.go` to test `SIGNOZ_APISERVER_LOGGING_EXCLUDED__ROUTES`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 4aa3ffc11aced89c93edcb8ab741b3e683d9bbfb. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/ee/query-service/app/server.go
+++ b/ee/query-service/app/server.go
@@ -330,7 +330,7 @@ func (s *Server) createPrivateServer(apiHandler *api.APIHandler) (*http.Server, 
 		s.serverOptions.Config.APIServer.Timeout.Max,
 	).Wrap)
 	r.Use(middleware.NewAnalytics(zap.L()).Wrap)
-	r.Use(middleware.NewLogging(zap.L(), s.serverOptions.Config.APIServer.Logging.ExcludedApiRoutes).Wrap)
+	r.Use(middleware.NewLogging(zap.L(), s.serverOptions.Config.APIServer.Logging.ExcludedRoutes).Wrap)
 
 	apiHandler.RegisterPrivateRoutes(r)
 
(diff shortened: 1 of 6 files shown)


pkg/apiserver/config.go
```diff
@@ -20,6 +21,11 @@ type Timeout struct {
 	ExcludedRoutes []string `mapstructure:"excluded_routes"`
 }
 
+type Logging struct {
+	// The list of routes that are excluded from the logging
+	ExcludedApiRoutes []string `mapstructure:"excluded_api_routes"`
```
grandwizard28, line 26:
```suggestion
	ExcludedRoutes []string `mapstructure:"excluded_routes"`
```

We should be as consistent as possible. We used ExcludedRoutes for Timeout, let's use the same here.

ee/query-service/app/server.go
```diff
@@ -330,7 +330,7 @@ func (s *Server) createPrivateServer(apiHandler *api.APIHandler) (*http.Server,
 		s.serverOptions.Config.APIServer.Timeout.Max,
```
grandwizard28, line 330: Also, a reminder to update the `conf/example.yaml` with this updated configuration for apiserver.

grandwizard28 (changes requested):

grandwizard28 (approved):

PR #6923, opened 2025-01-26
fix: telemetry store

nityanandagohain:
Fixes: https://github.com/SigNoz/platform-pod/issues/401

adds telemetry store package and use is as a part of signoz.

Major changes 
* The reason, telemtrystore provider is handing depricated ClickHouseUrl is because we need to parse it using clickhouse function.
* Add telemetry store package which initalizes ch connection
* the ch connection is the wrapper which adds default query settings as well.
* removed old clickhouse connection configuration flags.
* removed OptimizeReadInOrderRegex and OptimizeReadInOrderRegexCompiled from wrapper. as they are not used and were added during logs projections  

----

> [!IMPORTANT]
> Introduces a new telemetry store package for ClickHouse connection management, replacing old configurations and handling deprecated flags for backward compatibility.
> 
>   - **Behavior**:
>     - Introduces `telemetrystore` package for ClickHouse connection management.
>     - Removes old ClickHouse connection flags and configurations.
>     - Handles deprecated `ClickHouseUrl` and connection flags for backward compatibility.
>   - **Configuration**:
>     - Adds `telemetrystore` configuration to `config.go`.
>     - Updates `example.yaml` with `telemetrystore` settings.
>   - **Server Initialization**:
>     - Updates `NewServer` in `server.go` to use `telemetrystore` for ClickHouse connections.
>   - **Testing**:
>     - Updates test cases in `querier_test.go`, `v2/querier_test.go`, and `threshold_rule_test.go` to use new telemetry store.
>     - Adds mock telemetry store provider in `telemetrystoretest` package.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 2e713b47c73f2dc2320c44d531db6b7bfff6a6cc. It will automatically update as commits are pushed.</sup>


pkg/telemetrystore/clickhousetelemetrystore/wrapper.go
```diff
@@ -0,0 +1,116 @@
...
+	"context"
+	"encoding/json"
+
+	"github.com/ClickHouse/clickhouse-go/v2"
+	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
+	"go.signoz.io/signoz/pkg/query-service/common"
+	"go.signoz.io/signoz/pkg/telemetrystore"
+)
+
+type clickHouseWrapper struct {
```
grandwizard28, line 13: This seems to be satisfying some interface. Can we mention the interface here as well?

Also lets have a pointer implement the interface and create pointer receivers for this.

grandwizard28 (changes requested):

pkg/signoz/config.go
```diff
@@ -35,16 +36,28 @@ type Config struct {
...
 	// API Server config
 	APIServer apiserver.Config `mapstructure:"apiserver"`
+
+	// TelemetryStore config
+	TelemetryStore telemetrystore.Config `mapstructure:"telemetrystore"`
+}
+
+// DepricatedFlags are the flags that are deprecated and scheduled for removal.
+// These flags are used to ensure backward compatibility with the old flags.
+type DepricatedFlags struct {
```
grandwizard28, line 46:
```suggestion
type DeprecatedFlags struct {
```

pkg/signoz/config.go
```diff
@@ -35,16 +36,28 @@ type Config struct {
...
+	MaxIdleConns int
+	MaxOpenConns int
+	DialTimeout  time.Duration
 }
 
-func NewConfig(ctx context.Context, resolverConfig config.ResolverConfig) (Config, error) {
+func NewConfig(ctx context.Context, resolverConfig config.ResolverConfig, depricatedFlags DepricatedFlags) (Config, error) {
```
grandwizard28, line 52:
```suggestion
func NewConfig(ctx context.Context, resolverConfig config.ResolverConfig, deprecatedFlags DeprecatedFlags) (Config, error) {
```

pkg/telemetrystore/telemetrystorehook/clickhousesettings.go
```diff
@@ -0,0 +1,85 @@
+package telemetrystorehook
+
```
grandwizard28, line 2: just `settings.go` as the same of the file.

pkg/telemetrystore/telemetrystore.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"context"
+
+	"github.com/ClickHouse/clickhouse-go/v2"
+	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
+)
+
+type TelemetryStore interface {
+	// Returns the SigNoz Wrapper for Clickhouse
+	ClickHouse() clickhouse.Conn
```
grandwizard28, line 12:
```suggestion
	ClickHouseDB() clickhouse.Conn
```
To keep it consistent with sqlstore.SQLStore {SQLDB} interface

grandwizard28 (changes requested):

pkg/telemetrystore/telemetrystore.go
grandwizard28: Or if you feel we can keep something more idiomatic. Any suggestions?

nityanandagohain: Don't have an issue. Let's keep it ClicHouseDB if required we can change it later.

grandwizard28 (approved):

PR #6956, opened 2025-01-29
fix: apply x axis date/time formatting only to panel types that have date/time in x axis

ahmadshaheer:
### Summary
- previously the uPlotXAxisValuesFormat was being applied to all panel types, which caused some graphs to break, e.g. in the before recording

#### Screenshots
Before:

https://github.com/user-attachments/assets/80d31604-f3f6-418a-997b-747f3a27dc02

After:

https://github.com/user-attachments/assets/11432e42-bd28-45de-87a9-ada12498ec55

----

> [!IMPORTANT]
> Apply `uPlotXAxisValuesFormat` only to panel types with date/time x-axis by updating `getAxes` function and its usage across multiple files.
> 
>   - **Behavior**:
>     - Apply `uPlotXAxisValuesFormat` only to panel types with date/time x-axis in `getAxes`.
>     - Affects `PANEL_TYPES.TIME_SERIES`, `PANEL_TYPES.BAR`, `PANEL_TYPES.PIE`.
>   - **Functions**:
>     - Update `getAxes` to accept `panelType` and conditionally apply `uPlotXAxisValuesFormat`.
>   - **Files**:
>     - Modify `AnomalyAlertEvaluationView.tsx`, `BillingUsageGraph.tsx`, `getUplotChartOptions.ts`, and `getUplotHistogramChartOptions.ts` to pass `panelType` to `getAxes`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 905d469951f4c92d3b6f754889ee794bb65bd0f1. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/src/lib/uPlotLib/utils/getAxes.ts
+++ b/frontend/src/lib/uPlotLib/utils/getAxes.ts
@@ -1,12 +1,27 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 // @ts-nocheck
 import { getToolTipValue } from 'components/Graph/yAxisConfig';
+import { PANEL_TYPES } from 'constants/queryBuilder';
 
 import { uPlotXAxisValuesFormat } from './constants';
 import getGridColor from './getGridColor';
 
+const PANEL_TYPES_WITH_X_AXIS_DATETIME_FORMAT = [
+	PANEL_TYPES.TIME_SERIES,
+	PANEL_TYPES.BAR,
+	PANEL_TYPES.PIE,
+];
+
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const getAxes = (isDarkMode: boolean, yAxisUnit?: string): any => [
+const getAxes = ({
+	isDarkMode,
+	yAxisUnit,
+	panelType,
+}: {
+	isDarkMode: boolean;
+	yAxisUnit?: string;
+	panelType?: PANEL_TYPES;
+}): any => [
 	{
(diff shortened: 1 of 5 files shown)


frontend/src/lib/uPlotLib/utils/getAxes.ts
```diff
@@ -19,7 +34,11 @@ const getAxes = (isDarkMode: boolean, yAxisUnit?: string): any => [
 			width: 0.3, // Width of the tick lines,
 			show: true,
 		},
-		values: uPlotXAxisValuesFormat,
```
SagarRajput-7, line 22: Have we covered all cases of non-supported Panel consumption? I would have liked it to implement in a more reverse way to what you done, like have a prop and apply only to the panels you want to add, and for all the rest cases should work without touching anything

SagarRajput-7 (approved):

frontend/src/lib/uPlotLib/utils/getAxes.ts
ahmadshaheer: Yes, @SagarRajput-7,

we are adding values: `uPlotXAxisValuesFormat` based on the following

```const PANEL_TYPES_WITH_X_AXIS_DATETIME_FORMAT = [
	PANEL_TYPES.TIME_SERIES,
	PANEL_TYPES.BAR,
	PANEL_TYPES.PIE,
];

PR #7018, opened 2025-02-03
feat(sqlmigration): consolidate all sqlmigrations into one package

grandwizard28:
### Summary

consolidate all sqlmigrations into one package


pkg/query-service/utils/testutils.go
```diff
@@ -19,11 +26,39 @@ func NewTestSqliteDB(t *testing.T) (testDB *sqlx.DB, testDBFilePath string) {
...
 	testDBFile.Close()
 
-	testDB, err = sqlx.Open("sqlite3", testDBFilePath)
+	sqlstore, err := sqlitesqlstore.New(context.Background(), providertest.NewSettings(), sqlstore.Config{Provider: "sqlite", Sqlite: sqlstore.SqliteConfig{Path: testDBFilePath}})
 	if err != nil {
-		t.Fatalf("could not open test db sqlite file: %v", err)
+		t.Fatalf("could not create test db sqlite store: %v", err)
 	}
 
+	sqlmigrations, err := sqlmigration.New(
```
nityanandagohain, line 34: nit: is this the right place for this ?

pkg/signoz/config.go
```diff
@@ -100,6 +126,12 @@ func mergeAndEnsureBackwardCompatibility(config *Config, deprecatedFlags Depreca
 			fmt.Println("Error parsing CONTEXT_TIMEOUT_MAX_ALLOWED, using default value of 600s")
 		}
 	}
+
+	if os.Getenv("STORAGE") != "" {
```
nityanandagohain, line 130: not needed, let's remove this

nityanandagohain (commented):
Looks good to me. Few minor comments. 

I would suggest deploying this to a fresh instance and testing before merging. Just to avoid any misses.

pkg/query-service/utils/testutils.go
grandwizard28: This testutils package is unfortunately being used in a lot of places.

We need better ways to write tests.

pkg/signoz/config.go
grandwizard28: We're setting this in our charts and at other places. Does it make sense to deprecate this as well in line with other deprecations?

nityanandagohain: Hmm okay, don't see any harm keeping it. We can remove it all together.

pkg/query-service/utils/testutils.go
nityanandagohain: I was saying something like should we move this to the sqlmigration package itself ?

nityanandagohain: this is to make sure that the test doesn't fail

nityanandagohain (approved):

PR #7060, opened 2025-02-06
Multi-Tenancy

mgilham:
### Summary

This PR offers shared-instance multi-tenancy for SigNoz. It is not ready to be merged in its current form, but it is functional, and it meets my company's needs for multi-tenancy in our environment.

I am opening the PR to get feedback from owners and see if this is a direction the project wants to go, and hopefully find alignment before doing more preparation for merging. I'm not picky about or attached to the design. If there are aspects of design you'd like to see changed, I can spend some amount of time to make changes. I'm on #contributing in Slack.

#### Related Issues / PR's

- https://github.com/SigNoz/signoz/discussions/1471
- Multi-Tenancy has also been listed on the product roadmap on the website.

#### Design Sketch

The scope of changes include:
- identifying the tenant associated with a request
- authorizing and isolating access to telemetry data in ClickHouse for the tenant
- authorizing and isolating access to user data in sqlite by tenant

##### Tenant identity
- We assume there is exactly one tenant associated with a given request for a given user. The tenant is identified by a string.
- This current implementation assumes the tenant is the Organization.
- The tenant is plumbed through the system where needed, in some places as a direct function parameter, and in others as a Context value.
- The tenant is sent through Context to the custom Prometheus ClickHouse metrics reader and applied in the SQL there as well.
- Tenant is also applied as a key to the query cache lookups.

##### Telemetry data
- We define ClickHouse parameterized views to "protect" the base tables. The views take the tenant as a parameter.
  - The views are used as drop-in replacements for the base tables in many places in the query service.
  - Some of the views also take the time window as parameters for performance reasons. In theory this might not be necessary as the ClickHouse query optimizer could push the predicates down, but when I checked for that I was not seeing reassuring query plans.
  - As an alternative to parameterized views, we could include the tenant filter in the dynamic SQL built inside the query service. There are pros and cons to both approaches we could discuss further (security, coupling, flexibility, performance, etc.)
- The views can be defined differently on the back end to implement tenancy in different ways. This may or may not be an advantage depending on the direction the project wants to take.
  - This gist contains "noop" views I used for local testing: https://gist.github.com/mgilham/55c4c746eb2d9b6320b77bca3b39bf90
  - In my company's full environment, we take the tenant to be either the Kubernetes namespace from the resource tags, or a prefix of the otel `service.name`.
- In some cases we add columns in or on top of the existing base tables to materialize the tenant identifier.
- As an alternative, we could simplify the design and view logic by establishing a common standard column for all relevant tables. This would require more table changes, but the simplicity could be worthwhile. For the telemetry intake, I would recommend considering the otel `service.namespace`, which would help in de-duplicating `service.name` across tenants.

##### User data in sqllite
- Filters are added to restrict all CRUD operations to the organization of the user making the request.

#### Open Items
- Custom user ClickHouse queries are not yet isolated.
  - This is mostly implemented in this PR, but the version of ClickHouse would need to upgrade to support SQL security settings to allow users to query the views without access to the base tables.
  - Custom queries are parsed to ensure the `tenant` parameter to the views match the authenticated tenant for the request. This parsing could be improved upon for generality, or discarded in favor of an alternate solution, e.g. custom per-tenant views.
- Tests still need to be updated
- Infra metrics
- SigNoz internal telemetry for multiple orgs?
- New Role for a meta-admin that can access data for all tenants?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -231,4 +231,5 @@ require (
 	k8s.io/utils v0.0.0-20240711033017-18e509b52bc8 // indirect
 )
 
-replace github.com/prometheus/prometheus => github.com/SigNoz/prometheus v1.12.0
+//replace github.com/prometheus/prometheus => github.com/SigNoz/prometheus v1.12.0
+replace github.com/prometheus/prometheus => ../prometheus
--- a/pkg/query-service/app/auth.go
+++ b/pkg/query-service/app/auth.go
@@ -46,6 +46,7 @@ func (am *AuthMiddleware) ViewAccess(f func(http.ResponseWriter, *http.Request))
 			return
 		}
 		ctx := context.WithValue(r.Context(), constants.ContextUserKey, user)
+		ctx = context.WithValue(ctx, constants.ContextTenantKey, user.Organization)
 		r = r.WithContext(ctx)
 		f(w, r)
 	}
@@ -69,6 +70,7 @@ func (am *AuthMiddleware) EditAccess(f func(http.ResponseWriter, *http.Request))
 			return
 		}
 		ctx := context.WithValue(r.Context(), constants.ContextUserKey, user)
+		ctx = context.WithValue(ctx, constants.ContextTenantKey, user.Organization)
 		r = r.WithContext(ctx)
 		f(w, r)
 	}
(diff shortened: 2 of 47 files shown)


grandwizard28:
Hi @mgilham,
Thank you for the PR! We really appreciate the effort you have put in here.

This PR has come at a time when we are also in the process of implementing multi-tenancy. At this point, this looks very different from the approach we are considering.

That being said, we are still in the early stages of our experiments with multi-tenancy. We will definitely revisit this PR when we get here :)

therealpandey:
Hi @mgilham,
Just wanted to get you an update, we are very close to finalising our multi-tenant sql schema and structure.

basanth-ivoyant:
@therealpandey - would this feature allow for different retention periods for different tenants?

Issue #7082, opened 2025-02-11
Metric details api | summary view apis


aniketio-ctrl:
## Is your feature request related to a problem?

*Please describe.*

PR #7083, opened 2025-02-11
feat(summary view): added metric details apis | 7082

aniketio-ctrl:
### Summary

Added metrics details apis

#### Related Issues / PR's

https://github.com/SigNoz/signoz/issues/7082

### DIFF
--- a/pkg/query-service/app/clickhouseReader/reader.go
+++ b/pkg/query-service/app/clickhouseReader/reader.go
@@ -5334,18 +5334,6 @@ func (r *ClickHouseReader) GetTotalTimeSeriesForMetricName(ctx context.Context, 
 	return timeSeriesCount, cardinality, fingerprint, nil
 }
 
-func (r *ClickHouseReader) GetActiveTimeSeriesForMetricName(ctx context.Context, metricName string, duration time.Duration) (uint64, *model.ApiError) {
-	milli := time.Now().Add(-duration).UnixMilli()
-	query := fmt.Sprintf("SELECT count(DISTINCT fingerprint) FROM %s.%s WHERE metric_name = '%s' and unix_milli >= ?", signozMetricDBName, signozSampleTableName, metricName)
-	var timeSeries uint64
-	// Using QueryRow instead of Select since we're only expecting a single value
-	err := r.db.QueryRow(ctx, query, milli).Scan(&timeSeries)
-	if err != nil {
-		return 0, &model.ApiError{Typ: "ClickHouseError", Err: err}
-	}
-	return timeSeries, nil
-}
-
 func (r *ClickHouseReader) GetAttributesForMetricName(ctx context.Context, metricName string) (*[]metrics_explorer.Attribute, *model.ApiError) {
 	query := fmt.Sprintf(`
         SELECT 
@@ -5395,4 +5383,16 @@ func (r *ClickHouseReader) GetAttributesForMetricName(ctx context.Context, metri
 	}
 
 	return &attributesList, nil
+}
+
+func (r *ClickHouseReader) GetActiveTimeSeriesForMetricName(ctx context.Context, metricName string, duration time.Duration) (uint64, *model.ApiError) {
+	milli := time.Now().Add(-duration).UnixMilli()
+	query := fmt.Sprintf("SELECT count(DISTINCT fingerprint) FROM %s.%s WHERE metric_name = '%s' and unix_milli >= ?", signozMetricDBName, signozSampleTableName, metricName)
+	var timeSeries uint64
+	// Using QueryRow instead of Select since we're only expecting a single value
+	err := r.db.QueryRow(ctx, query, milli).Scan(&timeSeries)
+	if err != nil {
+		return 0, &model.ApiError{Typ: "ClickHouseError", Err: err}
+	}
+	return timeSeries, nil
 }
--- a/pkg/query-service/app/clickhouseReader/reader.go
+++ b/pkg/query-service/app/clickhouseReader/reader.go
@@ -1184,7 +1184,7 @@ func (r *ClickHouseReader) GetUsage(ctx context.Context, queryParams *model.GetU
 
 func (r *ClickHouseReader) SearchTracesV2(ctx context.Context, params *model.SearchTracesParams,
 	smartTraceAlgorithm func(payload []model.SearchSpanResponseItem, targetSpanId string,
-		levelUp int, levelDown int, spanLimit int) ([]model.SearchSpansResult, error)) (*[]model.SearchSpansResult, error) {
+	levelUp int, levelDown int, spanLimit int) ([]model.SearchSpansResult, error)) (*[]model.SearchSpansResult, error) {
 	searchSpansResult := []model.SearchSpansResult{
 		{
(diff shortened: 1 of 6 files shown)


pkg/query-service/app/metricsexplorer/summary.go
```diff
@@ -66,3 +71,128 @@ func (receiver *SummaryService) FilterValues(ctx context.Context, params *metric
...
+	)
+
+	// Create a context with cancellation to stop other goroutines on the first error
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	handleError := func(err *model.ApiError) {
+		select {
+		case errCh <- err: // Send the error if no error was sent yet
+			cancel() // Cancel other goroutines
```
srikanthccv, line 89: Imagine this case: Call 1 operation failed, and an error was sent to the channel `errCh`. Since we are cancelling the context, the remaining will also fail and return errors, which also try to send to the channel. What is the behaviour, in this case, with the current code, and what is the expected behaviour?

pkg/query-service/app/metricsexplorer/summary.go
```diff
@@ -66,3 +71,128 @@ func (receiver *SummaryService) FilterValues(ctx context.Context, params *metric
...
+		default:
+		}
+	}
+
+	wg.Add(6)
```
srikanthccv, line 94: Would it be better to add `wg.Add(1)` inside the routine than the fixed number here?

pkg/query-service/app/metricsexplorer/summary.go
aniketio-ctrl: if we cancel other go routines with context ancel they will return error if the channel is full it could lead to deadlock situation what we can do is create a boolean to check if error was sent then no more errors can be sent to the error channel

pkg/query-service/app/metricsexplorer/summary.go
```diff
@@ -66,3 +71,128 @@ func (receiver *SummaryService) FilterValues(ctx context.Context, params *metric
...
+	//TODO: ADD ALERTS CONFIG
+
+	// Wait for all goroutines to complete
+	wg.Wait()
+	close(errCh)
+
+	// If an error occurred, return immediately
+	if apiErr := <-errCh; apiErr != nil {
+		return metrics_explorer.MetricDetailsDTO{}, apiErr
+	}
```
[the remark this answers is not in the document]
aniketio-ctrl: yeah you are right if we read from errch after closing it we will get nil, so error handling might not be correct

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -5310,3 +5310,89 @@ func (r *ClickHouseReader) GetAllMetricFilterUnits(ctx context.Context, req *met
 	}
 	return &response, nil
 }
+
+func (r *ClickHouseReader) GetMetricsDataPointsAndLastReceived(ctx context.Context, metricName string) (uint64, uint64, *model.ApiError) {
+	query := fmt.Sprintf("SELECT COUNT(*) AS data_points, MAX(unix_milli) AS last_received_time FROM %s.%s WHERE metric_name = ?", signozMetricDBName, signozSampleTableName)
```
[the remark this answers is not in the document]
aniketio-ctrl: already discussed.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -5298,3 +5298,87 @@ func (r *ClickHouseReader) GetAllMetricFilterUnits(ctx context.Context, req *met
...
+		return 0, 0, &model.ApiError{Typ: "ClickHouseError", Err: err}
+	}
+	return timeSeriesCount, cardinality, nil
+}
+
+func (r *ClickHouseReader) GetAttributesForMetricName(ctx context.Context, metricName string) (*[]metrics_explorer.Attribute, *model.ApiError) {
+	query := fmt.Sprintf(`
+SELECT 
+    kv.1 AS key,
+    arrayMap(x -> replaceAll(x, '"', ''), groupUniqArray(kv.2)) AS values,
```
srikanthccv, line 5332: Use trim `trim(BOTH '"' FROM x)`

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -5298,3 +5298,87 @@ func (r *ClickHouseReader) GetAllMetricFilterUnits(ctx context.Context, req *met
...
+    length(groupUniqArray(kv.2)) AS valueCount
+FROM %s.%s
+ARRAY JOIN arrayFilter(x -> NOT startsWith(x.1, '__'), JSONExtractKeysAndValuesRaw(labels)) AS kv
+WHERE metric_name = 'system_memory_usage'
+GROUP BY kv.1
+ORDER BY valueCount DESC;
+    `, signozMetricDBName, signozTSTableNameV41Day)
```
srikanthccv, line 5339: Use 1 week table

pkg/query-service/app/metricsexplorer/summary.go
```diff
@@ -68,3 +74,104 @@ func (receiver *SummaryService) FilterValues(ctx context.Context, params *metric
...
+			return err
+		}
+		metricDetailsDTO.TimeSeriesTotal = totalSeries
+		metricDetailsDTO.Cardinality = cardinality
+		return nil
+	})
+
+	// Call 4: GetActiveTimeSeriesForMetricName
+	g.Go(func() error {
+		activeSeries, err := receiver.reader.GetActiveTimeSeriesForMetricName(ctx, metricName, 30*time.Minute)
```
srikanthccv, line 122: We need at least 1 hour because we maintain hourly precision. Let's use 2 hours

srikanthccv (commented):
Some questions, mostly LGTM

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -5298,3 +5298,87 @@ func (r *ClickHouseReader) GetAllMetricFilterUnits(ctx context.Context, req *met
...
+	}
+	return dataPoints, uint64(lastRecievedTimestamp), nil // Convert to uint64 before returning
+}
+
+func (r *ClickHouseReader) GetTotalTimeSeriesForMetricName(ctx context.Context, metricName string) (uint64, uint64, *model.ApiError) {
+	query := fmt.Sprintf(`SELECT 
+     uniq(arrayJoin(arrayMap(x -> x.2, arrayFilter(x -> NOT startsWith(x.1, '__'), JSONExtractKeysAndValuesRaw(labels))))) AS cardinality,
+    count(DISTINCT fingerprint) AS timeSeriesCount
+FROM %s.%s
+WHERE metric_name = ?;`, signozMetricDBName, signozTSTableNameV41Day)
```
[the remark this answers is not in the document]
aniketio-ctrl: cardinality here is all the unique label values exist, and time series count, define total no of different time series we are getting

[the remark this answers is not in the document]
srikanthccv: metric cardinality refers to the number of unique time series that are produced by a combination of metric names and their associated labels. The unique label values exist is not the same thing as metric cardinality right?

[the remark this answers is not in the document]
aniketio-ctrl: yeah initially i only thought that but there were two values, then there would be no diff in cardinality and total time series

srikanthccv (approved):

PR #7055, opened 2025-02-06
fix: move migrations to bun

nityanandagohain:
Move migrations to bun by declaring structs.

I have not removed the old structs which are here and there as of now.

Few things are not working directly with bun right now will have to figure out.
* checks on value of column of organizations table didn't work. tried a lot of apporaches but couldn't get it working.

FIxes https://github.com/SigNoz/platform-pod/issues/444

----

> [!IMPORTANT]
> Migrated SQL migrations to Bun ORM by defining structs for tables and added corresponding models in `pkg/types`.
> 
>   - **Migration to Bun ORM**:
>     - Replaced raw SQL table creation with Bun ORM in `pkg/sqlmigration/000_add_data_migrations.go`, `001_add_organization.go`, `002_add_preferences.go`, and 7 other files.
>     - Defined structs for each table with fields, types, and constraints using Bun tags.
>     - Added foreign key constraints where applicable.
>   - **New Models**:
>     - Added new model files in `pkg/types` for `Agent`, `Dashboard`, `DataMigration`, `Integration`, `License`, `Organization`, `PersonalAccessToken`, `Pipeline`, `Preference`, and `SavedView`.
>     - Each model corresponds to a table and includes fields with Bun tags for ORM mapping.
>   - **Constraints and Issues**:
>     - Some constraints, like check constraints in `Organization`, are noted as not working.
>     - Foreign key constraints are added but some cascade options are not functioning as expected.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 1a78c8ae88bc8d18ec2c3f17fe620be28de75b14. It will automatically update as commits are pushed.</sup>

PR #7089, opened 2025-02-11
fix: add migration and postgres provider

nityanandagohain:
fixes https://github.com/SigNoz/platform-pod/issues/453

follow up of https://github.com/SigNoz/signoz/pull/7055

This makes sure that the migrations run on postgres.
Will rebase on the other pr is merged.

The API's work as expected after this change and no changes required on the API side.

----

> [!IMPORTANT]
> Add PostgreSQL support and ensure migrations run on PostgreSQL, with updates to provider configuration and migration scripts.
> 
>   - **PostgreSQL Support**:
>     - Add PostgreSQL provider in `pkg/sqlstore/postgressqlstore/provider.go`.
>     - Update `pkg/sqlstore/config.go` to include `PostgresConfig`.
>     - Add `pgdialect` to `go.mod` and `go.sum`.
>   - **Migrations**:
>     - Add `NewModifyDatetimeFactory` in `pkg/sqlmigration/011_modify_datetime.go` to handle datetime modifications for SQLite.
>     - Update `created_at` and `updated_at` fields in `pkg/sqlmigration/003_add_dashboards.go`, `004_add_saved_views.go`, and `005_add_agents.go` to remove `type:datetime`.
>   - **Miscellaneous**:
>     - Update dependencies in `go.mod` and `go.sum` to include PostgreSQL related packages.
>     - Comment out `postgressqlstore.NewFactory()` in `pkg/signoz/provider.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for d3040d3272471b825f9b8ab6e7e8b272a5f0b8b7. It will automatically update as commits are pushed.</sup>


pkg/sqlstore/pgstore/provider.go
```diff
@@ -0,0 +1,54 @@
+package pgstore
```
grandwizard28, line 1: The name of the package should be `postgressqlstore` just like `sqlitesqlstore`

pkg/sqlstore/pgstore/provider.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"context"
+	"database/sql"
+
+	"github.com/jmoiron/sqlx"
+	_ "github.com/lib/pq"
```
grandwizard28, line 8:
```suggestion
	_ "github.com/lib/pq"
```
`lib/pq`'s last commit was in 2023 and has gone into maintenance. Let's use either `pgdriver` from bun or https://github.com/jackc/pgx.

pkg/sqlmigration/003_add_dashboards.go
```diff
@@ -34,9 +34,9 @@ func (migration *addDashboards) Up(ctx context.Context, db *bun.DB) error {
 			bun.BaseModel `bun:"table:dashboards"`
```
grandwizard28, line 34: There is no need to change older migrations now.

grandwizard28 (changes requested):

pkg/sqlmigration/003_add_dashboards.go
nityanandagohain: this is required so that in postgres when new one is created it doesn't throw an error. It won't affect the existing sqlite data or the new ones.

grandwizard28: Ahhhh!

Issue #6947, opened 2025-01-27
Dashboard panel not visible in dashboard details


srikanthccv:
Dashboard getting saved with empty layout detail due to which on the dashboard page we are not seeing some dashboards 

Thread - https://signoz-team.slack.com/archives/C02TJ466H8U/p1728321178786499

related conversation thread - https://signoz-team.slack.com/archives/C02C7P6KE06/p1728319342518089

SagarRajput-7:
fixed with this - https://github.com/SigNoz/signoz/pull/7106

PR #7106, opened 2025-02-13
fix: fixed widgets not visible after saving in the dashboard detail

SagarRajput-7:
### Summary

Here we had a race condition in maintaining 2 interdependent variables `dashboardLayout` and `layouts` caused by the parallel execution of the get and put calls triggered from different sources.

> The get API gets called with the updated values and returns the correct result. Meanwhile, the put call doesn't wait for the get call to finish and update the variables; rather, it continues with the older value. In those cases when the put call finishes after the get call, it overrides the updated value. Hence, we see the widget coming and then going away.

#### Related Issues / PR's

- https://github.com/SigNoz/signoz/issues/6947

Solves following customer issues:
- https://signoz-team.slack.com/archives/C02C7P6KE06/p1728319342518089
- https://signoz-team.slack.com/archives/C02B85R4R0A/p1728367047032009

#### Screenshots

Before:
--------------
https://github.com/user-attachments/assets/c6deee0e-3651-4985-82fd-0d7264bfad30

After: 
--------------
https://github.com/user-attachments/assets/b9eb7c39-692e-4254-8855-7655bf9d4464

----

> [!IMPORTANT]
> Fix widgets visibility issue by adding `isDashboardFetching` check in `GridCardLayout.tsx` and updating dashboard context in `Dashboard.tsx`.
> 
>   - **Behavior**:
>     - Add `isDashboardFetching` check in `GraphLayout` in `GridCardLayout.tsx` to prevent saving layout during fetch.
>     - Modify `useDashboard` in `Dashboard.tsx` to set `isDashboardFetching` state during dashboard data fetch.
>   - **Context**:
>     - Add `isDashboardFetching` to `IDashboardContext` in `types.ts` and `Dashboard.tsx`.
>   - **Misc**:
>     - Update `DashboardProvider` in `Dashboard.tsx` to manage `isDashboardFetching` state.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 98ec6168e3e06008b538fdc0e5949f18dd77294c. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/src/providers/Dashboard/Dashboard.tsx
+++ b/frontend/src/providers/Dashboard/Dashboard.tsx
@@ -73,6 +73,7 @@ const DashboardContext = createContext<IDashboardContext>({
 	setDashboardQueryRangeCalled: () => {},
 	selectedRowWidgetId: '',
 	setSelectedRowWidgetId: () => {},
+	isDashboardFetching: false,
 });
 
 interface Props {
@@ -192,6 +193,8 @@ export function DashboardProvider({
 	const { t } = useTranslation(['dashboard']);
 	const dashboardRef = useRef<Dashboard>();
 
+	const [isDashboardFetching, setIsDashboardFetching] = useState<boolean>(false);
+
 	const mergeDBWithLocalStorage = (
 		data: Dashboard,
 		localStorageVariables: any,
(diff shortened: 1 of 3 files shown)


amlannandy (approved):

frontend/src/providers/Dashboard/Dashboard.tsx
```diff
@@ -256,10 +259,16 @@ export function DashboardProvider({
...
+			queryFn: async () => {
+				setIsDashboardFetching(true);
+				try {
+					return await getDashboard({
+						uuid: dashboardId,
+					});
+				} finally {
+					setIsDashboardFetching(false);
+				}
+			},
```
ahmadshaheer, line 271: can we achieve the same using `dashboardResponse.isFetching`

SagarRajput-7: no i tired it multiple tiime this was not working its always gives false

ahmadshaheer: if possible, can we revisit this and check why is isFetching is not working as expected :slightly_smiling_face: 

approving the PR :+1:

ahmadshaheer (approved):

Issue #7011, opened 2025-02-03
handle invalid invitation token gracefully in signup page


prashant-shahi:
## Bug description

In case of invalid/used/expired invitation tokens, we still show the Sign Up page which is an invalid state to be at.

As the outcome would be this:

<img width="843" alt="Image" src="https://github.com/user-attachments/assets/cdc4543b-8df4-4c64-ae9f-a55022c815d3" />

## Expected behavior

Ideally, we should redirect the users to Login Page where we should display the error message: "invalid/userd/expired invitation token. Please contact your admin."

## How to reproduce

1. Create an user invitation
2. Use the invitation once, log out, and use the same link again
3. Or update the invitation token from the URL to something else

## Version information 
* **Signoz version**: `latest`
* **Browser version**: NA
* **Your OS and version**: NA
* **Your CPU Architecture**(ARM/Intel):  NA

## Additional context

NA

AdityaPimpalkar:
Hi @prashant-shahi I'd like to give this a try, can you assign it to me? :) Thanks!

prashant-shahi:
Hi @AdityaPimpalkar! That's great to hear!

Let us know if you have any questions or are in need of assistance.

AdityaPimpalkar:
Hi @prashant-shahi Thanks!

I'm having trouble setting up the query-service locally. I have started a thread for it on the Signoz community slack here - https://signoz-community.slack.com/archives/C01LWQ8KS7M/p1739444440393049?thread_ts=1739444229.942839&cid=C01LWQ8KS7M 

Any help will be greatly appreciated! Thanks again!

OxDTH:
Hi @YounixM , I think no one is working on this issue so can you please assign it to me?

AviSharmaaa:
Hi @YounixM / @prashant-shahi, I did a full recon of the codebase before commenting.

Key findings:
- /signup?token= is a stale URL. The new invite flow (post PR #10600) sends users to /password-reset?token=. Nothing in the frontend or backend still generates /signup?token= links (confirmed via grep).
- There is no public v2 endpoint to validate an invite token, so no API call is needed.
- Login reads errors via URL query params callbackauthnerr and message at container/Login/index.tsx:54-56, not via location.state.

[...]

PR #7126, opened 2025-02-16
fix: handle invalid invitation token gracefully in signup page

AdityaPimpalkar:
### Summary

If the sign up fails, we redirect the users to the Login Page where we should display the error message: "invalid/used/expired invitation token. Please contact your admin."

#### Related Issues / PR's

#7011 

#### Screenshots
https://github.com/user-attachments/assets/0e3fd09d-c500-40f8-9ad3-0e6870f739c1

`frontend/src/pages/SignUp/SignUp.tsx`
`pkg/query-service/auth/auth.go`

### DIFF
--- a/pkg/query-service/auth/auth.go
+++ b/pkg/query-service/auth/auth.go
@@ -452,7 +452,7 @@ func RegisterInvitedUser(ctx context.Context, req *RegisterRequest, nopassword b
 	invite, err := ValidateInvite(ctx, req)
 	if err != nil {
 		zap.L().Error("failed to validate invite token", zap.Error(err))
-		return nil, model.UnavailableError(fmt.Errorf("invalid/used/expired invitation token, please contact your admin"))
+		return nil, model.ForbiddenError(fmt.Errorf("invalid/used/expired invitation token, please contact your admin"))
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/query-service/auth/auth.go
```diff
@@ -452,7 +452,7 @@ func RegisterInvitedUser(ctx context.Context, req *RegisterRequest, nopassword b
 	invite, err := ValidateInvite(ctx, req)
 	if err != nil {
 		zap.L().Error("failed to validate invite token", zap.Error(err))
-		return nil, model.BadRequest(model.ErrSignupFailed{})
+		return nil, model.ForbiddenError(fmt.Errorf("invalid/used/expired invitation token, please contact your admin"))
```
AdityaPimpalkar, line 455: Redirecting to Login page only on `Forbidden` errors. I'm assuming we don't want to redirect on any and all errors thrown by sign-up API, that would be a bad experience. So throwing a `Forbidden` instead of `BadRequest` made sense here.

If not then I'm open to suggestions! :)

therealpandey:
Hi @AdityaPimpalkar,
Thanks for the PR. We've made changes to the entire Member and Invite flow which takes care of this.

Issue #6614, opened 2024-12-11
3 front-end bugs


941477276:
## bug1：New alarm channel has no form verification
When adding an alarm channel, the alarm channel name, web address, and email are not verified in the form, resulting in an interface error.
![Image](https://github.com/user-attachments/assets/0d1b7a18-3d5e-46c0-99fc-577ab74c3bca)

## bug2：After searching for a card in the dashboard, clicking the Clear button does not execute the search again
![Image](https://github.com/user-attachments/assets/604c39f4-ca5d-4e18-a8cd-0f936af235ae)

## bug3：The useAlertHistoryQueryParams hooks is not accurate in judging whether relativeTime is null
If the address is `http://localhost:3301/alerts/history?ruleId=1&relativeTime=null`, the relativeTime obtained by `params.get(QueryParams.relativeTime)` is a string of `'null'` instead of `null`, so the statement after `??` will not be executed
![Image](https://github.com/user-attachments/assets/30bdecf7-0a83-4a4b-a0b1-8a63d8afa878)

Fix：
![Image](https://github.com/user-attachments/assets/9c666d2f-07da-47b7-9619-79efa5e6258c)

## Version information 
* **Signoz version**: 0.5.7
* **Browser version**: Chrome 127.0.6533.89
* **Your OS and version**: Windows 10
* **Your CPU Architecture**(ARM/Intel):  Intel(R) Core(TM) i5-10600

pranay01:
@941477276 can you share where have you setup this and docs you followd?

941477276:
Share the repaired code file and location?

PR #7135, opened 2025-02-17
fix: add form validation to alert channel form

AdityaPimpalkar:
### Summary

When adding an alarm channel, the alarm channel name, web address, and email are not verified in the form, resulting in an interface error.

#### Related Issues / PR's

#6614 - bug 1

#### Screenshots

https://github.com/user-attachments/assets/121ee27b-32ad-4af7-8aae-95e01671bed3

`frontend/src/container/FormAlertChannels/index.tsx`

### DIFF
--- a/frontend/src/container/FormAlertChannels/index.tsx
+++ b/frontend/src/container/FormAlertChannels/index.tsx
@@ -6,6 +6,7 @@ import ROUTES from 'constants/routes';
 import {
 	ChannelType,
 	EmailChannel,
+	MsTeamsChannel,
 	OpsgenieChannel,
 	PagerChannel,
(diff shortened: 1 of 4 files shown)


frontend/src/container/FormAlertChannels/index.tsx
```diff
@@ -84,12 +96,45 @@ function FormAlertChannels({
 		}
 	};
 
+	const isValidForm = (): boolean => {
```
AdityaPimpalkar, line 99: May have missed some more fields that are required for validation. I have added checks to the fields which I think are important at first glance, but let me know if I have missed any! :)

PR #7155, opened 2025-02-19
Fix: Multitenancy support for ORG

nityanandagohain:
Fixes https://github.com/SigNoz/platform-pod/issues/459

* removes user flags from backend code and db and also the frontend code is removed for that.
* ingestion keys should be as it is since we will remove it soon.
* Appdex settings has org id now. ( uses org id from context to populate the db)
  * The migration has to create a new table and drop the old one to change the primary key
* Timestamp datatype is changed for org, users and invites. Invites UI is updated to fix it to render properly. others are not required. 

---

Note: This PR doesn't cover anything related to creating multiple ORG. that will be tackled in the end.
Also org id is not enforced on groups as right now groups are initalized during startup we will have to solve for it differently

----

> [!IMPORTANT]
> Enhances multitenancy support by removing user flags, updating timestamp data types, and refactoring user-related code to include organization IDs.
> 
>   - **Behavior**:
>     - Removes user flags from backend and frontend.
>     - Updates `AppDao` to handle organization IDs for Apdex settings.
>     - Changes timestamp data types for `org`, `users`, and `invites`.
>   - **Database**:
>     - Drops `user_flags` table.
>     - Adds `org_id` to `groups` table.
>     - Modifies `apdex_settings` to include `org_id`.
>     - Migrates `created_at` to timestamp and `is_anonymous` to boolean.
>   - **Code Refactoring**:
>     - Replaces `basemodel.User` with `types.User` in multiple files.
>     - Updates `GetUserFromRequestContext` to return `types.GettableUser`.
>     - Refactors `auth.go`, `dao/interface.go`, and `dao/sqlite/auth.go` to use new user model.
>   - **Frontend**:
>     - Removes `setFlags.ts` and related components.
>     - Updates `Members/index.tsx` to handle new timestamp format.
>   - **Testing**:
>     - Updates integration tests to align with new user and org models.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 52aac60642f007b53a0b97235b6a5f37b4bcc7cc. It will automatically update as commits are pushed.</sup>


pkg/types/organization.go
```diff
@@ -6,73 +6,74 @@ import (
 	"github.com/uptrace/bun"
 )
 
+type AuditableModel struct {
+	CreatedAt time.Time `bun:"created_at" json:"createdAt"`
+	CreatedBy string    `bun:"created_by" json:"createdBy"`
+	UpdatedAt time.Time `bun:"updated_at" json:"updatedAt"`
+	UpdatedBy string    `bun:"updated_by" json:"updatedBy"`
+}
```
grandwizard28, line 14: We should separate this into 2:
1. One struct containing `CreatedAt` and `UpdatedAt`. (call it TimeAuditable)
2. One struct containing `CreatedBy` and `UpdatedBy` (call it UserAuditable)

For example:

- `Organization` needs only struct 1. 
- `Group` **probably** needs struct 1. (need to check this and confirm whether 2 will make sense)
- `User` needs only struct 1.

grandwizard28 (changes requested):

pkg/types/organization.go
nityanandagohain: *  organization settings can be updated by someone no.
* Group needs both. right now we tread group as role, but it should have both.
* same for user, a user role can be updated so it should have both.

frontend/src/container/OrganizationSettings/Members/index.tsx
```diff
@@ -281,7 +281,7 @@ function Members(): JSX.Element {
 				const { joinedOn } = record;
 				return (
 					<Typography>
-						{dayjs.unix(Number(joinedOn)).format(DATE_TIME_FORMATS.MONTH_DATE_FULL)}
+						{dayjs(joinedOn).format(DATE_TIME_FORMATS.MONTH_DATE_FULL)}
```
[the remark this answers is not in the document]
nityanandagohain: so I have updated the code to return time instead of unix timestamp. Those are the backend changes.

pkg/types/organization.go
nityanandagohain: done

pkg/sqlmigration/012_modify_org_domain.go
```diff
@@ -27,6 +27,11 @@ func (migration *modifyOrgDomain) Register(migrations *migrate.Migrations) error
 }
 
 func (migration *modifyOrgDomain) Up(ctx context.Context, db *bun.DB) error {
+	// only run this for old sqlite db
+	if db.Dialect().Name().String() != "sqlite" {
```
grandwizard28, line 31:
```suggestion
	if db.Dialect().Name() != dialect.SQLite {
```

pkg/sqlmigration/sqlmigration.go
```diff
@@ -86,3 +86,93 @@ func WrapIfNotExists(ctx context.Context, db *bun.DB, table string, column strin
 		return q.Err(ErrNoExecute)
 	}
 }
+
+func GetColumnType(ctx context.Context, tx bun.Tx, table string, column string) (string, error) {
```
grandwizard28, line 90:
```suggestion
func GetColumnType(ctx context.Context, bun bun.IDB, table string, column string) (string, error) {
```

There's a bun interface which accepts both transactions and the database. We should use that here.

pkg/sqlmigration/sqlmigration.go
```diff
@@ -86,3 +86,93 @@ func WrapIfNotExists(ctx context.Context, db *bun.DB, table string, column strin
...
+	var columnType string
+	var query string
+
+	if tx.Dialect().Name() == dialect.SQLite {
+		query = `SELECT type FROM pragma_table_info('` + table + `') WHERE name = '` + column + `'`
```
grandwizard28, line 95:
```suggestion
err := db.
			NewSelect().
			ColumnExpr("name").
			Table("pragma_table_info").
			Where("arg = ?", table).
			Where("name = ?", column).
			Scan(ctx, &result)
		if err != nil {
			if err == sql.ErrNoRows {
				return q
			}
			return q.Err(err)
		}
```

grandwizard28 (changes requested):

grandwizard28:
Seems we are doing a lot of stuff specific to sqlite and postgres. Let's sync on creating dedicated dialects for them instead of this if,else stuff.

frontend/src/container/OrganizationSettings/Members/index.tsx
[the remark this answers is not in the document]
ahmadshaheer: Got it @nityanandagohain, @grandwizard28.

I was curious, if we need to add zero date value here as well. if so, what should we display if we get zero value for date.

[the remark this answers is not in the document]
nityanandagohain: So this value will always be populated if there is an user.

pkg/sqlstore/postgressqlstore/dialect.go
```diff
@@ -0,0 +1,116 @@
+package postgressqlstore
```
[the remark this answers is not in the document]
nityanandagohain: its correct

pkg/types/organization.go
```diff
@@ -6,73 +6,77 @@ import (
 	"github.com/uptrace/bun"
```
grandwizard28, line 6: @nityanandagohain can we split this file now?

- One file auditable.go containing auditable models

- One file organization.go

- One file user.go (groups, invites can be kept in this file as this relate to user)

grandwizard28 (changes requested):

pkg/types/organization.go
nityanandagohain: done.

grandwizard28 (approved):

PR #7159, opened 2025-02-20
fix: materialize columns api fixed

nityanandagohain:
Fixes https://github.com/SigNoz/engineering-pod/issues/2135


srikanthccv:
@nityanandagohain the keys are showing up twice in the list, can you also check that once?

nityanandagohain:
Ahh @srikanthccv that has happened because in the mv's we use to have the types in caps i.e `Float64` etc but we moved to all smallcase in this pr through code https://github.com/SigNoz/signoz-otel-collector/commit/33af7105c8edef2fcb468922de0dba3c750c3b81

Now since the the keys tables for attribute and resource doesn't have any ttl they  have the old keys as well. That is why it is happening for instances which are old and not the new ones.

Let me create a migration in otel collector that will truncate these tables.

nityanandagohain:
@srikanthccv going ahead with merging this for now.

Issue #7164, opened 2025-02-20
Stage and run query should work regardless if the filters are not changed


nityanandagohain:
The stage and run query doesn't run if the filters or timerange is not changed.

It should work regardless.

PR #6701, opened 2024-12-23
Feat: back navigation support throughout the app

ahmadshaheer:
### Summary
Handled back navigation throughout the app 

Note: This PR handles the back navigation throughout the app wherever we have handled state management with query params, there are some pages where the query params are not being added, for those cases, the query params should first be added in the URL then we can handle the back navigation.

#### Screenshots

NA

#### Affected Areas and Manually Tested Areas
- Services 
    - Go to a service, and navigate back to the services page
    - Select environment from the dropdown, and go back, the environments should be removed
    - Filter services by any of the filters (e.g. host.name), and go back, the filters should be removed
    - Update the time range, and go back, the time range picker should retain the previous time range

- Service details
    - Select any range in any of the graphs, and go back, the graph and time picker should retain the previous state
    - Click on any key operation, it should redirect to traces page, and go back, the service details page should keep it's last state
    - Switch between the different tabs (i.e. Overview, DB Calls, External Metrics), and go back, the most recent tab should be selected
    - Click on any point in the graph, click on go to traces, it should redirect to traces page, and go back, the service details page should keep it's last state
    - Click on any point in the graph, click on go to logs, it should redirect to logs page, and go back, the service details page should keep it's last state
    
- Query Builder
    - Select any filters, stage query, and go back, the previous query should be retained
    - Add Group by filter, stage query, and go back, the previous query should be retained
    - Add Limit filter, stage query, and go back, the previous query should be retained
    - Add Having filter, stage query, and go back, the previous query should be retained
    - Change the aggregate every, stage query, and go back, the previous query should be retained
    - Clone the query, stage query, and go back, the previous query should be retained
    - Hide the query, stage query, and go back, the previous query should be retained
    - Add new query, stage query,  and go back, the previous query should be retained
    - Add a new formula, stage query, and go back, the previous query should be retained
    
- Traces explorer / Logs explorer
    - Switch between the different tabs, and go back, the most recent tab should be selected
    - Change the range picker, go back, the range picker should retain the previous state
    - Create an alert from traces explorer, go back, it should redirect to traces explorer page
    - Add to dashboard, go back, it should redirect to traces explorer page
    - Choose any quick filters, go back, the previous state should be retained
    - Choose a saved view, go back, the previous state should be retained

- Dashboards
    - Go to a dashboard, and press back button, it should redirect to dashboard list
    - Dashboard
        - Click on new panel, choose a type, go back, it should redirect to dashboard page
        - Change the time range, go back, it should retain the previous state
[...]

PR #7187, opened 2025-02-26
fix: disabled same url check for redirect in logs explorer

sawhil:
#### Related Issues / PR's

Resolves : https://github.com/SigNoz/signoz/issues/7164

#### Screenshots

NA

----

> [!IMPORTANT]
> Adds `preventSameUrlNavigation` option to `useSafeNavigate` and applies it in `QueryBuilderProvider` to allow same URL navigation for specific data sources.
> 
>   - **Behavior**:
>     - Adds `preventSameUrlNavigation` option to `useSafeNavigate` in `useSafeNavigate.ts`.
>     - Default is `true`, preventing navigation to the same URL.
>     - In `QueryBuilder.tsx`, disables same URL navigation prevention for `DataSource.LOGS` and `DataSource.TRACES`.
>   - **Functions**:
>     - Modifies `useSafeNavigate` to accept `UseSafeNavigateProps` with `preventSameUrlNavigation`.
>     - Updates `safeNavigate` logic to respect `preventSameUrlNavigation` flag.
>   - **Misc**:
>     - Updates `QueryBuilderProvider` to use the new `useSafeNavigate` option.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 7739a44f88d226e95a24a0117f763b4666cfcd11. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/src/hooks/useSafeNavigate.ts
+++ b/frontend/src/hooks/useSafeNavigate.ts
@@ -13,7 +13,7 @@ interface SafeNavigateParams {
 }
 
 interface UseSafeNavigateProps {
-	enableSameURLCheck?: boolean;
+	preventSameUrlNavigation?: boolean;
 }
 
 const areUrlsEffectivelySame = (url1: URL, url2: URL): boolean => {
@@ -83,7 +83,9 @@ const isDefaultNavigation = (currentUrl: URL, targetUrl: URL): boolean => {
 	return newKeys.length > 0;
 };
 export const useSafeNavigate = (
-	{ enableSameURLCheck }: UseSafeNavigateProps = { enableSameURLCheck: true },
+	{ preventSameUrlNavigation }: UseSafeNavigateProps = {
+		preventSameUrlNavigation: true,
+	},
 ): {
 	safeNavigate: (
 		to: string | SafeNavigateParams,
@@ -114,7 +116,7 @@ export const useSafeNavigate = (
 			const urlsAreSame = areUrlsEffectivelySame(currentUrl, targetUrl);
 			const isDefaultParamsNavigation = isDefaultNavigation(currentUrl, targetUrl);
 
-			if (enableSameURLCheck && urlsAreSame) {
+			if (preventSameUrlNavigation && urlsAreSame) {
 				return;
 			}
 
@@ -135,7 +137,7 @@ export const useSafeNavigate = (
 				);
 			}
 		},
-		[navigate, location.pathname, location.search, enableSameURLCheck],
+		[navigate, location.pathname, location.search, preventSameUrlNavigation],
 	);
 
 	return { safeNavigate };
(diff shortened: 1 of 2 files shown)


frontend/src/providers/QueryBuilder.tsx
```diff
@@ -763,7 +763,9 @@ export function QueryBuilderProvider({
 		[panelType, stagedQuery],
 	);
 
-	const { safeNavigate } = useSafeNavigate();
+	const { safeNavigate } = useSafeNavigate({
+		enableSameURLCheck: !(initialDataSource === DataSource.LOGS),
```
SagarRajput-7, line 767: Have we checked the same on traces page? Why only logs here.

sawhil: For now, just did for logs. After Shaheer shares his reasoning for the same change. Will either remove the whole check or add the flag from required sources.

ahmadshaheer: @sawhil ,

in the `areUrlsEffectivelySame` function, we are checking if the urls are effectively the same i.e. if the composite query and other params are the same. if so, we prevent the navigation. (https://github.com/SigNoz/signoz/pull/6701)

I think we need to add the `enableSameURLCheck` flag for traces as well.

frontend/src/hooks/useSafeNavigate.ts
```diff
@@ -78,7 +82,9 @@ const isDefaultNavigation = (currentUrl: URL, targetUrl: URL): boolean => {
 
 	return newKeys.length > 0;
 };
-export const useSafeNavigate = (): {
+export const useSafeNavigate = (
+	{ enableSameURLCheck }: UseSafeNavigateProps = { enableSameURLCheck: true },
```
ahmadshaheer, line 86: can we slightly change this to make it more readable?

e.g. 

we can change this to 
```tsx
	{ preventSameUrlNavigation }: UseSafeNavigateProps = { preventSameUrlNavigation: true },
```

if we want to enable same url navigation in the consumers. we could say

```tsx
	const { safeNavigate } = useSafeNavigate({
		preventSameUrlNavigation: false,
	});
```

sawhil: done

frontend/src/providers/QueryBuilder.tsx
sawhil: @nityanandagohain @YounixM For now I have disabled the same URL check for traces and logs for now.
This can be reviewed now.

ahmadshaheer (approved):
Thanks for making the change. LGTM! :+1: :slightly_smiling_face:

SagarRajput-7 (approved):

YounixM (approved):

PR #7191, opened 2025-02-27
fix(FE): add sorting to the timestamp and duration column in trace explorer

pranjalS9:
#### Screenshots

NA

### DIFF
--- a/frontend/src/container/TracesExplorer/ListView/utils.tsx
+++ b/frontend/src/container/TracesExplorer/ListView/utils.tsx
@@ -59,6 +59,9 @@ export const getListColumns = (
 			key: 'date',
 			title: 'Timestamp',
 			width: 145,
+			sorter: (a, b): number =>
+				new Date(a.date).getTime() - new Date(b.date).getTime(),
+			sortDirections: ['ascend', 'descend'],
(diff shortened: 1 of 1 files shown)


YounixM:
Thanks for the PR @pranjalS9 .

pranjalS9:
Can someone please check and merge PR please?
@YounixM @SagarRajput-7 @ahmadshaheer @sawhil

YounixM:
Thanks for PR @pranjalS9 . 

We will review this PR soon.

SagarRajput-7:
@pranjalS9 - current code and approach looks good, but this sorting is at the UI.

We shouldn't implement local sorting for this table since it's paginated, and the data is fetched through API requests. Local sorting only makes sense when the client already has the entire dataset. To illustrate, imagine a service has 100k spans in the last 15 minutes. Since we can't display them all, we use pagination through the list API. If users want to see the slowest span, they're looking for the slowest span in the entire last 15 minutes, not just the slowest span among the 200 spans on the current page. 

Please see if we can implement server-side sorting

I earlier worked on this - do take a look for reference - https://github.com/SigNoz/signoz/pull/5252

Thanks!

Issue #7183, opened 2025-02-25
LOGS[cache]: The start and end timestamp with stepintervalue causes wrong cache entry


nityanandagohain:
The start and end timestamp with stepintervalue causes wrong cache entry

```

Test 1

* Last 1 day - 43k 
* Last 3 days - 43k, 70, 43k

Cache
* First - start 1740399572000, end 1740485972000
    cache misses = 1740399572000 , end 1740485972000
* Second - start 1740226554000, end 1740485754000
    cache misses = 1740226554000, end 1740399572000


--------------------------------
Test 2

* Last 3 day - 43k, 43k, 43k
* Last 1 day - 43k

Cache
* First - start 1740226554000, end 1740485754000
    cache misses = 1740226554000, 1740485754000
* Second - start 1740399572000, end 1740485972000
    cache misses = 1740485754000, 1740485972000
```

PR #7212, opened 2025-03-04
Store complete intervals in cache and update logic for response

nityanandagohain:
part of #7183 

Follow up of https://github.com/SigNoz/signoz/pull/7201

* This PR makes sure data correct data gets stored in cache and make sure that response returns all the data without any filtering.
Main function to review is `filterSeriesPoints`

Changes are not added to querier v1, once this is reviewed I will copy it to v1.

image

Case where end is equal to a complete window end
image

----

> [!IMPORTANT]
> Enhance caching logic to store complete intervals and update response handling in `runBuilderQuery` for accurate data retrieval.
> 
>   - **Behavior**:
>     - Update `runBuilderQuery` in `helper.go` and `v2/helper.go` to store complete intervals in cache and ensure response returns all data.
>     - Introduce `FilterSeriesPoints` in `query_range.go` to filter series points for caching.
>     - Add `GetSeriesFromCachedDataV2` in `query_range.go` to handle series data retrieval without removing points.
>   - **Caching**:
>     - Update `FindMissingTimeRangesV2` and `MergeWithCachedSeriesDataV2` in `query_range_cache.go` for improved cache handling.
>     - Add `StoreSeriesInCache` in `query_range_cache.go` to store merged data in cache.
>   - **Interfaces**:
>     - Update `QueryCache` interface in `interface.go` to include new methods for cache handling.
>   - **Testing**:
>     - Add test cases in `query_range_test.go` for `FilterSeriesPoints` and `GetSeriesFromCachedDataV2`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for da627b977955bb4957cd83065faef4cddeb344ec. It will automatically update as commits are pushed.</sup>


pkg/query-service/querycache/query_range_cache.go
```diff
@@ -49,6 +49,92 @@ func WithFluxInterval(fluxInterval time.Duration) QueryCacheOption {
...
+			continue
+		}
+		// Stop processing if we've reached the end time
+		if data.Start >= end {
+			break
+		}
+
+		// Add missing range if there's a gap
+		if currentTime < data.Start {
+			missingRanges = append(missingRanges, MissInterval{Start: currentTime, End: min(data.Start, end)})
```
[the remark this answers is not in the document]
nityanandagohain: it's builtin

pkg/query-service/app/querier/v2/helper_test.go
```diff
@@ -0,0 +1,283 @@
...
+						{Timestamp: 1609473600000, Value: 1.0}, // 01 Jan 2021 04:00:00 UTC
+					},
+				},
+			},
+			expectedPoints: []*v3.Series{},
+			expectedStart:  1609473600000,
+			expectedEnd:    1609475400000,
+		},
+		{
+			name:         "respect actual empty series",
```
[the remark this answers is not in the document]
nityanandagohain: not helpful

pkg/query-service/app/querier/v2/helper.go
```diff
@@ -79,6 +80,76 @@ func prepareLogsQuery(_ context.Context,
...
+		if series.Points[0].Timestamp != missStart && series.Points[0].Timestamp < missStart {
+			// Remove the first point
+			points = points[1:]
+		}
+
+		// filter the last point if it is not a complete aggregation window
+		if !endCompleteWindow && series.Points[len(series.Points)-1].Timestamp == missEnd-(missEnd%stepMs) {
+			// Remove the last point
+			points = points[:len(points)-1]
+		}
```
srikanthccv, line 138: this will panic if the len of points is zero

pkg/query-service/app/querier/v2/helper.go
```diff
@@ -138,13 +210,36 @@ func (q *querier) runBuilderQuery(
...
+			filteredSeries, startTime, endTime := filterSeriesPoints(series, miss.Start, miss.End, builderQuery.StepInterval)
+
+			// making sure that empty range doesn't doesn't enter the cache
+			if len(filteredSeries) > 0 {
+				filteredMissedSeries = append(filteredMissedSeries, querycache.CachedSeriesData{
+					Data:  filteredSeries,
+					Start: startTime,
+					End:   endTime,
+				})
+			}
```
srikanthccv, line 223: Why do we have an additional check here? If the original query returned empty results, the filteredSeries would list with zero entries?

pkg/query-service/app/querier/v2/helper.go
```diff
@@ -138,13 +210,36 @@ func (q *querier) runBuilderQuery(
...
 				End:   miss.End,
 			})
 		}
-		mergedSeries := q.queryCache.MergeWithCachedSeriesData(cacheKeys[queryName], missedSeries)
+
+		filteredMergedSeries, err := q.queryCache.MergeWithCachedSeriesDataV2(cacheKeys[queryName], filteredMissedSeries)
+		if err != nil {
+			filteredMergedSeries = filteredMissedSeries
+		}
+		q.queryCache.StoreSeriesInCache(cacheKeys[queryName], filteredMergedSeries)
```
srikanthccv, line 237: What's the rationale for `MergeWithCachedSeriesDataV2` return err?

The error here is not the kind of error that the caller is expected to know and handle. As in if the calle doesn't know how to handle an error and should propagate up, then it makes sense to return an error to the caller so that the caller has control flow to handle it. In this case, error comes from the json.Unmarshal, which is the internal detail of the calle and it knows what to do in that case. Why do we need to prop that up and do the exact thing in caller.

pkg/query-service/app/querier/v2/helper.go
```diff
@@ -138,13 +210,36 @@ func (q *querier) runBuilderQuery(
...
+
+		mergedSeries, err := q.queryCache.MergeWithCachedSeriesDataV2(cacheKeys[queryName], missedSeries)
+		if err != nil {
+			mergedSeries = missedSeries
+		}
 
 		resultSeries := common.GetSeriesFromCachedData(mergedSeries, start, end)
```
srikanthccv, line 244: The logic in the function `GetSeriesFromCachedData` filters out the items that have timestamp < start but it's possible that the first point has timestamp < start which is not returned to client?

pkg/query-service/app/querier/v2/helper.go
nityanandagohain: filteredSeries will return empty series, but we don't want overall empty series to be appended as well.

so lets say filteredSeries is empty, if we don't add an if condition the append function will still append it.

pkg/query-service/app/querier/v2/helper.go
nityanandagohain: yeah, the getSeriesFromCache data actually filters that item, thanks for the catch.

pkg/query-service/app/querier/v2/helper.go
nityanandagohain: ideally this will never happen as if there is just one point and it is captured by the first one then this second condition will not even evalues as we are doint it on series.Points .

I don't see a case though which this can happen so I didn't add it.

pkg/query-service/app/querier/v2/helper.go
nityanandagohain: yeah make sense to fix the MergeWithCachedSeriesDataV2

pkg/query-service/app/querier/v2/helper.go
srikanthccv: > but we don't want overall empty series to be appended as well.

Why? A query could return an empty result which is it's actual result that should be cached?

nityanandagohain: Aa right, so this is the case when the actual series lenth is 0. I have added an or condition.

srikanthccv:
as the filter uses the `<=` here https://github.com/SigNoz/signoz/blob/main/pkg/query-service/app/logs/v4/query_builder.go#L362. Here is another case that's not handled properly.


```
		{
			name:         "Some test case",
			missStart:    1609466400000, // 01 Jan 2021 02:00:00 UTC
			missEnd:      1609470000000, // 01 Jan 2021 03:00:00 UTC
			stepInterval: 3600,          // 1 hour
			seriesList: []*v3.Series{
				{
					Points: []v3.Point{
						{Timestamp: 1609466400000, Value: 2.0}, // 01 Jan 2021 02:00:00 UTC
						{Timestamp: 1609470000000, Value: 3.0}, // 01 Jan 2021 03:00:00 UTC
					},
				},
			},
			expectedPoints: []*v3.Series{
				{
					Points: []v3.Point{
						{Timestamp: 1609466400000, Value: 2.0},
					},
				},
			},
			expectedStart: 1609466400000,
			expectedEnd:   1609470000000,
		},
```

If there is a data in the DB whose timestamp matches with the end, then CH will return data for it and it will be incorrectly stored in the cache even though it doesn't cover the full range but only exact timestamp that could also be the timestamp that is multiple of step interval.

pkg/query-service/app/querier/helper.go
```diff
@@ -80,6 +81,78 @@ func prepareLogsQuery(_ context.Context,
 	return query, err
 }
 
+// New function to filter series points
+func filterSeriesPoints(seriesList []*v3.Series, missStart, missEnd int64, stepInterval int64) ([]*v3.Series, int64, int64) {
```
srikanthccv, line 85: This should be shared instead of copying here because I don't see any difference?

srikanthccv (commented):
Final question. Otherwise, LGTM.

pkg/query-service/app/querier/helper.go
nityanandagohain: yeah, better to move it common

srikanthccv (approved):

PR #7235, opened 2025-03-06
feat(summary): added update metrics metadata api

aniketio-ctrl:
### Summary

Added update metrics metadata api, which is used to update metrics metadata , we are storing this info in a new table and we are going to fetch this details from cache whenever we need details regarding updated metrics, we are also pre loading cache after the start of our clickhouse reader

----

> [!IMPORTANT]
> Added an API to update metrics metadata, including server initialization for preloading metadata and enhancements to the reader interface and ClickHouseReader.
> 
>   - **API Addition**:
>     - Added `UpdateMetricsMetadata` API in `http_handler.go` to update metrics metadata.
>     - Added `ParseUpdateMetricsMetadataParams` in `parser.go` for parsing request parameters.
>   - **Server Initialization**:
>     - Preload metrics metadata in `server.go` using `PreloadMetricsMetadata()`.
>   - **Reader Interface**:
>     - Added `UpdateMetricsMetadata`, `DeleteMetricsMetadata`, `GetUpdatedMetricsMetadata`, and `CheckForLabelsInMetric` to `interfaces.go`.
>   - **ClickHouseReader**:
>     - Implemented `UpdateMetricsMetadata`, `DeleteMetricsMetadata`, `GetUpdatedMetricsMetadata`, and `PreloadMetricsMetadata` in `reader.go`.
>   - **Model Changes**:
>     - Added `UpdateMetricsMetadata` struct in `updatedMetricsMetadata.go`.
>     - Added `UpdateMetricsMetadataRequest` in `metrics_explorer/summary.go`.
>   - **Miscellaneous**:
>     - Updated dependencies in `go.mod` and `go.sum`.
>     - Minor logging and error handling improvements in `server.go` and `logging.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for e6964bddd1ab0472690cb69d568627cafb13a21e. It will automatically update as commits are pushed.</sup>


pkg/query-service/app/metricsexplorer/parser.go
```diff
@@ -82,8 +83,16 @@ func ParseInspectMetricsParams(r *http.Request) (*metrics_explorer.InspectMetric
...
 	}
 	return &inspectMetricParams, nil
 }
+
+func ParseUpdateMetricsMetadataParams(r *http.Request) (*metrics_explorer.UpdateMetricsMetadataRequest, *model.ApiError) {
+	var updateMetricsMetadataReq metrics_explorer.UpdateMetricsMetadataRequest
+	if err := json.NewDecoder(r.Body).Decode(&updateMetricsMetadataReq); err != nil {
+		return nil, &model.ApiError{Typ: model.ErrorBadData, Err: fmt.Errorf("cannot parse the request body: %v", err)}
+	}
+	return &updateMetricsMetadataReq, nil
```
srikanthccv, line 97: Let's add validation here. The update metric type should be one of the valid metric types, same for temporality. A gauge can't have temporality. If a metric is being updated to Histogram, and if there are not buckets in the data, we should reject etc... all sensible validations.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -6569,3 +6581,69 @@ LIMIT 40`,
...
+		return metricsMetadata, nil
+	}
+	return nil, nil
+}
+
+func (r *ClickHouseReader) PreloadMetricsMetadata(ctx context.Context) []error {
+	var allMetricsMetadata []model.UpdateMetricsMetadata
+	var errorList []error
+	// Fetch all rows from ClickHouse
+	query := fmt.Sprintf(`SELECT metric_name, type, description FROM %s.%s;`, signozMetricDBName, signozUpdatedMetricsMetadataTable)
```
srikanthccv, line 6635: Do we need a description in cache? Shouldn't there be temporality?

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -6569,3 +6581,69 @@ LIMIT 40`,
...
+}
+
+func (r *ClickHouseReader) GetUpdatedMetricsMetadata(ctx context.Context, metricName string) (*model.UpdateMetricsMetadata, *model.ApiError) {
+	metricsMetadata := new(model.UpdateMetricsMetadata)
+	retrieve, err := r.cache.Retrieve(ctx, constants.UpdatedMetricsMetadataCachePrefix+metricName, metricsMetadata, true)
+	if err != nil {
+		query := fmt.Sprintf(`SELECT metric_name, type, description
+FROM %s.%s
+WHERE metric_name = ?;`, signozMetricDBName, signozUpdatedMetricsMetadataTable)
+		err := r.db.Select(ctx, query, metricName)
```
srikanthccv, line 6620: This is incomplete? What are we reading the result into and returning when there is a cache miss?

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -6569,3 +6581,69 @@ LIMIT 40`,
...
+	return nil
+}
+
+func (r *ClickHouseReader) UpdateMetricsMetadata(ctx context.Context, req *model.UpdateMetricsMetadata) *model.ApiError {
+	apiErr := r.DeleteMetricsMetadata(ctx, req.MetricName)
+	if apiErr != nil {
+		return apiErr
+	}
+	insertQuery := fmt.Sprintf(`INSERT INTO %s.%s (metric_name, type, description, created_at)
+VALUES ( ?, ?, ?, ?);`, signozMetricDBName, signozUpdatedMetricsMetadataTable)
```
srikanthccv, line 6601: unit and temporality are missing?

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -6569,3 +6581,69 @@ LIMIT 40`,
...
+	err := r.db.Exec(ctx, insertQuery, req.MetricName, req.MetricType, req.Description, time.Now().UnixMilli())
+	if err != nil {
+		return &model.ApiError{Typ: "ClickHouseError", Err: err}
+	}
+	err = r.cache.Store(ctx, constants.UpdatedMetricsMetadataCachePrefix+req.MetricName, req, -1)
+	if err != nil {
+		return &model.ApiError{Typ: "ClickHouseError", Err: err}
+	}
```
srikanthccv, line 6609: This is not a ClickHouseError and I don't think we should return an error to the user if the cache store fails.

srikanthccv (commented):
We have `FetchTemporality` that should be updated as well.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -6569,3 +6601,146 @@ LIMIT 40`,
...
+func (r *ClickHouseReader) GetUpdatedMetricsMetadata(ctx context.Context, metricName string) (*model.UpdateMetricsMetadata, *model.ApiError) {
+	metricsMetadata := new(model.UpdateMetricsMetadata)
+
+	// Try to get from cache first
+	retrieveStatus, err := r.cache.Retrieve(ctx, constants.UpdatedMetricsMetadataCachePrefix+metricName, metricsMetadata, true)
+	if err == nil && retrieveStatus == cache.RetrieveStatusHit {
+		return metricsMetadata, nil
+	}
+
+	zap.L().Error("Error getting metrics updated metadata from cache ", zap.String("metric_name", metricName), zap.Error(err))
```
srikanthccv, line 6703: This should not be an error log. It can be just be info log saying entry doesn't exist in cache.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -6569,3 +6601,146 @@ LIMIT 40`,
...
+		&metricsMetadata.MetricName,
+		&metricsMetadata.MetricType,
+		&metricsMetadata.Description,
+		&metricsMetadata.Temporality,
+		&metricsMetadata.IsMonotonic,
+		&metricsMetadata.Unit,
+	)
+	if err != nil {
+		return nil, &model.ApiError{Typ: "ClickhouseErr", Err: fmt.Errorf("error querying metrics metadata: %v", err)}
+	}
```
srikanthccv, line 6719: What happens if there is no entry for the passed metric in the updated metadata table?

pkg/query-service/app/metricsexplorer/parser.go
```diff
@@ -82,8 +85,66 @@ func ParseInspectMetricsParams(r *http.Request) (*metrics_explorer.InspectMetric
...
+		return nil, &model.ApiError{Typ: model.ErrorBadData, Err: fmt.Errorf("cannot parse the request body: %v", err)}
+	}
+	updateMetricsMetadataReq.MetricName = mux.Vars(r)["metric_name"]
+
+	if updateMetricsMetadataReq.Description == "" {
+		return nil, &model.ApiError{
+			Typ: model.ErrorBadData,
+			Err: fmt.Errorf("description is required"),
+		}
+	}
```
srikanthccv, line 106: Let's not make it mandatory to provide a description

pkg/query-service/app/metricsexplorer/parser.go
```diff
@@ -82,8 +85,66 @@ func ParseInspectMetricsParams(r *http.Request) (*metrics_explorer.InspectMetric
...
+		}
+
+		if updateMetricsMetadataReq.IsMonotonic && updateMetricsMetadataReq.Temporality == v3.Delta {
+			return nil, &model.ApiError{
+				Typ: model.ErrorBadData,
+				Err: fmt.Errorf("histogram metrics with Delta temporality can't be monotonically increased"),
+			}
+		}
+	case v3.MetricTypeGauge:
+	case v3.MetricTypeSummary:
```
srikanthccv, line 138: This should be `case v3.MetricTypeGauge, v3.MetricTypeSummary:` because

```
	case v3.MetricTypeGauge:
	case v3.MetricTypeSummary:

```

different from
```
case v3.MetricTypeGauge, v3.MetricTypeSummary:
```

pkg/query-service/app/metricsexplorer/parser.go
aniketio-ctrl: let the client send the same one, otherwise partial metadata would be there in cache and partial in tables,

pkg/query-service/model/updatedMetricsMetadata.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+type UpdateMetricsMetadata struct {
+	MetricName  string         `json:"metricName" ch:"metric_name"`
+	MetricType  v3.MetricType  `json:"metricType" ch:"type"`
+	Description string         `json:"description" ch:"description"`
+	Unit        string         `json:"unit" ch:"unit"`
+	Temporality v3.Temporality `json:"temporality" ch:"temporality"`
+	IsMonotonic bool           `json:"is_monotonic" ch:"is_monotonic"`
+	CreatedAt   time.Time      `json:"created_at" ch:"created_at"`
```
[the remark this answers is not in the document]
aniketio-ctrl: yeah it will convert it into int as seconds

pkg/query-service/app/metricsexplorer/parser.go
srikanthccv: It's possible some metrics don't have the description in the first place. Why would we reject the request, and why does the client need to send "same one". The client can send a partial payload updating the type, but in return they can't get the "description is required" where there is no description or no change in the description.

aniketio-ctrl: my concern is that once the user has updated metadata, we should be storing it in once plase , so every time we should be fetching metadata, it should come first from our table, and if there is no description, for that metrics, let them send empty one, we will consider empty one ill remove the check but metadata source of truth should be first from that table

srikanthccv (dismissed):

amlannandy (approved):

PR #7241, opened 2025-03-07
New autocomplete endpoint with filters

shivanshuraj1333:


pkg/query-service/model/v3/v3.go
```diff
@@ -366,9 +412,10 @@ func (a AttributeKey) Validate() error {
 }
 
 type FilterAttributeValueResponse struct {
-	StringAttributeValues []string      `json:"stringAttributeValues"`
-	NumberAttributeValues []interface{} `json:"numberAttributeValues"`
-	BoolAttributeValues   []bool        `json:"boolAttributeValues"`
+	StringAttributeValues []string                      `json:"stringAttributeValues"`
+	NumberAttributeValues []interface{}                 `json:"numberAttributeValues"`
+	BoolAttributeValues   []bool                        `json:"boolAttributeValues"`
+	RelatedValues         *FilterAttributeValueResponse `json:"relatedValues"`
```
srikanthccv, line 418: This is precisely why the frontend breaks. The frontend code is written in a way that we can't add additional fields to the response structure.

shivanshuraj1333: Making `RelatedValues` optional in the response, and only add it when we are calling the new API should fix the problem.

srikanthccv (approved):

PR #7323, opened 2025-03-17
fix: css style override in planned downtime table

### LINKED ISSUE #7214: CSS style override causing layout issues on planned downtime page
Image

tarunsai217:
## Description
This PR fixes the CSS style override issue in the Planned Downtime page where gloabl scoped table cell styles from InfraMonitoringK8s.styles.scss were overriding component-specific styles, causing layout issues.

## Changes
- Added specific CSS rules with !important flags to ensure the planned downtime table cells maintain their correct width
- Fixed the table cell width to 736px as intended by the component design

## Related Issue
Fixes #7214

## Testing
- Verified that the table cells maintain the correct width
- Confirmed that the layout no longer breaks after navigation

## Screenshots 
Screenshot 2025-03-17 at 8 56 36 AM

### DIFF
--- a/frontend/src/container/PlannedDowntime/PlannedDowntime.styles.scss
+++ b/frontend/src/container/PlannedDowntime/PlannedDowntime.styles.scss
@@ -274,7 +274,9 @@
 	.ant-table-cell {
 		padding: 0 !important;
 		border: 0 !important;
-		width: 736px;
+		min-width: 736px !important;
+                max-width: 736px !important;
+                width: 736px !important;
 	}
 
 	.ant-table-tbody {


SagarRajput-7:
@tarunsai217 - though the fix independently works, adding `min-width` with `!important` does fix the issue, but it doesn't get to the root of the problem, that is, the style was getting influenced by having `antd` style override globally.

Hence, your fix would have solved it but couldn't ensure that other places are not broken.

Fix - https://github.com/SigNoz/signoz/pull/7425

Thanks for the contribution and effort put in.

PR #7341, opened 2025-03-17
fix(license): add check for current route for workspace access

vikrantgupta25:
### Summary

- add check for current route for workspace access


vikrantgupta25:
contributes to - https://github.com/SigNoz/platform-pod/issues/467

YounixM (approved):

PR #7378, opened 2025-03-20
fix: provide proper naming to signoz package to publish it

shivanshuraj1333:
fixes: https://github.com/SigNoz/signoz/issues/7379


nityanandagohain (approved):
I don't see any issue changing this, but @srikanthccv @prashant-shahi  please add your thoughts as well if you think there will be some issue with this change

PR #7415, opened 2025-03-24
fix: stage and run query button same url navigation enabled

sawhil:
### Summary

**Bug** : Stage and Run query wouldn't run the query if there are no changes in the `URL` which in our case is controlled by the QB and time filters.

**Cause behind the bug** : A new useSafeNavigate hook was recently implemented across the product to replace the manual navigation with this hook. It basically does some type checking on the target url and safely creates it and then navigates to the created target URL.
It aslo has a safety check which checks if the URL has changed from the previous refresh or not to avoid unnecessary refreshes. And the same was happening with QB as well.

Previously to fix this I added a boolean param to control the same url check `preventSameUrlNavigation`, so as not to effect other places where this hook was being used. After discussion back then we disabled the same url check from Logs Explorer and Traces Explorer.

Now we are disabling the same URL check directly from QB. Now wherever we will have QB we won't be having the same URL check.

Ref Code : 
image
 

#### Related Issues / PR's

Fixes : https://github.com/SigNoz/signoz/issues/7164

#### Screenshots

NA

Impacted places : 

* Logs QB
* Traces QB
* Dashboard panel creation / edit QB
* Alert creation QB

@nityanandagohain Please mention any more areas to test If I missed some areas

----

> [!IMPORTANT]
> Allow same URL navigation in `QueryBuilder.tsx` by setting `preventSameUrlNavigation` to `false` in `useSafeNavigate`.
> 
>   - **Behavior**:
>     - In `QueryBuilder.tsx`, `useSafeNavigate` now allows navigation to the same URL by setting `preventSameUrlNavigation` to `false`.
>     - Previously, navigation to the same URL was prevented for `DataSource.LOGS` and `DataSource.TRACES`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 20b01a3d334adffc1342a1ccd35642a3012b475f. It will automatically update as commits are pushed.</sup>


nityanandagohain:
@sawhil do mention why this change was added in the first place and what happens since we are disabling it.

nityanandagohain (approved):

PR #7450, opened 2025-03-27
fix: added auth token refresh logic in event source provider error handler

sawhil:
#### Screenshots

NA

### DIFF
--- a/frontend/src/container/LiveLogs/LiveLogsContainer/index.tsx
+++ b/frontend/src/container/LiveLogs/LiveLogsContainer/index.tsx
@@ -50,6 +50,8 @@ function LiveLogsContainer(): JSX.Element {
 		handleCloseConnection,
 		initialLoading,
 		isConnectionLoading,
+		isConnectionError,
+		reconnectDueToError,
 	} = useEventSource();
 
 	const compositeQuery = useGetCompositeQueryParam();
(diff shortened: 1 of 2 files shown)


nityanandagohain (approved):

SagarRajput-7 (approved):

frontend/src/container/LiveLogs/LiveLogsContainer/index.tsx
```diff
@@ -86,8 +85,8 @@ function LiveLogsContainer(): JSX.Element {
 	);
 
 	const handleError = useCallback(() => {
-		notifications.error({ message: 'Sorry, something went wrong' });
-	}, [notifications]);
+		console.error('Sorry, something went wrong');
```
YounixM, line 88: Should we update the error message to be something more informative so that we can use this to debug in future ?

frontend/src/container/LiveLogs/LiveLogsContainer/index.tsx
```diff
@@ -153,6 +152,23 @@ function LiveLogsContainer(): JSX.Element {
...
+	useEffect((): (() => void) | undefined => {
+		if (isConnectionError && reconnectDueToError && compositeQuery) {
+			// Small delay to prevent immediate reconnection attempts
+			const reconnectTimer = setTimeout(() => {
+				handleStartNewConnection(compositeQuery);
+			}, 1000);
+
+			return (): void => clearTimeout(reconnectTimer);
+		}
+		return undefined;
```
YounixM, line 164: Why do we need an explicit return here ?

frontend/src/providers/EventSource.tsx
```diff
@@ -63,16 +71,49 @@ export function EventSourceProvider({
...
+			notifications.error({ message: 'Sorry, something went wrong' });
+			// If token refresh failed, logout the user
+			if (!eventSourceRef.current) return;
+			eventSourceRef.current.close();
+			setIsConnectionError(true);
+			Logout();
+		} catch (error) {
+			// If there was an error during token refresh, we'll just
+			// let the component handle the error state
+			notifications.error({ message: 'Sorry, something went wrong' });
```
YounixM, line 109: Error messages should be informative (if they can be), or should convey steps the user can take post to error

frontend/src/container/LiveLogs/LiveLogsContainer/index.tsx
sawhil: This is a component level error message, since we are now handling the error into the event context. I have added the proper error message there.
This is just in case of fallback now.

frontend/src/providers/EventSource.tsx
sawhil: FIxed this

YounixM (approved):

Issue #7384, opened 2025-03-20
Intermediate 404 page on visiting SigNoz URL


makeavish:
Intermediate 404 page on visiting deployment URL
<https://www.loom.com/share/a74816db8b8649beb68acf4125ff5a6c?sid=e590d2f2-fced-42cd-9ffd-dddad81ab6b7>
Maybe due to redirection

[Slack Message](https://signoz-team.slack.com/archives/C02TJ466H8U/p1742465999103329?thread_ts=1742465999.103329&cid=C02TJ466H8U)

PR #7454, opened 2025-03-27
fix: handle 404 redirection on root route

aks07:
### Summary

This fixes the intermittent "Not Found" screen that appears on the root route before the appropriate redirection is triggered.

#### Related Issues / PR's

fixes: https://github.com/SigNoz/signoz/issues/7384

#### Screenshots

NA

----

> [!IMPORTANT]
> Adds a root path route in `AppRoutes/index.tsx` to render `Home` and prevent 404 errors before redirection.
> 
>   - **Behavior**:
>     - Adds a new `Route` for the root path `/` in `AppRoutes/index.tsx` to render the `Home` component.
>     - Prevents 404 error on root path before redirection.
>   - **Misc**:
>     - Imports `Home` from `pageComponents` in `AppRoutes/index.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for d815c847eede1ce645b0d850954b5de3fc0c0696. It will automatically update as commits are pushed.</sup>


frontend/src/AppRoutes/index.tsx
```diff
@@ -310,6 +311,7 @@ function App(): JSX.Element {
 																/>
 															))}
 
+															<Route exact path="/" component={Home} />
```
SagarRajput-7, line 314: instead of this - we can have `path: ['/', ROUTES.HOME],` in the routes.ts

aks07: Defining the root route in defaultRoutes would prevent the redirection from / to /home, as it would then be considered an existing programmed route, and the current logic does not apply redirection in such cases.

SagarRajput-7 (approved):

YounixM (approved):

PR #7518, opened 2025-04-03
feat(sqlmigration): update the user related tables according to new schema

vikrantgupta25:
### Summary

- update the user related tables according to the new schema 
- fixed the issue where the delete user was failing because of presence of PAT or reset password request


vikrantgupta25:
contributes to - https://github.com/SigNoz/platform-pod/issues/607

grandwizard28 (approved):

PR #7564, opened 2025-04-09
chore: add where clause visitor implementation for query expression

srikanthccv:
### Summary

Uses the visitor implementation to prepare the Where clause. Index filter and resource fingerprint filter will be added in follow-up.

----

> [!IMPORTANT]
> Add `WhereClauseVisitor` for ClickHouse WHERE clause generation and remove `HASNONE` function from grammar.
> 
>   - **Behavior**:
>     - Implements `WhereClauseVisitor` to convert filter expressions into ClickHouse WHERE clauses in `where_clause_visitor.go`.
>     - Removes `HASNONE` function from grammar in `FilterQuery.g4` and related files.
>     - Adds `QueryStringToKeysSelectors` in `query_to_keys.go` to convert query strings to field key selectors.
>   - **Tests**:
>     - Adds tests for `QueryStringToKeysSelectors` in `query_to_keys_test.go`.
>     - Adds tests for `WhereClauseVisitor` in `where_clause_visitor_test.go`.
>     - Adds tests for `GetCondition` and JSON body search in `condition_builder_test.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 505a6970a412b96ea02405fb5a66ae962c171352. It will automatically update as commits are pushed.</sup>


pkg/parser/grammar/where_clause_visitor.go
```diff
@@ -0,0 +1,587 @@
...
+	// for the body json search, we need to add search on the body field even
+	// if there is a field with the same name as attribute/resource attribute
+	// Since it will ORed with the fieldKeysForName, it will not result empty
+	// when either of them have values
+	if strings.HasPrefix(fieldKey.Name, telemetrylogs.BodyJSONStringSearchPrefix) {
+		fieldKeysForName = append(fieldKeysForName, fieldKey)
+	}
+
+	// TODO(srikanthccv): do we want to return an error here?
+	// should we infer the type and auto-magically build a key for expression?
```
nityanandagohain, line 564: this will happen when the it doesn't contain in the attribute keys  right ?

nityanandagohain (commented):
minor comments

pkg/parser/grammar/where_clause_visitor.go
srikanthccv: >this will happen when the it doesn't contain in the attribute keys right ?

When the key is not present in the `fieldKeys` for some reason i.e key we have never seen or missed in keys table for any reason.

>Also where will the enrichment logic be present ?

We can't enrich a key with one type or data type because there can be more than one combination, and in that case, we query across all the combinations and return the results along with the warning. If the question is where will the `fieldKeys` come from, they will be queried from DB and passed as args.

pkg/parser/grammar/where_clause_visitor_test.go
```diff
@@ -0,0 +1,639 @@
...
+		})
+		if err != nil {
+			t.Errorf("Error converting query to ClickHouse: %v", err)
+		}
+		if chQuery != c.expectedSearchString {
+			t.Errorf("Expected %s, got %s", c.expectedSearchString, chQuery)
+		}
+		if !reflect.DeepEqual(chQueryArgs, c.expectedSearchArgs) {
+			for i, arg := range chQueryArgs {
+				fmt.Printf("Expected %v with type %T, got %v with type %T\n", c.expectedSearchArgs[i], c.expectedSearchArgs[i], arg, arg)
```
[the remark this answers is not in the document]
srikanthccv: Just below there is `t.Errorf`. I kept the Printf for detailed debugging when it fails. I will change to t.Logf

pkg/parser/grammar/where_clause_visitor.go
nityanandagohain: Hmm, so previously if we didn't find anything then we fallback to string though. While I don't have much issue with this new implementation as that is a rare case.

nityanandagohain: Add one error test case as well, so that in future if someone changes something we know that error was expected.

srikanthccv: We have a retention of 15 days for keys, but the main data retention is customisable, which could potentially create a situation where keys exist in the main data but the keys don't because of key retention. The error is probably fine in a genuine error case, but in other cases, like retention, we can't return an error. I will think about what we should do about it (assuming String could result in no results when the results are expected) and address it separately.

nityanandagohain: ok, then let's just add an error test case for now and merge this PR.

srikanthccv: There is already one https://github.com/SigNoz/signoz/pull/7564/files#diff-e1250b57770f50babc406cd27231128fad68f212e8705a7dbc004db6131da3faR590. See line 590 in test file.

Issue #7484, opened 2025-03-30
/status page incorrectly shows an upgrade is available


flawaetz:
## Bug description

The `/status` page incorrectly shows that an upgrade is available due to what looks like a `v` being present in one version number and not the other.

<img width="949" alt="Image" src="https://github.com/user-attachments/assets/40dbaebb-b42e-4955-a6f2-739c087e7ea4" />

## Expected behavior

Recognize that the versions are the same.

## How to reproduce

1. Install signoz via Linux standalone binary method.
2. Check /status page

## Version information 
* **Signoz version**:
* 0.76.2
* **Browser version**:
* Firefox latest
* **Your OS and version**:
* Mac OS 15.3
* **Your CPU Architecture**(ARM/Intel): 
ARM

vikrantgupta25:
@grandwizard28 please have a look !

saravadeo:
I've opened a fix for this in #10411. The root cause was that `checkVersionState` compared the backend version (e.g. `0.76.2`) with the GitHub release tag name (e.g. `v0.76.2`) using a simple string equality check — the `v` prefix mismatch caused a false "upgrade available" prompt. The fix normalizes both version strings by stripping the `v` prefix before comparison.

PR #7594, opened 2025-04-13
fix(status): normalize version strings to prevent false upgrade alert

Ayush9026:
### Summary

The status page showed a false upgrade alert due to one version having a 'v' prefix (e.g., v0.76.2) and the other not (0.76.2), causing a mismatch

#### Related Issues / PR's
https://github.com/SigNoz/signoz/issues/7484

## Fix
Normalize both version strings before comparison by stripping the leading 

----

> [!IMPORTANT]
> Fixes false upgrade alert by normalizing version strings in `checkVersionState` in `app.ts`.
> 
>   - **Behavior**:
>     - Fixes false upgrade alert by normalizing version strings in `checkVersionState` in `app.ts`.
>     - Strips 'v' prefix from version strings before comparison.
>   - **Functions**:
>     - Adds `normalizeVersion` function within `checkVersionState` to handle version string normalization.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 3efd25ed84ae3815ee8aa8c1cf12e8f18c4dff80. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/src/utils/app.ts
+++ b/frontend/src/utils/app.ts
@@ -28,8 +28,14 @@ export const checkVersionState = (
 	currentVersion: string,
 	latestVersion: string,
 ): boolean => {
-	const versionCore = currentVersion?.split('-')[0];
-	return versionCore === latestVersion;
+	// Normalize versions by removing 'v' prefix if present
+	const normalizeVersion = (version: string): string => 
+		version?.startsWith('v') ? version.slice(1) : version;
+	
+	const currentNormalized = normalizeVersion(currentVersion?.split('-')[0]);
+	const latestNormalized = normalizeVersion(latestVersion);
+	
+	return currentNormalized === latestNormalized;
 };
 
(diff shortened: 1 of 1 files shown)


frontend/src/utils/app.ts
```diff
@@ -28,8 +28,14 @@ export const checkVersionState = (
 	currentVersion: string,
 	latestVersion: string,
 ): boolean => {
-	const versionCore = currentVersion?.split('-')[0];
-	return versionCore === latestVersion;
+	// Normalize versions by removing 'v' prefix if present
```
YounixM, line 31: Move the function out of the current scope to make it reusable

YounixM:
@Ayush9026 : Added a minor comment. Please address it.

Issue #7651, opened 2025-04-16
BE: Aggreation on floating point fields returns data in scientific notation.


nityanandagohain:
If a field is float and we group by that field then the data is returned in scientific notation.

same issue is seen in the trace detail attributs as well

![Image](https://github.com/user-attachments/assets/0c06c546-a728-46b1-abfa-87617e691ad4)

PR #7653, opened 2025-04-16
fix: proper formatting of floating point

nityanandagohain:
Fixes https://github.com/SigNoz/signoz/issues/7651

The `%v` format specifier covert floating values to scientific notation randomly.

eg:- 1240000 is converted to "1.24e+06"

The fix makes sure to use %f and when the number has nothing after decimal print it as integer value.

-----
Before

Trace detail :
image

Log aggregation : 
image

-----
After

Trace detail: 
image

Log aggreagtion: 
image

----

> [!IMPORTANT]
> Replaces `%v` with `strconv.FormatFloat` in `reader.go` to ensure consistent floating point formatting, addressing issue #7651.
> 
>   - **Behavior**:
>     - Replaces `%v` with `strconv.FormatFloat` for floating point numbers in `reader.go` to ensure consistent formatting.
>     - Formats numbers as integers if they have no decimal part.
>   - **Functions**:
>     - Updates `GetWaterfallSpansForTraceWithMetadata`, `readRow`, and `SearchTracesV2` to use `strconv.FormatFloat`.
>     - Similar changes in 8 other functions.
>   - **Misc**:
>     - Fixes issue #7651 related to inconsistent floating point formatting.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 071346686275ea97bc901e172ef085dc68f91e6e. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/pkg/query-service/utils/format.go
+++ b/pkg/query-service/utils/format.go
@@ -389,3 +389,10 @@ func NormalizeMap(data map[string]uint64) map[string]float64 {
 
 	return normalized
 }
+
+func FormatFloat(f float64) string {
+	if f == math.Trunc(f) {
+		return fmt.Sprintf("%.0f", f)
+	}
+	return fmt.Sprintf("%g", f)
+}
--- a/pkg/query-service/utils/format.go
+++ b/pkg/query-service/utils/format.go
@@ -389,10 +389,3 @@ func NormalizeMap(data map[string]uint64) map[string]float64 {
 
 	return normalized
 }
-
-func FormatFloat(f float64) string {
-	if f == math.Trunc(f) {
-		return fmt.Sprintf("%.0f", f)
-	}
-	return fmt.Sprintf("%g", f)
-}
--- a/.github/workflows/build-enterprise.yaml
+++ b/.github/workflows/build-enterprise.yaml
@@ -73,7 +73,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: frontend/.env
-          key: dotenv-${{ github.sha }}
+          key: enterprise-dotenv-${{ github.sha }}
   js-build:
     uses: signoz/primus.workflows/.github/workflows/js-build.yaml@main
     needs: prepare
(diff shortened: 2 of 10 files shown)


eKuG (commented):
Do we need test cases for the same? Lets see if we can do Sanity checks. Rest the functionality LGTM

eKuG (dismissed):

nityanandagohain:
Yep, added tests.

pkg/query-service/utils/format.go
```diff
@@ -389,3 +389,10 @@ func NormalizeMap(data map[string]uint64) map[string]float64 {
 
 	return normalized
 }
+
+func FormatFloat(f float64) string {
+	if f == math.Trunc(f) {
+		return fmt.Sprintf("%.0f", f)
+	}
+	return fmt.Sprintf("%g", f)
```
srikanthccv, line 397: As I understand, this change is trying to format the non-decimal floats. The decimal floats will still be formatted as they are now. Is that the desired outcome https://go.dev/play/p/trAh_ppYSRZ?

There will be people complaining the same about floats with decimal floating points?

srikanthccv: https://pkg.go.dev/fmt#:~:text=The%20default%20format,p%0Apointer%3A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%25p

nityanandagohain: Yeah, if it's an actual floating point then it's not doing anything else, but if I use %f and the number of zeros need to be specified and if there are less zeros in the actual number then extras will be added.

srikanthccv: We can use https://pkg.go.dev/strconv#FormatFloat?

nityanandagohain: what value of `fmt` are you suggesting to pass in the above ?

srikanthccv: (value, `f`, -1, 64).

nityanandagohain: interesting, thanks.

nityanandagohain: Made the change

srikanthccv (approved):

Issue #6188, opened 2024-10-14
Custom options for quick filters


srikanthccv:
- Should it be org and/or team and/or personal level configuration?
- Can a fixed set of custom fields be shown across the signal i.e the same set of custom options in logs, metrics, or anything in the product without repeating?

makeavish:
Allowing users to have add custom quick filters is planned for v1 of traces explorer quick filters.
cc: @GeekBoySupreme

makeavish:
Has been asked by many customers previously.

Recently this ICP/paying user asked as well: https://app.intercom.com/a/inbox/wmkurvb7/inbox/shared/all/conversation/5687#part_id=initial-part-5687-2510794044

nityanandagohain:
We will need a proper PRD on how to store these user /org prefernces if it's not there.

eKuG:
[PRD](https://www.notion.so/signoz/PRD-Custom-Options-for-Quick-Filters-1d2fcc6bcd198012aa45cf4642258268?pvs=4)

PR #7660, opened 2025-04-17
feat: added support for custom quick filters

eKuG:
### Summary

Added support for custom quick filters
The Quick filters would be supported universally. Its on Org Level. The quick filters applied on the Traces page should persist on the logs page as well.

#### Related Issues / PR's

[Related Issue](https://github.com/SigNoz/signoz/issues/6188)

#### Sample PUT and Get Curls

PUT Request:
```
curl --location --request PUT 'http://localhost:8080/api/v1/org/preferences/TRACES_QUICK_FILTERS_MAPPING' \
--header 'Authorization: <Token>' \
--data '{
    "preference_value": "[{\"key\":\"service.name\",\"datatype\":\"string\",\"type\":\"resource\"},{\"key\":\"trace_id\",\"datatype\":\"string\",\"type\":\"tag\"}]"
}'
```

GET Request:
```
curl --location 'http://localhost:8080/api/v1/org/preferences/TRACES_QUICK_FILTERS_MAPPING' \
--header 'Authorization: <Token>'
```

Get Request Response:
```
{
    "status": "success",
    "data": {
        "preference_id": "TRACES_QUICK_FILTERS_MAPPING",
        "preference_value": "[{\"key\":\"service.name\",\"datatype\":\"string\",\"type\":\"resource\"},{\"key\":\"trace_id\",\"datatype\":\"string\",\"type\":\"tag\"}]"
    }
}
```

----

> [!IMPORTANT]
> Added support for custom quick filters at the organization level, with persistence across Traces and Logs pages.
> 
>   - **Behavior**:
>     - Added support for custom quick filters at the organization level, persisting across Traces and Logs pages.
>     - Introduced `TRACES_QUICK_FILTERS_MAPPING` and `LOGS_QUICK_FILTERS_MAPPING` preferences in `preference.go`.
>   - **Validation**:
>     - Added `PreferenceValueTypeAttributeKeys` for JSON validation in `value.go`.
>     - Implemented validation logic for attribute keys in `IsValidValue()` in `preference.go`.
>   - **Misc**:
>     - Updated `SanitizeValue()` in `preference.go` to handle JSON type.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for b6445bc0ad5d6eb6bc99be7e53c06b1a00629388. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


pkg/types/preferencetypes/value.go
```diff
@@ -21,3 +22,9 @@ type PreferenceWithValue struct {
 	Preference
 	Value interface{} `json:"value"`
 }
+
+type PreferenceKeyDefinition struct {
```
nityanandagohain, line 26: use v3.AttributeKey instead

pkg/types/preferencetypes/preference.go
```diff
@@ -222,6 +232,27 @@ func (p *Preference) IsValidValue(preferenceValue interface{}) error {
 				return errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, fmt.Sprintf("the preference value is not in the range specified, min: %v , max:%v", p.Range.Min, p.Range.Max))
 			}
 		}
+	case PreferenceValueTypeJSON:
```
nityanandagohain, line 235: it should be something like preferenceValuteTypeAttributeKeys instead of generic json.

pkg/types/preferencetypes/preference.go
```diff
@@ -222,6 +232,27 @@ func (p *Preference) IsValidValue(preferenceValue interface{}) error {
...
+		var parsed []PreferenceKeyDefinition
+		if err := json.Unmarshal([]byte(strVal), &parsed); err != nil {
+			return errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "invalid JSON format for structured preference: %v", err)
+		}
+
+		for _, def := range parsed {
+			if def.Key == "" || def.DataType == "" || def.Type == "" {
+				return errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "each item must have non-empty key, datatype, and type")
+			}
+			if !isValidDataType(def.DataType) {
```
nityanandagohain, line 249: Validate function already exists for v3.AttributeKey

nityanandagohain (changes requested):

nityanandagohain:
Also I would suggestion to use someting else intead of `QUICK_FILTERS_MAPPING` , the reason is there should be an way to differentiate between quickfitlers for traces/logs/infra monitoring etc.

Also the validation should be based on the the key as well in this case.

eKuG:
> Also I would suggestion to use someting else intead of `QUICK_FILTERS_MAPPING` , the reason is there should be an way to differentiate between quickfitlers for traces/logs/infra monitoring etc.
> 
> Also the validation should be based on the the key as well in this case.

According to me, these filters should be persisted across the signals due to which I have made it "QUICK_FILTERS_MAPPING" @nityanandagohain

nityanandagohain:
Attributes across signals are different. Check the current quick filters for traces and logs as well.

eKuG:
> Attributes across signals are different. Check the current quick filters for traces and logs as well.

Valid point. Fair enough. I can change that

vikrantgupta25 (changes requested):
It looks like we will end up introducing a lot of custom validations in the preferences. Today it started with quick filter attribute validations again it will end up opening the flood gates for everything. Quick filters will have a lot of future implications in terms of materialised columns, facet creations which won't scale with it's presence inside preferences. 

Quick filters is an umbrella domain entity over the aggregate attributes root entity, which contains a lifecycle of it's own. Hence it makes sense for quick filters to contain its lifecycle in itself rather than relying on the preferences. Another downside of preferences is that anytime you update even a single filter you will end up validating the whole bunch of filters each time. 

I suggest we create a table for quick filters and let it run its own course. This will give you a lot of freedom for custom validations, api client validations and ease of maintaining the lifecycle of the same. Maybe later you wanna put validation for a key that is present in telemetry store etc as well.

eKuG:
> I suggest we create a table for quick filters and let it run its own course.

@nityanandagohain what is your opinion? If we are planning for a new table **which I initially proposed**, we need a proper timeline for the same, which I dont have the bandwidth for currently. Let me know how you feel.

PR #7673, opened 2025-04-19
feat: show notification for alert rule ID logic change

YounixM:
Notification:

```We’ve updated alert rule IDs from integers to UUIDs. Notifications fired before April 23, 2025, will continue to work seamlessly, and future notifications will use the new UUID-based rule IDs.```

### DIFF
--- a/frontend/src/container/Home/Home.tsx
+++ b/frontend/src/container/Home/Home.tsx
@@ -647,7 +647,7 @@ export default function Home(): JSX.Element {
 					<div className="home-notifications-container">
 						<div className="notification">
 							<Alert
-								message="We’ve updated alert rule IDs from integers to UUIDs. Notifications fired before April 23, 2025, will continue to work seamlessly, and future notifications will use the new UUID-based rule IDs."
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by vikrantgupta25 — DISMISSED** [review]

File: frontend/src/container/Home/Home.tsx [hunk 1]
```diff
@@ -644,6 +644,16 @@ export default function Home(): JSX.Element {
 				</div>
 
 				<div className="home-right-content">
+					<div className="home-notifications-container">
+						<div className="notification">
+							<Alert
+								message="We’ve updated alert rule IDs from integers to UUIDs. Notifications fired before April 23, 2025, will continue to work seamlessly, and future notifications will use the new UUID-based rule IDs."
```
srikanthccv, line 650:
```suggestion
								message="We're transitioning alert rule IDs from integers to UUIDs on April 23, 2025. Both old and new alert links will continue to work after this change - existing notifications using integer IDs will remain functional while new alerts will use the UUID format."
```

bit of wording change

vikrantgupta25 (approved):

Issue #7286, opened 2025-03-12
Trace detail page should render right number of spans


shivanshuraj1333:
The trace detail page should accurately render the correct number of spans, regardless of duplicate entries. Otherwise, it creates confusion by giving the impression that some spans might be missing.

OSS user reported it here https://signoz-community.slack.com/archives/C01HWQ1R0BC/p1740940431302109

Reported internally too.

A bigger problem is, how to take care of duplication of spans, but there's no easy way to control that at ingestion, it can happen at the SDK level, at the collector on the user side, or at the consumer collector while processing the batch. That issue is captured here https://github.com/SigNoz/signoz/issues/7281

Issue #7281, opened 2025-03-12
Traces[Pipeline]: Duplication of spans in Clickhouse DB


eKuG:
## Bug description

Duplication of spans in Clickhouse DB

<img width="1212" alt="Image" src="https://github.com/user-attachments/assets/5cfb83f0-03a7-4e84-94a0-1f4fe18b8d9c" />

The span with `span_id = b858a4c8def6d879` has 3 duplicated spans associated to it. 

## Expected behavior

1 span_id should corelate to 1 span. 

[...]

PR #7711, opened 2025-04-23
use count instead of count distinct

shivanshuraj1333:


pkg/query-service/app/integrations/thirdPartyApi/translator.go
```diff
@@ -61,7 +61,7 @@ func BuildDomainList(thirdPartyApis *ThirdPartyApis) (*v3.QueryRangeParamsV3, er
 		Legend:            "endpoints",
 		DataSource:        v3.DataSourceTraces,
 		StepInterval:      defaultStepInterval,
-		AggregateOperator: v3.AggregateOperatorCountDistinct,
+		AggregateOperator: v3.AggregateOperatorCount,
 		AggregateAttribute: v3.AttributeKey{
 			Key:      "http.url",
 			DataType: v3.AttributeKeyDataTypeString,
```
srikanthccv, line 67: why was this changed?

shivanshuraj1333: I'm not sure if this runs on single thread or multiple threads, we don't need to change this if so why?

srikanthccv: Correctness is a non-negotiable requirement. The existing aggregation is countDistinct(http.url), which you're changing to count. These are not equivalent operations. However, countDistinct(span_id) would be the same as count because each row in the table already represents a unique span.

srikanthccv: > that the span ids are distinct, but there could be cases where the spans are not distinct, right?

When does this happen?

shivanshuraj1333: > When does this happen?

I encountered this here https://github.com/SigNoz/signoz/issues/7286, more detailed issue is here https://github.com/SigNoz/signoz/issues/7281

srikanthccv: Yes, as you can see this is not specific to your current feature implementation but rather points to a broader issue that needs to be addressed separately.

shivanshuraj1333: yes, can you re-review this PR?

srikanthccv:
What makes a http span to be candidate for API Monitoring? does the existence of `http.url` alone makes it a candidate span or should there be span kind filter?

shivanshuraj1333:
We could add `kind_string = client` as an additional check, by while going through the example data, I found out that http.url is only added in those cases.

Can there be cases when span kind is something else but still http.url is present in the instrumentation?

PR #7747, opened 2025-04-28
fix(organization): return display name instead of name for organization

vikrantgupta25:
### Summary

- return display name instead of name for organization


grandwizard28 (approved):

grandwizard28:
Part of https://github.com/SigNoz/platform-pod/issues/704

PR #7768, opened 2025-04-29
fix(sentry): initialize sentry only once

YounixM:
- initialize sentry only once


vikrantgupta25:
contributes to - https://github.com/SigNoz/platform-pod/issues/742

PR #7805, opened 2025-05-01
feat(cache): multi-tenant cache 

vikrantgupta25:
### Summary

- add support for multi-tenant cache 
- deprecate the old cache module and use the new cache module in the product 
- update the query_range_cache to use the new cache


vikrantgupta25:
contributes to - https://github.com/SigNoz/platform-pod/issues/506

pkg/cache/cache.go
```diff
@@ -6,9 +6,11 @@ import (
 	"fmt"
 	"reflect"
 	"time"
+
+	v3 "github.com/SigNoz/signoz/pkg/query-service/model/v3"
+	"github.com/SigNoz/signoz/pkg/valuer"
 )
 
-// cacheable entity
 type CacheableEntity interface {
```
grandwizard28, line 14:
```suggestion
type Cacheable interface {
```

pkg/cache/cache.go
```diff
@@ -59,13 +60,19 @@ func (s RetrieveStatus) String() string {
 	}
 }
 
-// cache interface
```
grandwizard28, line 62: The above structs should be moved to the `types` package under `cachetypes`. And we should get rid of RetreiveStatus. It's just some fancy code which was copy pasted from somewhere which doesn't make sense.

pkg/cache/memorycache/provider.go
```diff
@@ -23,25 +25,18 @@ func New(ctx context.Context, settings factory.ProviderSettings, config cache.Co
...
-}
-
-// Store stores the data in the cache
-func (c *provider) Store(_ context.Context, cacheKey string, data cache.CacheableEntity, ttl time.Duration) error {
+func (c *provider) Set(_ context.Context, orgID valuer.UUID, cacheKey string, data cache.CacheableEntity, ttl time.Duration) error {
 	// check if the data being passed is a pointer and is not nil
 	rv := reflect.ValueOf(data)
 	if rv.Kind() != reflect.Pointer || rv.IsNil() {
 		return cache.WrapCacheableEntityErrors(reflect.TypeOf(data), "inmemory")
 	}
```
grandwizard28, line 33: This can also be moved as a helper function in the types package. It is unecessarily repeated across all functions.

pkg/cache/cache.go
grandwizard28: Can we also think of adding a function `Key(orgID valuer.UUID)` to this interface? The thought process is to somehow let the cacheables decide what key to want to have whilst ensuring that orgID is always present.

pkg/cache/memorycache/provider.go
```diff
@@ -23,25 +25,18 @@ func New(ctx context.Context, settings factory.ProviderSettings, config cache.Co
...
 }
 
-// Connect does nothing
-func (c *provider) Connect(_ context.Context) error {
-	return nil
-}
-
-// Store stores the data in the cache
-func (c *provider) Store(_ context.Context, cacheKey string, data cache.CacheableEntity, ttl time.Duration) error {
+func (c *provider) Set(_ context.Context, orgID valuer.UUID, cacheKey string, data cache.CacheableEntity, ttl time.Duration) error {
```
grandwizard28, line 28: On line 25, the new config does not support camel casing. (Technically it does) We want to semantically keep it all lowercase with `underscores` only. Suggestions:

```
flush_interval:
cleanup_interval:
```

grandwizard28 (changes requested):

pkg/cache/cache.go
vikrantgupta25: but this gives freedom to each entity to decide to put orgID or not which is dangerous

pkg/cache/memorycache/provider.go
vikrantgupta25: done

pkg/cache/memorycache/provider.go
vikrantgupta25: done

pkg/cache/cache.go
vikrantgupta25: done

pkg/types/cachetypes/cacheable.go
```diff
@@ -0,0 +1,33 @@
...
+	return errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "%s: (nil \"%s\")", caller, rt.String())
+}
+
+func ValidatePointer(dest any) error {
+	rv := reflect.ValueOf(dest)
+	if rv.Kind() != reflect.Pointer || rv.IsNil() {
+		return WrapCacheableErrors(reflect.TypeOf(dest), "inmemory")
+	}
+	return nil
+}
```
grandwizard28, line 33:
```suggestion
func ValidatePointer(dest any, caller string) error {
	rv := reflect.ValueOf(dest)
	if rv.Kind() != reflect.Pointer || rv.IsNil() {
		return WrapCacheableErrors(reflect.TypeOf(dest), caller)
	}
	return nil
}
```

pkg/query-service/querycache/query_range_cache.go
```diff
@@ -27,6 +30,17 @@ type CachedSeriesData struct {
 	Data  []*v3.Series `json:"data"`
 }
 
+type CacheableSeriesData struct {
```
grandwizard28, line 33: Please add

`var _ cachetypes.Cacheable = (*CacheableSeriesData)(nil)`

grandwizard28 (changes requested):

grandwizard28:
For testing cache, there needs to be an explicit `cachetest` package. It can directly embed the memorycache but the package needs to be visible and different @vikrantgupta25

pkg/query-service/querycache/query_range_cache.go
vikrantgupta25: done

pkg/types/cachetypes/cacheable.go
vikrantgupta25: done

grandwizard28 (approved):

PR #7836, opened 2025-05-06
fix(ruler): make the rule history work for the older integer rule id's 

vikrantgupta25:
### Summary

- make the rule history work for the older rule integer id's as well


vikrantgupta25:
contributes to - https://github.com/SigNoz/platform-pod/issues/773

vikrantgupta25:
closed due to - https://github.com/SigNoz/platform-pod/issues/773#issuecomment-2853928504

PR #7863, opened 2025-05-08
fix: changed the keys in the default quick filters to actual keys in …

eKuG:
### Summary

changed the keys in the default quick filters to actual keys in v3.AttributeKeys

----

> [!IMPORTANT]
> Update quick filter keys to match `v3.AttributeKeys` and add SQL migration to recreate filters with updated keys.
> 
>   - **Behavior**:
>     - Updates keys in `NewDefaultQuickFilter` in `filter.go` to match `v3.AttributeKeys`.
>     - Adds `NewUpdateQuickFiltersFactory` in `provider.go` to handle quick filter updates.
>     - New SQL migration `031_update_quick_filters.go` deletes existing quick filters and recreates them with updated keys for each organization.
>   - **Migration**:
>     - `Up` function in `031_update_quick_filters.go` deletes all quick filters and recreates them using `NewDefaultQuickFilter`.
>     - `Down` function in `031_update_quick_filters.go` is a no-op.
>   - **Misc**:
>     - Adds `NewUpdateQuickFiltersFactory` to `NewSQLMigrationProviderFactories` in `provider.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for d421e40f0bb11d052f151b0a5debc715713510b9. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


grandwizard28:
@eKuG you also need to think about those users for whom the defaults have already been created.

eKuG:
> @eKuG you also need to think about those users for whom the defaults have already been created.

Yes I know, I'm still thinking whether it will affect them or not? Can we handle it or not? Otherwise I'll add an update migration script. @grandwizard28

eKuG:
@grandwizard28 you can review now

grandwizard28 (dismissed):

grandwizard28:
Approved for types @vikrantgupta25

pkg/sqlmigration/031_update_quick_filters.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return err
+	}
+
+	defer tx.Rollback()
+
+	// Delete all existing quick filters
+	_, err = tx.NewDelete().
+		Table("quick_filter").
+		Where("1=1"). // Delete all rows
```
vikrantgupta25, line 50: why is this where clause required ?

pkg/sqlmigration/031_update_quick_filters.go
```diff
@@ -0,0 +1,100 @@
...
+			return err
+		}
+
+		// Insert all filters for this organization
+		_, err = tx.NewInsert().
+			Model(&storableQuickFilters).
+			Exec(ctx)
+
+		if err != nil {
+			if errors.Ast(migration.store.WrapAlreadyExistsErrf(err, errors.CodeAlreadyExists, "Quick Filter already exists"), errors.TypeAlreadyExists) {
```
vikrantgupta25, line 87: why is this check required when already we are dropping all the data before the migration

pkg/sqlmigration/031_update_quick_filters.go
```diff
@@ -0,0 +1,100 @@
...
+				// Skip if filters already exist for this org
+				continue
+			}
+			return err
+		}
+	}
+
+	return tx.Commit()
```
vikrantgupta25, line 95: follow the same pattern as in line 65. you can use both of them but decide on one and then use it everywhere

vikrantgupta25 (changes requested):

pkg/sqlmigration/031_update_quick_filters.go
eKuG: Ack

pkg/sqlmigration/031_update_quick_filters.go
eKuG: Its precautionary

pkg/sqlmigration/031_update_quick_filters.go
eKuG: "bun: Update and Delete queries require at least one Where"

^Thats why we have used "Where"

grandwizard28 (approved):

PR #7901, opened 2025-05-12
chore(frontend): update the api folder structure

vikrantgupta25:
### Summary

- the idea behind the new folder structure is that it closely resembles the `REST` structure
- this helps in increasing the readability of the code, auto structuring the code into modules as the API endpoints dictate the same 
- exact replication of module structure from backend to frontend without any extra efforts


vikrantgupta25:
contributes to - https://github.com/SigNoz/platform-pod/issues/769

frontend/src/api/v1/user/preferences/get.ts
```diff
@@ -8,7 +8,7 @@ const getPreference = async (): Promise<
 	SuccessResponse<PayloadProps> | ErrorResponse
 > => {
 	try {
-		const response = await axios.get(`/userPreferences`);
+		const response = await axios.get(`/user/preferences`);
```
SagarRajput-7, line 11: expected?

SagarRajput-7 (approved):

frontend/src/api/v1/user/preferences/get.ts
vikrantgupta25: yes

PR #7952, opened 2025-05-15
chore(sqlmigration): make the timestamp column non nullable

vikrantgupta25:
contributes to - https://github.com/SigNoz/platform-pod/issues/731


vikrantgupta25:
closing this as not picking this task, it will be a huge effort task for a very low impact. will do it later when we move to postgres

PR #7972, opened 2025-05-18
chore: add aggregation expr rewriter and exhaustive tests for logs filter

srikanthccv:
## 📄 Summary

1. Agg expr rewrite
   - We accept bare expression for the aggregation, such as `sum(total_orders)`. This expression needs to be transformed to use the right columns on the table. The changes in `pkg/querybuilder/agg_rewrite.go` implements this. It does so by wrapping the user given expression with SELECT clause, then parse, then visit the AST, then rewrite the arg with right expression.

2. Changes to grammar
    - Searching on simple numbers without quotes breaks, this is because lexer prepares a NUMBER token which parser doesn't recognize as primary, while we could ask them to add quotes, it would be better to support values also as primary.
    - Allow to support signed numbers
    - Fix issues with brackets in the key of the filter
3. Condition builder
    - Change from `IN` to `= + OR` to trigger the index if possible
    - Change from `NOT IN` to `!= + AND` to trigger the index if possible
    - Add `EXISTS` for non-negative predicates
    - Fix JSON paths to account for `-` in the key
    - Logs specific fixes such as body searches are always `LOWER(body) LIKE LOWER()`
4. Error Listener
    - Improve the error listener with as much context as possible
 5. Where clause visitor
     - Address some unhandled cases

TODO: Indexs filter for json and resources table filter

----

> [!IMPORTANT]
> Enhances query building with aggregation expression rewriting, grammar updates, optimized condition building, and comprehensive testing for logs and telemetry data.
> 
>   - **Aggregation Expression Rewriting**:
>     - Implements `agg_rewrite.go` to transform aggregation expressions like `sum(total_orders)` to use correct table columns.
>     - Introduces `AggExprRewriter` with methods `Rewrite` and `RewriteMultiple` for expression parsing and rewriting.
>   - **Grammar Enhancements**:
>     - Updates `FilterQuery.g4` to support signed numbers and treat values as primary expressions.
>     - Fixes bracket handling in filter keys.
>   - **Condition Building**:
>     - Changes `IN` to `= + OR` and `NOT IN` to `!= + AND` for index optimization in `condition_builder.go`.
>     - Adds `EXISTS` handling for non-negative predicates.
>     - Adjusts JSON path handling for keys with `-`.
>   - **Error Handling**:
>     - Enhances error listener to provide detailed context in `filter_error_listener.go`.
>   - **Testing**:
>     - Adds comprehensive tests in `agg_rewrite_test.go`, `filter_expr_logs_test.go`, and `filter_expr_logs_body_json_test.go` to validate new functionalities.
>   - **Miscellaneous**:
>     - Introduces `AddDefaultExistsFilter` in `builder_elements.go` to manage default existence checks for positive predicates.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 316a27c023b3e1e0a064c3e723b116a09217bc89. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


pkg/querybuilder/filter_error_listener.go
```diff
@@ -0,0 +1,235 @@
...
+
+// textSlice pulls raw input text between two token indexes.
+func textSlice(ts antlr.TokenStream, start, stop int) string {
+	var b strings.Builder
+	for i := start; i <= stop && i >= 0; i++ {
+		if tok := ts.Get(i); tok != nil && tok.GetTokenType() != antlr.TokenEOF {
+			b.WriteString(tok.GetText())
+		}
+	}
+	return b.String()
```
eKuG, line 234: By adding && i < size to the loop condition, the function would prevent potential index out-of-bounds issues when the stop parameter exceeds the token stream's length.

```
size := ts.Size()  // Get the stream size
    for i := start; i <= stop && i >= 0 && i < size; i++ {
```

srikanthccv: That should not happen, the stop wouldn't exceed the stream length.

nityanandagohain (approved):
Overall looks good to me!

grandwizard28 (approved):

PR #7988, opened 2025-05-20
chore: add statement builder for traces

srikanthccv:


pkg/types/telemetrytypes/testutil/stubs/metadata_store_stub.go
```diff
@@ -0,0 +1,251 @@
+package telemetrytypes
```
grandwizard28, line 1: Can you move this to `telemetrytypestest` package.

We are keeping all mocks/stubs/interface implementation for unit testing in a package suffixed with `test`.

See:
- `pkg/types/alertmanagertypes/alertmanagertypestest`
- `pkg/sqlstore/sqlstoretest`
- `pkg/sqlmigration/sqlmigrationtest`
- `pkg/instrumentation/instrumentationtest`

grandwizard28 (changes requested):

PR #7959, opened 2025-05-16
chore(api-key): refactor API key 

nityanandagohain:
- refactor the api key's according to the new schema and add the same to user module 

Contributes to - https://github.com/SigNoz/platform-pod/issues/702

PR #8000, opened 2025-05-21
fix(api-key): use role from the api key not from the user

vikrantgupta25:


vikrantgupta25:
closing in favor of  - https://github.com/SigNoz/signoz/pull/7959

PR #8013, opened 2025-05-22
fix: allow non expireable API key

nityanandagohain:
Allow API keys where expires_at is 0, which means it never expires.


SagarRajput-7 (approved):

grandwizard28 (approved):

vikrantgupta25:
contributes to - https://github.com/SigNoz/platform-pod/issues/702

PR #8020, opened 2025-05-23
chore: add traces statement builder base

srikanthccv:
## 📄 Summary

Adds the base statement builder that implements all the elements of query builder.

What is yet to added

1. exists condition for group by (I am re-collecting all the context so we can take a final call)
2. exists condition for aggregate attribute (this is detail of aggregate expression rewrite logic, independent)
3. trace querying ( I will let @eKuG implement this)

### DIFF
--- a/pkg/querybuilder/resourcefilter/statement_builder.go
+++ b/pkg/querybuilder/resourcefilter/statement_builder.go
@@ -1,0 +1,133 @@
+package resourcefilter
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/SigNoz/signoz/pkg/errors"
+	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
+	"github.com/SigNoz/signoz/pkg/types/telemetrytypes"
+	"github.com/huandu/go-sqlbuilder"
+)
+
+type ResourceFilterStatementBuilderOpts struct {
+	FieldMapper      qbtypes.FieldMapper
+	ConditionBuilder qbtypes.ConditionBuilder
+	Compiler         qbtypes.FilterCompiler
+}
+
+var (
+	ErrUnsupportedSignal = errors.NewInvalidInputf(errors.CodeInvalidInput, "unsupported signal type")
+)
+
+// Configuration for different signal types
+type signalConfig struct {
+	dbName    string
+	tableName string
+}
+
+var signalConfigs = map[telemetrytypes.Signal]signalConfig{
+	telemetrytypes.SignalTraces: {
+		dbName:    TracesDBName,
+		tableName: TraceResourceV3TableName,
+	},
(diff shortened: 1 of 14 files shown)


pkg/querybuilder/resourcefilter/statement_builder.go
```diff
@@ -0,0 +1,133 @@
...
+	b.addTimeFilter(sb, start, end)
+	return nil
+}
+
+// addTimeFilter adds time-based filtering conditions
+func (b *resourceFilterStatementBuilder[T]) addTimeFilter(sb *sqlbuilder.SelectBuilder, start, end uint64) {
+	// Convert nanoseconds to seconds and adjust start bucket
+	const (
+		nsToSeconds      = 1000000000
+		bucketAdjustment = 1800 // 30 minutes
```
nityanandagohain, line 123: nit: move this to the options of the builder ?

pkg/querybuilder/resourcefilter/tables.go
```diff
@@ -0,0 +1,8 @@
+package resourcefilter
+
+const (
+	TracesDBName             = "signoz_traces"
+	TraceResourceV3TableName = "distributed_traces_v3_resource"
+	LogsDBName               = "signoz_logs"
```
nityanandagohain, line 6: nit: combine the db names to a single source if possible ?

pkg/querybuilder/time.go
```diff
@@ -0,0 +1,18 @@
+package querybuilder
+
+import "math"
+
+// ToNanoSecs takes epoch and returns it in ns
+func ToNanoSecs(epoch uint64) uint64 {
```
nityanandagohain, line 6: lets add a basic test file atleast for happy flow ?

nityanandagohain (commented):
minor comments

pkg/querybuilder/resourcefilter/statement_builder.go
srikanthccv: I don't see a good reason to do it because it's not something that changes, definitely not without changing the way we do things. If the argument in making it an option is that it can be changed easily later, then it's the easier part to update if we change the way we bucket.

nityanandagohain: Yeah makes sense, let's keep it this way.

pkg/querybuilder/resourcefilter/tables.go
srikanthccv: Probably, Where do suggest we put them.

pkg/querybuilder/time.go
nityanandagohain: I think a test file will be better with just the basic ones, i.e seconds, milliseconds, microseconds, and ns to ns.

nityanandagohain (approved):

PR #8028, opened 2025-05-24
chore: add querier base implementation

srikanthccv:
## 📄 Summary

- refactored to remove the opts and pass required dependency (even if not all required every time) as args
- changed the response types to pointers
- removed filter compiler because it's not adding anything
- add the implementation `builderQuery` and `chSQLQuery` for reading the query result
- added the `querier` which functions as sort orchestrator and abstract away the query details of query range request.
- added logic in `consume.go` to read from the clickhouse based on the RequestType
- handled the mixed types in group by and aggregation i.e if someone aggregates on field `X` which has float64 and string, we try to convert using `toFloat64OrNull`.
- address some gaps like exists for "intrinsic" fields aka fields that are natural to signal from OTLP message

----

> [!IMPORTANT]
> Adds a base implementation for a querier with support for various query types and refactors existing query handling logic.
> 
>   - **Behavior**:
>     - Introduces `querier` as a sort orchestrator to abstract query details in `querier.go`.
>     - Implements `builderQuery` and `chSQLQuery` for query execution in `builder_query.go` and `clickhouse_query.go`.
>     - Adds logic to handle mixed types in group by and aggregation in `consume.go`.
>     - Supports reading from ClickHouse based on `RequestType` in `consume.go`.
>   - **Refactoring**:
>     - Removes `filter compiler` as it was deemed unnecessary.
>     - Refactors to pass dependencies as arguments instead of using `opts`.
>   - **Models**:
>     - Adds models for `TimeSeriesData`, `ScalarData`, and `RawData` in `resp.go`.
>     - Updates `Result` and `ExecStats` structures in `query.go`.
>   - **Misc**:
>     - Changes response types to pointers for efficiency.
>     - Addresses gaps for intrinsic fields from OTLP messages in `condition_builder.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 92b4bb6fdd54e12ea134b0e8799378576937f864. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


grandwizard28 (dismissed):

pkg/querier/querier.go
```diff
@@ -0,0 +1,96 @@
...
+
+			case qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]:
+				bq := newBuilderQuery(q.telemetryStore, q.logStmtBuilder, spec, qbtypes.TimeRange{From: req.Start, To: req.End}, req.RequestType)
+				queries[query.Name] = bq
+
+			case qbtypes.QueryBuilderQuery[qbtypes.MetricAggregation]:
+				bq := newBuilderQuery(q.telemetryStore, q.metricStmtBuilder, spec, qbtypes.TimeRange{From: req.Start, To: req.End}, req.RequestType)
+				queries[query.Name] = bq
+			default:
+				return nil, errors.WrapInvalidInputf(nil, errors.CodeInvalidInput, "unsupported builder spec type %T", query.Spec)
```
grandwizard28, line 75:
```suggestion
				return nil, errors.NewInvalidInputf(nil, errors.CodeInvalidInput, "unsupported builder spec type %T", query.Spec)
```

grandwizard28 (changes requested):

PR #8086, opened 2025-05-28
fix(saml): do not fetch the claims and use orgID from domain

vikrantgupta25:


vikrantgupta25:
contributes to - https://github.com/SigNoz/platform-pod/issues/797

PR #8109, opened 2025-05-31
fix: handle alert list updates

YounixM:
## 📄 Summary

Adding a new alert channel didn't update the alert list.
We are maintaining a local state of alertChannels that conflicts with the data from API when there is a new addition or a deletion.

---

## ✅ Changes

Remove duplicated state and render alerts using the API data.
Invalidate query and refetch on delete

This would maintain a single source of truth for alert channels.

## 🧪 How to Test

1.  Add a new channel
2. Delete  a channel
3. List should be updated correctly

### DIFF
--- a/frontend/src/container/AllAlertChannels/Delete.tsx
+++ b/frontend/src/container/AllAlertChannels/Delete.tsx
@@ -1,14 +1,15 @@
 import { Button } from 'antd';
 import { NotificationInstance } from 'antd/es/notification/interface';
 import deleteChannel from 'api/channels/delete';
-import { Dispatch, SetStateAction, useState } from 'react';
+import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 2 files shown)


frontend/src/container/AllAlertChannels/Delete.tsx
```diff
@@ -21,7 +22,8 @@ function Delete({ notifications, setChannels, id }: DeleteProps): JSX.Element {
 				message: 'Success',
 				description: t('channel_delete_success'),
 			});
-			setChannels((preChannels) => preChannels.filter((e) => e.id !== id));
+			// Invalidate and refetch
+			queryClient.invalidateQueries(['getChannels']);
```
vikrantgupta25, line 26: we need to refresh the alerts list on create / edit / delete all three

vikrantgupta25 (changes requested):

vikrantgupta25 (approved):

PR #8121, opened 2025-06-02
fix(license): fixes for license service

vikrantgupta25:
- send the `TERMINATED` licenses as well in the license API call 
- refresh the features when refreshing the license 
- show manage licenses for users on basic plan


vikrantgupta25:
contributes to - https://github.com/SigNoz/platform-pod/issues/632

grandwizard28 (approved):

PR #8130, opened 2025-06-03
chore: add range query impl for promql

srikanthccv:
## 📄 Summary

Add query exec querier implementation for promql query range

----

> [!IMPORTANT]
> Implements PromQL range query execution with error handling and updates error codes, types, and HTTP mappings.
> 
>   - **Behavior**:
>     - Implements `Execute()` in `promql_query.go` for PromQL range queries, handling execution and error mapping.
>     - Maps PromQL errors to custom error types, including `ErrQueryCanceled`, `ErrQueryTimeout`, and `ErrStorage`.
>     - Returns query results as `qbv5.Result` with time series data and warnings.
>   - **Errors**:
>     - Adds `CodeCanceled` and `CodeTimeout` to `code.go`.
>     - Adds `TypeCanceled` and `TypeTimeout` to `type.go`.
>   - **HTTP Handling**:
>     - Updates `Error()` in `render.go` to map `TypeCanceled` to status 499 and `TypeTimeout` to 504.
>   - **Models**:
>     - Adds `Step` and `Stats` fields to `PromQuery` in `prom_query.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 775c17589ebe77c0947e15ed13e1f5012fe37b27. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/pkg/query-service/app/metrics/v4/helpers/clauses.go
+++ b/pkg/query-service/app/metrics/v4/helpers/clauses.go
@@ -2,8 +2,9 @@ package helpers
 
 import (
 	"fmt"
-	"github.com/SigNoz/signoz/pkg/query-service/utils"
 	"strings"
+
+	"github.com/SigNoz/signoz/pkg/query-service/utils"
 
 	v3 "github.com/SigNoz/signoz/pkg/query-service/model/v3"
 )
--- a/pkg/telemetrymetrics/const.go
+++ b/pkg/telemetrymetrics/const.go
@@ -1,6 +1,7 @@
 package telemetrymetrics
 
 var IntrinsicFields = []string{
+	"__normalized",
 	"temporality",
 	"metric_name",
 	"type",
--- a/pkg/telemetrymetrics/field_mapper.go
+++ b/pkg/telemetrymetrics/field_mapper.go
@@ -3,6 +3,7 @@ package telemetrymetrics
 import (
 	"context"
 	"fmt"
+	"slices"
 
 	schema "github.com/SigNoz/signoz-otel-collector/cmd/signozschemamigrator/schema_migrator"
 	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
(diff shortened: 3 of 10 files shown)


pkg/http/render/render.go
```diff
@@ -60,6 +66,10 @@ func Error(rw http.ResponseWriter, cause error) {
 		httpCode = http.StatusNotImplemented
 	case errors.TypeForbidden:
 		httpCode = http.StatusForbidden
+	case errors.TypeCanceled:
+		httpCode = statusClientClosedConnection
+	case errors.TypeTimeout:
+		httpCode = http.StatusServiceUnavailable
```
therealpandey: 504 seems more appropriate.
503 would indicate the server itself is unavailable.

You can also think about 408 since the timeout is set in the client side header. Can the client still get a response by increasing the timeout header? If yes, then 408.
Otherwise 504.

srikanthccv: I think 503 is more appropriate

> Common causes are that a server is down for maintenance or overloaded

> In overload cases, some server-side applications will reject requests with a 503 status when resource thresholds like memory, CPU, or connection pool limits are met.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/503

I don't think 408 makes sense because server is not waiting for something from client to return request timeout with client error here.

therealpandey: Okay 504 then 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/504

503 would mean I cannot do anything on the server. You want to convey that the query timed out but you can still login.

srikanthccv: I still think 503 would be ideal (with retry-after) for this server to send than send gatewaytimeout.

grandwizard28 (approved):

pkg/http/render/render.go
srikanthccv: had a discussion on chat, without explicit `retry-after` there is no point in sending the 503 which would be interpreted as service unavailable which is a different meaning altogether. If we ever implement the `retry-after` reliably then maybe revisit. but for now 504 is more appropriate. 

also

>Note: The existence of the 503 status code does not imply that a server has to use it when becoming overloaded. Some servers might simply refuse the connection.

https://www.rfc-editor.org/rfc/rfc9110.html#name-503-service-unavailable

aniketio-ctrl (approved):

PR #8165, opened 2025-06-05
feat: added support for trace_operators in query range v5

eKuG:
## 📄 Summary

Added support for trace_operators in query range v5
Added tests for the same.
For deciding precedence we are using `expression tree`, Refer : [Exporession Tree](https://www.geeksforgeeks.org/expression-tree/)
Also, only 1 trace-operator expression would be allowed throughout the whole query builder.
Sample Trace Operator query in Query Range : 

```
QueryRangeRequest{
SchemaVersion: "v1",
Start:         1640995200000,
End:           1640998800000,
RequestType:   RequestTypeTimeSeries,
CompositeQuery: CompositeQuery{
	Queries: []QueryEnvelope{
		{
			Name: "A",
			Type: QueryTypeBuilder,
			Spec: QueryBuilderQuery[TraceAggregation]{
				Signal: telemetrytypes.SignalTraces,
				Filter: &Filter{
					Expression: "service.name = 'checkoutservice'",
				},
			},
		},
		{
			Name: "B",
			Type: QueryTypeBuilder,
			Spec: QueryBuilderQuery[TraceAggregation]{
				Signal: telemetrytypes.SignalTraces,
				Filter: &Filter{
					Expression: "hasError = true",
				},
			},
		},
		{
			Name: "T1",
			Type: QueryTypeTraceOperator,
			Spec: QueryBuilderTraceOperator{
				Name:       "trace_flow_analysis",
				Expression: "A => B",
				Filter: &Filter{
					Expression: "trace_duration > 200ms AND span_count >= 5",
				},
				OrderBy: &TraceOrderBy{
					Field:     TraceOrderByTraceDuration,
					Direction: OrderDirectionDesc,
				},
				Limit:     100,
				PageToken: "eyJsYXN0X3RyYWNlX2lkIjoiYWJjZGVmIn0=",
			},
		},
	},
},
}
```

---

## ✅ Changes

- [x] Feature: Brief description

---

## 📋 Checklist

- [x] Dev Review

---

----

> [!IMPORTANT]
> Add support for trace operators in query range v5 with parsing, validation, and tests.
> 
>   - **Feature**:
>     - Add `QueryTypeTraceOperator` to `query_type.go` for trace operator support.
>     - Implement `QueryBuilderTraceOperator` in `trace_operator.go` for parsing and validating trace operator expressions.
>     - Support precedence in expressions using an expression tree.
>     - Allow only one trace operator expression per query builder.
>   - **Validation**:
>     - Add `ValidateTraceOperator` and `ValidateUniqueTraceOperator` in `trace_operator.go` to ensure correct usage.
>     - Validate orderBy fields and pagination limits.
>   - **Tests**:
>     - Add tests in `req_test.go` for various trace operator scenarios, including valid and invalid cases.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for fa0183ff206ea0a86193258ff30116a1c4a9f9bf. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


pkg/types/querybuildertypes/querybuildertypesv5/trace_operator.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+// QueryBuilderTraceOperator represents a trace operator query
+type QueryBuilderTraceOperator struct {
+	Name     string `json:"name"`
+	Disabled bool   `json:"disabled,omitempty"`
+
+	Expression string `json:"expression"`
+
+	TraceFilters *TraceFilters `json:"traceFilters,omitempty"`
```
srikanthccv: https://google.aip.dev/160

pkg/types/querybuildertypes/querybuildertypesv5/trace_operator.go
```diff
@@ -0,0 +1,360 @@
...
+
+	SelectFields []telemetrytypes.TelemetryFieldKey `json:"selectFields,omitempty"`
+	Order        []OrderBy                          `json:"order,omitempty"`
+	Limit        int                                `json:"limit,omitempty"`
```
srikanthccv: limit alone won't be enough for pagination, there should be other info. Of all the alternatives, cursor/token based pagination is recommended and common for paginated APIs https://google.aip.dev/158

pkg/types/querybuildertypes/querybuildertypesv5/trace_operator.go
```diff
@@ -0,0 +1,360 @@
...
+func (q *QueryBuilderTraceOperator) ParseExpression() error {
+	if q.Expression == "" {
+		return errors.WrapInvalidInputf(
+			nil,
+			errors.CodeInvalidInput,
+			"expression cannot be empty",
+		)
+	}
+
+	parsed, err := parseTraceExpression(q.Expression)
```
srikanthccv, line 87: Please write some tests for `parseTraceExpression` and show how the complex expression get transformed.

pkg/types/querybuildertypes/querybuildertypesv5/trace_operator.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+// TraceFilters represents trace-level conditions
+type TraceFilters struct {
+	SpanCount     *TraceFilterCondition `json:"span_count,omitempty"`
+	TraceDuration *TraceFilterCondition `json:"trace_duration,omitempty"`
+}
+
+// QueryBuilderTraceOperator represents a trace operator query
+type QueryBuilderTraceOperator struct {
```
srikanthccv, line 32: Things that are not clear yet

- Which spans get returned in the list (and how are indicated)
- What does the aggregation request look like for the trace operator queries? What are the aggregations supported? And what is allowed in GROUP BY (as in, do we allow grouping by an attribute in an indirect child with aggregation on some parent at several levels higher)? or what is allowed and where is it provided in the payload?

eKuG: - Traces Tab:
  - Input: A => B expression
  - Returns: Complete traces that satisfy the trace operator logic
 
- List View : 
  - Input: A => B expression
  -Returns: Only spans that match Query A (the "left side" of the expression)

eKuG: - Aggregations - INHERITED
  - Trace operators inherit aggregations from the base queries (A, B, C, etc.)
  - No additional aggregations are defined at the trace operator level
  - The trace operator logic determines which traces/spans to include, then applies inherited aggregations
  
- GROUP BY - NOT SUPPORTED
  - Trace operators do not support GROUP BY clauses
  - Grouping should be done at the base query level if needed
  - This keeps trace operator logic focused purely on trace relationships

srikanthccv: > Returns: Only spans that match Query A (the "left side" of the expression)

Why is it specifically `A` and not any other query in the expression?

> Trace operators inherit aggregations from the base queries (A, B, C, etc.)

> GROUP BY - NOT SUPPORTED

>Trace operators do not support GROUP BY clauses
>Grouping should be done at the base query level if needed
>This keeps trace operator logic focused purely on trace relationships

Can you share simple excalidraw sketch how you are envisioning this feature looks like for end user and how they are going to satisfy their use-cases? I am not able to follow this

eKuG: > Why is it specifically A and not any other query in the expression?

Technically there is going to be a main query and then there are going to be some sub queries related to it. The main query would always be the left most part of the trace_operator due to which all the spans would be selected from the "left side" of this expression

eKuG: Screenshot 2025-06-10 at 3 20 18 PM

This is what I'm envisioning for the aggregation on queries. @srikanthccv

srikanthccv: We have the following operators

```
	TraceOperatorDirectDescendant   = TraceOperatorType{valuer.NewString("=>")}
	TraceOperatorIndirectDescendant = TraceOperatorType{valuer.NewString("->")}
	TraceOperatorAnd                = TraceOperatorType{valuer.NewString("&&")}
	TraceOperatorOr                 = TraceOperatorType{valuer.NewString("||")}
	TraceOperatorNot                = TraceOperatorType{valuer.NewString("NOT")}

```


Let's take a simple use-case


A: service=frontend
B: service=payment-api status=error

1. I am interested in seeing the erroring payment-api spans that are initiated from the frontend. 


This is very basic example. We can write number of use-cases where A is just a condition to satisfy but not the span we are interested in. So it would be of no use to see just A when I am interested in spans from B.


2. Regarding the aggregation.

Let's take another simple use-case that trace operators supposed to solve as it brings the whole trace context for user to work with

A: service=frontend
B: service=payment-api status=error

I am interested in the p99 of e2e latency grouped by a payment_id which is only available in B span (i.e `p99(duration_nano) group by B.payment_id`). How does that look like?

eKuG: @srikanthccv We should be able to support Point 1 after the ClickHouse upgrade, as CTEs make this type of querying possible. I’ve tested it, and it looks like we can achieve good performance, though we may need to impose some limitations.

pkg/types/querybuildertypes/querybuildertypesv5/trace_operator.go
```diff
@@ -0,0 +1,441 @@
...
+// TraceOperatorType defines the type of trace operator
+type TraceOperatorType struct{ valuer.String }
+
+var (
+	TraceOperatorDirectDescendant   = TraceOperatorType{valuer.NewString("=>")}
+	TraceOperatorIndirectDescendant = TraceOperatorType{valuer.NewString("->")}
+	TraceOperatorAnd                = TraceOperatorType{valuer.NewString("&&")}
+	TraceOperatorOr                 = TraceOperatorType{valuer.NewString("||")}
+	TraceOperatorNot                = TraceOperatorType{valuer.NewString("NOT")}
+	TraceOperatorExclude            = TraceOperatorType{valuer.NewString("NOT")}
```
srikanthccv, line 20: These are redundant

pkg/types/querybuildertypes/querybuildertypesv5/trace_operator.go
```diff
@@ -0,0 +1,441 @@
...
+var (
+	TraceOrderBySpanCount     = TraceOrderByField{valuer.NewString("span_count")}
+	TraceOrderByTraceDuration = TraceOrderByField{valuer.NewString("trace_duration")}
+)
+
+// TraceOrderBy represents ordering for trace operators
+type TraceOrderBy struct {
+	Field     TraceOrderByField `json:"field"`
+	Direction OrderDirection    `json:"direction"`
+}
```
srikanthccv: Please use the existing `OrderBy` type

srikanthccv (commented):
As discussed, please update `QueryBuilderTraceOperator` to support the aggregation at the trace level.

pkg/types/querybuildertypes/querybuildertypesv5/trace_operator.go
eKuG: No, Not can have only 1 Query such as (NOT A)
Exclude should have both Queries such as (A NOT B)

pkg/types/querybuildertypes/querybuildertypesv5/trace_operator.go
eKuG: Ack

pkg/types/querybuildertypes/querybuildertypesv5/trace_operator.go
srikanthccv: A `NOT` is negation and is used differently depending on context. `A NOT B` is `A AND (NOT B)`. Please explain how they are different.

eKuG: Yes, whenever we are traversing, we are checking if the query is complete which means
A valid query should look something like `A AND B`.
This would not be a valid query `AND A` or `OR B` in itself. But for `NOT A`, there's an exception where left query can be empty.

srikanthccv: That doesn't answer the question. I am not able to follow how are they different operators?

eKuG: A single NOT operator could technically serve both the unary form `NOT A` and the binary form `A NOT B`.

However, I introduced a dedicated Exclude operator for the latter to keep the grammar explicit and to simplify parsing and traversal.

pkg/types/querybuildertypes/querybuildertypesv5/trace_operator.go
```diff
@@ -0,0 +1,428 @@
...
+	Expression string `json:"expression"`
+
+	Filter *Filter `json:"filter,omitempty"`
+
+	// User-configurable span return strategy - which query's spans to return
+	ReturnSpansFrom string `json:"returnSpansFrom,omitempty"`
+
+	// Trace-specific ordering (only span_count and trace_duration allowed)
+	Order       []OrderBy        `json:"orderBy,omitempty"`
+	Aggregation TraceAggregation `json:"aggregations,omitempty"`
```
srikanthccv:
```suggestion
	Aggregations []TraceAggregation `json:"aggregations,omitempty"`
```

aggregations or aggregation?


And any aggregation without an interval and group by is useless, please add `stepInterval` and `groupby`. I already mentioned the use case here with explicit group by paymentId here https://github.com/SigNoz/signoz/pull/8165#discussion_r2141607077

pkg/types/querybuildertypes/querybuildertypesv5/trace_operator.go
```diff
@@ -0,0 +1,428 @@
...
+
+	// AIP-158 compliant pagination
+	Limit     int    `json:"limit,omitempty"`
+	PageToken string `json:"page_token,omitempty"`
```
srikanthccv: for consistency https://github.com/SigNoz/signoz/blob/3f128f0f1d4c7ec24a2805bfe57671051396870e/pkg/types/querybuildertypes/querybuildertypesv5/builder_query.go#L47


```suggestion
	Cursor string `json:"cursor,omitempty"`
```

pkg/types/querybuildertypes/querybuildertypesv5/trace_operator.go
srikanthccv: There is no such thing as binary NOT. A negation is always a unary https://en.wikipedia.org/wiki/Negation. I will end this discussion here to keep things moving, especially since this is not end-user facing.

pkg/types/querybuildertypes/querybuildertypesv5/trace_operator.go
eKuG: Yes makes a lot of sense. Did not account for a group by in the aggregations

pkg/query-service/constants/constants.go
```diff
@@ -50,6 +50,7 @@ const LogsTTL = "logs"
 const SpanSearchScopeRoot = "isroot"
 const SpanSearchScopeEntryPoint = "isentrypoint"
 const OrderBySpanCount = "span_count"
+const OrderByTraceDuration = "trace_duration"
```
therealpandey: Please remove this

therealpandey: No outer `query-service` import in the new types package. Redefine and clear them in the builder.

grandwizard28 (changes requested):

pkg/query-service/constants/constants.go
eKuG: Why? I need it actually

PR #8181, opened 2025-06-09
chore(sqlmigration): drop the rule history and data migrations table

vikrantgupta25:
- drop the deprecated rule history and data migrations table


pkg/query-service/app/http_handler.go
```diff
@@ -660,7 +660,13 @@ func Intersection(a, b []int) (c []int) {
 }
 
 func (aH *APIHandler) getRule(w http.ResponseWriter, r *http.Request) {
-	id := mux.Vars(r)["id"]
+	idStr := mux.Vars(r)["id"]
+	id, err := valuer.NewUUID(idStr)
+	if err != nil {
```
vikrantgupta25, line 665: since the entire handler uses the older success / error structures didn't update it to the latest structures as didn't wanna expand the scope here.

PR #8229, opened 2025-06-12
chore: added preloaded metrics metadata at first api call

aniketio-ctrl:
## 📄 Summary

1.) Removed the pre loaded cache from application startup/
2.) Added the preloaded cache in `GetUpdatedMetricsMetadata` function
3.) Added the cache key to check if we have already loaded the metadata cache.
4.) Removed the unnecessary DB check, db check only on cache errors not on key miss errors.
5.) Whenever we update the metrics it will always set the cache, so we dont need to check in case of cache miss.

---

Test Cases ->
1.) Empty update metadata sanity
2.) First cache metadata at first call sanity
3.)Updated metadata than boot up again to check sanity, that cache is loading.
4.) Multiple update on same metrics sanity

----

> [!IMPORTANT]
> Optimized metrics metadata caching by loading cache on first API call instead of startup, with updated logic in `reader.go` and added test cases.
> 
>   - **Caching Logic**:
>     - Removed preloaded cache from application startup in `server.go`.
>     - Added cache loading in `GetUpdatedMetricsMetadata` in `reader.go`.
>     - Cache key added to check if metadata cache is already loaded.
>     - Removed unnecessary DB check on cache errors, only checks on cache errors.
>     - Cache is set whenever metrics are updated, eliminating the need for cache miss checks.
>   - **Test Cases**:
>     - Added test cases for empty update metadata, first cache load, cache loading after update, and multiple updates on the same metrics.
>   - **Misc**:
>     - Removed redundant code in `helper.go` and `threshold_rule_test.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 7353b86bb0573442e8eecb43d2afbdce58aa8f31. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -6233,45 +6233,65 @@ func (r *ClickHouseReader) CheckForLabelsInMetric(ctx context.Context, metricNam
 	return hasLE, nil
 }
 
-func (r *ClickHouseReader) PreloadMetricsMetadata(ctx context.Context, orgID valuer.UUID) []error {
+func (r *ClickHouseReader) PreloadMetricsMetadata(ctx context.Context, orgID valuer.UUID) ([]string, *model.ApiError) {
```
vikrantgupta25: use the latest `error` here

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -6233,45 +6233,65 @@ func (r *ClickHouseReader) CheckForLabelsInMetric(ctx context.Context, metricNam
...
+	var errorMetricsList []string
 	// Fetch all rows from ClickHouse
 	query := fmt.Sprintf(`SELECT metric_name, type, description , temporality, is_monotonic, unit
 		FROM %s.%s;`, signozMetricDBName, signozUpdatedMetricsMetadataTable)
 	valueCtx := context.WithValue(ctx, "clickhouse_max_threads", constants.MetricsExplorerClickhouseThreads)
 	err := r.db.Select(valueCtx, &allMetricsMetadata, query)
 	if err != nil {
-		errorList = append(errorList, err)
-		return errorList
+		return nil, &model.ApiError{Typ: "ClickHouseError", Err: fmt.Errorf("error in getting updated metadata: %v", err)}
```
vikrantgupta25: use new errors package

vikrantgupta25 (dismissed):

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -6233,45 +6233,65 @@ func (r *ClickHouseReader) CheckForLabelsInMetric(ctx context.Context, metricNam
...
-	return errorList
+	return errorMetricsList, nil
 }
 
 func (r *ClickHouseReader) GetUpdatedMetricsMetadata(ctx context.Context, orgID valuer.UUID, metricNames ...string) (map[string]*model.UpdateMetricsMetadata, *model.ApiError) {
 	cachedMetadata := make(map[string]*model.UpdateMetricsMetadata)
 	var missingMetrics []string
 
+	preCacheLoaded := new(model.CacheLoaded)
+	err := r.cache.Get(ctx, orgID, constants.METRICS_UPDATED_METADATA_CACHE_LOADED_KEY, preCacheLoaded, false)
```
vikrantgupta25: why are we doing this ? On the call request the metric from cache based on the key and if the key isn't found then make the call and cache it . we do not need to store another key in cache to store boolean

pkg/query-service/model/updatedMetricsMetadata.go
```diff
@@ -22,3 +22,12 @@ func (c *UpdateMetricsMetadata) MarshalBinary() (data []byte, err error) {
 func (c *UpdateMetricsMetadata) UnmarshalBinary(data []byte) error {
 	return json.Unmarshal(data, c)
 }
+
```
vikrantgupta25: not needed

pkg/query-service/constants/constants.go
```diff
@@ -668,3 +668,4 @@ func GetDefaultSiteURL() string {
 }
 
 const DotMetricsEnabled = "DOT_METRICS_ENABLED"
+const METRICS_UPDATED_METADATA_CACHE_LOADED_KEY = "METRICS_UPDATED_METADATA_CACHE_LOADED"
```
vikrantgupta25: not needed

pkg/query-service/app/clickhouseReader/reader.go
aniketio-ctrl: To avoid unnecessary calls to ClickHouse for each metric name, we use PreloadMetricsMetadata to load all updated metrics metadata in advance. This way, we can check if a metric name exists in the preloaded metadata and only consider it "updated" if it's present—eliminating redundant queries. Since we always cache the metadata whenever it's updated, relying on the cache ensures efficient lookups without repeatedly hitting the database.

pkg/query-service/app/clickhouseReader/reader.go
aniketio-ctrl: which error struct?

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -2342,41 +2342,25 @@ func (r *ClickHouseReader) GetTotalLogs(ctx context.Context) (uint64, error) {
...
 
-	rows, err := r.db.Query(ctx, query, metricNames)
-	if err != nil {
-		return nil, err
+	// Batch fetch all metadata at once
+	metadataMap, apiErr := r.GetUpdatedMetricsMetadata(ctx, orgID, metricNames...)
+	if apiErr != nil {
+		zap.L().Warn("Failed to fetch updated metrics metadata", zap.Error(apiErr))
+		// best-effort return, not failing outright
+		return metricNameToTemporality, nil
```
srikanthccv: Why are we returning empty here instead of an error? The issue with not returning an error is that the caller receives `FetchTemporality` call was success but without data resulting in incorrect queries.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3037,67 +3021,80 @@ func (r *ClickHouseReader) QueryDashboardVars(ctx context.Context, query string)
...
+	if apiError != nil {
+		return &response, fmt.Errorf("error getting updated metrics metadata: %s", apiError.Error())
+	}
+
+	seen := make(map[string]struct{})
+	for _, name := range metricNames {
+		metadata, ok := metadataMap[name]
+		if !ok {
+			continue
 		}
```
srikanthccv: This should not happen; a metric's metadata should always exist if a metric exists.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3189,72 +3186,67 @@ func (r *ClickHouseReader) GetMetricMetadata(ctx context.Context, orgID valuer.U
...
+	var (
+		deltaExists bool
+		isMonotonic bool
+		temporality string
+		description string
+		metricType  string
+		unit        string
+	)
+
+	if metadata, exists := metadataMap[metricName]; exists {
```
srikanthccv: same question as above

pkg/query-service/app/querier/helper.go
```diff
@@ -213,17 +213,6 @@ func (q *querier) runBuilderQuery(
...
-	if builderQuery.DataSource == v3.DataSourceMetrics && !q.testingMode {
-		metadata, apiError := q.reader.GetUpdatedMetricsMetadata(ctx, orgID, builderQuery.AggregateAttribute.Key)
-		if apiError != nil {
-			zap.L().Error("Error in getting metrics cached metadata", zap.Error(apiError))
-		}
-		if updatedMetadata, exist := metadata[builderQuery.AggregateAttribute.Key]; exist {
-			builderQuery.AggregateAttribute.Type = v3.AttributeKeyType(updatedMetadata.MetricType)
-			builderQuery.Temporality = updatedMetadata.Temporality
-		}
-	}
```
srikanthccv, line 225: Where is this now moved?

pkg/query-service/rules/threshold_rule_test.go
```diff
@@ -1224,7 +1223,6 @@ func TestThresholdRuleUnitCombinations(t *testing.T) {
 
 	for idx, c := range cases {
 		rows := cmock.NewRows(cols, c.values)
-		telemetryStore.Mock().ExpectQuery(".*").WillReturnError(fmt.Errorf("error"))
```
srikanthccv, line 1227: Why is this removed?

pkg/query-service/app/querier/helper.go
aniketio-ctrl: they wont need updation now as we are updating temporality at the start of the query range

pkg/query-service/rules/threshold_rule_test.go
aniketio-ctrl: due to shift in updated metadata and fetch temporality calls, testcases were failing due to this

pkg/query-service/app/clickhouseReader/reader.go
aniketio-ctrl: yeah , but its good to have ok, or exist checks with map, removed it though

pkg/query-service/app/clickhouseReader/reader.go
aniketio-ctrl: yeah , but its good to have ok, or exist checks with map, removed it though

pkg/query-service/rules/threshold_rule_test.go
srikanthccv: I can follow that much already from the code. Please help me understand what was the query earlier that expected error but no select and error needed now.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -5918,15 +5884,57 @@ func (r *ClickHouseReader) GetUpdatedMetricsMetadata(ctx context.Context, orgID
...
+				&metadata.Description,
+				&metadata.Temporality,
+				&metadata.IsMonotonic,
+				&metadata.Unit,
+			); err != nil {
+				return cachedMetadata, &model.ApiError{Typ: "ClickhouseErr", Err: fmt.Errorf("error scanning fallback metadata: %v", err)}
+			}
+
+			cacheKey := constants.UpdatedMetricsMetadataCachePrefix + metadata.MetricName
+			if cacheErr := r.cache.Set(ctx, orgID, cacheKey, metadata, -1); cacheErr != nil {
```
srikanthccv: `-1` is invalid for redis cache.

pkg/query-service/rules/threshold_rule_test.go
aniketio-ctrl: While Prepare metrics query we call getUpdatedMetricsMetadata first, above call was to throw error from there.
Now we don't call that function because we always get updated temporality, so thats why we need to remove that line

https://github.com/SigNoz/signoz/pull/8229/files#r2159820088

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3170,18 +3166,17 @@ func (r *ClickHouseReader) GetMetricMetadata(ctx context.Context, orgID valuer.U
...
-		isMonotonic = metadata.IsMonotonic
-		description = metadata.Description
-		unit = metadata.Unit
+	metadata := metadataMap[metricName]
 
-		if temporality == string(v3.Delta) {
-			deltaExists = true
-		}
-	}
+	metricType = string(metadata.MetricType)
```
srikanthccv, line 2843: This panics if the `GetMetricMetadata` is requested for a metric that doesn't exist (rare but possible). Return Not-found error if the metric name doesn't exist.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -5885,7 +5885,7 @@ func (r *ClickHouseReader) GetUpdatedMetricsMetadata(ctx context.Context, orgID
 			}
 
 			cacheKey := constants.UpdatedMetricsMetadataCachePrefix + metadata.MetricName
-			if cacheErr := r.cache.Set(ctx, orgID, cacheKey, metadata, -1); cacheErr != nil {
+			if cacheErr := r.cache.Set(ctx, orgID, cacheKey, metadata, 0); cacheErr != nil {
```
srikanthccv, line 5891: With this change, there will be bunch of warnings with message `zero value for TTL found. defaulting to the base TTL`. However, the cache interface doesn't provide a unified way to set non-expiration independent of the provider so I would suggest create an issue for that.

srikanthccv (commented):
Overall looks good to me, minor comments.

pkg/query-service/app/clickhouseReader/reader.go
aniketio-ctrl: https://github.com/SigNoz/signoz/issues/8439

PR #8237, opened 2025-06-13
chore: disallow unknown fields and address gaps

srikanthccv:
## 📄 Summary

1. Reject requests with unknown fields (with suggestions) https://brandur.org/disallow-unknown-fields
2. Fix an issue with the cache where the bucket start and end are not adjusted to the step. One or both ends of the time range requested in the query range can be incomplete intervals (see tests bucket_cache_step_test.go for examples); they can't be cached, `Put` is updated to handle it.
3. Add shiftby
4. Add some validations for the query range request
5. Address some gaps in the statement builder implementation

----

> [!IMPORTANT]
> Enhance query handling by adding validation for unknown fields, improving caching for complete intervals, and introducing time shift functionality.
> 
>   - **Validation**:
>     - Reject requests with unknown fields, providing suggestions for corrections using `UnmarshalJSONWithContext` in `json_decoder.go`.
>     - Add `Validate()` method in `validation.go` to ensure query range requests are valid.
>   - **Caching**:
>     - Update `Put()` in `bucket_cache.go` to adjust start and end times for complete intervals.
>     - Add `bucket_cache_step_test.go` to test step alignment in caching.
>   - **Query Handling**:
>     - Add `ShiftBy` field in `builder_query.go` to handle time shifts in queries.
>     - Modify `executeWithContext()` in `builder_query.go` to adjust time ranges based on `ShiftBy`.
>   - **Error Handling**:
>     - Improve error messages for JSON unmarshaling in `json_decoder.go`.
>     - Add tests in `req_error_test.go` to verify error messages and suggestions.
>   - **Miscellaneous**:
>     - Add `FetchTemporalityMulti()` in `telemetrymetadata/metadata.go` to fetch metric temporality.
>     - Update `req.go` to handle new query types and improve error handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for ac85a39aa4c2798721425d549769a54742293fe4. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


pkg/types/querybuildertypes/querybuildertypesv5/json_decoder.go
```diff
@@ -0,0 +1,163 @@
...
+	dec.DisallowUnknownFields()
+
+	err := dec.Decode(target)
+	if err == nil {
+		// No error, successful unmarshal
+		return nil
+	}
+
+	// Check if it's an unknown field error
+	if strings.Contains(err.Error(), "unknown field") {
```
therealpandey, line 32:
```suggestion
	if strings.Contains(err.Error(), "unknown field") {
		return err
	}
```
This usecase is ripe for defining a custom error code.

You can define a ErrCodeQuerierUnknownField = errors.MustNewCode("querier_unknown_field") and then call errors.Asc().

It's not very maintainable when you call string.Contains.

grandwizard28 (changes requested):

therealpandey:
Also, @srikanthccv, you can think about defining custom error codes for the querier.

```
var (
ErrCodeUnknownField
ErrCodeMalformedQuery
ErrCodeQueryBuilder.....
)
```

Please read https://github.com/SigNoz/signoz/blob/main/docs/contributing/go/errors.md#code

pkg/types/querybuildertypes/querybuildertypesv5/json_decoder.go
srikanthccv: I wouldn't have used `string.Contains` if I had an option. This is an error that comes from the JSON library. There exists an error sentinel for this in experimental v2 json, but not in current json from stdlib, so I had to go with `strings.Contains(err.Error(), "unknown field")`

srikanthccv: https://github.com/golang/go/blob/master/src/encoding/json/v2/errors.go#L39


https://github.com/golang/go/blob/master/src/encoding/json/decode.go#L737

grandwizard28 (dismissed):

srikanthccv:
I messed up the rebase. All commits will be squashed.

PR #8238, opened 2025-06-13
chore: find contradictory condition keys in expression

srikanthccv:
## 📄 Summary

More DX improvements; occasionally, people may end up writing filter conditions that will never become true because of contradictory conditions within the expression. This change adds the initial version of the condition analyser that detects if there is any chance of such a case and what field keys are responsible for it. See tests for examples.

### DIFF
--- a/pkg/querybuilder/never_true_test.go
+++ b/pkg/querybuilder/never_true_test.go
@@ -1,0 +1,399 @@
+package querybuilder
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+)
+
+func TestContradictionDetection(t *testing.T) {
+	tests := []struct {
+		name             string
+		query            string
+		hasContradiction bool
+		expectedErrors   []string
+	}{
+		{
+			name:             "Simple equality contradiction",
+			query:            `service.name = 'redis' service.name='route' http.status_code=200`,
+			hasContradiction: true,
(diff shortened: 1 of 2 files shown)


pkg/querybuilder/never_true_test.go
```diff
@@ -0,0 +1,395 @@
...
+func TestContradictionDetection(t *testing.T) {
+	tests := []struct {
+		name             string
+		query            string
+		hasContradiction bool
+		expectedErrors   []string
+	}{
+		{
+			name:             "Simple equality contradiction",
+			query:            `service.name = 'redis' service.name='route' http.status_code=200`,
```
nityanandagohain, line 17: how it this a correct query ?

srikanthccv: It is not, that's why the `hasContradiction` is true and the contradicting key is `service.name`

nityanandagohain: No I mean shouln't it be

```
service.name = 'redis' AND  service.name='route' AND http.status_code=200
```

srikanthccv: An implicit `AND` is assumed in the absence of logical conjunction operator for user's convenience.

nityanandagohain (approved):

PR #8263, opened 2025-06-16
fix(8232): added fix for error graph in services tab

aniketio-ctrl:
## 📄 Summary
Added Fix for error graph it was earlier using service_name instead of status code.

Tested ss - > 

Screenshot 2025-06-16 at 7 50 07 PM


srikanthccv:
While at it, please check all the Service details tab panels once again to see if there are any such gaps (in Overview, DB, and external tabs).

aniketio-ctrl:
> While at it, please check all the Service details tab panels once again to see if there are any such gaps (in Overview, DB, and external tabs).

yupp checked it,everything is working

one of the call is making in overview(latency) is with traces and was not working with not in filter with signoz.collector.id

srikanthccv:
This `signoz.collector.id` can be ignored.

Issue #8392, opened 2025-06-30
Exceptions page filter bug


ch0s1n:
## Bug description

I found a bug on the expectation page. After selecting the env filter conditions, clicking the paging filter conditions will clear.

makeavish:
Thanks for reporting

nityanandagohain:
Fixed in https://github.com/SigNoz/signoz/pull/8300

PR #8300, opened 2025-06-19
fix: retain filters during pagination and field filtering on exceptions page

scout9ll:
…
## 📄 Summary

Fixes a bug where filter conditions from the left sidebar (`compositeQuery`) were lost when performing pagination or field-based filtering on the Exceptions page.

This issue caused unexpected data results when users switched pages or adjusted table filters, because the existing composite filters were not preserved.

before:
gif1

after:
gif2

---

## ✅ Changes

- [x] Bug fix: Retain `compositeQuery` (left sidebar filters) when paginating or applying table-level filters on the Exceptions page

---

## 👥 Reviewers

> Tag the relevant teams for review:

- frontend

---

## 🧪 How to Test
1. On the **Exceptions page**, apply one or more filters using the **left-hand Filter Bar** 
2. Click a pagination control to navigate to another page.
3. Or, perform a search or sorting action on one of the table columns.
4. Verify:
   - The filters from the left sidebar remain active.
   - The table data is correctly filtered based on those sidebar filters.
---

## 🔍 Related Issues

Fixes #8288

---

## 📸 Screenshots / Screen Recording (if applicable / mandatory for UI related changes)

---

## 📋 Checklist

- [x] Dev Review
- [x] Manually tested the changes

---

## 👀 Notes for Reviewers

----

> [!IMPORTANT]
> Fixes bug in `AllErrors` to retain `compositeQuery` during pagination and filtering on the Exceptions page.
> 
>   - **Behavior**:
>     - Fixes bug in `AllErrors` component to retain `compositeQuery` during pagination and field filtering.
>     - Updates `handleSearch` and `onChangeHandler` to include `compositeQuery` in query parameters.
>   - **Utilities**:
>     - Adds `compositeQuery` to `QueryParams` type and `urlKey` in `utils.ts`.
>     - Updates `getFilterString` to handle `compositeQuery` in `index.tsx` and `utils.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for eb3c22082afeed1ea5e88552de8259bb62deb1df. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #8396, opened 2025-06-30
fix: fix filter compositerquery change

ch0s1n:
Hi team,

This PR fixes Issue #8392, where filter conditions on the Exception page were lost after performing pagination.

After investigation, the root cause was that the compositeQuery parameter was not being passed correctly during pagination. As a result, active filters were cleared when switching pages. This fix ensures that the existing compositeQuery is preserved and passed during pagination, maintaining the expected filter behavior.

✅ How to Test
Go to the Exception page.

Apply one or more filter conditions (e.g., time range, error type).

Navigate to another page via pagination.

Verify that the filters remain applied and data is correctly rendered.

Let me know if you have any questions or feedback. Thanks!

----

> [!IMPORTANT]
> Fixes filter persistence during pagination on the Exception page by correctly handling `compositeQuery` parameter.
> 
>   - **Behavior**:
>     - Fixes issue where filters were lost during pagination on the Exception page by ensuring `compositeQuery` is preserved.
>     - Updates `AllErrors` component in `index.tsx` to include `compositeQuery` in query parameters during pagination.
>   - **Functions**:
>     - Adds `getCompositeQuery` function in `utils.ts` to handle `compositeQuery` parameter.
>     - Updates `extractFilterValues` in `utils.ts` to include `compositeQuery`.
>   - **Constants**:
>     - Adds `COMPOSITE_QUERY_FILTER_NAME` to `constant.ts` for consistent reference.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for cb2b24090008fee6d6a8bd742fdcec98818cb23a. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


frontend/src/container/AllError/index.tsx
```diff
@@ -59,6 +60,7 @@ import {
 } from './utils';
 
 type QueryParams = {
+	compositeQuery: string;
```
ahmadshaheer, line 63: nit: instead of redefining the compositeQuery, can you please use the existing one from https://github.com/SigNoz/signoz/blob/cb2b24090008fee6d6a8bd742fdcec98818cb23a/frontend/src/constants/query.ts#L21-L22?

ahmadshaheer (approved):
Thanks for making these changes, @ch0s1n :pray:  :raised_hands: . added a small comment. rest LGTM!

frontend/src/container/AllError/index.tsx
ch0s1n: Thank you for the suggestion!

I believe there might be a misunderstanding here — in this context, I'm declaring the type of the compositeQuery parameter, not assigning a value to it. So using QueryConstants.compositeQuery (which is a value from the enum) doesn't seem appropriate here, since it's not a type.

Perhaps I'm not fully understanding your intention — would you mind clarifying how you'd recommend changing it? If there's a specific way you think I should modify this declaration, I'd really appreciate a concrete example.

Thanks again!

ch0s1n: Just to confirm — do you mean we should remove the QueryParams type defined on line 62 and instead directly use the QueryParams enum from constants/query.ts?

If so, I noticed that the existing enum in constants/query.ts doesn't yet cover all the query parameters used within the AllError component. Currently, some of those parameters are defined in the local constants file at src/container/AllError/constant.ts.

If that's what you’re suggesting, I'm happy to migrate the missing constants from the local file into constants/query.ts to centralize all query keys in one place. Please let me know if that aligns with your expectation.

If that sounds good, I can include that change in this PR as well.

ahmadshaheer: sorry for the confusion, @ch0s1n, I had added the comment on the incorrect line. I meant we could use `QueryParams.compositeQuery` rather than redefining the `COMPOSITE_QUERY_FILTER_NAME` in https://github.com/SigNoz/signoz/blob/cb2b24090008fee6d6a8bd742fdcec98818cb23a/frontend/src/container/AllError/constant.ts#L4-L5

i.e. we could refactor

https://github.com/SigNoz/signoz/blob/cb2b24090008fee6d6a8bd742fdcec98818cb23a/frontend/src/container/AllError/utils.ts#L195-L200


to 

```ts
if (filters[QueryParams.compositeQuery]) {
	filterValues.compositeQuery = extractSingleFilterValue(
		QueryParams.compositeQuery,
		filters,
	);
}

```

ahmadshaheer: it's a nit change. we can merge the PR even without the above change I guess.

aks07 (approved):
LGTM

ahmadshaheer:
Hi @ch0s1n, thank you so much for your contribution! 🙏

We're closing this PR because a similar one was raised before this and was recently merged: https://github.com/SigNoz/signoz/pull/8300.

We really appreciate your effort and would love to see more contributions from you in the future. Thanks again for taking the time to work on this!

Issue #8408, opened 2025-07-01
[memorycache] data race in memorycache


grandwizard28:
## Description

The `memorycache` is returning the same object which is causing data race. when multiple processes are trying to access that object.

PR #8414, opened 2025-07-02
fix(memorycache): add a cloneable interface

therealpandey:
## 📄 Summary

- get and set a clone of the underlying pointer to prevent race conditions

----

> [!IMPORTANT]
> Add a cloneable interface to memory cache to prevent race conditions by cloning objects, with updated `Set` and `Get` methods and new tests.
> 
>   - **Behavior**:
>     - Add `Cloneable` interface in `cachetypes/cacheable.go` for deep copying cacheable objects.
>     - Modify `Set` and `Get` methods in `memorycache/provider.go` to clone objects implementing `Cloneable` to prevent race conditions.
>     - Handle non-cloneable objects by serializing/deserializing in `Set` and `Get` methods.
>   - **Tests**:
>     - Add tests in `memorycache/provider_test.go` for `Set` and `Get` with cloneable and non-cloneable objects.
>     - Add concurrent access test to ensure thread safety with cloneable objects.
>   - **Misc**:
>     - Rename import `go_cache` to `gocache` in `memorycache/provider.go`.
>     - Update error handling in `CheckCacheablePointer` in `cachetypes/cacheable.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for e781e99a2e0d3935542fadf89c864aff60a52d9c. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/deploy/docker-swarm/docker-compose.ha.yaml
+++ b/deploy/docker-swarm/docker-compose.ha.yaml
@@ -174,7 +174,7 @@ services:
       # - ../common/clickhouse/storage.xml:/etc/clickhouse-server/config.d/storage.xml
   signoz:
     !!merge <<: *db-depend
-    image: signoz/signoz:v0.88.0
+    image: signoz/signoz:v0.88.1
     command:
       - --config=/root/config/prometheus.yml
     ports:
--- a/deploy/docker-swarm/docker-compose.yaml
+++ b/deploy/docker-swarm/docker-compose.yaml
@@ -115,7 +115,7 @@ services:
       # - ../common/clickhouse/storage.xml:/etc/clickhouse-server/config.d/storage.xml
   signoz:
     !!merge <<: *db-depend
-    image: signoz/signoz:v0.88.0
+    image: signoz/signoz:v0.88.1
     command:
       - --config=/root/config/prometheus.yml
     ports:
--- a/deploy/docker/docker-compose.ha.yaml
+++ b/deploy/docker/docker-compose.ha.yaml
@@ -177,7 +177,7 @@ services:
       # - ../common/clickhouse/storage.xml:/etc/clickhouse-server/config.d/storage.xml
   signoz:
     !!merge <<: *db-depend
-    image: signoz/signoz:${VERSION:-v0.88.0}
+    image: signoz/signoz:${VERSION:-v0.88.1}
     container_name: signoz
     command:
       - --config=/root/config/prometheus.yml
--- a/deploy/docker/docker-compose.yaml
+++ b/deploy/docker/docker-compose.yaml
@@ -110,7 +110,7 @@ services:
       # - ../common/clickhouse/storage.xml:/etc/clickhouse-server/config.d/storage.xml
   signoz:
     !!merge <<: *db-depend
-    image: signoz/signoz:${VERSION:-v0.88.0}
+    image: signoz/signoz:${VERSION:-v0.88.1}
     container_name: signoz
     command:
       - --config=/root/config/prometheus.yml
--- a/frontend/src/container/AppLayout/index.tsx
+++ b/frontend/src/container/AppLayout/index.tsx
@@ -63,7 +63,6 @@ import {
 } from 'types/api/licensesV3/getActive';
 import AppReducer from 'types/reducer/app';
 import { USER_ROLES } from 'types/roles';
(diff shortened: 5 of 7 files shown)


pkg/cache/memorycache/provider_test.go
```diff
@@ -8,242 +8,137 @@ import (
...
-	orgID := valuer.GenerateUUID()
-	assert.NoError(t, c.Set(context.Background(), orgID, "key", storeCacheableEntity, 10*time.Second))
-	c.Delete(context.Background(), orgID, "key")
+	// confirm that the cached cacheable is equal to the original cacheable
+	assert.Equal(t, cacheable, cachedCacheable)
 
-	err = c.Get(context.Background(), orgID, "key", retrieveCacheableEntity, false)
-	assert.Error(t, err)
+	// confirm that the cached cacheable is a different pointer
+	assert.NotSame(t, cacheable, cachedCacheable)
```
therealpandey: Added one unit test here.

therealpandey:
Fixes https://github.com/SigNoz/signoz/issues/8408

therealpandey:
Closes https://github.com/SigNoz/platform-pod/issues/506

grandwizard28:
Part of https://github.com/SigNoz/platform-pod/issues/506

pkg/querier/bucket_cache_test.go
```diff
@@ -1408,3 +1409,33 @@ func TestBucketCache_NoCache(t *testing.T) {
 	// The actual NoCache logic is implemented in querier.run(), not in bucket cache
 	// This test verifies that the cache works normally and NoCache bypasses it at a higher level
 }
+
+func TestBucketCache_Clone(t *testing.T) {
```
therealpandey: @srikanthccv added one unit test here.

pkg/query-service/querycache/query_range_cache.go
```diff
@@ -37,6 +37,12 @@ type CacheableSeriesData struct {
 	Series []CachedSeriesData
 }
 
+func (c *CacheableSeriesData) Clone() cachetypes.Cacheable {
+	return &CacheableSeriesData{
+		Series: c.Series,
```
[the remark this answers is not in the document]
srikanthccv: The `Data` of `CachedSeriesData` is a pointer array, so this is relevant as the panic trigger comes from the mutating the `[]Point` of the v3.Series, which is a pointer here. 

```
copyData &{[{1500 2500 [0x14000114e40 0x14000114e80]}]} oldData {[{1500 2500 [0x14000114e40 0x14000114e80]}]}
```

pkg/types/cachetypes/cacheable.go
```diff
@@ -10,24 +10,23 @@ import (
 type Cacheable interface {
 	encoding.BinaryMarshaler
 	encoding.BinaryUnmarshaler
+	Clone() Cacheable
```
srikanthccv, line 18: Thinking out loud as it sets a precedent. The implementor needs to maintain three methods to make their type truly cacheable, which I think is a kind of leaky abstraction, because different cache providers have different needs that are pushed to the implementor. 

- For the most part data we cache is relatively small, a few hundred time series with a few hundred data points. Should we stick just to using the binary serializer already enforced?
- The large span case is unique since the data is immutable, where the cost of ser/deser is unnecessary overhead, would it make sense to create a purpose-built solution for immutable types?

nityanandagohain: But if we move away from inmemory cache will it still hold true for something like redis, where even for the large span ser/deser will happen ?

srikanthccv: By moving away, did you mean we get rid of the `in_memory` as provider altogether, or did you mean we use redis as backend for our cloud/enterprise offering? If it's the latter the concern still holds as it's about the interface. If it's the former, then I don't think that's a good idea because it means no caching at all or one needs to spin up another process, which can affect the adoption, while I don't have the numbers but the in-memory caching can be pretty useful for most small-medium OSS deployments so I think it as a provider should remain.

therealpandey: Agreed!

PR #8518, opened 2025-07-14
chore(dot-metrics): added log line to check for those queries who are still using normalized metrics

aniketio-ctrl:
## 📄 Summary

Added log line to check for those queries who are still using normalized metrics
We will create a map of normalized metrics map to unnormalized metrics map, and check if the 
1.) metrics is in the keys -> if key and value is same then means its a unnormalized metrics name otherwise its a normalized metrics name.
2.) metrics is not in the keys - > means its a unnormalized metrics name

---

ex:

- `backend`

## 🧪 How to Test

1. Try calling query range you will get the log lines if you are using normalized metrics name

## 👀 Notes for Reviewers

Its temporary to check all clients wether they are migrated successfully or not.

----

> [!IMPORTANT]
> Adds logging for queries using normalized metrics and introduces `GetNormalizedStatus()` to check metric normalization status.
> 
>   - **Behavior**:
>     - Adds logging for queries using normalized metrics in `ValidateMetricNames()` in `helper.go` and `QueryRange()` in `querier.go`.
>     - Introduces `GetNormalizedStatus()` in `reader.go` to check metric normalization status.
>   - **Models**:
>     - Adds `MetricsNormalizedMap` in `normalizedMetrics.go` to represent metric normalization status.
>   - **Interfaces**:
>     - Updates `Reader` interface in `interface.go` to include `GetNormalizedStatus()`.
>   - **Constants**:
>     - Adds constants `NormalizedMetricsMapCacheKey`, `NormalizedMetricsMapQueryThreads`, `NormalizedMetricsMapRegex`, and `NormalizedMetricsMapQuantileRegex` in `constants.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 5ad633051a264b1c190dc7bd1fe5e31cf1e9dae4. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -6046,3 +6046,86 @@ func (r *ClickHouseReader) SearchTraces(ctx context.Context, params *model.Searc
...
+	// Try cache first
+	resultDto := new(model.NormalizedMetricsMap)
+	cacheKey := constants.NormalizedMetricsMapCacheKey
+
+	if err := r.cache.Get(ctx, orgID, cacheKey, resultDto, true); err == nil {
+		return resultDto.MetricsMap, nil
+	}
+
+	valueCtx := context.WithValue(ctx, "clickhouse_max_threads", constants.NormalizedMetricsMapQueryThreads)
+	query := "SELECT DISTINCT metric_name, type, toUInt8(__normalized) FROM %s.%s"
```
srikanthccv: Why are we loading all the metrics into cache regardless of whether they are used or not?

aniketio-ctrl: to create a map instead of calling every time in query-service, we can load all at once, for a user metrics wont be more than 1000, so it would be much better to store all at once

srikanthccv: Many users have ~5k metrics in total, and some even more. We are going to load them all into memory (because we are using memory cache, even though very few of the metrics are used.

aniketio-ctrl: 1.)for every new query we will be adding latency in the query range, the reason i gone with loading all the metrics, is beacuse no of metrics is gonna be finite, for our current use case, it wont gonna scale, from 5k to 5m, 
2.) its not that very few metrics we are gonna use, its that we have to check every query of any alert and dashboard, we will consider any new metrics to be normalized then will check wether its correct or not

srikanthccv: >1.)for every new query we will be adding latency in the query range, the reason i gone with loading all the metrics, is beacuse no of metrics is gonna be finite, for our current use case, it wont gonna scale, from 5k to 5m,

Even when it's finite, when you load 5k metrics into memory for each user, it going to take up unnecessary memory when there are hundreds of users. 

> its not that very few metrics we are gonna use, its that we have to check every query of any alert and dashboard, we will consider any new metrics to be normalized then will check wether its correct or not

What is the problem with checking from every metric? We already do this to `FetchTemporality` and then cache the result of just the metric in use.

aniketio-ctrl: added the single metrics call add updated the function

srikanthccv: Have you looked at the `FetchTemporality` implementation? It doesn't make a single call each metric query. It collect the list of metric names and fetch them all at once. Please reduce the DB roundtrips to one for each query range.

aniketio-ctrl: added the search for list, reduced the db round trips

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -6046,3 +6047,137 @@ func (r *ClickHouseReader) SearchTraces(ctx context.Context, params *model.Searc
...
+
+	metricsToRegex := func(name string) string {
+		tokens := strings.FieldsFunc(strings.ToLower(name), func(r rune) bool {
+			return !unicode.IsLetter(r) && !unicode.IsDigit(r)
+		})
+		if len(tokens) == 0 {
+			return "^$"
+		}
+		return "^" + strings.Join(tokens, "[^a-z0-9]*")
+	}
```
srikanthccv: Did you test the performance for this? The regex matching is 5x slower than directly using the metric and normalised metric name. For example, instead of building a regex for `process.cpu.time`, why can't we search on ['process.cpu.time', 'process_cpu_time']? We know the normalised metric name in the database, if any.

aniketio-ctrl: removed the regex added simple in query on metirc names

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -6046,3 +6046,95 @@ func (r *ClickHouseReader) SearchTraces(ctx context.Context, params *model.Searc
...
+          GROUP BY metric_name, __normalized`,
+		signozMetricDBName, signozTSTableNameV41Day, placeholders,
+	)
+
+	args := make([]any, len(uncached))
+	for i, v := range uncached {
+		args[i] = v
+	}
+
+	rows, err := r.db.Query(ctx, q, args...)
```
srikanthccv: what is the point of args if you are using `placeholders`

pkg/query-service/app/querier/v2/helper.go
```diff
@@ -276,3 +277,59 @@ func (q *querier) runBuilderQuery(
...
+			return
+		}
+		for k, m := range metrics {
+			if m {
+				continue
+			} else {
+				zap.L().Warn("using normalized metric name", zap.String("metrics", k))
+				continue
+			}
+		}
```
srikanthccv: Please use readable names, what should the reader interpret as just reading the `k, m`?

srikanthccv (commented):
Minor comments, otherwise LGTM.

srikanthccv (approved):

PR #8563, opened 2025-07-19
fix(prom-dup-labels): added fingerprint in prom labels

aniketio-ctrl:
## 📄 Summary
The issue stems from the __resource.schema_url__ label. When this label is removed, Collector treats the data as a new time‑series instead of appending to the existing one.

t1 – has __resource.schema_url__; last sample at time1

t2 – identical to t1 except __resource.schema_url__ is missing or empty; last sample at time2 (with time1 < time2)

If you query a range time3 < time1 to time4 > time2 both series are returned. The seriesSetFilter.labels() call then strips labels whose values are empty, so t1 and t2 end up with identical label sets. Prometheus detects two series with the same labels and raises a “duplicate series” error.

A straightforward fix is to add a stable, unique label—such as each series’ fingerprint—when constructing the Prometheus labels. That prevents collisions even after __resource.schema_url__ is dropped.

---

----

> [!IMPORTANT]
> Add `fingerprint` label to Prometheus series to prevent duplicate series errors and remove it after query execution.
> 
>   - **Behavior**:
>     - Add `fingerprint` label to Prometheus series in `unmarshalLabels()` in `json.go` to prevent duplicate series errors.
>     - Remove `fingerprint` label after query execution in `RemoveExtraLabels()` in `utils.go`.
>   - **Functions**:
>     - Update `unmarshalLabels()` in `json.go` to accept `fingerprint` and append it to labels.
>     - Add `RemoveExtraLabels()` in `utils.go` to remove `fingerprint` label from query results.
>   - **Tests**:
>     - Add `TestClient_getFingerprintsFromClickhouseQuery` in `client_query_test.go` to verify fingerprint handling.
>     - Add `TestClient_QuerySamples` in `client_query_test.go` to test sample queries with fingerprints.
>   - **Misc**:
>     - Define `FingerprintAsPromLabelName` constant in `constants.go`.
>     - Update `GetInstantQueryMetricsResult()` and `GetQueryRangeResult()` in `reader.go` to call `RemoveExtraLabels()`.
>     - Update `RunAlertQuery()` in `prom_rule.go` to call `RemoveExtraLabels()`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 61d7934a069f2926cf4a316c09e570e1e002dd44. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


srikanthccv:
The issue with passing this to labels is that it will also be present result labels if the query doesn't do any aggregation, for instance, if there is an `up` query as `up{"host.name"="my-host"}`, then the "fingerprint" will also present in the result labels which is unnecessary, can we get rid of it if exist in the result of the query?

aniketio-ctrl:
> The issue with passing this to labels is that it will also be present result labels if the query doesn't do any aggregation, for instance, if there is an `up` query as `up{"host.name"="my-host"}`, then the "fingerprint" will also present in the result labels which is unnecessary, can we get rid of it if exist in the result of the query?

make sense, removed it from result series

srikanthccv:
Please check in the pkg/query-service/rules/prom_rule as well. Separately, Please check if prometheus ingests the labels with empty values (and then complain at the query time) or does it drop at them at the write stage.

aniketio-ctrl:
> Please check in the pkg/query-service/rules/prom_rule as well. Separately, Please check if prometheus ingests the labels with empty values (and then complain at the query time) or does it drop at them at the write stage.

removed from all possible places, except from qb v5

What do you mean by ingests the labels with empty values?

srikanthccv:
> What do you mean by ingests the labels with empty values?

On a standalone Prometheus setup, if Prometheus server scrapes an endpoint with a label having empty value and label missing, does it throw error at the query time or does it prevent this from happening at the write so there won't be an issue when querying.

srikanthccv:
>removed from all possible places, except from qb v5


Good, please add tests

aniketio-ctrl:
> > removed from all possible places, except from qb v5
> 
> Good, please add tests

done

aniketio-ctrl:
> > What do you mean by ingests the labels with empty values?
> 
> On a standalone Prometheus setup, if Prometheus server scrapes an endpoint with a label having empty value and label missing, does it throw error at the query time or does it prevent this from happening at the write so there won't be an issue when querying.

yeah scrape endpoint will show empty labels, but write when check in tsdb WAL files, i can see that empty labels are not stored there.

https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels:~:text=Labels%20with%20an%20empty%20label%20value%20are%20considered%20equivalent%20to%20labels%20that%20do%20not%20exist.

pkg/prometheus/utils.go
```diff
@@ -0,0 +1,49 @@
+package prometheus
+
+import (
+	"fmt"
+	"github.com/prometheus/prometheus/promql"
+)
+
+func RemoveExtraLabels(res *promql.Result, labelsToRemove ...string) error {
```
srikanthccv, line 8: Please add tests for this.

aniketio-ctrl: done

PR #8615, opened 2025-07-25
fix: changes in code to support ch 25.5

nityanandagohain:
Changes in the trace code to replace the alias columns by the actual column in the builder

----

> [!IMPORTANT]
> Replace alias columns with actual column names in SQL queries and update related functions and tests.
> 
>   - **Behavior**:
>     - Replace alias columns with actual column names in `clickhouse_queries.go`, `query.go`, and `reader.go`.
>     - Remove `BuildTracesFilter` function in `query_builder.go` and integrate its functionality into `BuildTracesFilterQuery`.
>   - **Functions**:
>     - Update `getColumnName()` in `query_builder.go` to support alias replacement.
>     - Modify `BuildTracesFilterQuery()` in `query_builder.go` to handle alias replacement and remove `BuildTracesFilter`.
>   - **Constants**:
>     - Add new mappings in `constants.go` for deprecated and new trace fields.
>   - **Tests**:
>     - Update tests in `query_builder_test.go` to reflect changes in column name handling and alias replacement.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 6f300d3d5207849b53037b6de21be3a98dc6dbb8. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


eKuG (approved):
LGTM

pkg/query-service/app/traces/v4/query_builder.go
```diff
@@ -78,29 +86,30 @@ func getSelectLabels(groupBy []v3.AttributeKey) string {
 func existsSubQueryForFixedColumn(key v3.AttributeKey, op v3.FilterOperator) (string, error) {
 	if key.DataType == v3.AttributeKeyDataTypeString {
 		if op == v3.FilterOperatorExists {
-			return fmt.Sprintf("%s %s ''", getColumnName(key), tracesOperatorMappingV3[v3.FilterOperatorNotEqual]), nil
+			return fmt.Sprintf("%s %s ''", getColumnName(key, false), tracesOperatorMappingV3[v3.FilterOperatorNotEqual]), nil
 		} else {
-			return fmt.Sprintf("%s %s ''", getColumnName(key), tracesOperatorMappingV3[v3.FilterOperatorEqual]), nil
+			return fmt.Sprintf("%s %s ''", getColumnName(key, false), tracesOperatorMappingV3[v3.FilterOperatorEqual]), nil
```
srikanthccv: Why is it `false` here but `true` for all items in `BuildTracesFilterQuery`?

srikanthccv (commented):
Should you do this for v3 as well because there are some people who rely on `/v3/query_range`?

nityanandagohain:
the v3 query builder is stopped some time back, I am removing the code step by step https://github.com/SigNoz/signoz/blob/210393e28133405548f229e3cbc9142cab870298/pkg/query-service/app/traces/v3/query_builder.go#L361

pkg/query-service/app/traces/v4/query_builder.go
nityanandagohain: thanks for the catch, addressed it with tests

srikanthccv:
> the v3 query builder is stopped some time back

Are they don't work or are they redirected to v4 and v3 requests work with v4 query builder code?

nityanandagohain:
We use the V4 query builder, by default. Previously we allowed the users to use the old v3 query builder which used the old tables with a parameter, but we have stopped writing to those tables and we use the V4 by default.

srikanthccv:
I don't mean to ask related traces schema migration. I meant it from the API perspective, we have `/v3/query_range` and `/v4/query_range`. The difference between them was the metrics query builder, do the requests to `/v3/query_range` with traces work?

nityanandagohain:
Yeah, the changes applies to that as well, i.e the keys will be mapped.

srikanthccv (approved):

PR #8642, opened 2025-07-28
chore: add sql migration for dashboards, alerts, and saved views

srikanthccv:
## 📄 Summary

To reliably migrate the alerts and dashboards, we need access to the telemetrystore to fetch some metadata and while doing migration, I need to log some stuff to fix stuff later.

Key changes:
- Modified the migration to include telemetrystore and a logging provider (open to using a standard logger instead)
- To avoid the previous issues with imported dashboards failing during migration, I've ensured that imported JSON files are automatically transformed when migration is active
- Implemented detailed logic to handle dashboard migration cleanly and prevent unnecessary errors
- Separated the core migration logic from SQL migration code, as users from the dot metrics migration requested shareable code snippets for local migrations. This modular approach allows others to easily reuse the migration functionality.

Known: I didn't register the migration yet in this PR, and will not merge this yet, so please review with that in mid.

----

> [!IMPORTANT]
> Adds SQL migration for dashboards, alerts, and saved views with modular logic and telemetry store integration.
> 
>   - **Migration**:
>     - Adds `046_update_dashboard_alert_and_saved_view_v5.go` for SQL migration of dashboards, alerts, and saved views.
>     - Introduces `NewQueryBuilderV5MigrationFactory` in `sqlmigration` to handle migration logic.
>     - Separates core migration logic from SQL migration code for reusability.
>   - **Telemetry Store**:
>     - Updates `GetKeys`, `GetKeysMulti`, `GetRelatedValues`, and `GetAllValues` in `telemetrymetadata` to return a completion status.
>     - Adjusts `metadata_store.go` in `telemetrytypestest` to reflect changes in telemetry store interface.
>   - **Query Building**:
>     - Modifies `Build` methods in `metricQueryStatementBuilder` and `traceQueryStatementBuilder` to use updated telemetry store methods.
>     - Adds `Copy` methods to various query builder types in `querybuildertypesv5` for deep copying.
>   - **Miscellaneous**:
>     - Adds `QBV5Enabled` flag in `querybuilder/init.go` to toggle new query builder version.
>     - Updates `README.md` for typo fixes and minor logging changes in `app.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for ad96d5a6fc281ee3b468d9caaa86737e5a5b855b. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


pkg/sqlmigration/046_update_dashboard_alert_and_saved_view_v5.go
```diff
@@ -0,0 +1,297 @@
...
+
+		if updated {
+			dataJSON, err := json.Marshal(data)
+
+			if err != nil {
+				return err
+			}
+
+			_, err = tx.NewUpdate().
+				Model((*types.SavedView)(nil)).
```
vikrantgupta25: please use Table() here or define the saved views type here itself inline. we don't wanna bring imports from types in migration

pkg/sqlmigration/046_update_dashboard_alert_and_saved_view_v5.go
```diff
@@ -0,0 +1,297 @@
...
+	return nil
+}
+
+func (migration *queryBuilderV5Migration) migrateSavedViews(
+	ctx context.Context,
+	tx bun.Tx,
+	logsKeys []string,
+	tracesKeys []string,
+) error {
+	var savedViews []*types.SavedView
```
vikrantgupta25: define the saved views type inline do not import from types package. this causes issues when the types evolve.

pkg/transition/migrate_dashboard.go
```diff
@@ -0,0 +1,95 @@
...
+	return &dashboardMigrateV5{
+		migrateCommon: migrateCommon{
+			ambiguity: ambiguity,
+			logger:    logger,
+		},
+	}
+}
+
+func (m *dashboardMigrateV5) Migrate(ctx context.Context, dashboardData map[string]any) bool {
+
```
vikrantgupta25: NIT: let's not leave lines empty post the function signature, maintains uniformity across codebase

pkg/transition/migrate_alert.go
```diff
@@ -0,0 +1,123 @@
...
+	compositeQuery, ok := ruleCondition["compositeQuery"].(map[string]any)
+	if !ok {
+		m.logger.WarnContext(ctx, "didn't find composite query")
+		return updated
+	}
+
+	if compositeQuery["queries"] == nil {
+		compositeQuery["queries"] = []any{}
+	}
+	m.logger.InfoContext(ctx, "setup empty list")
```
vikrantgupta25: move this inside if condition

pkg/transition/migrate_dashboard.go
```diff
@@ -0,0 +1,95 @@
+// nolint
```
vikrantgupta25, line 1: why no lint ?

srikanthccv: https://github.com/go-simpler/sloglint/issues/97

pkg/sqlmigration/046_update_dashboard_alert_and_saved_view_v5.go
srikanthccv: There are a lot of `types` imported in migration already. I would prefer to keep it this way.

pkg/sqlmigration/046_update_dashboard_alert_and_saved_view_v5.go
srikanthccv: Same, there are already types in sqlmigration. This change is not adding as first step and I think the idea of re-define bad idea.

vikrantgupta25: > I think the idea of re-define bad idea.

it's needed because the types can evolve and when the same type is being referenced in the older migration which expects it to be in certain shape or form it doesn't. hence when you try to run migrations from scratch things fail.

pkg/sqlmigration/046_update_dashboard_alert_and_saved_view_v5.go
vikrantgupta25: we are using the Table command in the same file already 😅

pkg/sqlmigration/046_update_dashboard_alert_and_saved_view_v5.go
srikanthccv: If it changes in a backward-incompatible then it's breaking change (for someone using it) and failure is expected so the type should be versioned?

Imagine I had a dashboard/alert/saved view with structured payload (not the one we have now where everything is simple string) defined with clear types, would you want the author each migration to inline define entire type here?

srikanthccv: > Imagine I had a dashboard/alert/saved view with structured payload

By this I mean a concrete go structure defining the entire structure of the data.

srikanthccv:
Using the "version" identifier to skip the already migrated assets, also using the same for dashboard import here https://github.com/SigNoz/signoz/pull/8642/files#diff-661817061c84fac3a977e265b5516a443a8c9c2565b45935344d5f4ba139749dR55

pkg/sqlmigration/046_update_dashboard_alert_and_saved_view_v5.go
vikrantgupta25: > If it changes in a backward-incompatible then it's breaking change (for someone using it) and failure is expected so the type should be versioned?

yes but when the schema generally is in evolving stage it's bound to have breaking changes. and we don't have very well versioned types as of now

vikrantgupta25 (dismissed):

vikrantgupta25 (approved):
LGTM

PR #8644, opened 2025-07-28
fix(added-backticks): added backticks for hyphen also

aniketio-ctrl:
## 📄 Summary

Added backtick so that clickhouse queries can work in metrics, so when use as alias we can surrond the variable with backtick

---

### DIFF
--- a/pkg/query-service/utils/format.go
+++ b/pkg/query-service/utils/format.go
@@ -245,7 +245,7 @@ func ClickHouseFormattedMetricNames(v interface{}) string {
 }
 
 func AddBackTickToFormatTag(str string) string {
-	if strings.Contains(str, ".") {
+	if strings.Contains(str, ".") || strings.Contains(str, "-") {
 		return "`" + str + "`"
 	} else {
 		return str


pkg/query-service/utils/format.go
```diff
@@ -245,8 +245,8 @@ func ClickHouseFormattedMetricNames(v interface{}) string {
 }
 
 func AddBackTickToFormatTag(str string) string {
-	if strings.Contains(str, ".") {
-		return "`" + str + "`"
+	if strings.Contains(str, ".") || strings.Contains(str, "-") {
+		if strings.HasPrefix(str, "`") && strings.HasSuffix(str, "`") { return str } else { return "`" + str + "`" }
```
srikanthccv, line 249: `if strings.HasPrefix(str, "`") && strings.HasSuffix(str, "`")` is this added to avoid double "`"? When does that happen?

aniketio-ctrl: no it doesn't happen check was suggested by elipses just a one more safe check, then main thing was clickhouse queries was not working with alias name having '-'

PR #8667, opened 2025-07-30
feat(telemetrymeter): add support for telemetry meter 

vikrantgupta25:
###  Summary

- add telemetry meter metrics support

contributes to - https://github.com/SigNoz/platform-pod/issues/986

### DIFF
--- a/pkg/types/quickfiltertypes/filter.go
+++ b/pkg/types/quickfiltertypes/filter.go
@@ -35,7 +35,7 @@ var (
 	SignalLogs          = Signal{valuer.NewString("logs")}
 	SignalApiMonitoring = Signal{valuer.NewString("api_monitoring")}
 	SignalExceptions    = Signal{valuer.NewString("exceptions")}
-	SignalMeterExplorer = Signal{valuer.NewString("meter_explorer")}
+	SignalMeter         = Signal{valuer.NewString("meter")}
 )
 
 // NewSignal creates a Signal from a string
@@ -49,8 +49,8 @@ func NewSignal(s string) (Signal, error) {
 		return SignalApiMonitoring, nil
 	case "exceptions":
 		return SignalExceptions, nil
-	case "meter_explorer":
-		return SignalMeterExplorer, nil
+	case "meter":
+		return SignalMeter, nil
 	default:
 		return Signal{}, errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "invalid signal: %s", s)
 	}
@@ -181,7 +181,7 @@ func NewDefaultQuickFilter(orgID valuer.UUID) ([]*StorableQuickFilter, error) {
 		{"key": "k8s.pod.name", "dataType": "string", "type": "resource"},
 	}
 
-	meterExplorerFilters := []map[string]interface{}{
+	meterFilters := []map[string]interface{}{
 		{"key": "deployment.environment", "dataType": "float64", "type": "Sum"},
 		{"key": "service.name", "dataType": "float64", "type": "Sum"},
 		{"key": "host.name", "dataType": "float64", "type": "Sum"},
@@ -207,9 +207,9 @@ func NewDefaultQuickFilter(orgID valuer.UUID) ([]*StorableQuickFilter, error) {
 		return nil, errors.Wrapf(err, errors.TypeInternal, errors.CodeInternal, "failed to marshal exceptions filters")
 	}
 
-	meterExplorerJSON, err := json.Marshal(meterExplorerFilters)
+	meterJSON, err := json.Marshal(meterFilters)
 	if err != nil {
-		return nil, errors.Wrapf(err, errors.TypeInternal, errors.CodeInternal, "failed to marshal meter explorer filters")
+		return nil, errors.Wrapf(err, errors.TypeInternal, errors.CodeInternal, "failed to marshal meter filters")
 	}
 
 	timeRightNow := time.Now()
@@ -268,8 +268,8 @@ func NewDefaultQuickFilter(orgID valuer.UUID) ([]*StorableQuickFilter, error) {
 				ID: valuer.GenerateUUID(),
 			},
 			OrgID:  orgID,
-			Filter: string(meterExplorerJSON),
-			Signal: SignalMeterExplorer,
+			Filter: string(meterJSON),
+			Signal: SignalMeter,
 			TimeAuditable: types.TimeAuditable{
 				CreatedAt: timeRightNow,
 				UpdatedAt: timeRightNow,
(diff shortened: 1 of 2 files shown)


pkg/types/telemetrytypes/signal.go
```diff
@@ -10,5 +10,6 @@ var (
 	SignalTraces      = Signal{valuer.NewString("traces")}
 	SignalLogs        = Signal{valuer.NewString("logs")}
 	SignalMetrics     = Signal{valuer.NewString("metrics")}
+	SignalMeter       = Signal{valuer.NewString("meter")}
```
srikanthccv: I don't think we should refer to this as signal. The intent is to map the signals from https://opentelemetry.io/docs/concepts/signals/ to our telemetry. Can you move this identification to somewhere else, perhaps under `MetricAggregation` definition?

pkg/telemetrymeter/statement_builder.go
```diff
@@ -0,0 +1,363 @@
...
+	end uint64,
+	_ qbtypes.RequestType,
+	query qbtypes.QueryBuilderQuery[qbtypes.MetricAggregation],
+	variables map[string]qbtypes.VariableItem,
+) (*qbtypes.Statement, error) {
+	keySelectors := telemetrymetrics.GetKeySelectors(query)
+	keys, err := b.metadataStore.GetKeysMulti(ctx, keySelectors)
+	if err != nil {
+		return nil, err
+	}
```
srikanthccv: I don't remember why we added the `keys` in the metric statement builder since we don't need the field context or data type similar to logs/traces. I think we can remove this.

pkg/querybuilder/time.go
```diff
@@ -61,6 +61,23 @@ func MinAllowedStepInterval(start, end uint64) uint64 {
...
 
+func RecommendedStepIntervalForMeter(start, end uint64) uint64 {
+	start = ToNanoSecs(start)
+	end = ToNanoSecs(end)
+
+	step := (end - start) / RecommendedNumberOfPoints / 1e9
+
+	// for meter queries the minimum step interval allowed is 1 day as this is our granularity
+	if step < 86400 {
+		return 86400
```
srikanthccv: do you want to always use the 1d table? When do you plan to use 1h default table?

pkg/telemetrymeter/statement_builder.go
```diff
@@ -0,0 +1,363 @@
...
+		for _, g := range query.GroupBy {
+			wrapped.SelectMore(fmt.Sprintf("`%s`", g.TelemetryFieldKey.Name))
+		}
+		wrapped.SelectMore(fmt.Sprintf("%s AS per_series_value", rateExpr))
+		wrapped.From(fmt.Sprintf("(%s) WINDOW rate_window AS (PARTITION BY fingerprint ORDER BY fingerprint, ts)", innerQuery))
+		q, args := wrapped.BuildWithFlavor(sqlbuilder.ClickHouse, innerArgs...)
+		return fmt.Sprintf("__temporal_aggregation_cte AS (%s)", q), args, nil
+
+	case metrictypes.TimeAggregationIncrease:
+		incExpr := fmt.Sprintf(telemetrymetrics.IncreaseWithoutNegative, start, start)
```
srikanthccv, line 326: We probably want to be more strict here (point by point lagInFram than the lagInFrame of `max(max)`) than using the `RateWithoutNegative`/`telemetrymetrics.IncreaseWithoutNegative` defined for metrics. We tolerate a bit of inaccuracies for repeated counter resets in the window. The same might not be acceptable for metering. I will write the detail what I am referring to and share.

pkg/query-service/app/http_handler.go
```diff
@@ -4213,6 +4213,8 @@ func (aH *APIHandler) autocompleteAggregateAttributes(w http.ResponseWriter, r *
 	switch req.DataSource {
 	case v3.DataSourceMetrics:
 		response, err = aH.reader.GetMetricAggregateAttributes(r.Context(), orgID, req, false)
+	case v3.DataSourceMeter:
+		response, err = aH.reader.GetMeterAggregateAttributes(r.Context(), orgID, req)
```
srikanthccv: Same question as signal below. I think this should not be another data source or signal.

pkg/querybuilder/time.go
vikrantgupta25: never for now, we will be using 1D tables for all intents and purposes

pkg/telemetrymeter/statement_builder.go
vikrantgupta25: as discussed will pick this later as we don't support cumulative metrics for meter

pkg/telemetrymeter/statement_builder.go
vikrantgupta25: will remove from both meter builder and metric statement builder together

pkg/types/telemetrytypes/signal.go
vikrantgupta25: removed

pkg/querybuilder/time.go
```diff
@@ -61,6 +61,32 @@ func MinAllowedStepInterval(start, end uint64) uint64 {
 	return step - step%5
 }
 
+func RecommendedStepIntervalForMeter(start, end uint64) uint64 {
+	start = ToNanoSecs(start)
+	end = ToNanoSecs(end)
+
+	step := (end - start) / RecommendedNumberOfPoints / 1e9
+
+	// for meter queries the minimum step interval allowed is 1 day as this is our granularity
```
srikanthccv:
```suggestion
	// for meter queries the minimum step interval allowed is 1 hour as this is our granularity
```

srikanthccv (approved):

srikanthccv (dismissed):

pkg/types/quickfiltertypes/filter.go
```diff
@@ -35,6 +35,7 @@ var (
 	SignalLogs          = Signal{valuer.NewString("logs")}
 	SignalApiMonitoring = Signal{valuer.NewString("api_monitoring")}
 	SignalExceptions    = Signal{valuer.NewString("exceptions")}
+	SignalMeterExplorer = Signal{valuer.NewString("meter_explorer")}
```
therealpandey:
```suggestion
	SignalMeter = Signal{valuer.NewString("meter")}
```

Here signal means the page/the sidenav button. We can get rid of `_explorer`

vikrantgupta25: done, we had the side nav name as meter explorer hence this but we are gonna rename that as well so better to use generic term here

PR #8689, opened 2025-08-03
chore: add endpoint to replace varibales

srikanthccv:
## 📄 Summary

- When creating an alert from the dashboard, the variable used in the search should be replaced with the value of the variable selected at the time of creation. Adds an endpoint that takes the request and returns the same back with variables replaced. There is no good reason this is n/w call, but till the time it can be implemented on the frontend, this endpoint helps to take care of that job.
- Fix `OR` b/w resource attribute and span/log attribute. Avoid short-circuiting by removing the "true".
- Round the float to 3 decimal places

----

> [!IMPORTANT]
> Adds endpoint for variable replacement in alerts, fixes OR logic, and rounds floats to 3 decimals.
> 
>   - **Behavior**:
>     - Adds `/api/v5/substitute_vars` endpoint in `api.go` to replace variables in alert creation requests.
>     - Fixes logical OR handling between resource and span/log attributes in `statement_builder.go`.
>     - Rounds float values to 3 decimal places in `resp.go`.
>   - **Variable Replacement**:
>     - Implements `ReplaceVariables` function in `variable_replace_visitor.go` to replace variables in expressions.
>     - Adds tests for variable replacement in `variable_replace_visitor_test.go`.
>   - **Misc**:
>     - Updates `QueryRangeResponse` and `ScalarData` to round float values in `resp.go`.
>     - Adjusts `PrepareWhereClause` function calls across multiple files for new logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for c88c09f9b63b35d3f6d1d3340254c29fbaa1cdb7. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/ee/anomaly/seasonal.go
+++ b/ee/anomaly/seasonal.go
@@ -50,11 +50,13 @@ func (p *BaseSeasonalProvider) getQueryParams(req *AnomaliesRequest) *anomalyQue
 
 func (p *BaseSeasonalProvider) toTSResults(ctx context.Context, resp *qbtypes.QueryRangeResponse) []*qbtypes.TimeSeriesData {
 
+	tsData := []*qbtypes.TimeSeriesData{}
+
 	if resp == nil {
 		p.logger.InfoContext(ctx, "nil response from query range")
+		return tsData
 	}
 
-	tsData := []*qbtypes.TimeSeriesData{}
 	for _, item := range resp.Data.Results {
 		if resultData, ok := item.(*qbtypes.TimeSeriesData); ok {
 			tsData = append(tsData, resultData)


pkg/querybuilder/where_clause_visitor.go
```diff
@@ -120,12 +120,16 @@ func PrepareWhereClause(query string, opts FilterExprVisitorOpts) (*sqlbuilder.W
 		combinedErrors := errors.Newf(
 			errors.TypeInvalidInput,
 			errors.CodeInvalidInput,
-			"found %d errors while parsing the search expression",
+			"Found %d errors while parsing the search expression. Refer to the troubleshooting guide <a href=\"https://signoz.io/docs/userguide/search-troubleshooting/\" target=\"_blank\" rel=\"noopener noreferrer\">here</a>",
```
therealpandey: Maybe use `WithUrl` here? The new error structure has support for url just for these sort of usecases

srikanthccv: The current component on the frontend is not usable across the product. @SagarRajput-7 @ahrefabhi said they would prefer html contents in the error till that supposed gets built https://signoz-team.slack.com/archives/C03AXVC598Q/p1753679361674159

PR #8734, opened 2025-08-07
fix: added fix for extractquerypararms when value is string in multivalue operators

ahrefabhi:
…alue operator

## 📄 Summary

---

## ✅ Changes

- Added fix for extractQueryPararms when value is string in multivalue operators.

### DIFF
--- a/frontend/src/utils/queryContextUtils.ts
+++ b/frontend/src/utils/queryContextUtils.ts
@@ -1279,6 +1279,15 @@ export function extractQueryPairs(query: string): IQueryPair[] {
 						if (allTokens[iterator].type === closingToken) {
 							multiValueEnd = allTokens[iterator].stop;
 						}
+					} else if (isValueToken(allTokens[iterator].type)) {
+						valueList.push(allTokens[iterator].text);
+						valuesPosition.push({
+							start: allTokens[iterator].start,
+							end: allTokens[iterator].stop,
+						});
+						multiValueStart = allTokens[iterator].start;
+						multiValueEnd = allTokens[iterator].stop;
+						iterator += 1;
 					}
 
 					currentPair.valuesPosition = valuesPosition;


srikanthccv:
Please help me understand why are we mocking the antlr?

frontend/src/components/QueryBuilderV2/utils.test.ts
```diff
@@ -0,0 +1,387 @@
...
+					id: '1',
+					key: { id: 'service.name', key: 'service.name', type: 'string' },
+					op: OPERATORS['='],
+					value: 'test-service',
+				},
+			],
+			op: 'AND',
+		};
+
+		const result = convertFiltersToExpressionWithExistingQuery(
```
srikanthccv, line 35: This depends on `convertFiltersToExpression`, which has an incorrect implementation https://github.com/SigNoz/signoz/pull/8804#issuecomment-3194650213

frontend/src/components/QueryBuilderV2/utils.test.ts
```diff
@@ -0,0 +1,387 @@
...
+			existingQuery,
+		);
+
+		// The function should return the original expression for now (until we fix the replacement logic)
+		expect(result.filters).toBeDefined();
+		expect(result.filter).toBeDefined();
+		expect(result.filter.expression).toBe("service.name = 'old-service'");
+		expect(mockExtractQueryPairs).toHaveBeenCalledWith(
+			"service.name = 'old-service'",
+		);
```
srikanthccv: >		// The function should return the original expression for now (until we fix the replacement logic)



What does this mean? The replacement logic not working?

frontend/src/components/QueryBuilderV2/utils.test.ts
```diff
@@ -0,0 +1,387 @@
...
+	});
+
+	test('should handle NOT IN operator conversion from equals', () => {
+		const filters = {
+			items: [
+				{
+					id: '1',
+					key: { id: 'service.name', key: 'service.name', type: 'string' },
+					op: OPERATORS.IN,
```
srikanthccv: This is not `NOT IN`

frontend/src/components/QueryBuilderV2/utils.test.ts
```diff
@@ -0,0 +1,387 @@
...
+			filters,
+			existingQuery,
+		);
+
+		expect(result.filters).toBeDefined();
+		expect(result.filter).toBeDefined();
+		// The function is currently returning the new value but with extra characters
+		expect(result.filter.expression).toBe(
+			"service.name IN ['service1', 'service2']",
+		);
```
srikanthccv, line 160: What are the "extra characters"?

frontend/src/components/QueryBuilderV2/utils.test.ts
```diff
@@ -0,0 +1,387 @@
...
+				key: 'service.name',
+				operator: 'IN',
+				value: "['old-service']",
+				valueList: ["'old-service'"],
+				valuesPosition: [
+					{
+						start: 17,
+						end: 29,
+					},
+				],
```
srikanthccv, line 133: The problem with these mocked values is that it's hard to verify whether the mocked response is valid. And why do we need to mock in the first place?

frontend/src/components/QueryBuilderV2/utils.test.ts
ahrefabhi: added a fix for that in this PR

srikanthccv: That last https://github.com/SigNoz/signoz/pull/8734/commits/339d81a240e1e13fa0062deeaa5532967ef0cccf is not enough. For example, in the old filters, the operator for `NOT IN` was `nin` which is invalid in the new component.

PR #8807, opened 2025-08-14
fix: update live tail api

nityanandagohain:
Fixes https://github.com/SigNoz/engineering-pod/issues/2809

eg:-
```
/api/v3/logs/livetail?start=1754910151000&filter=service.name=newlinetab1
```

two query params
start = start timestamp
fitler = logs filter

----

> [!IMPORTANT]
> Adds new live tailing API for logs with `RequestTypeRawStream`, updating query building and validation logic, and removing old implementations.
> 
>   - **Behavior**:
>     - Adds `QueryRawStream` method in `api.go` for live tailing logs with `start` and `filter` query params.
>     - Introduces `RequestTypeRawStream` in `request_type.go` for live streaming logs.
>     - Removes old live tailing methods `LiveTailLogsV3` and `LiveTailLogsV4` from `clickhouseReader/reader.go`.
>   - **Interfaces**:
>     - Adds `QueryRawStream` to `Querier` interface in `interfaces.go`.
>     - Removes `LiveTailLogsV3` and `LiveTailLogsV4` from `Reader` interface in `interfaces.go`.
>   - **Query Building**:
>     - Updates `Build` method in `statement_builder.go` to handle `RequestTypeRawStream`.
>     - Adjusts time filter logic in `addTimeFilter` in `resourcefilter/statement_builder.go` and `telemetrylogs/statement_builder.go`.
>   - **Validation**:
>     - Updates `Validate` method in `validation.go` to support `RequestTypeRawStream` without end time.
>   - **Misc**:
>     - Updates tests in `stmt_builder_test.go` to include `RequestTypeRawStream` scenarios.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 158b516cfd3e057b539d4c74d942ede0a935274d. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


pkg/querier/api.go
```diff
@@ -85,6 +89,134 @@ func (a *API) QueryRange(rw http.ResponseWriter, req *http.Request) {
...
+func (a *API) QueryRawStream(rw http.ResponseWriter, req *http.Request) {
+	ctx := req.Context()
+
+	// get the param from url and add it to body
+	startQ := req.URL.Query().Get("start")
+	filterQ := req.URL.Query().Get("filter")
+
+	start, err := strconv.ParseUint(startQ, 10, 64)
+	if err != nil {
+		start = 0
```
srikanthccv, line 101: Would it make sense to throw instead of 0?

pkg/querier/api.go
```diff
@@ -85,6 +89,134 @@ func (a *API) QueryRange(rw http.ResponseWriter, req *http.Request) {
...
+
+			render.Error(rw, errors.NewInternalf(
+				errors.CodeInternal,
+				"Something went wrong on our end. It's not you, it's us. Our team is notified about it. Reach out to support if issue persists.",
+			))
+		}
+	}()
+
+	// just for validation as we don't use end
+	queryRangeRequest.End = queryRangeRequest.Start + 1
```
srikanthccv: Would it make sense to remove this, and instead skip end check for `RequestTypeRawStream` since we already skip certain validations there?

pkg/querier/api.go
srikanthccv: I see that later the now() is used, I guess this is fine.

pkg/querier/querier.go
```diff
@@ -294,6 +297,103 @@ func (q *querier) QueryRange(ctx context.Context, orgID valuer.UUID, req *qbtype
 	return qbResp, qbErr
 }
 
+func (q *querier) QueryRangeRawStream(ctx context.Context, orgID valuer.UUID, req *qbtypes.QueryRangeRequest, client *qbtypes.RawStream) {
```
srikanthccv: How about naming this also `QueryRawStream` since it doesn't really have a "range" unlike other queries.

pkg/querier/querier.go
```diff
@@ -294,6 +297,103 @@ func (q *querier) QueryRange(ctx context.Context, orgID valuer.UUID, req *qbtype
...
+		PanelType:       req.RequestType.StringValue(),
+	}
+
+	for idx, query := range req.CompositeQuery.Queries {
+		event.QueryType = query.Type.StringValue()
+		if query.Type == qbtypes.QueryTypeBuilder {
+			switch spec := query.Spec.(type) {
+			case qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]:
+				event.FilterApplied = spec.Filter != nil && spec.Filter.Expression != ""
+				req.CompositeQuery.Queries[idx].Spec = spec
```
srikanthccv: I think this line is not needed as it's re-assigning the same spec back.

pkg/querier/api.go
```diff
@@ -85,6 +89,134 @@ func (a *API) QueryRange(rw http.ResponseWriter, req *http.Request) {
...
+	}
+	// create the v5 request param
+	queryRangeRequest := qbtypes.QueryRangeRequest{
+		Start:       start,
+		RequestType: qbtypes.RequestTypeRawStream,
+		CompositeQuery: qbtypes.CompositeQuery{
+			Queries: []qbtypes.QueryEnvelope{
+				{
+					Type: qbtypes.QueryTypeBuilder,
+					Spec: qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]{
```
srikanthccv, line 111: Can you add `Signal` field spec so the `Fingerprint` of the ensures it return empty string to indicate caching is not supported.

srikanthccv: And perhaps some Name.

pkg/querier/querier.go
```diff
@@ -294,6 +297,103 @@ func (q *querier) QueryRange(ctx context.Context, orgID valuer.UUID, req *qbtype
...
+			}
+		} else {
+			// return if it's not of type query builder
+			client.Error <- errors.NewInvalidInputf(errors.CodeInvalidInput, "unsupported query type %s", query.Type)
+			return
+		}
+	}
+
+	queries := make(map[string]qbtypes.Query)
+	steps := make(map[string]qbtypes.Step)
```
srikanthccv: you can remove this and pass `nil` map since non-cachable types don't require step interval.

pkg/types/querybuildertypes/querybuildertypesv5/request_type.go
```diff
@@ -14,6 +14,8 @@ var (
 	RequestTypeTimeSeries = RequestType{valuer.NewString("time_series")}
 	// [][]any, SQL result set, but paginated, example: list view
 	RequestTypeRaw = RequestType{valuer.NewString("raw")}
+	// [][]any, SQL result set, but live, example: live list view
+	RequestTypeRawStream = RequestType{valuer.NewString("rawStream")}
```
srikanthccv: let's use `raw_stream` instead to remain consistent

pkg/querier/querier.go
```diff
@@ -50,15 +52,16 @@ func New(
...
+		logger:                 querierSettings.Logger(),
+		telemetryStore:         telemetryStore,
+		metadataStore:          metadataStore,
+		promEngine:             promEngine,
+		traceStmtBuilder:       traceStmtBuilder,
+		logStmtBuilder:         logStmtBuilder,
+		metricStmtBuilder:      metricStmtBuilder,
+		meterStmtBuilder:       meterStmtBuilder,
+		bucketCache:            bucketCache,
+		liveDataRefreshSeconds: 5,
```
srikanthccv, line 68: Do you see this to be configurable? If so, good idea to add here https://github.com/SigNoz/signoz/blob/main/pkg/querier/config.go#L10C1-L18C2

nityanandagohain: lets keep it hardcoded for now. I don't see myself configuring it somewhere.

pkg/querier/querier.go
```diff
@@ -294,6 +297,100 @@ func (q *querier) QueryRange(ctx context.Context, orgID valuer.UUID, req *qbtype
...
+	defer ticker.Stop()
+
+	// we are creating a custom ticker wrapper to trigger it instantly
+	tick := make(chan time.Time, 1)
+	tick <- time.Now() // initial tick
+	go func() {
+		for t := range ticker.C {
+			tick <- t
+		}
+	}()
```
tushar-signoz, line 377: How about something like this ?
Something like 
    timer := time.NewTimer(0)
    defer timer.Stop()

    for {
        select {
        case t := <-timer.C:
            doWork(t)

            if !timer.Stop() {. // To remove stale ticks, if query takes a huge amount of time.
                select { case <-timer.C: default: }
            }
            timer.Reset(d)

        case <-ctx.Done():
            return
        }
    }
    
Another note to be thought of is, what if the query to CH takes liveDataRefreshSeconds. The timer implementation make sure, that it waits +liveDataRefreshSeconds on top of whatever time query took.

pkg/querier/api.go
```diff
@@ -85,6 +89,134 @@ func (a *API) QueryRange(rw http.ResponseWriter, req *http.Request) {
...
+				},
+			},
+		},
+	}
+
+	claims, err := authtypes.ClaimsFromContext(ctx)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
```
tushar-signoz, line 148: We do have a midlleware to handle this, is there a reason to do it again ?

pkg/querier/querier.go
```diff
@@ -294,6 +297,100 @@ func (q *querier) QueryRange(ctx context.Context, orgID valuer.UUID, req *qbtype
...
+				continue
+			}
+			data := qbResp.Data.Results[0].(*qbtypes.RawData)
+			for i := len(data.Rows) - 1; i >= 0; i-- {
+				client.Logs <- data.Rows[i]
+				if i == 0 {
+					tsStart = uint64(data.Rows[i].Timestamp.UnixNano())
+					updatedLogID = data.Rows[i].Data["id"].(string)
+				}
+			}
```
tushar-signoz, line 411: Since in the query parameter we have specified, DirectionDesc. Should we be iterating from the end ? Im not sure what is the rendering logic is on the UI so you can correct me here. 

We have specified 500 limit in the query and we are using, last logID as a cursor, should we have check if we are lagging too much, it should skip some logs ?

joriskoris:
Is this getting addressed anytime soon? Trying to make a decision on whether I should revert to a previous working version.

nityanandagohain:
@joriskoris please wait for the next release on wednesday, the UI is getting revamped as well.

pkg/querier/querier.go
nityanandagohain: We want to send older logs first and then newer logs, it's always the latest 500 logs thats why limit 500. Eg from the last cursor to the current one more than 500 is ingested then the latest 500 will be sent back.

pkg/querier/api.go
nityanandagohain: So for using anything from the context we use this function that stores the data in context.

pkg/querier/querier.go
nityanandagohain: Thanks for the suggestion it seems like a good enhancement, but I am expecting this livetail query to be very fast as we are updating the timestamp and id every time. So we can pick this as an improvement later.

srikanthccv (approved):

PR #8816, opened 2025-08-16
chore: add multiple thresholds support

aniketio-ctrl:
**Added support for multiple thresholds in alerting**

Users can now create multiple thresholds. Currently, we support only basic thresholds.
warning : For previous alert one more label will get attached "threshold: critical"
Example for better understanding:
Suppose we define two thresholds:

Warning when the value is ≥ 2

Critical when the value is ≥ 15

Now, consider a query that returns three time series with constant values: 5, 18, and 24.

In this case, a single alert message will be generated containing three sections:

2 Critical alerts for values 18 and 24

1 Warning alert for value 5

Things remaining, handle threshold while creating template with unit

----

> [!IMPORTANT]
> Support for multiple thresholds in alerting added, allowing users to define different alert levels such as 'Warning' and 'Critical'.
> 
>   - **Behavior**:
>     - Support for multiple thresholds in alerting added, allowing users to define different alert levels (e.g., 'Warning', 'Critical').
>     - `RuleThresholdData` structure introduced to hold multiple thresholds.
>     - Alerting logic in `anomaly.go`, `base_rule.go`, and `manager.go` updated to handle multiple thresholds.
>     - Alerts now include a 'threshold.name' label to indicate the threshold level.
>   - **Tests**:
>     - New test cases added in `promrule_test.go` and `threshold_rule_test.go` to verify multiple threshold functionality.
>     - Existing tests updated to accommodate changes in alerting logic.
>   - **Misc**:
>     - Removed YAML support for rule parsing, now only JSON is supported.
>     - Minor refactoring and code cleanup in `api_params.go` and `threshold.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for ded8b346235333b8e1a0c82205fbb1318d8e2b79. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


pkg/query-service/rules/base_rule.go
```diff
@@ -85,6 +86,9 @@ type BaseRule struct {
 	TemporalityMap map[string]map[v3.Temporality]bool
 
 	sqlstore sqlstore.SQLStore
+
+	cronExpression string
+	cronEnabled    bool
```
srikanthccv: Please don't add unrelated changes.

pkg/types/ruletypes/alerting.go
```diff
@@ -105,6 +107,294 @@ const (
...
+	// convert the target value to the y-axis unit
+	value := unitConverter.Convert(converter.Value{
+		F: *b.target,
+		U: converter.Unit(b.targetUnit),
+	}, converter.Unit(b.ruleUnit))
+	return value.F
+}
+
+func (b BasicRuleThreshold) RecoveryTarget() float64 {
+	if b.recoveryTarget == nil { return 0 } else { return *b.recoveryTarget }
```
srikanthccv: 1. Why is "0"? How do you differentiate b/w the no recovery value vs recovery with 0 value?
2. Why is there no unit handling for recovery value?

pkg/types/ruletypes/alerting.go
```diff
@@ -105,6 +107,294 @@ const (
...
+
+	lbls = append(lbls, qslabels.Label{Name: "threshold", Value: b.name})
+
+	series.Points = removeGroupinSetPoints(series)
+
+	// nothing to evaluate
+	if len(series.Points) == 0 {
+		return alertSmpl, false
+	}
+
```
srikanthccv: Where is the `RequireMinPoints` check?

pkg/types/ruletypes/api_params.go
```diff
@@ -140,6 +140,11 @@ func ParseIntoRule(initRule PostableRule, content []byte, kind RuleDataKind) (*P
 		return nil, err
 	}
 
+	//added alerts v2 fields
+	if len(rule.RuleCondition.Thresholds) == 0 {
+		rule.RuleCondition.Thresholds = append(rule.RuleCondition.Thresholds,
+			NewBasicRuleThreshold(CriticalThresholdName, rule.RuleCondition.Target, nil, rule.RuleCondition.MatchType, rule.RuleCondition.CompareOp, rule.RuleCondition.SelectedQuery, rule.RuleCondition.TargetUnit, rule.RuleCondition.CompositeQuery.Unit))
```
srikanthccv: Why is it hardcoded with `CriticalThresholdName`?

pkg/query-service/rules/prom_rule.go
```diff
@@ -151,84 +151,86 @@ func (r *PromRule) Eval(ctx context.Context, ts time.Time) (interface{}, error)
 	var alerts = make(map[uint64]*ruletypes.Alert, len(res))
 
 	for _, series := range res {
-		l := make(map[string]string, len(series.Metric))
-		for _, lbl := range series.Metric {
-			l[lbl.Name] = lbl.Value
-		}
+		for _, ruleThreshold := range r.Thresholds() {
```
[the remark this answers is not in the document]
srikanthccv: Please explain, this is a valid question.

pkg/query-service/rules/threshold_rule_test.go
```diff
@@ -1708,6 +1710,7 @@ func TestThresholdRuleLogsLink(t *testing.T) {
 		postableRule.RuleCondition.Target = &c.target
 		postableRule.RuleCondition.CompositeQuery.Unit = c.yAxisUnit
 		postableRule.RuleCondition.TargetUnit = c.targetUnit
+		postableRule.RuleCondition.Thresholds = []ruletypes.RuleThreshold{ruletypes.NewBasicRuleThreshold(postableRule.AlertName, &c.target, nil, ruletypes.MatchType(c.matchType), ruletypes.CompareOp(c.compareOp), postableRule.RuleCondition.SelectedQuery, c.targetUnit, postableRule.RuleCondition.CompositeQuery.Unit)}
```
srikanthccv: Can you please add tests which makes use of the `ParseIntoRule` and test it? It would ensure that code path is covered.

pkg/types/ruletypes/alerting.go
```diff
@@ -105,6 +107,294 @@ const (
...
+func (b BasicRuleThreshold) ShouldAlert(series v3.Series) (Sample, bool) {
+	var shouldAlert bool
+	var alertSmpl Sample
+	var lbls qslabels.Labels
+
+	for name, value := range series.Labels {
+		lbls = append(lbls, qslabels.Label{Name: name, Value: value})
+	}
+
+	lbls = append(lbls, qslabels.Label{Name: "threshold", Value: b.name})
```
srikanthccv: use `threshold.name` instead

srikanthccv: Is this label going to be removed before notification message delivery? Or is it going to go into the notification content the end user receives?

pkg/types/ruletypes/alerting.go
srikanthccv: will recovery thresholds be implemented separately?

pkg/types/ruletypes/alerting.go
```diff
@@ -118,6 +408,109 @@ type RuleCondition struct {
...
+type BasicRuleThresholdJSON struct {
+	Name           string    `json:"name"`
+	Target         *float64  `json:"target"`
+	TargetUnit     string    `json:"targetUnit,omitempty"`
+	RuleUnit       string    `json:"ruleUnit,omitempty"`
+	RecoveryTarget *float64  `json:"recoveryTarget,omitempty"`
+	MatchType      MatchType `json:"matchType"`
+	CompareOp      CompareOp `json:"op"` // Note: using "op" to match existing format
+	SelectedQuery  string    `json:"selectedQuery,omitempty"`
+}
```
srikanthccv: Why is there no channel here? Because if one want to send warning and critical to different destinations, where would that information go?

pkg/types/ruletypes/alerting.go
aniketio-ctrl: yes, because that recovery threshold needs to send resolved alerts

pkg/types/ruletypes/api_params.go
aniketio-ctrl: because for previous alerts who dont use multi thresholds i have considered single threshold as crtitical

pkg/types/ruletypes/alerting.go
aniketio-ctrl: it woud be part of notification content to let user know which threshsold triggered this

pkg/types/ruletypes/alerting.go
aniketio-ctrl: it is implemented in notification routing pr, we will merge this first post that we will add that in notification routing

pkg/query-service/rules/prom_rule.go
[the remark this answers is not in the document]
aniketio-ctrl: done

pkg/types/ruletypes/alerting.go
aniketio-ctrl: was going to add with subsequent prs that will handle advanced options, but for now added it here

pkg/query-service/rules/threshold_rule_test.go
aniketio-ctrl: added the tests

pkg/types/ruletypes/alerting.go
aniketio-ctrl: no added it before threshold evaluation

pkg/types/ruletypes/api_params.go
srikanthccv: There was single threshold but which kind of threshold it is informed with Severity. 

Screenshot 2025-08-30 at 5 37 28 PM

pkg/query-service/rules/prom_rule.go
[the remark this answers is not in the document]
srikanthccv: I don't what the "done" here mean. I was expecting an explanation.

pkg/types/ruletypes/api_params.go
aniketio-ctrl: yeah we can add severity labels

aniketio-ctrl: done

pkg/query-service/rules/prom_rule.go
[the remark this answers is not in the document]
aniketio-ctrl: It was a miss, added the case for breaking the loop

pkg/types/ruletypes/api_params.go
srikanthccv: What happens to the existing alerts with severity in the UI with new revamp?

aniketio-ctrl: For existing alerts ,when we populate them at runtime in task they will be parsed with new threshold structures.

ee/query-service/rules/anomaly.go
```diff
@@ -253,9 +253,18 @@ func (r *AnomalyRule) buildAndRunQuery(ctx context.Context, orgID valuer.UUID, t
...
 	for _, series := range queryResult.AnomalyScores {
-		smpl, shouldAlert := r.ShouldAlert(*series)
-		if shouldAlert {
-			resultVector = append(resultVector, smpl)
+		if r.Condition() != nil && r.Condition().RequireMinPoints {
+			if len(series.Points) < r.Condition().RequiredNumPoints {
+				r.logger.InfoContext(ctx, "not enough data points to evaluate series, skipping", "ruleid", r.ID(), "numPoints", len(series.Points), "requiredPoints", r.Condition().RequiredNumPoints)
+				continue
+			}
+		}
```
srikanthccv, line 260: This check was part of the base rule that inside the `ShouldAlert`. Whether or not to alert was on the basis of the condition as a whole. What is the reason to move that to threshold and repeat `len(series.Points) < r.Condition().RequiredNumPoints` twice in the `ee/query-service/rules/anomaly.go` and `pkg/query-service/rules/threshold_rule.go` and soon in prom_rule.go

ee/query-service/rules/anomaly.go
```diff
@@ -253,9 +253,18 @@ func (r *AnomalyRule) buildAndRunQuery(ctx context.Context, orgID valuer.UUID, t
...
+		for _, threshold := range r.Thresholds() {
+			smpl, shouldAlert := threshold.ShouldAlert(*series)
+			if shouldAlert {
+				resultVector = append(resultVector, smpl)
+				break //return when one threshold is alerted
```
srikanthccv: Why is there a break here? Imagine the alert is setup to send warn on 80% and then on 95% critical. the following is how things went

hh:mm

6:10 - warning
6:29 - critical
6:30 - critical


The break/return here means we don't send the `warning` anymore which means the alert manager won't get the warning alert => the warning alert sent at 6:10 is now resolved which doesn't make sense because if it is continue to in critical, the warning state can't be resolved. 

It is the responsibility of the rule runner to send the alerts.

pkg/query-service/rules/base_rule.go
```diff
@@ -210,6 +211,32 @@ func (r *BaseRule) TargetVal() float64 {
...
+			// For "below" operations, sort ascending (lower values first)
+			return targetI < targetJ
+		default:
+			// For equal/not equal operations, use descending as default
+			return targetI > targetJ
+		}
+	})
+
+	return thresholds
+}
```
srikanthccv: Why is defined on the base rule?

pkg/query-service/rules/prom_rule.go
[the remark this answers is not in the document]
srikanthccv: https://github.com/SigNoz/signoz/pull/8816/files#r2331386581

pkg/query-service/rules/threshold_rule.go
```diff
@@ -484,9 +484,18 @@ func (r *ThresholdRule) buildAndRunQuery(ctx context.Context, orgID valuer.UUID,
...
+				r.logger.InfoContext(ctx, "not enough data points to evaluate series, skipping", "ruleid", r.ID(), "numPoints", len(series.Points), "requiredPoints", r.Condition().RequiredNumPoints)
+				continue
+			}
+		}
+		for _, threshold := range r.Thresholds() {
+			smpl, shouldAlert := threshold.ShouldAlert(*series)
+			if shouldAlert {
+				resultVector = append(resultVector, smpl)
+				break
+			}
```
srikanthccv: Same question as above the reason to repeat the `len(series.Points) < r.Condition().RequiredNumPoints` and have should alert on threshold as separate from the condition.

pkg/types/ruletypes/alerting.go
```diff
@@ -105,6 +107,298 @@ const (
...
+	Name() string
+	Target() float64
+	RecoveryTarget() float64
+
+	MatchType() MatchType
+	CompareOp() CompareOp
+
+	SelectedQuery() string
+	ShouldAlert(series v3.Series) (Sample, bool)
+}
```
srikanthccv: Why is this interface added? It is very focused for the basic rule threshold. Which defeats the purpose of interface.

pkg/types/ruletypes/alerting.go
```diff
@@ -118,6 +412,109 @@ type RuleCondition struct {
 	SelectedQuery     string             `json:"selectedQueryName,omitempty"`
 	RequireMinPoints  bool               `yaml:"requireMinPoints,omitempty" json:"requireMinPoints,omitempty"`
 	RequiredNumPoints int                `yaml:"requiredNumPoints,omitempty" json:"requiredNumPoints,omitempty"`
+	Thresholds        []RuleThreshold    `yaml:"thresholds,omitempty" json:"thresholds,omitempty"`
```
srikanthccv: I am unable to follow why is it an array of interfaces here?

srikanthccv: And one would have to call `NewX` to create a type that implements the interface which is very focused towards the "basic" threshold?

ee/query-service/rules/anomaly.go
aniketio-ctrl: Checking to be run on minimum no of points for each series should be part of evaluation, it shoudn't be the responsibility of threshold to do that, thats why they are part of buildAndRunQuery which therefore is the part of eval function, should alert should only have one responsibility to verify wether a threshold is breached or not.

pkg/query-service/rules/base_rule.go
aniketio-ctrl: Because of the current structure compareOp(), matchType() are also part of base rule, with future api changes of kind and spec structure it would be moved.

ee/query-service/rules/anomaly.go
srikanthccv: > it shoudn't be the responsibility of threshold to do that

I didn't say it should be responsibility of threshold. I said "Whether or not to alert was on the basis of the condition as a whole"

> which therefore is the part of eval function, should alert should only have one responsibility to verify wether a threshold is breached or not
 
Where is the "eval function" you refer to? All of the logic is clubbed under `buildAndRunQuery`.

pkg/query-service/rules/base_rule.go
srikanthccv: >Because of the current structure compareOp(), matchType() are also part of base rule, with future api changes of kind and spec structure it would be moved.

Please share where it would be moved.

ee/query-service/rules/anomaly.go
aniketio-ctrl: > Where is the "eval function" you refer to? All of the logic is clubbed under buildAndRunQuery.

r.Eval( specific rule eval function) calls buildAndRunQuery 

> Whether or not to alert was on the basis of the condition as a whole. What is the reason to move that to threshold and repeat

if i understand correctly regarding this, you are asking the reason of moving requireMinPoint check to threshold and repeat, Its not part of threshold , we have moved it from threshold to Eval function -> build and run query, because of the reason that shoudlAlert be only have one responsibility to check wether or not the threshold is breached or not.

pkg/query-service/rules/base_rule.go
aniketio-ctrl: it would be part of basic rule threshold.

pkg/types/ruletypes/alerting.go
aniketio-ctrl: So that in future if the requirements arise to have multiple threshold with different match and compare operator

pkg/types/ruletypes/threshold.go
```diff
@@ -0,0 +1,374 @@
...
+	"github.com/SigNoz/signoz/pkg/query-service/utils/labels"
+	"math"
+	"sort"
+)
+
+type ThresholdKind string
+
+const (
+	BasicThresholdKind ThresholdKind = "basic"
+)
```
srikanthccv: Let's use the `valuer.String`

```
type ThresholdKind struct {
	valuer.String
}

var (
	BasicThresholdKind       = ThresholdKind{valuer.NewString("basic")}
)

```

pkg/types/ruletypes/threshold.go
```diff
@@ -0,0 +1,374 @@
...
+
+type RuleThresholdData struct {
+	Kind ThresholdKind `json:"kind"`
+	Spec any           `json:"spec"`
+}
+
+func (r *RuleThresholdData) UnmarshalJSON(data []byte) error {
+	var raw map[string]json.RawMessage
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
```
srikanthccv: Use the errors package with more detailed message

```
			return errors.NewInvalidInputf(
				errors.CodeInvalidInput,
				"......",
				err,
			)
```

pkg/types/ruletypes/threshold.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+	if err := json.Unmarshal(raw["kind"], &r.Kind); err != nil {
+		return err
```
srikanthccv: Same, don't return the naked errors.

pkg/types/ruletypes/threshold.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+	switch r.Kind {
+	case BasicThresholdKind:
+		var basicThresholds BasicRuleThresholds
+		if err := json.Unmarshal(raw["spec"], &basicThresholds); err != nil {
+			return fmt.Errorf("failed to unmarshal basic thresholds: %v",
```
srikanthccv: Same

pkg/types/ruletypes/threshold.go
```diff
@@ -0,0 +1,374 @@
...
+
+	default:
+		return fmt.Errorf("unknown threshold kind: %v", r.Kind)
+	}
+
+	return nil
+}
+
+type RuleThreshold interface {
+	ShouldAlert(series v3.Series) (Vector, error)
```
srikanthccv, line 54: Can we use `qbtypes.TimeSeries` in this implementation?

pkg/types/ruletypes/threshold.go
```diff
@@ -0,0 +1,374 @@
...
+	var result []v3.Point
+	for _, s := range series.Points {
+		if s.Timestamp >= 0 && !math.IsNaN(s.Value) && !math.IsInf(s.Value, 0) {
+			result = append(result, s)
+		}
+	}
+	return result
+}
+
+func (b BasicRuleThreshold) ShouldAlert(series v3.Series) (Sample, bool) {
```
srikanthccv, line 184: @srikanthccv to compare against existing implementation.

pkg/types/ruletypes/threshold.go
```diff
@@ -0,0 +1,374 @@
...
+			if series.Points[len(series.Points)-1].Value != b.Target() {
+				shouldAlert = true
+			}
+		}
+	}
+	return alertSmpl, shouldAlert
+}
+
+// BASE RULE WILL HAVE NO IDEA EXCEPTT RULETHREHSOLDS
+func (r *RuleThresholdData) GetRuleThreshold() (RuleThreshold, error) {
```
srikanthccv, line 405: No naked errors here also.

pkg/query-service/rules/base_rule.go
```diff
@@ -32,6 +32,8 @@ type BaseRule struct {
 	typ ruletypes.AlertType
 
 	ruleCondition *ruletypes.RuleCondition
+
+	Threshold ruletypes.RuleThreshold `yaml:"thresholds,omitempty" json:"thresholds,omitempty"`
```
srikanthccv: Why is there ``yaml:"thresholds,omitempty" json:"thresholds,omitempty"``?

pkg/types/ruletypes/threshold.go
srikanthccv: nvm, we keep the `v3.Series` till we get rid of v4 alerts completely.

pkg/types/ruletypes/threshold.go
aniketio-ctrl: added specific errors

pkg/types/ruletypes/threshold.go
srikanthccv: The code is exactly same

- the `RequireMinPoints` remain in the builder_query but not threshold - to be refactored with eval
- New `threshold.name` is added to labels.

srikanthccv:
The label `threshold.name` is added to the label set, which in turn is getting added to the related logs/traces link we generate.
- We should fix that
- Before we fix that, we need to understand why tests didn't catch that and update tests to ensure the links are correct

pkg/types/ruletypes/alerting.go
```diff
@@ -188,10 +188,7 @@ func (rc *RuleCondition) IsValid() bool {
 	}
 
 	if rc.QueryType() == v3.QueryTypeBuilder {
-		if rc.Target == nil {
-			return false
-		}
-		if rc.CompareOp == "" {
+		if rc.Thresholds == nil {
```
srikanthccv, line 191: What ensures that `Thresholds` condition is never nil for old alerts.

pkg/query-service/rules/manager.go
```diff
@@ -798,10 +798,12 @@ func (m *Manager) ListRuleStates(ctx context.Context) (*ruletypes.GettableRules,
 	for _, s := range storedRules {
 
 		ruleResponse := &ruletypes.GettableRule{}
-		if err := json.Unmarshal([]byte(s.Data), ruleResponse); err != nil { // Parse []byte to go struct pointer
+		postableRule, err := ruletypes.ParsePostableRule([]byte(s.Data))
+		if err != nil {
 			zap.L().Error("failed to unmarshal rule from db", zap.String("id", s.ID.StringValue()), zap.Error(err))
 			continue
 		}
+		ruleResponse.PostableRule = *postableRule
```
srikanthccv: Are there any other places like this?

aniketio-ctrl:
> The label `threshold.name` is added to the label set, which in turn is getting added to the related logs/traces link we generate.
> 
> * We should fix that
> * Before we fix that, we need to understand why tests didn't catch that and update tests to ensure the links are correct

yeah nice catch, checking that

pkg/types/ruletypes/alerting.go
aniketio-ctrl: Every older alerts will get parsed to postableRule via parsedIntoRule function which ensures, that thresholds field cant be nill

pkg/query-service/rules/manager.go
aniketio-ctrl: no there aren't any other places like this

pkg/types/ruletypes/alerting.go
aniketio-ctrl: One more thing that we can do is overwrite unmarshal json function for postablerule that will always make a check of handling thresholds

pkg/query-service/rules/manager.go
aniketio-ctrl: they are used in string() function of individual alerts but we are not using those

aniketio-ctrl: there were some missed places added there

aniketio-ctrl: removed yaml support and added validation inside json marshalling

PR #8836, opened 2025-08-18
fix: set source on add new query 

YounixM:
set source on add new query - meter explorer
fix light mode UI issues


frontend/src/providers/QueryBuilder.tsx
```diff
@@ -522,6 +523,8 @@ export function QueryBuilderProvider({
 		setCurrentQuery((prevState) => {
 			if (prevState.builder.queryData.length >= MAX_QUERIES) return prevState;
 
+			console.log('prevState', prevState.builder.queryData);
```
SagarRajput-7, line 526: remove console.log

PR #8855, opened 2025-08-19
fix(comment): add a dedicated comment parsing middleware

grandwizard28:
## 📄 Summary

add a dedicated comment parsing middleware. This removes duplication and double parsing of referrer.


therealpandey:
Resolves https://github.com/SigNoz/platform-pod/issues/1100

pkg/query-service/rules/prom_rule_task.go
```diff
@@ -369,12 +369,11 @@ func (g *PromRuleTask) Eval(ctx context.Context, ts time.Time) {
...
 			}(time.Now())
 
-			kvs := map[string]string{
-				"alertID": rule.ID(),
-				"source":  "alerts",
-				"client":  "query-service",
-			}
-			ctx = context.WithValue(ctx, common.LogCommentKey, kvs)
+			comment := ctxtypes.CommentFromContext(ctx)
+			comment.Set("alertID", rule.ID())
```
srikanthccv: change this to `alert_id` in consistency with `dashboardID` -> `dashboard_id` change?

pkg/types/ctxtypes/comment.go
```diff
@@ -0,0 +1,146 @@
...
+	widgetIDRegex        = regexp.MustCompile(`widgetId=([a-f0-9\-]+)`)
+	ruleRegex            = regexp.MustCompile(`/alerts/(new|edit)(?:\?.*)?$`)
+	ruleIDRegex          = regexp.MustCompile(`ruleId=(\d+)`)
+)
+
+type commentCtxKey struct{}
+
+type Comment struct {
+	vals map[string]string
+	mtx  sync.RWMutex
```
srikanthccv, line 27: Context values are usually immutable. It would be better to disallow the mutation and provide a way to create new comment with additional kvs.

srikanthccv (commented):
`email` and `client` in the query log are really helpful for understanding which user initiated the request and if they are coming from browser or API request. This helps us reach out if we need to communicate with them. Can we add them as well?

pkg/types/ctxtypes/comment.go
therealpandey: How do you want to handle same keys?

srikanthccv: I believe we can override. I can't think of any legitimate case (in this comment middleware context) where override is not the expected.

therealpandey: Addressed

pkg/query-service/rules/prom_rule_task.go
therealpandey: Addressed

srikanthccv (dismissed):

PR #8887, opened 2025-08-22
fix: use lower and convert re2 to string in fulltext

nityanandagohain:
* uses lower for index utilization
* if the string is invalid re2 then escape it.


pkg/querybuilder/fallback_expr.go
```diff
@@ -185,3 +186,12 @@ func FormatValueForContains(value any) string {
 		}
 	}
 }
+
+func FormatFullTextSearch(input string) string {
+	_, err := regexp.Compile(input)
+	if err != nil {
+		// Not a valid regex -> treat as literal substring
+		return regexp.QuoteMeta(input)
+	}
```
srikanthccv: nit:
```suggestion
    if _, err := regexp.Compile(input); err != nil {
        return regexp.QuoteMeta(input)
    }
```

pkg/telemetrylogs/condition_builder.go
```diff
@@ -99,7 +99,12 @@ func (c *conditionBuilder) conditionFor(
 		return sb.NotILike(tblFieldName, fmt.Sprintf("%%%s%%", value)), nil
 
 	case qbtypes.FilterOperatorRegexp:
-		return fmt.Sprintf(`match(%s, %s)`, tblFieldName, sb.Var(value)), nil
+		// check if the field is fullText, if yes then convert value to lower
+		valueStr := sb.Var(value)
+		if tblFieldName == DefaultFullTextColumn.Name {
+			valueStr = fmt.Sprintf("LOWER(%s)", valueStr)
+		}
```
srikanthccv: Can we move this logic to https://github.com/SigNoz/signoz/blob/eca13075e9f40f0ee2e001056f3cc6f536fc902e/pkg/telemetrylogs/condition_builder.go#L61-L68?

I think, then you can also remove the need to define `LOWER(body)` as column/field and the fields list can be just a list of actual columns on table.

srikanthccv (commented):
just one question around the need for `LOWER(body)` as a column and field otherwise looks good to me.

PR #8898, opened 2025-08-23
chore(meter): added product analytics for meter module

vikrantgupta25:
- added source in the query range requests and module name in the comment middleware

contributes to - https://github.com/SigNoz/platform-pod/issues/859

### DIFF
--- a/frontend/src/api/v5/queryRange/prepareQueryRangePayloadV5.ts
+++ b/frontend/src/api/v5/queryRange/prepareQueryRangePayloadV5.ts
@@ -66,9 +66,46 @@ function getSignalType(dataSource: string): 'traces' | 'logs' | 'metrics' {
 	return 'metrics';
 }
 
-/**
- * Creates base spec for builder queries
- */
+function isDeprecatedField(fieldName: string): boolean {
+	const deprecatedIntrinsicFields = [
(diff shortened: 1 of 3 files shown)


pkg/types/querybuildertypes/querybuildertypesv5/resp.go
```diff
@@ -19,6 +19,7 @@ type QBEvent struct {
 	LogsUsed        bool   `json:"logs_used,omitempty"`
 	MetricsUsed     bool   `json:"metrics_used,omitempty"`
 	TracesUsed      bool   `json:"traces_used,omitempty"`
+	Source          string `json:"source,omitempty"`
```
srikanthccv, line 22: https://github.com/SigNoz/signoz/blob/b5b513f1e0cf78a582c7343a826f8eb77e0a11dc/pkg/querier/api.go#L153-L162

Adding here doesn't automatically add to segment event. I think I did that to avoid marshal and then unmarshal to map any or something. either add or refactor to add source to event.

vikrantgupta25: added in the map for now, a little misleading as it seems all the properties of the event should be sent

Issue #8819, opened 2025-08-17
Do not sent type of the default columns as "attribute"


srikanthccv:
Issue: Trace field `Name` display N/A and can not filter using field Name.
Version: 0.92.1

Reason: Both field `name` and `attributes_string['name']` exists.
```                    "message": "Key `name` is ambiguous, found 2 different combinations of field context / data type: [name=name,context=attribute,datatype=string name=name,context=span,datatype=string]."```
Do you think this is a bug? I believe field `Name` should fallback to span context.

Ref: <https://signoz.io/docs/userguide/field-context-data-types/>

[Slack Message](https://signoz-community.slack.com/archives/C01HWQ1R0BC/p1755336730623109?thread_ts=1755336730.623109&cid=C01HWQ1R0BC)

The default list of selected columns sends the `fieldContext` as `attribute`. When an attribute conflicts with the selected column, it returns N/A because the value doesn't exist. Remove the fieldContext as an attribute for the default column in traces explorer.

PR #8902, opened 2025-08-24
chore: do not send field context as tag for deprecated fields

srikanthccv:
## 📄 Summary

Fixes https://github.com/SigNoz/signoz/issues/8819

----

> [!IMPORTANT]
> `createBaseSpec` in `prepareQueryRangePayloadV5.ts` now omits `fieldContext` for deprecated fields using the new `isDeprecatedField` function.
> 
>   - **Behavior**:
>     - In `createBaseSpec` in `prepareQueryRangePayloadV5.ts`, `fieldContext` is no longer added for deprecated fields.
>     - Introduces `isDeprecatedField` function to identify deprecated fields.
>   - **Functions**:
>     - Adds `isDeprecatedField` to check for deprecated intrinsic and calculated fields.
>     - Updates `createBaseSpec` to conditionally add `fieldContext` based on deprecation status.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 59e0166731c81bd504356bf9cbf25f2784e62281. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/src/api/v5/queryRange/prepareQueryRangePayloadV5.ts
+++ b/frontend/src/api/v5/queryRange/prepareQueryRangePayloadV5.ts
@@ -66,9 +66,46 @@ function getSignalType(dataSource: string): 'traces' | 'logs' | 'metrics' {
 	return 'metrics';
 }
 
-/**
- * Creates base spec for builder queries
- */
+function isDeprecatedField(fieldName: string): boolean {
+	const deprecatedIntrinsicFields = [
+		'traceID',
+		'spanID',
+		'parentSpanID',
+		'spanKind',
+		'durationNano',
+		'statusCode',
+		'statusMessage',
+		'statusCodeString',
+	];
+
+	const deprecatedCalculatedFields = [
+		'responseStatusCode',
+		'externalHttpUrl',
+		'httpUrl',
+		'externalHttpMethod',
+		'httpMethod',
+		'httpHost',
+		'dbName',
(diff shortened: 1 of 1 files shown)


frontend/src/api/v5/queryRange/prepareQueryRangePayloadV5.ts
```diff
@@ -140,16 +177,29 @@ function createBaseSpec(
...
+
+						const fieldObj: TelemetryFieldKey = {
+							name: fieldName,
+							fieldDataType:
+								column?.fieldDataType ?? (column?.dataType as FieldDataType),
+							signal: column?.signal ?? undefined,
+						};
+
+						// Only add fieldContext if the field is NOT deprecated
+						if (!isDeprecated && fieldName !== 'name') {
```
SagarRajput-7, line 192: Why does the 'name' need to be separately handled?

srikanthccv: `name` is a standard intrinsic field (not deprecated), but when we send it with `fieldContext: "tag"` in the payload, the system incorrectly tries to fetch it as an attribute value instead of the actual span name, resulting in empty values being returned. This is why we treat it here.

PR #8936, opened 2025-08-28
Fix: API for Download Buttton 

### LINKED ISSUE #5701: Ability to download all matching logs
The ability to download logs has several practical benefits. For instance, it allows attaching complete log records to tickets or issues, ensuring that future team members working on the problem have access to all relevant information without worrying about retention. A good example of this can be seen in the GitHub issue https://github.com/SigNoz/engineering-pod/issues/1508. There are valid use cases for downloading all logs that match specific criteria.

However, if users are primarily interested in further analysis or exploration of the logs, it might indicate areas where our product could be improved. In such cases, we should actively seek user feedback to enhance our current product and better meet their needs

tushar-signoz:
This pull request introduces a new export API endpoint, enhances the telemetry log query to support selecting specific fields, and adds utility and validation functions for telemetry field keys. The main changes are grouped below.

### API Enhancements

* Added a new export endpoint at `/api/v1/export` with open access, registered in the API router (`api.go`).

### Telemetry Log Query Improvements

* Modified the log query statement builder to support selecting specific fields via the `SelectFields` parameter, always including `id` and `timestamp` by default (`statement_builder.go`).

### Telemetry Field Key Utilities & Validation

* Added a `ValidateFieldKeyText` function to validate field key text formats and their components using regex, ensuring correct field context and data type (`field.go`).
* Introduced `GetKeyTextFromFieldKey` utility to generate a key text string from a `TelemetryFieldKey` struct (`field.go`).
* Imported the `regexp` package to support regex-based validation in field key handling (`field.go`).## 📄 Summary

---

## ✅ Changes

---

ex:

- `frontend`
- `backend`
- `devops`
- `bug`
- `enhancement`
- `ui`
- `test`

---

## 👥 Reviewers

- frontend / backend / devops

---

## 🧪 How to Test

1. ...
2. ...
3. ...

---

## 🔍 Related Issues

Fixes #5701 

---

---

## 📋 Checklist

- [ ] Dev Review

---

### DIFF
--- a/ee/query-service/app/api/api.go
+++ b/ee/query-service/app/api/api.go
@@ -102,6 +102,9 @@ func (ah *APIHandler) RegisterRoutes(router *mux.Router, am *middleware.AuthZ) {
 	router.HandleFunc("/api/v1/billing", am.AdminAccess(ah.getBilling)).Methods(http.MethodGet)
 	router.HandleFunc("/api/v1/portal", am.AdminAccess(ah.LicensingAPI.Portal)).Methods(http.MethodPost)
 
+	// export
+	router.HandleFunc("/api/v1/export", am.OpenAccess(ah.Export)).Methods(http.MethodGet)
+
 	// v3
 	router.HandleFunc("/api/v3/licenses", am.AdminAccess(ah.LicensingAPI.Activate)).Methods(http.MethodPost)
 	router.HandleFunc("/api/v3/licenses", am.AdminAccess(ah.LicensingAPI.Refresh)).Methods(http.MethodPut)
--- a/ee/query-service/app/api/export.go
+++ b/ee/query-service/app/api/export.go
@@ -1,0 +1,429 @@
+package api
+
+import (
+	"compress/gzip"
+	"encoding/csv"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"runtime"
+	"runtime/debug"
+	"strconv"
+	"strings"
+
+	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/http/render"
+	"github.com/SigNoz/signoz/pkg/types/authtypes"
+	"github.com/SigNoz/signoz/pkg/types/telemetrytypes"
+	"github.com/SigNoz/signoz/pkg/valuer"
+
+	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
+)
+
+const MAX_EXPORT_ROW_COUNT_LIMIT = 50000
+const DEFAULT_EXPORT_ROW_COUNT_LIMIT = 10000
+const MAX_EXPORT_BYTES_LIMIT = 10 * 1024 * 1024 * 1024 // 10 GB
+const CHUNK_SIZE = 1000
+
+func (aH *APIHandler) Export(rw http.ResponseWriter, req *http.Request) {
+
+	queryParams := req.URL.Query()
+
+	startTime, endTime, err := getExportQueryTimeRange(queryParams)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	limit, err := getExportQueryLimit(queryParams)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	format, err := getExportQueryFormat(queryParams)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	filterExpression := queryParams.Get("filter")
+
+	selectColumns, err := getExportQueryColumns(queryParams)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	source, err := getExportQuerySource(queryParams)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	claims, err := authtypes.ClaimsFromContext(req.Context())
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	flusher, ok := rw.(http.Flusher)
+	if !ok {
+		render.Error(rw, errors.Newf(errors.TypeUnsupported, errors.CodeUnsupported, "export is not supported"))
+		return
+	}
+
+	defer func() {
+		if r := recover(); r != nil {
+			stackTrace := string(debug.Stack())
+
+			aH.Signoz.Instrumentation.Logger().ErrorContext(req.Context(), "panic in export",
+				"error", r,
+				"user", claims.UserID,
+				"stacktrace", stackTrace,
+			)
+
+			render.Error(rw, errors.NewInternalf(
+				errors.CodeInternal,
+				"Something went wrong on our end. It's not you, it's us. Our team is notified about it. Reach out to support if issue persists.",
+			))
+		}
+	}()
+
+	queryRangeRequest := qbtypes.QueryRangeRequest{
+		Start:       startTime,
+		End:         endTime,
+		RequestType: qbtypes.RequestTypeRaw,
+		CompositeQuery: qbtypes.CompositeQuery{
+			Queries: []qbtypes.QueryEnvelope{
+				{
+					Type: qbtypes.QueryTypeBuilder,
+					Spec: nil,
+				},
+			},
+		},
+	}
+
+	switch source {
+	case "logs":
+		spec := qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]{
+			Signal: telemetrytypes.SignalLogs,
+			Name:   "raw",
+			Filter: &qbtypes.Filter{
+				Expression: filterExpression,
+			},
+			Limit:        limit,
+			SelectFields: selectColumns,
+			Order: []qbtypes.OrderBy{
+				{
+					Direction: qbtypes.OrderDirectionAsc,
+					Key: qbtypes.OrderByKey{
+						TelemetryFieldKey: telemetrytypes.TelemetryFieldKey{
+							Name:         "timestamp",
+							Materialized: true,
+						},
+					},
+				},
+				{
+					Direction: qbtypes.OrderDirectionAsc,
+					Key: qbtypes.OrderByKey{
+						TelemetryFieldKey: telemetrytypes.TelemetryFieldKey{
+							Name:         "id",
+							Materialized: true,
+						},
+					},
+				},
+			},
+		}
+
+		queryRangeRequest.CompositeQuery.Queries[0].Spec = spec
+		err = aH.exportLogs(rw, req, &queryRangeRequest, format)
+		if err != nil {
+			render.Error(rw, err)
+			return
+		}
+
+	}
+
+	flusher.Flush()
+}
+
+func (aH *APIHandler) exportLogs(rw http.ResponseWriter, req *http.Request, rangeRequest *qbtypes.QueryRangeRequest, format string) error {
+
+	// Set up response headers
+	rw.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=\"data_exported.%s.gz\"", format))
+	rw.Header().Set("Cache-Control", "no-cache")
+	rw.Header().Set("Content-Encoding", "gzip")
+
+	rw.Header().Add("Trailer", "X-Total-Bytes, X-Total-Rows, X-Response-Complete")
+
+	spec := rangeRequest.CompositeQuery.Queries[0].Spec.(qbtypes.QueryBuilderQuery[qbtypes.LogAggregation])
+	rowCountLimit := spec.Limit
+
+	// Set up the gzip writer first
+	gzipWriter := gzip.NewWriter(rw)
+
+	// Get claims for querier
+	claims, err := authtypes.ClaimsFromContext(req.Context())
+	if err != nil {
+		return err
+	}
+
+	// Create orgID
+	orgID, err := valuer.NewUUID(claims.OrgID)
+	if err != nil {
+		return err
+	}
+
+	var (
+		csvWriter *csv.Writer
+		encoder   *json.Encoder
+	)
+
+	switch format {
+	case "csv":
+		rw.Header().Set("Content-Type", "text/csv")
+		csvWriter = csv.NewWriter(gzipWriter)
+
+	case "jsonl":
+		rw.Header().Set("Content-Type", "application/x-ndjson")
+		encoder = json.NewEncoder(gzipWriter)
+		encoder.SetEscapeHTML(false)
+	}
+
+	rowChan := make(chan *qbtypes.RawRow, runtime.NumCPU())
+	errChan := make(chan error, 1)
+	doneChan := make(chan any)
+
+	go func() {
+		defer close(doneChan)
+
+		var header []string
+
+		headerToIndexMapping := make(map[string]int, len(header))
+
+		for row := range rowChan {
+			switch format {
+			case "csv":
+				if header == nil {
+					// Initialize and write header for CSV
+
+					header = constructCSVHeaderFromQueryResponse(row.Data)
+
+					if err := csvWriter.Write(header); err != nil {
+						errChan <- fmt.Errorf("error writing CSV header: %w", err)
+						return
+					}
+
+					for i, col := range header {
+						headerToIndexMapping[col] = i
+					}
+				}
+				record := constructCSVRecordFromQueryResponse(row.Data, headerToIndexMapping)
+
+				if err := csvWriter.Write(record); err != nil {
+					errChan <- fmt.Errorf("error writing CSV record: %w", err)
+					return
+				}
+			case "jsonl":
+				// Handle JSON format (JSONL - one object per line)
+				if err := encoder.Encode(row.Data); err != nil {
+					errChan <- fmt.Errorf("error encoding JSON: %w", err)
+					return
+				}
+			}
+		}
+
+		if csvWriter != nil {
+			csvWriter.Flush()
+			if err := csvWriter.Error(); err != nil {
+				errChan <- fmt.Errorf("error flushing CSV writer: %w", err)
+				return
+			}
+		}
+
+		if gzipWriter != nil {
+			if err := gzipWriter.Flush(); err != nil {
+				errChan <- fmt.Errorf("error flushing gzip writer: %w", err)
+				return
+			}
+			if err := gzipWriter.Close(); err != nil {
+				errChan <- fmt.Errorf("error closing gzip writer: %w", err)
+				return
+			}
+		}
+
+	}()
+
+	totalBytes := uint64(0)
+	rowCount := 0
+
+	cursor := ""
+	for rowCount < rowCountLimit && totalBytes < MAX_EXPORT_BYTES_LIMIT {
+		spec.Limit = min(CHUNK_SIZE, rowCountLimit-rowCount)
+		rangeRequest.CompositeQuery.Queries[0].Spec = spec
+
+		if cursor != "" {
+			spec.Cursor = cursor
+		}
+
+		response, err := aH.Signoz.Querier.QueryRange(req.Context(), orgID, rangeRequest)
+		if err != nil {
+			errChan <- err
+			break
+		}
+
+		for _, result := range response.Data.Results {
+			resultData, ok := result.(*qbtypes.RawData)
+			if !ok {
+				return errors.NewInvalidInputf(errors.CodeInvalidInput, "expected RawData, got %T", result)
+			}
+
+			cursor = resultData.NextCursor
+
+			for _, row := range resultData.Rows {
+				select {
+				case rowChan <- row:
+				case <-req.Context().Done():
+					return req.Context().Err()
+				}
+			}
+
+			rowCount += len(resultData.Rows)
+		}
+		totalBytes += response.Meta.BytesScanned
+
+	}
+	rw.Header().Set("X-Total-Rows", strconv.Itoa(rowCount))
+	rw.Header().Set("X-Total-Bytes", strconv.FormatUint(totalBytes, 10))
+	rw.Header().Set("X-Response-Complete", strconv.FormatBool(rowCount < rowCountLimit))
+	close(rowChan)
+
+	// Wait for completion or error
+	select {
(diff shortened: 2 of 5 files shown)


ee/query-service/app/api/export.go
```diff
@@ -0,0 +1,429 @@
+package api
```
nityanandagohain: please move this package api and logic to `pkg/modules` we are refactoring and moving things there. You can check the example package `pkg/modules/user`

ee/query-service/app/api/api.go
```diff
@@ -102,6 +102,9 @@ func (ah *APIHandler) RegisterRoutes(router *mux.Router, am *middleware.AuthZ) {
 	router.HandleFunc("/api/v1/billing", am.AdminAccess(ah.getBilling)).Methods(http.MethodGet)
 	router.HandleFunc("/api/v1/portal", am.AdminAccess(ah.LicensingAPI.Portal)).Methods(http.MethodPost)
 
+	// export
+	router.HandleFunc("/api/v1/export", am.OpenAccess(ah.Export)).Methods(http.MethodGet)
```
nityanandagohain: let's put this endpoint for both ee and community, add it to http_handler.go. Also let's add viewAccess instead of openAccess. As openAccess will allow downloding without any authentication key.

ee/query-service/app/api/export.go
```diff
@@ -0,0 +1,429 @@
...
+			Signal: telemetrytypes.SignalLogs,
+			Name:   "raw",
+			Filter: &qbtypes.Filter{
+				Expression: filterExpression,
+			},
+			Limit:        limit,
+			SelectFields: selectColumns,
+			Order: []qbtypes.OrderBy{
+				{
+					Direction: qbtypes.OrderDirectionAsc,
```
nityanandagohain: this is a thing to consider, if we use ascending and the limit hits then I would get the starting logs and not the latest logs. wich might not be an ideal case. Let's start with desc ?

ee/query-service/app/api/export.go
```diff
@@ -0,0 +1,429 @@
...
+		rw.Header().Set("Content-Type", "text/csv")
+		csvWriter = csv.NewWriter(gzipWriter)
+
+	case "jsonl":
+		rw.Header().Set("Content-Type", "application/x-ndjson")
+		encoder = json.NewEncoder(gzipWriter)
+		encoder.SetEscapeHTML(false)
+	}
+
+	rowChan := make(chan *qbtypes.RawRow, runtime.NumCPU())
```
nityanandagohain: why is the size NumCPU?

ee/query-service/app/api/export.go
```diff
@@ -0,0 +1,429 @@
...
+
+		for _, result := range response.Data.Results {
+			resultData, ok := result.(*qbtypes.RawData)
+			if !ok {
+				return errors.NewInvalidInputf(errors.CodeInvalidInput, "expected RawData, got %T", result)
+			}
+
+			cursor = resultData.NextCursor
+
+			for _, row := range resultData.Rows {
```
nityanandagohain: break if empty rows?

ee/query-service/app/api/export.go
```diff
@@ -0,0 +1,429 @@
...
+						TelemetryFieldKey: telemetrytypes.TelemetryFieldKey{
+							Name:         "id",
+							Materialized: true,
+						},
+					},
+				},
+			},
+		}
+
+		queryRangeRequest.CompositeQuery.Queries[0].Spec = spec
```
nityanandagohain: ctx := context.WithValue(ctx, "clickhouse_max_threads", 2)

ee/query-service/app/api/export.go
tushar-signoz: Sure, lets discuss on this. Haven't changed this yet

ee/query-service/app/api/export.go
tushar-signoz: Modified it to 1, it was from an earlier implementation

ee/query-service/app/api/export.go
tushar-signoz: Loop logic will take care of it and later there is a counter to break

ee/query-service/app/api/export.go
tushar-signoz: "clickhouse_max_threads" is there a const declared for this ?

pkg/modules/export/implexport/handler.go
```diff
@@ -0,0 +1,494 @@
...
+//   - columns (optional): Specific columns to include in export
+//     Default: ["timestamp", "id", "body"]
+//     Format: ["context.field:type", "context.field", "field"]
+//
+//   - order_by (optional): Sorting specification ["column:direction" or "context.field:type:direction"]
+//     Direction: "asc" or "desc"
+//     Default: ["timestamp:asc", "id:asc"]
+//
+// Response Headers:
+//   - Content-Type: "text/csv" or "application/jsonl"
```
nityanandagohain: nit: are these helpful or unnecessary ?

pkg/modules/export/implexport/handler.go
```diff
@@ -0,0 +1,494 @@
...
+		handler.exportLogs(rw, r)
+	case "traces":
+		handler.exportTraces(rw, r)
+	case "metrics":
+		handler.exportMetrics(rw, r)
+	default:
+		render.Error(rw, errors.NewInvalidInputf(errors.CodeInvalidInput, "invalid source: must be logs"))
+	}
+
+	flusher, ok := rw.(http.Flusher)
```
nityanandagohain: dont think this is required anymore

pkg/modules/export/implexport/handler.go
```diff
@@ -0,0 +1,494 @@
...
+		totalBytes += uint64(len(str))
+	}
+	return totalBytes
+}
+
+// getExportQueryOrderBy parses the "order_by" query parameters and returns a slice of OrderBy structs.
+// Each "order_by" parameter should be in the format "column:direction"
+// Each "column" should be a valid telemetry field key in the format "context.field:type" or "context.field" or "field"
+func getExportQueryOrderBy(queryParams url.Values) (orderBy []qbtypes.OrderBy, err error) {
+	orderByParams := queryParams["order_by"]
```
nityanandagohain: we can keep this simple by just accepting one ordery by.

when the user selects timestamp as order by key in that case we add `id` as a secondary order by key internally ( unknown to user). for other atttributes just the order by will be single key.

* so we can accept a single value in order by and then if its timestamp then use id in the query else don't use id in the order by caluse.

ee/query-service/app/api/export.go
nityanandagohain: don't think there is a constant, you can create one.

ee/query-service/app/api/export.go
nityanandagohain: Did you check how other platform does it ?

nityanandagohain (commented):
I have added some comments, also let's bring the mockgen in a separete PR, it might require buy in for people in platform as well.

ee/query-service/app/api/export.go
tushar-signoz: Changing default to timestamp:desc

pkg/modules/export/implexport/handler.go
tushar-signoz: If we decide to add swagger, swagger should be able to pick these comments. 
And documenting doesn't harm

pkg/modules/export/implexport/handler.go
tushar-signoz: Done
I'll revisit this parsing logic as well later

pkg/modules/export/implexport/handler.go
tushar-signoz: Removed

pkg/types/telemetrytypes/field.go
```diff
@@ -39,6 +40,36 @@ func (f TelemetryFieldKey) String() string {
...
+// Invalid formats:
+// - "name:"            (Type separator without type)
+// - ".name"            (Leading dot)
+// - "http..status"     (Consecutive dots)
+// - "name:type:extra"  (Multiple type separators)
+//
+// Note: If a data type is specified, a context must also be specified
+var fieldTextRegex = regexp.MustCompile(`^(?:(\w+)\.(?:[\w-]+(?:\.[\w-]+)*?)(?::(\w+))?|[\w-]+)$`)
+
+func ValidateFieldKeyText(key string) error {
```
nityanandagohain: @srikanthccv incase you have any suggestion here to add

nityanandagohain (approved):
lgtm

nityanandagohain:
Also a thought from the product point if we should download in plain text format instead of downloading it as a gzip. That way user won't need a tool to unzip the data.

pkg/types/telemetrytypes/field.go
srikanthccv: @tushar-signoz , the regex here is incorrect and deviates from the grammar rules we have defined here https://github.com/SigNoz/signoz/blob/main/grammar/FilterQuery.g4. Let's remove the key validation here. If the key is invalid, either the WhereClauseVisitor or StatementBuilder will return error.

pkg/telemetrylogs/const.go
```diff
@@ -16,51 +44,59 @@ var (
...
 			FieldContext:  telemetrytypes.FieldContextLog,
 			FieldDataType: telemetrytypes.FieldDataTypeString,
+			Materialized:  true,
 		},
 		"trace_id": {
 			Name:          "trace_id",
 			Signal:        telemetrytypes.SignalLogs,
 			FieldContext:  telemetrytypes.FieldContextLog,
 			FieldDataType: telemetrytypes.FieldDataTypeString,
+			Materialized:  true,
```
srikanthccv: Is there any reason why we adding these as "Materialized"? These are not materialised. They are intrinsic fields of the log. We should keep the "materialized" true only for the actually materialised column of the ClickHouse.

pkg/modules/rawdataexport/implrawdataexport/module.go
```diff
@@ -0,0 +1,96 @@
...
+
+const MAX_EXPORT_ROW_COUNT_LIMIT = 50000
+const DEFAULT_EXPORT_ROW_COUNT_LIMIT = 10000
+const MAX_EXPORT_BYTES_LIMIT = 10 * 1024 * 1024 * 1024 // 10 GB
+const CLICKHOUSE_CONTEXT_EXPORT_MAX_THREADS = 2
+const CHUNK_SIZE = 5000
+
+type ctxKey string
+
+const CLICKHOUSE_CONTEXT_MAX_THREADS_KEY ctxKey = "clickhouse_max_threads"
```
srikanthccv: https://go.dev/doc/effective_go#mixed-caps please

pkg/modules/rawdataexport/implrawdataexport/module.go
```diff
@@ -0,0 +1,96 @@
...
+		rowCount := 0
+
+		for rowCount < rowCountLimit {
+			spec.Limit = min(CHUNK_SIZE, rowCountLimit-rowCount)
+			spec.Offset = rowCount
+
+			rangeRequest.CompositeQuery.Queries[0].Spec = spec
+
+			// Set clickhouse max threads
+			ctx := context.WithValue(ctx, CLICKHOUSE_CONTEXT_MAX_THREADS_KEY, CLICKHOUSE_CONTEXT_EXPORT_MAX_THREADS)
```
srikanthccv: Also mixed-caps and can you please add this `CLICKHOUSE_CONTEXT_MAX_THREADS_KEY` ctxtypes?

pkg/modules/rawdataexport/implrawdataexport/module.go
```diff
@@ -0,0 +1,96 @@
...
+			if err != nil {
+				errChan <- err
+				return
+			}
+
+			newRowsCount := 0
+			for _, result := range response.Data.Results {
+				resultData, ok := result.(*qbtypes.RawData)
+				if !ok {
+					errChan <- errors.NewInvalidInputf(errors.CodeInvalidInput, "expected RawData, got %T", result)
```
srikanthccv: Must not happen, but if ever happens. It should be an internal error than invalid input because there is nothing invalid about this from end user side.

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -0,0 +1,499 @@
...
+			},
+			{
+				Direction: qbtypes.OrderDirectionDesc,
+				Key: qbtypes.OrderByKey{
+					TelemetryFieldKey: telemetrytypes.TelemetryFieldKey{
+						Name: telemetrylogs.LogsV2IDColumn,
+					},
+				},
+			},
+		}, nil
```
srikanthccv: Can you define this under `pkg/telemetrylogs` as default order by and use it here?

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -0,0 +1,499 @@
...
+
+	orderDirection, ok := qbtypes.OrderDirectionMap[direction]
+	if !ok {
+		return nil, errors.NewInvalidInputf(errors.CodeInvalidInput, "invalid order_by direction: %s, should be one of %s, %s", direction, qbtypes.OrderDirectionAsc, qbtypes.OrderDirectionDesc)
+	}
+
+	if err := telemetrytypes.ValidateFieldKeyText(column); err != nil {
+		return nil, err
+	}
+	orderByKey := telemetrytypes.GetFieldKeyFromKeyText(column)
```
srikanthccv: Are we restricting the order by to only one column in the API? I know the current UI has only one column but that was an incomplete implementation. We should support order by more than one column?

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -0,0 +1,499 @@
...
+
+	orderBy = append(orderBy, qbtypes.OrderBy{
+		Key: qbtypes.OrderByKey{
+			TelemetryFieldKey: orderByKey,
+		},
+		Direction: orderDirection,
+	})
```
srikanthccv: There is mixed use of naked returns. Can we please avoid using the named return values? 

https://go.dev/wiki/CodeReviewComments#named-result-parameters

>Naked returns are okay if the function is a handful of lines. Once it’s a medium sized function, be explicit with your return values. Corollary: it’s not worth it to name result parameters just because it enables you to use naked returns. Clarity of docs is always more important than saving a line or two in your function.

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -0,0 +1,499 @@
...
+//
+//   - end (required): End time for query (Unix timestamp in nanoseconds)
+//
+//   - limit (optional): Maximum number of rows to export
+//     Constraints: Must be positive and cannot exceed MAX_EXPORT_ROW_COUNT_LIMIT
+//
+//   - filter (optional): Filter expression to apply to the query
+//
+//   - columns (optional): Specific columns to include in export
+//     Default: ["timestamp", "id", "body"]
```
srikanthccv: @nityanandagohain should we send everything instead of just these three?

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -0,0 +1,499 @@
...
+				return fmt.Errorf("error writing CSV record: %w", err)
+			}
+
+			totalBytes += getsizeOfStringSlice(record)
+			if totalBytes > MAX_EXPORT_BYTES_LIMIT {
+				return nil
+			}
+		case err := <-errChan:
+			if err != nil {
+				return fmt.Errorf("error processing row: %w", err)
```
srikanthccv:
```suggestion
				return err
```

otherwise render.Error throws unknown error.

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -0,0 +1,499 @@
...
+			if _, err := gzipWriter.Write([]byte("\n")); err != nil {
+				return fmt.Errorf("error writing JSON newline: %w", err)
+			}
+
+			if totalBytes > MAX_EXPORT_BYTES_LIMIT {
+				return nil
+			}
+		case err := <-errChan:
+			if err != nil {
+				return fmt.Errorf("error processing row: %w", err)
```
srikanthccv: Same

pkg/modules/rawdataexport/implrawdataexport/handler.go
nityanandagohain: Yeah Srikanth, I suggested to revert back to one order by. For raw data didn't see much usecase. Let's consider this as a improvement if user comes up ? or we decide to have multiple order by in the UI for raw data.

pkg/modules/rawdataexport/implrawdataexport/handler.go
nityanandagohain: so the issue then is, we don't know what fields to extract from attributes,  either we get all the fileds from the keys table, which can be huge and unnecessary

The frontend is expected to send everything from the columns that it is rendering in the UI.

srikanthccv: I should have clarified it, I meant to say you don't look at the specific keys but instead send entire attributes object back to user. Otherwise this endpoint wouldn't be useful beyond UI trigger? If I want to download matching logs with complete data, what is the recommended way? 

```
[
  {
    "timestamp": <uint64>,
    "trace_id": <hex string>,
    "span_id": <hex string>,
    "trace_flags": <int>,
    "severity_text": <string>,
    "severity_number": <int>,
    "attributes": <map>,
    "resources": <map>,
    "body": <string>
  }
]
```

nityanandagohain: So in case of csv also you are suggesting to put the data as a map is it ?

your usecase is a valid one though @tushar-signoz

nityanandagohain: Checked and seems like it's a good enough way that we dump the map.

so here is what we can do.
* if the selectFields is not present then it means selecting everything and data which can contain maps as well.
* if the selectFields are specified then just add the select fields.

@tushar-signoz think about this.

pkg/modules/rawdataexport/implrawdataexport/module.go
tushar-signoz: Can you elaborate where to add this ?

pkg/modules/rawdataexport/implrawdataexport/module.go
tushar-signoz: Good catch

pkg/modules/rawdataexport/implrawdataexport/handler.go
tushar-signoz: ack

pkg/modules/rawdataexport/implrawdataexport/handler.go
tushar-signoz: Im wrapping the err coming from downstream and then returning it. Didn't get you how reder.Error throws unknown error. Is there some sort of logic in render.Error that classifies errors ?

pkg/modules/rawdataexport/implrawdataexport/module.go
srikanthccv: [github.com/SigNoz/signoz/pkg/types/ctxtypes](https://github.com/SigNoz/signoz/tree/main/pkg/types/ctxtypes)

pkg/modules/rawdataexport/implrawdataexport/handler.go
srikanthccv: The error handler here https://github.com/SigNoz/signoz/blob/7602d863dd8e687ca29b2a43219a13040f5113aa/pkg/http/render/render.go#L52 won't be able to clearly handle it if we wrap it like this. The downstream errors have code, type, and message, etc.. If you pass the err as-is, they will be mapped to the appropriate status codes.

tushar-signoz: ack

pkg/modules/rawdataexport/implrawdataexport/handler.go
tushar-signoz: Hey, added a default order const but not is pkg/telemetrylogs, importing qbtypes in telemetrylogs, might create cyclic import issues later.

nityanandagohain: @tushar-signoz we aready import qbtypes in `pkg/telemetrylogs`

therealpandey:
@tushar-signoz please take care of https://github.com/SigNoz/platform-pod/issues/1161

pkg/modules/rawdataexport/implrawdataexport/handler.go
tushar-signoz: Done

pkg/types/ctxtypes/clickhouse.go
```diff
@@ -0,0 +1,14 @@
+package ctxtypes
```
therealpandey, line 1: Can we reuse the function from upstream?

You can call the upstream and mutate the ctx.

tushar-signoz: @grandwizard28 Can you elaborate a bit more, since I don't have context on entire codebase, which function are you suggesting to be reused ?

nityanandagohain: I guess pandey is taking about something like this


```
"github.com/ClickHouse/clickhouse-go/v2"



ctx := clickhouse.Context(context.Background(), clickhouse.WithSettings(clickhouse.Settings{
		"clickhouse_max_threads": 1,
	}))
```

grandwizard28 (approved):

pkg/types/ctxtypes/clickhouse.go
therealpandey: Unblocking you guys. Will raise a separate PR to fix this inconsistency.

nityanandagohain (approved):

srikanthccv (approved):

PR #8943, opened 2025-08-28
feat: added limit and timestamp to the trace query builder

eKuG:
## 📄 Summary

Added limit and timestamp to the trace query builder
Fixes : https://signoz-team.slack.com/archives/C08814ED9S5/p1756370568448909

---

## ✅ Changes

- [x] Bug fix: Brief description

---

### DIFF
--- a/pkg/telemetrytraces/statement_builder.go
+++ b/pkg/telemetrytraces/statement_builder.go
@@ -361,6 +361,11 @@ func (b *traceQueryStatementBuilder) buildTraceQuery(
 	distSB := sqlbuilder.NewSelectBuilder()
 	distSB.Select("trace_id")
 	distSB.From(fmt.Sprintf("%s.%s", DBName, SpanIndexV3TableName))
+	distSB.Where(
+		distSB.GE("timestamp", fmt.Sprintf("%d", start)),
+		distSB.L("timestamp", fmt.Sprintf("%d", end)),
+		distSB.GE("ts_bucket_start", startBucket),
+		distSB.LE("ts_bucket_start", endBucket))
 
 	var (
 		cteFragments []string
@@ -401,6 +406,12 @@ func (b *traceQueryStatementBuilder) buildTraceQuery(
 	// order by duration and limit 1 per trace
 	innerSB.OrderBy("duration_nano DESC")
 	innerSB.SQL("LIMIT 1 BY trace_id")
+
+	if query.Limit > 0 {
+		innerSB.Limit(query.Limit)
+	} else {
+		innerSB.Limit(100)
+	}
 
 	innerSQL, innerArgs := innerSB.BuildWithFlavor(sqlbuilder.ClickHouse)
 
@@ -456,6 +467,7 @@ func (b *traceQueryStatementBuilder) buildTraceQuery(
 		stmt.WarningsDocURL = preparedWhereClause.WarningsDocURL
 	}
 
+	b.logger.Info(stmt.Query)
 	return stmt, nil
 }
 


pkg/telemetrytraces/stmt_builder_test.go
```diff
@@ -547,8 +547,8 @@ func TestStatementBuilderTraceQuery(t *testing.T) {
...
...n_at_ts_bucket_start <= ?), __toe AS (SELECT trace_id FROM signoz_traces.distributed_signoz_index_v3 WHERE timestamp >= ? AND timestamp < ? AND ts_bucket_start >= ? AND ts_bucket_start <= ? AND resource_fingerprint GLOBAL IN (SELECT fingerprint FROM __resource_filter) AND true AND timestamp >= ? AND timestamp < ? AND ts_bucket_start >= ? AND ts_bucket_start <= ?), __toe_duration_sorted AS (SELECT trace_id, duration_nano, resource_string_service$$name as `service.name`, name FROM signoz_traces.distributed_signoz_index_v3 WHERE parent_span_id = '' AND timestamp >= ? AND timestamp < ? AND ts_bucket_start >= ? AND ts_bucket_start <= ? ORDER BY duration_nano DESC LIMIT 1 BY trace_id LIMIT ?) SELECT __toe_duration_sorted.`service.name` AS `service.name`, __toe_duration_sorted.name AS `name`, count() AS span_count, __toe_duration_sorted.duration_nano AS `duration_nano`, __toe_duration_sorted.trace_id AS `trace_id` FROM __toe INNER JOIN __toe_duration_sorted ON __toe.trace_id = __toe_duration_sorted.trace_id GROUP BY trace_id, duration_nano, name, `service.name` ORDER BY duration_nano DESC LIMIT 1 BY trace_id LIMIT ? SETTINGS distributed_product_mode='allow', max_memory_usage=10000000000",
```
nityanandagohain: I see the timestamp filters getting applied twice 
image


Also the limit by tace_id doesn't seem to be correct, check "test noop trace view" in query_builder_test.go in the v4

eKuG: LIMIT 1 by trace_id == `select distinct on (trace_id)`

Also we are limiting the inner query because we had also done the same thing in v3 as well. Refer to https://github.com/SigNoz/signoz/blob/main/pkg/query-service/app/traces/v4/query_builder_test.go#L742

This will limit the select of traces earlier in the query and then apply all the necessary filters.

nityanandagohain: but the example you shared, the order by is `span_count` and it is not used anywhere from the ui.

nityanandagohain: let me try the query and get back

nityanandagohain: okay it's seems correct, but if you are adding limit here then you will also need to add offset ? please add an example what will it look if the offset is 100 and limit 10

eKuG: Yes, the offset should be added in the inner query, I have made the required changes.
@nityanandagohain

nityanandagohain: please add test case as well, so that any change in future will be caught in that test.

eKuG: Added Tests cases as well

srikanthccv:
This is incorrect and won't work for any case where the filters are applied. Let's take a simple case of 2 independent services: 1. `api`, 2. `job`. The service `api` serves the web requests, and `job` does some background job processing all the time. Let's say `job` by nature has long-running tasks and a high duration_nano.

Then the user applied the filter `service.name='api'` in the trace view. Since the `job` high duration_nano, the top 100 will always be `job` traces and doesn't include any traces from the `api`; as a result, the join will return no result.

srikanthccv:
I don't know why this was closed, I assumed it was discussed with @nityanandagohain and perhaps , there would be follow-up PR but I haven't seen one. I hope this is not forgotten just like that. @nityanandagohain please help me understand the next steps.

eKuG:
@srikanthccv I closed this PR because I wanted to take more time to develop a proper solution — as you correctly pointed out earlier, the previous approach wasn’t right. I’ll be opening a new PR soon once the revised solution is ready. I’m already tracking this as an item on my board and will inform you once the new PR is up.

PR #8955, opened 2025-08-29
feat: add span percentile for traces

eKuG:
## 📄 Summary

Add span percentile for traces

Request : Get Span Details
```
curl --location 'localhost:8080/api/v1/span_percentile' \
--data '{
    "spanDuration": 216401,
    "name": "oteldemo.ProductCatalogService/ListProducts",
    "serviceName": "productcatalogservice",
    "resourceAttributes": {
        "deployment.environment": "otel-demo"
    },
    "start":1760513940000,
    "end": 1760517540000
}'
```

Response:
```
{
    "status": "success",
    "data": {
        "percentiles": {
            "p50": 135289,
            "p90": 175302,
            "p99": 374719
        },
        "position": {
            "percentile": 95.24,
            "description": "faster than 95.2% of spans"
        }
    }
}
```

---

## ✅ Changes

- [x] Feature: Brief description

---

## 📋 Checklist

- [x] Dev Review

---

## 👀 Notes for Reviewers

----

> [!IMPORTANT]
> Add span percentile calculation feature for traces with new API routes and modules.
> 
>   - **Feature**:
>     - Add span percentile calculation for traces.
>     - Introduce `GetSpanPercentileDetails` in `implspanpercentile/handler.go`.
>     - Add `buildSpanPercentileQuery` in `implspanpercentile/translator.go`.
>   - **API**:
>     - Register new route `/api/v1/span-percentile/details` in `http_handler.go`.
>     - Add `RegisterSpanPercentileRoutes` in `http_handler.go`.
>   - **Modules**:
>     - Add `spanpercentile` module in `signoz/module.go` and `signoz/handler.go`.
>     - Implement `spanpercentile.Module` in `implspanpercentile/module.go`.
>   - **Validation**:
>     - Add `Validate` method in `spanpercentiletypes/spanpercentile.go`.
>   - **Tests**:
>     - Add tests for span percentile query building in `implspanpercentile/translator_cte_builder_test.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for b75bdc9ceda3c744e5a8bbcc49806f9607be666c. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/pkg/query-service/app/http_handler.go
+++ b/pkg/query-service/app/http_handler.go
@@ -625,6 +625,8 @@ func (aH *APIHandler) RegisterRoutes(router *mux.Router, am *middleware.AuthZ) {
 	// Export
 	router.HandleFunc("/api/v1/export_raw_data", am.ViewAccess(aH.Signoz.Handlers.RawDataExport.ExportRawData)).Methods(http.MethodGet)
 
+	router.HandleFunc("/api/v1/span_percentile", am.ViewAccess(aH.Signoz.Handlers.SpanPercentile.GetSpanPercentileDetails)).Methods(http.MethodPost)
+
 }
 
 func (ah *APIHandler) MetricExplorerRoutes(router *mux.Router, am *middleware.AuthZ) {
@@ -2172,10 +2174,6 @@ func (aH *APIHandler) RegisterThirdPartyApiRoutes(router *mux.Router, am *middle
 
 	overviewRouter.HandleFunc("/list", am.ViewAccess(aH.getDomainList)).Methods(http.MethodPost)
 	overviewRouter.HandleFunc("/domain", am.ViewAccess(aH.getDomainInfo)).Methods(http.MethodPost)
-}
-
-func (aH *APIHandler) RegisterSpanPercentileRoutes(router *mux.Router, am *middleware.AuthZ) {
-	router.HandleFunc("/api/v1/span-percentile", am.ViewAccess(aH.Signoz.Handlers.SpanPercentile.GetSpanPercentileDetails)).Methods(http.MethodPost)
 }
 
 // not using md5 hashing as the plain string would work
--- a/pkg/types/spanpercentiletypes/spanpercentile.go
+++ b/pkg/types/spanpercentiletypes/spanpercentile.go
@@ -5,10 +5,10 @@ import (
 )
 
 type SpanPercentileRequest struct {
-	DurationNano       int64             `json:"span_duration"`
+	DurationNano       int64             `json:"spanDuration"`
 	Name               string            `json:"name"`
-	ServiceName        string            `json:"service_name"`
-	ResourceAttributes map[string]string `json:"resource_attributes"`
+	ServiceName        string            `json:"serviceName"`
+	ResourceAttributes map[string]string `json:"resourceAttributes"`
 	Start              uint64            `json:"start"`
 	End                uint64            `json:"end"`
 }
--- a/ee/query-service/app/server.go
+++ b/ee/query-service/app/server.go
@@ -228,7 +228,6 @@ func (s *Server) createPublicServer(apiHandler *api.APIHandler, web web.Web) (*h
 	apiHandler.RegisterThirdPartyApiRoutes(r, am)
 	apiHandler.MetricExplorerRoutes(r, am)
 	apiHandler.RegisterTraceFunnelsRoutes(r, am)
-	apiHandler.RegisterSpanPercentileRoutes(r, am)
 
 	c := cors.New(cors.Options{
 		AllowedOrigins: []string{"*"},
--- a/pkg/query-service/app/server.go
+++ b/pkg/query-service/app/server.go
@@ -204,7 +204,6 @@ func (s *Server) createPublicServer(api *APIHandler, web web.Web) (*http.Server,
 	api.RegisterQueryRangeV5Routes(r, am)
 	api.RegisterMessagingQueuesRoutes(r, am)
 	api.RegisterThirdPartyApiRoutes(r, am)
-	api.RegisterSpanPercentileRoutes(r, am)
 	api.MetricExplorerRoutes(r, am)
 	api.RegisterTraceFunnelsRoutes(r, am)
 


pkg/query-service/app/http_handler.go
```diff
@@ -5093,6 +5103,53 @@ func (aH *APIHandler) getDomainInfo(w http.ResponseWriter, r *http.Request) {
 	aH.Respond(w, finalResult)
 }
 
+// getSpanPercentileDetails handles requests for span percentile analysis
+func (aH *APIHandler) getSpanPercentileDetails(w http.ResponseWriter, r *http.Request) {
+	claims, err := authtypes.ClaimsFromContext(r.Context())
+	if err != nil {
+		render.Error(w, err)
```
nityanandagohain: why are we mixing errors again ?

pkg/query-service/app/http_handler.go
```diff
@@ -5093,6 +5103,53 @@ func (aH *APIHandler) getDomainInfo(w http.ResponseWriter, r *http.Request) {
...
+	queryRangeRequest, err := spanpercentile.BuildSpanPercentileQuery(spanPercentileRequest)
+	if err != nil {
+		zap.L().Error("Failed to build span percentile query", zap.Error(err))
+		apiErrObj := errorsV2.Newf(errorsV2.TypeInvalidInput, errorsV2.CodeInvalidInput, err.Error())
+		render.Error(w, apiErrObj)
+		return
+	}
+
+	if err := queryRangeRequest.Validate(); err != nil {
+		zap.L().Error("Query validation failed", zap.Error(err))
```
nityanandagohain: nit: do you think these requires to be logged ?

pkg/modules/spanpercentile/translator.go
```diff
@@ -0,0 +1,116 @@
...
+	mainBuilder.SQL("FROM target_span t")
+	mainBuilder.SQL(fmt.Sprintf("CROSS JOIN %s.%s s", telemetrytraces.DBName, telemetrytraces.SpanIndexV3TableName))
+
+	mainBuilder.Where(
+		mainBuilder.GE("s.timestamp", fmt.Sprintf("%d", startNano)),
+		mainBuilder.L("s.timestamp", fmt.Sprintf("%d", endNano)),
+		mainBuilder.GE("s.ts_bucket_start", startBucketSec),
+		mainBuilder.LE("s.ts_bucket_start", endBucketSec),
+		"s.name = t.name",
+		"s.`resource_string_service$name` = t.service_name",
```
nityanandagohain: this is incorrect ,

`resource_string_service$$name`

pkg/modules/spanpercentile/implspanpercentile/translator.go
```diff
@@ -0,0 +1,116 @@
...
+	endNano := querybuilder.ToNanoSecs(req.End)
+	startBucketSec := req.Start/1000 - querybuilder.BucketAdjustment
+	endBucketSec := req.End / 1000
+
+	targetSpanBuilder := sqlbuilder.NewSelectBuilder()
+	targetSpanBuilder.Select(
+		"duration_nano",
+		"name",
+		"resource_string_service$$name as `service_name`",
+		"resources_string['deployment.environment'] as deployment_environment",
```
nityanandagohain: Whenever you are using resource based filtering, please add why fingerprint is not used if we don't use it.

pkg/modules/spanpercentile/translator.go
```diff
@@ -0,0 +1,116 @@
...
+import (
+	"fmt"
+	"github.com/SigNoz/signoz/pkg/querybuilder"
+	"github.com/SigNoz/signoz/pkg/telemetrytraces"
+	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
+	"github.com/SigNoz/signoz/pkg/types/spanpercentiletypes"
+	"github.com/huandu/go-sqlbuilder"
+)
+
+func BuildSpanPercentileQuery(req *spanpercentiletypes.SpanPercentileRequest) (*qbtypes.QueryRangeRequest, error) {
```
nityanandagohain: please add test for this module to check the queries getting built.

pkg/query-service/app/http_handler.go
```diff
@@ -5093,6 +5103,53 @@ func (aH *APIHandler) getDomainInfo(w http.ResponseWriter, r *http.Request) {
 	aH.Respond(w, finalResult)
 }
 
+// getSpanPercentileDetails handles requests for span percentile analysis
+func (aH *APIHandler) getSpanPercentileDetails(w http.ResponseWriter, r *http.Request) {
```
nityanandagohain: since we have already created module, let's move the handler to the module itself.

nityanandagohain (commented):
Have left some comments, I would request to take care of errors as we have been mixing old and new errors multiple times in the last few pr's

pkg/modules/spanpercentile/translator.go
```diff
@@ -0,0 +1,116 @@
...
+		mainBuilder.L("s.timestamp", fmt.Sprintf("%d", endNano)),
+		mainBuilder.GE("s.ts_bucket_start", startBucketSec),
+		mainBuilder.LE("s.ts_bucket_start", endBucketSec),
+		"s.name = t.name",
+		"s.`resource_string_service$name` = t.service_name",
+		"s.resources_string['deployment.environment'] = t.deployment_environment",
+	)
+
+	if req.Filter != nil && req.Filter.Expression != "" {
+		mainBuilder.Where(fmt.Sprintf("(%s)", req.Filter.Expression))
```
nityanandagohain: how is this going to work with somthing like service.name='myservice' ?

pkg/modules/spanpercentile/implspanpercentile/translator.go
```diff
@@ -0,0 +1,208 @@
...
+	endBucket := b.end / querybuilder.NsToSeconds
+
+	sb := sqlbuilder.NewSelectBuilder()
+	sb.Select("fingerprint")
+	sb.From(fmt.Sprintf("%s.%s", resourcefilter.TracesDBName, resourcefilter.TraceResourceV3TableName))
+	sb.Where(
+		sb.GE("seen_at_ts_bucket_start", fmt.Sprintf("%d", startBucket)),
+		sb.LE("seen_at_ts_bucket_start", fmt.Sprintf("%d", endBucket)),
+	)
+	sb.Where(fmt.Sprintf("(%s)", b.filter.Expression))
```
nityanandagohain: can you directly add filter to a where clause ? we do a lot of metadata update and traformation before building a where clause from expression

pkg/modules/spanpercentile/implspanpercentile/translator.go
```diff
@@ -0,0 +1,208 @@
...
+	b.addCTE("base_spans", sql, args)
+}
+
+func (b *spanPercentileCTEBuilder) buildTargetSpanCTE() {
+	sb := sqlbuilder.NewSelectBuilder()
+	sb.Select(
+		"duration_nano",
+		"name",
+		"`service.name` as service_name",
+		"resources_string['deployment.environment'] as deployment_environment",
```
nityanandagohain: use "resource.`deployment.environment`" as deployment_environment.

pkg/modules/spanpercentile/implspanpercentile/translator.go
```diff
@@ -0,0 +1,208 @@
...
+		"quantile(0.99)(s.duration_nano) as p99_duration_ms",
+		"count(*) as total_spans_in_group",
+	)
+
+	sb.SQL("FROM target_span t")
+	sb.SQL("CROSS JOIN base_spans s")
+	sb.Where(
+		"s.name = t.name",
+		"s.`service.name` = t.service_name",
+		"s.resources_string['deployment.environment'] = t.deployment_environment",
```
nityanandagohain: we can use same resource json

pkg/modules/spanpercentile/implspanpercentile/translator_cte_builder_test.go
```diff
@@ -0,0 +1,339 @@
...
+			expectedErr: nil,
+		},
+		{
+			name: "span percentile query with resource filter",
+			request: &spanpercentiletypes.SpanPercentileRequest{
+				SpanID: "span987654321fedcba",
+				Start:  1640995200000,
+				End:    1640995800000,
+				Filter: &qbtypes.Filter{
+					Expression: "simpleJSONExtractString(labels, 'service.name') = 'frontend' AND labels LIKE '%service.name%' AND labels LIKE '%service.name\":\"frontend%'",
```
nityanandagohain: we cannot write these kind of queries from the UI

we can only write like this
`service.name = 'frontend'`

pkg/modules/spanpercentile/implspanpercentile/translator_cte_builder_test.go
```diff
@@ -0,0 +1,339 @@
...
...vice.name` FROM signoz_traces.distributed_signoz_index_v3 WHERE timestamp >= '1641000000000000000' AND timestamp < '1641003600000000000' AND ts_bucket_start >= 1640998200 AND ts_bucket_start <= 1641003600 AND resource_fingerprint GLOBAL IN (SELECT fingerprint FROM __resource_filter)), target_span AS (SELECT duration_nano, name, `service.name` as service_name, resources_string['deployment.environment'] as deployment_environment FROM base_spans WHERE span_id = 'complex-filter-span' LIMIT 1) SELECT 'complex-filter-span' as span_id, t.duration_nano, t.duration_nano as duration_ms, t.name as span_name, t.service_name, t.deployment_environment, round((sum(multiIf(s.duration_nano < t.duration_nano, 1, 0)) * 100.0) / count(*), 2) as percentile_position, quantile(0.50)(s.duration_nano) as p50_duration_ms, quantile(0.90)(s.duration_nano) as p90_duration_ms, quantile(0.99)(s.duration_nano) as p99_duration_ms, count(*) as total_spans_in_group FROM target_span t CROSS JOIN base_spans s WHERE s.name = t.name AND s.`service.name` = t.service_name AND s.resources_string['deployment.environment'] = t.deployment_environment GROUP BY t.duration_nano, t.name, t.service_name, t.deployment_environment",
```
nityanandagohain: we don't have any `deployment.environment` column

pkg/modules/spanpercentile/implspanpercentile/handler.go
```diff
@@ -0,0 +1,60 @@
...
+func (h *handler) GetSpanPercentileDetails(w http.ResponseWriter, r *http.Request) {
+	claims, err := authtypes.ClaimsFromContext(r.Context())
+	if err != nil {
+		render.Error(w, errorsV2.New(errorsV2.TypeInvalidInput, errorsV2.CodeInvalidInput, err.Error()))
+		return
+	}
+
+	spanPercentileRequest, err := parseSpanPercentileRequestBody(r)
+	if err != nil {
+		render.Error(w, errorsV2.New(errorsV2.TypeInvalidInput, errorsV2.CodeInvalidInput, err.Error()))
```
nityanandagohain: directly render no need to create new error

pkg/modules/spanpercentile/implspanpercentile/handler.go
```diff
@@ -0,0 +1,60 @@
...
+func NewHandler(module spanpercentile.Module) spanpercentile.Handler {
+	return &handler{
+		module: module,
+	}
+}
+
+func (h *handler) GetSpanPercentileDetails(w http.ResponseWriter, r *http.Request) {
+	claims, err := authtypes.ClaimsFromContext(r.Context())
+	if err != nil {
+		render.Error(w, errorsV2.New(errorsV2.TypeInvalidInput, errorsV2.CodeInvalidInput, err.Error()))
```
nityanandagohain: same for all the error handing in this handler

pkg/modules/spanpercentile/implspanpercentile/module.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func NewModule(querier querier.Querier, preferenceModule preference.Module) spanpercentile.Module {
+	return &module{
+		querier:          querier,
+		preferenceModule: preferenceModule,
+	}
+}
+
+func (m *module) GetSpanPercentileDetails(ctx context.Context, orgID valuer.UUID, userID valuer.UUID, req *spanpercentiletypes.SpanPercentileRequest) (*spanpercentiletypes.SpanPercentileResponse, error) {
```
nityanandagohain: GetSpanPercentile

pkg/modules/spanpercentile/implspanpercentile/module.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(req.ResourceAttributes) > 0 {
+		attrKeys := make([]any, 0, len(req.ResourceAttributes))
+		for key := range req.ResourceAttributes {
+			attrKeys = append(attrKeys, key)
+		}
+		_ = m.preferenceModule.UpdateByUser(ctx, userID, preferencetypes.NameSpanPercentileResourceAttributes, attrKeys)
```
nityanandagohain: so frontend won't send a request for saving preference right ?

pkg/modules/spanpercentile/implspanpercentile/module.go
```diff
@@ -0,0 +1,160 @@
...
+	row := scalarData.Data[0]
+
+	columnMap := make(map[string]int)
+	for i, col := range scalarData.Columns {
+		columnMap[col.Name] = i
+	}
+
+	p50Idx, ok := columnMap["p50_duration_nano"]
+	if !ok {
+		return nil, errors.New(errors.TypeInvalidInput, errors.CodeInvalidInput, "missing p50_duration_nano column")
```
nityanandagohain: how are all of these invalid input errors, these errors means your query building is wrong ?

pkg/modules/spanpercentile/implspanpercentile/module.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	p90, err := toInt64(row[p90Idx])
+	if err != nil {
+		return nil, errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "invalid p90 value: %v", err)
+	}
+	p99, err := toInt64(row[p99Idx])
+	if err != nil {
+		return nil, errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "invalid p99 value: %v", err)
+	}
+	position, err := toFloat64(row[positionIdx])
```
nityanandagohain, line 97: why is this float and the other ones int ?

pkg/modules/spanpercentile/implspanpercentile/statement_builder.go
```diff
@@ -0,0 +1,132 @@
...
+
+	for _, key := range attrKeys {
+		value := req.ResourceAttributes[key]
+		resourceKey := &telemetrytypes.TelemetryFieldKey{
+			Name:          key,
+			FieldContext:  telemetrytypes.FieldContextResource,
+			FieldDataType: telemetrytypes.FieldDataTypeString,
+			Materialized:  false,
+		}
+		fieldName, err := fieldMapper.FieldFor(ctx, resourceKey)
```
nityanandagohain: fingerprint is not getting generated

pkg/modules/spanpercentile/implspanpercentile/statement_builder.go
```diff
@@ -0,0 +1,132 @@
+package implspanpercentile
```
nityanandagohain, line 1: file name should be statement builder ?

pkg/modules/spanpercentile/implspanpercentile/statement_builder.go
```diff
@@ -0,0 +1,132 @@
...
+			Materialized:  false,
+		}
+		fieldName, err := fieldMapper.FieldFor(ctx, resourceKey)
+		if err != nil {
+			fieldName = fmt.Sprintf("s.resources_string['%s']", key)
+		}
+		sb.Where(sb.Equal(fieldName, value))
+	}
+
+	sb.SQL("SETTINGS max_threads = 12, min_bytes_to_use_direct_io = 1, use_query_cache = 0, enable_filesystem_cache = 0, use_query_condition_cache = 0")
```
nityanandagohain: set this at context level also add a timeout of the query at 10s

nityanandagohain:
I have added some comments @eKuG

pkg/modules/spanpercentile/implspanpercentile/module.go
nityanandagohain: we need to reset this if the length of attributes is 0

pkg/modules/spanpercentile/implspanpercentile/statement_builder_test.go
```diff
@@ -0,0 +1,291 @@
...
+				DurationNano: 206154,
+				Name:         "OPTIONS",
+				ServiceName:  "analytics-prod-v4",
+				ResourceAttributes: map[string]string{
+					"deployment.environment": "prod",
+				},
+				Start: 1640995200000,
+				End:   1640995800000,
+			},
+			expectedQueryParts: []string{
```
nityanandagohain: let's write full queries for some tests

pkg/modules/spanpercentile/implspanpercentile/module.go
eKuG: Yes

pkg/modules/spanpercentile/implspanpercentile/module.go
eKuG: Span percentile can be something like 95.24 which would be rendered in the frontend as P95, but we will give the full value.

pkg/modules/spanpercentile/implspanpercentile/statement_builder.go
eKuG: Ack

pkg/modules/spanpercentile/implspanpercentile/module.go
nityanandagohain: I mean `row[p99Idx]` will be int and `row[positionIdx]` will be float ? the values will be floating point right ?

eKuG: The only floating point value would be for percentile position which can have decimal positions. Other values of p99, p95 would be in nano seconds and should not be in floats

eKuG:
@nityanandagohain please review

pkg/modules/spanpercentile/implspanpercentile/handler.go
```diff
@@ -0,0 +1,59 @@
...
+
+	spanPercentileRequest, err := parseSpanPercentileRequestBody(r)
+	if err != nil {
+		render.Error(w, err)
+		return
+	}
+
+	result, err := h.module.GetSpanPercentile(r.Context(), valuer.MustNewUUID(claims.OrgID), valuer.MustNewUUID(claims.UserID), spanPercentileRequest)
+	if err != nil {
+		apiErrObj := errorsV2.New(errorsV2.TypeInvalidInput, errorsV2.CodeInvalidInput, err.Error())
```
nityanandagohain: this is also a redundant one right ?

pkg/modules/spanpercentile/implspanpercentile/statement_builder.go
```diff
@@ -0,0 +1,164 @@
...
+	)
+
+	sb.Where(sb.Equal("s.name", req.Name))
+
+	// Use fingerprint-based filtering when resource attributes are present
+	if len(req.ResourceAttributes) > 0 {
+		sb.Where("s.resource_fingerprint GLOBAL IN (SELECT fingerprint FROM __resource_filter)")
+	} else {
+		// Fallback to direct service.name filtering when no additional attributes
+		sb.Where(sb.Equal("s.resource_string_service$$name", req.ServiceName))
```
nityanandagohain: if service name is present then we should use the resource fingerprint since it is a resource attribute. I.e a resource filter will always be present in case of span percentile.

pkg/modules/spanpercentile/implspanpercentile/statement_builder_test.go
```diff
@@ -0,0 +1,360 @@
...
+				"206154 AS duration_nano",
```
nityanandagohain: still don't see a full sql query generated for the test, please add that.

pkg/modules/spanpercentile/implspanpercentile/module.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil, errors.New(errors.TypeInternal, errors.CodeInternal, "missing p99_duration_nano column")
+	}
+	positionIdx, ok := columnMap["percentile_position"]
+	if !ok {
+		return nil, errors.New(errors.TypeInternal, errors.CodeInternal, "missing percentile_position column")
+	}
+
+	p50, err := toInt64(row[p50Idx])
+	if err != nil {
+		return nil, errors.New(errors.TypeNotFound, errors.CodeNotFound, "no spans found matching the specified criteria")
```
nityanandagohain, line 87: this part is a bit unclear to me, when you run the query you explicitly add a query to calculate p50, p90, p99, and in what case you will get empty or no value and should we return an error or should keep P50 = 0 in the response.

pkg/modules/spanpercentile/implspanpercentile/module.go
```diff
@@ -0,0 +1,191 @@
...
+		attrKeys = append(attrKeys, key)
+	}
+	_ = m.preferenceModule.UpdateByUser(ctx, userID, preferencetypes.NameSpanPercentileResourceAttributes, attrKeys)
+
+	return transformToSpanPercentileResponse(result)
+}
+
+func transformToSpanPercentileResponse(queryResult *qbtypes.QueryRangeResponse) (*spanpercentiletypes.SpanPercentileResponse, error) {
+	if len(queryResult.Data.Results) == 0 {
+		return nil, errors.New(errors.TypeInvalidInput, errors.CodeInvalidInput, "no data returned from query")
```
nityanandagohain: these should all be internal errors right ?

pkg/modules/spanpercentile/implspanpercentile/module.go
```diff
@@ -0,0 +1,191 @@
...
+	switch v := val.(type) {
+	case int64:
+		return v, nil
+	case int:
+		return int64(v), nil
+	case uint64:
+		return int64(v), nil
+	case float64:
+		if v != v {
+			return 0, errors.New(errors.TypeInvalidInput, errors.CodeInvalidInput, "value is NaN")
```
nityanandagohain: even here what will the client do with invalid input, give it a defaul or keep it nil.

nityanandagohain: Also I would suggest to check onence if we can ignore this function altogher and see if there is something that we can set on the query level and just extract or fail at once.

like in this case this is missing a lot like uint8, uint16 , int8, int16 etc etc.


suggesting this both for toInt64 and toFloat64

pkg/modules/spanpercentile/implspanpercentile/statement_builder.go
```diff
@@ -0,0 +1,153 @@
...
+	return &qbtypes.QueryRangeRequest{
+		SchemaVersion: "v5",
+		Start:         req.Start,
+		End:           req.End,
+		RequestType:   qbtypes.RequestTypeScalar,
+		CompositeQuery: qbtypes.CompositeQuery{
+			Queries: []qbtypes.QueryEnvelope{queryEnvelope},
+		},
+		FormatOptions: &qbtypes.FormatOptions{
+			FormatTableResultForUI: true,
```
nityanandagohain, line 115: any use of this option in our case ?

pkg/modules/spanpercentile/implspanpercentile/statement_builder_test.go
nityanandagohain: these tests are redundant now  since we already have full test let's keep that and maintail useful ones.

pkg/modules/spanpercentile/implspanpercentile/statement_builder.go
```diff
@@ -0,0 +1,153 @@
...
+	sb.Select(
+		fmt.Sprintf("%d AS duration_nano", req.DurationNano),
+		fmt.Sprintf("'%s' AS span_name", escapeSQLString(req.Name)),
+		fmt.Sprintf("'%s' AS service_name", escapeSQLString(req.ServiceName)),
+	)
+
+	for _, key := range attrKeys {
+		value := req.ResourceAttributes[key]
+		escapedKey := escapeResourceAttr(key)
+		sb.SelectMore(fmt.Sprintf("'%s' AS %s", escapeSQLString(value), escapedKey))
```
nityanandagohain: what is the use of this ? I don't see it being used anywhere in the response

pkg/types/preferencetypes/preference.go
```diff
@@ -163,6 +163,15 @@ func NewAvailablePreference() map[Name]Preference {
...
 			Value:         MustNewValue([]any{}, ValueTypeArray),
 		},
+		NameSpanPercentileResourceAttributes: {
+			Name:          NameSpanPercentileResourceAttributes,
+			Description:   "Additional resource attributes for span percentile filtering (beyond mandatory name and service.name).",
+			ValueType:     ValueTypeArray,
+			DefaultValue:  MustNewValue([]any{"deployment.environment"}, ValueTypeArray),
+			AllowedScopes: []Scope{ScopeUser},
+			AllowedValues: []string{},
+			Value:         MustNewValue([]any{"deployment.environment"}, ValueTypeArray),
```
nityanandagohain, line 173: why are we specifying `deployment.environment` here ?

pkg/modules/spanpercentile/implspanpercentile/statement_builder_test.go
```diff
@@ -0,0 +1,415 @@
...
+		input    string
+		expected string
+	}{
+		{
+			input:    "simple string",
+			expected: "simple string",
+		},
+		{
+			input:    "string with 'quote'",
+			expected: "string with ''quote''",
```
nityanandagohain: `string with \'quote\'`

pkg/modules/spanpercentile/implspanpercentile/statement_builder.go
```diff
@@ -0,0 +1,153 @@
...
+
+	for _, key := range attrKeys {
+		value := resourceAttributes[key]
+		conditions = append(conditions, fmt.Sprintf("%s = '%s'", key, escapeSQLString(value)))
+	}
+
+	return strings.Join(conditions, " AND ")
+}
+
+func escapeResourceAttr(attr string) string {
```
nityanandagohain: not needed

pkg/modules/spanpercentile/implspanpercentile/statement_builder.go
```diff
@@ -0,0 +1,153 @@
...
+
+	sb.Where("s.resource_fingerprint GLOBAL IN (SELECT fingerprint FROM __resource_filter)")
+
+	sb.SQL("SETTINGS distributed_product_mode='allow', max_memory_usage=10000000000, max_execution_time=10")
+
+	mainSQL, mainArgs := sb.BuildWithFlavor(sqlbuilder.ClickHouse)
+
+	finalSQL := querybuilder.CombineCTEs(cteFragments) + mainSQL
+	finalArgs := querybuilder.PrependArgs(cteArgs, mainArgs)
+	interpolatedQuery := interpolateArgs(finalSQL, finalArgs)
```
nityanandagohain: what is the usecase of using query_range v5 when we are already using interpolateArgs, can't we directly query clickhouse ?

nityanandagohain: Also let's not interpolate ourself and leave it to the package and send them as args itself. As doing this we might land into more issues that might pop up due to escaping and there can be a potential for sql injections.

nityanandagohain: The other way is also you can avoid preparing the QB and use query range directly for calculating the percentile.

```
qbtypes.QueryBuilderQuery[qbtypes.TraceAggregation]{
        Signal:       telemetrytypes.SignalTraces,
        Aggregations: []qbtypes.TraceAggregation{
            {
                Expression: "p50(duration_nano)",
                Alias: "p50_duration_nano",
            },
            {
                Expression: "p90(duration_nano)",
                Alias: "p90_duration_nano",
            },
            {
                Expression: "p99(duration_nano)",
                Alias: "p99_duration_nano",
            },
            {
                Expression: "countIf(s.duration_nano <= 500000)) / count()",
                Alias: "percentile_position",
            },
        },
        Filter: &qbtypes.Filter{
            Expression: "...",
        },
    },
    ```

File: pkg/modules/spanpercentile/implspanpercentile/module.go (hunk 26 above)
Comment (eKuG) [reply pkg/modules/spanpercentile/implspanpercentile/module.go:87]: This can happen when the user clicks on a missing span. Then there will be a response with NaN

File: pkg/types/preferencetypes/preference.go (hunk 31 above)
Comment (eKuG) [reply pkg/types/preferencetypes/preference.go:173]: Because that would be the default selection if no attribute is selected other than name and service name

Comment (nityanandagohain) [reply pkg/types/preferencetypes/preference.go:173]: so we are using `name`, `service.name` and `deployment.environement` as the default keys ?

File: pkg/modules/spanpercentile/implspanpercentile/module.go (hunk 26 above)
Comment (nityanandagohain) [reply pkg/modules/spanpercentile/implspanpercentile/module.go:87]: but in that case we shouldn't run the query right, as the duration will be zero ?

File: pkg/modules/spanpercentile/implspanpercentile/statement_builder.go (hunk 34 above)
Comment (eKuG) [reply pkg/modules/spanpercentile/implspanpercentile/statement_builder.go]: > what is the usecase of using query_range v5 when we are already using interpolateArgs, can't we directly query clickhouse ?

We were doing that earlier and then you asked me to change it QB V5 so that we can use resource fingerprinting. So we changed it to use Query builder

Comment (eKuG) [reply pkg/modules/spanpercentile/implspanpercentile/statement_builder.go]: > The other way is also you can avoid preparing the QB and use query range directly for calculating the percentile.

Not the way we want to follow. Different package altogether, I feel we should directly make the query and pass it as it is as It is more performant as well

File: pkg/modules/spanpercentile/implspanpercentile/statement_builder.go (hunk 30 above)
Comment (eKuG) [reply pkg/modules/spanpercentile/implspanpercentile/statement_builder.go]: This is for building the Span Percentile Query

File: pkg/modules/spanpercentile/implspanpercentile/statement_builder.go (hunk 29 above)
Comment (eKuG) [reply pkg/modules/spanpercentile/implspanpercentile/statement_builder.go:115]: Since span percentile returns a single ClickHouseSQL query with this flag set to true, the querier will return the raw ClickHouse result directly without any additional post-processing



 What FormatTableResultForUI: true does:

  1. For ClickHouse SQL queries: When there's a single ClickHouseSQL query with this flag set, it simply returns the raw result value directly.
  2. For other query types: It formats multiple query results into a unified table structure that's easier for the UI to render.

File: pkg/modules/spanpercentile/implspanpercentile/statement_builder.go [hunk 35]
```diff
@@ -0,0 +1,122 @@
...
+		CompositeQuery: qbtypes.CompositeQuery{
+			Queries: []qbtypes.QueryEnvelope{queryEnvelope},
+		},
+		FormatOptions: &qbtypes.FormatOptions{
+			FormatTableResultForUI: true,
+		},
+	}, nil
+}
+
+func escapeSQLString(s string) string {
```
nityanandagohain: this seems incorrect and AI generated, check `QUOTED_TEXT` in `FilterQuery.g4` for the grammar, we just need to escape using `\n`

So you can just use `strings.ReplaceAll(s, "'", "\'")` and escapeSQLString function is not even needed.

pkg/modules/spanpercentile/implspanpercentile/statement_builder_test.go
```diff
@@ -0,0 +1,148 @@
...
+				ServiceName:  "inventory-service",
+				ResourceAttributes: map[string]string{
+					"cloud.platform":         "aws",
+					"deployment.environment": "staging",
+					"k8s.cluster.name":       "staging-cluster",
+				},
+				Start: 1640995200000,
+				End:   1640995800000,
+			},
+			expectedFilterExpr: "service.name = 'inventory-service' AND name = 'DELETE /api/items' AND cloud.platform = 'aws' AND deployment.environment = 'staging' AND k8s.cluster.name = 'staging-cluster'",
```
nityanandagohain, line 107: modify the test to check everything, i.e filter and group by and aggregations.

pkg/modules/spanpercentile/implspanpercentile/module.go
```diff
@@ -0,0 +1,177 @@
...
+type module struct {
+	querier          querier.Querier
+	preferenceModule preference.Module
+}
+
+func NewModule(
+	querier querier.Querier,
+	preferenceModule preference.Module,
+	_ factory.ProviderSettings,
+	_ telemetrytypes.MetadataStore,
```
nityanandagohain: MetadataStore is not required now.

nityanandagohain (dismissed):

pkg/types/spanpercentiletypes/spanpercentile.go
```diff
@@ -0,0 +1,43 @@
+package spanpercentiletypes
+
+import (
+	"github.com/SigNoz/signoz/pkg/errors"
+)
+
+type SpanPercentileRequest struct {
+	DurationNano       int64             `json:"span_duration"`
```
therealpandey:
```suggestion
	DurationNano       int64             `json:"spanDuration"`
```

PascalCase for requests. See other request response JSONs

pkg/query-service/app/http_handler.go
```diff
@@ -2174,6 +2174,10 @@ func (aH *APIHandler) RegisterThirdPartyApiRoutes(router *mux.Router, am *middle
 	overviewRouter.HandleFunc("/domain", am.ViewAccess(aH.getDomainInfo)).Methods(http.MethodPost)
 }
 
+func (aH *APIHandler) RegisterSpanPercentileRoutes(router *mux.Router, am *middleware.AuthZ) {
+	router.HandleFunc("/api/v1/span-percentile", am.ViewAccess(aH.Signoz.Handlers.SpanPercentile.GetSpanPercentileDetails)).Methods(http.MethodPost)
+}
```
therealpandey: Do we really need this? You can register these in the main `Register` function.

therealpandey: Also `/api/v1/span_percentiles`. We're using `underscore` based routes, not `-` based.

therealpandey (approved):

PR #8958, opened 2025-08-30
fix: don't skip resource filter in main table for OR queries

nityanandagohain:
Fixes https://github.com/SigNoz/engineering-pod/issues/2889

* Don't skip resource filters in the main table if there is an OR clause.
* Remove the resource filter when there is a or caluse as there is a huge difference in perf ( not much difference if perf but felt unnecessary)

----

> [!IMPORTANT]
> Fixes issue with resource filters being skipped in OR queries by updating `PrepareWhereClause()` logic and adding relevant tests.
> 
>   - **Behavior**:
>     - In `PrepareWhereClause()` in `where_clause_visitor.go`, ensure `SkipResourceFilter` is set to `false` if an OR clause is present in the query.
>     - This change prevents skipping resource filters in the main table for OR queries, improving query accuracy and performance.
>   - **Tests**:
>     - Add test case "Time series with OR b/w resource attr and attribute filter" in `stmt_builder_test.go` for `telemetrylogs` and `telemetrytraces` to verify the new behavior.
>     - Update expected query and arguments in test cases to reflect the inclusion of resource filters in OR queries.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for b9e93b7f9c71ce4a363e821edba42a5c6b127e02. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/pkg/querybuilder/where_clause_visitor.go
+++ b/pkg/querybuilder/where_clause_visitor.go
@@ -95,13 +95,6 @@ func PrepareWhereClause(query string, opts FilterExprVisitorOpts) (*PreparedWher
 		opts.Builder = sb
 	}
 
-	// override skipResourceFilter if the expression contains OR
-	if strings.Contains(strings.ToLower(query), " or ") {
-		opts.SkipResourceFilter = false
-	}
-
-	visitor := newFilterExpressionVisitor(opts)
-
 	// Set up error handling
 	lexerErrorListener := NewErrorListener()
 	lexer.RemoveErrorListeners()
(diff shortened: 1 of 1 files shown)


pkg/querybuilder/where_clause_visitor.go
```diff
@@ -95,6 +95,11 @@ func PrepareWhereClause(query string, opts FilterExprVisitorOpts) (*PreparedWher
 		opts.Builder = sb
 	}
 
+	// override skipResourceFilter if the expression contains OR
+	if strings.Contains(strings.ToLower(query), " or ") {
+		opts.SkipResourceFilter = false
+	}
```
srikanthccv: Can we iterate over the token stream and check for the type `grammar.FilterQueryLexerOR` instead?

nityanandagohain: thanks, done 👍

srikanthccv (approved):

PR #8974, opened 2025-09-01
chore: add metric temporality cache

### LINKED ISSUE #8961: Metric temporality cache-aside with soft/hard TTL
- Hit & fresh (< softTTL) then return.
- Hit but stale (softTTL <= age < hardTTL), then return stale and kick off the go routine for refresh.
- Miss or > hardTTL -> load from DB, set cache
- Add jitter to avoid herd behaviour (TTL +/-X%)

What should be soft and hard TTL for each metric type?

srikanthccv:
## 📄 Summary

Fixes https://github.com/SigNoz/signoz/issues/8961


srikanthccv:
stale

PR #8975, opened 2025-09-01
fix: resource filter breaks under NOT with non-resource predicates

### LINKED ISSUE #8964: Resource fingerprint extraction breaks under `NOT` with non-resource predicates
When building the __resource_filter CTE for logs/traces, our resource-only predicate extraction mishandles expressions that contain `NOT (...)` around sub-expressions that have no resource terms. This yields an empty resource fingerprint set leading to incorrect result.

```
NOT (message CONTAINS 'foo' AND hasToken(body, 'bar'))
```

The expression has no resource terms. Projected to the resource plane, the inner becomes `true AND true` => `NOT (true)` => no resource fingerprints, when the fingerprints should have been entire set.

srikanthccv:
## 📄 Summary

Fixes https://github.com/SigNoz/signoz/issues/8964


srikanthccv:
Superseded by https://github.com/SigNoz/signoz/pull/10165

PR #8982, opened 2025-09-02
feat(authz): implement the current usecases in openfga

vikrantgupta25:


pkg/signoz/signoz.go
```diff
@@ -264,6 +266,12 @@ func New(
 		return nil, err
 	}
 
+	authzProviderFactory := openfgaauthz.NewProviderFactory(config.SQLStore)
```
vikrantgupta25: i ll remove this. just for some testing

pkg/authz/openfgaschema/schema.go
```diff
@@ -0,0 +1,32 @@
...
+
+	openfgapkgtransformer "github.com/openfga/language/pkg/go/transformer"
+)
+
+var (
+	//go:embed base.fga
+	baseDSL string
+)
+
+type Schema interface {
```
therealpandey: Move this to `pkg/authz/schema.go`. The interface needs to be decoupled from the implemention.

grandwizard28 (approved):

PR #9020, opened 2025-09-06
chore: enable alertmanager metrics collection with instrumentation::metrics

aniketio-ctrl:
## 📄 Summary
Exposed Alertmanager metrics at :9093/metrics and added a signoz_ prefix to all metrics using a custom registerer

----

> [!IMPORTANT]
> Exposes Alertmanager metrics with a `signoz_` prefix and integrates custom Prometheus registry support.
> 
>   - **Behavior**:
>     - Exposes Alertmanager metrics at `:9093/metrics` with a `signoz_` prefix using a custom registerer in `server.go`.
>     - Adds `metric.go` to create Prometheus metric readers with a custom registry.
>   - **Instrumentation**:
>     - Updates `sdk.go` to use `meterProviderWithCustomRegistry()` for setting up metrics with a custom Prometheus registry.
>   - **Misc**:
>     - Removes indirect comments for `prometheus` and `metric` packages in `go.mod`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for a4e0379ac4ab8fe9b94792bad1ba1a4f6fd764c5. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


pkg/alertmanager/alertmanagerserver/server.go
```diff
@@ -62,6 +64,9 @@ type Server struct {
 	tmpl              *template.Template
 	wg                sync.WaitGroup
 	stopc             chan struct{}
+
+	// metrics server for exposing prometheus metrics
```
therealpandey: There is a prometheus registry present in instrumentation. Please reuse.

grandwizard28 (changes requested):

pkg/alertmanager/alertmanagerserver/server.go
therealpandey: Also, please ensure that the metrics are org namespaced otherwise the registry will panic if same metrics are registered twice.

aniketio-ctrl: yes i am using scoped registry only wrapping it with prefix : signoz_

pkg/instrumentation/metric.go
srikanthccv, line 1: did a side-by-side comparison, this is largely a copy of upstream without the period readers

srikanthccv (dismissed):

pkg/alertmanager/alertmanagerserver/server.go
srikanthccv: @aniketio-ctrl I don't think you understood his point; alertmanager/service/newServer creates one server for each org (which uses the same registry). You should instead use `WrapRegistererWith` with orgId as label in service.go.

pkg/instrumentation/metric.go
therealpandey: Upstream does support exposing the prometheus registry directly.

We will have to write something of our own or contribute to the upstream.

pkg/alertmanager/alertmanagerserver/server.go
aniketio-ctrl: added the org_id as a label

pkg/instrumentation/metric.go
srikanthccv: I assume you meant "does not support". Yes, I spent some time digging through it and there weren't much options. I expected there would be a way using the bridge provided here https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/bridges/prometheus, however, one can only pass the reader at the time meter provider instantiation which again means no way to use bridge with the `contribsdkconfig.NewSDK`.

srikanthccv (approved):

pkg/instrumentation/metric.go
```diff
@@ -0,0 +1,160 @@
...
+			}
+			opts = append(opts, otelprom.WithResourceAsConstantLabels(attribute.NewDenyKeysFilter(keys...)))
+		}
+	}
+
+	// Use custom registry instead of creating a new one
+	opts = append(opts, otelprom.WithRegisterer(customRegistry))
+
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.HandlerFor(customRegistry, promhttp.HandlerOpts{Registry: customRegistry}))
```
therealpandey, line 77: We also read the path from config. Please fix this

aniketio-ctrl: ok, yeah updating that

Issue #8905, opened 2025-08-24
NO_COMMON_TYPE exception on resource filters with numeric values


srikanthccv:
We convert all resource attributes into string values. When a search is made on a resource attribute with a numeric value (the actual type from the user side), the query fails with the NO_COMMON_TYPE exception.

```
code: 386, message: There is no supertype for types String, UInt16 because some of them are String/FixedString/Enum and some of them are not
```

Prepared query
```
WITH __resource_filter AS (SELECT fingerprint FROM signoz_logs.distributed_logs_v2_resource WHERE ((simpleJSONExtractString(labels, 'some.key.here') = ? AND labels LIKE ? AND labels LIKE ?) AND (simpleJSONExtractString(labels, 'deployment.environment') = ? AND labels LIKE ? AND labels LIKE ?)) AND seen_at_ts_bucket_start >= ? AND seen_at_ts_bucket_start <= ?) SELECT timestamp, id, trace_id, span_id, trace_flags, severity_text, severity_number, scope_name, scope_version, body, attributes_string, attributes_number, attributes_bool, resources_string, scope_string FROM signoz_logs.distributed_logs_v2 WHERE resource_fingerprint GLOBAL IN (SELECT fingerprint FROM __resource_filter) AND true AND timestamp >= ? AND timestamp < ? AND ts_bucket_start >= ? AND ts_bucket_start <= ? ORDER BY timestamp AS `timestamp` desc, id AS `id` desc LIMIT ?
```

Prepared args
```
[12345,\"%some.key.here%\",\"%12345%\",\"staging\",\"%deployment.environment%\",\"%deployment.environment\\\":\\\"staging%\",1756051218,1756056618,\"1756053018000000000\",\"1756056618000000000\",1756051218,1756056618,100]
```

PR #9028, opened 2025-09-08
fix: exception on resource filters with numeric values

nityanandagohain:
Fixes https://github.com/SigNoz/signoz/issues/8905 .

### DIFF
--- a/pkg/querybuilder/resourcefilter/condition_builder.go
+++ b/pkg/querybuilder/resourcefilter/condition_builder.go
@@ -31,7 +31,7 @@ func valueForIndexFilter(op qbtypes.FilterOperator, key *telemetrytypes.Telemetr
 		// assuming array will always be for in and not in
 		values := make([]string, 0, len(v))
 		for _, v := range v {
-			values = append(values, fmt.Sprintf(`%%%s":"%s%%`, key.Name, v))
+			values = append(values, fmt.Sprintf(`%%%s":"%s%%`, key.Name, querybuilder.FormatValueForContains(v)))
 		}
 		return values
 	}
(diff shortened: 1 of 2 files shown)


srikanthccv:
I am wondering if it should be the other way around, i.e simpleJSONExtractNumber should be used instead of simpleJSONExtractString. Why? It makes sense to convert any type to a string if the operator used is (I)(NOT)(LIKE) or (NOT)(CONTAINS). However, the same can't be true for other operators such as `>=`, `>`, `<=`, `<` or between, as the way comparison works on them is chronological comparison rather than the numeric comparison. For instance, `'27' < '4'` is true.

nityanandagohain:
As discussed we are okay with chronological ordering for `>=, >, <=, <` as of now.

pkg/querybuilder/resourcefilter/condition_builder.go
```diff
@@ -61,7 +61,12 @@ func (b *defaultConditionBuilder) ConditionFor(
 		qbtypes.FilterOperatorILike,
 		qbtypes.FilterOperatorNotILike,
 		qbtypes.FilterOperatorLike,
-		qbtypes.FilterOperatorNotLike:
```
srikanthccv, line 64: We should keep the `FilterOperatorNotLike`, and additionally need to include `FilterOperatorRegexp` and `FilterOperatorNotRegexp`.

At this point, we should just update the `FormatValueForContains` to return the right value for all inputs instead of `fmt.Sprintf("%v", value)` in its implementation.

nityanandagohain: Thanks, have added them.


Didn't get your second part though, at this point we are just excluding the between and IN operators here.

srikanthccv: Right, and then we format them separately below. It's cleaner if we make the `value` get typecasted at the beginning regardless of `FilterOperator`.

nityanandagohain: makes sense, added  the changes.

srikanthccv (approved):

PR #9054, opened 2025-09-10
fix: show precision for value panel

SagarRajput-7:
## 📄 Summary

---

## ✅ Changes

---

ex:

- `frontend`
- `backend`
- `devops`
- `bug`
- `enhancement`
- `ui`
- `test`

---

## 👥 Reviewers

- frontend / backend / devops

---

## 🧪 How to Test

1. ...
2. ...
3. ...

---

## 🔍 Related Issues

Closes #

---

---

## 📋 Checklist

- [ ] Dev Review

---

### DIFF
--- a/frontend/src/container/GridValueComponent/index.tsx
+++ b/frontend/src/container/GridValueComponent/index.tsx
@@ -11,6 +11,30 @@ import { EQueryType } from 'types/common/dashboard';
 import { TitleContainer, ValueContainer } from './styles';
 import { GridValueComponentProps } from './types';
 
+// Function to preserve original precision while applying units
+const formatValueWithPrecision = (
+	value: number,
+	yAxisUnit?: string,
+): string => {
+	if (!yAxisUnit) {
+		return value.toString();
+	}
+
+	// Get the formatted value with unit but without precision loss
+	try {
+		const formattedValue = getYAxisFormattedValue(String(value), yAxisUnit);
+		// Extract the unit part from the formatted value
+		const unitMatch = formattedValue.match(/([KMB]?)$/);
+		const unit = unitMatch ? unitMatch[1] : '';
+
+		// Return the original value with the unit appended
+		return `${value}${unit}`;
+	} catch (error) {
+		console.error('Error formatting value:', error);
+		return value.toString();
+	}
+};
+
 function GridValueComponent({
 	data,
 	title,
@@ -97,11 +121,7 @@ function GridValueComponent({
 				<ValueGraph
 					thresholds={thresholds || []}
 					rawValue={value}
-					value={
-						yAxisUnit
-							? getYAxisFormattedValue(String(value), yAxisUnit)
-							: value.toString()
-					}
+					value={formatValueWithPrecision(value, yAxisUnit)}
 				/>
 			</ValueContainer>
 			<ContextMenu


frontend/src/components/Graph/__tests__/yAxisConfig.test.ts
```diff
@@ -0,0 +1,134 @@
...
+			expect(getYAxisFormattedValue('invalid', 'none')).toBe('NaN');
+			expect(getYAxisFormattedValue('', 'none')).toBe('NaN');
+			expect(getYAxisFormattedValue('abc123', 'none')).toBe('NaN');
+		});
+
+		it('should handle very small numbers', () => {
+			expect(getYAxisFormattedValue('0.0001', 'none')).toBe('0');
+			expect(getYAxisFormattedValue('-0.0001', 'none')).toBe('0');
+			expect(getYAxisFormattedValue('0.000000001', 'none')).toBe('0');
+		});
```
srikanthccv: the test says it should handle but it doesn't really handle it. the `0.0001` is very much a valid and significant value in observabilty. It's not a zero.

srikanthccv: @ahrefabhi you should catch these

SagarRajput-7: This test case, as per the current implementation, `defaults to 3 decimal` this expectation is correct. But since we have a thread to make the final decision on the precision. - [thread](https://signoz-team.slack.com/archives/C080PFAFHS4/p1757911197743979), hence according to that the test expectation is also expected to change

srikanthccv: That's not the point. Can he see the test and reason that `0.0001` rounded to `0` is acceptable or not. Given the domain we are in, we expect our people to think, ask questions and make the right choices.

frontend/src/components/Graph/yAxisConfig.ts
```diff
@@ -1,49 +1,100 @@
 import { formattedValueToString, getValueFormat } from '@grafana/data';
+import { isNaN } from 'lodash-es';
 
+/**
+ * Formats a number for display, showing up to 3 significant decimal places.
+ * It avoids scientific notation and removes unnecessary trailing zeros.
```
srikanthccv:
```suggestion
 * Formats a number for display, preserving leading zeros after the decimal point
 * and showing up to 5 digits after the first non-zero decimal digit.
 * It avoids scientific notation and removes unnecessary trailing zeros.
```

frontend/src/components/Graph/yAxisConfig.ts
```diff
@@ -1,49 +1,100 @@
...
+ *
+ * @example
+ * formatWithSignificantDecimals(1.2345); // "1.234"
+ * formatWithSignificantDecimals(0.0012345); // "0.00123"
+ * formatWithSignificantDecimals(5.0); // "5"
+ *
+ * @param value The number to format.
+ * @returns The formatted string.
+ */
+const formatWithSignificantDecimals = (value: number): string => {
```
srikanthccv:
```suggestion
const formatDecimalWithLeadingZeros = (value: number): string => {
```

frontend/src/components/Graph/yAxisConfig.ts
```diff
@@ -1,49 +1,100 @@
...
+
+	const [integerPart, decimalPart = ''] = numStr.split('.');
+
+	// If there's no decimal part, the integer part is the result.
+	if (!decimalPart) {
+		return integerPart;
+	}
+
+	// Find the index of the first non-zero digit in the decimal part.
+	const firstSignificantIndex = decimalPart.search(/[^0]/);
```
srikanthccv:
```suggestion
	const firstNonZeroIndex = decimalPart.search(/[^0]/);
```

frontend/src/components/Graph/yAxisConfig.ts
```diff
@@ -1,49 +1,100 @@
...
+	// Handle non-numeric or special values first.
+	if (isNaN(numValue)) return 'NaN';
+	if (numValue === Infinity) return '∞';
+	if (numValue === -Infinity) return '-∞';
+
+	const decimalPlaces = value.split('.')[1]?.length || undefined;
+
+	// Use high-precision formatter for the 'none' format.
+	if (format === 'none') {
+		return formatWithSignificantDecimals(numValue);
```
srikanthccv:
```suggestion
		return formatDecimalWithLeadingZeros(numValue);
```

frontend/src/components/Graph/yAxisConfig.ts
```diff
@@ -1,49 +1,100 @@
...
+	// Find the index of the first non-zero digit in the decimal part.
+	const firstSignificantIndex = decimalPart.search(/[^0]/);
+
+	// If the decimal part consists only of zeros, return just the integer part.
+	if (firstSignificantIndex === -1) {
+		return integerPart;
+	}
+
+	// Determine the number of decimals to keep: leading zeros + up to 3 significant digits.
+	const decimalsToKeep = firstSignificantIndex + 3;
```
srikanthccv:
```suggestion
	// Determine the number of decimals to keep: leading zeros + up to 5 significant digits.
	const decimalsToKeep = firstSignificantIndex + 5;
```

frontend/src/components/Graph/yAxisConfig.ts
```diff
@@ -1,49 +1,100 @@
...
-					}
-				} else if (nonZeroCtr) {
-					decimalPrecision = idx;
-					break;
-				}
-			}
+		const formatter = getValueFormat(format);
+		const formattedValue = formatter(
+			numValue,
+			decimalPlaces && decimalPlaces >= 3 ? decimalPlaces : 3,
```
srikanthccv:
```suggestion
			decimalPlaces && decimalPlaces >= 5 ? decimalPlaces : 5,
```

frontend/src/components/Graph/yAxisConfig.ts
```diff
@@ -1,49 +1,100 @@
...
-				parseFloat(value),
-				decimalPrecision,
-				undefined,
-				undefined,
-			),
-		);
+		return formattedValueToString(formattedValue);
 	} catch (error) {
-		console.error(error);
+		console.error('Error applying formatter:', error);
```
srikanthccv: Log to sentry

frontend/src/components/Graph/yAxisConfig.ts
```diff
@@ -1,49 +1,107 @@
...
+		const formatter = getValueFormat(format);
+		const formattedValue = formatter(
+			numValue,
+			decimalPlaces && decimalPlaces >= 5 ? decimalPlaces : 5,
+			undefined,
+		);
+		if (formattedValue.text && formattedValue.text.includes('.')) {
+			formattedValue.text = formatDecimalWithLeadingZeros(
+				parseFloat(formattedValue.text),
+			);
```
srikanthccv, line 139: What's going on here and why is it needed?

frontend/src/components/Graph/yAxisConfig.ts
```diff
@@ -1,49 +1,107 @@
...
+			}`,
+			level: 'error',
+		});
+		// Fallback
+		return numValue.toString();
 	}
-	return `${parseFloat(value)}`;
 };
 
 export const getToolTipValue = (value: string, format?: string): string => {
```
srikanthccv: Why are tooltip values not updated to follow the same precision logic?

frontend/src/components/Graph/yAxisConfig.ts
SagarRajput-7: @srikanthccv - I didn't get which particular part you are referring to, I have explained the process, hope this will cover your question.

Here,
Just before this, we pass the value through the formatter from the `grafana/data` library, with a minimum of 5 digits (or the original precision if it's 5 or more digits). When the y-axis unit is provided, this is accordingly converted to the relevant format.

So now, when we come to this particular step (above tagged portion), we will have values in different formats:
- Decimal values: `0.25` → `0.25000`, `0.256787` → `0.256787`
- Scientific notation: `0.000000250034` → `2.50034e-7` (for very small values)

Then function `formatDecimalWithLeadingZeros` performs our desired operations on values that contain decimal points:
- preserving leading zeros after the decimal point
- showing up to 5 digits after the first non-zero decimal digit
- removing unnecessary trailing zeros

The condition `if (formattedValue.text && formattedValue.text.includes('.'))` ensures this custom formatting is applied to both decimal and scientific notation values that contain a decimal point:
- Values like `0.000000250034` → `2.50034e-7` (from Grafana formatter) → `0.00000025003` (custom formatting applied)
- Values like `0.000000001` → `1e-9` (Grafana) → `1e-9` (no decimal point, skips custom formatting)

This selective application (if case) allows us to enhance precision for values that can be meaningfully represented in decimal format while preserving scientific notation for extremely small values that would be unreadable in decimal form.

frontend/src/components/Graph/yAxisConfig.ts
SagarRajput-7: Done

SagarRajput-7: I had this in mind, but at the end, after all the changes, test cases and review comments, this got missed.

frontend/src/components/Graph/yAxisConfig.ts
```diff
@@ -1,46 +1,110 @@
...
-		console.error(error);
+	// Use high-precision formatter for the 'none' format.
+	if (format === 'none') {
+		return formatDecimalWithLeadingZeros(numValue);
 	}
-	return `${parseFloat(value)}`;
-};
 
-export const getToolTipValue = (value: string, format?: string): string => {
+	// For all other standard formats, delegate to grafana/data's built-in formatter.
```
ahrefabhi, line 111: Do we want to keep grafana's reference in the comment?

SagarRajput-7: We can keep it as we are calling the package `grafana/data` for this.

SagarRajput-7:
As discussed we need to add config option to give user choice to decide the precision and have default as 2 - https://signoz-team.slack.com/archives/C080PFAFHS4/p1761447738656399?thread_ts=1760264972.833439&cid=C080PFAFHS4

PR #9089, opened 2025-09-14
fix: normalize context-prefixed field keys

### LINKED ISSUE #8917: /fields/keys and /fields/values do not work if the context is specified in the key
- try using `resource.service.name = ` vs `service.name = ` and see the response from the API for values
- try using `resource.servic` vs `servic` and see the response from the API for keys

niladrix719:
## 📄 Summary

Normalize context‑prefixed field keys (e.g. `resource.service.name`) so `/api/v1/fields/keys` and `/api/v1/fields/values` return the same results as with the canonical `service.name`. Previously, prefixed forms returned empty results for both autocomplete (keys) and value enumeration.

---

## ✅ Changes

- [X] Bug fix: Added key normalization in metadata lookup so resource.<key> aliases resolve to their canonical materialized keys (e.g. `service.name`), applied to both keys and values endpoints.

---

## 🧪 How to Test

1. Ingest at least one trace span with service.name (e.g. demo-backend).
2. Curl (both should match):
`curl "$BASE/api/v1/fields/values?signal=traces&name=service.name`"
`curl "$BASE/api/v1/fields/values?signal=traces&name=resource.service.name"`
3. Autocomplete / keys (both should include service.name):
`curl "$BASE/api/v1/fields/keys?signal=traces&searchText=servic"`
`curl "$BASE/api/v1/fields/keys?signal=traces&searchText=resource.servic"`
4. UI (Traces Explorer):
Run `service.name` = 'demo-backend' and note span count.
Run `resource.service.name` = 'demo-backend' → same count expected.
---

## 🔍 Related Issues

Closes #8917

---

https://github.com/user-attachments/assets/b10bed18-ca8b-462f-ad92-d5ac290be190

---

## 📋 Checklist

- [x] Dev Review
- [X] Manually tested the changes

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> Ensures context‑prefixed keys resolve to their canonical forms with correct context, improving consistency across field key/value lookups.
> 
> - Infer and normalize context from `searchText` and `name` in `parse.go` when `fieldContext` is unspecified, guarded by `isContextValidForSignal` for logs/traces/metrics
> - Add `isContextValidForSignal` helper to restrict inferred contexts per signal
> - Update integration tests for logs and traces to validate context‑prefixed key discovery and value enumeration; verify `metric.` is not treated as a context for logs and include `metric.domain_id` attribute in fixtures
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8cdbc61ec68896022a1192d63810ca47fb626b2d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/types/querybuildertypes/querybuildertypesv5/formula.go
+++ b/pkg/types/querybuildertypes/querybuildertypesv5/formula.go
@@ -153,10 +153,28 @@ func NewFormulaEvaluator(expressionStr string, canDefaultZero map[string]bool) (
 		return nil, errors.NewInvalidInputf(errors.CodeInvalidInput, "failed to parse expression")
 	}
 
+	// Normalize canDefaultZero keys to match variable casing from expression
+	normalizedCanDefaultZero := make(map[string]bool)
+	vars := expression.Vars()
+	for _, variable := range vars {
+		// If exact match exists, use it
+		if val, ok := canDefaultZero[variable]; ok {
+			normalizedCanDefaultZero[variable] = val
+			continue
+		}
+		// Otherwise try case-insensitive lookup
+		for k, v := range canDefaultZero {
+			if strings.EqualFold(k, variable) {
+				normalizedCanDefaultZero[variable] = v
+				break
+			}
+		}
+	}
+
 	evaluator := &FormulaEvaluator{
 		expression:     expression,
-		variables:      expression.Vars(),
-		canDefaultZero: canDefaultZero,
+		variables:      vars,
+		canDefaultZero: normalizedCanDefaultZero,
 		aggRefs:        make(map[string]aggregationRef),
 	}
 
@@ -281,6 +299,16 @@ func (fe *FormulaEvaluator) buildSeriesLookup(timeSeriesData map[string]*TimeSer
 		// We are only interested in the time series data for the queries that are
 		// involved in the formula expression.
 		data, exists := timeSeriesData[aggRef.QueryName]
+		if !exists {
+			// try case-insensitive lookup
+			for k, v := range timeSeriesData {
+				if strings.EqualFold(k, aggRef.QueryName) {
+					data = v
+					exists = true
+					break
+				}
+			}
+		}
 		if !exists {
 			continue
 		}
--- a/pkg/types/querybuildertypes/querybuildertypesv5/formula_test.go
+++ b/pkg/types/querybuildertypes/querybuildertypesv5/formula_test.go
@@ -864,6 +864,158 @@ func TestComplexExpression(t *testing.T) {
 	}
 }
 
+func TestCaseInsensitiveQueryNames(t *testing.T) {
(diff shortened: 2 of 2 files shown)


srikanthccv:
Briefly looked at it, looks correct. Please add integration tests here https://github.com/SigNoz/signoz/tree/main/tests/integration/src/querier

niladrix719:
Tests are added, and ready for review

srikanthccv:
I am wondering if we should do this at https://github.com/SigNoz/signoz/blob/main/pkg/apis/fields/parse.go and state that the caller of the metadata store must take care of the details to ensure that the `key` is correct, as it was called with the name and other details.

niladrix719:
There will be an import cycle issue between the fields package and the telemetry packages if we move the normalization into parse.go, but I think we can safely move it to https://github.com/SigNoz/signoz/blob/main/pkg/types/telemetrytypes/field.go and do the same thing, WDYT?

srikanthccv:
Why will there be an import-cycle because of the fields `parse.go` because we already import the `telemetrytypes` package there?

The suggestions are powered by `api.go` in that package. The request body is parsed and then calls the metadata store https://github.com/SigNoz/signoz/blob/49306cbe3d233da8caff8bd0f7c3e5d9278a6d5a/pkg/apis/fields/api.go#L72 and https://github.com/SigNoz/signoz/blob/49306cbe3d233da8caff8bd0f7c3e5d9278a6d5a/pkg/apis/fields/api.go#L103 
However, it doesn't account for the fact that field context and data type can be present as part of the search text https://github.com/SigNoz/signoz/blob/49306cbe3d233da8caff8bd0f7c3e5d9278a6d5a/pkg/apis/fields/parse.go#L81

Let's update the parse to use `GetFieldKeyFromKeyText` from `telemetrytypes`

niladrix719:
Got it! Sorry for the late response. I’ve updated `parse.go`

srikanthccv:
Great, I appreciate you getting back to this. I briefly reviewed it, and it appears to be good to me. Will review thoroughly sometime later. 

On a side note, if you fully understand all these changes you have been making to the codebase, Good job!

srikanthccv:
Here are some cases where it fails, imagine there is attribute key `metric.domain_id` in log attributes. When the "signal" is "logs" in the request payload, and the text is `metric.count`, it will search for `count` instead of `metric.domain_id`. Let me know if this is clear?

srikanthccv (approved):
Thanks; 

there is still a chance that some attributes in the logs/traces/metrics have the other context as prefix, example, `resource.some-key` as legit attribute, that falls under `attribute` context but won't be working with this change, however, it should be rare, we will see when such instances come up, if they come up, the querying in `metadata.go` needs to change, or shouldn't it not is question for later?

PR #9099, opened 2025-09-16
fix: add trace perf improvement only when there is any filter

nityanandagohain:
We added a fix to read only the required traces, but it doesn't help in cases where the filter is empty and results in more memory usage.


srikanthccv (approved):
Notes from the discussion

- It's not clear why it's faster
- It was observed that it becomes slower as the number of possible traces of interest (__toe) is in the range of a few hundred thousand. Otherwise, it becomes slower. It's not clear why it becomes slower
- Regardless, the nature of the query is that it will be slower and memory-intensive beyond a certain limit, so there needs to be limits to ensure it won't overload the entire system
- Without the `trace_id IN __toe`, the rows on the right side of the join are going to be as many traces in the time range, which CH loads entirely into memory and is known to be bad. I suggest changing the query so that the right side of the join is the smaller table. 
- Nitya to spend time to find the limits and find any further optimisation possible.

PR #9131, opened 2025-09-19
feat: add evaluation cadence component for alerts v2

amlannandy:
## 📄 Summary

- Add the EvaluationCadence component
- Add the AdvancedOptionItem component as we're using its styles as well

---

## ✅ Changes

---

ex:

- `frontend`
- `backend`
- `devops`
- `bug`
- `enhancement`
- `ui`
- `test`

---

## 👥 Reviewers

- frontend / backend / devops

---

## 🧪 How to Test

1. ...
2. ...
3. ...

---

## 🔍 Related Issues

Closes #

---

https://github.com/user-attachments/assets/b4b5777e-55fb-466c-9009-cb209a210220

---

## 📋 Checklist

- [ ] Dev Review

---

## 👀 Notes for Reviewers

----

> [!IMPORTANT]
> Add `EvaluationCadence` component with custom scheduling and preview features for alerts, including tests and styles.
> 
>   - **Components**:
>     - Add `EvaluationCadence` component in `EvaluationCadence.tsx` for managing alert evaluation cadence.
>     - Add `EditCustomSchedule` component in `EditCustomSchedule.tsx` for editing custom schedules.
>     - Add `EvaluationCadenceDetails` component in `EvaluationCadenceDetails.tsx` for detailed schedule settings.
>     - Add `EvaluationCadencePreview` component in `EvaluationCadencePreview.tsx` for previewing schedules.
>     - Add `AdvancedOptionItem` component for reusable UI elements.
>   - **Styles**:
>     - Add styles in `styles.scss` for `EvaluationCadence` and `AdvancedOptionItem` components.
>   - **Tests**:
>     - Add unit tests for `AdvancedOptionItem`, `EditCustomSchedule`, `EvaluationCadence`, `EvaluationCadenceDetails`, and `EvaluationCadencePreview` in `__tests__` directory.
>   - **Constants**:
>     - Update `constants.ts` with new options for evaluation cadence.
>   - **Context**:
>     - Update `constants.ts` in `context` to include initial states for new components.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 9559dca224fd755c783696b6bc63a731be78cc9c. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EditCustomSchedule.tsx
+++ b/frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EditCustomSchedule.tsx
@@ -1,0 +1,104 @@
+import { Button, Typography } from 'antd';
+import { useCreateAlertState } from 'container/CreateAlertV2/context';
+import { INITIAL_ADVANCED_OPTIONS_STATE } from 'container/CreateAlertV2/context/constants';
+import { IEditCustomScheduleProps } from 'container/CreateAlertV2/EvaluationSettings/types';
+import { Calendar1, Edit, X } from 'lucide-react';
+import { useMemo } from 'react';
+
+function EditCustomSchedule({
+	setIsEvaluationCadenceDetailsVisible,
+	setIsPreviewVisible,
+}: IEditCustomScheduleProps): JSX.Element {
+	const { advancedOptions, setAdvancedOptions } = useCreateAlertState();
+
+	const displayText = useMemo(() => {
+		if (advancedOptions.evaluationCadence.mode === 'custom') {
+			return (
+				<Typography.Text>
+					<Typography.Text>Every</Typography.Text>
+					<Typography.Text className="highlight">
+						{advancedOptions.evaluationCadence.custom.repeatEvery
+							.charAt(0)
+							.toUpperCase() +
+							advancedOptions.evaluationCadence.custom.repeatEvery.slice(1)}
+					</Typography.Text>
+					{advancedOptions.evaluationCadence.custom.repeatEvery !== 'day' && (
+						<>
+							<Typography.Text>on</Typography.Text>
+							<Typography.Text className="highlight">
+								{advancedOptions.evaluationCadence.custom.occurence
+									.map(
+										(occurence) => occurence.charAt(0).toUpperCase() + occurence.slice(1),
+									)
+									.join(', ')}
+							</Typography.Text>
+						</>
+					)}
+					<Typography.Text>at</Typography.Text>
+					<Typography.Text className="highlight">
+						{advancedOptions.evaluationCadence.custom.startAt}
+					</Typography.Text>
+				</Typography.Text>
+			);
+		}
+		return (
+			<Typography.Text>
+				<Typography.Text>Starting on</Typography.Text>
+				<Typography.Text className="highlight">
+					{advancedOptions.evaluationCadence.rrule.date?.format('DD/MM/YYYY')}
+				</Typography.Text>
+				<Typography.Text>at</Typography.Text>
+				<Typography.Text className="highlight">
+					{advancedOptions.evaluationCadence.rrule.startAt}
+				</Typography.Text>
+			</Typography.Text>
+		);
+	}, [advancedOptions.evaluationCadence]);
+
+	const handleEdit = (): void => {
+		setIsEvaluationCadenceDetailsVisible(true);
+	};
+
+	const handlePreview = (): void => {
+		setIsPreviewVisible(true);
+	};
+
+	const handleDiscard = (): void => {
+		setIsEvaluationCadenceDetailsVisible(false);
+		setAdvancedOptions({
+			type: 'SET_EVALUATION_CADENCE',
+			payload: INITIAL_ADVANCED_OPTIONS_STATE.evaluationCadence,
+		});
+		setAdvancedOptions({
+			type: 'SET_EVALUATION_CADENCE_MODE',
+			payload: 'default',
+		});
+	};
+
+	return (
+		<div className="edit-custom-schedule">
+			{displayText}
+			<div className="button-row">
+				<Button.Group>
+					<Button type="default" onClick={handleEdit}>
+						<Edit size={12} />
+						<Typography.Text>Edit custom schedule</Typography.Text>
+					</Button>
+					<Button type="default" onClick={handlePreview}>
+						<Calendar1 size={12} />
+						<Typography.Text>Preview</Typography.Text>
+					</Button>
+					<Button
+						data-testid="discard-button"
+						type="default"
+						onClick={handleDiscard}
+					>
+						<X size={12} />
+					</Button>
+				</Button.Group>
+			</div>
+		</div>
+	);
+}
+
+export default EditCustomSchedule;
--- a/frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EditCustomSchedule.tsx
+++ b/frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EditCustomSchedule.tsx
@@ -2,7 +2,7 @@ import { Button, Typography } from 'antd';
 import { useCreateAlertState } from 'container/CreateAlertV2/context';
 import { INITIAL_ADVANCED_OPTIONS_STATE } from 'container/CreateAlertV2/context/constants';
 import { IEditCustomScheduleProps } from 'container/CreateAlertV2/EvaluationSettings/types';
-import { Calendar1, Edit, X } from 'lucide-react';
+import { Calendar1, Edit, Trash } from 'lucide-react';
 import { useMemo } from 'react';
 
 function EditCustomSchedule({
@@ -93,7 +93,7 @@ function EditCustomSchedule({
 						type="default"
 						onClick={handleDiscard}
 					>
-						<X size={12} />
+						<Trash size={12} />
 					</Button>
 				</Button.Group>
 			</div>
--- a/frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadenceDetails.tsx
+++ b/frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadenceDetails.tsx
@@ -3,8 +3,8 @@ import TextArea from 'antd/lib/input/TextArea';
 import classNames from 'classnames';
 import { useCreateAlertState } from 'container/CreateAlertV2/context';
 import { AdvancedOptionsState } from 'container/CreateAlertV2/context/types';
-import { Calendar, Code, Edit3Icon, Info } from 'lucide-react';
-import { useMemo, useState } from 'react';
+import { Code, Edit3Icon } from 'lucide-react';
+import { useEffect, useMemo, useState } from 'react';
 
 import {
 	EVALUATION_CADENCE_REPEAT_EVERY_MONTH_OPTIONS,
@@ -19,16 +19,23 @@ import {
 	buildAlertScheduleFromRRule,
 	isValidRRule,
 } from '../utils';
+import { ScheduleList } from './EvaluationCadencePreview';
 
 function EvaluationCadenceDetails({
 	setIsOpen,
+	setIsCustomScheduleButtonVisible,
 }: IEvaluationCadenceDetailsProps): JSX.Element {
 	const { advancedOptions, setAdvancedOptions } = useCreateAlertState();
 	const [evaluationCadence, setEvaluationCadence] = useState<
 		AdvancedOptionsState['evaluationCadence']
 	>({
 		...advancedOptions.evaluationCadence,
+		mode: 'custom',
 	});
+
+	const [searchTimezoneString, setSearchTimezoneString] = useState('');
+	const [occurenceSearchString, setOccurenceSearchString] = useState('');
+	const [repeatEverySearchString, setRepeatEverySearchString] = useState('');
 
 	const tabs = [
 		{
@@ -51,6 +58,28 @@ function EvaluationCadenceDetails({
 			? EVALUATION_CADENCE_REPEAT_EVERY_WEEK_OPTIONS
 			: EVALUATION_CADENCE_REPEAT_EVERY_MONTH_OPTIONS;
 
+	useEffect(() => {
+		if (!evaluationCadence.custom.occurence.length) {
+			const today = new Date();
+			const dayOfWeek = today.getDay();
+			const dayOfMonth = today.getDate();
+
+			const occurence =
+				evaluationCadence.custom.repeatEvery === 'week'
+					? EVALUATION_CADENCE_REPEAT_EVERY_WEEK_OPTIONS[dayOfWeek].value
+					: dayOfMonth.toString();
+
+			setEvaluationCadence({
+				...evaluationCadence,
+				custom: {
+					...evaluationCadence.custom,
+					occurence: [occurence],
+				},
+			});
+		}
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [evaluationCadence.custom.repeatEvery]);
+
 	const EditorView = (
 		<div className="editor-view" data-testid="editor-view">
 			<div className="select-group">
@@ -69,6 +98,9 @@ function EvaluationCadenceDetails({
 						})
 					}
 					placeholder="Select repeat every"
+					showSearch
+					searchValue={repeatEverySearchString}
+					onSearch={setRepeatEverySearchString}
 				/>
 			</div>
 			{evaluationCadence.custom.repeatEvery !== 'day' && (
@@ -88,6 +120,9 @@ function EvaluationCadenceDetails({
 							})
 						}
 						placeholder="Select day(s)"
+						showSearch
+						searchValue={occurenceSearchString}
+						onSearch={setOccurenceSearchString}
 					/>
 				</div>
 			)}
@@ -121,6 +156,9 @@ function EvaluationCadenceDetails({
 						})
 					}
 					placeholder="Select timezone"
+					onSearch={setSearchTimezoneString}
+					searchValue={searchTimezoneString}
+					showSearch
 				/>
 			</div>
 		</div>
@@ -177,10 +215,7 @@ function EvaluationCadenceDetails({
 
 	const handleDiscard = (): void => {
 		setIsOpen(false);
-		setAdvancedOptions({
-			type: 'SET_EVALUATION_CADENCE_MODE',
-			payload: 'default',
-		});
+		setIsCustomScheduleButtonVisible(true);
 	};
 
 	const handleSaveCustomSchedule = (): void => {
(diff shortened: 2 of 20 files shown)


frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadencePreview.tsx
```diff
@@ -0,0 +1,107 @@
...
+							<div className="no-schedule" data-testid="no-schedule">
+								<Info size={32} />
+								<Typography.Text>
+									Please fill the relevant information to generate a schedule
+								</Typography.Text>
+							</div>
+						)}
+					</div>
+				</div>
+			</div>
```
srikanthccv, line 113: Is there a reason why this is repeated in both `EvaluationCadenceDetails.tsx` and here?

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadenceDetails.tsx
```diff
@@ -0,0 +1,350 @@
...
+						setEvaluationCadence({
+							...evaluationCadence,
+							custom: {
+								...evaluationCadence.custom,
+								timezone: value,
+							},
+						})
+					}
+					placeholder="Select timezone"
+				/>
```
srikanthccv, line 171: Imagine the pain one has to go through select any timezone that's not a immediate one in the list. They would have to scroll up and down to figure out. Every select should be searchable.

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadenceDetails.tsx
```diff
@@ -0,0 +1,350 @@
...
+							...evaluationCadence,
+							custom: {
+								...evaluationCadence.custom,
+								repeatEvery: value,
+								occurence: [],
+							},
+						})
+					}
+					placeholder="Select repeat every"
+				/>
```
srikanthccv, line 113: Same as above for search

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadenceDetails.tsx
```diff
@@ -0,0 +1,350 @@
...
+			</div>
+			{evaluationCadence.custom.repeatEvery !== 'day' && (
+				<div className="select-group">
+					<Typography.Text>ON DAY(S)</Typography.Text>
+					<Select
+						options={occurenceOptions}
+						value={evaluationCadence.custom.occurence || null}
```
srikanthccv, line 120: When there is no `occurence`, pass the current day as the value. It may not be the day that user wants to set, however, if it is indeed the day they are interested in then it's one less interaction they need to perform.

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadenceDetails.tsx
```diff
@@ -0,0 +1,350 @@
...
+							})
+						}
+						placeholder="Select day(s)"
+					/>
+				</div>
+			)}
+			<div className="select-group">
+				<Typography.Text>AT</Typography.Text>
+				<TimeInput
+					value={evaluationCadence.custom.startAt}
```
srikanthccv, line 141: Same reasoning as above, set the startAt as the current user time.

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadenceDetails.tsx
```diff
@@ -0,0 +1,350 @@
...
+							},
+						})
+					}
+				/>
+			</div>
+			<div className="select-group">
+				<Typography.Text>TIMEZONE</Typography.Text>
+				<Select
+					options={TIMEZONE_DATA}
+					value={evaluationCadence.custom.timezone || null}
```
srikanthccv, line 157: Same, populate the user timezone if the value is not present.

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadenceDetails.tsx
```diff
@@ -0,0 +1,350 @@
...
+			</div>
+		</div>
+	);
+
+	const RRuleView = (
+		<div className="rrule-view" data-testid="rrule-view">
+			<div className="select-group">
+				<Typography.Text>STARTING ON</Typography.Text>
+				<DatePicker
+					value={evaluationCadence.rrule.date}
```
srikanthccv, line 181: Same, use the current date if not present.

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadenceDetails.tsx
```diff
@@ -0,0 +1,350 @@
...
+							},
+						})
+					}
+					placeholder="Select date"
+				/>
+			</div>
+			<div className="select-group">
+				<Typography.Text>AT</Typography.Text>
+				<TimeInput
+					value={evaluationCadence.rrule.startAt}
```
srikanthccv, line 197: Use the current time as `startAt` instead of `00:00:00`

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadenceDetails.tsx
```diff
@@ -0,0 +1,350 @@
...
+			/>
+		</div>
+	);
+
+	const handleDiscard = (): void => {
+		setIsOpen(false);
+		setAdvancedOptions({
+			type: 'SET_EVALUATION_CADENCE_MODE',
+			payload: 'default',
+		});
```
srikanthccv: This is incorrect. Just because they are discarding doesn't mean they want to go back to `default` mode. You can verify this by adding a custom schedule, then go to edit mode and then discard. It will be clear.

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EditCustomSchedule.tsx
```diff
@@ -0,0 +1,104 @@
...
+					<Button type="default" onClick={handlePreview}>
+						<Calendar1 size={12} />
+						<Typography.Text>Preview</Typography.Text>
+					</Button>
+					<Button
+						data-testid="discard-button"
+						type="default"
+						onClick={handleDiscard}
+					>
+						<X size={12} />
```
srikanthccv: I think this should be a delete icon instead but let's come back to this point later.

frontend/src/container/CreateAlertV2/EvaluationSettings/__tests__/EvaluationCadencePreview.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const mockSetIsOpen = jest.fn();
+
+describe('EvaluationCadencePreview', () => {
+	it('should render list of dates when schedule is generated', () => {
+		render(<EvaluationCadencePreview isOpen setIsOpen={mockSetIsOpen} />);
+		expect(screen.getByTestId('schedule-preview')).toBeInTheDocument();
+	});
+
+	it('should render empty state when no schedule is generated', () => {
```
srikanthccv, line 23: Please add a test where there are schedule. The tests should include
- custom
  - day
  - week
     - use some day
  - month
     - use 31 day
     - use 28th day of Feb
 - with custom at and timezone
- rrule

frontend/src/container/CreateAlertV2/EvaluationSettings/__tests__/EvaluationCadenceDetails.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+						occurence: ['1'],
+					},
+				},
+			},
+		} as any);
+		render(<EvaluationCadenceDetails isOpen setIsOpen={mockSetIsOpen} />);
+
+		// Verify that the schedule preview is shown because days are selected
+		expect(screen.getByTestId(SCHEDULE_PREVIEW_TEST_ID)).toBeInTheDocument();
+		expect(screen.queryByTestId(NO_SCHEDULE_TEST_ID)).not.toBeInTheDocument();
```
srikanthccv, line 228: Is it general practice that you just expect the elemets to be there with id or should we assert on text of the elemente rendered?

frontend/src/container/CreateAlertV2/EvaluationSettings/__tests__/EvaluationCadenceDetails.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+	it('should render the evaluation cadence details component with editor mode in daily occurence by default', () => {
+		render(<EvaluationCadenceDetails isOpen setIsOpen={mockSetIsOpen} />);
+		expect(screen.getByText('Add Custom Schedule')).toBeInTheDocument();
+
+		expect(screen.getByTestId(EDITOR_VIEW_TEST_ID)).toBeInTheDocument();
+		expect(screen.queryByTestId('rrule-view')).not.toBeInTheDocument();
+
+		expect(screen.getByText('REPEAT EVERY')).toBeInTheDocument();
+		expect(screen.getByText('AT')).toBeInTheDocument();
+		expect(screen.getByText('TIMEZONE')).toBeInTheDocument();
```
srikanthccv, line 73: Ok, I see we expect text to be in the document

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadencePreview.tsx
amlannandy: No I should have extracted and make it a common util. Will do so now

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadenceDetails.tsx
amlannandy: Yes here the browser timezone is selected by default

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadenceDetails.tsx
amlannandy: Ah got it. Yeah this is a bug. Will fix it.

frontend/src/container/CreateAlertV2/EvaluationSettings/__tests__/EvaluationCadenceDetails.test.tsx
amlannandy: Usually I tried to assert by the UI text itself. For complicated components, I used test id. So it's a mixture

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EditCustomSchedule.tsx
amlannandy: I think the cross icon makes more sense as we're just discarding the staged changes rather than deleting something.

frontend/src/container/CreateAlertV2/EvaluationSettings/__tests__/EvaluationCadencePreview.test.tsx
amlannandy: Ah okay. I didn't test the schedule content logic itself as we're already testing and verifying the buildSchedule util itself. But I'll add similar tests here as well

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EvaluationCadenceDetails.tsx
amlannandy: yes we already select the current date by default here

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationCadence/EditCustomSchedule.tsx
srikanthccv: This isn't discard right? This is the cross button after the "[Edit custom schedule] [Preview] X". What is the intent of this cross icon here? I imagined it's for deleting the custom schedule.

frontend/src/container/CreateAlertV2/EvaluationSettings/__tests__/EvaluationCadencePreview.test.tsx
srikanthccv: My thoughts on this
- the `buildSchedule` you should test for the logical correctness of the `buildSchedule`, which you already did in previour PR.
- Here you should test for the rendered text (i.e main goal of the test here should be text tests). While the point that we tested it already in `buildSchedule` makes some sense, you shouldn't assume that a correct `buildSchedule` means a correctly rendered schedule. You could have a correct `buildSchedule` and may not render anything or render something incorrectly because rendering has it's own logic. So the test here should target text beyone just the presense of some Element.

amlannandy: Okay. I've added a test for it. Please let me know if that works

amlannandy:
@srikanthccv Have addressed your comments

frontend/src/container/CreateAlertV2/context/constants.ts
```diff
@@ -111,7 +111,7 @@ export const INITIAL_EVALUATION_WINDOW_STATE: EvaluationWindowState = {
 	windowType: 'rolling',
 	timeframe: '5m0s',
 	startingAt: {
-		time: '00:00:00',
+		time: dayjs().format('HH:mm:ss'),
```
srikanthccv, line 114: Using `dayjs().format('HH:mm:ss')` in the constants will cause unexpected behavior. These values get set when the alert assets load(is this correct understanding?). If someone has the app open without refreshing, the input fields will show the time from when the page initially loaded - potentially hours ago. This will confuse users who see outdated timestamps and wonder where these values came from. They'll likely contact support asking why the form is pre-filled with old times. Instead, if we populate the fields with the current time when they actually create an alert, it's clear and intuitive. What are your thoughts?

frontend/src/container/CreateAlertV2/EvaluationSettings/__tests__/AdvancedOptionItem.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+		expect(inputElement).toHaveValue(TEST_VALUE);
+
+		// Toggle off
+		await user.click(switchElement);
+		expect(screen.queryByTestId(TEST_INPUT_TEST_ID)).not.toBeInTheDocument();
+
+		// Toggle back on - input should be recreated (fresh state)
+		await user.click(switchElement);
+		const inputElementAgain = screen.getByTestId(TEST_INPUT_TEST_ID);
+		expect(inputElementAgain).toHaveValue(''); // Fresh input, no previous state
```
srikanthccv: Should we lose the work b/w toggles?

It would make sense to start with fresh state or reset when saving the alert. However, should the intermediate toggles lose the work?

frontend/src/container/CreateAlertV2/EvaluationSettings/__tests__/EditCustomSchedule.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+import { INITIAL_ADVANCED_OPTIONS_STATE } from 'container/CreateAlertV2/context/constants';
+
+import { TIMEZONE_DATA } from '../constants';
+import EditCustomSchedule from '../EvaluationCadence/EditCustomSchedule';
+
+const mockSetAdvancedOptions = jest.fn();
+jest.spyOn(alertState, 'useCreateAlertState').mockReturnValue({
+	advancedOptions: INITIAL_ADVANCED_OPTIONS_STATE,
+	setAdvancedOptions: mockSetAdvancedOptions,
+} as any);
```
srikanthccv: question: why `as any` here?

srikanthccv (commented):
Overall LGTM, few questions.

frontend/src/container/CreateAlertV2/EvaluationSettings/__tests__/EditCustomSchedule.test.tsx
amlannandy: EditCustomSchedule component only uses `advancedOptions` and `setAdvancedOptions`. So I have only mocked those 2 values and omitted the other fields. Mocking all the context values isn't required here, so added `as any` here to remove the warning.

srikanthccv: I am curious what's warning? 

> So I have only mocked those 2 values and omitted the other fields. Mocking all the context values isn't required here

Right, however, we could instead have consistent whole context mock that can be shared?

frontend/src/container/CreateAlertV2/EvaluationSettings/__tests__/AdvancedOptionItem.test.tsx
amlannandy: Okay that makes sense. Will change it

frontend/src/container/CreateAlertV2/context/constants.ts
amlannandy: You mean the time when we actually make the API call and "save" the alert? Wouldn't that be confusing if the user had earlier seen a different time in the alert schedule?
Maybe what we can do is, when user expands the EvaluationCadence component, we set the current time to exactly that moment?

srikanthccv: > Maybe what we can do is, when user expands the EvaluationCadence component, we set the current time to exactly that moment?

This is what I meant. Right now the logic is, the time set at the time of `constants` init which is never reset until the reload happens that can be confusing.

amlannandy: Got it. That makes sense

frontend/src/container/CreateAlertV2/EvaluationSettings/__tests__/EditCustomSchedule.test.tsx
amlannandy: I meant it will throw a type error. But yeah I can define a single constant and use it everywhere I guess. We just override it on test case level when we need to mock something different.

srikanthccv: Are you aware of the test fixtures https://stackoverflow.com/questions/12071344/what-are-fixtures-in-programming? Is this a thing in frontend testing world?

srikanthccv: The way I see it, we could have a test fixture for useCreateAlertState that serves as a mock across all our test files that require it.

amlannandy:
@srikanthccv Addressed your comments

srikanthccv (approved):

srikanthccv (approved):

PR #9136, opened 2025-09-19
feat(authz): build role module

vikrantgupta25:
contributes to - https://github.com/SigNoz/platform-pod/issues/1188


pkg/modules/role/role.go
```diff
@@ -0,0 +1,66 @@
...
+	"net/http"
+
+	"github.com/SigNoz/signoz/pkg/types/authtypes"
+	"github.com/SigNoz/signoz/pkg/types/roletypes"
+	"github.com/SigNoz/signoz/pkg/valuer"
+)
+
+type Module interface {
+	// Creates the role metadata
+	Create(context.Context, valuer.UUID, *roletypes.PostableRole) (*roletypes.GettableRole, error)
```
therealpandey:
```suggestion
	// Creates the role metadata
	Create(context.Context, valuer.UUID, <split-this>) (*roletypes.Role, error)
```
Postable, Gettable and Updatable types are just for the http layer. The Module works on a central entity `Role`. It can take a role and return a role. It can internally play with StorableRole.

Flow
```
PostableRole -> decomposed -> passed to module -> storable role -> role -> sent back to handler -> gettable role -> user
```

Everyone needs to have Role. Now if Role == StorableRole, that is okay. If Role == GettableRole, that is also okay.

pkg/types/authtypes/resource.go
```diff
@@ -12,26 +12,36 @@ type resource struct {
 	name Name
 }
 
-func MustNewResource(name string) Typeable {
-	return &resource{name: MustNewName(name)}
+func NewResource(name Name) (Typeable, error) {
+	return &resource{name: name}, nil
```
therealpandey: This Typeable implementation is conflicting. For instance, there is an actual Resource defined below and this small `resource` is just a typeable implementation

Let's move them to `typeableResource` in file `typeable_resource.go`

pkg/types/authtypes/transaction.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return err
+	}
+
+	_transaction, err := NewTransaction(shadow.Relation, shadow.Object)
+	if err != nil {
+		return err
+	}
+
+	*transaction = *_transaction
```
therealpandey: Please. No underscores

grandwizard28 (changes requested):

pkg/modules/role/role.go
vikrantgupta25: done

pkg/types/authtypes/transaction.go
vikrantgupta25: done

pkg/types/authtypes/resource.go
vikrantgupta25: Actual Resource is also an instance of the small `resource`, didn't understand the idea here..

therealpandey: There is a Resource type in transaction. Whats the difference?

pkg/modules/dashboard/impldashboard/module.go
```diff
@@ -222,3 +223,7 @@ func (module *module) Collect(ctx context.Context, orgID valuer.UUID) (map[strin
 
 	return dashboardtypes.NewStatsFromStorableDashboards(dashboards), nil
 }
+
+func (module *module) MustGetTypeables() []authtypes.Typeable {
```
therealpandey, line 227: Unit test please. Otherwise this will fail at runtime.

grandwizard28 (changes requested):

pkg/modules/dashboard/impldashboard/module.go
vikrantgupta25: will add detailed tests once i setup the CLI and more registrations

grandwizard28 (approved):

Issue #9080, opened 2025-09-13
Cursor jumps to beginning of the query randomly


srikanthccv:
See the conversation here https://signoz-team.slack.com/archives/C080TMJ8K32/p1757703273284969 and https://signoz-team.slack.com/archives/C080TMJ8K32/p1756306252625419

Ms-Kulkarni:
Can i get access to this conversation please!

PR #9140, opened 2025-09-20
fix: added fix for cursor jump in QB

ahrefabhi:
Fix: [ISSUE:9080](https://github.com/SigNoz/signoz/issues/9080)

### Current Implementation: 
Previously, the IbuilderQuery was updated on every keystroke in the CodeMirror editor. This approach led to unnecessary remounting of components and creation of new objects on each key stroke, as the currentQuery was updated every time.

### Solution: 
The new implementation delays the update of the filter expression in the builder query. It now only updates when the editor is unfocused or when a query is run using the hotkey (Command + Enter). This improves performance by reducing unnecessary updates and operations with every keystroke.

Before: 

https://github.com/user-attachments/assets/0a3db40e-7788-4f85-9c27-1d5ac7d4614c

After:

https://github.com/user-attachments/assets/fd1126f3-296f-4430-bff2-fe8561b1d075

----

> [!IMPORTANT]
> Improves `QuerySearch` performance by updating query only on blur or hotkey, reducing unnecessary updates.
> 
>   - **Behavior**:
>     - Delays `query` update in `QuerySearch` to occur only on editor blur or hotkey run (Command + Enter).
>     - Prevents unnecessary updates on every keystroke.
>   - **State Management**:
>     - Adds `isFocused`, `hasInteractedWithQB`, and `shouldRunQueryPostUpdate` states in `QuerySearch`.
>   - **Testing**:
>     - Adds `QuerySearch.test.tsx` to test new behavior, including `onChange` and `onRun` triggers.
>     - Mocks `CodeMirror` to simulate editor interactions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 3161af20054f12276f0ee4b0d4d23d00d34ef19f. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/src/components/QueryBuilderV2/QueryV2/__tests__/QuerySearch.test.tsx
+++ b/frontend/src/components/QueryBuilderV2/QueryV2/__tests__/QuerySearch.test.tsx
@@ -7,14 +7,11 @@ import { initialQueriesMap } from 'constants/queryBuilder';
 import * as UseQBModule from 'hooks/queryBuilder/useQueryBuilder';
 import React from 'react';
 import { render, screen, userEvent, waitFor } from 'tests/test-utils';
-// Mock API suggestion endpoints to avoid network calls
 import type { QueryKeyDataSuggestionsProps } from 'types/api/querySuggestions/types';
 import { DataSource } from 'types/common/queryBuilder';
 
-// Import the component and mocked APIs after mocks are set up
 import QuerySearch from '../QuerySearch/QuerySearch';
 
-// Mock infrastructure/business dependencies locally for this test
 jest.mock('hooks/useDarkMode', () => ({
 	useIsDarkMode: (): boolean => false,
 }));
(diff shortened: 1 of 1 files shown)


srikanthccv:
@ahrefabhi can you please describe what was the reason for jump behaviour and how this change is fixing it?

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
```diff
@@ -1209,6 +1228,25 @@ function QuerySearch({
 		</div>
 	);
 
+	// Effect to handle query run after update
+	useEffect(
+		() => {
+			// Only run the query post updating the filter expression.
+			// This runs the query in the next update cycle of react, when it's guaranteed that the query is updated.
+			// Because both the things are sequential and react batches the updates so it was still taking the old query.
```
srikanthccv, line 1236: Where is this guaranteed @ahrefabhi? i.e `handleRunQuery` is called only after the expression is updated? isn't `onChange(query)` async update?

Issue #5291, opened 2024-06-20
Pre-populate the alert condition based on the selected `Reduce To` and `Thresholds`


srikanthccv:
We allow users to create an alert directly from the value panel. When they do, they're redirected to a new alert creation page with their query builder state fully preserved. The default alert condition is set to _**A**_ is _**above**_ the threshold _**at least once**_ in the last 5 minutes. However, we can improve this by better auto-populating the alert condition. To make it easier for users, we can leverage two features: 1. Reduce to and 2. Thresholds.

## Only one query is used in the panel

- Reduce to used and no thresholds set:

    - Sum:
        - Query is _A_
        - No change in the default Condition
        - Occurrence change is _**in total**_
        - No change in default Eval window
    - Avg:
        - Query is _A_
        - No change in the default Condition
        - Occurrence change is _**on average**_
        - No change in default Eval window
    - Max:
        - Query is _A_
        - No change in the default Condition
        - No change in default Occurrence
        - No change in default Eval window
    - Min:
        - Query is _A_
        - No change in the default Condition
        - No change in default Occurrence
        - No change in default Eval window
    - Latest:
        - Query is _A_
        - No change in the default Condition
        - No change in default Occurrence
        - No change in default Eval window

- When both Reduce to and thresholds are used: If thresholds are set for a panel, select the threshold item with the highest danger level. For instance, if there are two threshold items—one with a RED color and one with a YELLOW color—choose the RED item.

    - Sum:
        - Query is _A_
        - Condition changes to _**operator used in threshold item**_
        - Occurrence change is _**in total**_
        - No change in default Eval window
        - Rule threshold value changes to _**value of threshold item**_
        - Rule threshold unit changes to _**unit of threshold item**_
    - Avg:
        - Query is _A_
        - Condition changes to _**operator used in threshold item**_
        - Occurrence change is _**on average**_
        - No change in default Eval window
        - Rule threshold value changes to _**value of threshold item**_
        - Rule threshold unit changes to _**unit of threshold item**_
    - Max:
        - Query is _A_
        - Condition changes to _**operator used in threshold item**_
        - No change in default Occurrence
        - No change in default Eval window
        - Rule threshold value changes to _**value of threshold item**_
        - Rule threshold unit changes to _**unit of threshold item**_
    - Min:
        - Query is _A_
        - Condition changes to _**operator used in threshold item**_
        - No change in default Occurrence
        - No change in default Eval window
[...]

PR #9142, opened 2025-09-20
feat: enable users to query meter specific data in alerts

YounixM:
## 📄 Summary

---

## ✅ Changes

---

ex:

- `frontend`
- `backend`
- `devops`
- `bug`
- `enhancement`
- `ui`
- `test`

---

## 👥 Reviewers

- frontend / backend / devops

---

## 🧪 How to Test

1. ...
2. ...
3. ...

---

## 🔍 Related Issues

Closes #

---

---

## 📋 Checklist

- [ ] Dev Review

---

### DIFF
--- a/frontend/src/components/QueryBuilderV2/QueryBuilderV2.tsx
+++ b/frontend/src/components/QueryBuilderV2/QueryBuilderV2.tsx
@@ -22,6 +22,8 @@ export const QueryBuilderV2 = memo(function QueryBuilderV2({
 	showOnlyWhereClause = false,
 	showTraceOperator = false,
 	version,
+	onSignalSourceChange,
+	signalSourceChangeEnabled = false,
 }: QueryBuilderProps): JSX.Element {
 	const {
 		currentQuery,
@@ -175,6 +177,8 @@ export const QueryBuilderV2 = memo(function QueryBuilderV2({
 							queryVariant={config?.queryVariant || 'dropdown'}
 							showOnlyWhereClause={showOnlyWhereClause}
 							isListViewPanel={isListViewPanel}
+							onSignalSourceChange={onSignalSourceChange || ((): void => {})}
+							signalSourceChangeEnabled={signalSourceChangeEnabled}
 						/>
 					) : (
 						currentQuery.builder.queryData.map((query, index) => (
@@ -194,6 +198,8 @@ export const QueryBuilderV2 = memo(function QueryBuilderV2({
 								showOnlyWhereClause={showOnlyWhereClause}
 								isListViewPanel={isListViewPanel}
 								signalSource={config?.signalSource || ''}
+								onSignalSourceChange={onSignalSourceChange || ((): void => {})}
+								signalSourceChangeEnabled={signalSourceChangeEnabled}
 							/>
 						))
 					)}
--- a/frontend/src/components/QueryBuilderV2/QueryV2/MetricsSelect/MetricsSelect.styles.scss
+++ b/frontend/src/components/QueryBuilderV2/QueryV2/MetricsSelect/MetricsSelect.styles.scss
@@ -1,5 +1,10 @@
-.metrics-select-container {
+.metrics-source-select-container {
 	margin-bottom: 8px;
+	display: flex;
+	flex-direction: row;
+	align-items: flex-start;
+	gap: 8px;
+	width: 100%;
 
 	.ant-select-selector {
 		width: 100%;
@@ -42,7 +47,7 @@
 }
 
 .lightMode {
-	.metrics-select-container {
+	.metrics-source-select-container {
 		.ant-select-selector {
 			border: 1px solid var(--bg-vanilla-300) !important;
 			background: var(--bg-vanilla-100);
--- a/frontend/src/components/QueryBuilderV2/QueryV2/MetricsSelect/MetricsSelect.tsx
+++ b/frontend/src/components/QueryBuilderV2/QueryV2/MetricsSelect/MetricsSelect.tsx
@@ -1,20 +1,31 @@
 import './MetricsSelect.styles.scss';
 
+import { Select } from 'antd';
 import { AggregatorFilter } from 'container/QueryBuilder/filters';
 import { useQueryOperations } from 'hooks/queryBuilder/useQueryBuilderOperations';
-import { memo } from 'react';
+import { memo, useState } from 'react';
 import { IBuilderQuery } from 'types/api/queryBuilder/queryBuilderData';
+import { SelectOption } from 'types/common/select';
+
+export const SOURCE_OPTIONS: SelectOption<string, string>[] = [
+	{ value: 'metrics', label: 'Metrics' },
+	{ value: 'meter', label: 'Meter' },
+];
 
 export const MetricsSelect = memo(function MetricsSelect({
 	query,
 	index,
 	version,
 	signalSource,
+	onSignalSourceChange,
+	signalSourceChangeEnabled = false,
 }: {
 	query: IBuilderQuery;
 	index: number;
 	version: string;
 	signalSource: 'meter' | '';
+	onSignalSourceChange: (value: string) => void;
+	signalSourceChangeEnabled: boolean;
 }): JSX.Element {
 	const { handleChangeAggregatorAttribute } = useQueryOperations({
 		index,
@@ -22,8 +33,25 @@ export const MetricsSelect = memo(function MetricsSelect({
 		entityVersion: version,
 	});
 
+	const [source, setSource] = useState<string>(
+		signalSource === 'meter' ? 'meter' : 'metrics',
+	);
+
+	const handleSignalSourceChange = (value: string): void => {
+		setSource(value);
+		onSignalSourceChange(value);
+	};
+
 	return (
-		<div className="metrics-select-container">
+		<div className="metrics-source-select-container">
+			<Select
+				className="source-selector"
+				placeholder="Source"
+				options={SOURCE_OPTIONS}
+				value={source}
+				onChange={handleSignalSourceChange}
+				disabled={!signalSourceChangeEnabled}
+			/>
 			<AggregatorFilter
 				onChange={handleChangeAggregatorAttribute}
 				query={query}
--- a/frontend/src/components/QueryBuilderV2/QueryBuilderV2.styles.scss
+++ b/frontend/src/components/QueryBuilderV2/QueryBuilderV2.styles.scss
@@ -398,7 +398,7 @@
 	}
 
 	.qb-search-container {
-		.metrics-select-container {
+		.metrics-container {
 			margin-bottom: 12px;
 		}
 	}
--- a/frontend/src/components/QueryBuilderV2/QueryV2/QueryV2.tsx
+++ b/frontend/src/components/QueryBuilderV2/QueryV2/QueryV2.tsx
@@ -33,7 +33,13 @@ export const QueryV2 = memo(function QueryV2({
 	showOnlyWhereClause = false,
 	signalSource = '',
 	isMultiQueryAllowed = false,
-}: QueryProps & { ref: React.RefObject<HTMLDivElement> }): JSX.Element {
+	onSignalSourceChange,
+	signalSourceChangeEnabled = false,
+}: QueryProps & {
+	ref: React.RefObject<HTMLDivElement>;
+	onSignalSourceChange: (value: string) => void;
+	signalSourceChangeEnabled: boolean;
+}): JSX.Element {
 	const { cloneQuery, panelType } = useQueryBuilder();
 
 	const showFunctions = query?.functions?.length > 0;
@@ -207,12 +213,14 @@ export const QueryV2 = memo(function QueryV2({
 					<div className="qb-elements-container">
 						<div className="qb-search-container">
 							{dataSource === DataSource.METRICS && (
-								<div className="metrics-select-container">
+								<div className="metrics-container">
 									<MetricsSelect
 										query={query}
 										index={index}
 										version={ENTITY_VERSION_V5}
 										signalSource={signalSource as 'meter' | ''}
+										onSignalSourceChange={onSignalSourceChange}
+										signalSourceChangeEnabled={signalSourceChangeEnabled}
 									/>
 								</div>
 							)}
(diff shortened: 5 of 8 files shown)


YounixM:
cursor review

vikrantgupta25:
bugbot run

vikrantgupta25 (approved):

vikrantgupta25 (dismissed):

frontend/src/components/QueryBuilderV2/QueryV2/MetricsSelect/MetricsSelect.tsx
```diff
@@ -31,8 +49,69 @@ export const MetricsSelect = memo(function MetricsSelect({
 		},
 		[handleChangeAggregatorAttribute, attributeKeys],
 	);
+
+	const { updateAllQueriesOperators, handleSetQueryData } = useQueryBuilder();
+
+	const [source, setSource] = useState<string>('');
```
aks07: Seems like this could have been a useMemo. onSignalSourceChange is setting signalSource and we can derive this value from its change.

frontend/src/components/QueryBuilderV2/QueryV2/MetricsSelect/MetricsSelect.tsx
```diff
@@ -31,8 +49,69 @@ export const MetricsSelect = memo(function MetricsSelect({
...
+				PANEL_TYPES.BAR,
+				DataSource.METRICS,
+				'',
+			),
+		[updateAllQueriesOperators],
+	);
+
+	const handleSignalSourceChange = (value: string): void => {
+		setSource(value);
+		onSignalSourceChange(value);
```
aks07, line 83: onSignalSourceChange name is misleading. It suggests that the execution of this fn is done after the signalSource changes, but this is actually changing the state.

frontend/src/components/QueryBuilderV2/QueryBuilderV2.tsx
```diff
@@ -193,7 +197,9 @@ export const QueryBuilderV2 = memo(function QueryBuilderV2({
 								queryVariant={config?.queryVariant || 'dropdown'}
 								showOnlyWhereClause={showOnlyWhereClause}
 								isListViewPanel={isListViewPanel}
-								signalSource={config?.signalSource || ''}
+								signalSource={query.source as 'meter' | ''}
```
aks07, line 200: Lets avoid inline type def. signalSource should also follow the same type (currently, this is string in places and 'meter' | '' in other)

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationSettings.tsx
```diff
@@ -1,20 +1,72 @@
...
 
+	const { currentQuery } = useQueryBuilder();
+
+	useEffect(() => {
+		// loop through currenttQuery and find the query that matches the selected query
+		const query = currentQuery?.builder?.queryData.find(
+			(query) => query.queryName === thresholdState.selectedQuery,
+		);
+
+		if (query && query.source === 'meter') {
```
aks07: We are calling 3 separete actions here. It works since react would batch the state updates, but the reducer still runs 3 times and defeats the purpose of creating a reducer in the first place. Ideally, we should create a new action for this state update which combines these state changes.

frontend/src/container/CreateAlertV2/context/index.tsx
```diff
@@ -122,7 +124,17 @@ export function CreateAlertProvider(
...
+		if (ingestionLimitFromURL) {
+			const thresholds = cloneDeep(INITIAL_ALERT_THRESHOLD_STATE.thresholds);
+
+			thresholds[0].thresholdValue = parseInt(ingestionLimitFromURL, 10);
+			setThresholdState({
+				type: 'SET_THRESHOLDS',
+				payload: thresholds,
+			});
+		}
+	}, [alertType, ingestionLimitFromURL]);
```
aks07: If this effect needs to run on alertType change, we can add this to handleAlertTypeChange and avoid the useEffect altogether.

aks07 (dismissed):
Added few comment. Approving this since its urgent.

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationSettings.tsx
```diff
@@ -1,20 +1,72 @@
...
+			setThresholdState({
+				type: 'SET_MATCH_TYPE',
+				payload: AlertThresholdMatchType.IN_TOTAL,
+			});
+		} else {
+			setEvaluationWindow({
+				type: 'SET_INITIAL_STATE',
+				payload: INITIAL_EVALUATION_WINDOW_STATE,
+			});
+		}
```
srikanthccv: This wipes out the existing user config for alerts that are already created. That would come as a surprise because they didn't modify it themselves. Flow when it's overridden and potentially at risk of updating the DB entry as well.

1. Create an alert with a new experience -> Slack notification -> Click on alert -> Save it (out of habit, even though they didn't modify anything) -> later realise the window state is reset
2. Create an alert with a new experience -> Slack notification -> Click on alert -> Update query to exclude something -> Save it -> later realise the window state is reset

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationSettings.tsx
srikanthccv: Looking at it, the above reset problem exists if the meter is already created.

frontend/src/constants/query.ts
```diff
@@ -51,4 +51,5 @@ export enum QueryParams {
 	thresholds = 'thresholds',
 	selectedExplorerView = 'selectedExplorerView',
 	variables = 'variables',
+	ingestionLimit = 'ingestionLimit',
```
srikanthccv: @YounixM, we have `thresholds` to pass the threshold data if we coming to alert from somewhere in the product, can we use that instead of this new addition?

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationSettings.tsx
srikanthccv: Looking at the code (just the diff in this PR), it was clear this would reset the user config. However, I've learned that due to side effects, there's a race condition where the outcome is unpredictable. In some cases, the correct window values from rule data may override this (no issue in this case). There is also a risk that this initial state trigger could override the correct values, resulting in a silent misconfiguration. It can happen. Here is an example from local.

image

YounixM: There is a race condition on `RESET`. The reset is being called after other actions and that's making the state unpredictable.

YounixM: The useEffect gets triggered on change on selectedQuery of the eval window. When changed, the requirement is to reset the existing values based on the selectedQuery signal.

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationSettings.tsx
vikrantgupta25: this is the issue in a couple of places with this reducer. Srikanth is gonna create a ticket for fixing this reducer actions everywhere.

frontend/src/components/QueryBuilderV2/QueryV2/MetricsSelect/MetricsSelect.styles.scss
```diff
@@ -1,5 +1,10 @@
-.metrics-select-container {
+.metrics-source-select-container {
 	margin-bottom: 8px;
+	display: flex;
+	flex-direction: row;
+	align-items: flex-start;
+	gap: 8px;
+	width: 100%;
```
amlannandy, line 7: Maybe give a fixed width to the select to avoid the shift?

https://github.com/user-attachments/assets/b2133b6e-0f66-4715-82b3-79105317d040

frontend/src/components/QueryBuilderV2/QueryV2/MetricsSelect/MetricsSelect.tsx
```diff
@@ -31,8 +49,69 @@ export const MetricsSelect = memo(function MetricsSelect({
...
+		setSource(signalSource === 'meter' ? 'meter' : 'metrics');
+	}, [signalSource]);
+
+	const defaultMeterQuery = useMemo(
+		() =>
+			updateAllQueriesOperators(
+				initialQueryMeterWithType,
+				PANEL_TYPES.BAR,
+				DataSource.METRICS,
+				'meter' as 'meter' | '',
```
amlannandy, line 66: Is this typecasting necessary?

frontend/src/container/CreateAlertV2/context/index.tsx
```diff
@@ -122,7 +124,17 @@ export function CreateAlertProvider(
 		setThresholdState({
 			type: 'RESET',
 		});
-	}, [alertType]);
+
+		if (ingestionLimitFromURL) {
```
amlannandy: Like Aditya said, we can move the whole useEffect logic into `handleAlertTypeChange`.

frontend/src/container/MeterExplorer/Explorer/Explorer.tsx
```diff
@@ -50,6 +51,15 @@ function Explorer(): JSX.Element {
...
 	);
 
+	useEffect(() => {
+		handleSetQueryData(0, {
+			...initialQueryMeterWithType.builder.queryData[0],
+			source: 'meter',
+		});
+
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [handleSetQueryData]);
```
amlannandy: Is having `handleSetQueryData` in the dependency array necessary here?

frontend/src/container/IngestionSettings/MultiIngestionSettings.tsx
```diff
@@ -694,6 +703,65 @@ function MultiIngestionSettings(): JSX.Element {
...
+			},
+		};
+
+		const stringifiedQuery = JSON.stringify(query);
+
+		history.push(
+			`${ROUTES.ALERTS_NEW}?showNewCreateAlertsPage=true&${
+				QueryParams.compositeQuery
+			}=${encodeURIComponent(stringifiedQuery)}&${
+				QueryParams.ingestionLimit
```
amlannandy: Reminder to use the thresholds param instead

amlannandy: You can refer to `frontend/src/container/FormAlertRules/usePrefillAlertConditions.ts` and its usage in `frontend/src/container/FormAlertRules/index.tsx` as a reference

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationSettings.tsx
```diff
@@ -1,20 +1,72 @@
...
+				payload: CumulativeWindowTimeframes.CURRENT_DAY,
+			});
+
+			setEvaluationWindow({
+				type: 'SET_STARTING_AT',
+				payload: {
+					time: '00:00:00',
+					number: '0',
+					timezone: 'UTC',
+					unit: 'minutes',
```
amlannandy: We can use the MINUTE enum from `UniversalYAxisUnit`

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationSettings.tsx
```diff
@@ -1,20 +1,72 @@
...
+				type: 'SET_TIMEFRAME',
+				payload: CumulativeWindowTimeframes.CURRENT_DAY,
+			});
+
+			setEvaluationWindow({
+				type: 'SET_STARTING_AT',
+				payload: {
+					time: '00:00:00',
+					number: '0',
+					timezone: 'UTC',
```
amlannandy: We can use the appropriate value from `TIMEZONE_DATA`

frontend/src/container/CreateAlertV2/EvaluationSettings/EvaluationSettings.tsx
```diff
@@ -1,20 +1,72 @@
...
+				payload: 'cumulative',
+			});
+
+			setEvaluationWindow({
+				type: 'SET_TIMEFRAME',
+				payload: CumulativeWindowTimeframes.CURRENT_DAY,
+			});
+
+			setEvaluationWindow({
+				type: 'SET_STARTING_AT',
```
amlannandy: I think we can convert these 3 reducer calls into a single `SET_INITIAL_STATE` call. We can rename this action to something else as well that's more appropriate. Or create a new action as well.

frontend/src/constants/query.ts
YounixM: Done ✅

frontend/src/container/CreateAlertV2/context/index.tsx
YounixM: We need it on alertTypeChange as well as redirect

amlannandy (approved):

frontend/src/components/QueryBuilderV2/QueryBuilderV2.tsx
```diff
@@ -175,6 +177,8 @@ export const QueryBuilderV2 = memo(function QueryBuilderV2({
 							queryVariant={config?.queryVariant || 'dropdown'}
 							showOnlyWhereClause={showOnlyWhereClause}
 							isListViewPanel={isListViewPanel}
```
srikanthccv: curious, why no `signalSource` prop here?

frontend/src/container/CreateAlertV2/context/index.tsx
```diff
@@ -122,7 +127,28 @@ export function CreateAlertProvider(
...
+			}
+
+			setEvaluationWindow({
+				type: 'SET_INITIAL_STATE_FOR_METER',
+			});
+
+			setThresholdState({
+				type: 'SET_MATCH_TYPE',
+				payload: AlertThresholdMatchType.IN_TOTAL,
+			});
```
srikanthccv, line 149: Not an issue for now since only meter alerts implement the `thresholds` URL param, but the fall back to meter will be problematic when we implement https://github.com/SigNoz/signoz/issues/5291.

frontend/src/components/QueryBuilderV2/QueryBuilderV2.tsx
srikanthccv, line 179: curious, why `signalSource` is not here, but it exists below? and it appears to be part of the `filterConfigs` as well. Overall, it's confusing and unclear

vikrantgupta25:
@YounixM there were couple of non blocking comments from @srikanthccv please reply/address them when you see this. Thanks!

PR #9153, opened 2025-09-22
feat: add history.push and safeNavigate method overrides

manika-signoz:
## 📄 Summary

---

## ✅ Changes

---

ex:

- `frontend`
- `backend`
- `devops`
- `bug`
- `enhancement`
- `ui`
- `test`

---

## 👥 Reviewers

- frontend / backend / devops

---

## 🧪 How to Test

1. ...
2. ...
3. ...

---

## 🔍 Related Issues

Closes #

---

---

## 📋 Checklist

- [ ] Dev Review

---

### DIFF
--- a/frontend/src/hooks/useSafeNavigate.ts
+++ b/frontend/src/hooks/useSafeNavigate.ts
@@ -83,6 +83,84 @@ const isDefaultNavigation = (currentUrl: URL, targetUrl: URL): boolean => {
 
 	return newKeys.length > 0;
 };
+// Helper function to determine if an argument is an event
+const isEventObject = (arg: any): boolean =>
+	arg &&
+	(arg instanceof MouseEvent ||
+		arg instanceof KeyboardEvent ||
+		arg.nativeEvent instanceof MouseEvent ||
+		arg.nativeEvent instanceof KeyboardEvent ||
+		arg.metaKey !== undefined ||
+		arg.ctrlKey !== undefined);
+
+// Helper function to extract options from arguments
+const extractOptions = (
+	optionsOrEvent?:
+		| NavigateOptions
+		| React.MouseEvent
+		| MouseEvent
+		| KeyboardEvent,
+	options?: NavigateOptions,
+): NavigateOptions => {
+	const isEvent = isEventObject(optionsOrEvent);
+	const actualOptions = isEvent ? options : (optionsOrEvent as NavigateOptions);
+
+	const shouldOpenInNewTab =
+		isEvent &&
+		((optionsOrEvent as any).metaKey || (optionsOrEvent as any).ctrlKey);
+
+	return {
+		...actualOptions,
+		newTab: shouldOpenInNewTab || actualOptions?.newTab,
+	};
+};
+
+// Helper function to create target URL
+const createTargetUrl = (
+	to: string | SafeNavigateParams,
+	location: { pathname: string; search: string },
+): URL => {
+	if (typeof to === 'string') {
+		return new URL(to, window.location.origin);
+	}
+	return new URL(
+		`${to.pathname || location.pathname}${to.search || ''}`,
+		window.location.origin,
+	);
+};
+
+// Helper function to handle new tab navigation
+const handleNewTabNavigation = (
+	to: string | SafeNavigateParams,
+	location: { pathname: string; search: string },
+): void => {
+	const targetPath =
+		typeof to === 'string'
+			? to
+			: `${to.pathname || location.pathname}${to.search || ''}`;
+	window.open(targetPath, '_blank');
+};
+
+// Helper function to perform navigation
+const performNavigation = (
+	to: string | SafeNavigateParams,
+	navigationOptions: NavigateOptions,
+	navigate: any,
+	location: { pathname: string; search: string },
+): void => {
+	if (typeof to === 'string') {
+		navigate(to, navigationOptions);
+	} else {
+		navigate(
+			{
+				pathname: to.pathname || location.pathname,
+				search: to.search,
+			},
+			navigationOptions,
+		);
+	}
+};
+
 export const useSafeNavigate = (
 	{ preventSameUrlNavigation }: UseSafeNavigateProps = {
 		preventSameUrlNavigation: true,
@@ -90,6 +168,11 @@ export const useSafeNavigate = (
 ): {
 	safeNavigate: (
 		to: string | SafeNavigateParams,
+		optionsOrEvent?:
+			| NavigateOptions
+			| React.MouseEvent
+			| MouseEvent
+			| KeyboardEvent,
 		options?: NavigateOptions,
 	) => void;
 } => {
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/useSafeNavigate.ts
```diff
@@ -83,44 +83,122 @@ const isDefaultNavigation = (currentUrl: URL, targetUrl: URL): boolean => {
...
+			? to
+			: `${to.pathname || location.pathname}${to.search || ''}`;
+	window.open(targetPath, '_blank');
+};
+
+// Helper function to perform navigation
+const performNavigation = (
+	to: string | SafeNavigateParams,
+	navigationOptions: NavigateOptions,
+	navigate: any,
```
ahmadshaheer: can we improve the type safety here?

ahmadshaheer (commented):
can we improve the typing in useSafeNavigate? rest lgtm.

frontend/src/lib/history.ts
```diff
@@ -1,3 +1,64 @@
...
+	// Check if second argument is an event object
+	const isEvent =
+		eventOrState &&
+		(eventOrState instanceof MouseEvent ||
+			eventOrState instanceof KeyboardEvent ||
+			eventOrState.nativeEvent instanceof MouseEvent ||
+			eventOrState.nativeEvent instanceof KeyboardEvent ||
+			eventOrState.metaKey !== undefined ||
+			eventOrState.ctrlKey !== undefined);
+
```
ahmadshaheer: nit: should we use the isEventObject from useSafeNavigate?

frontend/src/hooks/useSafeNavigate.ts
manika-signoz: removed any from the file and used types from react-router-dom-v5-compat wherever possible

frontend/src/lib/history.ts
manika-signoz: done

frontend/__mocks__/useSafeNavigate.ts
```diff
@@ -18,6 +18,26 @@ interface UseSafeNavigateReturn {
...
+	return (
+		arg instanceof MouseEvent ||
+		arg instanceof KeyboardEvent ||
+		('nativeEvent' in arg &&
+			(arg.nativeEvent instanceof MouseEvent ||
+				arg.nativeEvent instanceof KeyboardEvent)) ||
+		'metaKey' in arg ||
+		'ctrlKey' in arg
+	);
+};
```
ahmadshaheer: do we need to mock this? the mocked version seems to be the same as the actual isEventObject in src/hooks/useSafeNavigate.ts https://github.com/SigNoz/signoz/blob/a1115ac65b5d29aed2c38ee76ace1daca26790f4/frontend/src/hooks/useSafeNavigate.ts#L101-L114

manika-signoz: done, was getting a circular dep before so had copied the method, moved this to a new file now

frontend/src/utils/isEventObject.ts
```diff
@@ -0,0 +1,21 @@
...
+	if (!arg || typeof arg !== 'object') return false;
+
+	return (
+		arg instanceof MouseEvent ||
+		arg instanceof KeyboardEvent ||
+		('nativeEvent' in arg &&
+			(arg.nativeEvent instanceof MouseEvent ||
+				arg.nativeEvent instanceof KeyboardEvent)) ||
+		'metaKey' in arg ||
+		'ctrlKey' in arg
```
srikanthccv, line 19: Why do we have these two loose checks here? Purely from the correctness perspective, this `isEventObject` incorrectly returns the regular object with `metaKey` as EventObject. Lets say this function is called with something like `{ metaKey: 'someString', ... }` then it returns `true`, which is incorrect. If it's a genuine event, it would have been already detected with the `instanceof` right? Is there a case where an event object doesn't eval true for 

```
		arg instanceof MouseEvent ||
		arg instanceof KeyboardEvent ||
```
but 

```
		'metaKey' in arg ||
		'ctrlKey' in arg

```

only works.

frontend/src/hooks/useSafeNavigate.ts
```diff
@@ -83,44 +90,112 @@ const isDefaultNavigation = (currentUrl: URL, targetUrl: URL): boolean => {
...
+// Helper function to create target URL
+const createTargetUrl = (
+	to: string | SafeNavigateParams,
+	location: Location,
+): URL => {
+	if (typeof to === 'string') {
+		return new URL(to, window.location.origin);
+	}
+	return new URL(
+		`${to.pathname || location.pathname}${to.search || ''}`,
```
srikanthccv, line 124: What is the reason for fall back to `location.pathname` but for search the fallback is empty string and not `location.search`?

frontend/src/hooks/useSafeNavigate.ts
```diff
@@ -83,44 +90,112 @@ const isDefaultNavigation = (currentUrl: URL, targetUrl: URL): boolean => {
...
+
+// Helper function to handle new tab navigation
+const handleNewTabNavigation = (
+	to: string | SafeNavigateParams,
+	location: Location,
+): void => {
+	const targetPath =
+		typeof to === 'string'
+			? to
+			: `${to.pathname || location.pathname}${to.search || ''}`;
```
srikanthccv, line 137: Same question here.

frontend/src/lib/history.ts
```diff
@@ -1,3 +1,58 @@
...
+}
+
+// Create enhanced history with overridden push method
+const history = baseHistory as EnhancedHistory;
+
+// Store the original push method
+history.originalPush = baseHistory.push;
+
+// Override push to handle meta/ctrl key events and location objects
+history.push = function (
```
srikanthccv, line 35: I am not able to follow why `history.push` should bother about the events at all? This is a code smell i.e When the `sageNavigate` is already handling the navigation including opening in the new tab, why is there an event handling in `history.push`? I don't think it should be concern of `history` i.e history should not handling UI behaviour. It should just manage the navigation state.

srikanthccv:
@manika-signoz @ahmadshaheer left some questions. Please check.

frontend/src/lib/history.ts
srikanthccv: If the original library didn't do it (the idea of opening in new tab with `push` API), then there is a reason. This is violation of separation of concerns. I looked if any other projects had to resort to override and I couldn't find any, affirms my belief that push should not handle events. It should just manage the data layer of pushing entries to stack. 

Here is more detail why this is a violation of separation of concerns.

The history module is intended and should do just one job; manage navigation state (URL history stack). However, now it is getting overridden to multiple things for no good reason. 

1. Navigation state management (its original purpose)
2. UI interaction handling (interpreting user input to decide behavior)

This is the problem at the core of overriding it.

```
history.push('/page', obj);
```

Now, in this example, any or all of following could happen

- Update the URL/history stack
- Open in new tab vs same tab
- Interpret mouse/keyboard events

Please read about it.

Issue #9110, opened 2025-09-17
QB V5: Incorrect implementation for DataTypeCollisionHandledFieldName


nityanandagohain:
As of now the query becomes (toString(severity_number) >= ‘14’) , as based on the value we are converting the key.

https://github.com/SigNoz/signoz/blob/c477ec65da88484f55574b25eba1838de906e42e/pkg/types/telemetrytypes/field.go#L168 this won’t be valid for top level keys.  Should we handle collision based on what different key types are available rather than taking decision on user provided input. Here since user sent 14 as string it should check what types of severity_number is available and do collsion handling based on that which will be none.

Issue #8963, opened 2025-08-31
Lexicographical vs numerical compare for string encoded number comparisions


srikanthccv:
Should it be a numerical comparison? Does it ever make sense to perform lexicographical comparison?

PR #9154, opened 2025-09-23
fix: use numerical comparison instead of lexicographical for string-encoded numbers

nityanandagohain:
Fixes https://github.com/SigNoz/signoz/issues/9110

Fixes https://github.com/SigNoz/signoz/issues/8963

----

> [!IMPORTANT]
> Introduces `DataTypeCollisionHandledFieldName` to handle mixed data types in queries, updating condition builders and adding tests.
> 
>   - **Behavior**:
>     - Introduces `DataTypeCollisionHandledFieldName` in `pkg/querybuilder/fallback_expr.go` to handle mixed data types in queries.
>     - Updates `CollisionHandledFinalExpr` in `fallback_expr.go` to use the new function.
>     - Integrates `DataTypeCollisionHandledFieldName` into condition builders in `condition_builder.go` files for `telemetrylogs`, `telemetrymetadata`, and `telemetrytraces`.
>   - **Tests**:
>     - Adds `TestDataTypeCollisionHandledFieldName` in `fallback_expr_test.go` to test the new function.
>     - Updates `filter_expr_logs_test.go` to include scenarios for data type collisions.
>     - Adds integration tests in `a_logs.py` to verify behavior with mixed data types.
>   - **Misc**:
>     - Adds `IsComparisonOperator` method to `FilterOperator` in `builder_elements.go` to support the new function.
>     - Removes old `DataTypeCollisionHandledFieldName` from `field.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 29dd9db89726762b1639fd45e6f40adde2051d7b. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


pkg/types/telemetrytypes/field.go
```diff
@@ -175,11 +175,10 @@ func DataTypeCollisionHandledFieldName(key *TelemetryFieldKey, value any, tblFie
 		// is used which prepares the
 		// final query as attributes_number['http.status_code'] = 200 giving this error
 		// This following is one way to workaround it
+
+		// if the key is a number, the value is a string, we will let clickHouse handle the conversion
 		case float32, float64:
 			tblFieldName = castFloatHack(tblFieldName)
-		case string:
-			// try to convert the number attribute to string
-			tblFieldName = castString(tblFieldName) // numeric col vs string literal
```
srikanthccv: On a second thought, removing it entirely is not going to work. We should be selective about when we cast. For instance the following does not work with this change but one would expect to work.  
```
severity_number REGEXP '9'
```

A more realistic example would be something that use http code like 

```
http.response.status_code LIKE '5__'
``` 

Of course it could be re-written as `http.response.status_code >= 500` but both should work. So I think we should remove the cast if the `value` is number but as string otherwise we should keep `castString(tblFieldName)`

nityanandagohain: >remove the cast if the value is number

But then our original issue is still not solved right ? where if my query is `severity_number > '9'`

```
SELECT '9' > '19'

Query id: 7d0d71f0-c68a-4689-ba8d-52ad4579bdc2

┌─greater('9', '19')─┐
│                  1 │
└────────────────────┘

```

nityanandagohain: have updated the code to cast only when it's not a comparison operator.

srikanthccv:
@nityanandagohain let's add integration test for this. The change here relies on ClickHouse to handle the comparison. Having a test would tell what all behaviours change from one version to another, so let's add integration test otherwise we won't be able to recall all these details when we want to bump CH version.

nityanandagohain:
Done @srikanthccv

pkg/types/telemetrytypes/field.go
```diff
@@ -137,7 +137,7 @@ type FieldValueSelector struct {
 	Limit         int    `json:"limit"`
 }
 
-func DataTypeCollisionHandledFieldName(key *TelemetryFieldKey, value any, tblFieldName string) (string, any) {
+func DataTypeCollisionHandledFieldName(key *TelemetryFieldKey, value any, tblFieldName string, isComparisonOperator bool) (string, any) {
```
srikanthccv: can we pass the operator directly and use `IsComparisonOperator` inside `DataTypeCollisionHandledFieldName` instead. It will give `DataTypeCollisionHandledFieldName` access to the exact operator used and any functions available.

pkg/types/telemetrytypes/field.go
```diff
@@ -175,11 +175,15 @@ func DataTypeCollisionHandledFieldName(key *TelemetryFieldKey, value any, tblFie
...
+		// if the key is a number, the value is a string, we will let clickHouse handle the conversion
 		case float32, float64:
 			tblFieldName = castFloatHack(tblFieldName)
 		case string:
-			// try to convert the number attribute to string
-			tblFieldName = castString(tblFieldName) // numeric col vs string literal
+			if !isComparisonOperator {
+				// try to convert the number attribute to string
+				tblFieldName = castString(tblFieldName) // numeric col vs string literal
+			}
```
srikanthccv: We should also check if the `value` is indeed a number and if it's not a number we should cast it? For instance, if somebody does `severity_number > 'ERROR'` should we 5xx or cast and compare?

pkg/types/telemetrytypes/field.go
nityanandagohain: Passing the operator results in import cycle, so I used this method.

pkg/types/telemetrytypes/field.go
nityanandagohain: ahh right let me add that.

pkg/types/telemetrytypes/field.go
srikanthccv: Should the function `DataTypeCollisionHandledFieldName` even be here in that case; We can move to https://github.com/SigNoz/signoz/tree/main/pkg/querybuilder?

nityanandagohain: fallback_expr.go will be the right place right ?

srikanthccv: Yes

srikanthccv (commented):
Thanks, LGTM. Taking some time to further process the change.

srikanthccv:
The thing with relying on ClickHouse is that we need to learn all the ways it works and breaks (I asked the time for this exact reason here https://github.com/SigNoz/signoz/pull/9154#pullrequestreview-3298645520). While the most of the queries work, this one doesn't work `severity_number >= '2.0'`. One could argue that this should happen unless someone wants to break it deliberately. To that, I would say it doesn't matter whether or not it is the common case or breaking case, it is expected to not throw error. 


Should the change handle it? Something like this?

```
			if !operator.IsComparisonOperator() || !isNumber {
				// try to convert the number attribute to string
				tblFieldName = castString(tblFieldName) // numeric col vs string literal
			} else {
				tblFieldName = castFloat(tblFieldName)
			}

```

nityanandagohain:
> would say it doesn't matter whether or not it is the common case or breaking case, it is expected to not throw error

exactly it shouldn't matter. 

> Should the change handle it? Something like this?
Instead should we make value = the parsed value to float64 ?

nityanandagohain:
I tested it out, it doesn't throw an error, it just gives me incorrect result.

in this case

```
 (severity_number > '2.0')
   ┌─severity_number─┬─__result_0─┐
1. │ 9               │       1874 │
   └─────────────────┴────────────┘
```


while


```
(toFloat64(severity_number) > '2.0')
   ┌─severity_number─┬─__result_0─┐
1. │ 9               │       1874 │
2. │ 17              │          6 │
   └─────────────────┴────────────┘
```

So seems like a valid change

nityanandagohain:
added the changes @srikanthccv

pkg/querybuilder/fallback_expr_test.go
```diff
@@ -0,0 +1,162 @@
...
+				FieldDataType: telemetrytypes.FieldDataTypeFloat64,
+			},
+			value:             "9.11",
+			tblFieldName:      "attribute_float64_http$$request$$duration",
+			expectedFieldName: "toFloat64(attribute_float64_http$$request$$duration)",
+			expectedValue:     "9.11",
+			operator:          qbtypes.FilterOperatorGreaterThan,
+		},
+		{
+			// we won't cast to string if it's a comparison operator
```
srikanthccv: The comment is misleading here.

pkg/telemetrylogs/filter_expr_logs_test.go
```diff
@@ -424,6 +424,16 @@ func TestFilterExprLogs(t *testing.T) {
 			expectedArgs:          []any{"critical", "resolved", "open", true},
 			expectedErrorContains: "",
 		},
+		{
+			// this will result in failure from the DB side.
+			// user will have to use attribute.status:string > open
```
srikanthccv, line 429: Same misleading comment as we are casting so there won't be failure.

nityanandagohain: thanks corrected them, starting the integration test again

srikanthccv (approved):
Thanks, discussed with @nityanandagohain on our approach to what change requires integration. We want to try being judicious with nature of the change; add integration test if the change relies on ClickHouse to do certain job or change itself involves multiple interconnected components that make assumptions that could fail. We will see how this changes as we add more integration tests slowly as we decided do with TTL also.

@tushar-signoz had a question about the doubling down on Python in our discussion. That's a good food for thought @grandwizard28  because
1. we re-create the insertion logic in here which keeps diverging, if it was go we could make use of the `inmemoey` receiver to simulate the collector service ingestion and let the exporter handle the ingestion
2. re-using the types (a selling point to some extension but if we have solution that generate the types for all languages then maybe not a problem but that's not clear when we have that.

PR #9195, opened 2025-09-25
chore: notification routing | added notificaiton routing via expression based routes

aniketio-ctrl:
1.) Added expression-based notification routing to replace config based alertmanager routing rules
2.) Uses expr-lang/expr library for flexible route matching
3.) Single route as well as single rule can target multiple notification channels.
4.) Added handling for multi threshold routing
5.) Added CRUD APIs: Complete route management.
6.) Added Inhibition for rule with mutliple thresholds 

----

> [!IMPORTANT]
> This PR introduces expression-based notification routing, supporting multiple channels and thresholds, with CRUD APIs and SQL migration for route management.
> 
>   - **Behavior**:
>     - Introduces expression-based notification routing, replacing config-based routing.
>     - Supports multiple notification channels per route and rule.
>     - Adds multi-threshold routing and inhibition for rules with multiple thresholds.
>   - **APIs**:
>     - Adds CRUD APIs for route management in `api.go`.
>     - Updates `TestAlert` function in `alertmanager.go` to handle multiple alerts and receivers.
>   - **Models**:
>     - Adds `RoutePolicy` and related models in `expressionroute.go`.
>     - Updates `NotificationConfig` in `config.go` to include `UsePolicy` and `GroupByAll`.
>   - **Migrations**:
>     - Adds SQL migration `049_add_route_policy.go` to create `route_policy` table and migrate existing rules.
>   - **Tests**:
>     - Adds tests for new routing logic in `api_params_test.go` and `threshold_test.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for e5de6999cbbe83234198e761f429d75452ccd1ac. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


pkg/types/ruletypes/threshold.go
```diff
@@ -64,10 +70,24 @@ type BasicRuleThreshold struct {
 	MatchType      MatchType `json:"matchType"`
 	CompareOp      CompareOp `json:"op"`
 	SelectedQuery  string    `json:"selectedQuery"`
```
srikanthccv: This should be removed from here.

pkg/types/ruletypes/threshold.go
srikanthccv: Make all these functions `GetName`, `Target` etc. on `BasicRuleThreshold` internal.

pkg/types/ruletypes/api_params.go
```diff
@@ -68,6 +70,7 @@ type NotificationSettings struct {
 	NotificationGroupBy []string           `json:"notificationGroupBy,omitempty"`
 	ReNotifyInterval    Duration           `json:"renotify,omitempty"`
```
srikanthccv: - How does one completely disable the re-notification?
- How does one disable the grouping altogether?

pkg/types/alertmanagertypes/expressionroute.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+type RouteStore interface {
+	GetByID(ctx context.Context, orgId string, id string) (*ExpressionRoute, error)
+	Create(ctx context.Context, route *ExpressionRoute) error
+	CreateBatch(ctx context.Context, routes []*ExpressionRoute) error
+	Delete(ctx context.Context, orgId string, id string) error
+	GetAllByKindAndOrgID(ctx context.Context, orgID string, kind ExpressionKind) ([]*ExpressionRoute, error)
+	GetAllByName(ctx context.Context, orgID string, name string) ([]*ExpressionRoute, error)
+	DeleteRouteByName(ctx context.Context, orgID string, name string) error
```
srikanthccv, line 138: There shouldn't be "Name" based deletions.

pkg/types/alertmanagertypes/expressionroute.go
```diff
@@ -0,0 +1,140 @@
...
+
+	return nil
+}
+
+type RouteStore interface {
+	GetByID(ctx context.Context, orgId string, id string) (*ExpressionRoute, error)
+	Create(ctx context.Context, route *ExpressionRoute) error
+	CreateBatch(ctx context.Context, routes []*ExpressionRoute) error
+	Delete(ctx context.Context, orgId string, id string) error
+	GetAllByKindAndOrgID(ctx context.Context, orgID string, kind ExpressionKind) ([]*ExpressionRoute, error)
```
srikanthccv:
```suggestion
	GetAllByKind(ctx context.Context, orgID string, kind ExpressionKind) ([]*ExpressionRoute, error)
```

AndOrgID is redundant here. OrgID as a parameter is expected and used in multiple function in the same interface with `AndOrgID` part.

pkg/types/alertmanagertypes/expressionroute.go
```diff
@@ -0,0 +1,140 @@
...
+		return errors.NewInvalidInputf(errors.CodeInvalidInput, "expression is required")
+	}
+
+	if er.Name == "" {
+		return errors.NewInvalidInputf(errors.CodeInvalidInput, "name is required")
+	}
+
+	if er.OrgID == "" {
+		return errors.NewInvalidInputf(errors.CodeInvalidInput, "organization ID is required")
+	}
```
srikanthccv, line 111: Why is this part of the `ExpressionRoute` json?

pkg/types/alertmanagertypes/expressionroute.go
```diff
@@ -0,0 +1,140 @@
...
+	types.TimeAuditable
+	types.UserAuditable
+
+	// Core routing fields
+	Expression     string         `bun:"expression,type:text,notnull" json:"expression"`
+	ExpressionKind ExpressionKind `bun:"kind,type:text" json:"kind"`
+
+	// Action configuration (stored as JSON)
+	Channels []string `bun:"channels,type:jsonb" json:"channels"`
+	Priority string   `bun:"priority,type:text" json:"priority"`
```
srikanthccv: What is this for?

pkg/types/alertmanagertypes/expressionroute.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	if len(req.Actions.Channels) == 0 {
+		return fmt.Errorf("channels required")
+	}
+	return nil
+}
+
+// ToExpressionRoute converts PolicyRouteRequest to ExpressionRoute
+func (req *PolicyRouteRequest) ToExpressionRoute(orgID, userID string, kind ExpressionKind) *ExpressionRoute {
```
srikanthccv: Where is this used and what is the purpose?

pkg/types/alertmanagertypes/expressionroute.go
```diff
@@ -0,0 +1,140 @@
...
+
+	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/types"
+	"github.com/SigNoz/signoz/pkg/valuer"
+	"github.com/uptrace/bun"
+)
+
+type Actions struct {
+	Channels []string `json:"channels"`
+	Priority string   `json:"priority"`
```
srikanthccv: Same question as earlier. Why is this needed?

pkg/sqlmigration/049_add_notification_routes.go
```diff
@@ -0,0 +1,243 @@
...
+			},
+			userAuditable: userAuditable{
+				CreatedBy: r.CreatedBy,
+				UpdatedBy: r.UpdatedBy,
+			},
+			Expression:     expression,
+			ExpressionKind: "rule",
+			Channels:       gettableRule.PreferredChannels,
+			Name:           r.ID.StringValue(),
+			Description:    fmt.Sprintf("Auto-migrated route from rule %s", gettableRule.AlertName),
```
srikanthccv: Remove this description.

pkg/sqlmigration/049_add_notification_routes.go
```diff
@@ -0,0 +1,243 @@
...
+
+	routes := migration.convertRulesToRoutes(rules, channelsByOrg)
+
+	if len(routes) > 0 {
+		_, err = db.NewInsert().
+			Model(&routes).
+			Exec(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to insert notification routes: %w", err)
+		}
```
srikanthccv: This makes the migration nor idempotent. If the script fails in middle for some reason, and re-runs, this inserts duplicate entries. 

Check the entire migration idempotency.

pkg/query-service/rules/manager.go
```diff
@@ -350,39 +350,43 @@ func (m *Manager) EditRule(ctx context.Context, ruleStr string, id valuer.UUID)
...
+				}
+				amConfig, err := m.alertmanager.GetConfig(ctx, orgID.StringValue())
+				if err != nil {
+					return err
+				}
+				// move ruleId inhibitor deletion to alertmanager
+				err = amConfig.DeleteRuleIdInhibitor(id.StringValue())
+				if err != nil {
+					return err
+				}
```
srikanthccv: Please move.

pkg/query-service/rules/manager.go
srikanthccv, line 1: Please add addl manager tests for how the following changes affect default schema version and new schema version.

pkg/query-service/app/http_handler.go
```diff
@@ -492,6 +491,12 @@ func (aH *APIHandler) RegisterRoutes(router *mux.Router, am *middleware.AuthZ) {
 	router.HandleFunc("/api/v1/channels", am.EditAccess(aH.AlertmanagerAPI.CreateChannel)).Methods(http.MethodPost)
 	router.HandleFunc("/api/v1/testChannel", am.EditAccess(aH.AlertmanagerAPI.TestReceiver)).Methods(http.MethodPost)
 
+	router.HandleFunc("/api/v1/notification-policy", am.ViewAccess(aH.AlertmanagerAPI.GetAllNotificationPolicies)).Methods(http.MethodGet)
+	router.HandleFunc("/api/v1/notification-policy/{id}", am.ViewAccess(aH.AlertmanagerAPI.GetNotificationPolicyByID)).Methods(http.MethodGet)
+	router.HandleFunc("/api/v1/notification-policy", am.AdminAccess(aH.AlertmanagerAPI.CreateNotificationPolicy)).Methods(http.MethodPost)
+	router.HandleFunc("/api/v1/notification-policy/{id}", am.AdminAccess(aH.AlertmanagerAPI.DeleteNotificationPolicyByID)).Methods(http.MethodDelete)
+	router.HandleFunc("/api/v1/notification-policy/{id}", am.AdminAccess(aH.AlertmanagerAPI.UpdateNotificationPolicy)).Methods(http.MethodPut)
```
srikanthccv:
```suggestion
	router.HandleFunc("/api/v1/route_policies", am.ViewAccess(aH.AlertmanagerAPI.GetAllNotificationPolicies)).Methods(http.MethodGet)
	router.HandleFunc("/api/v1/route_policies/{id}", am.ViewAccess(aH.AlertmanagerAPI.GetNotificationPolicyByID)).Methods(http.MethodGet)
	router.HandleFunc("/api/v1/route_policies", am.AdminAccess(aH.AlertmanagerAPI.CreateNotificationPolicy)).Methods(http.MethodPost)
	router.HandleFunc("/api/v1/route_policies/{id}", am.AdminAccess(aH.AlertmanagerAPI.DeleteNotificationPolicyByID)).Methods(http.MethodDelete)
	router.HandleFunc("/api/v1/route_policies/{id}", am.AdminAccess(aH.AlertmanagerAPI.UpdateNotificationPolicy)).Methods(http.MethodPut)
```

pkg/alertmanager/alertmanager.go
```diff
@@ -59,6 +61,18 @@ type Alertmanager interface {
...
 
+	// Notification Policy CRUD
+	CreateNotificationRoute(ctx context.Context, route *alertmanagertypes.PolicyRouteRequest) error
+	CreateNotificationRoutes(ctx context.Context, routeRequests []*alertmanagertypes.PolicyRouteRequest) error
+	GetNotificationRouteByID(ctx context.Context, routeID string) (*alertmanagertypes.ExpressionRoute, error)
+	GetAllNotificationRoutes(ctx context.Context) ([]*alertmanagertypes.ExpressionRoute, error)
+	UpdateNotificationRouteByID(ctx context.Context, routeID string, route *alertmanagertypes.PolicyRouteRequest) error
+	DeleteNotificationRouteByID(ctx context.Context, routeID string) error
+	DeleteAllNotificationRoutesByName(ctx context.Context, names string) error
+	UpdateAllNotificationRoutesByName(ctx context.Context, names string, routes []*alertmanagertypes.PolicyRouteRequest) error
```
srikanthccv: No updates/deletes by Name, even if the name if ruleId.

pkg/alertmanager/alertmanager.go
```diff
@@ -59,6 +61,18 @@ type Alertmanager interface {
 
 	DeleteNotificationConfig(ctx context.Context, orgID valuer.UUID, ruleId string) error
 
+	// Notification Policy CRUD
+	CreateNotificationRoute(ctx context.Context, route *alertmanagertypes.PolicyRouteRequest) error
```
srikanthccv:
```suggestion
	CreateRoutePolicy(ctx context.Context, route *alertmanagertypes.PolicyRouteRequest) error
```

The rest of the below likewise.

pkg/alertmanager/signozalertmanager/provider.go
```diff
@@ -211,3 +215,198 @@ func (provider *provider) DeleteNotificationConfig(ctx context.Context, orgID va
...
+	}
+
+	// Create updated route with new values but preserve ID and audit fields
+	updatedRoute := &alertmanagertypes.ExpressionRoute{
+		Expression:     route.Expression,
+		ExpressionKind: alertmanagertypes.PolicyBasedExpression,
+		Priority:       route.Actions.Priority,
+		Name:           route.Name,
+		Description:    route.Description,
+		Enabled:        true,
```
srikanthccv: Why is it fixed to `true`?

pkg/alertmanager/nfmanager/notificationmanager.go
```diff
@@ -2,12 +2,27 @@
...
 	SetNotificationConfig(orgID string, ruleID string, config *alertmanagertypes.NotificationConfig) error
 	DeleteNotificationConfig(orgID string, ruleID string) error
+
+	// Route Policy CRUD
+	CreateRoute(ctx context.Context, orgID string, route *alertmanagertypes.ExpressionRoute) error
+	CreateRoutes(ctx context.Context, orgID string, routes []*alertmanagertypes.ExpressionRoute) error
+	GetRouteByID(ctx context.Context, orgID string, routeID string) (*alertmanagertypes.ExpressionRoute, error)
+	GetAllRoutes(ctx context.Context, orgID string) ([]*alertmanagertypes.ExpressionRoute, error)
+	DeleteRoute(ctx context.Context, orgID string, routeID string) error
+	DeleteAllRoutesByName(ctx context.Context, orgID string, name string) error
```
srikanthccv: Same, no name based methods.

srikanthccv:
Let's add alert manager flow tests. The test should call PutAlerts and we should have assertion on the final notification side.

pkg/alertmanager/api.go
```diff
@@ -273,3 +274,112 @@ func (api *API) CreateChannel(rw http.ResponseWriter, req *http.Request) {
...
+	ctx, cancel := context.WithTimeout(req.Context(), 30*time.Second)
+	defer cancel()
+
+	body, err := io.ReadAll(req.Body)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+	defer req.Body.Close()
+	var policy alertmanagertypes.PolicyRouteRequest
```
srikanthccv: `alertmanagertypes.PolicyRouteRequest` is used but it doesn't have json tags. 

```
type PolicyRouteRequest struct {
	Expression  string
	Actions     Actions
	Name        string
	Description string
	Tags        []string
	Kind        ExpressionKind
}
```

pkg/alertmanager/api.go
```diff
@@ -273,3 +274,112 @@ func (api *API) CreateChannel(rw http.ResponseWriter, req *http.Request) {
...
+	}
+
+	render.Success(rw, http.StatusNoContent, nil)
+}
+
+func (api *API) GetAllNotificationPolicies(rw http.ResponseWriter, req *http.Request) {
+	ctx, cancel := context.WithTimeout(req.Context(), 30*time.Second)
+	defer cancel()
+
+	policies, err := api.alertmanager.GetAllNotificationRoutes(ctx)
```
srikanthccv: So here in the response the ExpressionRote is json serialised and return in the response, and has the different structure with "actions" in req and "channels" in response. Why is that?

srikanthccv: We clerkly have a gettable rule and portable rule to take reference, why is that not followed and instead random flatten b/w request payload and response payload?

pkg/query-service/app/http_handler.go
aniketio-ctrl: https://developers.google.com/search/docs/crawling-indexing/url-structure?hl=en&visit_id=638946111977676942-1160211750&rd=1#:~:text=example.com/%F0%9F%A6%99%E2%9C%A8-,Use%20hyphens%20to%20separate%20words,-We%20recommend%20separating,

though for making consistency with other urls let me update that.

pkg/query-service/rules/manager.go
srikanthccv: Just re-upping on it. I need additional rule manager tests here. There is a basic thing missing here which causes the bugs.

Let me give one example. Imagine someone updating the preferred channels, in that case, the `notificationSettings` will be nil and route policy won't be updated with the new set of receivers.

pkg/sqlmigration/049_add_notification_routes.go
```diff
@@ -0,0 +1,256 @@
...
+	UpdatedAt time.Time `bun:"updated_at" json:"updatedAt"`
+}
+
+type userAuditable struct {
+	CreatedBy string `bun:"created_by,type:text" json:"createdBy"`
+	UpdatedBy string `bun:"updated_by,type:text" json:"updatedBy"`
+}
+
+type expressionRoute struct {
+	bun.BaseModel `bun:"table:notification_routes"`
```
srikanthccv: `route_policy`

Let's use one name how refer to them "Route policy" and we want our table names to be singular.

pkg/sqlmigration/049_add_notification_routes.go
```diff
@@ -0,0 +1,256 @@
...
+	identifiable
+	timeAuditable
+	userAuditable
+
+	Expression     string `bun:"expression,type:text,notnull" json:"expression"`
+	ExpressionKind string `bun:"kind,type:text" json:"kind"`
+
+	Channels []string `bun:"channels,type:jsonb" json:"channels"`
+	Priority string   `bun:"priority,type:text" json:"priority"`
```
srikanthccv: It's not clear why this is still there.

pkg/sqlmigration/049_add_notification_routes.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+
+		if existingRouteCount > 0 {
+			continue
+		}
+
+		routeList := migration.convertRulesToRoutes([]*rule{r}, channelsByOrg)
+		if len(routeList) > 0 {
+			_, err = db.NewInsert().
+				Model(&routeList[0]).
```
srikanthccv: incorrect because of pointer-to-pointer

pkg/query-service/rules/manager.go
aniketio-ctrl: done

pkg/alertmanager/nfmanager/notificationmanager.go
aniketio-ctrl: as discussed for now we are going with the name field post will create rule to route mapping

pkg/alertmanager/alertmanagerserver/dispatcher.go
```diff
@@ -543,21 +549,35 @@ func deepCopyRouteOpts(opts dispatch.RouteOpts, renotify time.Duration) dispatch
 	return newOpts
 }
 
-func getGroupLabels(alert *types.Alert, groups map[model.LabelName]struct{}) model.LabelSet {
+func getGroupLabels(alert *types.Alert, groups map[model.LabelName]struct{}, groupByAll bool) model.LabelSet {
 	groupLabels := model.LabelSet{}
 	for ln, lv := range alert.Labels {
-		if _, ok := groups[ln]; ok {
+		if _, ok := groups[ln]; ok || groupByAll {
```
srikanthccv, line 556: This will be problematic from the alertmanager side. Group by disabled is different from putting all labels in `groupLabels`. Probably not an immediate concern, but create a separate issue.

pkg/alertmanager/alertmanagerserver/server_e2e_test.go
```diff
@@ -0,0 +1,223 @@
...
+		require.NoError(t, err)
+		require.Len(t, alerts, 3, "Expected 3 active alerts")
+
+		for _, alert := range alerts {
+			require.Equal(t, "high-cpu-usage", alert.Alert.Labels["ruleId"])
+			require.NotEmpty(t, alert.Alert.Labels["severity"])
+			require.Contains(t, []string{"critical", "warning"}, alert.Alert.Labels["severity"])
+			require.Equal(t, "prod-cluster", alert.Alert.Labels["cluster"])
+			require.NotEmpty(t, alert.Alert.Labels["instance"])
+		}
```
srikanthccv, line 177: This is not the same thing as verifying alert reaching the notify stage. This just gets the alerts from the alerts provider. I explicitly state the goal here https://github.com/SigNoz/signoz/pull/9195#issuecomment-3339599416 and re-iterated it number of times.

pkg/alertmanager/alertmanagerserver/server_e2e_test.go
```diff
@@ -0,0 +1,223 @@
...
+		require.Contains(t, notifConfig.NotificationGroup, model.LabelName("instance"))
+	})
+
+	t.Run("verify_alert_groups_and_stages", func(t *testing.T) {
+		time.Sleep(2 * time.Second)
+
+		alertGroups, _ := server.dispatcher.Groups(
+			func(route *dispatch.Route) bool { return true },                         // Accept all routes
+			func(alert *alertmanagertypes.Alert, now time.Time) bool { return true }, // Accept all alerts
+		)
```
srikanthccv, line 209: I mentioned clearly this test should not be working with the internal component of the server.

pkg/alertmanager/alertmanagerserver/distpatcher_test.go
```diff
@@ -482,13 +544,12 @@ route:
 		totalAggrGroups += len(groups)
 	}
 
-	require.Equal(t, 7, totalAggrGroups, "Should have exactly 7 aggregation groups")
+	require.Equal(t, 4, totalAggrGroups, "Should have exactly 6 aggregation groups")
```
srikanthccv:
```suggestion
	require.Equal(t, 4, totalAggrGroups, "Should have exactly 4 aggregation groups")
```

pkg/alertmanager/alertmanagerserver/distpatcher_test.go
```diff
@@ -482,13 +544,12 @@ route:
...
 
 	// Verify specific route group counts
 	expectedGroupCounts := map[string]int{
-		"{}/{ruleId=~\"ruleId-OtherAlert|ruleId-TestingAlert\"}/0":   2, // OtherAlert + TestingAlert
-		"{}/{ruleId=~\"ruleId-HighLatency|ruleId-HighErrorRate\"}/1": 4, // 3 HighErrorRate + 1 HighLatency
-		"{}/{ruleId=\"ruleId-HighLatency\"}/2":                       1, // 1 HighLatency group
+		"{__receiver__=\"slack\"}":     2, // OtherAlert + TestingAlert
```
srikanthccv: Why does `OtherAlert + TestingAlert` mean here? The `TestingAlert` is not there in the groups at all.

pkg/alertmanager/nfmanager/nfroutingstore/sqlroutingstore/store.go
```diff
@@ -0,0 +1,93 @@
...
+		sqlstore: sqlstore,
+	}
+}
+
+func (store *store) GetByID(ctx context.Context, orgId string, id string) (*routeTypes.ExpressionRoute, error) {
+	route := new(routeTypes.ExpressionRoute)
+	err := store.sqlstore.BunDBCtx(ctx).NewSelect().Model(route).Where("id = ?", id).Where("org_id = ?", orgId).Scan(ctx)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, store.sqlstore.WrapNotFoundErrf(err, errors.CodeNotFound, "expression route with ID: %s does not exist", id)
```
srikanthccv:
```suggestion
			return nil, store.sqlstore.WrapNotFoundErrf(err, errors.CodeNotFound, "routing policy with ID: %s does not exist", id)
```

pkg/alertmanager/nfmanager/nfroutingstore/sqlroutingstore/store.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+		return nil, errors.Wrapf(err, errors.TypeInternal, errors.CodeInternal, "unable to fetch expression route with ID: %s", id)
```
srikanthccv:
```suggestion
		return nil, errors.Wrapf(err, errors.TypeInternal, errors.CodeInternal, "unable to fetch routing policy with ID: %s", id)
```

pkg/alertmanager/nfmanager/nfroutingstore/sqlroutingstore/store.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	return route, nil
+}
+
+func (store *store) Create(ctx context.Context, route *routeTypes.ExpressionRoute) error {
+	_, err := store.sqlstore.BunDBCtx(ctx).NewInsert().Model(route).Exec(ctx)
+	if err != nil {
+		return errors.NewInternalf(errors.CodeInternal, "error creating expression route with ID: %s", route.ID)
```
srikanthccv: Same, please fix the error message that will not confuse the user. Let's refer to them as "routing policy" across. There are like 5 different ways it's being referred.

pkg/alertmanager/nfmanager/nfroutingstore/sqlroutingstore/store.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	return nil
+}
+
+func (store *store) CreateBatch(ctx context.Context, route []*routeTypes.ExpressionRoute) error {
+	_, err := store.sqlstore.BunDBCtx(ctx).NewInsert().Model(&route).Exec(ctx)
+	if err != nil {
+		return errors.NewInternalf(errors.CodeInternal, "error creating expression routes: %v", err)
```
srikanthccv: Same

pkg/alertmanager/nfmanager/nfroutingstore/sqlroutingstore/store.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	return nil
+}
+
+func (store *store) Delete(ctx context.Context, orgId string, id string) error {
+	_, err := store.sqlstore.BunDBCtx(ctx).NewDelete().Model((*routeTypes.ExpressionRoute)(nil)).Where("org_id = ?", orgId).Where("id = ?", id).Exec(ctx)
+	if err != nil {
+		return errors.Wrapf(err, errors.TypeInternal, errors.CodeInternal, "unable to delete expression route with ID: %s", id)
```
srikanthccv:
```suggestion
		return errors.Wrapf(err, errors.TypeInternal, errors.CodeInternal, "unable to delete routing policy with ID: %s", id)
```

pkg/alertmanager/nfmanager/rulebasednotification/provider.go
```diff
@@ -101,3 +108,147 @@ func (r *provider) DeleteNotificationConfig(orgID string, ruleID string) error {
 
 	return nil
 }
+
+func (r *provider) CreateRoute(ctx context.Context, orgID string, route *alertmanagertypes.ExpressionRoute) error {
```
srikanthccv:
```suggestion
func (r *provider) CreateRoutePolicy(ctx context.Context, orgID string, route *alertmanagertypes.ExpressionRoute) error {
```

Since it's not "Route", to avoid confusion. Call them what they actually do i.e create rout policy.

pkg/alertmanager/nfmanager/rulebasednotification/provider.go
```diff
@@ -101,3 +108,147 @@ func (r *provider) DeleteNotificationConfig(orgID string, ruleID string) error {
...
+	if route == nil {
+		return errors.NewInvalidInputf(errors.CodeInvalidInput, "route cannot be nil")
```
srikanthccv:
```suggestion
		return errors.NewInvalidInputf(errors.CodeInvalidInput, "route policy cannot be nil")
```

pkg/alertmanager/nfmanager/rulebasednotification/provider.go
```diff
@@ -101,3 +108,147 @@ func (r *provider) DeleteNotificationConfig(orgID string, ruleID string) error {
...
+
+	err := route.Validate()
+	if err != nil {
+		return errors.NewInvalidInputf(errors.CodeInvalidInput, "Invalid route: %v", err)
+	}
+
+	return r.routeStore.Create(ctx, route)
+}
+
+func (r *provider) CreateRoutes(ctx context.Context, orgID string, routes []*alertmanagertypes.ExpressionRoute) error {
```
srikanthccv:
```suggestion
func (r *provider) CreateRoutePolicies(ctx context.Context, orgID string, routes []*alertmanagertypes.ExpressionRoute) error {
```

pkg/alertmanager/nfmanager/rulebasednotification/provider.go
```diff
@@ -101,3 +108,147 @@ func (r *provider) DeleteNotificationConfig(orgID string, ruleID string) error {
...
+
+func (r *provider) GetRouteByID(ctx context.Context, orgID string, routeID string) (*alertmanagertypes.ExpressionRoute, error) {
+	if routeID == "" {
+		return nil, errors.NewInvalidInputf(errors.CodeInvalidInput, "routeID cannot be empty")
+	}
+
+	return r.routeStore.GetByID(ctx, orgID, routeID)
+}
+
+func (r *provider) GetAllRoutes(ctx context.Context, orgID string) ([]*alertmanagertypes.ExpressionRoute, error) {
```
srikanthccv:
```suggestion
func (r *provider) GetAllRoutePolicies(ctx context.Context, orgID string) ([]*alertmanagertypes.ExpressionRoute, error) {
```

pkg/alertmanager/nfmanager/rulebasednotification/provider.go
```diff
@@ -101,3 +108,147 @@ func (r *provider) DeleteNotificationConfig(orgID string, ruleID string) error {
...
+			return errors.NewInvalidInputf(errors.CodeInvalidInput, "route cannot be nil")
+		}
+		if err := route.Validate(); err != nil {
+			return errors.NewInvalidInputf(errors.CodeInvalidInput, "route with name %s: %s", route.Name, err.Error())
+		}
+	}
+	return r.routeStore.CreateBatch(ctx, routes)
+}
+
+func (r *provider) GetRouteByID(ctx context.Context, orgID string, routeID string) (*alertmanagertypes.ExpressionRoute, error) {
```
srikanthccv:
```suggestion
func (r *provider) GetRoutePolicyByID(ctx context.Context, orgID string, routeID string) (*alertmanagertypes.ExpressionRoute, error) {
```

pkg/alertmanager/nfmanager/notificationmanager.go
srikanthccv:
```suggestion
	CreateRoutePolicy(ctx context.Context, orgID string, route *alertmanagertypes.ExpressionRoute) error
	CreateRoutePolicies(ctx context.Context, orgID string, routes []*alertmanagertypes.ExpressionRoute) error
	GetRoutePolicyByID(ctx context.Context, orgID string, routeID string) (*alertmanagertypes.ExpressionRoute, error)
	GetAllRoutePolicies(ctx context.Context, orgID string) ([]*alertmanagertypes.ExpressionRoute, error)
	DeleteRoutePolicy(ctx context.Context, orgID string, routeID string) error
	DeleteAllRoutePoliciesByName(ctx context.Context, orgID string, name string) error
```

pkg/alertmanager/api.go
```diff
@@ -273,3 +274,125 @@ func (api *API) CreateChannel(rw http.ResponseWriter, req *http.Request) {
...
+		return
+	}
+	defer req.Body.Close()
+	var policy alertmanagertypes.PostableExpressionRoute
+	err = json.Unmarshal(body, &policy)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
```
srikanthccv, line 294:
```suggestion
policy.ExpressionKind = alertmanagertypes.PolicyBasedExpression
```

pkg/alertmanager/api.go
```diff
@@ -273,3 +274,125 @@ func (api *API) CreateChannel(rw http.ResponseWriter, req *http.Request) {
...
```
srikanthccv, line 388:
```suggestion
policy.ExpressionKind = alertmanagertypes.PolicyBasedExpression
```

pkg/alertmanager/service.go
```diff
@@ -61,6 +64,7 @@ func New(
 }
 
 func (service *Service) SyncServers(ctx context.Context) error {
+	compat.InitFromFlags(service.settings.Logger(), featurecontrol.NoopFlags{})
```
srikanthccv, line 67: What is this for?

pkg/alertmanager/api.go
```diff
@@ -273,3 +274,125 @@ func (api *API) CreateChannel(rw http.ResponseWriter, req *http.Request) {
 
 	render.Success(rw, http.StatusNoContent, nil)
 }
+
+func (api *API) CreateNotificationPolicy(rw http.ResponseWriter, req *http.Request) {
```
srikanthccv:
```suggestion
func (api *API) CreateRoutePolicy(rw http.ResponseWriter, req *http.Request) {
```

pkg/alertmanager/api.go
```diff
@@ -273,3 +274,125 @@ func (api *API) CreateChannel(rw http.ResponseWriter, req *http.Request) {
...
+	result, err := api.alertmanager.CreateRoutePolicy(ctx, &policy)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	render.Success(rw, http.StatusCreated, result)
+}
+
+func (api *API) GetAllNotificationPolicies(rw http.ResponseWriter, req *http.Request) {
```
srikanthccv:
```suggestion
func (api *API) GetAllRoutePolicies(rw http.ResponseWriter, req *http.Request) {
```

pkg/alertmanager/api.go
```diff
@@ -273,3 +274,125 @@ func (api *API) CreateChannel(rw http.ResponseWriter, req *http.Request) {
...
+	policies, err := api.alertmanager.GetAllRoutePolicies(ctx)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	render.Success(rw, http.StatusOK, policies)
+}
+
+func (api *API) GetNotificationPolicyByID(rw http.ResponseWriter, req *http.Request) {
```
srikanthccv:
```suggestion
func (api *API) GetRoutePolicyByID(rw http.ResponseWriter, req *http.Request) {
```

pkg/alertmanager/api.go
```diff
@@ -273,3 +274,125 @@ func (api *API) CreateChannel(rw http.ResponseWriter, req *http.Request) {
...
+	policy, err := api.alertmanager.GetRoutePolicyByID(ctx, policyID)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	render.Success(rw, http.StatusOK, policy)
+}
+
+func (api *API) DeleteNotificationPolicyByID(rw http.ResponseWriter, req *http.Request) {
```
srikanthccv:
```suggestion
func (api *API) DeleteRoutePolicyByID(rw http.ResponseWriter, req *http.Request) {
```

pkg/alertmanager/api.go
```diff
@@ -273,3 +274,125 @@ func (api *API) CreateChannel(rw http.ResponseWriter, req *http.Request) {
...
+	err := api.alertmanager.DeleteRoutePolicyByID(ctx, policyID)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	render.Success(rw, http.StatusNoContent, nil)
+}
+
+func (api *API) UpdateNotificationPolicy(rw http.ResponseWriter, req *http.Request) {
```
srikanthccv:
```suggestion
func (api *API) UpdateRoutePolicy(rw http.ResponseWriter, req *http.Request) {
```

pkg/query-service/rules/base_rule.go
```diff
@@ -221,10 +203,6 @@ func (r *BaseRule) HoldDuration() time.Duration {
 	return r.holdDuration
 }
 
-func (r *BaseRule) TargetVal() float64 {
```
srikanthccv, line 224: 👍

pkg/sqlmigration/049_add_notification_routes.go
```diff
@@ -0,0 +1,290 @@
...
+		"CREATE INDEX IF NOT EXISTS idx_notification_routes_name ON notification_routes (name)",
+		"CREATE INDEX IF NOT EXISTS idx_notification_routes_org_route_name ON notification_routes (org_id, name)",
+	}
+
+	for _, indexSQL := range indexes {
+		_, err := db.ExecContext(ctx, indexSQL)
+		if err != nil {
+			return err
+		}
+	}
```
srikanthccv: Do it in transaction, and see how to create index using `CreateIndex` instead of the pain strings list

pkg/sqlmigration/049_add_notification_routes.go
```diff
@@ -0,0 +1,290 @@
...
+			userAuditable: userAuditable{
+				CreatedBy: r.CreatedBy,
+				UpdatedBy: r.UpdatedBy,
+			},
+			Expression:     expression,
+			ExpressionKind: "rule",
+			Channels:       gettableRule.PreferredChannels,
+			Name:           r.ID.StringValue(),
+			Description:    "",
+			Enabled:        !gettableRule.Disabled,
```
srikanthccv: The Enabled is set based on the rule disabled, however, rule manager doesn't handle the enable/disable to update it. 

Given that enabled is not actually, everything worked without an issue however, if the enabled is used then things break.

And why is that route expression is enabled/disabled based on the rule enabled/disabled?

pkg/types/alertmanagertypes/config.go
```diff
@@ -403,8 +439,10 @@ func init() {
 
 // NotificationConfig holds configuration for alert notifications timing.
 type NotificationConfig struct {
-	NotificationGroup map[model.LabelName]struct{}
-	Renotify          ReNotificationConfig
+	NotificationGroup  map[model.LabelName]struct{}
+	Renotify           ReNotificationConfig
+	NotificationPolicy bool
+	GroupByAll         bool
```
srikanthccv:
```suggestion
	NotificationGroup  map[model.LabelName]struct{}
	Renotify           ReNotificationConfig
	UsePolicy bool
	GroupByAll         bool
```

pkg/types/alertmanagertypes/expressionroute.go
srikanthccv: This is not answered yet.

pkg/types/ruletypes/api_params.go
```diff
@@ -65,21 +67,96 @@ type PostableRule struct {
 }
 
 type NotificationSettings struct {
-	NotificationGroupBy []string           `json:"notificationGroupBy,omitempty"`
-	ReNotifyInterval    Duration           `json:"renotify,omitempty"`
-	AlertStates         []model.AlertState `json:"alertStates,omitempty"`
+	NotificationGroupBy []string `json:"notificationGroupBy,omitempty"`
+	Renotify            Renotify `json:"renotify,omitempty"`
+	NotificationPolicy  bool     `json:"notificationPolicy,omitempty"`
```
srikanthccv:
```suggestion
	GroupBy []string `json:"groupBy,omitempty"`
	Renotify            Renotify `json:"renotify,omitempty"`
	UsePolicy  bool     `json:"usePolicy,omitempty"`
```

pkg/types/ruletypes/threshold.go
srikanthccv: The `RuleUnit` shouldn't be here. The `RuleUnit` comes from the composite query.

srikanthccv: Not addressed

pkg/alertmanager/service.go
aniketio-ctrl: this is for dot enabling in matchers,

pkg/types/alertmanagertypes/expressionroute.go
aniketio-ctrl: so to validate we dont have empty org id while dumping into databse

pkg/alertmanager/alertmanagerserver/dispatcher.go
aniketio-ctrl: done https://github.com/SigNoz/signoz/issues/9214

pkg/alertmanager/alertmanagerserver/server_e2e_test.go
srikanthccv: https://github.com/SigNoz/signoz/issues/9221

pkg/alertmanager/alertmanagerserver/server_e2e_test.go
srikanthccv: https://github.com/SigNoz/signoz/issues/9221

ee/query-service/rules/manager.go
```diff
@@ -126,7 +126,7 @@ func TestNotification(opts baserules.PrepareTestRuleOptions) (int, *basemodel.Ap
 	if parsedRule.RuleType == ruletypes.RuleTypeThreshold {
 
 		// add special labels for test alerts
-		parsedRule.Annotations[labels.AlertSummaryLabel] = fmt.Sprintf("The rule threshold is set to %.4f, and the observed metric value is {{$value}}.", *parsedRule.RuleCondition.Target)
+		//parsedRule.Annotations[labels.AlertSummaryLabel] = fmt.Sprintf("The rule threshold is set to %.4f, and the observed metric value is {{$value}}.", *parsedRule.RuleCondition.Target)
```
srikanthccv:
```suggestion
```

pkg/query-service/rules/test_notification.go
```diff
@@ -38,7 +38,8 @@ func defaultTestNotification(opts PrepareTestRuleOptions) (int, *model.ApiError)
 	if parsedRule.RuleType == ruletypes.RuleTypeThreshold {
 
 		// add special labels for test alerts
-		parsedRule.Annotations[labels.AlertSummaryLabel] = fmt.Sprintf("The rule threshold is set to %.4f, and the observed metric value is {{$value}}.", *parsedRule.RuleCondition.Target)
+		//we dont need this label as of now
+		//parsedRule.Annotations[labels.AlertSummaryLabel] = fmt.Sprintf("The rule threshold is set to %.4f, and the observed metric value is {{$value}}.", 0)
```
srikanthccv:
```suggestion
```

pkg/types/ruletypes/api_params.go
```diff
@@ -143,15 +219,25 @@ func (r *PostableRule) processRuleDefaults() error {
...
+			r.NotificationSettings = &NotificationSettings{
+				Renotify: Renotify{
+					Enabled:          true,
+					ReNotifyInterval: Duration(4 * time.Hour),
+					AlertStates:      []model.AlertState{model.StateFiring},
+				},
+			}
+			if r.RuleCondition.AlertOnAbsent {
+				r.NotificationSettings.Renotify.AlertStates = append(r.NotificationSettings.Renotify.AlertStates, model.StateNoData)
+			}
```
srikanthccv, line 240: If you add it here by default, you need to make sure that they are set to `nil` in the `MarshalJSON` for `DefaultSchemaVersion`.

pkg/sqlmigration/049_add_notification_routes.go
```diff
@@ -0,0 +1,341 @@
...
+		Index("idx_route_policy_name").
+		Column("name").
+		IfNotExists().
+		Exec(ctx)
+	if err != nil {
+		return err
+	}
+
+	_, err = tx.NewCreateIndex().
+		Model((*expressionRoute)(nil)).
```
vikrantgupta25: we don't use this anymore. please refer to https://github.com/SigNoz/signoz/blob/main/pkg/sqlmigration/048_update_ttl_setting.go

please use the sqlschema for the same

vikrantgupta25 (changes requested):

pkg/types/ruletypes/api_params.go
aniketio-ctrl: done

pkg/sqlmigration/049_add_notification_routes.go
```diff
@@ -0,0 +1,300 @@
...
+	"github.com/uptrace/bun"
+	"github.com/uptrace/bun/migrate"
+)
+
+// Shared types for migration
+type identifiable struct {
+	ID valuer.UUID `json:"id" bun:"id,pk,type:text"`
+}
+
+type timeAuditable struct {
```
vikrantgupta25: don't define them inline use them from types

pkg/sqlmigration/049_add_route_policy.go
```diff
@@ -0,0 +1,300 @@
...
+
+type expressionRoute struct {
+	bun.BaseModel `bun:"table:route_policy"`
+	identifiable
+	timeAuditable
+	userAuditable
+
+	Expression     string `bun:"expression,type:text,notnull" json:"expression"`
+	ExpressionKind string `bun:"kind,type:text" json:"kind"`
+
```
vikrantgupta25, line 31: you don't need json tags here and remove notnull directives as well

pkg/sqlmigration/049_add_route_policy.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+type addRoutePolicies struct {
+	sqlstore  sqlstore.SQLStore
+	sqlschema sqlschema.SQLSchema
+	logger    *slog.Logger
+}
+
+func NewAddRoutePolicyFactory(sqlstore sqlstore.SQLStore, sqlschema sqlschema.SQLSchema) factory.ProviderFactory[SQLMigration, Config] {
+	return factory.NewProviderFactory(factory.MustNewName("add_route_policy"), func(ctx context.Context, providerSettings factory.ProviderSettings, config Config) (SQLMigration, error) {
```
vikrantgupta25, line 59: match the name with the file name

pkg/sqlmigration/049_add_notification_routes.go
```diff
@@ -0,0 +1,300 @@
...
+	channelsByOrg := make(map[string][]string)
+	for _, ch := range channels {
+		channelsByOrg[ch.OrgID] = append(channelsByOrg[ch.OrgID], ch.Name)
+	}
+
+	return channelsByOrg, nil
+}
+
+func (migration *addRoutePolicies) Down(ctx context.Context, db *bun.DB) error {
+	table := &sqlschema.Table{
```
vikrantgupta25: we don't define the down migration remove this

pkg/sqlmigration/049_add_route_policy.go
```diff
@@ -0,0 +1,300 @@
...
+		Columns: []*sqlschema.Column{
+			{Name: "id", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "created_at", DataType: sqlschema.DataTypeTimestamp, Nullable: false},
+			{Name: "updated_at", DataType: sqlschema.DataTypeTimestamp, Nullable: false},
+			{Name: "created_by", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "updated_by", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "expression", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "kind", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "channels", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "name", DataType: sqlschema.DataTypeText, Nullable: false},
```
vikrantgupta25, line 109: display_name is what we follow for human readable names

pkg/sqlmigration/049_add_notification_routes.go
```diff
@@ -0,0 +1,300 @@
...
+			{Name: "id", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "created_at", DataType: sqlschema.DataTypeTimestamp, Nullable: false},
+			{Name: "updated_at", DataType: sqlschema.DataTypeTimestamp, Nullable: false},
+			{Name: "created_by", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "updated_by", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "expression", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "kind", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "channels", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "name", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "description", DataType: sqlschema.DataTypeText, Nullable: false},
```
vikrantgupta25: why is description non nullable ?

pkg/sqlmigration/049_add_route_policy.go
```diff
@@ -0,0 +1,300 @@
...
+		Where("deleted = ?", 0).
+		Scan(ctx)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil // No rules to migrate
+		}
+		return errors.NewInternalf(errors.CodeInternal, "failed to fetch rules")
+	}
+
+	if len(rules) == 0 {
```
vikrantgupta25: this shouldn't be the case given 157 ?

vikrantgupta25 (changes requested):

pkg/sqlmigration/049_add_route_policy.go
```diff
@@ -0,0 +1,300 @@
...
+			Exec(ctx)
+		if err != nil {
+			return errors.NewInternalf(errors.CodeInternal, "failed to insert notification routes")
+		}
+	}
+
+	return nil
+}
+
+func (migration *addRoutePolicies) convertRulesToRoutes(rules []*rule, channelsByOrg map[string][]string) ([]*expressionRoute, error) {
```
vikrantgupta25, line 186: why do we need a slice here for rules ? we are creating a rule slice for a single rule above

pkg/sqlmigration/049_add_notification_routes.go
```diff
@@ -0,0 +1,300 @@
...
+	if err != nil {
+		return errors.NewInternalf(errors.CodeInternal, "fetching channels error: %v", err)
+	}
+
+	var routesToInsert []*expressionRoute
+
+	for _, r := range rules {
+		existingRouteCount, err := tx.NewSelect().
+			Model((*expressionRoute)(nil)).
+			Where("name = ? AND kind = ? AND org_id = ?", r.ID.StringValue(), "rule", r.OrgID).
```
vikrantgupta25: we can have idempotency just by checking the existence of a table ?

pkg/sqlmigration/049_add_notification_routes.go
aniketio-ctrl: for rules we are storing empty descriptions, and for policy we are storing with descriptions, changing it

pkg/sqlmigration/049_add_route_policy.go
aniketio-ctrl: removed

pkg/sqlmigration/049_add_route_policy.go
aniketio-ctrl: done

pkg/sqlmigration/049_add_route_policy.go
aniketio-ctrl: done

pkg/sqlmigration/049_add_route_policy.go
aniketio-ctrl: now we are passing whole slice

pkg/sqlmigration/049_add_route_policy.go
vikrantgupta25: still there ?

pkg/sqlmigration/049_add_route_policy.go
```diff
@@ -0,0 +1,258 @@
...
+	var routes []*expressionRoute
+	var errorList []error
+	for _, r := range rules {
+		var gettableRule ruletypes.GettableRule
+		if err := json.Unmarshal([]byte(r.Data), &gettableRule); err != nil {
+			errorList = append(errorList, errors.NewInvalidInputf(errors.CodeInvalidInput, "failed to unmarshal gettableRule for rule %s : error : %v", r.ID.StringValue(), err))
+			continue
```
vikrantgupta25: this continue won't do anything as even a single error will fail the transaction

pkg/sqlmigration/049_add_route_policy.go
```diff
@@ -0,0 +1,258 @@
...
+
+func (migration *addRoutePolicies) Register(migrations *migrate.Migrations) error {
+	if err := migrations.Register(migration.Up, migration.Down); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (migration *addRoutePolicies) Up(ctx context.Context, db *bun.DB) error {
```
vikrantgupta25, line 80: the migration isn't idempotent

aniketio-ctrl: added the get check below

vikrantgupta25 (approved):

pkg/sqlmigration/049_add_route_policy.go
```diff
@@ -0,0 +1,254 @@
...
+			{Name: "expression", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "kind", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "channels", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "name", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "description", DataType: sqlschema.DataTypeText, Nullable: true},
+			{Name: "enabled", DataType: sqlschema.DataTypeBoolean, Nullable: false, Default: "true"},
+			{Name: "tags", DataType: sqlschema.DataTypeText, Nullable: true},
+			{Name: "org_id", DataType: sqlschema.DataTypeText, Nullable: false},
+		},
+		PrimaryKeyConstraint: &sqlschema.PrimaryKeyConstraint{
```
vikrantgupta25, line 115: fk constraint for org_id is missing!

vikrantgupta25 (changes requested):

pkg/sqlmigration/049_add_route_policy.go
```diff
@@ -0,0 +1,254 @@
...
+				UpdatedAt: r.UpdatedAt,
+			},
+			UserAuditable: types.UserAuditable{
+				CreatedBy: r.CreatedBy,
+				UpdatedBy: r.UpdatedBy,
+			},
+			Expression:     expression,
+			ExpressionKind: "rule",
+			Channels:       gettableRule.PreferredChannels,
+			Name:           r.ID.StringValue(),
```
vikrantgupta25, line 221: name is the ID for the rule ?

pkg/sqlmigration/049_add_route_policy.go
```diff
@@ -0,0 +1,254 @@
...
+			Description:    "",
+			Enabled:        true,
+			Tags:           []string{},
```
vikrantgupta25: since there are nullable you can remove this and for description as well

pkg/sqlmigration/049_add_route_policy.go
```diff
@@ -0,0 +1,254 @@
...
+			OrgID:          r.OrgID,
+		}
+		routes = append(routes, route)
+	}
+	return routes, nil
+}
+
+func (migration *addRoutePolicies) getAllChannelsInTx(ctx context.Context, tx bun.Tx) (map[string][]string, error) {
```
vikrantgupta25: getAllChannels should be fine as all of the operations run in tx

vikrantgupta25 (changes requested):

pkg/sqlmigration/049_add_route_policy.go
```diff
@@ -0,0 +1,254 @@
...
+
+	Expression     string `bun:"expression,type:text"`
+	ExpressionKind string `bun:"kind,type:text"`
+
+	Channels []string `bun:"channels,type:jsonb"`
+
+	Name        string   `bun:"name,type:text"`
+	Description string   `bun:"description,type:text"`
+	Enabled     bool     `bun:"enabled,type:boolean,default:true"`
+	Tags        []string `bun:"tags,type:jsonb"`
```
vikrantgupta25: since you are using tags of type text then use text here not jsonb

vikrantgupta25 (changes requested):

pkg/sqlmigration/049_add_route_policy.go
```diff
@@ -0,0 +1,254 @@
...
+
+	return nil
+}
+
+func (migration *addRoutePolicies) convertRulesToRoutes(rules []*rule, channelsByOrg map[string][]string) ([]*expressionRoute, error) {
+	var routes []*expressionRoute
+	for _, r := range rules {
+		var gettableRule ruletypes.GettableRule
+		if err := json.Unmarshal([]byte(r.Data), &gettableRule); err != nil {
+			return nil, errors.NewInternalf(errors.CodeInternal, "failed to unmarshal rule data for rule ID %s: %v", r.ID, err)
```
vikrantgupta25, line 191: would it be better to log the failures and migrate them manually on case basis rather than failing the entire migration? Thoughts @srikanthccv ?

pkg/sqlmigration/049_add_route_policy.go
```diff
@@ -0,0 +1,254 @@
...
+		if v, ok := gettableRule.Labels["severity"]; ok {
+			severity = v
+		}
+		expression := fmt.Sprintf(`%s == "%s" && %s == "%s"`, "threshold.name", severity, "ruleId", r.ID.String())
+		route := &expressionRoute{
+			Identifiable: types.Identifiable{
+				ID: valuer.GenerateUUID(),
+			},
+			TimeAuditable: types.TimeAuditable{
+				CreatedAt: r.CreatedAt,
```
vikrantgupta25: use current time as they are created and updated just now

pkg/sqlmigration/049_add_route_policy.go
srikanthccv: We discussed this, we would want to migration to fail and fix the broken alert than letting it slide. I checked on the nightswatch if there are any such changes and there were none https://github.com/SigNoz/signoz/blob/d3be2632b6597f07a21d7e028ef7b9b4120c4f2c/pkg/query-service/rules/manager.go#L278

So there won't be any failure but if it fails, let it fail so we can correct the rule entry.

pkg/sqlmigration/049_add_route_policy.go
aniketio-ctrl: yes

aniketio-ctrl: for now we are using rule id as a name, later will migrate to ruleId to routeid mapping table

srikanthccv: Yes, for the fixed channels based alert `Name` of policy is the rule ID.

pkg/sqlmigration/049_add_route_policy.go
therealpandey: Okay

pkg/signoz/signoz.go
```diff
@@ -49,6 +51,7 @@ type SigNoz struct {
 	StatsReporter   statsreporter.StatsReporter
 	Modules         Modules
 	Handlers        Handlers
+	RouteStore      alertmanagertypes.RouteStore
```
therealpandey: Remove this 
```suggestion
	RouteStore      alertmanagertypes.RouteStore
```

pkg/signoz/signoz.go
```diff
@@ -334,5 +339,6 @@ func New(
 		Sharder:         sharder,
 		Modules:         modules,
 		Handlers:        handlers,
+		RouteStore:      routeStore,
```
therealpandey:
```suggestion
		RouteStore:      routeStore,
```

remove this

vikrantgupta25:
guys if we are gonna join or query the route table based on rule id and use the name field for the same, add the same as fk relation and remove org_id relation

aniketio-ctrl:
> guys if we are gonna join or query the route table based on rule id and use the name field for the same, add the same as fk relation and remove org_id relation

we can add that table later it wont require any mutations to the table, the name column would still be there

srikanthccv:
I ran (and re-ran) the migration, and tested the functionality once again on the latest commit.

```
CREATE TABLE IF NOT EXISTS "route_policy" ("id" TEXT NOT NULL, "created_at" TIMESTAMP NOT NULL, "updated_at" TIMESTAMP NOT NULL, "created_by" TEXT NOT NULL, "updated_by" TEXT NOT NULL, "expression" TEXT NOT NULL, "kind" TEXT NOT NULL, "channels" TEXT NOT NULL, "name" TEXT NOT NULL, "description" TEXT, "enabled" BOOLEAN NOT NULL DEFAULT true, "tags" TEXT, "org_id" TEXT NOT NULL, CONSTRAINT "pk_route_policy" PRIMARY KEY ("id"), CONSTRAINT "fk_route_policy_org_id" FOREIGN KEY ("org_id") REFERENCES "organizations" ("id"));
```

grandwizard28 (approved):

vikrantgupta25 (approved):

PR #9236, opened 2025-10-02
chore: added normal index in sql schema

aniketio-ctrl:
## 📄 Summary

  The sqlschema package only supported UniqueIndex, but many migrations need regular non-unique indexes for performance optimization without enforcing uniqueness constraints. We Introduced Normal Index for this


aniketio-ctrl:
Not needed, currently working with unique index only, due to not much data in the tables

Issue #5676, opened 2024-08-12
Show unsupported mesaage when Summay metric type is used with query builder


srikanthccv:
The query builder doesn't support the summary metric type because aggregating summary metrics is illogical. At present, attempting to use the query builder with a summary metric results in an unspecified error. We should modify this to display a clear "unsupported type" error message instead.

srikanthccv:
While it is almost always incorrect to aggregate the summary, we will let them do it and instead show warning and link to docs why they shouldn't use Summary in general.

PR #9248, opened 2025-10-04
fix: summary metric type works with query builder

srikanthccv:
## 📄 Summary

Fixes https://github.com/SigNoz/signoz/issues/5676


srikanthccv:
Fixed in https://github.com/SigNoz/signoz/pull/10458

PR #9288, opened 2025-10-07
fix: anomaly with below operator negates the target

srikanthccv:
## 📄 Summary

With the recent enhancements, the comparison is moved to the `ShouldAlert` of the `Threshold` object from the rule. The thresholds are populated for the existing alerts during the Unmarshal. This works every time except for the anomaly alert +  operator is below. When the operator is below, we need to negate the target value, i.e if the expected value is 2 deviations below means the z-score threshold is -2. This negation is taking place in the `NewAnomalyRule`, which makes it of no use as the threshold has been created. 

While fixing the issue, also notice that the stdDev is not using the current season series. Fixed it.

Unit tests don't catch these subtle bugs, so I didn't add any here.

----

> [!IMPORTANT]
> Fixes anomaly detection bug with `ValueIsBelow` operator by moving target negation logic and correcting standard deviation calculation.
> 
>   - **Behavior**:
>     - Fixes anomaly detection with `ValueIsBelow` operator by negating target value in `processRuleDefaults()` in `api_params.go`.
>     - Corrects standard deviation calculation to use `currentSeasonSeries` in `getBounds()` and `getAnomalies()` in `seasonal.go`.
>   - **Tests**:
>     - Adds `TestAnomalyNegationShouldAlert` in `api_params_test.go` to verify anomaly rule behavior with `ValueIsBelow` and other operators.
>   - **Misc**:
>     - Removes redundant negation logic from `NewAnomalyRule()` in `anomaly.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 2193bc826ae0224136f44eee70a22241f57651ce. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/ee/anomaly/seasonal.go
+++ b/ee/anomaly/seasonal.go
@@ -232,7 +232,7 @@ func (p *BaseSeasonalProvider) getPredictedSeries(
 // moving avg of the previous period series + z score threshold * std dev of the series
 // moving avg of the previous period series - z score threshold * std dev of the series
 func (p *BaseSeasonalProvider) getBounds(
-	series, predictedSeries *qbtypes.TimeSeries,
+	series, predictedSeries, weekSeries *qbtypes.TimeSeries,
 	zScoreThreshold float64,
 ) (*qbtypes.TimeSeries, *qbtypes.TimeSeries) {
 	upperBoundSeries := &qbtypes.TimeSeries{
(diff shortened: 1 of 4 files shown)


therealpandey:
How about a unit test on the process rule defaults? @srikanthccv

srikanthccv:
At the moment, it didn't feel like the right thing to test on, but the existence of the test has utility to it. Added.

srikanthccv:
I will push another commit with test on shouldAlert for anomaly; the last commit test doesn't make sense (no point in adding test for the internal detail).

srikanthccv:
Added tests for `ShouldAlert` @grandwizard28, PTAL.

PR #9303, opened 2025-10-10
feat(logs-context): implement priority-based resource attribute selection

### LINKED ISSUE #9147: LOG: Logs context filter selection logic
Improvement over https://github.com/SigNoz/signoz/issues/8504#issuecomment-3061153364.

One we released this, we have seen then some users have k8s, cloud, host in resource attributes even this all selected results in a query which exceeds the limit.

We should come with a strategy.

1) Always include - `service.name` and `deployment.environment/env/environment` if present in resource filters.
2) Select only one from the list, if k8s is found then don't move to the next. And from k8s just pick the first found
```
{
  "kubernetes": {
    "priority_order": [
      "k8s.pod.uid",
      "k8s.pod.name",
      "k8s.deployment.name"
    ],
    "selection_rule": "pick_first_available"
  },
  "cloud": {
    "priority_order": [
      "cloud.resource_id",
      "cloud.provider",
      "cloud.region"
    ],
    "selection_rule": "pick_first_available"
  },
  "host": {
    "priority_order": [
      "host.id",
      "host.name"
    ],
    "selection_rule": "pick_first_available"
  },
  "container": {
    "priority_order": [
      "container.id",
      "container.name"
    ],
    "selection_rule": "pick_first_available"
  }
}
```

If the resource filter is still empty after the above. We keep this logic https://github.com/SigNoz/signoz/issues/8504#issuecomment-3061153364 and remove the k8s, cloud, host, container from the logic as it is already added above.

ahmadshaheer:
## 📄 Summary
This PR introduces a priority-based approach for selecting resource attributes in the logs context. It improves the relevance and accuracy of log context enrichment by ensuring that higher-priority resource attributes are preferred during selection.

---

## 🔍 Related Issues

Closes #9147 

---

Before:
Screenshot 2025-10-10 094311

After:
Screenshot 2025-10-10 094334

---

## 📋 Checklist

- [x] Dev Review

---

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -32,6 +32,10 @@ linters-settings:
   iface:
     enable:
       - identical
+  forbidigo:
+    forbid:
+      - fmt.Errorf
+      - ^(fmt\.Print.*|print|println)$
 issues:
   exclude-dirs:
     - "pkg/query-service"
--- a/ee/licensing/config.go
+++ b/ee/licensing/config.go
@@ -1,10 +1,10 @@
 package licensing
 
 import (
-	"fmt"
 	"sync"
 	"time"
 
+	"github.com/SigNoz/signoz/pkg/errors"
 	"github.com/SigNoz/signoz/pkg/licensing"
 )
 
@@ -18,7 +18,7 @@ func Config(pollInterval time.Duration, failureThreshold int) licensing.Config {
 	once.Do(func() {
 		config = licensing.Config{PollInterval: pollInterval, FailureThreshold: failureThreshold}
 		if err := config.Validate(); err != nil {
-			panic(fmt.Errorf("invalid licensing config: %w", err))
+			panic(errors.WrapInternalf(err, errors.CodeInternal, "invalid licensing config"))
 		}
 	})
 
--- a/ee/zeus/config.go
+++ b/ee/zeus/config.go
@@ -1,10 +1,10 @@
 package zeus
 
 import (
-	"fmt"
 	neturl "net/url"
 	"sync"
 
+	"github.com/SigNoz/signoz/pkg/errors"
 	"github.com/SigNoz/signoz/pkg/zeus"
 )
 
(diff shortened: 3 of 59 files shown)


frontend/src/container/LogsExplorerContext/useInitialQuery.ts
```diff
@@ -2,14 +2,31 @@ import { convertFiltersToExpression } from 'components/QueryBuilderV2/utils';
...
+	TagFilterItem,
+} from 'types/api/queryBuilder/queryBuilderData';
 import { DataSource } from 'types/common/queryBuilder';
 
 import { getFiltersFromResources } from './utils';
 
-const RESOURCE_STARTS_WITH_REGEX = /^(k8s|cloud|host|deployment)/; // regex to filter out resources that start with the specified keywords
-const RESOURCE_CONTAINS_REGEX = /(env|service|file|container|tenant)/; // regex to filter out resources that contains the spefied keywords
+const FALLBACK_STARTS_WITH_REGEX = /^(deployment)/;
+const FALLBACK_CONTAINS_REGEX = /(service|file|tenant)/;
```
nityanandagohain: let keep the regex, in fallback we can use anything apart form the attributes defined in k8s specs.

ahmadshaheer: done

nityanandagohain (dismissed):

frontend/src/container/LogsExplorerContext/useInitialQuery.ts
```diff
@@ -20,15 +37,70 @@ const useInitialQuery = (log: ILog): Query => {
...
+			const { key } = item.key;
+
+			return (
+				!SERVICE_AND_ENVIRONMENT_KEYS.includes(key) &&
+				(FALLBACK_STARTS_WITH_REGEX.test(key) || FALLBACK_CONTAINS_REGEX.test(key))
+			);
+		});
+	};
+
+	const updateFilters = (filters: TagFilter): TagFilter => {
```
aks07: Lets move these util functions into a util file. These operate independantly and dont req to be in the hook

frontend/src/container/LogsExplorerContext/useInitialQuery.ts
```diff
@@ -20,15 +37,70 @@ const useInitialQuery = (log: ILog): Query => {
...
+	): TagFilterItem[] =>
+		items.filter(
+			(item) => item.key?.key && isServiceOrEnvironmentAttribute(item.key.key),
+		);
+
+	// Check each category in priority order (k8s → cloud → host → container)
+	const findFirstPriorityItem = (
+		items: TagFilterItem[],
+	): TagFilterItem | undefined =>
+		PRIORITY_CATEGORIES.flatMap((categoryPriorities) =>
```
aks07: This looks a bit complicated logic. Lets see if we can simplify this a bit.

frontend/src/container/LogsExplorerContext/useInitialQuery.ts
```diff
@@ -20,15 +37,70 @@ const useInitialQuery = (log: ILog): Query => {
...
+			categoryPriorities.map((priorityKey) =>
+				items.find((item) => item.key?.key === priorityKey),
+			),
+		).find(Boolean);
+
+	const getFallbackItems = (
```
aks07: This function should not worry about if priorityItems have been found. It should be a single responsibility function. This way it is easier to extend the logic going forward if we need to. In its current state it will become hard to extend the filter logic.

aks07 (changes requested):
Added some comments. Regarding tests, we should also test duplicate attr are not present in the filters and also test for certain attributes to not be final fiterItems.
Eg. if k8s.pod.uid is there, k8s.pod.name is not needed.

PR #9308, opened 2025-10-10
fix: update validation for select and group by

nityanandagohain:
Fixes: https://github.com/SigNoz/engineering-pod/issues/3019

Previously the query went to DB and resulted in 500 errors, now we are blocking it at validation layer.

We will need frontend changes as well i.e for 400 errors we should show the options to add and remove columns.

image

### DIFF
--- a/pkg/types/querybuildertypes/querybuildertypesv5/validation.go
+++ b/pkg/types/querybuildertypes/querybuildertypesv5/validation.go
@@ -134,6 +134,11 @@ func (q *QueryBuilderQuery[T]) Validate(requestType RequestType) error {
 		return err
 	}
 
+	// Validate GroupBy
+	if err := q.validateGroupByFields(); err != nil {
+		return err
+	}
+
 	if requestType != RequestTypeRaw && requestType != RequestTypeTrace && len(q.Aggregations) > 0 {
 		if err := q.validateOrderByForAggregation(); err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


srikanthccv:
> timestamp field is not supported in selectFields, it's added by default where needed

@nityanandagohain , instead of throwing error for this, I think, when the selected field name is "timestamp" without explicitly context, it should always default to span/log timestamp.

nityanandagohain:
@srikanthccv  in this case it's with explicit context. I am thinking that in the `adjustKeys` I can override some fields regardless of what context they have.

This is the Payload UI sends.

```
{
    "schemaVersion": "v1",
    "start": 1761260368000,
    "end": 1761281968000,
    "requestType": "raw",
    "compositeQuery": {
        "queries": [
            {
                "type": "builder_query",
                "spec": {
                    "name": "A",
                    "signal": "traces",
                    "stepInterval": null,
                    "disabled": false,
                    "limit": 10,
                    "offset": 0,
                    "order": [
                        {
                            "key": {
                                "name": "timestamp"
                            },
                            "direction": "desc"
                        }
                    ],
                    "having": {
                        "expression": ""
                    },
                    "selectFields": [
                        {
                            "name": "timestamp",
                            "fieldDataType": "string",
                            "signal": "traces",
                            "fieldContext": "attribute"
                        }
                    ]
                }
            }
        ]
    },
    "formatOptions": {
        "formatTableResultForUI": false,
        "fillGaps": false
    },
    "variables": {}
}
```

srikanthccv:
> I am thinking that in the adjustKeys I can override some fields regardless of what context they have.

Let's not override to disregard the context all together, instead let's fix the frontend implementation to send the right context. We need to preserve context handling because there are valid scenarios where users intentionally use `timestamp` as an attribute for their specific requirements.

nityanandagohain:
Hmm so are you are suggesting that keep everything as it is on the backend, i.e return the timestamp if it's present in attribute as well. But on the frontend restrict it from getting added to selectField, and only allow the the span timestamp to be selected even though the metadata API returns two timestamp keys.

srikanthccv:
Please check this conversation @nityanandagohain https://signoz-team.slack.com/archives/C080TMJ8K32/p1753029200930109

srikanthccv:
This is addressed by @tushar-signoz's changes. Not relevant anymore. Hence closing.

PR #9350, opened 2025-10-16
fix: formatted threshold unit in description and summary

aniketio-ctrl:
## 📄 Summary

We were formatting the threshold value twice , due to which in the alert description we were getting values incorrect, 
---

## ✅ Changes

- [ ] Bug fix: 

## 👥 Reviewers

- backend

---

## 🧪 How to Test

1. Try testing the Rule with different unit as y axis and threshold unit,
2. Both the crossed value and threshold value should be in same unit 
3. The value of the threshold and the unit should be same as user described in description

### DIFF
--- a/frontend/src/components/QueryBuilderV2/QueryBuilderV2.styles.scss
+++ b/frontend/src/components/QueryBuilderV2/QueryBuilderV2.styles.scss
@@ -398,7 +398,7 @@
 	}
 
 	.qb-search-container {
-		.metrics-select-container {
+		.metrics-container {
 			margin-bottom: 12px;
 		}
 	}
--- a/frontend/src/components/QueryBuilderV2/QueryBuilderV2.tsx
+++ b/frontend/src/components/QueryBuilderV2/QueryBuilderV2.tsx
@@ -22,6 +22,8 @@ export const QueryBuilderV2 = memo(function QueryBuilderV2({
 	showOnlyWhereClause = false,
 	showTraceOperator = false,
 	version,
+	onSignalSourceChange,
+	signalSourceChangeEnabled = false,
 }: QueryBuilderProps): JSX.Element {
 	const {
 		currentQuery,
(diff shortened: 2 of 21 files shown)


srikanthccv:
Why does this not have unit tests?

aniketio-ctrl:
> Why does this not have unit tests?

added

pkg/types/ruletypes/threshold_test.go
```diff
@@ -0,0 +1,308 @@
...
+			},
+			series: v3.Series{
+				Labels: map[string]string{"service": "test"},
+				Points: []v3.Point{
+					{Value: 150000, Timestamp: 1000}, // 150000ms = 150s
+				},
+			},
+			ruleUnit:       "ms",
+			shouldAlert:    true,
+			expectedTarget: 100000,
```
srikanthccv: Why is the `expectedTarget` 100000? Are these AI generated tests without self-review?

aniketio-ctrl: Why is the expectedTarget 100000? 
Missed to clean up expectedTarget.
 Are these AI generated tests without self-review?
Only used I Ai for formatting.

srikanthccv:
>We were formatting the threshold value twice , due to which in the alert description we were getting values incorrect,

This is really not correct root cause analysis. Earlier (prior to the v2 enhancements) also were formatting twice https://github.com/SigNoz/signoz/blob/9a3a8c8305b8ca9e493e0b12bc19db47775bc809/pkg/query-service/rules/threshold_rule.go#L432

The `targetVal` implementation here https://github.com/SigNoz/signoz/blob/9a3a8c8305b8ca9e493e0b12bc19db47775bc809/pkg/query-service/rules/base_rule.go#L154-L168

I will check.

srikanthccv:
There is no "double" formatting. The threshold value was converted to the underlying data unit, so the `V` and `Target` are on the same base unit. However, the formatter is still called with the target unit, which resulted in this. The `Format` should have been called with the rule unit. I am going to accept this change for now.

srikanthccv (approved):

PR #9381, opened 2025-10-21
chore: JSON Logs Query Experience

piyushsingariya:
## 📄 Summary

This PR intends to bring filter and Group By capabilities on JSON logs body using ClickHouse's JSON datatype

---

## ✅ Changes

- [x] Feature: JSON Logs filteration and GroupBy

---

## 🔍 Related Issues

Closes SigNoz/engineering-pod#2990

---

## 📋 Checklist

- [ ] Dev Review

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> Introduces full JSON logs query support leveraging ClickHouse JSON, with end-to-end plumbing, metadata, and UX improvements.
> 
> - New JSON-aware condition and statement builders for logs, plus resource filter tweaks; body keys discovered via metadata and excluded from resource filters
> - `field_mapper` and schema updated to use `body_json` and `body_json_promoted`; array-paths restricted for aggregation/group-by; extensive unit tests added
> - Querier: reads JSON columns as bytes, unmarshals via `sonic`, and merges `body_json` + promoted into `body`; new `seed` helper
> - Query-service: default normalize-body pipeline suggested; log preview uses `Body.FromRaw`
> - Telemetry metadata: fetches JSON paths/indexes; supports promoted paths; robust path planning for arrays/dynamic
> - Infra: add `github.com/bytedance/sonic`; ClickHouse setting `output_format_native_write_json_as_string=1`; CI adds CH 25.10.1
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f608b940ad794644b4e01056fb5bdbfe15f2893f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -2067,6 +2067,431 @@ paths:
       summary: Get features
       tags:
       - features
+  /api/v2/metric/alerts:
+    get:
+      deprecated: false
+      description: This endpoint returns associated alerts for a specified metric
+      operationId: GetMetricAlerts
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                properties:
+                  data:
+                    $ref: '#/components/schemas/MetricsexplorertypesMetricAlertsResponse'
+                  status:
+                    type: string
+                type: object
+          description: OK
+        "400":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Bad Request
+        "401":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Unauthorized
+        "403":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Forbidden
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Internal Server Error
+      security:
+      - api_key:
+        - VIEWER
+      - tokenizer:
+        - VIEWER
+      summary: Get metric alerts
+      tags:
+      - metrics
+  /api/v2/metric/dashboards:
+    get:
+      deprecated: false
+      description: This endpoint returns associated dashboards for a specified metric
+      operationId: GetMetricDashboards
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                properties:
+                  data:
+                    $ref: '#/components/schemas/MetricsexplorertypesMetricDashboardsResponse'
+                  status:
+                    type: string
+                type: object
+          description: OK
+        "400":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Bad Request
+        "401":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Unauthorized
+        "403":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Forbidden
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Internal Server Error
+      security:
+      - api_key:
+        - VIEWER
+      - tokenizer:
+        - VIEWER
+      summary: Get metric dashboards
+      tags:
+      - metrics
+  /api/v2/metric/highlights:
+    get:
+      deprecated: false
+      description: This endpoint returns highlights like number of datapoints, totaltimeseries,
+        active time series, last received time for a specified metric
+      operationId: GetMetricHighlights
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                properties:
+                  data:
+                    $ref: '#/components/schemas/MetricsexplorertypesMetricHighlightsResponse'
+                  status:
+                    type: string
+                type: object
+          description: OK
+        "400":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Bad Request
+        "401":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Unauthorized
+        "403":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Forbidden
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Internal Server Error
+      security:
+      - api_key:
+        - VIEWER
+      - tokenizer:
+        - VIEWER
+      summary: Get metric highlights
+      tags:
+      - metrics
+  /api/v2/metrics/{metric_name}/metadata:
+    post:
+      deprecated: false
+      description: This endpoint helps to update metadata information like metric
+        description, unit, type, temporality, monotonicity for a specified metric
+      operationId: UpdateMetricMetadata
+      parameters:
+      - in: path
+        name: metric_name
+        required: true
+        schema:
+          type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/MetricsexplorertypesUpdateMetricMetadataRequest'
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                type: string
+          description: OK
+        "400":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Bad Request
+        "401":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Unauthorized
+        "403":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Forbidden
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Internal Server Error
+      security:
+      - api_key:
+        - EDITOR
+      - tokenizer:
+        - EDITOR
+      summary: Update metric metadata
+      tags:
+      - metrics
+  /api/v2/metrics/attributes:
+    post:
+      deprecated: false
+      description: This endpoint returns attribute keys and their unique values for
+        a specified metric
+      operationId: GetMetricAttributes
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/MetricsexplorertypesMetricAttributesRequest'
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                properties:
+                  data:
+                    $ref: '#/components/schemas/MetricsexplorertypesMetricAttributesResponse'
+                  status:
+                    type: string
+                type: object
+          description: OK
+        "400":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Bad Request
+        "401":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Unauthorized
+        "403":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Forbidden
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Internal Server Error
+      security:
+      - api_key:
+        - VIEWER
+      - tokenizer:
+        - VIEWER
+      summary: Get metric attributes
+      tags:
+      - metrics
+  /api/v2/metrics/metadata:
+    get:
+      deprecated: false
+      description: This endpoint returns metadata information like metric description,
+        unit, type, temporality, monotonicity for a specified metric
+      operationId: GetMetricMetadata
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                properties:
+                  data:
+                    $ref: '#/components/schemas/MetricsexplorertypesMetricMetadata'
+                  status:
+                    type: string
+                type: object
+          description: OK
+        "400":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Bad Request
+        "401":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Unauthorized
+        "403":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Forbidden
+        "404":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Not Found
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Internal Server Error
+      security:
+      - api_key:
+        - VIEWER
+      - tokenizer:
+        - VIEWER
+      summary: Get metric metadata
+      tags:
+      - metrics
+  /api/v2/metrics/stats:
+    post:
+      deprecated: false
+      description: This endpoint provides list of metrics with their number of samples
+        and timeseries for the given time range
+      operationId: GetMetricsStats
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/MetricsexplorertypesStatsRequest'
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                properties:
+                  data:
+                    $ref: '#/components/schemas/MetricsexplorertypesStatsResponse'
+                  status:
+                    type: string
+                type: object
+          description: OK
+        "400":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Bad Request
+        "401":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Unauthorized
+        "403":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Forbidden
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Internal Server Error
+      security:
+      - api_key:
+        - VIEWER
+      - tokenizer:
+        - VIEWER
+      summary: Get metrics statistics
+      tags:
+      - metrics
+  /api/v2/metrics/treemap:
+    post:
+      deprecated: false
+      description: This endpoint returns a treemap visualization showing the proportional
+        distribution of metrics by sample count or time series count
+      operationId: GetMetricsTreemap
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/MetricsexplorertypesTreemapRequest'
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                properties:
+                  data:
+                    $ref: '#/components/schemas/MetricsexplorertypesTreemapResponse'
+                  status:
+                    type: string
+                type: object
+          description: OK
+        "400":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Bad Request
+        "401":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Unauthorized
+        "403":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Forbidden
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Internal Server Error
+      security:
+      - api_key:
+        - VIEWER
+      - tokenizer:
+        - VIEWER
+      summary: Get metrics treemap
+      tags:
+      - metrics
   /api/v2/orgs/me:
     get:
       deprecated: false
@@ -2586,6 +3011,202 @@ components:
           nullable: true
           type: object
       type: object
+    MetricsexplorertypesMetricAlert:
+      properties:
+        alertId:
+          type: string
+        alertName:
+          type: string
+      type: object
+    MetricsexplorertypesMetricAlertsResponse:
+      properties:
+        alerts:
+          items:
+            $ref: '#/components/schemas/MetricsexplorertypesMetricAlert'
+          nullable: true
+          type: array
+      type: object
+    MetricsexplorertypesMetricAttribute:
+      properties:
+        key:
+          type: string
+        valueCount:
+          minimum: 0
+          type: integer
+        values:
+          items:
+            type: string
+          nullable: true
+          type: array
+      type: object
+    MetricsexplorertypesMetricAttributesRequest:
+      properties:
+        end:
+          nullable: true
+          type: integer
+        metricName:
+          type: string
+        start:
+          nullable: true
+          type: integer
+      type: object
+    MetricsexplorertypesMetricAttributesResponse:
+      properties:
+        attributes:
+          items:
+            $ref: '#/components/schemas/MetricsexplorertypesMetricAttribute'
+          nullable: true
+          type: array
+        totalKeys:
+          format: int64
+          type: integer
+      type: object
+    MetricsexplorertypesMetricDashboard:
+      properties:
+        dashboardId:
+          type: string
+        dashboardName:
+          type: string
+        widgetId:
+          type: string
+        widgetName:
+          type: string
+      type: object
+    MetricsexplorertypesMetricDashboardsResponse:
+      properties:
+        dashboards:
+          items:
+            $ref: '#/components/schemas/MetricsexplorertypesMetricDashboard'
+          nullable: true
+          type: array
+      type: object
+    MetricsexplorertypesMetricHighlightsResponse:
+      properties:
+        activeTimeSeries:
+          minimum: 0
+          type: integer
+        dataPoints:
+          minimum: 0
+          type: integer
+        lastReceived:
+          minimum: 0
+          type: integer
+        totalTimeSeries:
+          minimum: 0
+          type: integer
+      type: object
+    MetricsexplorertypesMetricMetadata:
+      properties:
+        description:
+          type: string
+        isMonotonic:
+          type: boolean
+        temporality:
+          type: string
+        type:
+          type: string
+        unit:
+          type: string
+      type: object
+    MetricsexplorertypesStat:
+      properties:
+        description:
+          type: string
+        metricName:
+          type: string
+        samples:
+          minimum: 0
+          type: integer
+        timeseries:
+          minimum: 0
+          type: integer
+        type:
+          type: string
+        unit:
+          type: string
+      type: object
+    MetricsexplorertypesStatsRequest:
+      properties:
+        end:
+          format: int64
+          type: integer
+        filter:
+          $ref: '#/components/schemas/Querybuildertypesv5Filter'
+        limit:
+          type: integer
+        offset:
+          type: integer
+        orderBy:
+          $ref: '#/components/schemas/Querybuildertypesv5OrderBy'
+        start:
+          format: int64
+          type: integer
+      type: object
+    MetricsexplorertypesStatsResponse:
+      properties:
+        metrics:
+          items:
+            $ref: '#/components/schemas/MetricsexplorertypesStat'
+          nullable: true
+          type: array
+        total:
+          minimum: 0
+          type: integer
+      type: object
+    MetricsexplorertypesTreemapEntry:
+      properties:
+        metricName:
+          type: string
+        percentage:
+          format: double
+          type: number
+        totalValue:
+          minimum: 0
+          type: integer
+      type: object
+    MetricsexplorertypesTreemapRequest:
+      properties:
+        end:
+          format: int64
+          type: integer
+        filter:
+          $ref: '#/components/schemas/Querybuildertypesv5Filter'
+        limit:
+          type: integer
+        mode:
+          type: string
+        start:
+          format: int64
+          type: integer
+      type: object
+    MetricsexplorertypesTreemapResponse:
+      properties:
+        samples:
+          items:
+            $ref: '#/components/schemas/MetricsexplorertypesTreemapEntry'
+          nullable: true
+          type: array
+        timeseries:
+          items:
+            $ref: '#/components/schemas/MetricsexplorertypesTreemapEntry'
+          nullable: true
+          type: array
+      type: object
+    MetricsexplorertypesUpdateMetricMetadataRequest:
+      properties:
+        description:
+          type: string
+        isMonotonic:
+          type: boolean
+        metricName:
+          type: string
+        temporality:
+          type: string
+        type:
+          type: string
+        unit:
+          type: string
+      type: object
     PreferencetypesPreference:
       properties:
         allowedScopes:
@@ -2647,6 +3268,35 @@ components:
           minimum: 0
           type: integer
       type: object
+    Querybuildertypesv5Filter:
+      properties:
+        expression:
+          type: string
+      type: object
+    Querybuildertypesv5OrderBy:
+      properties:
+        direction:
+          type: string
+        key:
+          $ref: '#/components/schemas/Querybuildertypesv5OrderByKey'
+      type: object
+    Querybuildertypesv5OrderByKey:
+      properties:
+        '-':
+          $ref: '#/components/schemas/TelemetrytypesJSONDataType'
+        description:
+          type: string
+        fieldContext:
+          type: string
+        fieldDataType:
+          type: string
+        name:
+          type: string
+        signal:
+          type: string
+        unit:
+          type: string
+      type: object
     Querybuildertypesv5QueryData:
       properties:
         results:
@@ -2687,6 +3337,8 @@ components:
           $ref: '#/components/schemas/ErrorsJSON'
         status:
           type: string
+      type: object
+    TelemetrytypesJSONDataType:
       type: object
     TypesChangePasswordRequest:
       properties:
--- a/pkg/apiserver/signozapiserver/metricsexplorer.go
+++ b/pkg/apiserver/signozapiserver/metricsexplorer.go
@@ -1,0 +1,166 @@
+package signozapiserver
+
+import (
+	"net/http"
+
+	"github.com/SigNoz/signoz/pkg/http/handler"
+	"github.com/SigNoz/signoz/pkg/types"
+	"github.com/SigNoz/signoz/pkg/types/metricsexplorertypes"
+	"github.com/gorilla/mux"
+)
+
+func (provider *provider) addMetricsExplorerV2Routes(router *mux.Router) error {
+	if err := router.Handle("/api/v2/metrics/stats", handler.New(
+		provider.authZ.ViewAccess(provider.metricsExplorerHandler.GetStats),
+		handler.OpenAPIDef{
+			ID:                  "GetMetricsStats",
+			Tags:                []string{"metrics"},
+			Summary:             "Get metrics statistics",
+			Description:         "This endpoint provides list of metrics with their number of samples and timeseries for the given time range",
+			Request:             new(metricsexplorertypes.StatsRequest),
+			RequestContentType:  "application/json",
+			Response:            new(metricsexplorertypes.StatsResponse),
+			ResponseContentType: "application/json",
+			SuccessStatusCode:   http.StatusOK,
+			ErrorStatusCodes:    []int{http.StatusBadRequest, http.StatusUnauthorized, http.StatusInternalServerError},
+			Deprecated:          false,
+			SecuritySchemes:     newSecuritySchemes(types.RoleViewer),
+		})).Methods(http.MethodPost).GetError(); err != nil {
+		return err
+	}
+
+	if err := router.Handle("/api/v2/metrics/treemap", handler.New(
+		provider.authZ.ViewAccess(provider.metricsExplorerHandler.GetTreemap),
+		handler.OpenAPIDef{
+			ID:                  "GetMetricsTreemap",
+			Tags:                []string{"metrics"},
+			Summary:             "Get metrics treemap",
+			Description:         "This endpoint returns a treemap visualization showing the proportional distribution of metrics by sample count or time series count",
+			Request:             new(metricsexplorertypes.TreemapRequest),
+			RequestContentType:  "application/json",
+			Response:            new(metricsexplorertypes.TreemapResponse),
+			ResponseContentType: "application/json",
+			SuccessStatusCode:   http.StatusOK,
+			ErrorStatusCodes:    []int{http.StatusBadRequest, http.StatusUnauthorized, http.StatusInternalServerError},
+			Deprecated:          false,
+			SecuritySchemes:     newSecuritySchemes(types.RoleViewer),
+		})).Methods(http.MethodPost).GetError(); err != nil {
+		return err
+	}
+
+	if err := router.Handle("/api/v2/metrics/attributes", handler.New(
+		provider.authZ.ViewAccess(provider.metricsExplorerHandler.GetMetricAttributes),
+		handler.OpenAPIDef{
+			ID:                  "GetMetricAttributes",
+			Tags:                []string{"metrics"},
+			Summary:             "Get metric attributes",
+			Description:         "This endpoint returns attribute keys and their unique values for a specified metric",
+			Request:             new(metricsexplorertypes.MetricAttributesRequest),
+			RequestContentType:  "application/json",
+			Response:            new(metricsexplorertypes.MetricAttributesResponse),
+			ResponseContentType: "application/json",
+			SuccessStatusCode:   http.StatusOK,
+			ErrorStatusCodes:    []int{http.StatusBadRequest, http.StatusUnauthorized, http.StatusInternalServerError},
+			Deprecated:          false,
+			SecuritySchemes:     newSecuritySchemes(types.RoleViewer),
+		})).Methods(http.MethodPost).GetError(); err != nil {
+		return err
+	}
+
+	if err := router.Handle("/api/v2/metrics/metadata", handler.New(
+		provider.authZ.ViewAccess(provider.metricsExplorerHandler.GetMetricMetadata),
+		handler.OpenAPIDef{
+			ID:                  "GetMetricMetadata",
+			Tags:                []string{"metrics"},
+			Summary:             "Get metric metadata",
+			Description:         "This endpoint returns metadata information like metric description, unit, type, temporality, monotonicity for a specified metric",
+			Request:             nil,
+			RequestContentType:  "",
+			Response:            new(metricsexplorertypes.MetricMetadata),
+			ResponseContentType: "application/json",
+			SuccessStatusCode:   http.StatusOK,
+			ErrorStatusCodes:    []int{http.StatusBadRequest, http.StatusUnauthorized, http.StatusNotFound, http.StatusInternalServerError},
+			Deprecated:          false,
+			SecuritySchemes:     newSecuritySchemes(types.RoleViewer),
+		})).Methods(http.MethodGet).GetError(); err != nil {
+		return err
+	}
+
+	if err := router.Handle("/api/v2/metrics/{metric_name}/metadata", handler.New(
+		provider.authZ.EditAccess(provider.metricsExplorerHandler.UpdateMetricMetadata),
+		handler.OpenAPIDef{
+			ID:                  "UpdateMetricMetadata",
+			Tags:                []string{"metrics"},
+			Summary:             "Update metric metadata",
+			Description:         "This endpoint helps to update metadata information like metric description, unit, type, temporality, monotonicity for a specified metric",
+			Request:             new(metricsexplorertypes.UpdateMetricMetadataRequest),
+			RequestContentType:  "application/json",
+			Response:            nil,
+			ResponseContentType: "application/json",
+			SuccessStatusCode:   http.StatusOK,
+			ErrorStatusCodes:    []int{http.StatusBadRequest, http.StatusUnauthorized, http.StatusInternalServerError},
+			Deprecated:          false,
+			SecuritySchemes:     newSecuritySchemes(types.RoleEditor),
+		})).Methods(http.MethodPost).GetError(); err != nil {
+		return err
+	}
+
+	if err := router.Handle("/api/v2/metric/highlights", handler.New(
+		provider.authZ.ViewAccess(provider.metricsExplorerHandler.GetMetricHighlights),
+		handler.OpenAPIDef{
+			ID:                  "GetMetricHighlights",
+			Tags:                []string{"metrics"},
+			Summary:             "Get metric highlights",
+			Description:         "This endpoint returns highlights like number of datapoints, totaltimeseries, active time series, last received time for a specified metric",
+			Request:             nil,
+			RequestContentType:  "",
+			Response:            new(metricsexplorertypes.MetricHighlightsResponse),
+			ResponseContentType: "application/json",
+			SuccessStatusCode:   http.StatusOK,
+			ErrorStatusCodes:    []int{http.StatusBadRequest, http.StatusUnauthorized, http.StatusInternalServerError},
+			Deprecated:          false,
+			SecuritySchemes:     newSecuritySchemes(types.RoleViewer),
+		})).Methods(http.MethodGet).GetError(); err != nil {
+		return err
+	}
+
+	if err := router.Handle("/api/v2/metric/alerts", handler.New(
+		provider.authZ.ViewAccess(provider.metricsExplorerHandler.GetMetricAlerts),
+		handler.OpenAPIDef{
+			ID:                  "GetMetricAlerts",
+			Tags:                []string{"metrics"},
+			Summary:             "Get metric alerts",
+			Description:         "This endpoint returns associated alerts for a specified metric",
+			Request:             nil,
+			RequestContentType:  "",
+			Response:            new(metricsexplorertypes.MetricAlertsResponse),
+			ResponseContentType: "application/json",
+			SuccessStatusCode:   http.StatusOK,
+			ErrorStatusCodes:    []int{http.StatusBadRequest, http.StatusUnauthorized, http.StatusInternalServerError},
+			Deprecated:          false,
+			SecuritySchemes:     newSecuritySchemes(types.RoleViewer),
+		})).Methods(http.MethodGet).GetError(); err != nil {
+		return err
+	}
+
+	if err := router.Handle("/api/v2/metric/dashboards", handler.New(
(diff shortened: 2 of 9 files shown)


pkg/telemetrylogs/stmt_builder_test.go
```diff
@@ -232,7 +234,7 @@ func TestStatementBuilderListQuery(t *testing.T) {
...
+				Query: "WITH __resource_filter AS (SELECT fingerprint FROM signoz_logs.distributed_logs_v2_resource WHERE (simpleJSONExtractString(labels, 'service.name') = ? AND labels LIKE ? AND labels LIKE ?) AND seen_at_ts_bucket_start >= ? AND seen_at_ts_bucket_start <= ?) SELECT timestamp, id, trace_id, span_id, trace_flags, severity_text, severity_number, scope_name, scope_version, body_v2, promoted, attributes_string, attributes_number, attributes_bool, resources_string, scope_string FROM signoz_logs.distributed_logs_v2 WHERE resource_fingerprint GLOBAL IN (SELECT fingerprint FROM __resource_filter) AND true AND timestamp >= ? AND ts_bucket_start >= ? AND timestamp < ? AND ts_bucket_start <= ? LIMIT ?",
```
nityanandagohain: why are we removing body and just selecting body_v2, promoted ? shouldn't it be all three

pkg/telemetrylogs/stmt_builder_test.go
```diff
@@ -369,15 +373,16 @@ func TestStatementBuilderListQueryResourceTests(t *testing.T) {
...
+				Query: "WITH __resource_filter AS (SELECT fingerprint FROM signoz_logs.distributed_logs_v2_resource WHERE true AND seen_at_ts_bucket_start >= ? AND seen_at_ts_bucket_start <= ?) SELECT timestamp, id, trace_id, span_id, trace_flags, severity_text, severity_number, scope_name, scope_version, body_v2, promoted, attributes_string, attributes_number, attributes_bool, resources_string, scope_string FROM signoz_logs.distributed_logs_v2 WHERE resource_fingerprint GLOBAL IN (SELECT fingerprint FROM __resource_filter) AND (dynamicElement(body_v2.status, 'String') = ?) AND timestamp >= ? AND ts_bucket_start >= ? AND timestamp < ? AND ts_bucket_start <= ? LIMIT ?",
```
nityanandagohain: we will have to make it backward compatable, i.e there should be two version one will use the old style and the new one will use the new dynamicElement that can be switched with a config.

pkg/telemetrylogs/json_stmt_builder_test.go
```diff
@@ -0,0 +1,1121 @@
...
...t(`body_v2.interests[].entities[].reviews[].entries`.metadata, 'Array(JSON(max_dynamic_types=1, max_dynamic_paths=0))')), dynamicElement(`body_v2.interests[].entities[].reviews`.entries, 'Array(JSON(max_dynamic_types=2, max_dynamic_paths=0))')), dynamicElement(`body_v2.interests[].entities`.reviews, 'Array(JSON(max_dynamic_types=4, max_dynamic_paths=0))')), dynamicElement(`body_v2.interests`.entities, 'Array(JSON(max_dynamic_types=8, max_dynamic_paths=0))')), dynamicElement(body_v2.interests, 'Array(JSON(max_dynamic_types=16, max_dynamic_paths=0))')))) AS `body.interests[].entities[].reviews[].entries[].metadata[].positions[].unit` WHERE resource_fingerprint GLOBAL IN (SELECT fingerprint FROM __resource_filter) AND timestamp >= ? AND ts_bucket_start >= ? AND timestamp < ? AND ts_bucket_start <= ? AND (`body.education[].awards[].name`, `body.interests[].entities[].reviews[].entries[].metadata[].positions[].unit`) GLOBAL IN (SELECT `body.education[].awards[].name`, `body.interests[].entities[].reviews[].entries[].metadata[].positions[].unit` FROM __limit_cte) GROUP BY ts, `body.education[].awards[].name`, `body.interests[].entities[].reviews[].entries[].metadata[].positions[].unit`",
```
nityanandagohain: this is way too complex to understand and review.

pkg/telemetrylogs/json_stmt_builder_test.go
```diff
@@ -0,0 +1,1121 @@
...
...ge, 'Int64'))) AS `body.user.age`, toString(coalesce(dynamicElement(promoted.user.name, 'String'), dynamicElement(body_v2.user.name, 'String'))) AS `body.user.name`, count() AS __result_0 FROM signoz_logs.distributed_logs_v2 WHERE resource_fingerprint GLOBAL IN (SELECT fingerprint FROM __resource_filter) AND timestamp >= ? AND ts_bucket_start >= ? AND timestamp < ? AND ts_bucket_start <= ? GROUP BY `body.user.age`, `body.user.name` ORDER BY __result_0 DESC LIMIT ?) SELECT toStartOfInterval(fromUnixTimestamp64Nano(timestamp), INTERVAL 30 SECOND) AS ts, toString(coalesce(dynamicElement(promoted.user.age, 'Int64'), dynamicElement(body_v2.user.age, 'Int64'))) AS `body.user.age`, toString(coalesce(dynamicElement(promoted.user.name, 'String'), dynamicElement(body_v2.user.name, 'String'))) AS `body.user.name`, count() AS __result_0 FROM signoz_logs.distributed_logs_v2 WHERE resource_fingerprint GLOBAL IN (SELECT fingerprint FROM __resource_filter) AND timestamp >= ? AND ts_bucket_start >= ? AND timestamp < ? AND ts_bucket_start <= ? AND (`body.user.age`, `body.user.name`) GLOBAL IN (SELECT `body.user.age`, `body.user.name` FROM __limit_cte) GROUP BY ts, `body.user.age`, `body.user.name`",
```
nityanandagohain: coalesce will not help in performance, it will read both the columns thus resulting in slow pef.

pkg/telemetrylogs/stmt_builder_test.go
piyushsingariya: Older change

pkg/telemetrylogs/stmt_builder_test.go
piyushsingariya: It is. TLDR: older change

pkg/telemetrylogs/json_stmt_builder_test.go
piyushsingariya: I can put these into individual files, in formatted manner to be reviewed? but that will add up more work in developer productivity

nityanandagohain (commented):
Let's add some simpler test to start with eg:- in group by

* group by body.user.name
* group by body.user.name with limit 10
* Multiple group by

check what will group by these mean, and if it even makes sense in our case.
* group by body.data[].status 
* group by body.data[].status with limit 10

Exists on a key should be a simple is not NULL check eg:-
``` select payload.user.name, payload.user.xyz is NOT NULL  from example_json;```

File: pkg/querier/builder_query.go [hunk 5]
```diff
@@ -248,6 +249,40 @@ func (q *builderQuery[T]) executeWithContext(ctx context.Context, query string,
 		return nil, err
 	}
 
+	// merge body_v2 and promoted into body
+	if q.spec.Signal == telemetrytypes.SignalLogs {
```
nityanandagohain, line 254: Is postProcessResults the right place for this logic ?

pkg/querier/builder_query.go
```diff
@@ -248,6 +249,40 @@ func (q *builderQuery[T]) executeWithContext(ctx context.Context, query string,
...
+		switch typedPayload := payload.(type) {
+		case *qbtypes.RawData:
+			for _, rr := range typedPayload.Rows {
+				seeder := func() error {
+					body, ok := rr.Data["body_v2"].(map[string]any)
```
nityanandagohain: thinking about more self descriptive column names, `body_json`  and `body_json_promoted`. Might be more self descriptive column names for our use case.

Also we should use constants to define these names at one place

pkg/querier/builder_query.go
```diff
@@ -375,3 +410,18 @@ func decodeCursor(cur string) (int64, error) {
 	}
 	return strconv.ParseInt(string(b), 10, 64)
 }
+
+func seed(promoted map[string]any, body map[string]any) {
```
nityanandagohain, line 415: depth limit should be added here else it will traverse everything.

pkg/querier/consume.go
```diff
@@ -354,10 +354,22 @@ func readAsRaw(rows driver.Rows, queryName string) (*qbtypes.RawData, error) {
 	colTypes := rows.ColumnTypes()
 	colCnt := len(colNames)
 
+	// Helper that decides scan target per column based on DB type
+	makeScanTarget := func(i int) any {
+		dbt := strings.ToUpper(colTypes[i].DatabaseTypeName())
+		if strings.HasPrefix(dbt, "JSON") {
+			// Since the driver fails to decode JSON/Dynamic into native Go values, we read it as raw bytes
```
nityanandagohain, line 361: did you create a issue on their github ?

pkg/querier/consume.go
```diff
@@ -383,6 +395,28 @@ func readAsRaw(rows driver.Rows, queryName string) (*qbtypes.RawData, error) {
...
+			if strings.HasPrefix(strings.ToUpper(colTypes[i].DatabaseTypeName()), "JSON") {
+				switch x := val.(type) {
+				case []byte:
+					if len(x) > 0 {
+						var v any
+						if err := sonic.Unmarshal(x, &v); err == nil {
+							val = v
+						}
+					}
+				case string:
```
nityanandagohain: when will we reach to this second case ?

pkg/querier/signozquerier/provider.go
```diff
@@ -101,8 +101,14 @@ func newProvider(
 	)
 
 	// Create log statement builder
+	jqb, err := telemetrylogs.NewJSONQueryBuilder(ctx, telemetryStore, settings.Logger)
```
nityanandagohain: nit: needs better function naming

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -223,7 +227,142 @@ func NewReaderFromClickhouseConnection(
...
+		}
+		tokenIndex := schemamigrator.Index{
+			Name:        schemamigrator.JSONSubColumnIndexName(column, schemamigrator.IndexTypeTokenBF),
+			Expression:  schemamigrator.JSONSubColumnIndexExpr(column),
+			Type:        "tokenbf_v1(1024, 2, 0)",
+			Granularity: 1,
+		}
+		indexes := []schemamigrator.Index{ngramIndex, tokenIndex}
+		for _, index := range indexes {
+			query := fmt.Sprintf("ALTER TABLE %s.%s ON CLUSTER %s ADD %s",
```
nityanandagohain: should be idempotenet i.e `IF NOT EXISTS`

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -223,7 +227,142 @@ func NewReaderFromClickhouseConnection(
...
+	if len(columns) == 0 {
+		return nil
+	}
+	// We'll add indexes on the local logs table across cluster
+	// Index expression: lower(assumeNotNull(dynamicElement(promoted.<path>, 'String')))
+	for _, column := range columns {
+		ngramIndex := schemamigrator.Index{
+			Name:        schemamigrator.JSONSubColumnIndexName(column, schemamigrator.IndexTypeNGramBF),
+			Expression:  schemamigrator.JSONSubColumnIndexExpr(column),
+			Type:        "ngrambf_v1(4, 5000, 2, 0)",
```
nityanandagohain: keep this paramterized from the API, even the number of indexes. 

Also create and endpoint for GET/checking these.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -1327,9 +1466,21 @@ func (r *ClickHouseReader) setTTLLogs(ctx context.Context, orgID string, params
 			params.ToColdStorageDuration, params.ColdStorageVolume)
 	}
 
+	// TTL query for path_types table to follow the same TTL as logs
+	pathTypesLocal := fmt.Sprintf("%s.%s", r.logsDB, r.pathTypesLocalTable)
```
nityanandagohain: will require changes in TTL V2 and might require change in integration tests.

pkg/query-service/app/logparsingpipeline/controller.go
```diff
@@ -273,6 +308,13 @@ func (pc *LogParsingPipelineController) RecommendAgentConfig(
 		return nil, "", apiErr
 	}
 
+	// recommend default pipelines along with user created pipelines
+	defaultPipelines, err := pc.getDefaultPipelines()
```
nityanandagohain, line 299: so we are adding this pipeline by default, i.e we are adding this to all our old customers by default right ? 

Ideally we wanted to add this to all user over time.

pkg/query-service/model/logs.go
```diff
@@ -62,3 +64,38 @@ func GetLogFieldsV3(ctx context.Context, queryRangeParams *v3.QueryRangeParamsV3
...
 }
+
+type PromotePathsRequest struct {
+	Paths []PromotePathItem `json:"paths"`
+}
+
+type PromotePathItem struct {
+	Path    string `json:"path"`
+	Promote bool   `json:"promote"`
+	Indexed bool   `json:"indexed"`
```
nityanandagohain: set default index if not sent from the user, else let user configure what Index to be sent and how many of them.

pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -223,7 +227,142 @@ func NewReaderFromClickhouseConnection(
...
+
+// CreateJSONPathIndexes creates string ngram + token filter indexes on JSON path subcolumns for LIKE queries.
+func (r *ClickHouseReader) CreateJSONPathIndexes(ctx context.Context, columns []string) error {
+	if len(columns) == 0 {
+		return nil
+	}
+	// We'll add indexes on the local logs table across cluster
+	// Index expression: lower(assumeNotNull(dynamicElement(promoted.<path>, 'String')))
+	for _, column := range columns {
+		ngramIndex := schemamigrator.Index{
```
nityanandagohain: what will happen if the promoted path is a number

pkg/telemetrylogs/body_json_keys.go
```diff
@@ -0,0 +1,136 @@
...
+	} else {
+		// For full sync, we're always complete (no limit applied)
+		complete = true
+	}
+
+	return paths, complete, highestLastSeen, nil
+}
+
+// escapeForLike escapes special characters for LIKE queries
+func escapeForLike(s string) string {
```
nityanandagohain: check if this is really needed, we dont have something similar in the QB v5  as we just use `FormatValueForContains` function and nothing apart from that.

pkg/telemetrylogs/condition_builder.go
```diff
@@ -165,13 +179,14 @@ func (c *conditionBuilder) conditionFor(
 		}
 
 		var value any
-		switch column.Type {
-		case schema.JSONColumnType{}:
+		if column.IsJSONColumn() {
```
nityanandagohain: add why this is outside of the switch statement.

pkg/telemetrylogs/condition_builder.go
```diff
@@ -218,15 +233,15 @@ func (c *conditionBuilder) ConditionFor(
...
+	_ uint64,
+	_ uint64,
 ) (string, error) {
 	condition, err := c.conditionFor(ctx, key, operator, value, sb)
 	if err != nil {
 		return "", err
 	}
 
-	if operator.AddDefaultExistsFilter() {
+	if !(strings.HasPrefix(key.Name, BodyJSONStringSearchPrefix) && constants.BodyV2QueryEnabled) && operator.AddDefaultExistsFilter() {
```
nityanandagohain: nit: maybe add a differnet if condition just saying that if it new json then completed.

pkg/telemetrylogs/const.go
```diff
@@ -40,6 +42,7 @@ var (
 		FieldDataType: telemetrytypes.FieldDataTypeString,
 	}
 	BodyJSONStringSearchPrefix = `body.`
+	BodyV2Column               = `body_v2`
```
nityanandagohain: needed ?

pkg/telemetrylogs/json_plan.go
```diff
@@ -0,0 +1,315 @@
...
+	sep := arraySep
+	if n.Parent.isRoot {
+		sep = "."
+	}
+	return fmt.Sprintf("`%s%s%s`", parentAlias, sep, n.Name)
+}
+
+func (n *Node) FieldPath() string {
+	key := n.Name
+	if keycheck.IsBacktickRequired(key) {
```
nityanandagohain: I guess backtick can be added by default, don't see any harm in doing that.

pkg/telemetrylogs/json_plan.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+type BranchType string
+
+const (
+	BranchJSON    BranchType = "json"
+	BranchDynamic BranchType = "dynamic"
+)
+
+type TerminalConfig struct {
```
nityanandagohain: please move types to top.

pkg/telemetrylogs/json_plan.go
```diff
@@ -0,0 +1,315 @@
...
+	PreferArrayAtEnd bool
+	ElemType         telemetrytypes.JSONDataType
+	ValueType        telemetrytypes.JSONDataType
+	Operator         qbtypes.FilterOperator
+	Value            any
+}
+
+// chooseArrayPreference decides whether to search scalar or array at the terminal
+// and returns the preferred element type for the terminal comparison.
+func (node *Node) decideElemType(operator qbtypes.FilterOperator, valueType telemetrytypes.JSONDataType) (preferArray bool, elem telemetrytypes.JSONDataType) {
```
nityanandagohain: need unit tests

pkg/telemetrylogs/json_query_builder.go
```diff
@@ -0,0 +1,660 @@
...
+	lastSeen             uint64
+	promotedPaths        sync.Map     // map[string]struct{} of promoted JSON paths
+	stringIndexedColumns atomic.Value // map[string]string of string indexed columns
+
+	logger *slog.Logger
+}
+
+func NewJSONQueryBuilder(ctx context.Context,
+	telemetryStore telemetrystore.TelemetryStore, logger *slog.Logger) (*JSONQueryBuilder, error) {
+	builder := &JSONQueryBuilder{telemetryStore: telemetryStore, cache: sync.Map{}, logger: logger}
```
nityanandagohain: add limit to cache ? take a good number that should be there and then add it ? also make it configurable maybe

pkg/telemetrylogs/json_query_builder.go
```diff
@@ -0,0 +1,660 @@
...
+// IsPromoted reports whether a JSON path is present in the promoted paths set.
+func (b *JSONQueryBuilder) IsPromoted(path string) bool {
+	firstLeafNode := strings.Split(path, arraySep)[0]
+	_, ok := b.promotedPaths.Load(firstLeafNode)
+	return ok
+}
+
+// syncPromoted refreshes the promoted paths from ClickHouse and reconciles the set.
+func (b *JSONQueryBuilder) syncPromoted(ctx context.Context) error {
+	query := fmt.Sprintf("SELECT path FROM %s.%s", DBName, PromotedPathsTableName)
```
nityanandagohain: Add FINAL ?

pkg/telemetrylogs/json_stmt_builder_test.go
```diff
@@ -0,0 +1,996 @@
...
...l(fromUnixTimestamp64Nano(timestamp), INTERVAL 30 SECOND) AS ts, toString(`body.education[].awards[].type`) AS `body.education[].awards[].type`, count() AS __result_0 FROM signoz_logs.distributed_logs_v2 ARRAY JOIN arrayFlatten(arrayConcat(arrayMap(`body_v2.education`->arrayConcat(arrayMap(`body_v2.education[].awards`->dynamicElement(`body_v2.education[].awards`.type, 'String'), dynamicElement(`body_v2.education`.awards, 'Array(JSON(max_dynamic_types=8, max_dynamic_paths=0))')), arrayMap(`body_v2.education[].awards`->dynamicElement(`body_v2.education[].awards`.type, 'String'), arrayMap(x->assumeNotNull(dynamicElement(x, 'JSON')), arrayFilter(x->(dynamicType(x) = 'JSON'), dynamicElement(`body_v2.education`.awards, 'Array(Dynamic)'))))), dynamicElement(body_v2.education, 'Array(JSON(max_dynamic_types=16, max_dynamic_paths=0))')))) AS `body.education[].awards[].type` WHERE resource_fingerprint GLOBAL IN (SELECT fingerprint FROM __resource_filter) AND timestamp >= ? AND ts_bucket_start >= ? AND timestamp < ? AND ts_bucket_start <= ? AND (`body.education[].awards[].type`) GLOBAL IN (SELECT `body.education[].awards[].type` FROM __limit_cte) GROUP BY ts, `body.education[].awards[].type`",
```
nityanandagohain: as discussed we will  have to have a proper use case of group on keys inside array.

pkg/telemetrylogs/json_query_builder.go
```diff
@@ -0,0 +1,660 @@
...
+
+		expr := fmt.Sprintf("dynamicElement(%s, '%s')", node.FieldPath(), node.TerminalConfig.ElemType.StringValue())
+		if b.IsPromoted(path) {
+			if len(plan) < 2 {
+				return nil, errors.Newf(errors.TypeUnexpected, CodePromotedPlanMissing,
+					"plan length is less than 2 for promoted path: %s", path)
+			}
+
+			// promoted column first then body_v2 column
+			expr = fmt.Sprintf("coalesce(%s, %s)",
```
nityanandagohain: add a todo to change this once we have the new structure, which can be incorporated into the path table. This will be based on my changes of multi if

pkg/telemetrylogs/json_stmt_builder_test.go
```diff
@@ -0,0 +1,996 @@
...
+				require.Equal(t, c.expected.Query, q.Query)
+				require.Equal(t, c.expected.Args, q.Args)
+				require.Equal(t, c.expected.Warnings, q.Warnings)
+			}
+		})
+	}
+}
+
+/*
+		{
```
nityanandagohain: create a json file maybe ?

pkg/telemetrylogs/json_query_builder.go
```diff
@@ -0,0 +1,660 @@
...
+
+	case qbtypes.FilterOperatorContains, qbtypes.FilterOperatorNotContains, qbtypes.FilterOperatorLike, qbtypes.FilterOperatorILike, qbtypes.FilterOperatorNotLike:
+		if node.TerminalConfig.PreferArrayAtEnd {
+			return b.buildArrayMembershipCondition(node, operator, value, sb)
+		}
+		elemType := node.TerminalConfig.ElemType
+		fieldExpr := fmt.Sprintf("dynamicElement(%s, '%s')", fieldPath, elemType.StringValue())
+
+		// if elemType is string and this field path is string indexed, use the string indexing expression
+		if expr, found := b.stringIndexedColumns.Load().(map[string]string)[fieldPath]; found && elemType == telemetrytypes.String {
```
nityanandagohain: need unit test for what this means, as I don't think 
>this field path is string indexed, use the string indexing expression

this is correct way to go i.e use something different if indexed.

pkg/telemetrylogs/statement_builder.go
```diff
@@ -246,6 +255,13 @@ func (b *logQueryStatementBuilder) buildListQuery(
 			if query.SelectFields[index].Name == LogsV2TimestampColumn || query.SelectFields[index].Name == LogsV2IDColumn {
 				continue
 			}
+			// Keep SELECT projection raw for body JSON paths unless narrower typing is required by agg/filter
+			// Here, for plain SELECT, emit raw body_v2.path
+			if b.jsonBodyPrefix != "" && strings.HasPrefix(query.SelectFields[index].Name, b.jsonBodyPrefix) {
+				path := strings.TrimPrefix(query.SelectFields[index].Name, b.jsonBodyPrefix)
+				sb.SelectMore("body_v2." + path)
```
nityanandagohain: see if you can add tests, also it should check if new json is eanabled

pkg/telemetrylogs/statement_builder.go
```diff
@@ -255,8 +271,12 @@ func (b *logQueryStatementBuilder) buildListQuery(
 		}
 	}
 
-	// From table
-	sb.From(fmt.Sprintf("%s.%s", DBName, LogsV2TableName))
+	// From table (inject ARRAY JOINs if collected)
+	fromBase := fmt.Sprintf("%s.%s", DBName, LogsV2TableName)
+	if len(arrayJoinClauses) > 0 {
```
nityanandagohain: what does this mean ?

pkg/telemetrylogs/statement_builder.go
```diff
@@ -330,13 +350,35 @@ func (b *logQueryStatementBuilder) buildTimeSeriesQuery(
...
 	for _, gb := range query.GroupBy {
-		expr, args, err := querybuilder.CollisionHandledFinalExpr(ctx, &gb.TelemetryFieldKey, b.fm, b.cb, keys, telemetrytypes.FieldDataTypeString, b.jsonBodyPrefix, b.jsonKeyToKey)
-		if err != nil {
-			return nil, err
+		var expr string
+		var args []any
+		var err error
+
+		// For body JSON fields with feature flag enabled, use array join logic
+		if strings.HasPrefix(gb.TelemetryFieldKey.Name, BodyJSONStringSearchPrefix) && constants.BodyV2QueryEnabled {
```
nityanandagohain: constants.BodyV2QueryEnabled can be a part of condition builder struct ?

pkg/telemetrylogs/statement_builder.go
```diff
@@ -478,11 +524,33 @@ func (b *logQueryStatementBuilder) buildScalarQuery(
...
+			// Build array join info for this field
```
nityanandagohain: for these will it be clear to create v1/ v2 internal functions and then call whatever is required, even though code will be duplicated.

That way removing the flag will be cleaner.

nityanandagohain (commented):
Have added some comments.

I will have to go through the json plan and statement builder again with unit tests.

And here are few pointers for breaking down this PR. So that codes get merged and then we can integrate it later.
* API for get/update promoted paths can be a single PR.
* API from metadata get from promoted path and `body_keys_json.go` can be a individual PR.
* JSON Plan can be a single PR with unit tests
* statement_builder can be a single PR with unit test for **all the operators** as well, also think about  all the exisiting test cases added, and how helpful is it towards ensuring your code is correct vs someone looking at it and understanding.  Also if you have good test for all operators tests like TestStatementBuilderListQueryBodyMessage can have just one or two cases maybe and won't require multiple test cases.


Also you will have to add integration test as well, do allocate bandwith for that.

pkg/querier/builder_query.go
piyushsingariya: It needs to, else data will not appear from promoted

pkg/querier/consume.go
piyushsingariya: NO

pkg/query-service/app/clickhouseReader/reader.go
piyushsingariya: We're only indexing the String columns of these paths

pkg/telemetrylogs/json_plan.go
piyushsingariya: This is something different

pkg/telemetrylogs/json_stmt_builder_test.go
piyushsingariya: I believe it's oki here for quick access

pkg/query-service/app/clickhouseReader/reader.go
piyushsingariya: Supported numbers now

pkg/querier/builder_query.go
srikanthccv: We wanted to check if it's possible to do the merging of two JSON bodies in ClickHouse instead of this code block here.

pkg/telemetrystore/clickhousetelemetrystore/provider.go
```diff
@@ -32,6 +32,8 @@ func New(ctx context.Context, providerSettings factory.ProviderSettings, config
 	options.MaxIdleConns = config.Connection.MaxIdleConns
 	options.MaxOpenConns = config.Connection.MaxOpenConns
 	options.DialTimeout = config.Connection.DialTimeout
+	// This is to avoid the driver decoding issues with JSON columns
+	options.Settings["output_format_native_write_json_as_string"] = 1
```
therealpandey, line 36: This needs to work with `25.5.6` as well. Please test and check that it is not failing with `25.5.6`

deploy/docker-swarm/docker-compose.ha.yaml
```diff
@@ -11,7 +11,7 @@ x-common: &common
       max-file: "3"
 x-clickhouse-defaults: &clickhouse-defaults
   !!merge <<: *common
-  image: clickhouse/clickhouse-server:25.5.6
+  image: clickhouse/clickhouse-server:25.10.1
```
therealpandey: Please don't upgrade this. OSS will have their own 12.10.1 release plan.

deploy/docker-swarm/docker-compose.yaml
```diff
@@ -11,7 +11,7 @@ x-common: &common
       max-file: "3"
 x-clickhouse-defaults: &clickhouse-defaults
   !!merge <<: *common
-  image: clickhouse/clickhouse-server:25.5.6
+  image: clickhouse/clickhouse-server:25.10.1
```
therealpandey: Same, let's not update!

deploy/docker/docker-compose.ha.yaml
```diff
@@ -10,7 +10,7 @@ x-common: &common
 x-clickhouse-defaults: &clickhouse-defaults
   !!merge <<: *common
   # addding non LTS version due to this fix https://github.com/ClickHouse/ClickHouse/commit/32caf8716352f45c1b617274c7508c86b7d1afab
-  image: clickhouse/clickhouse-server:25.5.6
+  image: clickhouse/clickhouse-server:25.10.1
```
therealpandey: same!

deploy/docker/docker-compose.yaml
```diff
@@ -9,7 +9,7 @@ x-common: &common
       max-file: "3"
 x-clickhouse-defaults: &clickhouse-defaults
   !!merge <<: *common
-  image: clickhouse/clickhouse-server:25.5.6
+  image: clickhouse/clickhouse-server:25.10.1
```
therealpandey: same!

pkg/telemetrystore/clickhousetelemetrystore/provider.go
piyushsingariya: make sense.

deploy/docker-swarm/docker-compose.ha.yaml
piyushsingariya: Yes, I am gonna revert this change

pkg/telemetrylogs/field_mapper.go
```diff
@@ -106,23 +123,41 @@ func (m *fieldMapper) FieldFor(ctx context.Context, key *telemetrytypes.Telemetr
 		return "", err
 	}
 
+	if !querybuilder.BodyJSONQueryEnabled && key.FieldContext == telemetrytypes.FieldContextBody {
+		return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "Group by/Aggregation isn't available for the body column")
+	}
```
srikanthccv: The `FieldFor` shouldn't return this error. It should only be returned in the context of grouping / aggregation.

pkg/telemetrymetadata/body_json_metadata.go
srikanthccv, line 1: 1. The prepared query for fetching the paths in `buildGetBodyJSONPathsQuery` is incorrect, It doesn't have the wildcards similar to `buildListLogsJSONIndexesQuery`. Please check, https://github.com/SigNoz/signoz/blob/f6da9adb86a752083d0c8a0becc1e9d3c2fd1164/pkg/telemetrymetadata/body_json_metadata.go#L145
2. Return empty paths instead of error when no field selectors provided https://github.com/SigNoz/signoz/blob/f6da9adb86a752083d0c8a0becc1e9d3c2fd1164/pkg/telemetrymetadata/body_json_metadata.go#L120-L122
3. Would it ever be possible that mapping is not found _in real_ i.e with the user sent data that goes through out pipeline? If does, what is the most sensible choice to make, 1. return error 2. log error (and monitor) and skip the type? I think it should be log error and continue because returning error would mean one incorrect thing breaks the legit query were the missing map could not even be relevant (well, because we don't support in the first place) https://github.com/SigNoz/signoz/blob/f6da9adb86a752083d0c8a0becc1e9d3c2fd1164/pkg/telemetrymetadata/body_json_metadata.go#L81-L84 https://github.com/SigNoz/signoz/blob/f6da9adb86a752083d0c8a0becc1e9d3c2fd1164/pkg/telemetrymetadata/body_json_metadata.go#L195-L198
4. Same here, return empty rather than error when filtered paths is empty, (as returning error would be incorrect) https://github.com/SigNoz/signoz/blob/f6da9adb86a752083d0c8a0becc1e9d3c2fd1164/pkg/telemetrymetadata/body_json_metadata.go#L176-L178

pkg/telemetrymetadata/metadata.go
```diff
@@ -572,6 +572,14 @@ func (t *telemetryMetaStore) getLogsKeys(ctx context.Context, fieldKeySelectors
...
 	}
 
+	if querybuilder.BodyJSONQueryEnabled {
+		bodyJSONPaths, finished, err := t.getBodyJSONPaths(ctx, fieldKeySelectors) // LIKE for pattern matching
+		if err != nil {
+			t.logger.ErrorContext(ctx, "failed to extract body JSON paths", "error", err)
+		}
+		keys = append(keys, bodyJSONPaths...)
+		complete = complete && finished
+	}
```
srikanthccv, line 583: This block should make use of the `mapOfKeys` to ensure the returned list is populated correctly. For example, if there is a key `config.batch_size` in both attributes (from team 1) and body, then there should be one key with `[]*telemetrytypes.TelemetryFieldKey{}` than two in the `keys` list.

pkg/telemetrymetadata/metadata.go
```diff
@@ -1164,6 +1172,10 @@ func (t *telemetryMetaStore) getLogFieldValues(ctx context.Context, fieldValueSe
 		limit = 50
 	}
 
+	if strings.HasPrefix(fieldValueSelector.Name, telemetrytypes.BodyJSONStringSearchPrefix) {
+		return t.ListJSONValues(ctx, fieldValueSelector.Name, limit)
+	}
```
srikanthccv: This is going to be very expensive on production workloads, we shouldn't use main table for value suggestions.

pkg/telemetrylogs/json_condition_builder.go
```diff
@@ -0,0 +1,455 @@
...
+
+// emitPlannedCondition handles paths with array traversal
+func (c *conditionBuilder) emitPlannedCondition(plan *telemetrytypes.JSONAccessNode, operator qbtypes.FilterOperator, value any, sb *sqlbuilder.SelectBuilder) (string, error) {
+	// Build traversal + terminal recursively per-hop
+	compiled, err := c.recurseArrayHops(plan, operator, value, sb)
+	if err != nil {
+		return "", err
+	}
+
+	// sb.AddWhereClause(sqlbuilder.NewWhereClause().AddWhereExpr(sb.Args, compiled))
```
srikanthccv:
```suggestion
```

pkg/telemetrylogs/json_condition_builder.go
```diff
@@ -0,0 +1,455 @@
...
+		}
+	}
+	return types, nil
+}
+
+// BuildCondition builds the full WHERE condition for body_json JSON paths
+func (c *conditionBuilder) buildJSONCondition(ctx context.Context, key *telemetrytypes.TelemetryFieldKey,
+	operator qbtypes.FilterOperator, value any, sb *sqlbuilder.SelectBuilder) (string, error) {
+
+	plan, err := PlanJSON(ctx, key, operator, value, c.getTypes)
```
srikanthccv: Looking at the `PlanJSON` implementation, `getTypes` is invoked independently for each sub-path, which calls the `GetKeys` triggering a series of calls ~ 5 calls. That would be many too many small queries to DB for a non-trivial search expression. Is there any reason to pass the `getTypes` as callable than directly passing the required data for `PlanJSON` in one go?

pkg/telemetrylogs/json_condition_builder.go
```diff
@@ -0,0 +1,455 @@
...
+	// Use the parent's alias + current field name for the full path
+	fieldPath := node.FieldPath()
+
+	if node.TerminalConfig.ElemType.IsArray {
+		// switch operator for array membership checks
+		switch operator {
+		case qbtypes.FilterOperatorContains, qbtypes.FilterOperatorIn:
+			operator = qbtypes.FilterOperatorEqual
+		case qbtypes.FilterOperatorNotContains, qbtypes.FilterOperatorNotIn:
+			operator = qbtypes.FilterOperatorNotEqual
```
srikanthccv: (not)contains it makes sense, but `in` doesn't => `=` work, as it would result in invalid query as it build something like `x -> x = ?, ...` with args as array.

pkg/telemetrylogs/json_condition_builder.go
```diff
@@ -0,0 +1,455 @@
...
+			value = emptyValue
+		case qbtypes.FilterOperatorNotExists:
+			operator = qbtypes.FilterOperatorEqual
+			value = emptyValue
+		default:
+			// do nothing
+		}
+
+		cond, err := c.applyOperator(sb, indexedExpr, operator, value)
```
srikanthccv: The docs say,

>Returns the non-nullable value, if the original value was not NULL, otherwise an arbitrary value, if the input value is NULL. [Any](https://clickhouse.com/docs/sql-reference/data-types)

Then there is a chance, for the `EXISTS` operator the expression could give incorrect result depending on the value returned by `assumeNotNull`? Trying it on fiddle returns empty string for Nullable strings but the docs say arbitrary value which makes it possible it is not reliable.

pkg/telemetrylogs/json_condition_builder.go
```diff
@@ -0,0 +1,455 @@
...
+		if err != nil {
+			return "", err
+		}
+		conditions = append(conditions, cond)
+		// Switch operator to EXISTS
+		operator = qbtypes.FilterOperatorExists
```
srikanthccv, line 152: The flip should be based on the original operator than just switch to EXISTS

pkg/telemetrylogs/json_condition_builder.go
```diff
@@ -0,0 +1,455 @@
...
+	conditions := []string{}
+
+	elemType := node.TerminalConfig.ElemType
+	fieldExpr := fmt.Sprintf("dynamicElement(%s, '%s')", fieldPath, elemType.StringValue())
+	fieldExpr, value = querybuilder.DataTypeCollisionHandledFieldName(node.TerminalConfig.Key, value, fieldExpr, operator)
+
+	indexed := slices.ContainsFunc(node.TerminalConfig.Key.Indexes, func(index telemetrytypes.JSONDataTypeIndex) bool {
+		return index.Type == elemType && index.ColumnExpression == fieldPath
+	})
+	if elemType.IndexSupported && indexed {
```
srikanthccv, line 112: Looking at the if block, it is a variation of the main filter but with the `assumeNotNull`. Can you please share how the indexing is at play here?

pkg/telemetrylogs/json_condition_builder.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+	conditions = append(conditions, cond)
+	if len(conditions) > 1 {
+		return sb.And(conditions...), nil
+	}
+	return cond, nil
+}
+
+// buildArrayMembershipCondition handles array membership checks
+func (c *conditionBuilder) buildArrayMembershipCondition(node *telemetrytypes.JSONAccessNode, operator qbtypes.FilterOperator, value any, sb *sqlbuilder.SelectBuilder) (string, error) {
```
srikanthccv, line 167: note: Array membership check from client can come in form of `has(config.features, 'my-feature'), it should be handled.

pkg/telemetrylogs/json_condition_builder.go
```diff
@@ -0,0 +1,455 @@
...
+}
+
+// GroupByArrayJoinInfo contains information about array joins needed for GroupBy
+type GroupByArrayJoinInfo struct {
+	ArrayJoinClauses []string // ARRAY JOIN clauses to add to FROM clause
+	TerminalExpr     string   // Terminal field expression for SELECT/GROUP BY
+}
+
+// BuildGroupBy builds GroupBy information for body JSON fields using arrayConcat pattern
+func (c *conditionBuilder) BuildGroupBy(ctx context.Context, key *telemetrytypes.TelemetryFieldKey) (*GroupByArrayJoinInfo, error) {
```
srikanthccv, line 318: wrong place, why is there a `BuildGroupBy` on condition builder?

pkg/telemetrymetadata/metadata.go
piyushsingariya: Yes, we had the discussions to remove from the main flow

pkg/telemetrylogs/json_condition_builder.go
piyushsingariya: Make sense, No reason, it was miss from my side

pkg/telemetrylogs/json_condition_builder.go
piyushsingariya: args as array? Why would customer pass it as an array? Let's hop on a call?

pkg/telemetrylogs/json_condition_builder.go
piyushsingariya: Added an explanation on why this is being done

pkg/telemetrylogs/json_condition_builder.go
piyushsingariya: This was done, thinking we can offer GroupBy on fields inside arrays but me n @nityanandagohain decided this doesn't fit right in observability space more related to data engineering usecase

pkg/telemetrymetadata/metadata.go
piyushsingariya: sorted over call, parent function is making a map actually

pkg/telemetrylogs/field_mapper.go
```diff
@@ -109,20 +126,33 @@ func (m *fieldMapper) FieldFor(ctx context.Context, key *telemetrytypes.Telemetr
...
+			if key.Materialized {
+				oldKeyName = telemetrytypes.FieldKeyToMaterializedColumnName(key)
+				oldKeyNameExists := telemetrytypes.FieldKeyToMaterializedColumnNameForExists(key)
+				return fmt.Sprintf("multiIf(%s.`%s` IS NOT NULL, %s.`%s`::String, %s==true, %s, NULL)", column.Name, key.Name, column.Name, key.Name, oldKeyNameExists, oldKeyName), nil
+			}
 			return fmt.Sprintf("multiIf(%s.`%s` IS NOT NULL, %s.`%s`::String, mapContains(%s, '%s'), %s, NULL)", column.Name, key.Name, column.Name, key.Name, oldColumn.Name, key.Name, oldKeyName), nil
+		case telemetrytypes.FieldContextBody:
+			if strings.Contains(key.Name, telemetrytypes.ArraySep) || strings.Contains(key.Name, telemetrytypes.ArrayAnyIndex) {
+				return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "Group by/Aggregation isn't available for the Array Paths: %s", key.Name)
+			}
```
srikanthccv: This is incorrect, we will never know if the `FieldFor` is used for the `Group by/Aggregation`. It's not aware of the condition it will be used it. It's concern is to either return the field a key maps to on DB or return error.

pkg/telemetrylogs/field_mapper.go
```diff
@@ -109,20 +126,33 @@ func (m *fieldMapper) FieldFor(ctx context.Context, key *telemetrytypes.Telemetr
...
+			fieldExpr := BodyJSONColumnPrefix + fmt.Sprintf("`%s`", key.Name)
+			expr := fmt.Sprintf("dynamicElement(%s, '%s')", fieldExpr, key.JSONDataType.StringValue())
+			if key.Materialized {
+				promotedFieldExpr := BodyPromotedColumnPrefix + fmt.Sprintf("`%s`", key.Name)
+				expr = fmt.Sprintf("coalesce(%s, %s)", expr, fmt.Sprintf("dynamicElement(%s, '%s')", promotedFieldExpr, key.JSONDataType.StringValue()))
+			}
+			// returning qbtypes.ErrColumnNotFound is a hack that will trigger the fallback expr logic to include all the types for the key
+			return expr, qbtypes.ErrColumnNotFound
```
srikanthccv: This doesn't make sense, we shouldn't return both expression and error. Given a key, it either maps to something on DB, or doesn't, in which case an error is returned.

pkg/telemetrylogs/field_mapper.go
```diff
@@ -109,20 +126,33 @@ func (m *fieldMapper) FieldFor(ctx context.Context, key *telemetrytypes.Telemetr
...
+		default:
+			return "", errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "only resource context fields are supported for json columns, got %s", key.FieldContext.String)
```
srikanthccv:
```suggestion
			return "", errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "only resource/body context fields are supported for json columns, got %s", key.FieldContext.String)
```

pkg/querybuilder/fallback_expr.go
```diff
@@ -219,8 +219,36 @@ func DataTypeCollisionHandledFieldName(key *telemetrytypes.TelemetryFieldKey, va
 			// we don't have a toBoolOrNull in ClickHouse, so we need to convert the bool to a string
 			value = fmt.Sprintf("%t", v)
 		}
+	case telemetrytypes.FieldDataTypeFloat64,
+		telemetrytypes.FieldDataTypeArrayFloat64:
```
srikanthccv: Can you help me understand why is this separated out and uses the exactly same logic as the below `case` block.

pkg/querier/consume.go
srikanthccv: Please create one

pkg/querier/consume.go
srikanthccv: As we return bytes above, we don't need this case?

pkg/telemetrylogs/condition_builder.go
```diff
@@ -30,22 +31,34 @@ func (c *conditionBuilder) conditionFor(
...
-
 	column, err := c.fm.ColumnFor(ctx, key)
 	if err != nil {
 		return "", err
 	}
 
+	// For JSON columns, preserve the original value type (numeric, bool, etc.)
+	// Only format to string for non-JSON columns that need string formatting
+	isJSONColumn := column.IsJSONColumn() && querybuilder.BodyJSONQueryEnabled && key.FieldContext == telemetrytypes.FieldContextBody
+	if !isJSONColumn {
```
srikanthccv: Why is this skipped here for the body json column? If somebody does body.http.status_code LIKE 2, it should still work.

pkg/telemetrylogs/condition_builder.go
```diff
@@ -247,7 +258,7 @@ func (c *conditionBuilder) ConditionFor(
 		return "", err
 	}
 
-	if operator.AddDefaultExistsFilter() {
+	if !(key.FieldContext == telemetrytypes.FieldContextBody && querybuilder.BodyJSONQueryEnabled) && operator.AddDefaultExistsFilter() {
```
srikanthccv, line 250: question: how is the zero value condition taken care of `body.http.status_code < 500` should not include the rows which don't have the `body.http.status_code`

pkg/telemetrylogs/field_mapper.go
piyushsingariya: Error can be changed, but for now FieldFor can not work for Array Paths. Changing error

pkg/querybuilder/fallback_expr.go
piyushsingariya: I could have had something else here earlier, removing this

pkg/telemetrylogs/condition_builder.go
piyushsingariya: All paths inside JSON are Nullable, so doing `body.http.status_code < 500` itself is filtering rows where `body.http.status_code` exists

pkg/telemetrylogs/condition_builder.go
piyushsingariya: Because everything is converted to String by this function, hence I am skipping it

pkg/telemetrylogs/field_mapper.go
piyushsingariya: let's discuss this

pkg/telemetrylogs/json_condition_builder.go
piyushsingariya: discussed over call, marking resolved

pkg/telemetrystore/clickhousetelemetrystore/provider.go
piyushsingariya: This setting exists in 25.5.6

pkg/telemetrylogs/field_mapper.go
piyushsingariya: JSON logic branches just before FieldFor because accessing internals of JSON is different then direct FieldFor access.

Note: Check with Nitya on older arrayany `[*]` search logic trigger

pkg/telemetrylogs/json_condition_builder.go
piyushsingariya: Added a condition to address it, not testing the query that is going to be built, will fix later

pkg/telemetrylogs/condition_builder.go
piyushsingariya: handled in JSON

pkg/telemetrylogs/json_access_pb.go
```diff
@@ -80,14 +78,15 @@ func (pb *JSONAccessPlanBuilder) buildPlan(ctx context.Context, index int, paren
 			ValueType: telemetrytypes.MappingFieldDataTypeToJSONDataType[valueType],
 		}
 	} else {
+		var err error
```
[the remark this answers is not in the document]
piyushsingariya: Not valid, adjustKey is not even called for Where keys

srikanthccv (approved):

therealpandey (approved):

pkg/telemetrylogs/statement_builder.go
```diff
@@ -171,6 +168,29 @@ func (b *logQueryStatementBuilder) adjustKeys(ctx context.Context, keys map[stri
...
+					k.Materialized = fieldKeys[0].Materialized
+				}
+				k.Indexes = fieldKeys[0].Indexes
+
+				overallMatch = true // because we found a match
+			} else {
+				b.logger.InfoContext(ctx, "overriding the field context and data type", "key", k.Name)
+				k.FieldContext = telemetrytypes.FieldContextBody
+				k.FieldDataType = telemetrytypes.FieldDataTypeString
+				k.JSONDataType = &telemetrytypes.String
```
tushar-signoz, line 190: @piyushsingariya Help me understand this piece of code, Im getting merge conflicts for my PR

PR #9386, opened 2025-10-21
feat: allowing switching between views when groupby is present

YounixM:
remove group by from queries when switching from time series / table view to list / traces view.  This allows switching to list / traces view even when groupBy existed.

### DIFF
--- a/frontend/src/api/settings/setRetentionV2.ts
+++ b/frontend/src/api/settings/setRetentionV2.ts
@@ -8,7 +8,7 @@ const setRetentionV2 = async ({
 	type,
 	defaultTTLDays,
 	coldStorageVolume,
-	coldStorageDuration,
+	coldStorageDurationDays,
 	ttlConditions,
 }: PropsV2): Promise<SuccessResponseV2<PayloadPropsV2>> => {
 	try {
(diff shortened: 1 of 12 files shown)


nityanandagohain (changes requested):
please add it for table -> list view as well

nityanandagohain (dismissed):

frontend/src/pages/TracesExplorer/index.tsx
```diff
@@ -312,6 +310,8 @@ function TracesExplorer(): JSX.Element {
 
 	const [warning, setWarning] = useState<Warning | undefined>(undefined);
 
+	console.log('currentQuery in traces explorer', currentQuery);
```
nityanandagohain: can be removed?

srikanthccv:
There was some regression since the initial release of QB and This is not the right fix. When user switches b/w the view the relevant context is passed&respected and view changes instead of removing the parts from the actual query that was present because when they switch back again to time series, the earlier context is preserved without forcing user to re-enter the same.

nityanandagohain:
@srikanthccv we are going ahead with merging this as it will allow switching, have created a issue https://github.com/SigNoz/signoz/issues/9424 to address in the coming sprint.

nityanandagohain (approved):

aks07 (approved):

PR #9395, opened 2025-10-22
fix(sqlschema): postgres sqlschema get table operation

vikrantgupta25:
## Description

postgres sqlschema `GetTable` operation returned nil,nil if the table isn't created rather it should throw error in this case.

Contributes to : https://github.com/SigNoz/platform-pod/issues/583

### DIFF
--- a/ee/sqlschema/postgressqlschema/provider.go
+++ b/ee/sqlschema/postgressqlschema/provider.go
@@ -51,7 +51,7 @@ func (provider *provider) GetTable(ctx context.Context, tableName sqlschema.Tabl
 	columns := []struct {
 		ColumnName  string  `bun:"column_name"`
 		Nullable    bool    `bun:"nullable"`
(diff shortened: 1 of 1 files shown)


ee/sqlschema/postgressqlschema/provider.go
```diff
@@ -47,50 +48,45 @@ func (provider *provider) Operator() sqlschema.SQLOperator {
 }
 
 func (provider *provider) GetTable(ctx context.Context, tableName sqlschema.TableName) (*sqlschema.Table, []*sqlschema.UniqueConstraint, error) {
-	rows, err := provider.
+	columns := []struct {
+		ColumnName  string  `bun:"column_name"`
+		Nullable    bool    `bun:"nullable"`
+		SqlDataType string  `bun:"udt_name"`
```
therealpandey:
```suggestion
		SQLDataType string  `bun:"udt_name"`
```

PR #9418, opened 2025-10-27
fix: preserve SMTPRequireTLS during default merge (#8478)

### LINKED ISSUE #8478: [config] require tls for alertmanager (SIGNOZ_ALERTMANAGER_SIGNOZ_GLOBAL_SMTP__REQUIRE__TLS) is not working
I've been trying to get SMTP forwarding configured to work with alerting on an Ubuntu install of Signoz v0.84.1. When performing a test alert notification from the browser, I get this error:
unknown
establish connection to server: dial tcp 127.0.0.1:25: connect: connection refused

In /etc/environment I've configured it this way:

SIGNOZ_ALERTMANAGER_SIGNOZ_GLOBAL_SMTP__FROM="<an email address>"
SIGNOZ_ALERTMANAGER_SIGNOZ_GLOBAL_SMTP__SMARTHOST="<my mail server>:25"
SIGNOZ_ALERTMANAGER_SIGNOZ_GLOBAL_SMTP__REQUIRE__TLS="false"
SIGNOZ_ALERTMANAGER_SIGNOZ_GLOBAL_SMTP__TLS__CONFIG_INSECURE__SKIP__VERIFY="true"
* Private info replaced with < >

I've tested the email relay with telnet from the server and it sends fine on port 25. What am I missing?

pkalsi97:
## 📄 Summary
As mentioned in Issue (#8478), a few of people were facing issue to get SMTP forwarding configured to work when using 
```makefile
SIGNOZ_ALERTMANAGER_SIGNOZ_GLOBAL_SMTP__FROM=test@local \
SIGNOZ_ALERTMANAGER_SIGNOZ_GLOBAL_SMTP__SMARTHOST=:25 \
SIGNOZ_ALERTMANAGER_SIGNOZ_GLOBAL_SMTP__REQUIRE__TLS=false \
SIGNOZ_ALERTMANAGER_SIGNOZ_GLOBAL_SMTP__TLS__CONFIG_INSECURE__SKIP__VERIFY=true \
``` 
Referencing @therealpandey’s note about boolean handling: https://github.com/SigNoz/signoz/issues/8478#issuecomment-3299230883

The issue was with how Mergo Treats Zero values, Mergo only fills **zero-value** fields in the destination.
Since `false` is the zero value for `bool`, it always gets **replaced** by `true` from the source. Using pointers doesn’t help—`Merge` dereferences them and still treats `false` as zero.

https://github.com/SigNoz/signoz/blob/5f8cfbe474ad810fd209492e415e745de7279527/pkg/types/alertmanagertypes/config.go#L96

https://github.com/SigNoz/signoz/blob/5f8cfbe474ad810fd209492e415e745de7279527/pkg/types/alertmanagertypes/config.go#L170

In both instances, globalConfig.SMTPRequireTLS was always set to `true`, even when explicitly set to`false` by the user.

Neither `WithOverride` nor `WithOverwriteWithEmptyValue` can distinguish between an explicit `false` and an unset value, so only a manual or custom merge can **preserve** user intent.
#### Current Approach
Preserve the explicit field value before merge, restore it after in both of the mentioned instances:
```go
smtpRequireTls := globalConfig.SMTPRequireTLS
err := mergo.Merge(&globalConfig, config.DefaultGlobalConfig())
if err != nil {
    return nil, err
}
globalConfig.SMTPRequireTLS = smtpRequireTls
``` 
This minimal patch fixes the problem without altering struct definitions or wider behaviour but is not future proof.
Any future boolean or similar field requires another manual restore line.

---

## ✅ Changes
- [x] Bug fix: Fixes boolean merge issue where `SIGNOZ_ALERTMANAGER_SIGNOZ_GLOBAL_SMTP__REQUIRE__TLS=false` was ignored due to Mergo treating `false` as a zero value and overwriting it with the default `true`.

---

- `bug`
- `backend`
---

## 👥 Reviewers
- backend

---

## 🧪 How to Test

1. In the project root, open the `Makefile`.
   Add the following environment variables under the `go-run-community` target:

   ```bash
   SIGNOZ_ALERTMANAGER_SIGNOZ_GLOBAL_SMTP__FROM=test@local \
   SIGNOZ_ALERTMANAGER_SIGNOZ_GLOBAL_SMTP__SMARTHOST=:25 \
   SIGNOZ_ALERTMANAGER_SIGNOZ_GLOBAL_SMTP__REQUIRE__TLS=false \
   SIGNOZ_ALERTMANAGER_SIGNOZ_GLOBAL_SMTP__TLS__CONFIG_INSECURE__SKIP__VERIFY=true \
   ```
2. Start a new instance:

   ```bash
   make go-run-community
   ```
3. Create an admin account through the UI.
   The above environment variables should now be reflected in the `alertmanager_config` table inside `signoz.db`.
4. Run a mock SMTP server (e.g., MailHog) on the same port (`:25`).
5. Create a notification channel using this SMTP configuration and send a test notification.
   The test should succeed, confirming that `SMTPRequireTLS=false` is now respected.

---

## 🔍 Related Issues

Closes #8478

---

#### Recreation of Issue

https://github.com/user-attachments/assets/ca8a0678-b5c8-4b5c-a154-1b66dc7911e2

#### Explanation of solution

https://github.com/user-attachments/assets/b44a1afe-de78-4363-a692-bbf6b0bff220

#### Test Channels 

https://github.com/user-attachments/assets/a824a188-06d3-4964-bde2-ddebf71e81fb

#### Test Alerts

https://github.com/user-attachments/assets/9b585002-f508-4132-8d89-5b5343606f53

---

## 📋 Checklist

- [x] Dev Review

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> Preserves user-specified `SMTPRequireTLS` during default config merges and adds unit tests to verify it in both constructor and setter paths.
> 
> - **alertmanager types**:
>   - **Global merge fix**: In `NewDefaultConfig` and `SetGlobalConfig`, preserve `GlobalConfig.SMTPRequireTLS` across `mergo.Merge` to prevent overwriting explicit `false`.
>   - **Tests**: Add `TestNewDefaultConfigPreservesSMTPRequireTLS` and `TestSetGlobalConfigPreservesSMTPRequireTLS` to validate behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1006e3a091df223b8b1679d40fa77974f19b52fb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.devenv/docker/clickhouse/compose.yaml
+++ b/.devenv/docker/clickhouse/compose.yaml
@@ -42,7 +42,7 @@ services:
       timeout: 5s
       retries: 3
   schema-migrator-sync:
-    image: signoz/signoz-schema-migrator:v0.129.7
+    image: signoz/signoz-schema-migrator:v0.129.8
     container_name: schema-migrator-sync
     command:
       - sync
@@ -55,7 +55,7 @@ services:
         condition: service_healthy
     restart: on-failure
   schema-migrator-async:
-    image: signoz/signoz-schema-migrator:v0.129.7
+    image: signoz/signoz-schema-migrator:v0.129.8
     container_name: schema-migrator-async
     command:
       - async
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,7 +2,7 @@
 # Owners are automatically requested for review for PRs that changes code
 # that they own.
 
-/frontend/ @SigNoz/frontend @YounixM
+/frontend/ @YounixM @aks07
 /frontend/src/container/MetricsApplication @srikanthccv
 /frontend/src/container/NewWidget/RightContainer/types.ts @srikanthccv
 
--- a/.github/workflows/build-enterprise.yaml
+++ b/.github/workflows/build-enterprise.yaml
@@ -107,7 +107,6 @@ jobs:
         -X github.com/SigNoz/signoz/pkg/version.branch=${{ needs.prepare.outputs.branch }}
         -X github.com/SigNoz/signoz/ee/zeus.url=https://api.signoz.cloud
         -X github.com/SigNoz/signoz/ee/zeus.deprecatedURL=https://license.signoz.io
-        -X github.com/SigNoz/signoz/ee/query-service/constants.ZeusURL=https://api.signoz.cloud
         -X github.com/SigNoz/signoz/ee/query-service/constants.LicenseSignozIo=https://license.signoz.io/api/v1
         -X github.com/SigNoz/signoz/pkg/analytics.key=9kRrJ7oPCGPEJLF6QjMPLt5bljFhRQBr'
       DOCKER_BASE_IMAGES: '{"alpine": "alpine:3.20.3"}'
--- a/.github/workflows/build-staging.yaml
+++ b/.github/workflows/build-staging.yaml
@@ -106,7 +106,6 @@ jobs:
         -X github.com/SigNoz/signoz/pkg/version.branch=${{ needs.prepare.outputs.branch }}
         -X github.com/SigNoz/signoz/ee/zeus.url=https://api.staging.signoz.cloud
         -X github.com/SigNoz/signoz/ee/zeus.deprecatedURL=https://license.staging.signoz.cloud
-        -X github.com/SigNoz/signoz/ee/query-service/constants.ZeusURL=https://api.staging.signoz.cloud
         -X github.com/SigNoz/signoz/ee/query-service/constants.LicenseSignozIo=https://license.staging.signoz.cloud/api/v1
         -X github.com/SigNoz/signoz/pkg/analytics.key=9kRrJ7oPCGPEJLF6QjMPLt5bljFhRQBr'
       DOCKER_BASE_IMAGES: '{"alpine": "alpine:3.20.3"}'
--- a/.github/workflows/integrationci.yaml
+++ b/.github/workflows/integrationci.yaml
@@ -17,6 +17,7 @@ jobs:
           - bootstrap
           - passwordauthn
           - callbackauthn
+          - cloudintegrations
           - querier
           - ttl
         sqlstore-provider:
--- a/cmd/enterprise/.goreleaser.yaml
+++ b/cmd/enterprise/.goreleaser.yaml
@@ -31,7 +31,6 @@ builds:
       - -X github.com/SigNoz/signoz/pkg/version.branch={{ .Branch }}
       - -X github.com/SigNoz/signoz/ee/zeus.url=https://api.signoz.cloud
       - -X github.com/SigNoz/signoz/ee/zeus.deprecatedURL=https://license.signoz.io
-      - -X github.com/SigNoz/signoz/ee/query-service/constants.ZeusURL=https://api.signoz.cloud
       - -X github.com/SigNoz/signoz/ee/query-service/constants.LicenseSignozIo=https://license.signoz.io/api/v1
       - -X github.com/SigNoz/signoz/pkg/analytics.key=9kRrJ7oPCGPEJLF6QjMPLt5bljFhRQBr
     mod_timestamp: "{{ .CommitTimestamp }}"
--- a/deploy/docker-swarm/docker-compose.ha.yaml
+++ b/deploy/docker-swarm/docker-compose.ha.yaml
@@ -176,7 +176,7 @@ services:
       # - ../common/clickhouse/storage.xml:/etc/clickhouse-server/config.d/storage.xml
   signoz:
     !!merge <<: *db-depend
-    image: signoz/signoz:v0.99.0
+    image: signoz/signoz:v0.100.1
     command:
       - --config=/root/config/prometheus.yml
     ports:
@@ -209,7 +209,7 @@ services:
       retries: 3
   otel-collector:
     !!merge <<: *db-depend
-    image: signoz/signoz-otel-collector:v0.129.7
+    image: signoz/signoz-otel-collector:v0.129.8
     command:
       - --config=/etc/otel-collector-config.yaml
       - --manager-config=/etc/manager-config.yaml
@@ -233,7 +233,7 @@ services:
       - signoz
   schema-migrator:
     !!merge <<: *common
-    image: signoz/signoz-schema-migrator:v0.129.7
+    image: signoz/signoz-schema-migrator:v0.129.8
     deploy:
       restart_policy:
         condition: on-failure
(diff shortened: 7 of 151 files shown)


therealpandey:
@pkalsi97 can you write some unit tests for this?

pkalsi97:
@therealpandey, sure will do in some time.

pkalsi97:
@therealpandey 
I have added the unit test
```go
// Test both boolean states to ensure SMTPRequireTLS value is preserved.
testCases := []struct {
    name         string
    globalConfig GlobalConfig
    expect       bool
}{
    {"False", GlobalConfig{SMTPRequireTLS: false}, false},
    {"True", GlobalConfig{SMTPRequireTLS: true}, true},
}
```

**Observation:** If the user does not set `SMTPRequireTLS`, `globalConfig` defaults to `true`. Only two states exist: `true` or `false`.

therealpandey:
@pkalsi97,
When a user does not set this environment variable at all (unset case):
1. What happened before?
2. What will happen now?

pkalsi97:
@therealpandey 
### Question 
When a user does not set this environment variable at all (unset case):
1. What happened before?
`SMTPRequireTLS` was being set to `true` before `NewDefaultConfig` was called and after mergo.Merge still remained `true`
2. What will happen now?
Same behaviour as above.

### Conclusion
So if user does not set env for SMTP, default values are populated i.e `SMTPRequireTLS` will be set to `true`, and the PATCH made in this PR will not affect it. When ever `NewDefaultConfig` or `SetGlobalConfig` is called, the globalConfig argument they will receive will always have **SMTPRequireTLS = true**

Only incase User explicitly sets it SMTPRequireTLS = false, they only Patch will work to preserve that value.

### Explanation
As per my understanding of how AlertManager configuration is created, 
1. In **cmd/community/server.go**  `registerServer` executes
https://github.com/SigNoz/signoz/blob/7b1ca9a1a6273e576d517f5762effcb3994c6398/cmd/community/server.go#L35

3. In **cmd/config.go**   `NewSigNozConfig` calls 
https://github.com/SigNoz/signoz/blob/7b1ca9a1a6273e576d517f5762effcb3994c6398/cmd/config.go#L14

4. In **pkg/signoz/config.go**  `NewConfig` creates `configFactories` When this is called 
https://github.com/SigNoz/signoz/blob/7b1ca9a1a6273e576d517f5762effcb3994c6398/pkg/signoz/config.go#L151

5. in **pkg/alertmanager/config.go**  `NewConfigFactory()` further calls `newConfig()` which loads Config from `alertmanagerserver.NewConfig()`
https://github.com/SigNoz/signoz/blob/7b1ca9a1a6273e576d517f5762effcb3994c6398/pkg/alertmanager/config.go#L43

6. In **pkg/alertmanager/alertmanagerserver/config.go**  `newConfig()` return hardcoded Default values as `factory.Config`.
https://github.com/SigNoz/signoz/blob/7b1ca9a1a6273e576d517f5762effcb3994c6398/pkg/alertmanager/alertmanagerserver/config.go#L67

7. In **pkg/signoz/config.go**  merge overrides happenes where env variables are loaded then finally unmarshal into struct, apply deprecated override layer, validated and Return `Config`
https://github.com/SigNoz/signoz/blob/7b1ca9a1a6273e576d517f5762effcb3994c6398/pkg/signoz/config.go#L161


8. Finally In **cmd/community/server.go** Config is passed into signoz.New, and each provider receives its config when constructed.
https://github.com/SigNoz/signoz/blob/7b1ca9a1a6273e576d517f5762effcb3994c6398/cmd/community/server.go#L59

9. Based on the above when A new org account is created, in pkg/alertmanager/signozalertmanager/provider.go `SetDefaultConfig` gets called where `NewDefaultConfig` is called for the first time  and if in step 7 no env were found then `provider.config.Signoz.Config.Global.SMTPRequireTLS = true`
https://github.com/SigNoz/signoz/blob/7b1ca9a1a6273e576d517f5762effcb3994c6398/pkg/alertmanager/signozalertmanager/provider.go#L209

srikanthccv:
@pkalsi97 can you also check for other booleans from the config and fix them as well, example InsecureSkipVerify

pkalsi97:
@srikanthccv 
As per my understanding 

If we check where mergo.Merger is being used in the codebase, Which explains why `SMTPRequireTLS` was only affected 
```
[~/oss/signoz] grep -R '"dario.cat/mergo"' -n .
./pkg/types/alertmanagertypes/config.go:11:     "dario.cat/mergo"
``` 

I checked the following `bool` in there respective config.go 
| Component         | Field                          | Default (hardcoded) | After env + merge (shown in log) |
|-------------------|--------------------------------|----------------------|----------------------------------|
| Version           | Banner.Enabled                  | true                 | true                             |
| Instrumentation   | Traces.Enabled                  | false                | false                            |
| Instrumentation   | Metrics.Enabled                 | true                 | true                             |
| Analytics         | Enabled                         | false                | false                            |
| Web               | Enabled                         | true                 | false   |
| Prometheus        | ActiveQueryTracker.Enabled      | true                 | true                             |
| Emailing          | Enabled                         | false                | false                            |
| Emailing.SMTP.TLS | Enabled                         | false                | false                            |
| Emailing.SMTP.TLS | InsecureSkipVerify              | false                | false                            |
| StatsReporter     | Enabled                         | true                 | true                             |
| StatsReporter     | Collect.Identities              | true                 | true                             |

The only merger that can change these values happens 
https://github.com/SigNoz/signoz/blob/7b1ca9a1a6273e576d517f5762effcb3994c6398/pkg/signoz/config.go#L161

Where as `smtp_require_tls` value changes because `Alertmanager.Global` though a **second upstream merge** with  Prometheus Alertmanager’s default global config. 

These `bools` are being respected and don't need a fix like `smtp_require_tls`, if I am missing out on anything here please let me know.

pkalsi97:
@therealpandey Hi, can you please check in on this when you have availability.

pkg/types/alertmanagertypes/config.go
```diff
@@ -93,10 +93,12 @@ func NewConfigFromStoreableConfig(sc *StoreableConfig) (*Config, error) {
 }
 
 func NewDefaultConfig(globalConfig GlobalConfig, routeConfig RouteConfig, orgID string) (*Config, error) {
+	smtpRequireTls := globalConfig.SMTPRequireTLS
```
therealpandey:
```suggestion
	smtpRequireTLS := globalConfig.SMTPRequireTLS
```
Also, in one line can you add a code comment explaining why this is necessary? It's helpful for the next person to understand why this has been done

pkalsi97:
@therealpandey Added one line explaining the Why.

pkg/types/alertmanagertypes/config.go
therealpandey: Can you also address the variable casing? @pkalsi97

pkalsi97: @therealpandey  Done!

PR #9477, opened 2025-11-03
feat: update styles for percentile value

YounixM:
Minor Span percentile UI updates

Before:
Screenshot 2025-11-03 at 21 51 13

After:
Screenshot 2025-11-03 at 21 49 31

### DIFF
--- a/frontend/src/container/SpanDetailsDrawer/SpanDetailsDrawer.tsx
+++ b/frontend/src/container/SpanDetailsDrawer/SpanDetailsDrawer.tsx
@@ -388,17 +388,24 @@ function SpanDetailsDrawer(props: ISpanDetailsDrawerProps): JSX.Element {
 				setShouldUpdateUserPreference(false);
 			}
 		},
+		keepPreviousData: false,
+		cacheTime: 0, // no cache
 	});
 
 	// Prod Req - Wait for 2 seconds before fetching span percentile data on initial load
 	useEffect(() => {
+		setSpanPercentileData(null);
+		setIsSpanPercentilesOpen(false);
 		setInitialWaitCompleted(false);
 
 		const timer = setTimeout(() => {
 			setInitialWaitCompleted(true);
 		}, 2000); // 2-second delay
 
-		return (): void => clearTimeout(timer); // Cleanup on re-run or unmount
+		return (): void => {
+			// clean the old state around span percentile data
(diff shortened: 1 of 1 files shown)


SagarRajput-7:
I see the loader not properly handled in the span details drawer - (the whole drawer, on initial load)


https://github.com/user-attachments/assets/5db9dfc4-2131-4cb1-8f8d-529f4b731398

frontend/src/container/SpanDetailsDrawer/SpanDetailsDrawer.tsx
```diff
@@ -538,6 +539,16 @@ function SpanDetailsDrawer(props: ISpanDetailsDrawerProps): JSX.Element {
 		initialWaitCompleted,
 	]);
 
+	const loadingSpanPercentilesData = useMemo(
+		() => isLoadingSpanPercentilesData || isFetchingSpanPercentilesData,
+		[isLoadingSpanPercentilesData, isFetchingSpanPercentilesData],
+	);
+
+	const spanPercentileValue = useMemo(
```
ahrefabhi: Why is useMemo needed here?

YounixM:
@SagarRajput-7 - drawer loading is not the scope of the PR.

frontend/src/container/SpanDetailsDrawer/SpanDetailsDrawer.tsx
```diff
@@ -538,6 +539,16 @@ function SpanDetailsDrawer(props: ISpanDetailsDrawerProps): JSX.Element {
 		initialWaitCompleted,
 	]);
 
+	const loadingSpanPercentilesData = useMemo(
```
ahrefabhi: same, why is useMemo needed?

ahrefabhi: Let's not over optimise, this is an anti-pattern.

ahrefabhi (changes requested):

frontend/src/container/SpanDetailsDrawer/SpanDetailsDrawer.tsx
YounixM: No, useMemo wouldn't be helpful in both scenarios. Removing them

Issue #9491, opened 2025-11-05
Table panel non number keys as NaN


srikanthccv:
<img width="1504" height="130" alt="Image" src="https://github.com/user-attachments/assets/edcee089-db01-4d0e-8986-c696a5439519" />

PR #9492, opened 2025-11-05
fix: do not format for non aggregation columns

srikanthccv:
## 📄 Summary

Fixes https://github.com/SigNoz/signoz/issues/9491

The change https://github.com/SigNoz/signoz/pull/9054 introduced a regression where all values of the table data get formatter, as a result, the string values from group by get formatted as `NaN`. We should only format the aggregation columns. 

We need to recognise which should be formatted and which shouldn't be. Before this, we would return `undefined` for the columns where the explicit unit is not set (this includes when the unit is an empty string `{A: ''}`). In this commit, we update the `getColumnUnit` to return `undefined` only when the column key doesn't exist. If the column key exists, it represents an agg column, so we either return the unit defined by the user or `none`. The `getYAxisFormattedValue` is called only when the unit is not `undefined`. This effectively helps to distinguish which values of RowData to format and which should be regular text.

---

> [!NOTE]
> Formats table values only for columns with a defined unit and updates unit resolution to return "none" for empty units and undefined when absent.
> 
> - **Table formatting**:
>   - Only format cells when `getColumnUnit` returns a unit; stop formatting non-aggregation (no-unit) columns in `GridTableComponent`.
>   - Preserve raw numeric value in `_<column>_without_unit` for sorting/thresholds.
> - **Unit resolution** (`getColumnUnit`):
>   - Return `"none"` when a column key exists with empty string unit; return `undefined` only when the key is absent.
>   - Apply same logic for fallback from `queryName.expression` to `queryName`.
>   - Updated JSDoc to reflect new behavior.
> - **Tests/config**:
>   - Add `decimalPrecision: 2` to table widget test helper.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 254d6007d2a3e92524128722d085f59f7c569f0a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/src/container/GridTableComponent/index.tsx
+++ b/frontend/src/container/GridTableComponent/index.tsx
@@ -88,17 +88,13 @@ function GridTableComponent({
 					const newValue = { ...val };
 					Object.keys(val).forEach((k) => {
 						const unit = getColumnUnit(k, columnUnits);
-						// Apply formatting if:
-						// 1. Column has a unit defined, OR
-						// 2. decimalPrecision is specified (format all values)
-						const shouldFormat = unit || decimalPrecision !== undefined;
 
-						if (shouldFormat) {
+						if (unit) {
 							// the check below takes care of not adding units for rows that have n/a or null values
 							if (val[k] !== 'n/a' && val[k] !== null) {
 								newValue[k] = getYAxisFormattedValue(
 									String(val[k]),
-									unit || 'none',
+									unit,
 									decimalPrecision,
 								);
 							} else if (val[k] === null) {
--- a/frontend/src/lib/query/createTableColumnsFromQuery.ts
+++ b/frontend/src/lib/query/createTableColumnsFromQuery.ts
@@ -662,21 +662,23 @@ const generateTableColumns = (
  *
  * @param columnKey - The column identifier (could be queryName.expression or queryName)
  * @param columnUnits - The column units mapping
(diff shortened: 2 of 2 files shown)


srikanthccv:
I pushed a first commit that shows the failure of snapshot with the fix commit (second one) https://github.com/SigNoz/signoz/actions/runs/19102286673/job/54576990679

frontend/src/container/GridTableComponent/index.tsx
```diff
@@ -88,17 +88,13 @@ function GridTableComponent({
...
 
-						if (shouldFormat) {
+						if (unit) {
 							// the check below takes care of not adding units for rows that have n/a or null values
 							if (val[k] !== 'n/a' && val[k] !== null) {
 								newValue[k] = getYAxisFormattedValue(
 									String(val[k]),
-									unit || 'none',
+									unit,
 									decimalPrecision,
```
ahrefabhi, line 98: is decimalPrecision being undefined handled in getYAxisFormattedValue??

srikanthccv: Yes, it's handed as we default to `2` https://github.com/SigNoz/signoz/blob/1158e1199bc904bdf3a3ad210f99690d61d8c910/frontend/src/components/Graph/yAxisConfig.ts#L95

PR #9537, opened 2025-11-10
fix: removed the decimal places logic from getYAxisFormattedValue

SagarRajput-7:
## 📄 Summary

As the `DEFAULT_SIGNIFICANT_DIGITS` value is now increased to 15 digits, showing more than it and handling for full precision and decimalPlaces is not required and not ideal for real cases.

---

## ✅ Changes

---

ex:

- `frontend`
- `backend`
- `devops`
- `bug`
- `enhancement`
- `ui`
- `test`

---

## 👥 Reviewers

- frontend / backend / devops

---

## 🧪 How to Test

1. ...
2. ...
3. ...

---

## 🔍 Related Issues

- https://github.com/SigNoz/engineering-pod/issues/3252

---

---

## 📋 Checklist

- [ ] Dev Review

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> Simplifies FULL precision to always use default significant digits and adjusts unit/percent rounding, updating tests accordingly.
> 
> - **Frontend › Graph**:
>   - **Precision handling**: `getYAxisFormattedValue` now uses `DEFAULT_SIGNIFICANT_DIGITS` for `FULL` precision, removing previous `decimalPlaces` logic.
>   - **Custom format `none`**: Retains custom decimal formatter; applied after Grafana formatting when decimals present.
> - **Tests** (`frontend/src/components/Graph/__tests__/yAxisConfig.test.ts`):
>   - Adjust percent rounding expectation for `1.00555555559595876` to `1.005555555595959%`.
>   - Corrects `s` format with very small values to display `26.254299141484417 µs` instead of an overly large microseconds value.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4b033df252c4653e2fdb74decf7c1768c9f4cc9c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


frontend/src/components/Graph/__tests__/yAxisConfig.test.ts
```diff
@@ -369,3 +369,23 @@ describe('getYAxisFormattedValue - precision option tests', () => {
 		);
 	});
 });
+
+describe('getYAxisFormattedValue - edge case test', () => {
+	test('with number prop (-ve) testing', () => {
+		expect(getYAxisFormattedValue(12345.6789 as any, 'none', 3)).toBe(
```
SagarRajput-7: -ve testing purpose fully added unsupported type as input

frontend/src/components/Graph/yAxisConfig.ts
SagarRajput-7, line 97: parseFloat() won't break if a number is passed. It accepts both strings and numbers; if given a number, it converts it to a string first, then parses it.
However, the function signature is value: string, so:
- TypeScript will error if a number is passed directly
- At runtime, parseFloat(123) works (converts to "123" then parses)

cc: @aks07

frontend/src/components/Graph/yAxisConfig.ts
```diff
@@ -111,9 +109,7 @@ export const getYAxisFormattedValue = (
 	// For all other standard formats, delegate to grafana/data's built-in formatter.
 	const computeDecimals = (): number | undefined => {
 		if (precision === PrecisionOptionsEnum.FULL) {
-			return decimalPlaces && decimalPlaces >= DEFAULT_SIGNIFICANT_DIGITS
-				? decimalPlaces
```
SagarRajput-7, line 115: As the `DEFAULT_SIGNIFICANT_DIGITS` value is now increased to 15 digits, showing more than it and handling for full precision and decimalPlaces is not required and not ideal for real cases.

frontend/src/components/Graph/yAxisConfig.ts
SagarRajput-7, line 128: we should not be extending the try-catch block here, as try...catch block should have the smallest scope possible to protect only the volatile or external code. The initial calculation (parseFloat, isNaN, Infinity checks, and the format === 'none' call) are pure, simple, and internal logic.

frontend/src/components/Graph/yAxisConfig.ts
```diff
@@ -111,9 +109,7 @@ export const getYAxisFormattedValue = (
 	// For all other standard formats, delegate to grafana/data's built-in formatter.
```
aks07, line 104: Wrap the whole logic in try catch as it was before these changes.

SagarRajput-7: added

SagarRajput-7:
> > The split function, when consumed incorrectly by the consumer, violating the props type, can lead to the function breaking. This is not a bug in the function but rather an edge case handling that, if a consumer of this function incorrectly sends the value as a number, then it should not break.
> 
> Don't write and encourage this practice. It's a bottomless pit. How long are we going to keep adding the endless fallbacks. What's the point of typescript if the consumer can't get error on incorrect type?

I have not added fallback here, this is fix PR - https://github.com/SigNoz/signoz/pull/9539, here the consumer missed to have the values as strings, 

I agree and understand your point, hence we have not added other type support on the Typescript like, number etc. the prop is still a string type. In this particular PR, during the debugging we found that decimal places is something that we do not need now, as the Significant digit is 15, which is large enough, hence removed that, extended the try-catch and added test cases.

PR #9543, opened 2025-11-10
chore: metric name and group by extractor with CH and PromQL support

abhishekhugetech:
## 📄 Summary

We’re creating an interface called `QueryFilterExtractor` and the job for this interface would be exact information like metrics being filtered and table grouping in the provided query. 

[Check examples here](https://github.com/SigNoz/signoz/blob/4993a44ecc1eed1928d0d5a1611032d1d996a067/pkg/parser/queryfilterextractor/clickhouse_test.go)

---

## ✅ Changes

- [x] Feature: parsing query and extract useful info

---

ex:

- `frontend`
- `backend`
- `devops`
- `bug`
- `enhancement`
- `ui`
- `test`

---

## 👥 Reviewers

- frontend / backend / devops

---

## 🧪 How to Test

1. ...
2. ...
3. ...

---

## 🔍 Related Issues

Closes #

---

---

## 📋 Checklist

- [x] Dev Review

---


pkg/parser/queryfilterextractor/clickhouse_test.go
```diff
@@ -0,0 +1,954 @@
...
+		    ` + "`host_name`" + `,
+		    ts
+		ORDER BY
+		    ` + "`os.type`" + ` DESC,
+		    state ASC,
+		    ` + "`host_name`" + ` ASC,
+		    ts ASC
+					`,
+			wantMetrics: []string{"system.memory.usage"},
+			wantGroupBy: []string{"os.type", "state", "host_name", "ts"},
```
srikanthccv: This is correct, but to be able to exclude, we should know the original attributes right? otherwise, we don't know when the corresponding group initially ingested. For example, here `host_name` is the alias and the result label, but the lack of context about the original attribute `host.name` means we can't search for the metadata with the correct attribute name.

srikanthccv:
@abhishekhugetech , let me know when this is ready for re-review again.

pkg/parser/queryfilterextractor/clickhouse.go
```diff
@@ -0,0 +1,695 @@
...
+
+// ========================================
+// Metric Name Extraction
+// ========================================
+
+// fillMetricNamesFromExpr extracts metric names from various node types
+func (e *ClickHouseFilterExtractor) fillMetricNamesFromExpr(node clickhouse.Expr, metricNames map[string]bool) {
+	if node == nil {
+		return
+	}
```
srikanthccv: When and why would this node be nil?

pkg/parser/queryfilterextractor/clickhouse.go
```diff
@@ -0,0 +1,695 @@
...
+//
+// Unsupported operators (can be added later if needed):
+//   - "!=", "<>", "NOT IN": Negative filters. (e.g., metric_name != 'a')
+//   - "LIKE", "ILIKE": Pattern matching filters
+//   - "NOT LIKE", "NOT ILIKE": Negative pattern matching filters
+//   - "OR", "AND": Boolean operators as the Walk function will automatically traverse both sides
+//     of OR/AND operations and extract metrics from each branch. (e.g., metric_name='a' OR metric_name='b')
+func (e *ClickHouseFilterExtractor) fillMetricWithBinaryOpConditions(op *clickhouse.BinaryOperation, valueExpr clickhouse.Expr, metricNames map[string]bool) {
+	switch op.Operation {
+	case "=", "==":
```
srikanthccv: Use token kind consts, i.e `TokenKindSingleEQ` and `TokenKindDoubleEQ`

pkg/parser/queryfilterextractor/clickhouse.go
```diff
@@ -0,0 +1,695 @@
...
+		// metric_name = 'value' or metric_name = any(['a', 'b'])
+		// Skip if value side is a function call (per spec - function-wrapped literals are ignored, CH59)
```
srikanthccv: Can you help me understand the what is the spec you refer to here and what is the CH59 here?

pkg/parser/queryfilterextractor/clickhouse.go
```diff
@@ -0,0 +1,695 @@
...
+// ========================================
+// CTE and Subquery Extraction
+// ========================================
+
+// buildCTEMap builds a map of CTE names to their SelectQuery nodes by recursively
+// traversing all queries and their nested expressions
+func (e *ClickHouseFilterExtractor) buildCTEMap(query *clickhouse.SelectQuery, cteMap map[string]*clickhouse.SelectQuery) {
+	if query == nil {
+		return
+	}
```
srikanthccv: Same question here, if the `query` can be a nil, when would that happen?

pkg/parser/queryfilterextractor/clickhouse.go
```diff
@@ -0,0 +1,695 @@
...
+
+	return nil
+}
+
+// buildCTEMapFromExpr recursively extracts CTEs from various expression types
+func (e *ClickHouseFilterExtractor) buildCTEMapFromExpr(expr clickhouse.Expr, cteMap map[string]*clickhouse.SelectQuery) {
+	if expr == nil {
+		return
+	}
```
srikanthccv: I see this pattern throughout. I want to understand when would this happen, if it can happen, why is even the caller calling it with nil?

pkg/parser/queryfilterextractor/promql.go
```diff
@@ -0,0 +1,129 @@
...
+		return v, nil
+	case *parser.SubqueryExpr:
+		// Subqueries may contain VectorSelectors
+		return v, nil
+	case *parser.ParenExpr:
+		// Parentheses don't change semantics, continue traversal
+		return v, nil
+	case *parser.MatrixSelector:
+		// Matrix selectors wrap VectorSelectors
+		return v, nil
```
srikanthccv: So I see some node types handled as pass-through here, but at the bottom we have `return v, nil`, so what's the point of these node types handling in the switch case, because even if they don't exist, we just return the `v`? 

Or is it that we should be returning an error or something?

pkg/parser/queryfilterextractor/promql.go
```diff
@@ -0,0 +1,129 @@
...
+// in the case we'll be extracting the grouping keys from the outermost aggregation.
+func (v *promQLVisitor) visitAggregateExpr(ae *parser.AggregateExpr, path []parser.Node) {
+	// Count how many AggregateExpr nodes are in the path (excluding current node)
+	// This tells us the nesting level
+	nestingLevel := 0
+	for _, p := range path {
+		if _, ok := p.(*parser.AggregateExpr); ok {
+			nestingLevel++
+		}
+	}
```
srikanthccv, line 101: Just to follow, can you help me understand how the `path` is populated and what makes it reliable to decide the nesting level?

pkg/parser/queryfilterextractor/clickhouse_originparser.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	// If query has multiple select items, return blank string as we don't expect multiple select items
+	if len(selectStmt.SelectItems) > 1 {
+		return "", nil
+	}
+
+	if len(selectStmt.SelectItems) == 0 {
+		return "", errors.New(errors.TypeInvalidInput, errors.CodeInvalidInput, "SELECT query has no select items")
+	}
```
srikanthccv, line 117: All of the cases handled here, should ideally never occur right? as we are internally preparing a simple `SELECT expr` ourselves and invoking this `extractCHOriginFieldFromQuery`

srikanthccv:
The query parsing part is bit dense with logical flow. I will take time separately to follow it but mainly following the tests for correctness.

PR #9569, opened 2025-11-13
chore: converted querysearch codemirror component to uncontrolled component

ahrefabhi:
# Refactor: Convert QuerySearch CodeMirror Component to Uncontrolled

## 📄 Summary

This PR refactors the `QuerySearch` component to use an uncontrolled CodeMirror editor pattern, eliminating local query state management and fixing cursor jumping issues that occurred when the component was controlled.

### Refactoring
- **Removed local query state**: Eliminated redundant local state tracking of the query value
- **Uncontrolled CodeMirror**: CodeMirror now manages its own internal state via the editor ref instead of being controlled by React state

### DIFF
--- a/pkg/query-service/app/cloudintegrations/controller.go
+++ b/pkg/query-service/app/cloudintegrations/controller.go
@@ -116,7 +116,7 @@ func (c *Controller) GenerateConnectionUrl(ctx context.Context, orgId string, cl
 		return nil, model.WrapApiError(apiErr, "couldn't upsert cloud account")
 	}
 
-	agentVersion := "v0.0.6"
+	agentVersion := "v0.0.7"
 	if req.AgentConfig.Version != "" {
 		agentVersion = req.AgentConfig.Version
 	}


ahrefabhi:
> Examples where it breaks with other behaviour changes
> 
> * add a filter in list view page of traces service.name = 'frontend' and then switch to time series view
> * use the "Share" button we show on the top right of the page and use "Copy page link" to get the sharable URL and open that URL.
> 
> They should be caught is dev testing and we should write tests to ensure such problems are caught if we don't have tests already for them.



The above issues are fixed, added tests as well

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
```diff
@@ -79,6 +80,16 @@ const stopEventsExtension = EditorView.domEventHandlers({
 	},
 });
 
+interface QuerySearchProps {
+	placeholder?: string;
+	onChange: (value: string, syncExpression?: boolean) => void;
```
srikanthccv: Why do we have `syncExpression` still here?

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
```diff
@@ -119,29 +125,67 @@ function QuerySearch({
...
+
+			const newQuery = queryData.filter?.expression || '';
+			const currentQuery = getCurrentQuery();
+
+			/* eslint-disable-next-line sonarjs/no-collapsible-if */
+			if (newQuery !== currentQuery && !isFocused) {
+				// Prevent clearing a non-empty editor when queryData becomes empty temporarily
+				// Only update if newQuery has a value, or if both are empty (initial state)
+				if (newQuery || !currentQuery) {
+					updateEditorValue(newQuery, { skipOnChange: true });
```
srikanthccv, line 178: Is my understanding correct that we do this because the editor is getting destroyed, so we create a new editor with the  `newQuery` value, because `currentQuery` would be empty from the editor view state?

frontend/src/components/QueryBuilderV2/QueryV2/__tests__/QuerySearch.test.tsx
srikanthccv, line 1: I didn't find one. Can you please add a test for switching b/w tabs and ensure the search expression is correctly present? i.e., if the issue mentioned were to happen again for any reason, how would we know from the test https://github.com/SigNoz/signoz/pull/9569#pullrequestreview-3476697767?

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
ahrefabhi: removed

frontend/src/components/QueryBuilderV2/QueryV2/__tests__/QuerySearch.test.tsx
ahrefabhi: in both the cases we need to check if the querysearch component is getting initialised with the current query expression on mount of not, I've added a test for that and that checks for these cases.

TestName: `initializes CodeMirror with expression from queryData.filter.expression on mount`

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
ahrefabhi: this is to sync with the editor when query changes from external source like quick filter.

frontend/src/components/QueryBuilderV2/QueryV2/__tests__/QuerySearch.test.tsx
srikanthccv: I was talking about test case that exactly does the user interaction of switching the tabs. I created this https://github.com/SigNoz/signoz/issues/9647. Please add such cases. The on mount tests is no enough.

PR #9589, opened 2025-11-16
chore: fix undefined labels error in alerts

amlannandy:
## 📄 Summary

[Related Sentry Issue](https://signoz-io.sentry.io/issues/6952008637/?alert_rule_id=15272042&alert_type=issue&project=4506831143763968&referrer=slack)

- When creating alerts via API, it's possible to not pass the labels fields, so it can be undefined.
- But the frontend doesn't handle it completely. We handled it in the `AlertDef` interface earlier, but for triggered alerts, there is a separate interface which still had labels as always defined.
- Marking the `labels` field in it as having the potential to be undefined, and then handle typescript errors wherever found.

---

## 🔍 Related Issues

Closes [#3099](https://github.com/SigNoz/engineering-pod/issues/3099) 

---

https://github.com/user-attachments/assets/104915b7-0c2e-4cc7-a9ce-58034750109d

---

## 📋 Checklist

- [x] Dev Review


AshwinBhatkal:
1. Can we use the same interface at both the places? AlertDef and current usage
2. Is the passing of undefined labels via API intentional or a broken behaviour?

amlannandy:
> 1. Can we use the same interface at both the places? AlertDef and current usage
> 2. Is the passing of undefined labels via API intentional or a broken behaviour?
 
1. There are many different fields between them. So I would say no.
2. it's intentional. Basically, when user is creating an alert via API, `labels` is an optional field. We missed that while writing the interface for frontend usage.

PR #9593, opened 2025-11-17
feat(JSON): JSON Body Metadata

piyushsingariya:
## 📄 Summary

These changes includes some basic utilities that will be used in Body JSON Query Builder

---

## ✅ Changes

- [x] Chore: JSON Body utils

---

## 🔍 Related Issues

Closes #

---

## 📋 Checklist

- [ ] Dev Review

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> Introduces body JSON metadata (paths, values, indexes, promotions) and enables body-context JSON querying across logs, with refactored builders/types, new tests, and a collector dependency bump.
> 
> - **Body JSON Metadata (new)**:
>   - Add `telemetrymetadata`: list body JSON paths, values, indexes; detect promoted paths; shared table constants.
> - **Query & Condition Building**:
>   - Support `FieldContextBody` and `body.` prefix in filters, functions, and aggregations; remove `JsonBodyPrefix` plumbed arg.
>   - Update logs/traces condition builders and field mappers to new schema type enums and body JSON exists/array handling.
>   - Expose `BODY_JSON_QUERY_ENABLED` flag; tweak `PreparedWhereClause` fields; improve LIKE warnings.
> - **Types**:
>   - Add JSON data type model/mappings (`JSONDataType`, indexes) and extend `FieldDataType`; add `FieldContextBody` and `BodyJSONStringSearchPrefix`.
> - **Logs Query Pipeline**:
>   - Enhance statement builder to pass body JSON key resolution; keep full-text and resource filter integration.
> - **Tests**:
>   - Extensive updates/additions for body JSON queries, warnings, and condition building.
> - **Errors/Deps**:
>   - Add timeout error helpers; bump `signoz-otel-collector` to `v0.129.10-rc.9`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1bb60bbf082f4c96c340d91f4e158b84296388bd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/telemetrylogs/condition_builder.go
+++ b/pkg/telemetrylogs/condition_builder.go
@@ -182,8 +182,8 @@ func (c *conditionBuilder) conditionFor(
 				return sb.IsNull(tblFieldName), nil
 			}
 		case schema.ColumnTypeEnumLowCardinality:
-			switch elementType := column.Type.(schema.LowCardinalityColumnType).ElementType; elementType {
-			case schema.ColumnTypeString:
+			switch elementType := column.Type.(schema.LowCardinalityColumnType).ElementType; elementType.GetType() {
+			case schema.ColumnTypeEnumString:
 				value = ""
 				if operator == qbtypes.FilterOperatorExists {
 					return sb.NE(tblFieldName, value), nil
--- a/pkg/telemetrytraces/condition_builder.go
+++ b/pkg/telemetrytraces/condition_builder.go
@@ -182,8 +182,8 @@ func (c *conditionBuilder) conditionFor(
 				return sb.E(tblFieldName, value), nil
 			}
 		case schema.ColumnTypeEnumLowCardinality:
-			switch elementType := column.Type.(schema.LowCardinalityColumnType).ElementType; elementType {
-			case schema.ColumnTypeString:
+			switch elementType := column.Type.(schema.LowCardinalityColumnType).ElementType; elementType.GetType() {
+			case schema.ColumnTypeEnumString:
 				value = ""
 				if operator == qbtypes.FilterOperatorExists {
 					return sb.NE(tblFieldName, value), nil
--- a/pkg/telemetrylogs/field_mapper.go
+++ b/pkg/telemetrylogs/field_mapper.go
@@ -125,8 +125,8 @@ func (m *fieldMapper) FieldFor(ctx context.Context, key *telemetrytypes.Telemetr
 			return fmt.Sprintf("multiIf(%s.`%s` IS NOT NULL, %s.`%s`::String, mapContains(%s, '%s'), %s, NULL)", column.Name, key.Name, column.Name, key.Name, oldColumn.Name, key.Name, oldKeyName), nil
 		}
 	case schema.ColumnTypeEnumLowCardinality:
-		switch elementType := column.Type.(schema.LowCardinalityColumnType).ElementType; elementType {
-		case schema.ColumnTypeString:
+		switch elementType := column.Type.(schema.LowCardinalityColumnType).ElementType; elementType.GetType() {
+		case schema.ColumnTypeEnumString:
 			return column.Name, nil
 		default:
 			return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "exists operator is not supported for low cardinality column type %s", elementType)
@@ -140,8 +140,8 @@ func (m *fieldMapper) FieldFor(ctx context.Context, key *telemetrytypes.Telemetr
 			return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "key type %s is not supported for map column type %s", keyType, column.Type)
 		}
 
-		switch valueType := column.Type.(schema.MapColumnType).ValueType; valueType {
-		case schema.ColumnTypeString, schema.ColumnTypeBool, schema.ColumnTypeFloat64:
+		switch valueType := column.Type.(schema.MapColumnType).ValueType; valueType.GetType() {
+		case schema.ColumnTypeEnumString, schema.ColumnTypeEnumBool, schema.ColumnTypeEnumFloat64:
 			// a key could have been materialized, if so return the materialized column name
 			if key.Materialized {
 				return telemetrytypes.FieldKeyToMaterializedColumnName(key), nil


pkg/types/telemetrytypes/json_datatype.go
```diff
@@ -0,0 +1,40 @@
...
+	return jdt.str
+}
+
+var (
+	String       = JSONDataType{"String", false, ""}
+	Int64        = JSONDataType{"Int64", false, ""}
+	Float64      = JSONDataType{"Float64", false, ""}
+	Bool         = JSONDataType{"Bool", false, ""}
+	Dynamic      = JSONDataType{"Dynamic", false, ""}
+	ArrayString  = JSONDataType{"Array(Nullable(String))", true, "String"}
```
nityanandagohain: ArrayNullableString?

pkg/telemetrytraces/condition_builder.go
```diff
@@ -162,13 +162,15 @@ func (c *conditionBuilder) conditionFor(
 	case qbtypes.FilterOperatorExists, qbtypes.FilterOperatorNotExists:
 
 		var value any
-		switch column.Type {
-		case schema.JSONColumnType{}:
+		// schema.JSONColumnType{} now can not be used in switch cases, so we need to check if the column is a JSON column
```
nityanandagohain: we can keep the existing switch case ( which is much cleaner) if we change the struct to something like

```
type JSONColumnType struct {
	MaxDynamicPaths *uint
	MaxDynamicTypes *uint
	PromotedPaths   *PromotedPaths
}
type PromotedPaths []Column
```

Do you see any issues with this ?

pkg/telemetrylogs/body_json_metadata.go
```diff
@@ -0,0 +1,171 @@
+package telemetrylogs
```
nityanandagohain: filename can be `json_body_metadata.go`

pkg/telemetrylogs/body_json_keys.go
```diff
@@ -0,0 +1,171 @@
...
+// ExtractBodyPaths extracts body JSON paths from the path_types table
+// This function can be used by both JSONQueryBuilder and metadata extraction
+// uniquePathLimit: 0 for no limit, >0 for maximum number of unique paths to return
+//   - For startup load: set to 10000 to get top 10k unique paths
+//   - For lookup: set to 0 (no limit needed for single path)
+//   - For metadata API: set to desired pagination limit
+//
+// searchOperator: LIKE for pattern matching, EQUAL for exact match
+// Returns: (paths, error)
+func ExtractBodyPaths(ctx context.Context, telemetryStore telemetrystore.TelemetryStore,
```
nityanandagohain: `GetJsonBodyPaths`

pkg/telemetrylogs/body_json_keys.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	if rows.Err() != nil {
+		return nil, errors.Wrap(rows.Err(), errors.TypeInternal, errors.CodeInternal, "error iterating body JSON keys")
+	}
+
+	return paths, nil
+}
+
+func buildExtractBodyPathsQuery(searchTexts []string, uniquePathLimit int, operator qbtypes.FilterOperator) (string, []any) {
```
nityanandagohain: buildGetJsonPathsQuery will be a better name ?

pkg/telemetrylogs/body_json_keys.go
nityanandagohain: also uniquePathLimit should be just limit ?

pkg/telemetrylogs/body_json_metadata.go
```diff
@@ -0,0 +1,171 @@
...
+		var path string
+		var typesArray []string // ClickHouse returns array as []string
+		var lastSeen uint64
+
+		err = rows.Scan(&path, &typesArray, &lastSeen)
+		if err != nil {
+			return nil, errors.Wrap(err, errors.TypeInternal, errors.CodeInternal, "failed to scan body JSON key row")
+		}
+
+		set := utils.NewConcurrentSet[telemetrytypes.JSONDataType]()
```
nityanandagohain: are concurrent sets really required ? these feels unnecessary allocation of resource.

pkg/telemetrylogs/body_json_metadata.go
```diff
@@ -0,0 +1,171 @@
...
+		"path",
+		"groupArray(DISTINCT type) AS types",
+		"max(last_seen) AS last_seen",
+	).From(from)
+
+	// Add search filter if provided
+	if len(searchTexts) > 0 {
+		orClauses := []string{}
+		for _, searchText := range searchTexts {
+			if operator == qbtypes.FilterOperatorEqual {
```
nityanandagohain: will this just be confined to equal or !=, in, nin, >, < , >=, <=

pkg/telemetrylogs/body_json_keys_test.go
```diff
@@ -0,0 +1,264 @@
...
+			uniquePathLimit: 0,
+			operator:        qbtypes.FilterOperatorLike,
+			expectedSQL: []string{
+				"SELECT",
+				"path",
+				"groupArray(DISTINCT type) AS types",
+				"max(last_seen) AS last_seen",
+				fmt.Sprintf("FROM %s.%s", DBName, PathTypesTableName),
+				"GROUP BY path",
+				"ORDER BY max(last_seen) DESC",
```
nityanandagohain:
```
SELECT
    path,
    groupArrayDistinct(type) AS type,
    max(last_seen) AS last_seen
FROM path_types
GROUP BY path
ORDER BY max(last_seen) DESC

Query id: 2b5887c1-651b-4f86-87c0-601a641433f7


Elapsed: 0.005 sec.

Received exception from server (version 25.10.1):
Code: 184. DB::Exception: Received from localhost:9000. DB::Exception: Aggregate function max(last_seen) AS last_seen is found inside another aggregate function in query. (ILLEGAL_AGGREGATION)
```

correct query is 
```
SELECT
    path,
    groupArrayDistinct(type) AS type,
    max(last_seen) AS last_seen
FROM path_types
GROUP BY path
ORDER BY last_seen DESC
```

pkg/telemetrylogs/tables.go
```diff
@@ -8,4 +8,7 @@ const (
 	TagAttributesV2LocalTableName = "tag_attributes_v2"
 	LogAttributeKeysTblName       = "distributed_logs_attribute_keys"
 	LogResourceKeysTblName        = "distributed_logs_resource_keys"
+	PathTypesTableName            = "distributed_path_types"
```
nityanandagohain: names can be better

`distributed_json_body_paths`
`distributed_json_body_promoted_paths`

srikanthccv:
> Also this feels like the body_json_keys should be a part of metadata package instead and should also use FieldKeySelector struct

I checked the main PR to get some prior context and I agree. What do you think about keeping `JSONQueryBuilder` stateless; where it takes the relevant context from inputs and returns the condition? As I understand it, it's currently using the ClickHouse conn to build intelligent conditions based on: 1) indexing and 2) promotion.

How about we make `telemetrytypes.TelemetryFieldKey/FieldKeySelector` etc.. aware of this context and have the metadata package annotate the keys returned from `GetKeys`/`GetKeysMulti` with the JSON context? That way, `JSONQueryBuilder` would still have all the metadata needed to prepare the correct and performant query.

The way I view the work you're doing is that it's laying the foundation for all JSON work going forward. As such, it's better to keep `JSONQueryBuilder` stateless and independent of logs (at a high level) and ensure it remains pluggable if we want to introduce JSON support for span attributes. Let me know what you think.

pkg/telemetrylogs/body_json_metadata.go
piyushsingariya: It's just a map

pkg/telemetrylogs/body_json_metadata.go
piyushsingariya: Yes, cuz there're only two usecase for this as of now

piyushsingariya:
> What do you think about keeping `JSONQueryBuilder` stateless;

Can not be stateless



> How about we make `telemetrytypes.TelemetryFieldKey/FieldKeySelector` etc.. aware of this context and have the metadata package annotate the keys returned from `GetKeys`/`GetKeysMulti` with the JSON context? That way, `JSONQueryBuilder` would still have all the metadata needed to prepare the correct and performant query.
> 
> The way I view the work you're doing is that it's laying the foundation for all JSON work going forward. As such, it's better to keep `JSONQueryBuilder` stateless and independent of logs (at a high level) and ensure it remains pluggable if we want to introduce JSON support for span attributes. Let me know what you think.

Will do this later, right now JSONQueryBuilder is acting as a separate branch code that's flag control, it doesn't fit into the current rhythm of QB being stateless, QB would need large changes that I should not touch right now, which I had evaluated at the very start, hence built it in such way

My plan is to first overhaul Logs QB, and then based on chaos there afterwards I'll might touch the main QB

pkg/telemetrylogs/body_json_keys_test.go
piyushsingariya: I had tested it locally I forgot to update

pkg/telemetrylogs/body_json_keys.go
piyushsingariya: Right now there's no such usecase, will change later if required

pkg/telemetrytraces/condition_builder.go
piyushsingariya: Pointer to a Slice? Slice by itself a reference object, would create confusion for folks working with this

pkg/types/telemetrytypes/json_datatype.go
piyushsingariya: ??

srikanthccv:
Please help me understand why can't it be stateless.

pkg/telemetrylogs/tables.go
piyushsingariya: and we will create more tables for attributes later on? These can be central tables for both body and attributes later on?

piyushsingariya:
> Please help me understand why can't it be stateless.

It needs to know types for each paths existing in warehouse to form the actual queries.

e.g. Input value is string but the operator is Contains, existing type is Array of String, the operation would be to look into array of string contains the value give. 

This type of decision making is required in-flight.

pkg/telemetrylogs/tables.go
nityanandagohain: then 

```
distributed_json_paths
distributed_json_promoted_paths
```

pkg/telemetrylogs/body_json_metadata.go
nityanandagohain: By two usecase you mean ?

pkg/telemetrylogs/body_json_keys.go
nityanandagohain: What I meant was `Extract` doesn't seem the right naming, as we are just getting the data from DB

srikanthccv:
That's what we do in the condition builder + field mapper already. Prepare the right query based on the FieldContext + FieldDataType + Materialised info of the each that it gets. Can you please give an example why can't be the same done that it requires a state management.

pkg/types/telemetrytypes/json_datatype.go
nityanandagohain: I mean the naming.

pkg/telemetrytraces/condition_builder.go
nityanandagohain: hmm, I was just thinking of a common way to do this and not introduce a custom one again for JSON.

Will something like `column.Type.(type)` in the switch help in this case ?

piyushsingariya:
> Can you please give an example why can't be the same done that it requires a state management.

I just iterated over one above. Will give one more example.

something.something1[].name

something1 here could be a simple Array(JSON) or an Array(Dynamic) based on the state it will form a query to just look into Array(JSON) if it's Array(Dynamic) it will then filter JSON array out of Dynamic and then dive in for name field, if both it will look both cases

srikanthccv:
I get that, but what I am not able to follow is why there is a need to maintain a state. The relevant context would be available as part of the `TelemetryFieldKey`, and the condition can be built correctly based on the known context.

pkg/telemetrylogs/body_json_metadata.go
piyushsingariya: First for suggestion, and second for immediate lookups for patterns that aren't in cache

pkg/telemetrylogs/body_json_keys.go
piyushsingariya: uniquePathLimit is what I was referring to, name is changed

pkg/types/telemetrytypes/json_datatype.go
piyushsingariya: "Nullable" as keyword wasn't making sense for now atleast cuz JSON will always have Nullable Arrays

pkg/telemetrytraces/condition_builder.go
nityanandagohain: @srikanthccv do share if you have any suggestions here.

go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/ClickHouse/clickhouse-go/v2 v2.40.1
 	github.com/DATA-DOG/go-sqlmock v1.5.2
 	github.com/SigNoz/govaluate v0.0.0-20240203125216-988004ccc7fd
-	github.com/SigNoz/signoz-otel-collector v0.129.4
+	github.com/SigNoz/signoz-otel-collector v0.129.10-rc.4
```
khanbilal732: Done

khanbilal732 (approved):

khanbilal732 (approved):

pkg/query-service/app/logs/v4/query_builder.go
```diff
@@ -488,7 +488,7 @@ func buildLogsLiveTailQuery(mq *v3.BuilderQuery) (string, error) {
 	// the reader will add the timestamp and id filters
 	switch mq.AggregateOperator {
 	case v3.AggregateOperatorNoOp:
-		query := constants.LogsSQLSelectV2 + "from signoz_logs." + DISTRIBUTED_LOGS_V2 + " where "
+		query := constants.LogsSQLSelectV2() + "from signoz_logs." + DISTRIBUTED_LOGS_V2 + " where "
```
srikanthccv: Are we making v4 support the new json. I think we should just keep the support in /v5/query_range to keep the surface area limited. If we change it here and body json query enabled, then we need to make sure all of the /v4/query_range also supports the json querying where it touches?

pkg/query-service/app/logs/v4/query_builder_test.go
```diff
@@ -1210,9 +1211,8 @@ func TestPrepareLogsQuery(t *testing.T) {
 				t.Errorf("PrepareLogsQuery() error = %v, wantErr %v", err, tt.wantErr)
 				return
 			}
-			if got != tt.want {
-				t.Errorf("PrepareLogsQuery() = %v, want %v", got, tt.want)
-			}
+
+			assert.Equal(t, tt.want, got)
```
srikanthccv: Was is just a simple change with no reason or did the `got != tt.want` fail and we had to use `assert.Equal`?

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/SigNoz/signoz/pkg/errors"
 	grammar "github.com/SigNoz/signoz/pkg/parser/grammar"
+	"github.com/SigNoz/signoz/pkg/query-service/constants"
```
srikanthccv: is it possible to define the constant somewhere in `https://github.com/SigNoz/signoz/tree/main/pkg/querybuilder`. We try to avoid importing from `pkg/query-service` in other `pkg/...`

pkg/telemetrymetadata/body_json_metadata.go
```diff
@@ -0,0 +1,437 @@
...
+
+	return cleanIndexes, nil
+}
+
+func buildListLogsJSONIndexesQuery(cluster string, filters ...string) (string, []any) {
+	// Build a better query using GROUP BY to deduplicate at database level
+	// This aggregates all types per path and gets the max last_seen, then applies LIMIT
+	sb := sqlbuilder.Select(
+		"name", "type_full", "expr", "granularity",
+	).From(fmt.Sprintf("clusterAllReplicas('%s', %s)", cluster, SkipIndexTableName))
```
nityanandagohain, line 221: why is this needed, with replicated merge tree it's ensured that it will be same across all. you won't need to depulicate after that.

pkg/telemetrymetadata/body_json_metadata.go
```diff
@@ -0,0 +1,437 @@
...
+	if limit == 0 {
+		limit = defaultPathLimit
+	}
+	sb.Limit(limit)
+
+	query, args := sb.BuildWithFlavor(sqlbuilder.ClickHouse)
+	return query, args, limit, nil
+}
+
+func listLogsIndexesClean(ctx context.Context, conn clickhouse.Conn, cluster, path string) ([]telemetrytypes.JSONDataTypeIndex, error) {
```
nityanandagohain: can be a more meaningful name, here you are getting indexes in context of json.

pkg/telemetrymetadata/body_json_metadata.go
```diff
@@ -0,0 +1,437 @@
...
+		colExpr(telemetrytypes.ArrayInt64),
+		colExpr(telemetrytypes.ArrayFloat64),
+		colExpr(telemetrytypes.ArrayBool),
+		colExpr(telemetrytypes.ArrayDynamic),
+	).From(from)
+	sb.Where(fmt.Sprintf("%s IS NOT NULL", path))
+	sb.Limit(limit)
+
+	query, args := sb.BuildWithFlavor(sqlbuilder.ClickHouse)
+	rows, err := conn.Query(ctx, query, args...)
```
nityanandagohain: add very short timeout, searching something unique for suggestion across a large duration will take time and we should cancel it. Along with resource restrictions.

pkg/telemetrymetadata/tables.go
```diff
@@ -1,7 +1,12 @@
 package telemetrymetadata
 
+import "github.com/SigNoz/signoz-otel-collector/constants"
```
nityanandagohain: nit: import it as otelcollectorconst

pkg/telemetrytraces/condition_builder.go
nityanandagohain: @srikanthccv a reminder if you have a different suggestion here.

pkg/telemetrymetadata/body_json_metadata_test.go
```diff
@@ -0,0 +1,209 @@
...
+					Name:              "user",
+					SelectorMatchType: telemetrytypes.FieldSelectorMatchTypeFuzzy,
+				},
+				{
+					Name:              "admin",
+					SelectorMatchType: telemetrytypes.FieldSelectorMatchTypeFuzzy,
+				},
+			},
+			expectedSQL: []string{
+				"SELECT",
```
nityanandagohain: we can make this better just by looking at this
```
SELECT path groupArray(DISTINCT type) AS types max(last_seen) AS last_seen 
FROM signoz_metadata.distributed_json_path_types 
WHERE OR path = ? GROUP BY path ORDER BY last_seen DESC LIMIT ?
```
but the actual value that should be compared is 
```
SELECT path, groupArray(DISTINCT type) AS types, max(last_seen) AS last_seen 
FROM signoz_metadata.distributed_json_path_types 
WHERE (path = ? OR path = ?) GROUP BY path ORDER BY last_seen DESC LIMIT ?
```

now since your main paint here is to compare if the `Multiple search texts with LIKE operator` the test doesn't fully check it.

nityanandagohain (commented):
please change the expectedSql in metadata_test to be more human redabale and should be concrete enough for giving us more confidence.

I would request you to look at `stmt_builder_test.go`

pkg/telemetrymetadata/body_json_metadata.go
piyushsingariya: Understandable, but this is the right way to query a non distributed view table i.e. SkipIndexTableName

nityanandagohain: don't see any major concern just pointed out. I am okay with this as well

pkg/telemetrymetadata/body_json_metadata.go
piyushsingariya: 5 second timeout?

nityanandagohain: @srikanthccv we discussed something similar for trace operator as well where I thought of adding timeout and resource restrictions, if I apply similar philosophy here that there is not much actionable to do on the user side regardless if it fails, then we should not add timeout.

But  should we allow here as well without restrictions ? since in this case it might scan more data than required and  take up resouce.

Initally we didn't plan for adding adding value suggestion, but since we are going ahead with adding it thought we should decide it here. Let me know your thoughts.

ee/query-service/constants/constants.go
```diff
@@ -9,6 +9,7 @@ var LicenseAPIKey = GetOrDefaultEnv("SIGNOZ_LICENSE_API_KEY", "")
 var SaasSegmentKey = GetOrDefaultEnv("SIGNOZ_SAAS_SEGMENT_KEY", "")
 var FetchFeatures = GetOrDefaultEnv("FETCH_FEATURES", "false")
 var ZeusFeaturesURL = GetOrDefaultEnv("ZEUS_FEATURES_URL", "ZeusFeaturesURL")
+var BodyJSONQueryEnabled = GetOrDefaultEnv("BODY_JSON_QUERY_ENABLED", "false") == "true"
```
srikanthccv: nit: let's move this from here to https://github.com/SigNoz/signoz/tree/main/pkg/querybuilder

pkg/telemetrylogs/condition_builder.go
```diff
@@ -165,13 +166,15 @@ func (c *conditionBuilder) conditionFor(
...
-		case schema.JSONColumnType{}:
+		// schema.JSONColumnType{} now can not be used in switch cases, so we need to check if the column is a JSON column
+		if column.IsJSONColumn() {
 			if operator == qbtypes.FilterOperatorExists {
 				return sb.IsNotNull(tblFieldName), nil
 			} else {
 				return sb.IsNull(tblFieldName), nil
 			}
+		}
+		switch column.Type {
```
srikanthccv: The Type is changed to the point where it doesn't make sense to continue comparison with structs (going to be fragile) and some custom methods. Let's introduce the `Type` on ColumnType interface that returns the bare data type enum without args. What do you think?

Something like 

```
type ColumnTypeEnum string

const (
	TypeString              ColumnTypeEnum = "String"
    ...
	TypeFloat64             ColumnTypeEnum = "Float64"
	TypeBool                ColumnTypeEnum = "Bool"
	TypeIPv6                ColumnTypeEnum = "IPv6"
	TypeJSON                ColumnTypeEnum = "JSON"
    ...
	TypeDateTime64          ColumnTypeEnum = "DateTime64"
	TypeArray               ColumnTypeEnum = "Array"
	TypeMap                 ColumnTypeEnum = "Map"
	TypeTuple               ColumnTypeEnum = "Tuple"
    ...
	TypeEnum                ColumnTypeEnum = "Enum"
	TypeSimpleAggregateFunc ColumnTypeEnum = "SimpleAggregateFunction"
	TypeAggregateFunc       ColumnTypeEnum = "AggregateFunction"
)

type ColumnType interface {
	String() string
	Type() ColumnTypeEnum
}
```

```
switch column.Type.Type() {
case TypeString, TypeUInt64, TypeUInt32, TypeUInt8:
    // handle primitives
case TypeJSON:
    // handle JSON
case TypeMap:
    // handle map
```

pkg/telemetrytraces/condition_builder.go
srikanthccv: I suggest moving away from from comparing primitives or structs with comparable fields in the switch altogether given the how it evolved. Let me know what do you think https://github.com/SigNoz/signoz/pull/9593/files#r2573226459

pkg/telemetrymetadata/body_json_metadata.go
```diff
@@ -0,0 +1,440 @@
...
+		if err := rows.Scan(&path); err != nil {
+			return nil, errors.WrapInternalf(err, CodeFailLoadPromotedPaths, "failed to scan promoted path")
+		}
+		next[path] = struct{}{}
+	}
+
+	return next, nil
+}
+
+func ListJSONValues(ctx context.Context, conn clickhouse.Conn, path string, limit int) (*telemetrytypes.TelemetryFieldValues, bool, error) {
```
srikanthccv, line 288: Continuing the discussion from here https://github.com/SigNoz/signoz/pull/9593#discussion_r2571838129

IIUC, this is for the value suggestions, in that case, I don't think it's an acceptable idea to query the main table for suggestions because it's going to be _really_ slow and won't add any value if the suggestions aren't returned quickly. We have a separate table to keep the keys/values to enable the suggestions and other auto complete use-cases, I believe it should be either no suggestions or if we implement suggestions it should be from the keys/values meta tables. 


See note here https://github.com/SigNoz/engineering-pod/issues/2032#issuecomment-2741113526. Today the `severity_text` suggestions (simple direct field) are implemented on main table and it's slow and unnecessarily expensive. 

To give an idea

```
query:             select distinct severity_text from signoz_logs.logs_v2 where timestamp >= toInt64(toUnixTimestamp(now() - INTERVAL 48 HOUR)*1000000000) limit 50
query_duration_ms: 9541
log_comment:       {"auth_type":"jwt","http_path":"/logs/logs-explorer","module_name":"logs-explorer","org_id":"...","user_id":"..."}
```


This doesn't fall under the same category as the trace operator because there the trace operator query is the main query and if it times out it times and user need to tweak things here on the other hand, it's our product implementing the auto suggestions.

pkg/telemetrymetadata/body_json_metadata.go
```diff
@@ -0,0 +1,440 @@
...
+			})
+		}
+	}
+
+	return cleanIndexes, nil
+}
+
+func buildListLogsJSONIndexesQuery(cluster string, filters ...string) (string, []any) {
+	// Build a better query using GROUP BY to deduplicate at database level
+	// This aggregates all types per path and gets the max last_seen, then applies LIMIT
```
srikanthccv: nit: this comment is not relevant here

pkg/telemetrymetadata/body_json_metadata.go
```diff
@@ -0,0 +1,440 @@
...
+	sb.Where(sb.Or(
+		sb.ILike("expr", fmt.Sprintf("%%%s%%", querybuilder.FormatValueForContains(constants.BodyJSONColumnPrefix))),
+		sb.ILike("expr", fmt.Sprintf("%%%s%%", querybuilder.FormatValueForContains(constants.BodyPromotedColumnPrefix))),
+	))
+
+	filterExprs := []string{}
+	for _, filter := range filters {
+		filterExprs = append(filterExprs, sb.ILike("expr", fmt.Sprintf("%%%s%%", querybuilder.FormatValueForContains(filter))))
+	}
+	sb.Where(sb.Or(filterExprs...))
```
srikanthccv, line 234: question (assuming `filters` input is non-empty): should it be `OR` all `expo ILIKE ...` parts or should it be `OR` b/w (body, promoted) and then `AND` with list of filters like `(express ILIKE ... OR expr ILIKE ...) AND (expr ILIKE ... OR expr ILIKE ...)`

pkg/telemetrymetadata/body_json_metadata.go
```diff
@@ -0,0 +1,440 @@
...
+			Expression:  expr,
+			Granularity: int(granularity),
+		})
+	}
+
+	return indexes, nil
+}
+
+func ListPromotedPaths(ctx context.Context, conn clickhouse.Conn) (map[string]struct{}, error) {
+	query := fmt.Sprintf("SELECT path FROM %s.%s", DBName, PromotedPathsTableName)
```
srikanthccv, line 268: nit: add `FINAL` to clause to avoid reading same path again because the replacing is eventual with table engine. the following is map and this table might be small so it may not matter at all whether `FINAL` is added but good idea anyway.

pkg/telemetrymetadata/body_json_metadata.go
```diff
@@ -0,0 +1,440 @@
...
+		// replace [*] with []
+		fieldKeySelector.Name = strings.ReplaceAll(fieldKeySelector.Name, telemetrylogs.ArrayAnyIndex, telemetrylogs.ArraySep)
+		// Extract search text for body JSON keys
+		keyName := CleanPathPrefixes(fieldKeySelector.Name)
+		if fieldKeySelector.SelectorMatchType == telemetrytypes.FieldSelectorMatchTypeExact {
+			orClauses = append(orClauses, sb.Equal("path", keyName))
+		} else {
+			// Pattern matching for metadata API (defaults to LIKE behavior for other operators)
+			orClauses = append(orClauses, sb.Like("path", querybuilder.FormatValueForContains(keyName)))
+			limit += fieldKeySelector.Limit
```
srikanthccv: Should this be outside else?

pkg/telemetrymetadata/body_json_metadata.go
```diff
@@ -0,0 +1,440 @@
...
+
+		promoted, err := IsPathPromoted(ctx, telemetryStore.ClickhouseDB(), path)
+		if err != nil {
+			return nil, false, err
+		}
+
+		indexes, err := getJSONPathIndexes(ctx, telemetryStore, path)
+		if err != nil {
+			return nil, false, err
+		}
```
srikanthccv: It would be good idea to batch these to one query for all paths and reduce the number of DB roundtrips.

pkg/telemetrytraces/condition_builder.go
nityanandagohain: Yeah makes sense, the better way to add this is create a separete PR from this which moves to the new type comparison get it merged and then update this PR right, to keep changes in this PR minimal.

srikanthccv: Sounds good, but this PR is updating the collector hash to in-progress PR. So please check how do we want to get this merge.

nityanandagohain: Yeah, will get the collector changes merged, by pushing it behind a conifig and update the hash.

pkg/query-service/app/logs/v4/query_builder_test.go
piyushsingariya: assert.Equal shows the output better in terminal for Expected and Actual

pkg/telemetrymetadata/body_json_metadata.go
piyushsingariya: This table is a Merge Tree, afaik FINAL not is supported here??

pkg/query-service/app/logs/v4/query_builder.go
piyushsingariya: It's in v5 only, this constant was imported here, I'll change it

pkg/telemetrymetadata/body_json_metadata.go
srikanthccv: I thought this was ReplacingMergeTree. You can disregard the comment.

pkg/telemetrymetadata/body_json_metadata.go
piyushsingariya: Second one

pkg/telemetrymetadata/body_json_metadata.go
piyushsingariya: What does ` batch these to one query ` means?

srikanthccv: Right now there is a query (two to be precise) to DB for each path. Instead of making N number of database round trips for N paths, Just run one query to fetch the necessary for all paths. https://planetscale.com/blog/what-is-n-1-query-problem-and-how-to-solve-it

Does this make sense?

pkg/telemetrymetadata/body_json_metadata.go
srikanthccv: @nityanandagohain @piyushsingariya ICYMI. I don't think the main table based suggestions is a good idea.

nityanandagohain: Yeah, we didn't intially had it in plan and was added in the last minute. @piyushsingariya lets remove the value suggestion for now as we can pick it up in the next scope/improments that we will look into.

pkg/telemetrymetadata/body_json_metadata.go
piyushsingariya: ohh, "batch" had me confused, I thought I thought I am supposed to fire IsPathPromoted, getJSONPathIndexes in a batch.

pkg/telemetrymetadata/body_json_metadata.go
piyushsingariya: It's not integrated in flow, let's keep the function and reiterate on this later??

khanbilal732 (approved):

khanbilal732 (approved):

pkg/telemetrymetadata/body_json_metadata.go
```diff
@@ -0,0 +1,484 @@
...
+	sb.Where(sb.Or(
+		sb.ILike("expr", fmt.Sprintf("%%%s%%", querybuilder.FormatValueForContains(constants.BodyJSONColumnPrefix))),
+		sb.ILike("expr", fmt.Sprintf("%%%s%%", querybuilder.FormatValueForContains(constants.BodyPromotedColumnPrefix))),
+	))
+
+	filterExprs := []string{}
+	for _, filter := range filters {
+		filterExprs = append(filterExprs, sb.ILike("expr", fmt.Sprintf("%%%s%%", querybuilder.FormatValueForContains(filter))))
+	}
+	sb.Where(sb.Or(filterExprs...))
```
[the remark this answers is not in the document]
piyushsingariya: Select Builder handles this automatically

khanbilal732 (approved):

PR #9605, opened 2025-11-18
feat(authz): add stats reporting for public dashboards

vikrantgupta25:
## Summary 

- add stats reporting for public dashboards

Closes: https://github.com/SigNoz/platform-pod/issues/1326

### DIFF
--- a/pkg/modules/dashboard/impldashboard/store.go
+++ b/pkg/modules/dashboard/impldashboard/store.go
@@ -142,7 +142,6 @@ func (store *store) ListPublic(ctx context.Context, orgID valuer.UUID) ([]*dashb
 		Join("JOIN dashboard").
 		JoinOn("public_dashboard.dashboard_id = dashboard.id").
 		Where("dashboard.org_id = ?", orgID).
-		Where("org_id = ?", orgID).
 		Scan(ctx)
 	if err != nil {
 		return nil, err


pkg/modules/dashboard/impldashboard/module.go
```diff
@@ -58,6 +59,14 @@ func (module *module) Create(ctx context.Context, orgID valuer.UUID, createdBy s
 }
 
 func (module *module) CreatePublic(ctx context.Context, orgID valuer.UUID, publicDashboard *dashboardtypes.PublicDashboard) error {
+	storablePublicDashboard, err := module.store.GetPublic(ctx, publicDashboard.DashboardID.StringValue())
```
vikrantgupta25, line 62: this is needed to avoid unnecessary tuple creations, since create public in DB is last step so if it's created that means everything has been executed successfully

pkg/modules/dashboard/impldashboard/module.go
```diff
@@ -203,19 +212,14 @@ func (module *module) Delete(ctx context.Context, orgID valuer.UUID, id valuer.U
 	}
 
 	err = module.store.RunInTx(ctx, func(ctx context.Context) error {
-		err := module.store.Delete(ctx, orgID, id)
-		if err != nil {
+		err := module.DeletePublic(ctx, orgID, id)
```
vikrantgupta25, line 215: changed the delete public from store function to module function, as on delete of dashboard tuples should be cleared up as well, IN case the public_dashboard doesn't exists then move on

PR #9607, opened 2025-11-18
Fix: PromQL handle __all__ variables

### LINKED ISSUE #9601: Dynamic variable doesn't work with PromQL when `ALL` option is selected
## How to reproduce

1. Create a dynamic variable of type multi-select with `ALL` option enabled
2. Use it in an promql expression, example (`sum(rate({"system.cpu.time", "host.name"=~"$host.name"}[5m]))`) where `host.name` is the variable name
3. Verify if it works with selecting one host
4. Verify if it works with selecting two hosts
5. Verify if it works with selecting `ALL` for variable

nikhilmantri0902:
## 📄 Summary
Closes #9601 

---

## ✅ Changes

---

ex:

- `frontend`
- `backend`
- `devops`
- `bug`
- `enhancement`
- `ui`
- `test`

---

## 👥 Reviewers

- frontend / backend / devops

---

## 🧪 How to Test

1. ...
2. ...
3. ...

---

## 🔍 Related Issues

Closes #

---

---

## 📋 Checklist

- [ ] Dev Review

---

### DIFF
--- a/pkg/querier/promql_query.go
+++ b/pkg/querier/promql_query.go
@@ -87,7 +87,8 @@ func (q *promqlQuery) removeAllVarMatchers(query string, vars map[string]qbv5.Va
 	// Create visitor and walk the AST
 	visitor := &allVarRemover{allVars: allVars}
 	if err := parser.Walk(visitor, expr, nil); err != nil {
-		return "", errors.WrapInvalidInputf(err, errors.CodeInvalidInput, "invalid promql query %q", query)
+		q.logger.ErrorContext(context.TODO(), "unexpected error while removing __all__ variable matchers", "error", err, "query", query)
+		return "", errors.WrapInternalf(err, errors.CodeInternal, "error while removing __all__ variable matchers")
 	}
 
 	// Convert the modified AST back to a string


pkg/querier/promql_query.go
```diff
@@ -59,8 +60,52 @@ func (q *promqlQuery) Window() (uint64, uint64) {
...
+		return query, nil
+	}
+
+	// TODO(nikhilmantri0902, srikanthccv): should we return error if we fail to parse the query?
+	// or allow it for further processing?
+	// Parse the query into an AST
+	expr, err := parser.ParseExpr(query)
+	if err != nil {
+		return "", errors.WrapInternalf(err, errors.CodeInternal, "failed to parse PromQL query for __all__ removal")
+	}
```
srikanthccv: This wouldn't be appropriate error type. Right now, if the query input by user is invalid, we throw invalid error and say "invalid promql query ...". With this change, it will be internal which won't be right error type. Please check when does the `ParseExpr` return error?

pkg/querier/promql_query.go
```diff
@@ -59,8 +60,52 @@ func (q *promqlQuery) Window() (uint64, uint64) {
...
+
+	// Create visitor and walk the AST
+	visitor := &allVarRemover{allVars: allVars}
+	if err := parser.Walk(visitor, expr, nil); err != nil {
```
srikanthccv, line 89: Same, when does the `Walk` return error and how should they be handled?

pkg/querier/promql_query_parser.go
```diff
@@ -0,0 +1,63 @@
...
+// shouldRemoveMatcher checks if a matcher value contains a variable reference that has __all__ value.
+func (v *allVarRemover) shouldRemoveMatcher(value string) bool {
+	// Check for $var pattern
+	if strings.Contains(value, "$") {
+		for varName := range v.allVars {
+			if strings.Contains(value, "$"+varName) {
+				return true
+			}
+		}
+	}
```
srikanthccv, line 45: `v.allVars` is already a map so do direct lookup instead of loop.

pkg/querier/promql_query.go
nikhilmantri0902: if the promql expression is syntactically incorrect, it gives us error. Throwing an error with InvalidInput code now.

pkg/querier/promql_query.go
nikhilmantri0902: It returns error if the Visit method returns an error. In our case, the visit method is not returning an error. How should this error then be handled?

pkg/querier/promql_query_parser.go
nikhilmantri0902: done

pkg/querier/promql_query.go
srikanthccv: > It returns error if the Visit method returns an error. In our case, the visit method is not returning an error

That doesn't make sense.

nikhilmantri0902: As discussed logging this error and returning internal error.

Issue #9241, opened 2025-10-01
Status Error Message Should Not Display Fully, should be expandable.


gkarthi-signoz:
Currently in traces, entire status error message is displayed. Instead, should be truncated and have an expandable view where user can optionally see the entire message. 

https://github.com/user-attachments/assets/813d452d-f610-4b4f-809b-d1f151d11f59

makeavish:
@gkarthi-signoz Please open product issues on main SigNoz repo - https://github.com/SigNoz/signoz/issues
Transferring the issue

PR #9630, opened 2025-11-20
feat: add support for truncating long status_message in trace details and showing expandable popover on hover

ahmadshaheer:
## 🔍 Related Issues

Closes https://github.com/SigNoz/signoz/issues/9241

---

Before:

https://github.com/user-attachments/assets/12112848-b25a-459a-9fa2-310331e3f151

After:

https://github.com/user-attachments/assets/15815a0e-36a1-489c-8c86-e979e7035447

---

## 📋 Checklist

- [x] Dev Review


ahmadshaheer:
> I’m not sure this is the best way to display the error message. The component that was created seems to cater only to this specific case, whereas it should probably be a common component. Also, the error badge will now behave differently from the other badges, which already have inconsistent experiences across different places. We should involve the design team and address this properly. cc. @nityanandagohain @YounixM

@aks07, we have a similar UX in trace events as well. I guess it would be better to go ahead with reviewing and merging this PR. we can separately discuss with design team and work on unifying the UX for all attributes. 


https://github.com/user-attachments/assets/209dd262-70c3-4e38-bbeb-e0a54bfff98d

nityanandagohain:
@ahmadshaheer let's share with the design team if there is a better component to help with this now, else bring it to their radar and we can proceed with this.

PR #9637, opened 2025-11-20
fix: format numeric strings without quotes, preserve quoted values

ahrefabhi:
## Description

This PR fixes the value formatting logic in the query builder to properly handle numeric strings and preserve already-quoted values.

https://github.com/user-attachments/assets/d67a8934-e798-42df-b649-2e79dcdf16dd


srikanthccv (approved):

YounixM (approved):
Codewise - LGTM.

Issue #9398, opened 2025-10-23
Logs - Individual field in the body should not copy the whole body


YounixM:

ahmadshaheer:
@YounixM, can you please share more details about this?

nityanandagohain:
Check how much effort is required on this, if too much then we should actually look at replacing the entire component https://github.com/SigNoz/signoz/issues/9565

PR #9657, opened 2025-11-23
feat: add support for copying individual JSON tree nodes in log details

ahmadshaheer:
## 📄 Summary
Adds the ability to copy individual nodes from the JSON tree view in log details. Users can now click on any key-value pair in the JSON structure to copy it to their clipboard.

---

## 🔍 Related Issues

Closes https://github.com/SigNoz/signoz/issues/9398

---

Before:

https://github.com/user-attachments/assets/92602072-6ad1-48bc-a7d0-4eb470822f6f

After:

https://github.com/user-attachments/assets/b2250abf-3a92-49d1-9785-4c4632e30d63

---

## 📋 Checklist

- [x] Dev Review


frontend/src/container/LogDetailedView/BodyTitleRenderer.tsx
```diff
@@ -75,18 +80,53 @@ function BodyTitleRenderer({
...
+		[nodeKey, parentIsArray, setCopy, value, notifications],
+	);
 
 	return (
-		<TitleWrapper onMouseDown={handleTextSelection}>
-			<Dropdown menu={menu} trigger={['click']}>
-				<SettingOutlined style={{ marginRight: 8 }} className="hover-reveal" />
-			</Dropdown>
+		<TitleWrapper onClick={handleNodeClick}>
+			{typeof value !== 'object' && (
```
aks07, line 123: Why are we adding this condition: typeof value !== 'object'

frontend/src/container/LogDetailedView/BodyTitleRenderer.tsx
```diff
@@ -75,18 +80,53 @@ function BodyTitleRenderer({
 		onClick: onClickHandler,
 	};
 
-	const handleTextSelection = (e: React.MouseEvent): void => {
-		// Prevent tree node click when user is trying to select text
-		e.stopPropagation();
-	};
+	const handleNodeClick = useCallback(
```
aks07, line 83: Current implement is copying in the following format:
"service_meta.name": "demo-service"

Which is neither here nor there. Should it not be just "demo-service"? This is consistent with other details copying as well.

aks07: @nityanandagohain

aks07 (changes requested):
Now that we have added ability to copy object key, the user can no longer copy whole object. This could create an issue in my opinion.

aks07:
> Now that we have added ability to copy object key, the user can no longer copy whole object. This could create an issue in my opinion.

@nityanandagohain thoughts?

ahmadshaheer:
> Now that we have added ability to copy object key, the user can no longer copy whole object. This could create an issue in my opinion.

@aks07 the user can copy the whole JSON from the JSON view above the attributes IMO.

Screenshot 2025-11-24 182636

CC: @nityanandagohain

frontend/src/container/LogDetailedView/BodyTitleRenderer.tsx
ahmadshaheer: The condition `typeof value !== 'object' `ensures the filter dropdown (containing "Filter for" and "Filter out" options) is only displayed for primitive values (strings, numbers, booleans, etc.), not for objects or arrays.

nityanandagohain:
@ahmadshaheer I think the issue was different. If I drag and copy(cmd +c) one value of a certain key then the entire body gets copied, was that addressed ?

nityanandagohain:
@aks07 let's go ahead with this functionatily of being able to drag and select as well as click and select. The previous implementation was kind of useless so this looks good.

frontend/src/container/LogDetailedView/BodyTitleRenderer.tsx
nityanandagohain: since as of now it copies the nested keys as well, lets start with this, but @ahmadshaheer  do keep a issue where we should decided what should we do to handle an actual click.

PR #9660, opened 2025-11-24
style: metrics custom function css overflow issues

ishan621:
## 📄 Summary

---

## ✅ Changes

- [x] Feature: CSS - We resolved a UI issue in the Metrics panel → Custom Function section where long input values (e.g., for timeShift, cutOffTime) would overflow the input field, forcing horizontal scrolling.
- [x] Bug fix:  To fix this, we implemented a dynamic input resizing mechanism using a hidden mirror <span>. As the user types, the span calculates the rendered text width, and we update the input width in real time using a ref. The input now expands smoothly between the defined min (70px) and max (150px) widths, ensuring the full value remains visible without scrolling. This approach preserves layout consistency and significantly improves the typing experience.
- [x] We created a common tooltip component , on exceeding max width , user will be able to see a tooltip showing full value

---

## 🧪 How to Test

1. Open the staging environment: https://app.us.staging.signoz/
2. Navigate to Metrics from the left-side panel.
3. Go to the Explorer section inside Metrics.
4. In the query builder, use the Custom Function (fx) feature.
5. Select any function that accepts a value (e.g., timeShift, cutOffTime, etc.) and try typing different input lengths.
    Verify that the input field:
    * Automatically resizes between the defined min/max width
    * Does not overflow
    * Shows the complete value without requiring scrolling
    * On exceeding 150px, user will see ellipses and full values on hover on a tooltip
    * Maintains proper layout even when multiple custom functions are added

---

## 🔍 Related Issues

Closes # https://github.com/SigNoz/engineering-pod/issues/3292

---

Before: 

https://github.com/user-attachments/assets/6d5b72b8-9515-436d-9cf7-b93ef04777f2

After :

[

https://github.com/user-attachments/assets/7afe6d82-62d1-414e-be5c-957044ce75f1

]

---

## 📋 Checklist

- [x] Dev Review

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> Introduces an auto-resizing input with overflow tooltip and integrates it into Query Builder function arguments, with tests and style updates.
> 
> - **Components**:
>   - Add `OverflowInputToolTip` (`frontend/src/components/OverflowInputToolTip/OverflowInputToolTip.tsx`) with hidden mirror span for auto-width (70–150px) and tooltip on overflow; styles in `OverflowInputToolTip.scss`.
>   - Add unit tests `OverflowInputToolTip.test.tsx` covering overflow, disabled state, uncontrolled typing, and structure.
> - **Query Builder**:
>   - Replace plain input with `OverflowInputToolTip` in `QueryFunctions/Function.tsx`; manage local `value` state and `useMemo`-debounced `handleUpdateFunctionArgs`.
>   - Update styles in `QueryFunctions.styles.scss` (`.query-function-value` width to `70px`) and maintain seamless selector/button styling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ddbc6af11f8ce7adb38636b43eb7a4b49679085f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


frontend/src/container/QueryBuilder/components/QueryFunctions/Function.tsx
```diff
@@ -47,6 +49,22 @@ export default function Function({
...
+	useEffect(() => {
+		if (!mirrorRef.current || !inputRef.current?.input) return;
+
+		const mirrorWidth = mirrorRef.current.offsetWidth + 24; // padding
+		const newWidth = Math.min(150, Math.max(70, mirrorWidth));
+
+		// AntD input actual DOM element is inputRef.current.input
+		inputRef.current.input.style.width = `${newWidth}px`;
+	}, [value]);
+
```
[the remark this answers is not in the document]
ishan621: Added useMemo in debounce function, useMemo allows us to create the debounced function only once, and reuse the same function instance across renders

frontend/src/container/QueryBuilder/components/QueryFunctions/Function.tsx
```diff
@@ -47,6 +49,22 @@ export default function Function({
 		functionValue = funcData.args?.[0]?.value;
 	}
 
+	const [value, setValue] = useState<string>(
+		functionValue !== undefined ? String(functionValue) : '',
+	);
```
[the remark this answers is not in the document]
ishan621: In this component, the value state is initialized from funcData.args?.[0]?.value on mount. Since funcData.args is never updated by the parent while the component is mounted, the input value does not need to track changes from props. Adding a useEffect to synchronize props with state would be unnecessary and harmful: it could override the user’s input while typing, causing lag and inconsistent behavior.

Therefore, initializing the state once on mount is correct and sufficient, and the current implementation follows proper unidirectional data flow.

frontend/src/components/OverflowInputToolTip/OverflowInputToolTip.tsx
```diff
@@ -0,0 +1,70 @@
...
+			...rest
+		},
+		ref,
+	) => {
+		const inputRef = useRef<InputRef>(null);
+		const [isOverflowing, setIsOverflowing] = useState<boolean>(false);
+
+		useImperativeHandle(ref, () => inputRef.current as InputRef, []);
+
+		useEffect(() => {
```
aks07: Can we drive this through onChange? ie. onChange do the check done in this useEffect.
 Idea is to reduce the number of side effects that we add to the code.

frontend/src/container/QueryBuilder/components/QueryFunctions/Function.tsx
```diff
@@ -47,9 +49,25 @@ export default function Function({
...
-	const debouncedhandleUpdateFunctionArgs = debounce(
-		handleUpdateFunctionArgs,
-		500,
+	const [value, setValue] = useState<string>(
+		functionValue !== undefined ? String(functionValue) : '',
+	);
+	const inputRef = useRef<InputRef>(null);
+	const mirrorRef = useRef<HTMLSpanElement>(null);
+
+	useEffect(() => {
```
aks07: Same here

frontend/src/container/QueryBuilder/components/QueryFunctions/Function.tsx
```diff
@@ -89,14 +107,38 @@ export default function Function({
...
+						tooltipPlacement="top"
+						style={{
+							width: 70,
+							minWidth: 70,
+							maxWidth: 150,
+						}}
+					/>
+					<span
+						ref={mirrorRef}
+						style={{
```
aks07: Lets avoid inline styles and add a css class

aks07 (changes requested):

frontend/src/components/OverflowInputToolTip/OverflowInputToolTip.tsx
ishan621: We tried this to remove useEffect but it would be needed for mounting state to judge if input is overflowing while getting mounted , as input can have default values and disabled state check hence this useEffect would be needed

frontend/src/components/OverflowInputToolTip/OverflowInputToolTip.tsx
```diff
@@ -0,0 +1,68 @@
...
+			<Tooltip title={tooltipTitle} placement={tooltipPlacement}>
+				{/* eslint-disable-next-line react/jsx-props-no-spreading */}
+				<Input
+					{...rest}
+					value={value}
+					defaultValue={defaultValue}
+					onChange={onChange}
+					disabled={disabled}
+					ref={inputRef}
+					className="overflow-input"
```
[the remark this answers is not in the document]
ishan621: we don;t pass classnames in props

frontend/src/container/QueryBuilder/components/QueryFunctions/Function.tsx
ishan621: We moved the width logic out of useEffect and into onChange to get immediate width updates while typing, and we added useLayoutEffect to handle the initial sizing before paint. But useLayoutEffect ran too early—before AntD finished applying its styles—so the mirror element measured incorrect font metrics, causing the input to shrink or jump. Because useEffect runs after the component is fully painted with correct styles, we reverted to useEffect to avoid the UI glitch and ensure accurate width calculation.

frontend/src/container/QueryBuilder/components/QueryFunctions/Function.tsx
```diff
@@ -47,9 +49,41 @@ export default function Function({
...
+		const newWidth = Math.min(150, Math.max(70, mirrorWidth));
+
+		// AntD input actual DOM element is inputRef.current.input
+		inputRef.current.input.style.width = `${newWidth}px`;
+	}, [value]);
+
+	const debouncedhandleUpdateFunctionArgs = useMemo(
+		() => debounce(handleUpdateFunctionArgs, 500),
+		[handleUpdateFunctionArgs],
+	);
```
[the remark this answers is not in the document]
ishan621: We don’t clean up the debounced function because it only initializes when the component mounts (via useMemo) and never gets recreated on every render. Since the debounced callback fires only in response to user typing and not during unmount, there’s no pending async work triggered by the component lifecycle itself. In other words, the debounce isn’t scheduled automatically by mounting—it's scheduled only after the user types—so there’s no guaranteed pending invocation to cancel during unmount. That makes cleanup optional rather than required in this specific usage.

frontend/src/container/QueryBuilder/components/QueryFunctions/Function.tsx
```diff
@@ -89,14 +109,28 @@ export default function Function({
...
+							const newVal = event.target.value;
+							setValue(newVal);
+							debouncedhandleUpdateFunctionArgs(funcData, index, event.target.value);
+						}}
+						tooltipPlacement="top"
+						style={{
+							width: 70,
+							minWidth: 70,
+							maxWidth: 150,
+						}}
```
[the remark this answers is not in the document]
ishan621: it is a prop style that is being passed here

frontend/src/container/QueryBuilder/components/QueryFunctions/Function.tsx
```diff
@@ -89,14 +109,28 @@ export default function Function({
...
+						className="query-function-value"
+						autoFocus
+						value={value}
+						onChange={(event): void => {
+							const newVal = event.target.value;
+							setValue(newVal);
+							debouncedhandleUpdateFunctionArgs(funcData, index, event.target.value);
+						}}
+						tooltipPlacement="top"
+						style={{
```
aks07: Lets drive these styles through classname. OverflowInputTooltip should accept classname as prop.

frontend/src/container/QueryBuilder/components/QueryFunctions/Function.tsx
```diff
@@ -89,14 +109,28 @@ export default function Function({
...
+							width: 70,
+							minWidth: 70,
+							maxWidth: 150,
+						}}
+					/>
+					<span ref={mirrorRef} className="query-function-mirror">
```
aks07: Lets try to solve this without mirrorRef

frontend/src/container/QueryBuilder/components/QueryFunctions/Function.tsx
```diff
@@ -47,9 +51,25 @@ export default function Function({
...
-	const debouncedhandleUpdateFunctionArgs = debounce(
-		handleUpdateFunctionArgs,
-		500,
+	const [value, setValue] = useState<string>(
+		functionValue !== undefined ? String(functionValue) : '',
+	);
+	const inputRef = useRef<InputRef>(null);
+	const mirrorRef = useRef<HTMLSpanElement>(null);
+
+	useEffect(() => {
```
aks07: Ideally the common component should handle the input size change as it is core functionality of the component. Lets move this logic in the common component

frontend/src/container/QueryBuilder/components/QueryFunctions/Function.tsx
ishan621: Done

frontend/src/container/QueryBuilder/components/QueryFunctions/Function.tsx
ishan621: Moved mirrorRef logic into common component

frontend/src/container/QueryBuilder/components/QueryFunctions/Function.tsx
ishan621: We tried to resolve via mirrorRef but could not as by default input does not expose text width
For <input> elements:
* scrollWidth = full width of input element, not the text size
* No API exists to measure text width inside an input
* Browsers do not auto-shrink/grow <input> based on content

We went ahead with span as
Hidden <span> mirrors the actual text with the same font
A <span> can grow naturally based on the text rendered with the same CSS font + spacing.

frontend/src/container/QueryBuilder/components/QueryFunctions/Function.tsx
```diff
@@ -47,9 +51,14 @@ export default function Function({
 		functionValue = funcData.args?.[0]?.value;
 	}
 
-	const debouncedhandleUpdateFunctionArgs = debounce(
-		handleUpdateFunctionArgs,
-		500,
+	const [value, setValue] = useState<string>(
+		functionValue !== undefined ? String(functionValue) : '',
+	);
+	const inputRef = useRef<InputRef>(null);
```
aks07: why do we need to define the inputRef here as well?

ishan621: cleaned up code, removed unused input ref, forward ref , imperative handle from code

frontend/src/components/OverflowInputToolTip/OverflowInputToolTip.tsx
```diff
@@ -0,0 +1,74 @@
...
+	value,
+	defaultValue,
+	onChange,
+	disabled = false,
+	tooltipPlacement = 'top',
+	className,
+	minAutoWidth = 70,
+	maxAutoWidth = 150,
+	...rest
+}: OverflowTooltipInputProps): JSX.Element {
```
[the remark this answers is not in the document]
ishan621: removed ref from test files

aks07 (approved):

ishan621:
> One minor comment. Rest looks good

fixed

frontend/src/components/OverflowInputToolTip/OverflowInputToolTip.tsx
```diff
@@ -0,0 +1,76 @@
...
+		if (!input || !mirror) return;
+
+		// mirror text content
+		mirror.textContent = String(value ?? '') || ' ';
+
+		// measure + clamp
+		const mirrorWidth = mirror.offsetWidth + 24;
+		const newWidth = Math.min(maxAutoWidth, Math.max(minAutoWidth, mirrorWidth));
+		input.style.width = `${newWidth}px`;
+	}, [value, minAutoWidth, maxAutoWidth]);
```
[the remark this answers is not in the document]
ishan621: added for disabled also

PR #9669, opened 2025-11-25
feat: time aware dynamic field mapper

nityanandagohain:
Fixes https://github.com/SigNoz/engineering-pod/issues/3017

This PR makes sure that start and end timestamp is sent to the field mapper which then evalutates which column to use based on when it was added. As of now the logic is that if start is after the time when new column was added then we use the new column else both the columns ( for end < added date it could have been just the old column but in this case it's fine)

---

The exact change of this pr is added if the query start is after new resource json is added then just use the new resource json else use both the old and new column with multi if.

Future work https://github.com/SigNoz/engineering-pod/issues/3857

---

> [!NOTE]
> **Medium Risk**
> Touches core querybuilding/field-mapping paths across logs/metrics/metadata and changes multiple public interfaces (e.g., `PrepareWhereClause`, `FieldMapper`, `ConditionBuilder`), so regressions could alter query results or produce incorrect SQL for some time ranges. Added evolution-based column selection mitigates correctness but increases complexity and surface area.
> 
> **Overview**
> Adds *time-range aware* querybuilding by threading `ctx`, `StartNs`, and `EndNs` through `PrepareWhereClause`, aggregation rewriting, field mappers, and condition builders, replacing prior ad-hoc `context.Background()`/separate start/end params.
> 
> Introduces column evolution support for logs/resource fields: the logs field mapper can now choose the correct backing column(s) for a query window (single column when unambiguous; `multiIf` across old+new when needed), and logs condition building updates `exists`/`not exists` behavior accordingly.
> 
> Extends telemetry metadata to fetch per-field column evolution entries from a new ClickHouse table (`distributed_column_evolution_metadata`) and attach `Evolutions` onto returned keys; updates providers and adds/updates extensive unit tests to validate evolution selection and new SQL output.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit be12759524151afb97357385cb06bc2480d825b7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkg/querybuilder/agg_rewrite.go
```diff
@@ -77,7 +81,11 @@ func (r *aggExprRewriter) Rewrite(
 		return "", nil, errors.NewInternalf(errors.CodeInternal, "no SELECT items for %q", expr)
 	}
 
-	visitor := newExprVisitor(r.logger, keys,
+	visitor := newExprVisitor(
+		r.startNs,
+		r.endNs,
```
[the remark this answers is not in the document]
nityanandagohain: thanks

pkg/querybuilder/agg_rewrite.go
```diff
@@ -200,7 +216,7 @@ func (v *exprVisitor) VisitFunctionExpr(fn *chparser.FunctionExpr) error {
 				FullTextColumn:   v.fullTextColumn,
 				JsonBodyPrefix:   v.jsonBodyPrefix,
 				JsonKeyToKey:     v.jsonKeyToKey,
-            }, 0, 0,
+			}, 0, 0,
```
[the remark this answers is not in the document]
nityanandagohain: thanks

piyushsingariya (commented):
LGTM

pkg/telemetrylogs/field_mapper.go
```diff
@@ -112,17 +118,34 @@ func (m *fieldMapper) FieldFor(ctx context.Context, key *telemetrytypes.Telemetr
...
+		}
+
+		// get all evolution for the column
+		evolutions := m.evolutionMetadataStore.Get(ctx, orgID, baseColumn.Name)
+
+		// restricting now to just one entry where we know we changes from map to json
+		if len(evolutions) > 0 && evolutions[0].ReleaseTime.Before(tsStartTime) {
+			return fmt.Sprintf("%s.`%s`::String", column.Name, key.Name), nil
+		}
+
```
tushar-signoz, line 257: It would be good to add a debug statement for not finding evolutions or whichever case may be.

pkg/telemetrylogs/key_evolution_metadata.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	// Build query to fetch all key evolution metadata
+	sb := sqlbuilder.NewSelectBuilder()
+	sb.Select(
+		"base_column",
+		"base_column_type",
+		"new_column",
+		"new_column_type",
+		"release_time",
```
tushar-signoz: Can we create constants for these column names

pkg/telemetrylogs/field_mapper.go
```diff
@@ -112,17 +118,34 @@ func (m *fieldMapper) FieldFor(ctx context.Context, key *telemetrytypes.Telemetr
...
+			orgID, err = valuer.NewUUID(claims.OrgID)
+			if err != nil {
+				return "", errors.Wrapf(err, errors.TypeInvalidInput, errors.CodeInvalidInput, "invalid orgId %s", claims.OrgID)
+			}
+		}
+
+		// get all evolution for the column
+		evolutions := m.evolutionMetadataStore.Get(ctx, orgID, baseColumn.Name)
+
+		// restricting now to just one entry where we know we changes from map to json
```
tushar-signoz: nit typo: `changes` -> `changed`

pkg/telemetrylogs/key_evolution_metadata.go
```diff
@@ -0,0 +1,163 @@
...
+
+	logger.DebugContext(ctx, "Successfully fetched key evolution metadata from ClickHouse", "count", len(metadataByKey))
+}
+
+// Add adds a metadata key for the given key name and orgId.
+// This is primarily for testing purposes. In production, data should come from ClickHouse.
+func (k *KeyEvolutionMetadata) Add(ctx context.Context, orgId valuer.UUID, keyName string, key *telemetrytypes.KeyEvolutionMetadataKey) {
+	k.logger.WarnContext(ctx, "Add is not implemented for key evolution metadata")
+
+}
```
srikanthccv: Let's not do this. Since you already have the implementation for tests that implements the interface, you can populate the required data during the creation of new mock store rather adding `Add` to interface just to aid in testing.

pkg/telemetrylogs/key_evolution_metadata.go
```diff
@@ -0,0 +1,163 @@
...
+
+// Get retrieves all metadata keys for the given key name and orgId from cache.
+// Returns an empty slice if the key is not found in cache.
+func (k *KeyEvolutionMetadata) Get(ctx context.Context, orgId valuer.UUID, keyName string) []*telemetrytypes.KeyEvolutionMetadataKey {
+
+	cacheKey := KeyEvolutionMetadataCacheKeyPrefix + keyName
+	var cachedData CachedKeyEvolutionMetadata
+	if err := k.cache.Get(ctx, orgId, cacheKey, &cachedData); err != nil {
+		// Cache miss - fetch from ClickHouse and try again
```
srikanthccv: err here doesn't necessarily mean it's a miss, it could be some legit error from cache, please check `errors.Ast(NotFound` to confirm it's a miss, otherwise log error.

pkg/telemetrylogs/key_evolution_metadata.go
```diff
@@ -0,0 +1,163 @@
...
+		k.fetchFromClickHouse(ctx, orgId)
+
+		// Check cache again after fetching
+		if err := k.cache.Get(ctx, orgId, cacheKey, &cachedData); err != nil {
```
srikanthccv: By reading the function name `fetchFromClickHouse`, I wouldn't expect it to cache the data, let alone be aware that the cache is available. I would say, keep `fetchFromClickHouse` simple, fetching and returning from the ClickHouse and have the `cache.Set` logic here in the `Get`.

pkg/telemetrylogs/key_evolution_metadata.go
```diff
@@ -0,0 +1,163 @@
...
+			return nil
+		}
+	}
+
+	// Return a copy to prevent external modification
+	result := make([]*telemetrytypes.KeyEvolutionMetadataKey, len(cachedData.Keys))
+	copy(result, cachedData.Keys)
```
srikanthccv: Is this really needed? Aren't new objects created with each `Get` because the data is JSON serialised/deserialised?

pkg/telemetrylogs/key_evolution_metadata.go
```diff
@@ -0,0 +1,163 @@
...
+	store := k.telemetryStore
+	logger := k.logger
+
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+
+	if store.ClickhouseDB() == nil {
+		logger.WarnContext(ctx, "ClickHouse connection not available for key evolution metadata fetch")
+		return
+	}
```
srikanthccv: When would this happen? I don't think it should ever be possible to create `KeyEvolutionMetadata` without TelemetryStore dep?

pkg/telemetrylogs/field_mapper.go
```diff
@@ -112,17 +118,34 @@ func (m *fieldMapper) FieldFor(ctx context.Context, key *telemetrytypes.Telemetr
...
+		if len(evolutions) > 0 && evolutions[0].ReleaseTime.Before(tsStartTime) {
```
srikanthccv: It's named `evolutions`; 

- What does it mean to have more than 1 evolution?
- Why is that the 0th evolution release time is only considered?
- Why is that the nothing from the actual `evolutions` used besides the release_time, which brings to the question, what are the other columns in the table are supporting for?

srikanthccv (commented):
Thanks, have some questions.

pkg/telemetrylogs/key_evolution_metadata.go
nityanandagohain: yeah not required.

pkg/telemetrylogs/key_evolution_metadata.go
nityanandagohain: yes makes sense

pkg/telemetrylogs/field_mapper.go
nityanandagohain: added a statement in the evolution package itself

pkg/telemetrylogs/field_mapper.go
nityanandagohain: 1. 
Eg:- we change from 
* `resources_string`  new version `resource` at time T1
* `resources_string` new column `resource_v2` at time T2.

So basically here we are tying things to the base resources_string itself.

2. In this case if it's for resource we know we just have one, if we have to add the newer `resource_v2` we will have to update the logic here and add the new select command as well.

3. Thanks updated now, ideally here it should have used the evolutions[0].name . The types of both base and new one is not used as of now.

nityanandagohain:
Have added the changes and some test cases as wel, do have a look again.

pkg/telemetrylogs/field_mapper.go
srikanthccv: > The types of both base and new one is not used as of now

Does that mean it's a hardcoded way of doing the things; as in you know the type so you wrote the conditional code but that can be generically used by anyone else?

srikanthccv: > So basically here we are tying things to the base resources_string itself.

This also follows the same lines, the current mechanism is essentially aiding with release time with first entry but every time the type is evolved, the new code needs to be written, doesn't that defeat the purpose of the time aware evolution objectives to make it work dynamically based on the info from meta table?

nityanandagohain: Cool add discussed will add the changes in this mapper to make it more independent unless a completely new type is added.

pkg/telemetrylogs/field_mapper.go
```diff
@@ -193,12 +300,12 @@ func (m *fieldMapper) ColumnExpressionFor(
...
 		} else if len(keysForField) == 1 {
 			// we have a single key for the field, use it
-			colName, _ = m.FieldFor(ctx, keysForField[0])
+			colName, _ = m.FieldFor(ctx, tsStart, tsEnd, keysForField[0])
 		} else {
 			// select any non-empty value from the keys
 			args := []string{}
 			for _, key := range keysForField {
-				colName, _ = m.FieldFor(ctx, key)
+				colName, _ = m.FieldFor(ctx, tsStart, tsEnd, key)
```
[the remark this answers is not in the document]
nityanandagohain: not a change in this pr

pkg/telemetrylogs/field_mapper.go
nityanandagohain: Have made the changes, some minor changes might be needed based on how it is used in JSON, but overall it takes care of future things that will come up.
* adding multiple evolutions i.e resources_string, resource, resource_v2 etc.
* generic enough to suport other keys like attributes.
* Support for path which takes care of promoted path.

srikanthccv: Thanks

pkg/querybuilder/fallback_expr.go
```diff
@@ -44,7 +46,7 @@ func CollisionHandledFinalExpr(
 
 	addCondition := func(key *telemetrytypes.TelemetryFieldKey) error {
 		sb := sqlbuilder.NewSelectBuilder()
-		condition, err := cb.ConditionFor(ctx, key, qbtypes.FilterOperatorExists, nil, sb, 0, 0)
+		condition, err := cb.ConditionFor(ctx, key, qbtypes.FilterOperatorExists, nil, sb, startNs, endNs)
```
srikanthccv: nit: let's move the startNs, endNs to beginning of args list like we do for field for and rest of the places.

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -147,7 +148,7 @@ func PrepareWhereClause(query string, opts FilterExprVisitorOpts, startNs uint64
 	}
 
 	// Visit the parse tree with our ClickHouse visitor
-	cond := visitor.Visit(tree).(string)
+	cond := visitor.Visit(ctx, tree).(string)
```
srikanthccv: Let's make the `context.Context` as a visitor option rather than arg here to the `visitor.Visit`, why? because otherwise visitor doesn't satisfy the `FilterQueryVisitor` interface https://github.com/SigNoz/signoz/blob/479cba7dd296874c1b1277466b74cd4447144288/pkg/parser/grammar/filterquery_visitor.go#L8

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -174,69 +175,69 @@ func PrepareWhereClause(query string, opts FilterExprVisitorOpts, startNs uint64
 }
 
 // Visit dispatches to the specific visit method based on node type
-func (v *filterExpressionVisitor) Visit(tree antlr.ParseTree) any {
+func (v *filterExpressionVisitor) Visit(ctx context.Context, tree antlr.ParseTree) any {
```
srikanthccv: If we use context as option, we don't need to change the signature and still be able to use the right context.

pkg/telemetrylogs/condition_builder_test.go
```diff
@@ -242,7 +247,7 @@ func TestConditionFor(t *testing.T) {
 			},
 			operator:      qbtypes.FilterOperatorExists,
 			value:         nil,
-			expectedSQL:   "WHERE multiIf(resource.`service.name` IS NOT NULL, resource.`service.name`::String, mapContains(resources_string, 'service.name'), resources_string['service.name'], NULL) IS NOT NULL",
+			expectedSQL:   "WHERE multiIf(mapContains(resources_string, 'service.name'), resources_string['service.name'], NULL) IS NOT NULL",
```
srikanthccv: Why do we have `multiIf` here if there is only one column to look for?

pkg/types/telemetrytypes/telemetrytypestest/key_evolution_metadata.go
```diff
@@ -0,0 +1,40 @@
...
+// NewMockKeyEvolutionMetadataStore creates a new instance of MockKeyEvolutionMetadataStore with initialized maps
+func NewMockKeyEvolutionMetadataStore(metadata map[string]map[string][]*telemetrytypes.KeyEvolutionMetadata) *MockKeyEvolutionMetadataStore {
+	return &MockKeyEvolutionMetadataStore{
+		metadata: metadata,
+	}
+}
+
+// Get retrieves all metadata keys for the given key name and orgId.
+// Returns an empty slice if the key is not found.
+func (m *MockKeyEvolutionMetadataStore) Get(ctx context.Context, orgId valuer.UUID, keyName string) []*telemetrytypes.KeyEvolutionMetadata {
```
srikanthccv: nit: `orgId -> orgID` https://google.github.io/styleguide/go/decisions#initialisms

pkg/telemetrylogs/test_data.go
```diff
@@ -951,3 +953,23 @@ func buildCompleteFieldKeyMapCollision() map[string][]*telemetrytypes.TelemetryF
 	}
 	return keysMap
 }
+
+// mockKeyEvolutionMetadata builds a mock org-scoped key evolution metadata map for a column evolution.
+func mockKeyEvolutionMetadata(orgId valuer.UUID, releaseTime time.Time) map[string]map[string][]*telemetrytypes.KeyEvolutionMetadata {
```
srikanthccv: nit: find and replace `orgId` -> `orgID` please

pkg/telemetrylogs/key_evolution_metadata.go
```diff
@@ -0,0 +1,160 @@
...
+
+func NewKeyEvolutionMetadata(telemetryStore telemetrystore.TelemetryStore, cache cache.Cache, logger *slog.Logger) *KeyEvolutionMetadata {
+	return &KeyEvolutionMetadata{
+		cache:          cache,
+		telemetryStore: telemetryStore,
+		logger:         logger,
+	}
+}
+
+func (k *KeyEvolutionMetadata) fetchFromClickHouse(ctx context.Context, key string) []*telemetrytypes.KeyEvolutionMetadata {
```
srikanthccv: `key` as naming is bit confusing because I was thinking about the telemetryfieldkey like `service.name` but looks like the intent is `resources_string` or something like that?

pkg/telemetrylogs/filter_expr_logs_test.go
```diff
@@ -1,21 +1,33 @@
...
 // TestFilterExprLogs tests a comprehensive set of query patterns for logs search
 func TestFilterExprLogs(t *testing.T) {
-	fm := NewFieldMapper()
+	releaseTime := time.Date(2024, 1, 15, 10, 0, 0, 0, time.UTC)
+	ctx := context.Background()
+	orgId := valuer.GenerateUUID()
+	ctx = authtypes.NewContextWithClaims(ctx, authtypes.Claims{
+		OrgID: orgId.String(),
+	})
+	storeWithMetadata := telemetrytypestest.NewMockKeyEvolutionMetadataStore(mockKeyEvolutionMetadata(orgId, releaseTime))
```
srikanthccv: just curious, is there a reason why this test file has release time but others don't?

pkg/telemetrylogs/field_mapper.go
```diff
@@ -117,28 +123,130 @@ func (m *fieldMapper) getColumn(_ context.Context, key *telemetrytypes.Telemetry
...
 	case schema.ColumnTypeEnumJSON:
-		// json is only supported for resource context as of now
 		switch key.FieldContext {
 		case telemetrytypes.FieldContextResource:
-			oldColumn := logsV2Columns["resources_string"]
-			oldKeyName := fmt.Sprintf("%s['%s']", oldColumn.Name, key.Name)
+			// json is only supported for resource context as of now
+			if key.FieldContext != telemetrytypes.FieldContextResource {
+				return "", errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "only resource context fields are supported for json columns, got %s", key.FieldContext.String)
+			}
```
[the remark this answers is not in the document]
srikanthccv: .relevant and also below body context is also supported.

pkg/telemetrylogs/condition_builder.go
```diff
@@ -174,88 +179,120 @@ func (c *conditionBuilder) conditionFor(
...
 					return sb.NE(tblFieldName, value), nil
+				} else {
+					return sb.E(tblFieldName, value), nil
 				}
-				return sb.E(tblFieldName, value), nil
-			default:
-				return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "exists operator is not supported for low cardinality column type %s", elementType)
-			}
-		case schema.ColumnTypeEnumString:
-			value = ""
```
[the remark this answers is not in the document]
nityanandagohain: no it should just use the relevant ones

pkg/telemetrylogs/condition_builder.go
```diff
@@ -174,88 +175,106 @@ func (c *conditionBuilder) conditionFor(
...
+						return sb.E(leftOperand, true), nil
+					} else {
+						return sb.NE(leftOperand, true), nil
+					}
+				default:
+					return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "exists operator is not supported for map column type %s", valueType)
 				}
 			default:
-				return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "exists operator is not supported for map column type %s", valueType)
+				return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "exists operator is not supported for column type %s", column.Type)
```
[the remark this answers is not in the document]
nityanandagohain: this is not possible as evolutions are added to a a field struct only when it matches.

pkg/telemetrylogs/field_mapper.go
```diff
@@ -68,137 +69,279 @@ func NewFieldMapper() qbtypes.FieldMapper {
...
+	for _, evolution := range evolutions {
+		if evolution.ReleaseTime.After(tsStartTime) {
+			break
+		}
+		latestBaseEvolutionAcrossAll = evolution
+	}
+
+	if latestBaseEvolutionAcrossAll == nil {
+		return nil, nil, errors.Newf(errors.TypeInternal, errors.CodeInternal, "no base evolution found for columns %v", columns)
+	}
```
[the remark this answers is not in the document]
nityanandagohain: we expect the base evolution to be of epoch 0 always.

pkg/telemetrymetadata/metadata.go
```diff
@@ -1768,6 +1816,124 @@ func (t *telemetryMetaStore) fetchMeterSourceMetricsTemporality(ctx context.Cont
...
+			}
+		} else if selector.FieldName != "" {
+			// Only match by field name on any context
+			clauses = append(clauses,
+				sb.E("field_name", selector.FieldName),
+				sb.E("field_name", "__all__"),
+			)
+		}
+
+		sb.Where(sb.And(sb.E("signal", selector.Signal), sb.Or(clauses...)))
```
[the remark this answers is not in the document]
nityanandagohain: This is not the case, the query is something like field_context = resource and field_name = 'shusdf' or field_context='body' and filed_name ='user.name'

pkg/querybuilder/fallback_expr.go
```diff
@@ -115,12 +117,20 @@ func CollisionHandledFinalExpr(
 		stmts = append(stmts, colName)
 	}
 
+	// Remove empty strings from stmts and escape each one
+	escapedStmts := make([]string, 0, len(stmts))
 	for idx := range stmts {
-		stmts[idx] = sqlbuilder.Escape(stmts[idx])
+		if stmts[idx] != "" {
```
srikanthccv: When would there be empty statements?

pkg/querybuilder/fallback_expr.go
```diff
@@ -115,12 +117,20 @@ func CollisionHandledFinalExpr(
...
+			escapedStmts = append(escapedStmts, sqlbuilder.Escape(stmts[idx]))
+		}
+	}
+	stmts = escapedStmts
+	var multiIfStmt string
+	if len(stmts) > 1 {
+		multiIfStmt = fmt.Sprintf("multiIf(%s, NULL)", strings.Join(stmts, ", "))
+	} else {
+		multiIfStmt = stmts[0]
```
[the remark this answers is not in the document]
srikanthccv: @nityanandagohain will this ever happen?

pkg/querybuilder/resourcefilter/condition_builder.go
```diff
@@ -60,15 +60,15 @@ func (b *defaultConditionBuilder) ConditionFor(
 	// as we store resource values as string
 	formattedValue := querybuilder.FormatValueForContains(value)
 
-	column, err := b.fm.ColumnFor(ctx, key)
+	columns, err := b.fm.ColumnFor(ctx, startNs, endNs, key)
```
srikanthccv, line 64: Please add a note that there is only one column for resource fingerprints table so the evolution on main table doesn't effect the the checks here.

pkg/telemetrylogs/condition_builder.go
```diff
@@ -26,17 +26,18 @@ func NewConditionBuilder(fm qbtypes.FieldMapper, metadataStore telemetrytypes.Me
...
 	sb *sqlbuilder.SelectBuilder,
 ) (string, error) {
-	column, err := c.fm.ColumnFor(ctx, key)
+	columns, err := c.fm.ColumnFor(ctx, startNs, endNs, key)
 	if err != nil {
 		return "", err
 	}
 
-	if column.IsJSONColumn() && querybuilder.BodyJSONQueryEnabled {
+	if columns[0].IsJSONColumn() && querybuilder.BodyJSONQueryEnabled {
```
srikanthccv: This shouldn't be here but inside for each column? Take an example where the key went from String column into JSON column. The 0th column is string and 1st column is json so we can't decide here if the JSON condition can be used and return the json condition alone.

pkg/telemetrylogs/condition_builder.go
```diff
@@ -174,6 +175,22 @@ func (c *conditionBuilder) conditionFor(
...
+			fieldName := "__all__"
+			// we will use the corresponding column and its evolution entry for the query
+			newColumns, _, err := selectEvolutionsForColumns(columns, key.Evolutions, startNs, endNs, fieldName)
+			if err != nil {
+				return "", err
+			}
+			// this means the exists conditions are already taken care off, we don't need to add it
+			if len(newColumns) > 1 {
+				return "", nil
+			}
```
srikanthccv: Forgot the context, can you please help me understand how is this already taken care off? If so, please describe here itself how it's handled so the future reader has the detail along with the code.

pkg/telemetrylogs/condition_builder_test.go
```diff
@@ -11,14 +12,149 @@ import (
...
+						ColumnType:   "JSON()",
+						FieldContext: telemetrytypes.FieldContextResource,
+						FieldName:    "__all__",
+						ReleaseTime:  time.Date(2024, 1, 1, 0, 0, 0, 0, time.UTC),
+					},
+				},
+			},
+			operator:      qbtypes.FilterOperatorExists,
+			value:         nil,
+			expectedSQL:   "",
```
srikanthccv: In conjunction with the above question, please clarify why is this empty and how is it take care of?

pkg/telemetrylogs/filter_expr_logs_test.go
```diff
@@ -2386,7 +2394,7 @@ func TestFilterExprLogs(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(fmt.Sprintf("%s: %s", tc.category, limitString(tc.query, 50)), func(t *testing.T) {
 
-			clause, err := querybuilder.PrepareWhereClause(tc.query, opts, 0, 0)
+			clause, err := querybuilder.PrepareWhereClause(tc.query, opts, uint64(releaseTime.Add(-5*time.Minute).UnixNano()), uint64(releaseTime.Add(5*time.Minute).UnixNano()))
```
srikanthccv: Does it really matter what you pass here if there is no evolution in the first place? I am trying to understand why do we pass [release-5m, release+5m], is it to test without any evolution everything works as is? 

Look at the test `Very complex query example`, is still using the something like

```
multiIf(resource.`environment` IS NOT NULL, resource.`environment`::String, mapContains(resources_string, 'environment'), resources_string['environment'], NULL) <> 
```

I would expect it to be changed?

pkg/telemetrylogs/statement_builder.go
```diff
@@ -195,10 +195,13 @@ func (b *logQueryStatementBuilder) adjustKeys(ctx context.Context, keys map[stri
 			// check if all the key for the given field have been materialized, if so
 			// set the key to materialized
 			materilized := true
+			var evolutions []*telemetrytypes.EvolutionEntry
 			for _, key := range keys[k.Name] {
 				materilized = materilized && key.Materialized
+				evolutions = key.Evolutions
 			}
 			k.Materialized = materilized
+			k.Evolutions = evolutions
```
srikanthccv: This is happening only where there is no overall match, is there a case where evolutions should be added but is not added because there is overall match?

pkg/telemetrymetadata/metadata.go
```diff
@@ -581,9 +587,50 @@ func (t *telemetryMetaStore) getLogsKeys(ctx context.Context, fieldKeySelectors
 		keys = append(keys, bodyJSONPaths...)
 		complete = complete && finished
 	}
+
+	// fetch and add evolutions
+	metadataKeySelectors := getMetadataKeySelectors(keys)
```
srikanthccv: nit: rename to

`metadataKeySelectors` -> `evolutionMetadataKeySelectors`
`getMetadataKeySelectors` -> `getEvolutionMetadataKeySelectors`

to explicitly state these are for not general metadata but for key evolution

pkg/telemetrymetadata/metadata.go
```diff
@@ -581,9 +587,50 @@ func (t *telemetryMetaStore) getLogsKeys(ctx context.Context, fieldKeySelectors
...
+	var metadataKeySelectors []*telemetrytypes.EvolutionSelector
+	for _, keySelector := range keySelectors {
+		selector := &telemetrytypes.EvolutionSelector{
+			Signal:       keySelector.Signal,
+			FieldContext: keySelector.FieldContext,
+			FieldName:    "__all__", // Hardcoded for now
+		}
+		if keySelector.FieldContext == telemetrytypes.FieldContextBody {
+			selector.FieldName = keySelector.Name
+		}
```
srikanthccv: Why are we doing this hard-coded stuff here? We should use `keySelector.Name` as the `FieldName:`. The on the `fetchEvolutionEntryFromClickHouse` function ensure we fetch `__all__` entries as well. Am I missing something?

pkg/telemetrymetadata/metadata.go
```diff
@@ -581,9 +587,50 @@ func (t *telemetryMetaStore) getLogsKeys(ctx context.Context, fieldKeySelectors
...
+	evolutions, err := t.GetColumnEvolutionMetadataMulti(ctx, metadataKeySelectors)
+	if err != nil {
+		return nil, false, err
+	}
+	for i, key := range keys {
+		// Use the same logic as getMetadataKeySelectors to determine the selector key
+		fieldName := "__all__"
+		if keys[i].FieldContext == telemetrytypes.FieldContextBody {
+			fieldName = key.Name
+		}
```
srikanthccv: Same question, the consumers of the evolution shouldn't know there is something called as `__all__`, they should purely get evolutions back. What does it mean here? 

You should just pass the `key.Name` as `FieldName` all the time, and then it's the concern of the evolutions to return the right set of evolutions.

pkg/telemetrymetadata/metadata.go
```diff
@@ -1768,6 +1816,124 @@ func (t *telemetryMetaStore) fetchMeterSourceMetricsTemporality(ctx context.Cont
...
+func (c *CachedEvolutionEntry) MarshalBinary() ([]byte, error) {
+	return json.Marshal(c)
+}
+
+func (c *CachedEvolutionEntry) UnmarshalBinary(data []byte) error {
+	return json.Unmarshal(data, c)
+}
+
+func (k *telemetryMetaStore) fetchEvolutionEntryFromClickHouse(ctx context.Context, selectors []*telemetrytypes.EvolutionSelector) ([]*telemetrytypes.EvolutionEntry, error) {
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
srikanthccv: why?

pkg/telemetrymetadata/metadata.go
```diff
@@ -1768,6 +1816,124 @@ func (t *telemetryMetaStore) fetchMeterSourceMetricsTemporality(ctx context.Cont
...
+				// Only match context, accept any name
+				clauses = append(clauses, sb.E("field_context", selector.FieldContext))
+			}
+		} else if selector.FieldName != "" {
+			// Only match by field name on any context
+			clauses = append(clauses,
+				sb.E("field_name", selector.FieldName),
+				sb.E("field_name", "__all__"),
+			)
+		}
```
srikanthccv: This is convoluted code block. The `if` clause checks `FieldContext` and the else clause checks the `FieldName` which are independent pieces of info. Please simplify. Write a sequential code with handling one after other.

pkg/telemetrymetadata/metadata.go
srikanthccv: Why is there a `sb.Or`?

pkg/telemetrymetadata/metadata.go
```diff
@@ -1768,6 +1816,124 @@ func (t *telemetryMetaStore) fetchMeterSourceMetricsTemporality(ctx context.Cont
...
+	for _, selector := range selectors {
+		key := telemetrytypes.GetEvolutionMetadataUniqueKey(selector)
+		if _, ok := deduplicatedSelectors[key]; !ok {
+			deduplicatedSelectors[key] = selector
+		}
+	}
+	deduplicatedSelectorsList := make([]*telemetrytypes.EvolutionSelector, 0, len(deduplicatedSelectors))
+	for _, selector := range deduplicatedSelectors {
+		deduplicatedSelectorsList = append(deduplicatedSelectorsList, selector)
+	}
```
srikanthccv: Do we need this deduplication here? What happens if we remove this deduplication logic?

pkg/types/telemetrytypes/evolution_metadata.go
```diff
@@ -0,0 +1,44 @@
+package telemetrytypes
+
+import (
+	"context"
+	"time"
+
+	"github.com/SigNoz/signoz/pkg/valuer"
+)
+
+type EvolutionEntry struct {
```
srikanthccv, line 7: This is json serialised for cache but it doesn't have the explicit json tags, while not having the json tags works but it's a better idea to have tags if there is going to be json marshal of the structure.

pkg/types/telemetrytypes/evolution_metadata.go
```diff
@@ -0,0 +1,44 @@
...
+
+type EvolutionSelector struct {
+	Signal       Signal
+	FieldContext FieldContext
+	FieldName    string
+}
+
+type ColumnEvolutionMetadataStore interface {
+	Get(ctx context.Context, orgId valuer.UUID, keyName string) []*EvolutionEntry
+}
```
srikanthccv: This should be removed?

pkg/types/telemetrytypes/evolution_metadata.go
```diff
@@ -0,0 +1,44 @@
...
+func GetEvolutionMetadataUniqueKey(selector *EvolutionSelector) string {
+	return selector.Signal.StringValue() + ":" + selector.FieldContext.StringValue() + ":" + selector.FieldName
+}
+
+func GetEvolutionFromEvolutionsMap(key *TelemetryFieldKey, evolutions map[string][]*EvolutionEntry) []*EvolutionEntry {
+	var keyEvolutions []*EvolutionEntry
+	uniqueKey := GetEvolutionMetadataUniqueKey(&EvolutionSelector{
+		Signal:       key.Signal,
+		FieldContext: key.FieldContext,
+		FieldName:    "__all__",
```
srikanthccv: Why is this hard-coded to `__all__`. I get that it works for your case but it should be generic.

pkg/telemetrytraces/field_mapper.go
```diff
@@ -169,45 +169,46 @@ func NewFieldMapper() *defaultFieldMapper {
...
 func (m *defaultFieldMapper) getColumn(
 	_ context.Context,
+	_, _ uint64,
 	key *telemetrytypes.TelemetryFieldKey,
-) (*schema.Column, error) {
+) ([]*schema.Column, error) {
 	switch key.FieldContext {
 	case telemetrytypes.FieldContextResource:
-		return indexV3Columns["resource"], nil
+		return []*schema.Column{indexV3Columns["resource"]}, nil
```
srikanthccv, line 177: Discussed the problem with using the `resource` a single column for the cases where the retention is beyond the recent json resource rollout. Please create a ticket and get it addressed immediately follow up to this PR.

pkg/telemetrylogs/field_mapper.go
```diff
@@ -68,137 +69,279 @@ func NewFieldMapper() qbtypes.FieldMapper {
 	return &fieldMapper{}
 }
 
-func (m *fieldMapper) getColumn(_ context.Context, key *telemetrytypes.TelemetryFieldKey) (*schema.Column, error) {
+func (m *fieldMapper) getColumn(_ context.Context, key *telemetrytypes.TelemetryFieldKey) ([]*schema.Column, error) {
 	switch key.FieldContext {
 	case telemetrytypes.FieldContextResource:
-		return logsV2Columns["resource"], nil
+		columns := []*schema.Column{logsV2Columns["resource"], logsV2Columns["resources_string"]}
```
srikanthccv: Why is the order reversed here? Shouldn't the 0th entry be `resources_string`?

pkg/telemetrylogs/field_mapper.go
```diff
@@ -68,137 +69,279 @@ func NewFieldMapper() qbtypes.FieldMapper {
...
 			// check if the key has body JSON search
 			if strings.HasPrefix(key.Name, telemetrytypes.BodyJSONStringSearchPrefix) {
 				// Use body_json if feature flag is enabled and we have a body condition builder
 				if querybuilder.BodyJSONQueryEnabled {
-					return logsV2Columns[LogsV2BodyJSONColumn], nil
+					return []*schema.Column{logsV2Columns[LogsV2BodyJSONColumn]}, nil
 				}
 				// Fall back to legacy body column
-				return logsV2Columns["body"], nil
+				return []*schema.Column{logsV2Columns["body"]}, nil
```
srikanthccv, line 123: Just a note, unless the data is backfilled for then entirety of the retention period, a single column (i.e either old or new) shouldn't be returned from here as the data is present in both old and new.

pkg/telemetrylogs/field_mapper.go
```diff
@@ -68,137 +69,279 @@ func NewFieldMapper() qbtypes.FieldMapper {
...
+		return []*schema.Column{col}, nil
 	}
 
 	return nil, qbtypes.ErrColumnNotFound
 }
 
-func (m *fieldMapper) FieldFor(ctx context.Context, key *telemetrytypes.TelemetryFieldKey) (string, error) {
-	column, err := m.getColumn(ctx, key)
+// selectEvolutionsForColumns selects the appropriate evolution entries for each column based on the time range.
+// Assumes: evolutions are sorted and there's exactly one evolution per column.
```
srikanthccv: Why is it assumed that there is only one evolution for column? Or the assumption phrased incorrectly here?

pkg/telemetrylogs/field_mapper.go
srikanthccv, line 1: The diff on this file is non-trivial, pulling it locally and reviewing.

srikanthccv:
@nityanandagohain please resolve the conflicts. Or maybe it's a good idea to close the PR and create a one. Either way, let's try to get this change merged.

pkg/telemetrylogs/field_mapper.go
nityanandagohain: It didn't matter in the field_for as the final string  dependent on the evolutions, but keeping it in order.

pkg/telemetrylogs/condition_builder.go
```diff
@@ -228,34 +245,36 @@ func (c *conditionBuilder) conditionFor(
...
 		// skip adding exists filter for intrinsic fields
 		// with an exception for body json search
-		field, _ := c.fm.FieldFor(ctx, key)
+		field, _ := c.fm.FieldFor(ctx, startNs, endNs, key)
 		if slices.Contains(maps.Keys(IntrinsicFields), field) && key.FieldContext != telemetrytypes.FieldContextBody {
 			return condition, nil
 		}
 
-		existsCondition, err := c.conditionFor(ctx, key, qbtypes.FilterOperatorExists, nil, sb)
+		existsCondition, err := c.conditionFor(ctx, startNs, endNs, key, qbtypes.FilterOperatorExists, nil, sb)
```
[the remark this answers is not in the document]
nityanandagohain: We don't return empty string anymore.

pkg/telemetrymetadata/metadata.go
nityanandagohain: Not needed, the DB will take care of it.

pkg/telemetrymetadata/metadata.go
nityanandagohain: yeah these are not required.

pkg/types/telemetrytypes/evolution_metadata.go
nityanandagohain: yeah making all these generic.

pkg/telemetrymetadata/metadata.go
nityanandagohain: yeah this was unnecessary.

pkg/telemetrytraces/field_mapper.go
nityanandagohain: Yeah traces will still use the current multiIf, will address it in the next PR for traces.

pkg/telemetrylogs/condition_builder.go
nityanandagohain: As of now there will be one column, but this can be taken care in the next PR where we add the changes for JSON which will require some more changes to build the statements for normal vs json promoted differently and then combine them if required based on the evolutions

pkg/telemetrylogs/field_mapper.go
nityanandagohain: Yes can be taken up in the next PR as nothing changes for body since the evolutions table will be empty for it body context.

pkg/telemetrylogs/field_mapper.go
```diff
@@ -68,132 +69,269 @@ func NewFieldMapper() qbtypes.FieldMapper {
...
+
+				if key.KeyNameContainsArray() && !key.JSONDataType.IsArray {
+					return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "FieldFor not supported for nested fields; only supported for flat paths (e.g. body.status.detail) and paths of Array type: %s(%s)", key.Name, key.FieldDataType)
+				}
+				expr, err := m.buildFieldForJSON(key)
+				if err != nil {
+					return "", err
+				}
+
+				exprs = append(exprs, expr)
```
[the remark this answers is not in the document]
nityanandagohain: As of now there will be single column for body so we don't need an exists condition and this will be addressed in the next PR when we will add it.

pkg/telemetrylogs/field_mapper.go
nityanandagohain: ~~Updated the comment, as of now it handle duplicates, but if its an actual evolution with new version then it is not handled, will address in the next PR to get this moving forward. Let me know if you think otherwise.~~

pkg/telemetrylogs/filter_expr_logs_test.go
nityanandagohain: added evolutions and corrected it, previously it worked because even though I didn't add evolutions the order of columns in getColumn was such that it gave correct results.

We are passing ` [release-5m, release+5m` to make sure changes are minimal in this file and all the resource keys consider both the column which was the default case previously  but now it's due to evolutions.

While the field_mapper test takes care of different cases of evolutions.

I am okay with updating time-range where just new evolution is considered, but can be in the next PR as well.

pkg/telemetrylogs/field_mapper.go
```diff
@@ -68,132 +70,278 @@ func NewFieldMapper() qbtypes.FieldMapper {
...
+				return "", errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "only resource/body context fields are supported for json columns, got %s", key.FieldContext.String)
+			}
+
+		case schema.ColumnTypeEnumLowCardinality:
+			switch elementType := column.Type.(schema.LowCardinalityColumnType).ElementType; elementType.GetType() {
+			case schema.ColumnTypeEnumString:
+				exprs = append(exprs, column.Name)
+			default:
+				return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "exists operator is not supported for low cardinality column type %s", elementType)
+			}
```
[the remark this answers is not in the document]
nityanandagohain: https://github.com/SigNoz/signoz/pull/9669#discussion_r2763145124

Already mentioned It will be handled later in a separate issue and not a concern right now.

pkg/telemetrylogs/field_mapper.go
nityanandagohain: Added support for version which will support actual duplicate evolution with a new version.

nityanandagohain:
@srikanthccv bumping this for review, resolve the remaining conflict at once.

pkg/querybuilder/resourcefilter/condition_builder.go
```diff
@@ -60,15 +60,17 @@ func (b *defaultConditionBuilder) ConditionFor(
...
 	if err != nil {
 		return "", err
 	}
+	// resource evolution on main table doesn't affect this as we not changing the resource column in the resource fingerprint table.
+	column := columns[0]
```
tushar-signoz: Still would be a good idea to check for length and throw internal Error

pkg/querybuilder/collision.go
```diff
@@ -205,6 +206,7 @@ func AdjustKey(key *telemetrytypes.TelemetryFieldKey, keys map[string][]*telemet
 		key.Indexes = matchingKey.Indexes
 		key.Materialized = matchingKey.Materialized
 		key.JSONPlan = matchingKey.JSONPlan
+		key.Evolutions = matchingKey.Evolutions
```
tushar-signoz: There is a method for this, `key.Override()`

pkg/querybuilder/collision.go
```diff
@@ -153,6 +153,7 @@ func AdjustKey(key *telemetrytypes.TelemetryFieldKey, keys map[string][]*telemet
 			key.Indexes = intrinsicOrCalculatedField.Indexes
 			key.Materialized = intrinsicOrCalculatedField.Materialized
 			key.JSONPlan = intrinsicOrCalculatedField.JSONPlan
+			key.Evolutions = intrinsicOrCalculatedField.Evolutions
```
tushar-signoz: There is a method key.Update or key.Override. 
Can you use that method here and also the update the method to use Evolutions

pkg/querybuilder/resourcefilter/field_mapper.go
```diff
@@ -27,44 +27,48 @@ func NewFieldMapper() *defaultFieldMapper {
...
 	key *telemetrytypes.TelemetryFieldKey,
 ) (string, error) {
-	column, err := m.getColumn(ctx, key)
+	columns, err := m.getColumn(ctx, tsStart, tsEnd, key)
 	if err != nil {
 		return "", err
 	}
 	if key.FieldContext == telemetrytypes.FieldContextResource {
-		return fmt.Sprintf("simpleJSONExtractString(%s, '%s')", column.Name, key.Name), nil
+		return fmt.Sprintf("simpleJSONExtractString(%s, '%s')", columns[0].Name, key.Name), nil
```
tushar-signoz, line 60: Ideally this should return a multi If expression if len of columns > 1. even though it is given with the current situation that it is going to be a single column.

pkg/querybuilder/resourcefilter/field_mapper.go
```diff
@@ -27,44 +27,48 @@ func NewFieldMapper() *defaultFieldMapper {
...
 }
 
 func (m *defaultFieldMapper) ColumnExpressionFor(
 	ctx context.Context,
+	tsStart, tsEnd uint64,
 	key *telemetrytypes.TelemetryFieldKey,
 	_ map[string][]*telemetrytypes.TelemetryFieldKey,
 ) (string, error) {
-	colName, err := m.FieldFor(ctx, key)
+	colName, err := m.FieldFor(ctx, tsStart, tsEnd, key)
```
tushar-signoz: Can we rename colName to fieldExpression or fieldName as it is no longer a column

pkg/querybuilder/collision.go
tushar-signoz: There is a method for this

pkg/types/telemetrytypes/evolution_metadata.go
```diff
@@ -0,0 +1,25 @@
...
+
+type EvolutionSelector struct {
+	Signal       Signal
+	FieldContext FieldContext
+	FieldName    string
+}
+
+func GetEvolutionMetadataUniqueKey(selector *EvolutionSelector) string {
+	return selector.Signal.StringValue() + ":" + selector.FieldContext.StringValue() + ":" + selector.FieldName
+}
```
tushar-signoz: nit: this can be converted to selector.QualifiedName() and use ';' as separator.

pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
```diff
@@ -95,9 +97,51 @@ func (m *MockMetadataStore) GetKeysMulti(ctx context.Context, fieldKeySelectors
...
+				FieldName:    "__all__",
+			}
+
+			if keyEvolutions, ok := evolutions[telemetrytypes.GetEvolutionMetadataUniqueKey(selector)]; ok {
+				keys[i].Evolutions = keyEvolutions
+			}
+
+			selector.FieldName = key.Name
+			if keyEvolutions, ok := evolutions[telemetrytypes.GetEvolutionMetadataUniqueKey(selector)]; ok {
+				keys[i].Evolutions = keyEvolutions
```
tushar-signoz: Has this case been considered that
column 1 is changed to column 2 and key is moved from column 2 to column 3

pkg/querybuilder/collision.go
tushar-signoz: There's an override method defined on keys

pkg/telemetrylogs/field_mapper.go
```diff
@@ -61,137 +64,280 @@ var (
...
-			return m.buildFieldForJSON(key)
-		default:
-			return "", errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "only resource/body context fields are supported for json columns, got %s", key.FieldContext.String)
+	// Sort by ReleaseTime descending (newest first)
+	for i := 0; i < len(pairs)-1; i++ {
+		for j := i + 1; j < len(pairs); j++ {
+			if pairs[i].evolution.ReleaseTime.Before(pairs[j].evolution.ReleaseTime) {
+				pairs[i], pairs[j] = pairs[j], pairs[i]
+			}
 		}
```
tushar-signoz: slices.SortFunc

pkg/querybuilder/resourcefilter/condition_builder.go
srikanthccv: nit
```suggestion
	// resource evolution on main table doesn't affect this
	// as we have not changed the resource column in the resource fingerprint table.
	column := columns[0]
```

pkg/types/telemetrytypes/evolution_metadata.go
```diff
@@ -0,0 +1,25 @@
...
+	Version      uint32       `json:"version"`
+}
+
+type EvolutionSelector struct {
+	Signal       Signal
+	FieldContext FieldContext
+	FieldName    string
+}
+
+func GetEvolutionMetadataUniqueKey(selector *EvolutionSelector) string {
```
srikanthccv: question: will there ever be conflict with the unique key generated here?

context: some class of bugs arise because the assumption don't hold eventually and there are no things in place to ensure the assumption remains hold true.

pkg/telemetrymetadata/metadata.go
```diff
@@ -563,9 +569,48 @@ func (t *telemetryMetaStore) getLogsKeys(ctx context.Context, fieldKeySelectors
...
+	if err != nil {
+		return nil, false, err
+	}
+	for i, key := range keys {
+		// first check if there is evolutions that with field name as __all__
+		// then check for specific field name
+		selector := &telemetrytypes.EvolutionSelector{
+			Signal:       key.Signal,
+			FieldContext: key.FieldContext,
+			FieldName:    "__all__",
```
srikanthccv: `__all__` is detail that should be contained just within the `GetColumnEvolutionMetadataMulti`. Move this `__all__` logic into `GetColumnEvolutionMetadataMulti`. The return values of the `GetColumnEvolutionMetadataMulti` should be usable right away and downstream shouldn't need to change anything.

pkg/telemetrymetadata/metadata.go
```diff
@@ -1764,6 +1810,103 @@ func (t *telemetryMetaStore) fetchMeterSourceMetricsTemporality(ctx context.Cont
...
+			clause = sb.E("field_context", selector.FieldContext)
+		}
+
+		if selector.FieldName != "" {
+			clause = sb.And(clause,
+				sb.Or(sb.E("field_name", selector.FieldName), sb.E("field_name", "__all__")),
+			)
+		} else {
+			clause = sb.And(clause, sb.E("field_name", "__all__"))
+		}
```
srikanthccv: Is this needed, when will there be an empty field name for the evolution selector?

pkg/telemetrymetadata/metadata_test.go
```diff
@@ -384,3 +386,386 @@ func TestGetMetricFieldValuesIntrinsicBoolReturnsEmpty(t *testing.T) {
...
+	mock.ExpectQuery(clickHouseQueryPatternWithoutFieldName).WithArgs(telemetrytypes.SignalLogs, telemetrytypes.FieldContextResource, "__all__").WillReturnError(assert.AnError)
+
+	metadata := newTestTelemetryMetaStoreTestHelper(telemetryStore)
+	selector := &telemetrytypes.EvolutionSelector{
+		Signal:       telemetrytypes.SignalLogs,
+		FieldContext: telemetrytypes.FieldContextResource,
+	}
+	_, err := metadata.GetColumnEvolutionMetadataMulti(ctx, []*telemetrytypes.EvolutionSelector{selector})
+	require.Error(t, err)
+}
```
srikanthccv: Let's remove this test. It's doesn't really help us with anything meaningful.

pkg/telemetrymetadata/metadata_test.go
srikanthccv, line 1: @nityanandagohain the tests added here are not very useful. If the `metadata.GetColumnEvolutionMetadataMulti` had any complex logic, it would make sense to have some tests but right now it's really a simple logic to prepare and query the data and that's not something you would want to assert in tests unless there is a good reason. To be precise, one good rule of thumb is does refactoring internal details force you to touch tests, then they are bad tests. Imagine, you change the query which is fully functional but now you would have to change the tests as the expect a pattern.

I understand some existing tests slipped that may not be adding any value, and we should get rid of them as well but tests here are a maintenance burden than any substance. 

The coding assistant are very eager to write such tests and we should always look out for the maintenance of the tests and value they add.

Let me know what you think.

pkg/telemetrylogs/stmt_builder_test.go
```diff
@@ -189,14 +206,29 @@ func TestStatementBuilderTimeSeries(t *testing.T) {
...
+
+	// for each key of resource attribute add evolution metadata
+	for i, telemetryKeys := range keysMap {
+		for j, telemetryKey := range telemetryKeys {
+			if telemetryKey.FieldContext == telemetrytypes.FieldContextResource {
+				keysMap[i][j].Signal = telemetrytypes.SignalLogs
+				keysMap[i][j].Evolutions = mockEvolutionData(releaseTime)
+			}
+		}
+	}
```
srikanthccv: Let's define a `buildCompleteFieldKeyMap(releaseTime)` that returns what you are doing here. I see you find yourself populating the `Evolutions` in several test files.

pkg/telemetrylogs/filter_expr_logs_test.go
```diff
@@ -15,19 +17,33 @@ import (
...
 	cb := NewConditionBuilder(fm)
 
 	// Define a comprehensive set of field keys to support all test cases
 	keys := buildCompleteFieldKeyMap()
 
+	// for each key of resource attribute add evolution metadata
+	for i, telemetryKeys := range keys {
+		for j, telemetryKey := range telemetryKeys {
+			if telemetryKey.FieldContext == telemetrytypes.FieldContextResource {
+				keys[i][j].Evolutions = mockEvolutionData(releaseTime)
```
srikanthccv: ^

srikanthccv:
@tushar-signoz, you had some reservations about the change here. Please share written thoughts so we can evaluate if this change should go through or needs a revisit.

pkg/telemetrylogs/condition_builder.go
```diff
@@ -174,6 +178,31 @@ func (c *conditionBuilder) conditionFor(
...
+			}
+
+			if len(newColumns) == 0 {
+				return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "no valid evolution found for field %s in the given time range", key.Name)
+			}
+
+			// This mean tblFieldName is with multiIf, we just need to do a null check.
+			if len(newColumns) > 1 {
+				if operator == qbtypes.FilterOperatorExists {
+					return sb.IsNotNull(tblFieldName), nil
```
srikanthccv: I forgot the context, can you please help me understand why do we have this change here?

pkg/telemetrymetadata/metadata_test.go
nityanandagohain: Yeah, it's kind of useless I agreee. removing it.

pkg/telemetrymetadata/metadata.go
nityanandagohain: For this I will have to pass the keys to the `GetColumnEvolutionMetadataMulti` for the keys to be modified. Do you see any issue with that? or you mean something else.

pkg/telemetrymetadata/metadata.go
nityanandagohain: Yeah won't really happen.

pkg/types/telemetrytypes/evolution_metadata.go
nityanandagohain: On top of my mind I think datatype, but let me dig more here.

pkg/querybuilder/resourcefilter/field_mapper.go
nityanandagohain: I would want to keep it as it is( or address in a separate one) unless a very strong opinion on why we should add the default multiIf.

pkg/querybuilder/resourcefilter/field_mapper.go
nityanandagohain: yeah makes sense

pkg/querybuilder/resourcefilter/field_mapper.go
nityanandagohain: I see you suggested because we are renaming it to `fieldExpression` . But I think this can be as it is for now.

pkg/types/telemetrytypes/evolution_metadata.go
nityanandagohain: Thanks for the suggestion, makes sense.

pkg/querybuilder/collision.go
nityanandagohain: Here keeping the key.Name and the using the override.

pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
nityanandagohain: Didn't get you, can you explain a bit more here.

pkg/telemetrymetadata/metadata.go
srikanthccv: You already have access to keys? Can you share a snippet of what you are referring to?

nityanandagohain: So in this code https://github.com/SigNoz/signoz/blob/e1d73862c7e02b9b82252abaa3ebfd34a99b23d4/pkg/telemetrymetadata/metadata.go#L579 I am updating the keys based on the evolutions we got. So my question is adding the evolutions to keys should be a logic here or the metadata should do it by taking it in params.

nityanandagohain: Also making this a private function as I don't see it getting used otside of this. The key functions should returns fields with correct evolution/

pkg/telemetrylogs/stmt_builder_test.go
nityanandagohain: thanks makes sense.

pkg/telemetrylogs/condition_builder.go
```diff
@@ -174,6 +178,31 @@ func (c *conditionBuilder) conditionFor(
...
+				} else {
+					return sb.IsNull(tblFieldName), nil
+				}
+			}
+
+			// otherwise we have to find the correct exist operator based on the column type
+			column = newColumns[0]
+		}
```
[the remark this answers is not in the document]
nityanandagohain: if more than one is present then it is already taken care of in the above code where we return sb.IsNotNull(fieldExpression)

when exactly one column we do ahave support for schema.ColumnTypeEnumJSON

pkg/telemetrylogs/condition_builder.go
nityanandagohain: so when we know coulmns >0 that means there are multiple evolutions. Which also means that field_for returns a fieldExpression with multiIf in that case. So with that we just need to check null or not null for exists and not exists where a multiIf is present in the expression

pkg/telemetrylogs/field_mapper.go
```diff
@@ -61,137 +65,285 @@ var (
...
-			return fmt.Sprintf("%s['%s']", column.Name, key.Name), nil
-		default:
-			return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "exists operator is not supported for map column type %s", valueType)
+
+			switch valueType := column.Type.(schema.MapColumnType).ValueType; valueType.GetType() {
+			case schema.ColumnTypeEnumString, schema.ColumnTypeEnumBool, schema.ColumnTypeEnumFloat64:
+				// a key could have been materialized, if so return the materialized column name
+				if key.Materialized {
+					exprs = append(exprs, telemetrytypes.FieldKeyToMaterializedColumnName(key))
+					existExpr = append(existExpr, fmt.Sprintf("%s_exists = true", telemetrytypes.FieldKeyToMaterializedColumnName(key)))
```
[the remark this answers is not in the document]
nityanandagohain: yeah will change.

pkg/telemetrymetadata/metadata.go
nityanandagohain: Made the change.

pkg/types/telemetrytypes/evolution_metadata.go
nityanandagohain: getting clarity on JSON if datatype will be something that will be required., apart from that this looks good.

pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
nityanandagohain: Resolving this to move ahead

Issue #9701, opened 2025-11-27
TTL: TTL for logs not actually getting applied


nityanandagohain:
In the new shema for custom retention, updating the default of retention days in distributed table is missing.

PR #9702, opened 2025-11-27
fix: add defualt ttl to distributed_table

nityanandagohain:
When we just update the retention, the value of the local table gets updated but since it doen't get updated in the distributed table, it doesn't actually work.

It works when the user actually sets a custom value for retention since there the function gets evaluated.

Fixes https://github.com/SigNoz/signoz/issues/9701

---
We will need more robust integration test to check these kind of issues ( not covering in this PR)

---

> [!NOTE]
> Also update `_retention_days` and optional `_retention_days_cold` defaults on distributed logs tables when setting custom TTL.
> 
> - **Logs retention (custom TTL v2)**:
>   - Apply `_retention_days` default to distributed tables `logs_v2` and `logs_resource_v2` alongside local tables.
>   - When cold storage is configured, also set `_retention_days_cold` default on distributed tables.
>   - TTL alteration queries remain on local tables; distributed tables receive column default updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2fb6f798c7a760d0a2ad68d3d7839c08ff56ff3f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/query-service/app/clickhouseReader/reader.go
+++ b/pkg/query-service/app/clickhouseReader/reader.go
@@ -1663,6 +1663,10 @@ func (r *ClickHouseReader) SetTTLV2(ctx context.Context, orgID string, params *m
 		getLocalTableName(r.logsDB + "." + r.logsAttributeKeys),
 		getLocalTableName(r.logsDB + "." + r.logsResourceKeys),
 	}
+	distributedTableNames := []string{
+		r.logsDB + "." + r.logsTableV2,
+		r.logsDB + "." + r.logsResourceTableV2,
+	}
 
 	for _, tableName := range tableNames {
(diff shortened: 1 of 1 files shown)


tushar-signoz:
@nityanandagohain 
Integration tests will call the API to set the retention, sleep for few seconds and then will query clickhouse directly to fetch the retention. Add these tables in the list to be checked.
Here: https://github.com/SigNoz/signoz/blob/d01e6fc891daf491e7617ce9b868deb5aba5500d/tests/integration/src/ttl/a_ttl.py#L256

nityanandagohain:
So the best way will be to check the paritions in the integration tests, can you please pick up that in a  separete PR. @tushar-signoz 


https://github.com/SigNoz/engineering-pod/issues/3509

Issue #9647, opened 2025-11-21
Add tests for switching b/w tabs in explorer


srikanthccv:
I didn't find one. Can you please add a test for switching b/w tabs and ensure the search expression is correctly present? i.e., if the issue mentioned were to happen again for any reason, how would we know from the test https://github.com/SigNoz/signoz/pull/9569#pullrequestreview-3476697767?

_Originally posted by @srikanthccv in https://github.com/SigNoz/signoz/pull/9569#discussion_r2545373482_

ahrefabhi:
https://github.com/SigNoz/signoz/pull/9766

PR #9766, opened 2025-12-03
test: added test for testing codemirror state when switching tabs

ahrefabhi:
## 📄 Summary

---

## ✅ Changes

- Added test for codemirror context check when changing tabs

Closes https://github.com/SigNoz/signoz/issues/9647

### DIFF
--- a/frontend/src/components/QueryBuilderV2/utils.ts
+++ b/frontend/src/components/QueryBuilderV2/utils.ts
@@ -163,6 +163,10 @@ function formatSingleValueForFilter(
 		if (trimmed === 'true' || trimmed === 'false') {
 			return trimmed === 'true';
 		}
+
+		if (isQuoted(value)) {
+			return unquote(value);
+		}
 	}
 
 	// Return non-string values as-is, or string values that couldn't be converted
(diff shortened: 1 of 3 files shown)


frontend/src/container/LogExplorerQuerySection/LogExplorerQuerySection.test.tsx
srikanthccv, line 1: The view switch logic has been refactored recently. Can you update the tests to reflect that. For instance, take the following example

Image

This is very basic but as one switches from one view to another only selected parts of the query gets added and other get removed. Can we have the tests reflect and test that functionality. Let me know if this is clear or have any questions?

srikanthccv: The reason I mention is because the test here is only testing the search expression but rest of the parts are important as well to cover in tests.

AshwinBhatkal (approved):

frontend/src/container/LogExplorerQuerySection/LogExplorerQuerySection.test.tsx
ahrefabhi: added tests for other parts as well

YounixM (approved):

PR #9769, opened 2025-12-03
fix: restore click-to-copy for non-JSON body fields after #9657

ahmadshaheer:
## 📄 Summary
This PR restores the click-to-copy functionality for non-JSON body fields that was removed in PR #9657 when adding support for copying individual JSON tree nodes.

---

## 🔍 Related Issues

Closes https://github.com/SigNoz/engineering-pod/issues/3381

---

Before:

https://github.com/user-attachments/assets/a5098b7a-6d72-4aee-b33a-9cea950683a9

After:

https://github.com/user-attachments/assets/a6db4a00-ba2e-424c-aa8e-e766df7f1a48

---

## 📋 Checklist

- [x] Dev Review

### DIFF
--- a/frontend/src/container/SpanDetailsDrawer/SpanLogs/spanLogs.styles.scss
+++ b/frontend/src/container/SpanDetailsDrawer/SpanLogs/spanLogs.styles.scss
@@ -5,16 +5,20 @@
 	&-virtuoso {
 		background: rgba(171, 189, 255, 0.04);
 	}
-	&-list-container .logs-loading-skeleton {
+	&-list-container {
 		height: 100%;
-		border: 1px solid var(--bg-slate-500);
-		border-top: none;
-		color: var(--bg-vanilla-400);
-		display: flex;
-		flex-direction: column;
-		align-items: center;
-		justify-content: center;
-		padding: 8px 0;
+
+		.logs-loading-skeleton {
+			height: 100%;
+			border: 1px solid var(--bg-slate-500);
+			border-top: none;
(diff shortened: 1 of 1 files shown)


frontend/src/container/LogDetailedView/TableView/TableViewActions.tsx
```diff
@@ -80,23 +81,15 @@ const BodyContent: React.FC<{
 		);
 	}
 
-	if (record.field === 'body' && error) {
```
aks07, line 83: This code is required when JSON parsing for some reason fails. Why are we removing this? Is this related to copying non JSON body?

aks07 (changes requested):
Added a comment.

frontend/src/container/LogDetailedView/TableView/TableViewActions.tsx
ahmadshaheer: no, it's not related to copying non-json body. I guess it's better to remove it, since it's more useful to the user to see the json (even if we're unable to parse it) rather than an error message.

aks07: if we are unable to parse it, it throws an error and the whole page crashes. Hence this is in place

ahmadshaheer: oh, got it. thanks for catching this. reverting this change.

aks07 (approved):

PR #9824, opened 2025-12-17
fix: added fix for quick filter reset not working

ahrefabhi:
## 📄 Summary

---

## ✅ Changes

- Fix: This PR fixes quick filter resetting issue

## Tests:
- Added test for checkbox component where we test if filters are getting updated or not.
- Added test for QuerySearch component for testing if editorview.dispatch is getting called when we change the query expression.

closes https://github.com/SigNoz/engineering-pod/issues/3344


srikanthccv:
@ahrefabhi is writing tests for this.

PR #9825, opened 2025-12-17
chore: remove eval delay from safe rules

abhishekhugetech:
## 📄 Summary

This PR adds an evaluation-delay optimization for rule evaluation by computing evalDelay dynamically from each rule condition’s query + thresholds.

  Right now, rule evaluation applies a default `evalDelay` of 2 minutes to account for data that’s still “in motion” for the last 1–2 minutes.

  In many cases:

  1. Data for the current minute may still be on the client side getting pre-aggregated.
  2. Data for the previous minute may be in the pipeline or getting written to the database.

  This avoids false-positive alerts, but it also delays alerting unnecessarily when the data typically arrives in ~10–30s from sender → SigNoz DB.

  For example, for an alert with MatchType: `At least once`, CompareOperator: `above`, and Aggregation: `COUNT`, if the condition is true with partial data, it will remain true when the rest of the data arrives for the same duration, so evaluating without delay is safe and reduces detection latency.

[Notion doc for other such combination](https://www.notion.so/signoz/Necessity-of-Evaluation-Delay-2b5fcc6bcd198040a2c7f478867553ea)

---

## ✅ Changes

---

ex:

- `frontend`
- `backend`
- `devops`
- `bug`
- `enhancement`
- `ui`
- `test`

---

## 👥 Reviewers

- frontend / backend / devops

---

## 🧪 How to Test

1. ...
2. ...
3. ...

---

## 🔍 Related Issues

Closes #

---

---

## 📋 Checklist

- [ ] Dev Review

---


pkg/query-service/rules/eval_delay.go
```diff
@@ -0,0 +1,224 @@
...
+		// Group: Max, MaxIf
+		// Value can only go up or remain same.
+
+		if matchType == ruletypes.AtleastOnce || matchType == ruletypes.AllTheTimes {
+			if compareOp == ruletypes.ValueIsAbove || compareOp == ruletypes.ValueAboveOrEq {
+				return true
+			}
+		}
+
+	case metrictypes.TimeAggregationCount, metrictypes.TimeAggregationCountDistinct:
```
srikanthccv: Count distinct shouldn't be here right?

srikanthccv:
Will be addressed as part of https://github.com/SigNoz/engineering-pod/issues/3199

PR #9866, opened 2025-12-24
Feat: Calculate Services Details Using Span Metrics

nikhilmantri0902:
## 📄 Summary

Closes https://github.com/SigNoz/engineering-pod/issues/3479

---

## ✅ Changes

---

ex:

- `frontend`
- `backend`
- `devops`
- `bug`
- `enhancement`
- `ui`
- `test`

---

## 👥 Reviewers

- frontend / backend / devops

---

## 🧪 How to Test

1. ...
2. ...
3. ...

---

## 🔍 Related Issues

Closes #

---

---

## 📋 Checklist

- [ ] Dev Review

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> Enables span-metrics–based computation for services and operations, gated by a feature flag, and simplifies the frontend services list to use the new services API.
> 
> - **Backend (services module):** Integrates `flagger` and switches between traces vs span-metrics paths; adds span-metrics QueryRange builders for services, top ops, and entry-point ops; maps scalar results, handles single-bucket steps, and attaches top-level ops. Introduces synthetic metric-scope field `isTopLevelOperation` with condition support; wires flagger through `NewModules` and updates tests. Adds `LocalTopLevelOperationsTableName` constant.
> - **Frontend:** Home `ServiceMetrics` now fetches from `/api/v2/services` via `useQueryService`, removes legacy query-builder logic, and sorts by numeric `p99`; removes `TopOperationMetrics` and feature-gated rendering; `ServiceOverview` sets latency unit to `ms` when span metrics enabled; `ServiceApplication` always shows traces view.
> - **Dev/Config:** `make go-run-community` sets `SIGNOZ_FLAGGER_CONFIG_BOOLEAN_USE__SPAN__METRICS=true`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0193b4f9bfac2bb65793a24e541d485a5193b007. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


khanbilal732 (approved):

khanbilal732 (approved):

khanbilal732 (approved):
- [ ] 

- [ ]

srikanthccv:
@nikhilmantri0902 please resolve conflicts and make use of the new flagger integration for span metrics https://github.com/SigNoz/signoz/pull/9925.

khanbilal732 (approved):

pkg/telemetrymetrics/condition_builder.go
```diff
@@ -139,13 +141,52 @@ func (c *conditionBuilder) ConditionFor(
...
+	default:
+		return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "%s expects boolean value, got %T", MetricScopeFieldIsTopLevelOperation, value)
+	}
+	if !isTrue {
+		return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "%s can only be filtered with value 'true'", MetricScopeFieldIsTopLevelOperation)
+	}
+
+	startSec := int64(start / 1000)
+
+	// Note: Escape $$ to $$$$ to avoid sqlbuilder interpreting materialized $ signs
```
srikanthccv: Why is there this note here?

pkg/telemetrymetadata/metadata.go
```diff
@@ -698,6 +698,19 @@ func (t *telemetryMetaStore) getMetricsKeys(ctx context.Context, fieldKeySelecto
...
+		if selector.Name == "isTopLevelOperation" {
+			keys = append(keys, &telemetrytypes.TelemetryFieldKey{
+				Name:          "isTopLevelOperation",
+				Signal:        telemetrytypes.SignalMetrics,
+				FieldContext:  telemetrytypes.FieldContextUnspecified,
+				FieldDataType: telemetrytypes.FieldDataTypeBool,
+			})
+			break
+		}
+	}
```
srikanthccv: Why is this key always added here, regardless of any selection criteria?

pkg/modules/services/implservices/module.go
```diff
@@ -285,15 +372,233 @@ func (m *module) mapQueryRangeRespToServices(resp *qbtypes.QueryRangeResponse, s
...
+				GroupBy: groupByService,
+				Aggregations: []qbtypes.MetricAggregation{
+					{
+						MetricName:       "signoz_latency.sum",
+						Temporality:      metrictypes.Delta,
+						TimeAggregation:  metrictypes.TimeAggregationAvg,
+						SpaceAggregation: metrictypes.SpaceAggregationAvg,
+						ReduceTo:         qbtypes.ReduceToAvg,
+					},
+				},
```
srikanthccv: This is not an average latency

pkg/modules/services/implservices/module.go
```diff
@@ -285,15 +372,233 @@ func (m *module) mapQueryRangeRespToServices(resp *qbtypes.QueryRangeResponse, s
...
+	reqV5 := qbtypes.QueryRangeRequest{
+		Start:       startMs,
+		End:         endMs,
+		RequestType: qbtypes.RequestTypeScalar,
+		Variables:   variables,
+		CompositeQuery: qbtypes.CompositeQuery{
+			Queries: queries,
+		},
+		FormatOptions: &qbtypes.FormatOptions{
+			FormatTableResultForUI: true,
```
srikanthccv: Why is the `FormatTableResultForUI` set to true here when the request is prepared in backend code?

frontend/src/container/Home/Services/ServiceMetrics.tsx
srikanthccv, line 1: Please share the recording for the overview "latency" data when span metrics flag is enabled

pkg/modules/services/implservices/module.go
```diff
@@ -404,6 +709,550 @@ func (m *module) mapTopOpsQueryRangeResp(resp *qbtypes.QueryRangeResponse) []ser
...
+						TimeAggregation:  metrictypes.TimeAggregationIncrease,
+						SpaceAggregation: metrictypes.SpaceAggregationSum,
+						ReduceTo:         qbtypes.ReduceToSum,
+					},
+				},
+			},
+		},
+	}
+
+	reqV5 := qbtypes.QueryRangeRequest{
```
srikanthccv, line 834: Set noCache to true for all metrics based requsts prepared here.

pkg/telemetrymetrics/condition_builder.go
nikhilmantri0902: removed.

pkg/modules/services/implservices/module.go
nikhilmantri0902: added.

pkg/modules/services/implservices/module.go
nikhilmantri0902: removed and now all span metrics queries have the same pattern. including their response mapping functions.

pkg/modules/services/implservices/module.go
nikhilmantri0902: Correct this should have summed up and divided by total count represented by signoz_calls_total. But I am removing avg_latency altogether from here because we do not show it on UI today and its not needed actually.

pkg/modules/services/implservices/module.go
```diff
@@ -285,15 +372,233 @@ func (m *module) mapQueryRangeRespToServices(resp *qbtypes.QueryRangeResponse, s
...
+					{
+						MetricName:       "signoz_latency.sum",
+						Temporality:      metrictypes.Delta,
+						TimeAggregation:  metrictypes.TimeAggregationAvg,
+						SpaceAggregation: metrictypes.SpaceAggregationAvg,
+						ReduceTo:         qbtypes.ReduceToAvg,
+					},
+				},
+			},
+		},
```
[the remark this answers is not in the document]
nikhilmantri0902: removed altogether as not shown in UI

pkg/telemetrymetadata/metadata.go
nikhilmantri0902: I have now added it like this:
```
for _, selector := range fieldKeySelectors {
		for _, key := range telemetrymetrics.MetricScopeFieldDefinitions {
			if selector.Signal != telemetrytypes.SignalUnspecified && selector.Signal != telemetrytypes.SignalMetrics {
				continue
			}
			if selector.FieldContext != telemetrytypes.FieldContextUnspecified && selector.FieldContext != key.FieldContext {
				continue
			}

			if selector.FieldDataType != telemetrytypes.FieldDataTypeUnspecified && selector.FieldDataType != key.FieldDataType {
				continue
			}
			if matchesSelectorName(selector.Name, key.Name, selector.SelectorMatchType) {
				keys = append(keys, &key)
				break
			}
		}
	}
```

frontend/src/container/Home/Services/ServiceMetrics.tsx
nikhilmantri0902: Shared

nikhilmantri0902:
So we identified an issue while estimating percentile values for a time range as a scalar. 

We firstly specified a big step interval in querier ( 1 year ) so that data for each le bucket is summed up over the entire range correctly and attributed to a single timestamp. This raised an issue that internally in `querybuilder.AdjustedMetricTimeRange` function we adjust the start time to a very previous value because of the big step interval and also the aggregate tables used are for the widest range. This causes deviation from the user requested data. We were seeing different service names or operation names for example in their respective responses. 

We then went onto correct the stepInterval to be equal to the difference in end and start time. This worked better but there is a catch here:
https://clickhouse.com/docs/sql-reference/functions/date-time-functions#toStartOfInterval 

toStartOfInterval would produce more than one ts datapoints per group by entity because the calendar origins that are followed by toStartOfInterval are predefined. For example :
```
SELECT `toStartOfInterval(toDateTime('2023-01-15 14:30:00'), INTERVAL 1 MONTH)` 
```
gives result as : 
```
┌─toStartOfInt⋯alMonth(1))─┐
│               2023-01-01 │
└──────────────┘
```
and not 2022-12-15. This means that everytime an origin boundary is crossed (mid-month to mid-month, mid-week to mid-week ) you will get more than one ts data point for that. But this way the ultimate goal is compromised where we tried to get one datapoint per group by entity as the percentile value. We can go on to `reduceTo` these percentile values as we do for scalars, but that does not make sense as percentiles cannot be averaged, max-ed, min-ed, etc. because they lose their meaning. 

As a solution to this, we found out you can enforce an origin ( as the 3rd argument to toStartOfInterval ) which sets the calendar from this timestamp specified: 
```
toStartOfInterval(
      toDateTime(
        intDiv(unix_milli, 1000)
      ), 
      toIntervalSecond(604799), -- stepInterval
      toDateTime(intDiv(1767222678000, 1000)) --- this is the origin parameter.
    )
```
but here there is one caveat that the time range selected in the query should be all ahead of this origin timestamp. And the stepInterval has to be the exact difference between the requested start and end times, today in query builder specifying both affects the actual value of both inside the function querybuilder.AdjustedTimeRange so either u get 2 datapoints or your start time is less than the origin in which case the query errors out. 

We need to come up with a solution to this issue.

khanbilal732 (approved):

khanbilal732 (approved):

Issue #9282, opened 2025-10-05
Limit is still sent in the request payload after removing it


srikanthccv:
<img width="2896" height="1544" alt="Image" src="https://github.com/user-attachments/assets/722f8626-fa3d-4a15-b10e-11bebe930efe" />

I initial added a limit 10 to create the issue, then removed it show the presence of this bug.

## How to reproduce

- Create a query with the limit of 10
- Add order by asc on count
- Run the query
- Remove the limit and run again
- Observe the request payload

PR #9877, opened 2025-12-26
fix: added fix for limit still getting sent in payload even after removing

ahrefabhi:
## 📄 Summary

---

## ✅ Changes

- Fixes limit still getting sent in payload even after removing

---

## Cause

- In QB V5 we removed the use of pageSize but the code still uses it here and there, and on widget edit page we initialise the pageSize with `10`, which internally gets forwarded to the API as limit.
- Added a ticket to clean things up related to pageSize, will take that up in a new PR (Because that's a bigger change) https://github.com/SigNoz/engineering-pod/issues/3506

---

## Before

- The limit is still getting sent even if it's removed.

https://github.com/user-attachments/assets/99ff4b44-2df0-443e-8656-4608a3b6c428

## After
https://github.com/user-attachments/assets/bf2260b2-fe11-4820-890b-39d2e0b4ff02

Closes https://github.com/SigNoz/signoz/issues/9282

### DIFF
--- a/frontend/src/container/GridCardLayout/GridCard/index.tsx
+++ b/frontend/src/container/GridCardLayout/GridCard/index.tsx
@@ -167,7 +167,6 @@ function GridCardGraph({
 		}
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [updatedQuery]);
-	console.log('requestData', requestData);
 
 	const isLogsQuery = useMemo(
 		() =>


frontend/src/container/NewWidget/index.tsx
```diff
@@ -419,16 +415,12 @@ function NewWidget({
 	useEffect(() => {
 		if (stagedQuery) {
 			setIsLoadingPanelData(false);
-			const updatedStagedQuery = cloneDeep(stagedQuery);
-			if (updatedStagedQuery?.builder?.queryData?.[0]) {
-				updatedStagedQuery.builder.queryData[0].pageSize = 10;
```
AshwinBhatkal, line 424: Is this default limit of 10 enforced from the BE?

AshwinBhatkal: Ohh. Nvm. It is not required here

AshwinBhatkal (approved):

frontend/src/container/NewWidget/index.tsx
ahrefabhi: This is not required anymore

PR #9880, opened 2025-12-27
fix: fillZero usage in query builder and correct step values to ms

### LINKED ISSUE #9752: Populate start and end for the fill zero function
See the conversation here https://github.com/SigNoz/signoz/pull/9651. Even if the client sends `fillZero` it doesn't work.

niladrix719:
## 📄 Summary

- Fix fillZero usage in query builder
- Corrected step unit to ms
- Added integration tests for fillGaps and fillZero (logs, trances, metrics)

---

## ✅ Changes

- [x] Bug fix: Brief description

---

ex:

- `backend`
- `bug`
- `test`

---

## 👥 Reviewers

- backend

---

## 🧪 How to Test

1. Add fillZero in query builder functions list
2. Open metrics Explorer
3. Enter Metric name & add fillZero function
4. Run query and see response in network tab

---

## 🔍 Related Issues

Closes #9752 

---

https://github.com/user-attachments/assets/ee4cd137-561f-40e0-9168-7bd654aed153

---

## 📋 Checklist

- [x] Dev Review

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> **Fix step units to milliseconds and validate gap-filling behavior**
> 
> - Updates `funcFillZero` to align and iterate using `step` in ms (no extra ×1000) and returns fully filled series
> - Changes `StepIntervalForQuery` to return `Milliseconds()` for builder queries and PromQL `step`
> - Adjusts unit tests to use ms steps and verifies unaligned start/end behavior
> 
> **Integration tests and test utilities**
> 
> - Adds end-to-end tests for `fillGaps` and `fillZero` across logs, traces, and metrics (with and without `groupBy`, and within formulas)
> - Introduces `fixtures.metrics` and `src/querier/timeseries_utils.py` helpers; registers new fixture in `tests/integration/conftest.py`
> - Minor fixture improvements: optional timestamps defaults in logs/traces
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 52821ae8e84ea3f2b697ba4941f76cf6a3aca980. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/integration/src/querier/c_metrics.py
+++ b/tests/integration/src/querier/c_metrics.py
@@ -834,7 +834,7 @@ def test_metrics_fill_zero_formula_with_group_by(
                             "aggregations": [{
                                 "metricName": metric_name_a,
                                 "temporality": "cumulative",
-                                "timeAggregation": "sum",
+                                "timeAggregation": "increase",
                                 "spaceAggregation": "sum"
                             }],
                             "stepInterval": 60,
@@ -850,7 +850,7 @@ def test_metrics_fill_zero_formula_with_group_by(
                             "aggregations": [{
                                 "metricName": metric_name_b,
                                 "temporality": "cumulative",
-                                "timeAggregation": "sum",
+                                "timeAggregation": "increase",
                                 "spaceAggregation": "sum"
                             }],
                             "stepInterval": 60,


srikanthccv:
can you help me understand what was the actual issue

- the comment is misleading, https://github.com/SigNoz/signoz/issues/9752#issuecomment-3650695900, but the `step * 1000` handles the seconds part
- there is code to handle the zero args (which is how the client sends it, so that is also taken care of)

Then why was it not working?

niladrix719:
@srikanthccv The main issue was, step was being sent as a ms value params to `prepareFillZeroArgsWithStep` -
https://github.com/SigNoz/signoz/blob/2dc6febb3826c7cc599e83b3a7e912dbdb62c446/pkg/querier/postprocess.go#L173-L174

so this loop inside `funcFillZero` func
https://github.com/SigNoz/signoz/blob/2dc6febb3826c7cc599e83b3a7e912dbdb62c446/pkg/types/querybuildertypes/querybuildertypesv5/functions.go#L389

will have step value much higher(ms * 1000) that it will skip almost the entire time range in one iteration and exit immediately

pkg/querier/postprocess.go
```diff
@@ -936,7 +936,7 @@ func gcd(a, b int64) int64 {
 func (q *querier) prepareFillZeroArgsWithStep(functions []qbtypes.Function, req *qbtypes.QueryRangeRequest, step int64) []qbtypes.Function {
 	needsCopy := false
 	for _, fn := range functions {
-		if fn.Name == qbtypes.FunctionNameFillZero && len(fn.Args) == 0 {
+		if fn.Name == qbtypes.FunctionNameFillZero && len(fn.Args) < 3 {
```
srikanthccv: Let's assume the incoming args either present fully or not present.

pkg/querier/postprocess.go
```diff
@@ -950,14 +950,25 @@ func (q *querier) prepareFillZeroArgsWithStep(functions []qbtypes.Function, req
 	copy(updatedFunctions, functions)
 
 	for i, fn := range updatedFunctions {
-		if fn.Name == qbtypes.FunctionNameFillZero && len(fn.Args) == 0 {
```
srikanthccv, line 953: Following the above, let's keep this unmodified, if the args are absent completely, we fill them otherwise expect the incoming payload to contain data.

pkg/types/querybuildertypes/querybuildertypesv5/functions.go
```diff
@@ -91,25 +91,77 @@ func ApplyFunction(fn Function, result *TimeSeries) *TimeSeries {
...
-		if err != nil {
-			return result
+		return funcFillZero(result, start, end, step)
+	}
+	return result
+}
+
+// resolveFillZeroParams extracts start, end, and step for fillZero.
+// If start or end are missing, it falls back to the series' min and max timestamps.
+func resolveFillZeroParams(args []FunctionArg, series *TimeSeries) (int64, int64, int64, bool) {
```
srikanthccv: This should never happen because every query should have the start and end time stamp (which then means steps will also be created), otherwise it's not possible to run the query.

srikanthccv (commented):
thanks, some comments

niladrix719:
Thanks for the comments, I’ve addressed them

srikanthccv:
thank you @niladrix719, let's add some integration tests for this change, I will share the representative cases to cover to catch any future regressions.

niladrix719:
sure, sounds good

srikanthccv:
@niladrix719 

1. /v5/query_range

- with fillGaps enabled (without group by + one query) https://github.com/SigNoz/signoz/blob/b9f9c00da5655aca5dcda439f3cb819bbc57b735/pkg/types/querybuildertypes/querybuildertypesv5/req.go#L404
   - for metrics
   - traces
   - logs

- with fillGaps enabled (with group by + one query) https://github.com/SigNoz/signoz/blob/b9f9c00da5655aca5dcda439f3cb819bbc57b735/pkg/types/querybuildertypes/querybuildertypesv5/req.go#L404
   - for metrics
   - traces
   - logs

- with fillGaps enabled (without group by + 2 query (disabled) + formula) https://github.com/SigNoz/signoz/blob/b9f9c00da5655aca5dcda439f3cb819bbc57b735/pkg/types/querybuildertypes/querybuildertypesv5/req.go#L404
   - for metrics
   - traces
   - logs

- with fillGaps enabled (with group by + 2 query (disabled) + formula) https://github.com/SigNoz/signoz/blob/b9f9c00da5655aca5dcda439f3cb819bbc57b735/pkg/types/querybuildertypes/querybuildertypesv5/req.go#L404
   - for metrics
   - traces
   - logs
   

2. /v5/query_range - exactly same as above but by passing the "fillZero" function in the functions list instead of setting `fillGaps`

Looks intimidating, but should be fairly straightforward, let me know if you need any help. Thank you!

niladrix719:
Looks good, one question I don’t see any metrics file under `tests/integration/src/fixtures` folder, should I create one

srikanthccv:
Yes, let's create one. Look at the exporter code to see what tables to write data to https://github.com/SigNoz/signoz-otel-collector/blob/main/exporter/signozclickhousemetrics/exporter.go For this issue, we can scope the metric data to counter/sum type only https://opentelemetry.io/docs/specs/otel/metrics/data-model/#sums. Rest of the metric types we can separately add in another PR.

Let me know if you need any help. If you are not on our community slack, please join here http://signoz.io/slack. You can ping me there for any questions for quick response. I should be available anytime. thank you!

niladrix719:
ready for review

tests/integration/src/querier/a_logs.py
```diff
@@ -1333,3 +1333,668 @@ def test_datatype_collision(
...
+
+    values = series[0]["values"]
+    # With 5 minute range and 60s step, we expect ~5-6 data points
+    assert len(values) >= 5, f"Expected at least 5 values for gap filling, got {len(values)}"
+
+    # Verify gaps are filled with zeros
+    zero_count = sum(1 for v in values if v["value"] == 0)
+    assert zero_count >= 1, "Expected at least one zero-filled gap"
+
+
```
srikanthccv: This is good, let's also verify that the value for the minutes 1 and 3 are non-zero and rest are zero.

tests/integration/src/querier/a_logs.py
```diff
@@ -1333,3 +1333,668 @@ def test_datatype_collision(
...
+    assert len(aggregations) == 1
+
+    series = aggregations[0]["series"]
+    assert len(series) == 2, "Expected 2 series for 2 service groups"
+
+    # Verify each series has gap-filled values
+    for s in series:
+        values = s["values"]
+        assert len(values) >= 5, f"Expected at least 5 gap-filled values, got {len(values)}"
+
```
srikanthccv: Same, let's make the tests bit more tight by asserting the actual values seen at minute step intervals.

tests/integration/src/querier/a_logs.py
```diff
@@ -1333,3 +1333,668 @@ def test_datatype_collision(
...
+    """
+    now = datetime.now(tz=timezone.utc).replace(second=0, microsecond=0)
+    logs: List[Logs] = [
+        Logs(
+            timestamp=now - timedelta(minutes=3),
+            resources={"service.name": "test"},
+            attributes={"code.file": "test.py"},
+            body="Test log",
+            severity_text="INFO",
+        ),
```
srikanthccv, line 1543: Let's add one more log here with different time with another service name and use the service name filters in query A (`service.name = "test"`) and query B (`service.name = "another-test"`).

tests/integration/src/querier/a_logs.py
```diff
@@ -1333,3 +1333,668 @@ def test_datatype_collision(
...
+    assert response.status_code == HTTPStatus.OK
+    assert response.json()["status"] == "success"
+
+    results = response.json()["data"]["data"]["results"]
+    assert len(results) >= 1
+
+    # Find formula result and verify it has data
+    formula_result = next((r for r in results if r.get("name") == "F1"), results[0])
+    assert formula_result is not None
+
```
srikanthccv: When we don't know the value(s) beforehand, the tests can simply check for the existence of the data. However, here we know what data to expect, so to keep the tests complete, we should assert on the values wherever needed. 

Expanding on that for cases where we don't know the exact value (not this case, but hypothetically some case), there would still be a chance we know the delta between the expected and actual values, so such assertions should be written. Data existence tests should be a last resort. That way, we make the tests better by asserting on expected values, which catches correctness issues as well as fill gaps functionality. What do you think?

tests/integration/src/querier/a_logs.py
```diff
@@ -1333,3 +1333,668 @@ def test_datatype_collision(
...
+                        "spec": {
+                            "name": "A",
+                            "signal": "logs",
+                            "stepInterval": 60,
+                            "disabled": False,
+                            "having": {"expression": ""},
+                            "aggregations": [{"expression": "count()"}],
+                            "functions": [{
+                                "name": "fillZero",
+                                "args": [{"value": start_ms}, {"value": end_ms}, {"value": step_ms}]
```
srikanthccv: Let's remove this and test, because it is expected to be filled appropriately.

tests/integration/src/querier/a_logs.py
```diff
@@ -1333,3 +1333,668 @@ def test_datatype_collision(
...
+                            "name": "A",
+                            "signal": "logs",
+                            "stepInterval": 60,
+                            "disabled": False,
+                            "groupBy": [{"name": "service.name", "fieldDataType": "string", "fieldContext": "resource"}],
+                            "having": {"expression": ""},
+                            "aggregations": [{"expression": "count()"}],
+                            "functions": [{
+                                "name": "fillZero",
+                                "args": [{"value": start_ms}, {"value": end_ms}, {"value": step_ms}]
```
srikanthccv: Same

tests/integration/src/querier/a_logs.py
```diff
@@ -1333,3 +1333,668 @@ def test_datatype_collision(
...
+    now = datetime.now(tz=timezone.utc).replace(second=0, microsecond=0)
+    logs: List[Logs] = [
+        Logs(
+            timestamp=now - timedelta(minutes=3),
+            resources={"service.name": "test"},
+            attributes={"code.file": "test.py"},
+            body="Test log",
+            severity_text="INFO",
+        ),
+    ]
```
srikanthccv, line 1946: Same as the fillgaps, let's add some additional service data.

tests/integration/src/querier/a_logs.py
```diff
@@ -1333,3 +1333,668 @@ def test_datatype_collision(
...
+                    },
+                    {
+                        "type": "builder_formula",
+                        "spec": {
+                            "name": "F1",
+                            "expression": "A + B",
+                            "disabled": False,
+                            "functions": [{
+                                "name": "fillZero",
+                                "args": [{"value": start_ms}, {"value": end_ms}, {"value": step_ms}]
```
srikanthccv: Let's remove this.

tests/integration/src/querier/a_logs.py
```diff
@@ -1333,3 +1333,668 @@ def test_datatype_collision(
...
```
srikanthccv: Same

tests/integration/src/querier/a_logs.py
```diff
@@ -1333,3 +1333,668 @@ def test_datatype_collision(
...
+            },
+            "formatOptions": {"formatTableResultForUI": False, "fillGaps": False},
+        },
+    )
+
+    assert response.status_code == HTTPStatus.OK
+    assert response.json()["status"] == "success"
+
+    results = response.json()["data"]["data"]["results"]
+    assert len(results) >= 1
```
srikanthccv: No repeating everywhere, but let's ensure we assert on actual values.

tests/integration/src/querier/b_traces.py
srikanthccv, line 1: Similar comments as logs.

tests/integration/src/querier/c_metrics.py
srikanthccv, line 1: Here also, largely similar comments as logs for the better tests.

tests/integration/fixtures/metrics.py
```diff
@@ -0,0 +1,248 @@
...
+    """High-level metric representation. Produces both time series and sample entries."""
+    metric_name: str
+    labels: dict[str, str]
+    temporality: str
+    timestamp: datetime.datetime
+    value: float
+    flags: int
+    
+    time_series: MetricsTimeSeries
+    sample: MetricsSample
```
srikanthccv: make these two private

srikanthccv (commented):
Thanks for the comprehensive tests, left some comments, please take a look and let me know what do you think.

tests/integration/src/querier/a_logs.py
niladrix719: I agree

niladrix719:
🙌  Thanks for the comments, I have updated the tests

pkg/types/querybuildertypes/querybuildertypesv5/req.go
```diff
@@ -440,7 +440,8 @@ func (r *QueryRangeRequest) GetQueriesSupportingZeroDefault() map[string]bool {
 					if timeAgg == metrictypes.TimeAggregationCount ||
 						timeAgg == metrictypes.TimeAggregationCountDistinct ||
 						timeAgg == metrictypes.TimeAggregationRate ||
-						timeAgg == metrictypes.TimeAggregationIncrease {
+						timeAgg == metrictypes.TimeAggregationIncrease ||
+						timeAgg == metrictypes.TimeAggregationSum {
```
srikanthccv: `metrictypes.TimeAggregationSum` doesn't belong here. The reason `metrictypes.TimeAggregationSum` shouldn't be on this list comes down to how we handle missing data. The whole point of `GetQueriesSupportingZeroDefault` is to determine whether we can safely treat a non-existent value at timestamp `t` as zero when we need to; essentially, "if there's no data there, can we just call it zero?"

Now, here's why `Count` and its variants make perfect sense for this:

- **Count**: If you have an empty set, the count is definitively zero. Same goes for `CountDistinct` zero distinct values in a null set.
- **Rate**: At its core, rate is just `Count/Time`. So if count can be zero, rate can absolutely be zero too.
- **Increase**: This is really just another name for count, so the same logic applies, counting nothing gives you zero.

**But `Sum` is different.** With sum, you're trying to add up values that actually *exist*. If you have no data at all, an empty set, then the sum is *undefined*, not zero. You're not adding up zeros; you're adding up *nothing*, and there's no meaningful answer to "what's the sum of no values?"

Does that clarify?

tests/integration/src/querier/timeseries_utils.py
srikanthccv, line 1: These are not really _"fixtures"_,
- https://martinfowler.com/bliki/ObjectMother.html
- https://google.github.io/googletest/primer.html#same-data-multiple-tests
 
Let's move these function under `src/querier` only.

tests/integration/src/querier/c_metrics.py
```diff
@@ -0,0 +1,906 @@
...
+                "queries": [
+                    {
+                        "type": "builder_query",
+                        "spec": {
+                            "name": "A",
+                            "signal": "metrics",
+                            "aggregations": [{
+                                "metricName": metric_name,
+                                "temporality": "cumulative",
+                                "timeAggregation": "sum",
```
srikanthccv: For counter metrics, the valid `timeAggregation`s are `rate` and `increase`. Please either use `rate`/`increase` for `timeAggregation`. Similarly, please update the other tests as well.

More here about the metric types and aggregations https://signoz.io/docs/metrics-management/types-and-aggregation/

pkg/types/querybuildertypes/querybuildertypesv5/req.go
niladrix719: I initially added Sum because I noticed that in the logs and traces, we allow sum to default to zero, so I assumed metrics could also be treated the same way 😅, thanks for the on point review now I understand

tests/integration/src/querier/timeseries_utils.py
niladrix719: sure

pkg/types/querybuildertypes/querybuildertypesv5/req.go
srikanthccv: good catch! `sum` shouldn't be there as well. I will check why it was added there in the first place. but let's remove the newly added `sum` for metrics here.

niladrix719:
This should be good now, Thanks for all the reviews @srikanthccv 🙌

srikanthccv (approved):
Thanks, this is a good start; tests could be improved with parametrization and some additional enhancements to fixtures but those are for another ticket.

therealpandey (approved):

PR #9890, opened 2025-12-28
fix(querier): handle __all__ variable value as array in PromQL queries

### LINKED ISSUE #9888: Dynamic variable `ALL` doesn't work with PromQL
What do we do today,
- PromQL

   - Parse the the query and remove the filter altogether (since it's ALL), when the value for the variable is `__all__`
   - Recreate the query with removed filter as the new query and use it

It seems to be not working, either because the input is not simple `__all__` string or some other bug. Investigate and fix.

nishchay-veer:
## 📄 Summary

Fixes an issue where PromQL query filters with `__all__` variables were not being removed correctly when the variable value came as an array instead of a string.

---

## ✅ Changes

- [X] Bug fix: The `removeAllVarMatchers` function only checked if `v.Value` was a string `"__all__"`. 
https://github.com/SigNoz/signoz/blob/main/pkg/querier/promql_query.go#L71
However, due to JSON unmarshalling, the value can also arrive as:

```
[]interface{}{"__all__"}
[]string{"__all__"}
```
This caused the filter removal logic to fail silently, leaving unnecessary filters in the query.

---

## 🔍 Related Issues

Closes #9888 

---

## 📋 Checklist

- [X] Dev Review
- [X] Test cases added (Unit/ Integration / E2E)
- [X] Manually tested the changes

---

> [!NOTE]
> Fixes PromQL label-matcher removal when variables set to `__all__` arrive as arrays from JSON unmarshalling.
> 
> - Adds `isAllValue` to detect `__all__` in `string`, `[]any` (single element), and `[]string` (single element)
> - Updates `removeAllVarMatchers` to use `isAllValue` so matchers referencing such vars are dropped
> - Adds unit tests for `isAllValue` and expanded cases in `removeAllVarMatchers`, including array and non-matching edge cases
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f4cbe072cc44c0bd4beb6d7b93be6a2b095d2a56. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/querier/promql_query.go
+++ b/pkg/querier/promql_query.go
@@ -60,6 +60,28 @@ func (q *promqlQuery) Window() (uint64, uint64) {
 	return q.tr.From, q.tr.To
 }
 
+func isAllValue(value any) bool {
+	if strVal, ok := value.(string); ok && strVal == "__all__" {
+		return true
+	}
+
+	// check for slice types
+	switch v := value.(type) {
+	case []any:
+		if len(v) == 1 {
+			if strVal, ok := v[0].(string); ok && strVal == "__all__" {
+				return true
+			}
+		}
+	case []string:
+		if len(v) == 1 && v[0] == "__all__" {
+			return true
+		}
+	}
+
+	return false
+}
+
 // removeAllVarMatchers removes label matchers from a PromQL query that reference variables with __all__ value.
 // This method parses the query, walks the AST to remove matching matchers, and returns the modified query string.
 // If parsing or walking fails, it returns an error.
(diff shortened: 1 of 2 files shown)


srikanthccv:
bro, did you actually understand the issue or justing AI-ing you way through issues?

srikanthccv:
I am going to close this because it works currently from the querier functionality side.

nishchay-veer:
I looked at this test for ClickHouse query transformation here:
https://github.com/SigNoz/signoz/blob/main/pkg/variables/clickhouse/transfor_test.go#L30
and saw that we are handling slice form of __all__ as well and I thought the same thing should apply to PromQL query as well.

srikanthccv:
That's not yet integrated anywhere in the actual code yet for `pkg/querier`. It's always a good idea to verify the bug report before implementation. It's one thing to handle slice but it's completely another thing to imply that this fixes the linked issue (which also mentioned about investigation, as it was really verified at the time of creation).

PR #9892, opened 2025-12-28
feat(querier): support dynamic variable ALL for ClickHouse queries

### LINKED ISSUE #9889: Dynamic variable `ALL` doesn't work with ClickHouse queries
The code to transform the query from the original query to a new query with the condition replaced is present at https://github.com/SigNoz/signoz/blob/7c051601f25fa41502684c798d09f0c2932d7fb1/pkg/variables/clickhouse/transform.go#L28

It should be integrated with the 

1. https://github.com/SigNoz/signoz/blob/7c051601f25fa41502684c798d09f0c2932d7fb1/pkg/querier/clickhouse_query.go#L63
2. variable on the `/api/v1/query`

This is to be done in two steps

1. Passthrough logic
   - integrate with the transformer and check if the result query is the same as the incoming query for the existing queries; if not, log an error
   - for the queries that use a dynamic variable, ensure the filter is removed
2. Enable as the active substitution once all the cases of parsing and substitution are covered.

ixchio:
## What

This integrates the existing `QueryTransformer` with `renderVars()` to properly handle the `__all__` case for dynamic variables in ClickHouse queries.

## Why

When users select "ALL" for a dashboard variable, the filter should be removed entirely - not substituted with the literal `__all__` string (which matches nothing).

**Before**: `WHERE service_name = '__all__'` → returns nothing  
**After**: `WHERE` clause for that variable is removed → returns all

## How

1. Check if any dynamic variable has `__all__` value
2. Run `QueryTransformer` to parse SQL and remove those filter conditions
3. Log when transformation changes the query (for observability)
4. Fall back to original if parsing fails (safe rollout)

## Testing

- Added unit tests for helper functions
- Added integration test for the full `renderVars` flow with `__all__`
- All existing tests pass

Fixes #9889

---

> [!NOTE]
> Implements proper handling of dashboard variables set to `__all__` in ClickHouse queries.
> 
> - Integrates `QueryTransformer` into `renderVars()` to detect `__all__` dynamic variables and remove their filter clauses; logs transform results and falls back to original query on errors
> - Adds helpers `convertToVariableValues` and `hasAllVars`, plus unit and integration tests in `clickhouse_query_test.go`
> - Minor `deploy/install.sh` tweaks: more robust Docker start flow on macOS and small messaging/comment cleanups
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 32aabb051a9c2e06d0c732e183692d219d90e158. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/querier/clickhouse_query.go
+++ b/pkg/querier/clickhouse_query.go
@@ -15,6 +15,7 @@ import (
 	"github.com/SigNoz/signoz/pkg/querybuilder"
 	"github.com/SigNoz/signoz/pkg/telemetrystore"
 	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
+	chVariables "github.com/SigNoz/signoz/pkg/variables/clickhouse"
 )
 
 type chSQLQuery struct {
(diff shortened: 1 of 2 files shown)


srikanthccv:
See https://github.com/SigNoz/signoz/issues/9889#issuecomment-3916497305

Issue #9514, opened 2025-11-07
Clicking on "Alerts" erases the work


srikanthccv:
>It's very easy to navigate away from in-progress work and lose all of it. I clicked on "Alerts" while editing a panel because I was curious, and without any warning it took me to another page and cleared all of my work.

AshwinBhatkal:
rough notes:
new panel issue
back button should work
just changing filter without staging & run is erroneous

AshwinBhatkal:
Story so far:
1. Given that we will be in Edit mode, navigating to a whole different page in between is not ideal
2. Solution is to open a new tab when Alerts is clicked upon. We will add an icon next to the text to indicate this.
3. On clicking Alerts, we make a call to "substitute_vars" and then redirect to the alert creation page by creating the url from the response of this url
4. Before the change, we passed in the thresholds state to the navigation. Expectation is to show these thresholds on page but this does not seem to working as expected.
5. Also, the impact of this change proposed is also the panel in the main dashboard view. There is a difference in behaviour here. In this case, we don't pass in thresholds if they do exist. Point 2 will apply here as well

PR #9902, opened 2025-12-30
fix: alerts creation from dashboard and panel to happen in new tab

AshwinBhatkal:
## 📄 Summary

- When trying to create a new alert(rule) in the edit mode of a dashboard panel, the page would navigate to the creation of the alert(rule) but user would not be able to navigate back. This could sometimes cause them to lose all their unsaved work.
- This PR fixes that by opening this in a new tab instead of navigating it to another page in-situ.

---

## ✅ Changes

- [x] Trying to create an alert from View Panel in Dashboard page or Edit Panel Page would open a new tab instead of in app navigation
- [x] Removed the extra prop being pushed as state in the history object. This value was not being used. Will fix this state as part of https://github.com/SigNoz/signoz/issues/5291
- [x] Added icon `SquareArrowOutUpRight` in the 2 places mentioned above.

## 🧪 How to Test

Test can be executed exactly like the screen recording

---

## 🔍 Related Issues

Closes https://github.com/SigNoz/signoz/issues/9514

---

https://github.com/user-attachments/assets/58210ba1-09ae-401d-aef7-b23310ef2a90

Screenshot 2025-12-30 at 2 35 43 PM

Screenshot 2025-12-30 at 2 35 33 PM

---

## 📋 Checklist

- [x] Dev Review

### DIFF
--- a/frontend/src/components/Logs/ListLogView/index.tsx
+++ b/frontend/src/components/Logs/ListLogView/index.tsx
@@ -14,7 +14,7 @@ import { useIsDarkMode } from 'hooks/useDarkMode';
 // utils
 import { FlatLogData } from 'lib/logs/flatLogData';
 import { useTimezone } from 'providers/Timezone';
-import { useCallback, useMemo, useState } from 'react';
+import { memo, useCallback, useMemo } from 'react';
 // interfaces
 import { IField } from 'types/api/logs/fields';
 import { ILog } from 'types/api/logs/log';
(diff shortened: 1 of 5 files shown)


frontend/src/container/NewWidget/RightContainer/__tests__/RightContainer.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+jest.mock('lucide-react', () => ({
+	...jest.requireActual('lucide-react'),
+	ConciergeBell: (): JSX.Element => <svg data-testid="lucide-concierge-bell" />,
+	SquareArrowOutUpRight: (): JSX.Element => (
+		<svg data-testid="lucide-square-arrow-out-up-right" />
+	),
+	Plus: (): JSX.Element => <svg data-testid="lucide-plus" />,
+}));
+
+describe('RightContainer - Alerts Section', () => {
```
AshwinBhatkal, line 124: test is only for Right Container - Alerts display

PR #9912, opened 2026-01-02
fix(preference): rename columns for preference tables

vikrantgupta25:
## Summary

- rename preference_id and preference_value to name and value respectively 
- removed the cascading deletes and updates from preference tables 
- fixed the issue with opaque tokeniser with delete order with respect to user

closes: https://github.com/SigNoz/platform-pod/issues/1518

### DIFF
--- a/pkg/sqlschema/operator.go
+++ b/pkg/sqlschema/operator.go
@@ -82,42 +82,6 @@ func (operator *Operator) AddColumn(table *Table, uniqueConstraints []*UniqueCon
 	return sqls
 }
 
-func (operator *Operator) RenameColumn(table *Table, uniqueConstraints []*UniqueConstraint, oldColumnName ColumnName, newColumnName ColumnName) [][]byte {
-	index := operator.findColumnByName(table, oldColumnName)
-	// If the column does not exist, we do not need to rename it.
-	if index == -1 {
-		return [][]byte{}
-	}
-
-	column := table.Columns[index]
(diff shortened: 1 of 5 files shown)


pkg/sqlschema/operator.go
```diff
@@ -82,6 +82,42 @@ func (operator *Operator) AddColumn(table *Table, uniqueConstraints []*UniqueCon
 	return sqls
 }
 
+func (operator *Operator) RenameColumn(table *Table, uniqueConstraints []*UniqueConstraint, oldColumnName ColumnName, newColumnName ColumnName) [][]byte {
+	index := operator.findColumnByName(table, oldColumnName)
```
therealpandey: The table and uniqueConstraints were morphed in `RenameColumn`. The returning sqls do not match the updated table structure. Either return sqls which will recreate the table or just rename column here.

therealpandey (changes requested):

pkg/sqlschema/operator.go
vikrantgupta25: for sqlite and postgres it actually does update the underlying constraints automatically hence the update. but the current recreate table doesn't work due to duplicate constraints. need to think something else

vikrantgupta25: used go based migration

vikrantgupta25:
not adding integration tests as they are already added. thanks to @therealpandey

PR #9927, opened 2026-01-05
feat: improved keys adjustment logic

tushar-signoz:
## 📄 Summary
This pull request introduces a new utility file, `collision.go`, to the `querybuilder` package, which provides robust logic for handling duplicate and ambiguous field keys in query builder queries. The new functions ensure that queries are more resilient to naming collisions and ambiguities, especially when dealing with JSON and promoted fields. Additionally, several test cases have been updated to reflect improved handling and warning messages for ambiguous keys.

Key changes include:

**New functionality for key handling:**

* Added `collision.go` to `pkg/querybuilder`, introducing functions like `AdjustDuplicateKeys`, `AdjustKey`, and `AdjustKeysForAliasExpressions` to deduplicate keys, resolve ambiguities, and adjust key representations in queries. These utilities ensure that each key in `SelectFields`, `GroupBy`, and `OrderBy` is unique and as generic as possible, and help clarify ambiguous or aliased fields.

**Test updates for improved key handling:**

* Updated test cases in `json_stmt_builder_test.go` to use unprefixed field names (e.g., `"user.age"` instead of `"body.user.age"`) in group by clauses, reflecting changes in how keys are handled and represented in queries. [[1]](diffhunk://#diff-8d7b8c823783356c46f295b56d4628b2fbdd8848094a81cd07562a960a4cd989L48-R48) [[2]](diffhunk://#diff-8d7b8c823783356c46f295b56d4628b2fbdd8848094a81cd07562a960a4cd989L73-R73) [[3]](diffhunk://#diff-8d7b8c823783356c46f295b56d4628b2fbdd8848094a81cd07562a960a4cd989L129-R134)
* Enhanced warning messages in test expectations to include `jsondatatype` information, making ambiguity warnings more descriptive and aligned with the new key handling logic. [[1]](diffhunk://#diff-8d7b8c823783356c46f295b56d4628b2fbdd8848094a81cd07562a960a4cd989L243-R243) [[2]](diffhunk://#diff-8d7b8c823783356c46f295b56d4628b2fbdd8848094a81cd07562a960a4cd989L272-R272) [[3]](diffhunk://#diff-8d7b8c823783356c46f295b56d4628b2fbdd8848094a81cd07562a960a4cd989L287-R287)
---

## ✅ Changes

---

ex:

- `frontend`
- `backend`
- `devops`
- `bug`
- `enhancement`
- `ui`
- `test`

---

## 👥 Reviewers

- frontend / backend / devops

---

## 🧪 How to Test

1. ...
2. ...
3. ...

---

## 🔍 Related Issues

Closes #
https://github.com/SigNoz/engineering-pod/issues/3189
---

---

## 📋 Checklist

- [ ] Dev Review

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> Introduces generic key normalization/deduplication and smarter key resolution across queries.
> 
> - Adds `pkg/querybuilder/collision.go` with `AdjustDuplicateKeys`, `AdjustKey`, and alias-aware `AdjustKeysForAliasExpressions` to merge conflicting contexts/data types and reconcile intrinsic/calculated vs metadata keys
> - Refactors logs/traces statement builders to use the new adjustors, return the adjusted query, and build precise key selectors (including context/data type); logs key/action telemetry
> - Enhances validation to accept context-prefixed aggregation aliases in `orderBy`; adjusts `orderBy` keys to alias form for non-raw requests
> - Enriches `TelemetryFieldKey.String()` and adds `Equal`, enabling clearer ambiguity warnings (now include `jsondatatype`, `materialized`, indexes)
> - Updates tests (unit, integration) and test data to reflect new key handling, alias behavior, and improved warnings (e.g., remove `body.` prefixes, add mixed materialization cases)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit afea85ac2742fd20abfffe96e3bf963a2e52cf32. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/src/components/CustomTimePicker/CustomTimePicker.styles.scss
+++ b/frontend/src/components/CustomTimePicker/CustomTimePicker.styles.scss
@@ -131,6 +131,45 @@
 	border-top: 1px solid var(--bg-ink-200);
 	padding: 8px 14px;
 	.timezone-container {
+		display: flex;
+		align-items: center;
+		justify-content: space-between;
+
+		&__left {
+			display: flex;
+			align-items: center;
+			justify-content: center;
+			gap: 6px;
+
+			.timezone__name {
+				font-size: 12px;
+				line-height: 16px;
+				color: var(--bg-robin-400);
+				font-weight: 500;
+			}
+
+			.timezone__separator {
+				font-size: 12px;
+				line-height: 16px;
+				color: var(--bg-robin-300);
+				font-weight: 500;
+			}
+
+			.timezone__offset {
+				font-size: 11px;
+				line-height: 14px;
+				color: var(--bg-robin-400);
+				font-weight: 500;
+			}
+		}
+
+		&__right {
+			display: flex;
+			align-items: center;
+			justify-content: center;
+			gap: 6px;
+		}
+
 		&,
 		.timezone {
 			font-family: Inter;
@@ -138,6 +177,7 @@
 			line-height: 16px;
 			letter-spacing: -0.06px;
 		}
+
 		display: flex;
 		align-items: center;
 		color: var(--bg-vanilla-400);
@@ -156,18 +196,21 @@
 		}
 	}
 }
+
 .timezone-badge {
 	display: flex;
 	align-items: center;
 	justify-content: center;
 	padding: 0 4px;
 	border-radius: 2px;
-	background: rgba(171, 189, 255, 0.04);
-	color: var(--bg-vanilla-100);
-	font-size: 12px;
+	color: var(--bg-vanilla-400);
+	background-color: var(--bg-ink-200);
+	font-size: 9px;
 	font-weight: 400;
-	line-height: 16px;
-	letter-spacing: -0.06px;
+	line-height: 12px;
+	letter-spacing: -0.045px;
+	margin-right: 4px;
+	width: 72px;
 	cursor: pointer;
 }
 
@@ -183,6 +226,7 @@
 	font-size: 11px;
 	color: var(--bg-vanilla-400);
 	background-color: var(--bg-ink-200);
+	cursor: pointer;
 
 	&.is-live {
 		background-color: transparent;
@@ -238,11 +282,10 @@
 	.date-time-popover__footer {
 		border-color: var(--bg-vanilla-400);
 	}
+
 	.timezone-container {
 		color: var(--bg-ink-400);
-		&__clock-icon {
-			stroke: var(--bg-ink-400);
-		}
+
 		.timezone {
 			color: var(--bg-ink-100);
 			background: rgb(179 179 179 / 15%);
--- a/frontend/src/components/CustomTimePicker/CustomTimePicker.tsx
+++ b/frontend/src/components/CustomTimePicker/CustomTimePicker.tsx
@@ -4,7 +4,6 @@
 import './CustomTimePicker.styles.scss';
 
 import { Input, InputRef, Popover, Tooltip } from 'antd';
-import logEvent from 'api/common/logEvent';
 import cx from 'classnames';
 import { DATE_TIME_FORMATS } from 'constants/dateTimeFormats';
 import { DateTimeRangeType } from 'container/TopNav/CustomDateTimeModal';
@@ -22,9 +21,7 @@ import {
 	ChangeEvent,
 	Dispatch,
 	SetStateAction,
-	useCallback,
 	useEffect,
-	useMemo,
 	useRef,
 	useState,
 } from 'react';
@@ -113,22 +110,8 @@ function CustomTimePicker({
 
 	const [activeView, setActiveView] = useState<ViewType>(DEFAULT_VIEW);
 
-	const { timezone, browserTimezone } = useTimezone();
+	const { timezone } = useTimezone();
 	const activeTimezoneOffset = timezone.offset;
-	const isTimezoneOverridden = useMemo(
-		() => timezone.offset !== browserTimezone.offset,
-		[timezone, browserTimezone],
-	);
-
-	const handleViewChange = useCallback(
-		(newView: 'timezone' | 'datetime'): void => {
-			if (activeView !== newView) {
-				setActiveView(newView);
-			}
-			setOpen(true);
-		},
-		[activeView, setOpen],
-	);
 
 	const [isOpenedFromFooter, setIsOpenedFromFooter] = useState(false);
 
@@ -371,6 +354,7 @@ function CustomTimePicker({
 			startTime,
 			endTime,
 			DATE_TIME_FORMATS.UK_DATETIME_SECONDS,
+			timezone.value,
 		);
 
 		if (!isValidTimeRange) {
@@ -422,8 +406,8 @@ function CustomTimePicker({
 		</div>
 	);
 
-	const handleOpen = (e: React.SyntheticEvent): void => {
-		e.stopPropagation();
+	const handleOpen = (e?: React.SyntheticEvent): void => {
+		e?.stopPropagation?.();
 
 		if (showLiveLogs) {
 			setOpen(true);
@@ -436,12 +420,12 @@ function CustomTimePicker({
 		// reset the input status and error message as we reset the time to previous correct value
 		resetErrorStatus();
 
-		const startTime = dayjs(minTime / 1000_000).format(
-			DATE_TIME_FORMATS.UK_DATETIME_SECONDS,
-		);
-		const endTime = dayjs(maxTime / 1000_000).format(
-			DATE_TIME_FORMATS.UK_DATETIME_SECONDS,
-		);
+		const startTime = dayjs(minTime / 1000_000)
+			.tz(timezone.value)
+			.format(DATE_TIME_FORMATS.UK_DATETIME_SECONDS);
+		const endTime = dayjs(maxTime / 1000_000)
+			.tz(timezone.value)
+			.format(DATE_TIME_FORMATS.UK_DATETIME_SECONDS);
 
 		setInputValue(`${startTime} - ${endTime}`);
 	};
@@ -468,18 +452,6 @@ function CustomTimePicker({
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [location.pathname]);
 
-	const handleTimezoneHintClick = (e: React.MouseEvent): void => {
-		e.stopPropagation();
-		handleViewChange('timezone');
-		setIsOpenedFromFooter(false);
-		logEvent(
-			'DateTimePicker: Timezone picker opened from time range input badge',
-			{
-				page: location.pathname,
-			},
-		);
-	};
-
 	const handleInputBlur = (): void => {
 		resetErrorStatus();
 	};
@@ -498,19 +470,27 @@ function CustomTimePicker({
 		return '';
 	};
 
+	const focusInput = (): void => {
+		// Use setTimeout to wait for React to update the DOM and make input editable
+		setTimeout(() => {
+			const inputElement = inputRef.current?.input;
+			if (inputElement) {
+				inputElement.focus();
+				inputElement.select();
+			}
+		}, 100);
+	};
+
 	// Focus and select input text when popover opens
 	useEffect(() => {
 		if (open && inputRef.current) {
-			// Use setTimeout to wait for React to update the DOM and make input editable
-			setTimeout(() => {
-				const inputElement = inputRef.current?.input;
-				if (inputElement) {
-					inputElement.focus();
-					inputElement.select();
-				}
-			}, 0);
+			focusInput();
 		}
 	}, [open]);
+
+	const handleTimezoneChange = (): void => {
+		focusInput();
+	};
 
 	return (
 		<div className="custom-time-picker">
@@ -532,6 +512,7 @@ function CustomTimePicker({
 								customDateTimeVisible={defaultTo(customDateTimeVisible, false)}
 								onCustomDateHandler={defaultTo(onCustomDateHandler, noop)}
 								onSelectHandler={handleSelect}
+								onTimezoneChange={handleTimezoneChange}
 								onGoLive={defaultTo(onGoLive, noop)}
 								onExitLiveLogs={defaultTo(onExitLiveLogs, noop)}
 								options={items}
@@ -583,8 +564,8 @@ function CustomTimePicker({
 						prefix={getInputPrefix()}
 						suffix={
 							<div className="time-input-suffix">
-								{!!isTimezoneOverridden && activeTimezoneOffset && (
-									<div className="timezone-badge" onClick={handleTimezoneHintClick}>
+								{activeTimezoneOffset && (
+									<div className="timezone-badge">
 										<span>{activeTimezoneOffset}</span>
 									</div>
 								)}
--- a/frontend/src/components/CustomTimePicker/CustomTimePickerPopoverContent.tsx
+++ b/frontend/src/components/CustomTimePicker/CustomTimePickerPopoverContent.tsx
@@ -31,6 +31,7 @@ import { TimeRangeValidationResult } from 'utils/timeUtils';
 import CalendarContainer from './CalendarContainer';
 import { CustomTimePickerInputStatus } from './CustomTimePicker';
 import TimezonePicker from './TimezonePicker';
+import { Timezone } from './timezoneUtils';
 
 const TO_MILLISECONDS_FACTOR = 1000_000;
 
@@ -52,6 +53,7 @@ interface CustomTimePickerPopoverContentProps {
 		lexicalContext?: LexicalContext,
 	) => void;
 	onSelectHandler: (label: string, value: string) => void;
+	onTimezoneChange: (timezone: Timezone) => void;
 	onGoLive: () => void;
 	selectedTime: string;
 	activeView: 'datetime' | 'timezone';
@@ -101,6 +103,7 @@ function CustomTimePickerPopoverContent({
 	setCustomDTPickerVisible,
 	onCustomDateHandler,
 	onSelectHandler,
+	onTimezoneChange,
 	onGoLive,
 	selectedTime,
 	activeView,
@@ -208,6 +211,7 @@ function CustomTimePickerPopoverContent({
 					setActiveView={setActiveView}
 					setIsOpen={setIsOpen}
 					isOpenedFromFooter={isOpenedFromFooter}
+					onTimezoneSelect={onTimezoneChange}
 				/>
 			</div>
 		);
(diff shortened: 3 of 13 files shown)


pkg/telemetrylogs/statement_builder.go
```diff
@@ -199,6 +166,89 @@ func (b *logQueryStatementBuilder) adjustKeys(ctx context.Context, keys map[stri
...
+		// use the intrinsic/calculated field
+		if !match {
+			b.logger.InfoContext(ctx, "overriding the field context and data type", "key", key.Name)
+			key.FieldContext = IntrinsicFields[key.Name].FieldContext
+			key.FieldDataType = IntrinsicFields[key.Name].FieldDataType
+			key.Materialized = IntrinsicFields[key.Name].Materialized
+		} else {
+			// Here we have a key which is an intrinsic field but also exists in the metadata with the same name
+			// cannot do anything, so just return
+			return
```
srikanthccv: This situation should not be happening, if it does, it's likely an issue worth letting user know about it? And where do we handle it?

Let's take `body`, some users also send it as attribute, they almost always want the log body than attribute body. What do you think?

pkg/telemetrylogs/statement_builder.go
```diff
@@ -199,6 +166,89 @@ func (b *logQueryStatementBuilder) adjustKeys(ctx context.Context, keys map[stri
...
+		for _, metadataKey := range keys[key.Name] {
+			// Only consider keys that match the context and data type (if specified)
+			if (key.FieldContext == telemetrytypes.FieldContextUnspecified || key.FieldContext == metadataKey.FieldContext) &&
+				(key.FieldDataType == telemetrytypes.FieldDataTypeUnspecified || key.FieldDataType == metadataKey.FieldDataType) {
+				matchingKeys = append(matchingKeys, metadataKey)
+			}
+		}
+
+		if len(matchingKeys) == 0 {
+			// we do not have any matching keys, most likely user made a mistake, let QB handle it
```
srikanthccv: > let QB handle it

What happens to invalid keys in selected fields, group by, order by? Will it create a valid query and predictable results?

pkg/telemetrylogs/statement_builder.go
srikanthccv: See conv here https://signoz-team.slack.com/archives/C08AKRVE9FD/p1767677062994699?thread_ts=1767607943.517669&cid=C08AKRVE9FD.

srikanthccv:
discussed all the failure cases (that arise from the bad quality of data) here https://github.com/SigNoz/engineering-pod/issues/3193#issuecomment-3715954248. @tushar-signoz will write up a doc with details (and solutions to problems including the UX) and implement it.

pkg/telemetrylogs/statement_builder.go
tushar-signoz: Downstream throws an error, key not found.

pkg/telemetrylogs/statement_builder.go
```diff
@@ -217,6 +186,21 @@ func (b *logQueryStatementBuilder) adjustKeys(ctx context.Context, keys map[stri
...
+
+	// First check if it matches with any intrinsic fields
+	var intrinsicOrCalculatedField telemetrytypes.TelemetryFieldKey
+	if _, ok := IntrinsicFields[key.Name]; ok {
+		intrinsicOrCalculatedField = IntrinsicFields[key.Name]
+		return querybuilder.AdjustKey(ctx, key, keys, &intrinsicOrCalculatedField)
+	}
+
+	return querybuilder.AdjustKey(ctx, key, keys, nil)
+
```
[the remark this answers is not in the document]
tushar-signoz: This is because normalise is moved to json unmarshling

pkg/querybuilder/collision.go
```diff
@@ -0,0 +1,281 @@
...
+	globalUniqueKeysMap := map[string]telemetrytypes.TelemetryFieldKey{}
+
+	// for recording modifications
+	actions := []string{}
+
+	// SelectFields
+	for idx := range query.SelectFields {
+		if existingKey, ok := globalUniqueKeysMap[query.SelectFields[idx].Name]; !ok {
+			globalUniqueKeysMap[query.SelectFields[idx].Name] = query.SelectFields[idx]
+		} else {
```
piyushsingariya: I didn't found any specific reason to not iterate over the elements instead of idx 

```suggestion
	for _, field := range query.SelectFields {
		if existingKey, ok := globalUniqueKeysMap[field.Name]; !ok {
			globalUniqueKeysMap[field.Name] = field
		} else {
```

pkg/querybuilder/collision.go
```diff
@@ -0,0 +1,281 @@
...
+				existingKey.FieldContext = telemetrytypes.FieldContextUnspecified
+			}
+			if existingKey.FieldDataType != query.GroupBy[idx].FieldDataType && existingKey.FieldDataType != telemetrytypes.FieldDataTypeUnspecified {
+				// remove field data type in the map to make it generic
+				actions = append(actions, fmt.Sprintf("Removed field data type from %s for duplicate key %s", existingKey, query.GroupBy[idx].TelemetryFieldKey))
+				existingKey.FieldDataType = telemetrytypes.FieldDataTypeUnspecified
+			}
+			// Update the map with the modified key
+			globalUniqueKeysMap[query.GroupBy[idx].Name] = existingKey
+		}
```
piyushsingariya: The logic seems repeating? This can be a closure

pkg/querybuilder/collision.go
```diff
@@ -0,0 +1,281 @@
...
+	newSelectFields := make([]telemetrytypes.TelemetryFieldKey, 0, len(query.SelectFields))
+
+	seen := map[string]bool{}
+	for _, key := range query.SelectFields {
+		if !seen[key.Name] {
+			newSelectFields = append(newSelectFields, globalUniqueKeysMap[key.Name])
+			seen[key.Name] = true
+		} else {
+			actions = append(actions, fmt.Sprintf("Skipped duplicate SelectField key %s", key))
+		}
```
piyushsingariya, line 47: Can be a closure?

piyushsingariya (changes requested):

pkg/querybuilder/collision.go
tushar-signoz: Types are different, for select, groupBy and order so for this cannot use closure.

.vscode/settings.json
piyushsingariya, line 1: how come .vscode is not ignored? It should not appear right

pkg/querybuilder/collision.go
piyushsingariya: This is becoming quite a thing, loops repeating same operations at several places I believe. 

Let's introduce this.

```
type Keyed interface {
    GetKey() *telemetrytypes.TelemetryFieldKey
}
```

and use generics like this, without generics it can also be written differently just for reference added

```
func DedupAndRemap[T Keyed](
    items []T,
    global map[string]telemetrytypes.TelemetryFieldKey,
    action *[]string,
) []T {
    seen := map[string]bool{}
    out := make([]T, 0, len(items))

    for _, item := range items {
        name := item.GetKey().Name

        if seen[name] {
            *action = append(*action, fmt.Sprintf("Skipped duplicate key %s", name))
            continue
        }
        seen[name] = true

        newRef := global[name]
        updated := item.WithKey(newRef).(T)

        out = append(out, updated)
    }
    return out
}
```

tushar-signoz: Tried but couldn't do it without hampering the readability. Will take another tab at it later

pkg/querybuilder/collision.go
```diff
@@ -0,0 +1,256 @@
...
+
+	// Create a map to track unique keys across SelectFields, GroupBy, and OrderBy
+	globalUniqueKeysMap := map[string]telemetrytypes.TelemetryFieldKey{}
+
+	// for recording modifications
+	actions := []string{}
+
+	// SelectFields
+	for _, key := range query.SelectFields {
+		deduplicateKeys(key, globalUniqueKeysMap, &actions)
```
srikanthccv, line 23: TIL https://stackoverflow.com/questions/55520624/why-slice-values-can-sometimes-go-stale-but-never-map-values/55521138#55521138

pkg/querybuilder/collision.go
```diff
@@ -0,0 +1,256 @@
...
+			*actions = append(*actions, fmt.Sprintf("Removed field data type from %s for duplicate key %s", existingKey, key))
+			existingKey.FieldDataType = telemetrytypes.FieldDataTypeUnspecified
+		}
+		// Update the map with the modified key
+		keysMap[key.Name] = existingKey
+	}
+}
+
+
+func AdjustKey(ctx context.Context, key *telemetrytypes.TelemetryFieldKey, keys map[string][]*telemetrytypes.TelemetryFieldKey, intrinsicOrCalculatedField *telemetrytypes.TelemetryFieldKey) []string {
```
srikanthccv: We talked about this that `keys` already contains the intrinsic fields so this needs to be updated accordingly

pkg/querybuilder/collision.go
```diff
@@ -0,0 +1,256 @@
...
+			}
+			key.FieldContext = matchingKey.FieldContext
+			key.FieldDataType = matchingKey.FieldDataType
+			key.JSONDataType = matchingKey.JSONDataType
+			key.Indexes = matchingKey.Indexes
+			key.Materialized = matchingKey.Materialized
+
+			return actions
+		} else {
+			// multiple matching keys, set materialized only if all the keys are materialized
```
srikanthccv: I know this is from existing code, but shouldn't the materialized flag be determined per type + field context rather than across all keys? For example, http.status_code might have its number type materialized but not its string type, in which case, the query should use the appropriate column based on each type's materialization status, not a single value for the entire set?

srikanthccv (commented):
Initial pass, pulling the code locally to go through.

pkg/querybuilder/collision.go
tushar-signoz: Correct downstream should be handling this and shouldn't rely on materialised flag that we're setting or unsetting. I haven't looked downstream code with a fine comb yet. 
The ideology I have used in this section is to provide all the info downstream might be using (for eg: aggregating the indexes while Im not sure how they're being used ) same with materialised flag. 
Setting it true if it is always valid and setting it false will not throw any error as such only maybe a performance degradation.

That being said, this section might turn out to be redundant. Added a TODO.

pkg/querybuilder/collision.go
tushar-signoz: Done

pkg/telemetrytraces/statement_builder.go
```diff
@@ -246,6 +255,42 @@ func (b *traceQueryStatementBuilder) adjustKeys(ctx context.Context, keys map[st
 			keys[fieldKeyName] = append(keys[fieldKeyName], &fieldKey)
 		}
 	}
+
```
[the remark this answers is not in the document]
tushar-signoz: Its not an issue with the latest golang versions.

pkg/querybuilder/collision.go
```diff
@@ -0,0 +1,258 @@
...
+
+		// Check if there is any matching key in the metadata with the same name and it is not the same intrinsic/calculated field
+		match := false
+		for _, mapKey := range keys[key.Name] {
+			// Either field context is unspecified or matches
+			// and
+			// Either field data type is unspecified or matches
+			if (key.FieldContext == telemetrytypes.FieldContextUnspecified || mapKey.FieldContext == key.FieldContext) &&
+				(key.FieldDataType == telemetrytypes.FieldDataTypeUnspecified || mapKey.FieldDataType == key.FieldDataType) &&
+				mapKey.String() != intrinsicOrCalculatedField.String() {
```
srikanthccv: Why do we have `key.FieldContext == telemetrytypes.FieldContextUnspecified` / `key.FieldDataType == telemetrytypes.FieldDataTypeUnspecified`/ `mapKey.String() != intrinsicOrCalculatedField.String()` here?

We want a map key that exactly matches key?

pkg/querybuilder/collision.go
```diff
@@ -0,0 +1,258 @@
...
+
+		// We don't have a match, then it doesn't exist in attribute or resource attribute
+		// use the intrinsic/calculated field
+		if !match {
+			// This is the case where user is using an intrinsic/calculated field
+			// with a context or data type that may or may not match the intrinsic/calculated field
+			// and there is no matching key in the metadata with the same name
+			// So we can safely override the context and data type
+
+			if key.String() != intrinsicOrCalculatedField.String() {
```
srikanthccv: 1. This compare is brittle and shouldn't be used. The stringer interface implementation should be generally for debugging but not as a means of comparison.
2. Additionally, why do we need additional comparison here?

pkg/querybuilder/collision_test.go
```diff
@@ -0,0 +1,631 @@
...
+			expectedKey: telemetrytypes.TelemetryFieldKey{
+				Name:          "trace_id",
+				FieldContext:  telemetrytypes.FieldContextUnspecified,
+				FieldDataType: telemetrytypes.FieldDataTypeString,
+				Materialized:  true,
+			},
+			expectedActions: []string{
+				"Overriding key: name=trace_id,context=attribute,datatype=string to name=trace_id,datatype=string,materialized=true",
+			},
+			description: "Intrinsic field with no metadata match should use intrinsic field properties",
```
srikanthccv: What does it mean to have an "intrinsic field with no no metadata"? I am trying to understand what does this test target because there is a valid FieldContext for all intrinsic or calculated fields.

pkg/querybuilder/collision_test.go
```diff
@@ -0,0 +1,631 @@
...
+			intrinsicOrCalculatedField: &telemetrytypes.TelemetryFieldKey{
+				Name:          "trace_id",
+				FieldContext:  telemetrytypes.FieldContextUnspecified,
+				FieldDataType: telemetrytypes.FieldDataTypeString,
+				Materialized:  true,
+			},
+			expectedKey: telemetrytypes.TelemetryFieldKey{
+				Name:          "trace_id",
+				FieldContext:  telemetrytypes.FieldContextUnspecified,
+				FieldDataType: telemetrytypes.FieldDataTypeUnspecified,
```
srikanthccv: Why is the data type set to `Unspecified` here?

pkg/querybuilder/collision_test.go
```diff
@@ -0,0 +1,631 @@
...
+						FieldContext:  telemetrytypes.FieldContextAttribute,
+						FieldDataType: telemetrytypes.FieldDataTypeString,
+						Materialized:  true,
+					},
+					{
+						Name:          "field",
+						FieldContext:  telemetrytypes.FieldContextAttribute,
+						FieldDataType: telemetrytypes.FieldDataTypeString,
+						Materialized:  false,
+					},
```
srikanthccv: It should not be possible to have list of keys that exactly same context and data type with materialised is false/true. Or can it happen ever with `metadata.GetKeys`? If it doesn't happen, we should avoid tests as they could mislead.

pkg/querybuilder/collision_test.go
```diff
@@ -0,0 +1,631 @@
...
+				Materialized:  true,
+			},
+			expectedKey: telemetrytypes.TelemetryFieldKey{
+				Name:          "duration",
+				FieldContext:  telemetrytypes.FieldContextAttribute,
+				FieldDataType: telemetrytypes.FieldDataTypeUnspecified,
+				Materialized:  false,
+			},
+			expectedActions: []string{},
+			description:     "User explicitly specified attribute.duration, should prefer metadata over intrinsic",
```
srikanthccv: Why did the data type remained unspecified?

pkg/telemetrylogs/statement_builder.go
```diff
@@ -125,88 +139,79 @@ func getKeySelectors(query qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]) []
...
-				overallMatch = true // because we found a match
-			} else {
-				b.logger.InfoContext(ctx, "overriding the field context and data type", "key", k.Name)
-				k.FieldContext = telemetrytypes.FieldContextBody
-				k.FieldDataType = telemetrytypes.FieldDataTypeString
-				k.JSONDataType = &telemetrytypes.String
-			}
+	if len(aliasExpressions) > 0 {
+		for idx := range query.GroupBy {
+			revertContextForAliasKeys(&query.GroupBy[idx].TelemetryFieldKey)
```
srikanthccv: We shouldn't be doing this for group by but only order by?

pkg/telemetrylogs/statement_builder.go
```diff
@@ -125,88 +139,79 @@ func getKeySelectors(query qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]) []
...
+		what user meant. Here we take a call that we chose alias over fieldName because if user sepcifically wants to order or group by
+		a that field, a different alias can be chosen.
+	*/
+
+	aliasExpressions := map[string]bool{}
+	for _, agg := range query.Aggregations {
+		if agg.Alias != "" {
+			aliasExpressions[agg.Alias] = true
 		}
+	}
```
srikanthccv: Populate this only when the requestType is not raw.

pkg/telemetrylogs/statement_builder.go
```diff
@@ -125,88 +139,79 @@ func getKeySelectors(query qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]) []
...
-			return match
+func (b *logQueryStatementBuilder) adjustKeys(ctx context.Context, keys map[string][]*telemetrytypes.TelemetryFieldKey, query qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]) qbtypes.QueryBuilderQuery[qbtypes.LogAggregation] {
+	/*
+		Adjust keys for alias expressions in aggregations
+		For example, if user is using `body.count` as an alias for aggregation and
+		Uses it in orderBy or group by, upstream code will convert it to just `count` with fieldContext as Body
+		But we need to adjust it back to `body.count` with fieldContext as unspecified
+
+		NOTE": One ambiguity here is that if user specified alias `body.count` is also a valid field then we're not sure
+		what user meant. Here we take a call that we chose alias over fieldName because if user sepcifically wants to order or group by
```
srikanthccv: While alias is one case, it's not always the case right? Consider a list view with order by `body.user_id`, the order by will be transformed into `context=body,name=user_id`, so there is no question of alias here, now what we do should be based on the metadata rather than the alias?

pkg/querybuilder/collision.go
tushar-signoz: lets say user wants to select/order by `trace_id`. Now this could mean `log.trace_id` or `attribute.trace_id`.
So from the if condition  `if intrinsicOrCalculatedField != nil ` that the name matches with a top level field, our example of trace_id holds in this case. Now we have to check if it also matches with an attribute or resource, since the user didn't provide any context, we have to accommodate for that as well. 
`mapKey.String() != intrinsicOrCalculatedField.String()` is to ensure we are not matching with the intrinsic key again. 
We cannot to do a strict matching since user doesn't know about context prefixes. 
In simpler terms, we're checking if user definitely meant top level field without a doubt without user providing context or data type hint.

pkg/querybuilder/collision.go
tushar-signoz: Its not strictly needed, only to avoid logging if user has specified everything correctly and we don't have to change anything

I'll change `String()` compare.

pkg/querybuilder/collision.go
```diff
@@ -0,0 +1,258 @@
...
+			key.JSONDataType = matchingKey.JSONDataType
+			key.Indexes = matchingKey.Indexes
+			key.Materialized = matchingKey.Materialized
+
+			return actions
+		} else {
+			// multiple matching keys, set materialized only if all the keys are materialized
+			// TODO: This could all be redundant if it is not, it should be.
+			// Downstream query builder should handle multiple matching keys with their own metadata
+			// and not rely on this function to do so.
```
srikanthccv: This is correct, example, there could be two keys one is mat and other is not, then the downstream should be using the individual meta entry to prepare the right field for. Can you verify this once?

pkg/telemetrytraces/statement_builder.go
```diff
@@ -126,23 +126,29 @@ func getKeySelectors(query qbtypes.QueryBuilderQuery[qbtypes.TraceAggregation])
...
 	for idx := range query.Order {
 		keySelectors = append(keySelectors, &telemetrytypes.FieldKeySelector{
-			Name:         query.Order[idx].Key.Name,
-			Signal:       telemetrytypes.SignalTraces,
-			FieldContext: query.Order[idx].Key.FieldContext,
+			Name:          query.Order[idx].Key.Name,
+			Signal:        telemetrytypes.SignalTraces,
+			FieldContext:  query.Order[idx].Key.FieldContext,
+			FieldDataType: query.Order[idx].Key.FieldDataType,
 		})
```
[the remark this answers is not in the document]
tushar-signoz: getKeys already takes care of this

vikrantgupta25 (approved):

piyushsingariya (approved):

PR #9931, opened 2026-01-05
feat(authz): migrate existing rbac to openfga

vikrantgupta25:
### Summary 

- migrate existing rbac system to openfga

closes: https://github.com/SigNoz/platform-pod/issues/1278


pkg/authz/openfgaauthz/provider.go
```diff
@@ -334,3 +349,51 @@ func (provider *provider) getStoreIDandModelID() (string, string) {
 
 	return storeID, modelID
 }
+
+func (provider *provider) migrateExistingUsers(ctx context.Context) error {
```
vikrantgupta25: this doesn't feel right here and as a result of this `Store` interface came into pkg/authz, do something about this

pkg/authz/authz.go
```diff
@@ -29,3 +32,11 @@ type AuthZ interface {
 	// Lists the selectors for objects assigned to subject (s) with relation (r) on resource (s)
 	ListObjects(context.Context, string, authtypes.Relation, authtypes.Typeable) ([]*authtypes.Object, error)
 }
+
```
vikrantgupta25: result of `migrateExistingUsers` in provider to prevent circular imports.

pkg/authz/openfgaauthz/provider.go
vikrantgupta25: this needs to happen on startup, sqlmigration seems like a nice place to add but authz service starts post the migration is complete.

pkg/modules/user/impluser/module.go
```diff
@@ -413,13 +441,34 @@ func (module *Module) CreateFirstUser(ctx context.Context, organization *types.O
 		return nil, err
 	}
 
+	managedRoles := roletypes.NewManagedRoles(organization.ID)
```
vikrantgupta25, line 506: all of this needs to be done because we cannot run openfga writes and checks inside the transactions

pkg/authz/authz.go
vikrantgupta25: removed this and made setter and getter for role module

pkg/authz/openfgaauthz/provider.go
vikrantgupta25: sqlmigration is also hard because then we need to start and stop and independent service there, a service should not start without the sqlmigrations be completed as it can lead to bad schema access

pkg/signoz/signoz.go
```diff
@@ -377,9 +382,10 @@ func New(
 	}
 
 	// Initialize all modules
-	roleModule := implrole.NewModule(implrole.NewStore(sqlstore), authz, nil)
-	dashboardModule := dashboardModuleCallback(sqlstore, providerSettings, analytics, orgGetter, roleModule, queryParser, querier, licensing)
-	modules := NewModules(sqlstore, tokenizer, emailing, providerSettings, orgGetter, alertmanager, analytics, querier, telemetrystore, telemetryMetadataStore, authNs, authz, cache, queryParser, config, dashboardModule)
+	roleSetter := roleSetterCallback(sqlstore, authz, licensing, nil)
+	grant := implrole.NewGrant(implrole.NewStore(sqlstore), authz)
+	dashboard := dashboardModuleCallback(sqlstore, providerSettings, analytics, orgGetter, roleSetter, grant, queryParser, querier, licensing)
```
vikrantgupta25: import for roleSetter and grant module is gonna go away with https://github.com/SigNoz/platform-pod/issues/1530

vikrantgupta25: no one is gonna be importing role setter, everyone is gonna register them here. grant is also gonna be restricted to user module more or less let's see

pkg/types/roletypes/store.go
```diff
@@ -3,14 +3,16 @@ package roletypes
...
 	"github.com/SigNoz/signoz/pkg/valuer"
 )
 
 type Store interface {
 	Create(context.Context, *StorableRole) error
 	Get(context.Context, valuer.UUID, valuer.UUID) (*StorableRole, error)
-	GetByNameAndOrgID(context.Context, string, valuer.UUID) (*StorableRole, error)
+	GetByOrgIDAndName(context.Context, string, valuer.UUID) (*StorableRole, error)
 	List(context.Context, valuer.UUID) ([]*StorableRole, error)
+	ListUsersByOrgID(context.Context, valuer.UUID) ([]*types.User, error)
```
therealpandey: I think this is unused. Please remove.
```suggestion
```

Alternatively if we need this somewhere, better to call the userGetter?

pkg/types/roletypes/role.go
therealpandey:
```suggestion
	OrgID       valuer.UUID `json:"orgId"`
```

pkg/types/roletypes/role.go
```diff
@@ -32,8 +33,22 @@ var (
 )
```
therealpandey, line 33:
```suggestion
type RoleType struct { valuer.String }
```

pkg/modules/role/implrole/module.go
```diff
@@ -2,7 +2,6 @@ package implrole
 
 import (
 	"context"
-	"slices"
```
therealpandey, line 5: Rename file to `setter.go`

pkg/authz/openfgaauthz/provider.go
```diff
@@ -334,3 +353,51 @@ func (provider *provider) getStoreIDandModelID() (string, string) {
...
+	orgs, err := provider.orgGetter.ListByOwnedKeyRange(ctx)
```
therealpandey: On every signoz restart, we will keep executing this.

Can we check if this has already executed and skip?

therealpandey (changes requested):

pkg/types/roletypes/store.go
vikrantgupta25: this was before the separation of the module hence the DI wasn't an option but not needed now, removed

pkg/types/roletypes/role.go
vikrantgupta25: done

pkg/modules/role/implrole/module.go
vikrantgupta25: ahh it got reverted with the last commit, done

pkg/authz/openfgaauthz/provider.go
vikrantgupta25: since it's idempotant it's safe to run this, maintaining an inmemory flag is is is

pkg/http/middleware/authz.go
```diff
@@ -129,7 +251,18 @@ func (middleware *AuthZ) Check(next http.HandlerFunc, relation authtypes.Relatio
 			return
 		}
 
-		err = middleware.authzService.CheckWithTupleCreation(req.Context(), claims, orgId, relation, translation, typeable, selectors)
+		roles, err := middleware.roleGetter.ListByOrgIDAndNames(req.Context(), orgId, roles)
```
vikrantgupta25, line 254: this takes care of fine grained transactions for resource or role based for pkg

pkg/http/middleware/authz.go
```diff
@@ -129,7 +251,18 @@ func (middleware *AuthZ) Check(next http.HandlerFunc, relation authtypes.Relatio
...
+		if err != nil {
+			render.Error(rw, err)
+			return
+		}
+
+		roleSelectors := []authtypes.Selector{}
```
vikrantgupta25, line 260: if any handler forgets to pass the roles for pkg or passes empty slice and the cb is available for the build, then the authz server will fail with 5xx ( which is expected since we made a mistake ) but need to figure out how can this be pushed to make a build time check

vikrantgupta25: can do MustNewRoles() with name checks, but it's fine for now.

vikrantgupta25:
superseded by: https://github.com/SigNoz/signoz/pull/10121 and the next integrations PR

PR #9975, opened 2026-01-11
fix: add support for minimum data points in PromQL alerts

### LINKED ISSUE #8919: require minimum number of points config doesn't work for promql alerts
we allow ability to configure requiring minimum data points in each result group. However, this setting is not implemented for promql based alerts.

jatinderjit:
## 📄 Summary

The metric based alert rules allow configuring a minimum number of data points.
This works for Clickhouse, but not for PromQL queries. This PR fixes that.

---

## ✅ Changes

- Change `Rule.Eval` signature to return `int, error` instead of `any, error`.
- Add option to set `LookbackDelta` in Prometheus engine
- Bug fix: Fix PromQL alerts to respect the configuration to require minimum number of data points.

---

## 📝 Changelog 

**Deployment Type:** Cloud / OSS / Enterprise

**Type:** Bug Fix

**Description:** The frequency of alerts fired could reduce, depending on the configured rules. (Ref: the summary above).

---

## 🧪 How to Test

1. Create alert rules using PromQL and configure minimum data points required.
2. When the threshold breaches, the alert shouldn't fire if the number of samples are less than the configured number.
3. But it should fire if the number of samples exceed the configured number of data points.

---

## 🔍 Related Issues

- Closes #8919

---

## 📋 Checklist

- [ ] Dev Review

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> - **Min data points enforcement**: Introduces `RuleCondition.ShouldEval` and applies it in `anomaly.go`, `threshold_rule.go`, and `prom_rule.go` to skip series with fewer than required samples.
> - **API change**: Changes `Rule.Eval` to return `(int, error)`; updates implementations (`PromRule`, `ThresholdRule`, `AnomalyRule`) and callers in `manager.go` and test notification code.
> - **Prometheus config**: Adds `LookbackDelta` to `pkg/prometheus/config.go` and wires it into the engine in `engine.go`.
> - **Tests**: Adds/updates unit tests to cover min-points behavior for PromQL and Threshold rules; adjusts existing tests to new return type and minor cleanups.
> - **Misc**: Minor logging, typos, and comment improvements; small refactors (e.g., `hostFromSource` placement).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 85d2df3ae198d7791a083c7733eea68f424e9ff7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


srikanthccv:
Let's add some unit tests @jatinderjit.

pkg/prometheus/config.go
```diff
@@ -10,6 +14,10 @@ type ActiveQueryTrackerConfig struct {
 
 type Config struct {
 	ActiveQueryTrackerConfig ActiveQueryTrackerConfig `mapstructure:"active_query_tracker"`
+
+	// LookbackDelta determines the time since the last sample after which a time
+	// series is considered stale.
+	LookbackDelta time.Duration
```
srikanthccv: The `mapstructure` tag is missing here, i.e., not really configurable.

pkg/prometheus/engine.go
```diff
@@ -20,8 +20,9 @@ func NewEngine(logger *slog.Logger, cfg Config) *Engine {
 	return promql.NewEngine(promql.EngineOpts{
 		Logger:             logger,
 		Reg:                nil,
-		MaxSamples:         50000000,
-		Timeout:            time.Duration(2 * time.Minute),
+		MaxSamples:         5_0000_000,
+		Timeout:            2 * time.Minute,
 		ActiveQueryTracker: activeQueryTracker,
+		LookbackDelta:      cfg.LookbackDelta,
```
srikanthccv, line 26: What would be the behaviour if the lookback delta is not explicitly configured by user?

pkg/query-service/rules/base_rule.go
```diff
@@ -258,17 +255,6 @@ func (r *BaseRule) HoldDuration() time.Duration {
 	return r.holdDuration
 }
 
-func (r *ThresholdRule) hostFromSource() string {
```
srikanthccv, line 261: Hmm, this is supposed to be used by all alert types to generate a link to send in the notification. How is the link generation working in prom and anomaly rules?

pkg/query-service/rules/prom_rule_test.go
```diff
@@ -1341,3 +1344,197 @@ func TestMultipleThresholdPromRule(t *testing.T) {
...
+			requiredNumPoints: 2,
+			values: []sample{
+				{baseTime.Add(-10 * time.Minute), 100.0},
+				{baseTime.Add(-7 * time.Minute), 150.0},
+				{baseTime.Add(-6 * time.Minute), 250.0},
+			},
+			target:       200,
+			expectAlerts: 0,
+		},
+		{
```
srikanthccv: Can we add one more case here where it meets the target but gets skipped because it doesn't match the number of points in the series?

While this test here does sort of the same thing, but I think the `series` for this test comes out as 0 points, so no series to begin with?

pkg/query-service/rules/base_rule.go
jatinderjit: It's used for links to related logs/traces. These aren't being added for promql and anomaly.
I don't yet know about anomaly, but does it apply to promql?

pkg/query-service/rules/prom_rule_test.go
jatinderjit: Done

pkg/prometheus/config.go
jatinderjit: Thanks.. I missed it!
Added now.

pkg/prometheus/engine.go
jatinderjit: Prometheus' default is used ([ref](https://github.com/prometheus/prometheus/blob/664b25569945aa21f8cae8dfd4d83891aa668cef/promql/engine.go#L434)). Updated the doc comment.

pkg/query-service/rules/base_rule.go
srikanthccv: Not at the moment, but eventually one day, we should add related logs/traces/metrics for the promql alerts as well.

pkg/types/ruletypes/alerting.go
```diff
@@ -202,7 +202,15 @@ func (rc *RuleCondition) IsValid() bool {
 	return true
 }
 
-// QueryType is a short hand method to get query type
+// ShouldEval checks if the further series should be evaluated at all for alerts.
+func (rc *RuleCondition) ShouldEval(series *v3.Series) bool {
+	if rc == nil {
+		return true
+	}
```
srikanthccv, line 209: Technically, this should never be the case otherwise alert rule doesn't make sense but it can be there till the time we have full validation rules put in place.

PR #9976, opened 2026-01-11
test(integration): add cumulative counter rate tests

srikanthccv:
## 📄 Summary

- Add handwritten jsonl data for exact values matching in the response
- Add cumulative counter rate tests
  - group by endpoint
  - with filter
- Update fixture to write to missing tables
- Add functions to work with querier stuff easily


tests/integration/fixtures/querier.py
```diff
@@ -0,0 +1,259 @@
+from typing import Any, Dict, List, Optional
```
therealpandey, line 2: Can we move this file to the fixtures directory?

1. They seem like reusable functions which can be used across the test suite.

2. If we want to change the layout of the actual tests, it will help.

srikanthccv: It doesn't make sense to put them under fixtures. Nothing about them qualifies to put under fixture (things that manage shared state, pytest's dependency injection or that need setup/teardown). They are mainly used to assist with the querier functionality, If something needs them, they can always do `from src.querier import ...` like the querier tests themselves do.

therealpandey: 1. Fixtures are not just about shared state or pytest dependencies. They can contain functions which are used by the test suite in `src`. For example, `add_license` or `metrics_generators`.

2. In general, we do not want test suites importing one and another. Tests shouldn't depend on tests but on fixtures only.

3. The organization of the `src` or actual tests will keep evolving. We will keep moving tests from package A to package B etc. Having shared functionality in fixtures helps with that.

srikanthccv: > Fixtures are not just about shared state or pytest dependencies.

Yes, they are. They setup the thing needed to run the test. The test fixtures are not a dumping ground for any helpers.

> They can contain functions which are used by the test suite in src. For example, add_license or metrics_generators.

- metrics_generators are data fixtures that are yet to be used but they are going to fixtures some time later.

>In general, we do not want test suites importing one and another. Tests shouldn't depend on tests but on fixtures only.

>The organization of the src or actual tests will keep evolving. We will keep moving tests from package A to package B etc. Having shared functionality in fixtures helps with that.

Tests are not dependent on tests. Tests are dependent on package that offers a certain functionality. When we initially talked about choosing go vs Python, this was my point that we end writing the packages that would have avoided or should be in golang but we decided to write in Python. While at the moment the querier contains the flat tests, doesn't mean `src/querier` is just tests, it'a a package the contains tests and also offers a functionality that other packages depend on and other packages can also offer.

therealpandey: So where do you think `add_license` should be?

srikanthccv: My current understanding of licensing is limited so not sure about that, perhaps a license package `src/license`. I assume there is more to licensing than what is doing now with `add_license`?

therealpandey: But dont you think src/license should have licence related tests?

We're calling it src/<test-suite>. I think maybe the confusion is spreading from the term querier. It's not a package. It's a test suite name. You can tomorrow name it src/jsonlogs, src/cummetrics, src/histmetrics.

What do you think?

therealpandey (approved):

PR #9977, opened 2026-01-11
test(integration): add cache tests with cumulative counter metrics

srikanthccv:
## 📄 Summary

Covering the same cases as bucket cache tests but through integration tests

- Right overlap with cache
- Left overlap with cache
- Subset with cache
- Superset with cache
- Left + Right overlap with gap in middle with cache
- Sliding/Continuous data fetching


srikanthccv:
Superseded by https://github.com/SigNoz/signoz/pull/12813

PR #9981, opened 2026-01-12
feat: Pin log detail -> trace correlation button to top or more discoverable

ishan621:
## 📄 Summary

Users were unable to easily discover the Logs → Traces action since it was located at the bottom of the view. To improve discoverability and accessibility, we have moved (pinned) this action to the top of the interface.

## ✅ Changes

We now conditionally pin the trace_id attribute only when it is present in the log data. This ensures that trace-related context is highlighted for logs that are associated with traces, while avoiding unnecessary or empty pinned attributes for logs without a trace reference.

---

## 🧪 How to Test

1. Open app.us.staging.signoz.cloud and navigate to Logs Explorer https://app.us.staging.signoz.cloud/logs/logs-explorer
2. Open any log detail view 
3. If traceId is present, it should be pinned and onClick should redirect as expected

---

## 🔍 Related Issues

Closes #https://github.com/SigNoz/engineering-pod/issues/3356

---

Before: 

https://github.com/user-attachments/assets/95cfa3d1-0320-4a4e-8935-d05dd0484935

After: 
https://github.com/user-attachments/assets/252cc3b2-7694-4a21-9891-da89a3b624a2

## 📋 Checklist

- [x] Dev Review

### DIFF
--- a/frontend/src/lib/query/createTableColumnsFromQuery.ts
+++ b/frontend/src/lib/query/createTableColumnsFromQuery.ts
@@ -491,16 +491,18 @@ const fillDataFromList = (
 	listItem: ListItem,
 	columns: DynamicColumns,
 ): void => {
+	const listData = listItem.data || {};
 	columns.forEach((column) => {
 		if (isFormula(column.field)) return;
 
-		Object.keys(listItem.data).forEach((label) => {
+		Object.keys(listData).forEach((label) => {
 			if (column.dataIndex === label) {
-				if (listItem.data[label as ListItemKey] !== '') {
-					if (isObject(listItem.data[label as ListItemKey])) {
-						column.data.push(JSON.stringify(listItem.data[label as ListItemKey]));
+				const listValue = listData[label as ListItemKey];
+				if (listValue !== '') {
+					if (isObject(listValue)) {
+						column.data.push(JSON.stringify(listValue));
 					} else {
-						column.data.push(listItem.data[label as ListItemKey].toString());
+						column.data.push(listValue?.toString() ?? '');
 					}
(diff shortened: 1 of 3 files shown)


frontend/src/lib/query/createTableColumnsFromQuery.ts
```diff
@@ -491,16 +491,18 @@ const fillDataFromList = (
 	listItem: ListItem,
 	columns: DynamicColumns,
 ): void => {
+	const listData = listItem.data || {};
```
AshwinBhatkal, line 494: Is this a different change?

ishan621: This file failed ESLint null checks during commit; no logic changes were needed, but the lint issues have been fixed.

frontend/src/types/api/logs/log.ts
```diff
@@ -17,6 +17,7 @@ export interface ILog {
 	attributesFloat: Record<string, never>;
 	severity_text: string;
 	severity_number: number;
+	trace_id?: string;
```
AshwinBhatkal, line 20: what is the difference between trace_id and traceId

ishan621: trace_id often comes directly from the backend/OTLP payload; 
traceId is the camelCase version we use in the frontend/types. They refer to the same trace identifier, just with different naming conventions.
In this case , we had to have a check on backend payload response hence introduced trace_id in types. 

We can plan a cleanup to have same attributes name in frontend as they come from backend , but that would be bigger task which we can plan

PR #9982, opened 2026-01-12
chore: fill y axis unit in alerts coming from dashboards

amlannandy:
## 📄 Summary

- To fill y-axis unit in alerts when coming from dashboards.
- If we are in dashboards panel creation page and have a selected a y axis unit, then we go to alerts from there, then the y axis unit value should carry over.

Additional Info

- This flow is also working from Logs Explorer without any additional changes, as the query being exported already has the Y axis unit included
- But Traces Explorer doesn't do that, so it's not working there. We can probably file a separate ticket for that and assign to the relevant codeowner.

---

## 🧪 How to Test

1. Go to Create Panel page in Dashboards
2. Write a query and select a Y Axis Unit
3. Click on go to Alerts
4. It should land on Alerts with the value pre-filled
5. Please refer to recording for more reference

---

## 🔍 Related Issues

Closes https://github.com/SigNoz/engineering-pod/issues/2238

---

For Logs

https://github.com/user-attachments/assets/53ea441b-2f73-4533-ae0b-7d5cdd7de33e

For Traces

https://github.com/user-attachments/assets/2f43deda-c208-4e96-9d08-e17705fc9bb9

For Metrics

https://github.com/user-attachments/assets/6e84d4eb-ba65-4793-8346-0789c29e138e

---

## 📋 Checklist

- [x] Dev Review


frontend/src/constants/queryBuilder.ts
```diff
@@ -301,6 +301,7 @@ export const initialQueryState: QueryState = {
 	builder: initialQueryBuilderData,
 	clickhouse_sql: [initialClickHouseData],
 	promql: [initialQueryPromQLData],
+	unit: '',
```
amlannandy, line 304: When we were landing on alerts page, the compositeQuery from url was having the `unit` field ommitted from `currentQuery` due to the logic in `replaceIncorrectObjectFields` which would remove keys which are not present in the default state. So added the `unit` field in the default initial state

frontend/src/container/FormAlertRules/index.tsx
```diff
@@ -788,11 +788,18 @@ function FormAlertRules({
...
+	const source = useMemo(() => urlQuery.get(QueryParams.source) as YAxisSource, [
+		urlQuery,
+	]);
+
 	// Only update automatically when creating a new metrics-based alert rule
-	const shouldUpdateYAxisUnit = useMemo(
-		() => isNewRule && alertType === AlertTypes.METRICS_BASED_ALERT,
-		[isNewRule, alertType],
-	);
+	const shouldUpdateYAxisUnit = useMemo(() => {
```
amlannandy, line 796: When we come from dashboards for metrics based queries, we want to preserve the choice user made in dashboards instead of overriding it. Hence the check here.
Please let me know if there's a better way to handle this

AshwinBhatkal: This should be a good enough way

AshwinBhatkal: Should this be for all fields instead of just YAxis? I'm looking at targeting this one as well  - https://github.com/SigNoz/signoz/issues/5291

amlannandy: We should be be able to pass thresholds as well like this. Only concern would be if we're polluting the url too much. 
But we'll do this in a separate Pr right?

AshwinBhatkal: Yes. we can do it in a separate PR. The url is the best way in this case. Else we need some temporary BE store just to navigate.

AshwinBhatkal (approved):

PR #9985, opened 2026-01-12
chore: make settings drawer reusable

AshwinBhatkal:
## 📄 Summary

Part of consolidating code around and leading to dashboard variables

---

## ✅ Changes

- Made SettingsDrawer reusable. Children can now be passed into it.

---

https://github.com/user-attachments/assets/484a41d1-79f0-49ed-aa9b-98a82ee86894

---

## 📋 Checklist

- [x] Dev Review

---

> [!NOTE]
> Makes dashboard settings Drawer reusable and controlled, and integrates it across dashboard views.
> 
> - Introduces reusable `SettingsDrawer` (`isOpen`, `onClose`, `children`) in `DashboardDescription/SettingsDrawer/index.tsx` with new styles in `SettingsDrawer.styles.scss`
> - Updates `DashboardDescription` and `DashboardEmptyState` to show a "Configure" button, manage drawer state, and render `DashboardSettings` as the drawer’s content
> - Adds `VariablesSettingsTab` handle to reset the variables tab on drawer close; wires through `DashboardSettings` and `DashboardSettings/Variables`
> - Cleans up styles by removing drawer-related rules from `Description.styles.scss` and deleting obsolete `styles.ts`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c10d8bbd8f4d8ef5044b7b90afad93a1ef19d3e8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/src/container/DashboardContainer/DashboardDescription/SettingsDrawer/SettingsDrawer.styles.scss
+++ b/frontend/src/container/DashboardContainer/DashboardDescription/SettingsDrawer/SettingsDrawer.styles.scss
@@ -21,7 +21,7 @@
 					line-height: 20px; /* 142.857% */
 					letter-spacing: -0.07px;
 					padding-left: 16px;
-					border-left: 1px solid #161922;
+					border-left: 1px solid var(--bg-slate-500);
 				}
 
 				.ant-drawer-close {


frontend/src/container/DashboardContainer/DashboardSettings/Variables/index.tsx
```diff
@@ -117,6 +114,15 @@ function VariablesSetting({
...
 	};
 
+	// Reset state when component moves out of viewport
+	const containerRef = useOnOutOfViewport({
+		onOutOfViewport: () => {
+			setVariableViewMode(null);
+			setVariableEditData(null);
+		},
+		enabled: variableViewMode !== null,
+	});
```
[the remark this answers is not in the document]
AshwinBhatkal: This is okay for the current setup. Currently, the reset is required mainly on close. Given the other tabs, Overview & Publish, it does not make sense to retain the state when navigating between these tabs. So a reset is good in this case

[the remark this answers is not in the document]
AshwinBhatkal: Thinking of a solution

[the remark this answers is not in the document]
AshwinBhatkal: The best way would be to not allow navigation to the other tab in case of any edit that has happened. If they do still navigate, the state would be lost. If they don't make any change, it should be a pure switch

frontend/src/container/GridCardLayout/DashboardEmptyState/DashboardEmptyState.tsx
```diff
@@ -44,6 +52,15 @@ export default function DashboardEmptyState(): JSX.Element {
...
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [handleToggleDashboardSlider]);
+
+	const onConfigureClick = useCallback((): void => {
+		setIsSettingsDrawerOpen(true);
+	}, []);
+
+	const onSettingsDrawerClose = useCallback((): void => {
+		setIsSettingsDrawerOpen(false);
+	}, []);
```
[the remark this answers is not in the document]
AshwinBhatkal: Fixed. Thanks

frontend/src/container/DashboardContainer/DashboardDescription/SettingsDrawer/SettingsDrawer.styles.scss
```diff
@@ -0,0 +1,67 @@
...
+				.ant-drawer-title {
+					color: var(--bg-vanilla-400);
+					font-family: Inter;
+					font-size: 14px;
+					font-style: normal;
+					font-weight: 400;
+					line-height: 20px; /* 142.857% */
+					letter-spacing: -0.07px;
+					padding-left: 16px;
+					border-left: 1px solid #161922;
```
ahrefabhi: can you use variable `bg-slate-500` here instead of the hash value directly?

frontend/src/container/DashboardContainer/DashboardDescription/SettingsDrawer/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+}>;
+
+function SettingsDrawer({
+	children,
+	drawerTitle,
+	isOpen,
+	onClose,
+}: SettingsDrawerProps): JSX.Element {
+	return (
+		<Drawer
```
ahrefabhi, line 20: we should unmount the child component when the drawer is closed, add `destroyOnHidden` prop to the Drawer component

frontend/src/container/DashboardContainer/DashboardDescription/index.tsx
```diff
@@ -340,6 +348,18 @@ function DashboardDescription(props: DashboardDescriptionProps): JSX.Element {
 		publicDashboardResponse?.data,
 	]);
 
+	const onConfigureClick = useCallback((): void => {
```
ahrefabhi, line 351: is useCallback required here? we should not over-use useCallback or useMemo, this function is cheap to create on re-render so there is no need to memoize it.

can you check other places also where we've used useCallback

frontend/src/container/DashboardContainer/DashboardDescription/SettingsDrawer/SettingsDrawer.styles.scss
AshwinBhatkal: done

frontend/src/container/DashboardContainer/DashboardDescription/SettingsDrawer/index.tsx
AshwinBhatkal: I wanted to do the same. But this behaviour is intentionally maintained. This is along lines of good UX as well. I read a lot about this and then came to not adding that prop. See the screen recording. Specifically handled for it

frontend/src/container/DashboardContainer/DashboardDescription/index.tsx
AshwinBhatkal: It is required here
If you see below, these functions are being passed as props to the components. If we don't use useCallback here, on every render when function ref changes, the children also will re render

ahrefabhi (approved):

PR #9991, opened 2026-01-13
feat: export traces

tushar-signoz:
## 📄 Summary

This pull request introduces a new API endpoint for exporting raw trace and log data, along with the backend implementation to support flexible export queries and formats. The changes include OpenAPI documentation, handler wiring, and a comprehensive implementation for exporting traces and logs, including support for advanced composite queries and generic export logic.

**Key changes:**

#### API & Routing

* Added a new GET endpoint `/api/v1/export_raw_data` to the OpenAPI specification for exporting raw data for traces and logs, including response types and error handling.
* Registered the new handler in the API server provider, wiring up the `rawDataExportHandler` and its routes. [[1]](diffhunk://#diff-ce7e9c36290433fb906d4ccda6becaf3cceb6d8a00e0dce2ce78253fe0db3bb2R19) [[2]](diffhunk://#diff-ce7e9c36290433fb906d4ccda6becaf3cceb6d8a00e0dce2ce78253fe0db3bb2R43) [[3]](diffhunk://#diff-ce7e9c36290433fb906d4ccda6becaf3cceb6d8a00e0dce2ce78253fe0db3bb2R60-R63) [[4]](diffhunk://#diff-ce7e9c36290433fb906d4ccda6becaf3cceb6d8a00e0dce2ce78253fe0db3bb2R84) [[5]](diffhunk://#diff-ce7e9c36290433fb906d4ccda6becaf3cceb6d8a00e0dce2ce78253fe0db3bb2R104) [[6]](diffhunk://#diff-ce7e9c36290433fb906d4ccda6becaf3cceb6d8a00e0dce2ce78253fe0db3bb2R161-R164) [[7]](diffhunk://#diff-5def45b7ded5a95dd3df92a34e9322823f0f7ce0ba935610fcbc0d38d1847002R1-R28) [[8]](diffhunk://#diff-75f064d087295e3ead4026a5b3eef5b64f4f0b4006b662b9691dfcc8ebdfae32R21) [[9]](diffhunk://#diff-75f064d087295e3ead4026a5b3eef5b64f4f0b4006b662b9691dfcc8ebdfae32R51) [[10]](diffhunk://#diff-89b6da093da3aeb0a757415691ec84471db6329c7a97e05ae9b6cab457f56b56R250)
* Deprecated the previous registration of the export endpoint in the query service handler to avoid duplicate routing.

#### Export Functionality

* Implemented trace export logic in `exportTraces`, supporting both default and advanced composite queries, and output in CSV or JSONL formats. This includes validation of query limits, query types, and dynamic content headers.
* Refactored log export to use generic export functions (`exportRawDataCSV`, `exportRawDataJSONL`) for consistency and code reuse across logs and traces. [[1]](diffhunk://#diff-fe2c64b44243c4f5c2ff110ed68eca24762609ff84eaea3c658314977d93d227L209-R383) [[2]](diffhunk://#diff-fe2c64b44243c4f5c2ff110ed68eca24762609ff84eaea3c658314977d93d227L230-R397) [[3]](diffhunk://#diff-fe2c64b44243c4f5c2ff110ed68eca24762609ff84eaea3c658314977d93d227L271-R439)

#### Query Handling & Validation

* Added support for a `composite_query` parameter, allowing advanced, JSON-encoded queries for trace export, overriding filter/columns/order_by/limit if present. [[1]](diffhunk://#diff-fe2c64b44243c4f5c2ff110ed68eca24762609ff84eaea3c658314977d93d227R63-R69) [[2]](diffhunk://#diff-fe2c64b44243c4f5c2ff110ed68eca24762609ff84eaea3c658314977d93d227R96-R100)
* Implemented helper functions to parse and validate order-by parameters for traces, and to extract composite queries from request parameters.

#### Bug Fixes & Improvements

* Enabled trace export by updating the source validation logic. (F630b284L473R473)

These changes collectively enable robust, flexible raw data export for both logs and traces, with extensible query capabilities and improved code maintainability.

---

## ✅ Changes

---

## 📝 Changelog 

> Fill this only if the change affects users, APIs, UI, or documented behavior.
Mention as N/A for internal refactors or non-user-visible changes.

**Deployment Type:** Cloud / OSS / Enterprise

**Type:** Feature / Bug Fix / Maintenance

**Description:** Short, user-facing summary of the change

---

ex:

- `frontend`
- `backend`
- `devops`
- `bug`
- `enhancement`
- `ui`
- `test`

---

## 👥 Reviewers

- frontend / backend / devops

---

## 🧪 How to Test

1. ...
2. ...
3. ...

---

## 🔍 Related Issues

Closes #

---

---

## 📋 Checklist

- [ ] Dev Review

---


pkg/types/rawdataexport.go
```diff
@@ -0,0 +1,39 @@
...
+	Columns []string `json:"columns" schema:"columns"`
+
+	// OrderBy specifies the sorting order
+	// Format: "column:direction" or "context.field:type:direction"
+	// Direction can be "asc" or "desc"
+	OrderBy string `json:"order_by" schema:"order_by"`
+
+	// CompositeQuery is an advanced query specification as JSON-encoded QueryEnvelope array
+	// When provided, this overrides filter, columns, order_by, and limit parameters
+	CompositeQuery []qbtypes.QueryEnvelope `json:"composite_query" schema:"composite_query"`
```
srikanthccv: What does it mean to have more than one envelope here?

pkg/types/querybuildertypes/querybuildertypesv5/req.go
```diff
@@ -450,3 +451,74 @@ func (r *QueryRangeRequest) GetQueriesSupportingZeroDefault() map[string]bool {
...
+					Spec: spec.Copy(),
+				}
+			case QueryBuilderTraceOperator:
+				c.CompositeQuery.Queries[i] = QueryEnvelope{
+					Type: q.Type,
+					Spec: spec.Copy(),
+				}
+			default:
+				// Check if spec implements a Copy method
+				specValue := reflect.ValueOf(q.Spec)
```
srikanthccv: Let's remove this default; it can never reach a point where the spec type is not recognised?

pkg/telemetrytraces/const.go
```diff
@@ -373,4 +376,19 @@ var (
...
+				TelemetryFieldKey: DefaultFields["timestamp"],
+			},
+			Direction: qbtypes.OrderDirectionDesc,
+		},
+		{
+			Key: qbtypes.OrderByKey{
+				TelemetryFieldKey: DefaultFields["span_id"],
+			},
+			Direction: qbtypes.OrderDirectionDesc,
+		},
```
srikanthccv: Why is the `span_id` part of the order in the export but not the listing page?

pkg/query-service/app/http_handler.go
```diff
@@ -593,7 +593,7 @@ func (aH *APIHandler) RegisterRoutes(router *mux.Router, am *middleware.AuthZ) {
 	})).Methods(http.MethodGet)
 
 	// Export
-	router.HandleFunc("/api/v1/export_raw_data", am.ViewAccess(aH.Signoz.Handlers.RawDataExport.ExportRawData)).Methods(http.MethodGet)
+	// router.HandleFunc("/api/v1/export_raw_data", am.ViewAccess(aH.Signoz.Handlers.RawDataExport.ExportRawData)).Methods(http.MethodGet)
```
srikanthccv: nit: remove it if it's not needed

pkg/modules/rawdataexport/implrawdataexport/module.go
```diff
@@ -38,15 +154,21 @@ func (m *Module) ExportRawData(ctx context.Context, orgID valuer.UUID, rangeRequ
...
+			return
+		}
+
+		rowCountLimit := spec.Limit
 		rowCount := 0
 
 		for rowCount < rowCountLimit {
 			spec.Limit = min(ChunkSize, rowCountLimit-rowCount)
 			spec.Offset = rowCount
+			rangeRequest.CompositeQuery.Queries[traceOperatorIndex].Spec = spec
```
srikanthccv: Why is this assignment needed here. Isn't it redundant because the `rangeRequest.CompositeQuery.Queries[traceOperatorIndex].Spec` is already the exact same spec that was created at line 157.

pkg/modules/rawdataexport/implrawdataexport/module.go
```diff
@@ -23,8 +23,124 @@ func NewModule(querier querier.Querier) rawdataexport.Module {
...
+
+		appendQueryName := len(compositeQueries) > 1
+
+		for _, query := range compositeQueries {
+
+			dupRangeRequest := rangeRequest.Copy()
+			dupRangeRequest.CompositeQuery.Queries = []qbtypes.QueryEnvelope{query}
+			switch query.Spec.(type) {
+			case qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]:
+				exportRawDataForQueryBuilderQuery[qbtypes.LogAggregation](querier, contextWithTimeout, orgID, &dupRangeRequest, rowChan, errChan, doneChan, appendQueryName)
```
srikanthccv: Should we allow this? When would one want to have more than one query here?

pkg/modules/rawdataexport/implrawdataexport/module.go
```diff
@@ -23,8 +23,124 @@ func NewModule(querier querier.Querier) rawdataexport.Module {
...
+			case qbtypes.QueryBuilderQuery[qbtypes.MetricAggregation]:
```
srikanthccv: Metrics is not supported, let's remove it.

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -492,25 +663,78 @@ func getExportQueryOrderBy(queryParams url.Values) ([]qbtypes.OrderBy, error) {
...
+			},
+			Direction: orderBy[0].Direction,
+		})
+	}
+	return orderBy, nil
+}
+
+func getCompositeQueriesFromQueryParams(queryParams url.Values) ([]qbtypes.QueryEnvelope, error) {
+	// Check if composite_query parameter exists in query params
+	compositeQueryParams := queryParams["composite_query"]
```
srikanthccv: Not a good idea? Rather implement a POST method for the same API?

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -105,12 +118,169 @@ func (handler *handler) ExportRawData(rw http.ResponseWriter, r *http.Request) {
...
-func (handler *handler) exportMetrics(rw http.ResponseWriter, r *http.Request) {
+func (handler *handler) exportMetrics(rw http.ResponseWriter, _ *http.Request) {
 	render.Error(rw, errors.Newf(errors.TypeUnsupported, errors.CodeUnsupported, "metrics export is not yet supported"))
 }
 
 func (handler *handler) exportTraces(rw http.ResponseWriter, r *http.Request) {
-	render.Error(rw, errors.Newf(errors.TypeUnsupported, errors.CodeUnsupported, "traces export is not yet supported"))
+	// Set up response headers
+	rw.Header().Set("Cache-Control", "no-cache")
+	rw.Header().Set("Vary", "Accept-Encoding") // Indicate that response varies based on Accept-Encoding
```
srikanthccv: Do `exportTraces`/`exportLogs` deserve their own independent handler functions, one for each?

docs/api/openapi.yml
```diff
@@ -607,6 +607,34 @@ paths:
       summary: Update auth domain
       tags:
       - authdomains
+  /api/v1/export_raw_data:
```
srikanthccv: There is a recent change that added support for query params. See how it's done so the query params can be populated here.

pkg/types/querybuildertypes/querybuildertypesv5/req.go
tushar-signoz: IMO, since this method doesn't return any error, there's no way to bubble up that copy is not working, if someone ever adds a new spec and forgets to add it in the switch. 
Either we can chose to explicitly panic() if it doesn't match ( basically bubbling up the issue )
Or we can return a shallow copy which has a potential of creating bugs.
Or let the reflect thing be as is which still is not a perfect soln.

pkg/modules/rawdataexport/implrawdataexport/module.go
tushar-signoz: This is necessary, `spec` is a shallow copy of the actual spec when we type casted it. Im modifying the limit of the copy of the spec. Need to assign it back to make reflect it in the original spec.

tushar-signoz: https://go.dev/play/p/v1eJuC_lKVZ

tushar-signoz: Just to be sure, tested it

srikanthccv:
Let me know if this is ready for re-review.

tushar-signoz:
@srikanthccv Ready for review

pkg/types/querybuildertypes/querybuildertypesv5/req_getters.go
srikanthccv, line 1: absolutely no panic, anywhere, except the error is about a non-recoverable, stop everything situation (there is no such situation in this codebase yet)

pkg/apiserver/signozapiserver/rawdataexport.go
```diff
@@ -0,0 +1,46 @@
...
+		Request:             new(types.ExportRawDataQueryParams),
+		RequestContentType:  "application/json",
+		Response:            nil,
+		ResponseContentType: "application/json",
+		SuccessStatusCode:   http.StatusOK,
+		ErrorStatusCodes:    []int{http.StatusBadRequest},
+	})).Methods(http.MethodGet).GetError(); err != nil {
+		return err
+	}
+	if err := router.Handle("/api/v1/export_raw_data", handler.New(provider.authZ.ViewAccess(provider.rawDataExportHandler.ExportRawData), handler.OpenAPIDef{
```
nityanandagohain, line 15: This will just remain as an API right in our docs only the simple one will be integrated with frontend ?

From what it seems for traces we will have to use the advanced because of trace operator.

pkg/apiserver/signozapiserver/rawDataExport.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/gorilla/mux"
+)
+
+func (provider *provider) addRawDataExportRoutes(router *mux.Router) error {
+	if err := router.Handle("/api/v1/export_raw_data", handler.New(provider.authZ.ViewAccess(provider.rawDataExportHandler.ExportRawData), handler.OpenAPIDef{
+		ID:                  "HandleExportRawDataGET",
+		Tags:                []string{"logs", "traces"},
+		Summary:             "Export raw data",
+		Description:         "This endpoints allows simple query exporting raw data for traces and logs",
+		Request:             new(types.ExportRawDataQueryParams),
```
nityanandagohain: should it be `RequestQuery` instead, since it is a get request with url params.

request content type seems wrong as well

pkg/apiserver/signozapiserver/rawDataExport.go
```diff
@@ -0,0 +1,46 @@
...
+		RequestContentType:  "application/json",
+		Response:            nil,
+		ResponseContentType: "application/json",
+		SuccessStatusCode:   http.StatusOK,
+		ErrorStatusCodes:    []int{http.StatusBadRequest},
```
nityanandagohain: SecuritySchemes:     newSecuritySchemes(types.RoleViewer), ?

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -34,31 +35,27 @@ func NewHandler(module rawdataexport.Module) rawdataexport.Handler {
...
-// Query Parameters:
+// GET: Converts query params to QueryRangeRequest and delegates to common export logic. No composite_query/trace operator.
 //
-//   - source (optional): Type of data to export ["logs" (default), "metrics", "traces"]
-//     Note: Currently only "logs" is fully supported
+// GET Query Parameters:
 //
+//   - source (optional): ["logs" (default) or "traces"] - metrics not supported
 //   - format (optional): Output format ["csv" (default), "jsonl"]
+//   - start (required): Start time (Unix timestamp in nanoseconds)
```
nityanandagohain: I think these comments will just need another layer of maintainance and we can just keep openapi spec as the source of truth? what do you say

pkg/apiserver/signozapiserver/rawDataExport.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	if err := router.Handle("/api/v1/export_raw_data", handler.New(provider.authZ.ViewAccess(provider.rawDataExportHandler.ExportRawData), handler.OpenAPIDef{
+		ID:          "HandleExportRawDataPOST",
+		Tags:        []string{"logs", "traces"},
+		Summary:     "Export raw data",
+		Description: "This endpoints allows complex query exporting raw data for traces and logs",
+		Request: new(struct {
+			v5.QueryRangeRequest
+			types.ExportRawDataFormatQueryParam
+		}),
```
nityanandagohain: request will be v5.QueryRangeRequest and RequestQuery will be 			types.ExportRawDataFormatQueryParam

pkg/types/querybuildertypes/querybuildertypesv5/req_getters.go
```diff
@@ -0,0 +1,386 @@
+package querybuildertypesv5
+
+import "github.com/SigNoz/signoz/pkg/types/telemetrytypes"
+
+// GetExpression returns the expression string. Panics for Join, PromQuery and ClickHouseQuery.
+func (q *QueryEnvelope) GetExpression() string {
```
nityanandagohain, line 6: how are these getting used ?

nityanandagohain (commented):
Overall changes looks good, have some minor questions. 

Regarding getters and setters that we have added i don't have much opinion apart from what Srikanth has suggested to remove the panics.

Also please rename `rawDataExport`.

pkg/types/exporttypes/rawdataexport.go
```diff
@@ -0,0 +1,35 @@
+package types
+
+// ExportRawDataQueryParams represents the query parameters for the export raw data endpoint
+type ExportRawDataQueryParams struct {
+	ExportRawDataFormatQueryParam
+
+	// Source specifies the type of data to export: "logs" or "traces"
+	Source string `query:"source,default=logs" default:"logs" enum:"logs,traces"`
```
srikanthccv: Let's use Signal when it comes to referring to `logs`/`traces`/`metrics` and it shouldn't have default.

pkg/types/exporttypes/rawdataexport.go
```diff
@@ -0,0 +1,35 @@
...
+	Start uint64 `query:"start"`
+
+	// End is the end time for the query (Unix timestamp in nanoseconds)
+	End uint64 `query:"end"`
+
+	// Limit specifies the maximum number of rows to export
+	Limit int `query:"limit,default=10000" default:"10000" minimum:"1" maximum:"50000"`
+
+	// Filter is a filter expression to apply to the query
+	Filter string `query:"filter"`
```
srikanthccv: Let's use the same type from `qbtypes.Filter`

pkg/types/exporttypes/rawdataexport.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// Columns specifies the columns to include in the export
+	// Format: ["context.field:type", "context.field", "field"]
+	Columns []string `query:"columns"`
```
srikanthccv: Can this now be list of TelemetryFieldKey instead because it has `Normalize`?

pkg/types/exporttypes/rawdataexport.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// OrderBy specifies the sorting order
+	// Format: "column:direction" or "context.field:type:direction"
+	// Direction can be "asc" or "desc"
+	OrderBy string `query:"order_by"`
```
srikanthccv: Same let's use the enum from the query types.

pkg/types/exporttypes/rawdataexport.go
srikanthccv, line 1: We don't add types at the top level unless it's a core type. Since this is specific to exports, let's either define the `exporttypes` or put it under some other existing package types.

pkg/types/querybuildertypes/querybuildertypesv5/req_getters.go
srikanthccv: I think that's the point, there are generic getters and setters to work with query envelope easily. Not all of them need to be used at this point.

I really miss typescript here https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions.

pkg/types/querybuildertypes/querybuildertypesv5/req_getters.go
srikanthccv: just bumping this point

pkg/types/exporttypes/rawdataexport.go
tushar-signoz: Cannot change it now as this API is already rolled out. Will address this separately.

pkg/types/exporttypes/rawdataexport.go
tushar-signoz: Cannot change it now as this API is already rolled out. Will address this separately.

pkg/types/exporttypes/rawdataexport.go
tushar-signoz: Cannot change it now as this API is already rolled out. Will address this separately.

pkg/types/exporttypes/rawdataexport.go
tushar-signoz: Cannot change it now as this API is already rolled out. Will address this separately.

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -31,206 +32,184 @@ func NewHandler(module rawdataexport.Module) rawdataexport.Handler {
...
 	if err != nil {
 		render.Error(rw, err)
 		return
 	}
 
-	format, err := getExportQueryFormat(queryParams)
+	orgID, err := valuer.NewUUID(claims.OrgID)
 	if err != nil {
-		render.Error(rw, err)
+		render.Error(rw, errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "orgID is invalid"))
```
srikanthccv: I know you haven't this in the current PR, but shouldn't this be 401 instead? Also there is usage of `valuer.MustNewUUID`. @vikrantgupta25  why do we have potential panic in the request flow with `MustNewUUID` instead of rejecting the request? Or is that handled somewhere else?

pkg/types/exporttypes/rawdataexport.go
srikanthccv: Ok, let's do it this way.

- Keep the field around but deprecate it https://github.com/swaggest/jsonschema-go#field-tags. 
- Add the Signal field
- Update our frontend to use `Signal` if it's already using the `source`
- After param binding, map the `Source` to `Signal` if it's non-empty
- Track if anyone is actually using the `source`, remove it if none uses it

pkg/types/exporttypes/rawdataexport.go
srikanthccv: Same, let's plan in this PR only.

pkg/types/exporttypes/rawdataexport.go
srikanthccv: Same, let's plan in this PR only.

pkg/types/exporttypes/rawdataexport.go
srikanthccv: Same, let's plan in this PR only.

pkg/types/querybuildertypes/querybuildertypesv5/req_getters.go
```diff
@@ -0,0 +1,386 @@
...
+		return spec.Query
+	case ClickHouseQuery:
+		return spec.Query
+	}
+	return ""
+}
+
+// GetStep returns the PromQL step size.
+func (q *QueryEnvelope) GetStep() Step {
+	switch spec := q.Spec.(type) {
```
srikanthccv: There is `GetStepInterval` below, we should remove this `GetStep` and use the `GetStepInterval` for promql?

pkg/types/querybuildertypes/querybuildertypesv5/req_setters.go
```diff
@@ -0,0 +1,458 @@
...
+		spec.Query = query
+		q.Spec = spec
+	case ClickHouseQuery:
+		spec.Query = query
+		q.Spec = spec
+	}
+}
+
+// SetStep sets the PromQL step size, if applicable.
+func (q *QueryEnvelope) SetStep(step Step) {
```
srikanthccv: Same as above.

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -466,51 +377,24 @@ func getsizeOfStringSlice(slice []string) uint64 {
...
-		return nil, errors.NewInvalidInputf(errors.CodeInvalidInput, "invalid order_by format: %s, should be <column>:<direction>", orderByParam)
-	}
-
 	column := strings.Join(parts[:len(parts)-1], ":")
 	direction := parts[len(parts)-1]
 
-	orderDirection, ok := qbtypes.OrderDirectionMap[direction]
-	if !ok {
-		return nil, errors.NewInvalidInputf(errors.CodeInvalidInput, "invalid order_by direction: %s, should be one of %s, %s", direction, qbtypes.OrderDirectionAsc, qbtypes.OrderDirectionDesc)
-	}
```
srikanthccv, line 491: This removed code block is relevant?

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -427,9 +341,12 @@ func constructCSVRecordFromQueryResponse(data map[string]any, headerToIndexMappi
 				valueStr = v.String()
 
 			default:
-				// For all other complex types (maps, structs, etc.)
-				jsonBytes, _ := json.Marshal(v)
-				valueStr = string(jsonBytes)
+				jsonBytes, err := json.Marshal(v)
+				if err != nil {
+					valueStr = fmt.Sprintf("%v", v)
```
srikanthccv, line 316: Just curious, when does this happen?

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -466,51 +377,24 @@ func getsizeOfStringSlice(slice []string) uint64 {
...
-func getExportQueryOrderBy(queryParams url.Values) ([]qbtypes.OrderBy, error) {
-	orderByParam := queryParams.Get("order_by")
-
+// parseExportQueryOrderBy converts a bound order_by string to an OrderBy slice.
+// The string should be in the format "column:direction" and is assumed already validated.
+func parseExportQueryOrderBy(orderByParam string) []qbtypes.OrderBy {
 	orderByParam = strings.TrimSpace(orderByParam)
 	if orderByParam == "" {
-		return telemetrylogs.DefaultLogsV2SortingOrder, nil
+		return []qbtypes.OrderBy{}
```
srikanthccv: Is the default order by set by downstream?

pkg/modules/rawdataexport/implrawdataexport/handler.go
tushar-signoz: Very few cases, for eg: complex integers or map of non string keys like {1.1: "Float"}

pkg/modules/rawdataexport/implrawdataexport/handler.go
tushar-signoz: Validation part will be handled by the downstream

pkg/types/exporttypes/rawdataexport.go
```diff
@@ -0,0 +1,120 @@
...
+// ExportRawDataQueryParams represents the query parameters for the export raw data endpoint
+type ExportRawDataQueryParams struct {
+	ExportRawDataFormatQueryParam
+
+	// Signal specifies the type of data to export: "logs" or "traces"
+	Signal telemetrytypes.Signal `query:"signal" enum:"logs,traces" required:"true"`
+
+	// Source specifies the type of data to export: "logs" or "traces"
+	// Deprecated: Use Signal instead.
+	Source string `query:"source" deprecated:"true"`
```
srikanthccv: Add description that says "use signal instead" or something like that because the API client needs to what is the replacement for the substitute field.

Ans Same for other deprecated fields well.

pkg/types/querybuildertypes/querybuildertypesv5/req.go
```diff
@@ -393,6 +393,59 @@ func (r *QueryRangeRequest) HasOrderSpecified() bool {
 	return false
 }
 
+// UseDefaultOrderBy applies a default order unless query has an explicit order provided.
+func (r *QueryRangeRequest) UseDefaultOrderBy() {
```
srikanthccv, line 398: 1. Should it be a method on the `QueryEnvelope` instead?

2. Let's name it `UseDefaultOrderByForListQuery` or something similar to indicate that intent clearly because this is not something to use everywhere?

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
```diff
@@ -265,24 +282,25 @@ func (q *QueryBuilderQuery[T]) validateAggregations(requestType RequestType) err
...
-	// Validate limit
-	if q.Limit < 0 {
-		return errors.NewInvalidInputf(
-			errors.CodeInvalidInput,
-			"limit must be non-negative, got %d",
-			q.Limit,
-		)
-	}
+func (q *QueryBuilderQuery[T]) validateLimitAndPagination(cfg validationConfig) error {
+	if !cfg.skipLimitValidation {
```
srikanthccv: nit:

every other function has `if skip { return nil }` but this one has `if !skip { validate }`. this could also be `if skip { return nil }`

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
```diff
@@ -618,3 +633,11 @@ func validateQueryEnvelope(envelope QueryEnvelope, requestType RequestType) erro
 		)
 	}
 }
+
+func getValidationOptions(isAggregationQuery bool) []ValidationOption {
```
srikanthccv: Could this be `getValidationOptions(requestType RequestType)`? I think that's more cleaner because it reads naturally like "get validatation options for this request type` instead of `getValidationOptions(bool)` where one has to click and option to know what's the boolean is for?

pkg/types/exporttypes/rawdataexport.go
```diff
@@ -0,0 +1,120 @@
...
+		if columnStr == "" {
+			continue
+		}
+		columns = append(columns, telemetrytypes.GetFieldKeyFromKeyText(columnStr))
+	}
+	return columns
+}
+
+// parseExportQueryOrderBy converts a bound order_by string to an OrderBy slice.
+// The string should be in the format "column:direction" and is assumed already validated.
```
srikanthccv: this function states it's assumes already validated, but in the get request flow, binding -> normalize (the normalize step calls `parseExportQueryOrderBy` for `if len(p.Order) == 0 && len(p.OrderBy) > 0 {` ) -> validate, so this function could panic if OrderBy length is zero.

pkg/modules/rawdataexport/implrawdataexport/handler.go
srikanthccv: Let's keep the `MustNewUUID`, it is due for a change as mentioned by vikrant.

pkg/apiserver/signozapiserver/provider.go
```diff
@@ -221,6 +227,10 @@ func (provider *provider) AddToRouter(router *mux.Router) error {
 		return err
 	}
 
+	if err := provider.addRawDataExportRoutes(router); err != nil {
+    return err
+  }
```
srikanthccv: Please fix the formatting here.

pkg/apiserver/signozapiserver/rawdataexport.go
srikanthccv, line 1: lower case, single-word names

pkg/types/exporttypes/rawdataexport.go
```diff
@@ -0,0 +1,126 @@
...
+// parseExportQueryOrderBy converts a bound order_by string to an OrderBy slice.
+// The string should be in the format "column:direction" and validation is handled by the downstream.
+func parseExportQueryOrderBy(orderByParam string) []qbtypes.OrderBy {
+	orderByParam = strings.TrimSpace(orderByParam)
+	if orderByParam == "" {
+		return []qbtypes.OrderBy{}
+	}
+
+	parts := strings.Split(orderByParam, ":")
+	// Here we silently ignore the error as this is deprecated code path
```
srikanthccv: Should we silently ignore this?

pkg/types/exporttypes/rawdataexport.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	column := strings.Join(parts[:len(parts)-1], ":")
+	direction := parts[len(parts)-1]
+
+	return []qbtypes.OrderBy{
+		{
+			Key: qbtypes.OrderByKey{
+				TelemetryFieldKey: telemetrytypes.GetFieldKeyFromKeyText(column),
+			},
+			Direction: qbtypes.OrderDirectionMap[direction],
```
srikanthccv: Is this validated to ensure the `Direction` value won't be empty?

pkg/types/exporttypes/rawdataexport.go
```diff
@@ -0,0 +1,126 @@
...
+	FilterString string         `query:"filter"  deprecated:"true" description:"Deprecated: use filterExpression instead."`
+	Filter       qbtypes.Filter `query:"filterExpression" description:"The filter expression to apply to the query."`
+
+	// Columns specifies the columns to include in the export
+	// Format: ["context.field:type", "context.field", "field"]
+	// Deprecated: Use SelectFields instead.
+	Columns []string `query:"columns" deprecated:"true" description:"Deprecated: use selectFields instead."`
+
+	// SelectFields specifies the columns to include in the export
+	SelectFields []telemetrytypes.TelemetryFieldKey `query:"selectFields" description:"The columns to include in the export."`
```
srikanthccv: I couldn't find in the tests how the `filterExpression`, `selectFields` and `order` go in the get payload. It would be great if we can add tests for that.

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -31,206 +31,188 @@ func NewHandler(module rawdataexport.Module) rawdataexport.Handler {
...
+		}
+		params.Normalize()
+		if err := params.Validate(); err != nil {
+			render.Error(rw, err)
+			return
+		}
+		format = params.Format
+		queryRangeRequest = buildQueryRangeRequest(&params)
+	case http.MethodPost:
+		var formatParam exporttypes.ExportRawDataFormatQueryParam
```
srikanthccv: API design question: should there be format in params and rest in body for POST request? Should they all be part of the body instead?

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -31,206 +31,188 @@ func NewHandler(module rawdataexport.Module) rawdataexport.Handler {
...
+		case qbtypes.QueryBuilderQuery[qbtypes.LogAggregation],
+			qbtypes.QueryBuilderQuery[qbtypes.TraceAggregation],
+			qbtypes.QueryBuilderTraceOperator:
+			// Supported spec types
+		default:
+			return errors.NewInvalidInputf(errors.CodeInvalidInput, "unsupported query at index %d type: %T", idx, spec)
+		}
+	}
+
+	err := req.Validate(qbtypes.WithSkipLimitOffsetValidation())
```
srikanthccv: Should there be additional skip validation such as aggregations, group by here?

pkg/modules/rawdataexport/implrawdataexport/handler.go
tushar-signoz: Although nothings stops us to keep everything in body, Idea is to separate, `queryRangeRequest` and `export configs`. For example if we were to support exporting to S3 storages, that would be part of query params and QueryRangeRequest (post body) would remain the same.

srikanthccv: Asked Claude if this follows the REST-ful principles and good pattern, here is the response

---
This is actually a clean example of the pattern done right. The POST on `/api/v1/export_raw_data` uses query params appropriately:

**What's in the query param:** `format` (csv/jsonl) — this controls *how the response is delivered*, not what data is being queried. It's a meta-parameter about output representation.

**What's in the body:** `QueryRangeRequest` — the actual complex query with filters, time ranges, field selections, ordering, etc. This is the substantive payload that defines *what* you're asking for.

This separation makes sense. The `format` param is conceptually closer to an `Accept` header than to query data. You could even argue it belongs as a header or content negotiation, but a query param is pragmatic and widely understood — easier to use in browser testing, curl commands, and client code than fiddling with headers.

If I were nitpicking, the only thing I'd consider is whether `format` could move to an `Accept` header (`text/csv` vs `application/jsonl`), which would be more "REST-pure." But in practice, the query param approach is more developer-friendly and perfectly fine here. No issues with this design.

pkg/apiserver/signozapiserver/rawdataexport.go
srikanthccv: bump, this is not updated yet.

pkg/modules/rawdataexport/implrawdataexport/handler.go
```diff
@@ -31,206 +30,130 @@ func NewHandler(module rawdataexport.Module) rawdataexport.Handler {
...
-	rw.Header().Set("Transfer-Encoding", "chunked")
-
-	queryParams := r.URL.Query()
-
-	startTime, endTime, err := getExportQueryTimeRange(queryParams)
-	if err != nil {
-		render.Error(rw, err)
+	format := formatParam.Format
+	if err := json.NewDecoder(r.Body).Decode(&queryRangeRequest); err != nil {
+		render.Error(rw, errors.NewInvalidInputf(errors.CodeInvalidInput, "invalid request body: %v", err))
```
srikanthccv:
```suggestion
	if err := binding.JSON.BindBody(r.Body, &queryRangeRequest); err != nil {
		render.Error(rw, err)
```

srikanthccv: Can you please also update the `handler.md` to recommend using `binding.JSON.BindBody` for post request inside handler implementation?

tests/integration/src/rawexportdata/01_logs.py
```diff
@@ -0,0 +1,660 @@
...
+        ]
+    )
+
+    token = get_token(USER_ADMIN_EMAIL, USER_ADMIN_PASSWORD)
+
+    # Calculate timestamps in nanoseconds
+    start_ns = int((now - timedelta(minutes=5)).timestamp() * 1e9)
+    end_ns = int(now.timestamp() * 1e9)
+
+    body = QueryRangeRequest(
```
srikanthccv, line 110: Did you add this because `make_query_request` uses the query_range but you wanted the payload for `export_raw_data`?

tests/integration/fixtures/model.py
srikanthccv, line 1: Would it be better to keep these in `querier.py`?

tests/integration/src/rawexportdata/01_logs.py
tushar-signoz: Yes

tushar-signoz: Could have used/added `build_query_for_list()`, same payload would work for export_raw_data, but dataclass is more readable than the function

tests/integration/fixtures/model.py
tushar-signoz: `querier.py` already has similar named functions that's why chose to go with a different file

srikanthccv: `model.py` is too broad IMO.

PR #10020, opened 2026-01-16
chore: handle threshold unit scaling issues

amlannandy:
## 📄 Summary

- Update the `unitMapping` const to handle all new values of y axis unit values that we have introduced recently.
- We were not handling a few values of y axis units due to which scaling of values in the time series graph was not happening correctly when the Y Axis Unit Value and Threshold Value had different units
- How the flow works:
  - The graph calls `convertValue`. Inside that we check if it's a Universal Unit. If yes, we convert it to its grafana equivalent as the graph component uses grafana units for formatting and showcasing data.
  - After that the rest of the flow remains same.
  - For universal units that do not have a grafana equivalent, we have updated the `unitMapping` const to include them

## 🧪 How to Test

Please refer to the recording as it's clearer visually

---

## 🔍 Related Issues

Closes https://github.com/SigNoz/engineering-pod/issues/3684

---

Alerts V2

https://github.com/user-attachments/assets/a889b587-b150-492c-b735-396a948862a5

Old Alerts 

https://github.com/user-attachments/assets/600eede8-fe36-4ea8-b8cf-672143ac4ad8

---

## 📋 Checklist

- [x] Dev Review

### DIFF
--- a/cmd/community/server.go
+++ b/cmd/community/server.go
@@ -12,6 +12,8 @@ import (
 	"github.com/SigNoz/signoz/pkg/authn"
 	"github.com/SigNoz/signoz/pkg/authz"
 	"github.com/SigNoz/signoz/pkg/factory"
+	"github.com/SigNoz/signoz/pkg/gateway"
+	"github.com/SigNoz/signoz/pkg/gateway/noopgateway"
 	"github.com/SigNoz/signoz/pkg/licensing"
 	"github.com/SigNoz/signoz/pkg/licensing/nooplicensing"
 	"github.com/SigNoz/signoz/pkg/modules/dashboard"
@@ -89,6 +91,9 @@ func runServer(ctx context.Context, config signoz.Config, logger *slog.Logger) e
 		},
 		func(store sqlstore.SQLStore, settings factory.ProviderSettings, analytics analytics.Analytics, orgGetter organization.Getter, _ role.Module, queryParser queryparser.QueryParser, _ querier.Querier, _ licensing.Licensing) dashboard.Module {
 			return impldashboard.NewModule(impldashboard.NewStore(store), settings, analytics, orgGetter, queryParser)
+		},
+		func(_ licensing.Licensing) factory.ProviderFactory[gateway.Gateway, gateway.Config] {
+			return noopgateway.NewProviderFactory()
 		},
 	)
 	if err != nil {
--- a/cmd/enterprise/server.go
+++ b/cmd/enterprise/server.go
@@ -10,6 +10,7 @@ import (
 	"github.com/SigNoz/signoz/ee/authn/callbackauthn/samlcallbackauthn"
 	"github.com/SigNoz/signoz/ee/authz/openfgaauthz"
 	"github.com/SigNoz/signoz/ee/authz/openfgaschema"
+	"github.com/SigNoz/signoz/ee/gateway/httpgateway"
 	enterpriselicensing "github.com/SigNoz/signoz/ee/licensing"
 	"github.com/SigNoz/signoz/ee/licensing/httplicensing"
 	"github.com/SigNoz/signoz/ee/modules/dashboard/impldashboard"
@@ -22,6 +23,7 @@ import (
 	"github.com/SigNoz/signoz/pkg/authn"
 	"github.com/SigNoz/signoz/pkg/authz"
 	"github.com/SigNoz/signoz/pkg/factory"
+	"github.com/SigNoz/signoz/pkg/gateway"
 	"github.com/SigNoz/signoz/pkg/licensing"
 	"github.com/SigNoz/signoz/pkg/modules/dashboard"
 	pkgimpldashboard "github.com/SigNoz/signoz/pkg/modules/dashboard/impldashboard"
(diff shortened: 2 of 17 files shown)


frontend/src/container/FormAlertRules/RuleOptions.tsx
```diff
@@ -365,10 +364,10 @@ function RuleOptions({
 		</InlineSelect>
 	);
 
-	const selectedCategory = getCategoryByOptionId(currentQuery?.unit || '');
+	const selectedCategory = getCategoryByOptionId(yAxisUnit);
 
 	const categorySelectOptions = getCategorySelectOptionByName(
-		selectedCategory?.name,
+		selectedCategory || '',
```
AshwinBhatkal: can we ensure this is a string when it sets from the above function itself?

frontend/src/lib/getConvertedValue.ts
```diff
@@ -22,6 +29,51 @@ const unitsMapping = [
...
 			},
+			{
+				label: Y_AXIS_UNIT_NAMES[UniversalYAxisUnit.BITS],
+				value: UniversalYAxisUnit.BITS,
+				factor: 8,
+			},
+			{
+				label: Y_AXIS_UNIT_NAMES[UniversalYAxisUnit.KILOBITS],
+				value: UniversalYAxisUnit.KILOBITS,
+				factor: 8 * 1000,
```
AshwinBhatkal: Why not 1000 ^ 1, 1000 ^ 2 and so on?

.gitignore
```diff
@@ -104,7 +106,6 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
 lib64/
 parts/
```
amlannandy, line 112: @therealpandey I'm not sure why exactly we're excluding files under `lib` but it is affecting the frontend codebase as well. 
If we create any new files under `frontend/src/lib`, they are also not getting tracked.

I've removed this line of code for now. If it is still required, please let me know I'll update the gitignore to ignore this restriction for frontend only

frontend/src/lib/getConvertedValue.ts
amlannandy: Changed ✅

frontend/src/container/CreateAlertV2/AlertCondition/ThresholdItem.tsx
```diff
@@ -37,10 +37,7 @@ function ThresholdItem({
 		);
 		if (units.length === 0) {
 			component = (
-				<Tooltip
-					trigger="hover"
-					title="Please select a Y-axis unit for the query first"
-				>
+				<Tooltip trigger="hover" title="No compatible units available">
```
amlannandy, line 40: As this tooltip can show up in 2 cases
1. There is no base unit selected. Hence no category is valid to show units from.
2. There is a custom unit selected (e.g. {reason}) which is not going to be part of any category

frontend/src/lib/getConvertedValue.ts
```diff
@@ -15,12 +26,52 @@ const unitsMapping = [
 			{
 				label: 'bits(IEC)',
 				value: 'bits',
-				factor: 8, // 1 byte = 8 bits
+				factor: BIT_FACTOR,
```
amlannandy, line 32: This was incorrect earlier. As our base unit here is `bytes` with `factor = 1`.
So, bits should have a factor of `1/8` not `8`

frontend/src/container/FormAlertRules/RuleOptions.tsx
amlannandy: I've updated `getCategorySelectOptionByName` to handle false values. Does that work?

frontend/src/lib/getConvertedValue.ts
amlannandy: Have added a test for this as well `should convert bits to gigabytes`

.gitignore
amlannandy: Confirmed with @therealpandey that it's fine to remove it ✅

frontend/src/container/CreateAlertV2/AlertCondition/utils.tsx
```diff
@@ -47,9 +47,17 @@ export function getCategoryByOptionId(id: string): string | undefined {
 }
 
 export function getCategorySelectOptionByName(
-	name: string,
+	name: string | undefined,
```
AshwinBhatkal, line 50: Just checking - Is this typed properly? Where is the undefined value coming from or being saved?

frontend/src/lib/getConvertedValue.ts
```diff
@@ -1,3 +1,14 @@
+import {
+	UniversalUnitToGrafanaUnit,
+	Y_AXIS_UNIT_NAMES,
+} from 'components/YAxisUnitSelector/constants';
+import { UniversalYAxisUnit } from 'components/YAxisUnitSelector/types';
+import { isUniversalUnit } from 'components/YAxisUnitSelector/utils';
+
+// 1 byte = 8 bits
+// Or 1 bit = 1/8 bytes
+const BIT_FACTOR = 1 / 8;
```
AshwinBhatkal, line 10: Should you use the same constant for 1024 and 1000 as well?

frontend/src/container/CreateAlertV2/AlertCondition/ThresholdItem.tsx
amlannandy: Discussed on slack

frontend/src/container/CreateAlertV2/AlertCondition/utils.tsx
amlannandy: Yeah so `name` refers to the name of the category.
A custom unit like `{reason}` won't belong to any categroy. So `name` can be undefined in that case

PR #10026, opened 2026-01-17
chore: move types, constants to separate files, delete unused code

YounixM:
## Pull Request

---

To clean up unused code in Date Time Picker component

Helps with code maintenance

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Existing tests should pass
- Manual verification: verified all flows in the product.
- Edge cases covered: NA

---

NA

- Blast radius: NA
- Potential regressions:
- Rollback plan:

---

N/A

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

Let's do a quick sanity test of all flows where date time picker is present.

---

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,6 +6,15 @@
 > Why does this change exist?  
 > What problem does it solve, and why is this the right approach?
 
+
+
+#### Screenshots / Screen Recordings (if applicable)
+> Include screenshots or screen recordings that clearly show the behavior before the change and the result after the change. This helps reviewers quickly understand the impact and verify the update.
+
+
+#### Issues closed by this PR
+> Reference issues using `Closes #issue-number` to enable automatic closure on merge.
+
 ---
 
 ### ✅ Change Type


frontend/src/container/TopNav/DateTimeSelectionV2/types.ts
```diff
@@ -0,0 +1,80 @@
+type FiveMin = '5m';
```
AshwinBhatkal, line 1: nit: Why not Minutes instead of Min? Also, why not the s at the end - SixHours,etc?

AshwinBhatkal (approved):

frontend/src/container/TopNav/DateTimeSelectionV2/types.ts
YounixM: The scope was to just clean up the dead code so that it's easier to review. I will raise a separate PR for code rewrite as we clean this up more

PR #10005, opened 2026-01-14
feat: heatmap visualization panel for metrics/logs/traces

niladrix719:
## 📄 Summary

- Implemented a new heatmap visualization panel for displaying histogram bucket data
- Created a plugin that renders time-series data as heatmaps with proper bucket boundaries
- Added some new panel properties for heatmap like color scheme
- Added toolpit to view cell level information

Screenshot 2026-01-30 at 8 29 36 PM

fixes #6937 #8169
parts of #8171 
needs backend support #10029 #10030 

---

> [!NOTE]
> **Medium Risk**
> Medium risk: introduces a new panel/query type and modifies query payload shaping, response conversion, and drilldown filter construction, which could affect existing dashboards if assumptions about result formats change.
> 
> **Overview**
> Adds a new `PANEL_TYPES.HEATMAP` visualization, wiring it into dashboard panel selection, initial queries, explorer view mapping, and full-view visibility rules.
> 
> Extends the V5 query-range pipeline to support a new `requestType: 'heatmap'`, including payload mapping, response conversion, and type changes to carry heatmap bucket `bounds`/`values` without forcing legacy time-series coercion.
> 
> Implements a dedicated `HeatmapPanelWrapper` + `getUplotHeatmapChartOptions` to render bucketed time-series as a heatmap, with hover/tooltip, click + context-menu drilldowns (including min/max range filters), and a new per-widget `heatmapColorPalette` setting exposed in the widget right-side config.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2ceedc7e796118625423e0c80c0fca05f43f311a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #10030, opened 2026-01-18
feat: heatmap aggregation operator support for logs and traces

niladrix719:
## Pull Request

---

### 📄 Summary

- Added a heatmap(attr,N) aggregator function for logs and traces
- Added heatmap query builder function for both logs and traces

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature

---

- Tests added/updated: Tests are added for new implementation
- Manual verification: Manually tested and verified
- Edge cases covered: edge cases are considered

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

Parent PR #10005 

---


srikanthccv:
@niladrix719
- please add integration tests
- can heatmap and distribution data be cached?

pkg/query-service/model/v3/v3.go
```diff
@@ -67,6 +67,8 @@ const (
 	AggregateOperatorHistQuant90   AggregateOperator = "hist_quantile_90"
 	AggregateOperatorHistQuant95   AggregateOperator = "hist_quantile_95"
 	AggregateOperatorHistQuant99   AggregateOperator = "hist_quantile_99"
+	AggregateOperatorHeatmap       AggregateOperator = "heatmap"
+	AggregateOperatorDistribution  AggregateOperator = "distribution"
```
srikanthccv: Why are they added here in v3.go?

pkg/querybuilder/agg_funcs.go
```diff
@@ -229,6 +229,16 @@ var (
...
 	}
+	AggrFuncHeatmap = AggrFunc{
+		Name:        valuer.NewString("heatmap"),
+		FuncName:    "histogram",
+		RequireArgs: true, Numeric: true, MinArgs: 1, MaxArgs: 2,
+	}
+	AggrFuncDistribution = AggrFunc{
+		Name:        valuer.NewString("distribution"),
+		FuncName:    "histogram",
+		RequireArgs: true, Numeric: true, MinArgs: 1, MaxArgs: 2,
```
srikanthccv: Why does it have `MaxArgs: 2`?

pkg/querybuilder/agg_rewrite.go
```diff
@@ -15,6 +16,12 @@ import (
 	"github.com/huandu/go-sqlbuilder"
 )
 
+const (
+	defaultBuckets = "20"
+	// Histogram max buckets limit is 250
+	maxBuckets = 250
```
srikanthccv: Where is this max coming from?

pkg/querybuilder/agg_rewrite.go
```diff
@@ -15,6 +16,12 @@ import (
 	"github.com/huandu/go-sqlbuilder"
 )
 
+const (
+	defaultBuckets = "20"
```
srikanthccv: How did you arrive at this default value?

pkg/querybuilder/agg_rewrite.go
```diff
@@ -227,6 +234,23 @@ func (v *exprVisitor) VisitFunctionExpr(fn *chparser.FunctionExpr) error {
 		}
 	} else {
 		// Non-If functions: map every argument as a column/value
+		if name == AggrFuncHeatmap.Name.StringValue() || name == qbtypes.RequestTypeDistribution.StringValue() {
```
srikanthccv: Why is one condition is `name == AggrFuncHeatmap.Name` and other condition is `name == qbtypes.RequestTypeDistribution`?

pkg/querybuilder/agg_rewrite.go
```diff
@@ -227,6 +234,23 @@ func (v *exprVisitor) VisitFunctionExpr(fn *chparser.FunctionExpr) error {
...
+			buckets := defaultBuckets
+			if len(args) == 2 {
+				buckets = args[1].String()
+				val, err := strconv.Atoi(buckets)
+				if err != nil {
+					return errors.NewInvalidInputf(errors.CodeInvalidInput, "invalid bucket count '%s': must be a number", buckets)
+				}
+				if val <= 0 || val > maxBuckets {
+					return errors.NewInvalidInputf(errors.CodeInvalidInput, "buckets count %d exceeds range [1, %d]", val, maxBuckets)
```
srikanthccv:
```suggestion
					return errors.NewInvalidInputf(errors.CodeInvalidInput, "buckets count %d outside range [1, %d]", val, maxBuckets)
```

pkg/querybuilder/agg_rewrite.go
```diff
@@ -268,3 +292,47 @@ func parseFragment(sql string) (chparser.Expr, error) {
...
+	wrapped := fmt.Sprintf("SELECT %s", expr)
+	p := chparser.NewParser(wrapped)
+	stmts, err := p.ParseStmts()
+	if err != nil || len(stmts) == 0 {
+		return 0, err
+	}
+
+	sel, ok := stmts[0].(*chparser.SelectQuery)
+	if !ok || len(sel.SelectItems) == 0 {
+		return 0, nil
```
srikanthccv: Why does it return 0 _and_ nil? How should the caller interpret this return?

pkg/querybuilder/agg_rewrite.go
```diff
@@ -268,3 +292,47 @@ func parseFragment(sql string) (chparser.Expr, error) {
...
+	}
+
+	functionExpr, ok := sel.SelectItems[0].Expr.(*chparser.FunctionExpr)
+	if !ok {
+		return 0, nil
+	}
+
+	funcName := strings.ToLower(functionExpr.Name.Name)
+	if funcName != "heatmap" && funcName != "distribution" {
+		return 0, nil
```
srikanthccv: Same

pkg/telemetrylogs/statement_builder.go
```diff
@@ -83,6 +83,8 @@ func (b *logQueryStatementBuilder) Build(
 		return b.buildTimeSeriesQuery(ctx, q, query, start, end, keys, variables)
 	case qbtypes.RequestTypeScalar:
 		return b.buildScalarQuery(ctx, q, query, start, end, keys, false, variables)
+	case qbtypes.RequestTypeBucket, qbtypes.RequestTypeDistribution:
+		return b.buildBucketQuery(ctx, q, query, start, end, keys, variables, requestType)
```
srikanthccv: Why are they combined together?

pkg/telemetrylogs/statement_builder.go
```diff
@@ -688,3 +690,88 @@ func (b *logQueryStatementBuilder) buildResourceFilterCTE(
...
+	query qbtypes.QueryBuilderQuery[qbtypes.LogAggregation],
+	start, end uint64,
+	keys map[string][]*telemetrytypes.TelemetryFieldKey,
+	variables map[string]qbtypes.VariableItem,
+	requestType  qbtypes.RequestType,
+) (*qbtypes.Statement, error) {
+
+	if len(query.Aggregations) == 0 {
+		return nil, errors.NewInvalidInputf(errors.CodeInvalidInput, "at least one aggregation is required for query")
+	}
```
srikanthccv: Should not be here. It should have been rejected much earlier if the aggregations didn't exist. The `buildBucketQuery` expects that the query.Aggregations is present and valid.

pkg/telemetrylogs/statement_builder.go
```diff
@@ -688,3 +690,88 @@ func (b *logQueryStatementBuilder) buildResourceFilterCTE(
...
+
+	if requestType == qbtypes.RequestTypeDistribution {
+		sb.Select(fmt.Sprintf("%d AS ts", start))
+	} else {
+		sb.Select(fmt.Sprintf(
+			"toStartOfInterval(fromUnixTimestamp64Nano(timestamp), INTERVAL %d SECOND) AS ts",
+			int64(query.StepInterval.Seconds()),
+		))
+	}
```
srikanthccv: Split it up separately. The heatmap and distribution are not the same thing.

pkg/telemetrylogs/statement_builder.go
```diff
@@ -688,3 +690,88 @@ func (b *logQueryStatementBuilder) buildResourceFilterCTE(
...
+
+	aggExpr := query.Aggregations[0]
```
srikanthccv: Do we only expect one aggregation here?

pkg/telemetrytraces/statement_builder.go
srikanthccv, line 1: Same comments as logs.

pkg/telemetrytraces/trace_operator_cte_builder.go
srikanthccv, line 1: Same comments

pkg/types/querybuildertypes/querybuildertypesv5/request_type.go
```diff
@@ -20,4 +20,6 @@ var (
 	RequestTypeTrace = RequestType{valuer.NewString("trace")}
 	// []Bucket (struct{Lower,Upper,Count float64}), example: histogram
 	RequestTypeDistribution = RequestType{valuer.NewString("distribution")}
+	// BucketData (struct{Timestamps, BucketBounds, Counts}), example: bucket panel
+	RequestTypeBucket = RequestType{valuer.NewString("bucket")}
```
srikanthccv:
```suggestion
	RequestTypeHeatmap = RequestType{valuer.NewString("heatmap")}
```

pkg/types/querybuildertypes/querybuildertypesv5/resp.go
```diff
@@ -193,6 +193,28 @@ type RawStream struct {
 	Error chan error
 }
 
+type BucketData struct {
```
srikanthccv: Look at the `TimeSeries` type, do we need a custom type called `BucketData` or can `TimeSeries` be used?

pkg/types/querybuildertypes/querybuildertypesv5/resp.go
```diff
@@ -193,6 +193,28 @@ type RawStream struct {
...
+	BucketStarts []float64   `json:"bucketStarts,omitempty"`
+	BucketBounds []float64   `json:"bucketBounds"`
+	BucketCount  int         `json:"bucketCount,omitempty"`
+	Timestamps   []int64     `json:"timestamps"`
+	Counts       [][]float64 `json:"counts"`
+	Unit         string      `json:"unit,omitempty"`
+}
+
+type DistributionBucket struct {
+	Timestamp   int64   `json:"ts"`
```
srikanthccv: Why is there a timestamp here?

pkg/types/querybuildertypes/querybuildertypesv5/resp.go
```diff
@@ -193,6 +193,28 @@ type RawStream struct {
...
+	BucketStart float64 `json:"bucket_start"`
+	BucketEnd   float64 `json:"bucket_end"`
```
srikanthccv: What does the bucket start and end indicate?

pkg/querier/consume.go
srikanthccv, line 1: The code added looks unnecessarily complicated.

- Why is there a need to pass the bucketcount around?
- Is the attempt to merge the heatmap and the distribution reason for complex code?

pkg/querybuilder/agg_rewrite.go
niladrix719: clickhouse's histogram() function supports a maximum of 250 buckets, I think I will add a better comment there

pkg/telemetrylogs/statement_builder.go
niladrix719: Instead of duplicating the whole query logic, I kept a single bucket-query builder and used the requestType to slightly change the time selection and grouping.
Heatmap groups data by time intervals, while distribution collapses everything into a single time bucket. The rest of the aggregation logic stays identical, that was my intuition

pkg/telemetrylogs/statement_builder.go
niladrix719: okay I will create separate functions for both

pkg/types/querybuildertypes/querybuildertypesv5/resp.go
niladrix719: lower and upper bound of bucket range, I think we can rename them

pkg/querier/consume.go
niladrix719: > Why is there a need to pass the bucketcount around?

okay I will try to simplify it

> Is the attempt to merge the heatmap and the distribution reason for complex code?

I don’t think so, in fact both were merged quite seamlessly, but I’ll see if we can reduce the complexity of the code

pkg/querybuilder/agg_rewrite.go
niladrix719: I just picked it as a safe default, as histogram expects a compulsory bucket count

pkg/querybuilder/agg_funcs.go
niladrix719: the bucket count is optional, users can either rely on the default or provide it explicitly
- The first argument is the field/column
- The second argument is the bucket count

niladrix719:
Some major changes -

- I’ve separated the distribution support from here and will move it to another PR.
- Heatmap now uses the TimeSeriesValue type instead of creating a new HeatmapData type
- Instead of using the ClickHouse histogram function, I’ve implemented custom logic due to time-interval issues with bucket bins

srikanthccv:
> Instead of using the ClickHouse histogram function, I’ve implemented custom logic due to time-interval issues with bucket bins

Can you help me understand more about this? What is the time-interval issues you refer to and why does it force to implement custom logic instead of histogram function?

niladrix719:
> > Instead of using the ClickHouse histogram function, I’ve implemented custom logic due to time-interval issues with bucket bins
> 
> Can you help me understand more about this? What is the time-interval issues you refer to and why does it force to implement custom logic instead of histogram function?

When I use ClickHouse's histogram(40) with GROUP BY ts, it calculates 40 bins from the entire time range, then applies those same 40 bins to each time interval. The issue is that the output format makes it appear like I'm getting (number_of_intervals × 40) bins in the result, so the issue is that I am getting more bins than I expected

niladrix719:
Hey, on second thought, I was wrong about the histogram function, I was merging them in the wrong way, creating the confusion, I will update the code to use histogram again

srikanthccv:
Hi @niladrix719,

Can you help me with couple of questions similar to https://github.com/SigNoz/signoz/pull/10077#issuecomment-3787107612.

Please disclose the
- what AI tool is used (e.g. Claude Code, Cursor)
- to what extent that the work was AI-assisted

niladrix719:
GitHub Copilot for major understanding the codebase, debugging and writing proper code styles and tests, mainly

srikanthccv:
- How did you verify the correctness of the result produced?
- Why are bins created with equal width? 

Simple example, for `t1` there is one tuple `[1, 10, 2.5]` and `t2` there is `[1, 7, 3.5]` and you have bins created as `[1, 5]`, `[5, 10]`. then how does result from ClickHouse get mapped to local bins?

niladrix719:
I see the issue now. I opted for equal widths because I was rebinning. It would be better to keep them as it is without rebinning. Proportional rebinning is possible, but I think it's better to keep the raw values

PR #10032, opened 2026-01-18
feat(querybuilder): support using variables inside value

### LINKED ISSUE #10008: Variables can’t be composed inside strings / LIKE patterns in Query Builder (e.g. '$env-suffix', LIKE $env%)
In SigNoz dashboard Query Builder, variables work only as standalone RHS values (e.g. field = $var / field IN $var). There’s no way to compose variables as part of a string or build LIKE patterns inline.

I have a dashboard variable:
environment → values like prod, staging etc.
And a metric label / attribute:
cluster_name → values like prod-xyz, staging-xyz, etc.
I want to filter the panel based on $environment by composing the string:
cluster_name = '$environment-xyz'

Current behavior
Using cluster_name = 'prod-xyz' works.
Using cluster_name = '$environment-xyz' returns No Data (seems treated as a literal / variable parsing fails).

Create a separate variable for full cluster_name

srikanthccv:
### 📄 Summary

prior to this, variables work only as standalone RHS values (e.g. field = $var / field IN $var). this change adds support to use variables inside a value for pattern matching or substitution ex "$env-suffix" or LIKE "%$pattern-var-name%"

Fixes #10008


pkg/querybuilder/where_clause_visitor.go
```diff
@@ -516,6 +525,13 @@ func (v *filterExpressionVisitor) VisitComparison(ctx *grammar.ComparisonContext
 			}
 
 			if ok {
+				if varItem.Type == qbtypes.DynamicVariableType {
+					if all_, ok := varItem.Value.(string); ok && all_ == "__all__" {
```
tushar-signoz: nit: Can we use a constant for `"__all__" `.

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -798,7 +819,12 @@ func (v *filterExpressionVisitor) VisitValue(ctx *grammar.ValueContext) any {
 		// When the user writes an expression like `service.name=redis`
 		// The `redis` part is a VALUE context but parsed as a KEY token
 		// so we return the text as is
-		return ctx.KEY().GetText()
+		keyText := ctx.KEY().GetText()
+		// Check if this is a composed variable like $environment-xyz
+		if strings.HasPrefix(keyText, "$") {
```
tushar-signoz: Are we only support leading `$` sign ? Shouldn't it be Contains ?

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -928,3 +954,103 @@ func trimQuotes(txt string) string {
...
+
+	// find and replace variables by checking each variable name in the map
+	// process longer variable names first to handle cases with prefix substring
+	varNames := make([]string, 0, len(v.variables)*2)
+	for name := range v.variables {
+		varNames = append(varNames, name)
+		// add with $ prefix if not already present
+		if !strings.HasPrefix(name, "$") {
+			varNames = append(varNames, "$"+name)
+		}
```
tushar-signoz: This is not needed, its being handled later in the loop

pkg/querybuilder/where_clause_visitor_test.go
```diff
@@ -10,8 +10,113 @@ import (
...
+			name:  "variable composed with suffix",
+			input: "$environment-xyz",
+			variables: map[string]qbtypes.VariableItem{
+				"environment": {Value: "prod"},
+			},
+			expected: "prod-xyz",
+		},
+		{
+			name:  "variable in quoted string with suffix",
+			input: "$env-cluster",
```
tushar-signoz: input is not quoted

pkg/variables/variable_replace_visitor.go
```diff
@@ -517,6 +542,82 @@ func (v *variableReplacementVisitor) VisitKey(ctx *grammar.KeyContext) any {
...
+// Returns the interpolated string and a boolean indicating if any variable had __all__ value.
+func (v *variableReplacementVisitor) interpolateVariablesInString(s string) (string, bool) {
+	result := s
+
+	varNames := make([]string, 0, len(v.variables)*2)
+	for name := range v.variables {
+		varNames = append(varNames, name)
+		if !strings.HasPrefix(name, "$") {
+			varNames = append(varNames, "$"+name)
+		}
```
tushar-signoz: same

tushar-signoz (commented):
Looks good, minor nits

PR #10033, opened 2026-01-18
fix: do not sort in descending locally if the other is explicitly spe…

### LINKED ISSUE #9281: Table panel should not do any sorting on it's own by default
Image

srikanthccv:
…cified

Fixes #9281 

1. there is a piece of code that always sorts the result by first aggregation, leading to results in reverse order making users believe sorting is not working (it's working but the final result gets rearranged in desc order always)
2. so when we don't sort (some order is specified, the merging logic uses maps whose order is not guaranteed), so ensure we preserve the order, alternatively could import ordered maps but that doesn't make much sense here.

### DIFF
--- a/frontend/src/container/MySettings/LicenseSection/LicenseSection.styles.scss
+++ b/frontend/src/container/MySettings/LicenseSection/LicenseSection.styles.scss
@@ -1,0 +1,84 @@
+.license-section {
+	display: flex;
+	flex-direction: column;
+	gap: 16px;
+
+	.license-section-header {
+		display: flex;
+		flex-direction: column;
+		gap: 4px;
+
+		.license-section-title {
+			color: #fff;
+			font-family: Inter;
+			font-size: 16px;
+			font-style: normal;
+			line-height: 24px;
+			letter-spacing: -0.08px;
+		}
+	}
+
+	.license-section-content {
+		display: flex;
+		flex-direction: column;
+		gap: 16px;
+
+		.license-section-content-item {
+			padding: 16px;
+			border: 1px solid var(--Slate-500, #161922);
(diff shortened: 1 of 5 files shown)


tushar-signoz:
Hey @srikanthccv 
Do you have any default formatter or using `make py-fmt` and `make py-lint` ?

srikanthccv:
I am using `py-fmt` and `py-lint` targets, no default formatter.

tests/integration/src/querier/06_order_by_table_querier.py
```diff
@@ -0,0 +1,1004 @@
...
+                temporality="Unspecified",
+                type_="Gauge",
+                is_monotonic=False,
+                value=value,
+            )
+        )
+    return metrics
+
+
+def make_scalar_query_request(
```
vikrantgupta25, line 88: these should be in fixtures directory, can be re-used across

vikrantgupta25 (changes requested):

tests/integration/src/querier/06_order_by_table_querier.py
srikanthccv: I have another PR coming up, which has more functions, I will move it when things are more clear.

srikanthccv: Also this is my position, it's not a fixture https://github.com/SigNoz/signoz/pull/9976#discussion_r2680490591. For some reasons it will be moved.

srikanthccv: will be addressing in the https://github.com/SigNoz/signoz/pull/10032

Issue #9764, opened 2025-12-02
Improve error message for PromQL queries using old syntax with dotted metric names


dhopkins-omnicell:
## Problem

When writing PromQL queries with metric names containing dots (`.`) using the old syntax, the error message is unclear and doesn't guide users toward the solution.

## Current Behavior

**Query attempted:**
```promql
sum by (k8s.pod.name, deployment.environment) (
  increase(k8s.container.restarts{k8s.pod.name=~".*nifi.*"}[15m])
)
```

**Error received:**
```
invalid_input

invalid promql query "sum by (k8s.pod.name, deployment.environment) ( increase(k8s.container.restarts{k8s.pod.name=~\".*nifi.*\"}[15m]) )"
```

## Correct Working Query

For reference, the working query with proper syntax:
```promql
sum by ("k8s.pod.name", "deployment.environment") (
  increase({"k8s.container.restarts", "k8s.pod.name"=~".*nifi.*"}[15m])
)
```

## Environment
SigNoz Version: v0.97.1
Deployment: Self-hosted
Alert Creation: Via Terraform and UI

## Impact

- Users migrating from older versions or creating new alerts get stuck
- No indication that this is related to the UTF-8 syntax changes
- Requires community support or deep docs reading to resolve
- Slows down alert creation workflow

## Suggested Solution

Add parser-level detection for:

- Metric names with dots/special chars in old syntax
- Label names with dots not quoted
- Provide specific error messages with syntax examples
- Link to migration documentation

## Additional Context

This issue is particularly confusing because:
- The UTF-8 syntax change is relatively recent
- Documentation exists but isn't surfaced in the error
- Users copying queries from older tutorials/examples will hit this
- The error looks like a general parsing error, not a syntax version issue

PR #10034, opened 2026-01-18
chore: return original error message with hints for invalid promql query

srikanthccv:
one step towards better experience for https://github.com/SigNoz/signoz/issues/9764


tushar-signoz (approved):
LGTM

PR #9872, opened 2025-12-26
chore: validate alert payload before saving

abhishekhugetech:
## 📄 Summary

We're breaking the PR into smaller PRs for limiting the surface area for review and reducing blast radius.
- [chore: added Validate function for QueryBuilderFormula struct](https://github.com/SigNoz/signoz/pull/10041)
- [refactor: composite query validation](https://github.com/SigNoz/signoz/pull/10042)
- [chore: validation for ruleCondition](https://github.com/SigNoz/signoz/pull/10044)
- [chore: added new validations in postableRule struct](https://github.com/SigNoz/signoz/pull/10045)

---

## ✅ Changes

- [ ] Feature: Brief description
- [ ] Bug fix: Brief description

---

## 🏷️ Required: Add Relevant Labels

> ⚠️ **Manually add appropriate labels in the PR sidebar**  
Please select one or more labels (as applicable):

ex:

- `frontend`
- `backend`
- `devops`
- `bug`
- `enhancement`
- `ui`
- `test`

---

## 👥 Reviewers

> Tag the relevant teams for review:

- frontend / backend / devops

---

## 🧪 How to Test

1. ...
2. ...
3. ...

---

## 🔍 Related Issues

Closes #

---

## 📸 Screenshots / Screen Recording (if applicable / mandatory for UI related changes)

---

## 📋 Checklist

- [ ] Dev Review
- [ ] Test cases added (Unit/ Integration / E2E)
- [ ] Manually tested the changes

---

## 👀 Notes for Reviewers

- Disabling alerts won't work if existing alert has wrong values present (old alerts) 
- We had to apply the value for CH query template as the queries in template format were syntactically incorrect
- For validating the CH query we're using the `clickhouse-sql-parser` which doesn't complexly validates the syntax of CH queries, need to find some more robust approach to validate the CH query

---

> [!NOTE]
> Strengthens backend validation and error handling for alerts/rules and queries.
> 
> - Add `CompositeQuery.Validate()` with unit checks, PromQL and ClickHouse SQL syntax validation, and "no all-disabled" enforcement; new helpers in `validation.go` with tests
> - Validate formula/functions/join specs (`Function.Validate`, `FunctionName.Validate`, `JoinType.Validate`, `QueryBuilderJoin.Validate`, `QueryBuilderFormula.Validate`); export `GetQueryIdentifier`
[...]

PR #10042, opened 2026-01-19
refactor: composite query validation

abhishekhugetech:
## Pull Request

---

Part of [validate alert payload before saving](https://github.com/SigNoz/signoz/pull/9872) 
Contains refactor of CompositeQuery.Validate which earlier used to validate the older rule queries (builderQueries/chQueries/promQueries in compositeQuery) but now this will validate the newer query format (queries in compositeQuery).

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/pkg/query-service/model/v3/validation.go
+++ b/pkg/query-service/model/v3/validation.go
@@ -78,9 +78,9 @@ func validateClickHouseQuery(query string) error {
 	p := clickhouse.NewParser(queryBuffer.String())
 	_, err = p.ParseStmts()
 	if err != nil {
-		// TODO: errors returned here is errors.errorString, rather than using regex to parser the error
+		// TODO: errors returned here is errors.errorString
 		// we should think on using some other library that parses the CH query in more accurate manner,
-		// current CH parser only does very minimal checks.
+		// current CH parser does very minimal checks and on just the known keywords, without validating the syntax of given query.
 		// Sample Error: "line 0:36 expected table name or subquery, got ;\nSELECT department, avg(salary) FROM ;\n                                    ^\n"
 		return &QueryParseError{
 			ErrorMessage: err.Error(),


pkg/types/ruletypes/threshold.go
```diff
@@ -252,6 +252,15 @@ func (b BasicRuleThreshold) Validate() error {
 		errs = append(errs, errors.NewInvalidInputf(errors.CodeInvalidInput, "invalid match type: %s", string(b.MatchType)))
 	}
 
+	// Only validate unit if specified
+	if b.TargetUnit != "" {
```
abhishekhugetech, line 256: @srikanthccv this is something that might break for existing rules with invalid units, should we log this and see how this turns out in production?

pkg/query-service/model/v3/validation.go
srikanthccv, line 1: How much of the validation implemented in this PR can be part of the `validation.go` in `querybuildertypesv5/validation.go`?

pkg/types/ruletypes/threshold.go
srikanthccv: Yes, let's not put any validation errors on unit for now.

pkg/query-service/app/parser_test.go
```diff
@@ -323,1254 +317,3 @@ func TestParseFilterAttributeValueRequest(t *testing.T) {
 		assert.Equal(t, reqCase.expectedSearchText, filterAttrRequest.SearchText)
 	}
 }
```
abhishekhugetech, line 319: All these tests are for older version of query schema using `PromQueries`, `ClickHouseQueries` and `BuilderQueries`.

srikanthccv:
Will be addressed as part of https://github.com/SigNoz/engineering-pod/issues/3199

PR #10078, opened 2026-01-21
fix: update queryrange for consumer lag view to use dot notation if dotMetrics is enabled 

YounixM:
Consumer Lag View doesn't populate any info

When we shipped dotMetrics, we added logic to conditionally pass `kafka.consumer_group.lag` if dotMeterics is enabled or `kafka_consumer_group_lag` if dotMetrics is disabled  for the account. However, the same was not handled for Consumer Lag View and this broke for the functionality for the users. Making the same change for consumer lag view fixes the issue.

Before:

Screenshot 2026-01-21 at 23 59 34

After
Screenshot 2026-01-21 at 23 47 34

NA - Customer issue tracked via slack
---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

Developer miss and lack of test cases.

Regression

The query accepts the key to be `kafka.consumer_group.lag` if the dotMetrics FF is enabled for the account. Conditionally passing the `kafka.consumer_group.lag` key as aggregate Attribute fetches the required data for the view.

---

- Tests added/updated:
- Manual verification: ✅
- Edge cases covered:

---

No risk as this only fixes the consumer lag view and wouldn't impact any other area in the application

- Blast radius: Consumer Lag view
- Potential regressions:
- Rollback plan:

---

N/A

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---


YounixM:
> This is tested?

Yes @swapnil-signoz . Have tested it for demo and customers.

swapnil-signoz (approved):

PR #10086, opened 2026-01-22
fix: handle attribute collision

tushar-signoz:
### 📄 Summary
This pull request improves the handling of field key resolution in query building and updates the VSCode settings for Go development. The main focus is to make the field key matching logic in the `VisitKey` function more robust, especially when dealing with context and data type prefixes.

Field key resolution improvements:

* Refactored the logic in `VisitKey` (`pkg/querybuilder/where_clause_visitor.go`) to more accurately filter field keys by context and data type, ensuring that only relevant keys are considered. Now, keys are matched if their context and data type are either unspecified or explicitly match, and additional logic handles cases where the context might actually be part of the name (e.g., `span.div_num`).
* Removed redundant fallback logic that attempted to resolve keys by combining context and name, as this is now handled in the main matching logic.

Development environment configuration:

* Updated `.vscode/settings.json` to set the default code formatter for Go files to `golang.go`, improving the Go development experience in VSCode.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -103,9 +103,17 @@
 
 /tests/integration/ @vikrantgupta25
 
+# OpenAPI types generator
+
+/frontend/src/api @SigNoz/frontend-maintainers
+
 # Dashboard Owners
 
 /frontend/src/hooks/dashboard/ @SigNoz/pulse-frontend
+
+## Dashboard Types
+
+/frontend/src/api/types/dashboard/ @SigNoz/pulse-frontend
 
 ## Dashboard List
 
--- a/.github/workflows/goci.yaml
+++ b/.github/workflows/goci.yaml
@@ -65,6 +65,10 @@ jobs:
           set -ex
           sudo apt-get update
           sudo apt-get install -y gcc-aarch64-linux-gnu musl-tools
+      - name: node-install
+        uses: actions/setup-node@v5
+        with:
+          node-version: "22"
       - name: docker-community
         shell: bash
         run: |
--- a/.github/workflows/jsci.yaml
+++ b/.github/workflows/jsci.yaml
@@ -17,10 +17,23 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v4
+      - name: setup node
+        uses: actions/setup-node@v5
+        with:
+          node-version: "22"
       - name: install
         run: cd frontend && yarn install
       - name: tsc
         run: cd frontend && yarn tsc
+  tsc2:
+    if: |
+      (github.event_name == 'pull_request' && ! github.event.pull_request.head.repo.fork && github.event.pull_request.user.login != 'dependabot[bot]' && ! contains(github.event.pull_request.labels.*.name, 'safe-to-test')) ||
+      (github.event_name == 'pull_request_target' && contains(github.event.pull_request.labels.*.name, 'safe-to-test'))
+    uses: signoz/primus.workflows/.github/workflows/js-tsc.yaml@main
+    secrets: inherit
+    with:
+      PRIMUS_REF: main
+      JS_SRC: frontend
   test:
     if: |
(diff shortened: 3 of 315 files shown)


pkg/querybuilder/where_clause_visitor.go
```diff
@@ -809,32 +809,35 @@ func (v *filterExpressionVisitor) VisitKey(ctx *grammar.KeyContext) any {
...
+
+	if fieldKey.FieldContext != telemetrytypes.FieldContextUnspecified {
+
+		contextPrefixedFieldName := fmt.Sprintf("%s.%s", fieldKey.FieldContext.StringValue(), fieldKey.Name)
+		for _, item := range v.fieldKeys[contextPrefixedFieldName] {
+
+			// Match items where: user didn't specify data type OR data type matches. Context filtering not needed here since context was used to construct the lookup key.
+			if fieldKey.FieldDataType == telemetrytypes.FieldDataTypeUnspecified || item.FieldDataType == fieldKey.FieldDataType {
+				fieldKeysForName = append(fieldKeysForName, item)
 			}
```
piyushsingariya: nit: but there multiple empty newlines

pkg/querybuilder/where_clause_visitor_test.go
```diff
@@ -73,3 +73,551 @@ func TestPrepareWhereClause_EmptyVariableList(t *testing.T) {
...
+					},
+				},
+			},
+			expectedKeys: []telemetrytypes.TelemetryFieldKey{
+				{
+					Name:          "host",
+					FieldContext:  telemetrytypes.FieldContextResource,
+					FieldDataType: telemetrytypes.FieldDataTypeString,
+				},
+			},
```
piyushsingariya, line 398: doubt: both should've been expected right?

pkg/querybuilder/where_clause_visitor_test.go
```diff
@@ -73,3 +73,551 @@ func TestPrepareWhereClause_EmptyVariableList(t *testing.T) {
...
+					},
+				},
+				"custom_field": {
+					{
+						Name:          "custom_field",
+						FieldContext:  telemetrytypes.FieldContextAttribute,
+						FieldDataType: telemetrytypes.FieldDataTypeString,
+					},
+				},
+			},
```
piyushsingariya: doubt: since both are being considered now, how would a customer restrict the logic to not choose `attribute.custom_field`

pkg/querybuilder/where_clause_visitor_test.go
tushar-signoz: We prefer resource over attribute and throw warning

pkg/querybuilder/where_clause_visitor_test.go
tushar-signoz: This is a rare case but definitely possible that `attribute.count` and `count` both are valid attributes.
If we consider a case that both have same purposes, same data exists in either of columns. Then its fine even if user cannot restrict.
So we can restrict using `attribute.attribute.count` but not for the other field. Let me think on this.

tushar-signoz: 1 option I can think at top of my head is `log.attribute.count`

tushar-signoz: Hey, they can just say `custom_field`.

srikanthccv (approved):
Thanks for the visitor visit tests.

PR #10089, opened 2026-01-22
test(integration): fixture for notification channel

abhishekhugetech:
## Pull Request

---

### 📄 Summary

Contains fixtures for creation of notification channel.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/tests/integration/src/alerts/01_notification_channel.py
+++ b/tests/integration/src/alerts/01_notification_channel.py
@@ -1,7 +1,7 @@
+import time
 import uuid
 from http import HTTPStatus
 from typing import Callable, List
-import time
 
 import requests
 from wiremock.client import HttpMethods, Mapping, MappingRequest, MappingResponse
@@ -54,10 +54,10 @@ def test_webhook_notification_channel(
         send_resolved=True,
     )
 
-    # TODO: @abhishekhugetech
+    # TODO: @abhishekhugetech # pylint: disable=W0511
     # Time required for Org to be registered
     # in the alertmanager, default 1m.
-    # this will be fixed after https://github.com/SigNoz/engineering-pod/issues/3800
+    # this will be fixed after [https://github.com/SigNoz/engineering-pod/issues/3800]
     time.sleep(65)
 
     # Call test API for the notification channel
@@ -93,6 +93,6 @@ def test_webhook_notification_channel(
         f"Status code: {response.status_code}"
     )
     # Verify that the test notification was sent to the notification channel
-    assert response.json()["count"] == 1, (
-        f"Expected 1 test notification, got {response.json()['count']}"
-    )
+    assert (
+        response.json()["count"] == 1
+    ), f"Expected 1 test notification, got {response.json()['count']}"


tests/integration/fixtures/notification_channel.py
```diff
@@ -0,0 +1,115 @@
...
+            timeout=5,
+        )
+        assert response.status_code == HTTPStatus.CREATED
+        
+        channel_id = response.json()["data"]["id"]
+        channel_ids.append(channel_id)
+        return channel_id
+
+    def _delete_webhook_notification_channel(channel_id: str):
+        logger.info("Deleting channel: %s", {"channel_id": channel_id})
```
srikanthccv:
```suggestion
        logger.info("Deleting channel", {"channel_id": channel_id})
```

tests/integration/fixtures/notification_channel.py
```diff
@@ -0,0 +1,115 @@
...
+            headers={"Authorization": f"Bearer {admin_token}"},
+            timeout=5,
+        )
+        assert response.status_code == HTTPStatus.NO_CONTENT
+
+
+    yield _create_webhook_notification_channel
+    # delete the channel after the test
+    for channel_id in channel_ids:
+        _delete_webhook_notification_channel(channel_id)
```
srikanthccv: Would it make sense to wrap cleanup in try/except?

tests/integration/fixtures/notification_channel.py
```diff
@@ -0,0 +1,115 @@
...
+        channel_ids.append(channel_id)
+        return channel_id
+
+    def _delete_webhook_notification_channel(channel_id: str):
+        logger.info("Deleting channel: %s", {"channel_id": channel_id})
+        response = requests.delete(
+            signoz.self.host_configs["8080"].get(f"/api/v1/channels/{channel_id}"),
+            headers={"Authorization": f"Bearer {admin_token}"},
+            timeout=5,
+        )
```
srikanthccv, line 108: If deletion fails for reasons other than 204, it would be helpful while debugging if there is a log for it? We are comparing `response.status_code` to `HTTPStatus.NO_CONTENT`, so the only info we get from pytest is the value of response.status_code and no other info.

tests/integration/fixtures/notification_channel.py
```diff
@@ -0,0 +1,115 @@
...
+            json={
+                "name": channel_name,
+                "webhook_configs": [
+                    {"send_resolved": send_resolved, "url": webhook_url, "http_config": http_config}
+                ],
+            },
+            headers={"Authorization": f"Bearer {admin_token}"},
+            timeout=5,
+        )
+        assert response.status_code == HTTPStatus.CREATED
```
srikanthccv: Same as above here.

tests/integration/src/alerts/01_notification_channel.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+    assert response.status_code == HTTPStatus.NO_CONTENT
+
+    # Verify that the alert was sent to the notification channel
+    response = requests.post(
+        url=notification_channel.host_configs["8080"].get("/__admin/requests/count"),
+        json={"method": "POST", "url": webhook_endpoint_path},
+        timeout=5,
+    )
+    assert response.json()["count"] == 1
```
srikanthccv: nit: could add description

tests/integration/src/alerts/01_notification_channel.py
```diff
@@ -0,0 +1,78 @@
...
+        headers={"Authorization": f"Bearer {admin_token}"},
+        timeout=5,
+    )
+    assert response.status_code == HTTPStatus.NO_CONTENT
+
+    # Verify that the alert was sent to the notification channel
+    response = requests.post(
+        url=notification_channel.host_configs["8080"].get("/__admin/requests/count"),
+        json={"method": "POST", "url": webhook_endpoint_path},
+        timeout=5,
```
srikanthccv, line 89: I imagine the actual contents of the alert will be tested in later PRs.

tests/integration/fixtures/notification_channel.py
abhishekhugetech: without the `%s` the additional details won't be logged

tests/integration/src/alerts/01_notification_channel.py
abhishekhugetech: Yes

tests/integration/fixtures/notification_channel.py
```diff
@@ -0,0 +1,136 @@
...
+from wiremock.testing.testcontainer import WireMockContainer
+
+from fixtures import dev, types
+from fixtures.auth import USER_ADMIN_EMAIL, USER_ADMIN_PASSWORD
+from fixtures.logger import setup_logger
+
+logger = setup_logger(__name__)
+
+
+@pytest.fixture(name="notification_channel", scope="package")
```
vikrantgupta25, line 18: move this to `http.py` fixture, all the http mock fixtures reside there. use  this notification_channel for all other utils/fixtures.

srikanthccv: It would be prudent to not dump all fixture to `http.py` just because something happens to work with http. It's perfectly that this fixture stays as it is.

tests/integration/fixtures/notification_channel.py
```diff
@@ -0,0 +1,136 @@
...
+        pytestconfig,
+        "notification_channel",
+        lambda: types.TestContainerDocker(id="", host_configs={}, container_configs={}),
+        create,
+        delete,
+        restore,
+    )
+
+
+@pytest.fixture(name="create_webhook_notification_channel", scope="function")
```
vikrantgupta25, line 74: do you want to keep the scope for this as function ? you would be adding multiple tests for notification channels, which will end up cleaning and re-creating them again and again. Rather create this once and use across ?

tests/integration/fixtures/notification_channel.py
```diff
@@ -0,0 +1,136 @@
...
+            timeout=5,
+        )
+        if response.status_code != HTTPStatus.NO_CONTENT:
+            raise Exception( # pylint: disable=broad-exception-raised
+                f"Failed to delete channel, api returned {response.status_code} with response: {response.text}"
+            )
+
+    yield _create_webhook_notification_channel
+    # delete the channel after the test
+    for channel_id in channel_ids:
```
vikrantgupta25: if you let go of the scope function then you don't need to worry about the cleanup as well

tests/integration/fixtures/notification_channel.py
abhishekhugetech: Yes, we've decided to go with different notification channel for each alert test rule so the tests can be run in complete isolation. With this I don't think it should be problem on `cleaning and re-creating` as different channels would get created for different rule. That's why we've kept it `function` scoped so that once a test is completed the respective channel will be deleted.

vikrantgupta25: integrations tests are numbered to make them ordered and non isolated by design, we generally leave dangling resources for that matter, Trying to understand the usecase for keeping these isolated in terms of notification channels? Can you please help me with this understanding

abhishekhugetech: Okay, will change the scope of this fixture to package scoped as we've already decided to use different notification channel for each alert test rule.
Removing the cleanup from this fixture as we're not guaranteeing that the channel will be removed in cleanup. Now with this the fixture will only be responsible for creating the channel and not it's cleanup, this will give a predictable test env where we know that the created channel will persist across tests and can be re-used as per need.

srikanthccv: @abhishekhugetech, we don't control the alert manger, you don't want to share the notification channel b/w the tests. Otherwise it will require special logic to ensure the notification re-delivery of one alert should not interfere with other test.

vikrantgupta25 (approved):

PR #10094, opened 2026-01-22
fix: handle custom time ranges in timezones correctly

YounixM:
When a user has set a particular timezone and edit the custom time, the selected / entered custom time fallback to the browser timezone and sets incorrect custom time.

As we had not considered timezone for parsing the custom time, dayjs would fallback to the browser timezone and set the custom time wrt to the browser timezone, hence losing the timezone difference that we needed to maintain

Before:
https://github.com/user-attachments/assets/4e76be0f-0baa-4e6f-a136-af40718024bc

After:
https://drive.google.com/file/d/1pP_h1YVBbtE06AtgFTWM105J371OlktM/view?usp=sharing

https://github.com/SigNoz/engineering-pod/issues/3421
https://github.com/SigNoz/engineering-pod/issues/3175

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

During the initial implementation of timezone feature, the underlying expectations of the shadcn calendar were not taken into consideration and we were converting the timezone only in case of absolute time range selections - like Last 5 mins , Last 1 hours. It worked as we explicitly handled the time conversion for it and let the calendar component handle timezone functionality.

https://ui.shadcn.com/docs/components/radix/calendar

Screenshot 2026-01-22 at 20 34 10

As we improve our experience and code around datetime picker and timezone, we have now centralized the handling of custom datetime to a util function which converts the custom time correctly based on on the selected timezone.

- Dev miss, Not reading through the documentation thoroughly , Lack of testing

Centralized the handling of custom datetime to a util function which converts the custom time correctly based on on the selected timezone.

---

- Tests added/updated:
- Manual verification: ✅ 
- Edge cases covered:

---

- Blast radius: All things QueryRange and DateTime
- Potential regressions: DateTime isn't set correctly and user sees incorrect time range
- Rollback plan: Revert the PR

---

**N/A**

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

Manually Test all flows around the owned functionalities

---


frontend/src/components/CustomTimePicker/CustomTimePicker.tsx
```diff
@@ -498,20 +475,28 @@ function CustomTimePicker({
 		return '';
 	};
 
+	const focusInput = (): void => {
```
aks07, line 473: What is the reason behind increasing timeout from 0 to 100ms?

frontend/src/components/CustomTimePicker/CustomTimePicker.tsx
```diff
@@ -498,20 +470,28 @@ function CustomTimePicker({
...
+		// Use setTimeout to wait for React to update the DOM and make input editable
```
ishan621, line 474: can we use useLayoutEffect here ?

aks07:
Across the repo we use mixed timestamp conventions (sec vs ms) and apply .tz() directly at many call sites. That makes the intent unclear and increases risk of bugs (unit confusion, TZ parsing issues).

Suggestion:
- Standardize on epoch milliseconds internally.
- Add small, well-named helpers at boundaries:
  - `normalizeToEpochMs(ts: number | string): number`
  - `formatEpochMsWithTZ(epochMs: number, tz: string, format: string): string`
- Replace ad-hoc dayjs.tz(...) usages with these helpers and add unit tests for edge cases (seconds vs ms, ISO strings, invalid input, DST).

Example usage:
`formatEpochMsWithTZ(normalizeToEpochMs(startTime), timezone.value,DATE_TIME_FORMATS.DD_MMM_YYYY_HH_MM_SS)`

This can be a follow up PR and not a blocker

frontend/src/components/CustomTimePicker/CustomTimePicker.tsx
YounixM: In case of the timezone selector view, as we select a timezone, we need to convert the input value to the newly selected timezone trigger focus on the input to enable select all & set the cursor to the end like we do in case of normal input focus. Giving a 100ms delay guarantees that no re-render impacts that behaviour

frontend/src/components/CustomTimePicker/CustomTimePicker.tsx
YounixM: We can park it for later @ishan621

PR #10096, opened 2026-01-23
fix: show time in user selected timezone, overflow issue

YounixM:
fix: show time in user selected timezone, overflow issue, metadata loading

User selected Timezone is not being respected in trace details page and the datetime shown in is user browser timezone.

DateTimes will be shown with respect to user selected timezone.  

If the user had set UTC as the preferred timezone in the application and default timezone on the users system is UTC + 5:30, the date time were being shown in UTC + 5:30 and would create confusion. We now convert the time to user selected timezone in the application.

**Before:**
https://github.com/user-attachments/assets/5c1a9839-ebbb-4033-b3a6-31375542f5b4

**After:**

Screenshot 2026-01-23 at 10 51 41

https://github.com/user-attachments/assets/97ca5bda-af48-48db-a508-ad70991540a4

Closes https://github.com/SigNoz/engineering-pod/issues/3743

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

Timezone conversions across the application were not handled correctly for all instances of the datetime in the application. We are now incrementally fixing it. 

---

Convert the time to user selected timezone in the application.

---

- Tests added/updated: 
- Manual verification: ✅ 
- Edge cases covered: ✅ 

---

- Blast radius: Trace Details View
- Potential regressions: Incorrect time is shown to the user
- Rollback plan: We will fix the issues rather than rolling back the PR.

---

**N/A**

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

**N/A**

---

### DIFF
--- a/frontend/src/container/TraceMetadata/TraceMetadata.tsx
+++ b/frontend/src/container/TraceMetadata/TraceMetadata.tsx
@@ -106,11 +106,10 @@ function TraceMetadata(props: ITraceMetadataProps): JSX.Element {
 							<Tooltip title="Start timestamp">
 								<CalendarClock size={14} />
 							</Tooltip>
-							{startTimeInMs ? (
-								<Typography.Text className="text">{startTimeInMs}</Typography.Text>
-							) : (
-								<Typography.Text className="text">N/A</Typography.Text>
-							)}
+
+							<Typography.Text className="text">
+								{startTimeInMs || 'N/A'}
+							</Typography.Text>
 						</div>
 					</div>
 				)}


frontend/src/container/TraceMetadata/TraceMetadata.tsx
```diff
@@ -79,9 +106,11 @@ function TraceMetadata(props: ITraceMetadataProps): JSX.Element {
 							<Tooltip title="Start timestamp">
 								<CalendarClock size={14} />
 							</Tooltip>
-							<Typography.Text className="text">
-								{formatEpochTimestamp(startTime * 1000)}
-							</Typography.Text>
+							{startTimeInMs ? (
```
ahrefabhi: NIT:
<Typography.Text className="text">{startTimeInMs ? startTimeInMs : 'N/A'}</Typography.Text> ??

YounixM: {startTimeInMs || 'N/A'}

Going ahead with this

Issue #8931, opened 2025-08-26
Migrate quick filters to use /fields/keys and /fields/values


srikanthccv:
- replace the `autocomplete/attribute_keys with `/fields/keys`
- replace the `autocomplete/attribute_values with `/fields/values`
- integrate related values with the value fields

- [x] Migrate External APIs quick filters
- [x] Migrate Exceptions quick filters
- [x] Migrate Infrastructure quick filters
- [x] Migrate Logs explorer quick filters
- [x] Migrate Traces explorer quick filters

srikanthccv:
Issue discussion  - January 06
VIEW RECORDING: https://fathom.video/share/WbBsMsFZZVCBamYdS-DzhinZV6jotELL
Meeting Purpose

[To provide context for revamping the quick filter component.](https://fathom.video/share/WbBsMsFZZVCBamYdS-DzhinZV6jotELL?tab=summary&timestamp=7.0)

Key Takeaways

  - [Problem: Quick filters are outdated, polluting the UI with obsolete data (e.g., destroyed Kubernetes pods) and making excessive API calls.](https://fathom.video/share/WbBsMsFZZVCBamYdS-DzhinZV6jotELL?tab=summary&timestamp=1298.0)
  - [Solution: Migrate the component to a new API (/api/v3/autocomplete/field/values) that supports three critical features:](https://fathom.video/share/WbBsMsFZZVCBamYdS-DzhinZV6jotELL?tab=summary&timestamp=1111.0)
    1.  [Time-based filtering: Shows only active data for the selected time range.](https://fathom.video/share/WbBsMsFZZVCBamYdS-DzhinZV6jotELL?tab=summary&timestamp=910.0)
    2.  [Related values: Displays contextually relevant suggestions based on other active filters.](https://fathom.video/share/WbBsMsFZZVCBamYdS-DzhinZV6jotELL?tab=summary&timestamp=1236.0)
    3.  [Performance optimization: Reduces API calls by caching complete key lists locally.](https://fathom.video/share/WbBsMsFZZVCBamYdS-DzhinZV6jotELL?tab=summary&timestamp=698.0)
  - [Implementation: The migration must use lazy loading (e.g., fetching data only for filters in the viewport) to prevent performance issues from making multiple API calls simultaneously.](https://fathom.video/share/WbBsMsFZZVCBamYdS-DzhinZV6jotELL?tab=summary&timestamp=1819.0)

Topics

The Problem: Outdated Quick Filters

[...]

Issue #3748, opened 2023-10-15
Use io/os package instead of deprecated ioutil package functions


gaurav-gogia:
## Replace deprecated `ioutil` package functions with `os` or `io` package functions
I see that for the following files `ioutil` package is being used to perform I/O operations. However, that package has been deprecated as of [Go v1.16](https://pkg.go.dev/io/ioutil)

1. `server.go`
2. `auth.go`
3. `signozio.go`
4. `http_handler.go`
5. `reader.go`
6. `provision.go`
7. `config_parser_test.go`
8. `dynamic_config.go`
9. `telemetry.go`
10. `auth_test.go`
11. `cold_storage_test.go`

Simply replacing function calls such as `ioutil.ReadAll` with `io.ReadAll` counterpart should remove deprecation warnings.

PR #10110, opened 2026-01-27
Sig 8931 : Migrate quick filters to use /fields/keys and /fields/values

ishan621:
## Pull Request

---

1. 	The current quick filter component uses an old API (/api/v3/autocomplete/attribute/values) with two major flaws:
         No time-based filtering: It shows all historical data, not just what's active in the user's selected time range
          Impact: The UI is polluted with obsolete entities (e.g., destroyed Kubernetes pods), making it impossible to see the current state of infrastructure
2.        No context-aware suggestions: It shows all possible values for a key, ignoring other filters the user has already applied.
Impact: The UI is cluttered with irrelevant options, forcing users to manually filter through long lists

Before: 

https://github.com/user-attachments/assets/22695052-0d3b-460d-bfec-9ed61da96432

After: 

New video attached after time selection added in query Key

https://github.com/user-attachments/assets/3f5f6619-e600-4d4e-bc4f-037abaf56731

https://github.com/user-attachments/assets/4a8eb2d8-ae27-41e4-b505-a70988734c59

#### Issues closed by this PR
[#8931 , #3748](https://github.com/SigNoz/signoz/issues/8931) 
https://github.com/SigNoz/engineering-pod/issues/3748

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix

---

### 🐛 Bug Context
currentAttributeKeys currently contains both checked and unchecked attribute values in a single array. The UI relies on checkedValuesCount to insert a separator between checked and unchecked values.
However, currentAttributeKeys is derived by slicing based on visibleItemsCount. This causes a UI issue where unchecked values are not visible unless the user scrolls all the way to the bottom of the filter list.

Faulty assumption

#### Fix Strategy
We now have checked and unchecked both as separate arrays

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

Risk - Quick filters can break .
Recovery - It is not behind a feature flag hence rollback is the only option

- Blast radius: Wrong data shown to user
- Potential regressions: Needs complete testing on all pages which has query search or quick filters
- Rollback plan: Revert the changes 

---

### 📝 Changelog
Updated api/fields/values and keys api to pass startUnixMilli , endUnixMilli and existingQuery 

Feature

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit b7f83430786edb75b3a9313aded1a0eb62f44040. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
```diff
@@ -113,6 +116,20 @@ function QuerySearch({
 	const [isFocused, setIsFocused] = useState(false);
 	const editorRef = useRef<EditorView | null>(null);
 
+	// Get time range from Redux global state
+	const globalTime = useSelector<AppState, GlobalReducer>(
+		(state) => state.globalTime,
+	);
+
+	// Convert nanoseconds to milliseconds - use useMemo with both deps to avoid multiple updates
+	const timeRange = useMemo(
```
aks07: Can we calculate the start and end time directly in useKeyValueSuggestion hook. So that we would not need to pass it from every usage point?
Same is the case for existingQuery?

frontend/src/types/api/querySuggestions/types.ts
```diff
@@ -45,8 +47,11 @@ export interface QueryKeyValueRequestProps {
 	signal: 'traces' | 'logs' | 'metrics';
 	key: string;
 	searchText: string;
-	signalSource?: 'meter' | '';
+	signalSource?: string;
```
aks07: signalSource should not accept string. It can only be 'meter' or ''
Lets keep this as it is.

frontend/src/components/QuickFilters/tests/QuickFilters.test.tsx
```diff
@@ -224,8 +236,12 @@ describe('Quick Filters', () => {
 
 		render(<TestQuickFilters />);
 
-		// Prefer role if possible; if label text isn’t wired to input, clicking the label text is OK
-		const target = await screen.findByText('mq-kafka');
+		// Wait for the filter to load with data
+		const target = await screen.findByText('mq-kafka', {}, { timeout: 5000 });
+
+		// DEBUG: Print the DOM to see what's rendered
+		screen.debug();
```
aks07: remove this

srikanthccv:
Hi @ishan621, thanks for the PR.

1. There is no UX difference with this change. Let's say I select `severity_text = 'ERROR'`, then the related values are fetched and appear on the top, but it's not apparent as an end user that there are two kinds of values in the dropdown, i.e., associated/related values and all values. It looks like the previous values got rearranged. There should be a better UX here.
2. What does the UX look like when there are no keys/values for the selected time range
    - Before this change, the values for the keys did not have any time limit, so all values we have seen so far are fetched and shown. Lack of values for a key meant they didn't send the key in the first place. But with this change, lack of values doesn't necessarily mean they didn't send the key, but it could be that the key doesn't have data in the selected time range but has values overall. How is this handled?

ishan621:
> Hi @ishan621, thanks for the PR.
> 
> 1. There is no UX difference with this change. Let's say I select `severity_text = 'ERROR'`, then the related values are fetched and appear on the top, but it's not apparent as an end user that there are two kinds of values in the dropdown, i.e., associated/related values and all values. It looks like the previous values got rearranged. There should be a better UX here.
> 2. What does the UX look like when there are no keys/values for the selected time range
>    
>    * Before this change, the values for the keys did not have any time limit, so all values we have seen so far are fetched and shown. Lack of values for a key meant they didn't send the key in the first place. But with this change, lack of values doesn't necessarily mean they didn't send the key, but it could be that the key doesn't have data in the selected time range but has values overall. How is this handled?

@srikanthccv  

What does the UX look like when there are no keys/values for the selected time range -- will show 'no values found'

There is no UX difference with this change ---
does this UI looks better ? 
I see datadog also had similar UI/UX to show values , they don't show associated/related values separately.


https://github.com/user-attachments/assets/35f99a23-3a45-47ed-bf80-d1f52d92f8a8

frontend/src/components/QuickFilters/tests/QuickFilters.test.tsx
ishan621: done

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
ishan621: done

frontend/src/components/QuickFilters/QuickFilters.tsx
```diff
@@ -291,6 +291,27 @@ export default function QuickFilters(props: IQuickFiltersProps): JSX.Element {
 					/>
 				</div>
 			)}
+			<section className="filters-info">
+				<Tooltip
```
aks07, line 295: Image

Not sure if this is the right way to show this information. May be we ca update the existing quick filter settings bubble to show this info.
similar to this:

Image

[NOT A BLOCKER]

frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.tsx
```diff
@@ -112,70 +121,153 @@ export default function CheckboxFilter(props: ICheckboxProps): JSX.Element {
...
+
+			// Only update ref if we have valid API data or if filters actually changed
+			// Don't update if search returned 0 results to preserve unchecked values
+			const hasValidData = keyValueSuggestions && !isLoadingKeyValueSuggestions;
+			if (otherFiltersChanged || hasValidData) {
+				previousFiltersItemsRef.current = currentFiltersItems;
+			}
+
+			if (otherFiltersChanged) {
+				refetchKeyValueSuggestions();
```
aks07, line 179: Do we need this logic or can we just refetch if `currentFilterExpression` changes? Would it not handle for all the cases 🤔

frontend/src/styles.scss
```diff
@@ -592,6 +592,39 @@ body {
 		}
 	}
 
+	.ant-btn-text:hover,
+	.ant-btn-text:focus-visible {
+		background-color: var(--bg-vanilla-200) !important;
```
aks07, line 597: Are we handling these only for lightMode?

aks07:
When the time range changes, the filter suggestions are not reloading. Because these APIs take start and end time as parameters is this expected? @srikanthccv 

update - Ishan
Yes , we will be doing the same to have filter refetched after timerange update

frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.tsx
ishan621: Usecase is to call refetch on all other filters open . Example - you uncheck one value in severity_text filter, if i just refetch on currentFilterExpression it will refetch for all open filters including severity_text which will hamper user experience. User will see severity_text content change while he was on that filter - issue is what if he wanted to apply some other value from severity_text and due to all refetch , he is unable to see the values. 
Datadog also calls refetch for all other open filters except current

frontend/src/components/QuickFilters/QuickFilters.tsx
ishan621: this was UX approved but we can discuss

frontend/src/styles.scss
ishan621: Yes , antd dark theme matches our implementation by default

aks07:
Changes Look good. A couple of things to care of as enhancements:
1. Either move Adaptive filters heading to callout of quick filter settings or keep it as is and remove the settings callout. https://github.com/SigNoz/signoz/pull/10110#discussion_r2859271696
2. The logic for filters refetch can be directly tied to filterExpression changes and we can avoid expensive calculations which may not be needed.
https://github.com/SigNoz/signoz/pull/10110#discussion_r2859422906

Lets handle these in a separate ticket

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
```diff
@@ -454,7 +454,9 @@ function QuerySearch({
 				// Process the response data
 				const responseData = response.data as any;
 				const values = responseData.data?.values || {};
-				const stringValues = values.stringValues || [];
+				const relatedValues = values.relatedValues || [];
+				const stringValues =
+					relatedValues.length > 0 ? relatedValues : values.stringValues || [];
```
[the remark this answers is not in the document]
ishan621: expected

frontend/src/hooks/querySuggestions/useGetQueryKeyValueSuggestions.ts
```diff
@@ -20,18 +25,32 @@ export const useGetQueryKeyValueSuggestions = ({
...
+		AxiosError
+	>({
 		queryKey: [
 			'queryKeyValueSuggestions',
 			key,
 			signal,
 			searchText,
 			signalSource,
 			metricName,
+			timeRangeKey,
```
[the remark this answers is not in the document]
ishan621: intentional

frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.tsx
```diff
@@ -109,70 +120,153 @@ export default function CheckboxFilter(props: ICheckboxProps): JSX.Element {
...
-
 	const {
 		data: keyValueSuggestions,
 		isLoading: isLoadingKeyValueSuggestions,
+		refetch: refetchKeyValueSuggestions,
 	} = useGetQueryKeyValueSuggestions({
 		key: filter.attributeKey.key,
 		signal: filter.dataSource || DataSource.LOGS,
 		signalSource: 'meter',
+		searchText: searchText || '',
```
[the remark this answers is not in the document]
ishan621: required

PR #10113, opened 2026-01-27
fix: improve local setup commands

krsoninikhil:
## Pull Request
This PR fixes 2 issues in local setup, screenshot of the errors are also attached under screenshot heading:
1. **Issue**: Using local project path in docker volume mount for clickhouse was creating permissions issues because clickhouse trying to run chown command to change the ownership of the directory to the clickhouse user. Which was breaking due to permissions not available to the user.
**Fix:** This PR uses a docker managed volume to ensure permissions are managed by container runtime itself.

2. **Issue**:  If signoz project is cloned inside another git repo in the local system, then commitlint command breaks a it uses a relative url and changes the the director to frontend ( a non git root path). So git tries to find the .git directory in parent path and fails if it finds anything other than signoz root path.
**Fix**: Use absolute path of the file argument that git hook passes to the commitlint command.

My setup:
- I'm using colima instead of docker for mac -- https://github.com/abiosoft/colima
- I have a git directory .dotfiles in my home path to manage the dot files.
---

- It change dev docker compose to use docker volume to avoid permission issues.
- It use absolute path for commit linting

- Using project directory in docker volume for clickhouse create permissions problems since clickhouse tries to change the permissions of the directory to be owned by clickhouse user.

Error message on git commit if there is another git repo configured in anywhere in project path, in this example, I've a git repo in home director for managing dotfiles.

image

Error when using previously mounted path in clickhouse:
image

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

Explained above.

Updates the command and flags causing the issue.
---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:
No rollback required, it only affect dev setup.
---

N/A

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 18b542b8c506d0983463600e5a5d6cd5c8ef22a6. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/.husky/commit-msg
+++ b/frontend/.husky/commit-msg
@@ -1,7 +1,10 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-cd frontend && yarn run commitlint --edit $1
+# Convert $1 to absolute path before cd
+commit_msg_file="$(realpath "$1")"
+
+cd frontend && yarn run commitlint "$commit_msg_file"
 
 branch="$(git rev-parse --abbrev-ref HEAD)"
 


krsoninikhil:
Close, this might be user specific edge case.

PR #10131, opened 2026-01-28
test(integration): alert firing verification fixture

abhishekhugetech:
## Pull Request

---

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/tests/integration/fixtures/alertutils.py
+++ b/tests/integration/fixtures/alertutils.py
@@ -15,7 +15,7 @@ logger = setup_logger(__name__)
 
 def collect_webhook_firing_alerts(
     webhook_test_container: types.TestContainerDocker, notification_channel_name: str
-) -> List[dict[str, str]]:
+) -> List[types.FiringAlert]:
     # Prepare the endpoint path for the channel name, for alerts tests we have
     # used different paths for receiving alerts from each channel so that
     # multiple rules can be tested in isolation.
@@ -40,8 +40,7 @@ def collect_webhook_firing_alerts(
         alert_body = alert_body.replace("\n", "")
         alert_dict = json.loads(alert_body)  # parse the alert body into a dictionary
         for a in alert_dict["alerts"]:
-            labels = a["labels"]
-            alerts.append(labels)
+            alerts.append(types.FiringAlert(labels=a["labels"]))
     return alerts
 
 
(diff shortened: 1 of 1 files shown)


tests/integration/fixtures/alerts.py
```diff
@@ -108,3 +111,127 @@ def _insert_alert_data(
...
+                if verified_count == len(alert_expectations.expected_alerts):
+                    logger.info("Got expected number of alerts: %s", {"count": verified_count})
+                    return True
+            else:
+                # No alert is supposed to be fired if should_alert is False
+                if len(firing_alerts) > 0:
+                    break
+
+            # wait for some time before checking again
+            time.sleep(10)
```
abhishekhugetech: default alert eval is 1m, we're checking every 10s for newly fired alert

tests/integration/fixtures/alerts.py
```diff
@@ -108,3 +111,134 @@ def _insert_alert_data(
             insert_logs(logs)
 
     yield _insert_alert_data
+
+
+@pytest.fixture(name="collect_firing_alerts", scope="package")
+def collect_firing_alerts():
```
vikrantgupta25: what are the advantages of defining this as a fixture and not a helper function ?

tests/integration/fixtures/alerts.py
```diff
@@ -108,3 +111,134 @@ def _insert_alert_data(
...
+        # multiple rules can be tested in isolation.
+        rule_webhook_endpoint = f"/alert/{notification_channel_name}"
+
+        url = test_alert_container.host_configs["8080"].get("__admin/requests/find")
+        req = {
+            "method": "POST",
+            "url": rule_webhook_endpoint,
+        }
+        response = requests.post(url, json=req, timeout=5).json()
+        if len(response["requests"]) == 0:  # no alerts fired yet
```
vikrantgupta25: you can remove this check as the code below would do exactly the same. wanted to understand the rationale behind adding this

tests/integration/fixtures/alerts.py
vikrantgupta25: the advantage you can gain here is if you accept `test_alert_container` / `notification_channel` via fixture.

abhishekhugetech: Right no advantage as such, would be better to make `collect_firing_alerts` and `verify_alert_expectation` util functions

tests/integration/fixtures/alerts.py
abhishekhugetech: there were some other logic earlier so wanted to return early, can be remove now

tests/integration/fixtures/alerts.py
vikrantgupta25: please create `alertutils.py` for such functions

PR #10134, opened 2026-01-28
feat(authz): migrate rbac to authz

vikrantgupta25:
### 📄 Summary

- migrate existing rbac system to authz

closes: https://github.com/SigNoz/platform-pod/issues/1278


pkg/sqlmigration/062_migrate_public_dashboards.go
```diff
@@ -0,0 +1,183 @@
+package sqlmigration
```
therealpandey, line 1: Let's delete older tuples as well?

vikrantgupta25: already done in 61st migration

PR #10145, opened 2026-01-29
chore: variable store set up

AshwinBhatkal:
- Sets up an isolated store for variables
- Introduces a makeshift store that uses 'immer' to handle immutable state updates
- State is handled via the native hook useSyncExternalStore
- Hooks are added that will be replaced in the upcoming PRs

Documentation for the same - https://www.notion.so/signoz/Variables-V2-2effcc6bcd198038ac9ad37ce519fd0c

### DIFF
--- a/frontend/src/hooks/useDimensions.ts
+++ b/frontend/src/hooks/useDimensions.ts
@@ -16,13 +16,19 @@ export function useResizeObserver<T extends HTMLElement>(
 	});
 
 	useEffect(() => {
-		const handleResize = debounce((entries: ResizeObserverEntry[]) => {
-			const entry = entries[0];
-			if (entry) {
-				const { width, height } = entry.contentRect;
-				setSize({ width, height });
-			}
-		}, debounceTime);
+		const handleResize = debounce(
+			(entries: ResizeObserverEntry[]) => {
+				const entry = entries[0];
+				if (entry) {
+					const { width, height } = entry.contentRect;
+					setSize({ width, height });
+				}
+			},
+			debounceTime,
+			{
+				leading: true,
+			},
+		);
 
 		const ro = new ResizeObserver(handleResize);
 		const referenceNode = ref.current;


frontend/src/hooks/dashboard/useDashboardVariablesByType.ts
```diff
@@ -0,0 +1,30 @@
...
+	variableType: TVariableQueryType,
+	returnType: 'values',
+): IDashboardVariable[];
+export function useDashboardVariablesByType(
+	variableType: TVariableQueryType,
+	returnType?: 'entries',
+): [string, IDashboardVariable][];
+export function useDashboardVariablesByType(
+	variableType: TVariableQueryType,
+	returnType?: 'values' | 'entries',
```
ahrefabhi, line 19: can we move this to a enum, so that while consuming, we don't have to write literal strings, instead we can write. VariableReturnType.VALUES || VariableReturnType.ENTRIES

enum VariableReturnType {
   VALUES = 'values',
   ENTRIES = 'entries'
}

frontend/src/hooks/dashboard/useWidgetsByDynamicVariableId.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Widgets } from 'types/api/dashboard/getAll';
+
+import { useDashboardVariablesByType } from './useDashboardVariablesByType';
+
+/**
+ * Hook to get a map of dynamic variable IDs to widget IDs that use them.
+ * This is useful for determining which widgets need to be refreshed when a dynamic variable changes.
+ */
+export function useWidgetsByDynamicVariableId(): Record<string, string[]> {
+	const dynamicVariables = useDashboardVariablesByType('DYNAMIC', 'values');
```
ahrefabhi, line 14: use the enum we'll be defining for types

frontend/src/hooks/dashboard/useDashboardVariablesByType.ts
AshwinBhatkal: Discussed offline. Given only 2 types, this is good

frontend/src/hooks/dashboard/useWidgetsByDynamicVariableId.ts
AshwinBhatkal: same as above

YounixM:
> Sets up an isolated store for variables
> Introduces a makeshift store that uses 'immer' to handle immutable state updates
> State is handled via the native hook useSyncExternalStore
> Hooks are added that will be replaced in the upcoming PRs
> 

@AshwinBhatkal - Do we have the approach documented somewhere ? 

Want to understand what is the issue, what are the approaches considered, their pros and cons and why this approach was finalized?

YounixM (approved):

PR #10146, opened 2026-01-29
chore: dashboard container to use new variable store

AshwinBhatkal:
- Builds off this PR - https://github.com/SigNoz/signoz/pull/10145
- Makes use of all the hooks at places where variables are used
- Small refactors

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -19,6 +19,7 @@ pytest_plugins = [
     "fixtures.idp",
     "fixtures.idputils",
     "fixtures.notification_channel",
+    "fixtures.alerts",
 ]
 
 
--- a/tests/integration/fixtures/alerts.py
+++ b/tests/integration/fixtures/alerts.py
@@ -1,0 +1,54 @@
+from http import HTTPStatus
+from typing import Callable
+
+import pytest
+import requests
+
+from fixtures import types
+from fixtures.auth import USER_ADMIN_EMAIL, USER_ADMIN_PASSWORD
+from fixtures.logger import setup_logger
+
+logger = setup_logger(__name__)
+
+
+@pytest.fixture(name="create_alert_rule", scope="function")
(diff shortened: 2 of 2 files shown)


frontend/src/container/DashboardContainer/DashboardDescription/index.tsx
```diff
@@ -56,36 +57,18 @@ import { Base64Icons } from '../DashboardSettings/General/utils';
...
+	sanitizeDashboardData,
+} from './utils';
 
 import './Description.styles.scss';
 
 interface DashboardDescriptionProps {
 	handle: FullScreenHandle;
 }
 
-export function sanitizeDashboardData(
```
AshwinBhatkal, line 67: moved to utils file

frontend/src/container/DashboardContainer/DashboardDescription/utils.ts
```diff
@@ -1,3 +1,27 @@
...
+export function sanitizeDashboardData(
+	selectedData: DashboardData,
+): DashboardData {
+	if (!selectedData?.variables) {
+		return selectedData;
+	}
+
+	const updatedVariables = Object.entries(selectedData.variables).reduce(
+		(acc, [key, value]) => {
+			const { selectedValue: _selectedValue, ...rest } = value;
```
ahrefabhi, line 12: we don't want selectedValue here?

ahrefabhi (approved):

frontend/src/container/DashboardContainer/DashboardDescription/utils.ts
AshwinBhatkal: No. That is not required in the final dashboard JSON. selectedValue is only for retaining state on the UI. Right now I have just copied over the exact same function from frontend/src/container/DashboardContainer/DashboardDescription/index.tsx

Misleading? Yes. We will change this later

PR #10149, opened 2026-01-29
chore: preserve the original duration format

### LINKED ISSUE #9518: Preserving the original duration value provided
```sh

Run terraform apply -auto-approve tfplan
....: Modifying... [id=28964356136]
signoz_alert....: Modifying... [id=...]
...: Modifications complete after 1s [id=28964356136]
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to signoz_alert...., provider
│ "provider[\"registry.terraform.io/signoz/signoz\"]" produced an unexpected
│ new value: .eval_window: was cty.StringVal("60m0s"), but now
│ cty.StringVal("1h0m0s").
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
╵
Error: Terraform exited with code 1.
Error: Process completed with exit code 1.
```

jatinderjit:
## Pull Request

---

### 📄 Summary

Adds `TextDuration` type that preserves the input duration format input by the user.
If user entered "90m0s", they would get back "90m0s" instead of "1h30m0s".

This solves the Terraform provider issue that fails due to the mismatch.

#### Issues closed by this PR
- Closes #9518

---

### ✅ Change Type

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context

```
Run terraform apply -auto-approve tfplan
....: Modifying... [id=28964356136]
signoz_alert....: Modifying... [id=...]
...: Modifications complete after 1s [id=28964356136]
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to signoz_alert...., provider
│ "provider[\"registry.terraform.io/signoz/signoz\"]" produced an unexpected
│ new value: .eval_window: was cty.StringVal("60m0s"), but now
│ cty.StringVal("1h0m0s").
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
╵
Error: Terraform exited with code 1.
Error: Process completed with exit code 1.
```

#### Root Cause

If user inputs "60m0s", the value is parsed into `time.Duration` type. Serializing it again returns "1h0m0s", which is different from what Terraform expects, since it just compares string values.

#### Fix Strategy

We now preserve the input duration format, if available.

---

### 🧪 Testing Strategy

- Tests added/updated:
- Manual verification:
  - Tested [Terraform provider](https://github.com/SigNoz/terraform-provider-signoz) for the create/update/destroy operations.
- Edge cases covered:

---

- Blast radius:
- Potential regressions:

---

### 📝 Changelog

NA

| Field | Value |
|------|-------|
| Description | Preserves the user input duration format |

---

### 📋 Checklist
- [x] Manually tested

---

> [!NOTE]
> **Medium Risk**
> Touches shared rule evaluation/serialization types and scheduling math; while changes are mostly type swaps and conversions, any missed `.Duration()` usage could alter alert evaluation timing or API behavior.
> 
> **Overview**
> Introduces `valuer.TextDuration`, a new duration wrapper that **preserves the user-provided duration string** through JSON/DB round-trips (avoiding normalization like `90m` → `1h30m0s`).
> 
> Migrates alerting/rule APIs and internals from `time.Duration`/custom `ruletypes.Duration` to `TextDuration` across `ruletypes` (`PostableRule`, `Evaluation`, `NotificationSettings`, maintenance `Recurrence`), rule evaluation (`BaseRule`/`Rule` interface, manager task frequency, eval delay/window/hold logic), and updates query range calculations and tests accordingly; also removes the old `ruletypes.Duration` type and tightens a couple of task default-frequency checks and naming/comments.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c938fa03adb09d724f89ee20ef64e2aa6cba6e63. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkg/valuer/text_duration.go
srikanthccv, line 1: 1.
If someone does:
```go
td := types.NewTextDuration(90 * time.Minute)
// Later marshals it - "1h30m0s"
```
However, if they unmarshal "90m" and then remarshal, they still get "90m". This asymmetry is confusing and will surprise `TextDuration` users.

2.
Question: Can we use `text string` instead of `raw   []byte`? So we can get rid of code around trimming `"` quotes. The code as is works, but bit fragile. Not a strong opinion

3.

How about removing the `NewTextDuration` and introducing
```
func ParseDuration(s string) (TextDuration, error)    // preserves text
func FromStdDuration(d time.Duration) TextDuration    // no guarantees on preserving text
```

4.

Does it make sense to make `TextDuration` implement all the functions that `time.Duration` implements?

5.

Let's implement the `sql.Scanner`, `driver.Valuer`, `encoding.TextMarshaler`, and `encoding.TextUnmarshaler` because they work with DB and libraries (YAML, TOML, flag packages) check for text marshal automatically

jatinderjit: 1. Great catch! 🙌 \
  This is still an issue, but the distinct constructors might mitigate it to some extent.
2. Done. We now use `text string` instead of `raw []byte`.
3. I've added `ParseTextDuration`, but retained `NewTextDuration`. Reason: otherwise, the usage would be like `types.FromStdDuration(...)`, and it wouldn't be obvious from the name what type is being constructed. Does it make sense?
4. Not sure. Thoughts related to implementation:
    - One option is to embed `time.Duration`. But then the methods won't return `TextDuration`. It might be confusing. Probably it's better to manually implement them.
    - The `text` field would need to be reset.
    - If we were to replace every instance of `time.Duration` with this implementation, it might make life a bit easier. Otherwise the `Add`/`Sub` methods would always need to account for the current type (unless we embed).
5. Done

srikanthccv: > I've added ParseTextDuration, but retained NewTextDuration. Reason: otherwise, the usage would be like types.FromStdDuration(...), and it wouldn't be obvious from the name what type is being constructed. Does it make sense?

The thing is when I see a `NewX` I treat it as a representative way to create a new instance of a type. In this case, the `TextDuration` type. It doesn't sit well with me that `NewTextDuration` takes the `time.Duration`. How about removing the `NewTextDuration` and having `MustParseTextDuration` that returns the `TextDuration` and panic when it can't. The `NewTextDuration` API doesn't seem good to me.

> Not sure. Thoughts related to implementation

Let me rephrase the intent. How does one go about logging or accessing the `TextDuration` in epoch mills. 

- `td.Duration().Milliseconds()`
- `td.Milliseconds()`

Given some of the methods on https://pkg.go.dev/time#pkg-index `time.Duration` are common, would it make send that `TextDuration` duration implements them while also having `Duration()`?

jatinderjit: > How about removing the `NewTextDuration` and having `MustParseTextDuration`

1. We already have `MustParseTextDuration`. Not using it anywhere right now, but can use it everywhere where we're constructing from a static duration value.
2. We do have a few places where we need to construct this from `time.Duration` ([example](https://github.com/SigNoz/signoz/blob/274aeb557e401abbcd8519fc6841e63f52bf795a/pkg/query-service/rules/threshold_rule.go#L843-L843)). What do you suggest we do here?

    We can stringify it, and back, but it'd be weird! Or we rename `NewTextDuration` to `TextDurationFromDuration`?

---

> would it make sense that `TextDuration` duration implements them while also having `Duration()`?

Makes sense. I'll add helper methods to `TextDuration`.

pkg/types/text_duration.go
```diff
@@ -0,0 +1,139 @@
+package types
```
vikrantgupta25: Following up on the discussions: ( @srikanthccv @jatinderjit )

- I feel this belongs to the `valuer` interface since there is a pressing need for standardisation for time.Duration. We have multiple APIs which accepts all kinds of durations ( TTL in days, TTL in go strings, step interval in seconds to name a few ). Having this in valuer is a step towards the standardisation and can be incrementally improved and adopted across. This brings in predictable UX across the product. You can still use time.Duration where needed without resolving to valuer in cases where needed. Like we do with email and string valuers. 
- If it feels too early for a valuer you can move it inside **ruletypes/ alerttypes** since we do not add things at the root of the types package. 
- If you feel strongly about it being here as is, it's fine for now.

jatinderjit: Thanks @vikrantgupta25. I didn't know about the `Valuer` interface.
I've moved `TextDuration` to the package and implemented the interface.

Regarding standardisation of this Duration struct, I'm not sure if you meant replacing every instance of `time.Duration` with the "standardised" duration. If yes, then I kind of tried, but wouldn't recommend.
- It'd take away the niceties with duration operations (like `5 * time.Minute` and `<duration> / frequency`).
- Doesn't play nicely with `time.Time` (especially `Add`,  tickers,  and also `Sub` which would again need type conversion).

srikanthccv: @vikrantgupta25, since we are here, several things

1. I don't see a pressing need for the standardisation, If you do believe you should show with quantitative measure how that's a pressing need than a self assesment. Something can be done there and we get benefit out of it but that's not a pressing issue. (You can become a leader in the market and people praise for the top notch quality of the product all while having these various formats in the API, and precisely some of the these are accidental and some of them purpose specific than saying there is one and only format that everything has to be).
2. What makes a type to be in `valuer`? Is it the standardisation? (doesn't make sense to say that anything to standardise becomes a type in valuer?
3. Why do we not add things at the root? I am looking for a specific reason as to why? Where does a shared type that will be used by multiple modules?

vikrantgupta25: No @jatinderjit by standardisation i didn't mean to replace every instance of the same. `time.Duration` and `valuer.TextDuration` can co-exist complimenting each other where needed. The aim was not to replace all the instances with this change but to have set a standard of how we internally / externally handle the durations.

vikrantgupta25: @srikanthccv 

1. I mentioned couple of cases in the initial comment regarding the inconsistencies across the API's. From a user standpoint this is a better UX ( until something explicitly needs to deviate from the base ). This could potentially warrant a separate discussion hence the pointer 2 and 3. ( if you don't currently see it the same way, you can ideally go ahead with the current approach, giving a possibility if you see it the same way you could make the change. ) 

2. For all intents and purposes cross cutting module dependencies which demands standardisation makes any type a valuer candidate. This doesn't essentially mean that standing today we need to replace it everywhere / enforce it very strictly with everything going forward. The onus lies on the consumer and the reviewer if they feel it could help them in certain ways. Anything put in valuer doesn't neccessarily means that it should as of first commit solve 100 percent of the issues, but a genuine step towards the standardisation. 

That being said, if you guys feel this is fine where it is currently, we can keep it here for now. No blocker from my side currently.

srikanthccv: There is lack of specificity here and that's where I am trying to bring out of this discussion. "standardisation" is vague and as a result it puts us in a position where we can't say whether a type goes into `valuer`. As I see it, the `pkg/valuer` types are universal replacement to what their original types are, when I say universal replacement I mean there won't be use of the original type across entire codebase (not to be confused with whether that statement holds for codebase at a point in time, but the eventuality i.e entire codebase one will not have original type but the `pkg/valuer` custom type some day). If this is not what the `pkg/valuer` meant for, what is it for? 

Here is how I see the things

1. `pkg/valuer` -> universal replacement to the original type
2. `pkg/types` -> base types (types that we find using across the modules)
3. `pkg/types/mycustomtype` -> types that are contained to some scope of the work

This forms the basis I share with the team and how I review. What I am looking for is that assessment of above and a doc that guides author and reviewers when to put things where so that such discussions won't come up at the time of the review.

The reason I am using this discussion for that is I have been finding us doing a lot of [bike shedding](https://en.wikipedia.org/wiki/Law_of_triviality) and things on ad-hoc basis which I believe is not very productive. Let me know if this makes sense and how do we move forward.

vikrantgupta25: Let me create a discussion ticket where we close all such concerns and design principles and then have it in our codebase as the contributing/style guide to avoid these discussions on the PR. Also for the current PR i have no blocking concerns with the initial approach as well since we do lack a guide on how to justify the principle. 

Let's address the above valuer principle question in the discussion ticket itself.

srikanthccv: Alright, thanks.

pkg/valuer/text_duration.go
```diff
@@ -0,0 +1,159 @@
...
+		return errors.Wrap(err, errors.TypeInvalidInput, errors.CodeInvalidInput, "failed to parse duration text")
+	}
+	d.value = tmp
+	d.text = s
+	return nil
+}
+
+// Value implements [driver.Valuer] by delegating to the underlying duration.
+func (d TextDuration) Value() (driver.Value, error) {
+	return int64(d.value), nil
```
srikanthccv: This won't preserve the original duration across roundtrips i.e scan -> value -> scan. Can we please update the tests to ensure this works.

jatinderjit: That would mean we'll have to always save it as text in the database, right?
We're supporting parsing from int in `Scan` (and Unmarshal). If we just want to support text, should we remove support for that as well?

srikanthccv: Yes, let's remove that, otherwise, it doesn't make sense to introduce "TextDuration" if it can't preserve the original duration

pkg/valuer/text_duration.go
```diff
@@ -0,0 +1,159 @@
...
+	switch value := v.(type) {
+	case float64:
+		d.value = time.Duration(value)
+		d.text = ""
+		return nil
+
+	case string:
+		tmp, err := time.ParseDuration(value)
+		if err != nil {
+			return err
```
srikanthccv: Let's wrap both these errors with `errors.TypeInvalidInput`

```
	if err := json.Unmarshal(b, &v); err != nil {
		return err
	}
```

```
		tmp, err := time.ParseDuration(value)
		if err != nil {
			return errCollapse comment
```

pkg/valuer/text_duration.go
```diff
@@ -0,0 +1,159 @@
...
+// It serializes the duration value in a human-readable format (1h30m0s).
+// If the original text is available, it is returned as-is. Example: 90m is not normalized to 1h30m0s.
+func (d TextDuration) MarshalJSON() ([]byte, error) {
+	return json.Marshal(d.String())
+}
+
+// UnmarshalJSON implements the [encoding/json.Unmarshaler] interface.
+// It parses string or numeric durations, and stores the string representation.
+func (d *TextDuration) UnmarshalJSON(b []byte) error {
+	var v interface{}
```
srikanthccv: nit
```suggestion
	var v any
```

srikanthccv (commented):
thanks, LGTM, two minor comments

pkg/valuer/text_duration.go
jatinderjit: Done

srikanthccv (approved):

srikanthccv (approved):

PR #10150, opened 2026-01-29
Show both tag/resource variants for identical keys in pipeline filter options

aks07:
## Pull Request

---

Client-side dedupe in useOptions removed distinct suggestions that shared the same key by comparing only label and value. This collapsed entries like service.name coming from different sources (Type: tag vs resource, or different dataType).
Additionally, the dropdown rendered items with key={option.label}, causing React key collisions so only one of the duplicates appeared.

- Updated useOptions dedupe to include type and dataType in the uniqueness predicate, preserving options that differ by either dimension.
- Made dropdown item keys unique by using a composite key ${label}-${type}-${dataType} in QueryBuilderSearch, ensuring both variants render and are navigable.
- Added an integration test in PipelineListsView.test.tsx to mock /api/v3/autocomplete/attribute_keys and verify both service.name entries show with correct “Type: resource/tag” badges

---

Before: service.name shown only once. ie. resource type

image

After: service.name for both tag and resource shown in suggestions

image

https://github.com/SigNoz/engineering-pod/issues/3687
---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

- Tests added/updated: yes
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/frontend/src/container/PipelinePage/tests/PipelineListsView.test.tsx
+++ b/frontend/src/container/PipelinePage/tests/PipelineListsView.test.tsx
@@ -1,10 +1,16 @@
 /* eslint-disable sonarjs/no-duplicate-string */
-import { screen } from '@testing-library/react';
+import { screen, within } from '@testing-library/react';
 import { ENVIRONMENT } from 'constants/env';
 import { server } from 'mocks-server/server';
 import { rest } from 'msw';
 import { PreferenceContextProvider } from 'providers/preferences/context/PreferenceContextProvider';
-import { findByText, fireEvent, render, waitFor } from 'tests/test-utils';
+import {
+	findByText,
+	fireEvent,
+	render,
+	userEvent,
+	waitFor,
+} from 'tests/test-utils';
 import { DataTypes } from 'types/api/queryBuilder/queryAutocompleteResponse';
 
 import { pipelineApiResponseMockData } from '../mocks/pipeline';
@@ -289,6 +295,7 @@ describe('PipelinePage container test', () => {
 	});
 
 	it('should have populated form fields when edit pipeline is clicked', async () => {
+		const user = userEvent.setup({ pointerEventsCheck: 0 });
 		render(
 			<PreferenceContextProvider>
 				<PipelineListsView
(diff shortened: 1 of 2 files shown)


frontend/src/container/PipelinePage/tests/PipelineListsView.test.tsx
```diff
@@ -301,5 +318,53 @@ describe('PipelinePage container test', () => {
...
+				),
+			),
+		);
+
+		// Open Filter input and type to trigger suggestions
+		const input = document.querySelector(
+			'.ant-select-selector .ant-select-selection-search',
+		) as HTMLInputElement;
+
+		await fireEvent.click(input);
```
AshwinBhatkal: use userEvent
For next line, use findBy
https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find

frontend/src/container/PipelinePage/tests/PipelineListsView.test.tsx
```diff
@@ -301,5 +318,53 @@ describe('PipelinePage container test', () => {
...
+							],
+						},
+					}),
+				),
+			),
+		);
+
+		// Open Filter input and type to trigger suggestions
+		const input = document.querySelector(
+			'.ant-select-selector .ant-select-selection-search',
```
AshwinBhatkal: can we use test id here

aks07: i am not in favour of using test-id by default. It should be the last resort in my opinion.

frontend/src/container/PipelinePage/tests/PipelineListsView.test.tsx
aks07: i do get that userEvent is a better choice most of the times but there seems to be some issue with it in this case. 
Also findBy is again a choice that we can make and not necessary i feel.

AshwinBhatkal (dismissed):

PR #10165, opened 2026-02-02
fix: handle empty not() expression

tushar-signoz:
### 📄 Summary
Need to handle possible uses of NOT() operator

### 2. `not(False)` or `not("abc")` or `not(attribute.known.key)` or `not(resource.known.key)` or `not(unknown.key)` --> Something went wrong
The lexer parses above as Not of free text search which makes sense but while creating resource cte it breaks
```
WITH __resource_filter AS (
    SELECT
        fingerprint
    FROM
        signoz_logs.distributed_logs_v2_resource
    WHERE
        NOT () <- This is wrong
        AND seen_at_ts_bucket_start >= ?
        AND seen_at_ts_bucket_start <= ?
)
```
### `not(resource.known.key == 'xyz')` or `not(resource.known.key1 == 'xyz' and resource.known.key2 == 'xyz')` --> Something went wrong
In this case resource cte is created correctly however where clause of main query is broken
```
SELECT
    timestamp,
    id,
FROM
    signoz_logs.distributed_logs_v2
WHERE
    resource_fingerprint GLOBAL IN (
        SELECT
            fingerprint
        FROM
            __resource_filter
    )
    AND NOT ()                            <- THIS IS WRONG
    AND timestamp >= ?
    AND ts_bucket_start >= ?
    AND timestamp < ?
    AND ts_bucket_start <= ?
ORDER BY
    timestamp AS `timestamp` desc,
    id AS `id` desc
LIMIT
    ?
```
### `not(attribute.known.key == 'xyz')` or `not(attribute.known.key1 == 'xyz' and attribute.known.key2 == 'xyz')` -> NO RESULT

```
WITH __resource_filter AS (
    SELECT
        fingerprint
    FROM
        signoz_logs.distributed_logs_v2_resource
    WHERE
        NOT ((true))     <- THIS IS TECHNICALLY WRONG
        AND seen_at_ts_bucket_start >= ?
        AND seen_at_ts_bucket_start <= ?
)
```
### `not(attribute.known.key1 == 'xyz' and resource.known.key2 == 'xyz')` --> Works correctly

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> **Medium Risk**
> Changes core querybuilder boolean/`NOT` semantics and how ignored conditions are treated, which can subtly alter filtering behavior across logs/traces; mitigated by extensive new unit and integration coverage.
> 
> **Overview**
> Fixes filter-expression clause generation to prevent invalid `NOT ()` SQL and incorrect semantics when parts of an expression are skipped (resource-only filtering, full-text skipping, function-call skipping, or unknown keys).
> 
> `where_clause_visitor.go` now standardizes on `TrueConditionLiteral`, short-circuits `OR` when any branch is `true`, strips `true` no-ops from `AND`, treats ignored/missing comparisons and skipped full-text/function calls as `true`, and avoids wrapping/negating `true` so `NOT` never emits `not()`.
> 
> Adds a large new unit test suite for resource-filter statement building and expression visitation, updates existing logs/traces statement builder expectations, and introduces an integration test that fuzzes 10k filter expressions to ensure the server never returns HTTP 500.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1f672e15ac1a6670e6087a5fbd9131639f1c81f2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/querybuilder/fallback_expr.go
+++ b/pkg/querybuilder/fallback_expr.go
@@ -214,7 +214,10 @@ func DataTypeCollisionHandledFieldName(key *telemetrytypes.TelemetryFieldKey, va
 		case []any:
 			if allFloats(v) {
 				tblFieldName = castFloat(tblFieldName)
-			} else if hasString(v) {
+			} else {
+				// Any mix that is not all-floats (e.g. [bool, float64], [bool], all-strings)
+				// must be stringified: passing a Go bool as UInt8 against a String column
+				// causes ClickHouse error 386 "no supertype for String and UInt8".
 				_, value = castString(tblFieldName), toStrings(v)
 			}
 		case bool:


pkg/querybuilder/where_clause_visitor.go
```diff
@@ -280,6 +280,11 @@ func (v *filterExpressionVisitor) VisitUnaryExpression(ctx *grammar.UnaryExpress
 
 	// Check if this is a NOT expression
 	if ctx.NOT() != nil {
+		// If the inner expression is empty (filtered out), return empty
+		// to avoid generating invalid "not()" in ClickHouse
+		if result == "" {
+			return ""
+		}
```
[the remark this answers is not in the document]
tushar-signoz: It lacks testing in general, will invest time some other time.

srikanthccv:
@tushar-signoz, neither the issue nor the PR describe when this issue occurs. Without that, this is an attempt to fix something we don't fully understand, so we can't say if we are fixing the root cause. can we please get the reproducible case first and then come to the fix?

tushar-signoz:
Hey @srikanthccv 
Can you refer the updated PR description

tushar-signoz:
Hey @srikanthccv 
A gentle reminder on this

srikanthccv:
@tushar-signoz , let's add some integration tests, and target the mix of body full text searches that contain the word `not` and genuine `not` involved expressions.

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -291,6 +316,10 @@ func (v *filterExpressionVisitor) VisitPrimary(ctx *grammar.PrimaryContext) any
 	if ctx.OrExpression() != nil {
 		// This is a parenthesized expression
 		if condExpr, ok := v.Visit(ctx.OrExpression()).(string); ok && condExpr != "" {
+			// Don't wrap "true" in parentheses - it's a no-op condition
+			if condExpr == TrueConditionLiteral {
+				return TrueConditionLiteral
+			}
```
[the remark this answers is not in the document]
tushar-signoz: Not touching in this PR

srikanthccv:
@tushar-signoz interested in understanding the test methodology of the integration tests added. What were the main objective? Correctness?

tushar-signoz:
> @tushar-signoz interested in understanding the test methodology of the integration tests added. What were the main objective? Correctness?

test_not_filter_expression -> correctness
test_filter_expressions_no_server_error -> resiliency ~150 of these return 5XX

Was initially thinking of adding it in unit tests but that looked like polluting the codebase.

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -256,14 +261,26 @@ func (v *filterExpressionVisitor) VisitOrExpression(ctx *grammar.OrExpressionCon
...
+			unaryExpressionConditions = append(unaryExpressionConditions, condExpr)
 		}
 	}
 
+	// If there are no conditions and hasTrueConditionLiteral is set, return TrueConditionLiteral
 	if len(unaryExpressionConditions) == 0 {
+		if hasTrueConditionLiteral {
+			return TrueConditionLiteral
+		}
```
srikanthccv: Do we really need this conditional block?

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -280,6 +297,14 @@ func (v *filterExpressionVisitor) VisitUnaryExpression(ctx *grammar.UnaryExpress
 
 	// Check if this is a NOT expression
 	if ctx.NOT() != nil {
+		// If the inner expression is empty (filtered out), return empty
+		// to avoid generating invalid "not()" in ClickHouse
+		if result == "" {
+			return TrueConditionLiteral
+		}
+		if result == TrueConditionLiteral {
+			return result
```
srikanthccv:
```suggestion
		// If the inner expression is empty (filtered out), return true
		// to avoid generating invalid "not()" in ClickHouse
		if result == "" || result == TrueConditionLiteral {
			return TrueConditionLiteral
```

tests/integration/src/querier/08_filter_expression.py
```diff
@@ -0,0 +1,256 @@
...
+                        },
+                    }
+                ]
+            },
+            "formatOptions": {"formatTableResultForUI": False, "fillGaps": False},
+        },
+    )
+
+
+def _log_bodies_from_response(response_json: dict) -> Set[str]:
```
srikanthccv: Let's put this in the `querier.py` fixture, there we already have `get_scalar_*`, `get_all_series`, and `get_series_values`.

tests/integration/src/querier/08_filter_expression.py
```diff
@@ -0,0 +1,256 @@
...
+from fixtures.logs import Logs
+
+TESTDATA_DIR = os.path.join(os.path.dirname(__file__), "..", "..", "testdata")
+FILTER_EXPRESSIONS_FILE = os.path.join(TESTDATA_DIR, "filter_expressions_10000.txt")
+
+
+def _make_raw_logs_query(
+    signoz: types.SigNoz,
+    token: str,
+    filter_expression: str,
```
srikanthccv: Seems like very this wrapper that's similar to `make_query_request`. Let's use `make_query_request` only?

pkg/querybuilder/where_clause_visitor.go
tushar-signoz: So basically distinction is "" (empty) vs true. 
This would be required in case of cond1 OR cond2
Lets assume cond2 is what we're talking here, which results in `cond1 OR ""` or `cond1 OR true`. Both mean different things.

tushar-signoz: Let me double check on this

pkg/querybuilder/constants.go
```diff
@@ -4,6 +4,10 @@ import (
 	"os"
 )
 
+const (
+	TrueConditionLiteral = "true"
```
srikanthccv: Let's use some literal that can't collide with real SQL?

pkg/querybuilder/where_clause_visitor.go
srikanthccv: @tushar-signoz ,
- let's write down the case we discussed as code comment in detail here
- let's ensure there are integration tests covering it.

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -354,7 +380,10 @@ func (v *filterExpressionVisitor) VisitComparison(ctx *grammar.ComparisonContext
 	keys := v.Visit(ctx.Key()).([]*telemetrytypes.TelemetryFieldKey)
 
 	// if key is missing and can be ignored, the condition is ignored
-	if len(keys) == 0 && v.ignoreNotFoundKeys {
+	if len(keys) == 0 {
+		if v.ignoreNotFoundKeys {
+			return TrueConditionLiteral
+		}
 		return ""
 	}
```
srikanthccv, line 376: The change is not same as the earlier condition. If `keys` is nil and `ignoreNotFoundKeys` is false, it returns empty. While it doesn't happen in practice (because the keys either returns error or keys), let's not handle the empty keys here, it implies the empty keys are silently skipped.

pkg/querybuilder/resourcefilter/statement_builder_test.go
```diff
@@ -0,0 +1,671 @@
...
+				Args:  []any{"redis", "%service.name%", "%service.name\":\"redis%", expectedBucketStart, expectedBucketEnd},
+			},
+		},
+		{
+			name: "attribute filter with NOT and inner expression",
+			query: qbtypes.QueryBuilderQuery[qbtypes.TraceAggregation]{
+				Signal: telemetrytypes.SignalTraces,
+				Filter: &qbtypes.Filter{
+					// http.request.method is an attribute field, not a resource field
+					// so the condition returns "true", and NOT should also return "true" (not "NOT (true)")
```
srikanthccv, line 354:
```suggestion
					// http.request.method is an attribute field, not a resource field
					// so the condition returns "true", and NOT should also return "true" (not "NOT (true)")
					// Strictly, NOT(true) = false. But in this system, TrueConditionLiteral means 
					// "this condition is not evaluable here" and the negation of "not evaluable" is also
					// "not evaluable", so true is the right no-op. Returning false would incorrectly exclude all rows.
```

pkg/querybuilder/resourcefilter/statement_builder_test.go
```diff
@@ -0,0 +1,671 @@
...
+		},
+		{
+			name: "complex resource filter with mixed conditions for logs",
+			query: qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]{
+				Signal: telemetrytypes.SignalLogs,
+				Filter: &qbtypes.Filter{
+					// env and k8s.deployment.name are resource fields
+					// severity_text is an attribute field (returns true)
+					// Multiple grouped conditions with attribute fields
+					Expression: "env = 'prod' AND k8s.deployment.name = 'fnscrapers' AND severity_text = 'ERROR' AND severity_text = 'WARN' AND (severity_text = 'INFO' AND severity_text = 'DEBUG') AND (severity_text = 'TRACE' AND severity_text = 'FATAL') AND (severity_text = 'a' AND severity_text = 'b') AND (severity_text = 'c' AND severity_text = 'd') AND (severity_text = 'e' AND severity_text = 'f')",
```
srikanthccv:
```suggestion
					Expression: "env = 'prod' AND k8s.deployment.name = 'prod-deployment' AND severity_text = 'ERROR' AND severity_text = 'WARN' AND (severity_text = 'INFO' AND severity_text = 'DEBUG') AND (severity_text = 'TRACE' AND severity_text = 'FATAL') AND (severity_text = 'a' AND severity_text = 'b') AND (severity_text = 'c' AND severity_text = 'd') AND (severity_text = 'e' AND severity_text = 'f')",
```

pkg/querybuilder/resourcefilter/statement_builder_test.go
```diff
@@ -0,0 +1,671 @@
...
+			},
+		},
+		{
+			name: "NOT with multiple attribute fields should not generate NOT (true and true)",
+			query: qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]{
+				Signal: telemetrytypes.SignalLogs,
+				Filter: &qbtypes.Filter{
+					// http.request.method is an attribute field, not a resource field
+					// so the condition returns "true", and NOT should also return "true" (not "NOT (true)")
+					Expression: "not(http.request.method = 'POST' and http.request.method = 'GET')",
```
srikanthccv:
```suggestion
					Expression: "not(http.request.method = 'POST' OR http.request.method = 'GET')",
```

Let's feed good data to LLMs they take inspiration from existing tests.

pkg/querybuilder/resourcefilter/statement_builder_test.go
```diff
@@ -0,0 +1,671 @@
...
+			name: "resource filter with variable substitution",
+			query: qbtypes.QueryBuilderQuery[qbtypes.TraceAggregation]{
+				Signal: telemetrytypes.SignalTraces,
+				Filter: &qbtypes.Filter{
+					Expression: "service.name = $service_name",
+				},
+			},
+			variables: map[string]qbtypes.VariableItem{
+				"service_name": {
+					Value: "redis-manual",
```
srikanthccv:
```suggestion
					Expression: "service.name = $service.name",
				},
			},
			variables: map[string]qbtypes.VariableItem{
				"service.name": {
					Value: "redis-manual",
```

pkg/querybuilder/where_clause_visitor_test.go
```diff
@@ -659,3 +662,1448 @@ func TestVisitKey(t *testing.T) {
...
+//   of FieldContext. Opts: SkipFullTextFilter:false, SkipFunctionCalls:false,
+//   IgnoreNotFoundKeys:false, FullTextColumn:bodyCol.
+//
+// Key behavioral rules:
+//
+//   resourceConditionBuilder:
+//     attr key (a,b,c)          → TrueConditionLiteral
+//     resource key (x,y,z)      → "{name}_cond"
+//     full-text / function call → TrueConditionLiteral (skipped)
+//     unknown key               → "" (ignored; IgnoreNotFoundKeys=true)
```
srikanthccv: Do we need this `""` and `TrueConditionLiteral` distinction? Can everything be `TrueConditionLiteral`?

pkg/querybuilder/where_clause_visitor_test.go
```diff
@@ -659,3 +662,1448 @@ func TestVisitKey(t *testing.T) {
...
+	"z": {{Name: "z", FieldContext: telemetrytypes.FieldContextResource, FieldDataType: telemetrytypes.FieldDataTypeString}},
+	"ax": {{Name: "ax", FieldContext: telemetrytypes.FieldContextAttribute, FieldDataType: telemetrytypes.FieldDataTypeNumber},
+		{Name: "ax", FieldContext: telemetrytypes.FieldContextResource, FieldDataType: telemetrytypes.FieldDataTypeString}},
+	"by": {{Name: "by", FieldContext: telemetrytypes.FieldContextAttribute, FieldDataType: telemetrytypes.FieldDataTypeNumber},
+		{Name: "by", FieldContext: telemetrytypes.FieldContextResource, FieldDataType: telemetrytypes.FieldDataTypeString}},
+	"cz": {{Name: "cz", FieldContext: telemetrytypes.FieldContextAttribute, FieldDataType: telemetrytypes.FieldDataTypeNumber},
+		{Name: "cz", FieldContext: telemetrytypes.FieldContextResource, FieldDataType: telemetrytypes.FieldDataTypeString}},
+}
+
+type resourceConditionBuilder struct{ m map[string]string }
```
srikanthccv: I don't see where the `m` map used? Why does it exist?

pkg/querybuilder/where_clause_visitor_test.go
srikanthccv, line 1: There are ton of unit tests here, I was not able to go through all of them. It's really important to have tests but how many of the tests added here are redundant and test the right cases enough that they are not a burden for test maintenance?

tests/integration/src/querier/08_filter_expression.py
```diff
@@ -0,0 +1,289 @@
...
+TESTDATA_DIR = os.path.join(os.path.dirname(__file__), "..", "..", "testdata")
+FILTER_EXPRESSIONS_FILE = os.path.join(TESTDATA_DIR, "filter_expressions_10000.txt")
+
+
+@pytest.mark.parametrize(
+    "expression,expected_logs",
+    [
+        # NOT on resource attribute unique to alpha → beta only
+        pytest.param(
+            'NOT resource.f1 = "v10"',
```
srikanthccv: It was hard to reason about the change with `fx` and their mixed usage (they were distracting the main line of thought about the correctness reasoning). Please use the some real work example key names.

pkg/querybuilder/where_clause_visitor.go
tushar-signoz: Let me add a comment regarding error being logged by the VisitKey downstream. Reason for returning "" here is that code following this section has to do repeated checks on length ( 1 branch access keys[0] directly which panics in some cases ) ?

pkg/querybuilder/constants.go
tushar-signoz: Something like `__true__` and then generate the expression and at last step replace `__true__` with `true` ?
Don't think it adds any safety, because we have to convert it to `true` at some stage. 
Also this helps if filter expression contains 'true' keyword.

pkg/querybuilder/where_clause_visitor.go
srikanthccv: Sure

pkg/querybuilder/constants.go
srikanthccv: No strong point here other than potential conflicts with `true` down the line ...

pkg/querybuilder/resourcefilter/statement_builder_test.go
tushar-signoz: The tests needs `AND` operator. I'll change the other attribute to make this meaningful

tests/integration/src/querier/08_filter_expression.py
tushar-signoz: Done

pkg/querybuilder/where_clause_visitor_test.go
tushar-signoz: Done

tushar-signoz:
@srikanthccv 
Summary of what has changed since you last reviewed. 
1. Now only 1 value is used SkipConditionLiteral = '__skip__' instead of using `'true'`, `''`.
2. Once expression is generated  SkipConditionLiteral is replaced with TrueConditionLiteral
3. Unit tests are more organised
4. OR with skip -> skip
5. And with skip -> and 
6. NOT (skip) -> skip

pkg/querybuilder/where_clause_visitor.go
tushar-signoz: Condition is removed, it is not needed by using `SkipConditionLiteral` as return value

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -165,9 +165,12 @@ func PrepareWhereClause(query string, opts FilterExprVisitorOpts, startNs uint64
...
 
 	if cond == "" {
-		cond = "true"
+		cond = SkipConditionLiteral
 	}
 
+	// Replace all SkipConditionLiteral with TrueConditionLiteral
+	cond = strings.ReplaceAll(cond, SkipConditionLiteral, TrueConditionLiteral)
```
srikanthccv: This `strings.ReplaceAll` has a chance of colliding with the user input, while it's unlikely, but it's no impossible. We should have a better way instead of naive replace.

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -328,27 +350,27 @@ func (v *filterExpressionVisitor) VisitPrimary(ctx *grammar.PrimaryContext) any
 				searchText = valCtx.KEY().GetText()
 			} else {
 				v.errors = append(v.errors, fmt.Sprintf("unsupported value type: %s", valCtx.GetText()))
-				return ""
+				return SkipConditionLiteral
```
srikanthccv: Let's not return `SkipConditionLiteral` in error condition places? While the change made in the last commit to use only `SkipConditionLiteral` makes sense, the `""` returned in erroneous situations doesn't mean skip but rather restricted by the interface to return something. It is confusing to use `SkipConditionLiteral` in erroneous situations, as otherwise, for the reader it implies we skip error cases. What do you think?

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -328,27 +350,27 @@ func (v *filterExpressionVisitor) VisitPrimary(ctx *grammar.PrimaryContext) any
...
 			}
 		}
 		cond, err := v.conditionBuilder.ConditionFor(context.Background(), v.fullTextColumn, qbtypes.FilterOperatorRegexp, FormatFullTextSearch(searchText), v.builder, v.startNs, v.endNs)
 		if err != nil {
 			v.errors = append(v.errors, fmt.Sprintf("failed to build full text search condition: %s", err.Error()))
-			return ""
+			return SkipConditionLiteral
```
srikanthccv: Similarly rest of the file, does it make sense to not return `SkipConditionLiteral` errorneous cases?

pkg/querybuilder/where_clause_visitor_test.go
```diff
@@ -659,3 +662,950 @@ func TestVisitKey(t *testing.T) {
...
+				expr, _ := result.WhereClause.Build()
+				assert.Equal(t, tt.wantSB, expr, "conditionBuilder SQL mismatch:\n  want: %s\n   got: %s", tt.wantSB, expr)
+			}
+		})
+	}
+}
+
+// TestVisitComparison_UnknownKeys covers unknown key handling.
+// rsbOpts has IgnoreNotFoundKeys=true → VisitComparison returns "" which is silently
+// dropped from AND/OR; an all-"" AND collapses to TrueConditionLiteral.
```
srikanthccv: The phrasing here is not accurate?

pkg/querybuilder/where_clause_visitor_test.go
```diff
@@ -659,3 +662,950 @@ func TestVisitKey(t *testing.T) {
...
+	tests := []visitComparisonCase{
+		{
+			// RSB: "" dropped from AND; x_cond survives.
+			name:      "unknown key AND resource",
+			expr:      "unknown_key = 'val' AND x = 'x'",
+			wantRSB:   "WHERE x_cond",
+			wantErrSB: true,
+		},
+		{
+			// RSB: "" dropped from OR; only x_cond remains (no OR wrapper).
```
srikanthccv: This is also misleading comment

pkg/querybuilder/where_clause_visitor_test.go
```diff
@@ -659,3 +662,950 @@ func TestVisitKey(t *testing.T) {
...
+			// SB (no OR): a IN __all__→true; x filtered → ""; AND: no real conds → true.
+			name:    "attr IN allVariable AND resource",
+			expr:    "a IN $service AND x = 'x'",
+			wantRSB: "WHERE x_cond",
+			wantSB:  "WHERE true",
+		},
+		{
+			// NOT IN also resolves __all__ to TrueConditionLiteral.
+			name:    "NOT IN allVariable alone",
+			expr:    "x NOT IN $service",
```
srikanthccv, line 1428: This is technically incorrect but that's the behaviour at the moment so keep it around

pkg/querybuilder/where_clause_visitor.go
tushar-signoz: Done

pkg/querybuilder/where_clause_visitor.go
tushar-signoz: Done

pkg/querybuilder/where_clause_visitor.go
tushar-signoz: Done

pkg/querybuilder/where_clause_visitor_test.go
tushar-signoz: Fixed

pkg/querybuilder/where_clause_visitor_test.go
tushar-signoz: Fixed

tushar-signoz:
Need to fix merge issues
Edit: Done

Issue #9434, opened 2025-10-29
Use single reducer for form initialisation.


srikanthccv:
There are number of places in our code where we make multiple reducer calls. However, all of the these states are conceptually one "form state", a single reducer is cleaner architecturally.

```js
	useEffect(() => {
		if (isEditMode && initialAlertState) {
			setAlertState({
				type: 'SET_INITIAL_STATE',
				payload: initialAlertState.basicAlertState,
			});
			setThresholdState({
				type: 'SET_INITIAL_STATE',
				payload: initialAlertState.thresholdState,
			});
			setEvaluationWindow({
				type: 'SET_INITIAL_STATE',
				payload: initialAlertState.evaluationWindowState,
			});
			setAdvancedOptions({
				type: 'SET_INITIAL_STATE',
				payload: initialAlertState.advancedOptionsState,
			});
			setNotificationSettings({
				type: 'SET_INITIAL_STATE',
				payload: initialAlertState.notificationSettingsState,
			});
		}
	}, [initialAlertState, isEditMode]);
```

PR #10179, opened 2026-02-04
chore: use single reducer in create alerts flow

amlannandy:
### 📄 Summary

Originally, the create alerts v2 flow was using 5 different reducers for managing the form state.
This was an unoptimal design practice as we should be using a single reducer for a single form.
In this PR, we are combining the multiple reducers into a single one.
A new unified reducer is created and it uses the old reducer logic within itself

https://github.com/user-attachments/assets/dc0a08ae-7cb2-4ce1-8097-c60a96109c8e

Closes https://github.com/SigNoz/signoz/issues/9434

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🧪 Testing Strategy

The reducer is used for managing the form state. So, the testing is mostly surrounded around entering various form values, saving the alert, and making sure the correct payload is being sent.

---

As we're touching the entire create flow state, we have to ensure the form state updates and API call payloads are being done properly

### 📋 Checklist
- [x] Manually tested

---

> [!NOTE]
> **Medium Risk**
> Touches core state management for the alert creation form, so regressions could surface as missed updates, incorrect resets/initialization, or wrong payload composition despite being a refactor.
> 
> **Overview**
> Refactors the CreateAlertV2 flow to manage the entire form via a **single `createAlertReducer`** and unified `CreateAlertState`, replacing the previous pattern of separate reducers per form section.
> 
> This introduces `INITIAL_CREATE_ALERT_STATE` plus new slice/action plumbing (`CreateAlertSlice`, `CreateAlertReducerAction`), updates the provider to dispatch slice-specific actions and handle global `RESET`/`SET_INITIAL_STATE`, and renames the alert-def-to-local-state mapping output shape from `basicAlertState`/`thresholdState`… to `basic`/`threshold`… (with corresponding test updates).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0fd9934edfc09b1325c49f36b057b2dcc9e1351e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/src/components/QueryBuilderV2/QueryV2/__tests__/QueryV2.test.tsx
+++ b/frontend/src/components/QueryBuilderV2/QueryV2/__tests__/QueryV2.test.tsx
@@ -11,9 +11,14 @@ import {
 	IBuilderQuery,
 	Query,
 } from 'types/api/queryBuilder/queryBuilderData';
+import { QueryFunction } from 'types/api/v5/queryRange';
 import { EQueryType } from 'types/common/dashboard';
 import { UseQueryOperations } from 'types/common/operations.types';
-import { DataSource, QueryBuilderContextType } from 'types/common/queryBuilder';
+import {
+	DataSource,
+	QueryBuilderContextType,
+	QueryFunctionsTypes,
+} from 'types/common/queryBuilder';
 
 import '@testing-library/jest-dom';
 
@@ -45,12 +50,17 @@ const mockedUseQueryOperations = jest.mocked(
 
 describe('QueryBuilderV2 + QueryV2 - base render', () => {
 	let handleRunQueryMock: jest.MockedFunction<() => void>;
+	let handleQueryFunctionsUpdatesMock: jest.MockedFunction<() => void>;
+	let baseQBContext: QueryBuilderContextType;
 
 	beforeEach(() => {
 		const mockCloneQuery = jest.fn() as jest.MockedFunction<
 			(type: string, q: IBuilderQuery) => void
 		>;
 		handleRunQueryMock = jest.fn() as jest.MockedFunction<() => void>;
+		handleQueryFunctionsUpdatesMock = jest.fn() as jest.MockedFunction<
+			() => void
+		>;
 		const baseQuery: IBuilderQuery = {
 			queryName: 'A',
 			dataSource: DataSource.LOGS,
@@ -91,7 +101,7 @@ describe('QueryBuilderV2 + QueryV2 - base render', () => {
 		const updateQueriesData: QueryBuilderContextType['updateQueriesData'] = (q) =>
 			q;
 
-		mockedUseQueryBuilder.mockReturnValue(({
+		const baseContext = ({
 			currentQuery: currentQueryObj,
 			stagedQuery: null,
 			lastUsedQuery: null,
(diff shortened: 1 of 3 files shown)


frontend/src/container/CreateAlertV2/context/utils.tsx
```diff
@@ -251,3 +255,55 @@ export const notificationSettingsReducer = (
 			return state;
 	}
 };
+
+export const createAlertReducer = (
```
amlannandy, line 259: For the reducer logic, I ended up using the old reducer logic like this. I thought it helps it be more organized and readable. Please let me know if there are any concerns regarding this.

frontend/src/container/CreateAlertV2/context/constants.ts
```diff
@@ -196,3 +196,11 @@ export const INITIAL_NOTIFICATION_SETTINGS_STATE: NotificationSettingsState = {
 	description: NOTIFICATION_MESSAGE_PLACEHOLDER,
 	routingPolicies: false,
 };
+
+export const INITIAL_CREATE_ALERT_STATE = {
```
AshwinBhatkal, line 200: This initial state is not specific to create alert right? Should be alerts only?

frontend/src/container/CreateAlertV2/context/utils.tsx
AshwinBhatkal: What does the older reducer logic mean?

frontend/src/container/CreateAlertV2/context/constants.ts
AshwinBhatkal: If it should be alerts only, then we need to see how we name few of the constants as well

frontend/src/container/CreateAlertV2/context/utils.tsx
amlannandy: I meant I'm using the old reducers as slices inside the new combined reducers

frontend/src/container/CreateAlertV2/context/constants.ts
amlannandy: It's for the create and edit flow

AshwinBhatkal: Yeah. So we should indicate that. How?
Maybe AlertForm

For usages of CreateAlert will become AlertForm

frontend/src/container/CreateAlertV2/context/utils.tsx
AshwinBhatkal: That should be okay

frontend/src/container/CreateAlertV2/context/constants.ts
AshwinBhatkal: Or AlertEditor

AshwinBhatkal: Or you could go
CreateAlert -> AlertForm
EditAlert -> AlertForm

AshwinBhatkal: Or you could go
CreateALert -> AlertForm
EditAlert -> AlertForm

CreateAlert and EditAlert takes care of initializing and possibly saving and AlertForm is same across once initialized

amlannandy: as dicussed, we'll do this in a separate PR: https://github.com/SigNoz/signoz/issues/10195

PR #10187, opened 2026-02-04
feat(authn): root user

karanbalani:
### 📄 Summary

Support for special root user per org, configurable through env variables.

#### Issues closed by this PR
Closes: https://github.com/SigNoz/platform-pod/issues/1076

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

> [!NOTE]
> **High Risk**
> High risk because it introduces a new privileged identity type, changes login/session and token identity resolution behavior, and adds a new SQL migration/table used during startup and org creation.
> 
> **Overview**
> Introduces a new **per-organization root user** concept backed by a new `root_users` table and a `rootuser` module (store + reconciler + authentication) that can create/update the root user from config and ensure it has admin privileges.
> 
> Wires root users into core flows: root user reconciliation now runs on startup and when an org is created, session context/login prefers root user password auth when the email matches, user APIs block inviting reserved root emails and allow `GetMyUser` to return a root-user-shaped response, and the tokenizer store can resolve identities/tokens for `root_users` as well as normal `users`. Also updates config (`user.root`) + example YAML and adjusts `errors.TypeUnsupported` HTTP mapping to `422`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0cf13848921c10f04b4e3d19cb136cb294d99a70. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


karanbalani:
Closes: https://github.com/SigNoz/platform-pod/issues/1076

pkg/modules/rootuser/implrootuser/reconciler.go
```diff
@@ -0,0 +1,154 @@
...
+	if err != nil {
+		return err
+	}
+
+	r.settings.Logger().InfoContext(ctx, "Reconciler: Root user created for organization", "organization_id", orgID, "email", r.config.Email)
+
+	err = r.granter.Grant(ctx, orgID, roletypes.SigNozAdminRoleName, authtypes.MustNewSubject(authtypes.TypeableUser, rootUser.ID.StringValue(), rootUser.OrgID, nil))
+	if err != nil {
+		return err
+	}
```
karanbalani: This is a problem here, `authz` is a service which is not initialized/started at this point, so this code basically fails miserably.

Now to solve this, we can convert the reconciler into a service, but again, this reconciler service will have a dependency on the `authz` service, and the way we start the services in SigNoz is in a parallel fashion, which could be a race condition and this code may fail again.

Another way to solve this could be, to have the reconciler run inside the `server.go` each in community and enterprise, after `signoz.Start(ctx)`, but that doesn't avoid the race condition either.

@therealpandey @vikrantgupta25 - any help here?

pkg/tokenizer/tokenizerstore/sqltokenizerstore/store.go
```diff
@@ -34,6 +35,7 @@ func (store *store) Create(ctx context.Context, token *authtypes.StorableToken)
 }
 
 func (store *store) GetIdentityByUserID(ctx context.Context, userID valuer.UUID) (*authtypes.Identity, error) {
```
karanbalani, line 37: @therealpandey - requesting you to have a look at this change, this is critical.

pkg/modules/session/implsession/module.go
```diff
@@ -108,6 +111,18 @@ func (module *module) GetSessionContext(ctx context.Context, email valuer.Email,
 }
 
 func (module *module) CreatePasswordAuthNSession(ctx context.Context, authNProvider authtypes.AuthNProvider, email valuer.Email, password string, orgID valuer.UUID) (*authtypes.Token, error) {
+	// Root User Authentication
```
karanbalani, line 127: @therealpandey - another critical section

pkg/modules/rootuser/implrootuser/reconciler.go
karanbalani: Quick update: @vikrantgupta25 is making major changes in the authz init setup, probably, then we can solve this one.

pkg/modules/session/implsession/module.go
```diff
@@ -60,6 +63,19 @@ func (module *module) GetSessionContext(ctx context.Context, email valuer.Email,
...
+	// if this email is a root user email, we will only allow password authentication
+	if module.rootUserModule != nil {
+		rootUserContexts, err := module.getRootUserSessionContext(ctx, orgs, orgIDs, email)
+		if err != nil {
+			return nil, err
+		}
+		if rootUserContexts.Exists {
+			return rootUserContexts, nil
+		}
+	}
```
karanbalani, line 76: @therealpandey - another critical section

Makefile
```diff
@@ -108,6 +108,8 @@ go-run-community: ## Runs the community go backend server
 	SIGNOZ_TELEMETRYSTORE_PROVIDER=clickhouse \
 	SIGNOZ_TELEMETRYSTORE_CLICKHOUSE_DSN=tcp://127.0.0.1:9000 \
 	SIGNOZ_TELEMETRYSTORE_CLICKHOUSE_CLUSTER=cluster \
+	SIGNOZ_USER_ROOT_EMAIL=root@example.com \
```
karanbalani: once, testing is done, I will remove this

pkg/modules/rootuser/implrootuser/reconciler.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return &reconciler{
+		store:     store,
+		settings:  scopedSettings,
+		orgGetter: orgGetter,
+		config:    config,
+	}
+}
+
+func (r *reconciler) Reconcile(ctx context.Context) error {
```
karanbalani, line 33: @therealpandey - this logic currently is not handling the granting of Admin role to the user, we have move that logic in the `Authenticate` function for now.

pkg/modules/rootuser/implrootuser/reconciler.go
karanbalani: for ref: https://github.com/SigNoz/signoz/pull/10187#discussion_r2773525406

karanbalani:
Closing as this was picked up by @therealpandey in https://github.com/SigNoz/signoz/pull/10313

PR #10191, opened 2026-02-05
perf(service-map): use react-force-graph-2d dep to reduce bundle size

H4ad:
## Pull Request

Be more specific with the dependency of force graph to reduce bundle size, the bundle size drops from 21.39Mb to 19.39Mb.

Before:

image

After:

image

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 📋 Checklist
- [x] Manually tested


frontend/package.json
```diff
@@ -130,7 +130,7 @@
 		"react-dom": "18.2.0",
 		"react-drag-listview": "2.0.0",
 		"react-error-boundary": "4.0.11",
-		"react-force-graph": "^1.43.0",
+		"react-force-graph-2d": "^1.29.1",
```
ishan621, line 134: non blocking - do we need ^ here ?

H4ad: To keep consistent on what we had earlier, since there's no consensus about ^ or fixed deps, I kept the same it was previously.

PR #10202, opened 2026-02-05
feat: add support for simultaneous delta and cumulative temporality

### LINKED ISSUE #6175: Support for simultaneous delta and cumulative temporality
The current implementation only supports one temporality at a time. A metric has to be either delta or cumulative. This makes it hard to switch from one to another. For instance, I want to recommend adding a cumulative delta pipeline for metrics for end users. This would be fine for new users, but existing users can't transition without breaking things. Make query-service support both in any combination, i.e. both temporalities are sent, cumulative in the past but delta now, the delta in the past but cumulative now.

namanv3:
## Pull Request
PR for Issue #6175. This PR adds the handling of multiple temporalities for a metric, be it across different services for that metric, or a service changing its temporality, or just in general there being different time series for the same metric with different temporalities.

Doc with the new query and the process behind building it: https://www.notion.so/signoz/Handling-multiple-temporalities-2fdfcc6bcd1980338daffb5d285920cb?source=copy_link

#### Issues closed by this PR
Closes #6175 

### 📋 Checklist
- [x] Manually tested

---

> [!NOTE]
> **Medium Risk**
> Changes core metrics query generation and temporality detection logic, which can affect correctness/performance of `rate`/`increase` results across time ranges; mitigated by added integration tests for mixed-temporality cases.
> 
> **Overview**
> Adds support for metrics whose underlying time series report **multiple temporalities** (delta and cumulative) within the queried time range.
> 
> The querier/metadata store temporality lookup now takes the query time range and returns a new `metrictypes.Multiple` sentinel when both temporalities are present; the ClickHouse fetch is updated to filter by time range/table selection and to collect distinct temporalities per metric.
> 
> The metrics statement builder skips temporality filtering when `Multiple` and introduces a dedicated temporal-aggregation path that computes `rate`/`increase`/other aggregations per-point by branching on the stored `temporality` value, plus adds integration coverage and fixtures for mixed-temporality scenarios and makes `temporality` optional in test query builders.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 91a2bedb475cf93eb98f52456fb23e25bddcd2d5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/src/container/ApiMonitoring/Explorer/Domains/DomainList.tsx
+++ b/frontend/src/container/ApiMonitoring/Explorer/Domains/DomainList.tsx
@@ -1,7 +1,7 @@
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useSelector } from 'react-redux';
 import { LoadingOutlined } from '@ant-design/icons';
-import { Spin, Table, Typography } from 'antd';
+import { Spin, Table } from 'antd';
 import logEvent from 'api/common/logEvent';
 import cx from 'classnames';
 import QuerySearch from 'components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch';
@@ -14,11 +14,13 @@ import { useQueryOperations } from 'hooks/queryBuilder/useQueryBuilderOperations
 import { useShareBuilderUrl } from 'hooks/queryBuilder/useShareBuilderUrl';
 import { useListOverview } from 'hooks/thirdPartyApis/useListOverview';
 import { get } from 'lodash-es';
+import { MoveUpRight } from 'lucide-react';
 import { AppState } from 'store/reducers';
 import { BaseAutocompleteData } from 'types/api/queryBuilder/queryAutocompleteResponse';
 import { HandleChangeQueryDataV5 } from 'types/common/operations.types';
 import { DataSource } from 'types/common/queryBuilder';
 import { GlobalReducer } from 'types/reducer/globalTime';
+import DOCLINKS from 'utils/docLinks';
 
 import { ApiMonitoringHardcodedAttributeKeys } from '../../constants';
 import { DEFAULT_PARAMS, useApiMonitoringParams } from '../../queryParams';
@@ -125,51 +127,67 @@ function DomainList(): JSX.Element {
 					hardcodedAttributeKeys={ApiMonitoringHardcodedAttributeKeys}
 				/>
 			</div>
-			<Table
-				className={cx('api-monitoring-domain-list-table')}
-				dataSource={isFetching || isLoading ? [] : formattedDataForTable}
-				columns={columnsConfig}
-				loading={{
-					spinning: isFetching || isLoading,
-					indicator: <Spin indicator={<LoadingOutlined size={14} spin />} />,
-				}}
-				locale={{
-					emptyText:
-						isFetching || isLoading ? null : (
-							<div className="no-filtered-domains-message-container">
-								<div className="no-filtered-domains-message-content">
-									<img
-										src="/Icons/emptyState.svg"
-										alt="thinking-emoji"
-										className="empty-state-svg"
-									/>
+			{!isFetching && !isLoading && formattedDataForTable.length === 0 && (
+				<div className="no-filtered-domains-message-container">
+					<div className="no-filtered-domains-message-content">
+						<img
+							src="/Icons/emptyState.svg"
+							alt="thinking-emoji"
+							className="empty-state-svg"
+						/>
 
-									<Typography.Text className="no-filtered-domains-message">
-										This query had no results. Edit your query and try again!
-									</Typography.Text>
-								</div>
+						<div className="no-filtered-domains-message">
+							<div className="no-domain-title">
+								No External API calls detected with applied filters.
 							</div>
-						),
-				}}
-				scroll={{ x: true }}
-				tableLayout="fixed"
-				onRow={(record, index): { onClick: () => void; className: string } => ({
-					onClick: (): void => {
-						if (index !== undefined) {
-							const dataIndex = formattedDataForTable.findIndex(
-								(item) => item.key === record.key,
-							);
-							setSelectedDomainIndex(dataIndex);
-							setParams({ selectedDomain: record.domainName });
-							logEvent('API Monitoring: Domain name row clicked', {});
-						}
-					},
-					className: 'expanded-clickable-row',
-				})}
-				rowClassName={(_, index): string =>
-					index % 2 === 0 ? 'table-row-dark' : 'table-row-light'
-				}
-			/>
+							<div className="no-domain-subtitle">
+								Ensure all HTTP client spans are being sent with kind as{' '}
+								<span className="attribute">Client</span> and url set in{' '}
+								<span className="attribute">url.full</span> or{' '}
+								<span className="attribute">http.url</span> attribute.
+							</div>
+							<a
+								href={DOCLINKS.EXTERNAL_API_MONITORING}
+								target="_blank"
+								rel="noreferrer"
+								className="external-api-doc-link"
+							>
+								Learn how External API monitoring works in SigNoz{' '}
+								<MoveUpRight size={14} />
+							</a>
+						</div>
+					</div>
+				</div>
+			)}
+			{(isFetching || isLoading || formattedDataForTable.length > 0) && (
+				<Table
+					className="api-monitoring-domain-list-table"
+					dataSource={isFetching || isLoading ? [] : formattedDataForTable}
+					columns={columnsConfig}
+					loading={{
+						spinning: isFetching || isLoading,
+						indicator: <Spin indicator={<LoadingOutlined size={14} spin />} />,
+					}}
+					scroll={{ x: true }}
+					tableLayout="fixed"
+					onRow={(record, index): { onClick: () => void; className: string } => ({
+						onClick: (): void => {
+							if (index !== undefined) {
+								const dataIndex = formattedDataForTable.findIndex(
+									(item) => item.key === record.key,
+								);
+								setSelectedDomainIndex(dataIndex);
+								setParams({ selectedDomain: record.domainName });
+								logEvent('API Monitoring: Domain name row clicked', {});
+							}
+						},
+						className: 'expanded-clickable-row',
+					})}
+					rowClassName={(_, index): string =>
+						index % 2 === 0 ? 'table-row-dark' : 'table-row-light'
+					}
+				/>
+			)}
 			{selectedDomainIndex !== -1 && (
 				<DomainDetails
 					domainData={formattedDataForTable[selectedDomainIndex]}
--- a/frontend/src/container/ApiMonitoring/Explorer/Explorer.styles.scss
+++ b/frontend/src/container/ApiMonitoring/Explorer/Explorer.styles.scss
@@ -180,10 +180,59 @@
 
 	.no-filtered-domains-message {
 		margin-top: 8px;
+		display: flex;
+		gap: 8px;
+		flex-direction: column;
+
+		.no-domain-title {
+			color: var(--bg-vanilla-100, #fff);
+			font-family: Inter;
+			font-size: 14px;
+			font-style: normal;
+			font-weight: 500;
+			line-height: 20px; /* 142.857% */
+		}
+
+		.no-domain-subtitle {
+			color: var(--bg-vanilla-400, #c0c1c3);
+			font-family: Inter;
+			font-size: 14px;
+			font-style: normal;
+			font-weight: 400;
+			line-height: 20px; /* 142.857% */
+
+			.attribute {
+				font-family: 'Space Mono';
+			}
+		}
+
+		.external-api-doc-link {
+			display: flex;
+			flex-direction: row;
+			gap: 4px;
+			align-items: center;
+		}
 	}
 }
 
 .lightMode {
+	.no-filtered-domains-message-container {
+		.no-filtered-domains-message-content {
+			.no-filtered-domains-message {
+				.no-domain-title {
+					color: var(--text-ink-500);
+				}
+
+				.no-domain-subtitle {
+					color: var(--text-ink-400);
+
+					.attribute {
+						font-family: 'Space Mono';
+					}
+				}
+			}
+		}
+	}
 	.api-monitoring-domain-list-table {
 		.ant-table {
 			.ant-table-thead > tr > th {
--- a/frontend/src/utils/docLinks.ts
+++ b/frontend/src/utils/docLinks.ts
@@ -6,6 +6,8 @@ const DOCLINKS = {
 		'https://signoz.io/docs/product-features/trace-explorer/?utm_source=product&utm_medium=traces-explorer-trace-tab#traces-view',
 	METRICS_EXPLORER_EMPTY_STATE:
 		'https://signoz.io/docs/userguide/send-metrics-cloud/',
+	EXTERNAL_API_MONITORING:
+		'https://signoz.io/docs/external-api-monitoring/overview/',
 };
 
 export default DOCLINKS;


pkg/querier/builder_query.go
```diff
@@ -80,9 +80,16 @@ func (q *builderQuery[T]) Fingerprint() string {
...
+				if a.IsMultiTemporality {
+					temporalityStringVal = ""
+					for _, temporality := range a.TemporalityList {
+						temporalityStringVal += temporality.StringValue()
+					}
+				}
 				aggParts = append(aggParts, fmt.Sprintf("%s:%s:%s:%s",
 					a.MetricName,
-					a.Temporality.StringValue(),
+					temporalityStringVal, // query fingerprint I'm guessing is not the same as a time series' fingerprint
```
srikanthccv: yes, they are difference, query fingerprint is used a unique cache key.

pkg/querier/builder_query.go
```diff
@@ -80,9 +80,16 @@ func (q *builderQuery[T]) Fingerprint() string {
 			case qbtypes.LogAggregation:
 				aggParts = append(aggParts, a.Expression)
 			case qbtypes.MetricAggregation:
+				temporalityStringVal := a.Temporality.StringValue()
+				if a.IsMultiTemporality {
+					temporalityStringVal = ""
+					for _, temporality := range a.TemporalityList {
+						temporalityStringVal += temporality.StringValue()
+					}
```
srikanthccv: Have them joined with comma separated.

pkg/telemetrymetadata/metadata.go
```diff
@@ -1597,36 +1597,36 @@ func (t *telemetryMetaStore) GetAllValues(ctx context.Context, fieldValueSelecto
...
 	if metricName == "" {
-		return metrictypes.Unknown, errors.Newf(errors.TypeInternal, errors.CodeInternal, "metric name cannot be empty")
+		return []metrictypes.Temporality{metrictypes.Unknown}, errors.Newf(errors.TypeInternal, errors.CodeInternal, "metric name cannot be empty")
 	}
 
-	temporalityMap, err := t.FetchTemporalityMulti(ctx, metricName)
+	temporalityMap, err := t.FetchTemporalityMulti(ctx, queryTimeRangeStartTs, queryTimeRangeEndTs, metricName)
 	if err != nil {
-		return metrictypes.Unknown, err
+		return []metrictypes.Temporality{metrictypes.Unknown}, err
```
srikanthccv: Since the return type now a slice, the return value here can `nil, err` instead of `[]metrictypes.Temporality{metrictypes.Unknown}, err`

pkg/telemetrymetadata/metadata.go
```diff
@@ -1638,29 +1638,38 @@ func (t *telemetryMetaStore) FetchTemporalityMulti(ctx context.Context, metricNa
...
 	// Filter by metric names (in the temporality column due to data mix-up)
-	sb.Where(sb.In("metric_name", metricNames))
+	sb.Where(
+		sb.In("metric_name", metricNames),
+		sb.GTE("last_reported_unix_milli", queryTimeRangeStartTs),
+		sb.LT("last_reported_unix_milli", queryTimeRangeEndTs),
+	)
+
+	// group by last_reported_unix_milli to be able to sort by it
+	sb.GroupBy("metric_name", "temporality", "last_reported_unix_milli")
```
srikanthccv: Do we really need `last_reported_unix_milli` in group by? How does it work when there are 10 different values for `last_reported_unix_milli`? Will there be duplicate values in the result list?

pkg/telemetrymetadata/metadata.go
```diff
@@ -1693,25 +1702,34 @@ func (t *telemetryMetaStore) fetchMetricsTemporality(ctx context.Context, metric
...
 
-	// Group by metric name to get one temporality per metric
-	sb.GroupBy("metric_name")
+	// to get the latest seen temporality in zeroth index
+	sb.Desc().OrderBy("last_reported_unix_milli")
```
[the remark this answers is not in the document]
srikanthccv: Valid

pkg/telemetrymeter/statement_builder.go
```diff
@@ -42,6 +42,7 @@ func NewMeterQueryStatementBuilder(
 	}
 }
 
+// todo NAMAN: add handling for the scenario where there are multiple temporalities
```
srikanthccv: The meter metrics are generated by and will remain always delta, you can remove the todo.

pkg/types/querybuildertypes/querybuildertypesv5/builder_elements.go
```diff
@@ -381,8 +381,13 @@ type MetricAggregation struct {
...
 	// type of the metric
 	Type metrictypes.Type `json:"-"`
-	// temporality to apply to the query
+	// if false (default), Temporality will be used in the query, otherwise TemporalityList will be used
+	IsMultiTemporality bool `json:"-"`
+	// temporality to apply to the query if there is only one temporality to consider
+	// contains the latest seen temporality if there are multiple temporalities
 	Temporality metrictypes.Temporality `json:"temporality"`
+	// temporalities to apply to the query if there are multiple temporalities to consider
+	TemporalityList []metrictypes.Temporality `json:"temporality_list"`
```
srikanthccv: I thought about this, let's not introduce the two fields `IsMultiTemporality` and `TemporalityList` here. Instead, let's introduce a new virtual temporality with value `__all__`

```diff
diff --git a/pkg/types/metrictypes/metrictypes.go b/pkg/types/metrictypes/metrictypes.go
index 9a65031b8..f2950a706 100644
--- a/pkg/types/metrictypes/metrictypes.go
+++ b/pkg/types/metrictypes/metrictypes.go
@@ -19,6 +19,7 @@ var (
        Cumulative  = Temporality{valuer.NewString("cumulative")}
        Unspecified = Temporality{valuer.NewString("unspecified")}
        Unknown     = Temporality{valuer.NewString("")}
+       All         = Temporality{valuer.NewString("__all__")}
 )
```

And then let the `FetchTemporalityMulti` return `All` for a given metric where there is more than one active temporality and update the statement builder to decide how to prepare the query.

Why prefer this over the `IsMultiTemporality` and `TemporalityList`?
1. When user is querying the data, they don't really care about underlying temporality. The temporality is something they pay attention to at the time of setup, if they pay any attention at all. So what they are interested in the end result that accurately shows the data for them.
2. We can already see this detail spilling in the codebase. We find ourselves writing if checks and do something with the list. This comes with cognitive load when reading the code for no benefit.
3. When querying data, it's either a temporality is specified or nothing specified. So when it's specified, it will work as expected. When it's not specified the implication that the query should work with all temporalities. I don't see any legitimate case where one wants to specify the "list" of temporalities to consider. So for this reason, either it's `__all__` to indicate run on all or a specific one.


Let me know what you think.

pkg/telemetrymeter/statement_builder.go
namanv3: Then why do we even need `fetchMeterSourceMetricsTemporality`?

pkg/telemetrymetadata/metadata.go
namanv3: The group by is needed to ensure that the latest one comes first in the list, no other reason. I'm checking if your __All__ suggestion breaks anything or not. If it doesn't, then we won't need sorting anyway.

namanv3: And it will handle duplicates and return only once cuz the select distinct is on metric_name, temporality. Without that distinct the duplicates were also coming

pkg/types/querybuildertypes/querybuildertypesv5/builder_elements.go
namanv3: There are two places where the actual values are needed:
1. When calculating the cache fingerprint. Delta->Cumulative and Cumulative->Delta will end up as the same string in the fingerprint. I don't think this should be an issue
2. When checking if the temporality is unspecified during query validation (https://github.com/SigNoz/signoz/pull/10202/changes#diff-b385c29515dd9ae5c168c28b9b9e60e67a1eba321962e507d3e1ae0321fc028bR817). If the temporality is Unspecified, it is a gauge metric anyway, so it shouldn't be an issue per se, but this part of the code isn't reached via metrics explorer, so I'm not so sure if someone can somehow build an invalid query metric here

srikanthccv: #point2, it won't be coming from our frontend, but potentially from API requests, with `__all__` nothing needs to change, the original `if agg.Temporality == metrictypes.Unspecified {` remains as is

#point1, > Delta->Cumulative and Cumulative->Delta will end up as the same string in the fingerprint

I didn't understand this part, Please elaborate

namanv3: If a metric changes from delta to cumulative, we get a cache fingerprint for it with __All__. Now if it again shifts back to Delta, the cache fingerprint is still with __All__. However, it will probably not clash as the time range will be different anyway.

I think it should be alright to go ahead with __All__

namanv3: @srikanthccv let me know if I should go ahead with this suggestion then

srikanthccv: Sure, please go ahead and let's add integration tests for these cases.

pkg/telemetrymeter/statement_builder.go
srikanthccv: No good reason, other than to maintain the shape and implementation of the meter statement builder same as the metric statement builder. Either we can remove the todo code or implement the same in meter statement builder

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -528,6 +534,58 @@ func (b *MetricQueryStatementBuilder) buildTemporalAggCumulativeOrUnspecified(
...
+	sb.JoinWithOption(sqlbuilder.InnerJoin, timeSeriesCTE, "points.fingerprint = filtered_time_series.fingerprint")
+	sb.Where(
+		sb.In("metric_name", query.Aggregations[0].MetricName),
+		sb.GTE("unix_milli", start),
+		sb.LT("unix_milli", end),
+	)
+	sb.GroupBy("fingerprint", "ts", "temporality")
+	sb.GroupBy(querybuilder.GroupByKeys(query.GroupBy)...)
+	queryWithoutWindow, args := sb.BuildWithFlavor(sqlbuilder.ClickHouse, timeSeriesCTEArgs...)
+	queryWithWindowAndOrder := queryWithoutWindow + " WINDOW rate_window AS (PARTITION BY fingerprint ORDER BY fingerprint ASC, ts ASC) ORDER BY ts"
```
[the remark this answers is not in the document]
namanv3: Same fingerprint cannot have delta and cumulative both and fingerprint includes the temporality in its calculation. Resolving this comment

pkg/telemetrymeter/statement_builder.go
namanv3: removed todo

tests/integration/src/querier/08_metrics_increase_multi_temporality.py
```diff
@@ -0,0 +1,277 @@
...
+    # most values should be 0 (flat periods between increments)
+    assert (
+        count_zero_users >= 40
+    ), f"Expected >= 40 zero rate values for /users (sparse data), got {count_zero_users}"
+    # non-zero values should be 0.0167 (1/60 increment rate)
+    non_zero_users = [v["value"] for v in users_values if v["value"] != 0]
+    count_increment_rate = sum(1 for v in non_zero_users if v == 0.0167)
+    assert (
+        count_increment_rate >= 8
+    ), f"Expected >= 8 increment rate values (0.0167) for /users, got {count_increment_rate}"
```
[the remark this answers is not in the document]
namanv3: Fixed

pkg/types/metrictypes/metrictypes.go
```diff
@@ -19,6 +19,7 @@ var (
 	Cumulative  = Temporality{valuer.NewString("cumulative")}
 	Unspecified = Temporality{valuer.NewString("unspecified")}
 	Unknown     = Temporality{valuer.NewString("")}
+	Multiple    = Temporality{valuer.NewString("__multiple__")}
```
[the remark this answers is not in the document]
namanv3: Multiple is not supposed to be passed as a parameter, its a catch all, so it does not need a value

srikanthccv:
@vikrantgupta25 please review!

tests/integration/src/querier/08_metrics_increase_multi_temporality.py
```diff
@@ -0,0 +1,271 @@
...
+from fixtures.metrics import Metrics
+from fixtures.querier import (
+    build_builder_query,
+    get_all_series,
+    get_series_values,
+    make_query_request,
+)
+
+TESTDATA_DIR = os.path.join(os.path.dirname(__file__), "..", "..", "testdata")
+MULTI_TEMPORALITY_FILE = os.path.join(TESTDATA_DIR, "multi_temporality_counters_1h.jsonl")
```
vikrantgupta25: please use util from `utils.py` to achieve this ?

vikrantgupta25 (changes requested):
it seems like 07 / 08 have same setup, we can paramterize the test and have utils for comparison ? basically make these DRY if possible.

tests/integration/src/querier/08_metrics_increase_multi_temporality.py
namanv3: Done, also parametrised the tests

tests/integration/src/querier/07_metrics_multi_temporality.py
```diff
@@ -0,0 +1,333 @@
...
+    orders_values = endpoint_values["/orders"]
+    assert (
+        len(orders_values) >= 58
+    ), f"Expected >= 58 values for /orders, got {len(orders_values)}"
+    count_steady_orders = sum(1 for v in orders_values if v["value"] == stable_orders_value)
+    assert (
+        count_steady_orders >= 55
+    ), f"Expected >= 55 steady {time_aggregation} values (0.0833) for /orders, got {count_steady_orders}"
+    # check for counter reset effects - there should be some non-standard values
+    non_standard_orders = [v["value"] for v in orders_values if v["value"] != 0.0833]
```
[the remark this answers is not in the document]
namanv3: Fixed

srikanthccv (commented):
https://github.com/SigNoz/signoz/pull/10202/commits/b68219b8bc85c3a2d61f8e74f993f479424e3791

first `tf`, , , , , , , , , [t1, t2], , , , , , last `tl`

does this work if user is querying for the range `[t1, t2]` for the above timeline?

namanv3:
seems like we will need to use time series table for custom time ranges

tests/integration/src/querier/07_metrics_multi_temporality.py
```diff
@@ -0,0 +1,396 @@
...
+    ), f"Expected >= 53 steady {time_aggregation} values ({stable_checkout_value}) for /checkout, got {count_steady_checkout}"
+    # check that spike values exist (traffic spike +50/min at t40-t44)
+    count_spike_checkout = sum(1 for v in checkout_values if v["value"] == spike_checkout_value)
+    assert (
+        count_spike_checkout >= 4
+    ), f"Expected >= 4 spike {time_aggregation} values ({spike_checkout_value}) for /checkout, got {count_spike_checkout}"
+
+    # spike values should be consecutive
+    spike_indices = [
+        i for i, v in enumerate[Any](checkout_values) if v["value"] == spike_checkout_value
```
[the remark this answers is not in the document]
namanv3: Fixed

tests/integration/testdata/multi_temporality_counters_24h.jsonl
```diff
@@ -0,0 +1,196 @@
...
+{"metric_name":"http.request.count","labels":{"__temporality__":"Cumulative","service":"web","endpoint":"/products","status_code":"200"},"timestamp":"2025-01-10T23:00:00+00:00","value":470,"temporality":"Cumulative","type_":"Sum","is_monotonic":true,"flags":0,"description":"","unit":"","env":"default","resource_attrs":{},"scope_attrs":{}}
+{"metric_name":"http.request.count","labels":{"__temporality__":"Cumulative","service":"web","endpoint":"/products","status_code":"200"},"timestamp":"2025-01-10T23:30:00+00:00","value":480,"temporality":"Cumulative","type_":"Sum","is_monotonic":true,"flags":0,"description":"","unit":"","env":"default","resource_attrs":{},"scope_attrs":{}}
+{"metric_name":"http.request.count","labels":{"__temporality__":"Cumulative","service":"web","endpoint":"/products","status_code":"200"},"timestamp":"2025-01-11T24:00:00+00:00","value":490,"temporality":"Cumulative","type_":"Sum","is_monotonic":true,"flags":0,"description":"","unit":"","env":"default","resource_attrs":{},"scope_attrs":{}}
```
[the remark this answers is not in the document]
namanv3: Fixed

tests/integration/src/querier/07_metrics_multi_temporality.py
```diff
@@ -0,0 +1,307 @@
...
```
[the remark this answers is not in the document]
namanv3: Fixed

pkg/telemetrymetadata/metadata.go
```diff
@@ -1644,23 +1652,29 @@ func (t *telemetryMetaStore) FetchTemporalityMulti(ctx context.Context, metricNa
...
+		"temporality",
+	).
+		From(t.metricsDBName + "." + tsTableName)
 
 	// Filter by metric names (in the temporality column due to data mix-up)
-	sb.Where(sb.In("metric_name", metricNames))
+	sb.Where(
+		sb.In("metric_name", metricNames),
+		sb.GTE("unix_milli", queryTimeRangeStartTs),
+		sb.LT("unix_milli", queryTimeRangeEndTs),
```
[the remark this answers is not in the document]
namanv3: Fixed

pkg/telemetrymetadata/metadata.go
```diff
@@ -1644,23 +1652,29 @@ func (t *telemetryMetaStore) FetchTemporalityMulti(ctx context.Context, metricNa
 	return result, nil
 }
 
-func (t *telemetryMetaStore) fetchMetricsTemporality(ctx context.Context, metricNames ...string) (map[string]metrictypes.Temporality, error) {
-	result := make(map[string]metrictypes.Temporality)
+func (t *telemetryMetaStore) fetchMetricsTemporality(ctx context.Context, queryTimeRangeStartTs, queryTimeRangeEndTs uint64, metricNames ...string) (map[string][]metrictypes.Temporality, error) {
+	result := make(map[string][]metrictypes.Temporality)
+
+	_, _, _, tsTableName := telemetrymetrics.WhichTSTableToUse(queryTimeRangeStartTs, queryTimeRangeEndTs, nil)
```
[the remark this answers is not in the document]
namanv3: Done

pkg/telemetrymetadata/metadata.go
```diff
@@ -1693,25 +1707,30 @@ func (t *telemetryMetaStore) fetchMetricsTemporality(ctx context.Context, metric
...
 
 	// Filter by metric names (in the temporality column due to data mix-up)
-	sb.Where(sb.In("metric_name", metricNames))
+	sb.Where(
+		sb.In("metric_name", metricNames),
+		sb.GTE("unix_milli", queryTimeRangeStartTs),
+		sb.LT("unix_milli", queryTimeRangeEndTs),
+	)
 
-	// Group by metric name to get one temporality per metric
```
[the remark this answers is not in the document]
namanv3: Fixed

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -528,6 +534,57 @@ func (b *MetricQueryStatementBuilder) buildTemporalAggCumulativeOrUnspecified(
...
+	case metrictypes.TimeAggregationRate:
+		rateExpr := fmt.Sprintf(RateWithoutNegativeMultiTemporality, aggForDeltaTemporality, aggForCumulativeTemporality, aggForCumulativeTemporality, aggForCumulativeTemporality, start, aggForCumulativeTemporality, aggForCumulativeTemporality, start)
+		sb.SelectMore(rateExpr)
+	case metrictypes.TimeAggregationIncrease:
+		increaseExpr := fmt.Sprintf(IncreaseWithoutNegativeMultiTemporality, aggForDeltaTemporality, aggForCumulativeTemporality, aggForCumulativeTemporality, aggForCumulativeTemporality, aggForCumulativeTemporality, aggForCumulativeTemporality, start, start)
+		sb.SelectMore(increaseExpr)
+	default:
+		expr := fmt.Sprintf(OthersMultiTemporality, aggForDeltaTemporality, aggForCumulativeTemporality)
+		sb.SelectMore(expr)
+	}
```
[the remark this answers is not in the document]
namanv3: This is only for SUM metrics so the temporality can be either Delta or Cumulative.

pkg/telemetrymetadata/metadata.go
```diff
@@ -1693,25 +1707,33 @@ func (t *telemetryMetaStore) fetchMetricsTemporality(ctx context.Context, metric
 			temporality = metrictypes.Unknown
 		}
 
-		result[metricName] = temporality
+		result[metricName] = append(result[metricName], temporality)
```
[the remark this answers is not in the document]
namanv3: @srikanthccv can clickhouse have Unknown temporality? I don't see it on staging at least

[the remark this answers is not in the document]
namanv3: Done anyway

pkg/telemetrymetadata/metadata.go
```diff
@@ -1693,25 +1707,33 @@ func (t *telemetryMetaStore) fetchMetricsTemporality(ctx context.Context, metric
...
+		Distinct().
+		From(t.meterDBName + "." + t.meterFieldsTblName)
 
 	// Filter by metric names (in the temporality column due to data mix-up)
-	sb.Where(sb.In("metric_name", metricNames))
+	sb.Where(
+		sb.In("metric_name", metricNames),
+		sb.GTE("unix_milli", queryTimeRangeStartTs),
+		sb.LT("unix_milli", queryTimeRangeEndTs),
+	)
```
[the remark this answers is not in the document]
namanv3: Done

pkg/telemetrymetadata/metadata.go
```diff
@@ -1693,25 +1707,33 @@ func (t *telemetryMetaStore) fetchMetricsTemporality(ctx context.Context, metric
...
 
-	// Group by metric name to get one temporality per metric
-	sb.GroupBy("metric_name")
+	sb.GroupBy("metric_name", "temporality")
```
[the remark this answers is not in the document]
namanv3: Done

pkg/telemetrymeter/tables.go
```diff
@@ -57,6 +59,13 @@ func WhichSamplesTableToUse(
 
 }
 
+func AdjustStartTsForTSTable(start uint64, table string) uint64 {
```
vikrantgupta25:
```suggestion
func AdjustStartTsForSamplesTable(start uint64, table string) uint64 {
```

vikrantgupta25: there is no timeseries table for meter metrics

namanv3: Good catch thanks!

pkg/telemetrymetadata/metadata.go
```diff
@@ -1743,8 +1767,12 @@ func (t *telemetryMetaStore) fetchMeterSourceMetricsTemporality(ctx context.Cont
...
 			temporality = metrictypes.Unknown
 		}
-
-		result[metricName] = temporality
+		if temporality != metrictypes.Unknown {
+			result[metricName] = append(result[metricName], temporality)
+		}
+	}
+	if err := rows.Err(); err != nil {
+		return nil, errors.Wrapf(err, errors.TypeInternal, errors.CodeInternal, "error iterating over meter temporality rows")
```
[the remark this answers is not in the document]
namanv3: @srikanthccv mentioned that all meter data is delta so this cannot happen

pkg/telemetrymetadata/metadata.go
```diff
@@ -1692,26 +1707,35 @@ func (t *telemetryMetaStore) fetchMetricsTemporality(ctx context.Context, metric
...
 
 	sb := sqlbuilder.Select(
 		"metric_name",
-		"argMax(temporality, unix_milli) as temporality",
-	).From(t.meterDBName + "." + t.meterFieldsTblName)
+		"temporality",
+	).
+		From(t.meterDBName + "." + t.meterFieldsTblName)
 
+	adjustedStartTs := telemetrymeter.AdjustStartTsForSamplesTable(queryTimeRangeStartTs, t.meterFieldsTblName)
```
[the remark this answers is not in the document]
namanv3: Fields table means samples table, nothing wrong here

pkg/telemetrymetadata/metadata.go
```diff
@@ -1644,23 +1653,29 @@ func (t *telemetryMetaStore) FetchTemporalityMulti(ctx context.Context, metricNa
...
-	sb.Where(sb.In("metric_name", metricNames))
+	sb.Where(
+		sb.In("metric_name", metricNames),
+		sb.GTE("unix_milli", adjustedStartTs),
+		sb.LT("unix_milli", adjustedEndTs),
+	)
 
-	// Group by metric name to get one temporality per metric
-	sb.GroupBy("metric_name")
+	sb.GroupBy("metric_name", "temporality")
```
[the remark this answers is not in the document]
namanv3: The normalized form of a non-normalized data point would not have a different temporality anyway, so this is not a concern

pkg/telemetrymetrics/statement_builder.go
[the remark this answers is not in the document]
namanv3: Unspecified cannot come cuz this is only for counter metrics

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -528,6 +534,57 @@ func (b *MetricQueryStatementBuilder) buildTemporalAggCumulativeOrUnspecified(
...
+	sb.Where(
+		sb.In("metric_name", query.Aggregations[0].MetricName),
+		sb.GTE("unix_milli", start),
+		sb.LT("unix_milli", end),
+	)
+	sb.GroupBy("fingerprint", "ts", "temporality")
+	sb.GroupBy(querybuilder.GroupByKeys(query.GroupBy)...)
+	queryWithoutWindow, args := sb.BuildWithFlavor(sqlbuilder.ClickHouse, timeSeriesCTEArgs...)
+	queryWithWindowAndOrder := queryWithoutWindow + " WINDOW rate_window AS (PARTITION BY fingerprint ORDER BY fingerprint ASC, ts ASC) ORDER BY ts"
+	return fmt.Sprintf("__temporal_aggregation_cte AS (%s)", queryWithWindowAndOrder), args, nil
```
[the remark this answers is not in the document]
namanv3: This feature is not used so this is fine

pkg/telemetrymetadata/metadata.go
```diff
@@ -1630,8 +1630,12 @@ func (t *telemetryMetaStore) FetchTemporalityMulti(ctx context.Context, metricNa
...
 	// For metrics not found in the database, set to Unknown
 	for _, metricName := range metricNames {
-		if temporality, exists := metricsTemporality[metricName]; exists {
-			result[metricName] = temporality
+		if temporality, exists := metricsTemporality[metricName]; exists && len(temporality) > 0 {
+			if len(temporality) > 1 {
+				result[metricName] = metrictypes.Multiple
+			} else {
+				result[metricName] = temporality[0]
+			}
```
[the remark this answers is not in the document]
namanv3: A metric cannot have Unspecified along with Delta or Cumulative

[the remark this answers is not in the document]
srikanthccv: Anything could be possible in the wild, but we would call it bad data and try to get it fixed at the source rather than attempt to solve it in the query, especially for counter metrics.

PromQL is "untyped" in the sense that it doesn't care about the metric type and temporality. It will gladly run any function on any data even if they don't make sense, and puts the onus on the users to take care of these details.

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -21,6 +21,10 @@ const (
 	RateWithoutNegative     = `If((per_series_value - lagInFrame(per_series_value, 1, 0) OVER rate_window) < 0, per_series_value / (ts - lagInFrame(ts, 1, toDateTime(fromUnixTimestamp64Milli(%d))) OVER rate_window), (per_series_value - lagInFrame(per_series_value, 1, 0) OVER rate_window) / (ts - lagInFrame(ts, 1, toDateTime(fromUnixTimestamp64Milli(%d))) OVER rate_window))`
 	IncreaseWithoutNegative = `If((per_series_value - lagInFrame(per_series_value, 1, 0) OVER rate_window) < 0, per_series_value, ((per_series_value - lagInFrame(per_series_value, 1, 0) OVER rate_window) / (ts - lagInFrame(ts, 1, toDateTime(fromUnixTimestamp64Milli(%d))) OVER rate_window)) * (ts - lagInFrame(ts, 1, toDateTime(fromUnixTimestamp64Milli(%d))) OVER rate_window))`
 
+	RateWithoutNegativeMultiTemporality     = `IF(LOWER(temporality) LIKE LOWER('delta'), %s, IF((%s - lagInFrame(%s, 1, 0) OVER rate_window) < 0, %s / (ts - lagInFrame(ts, 1, toDateTime(fromUnixTimestamp64Milli(%d))) OVER rate_window), (%s - lagInFrame(%s, 1, 0) OVER rate_window) / (ts - lagInFrame(ts, 1, toDateTime(fromUnixTimestamp64Milli(%d))) OVER rate_window))) AS per_series_value`
```
srikanthccv, line 24: just double checking the `max(value) - lagInFrame(max(value), 1, 0)` does the same thing as sub-query `max_val - lagInFrame(max_val, 1, 0) FROM (SELECT max(value) as max_value ....)`. The former is bit hard to reason about.

```sql
SELECT
    ts,
    max(value) as max_val,
    lagInFrame(max(value), 1, 0) OVER (ORDER BY ts) as prev_max_val,
    max(value) - lagInFrame(max(value), 1, 0) OVER (ORDER BY ts) as diff
FROM
(
    SELECT 1 AS ts, 10 AS value
    UNION ALL SELECT 1, 8       -- dup, max picks 10
    UNION ALL SELECT 2, 30
    UNION ALL SELECT 2, 25      -- dup, max picks 30
    UNION ALL SELECT 3, 60
    UNION ALL SELECT 4, 5       -- restart! value dropped
    UNION ALL SELECT 5, 20
    UNION ALL SELECT 5, 18      -- dup, max picks 20
    UNION ALL SELECT 6, 50
)
GROUP BY ts
ORDER BY ts;
```

tests/integration/src/querier/07_metrics_multi_temporality.py
```diff
@@ -0,0 +1,396 @@
...
+    expected_value: float,
+) -> None:
+    now = datetime.now(tz=timezone.utc).replace(second=0, microsecond=0)
+    start_ms = int((now - timedelta(days=31)).timestamp() * 1000)
+    end_ms = int(now.timestamp() * 1000)
+    metric_name = f"test_{time_aggregation}_" + hex(random.getrandbits(32))[2:]
+
+    metrics = Metrics.load_from_file(
+        MULTI_TEMPORALITY_FILE_24h,
+        base_time=now - timedelta(minutes=600),
```
srikanthccv: Since this is only shifting the base by 10 hours, the delta values are pushed to the future timestamp and never take part in the query.

```
unix_milli >= x and unix_milli < y
```

Where `y` is the now in millis but all the delta values unix_millis are in future. Even some of the cumulative values are also pushed to the future.

tests/integration/src/querier/07_metrics_multi_temporality.py
```diff
@@ -0,0 +1,396 @@
...
+    count_increment_rate = sum(1 for v in non_zero_users if v == spike_users_value)
+    assert (
+        count_increment_rate >= 8
+    ), f"Expected >= 8 increment {time_aggregation} values ({spike_users_value}) for /users, got {count_increment_rate}"
+
+@pytest.mark.parametrize(
+    "time_aggregation, expected_value_at_30th_minute, expected_value_at_31st_minute, steady_value",
+    [
+        ("rate", 0.183, 0.183, 0.25),
+        ("increase", 11, 12, 15),
```
srikanthccv: The `steady_value` is a bit misleading, for the `api` service in multi temporality file, there is no steady value. While it is mostly occured value, there is an increase in `1` for every five minutes and a value with a jump.

The `/orders` endpoint of api has a `steady_value` that the whole `api` service doesn't have. I felt confused looking at the data where did the increase in `1` and other jump go.

srikanthccv (commented):
Thanks, overall LGTM, some final comments

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -528,6 +534,57 @@ func (b *MetricQueryStatementBuilder) buildTemporalAggCumulativeOrUnspecified(
...
+		sb.In("metric_name", query.Aggregations[0].MetricName),
+		sb.GTE("unix_milli", start),
+		sb.LT("unix_milli", end),
+	)
+	sb.GroupBy("fingerprint", "ts", "temporality")
+	sb.GroupBy(querybuilder.GroupByKeys(query.GroupBy)...)
+	queryWithoutWindow, args := sb.BuildWithFlavor(sqlbuilder.ClickHouse, timeSeriesCTEArgs...)
+	queryWithWindowAndOrder := queryWithoutWindow + " WINDOW rate_window AS (PARTITION BY fingerprint ORDER BY fingerprint ASC, ts ASC) ORDER BY ts"
+	return fmt.Sprintf("__temporal_aggregation_cte AS (%s)", queryWithWindowAndOrder), args, nil
+}
```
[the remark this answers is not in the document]
namanv3: The flag is never turned on, it isn't production ready. Maybe I should add this as a code comment.

tests/integration/src/querier/07_metrics_multi_temporality.py
namanv3: changed the var name

tests/integration/src/querier/07_metrics_multi_temporality.py
namanv3: changed it to 1441 and tested again, all good.

srikanthccv (approved):
thanks!

vikrantgupta25 (approved):

PR #10225, opened 2026-02-07
perf(grid-card): set cache time zero when auto-refresh is enabled

### LINKED ISSUE #2169: Chrome tab OOMs if dashboard page is kept open for long
Steps to reproduce:
1. Choose a dashboard with many panels
2. Set auto-refresh to 30s
3. Open the memory tab in chrome dev tools
4. Check memory slowly increasing to 300MBs in 15-20 mins

H4ad:
## Pull Request

This issue is caused by the react-query storing all the queries without releasing it fast enough.

This version of react-query is old, in the newer versions, we could use other kind of persistent implementations to allow have the cache without too much risk of overflow the memory.

Before:

image

After:

image

---

#### Issues closed by this PR

Fixes #2169

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context

To be able to test this issue:

- Open any dashboard
- Have actual data in the dashboard (more panels, faster you see the issue)
- Enable auto-refresh of 5s
- Press F12, go to Memory tab and get a Heap Snapshot
- Wait 1~2minutes
- Get another snapshot
- Select the second and compare with the first
- Search for query
- You should see a lot of queries stored (if the fix is not applied)

#### Root Cause

React query

#### Fix Strategy

Disable cache time that leads to react query to remove the query as soon it adds the item to the cache.

But we should also have another task just to update the version of react-query and then replace/improve the caching strategy.

---

### 📋 Checklist
- [x] Manually tested

---

### DIFF
--- a/frontend/src/container/DashboardContainer/DashboardVariablesSelection/QueryVariableInput.tsx
+++ b/frontend/src/container/DashboardContainer/DashboardVariablesSelection/QueryVariableInput.tsx
@@ -11,6 +11,10 @@ import { AppState } from 'store/reducers';
 import { VariableResponseProps } from 'types/api/dashboard/variables/query';
 import { GlobalReducer } from 'types/reducer/globalTime';
 
+import {
+	DASHBOARD_CACHE_TIME,
+	DASHBOARD_CACHE_TIME_ON_REFRESH_ENABLED,
+} from '../../../constants/queryCacheTime';
 import { variablePropsToPayloadVariables } from '../utils';
 import SelectVariableInput from './SelectVariableInput';
 import { useDashboardVariableSelectHelper } from './useDashboardVariableSelectHelper';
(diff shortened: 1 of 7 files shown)


frontend/src/container/GridCardLayout/GridCard/index.tsx
```diff
@@ -209,8 +211,10 @@ function GridCardGraph({
 		version || DEFAULT_ENTITY_VERSION,
 		{
 			queryKey: [
+				'dashboard-grid-card',
```
AshwinBhatkal: can you define a query key constant for this?

frontend/src/container/DashboardContainer/DashboardVariablesSelection/QueryVariableInput.tsx
```diff
@@ -184,6 +185,8 @@ function QueryVariableInput({
 					signal,
 				),
 			refetchOnWindowFocus: false,
+			// keep it low or zero, otherwise, when enabled auto-refresh, this causes OOM due to accumulated queries in cache
+			cacheTime: isAutoRefreshDisabled ? 30_000 : 0,
```
AshwinBhatkal: can you move this number 30_000 to a const? Can be reused

AshwinBhatkal (approved):

ahrefabhi (approved):

Issue #10192, opened 2026-02-05
Optimize Bundle Size


H4ad:
The following image represents the bundle separated by libs:

<img width="1922" height="964" alt="Image" src="https://github.com/user-attachments/assets/362fc424-d368-46a6-957d-f78275891d8b" />

We should aim reduce the bundle size as much as possible.

I started the work at #10191, which reduces the bundle size by 3Mb.

TODO:
- [x] react-force-graph -> to remove aframer: #10191 
- [ ] @grafana/data -> to remove moment-timezone: 
  - Started at https://github.com/grafana/grafana/pull/118243 but I'm not too optimistic since this is a huge change for Grafana. 
- [ ] react-lottie -> to remove lottie.
  - If we replace this by GIF, it will be heavier than the lib itself.
  - We only have the option to remove or keep it.
- [x] react-syntax-highlighter -> do we need that amount of langs?: #10228
- [x] posthog -> remove from the main bundle
  - After a little exploration, there's no way to remove from the bundle since it's always required for cloud.
  - Maybe we could reduce the bundle size by removing specific features but I don't know exactly what is being used.
- [x] chart.js -> remove from the main bundle: #10223
- [x] jszip/better-xlsx -> remove from the main bundle: #10229
- [x] query builder -> remove from main bundle: #10230
- [ ] better code split for Onboarding:
  - <details><summary>Info</summary><img width="809" height="676" alt="Image" src="https://github.com/user-attachments/assets/a302659d-4174-4019-9a70-d9fb4d6020fd" /></details>

And maybe we could add a github action to detect when we have huge changes in the bundles in each PR to avoid bloat the main bundle accidentally.

PR #10228, opened 2026-02-07
perf(bundle-size): be explicit when including new languages for react-syntax-highlighter

H4ad:
## Pull Request

This PR improves the bundle size for react-syntax-highlighter by including only the used languages (and themes).

I also added a little CI step to run to ensure the languages are tracked when new one is added inside any markdown of the codebase to ensure we have syntax highlight for these new languages.

This change saves about 670Kb.

---

### 📄 Summary

Before:

image

After:

image

#### Issues closed by this PR

Related: #10192

---

### ✅ Change Type

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/components/MarkdownRenderer/syntaxHighlighter.ts
+++ b/frontend/src/components/MarkdownRenderer/syntaxHighlighter.ts
@@ -1,0 +1,34 @@
+import { PrismLight as SyntaxHighlighter } from 'react-syntax-highlighter';
+import bash from 'react-syntax-highlighter/dist/esm/languages/prism/bash';
+import docker from 'react-syntax-highlighter/dist/esm/languages/prism/docker';
+import elixir from 'react-syntax-highlighter/dist/esm/languages/prism/elixir';
+import go from 'react-syntax-highlighter/dist/esm/languages/prism/go';
+import javascript from 'react-syntax-highlighter/dist/esm/languages/prism/javascript';
+import json from 'react-syntax-highlighter/dist/esm/languages/prism/json';
+import jsx from 'react-syntax-highlighter/dist/esm/languages/prism/jsx';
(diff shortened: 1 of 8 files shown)


frontend/src/components/MarkdownRenderer/syntaxHighlighter.ts
```diff
@@ -0,0 +1,34 @@
...
+SyntaxHighlighter.registerLanguage('go', go);
+SyntaxHighlighter.registerLanguage('javascript', javascript);
+SyntaxHighlighter.registerLanguage('js', javascript);
+SyntaxHighlighter.registerLanguage('json', json);
+SyntaxHighlighter.registerLanguage('jsx', jsx);
+SyntaxHighlighter.registerLanguage('rust', rust);
+SyntaxHighlighter.registerLanguage('swift', swift);
+SyntaxHighlighter.registerLanguage('ts', typescript);
+SyntaxHighlighter.registerLanguage('tsx', tsx);
+SyntaxHighlighter.registerLanguage('typescript', typescript);
```
ishan621, line 29: can we check if PrismLight internally handles tsx into typescript ?, similarly for js and javscript

H4ad: I followed the same pattern they have in the docs: https://github.com/react-syntax-highlighter/react-syntax-highlighter?tab=readme-ov-file#light-build

H4ad: But to answer your question, looks like it's a different thing for jsx and tsx.

PR #10241, opened 2026-02-09
feat: 3729 Add to alert flow from Logs Explorer doesn't work

ishan621:
## Pull Request

---

The Add to Alert functionality is currently not working as expected when creating an alert from the Logs Explorer.

While the user is on the Logs Explorer, query results are fetched using the query_range API and are ordered by timestamp, which works correctly for the list view. However, after adding an alert and navigating to the Alerts page, the alert is rendered in a time-series view.

The time-series view does not support ordering by raw timestamps and instead requires aggregation functions such as count(), sum(), etc. As a result, the query used in the Logs Explorer is incompatible with the alert’s time-series rendering, causing the functionality to behave incorrectly.

We now remove the orderBy clause from query when current dataSource is LOGS

Before : 

https://github.com/user-attachments/assets/5e2be59c-c3d4-4fb6-aede-7c3c8793a947

After: 

https://github.com/user-attachments/assets/604e858b-a8b6-4fc9-832c-190cab635451

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

faulty assumption -- Handling miss when creating an alert from LIST view 

---

Steps : 
1. Open app.us.staging.signoz.cloud 
2. Goto logs-explorer
3. Click on create an alert , previously it threw an error stating timestamp not allowed in orderBy clause.
4. Now it should work fine as expected and should render timeseries view of current query

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius: Add to alert functionality
- Potential regressions: Feature is not working, user unable to see created alert
- Rollback plan: It is not behind any feature flag hence only option would be to revert the code

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> **Medium Risk**
> Changes the query transformation used when navigating into alert creation, which could alter alert semantics (aggregation/orderBy) for logs and other explorers if edge cases aren’t covered.
> 
> **Overview**
> Fixes “Create alert” from Logs Explorer by sanitizing the query passed to the alert builder so it’s compatible with time-series alerts.
> 
> `handleConditionalQueryModification` now deep-clones and updates all query blocks, converting `aggregateOperator: NOOP` to `COUNT` (since alerts don’t support `NOOP`) and, for `DataSource.LOGS`, stripping `orderBy` to avoid timestamp ordering errors. The logic is wrapped in a `try/catch` fallback and the callback dependencies now include `sourcepage`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 27079d536d1d8c4c07223bf24dfd4b83c3299ba6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/src/container/ExplorerOptions/__tests__/ExplorerOptionWrapper.test.tsx
+++ b/frontend/src/container/ExplorerOptions/__tests__/ExplorerOptionWrapper.test.tsx
@@ -1,3 +1,4 @@
+import { useHistory } from 'react-router-dom';
 import { PANEL_TYPES } from 'constants/queryBuilder';
 import { MOCK_QUERY } from 'container/QueryTable/Drilldown/__tests__/mockTableData';
 import { useUpdateDashboard } from 'hooks/dashboard/useUpdateDashboard';
@@ -15,6 +16,11 @@ import { getExplorerToolBarVisibility } from '../utils';
 // Mock dependencies
 jest.mock('hooks/dashboard/useUpdateDashboard');
 
+jest.mock('react-router-dom', () => ({
+	...jest.requireActual('react-router-dom'),
+	useHistory: jest.fn(),
+}));
+
 jest.mock('../utils', () => ({
 	getExplorerToolBarVisibility: jest.fn(),
 	generateRGBAFromHex: jest.fn(() => 'rgba(0, 0, 0, 0.08)'),
@@ -29,6 +35,7 @@ const mockGetExplorerToolBarVisibility = jest.mocked(
 );
 
 const mockUseUpdateDashboard = jest.mocked(useUpdateDashboard);
+const mockUseHistory = jest.mocked(useHistory);
 
 // Mock data
 const TEST_QUERY_ID = 'test-query-id';
@@ -103,7 +110,6 @@ describe('ExplorerOptionWrapper', () => {
 	beforeEach(() => {
 		jest.clearAllMocks();
 		mockGetExplorerToolBarVisibility.mockReturnValue(true);
-
 		// Mock useUpdateDashboard to return a mutation object
 		mockUseUpdateDashboard.mockReturnValue(({
 			mutate: jest.fn(),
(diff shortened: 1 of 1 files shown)


frontend/src/container/ExplorerOptions/ExplorerOptions.tsx
```diff
@@ -173,22 +173,45 @@ function ExplorerOptions({
...
+						(item) => {
+							const updatedItem = { ...item };
+
+							if (updatedItem.aggregateOperator === StringOperators.NOOP) {
+								updatedItem.aggregateOperator = StringOperators.COUNT;
+							}
+
+							// Alerts do not support order by on logs explorer queries
+							if (sourcepage === DataSource.LOGS) {
+								updatedItem.orderBy = [];
```
aks07: This is removing orderBy from all the flows. Please check if we can do it in a better way.

ishan621: we have moved core logic to utils

frontend/src/container/ExplorerOptions/ExplorerOptions.tsx
```diff
@@ -173,22 +173,45 @@ function ExplorerOptions({
...
+							}
+
+							return updatedItem;
+						},
+					);
+				}
+
+				return JSON.stringify(modifiedQuery);
+			} catch (error) {
+				return JSON.stringify(defaultQuery || query);
```
aks07: If there is an error, lets not simply return default query rather lets throw error so it is easier to point out issues. Also, we should intentionally be adding try catch block knowing what errors we can expect here.

ishan621: handled better

ishan621:
> Add test case for this change

Done

PR #10243, opened 2026-02-09
feat: Filtering UI starts glitching when text is truncated

ishan621:
## Pull Request

---

Previously, when users hovered over quick filter values, a tooltip was displayed. However, since the tooltip was rendered as a fixed element, users who hovered over it and attempted to scroll were unable to scroll the page in the background, leading to a confusing and blocked interaction.

We have migrated from the Typography ellipsis/tooltip implementation to the standalone Tooltip component from Ant Design. Additionally, we’ve set the mouseLeaveDelay to 0, ensuring the tooltip disappears immediately when the cursor moves out of the tooltip trigger area, preventing it from blocking background interactions.

Before : 

https://github.com/user-attachments/assets/ef1d12c3-f69f-43b8-8e41-252cec33ab00

After : 

https://github.com/user-attachments/assets/db2755ab-c1bd-4379-a682-079a6bcee414

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

We were using typography from antd as tooltip , which does not have mouseDelay property hence did not vanish instantly as user left the area
edge case

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius: Minimal 
- Potential regressions: Tooltip wont work properly on quick filters if it fails
- Rollback plan: Code revert

---

It affects UX as now user wont face problem scrolling while tooltip is visible
> Use **N/A** for internal or non-user-facing changes

---

### 📋 Checklist
- [ ] Manually tested

---

> [!NOTE]
> **Low Risk**
> Small UI-only change limited to tooltip timing/behavior; main risk is minor regression in tooltip visibility timing for quick filter values.
> 
> **Overview**
> Updates the quick-filter checkbox value renderer to configure `Typography.Text` ellipsis tooltips with explicit `mouseEnterDelay` and `mouseLeaveDelay` (immediate hide) while keeping top placement, to prevent truncated-value tooltips from lingering and blocking interactions like background scrolling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 46afeb93a7ea0000ea5498a5d88400102ff72f8e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.tsx
+++ b/frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.tsx
@@ -650,7 +650,7 @@ export default function CheckboxFilter(props: ICheckboxProps): JSX.Element {
 													className="value-string"
 													ellipsis={{
 														tooltip: {
-															title: String(value),
+															placement: 'top',
 															mouseEnterDelay: 0.2,
 															mouseLeaveDelay: 0,
 														},
--- a/frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.styles.scss
+++ b/frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.styles.scss
@@ -65,9 +65,6 @@
 
 				.value-string {
 					width: 100%;
-					overflow: hidden;
-					text-overflow: ellipsis;
-					white-space: nowrap;
 				}
 
 				&.filter-disabled {


frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.tsx
```diff
@@ -646,12 +646,11 @@ export default function CheckboxFilter(props: ICheckboxProps): JSX.Element {
...
-													className="value-string"
-													ellipsis={{ tooltip: { placement: 'top' } }}
-												>
-													{String(value)}
-												</Typography.Text>
+												<Tooltip title={String(value)} mouseLeaveDelay={0}>
+													<Typography.Text className="value-string">
+														{String(value)}
+													</Typography.Text>
+												</Tooltip>
```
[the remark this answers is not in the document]
ishan621: that's ok , handling that would be over-engineering

frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.tsx
```diff
@@ -646,12 +646,11 @@ export default function CheckboxFilter(props: ICheckboxProps): JSX.Element {
 											{filter.customRendererForValue ? (
 												filter.customRendererForValue(value)
 											) : (
-												<Typography.Text
-													className="value-string"
-													ellipsis={{ tooltip: { placement: 'top' } }}
-												>
-													{String(value)}
-												</Typography.Text>
+												<Tooltip title={String(value)} mouseLeaveDelay={0}>
```
aks07: We should keep the mouse delay on mouse enter as this is better UX and the user wont see tooltip un-necessarily.

ishan621: we have introduced mouseEnterDelay (from antd) , with value 200ms , UI does not look jittery now

aks07 (approved):

PR #10253, opened 2026-02-09
test: added tests for uplotv2 utils

ahrefabhi:
### 📄 Summary

- **Add focused unit tests for `uPlotV2` utility functions** to validate data normalization, scale configuration, threshold handling, and fallback timestamp behavior.

---

### ✅ Change Type

- [ ] ✨ Feature  
- [ ] 🐛 Bug fix  
- [ ] ♻️ Refactor  
- [ ] 🛠️ Infra / Tooling  
- [x] 🧪 Test-only  

---

### 🧪 Testing Strategy

- **Tests added/updated**:
  - **`frontend/src/lib/uPlotV2/utils/__tests__/dataUtils.test.ts`**
    - Covers `isInvalidPlotValue`:
      - Treats `null`/`undefined`, `NaN`, `±Infinity`, and Infinity/NaN-like strings as invalid.
      - Accepts finite numbers and well-formed numeric strings as valid.
      - Leaves non-number, non-string types up to the caller (treated as valid for this helper).
    - Covers `normalizePlotValue`:
      - Returns `null` for anything considered invalid by `isInvalidPlotValue`.
      - Parses numeric strings into numbers.
      - Passes through valid numeric values unchanged.
  - **`frontend/src/lib/uPlotV2/utils/__tests__/scale.test.ts`**
    - `normalizeLogScaleLimits`: returns limits unchanged for non-log distributions, and snaps min/max/soft limits to powers of the log base for log scales.
    - `getDistributionConfig`: returns empty config for time scales, and correct linear/log distribution settings (including `distr` and `log`) for non-time scales.
    - `getRangeConfig`: computes `rangeConfig` with padding and modes, and sets `hardMinOnly`, `hardMaxOnly`, and `hasFixedRange` flags correctly.
    - `createRangeFunction`:
      - Returns `[null, null]` when there is no fixed range and no data.
      - Applies hard min/max limits when configured, even when skipping `uPlot.rangeNum`.
    - `adjustSoftLimitsWithThresholds`: returns original soft limits with no thresholds, and expands soft limits to include min/max threshold values when thresholds are provided.
    - `getFallbackMinMaxTimeStamp`: returns a sane 24-hour window (86400 seconds) ending near “now” with positive, ordered timestamps.
  - **`frontend/src/lib/uPlotV2/utils/__tests__/threshold.test.ts`**
    - `findMinMaxThresholdValues`:
      - Returns `[null, null]` for empty or missing thresholds (defensive behavior).
      - Computes min/max from `thresholdValue` when units are not provided.
      - Ignores thresholds without usable values or with unconvertible units, using only valid entries for min/max.
- **Manual verification**:
  - Not required; behavior is exercised via automated unit tests only.
- **Edge cases covered**:
  - Invalid numeric representations, Infinity/NaN variants, and non-standard input types.
  - Logarithmic vs linear distributions, missing data, and fixed vs non-fixed ranges.
  - Empty/missing threshold arrays, unknown units, and partially valid threshold lists.
  - Time fallback behavior for ranges when input data is unavailable.

---

### 📝 Changelog

- **Change Type**: Maintenance (test-only)  
- **Description**: N/A – internal test coverage improvements for `uPlotV2` utility functions.

### DIFF
--- a/.github/workflows/goci.yaml
+++ b/.github/workflows/goci.yaml
@@ -93,3 +93,13 @@ jobs:
         run: |
           go run cmd/enterprise/*.go generate openapi
           git diff --compact-summary --exit-code || (echo; echo "Unexpected difference in openapi spec. Run go run cmd/enterprise/*.go generate openapi locally and commit."; exit 1)
+      - name: node-install
+        uses: actions/setup-node@v5
+        with:
+          node-version: "22"
+      - name: install-frontend
+        run: cd frontend && yarn install
+      - name: generate-api-clients
+        run: |
+          cd frontend && yarn generate:api
+          git diff --compact-summary --exit-code || (echo; echo "Unexpected difference in generated api clients. Run yarn generate:api in frontend/ locally and commit."; exit 1)
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -4355,6 +4355,8 @@ components:
           type: string
         unit:
           type: string
+      required:
+      - name
       type: object
     Querybuildertypesv5QueryData:
       properties:
@@ -4427,6 +4429,9 @@ components:
             type: array
           nullable: true
           type: object
+      required:
+      - keys
+      - complete
       type: object
     TelemetrytypesGettableFieldValues:
       properties:
@@ -4434,6 +4439,9 @@ components:
           type: boolean
         values:
           $ref: '#/components/schemas/TelemetrytypesTelemetryFieldValues'
+      required:
+      - values
+      - complete
       type: object
     TelemetrytypesTelemetryFieldKey:
       properties:
@@ -4449,6 +4457,8 @@ components:
           type: string
         unit:
           type: string
+      required:
+      - name
       type: object
     TelemetrytypesTelemetryFieldValues:
       properties:
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -19,7 +19,7 @@
 		"commitlint": "commitlint --edit $1",
 		"test": "jest",
 		"test:changedsince": "jest --changedSince=main --coverage --silent",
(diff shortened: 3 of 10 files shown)


frontend/src/lib/uPlotV2/utils/__tests__/scale.test.ts
```diff
@@ -0,0 +1,201 @@
...
+			expect(hardMaxOnly).toBe(true);
+			expect(hasFixedRange).toBe(true);
+		});
+	});
+
+	describe('createRangeFunction', () => {
+		it('returns [dataMin, dataMax] when no fixed range and no data', () => {
+			const params = {
+				rangeConfig: {} as uPlot.Range.Config,
+				hardMinOnly: false,
```
AshwinBhatkal, line 109: hardMinOnly when set to true always required min value?

frontend/src/lib/uPlotV2/utils/__tests__/threshold.test.ts
```diff
@@ -0,0 +1,36 @@
+import { findMinMaxThresholdValues } from '../threshold';
+
+describe('findMinMaxThresholdValues', () => {
+	it('returns [null, null] when thresholds array is empty or missing', () => {
+		expect(findMinMaxThresholdValues([], 'ms')).toEqual([null, null]);
+		// @ts-expect-error intentional undefined to cover defensive branch
```
AshwinBhatkal: this may not be required, if we typed the function properly

frontend/src/lib/uPlotV2/utils/__tests__/scale.test.ts
ahrefabhi: if hardminOnly is true it'll prioratize the min value of passed otherwise it derives the dataMin from the series data

PR #10257, opened 2026-02-09
chore: shared utils update + API plumbing

AshwinBhatkal:
This PR lays the groundwork for the upcoming variable fetch state machine by introducing the following changes:

1. **Variable reference pattern utility** (`variableReference.ts`) — centralizes the regex logic for detecting variable references (`{{.var}}`, `{{var}}`, `$var`, `[[var]]`). Also provides helpers to extract all text strings from a widget `Query` and detect which variables a panel references. This replaces scattered inline regex construction and `.includes($name)` checks.

2. **API AbortSignal support** — adds an optional `AbortSignal` parameter to `dashboardVariablesQuery` and `getFieldValues` so callers can cancel in-flight requests when a variable's fetch cycle is superseded.

3. **`waitingMessage` prop for CustomSelect / CustomMultiSelect** — adds a new optional `waitingMessage` string prop that, when set, displays a loading indicator with the message in the dropdown footer. This suppresses the error/incomplete-data messages while a variable is waiting on dependencies.

None of these changes alter existing runtime behavior — they are additive types, props, and utility functions consumed by follow-up PRs.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Test have been added for the utils
- Manual verification: 
   - Verified `waitingMessage` renders correctly in the dropdown footer when passed. 
   - Verified AbortSignal propagates to axios calls.
- Edge cases covered: `textContainsVariableReference` guards against empty `text` or `variableName`.

---

- Blast radius: Minimal — all changes are additive. No existing call sites are modified to pass the new params yet.
- Potential regressions: None — new params are optional with no default behavior change.
- Rollback plan: Revert commit.

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

This is PR 1 of 4 in the variable fetch state machine series. It contains only additive, non-behavioral changes that subsequent PRs depend on:

- **PR 2** (store layer) will use `textContainsVariableReference` in `util.ts` and build the fetch state machine.
- **PR 3** (variable inputs) will pass `AbortSignal` to the API functions and `waitingMessage` to `SelectVariableInput`.
- **PR 4** (panel query gating) will use `getVariableReferencesInQuery` to determine which variables a panel depends on.

Key files to review:
- `variableReference.ts` — the core utility. Check the regex patterns match all 4 variable syntaxes.
- `CustomMultiSelect.tsx` / `CustomSelect.tsx` — the `waitingMessage` rendering and its interaction with existing `errorMessage` / `loading` / `showIncompleteDataMessage` states.


frontend/src/components/NewSelect/CustomMultiSelect.tsx
```diff
@@ -73,6 +73,7 @@ const CustomMultiSelect: React.FC<CustomMultiSelectProps> = ({
 	enableRegexOption = false,
 	isDynamicVariable = false,
 	showRetryButton = true,
+	waitingMessage,
```
ahrefabhi, line 76: can we rename this to `loadingMessage`?

frontend/src/components/NewSelect/CustomSelect.tsx
```diff
@@ -63,6 +63,7 @@ const CustomSelect: React.FC<CustomSelectProps> = ({
 	showIncompleteDataMessage = false,
 	showRetryButton = true,
 	isDynamicVariable = false,
+	waitingMessage,
```
ahrefabhi, line 66: here also `loadingMessage`?

frontend/src/components/NewSelect/CustomMultiSelect.tsx
AshwinBhatkal: This is actually waitingMessage itself. There is no loading that is happening. Happens in case when a query var options waits on its parent's options to be fetched

frontend/src/components/NewSelect/CustomSelect.tsx
AshwinBhatkal: Same as previous

frontend/src/components/NewSelect/CustomMultiSelect.tsx
AshwinBhatkal: I can control it before passing it as a prop.
Only thing with that is won't be able to customise waiting state specifically

ahrefabhi (approved):

srikanthccv (approved):

PR #10286, opened 2026-02-12
fix: added validations for having expression

tushar-signoz:
## Pull Request

---

### 📄 Summary
Validate having expression identifiers: The having expression now checks that it references valid aggregation expressions or aliases (for traces and logs). Invalid identifiers return an error listing valid keys. 

Metric having validation is deferred (TODO).

Reject aliasing in aggregation expressions: Aggregation expressions containing AS (case-insensitive) are rejected with a message to use the alias field instead. Applies to both trace and log aggregations.

Tests: Added TestQueryRangeRequest_ValidateHaving (16 cases) and TestQueryRangeRequest_ValidateAggregations (10 cases) covering nil/empty having, matching expression/alias, invalid identifiers, multiple aggregations, metric passthrough, and AS-in-expression detection across trace/log types.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---


tushar-signoz:
Hey @srikanthccv 
Ready for review. Have a look at the unit test file `pkg/querybuilder/having_rewriter_test.go` first, once that is finalised then you can review entire thing

pkg/querybuilder/having_rewriter_test.go
```diff
@@ -0,0 +1,736 @@
...
+			wantAdditional: []string{"Valid references are: [__result, __result0, count(), total]"},
+		},
+		{
+			name:       "expression not in column map",
+			expression: "sum(missing_field) > 100",
+			aggregations: []qbtypes.LogAggregation{
+				{Expression: "count()"},
+			},
+			wantErr:    true,
+			wantErrMsg: "Functions are not allowed in `Having` expression",
```
srikanthccv: I think this should be better. The "Functions are not allowed" message is not clear about why is function not allowed.

pkg/querybuilder/having_rewriter_test.go
```diff
@@ -0,0 +1,736 @@
...
+		},
+		{
+			name:       "one valid one invalid reference",
+			expression: "total > 100 AND ghost > 50",
+			aggregations: []qbtypes.LogAggregation{
+				{Expression: "count()", Alias: "total"},
+			},
+			wantErr:        true,
+			wantErrMsg:     "Invalid references in `Having` expression: [ghost]",
+			wantAdditional: []string{"Valid references are: [__result, __result0, count(), total]"},
```
srikanthccv, line 656: This is much better

pkg/querybuilder/having_rewriter_test.go
```diff
@@ -0,0 +1,736 @@
...
+		// --- Error: NOT syntax (must wrap comparison in parentheses) ---
+		{
+			name:       "NOT without parentheses",
+			expression: "NOT count() > 100",
+			aggregations: []qbtypes.LogAggregation{
+				{Expression: "count()"},
+			},
+			wantErr:        true,
+			wantErrMsg:     "Syntax error in `Having` expression",
+			wantAdditional: []string{"line 1:4 expecting one of {(, )} but got 'count'; line 1:10 expecting one of {(, ), AND, IDENTIFIER, NOT, number, quoted text} but got ')'; line 1:12 mismatched input '>' expecting <EOF>"},
```
srikanthccv: Does the actual sql parser give error for this case?

pkg/querybuilder/having_rewriter_test.go
```diff
@@ -0,0 +1,736 @@
...
+		// --- Error: malformed comparison (missing operand, operator, or connector) ---
+		{
+			name:       "two comparisons without boolean connector",
+			expression: "total > 100 count() < 500",
+			aggregations: []qbtypes.LogAggregation{
+				{Expression: "count()", Alias: "total"},
+			},
+			wantErr:        true,
+			wantErrMsg:     "Syntax error in `Having` expression",
+			wantAdditional: []string{"line 1:12 mismatched input 'count' expecting <EOF>"},
```
srikanthccv: In wherecaluse, we implicitly assume a missing connector as `AND`. We should do the same here for consistency.

pkg/querybuilder/having_rewriter_test.go
```diff
@@ -0,0 +1,736 @@
...
+		// --- Error: boolean literal as comparison value ---
+		{
+			name:       "boolean literal as comparison value",
+			expression: "count() > true",
+			aggregations: []qbtypes.LogAggregation{
+				{Expression: "count()"},
+			},
+			wantErr:        true,
+			wantErrMsg:     "Syntax error in `Having` expression",
+			wantAdditional: []string{"line 1:10 expecting one of {(, ), IDENTIFIER, number, quoted text} but got 'true'"},
```
srikanthccv: Why does it say _quoted text_ here?

pkg/querybuilder/having_rewriter_test.go
```diff
@@ -0,0 +1,736 @@
...
+		},
+		{
+			name:       "double NOT without valid grouping",
+			expression: "NOT NOT (count() > 100)",
+			aggregations: []qbtypes.LogAggregation{
+				{Expression: "count()"},
+			},
+			wantErr:        true,
+			wantErrMsg:     "Syntax error in `Having` expression",
+			wantAdditional: []string{"line 1:4 expecting one of {(, )} but got 'NOT'"},
```
srikanthccv: Same, what happens in this case in the SQL parser?

pkg/querybuilder/having_rewriter_test.go
tushar-signoz: "New aggregations cannot be defined and can only be referenced in `Having` expression"
I was considering this as well

pkg/querybuilder/having_rewriter_test.go
tushar-signoz: it works as a sql statement. However IMO we should reject such lazy typing

pkg/querybuilder/having_rewriter_test.go
tushar-signoz: it works as a sql statement. However IMO we should reject such lazy typing which has some ambiguity to it.

pkg/querybuilder/having_rewriter_test.go
tushar-signoz: Fixed

pkg/querybuilder/having_rewriter_test.go
tushar-signoz: Fixed

pkg/querybuilder/having_rewriter_test.go
srikanthccv: It will surprise users. Either we support it or at the least, we should identify and the error message should say something like "NOT requires parentheses" rather than the generic ANTLR position-based error

pkg/querybuilder/having_rewriter_test.go
srikanthccv: Can we make it return the following test case like details?

```
			wantErrMsg:     "Invalid references in `Having` expression: [ghost]",
			wantAdditional: []string{"Valid references are: [__result, __result0, count(), total]"},Collapse comment
```

pkg/querybuilder/having_expression_validator.go
```diff
@@ -0,0 +1,277 @@
...
+	// Layer 1 – Reject quoted string literals before parsing.
+	// `Having` expressions compare numeric aggregate results; string literals are not valid.
+	for _, ch := range original {
+		if ch == '\'' || ch == '"' {
+			return errors.NewInvalidInputf(
+				errors.CodeInvalidInput,
+				"`Having` expression contains string literals",
+			).WithAdditional("Aggregator results are numeric")
+		}
+	}
```
srikanthccv: So under no circumstances, the expression would have quoted literals?

srikanthccv (commented):
Some high level question, should the rewriter be separate from the validator? Instead, should be a visitor that validates and rewrites both as part of it?

grammar/HavingExpression.g4
```diff
@@ -0,0 +1,183 @@
...
+    ;
+
+fragment SIGN : [+-] ;
+
+// Numbers: optional sign, digits, optional decimal, optional scientific notation
+// E.g.: 100, -10, 0.5, 1.5e3, .75, -3.14
+NUMBER
+    : SIGN? DIGIT+ ('.' DIGIT*)? ([eE] SIGN? DIGIT+)?
+    | SIGN? '.' DIGIT+ ([eE] SIGN? DIGIT+)?
+    ;
```
srikanthccv: This creates creates ambiguity with arithmetic operators. Example, `count()-10 > 0`? Of course it could be written differently by user, but should we throw error or fix the grammar?

Could be solved by changing the factor?

```
  // Add unary sign to factor
  factor : (PLUS | MINUS) factor
         | LPAREN operand RPAREN
         | atom
         ;
```

grammar/HavingExpression.g4
```diff
@@ -0,0 +1,183 @@
...
+// Boolean logic (case-insensitive)
+NOT : [Nn][Oo][Tt] ;
+AND : [Aa][Nn][Dd] ;
+OR  : [Oo][Rr] ;
+
+// Boolean constants (case-insensitive)
+BOOL
+    : [Tt][Rr][Uu][Ee]
+    | [Ff][Aa][Ll][Ss][Ee]
+    ;
```
srikanthccv, line 221: Remove this?

grammar/HavingExpression.g4
```diff
@@ -0,0 +1,183 @@
...
+    | NUMBER
+    ;
+
+/*
+ * Aggregate function calls:
+ *   - count()
+ *   - sum(bytes)
+ *   - avg(duration_nano)
+ *
+ * Arguments are restricted to a single column identifier.
```
srikanthccv: comment is misleading about the argument being restricted to single column identifier, count_distinct can have more than one.

pkg/querybuilder/having_expression_validator.go
```diff
@@ -0,0 +1,400 @@
...
+	return v.visitOrExpression(ctx.OrExpression())
+}
+
+func (v *havingExpressionRewriteVisitor) visitOrExpression(ctx grammar.IOrExpressionContext) string {
+	andExprs := ctx.AllAndExpression()
+	parts := make([]string, len(andExprs))
+	for i, ae := range andExprs {
+		parts[i] = v.visitAndExpression(ae)
+	}
+	return strings.Join(parts, " OR ")
```
srikanthccv: Let's use `sqlbuilder` where appropriate

pkg/querybuilder/having_expression_validator.go
```diff
@@ -0,0 +1,400 @@
...
+// The grammar rule `primary ( AND primary | primary )*` allows adjacent primaries
+// without an explicit AND (implicit AND). Joining all of them with " AND " here is
+// equivalent to the old normalizeImplicitAND step.
+func (v *havingExpressionRewriteVisitor) visitAndExpression(ctx grammar.IAndExpressionContext) string {
+	primaries := ctx.AllPrimary()
+	parts := make([]string, len(primaries))
+	for i, p := range primaries {
+		parts[i] = v.visitPrimary(p)
+	}
+	return strings.Join(parts, " AND ")
```
srikanthccv: Same, wherever appropriate let's use sqlbuilder in preparing expression. Example it handles the empty primary part automatically.

grammar/HavingExpression.g4
tushar-signoz: This is used in "countIf(attribute.error = true) > 0" or any other valid aggregation expression

grammar/HavingExpression.g4
tushar-signoz: Fixed

grammar/HavingExpression.g4
tushar-signoz: Fixed, added a text case as well

tushar-signoz:
@srikanthccv Added a safe guard against white space. So if agg exp is `count_distinct(a, b)` and having `count_distinct(a,b)` (without space). This case wasn't working earlier. Updated the approach to handle this.

PR #10288, opened 2026-02-12
refactor: using QBv5 for connection status

swapnil-signoz:
## Pull Request

---

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/pkg/query-service/app/cloudintegrations/providerregistry.go
+++ b/pkg/query-service/app/cloudintegrations/providerregistry.go
@@ -3,9 +3,9 @@ package cloudintegrations
 import (
 	"log/slog"
 
+	"github.com/SigNoz/signoz/pkg/querier"
 	"github.com/SigNoz/signoz/pkg/query-service/app/cloudintegrations/implawsprovider"
 	integrationstore "github.com/SigNoz/signoz/pkg/query-service/app/cloudintegrations/store"
-	"github.com/SigNoz/signoz/pkg/query-service/interfaces"
 	"github.com/SigNoz/signoz/pkg/sqlstore"
 	"github.com/SigNoz/signoz/pkg/types/integrationtypes"
 )
@@ -13,15 +13,14 @@ import (
 func NewCloudProviderRegistry(
 	logger *slog.Logger,
 	store sqlstore.SQLStore,
-	reader interfaces.Reader,
-	querier interfaces.Querier,
+	querier querier.Querier,
 ) map[integrationtypes.CloudProviderType]integrationtypes.CloudProvider {
 	registry := make(map[integrationtypes.CloudProviderType]integrationtypes.CloudProvider)
 
 	accountsRepo := integrationstore.NewCloudProviderAccountsRepository(store)
 	serviceConfigRepo := integrationstore.NewServiceConfigRepository(store)
 
-	awsProviderImpl := implawsprovider.NewAWSCloudProvider(logger, accountsRepo, serviceConfigRepo, reader, querier)
+	awsProviderImpl := implawsprovider.NewAWSCloudProvider(logger, accountsRepo, serviceConfigRepo, querier)
 	registry[integrationtypes.CloudProviderAWS] = awsProviderImpl
 
 	return registry


pkg/query-service/app/cloudintegrations/implawsprovider/awsprovider.go
```diff
@@ -295,204 +293,256 @@ func (a *awsProvider) GetServiceDetails(ctx context.Context, req *integrationtyp
...
+			defer func() {
+				if r := recover(); r != nil {
+					a.logger.ErrorContext(
+						ctx, "panic while getting service metrics connection status",
+						"error", r,
+						"service", def.DefinitionMetadata.Id,
+					)
+				}
+			}()
 			defer wg.Done()
```
piyushsingariya: We've a recovery middleware inplace why this is needed?

pkg/query-service/app/cloudintegrations/implawsprovider/awsprovider.go
```diff
@@ -295,204 +293,256 @@ func (a *awsProvider) GetServiceDetails(ctx context.Context, req *integrationtyp
...
+			defer func() {
+				if r := recover(); r != nil {
+					a.logger.ErrorContext(
+						ctx, "panic while getting service logs connection status",
+						"error", r,
+						"service", def.DefinitionMetadata.Id,
+					)
+				}
+			}()
 			defer wg.Done()
```
piyushsingariya: this too? Why needed when middleware exists?

pkg/query-service/app/cloudintegrations/implawsprovider/awsprovider.go
```diff
@@ -295,204 +293,256 @@ func (a *awsProvider) GetServiceDetails(ctx context.Context, req *integrationtyp
...
+		}
+
+		if resp != nil && len(resp.Data.Results) < 1 {
+			continue
+		}
+
+		queryResponse, ok := resp.Data.Results[0].(*qbtypes.TimeSeriesData)
+		if !ok {
+			continue
+		}
```
piyushsingariya, line 432: Why continue, we should inform Customer something went wrong, and we (us) should be notified

pkg/query-service/app/cloudintegrations/implawsprovider/awsprovider.go
swapnil-signoz: As this is running inside a go routine, program panics even when the caller of the go routine has recovery.
So its just a practice to recover go routines with their own recovery.

pkg/query-service/app/cloudintegrations/implawsprovider/awsprovider.go
swapnil-signoz: Yes Ill return error in this case, as this unexpected.

pkg/query-service/app/cloudintegrations/implawsprovider/awsprovider.go
piyushsingariya: We can create a common function that does recovery and takes in a `callback func(err any)`, that would be a better practice?

swapnil-signoz: Yes I wanted to do that, but I couldn't find any global utils package that we use across the repo, so I did not pursue it.
But I can create something inside cloudintegrations package

swapnil-signoz: Pushed

pkg/query-service/app/cloudintegrations/implawsprovider/awsprovider.go
swapnil-signoz: Done

PR #10290, opened 2026-02-12
chore: variable fetch state machine

AshwinBhatkal:
### 📄 Summary
- Introduces a variable fetch state machine and supporting infrastructure to orchestrate dashboard variable fetching with correct dependency ordering.
- Currently, dashboard variables fetch in an order that blocks most of the elements on the UI. Whether it be other dependent variables, or it be panels
- Previously extracted shared utilities for variable reference detection are used here.
- Adds derived state (`variableTypes`, `dynamicVariableOrder`) to `dashboardVariablesStore`, and introduces a `getVariableDependencyContext` helper

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [x] 🧪 Test-only

---

---

- Tests added/updated: Yes
- Manual verification:
  - Dashboard variable dropdowns load in correct dependency order
  - Changing a parent variable cascades refetch to child variables
  - Dynamic variables with `allSelected=true` are treated as having a value
- Edge cases covered:
  - Chain dependencies (a → b → c): direct children start, deeper descendants wait
  - Parent failure cascades: descendants set to idle, dynamic variables still unlock
  - Mixed variable types: non-QUERY parents don't block, non-QUERY descendants aren't touched
  - Revalidating vs loading: variables with prior fetch data get `revalidating` state
  - Stale variable cleanup on re-initialization

---

- Blast radius: Dashboard variable fetching behavior
- Potential regressions: None expected — this PR adds the state machine infrastructure without changing existing fetch call sites.
- Rollback plan: Revert the commits. No migration or persisted state changes involved.

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- The `variableFetchStore` and `dashboardVariablesStore` are intentionally separate stores to avoid circular dependencies. The `getVariableDependencyContext()` function bridges them by reading a snapshot from `dashboardVariablesStore` and passing it as plain data to `variableFetchStore` actions.
- The `useVariableFetchState` hook is included but not yet consumed by any component — it will be wired up in a follow-up PR.
- `enqueueFetchOfAllVariables` handles two classes of variables differently: query variables respect their parent dependency graph, while dynamic variables gate on whether all variables already have `selectedValue` persisted (from localStorage/URL).


frontend/src/hooks/dashboard/utils.ts
```diff
@@ -132,7 +140,12 @@ export const createDynamicVariableToWidgetsMap = (
 						dynamicVariables.forEach((variable) => {
 							if (
 								variable.dynamicVariablesAttribute &&
-								queryData.filter?.expression?.includes(`$${variable.name}`) &&
+								variable.name &&
+								queryData.filter?.expression &&
+								textContainsVariableReference(
```
AshwinBhatkal, line 145: @ahrefabhi can you confirm if using this function is safe here? This includes check looked minimal

AshwinBhatkal: If you look at frontend/src/lib/dashboardVariables/variableReference.ts in the if condition, it has a lot more ways to identify the regex

ahrefabhi: @SagarRajput-7 can you see this? I'm not sure in what other way variable can be inside filter expression apart from $variable

SagarRajput-7: @ahrefabhi - currently is our codebase we support 4 types - 
*   {{.variableName}}  — dot prefix, optional whitespace
 *   {{variableName}}   — no dot, optional whitespace
 *   $variableName      — dollar prefix, word-boundary terminated
 *   [[variableName]]   — square brackets, optional whitespace
 
 but last i know we wanted to deprecate and hence oru documentation also shows only `$` but since as per the codebase this implementation of using `textContainsVariableReference` is cover these cases and is good.
 
 Refer - https://signoz.io/docs/userguide/manage-variables/

ahrefabhi (approved):

PR #10291, opened 2026-02-12
chore: replace prop drilling with fetch store

AshwinBhatkal:
### 📄 Summary
- Dashboard variable fetching was previously coordinated via prop drilling (`variablesToGetUpdated`, `setVariablesToGetUpdated`, `dependencyData`) through `DashboardVariableSelection → VariableItem → QueryVariableInput / DynamicVariableInput`. This made the data flow hard to trace and tightly coupled component rendering to fetch orchestration.

- This PR replaces that with a centralized `variableFetchStore` that manages the fetch lifecycle as a state machine (`idle → loading/revalidating → complete/error`, with `waiting` for dependency-blocked variables). Components now read their fetch state via `useVariableFetchState` hook instead of receiving it as props.

- Additionally, inline logic from `DynamicVariableInput` and `QueryVariableInput` was extracted into typed utility functions (`mergeUniqueStrings`, `buildExistingDynamicVariableQuery`, `settleVariableFetch`, `extractErrorMessage`), and variable-reference regex was replaced with the shared `textContainsVariableReference` utility from the previous PRs. 

- AbortController `signal` is now passed through to API calls for automatic cancellation of stale requests.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [x] 🧪 Test-only

---

- Tests added/updated: Yes
- Manual verification: Dashboard variable dropdowns load, cascade on value change, show "waiting" state for dependent variables
- Edge cases covered: ALL selected filtering, empty/null selectedValues, numeric vs string formatting, single-quote escaping, mixed type IN clauses

---

- Blast radius: Dashboard variable selection - fetch timing and cascading behavior
- Potential regressions: Variables not fetching, fetching in wrong order, stale data from cancelled requests not being properly handled.
- Rollback plan: Revert the commit

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- `variableFetchStore`, `variableFetchStoreUtils`, `useVariableFetchState`, and `variableReference` already have comprehensive test suites from prior PRs. The tests added here cover only the newly extracted utility functions in `util.ts`.
- The `variableSelectStrategyTypes.ts` change is a type-only simplification — it now references `VariableItemProps` types directly instead of duplicating the `onValueUpdate` signature.

### DIFF
--- a/frontend/src/container/PanelWrapper/constants.ts
+++ b/frontend/src/container/PanelWrapper/constants.ts
@@ -1,11 +1,11 @@
 import { PANEL_TYPES } from 'constants/queryBuilder';
-import BarPanel from 'container/DashboardContainer/visualization/panels/BarPanel/BarPanel';
 
 import TimeSeriesPanel from '../DashboardContainer/visualization/panels/TimeSeriesPanel/TimeSeriesPanel';
 import HistogramPanelWrapper from './HistogramPanelWrapper';
 import ListPanelWrapper from './ListPanelWrapper';
 import PiePanelWrapper from './PiePanelWrapper';
 import TablePanelWrapper from './TablePanelWrapper';
+import UplotPanelWrapper from './UplotPanelWrapper';
 import ValuePanelWrapper from './ValuePanelWrapper';
 
 export const PanelTypeVsPanelWrapper = {
@@ -16,7 +16,7 @@ export const PanelTypeVsPanelWrapper = {
 	[PANEL_TYPES.TRACE]: null,
 	[PANEL_TYPES.EMPTY_WIDGET]: null,
 	[PANEL_TYPES.PIE]: PiePanelWrapper,
-	[PANEL_TYPES.BAR]: BarPanel,
+	[PANEL_TYPES.BAR]: UplotPanelWrapper,
 	[PANEL_TYPES.HISTOGRAM]: HistogramPanelWrapper,
 };
 


frontend/src/container/DashboardContainer/DashboardVariablesSelection/DynamicVariableInput.tsx
```diff
@@ -192,13 +216,14 @@ function DynamicVariableInput({
...
 			variableData.dynamicVariablesAttribute,
+			variableFetchCycleId,
 		],
 		{
 			enabled:
-				variableData.type === 'DYNAMIC' &&
 				!!variableData.dynamicVariablesSource &&
-				!!variableData.dynamicVariablesAttribute,
-			queryFn: () =>
+				!!variableData.dynamicVariablesAttribute &&
```
ahrefabhi, line 232: can we convert this into a useMemo? with some comments

frontend/src/container/DashboardContainer/DashboardVariablesSelection/QueryVariableInput.tsx
```diff
@@ -169,27 +165,24 @@ function QueryVariableInput({
...
+			variableFetchCycleId,
 		],
 		{
 			enabled:
 				variableData &&
-				checkAPIInvocation(
-					variablesToGetUpdated,
-					variableData,
-					dependencyData?.parentDependencyGraph,
+				(isVariableFetching || (isVariableSettled && hasVariableFetchedOnce)),
```
ahrefabhi: nit: here also, after a certain point it become very hard to understand, it'll be good to have a useMemo, so that We can get an idea when this query willl be enabled based on the name of useMemo.

ahrefabhi (approved):

frontend/src/container/DashboardContainer/DashboardVariablesSelection/DynamicVariableInput.tsx
AshwinBhatkal: added comment inline

frontend/src/container/DashboardContainer/DashboardVariablesSelection/QueryVariableInput.tsx
AshwinBhatkal: added comment inline

ahrefabhi (approved):

ahrefabhi (approved):

PR #3780, opened 2023-10-20
chore: update logs/traces retention to 15days

makeavish:

PR #10292, opened 2026-02-12
chore: variables based panel fetching

AshwinBhatkal:
- Replaced the old `variablesToGetUpdated` queue and `widgetsByDynamicVariableId` prop drilling in `GridCard` with store-driven panel blocking via the `useIsPanelWaitingOnVariable` hook, and added `transitiveDescendants` pre-computation to `buildDependencyGraph`.
- Added unit tests for the same

N/A — test-only change.

#### Issues closed by this PR
Closes #3780

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [x] 🧪 Test-only

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered:
  - DYNAMIC variable with `allSelected=true` blocks panel regardless of `selectedValue`
  - QUERY variable with non-empty `selectedValue` does not block even while loading/revalidating
  - Variable missing from fetch store is treated as idle (no false blocking)
  - `error` state does not block panels
  - Empty array `selectedValue` is treated as empty (blocks), non-empty array does not block
  - Diamond dependency deduplicates transitive descendants correctly

---

- Blast radius: None — test-only, no production code changes.
- Potential regressions: None.
- Rollback plan: Revert commit.

---

### 📋 Checklist
- [x] Manually tested
- [ ] Breaking changes documented — N/A

### DIFF
--- a/frontend/src/container/GridCardLayout/GridCard/index.tsx
+++ b/frontend/src/container/GridCardLayout/GridCard/index.tsx
@@ -240,7 +240,6 @@ function GridCardGraph({
 				return failureCount < 2;
 			},
 			keepPreviousData: true,
-			// eslint-disable-next-line sonarjs/no-redundant-boolean
 			enabled: queryEnabledCondition,
 			refetchOnMount: false,
 			onError: (error) => {


frontend/src/container/GridCardLayout/GridCard/index.tsx
```diff
@@ -252,7 +240,8 @@ function GridCardGraph({
 				return failureCount < 2;
 			},
 			keepPreviousData: true,
-			enabled: queryEnabledCondition && !nonDynamicVariableChainToken,
+			// eslint-disable-next-line sonarjs/no-redundant-boolean
```
ahrefabhi: is this comment required?

PR #10299, opened 2026-02-14
fix: update rate/increase query and address several issues in builder…

srikanthccv:
… queries

### 📄 Summary

See this doc for metric query changes https://www.notion.so/signoz/State-of-Cumulative-Counters-306fcc6bcd19803b8ebee0efbf1c2faa

#### Issues closed by this PR

Fixes https://github.com/SigNoz/engineering-pod/issues/3907
Fixes https://github.com/SigNoz/engineering-pod/issues/3839
Fixes https://github.com/SigNoz/engineering-pod/issues/3837
Fixes https://github.com/SigNoz/engineering-pod/issues/3825
Fixes https://github.com/SigNoz/engineering-pod/issues/3195
Fixes https://github.com/SigNoz/engineering-pod/issues/3738
Fixes https://github.com/SigNoz/engineering-pod/issues/3801

### ✅ Change Type

- [x] 🐛 Bug fix

### 🐛 Bug Context

Empty time aggregation prepares an invalid query

#### Root Cause

validating at the incorrect call stack. 

#### Fix Strategy

Throw an invalid error; this itself doesn't solve the issue completely. There should be a fix on the frontend as well.

### 🧪 Testing Strategy

- Tests added/updated: yes
- Manual verification: yes
- Edge cases covered: yes

- Blast radius: should be limited
- Potential regressions:
- Rollback plan: no rollback

### 📋 Checklist
- [x] Tests added
- [x] Manually tested

### DIFF
--- a/conf/example.yaml
+++ b/conf/example.yaml
@@ -308,3 +308,14 @@ user:
       allow_self: true
       # The duration within which a user can reset their password.
       max_token_lifetime: 6h
+  root:
+    # Whether to enable the root user. When enabled, a root user is provisioned
+    # on startup using the email and password below. The root user cannot be
+    # deleted, updated, or have their password changed through the UI.
+    enabled: false
+    # The email address of the root user.
+    email: ""
+    # The password of the root user. Must meet password requirements.
+    password: ""
+    # The name of the organization to create or look up for the root user.
+    org_name: default
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -4678,6 +4678,8 @@ components:
           type: string
         id:
           type: string
+        isRoot:
+          type: boolean
         orgId:
           type: string
         role:
--- a/ee/query-service/app/api/api.go
+++ b/ee/query-service/app/api/api.go
@@ -45,7 +45,7 @@ type APIHandler struct {
 }
 
 // NewAPIHandler returns an APIHandler
-func NewAPIHandler(opts APIHandlerOptions, signoz *signoz.SigNoz) (*APIHandler, error) {
+func NewAPIHandler(opts APIHandlerOptions, signoz *signoz.SigNoz, config signoz.Config) (*APIHandler, error) {
 	baseHandler, err := baseapp.NewAPIHandler(baseapp.APIHandlerOpts{
 		Reader:                        opts.DataConnector,
 		RuleManager:                   opts.RulesManager,
@@ -58,7 +58,7 @@ func NewAPIHandler(opts APIHandlerOptions, signoz *signoz.SigNoz) (*APIHandler, 
 		Signoz:                        signoz,
 		QuerierAPI:                    querierAPI.NewAPI(signoz.Instrumentation.ToProviderSettings(), signoz.Querier, signoz.Analytics),
 		QueryParserAPI:                queryparser.NewAPI(signoz.Instrumentation.ToProviderSettings(), signoz.QueryParser),
-	})
+	}, config)
 
 	if err != nil {
 		return nil, err
--- a/ee/query-service/app/server.go
+++ b/ee/query-service/app/server.go
@@ -175,7 +175,7 @@ func NewServer(config signoz.Config, signoz *signoz.SigNoz) (*Server, error) {
 		GlobalConfig:                  config.Global,
 	}
 
-	apiHandler, err := api.NewAPIHandler(apiOpts, signoz)
+	apiHandler, err := api.NewAPIHandler(apiOpts, signoz, config)
 	if err != nil {
 		return nil, err
 	}
--- a/frontend/src/api/generated/services/sigNoz.schemas.ts
+++ b/frontend/src/api/generated/services/sigNoz.schemas.ts
@@ -1543,6 +1543,10 @@ export interface TypesUserDTO {
 	 */
 	id?: string;
 	/**
+	 * @type boolean
+	 */
+	isRoot?: boolean;
+	/**
 	 * @type string
 	 */
 	orgId?: string;
--- a/frontend/src/container/DashboardContainer/visualization/charts/ChartWrapper/ChartWrapper.tsx
+++ b/frontend/src/container/DashboardContainer/visualization/charts/ChartWrapper/ChartWrapper.tsx
@@ -23,6 +23,7 @@ export default function ChartWrapper({
 	width: containerWidth,
 	height: containerHeight,
 	showTooltip = true,
+	showLegend = true,
 	canPinTooltip = false,
 	syncMode,
 	syncKey,
@@ -36,6 +37,9 @@ export default function ChartWrapper({
 
 	const legendComponent = useCallback(
 		(averageLegendWidth: number): React.ReactNode => {
+			if (!showLegend) {
+				return null;
+			}
 			return (
 				<Legend
 					config={config}
@@ -44,7 +48,7 @@ export default function ChartWrapper({
 				/>
 			);
 		},
-		[config, legendConfig.position],
+		[config, legendConfig.position, showLegend],
 	);
 
 	const renderTooltipCallback = useCallback(
@@ -60,6 +64,7 @@ export default function ChartWrapper({
 	return (
 		<PlotContextProvider>
 			<ChartLayout
+				showLegend={showLegend}
 				config={config}
 				containerWidth={containerWidth}
 				containerHeight={containerHeight}
--- a/frontend/src/container/DashboardContainer/visualization/charts/Histogram/Histogram.tsx
+++ b/frontend/src/container/DashboardContainer/visualization/charts/Histogram/Histogram.tsx
@@ -1,0 +1,55 @@
+import { useCallback } from 'react';
+import ChartWrapper from 'container/DashboardContainer/visualization/charts/ChartWrapper/ChartWrapper';
+import HistogramTooltip from 'lib/uPlotV2/components/Tooltip/HistogramTooltip';
+import { buildTooltipContent } from 'lib/uPlotV2/components/Tooltip/utils';
+import {
+	HistogramTooltipProps,
+	TooltipRenderArgs,
+} from 'lib/uPlotV2/components/types';
+
+import { HistogramChartProps } from '../types';
+
+export default function Histogram(props: HistogramChartProps): JSX.Element {
+	const {
+		children,
+		renderTooltip: customRenderTooltip,
+		isQueriesMerged,
+		...rest
+	} = props;
+
+	const renderTooltip = useCallback(
+		(props: TooltipRenderArgs): React.ReactNode => {
+			if (customRenderTooltip) {
+				return customRenderTooltip(props);
+			}
+			const content = buildTooltipContent({
+				data: props.uPlotInstance.data,
+				series: props.uPlotInstance.series,
+				dataIndexes: props.dataIndexes,
+				activeSeriesIndex: props.seriesIndex,
+				uPlotInstance: props.uPlotInstance,
+				yAxisUnit: rest.yAxisUnit ?? '',
+				decimalPrecision: rest.decimalPrecision,
+			});
+			const tooltipProps: HistogramTooltipProps = {
+				...props,
+				timezone: rest.timezone,
+				yAxisUnit: rest.yAxisUnit,
+				decimalPrecision: rest.decimalPrecision,
+				content,
+			};
+			return <HistogramTooltip {...tooltipProps} />;
+		},
+		[customRenderTooltip, rest.timezone, rest.yAxisUnit, rest.decimalPrecision],
+	);
+
+	return (
+		<ChartWrapper
+			showLegend={!isQueriesMerged}
+			{...rest}
+			renderTooltip={renderTooltip}
+		>
+			{children}
+		</ChartWrapper>
+	);
+}
--- a/frontend/src/container/DashboardContainer/visualization/charts/types.ts
+++ b/frontend/src/container/DashboardContainer/visualization/charts/types.ts
@@ -7,6 +7,7 @@ interface BaseChartProps {
 	width: number;
 	height: number;
 	showTooltip?: boolean;
+	showLegend?: boolean;
 	timezone: string;
 	canPinTooltip?: boolean;
 	yAxisUnit?: string;
@@ -17,6 +18,7 @@ interface BaseChartProps {
 interface UPlotBasedChartProps {
 	config: UPlotConfigBuilder;
 	data: uPlot.AlignedData;
+	legendConfig: LegendConfig;
 	syncMode?: DashboardCursorSync;
 	syncKey?: string;
 	plotRef?: (plot: uPlot | null) => void;
@@ -27,13 +29,19 @@ interface UPlotBasedChartProps {
 
 export interface TimeSeriesChartProps
 	extends BaseChartProps,
+		UPlotBasedChartProps {}
+
+export interface HistogramChartProps
+	extends BaseChartProps,
 		UPlotBasedChartProps {
-	legendConfig: LegendConfig;
+	isQueriesMerged?: boolean;
 }
 
 export interface BarChartProps extends BaseChartProps, UPlotBasedChartProps {
-	legendConfig: LegendConfig;
 	isStackedBarChart?: boolean;
 }
 
-export type ChartProps = TimeSeriesChartProps | BarChartProps;
+export type ChartProps =
+	| TimeSeriesChartProps
+	| BarChartProps
+	| HistogramChartProps;
--- a/frontend/src/container/DashboardContainer/visualization/hooks/__tests__/useScrollWidgetIntoView.test.ts
+++ b/frontend/src/container/DashboardContainer/visualization/hooks/__tests__/useScrollWidgetIntoView.test.ts
@@ -1,0 +1,69 @@
+import { renderHook } from '@testing-library/react';
+import { useDashboard } from 'providers/Dashboard/Dashboard';
+
+import { useScrollWidgetIntoView } from '../useScrollWidgetIntoView';
+
+jest.mock('providers/Dashboard/Dashboard');
+
+type MockHTMLElement = {
+	scrollIntoView: jest.Mock;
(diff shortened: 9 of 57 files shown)


pkg/telemetrymetrics/statement_builder.go
```diff
@@ -5,67 +5,27 @@ import (
...
-						0  -- No next value either, can't interpolate
-				END
-			WHEN (per_series_value - lagInFrame(per_series_value, 1) OVER rate_window) < 0 THEN
-				-- Counter reset detected: the increase is the current value
-				per_series_value
-			ELSE 
-				-- Normal case: calculate increase
-				(per_series_value - lagInFrame(per_series_value, 1) OVER rate_window)
-		END`
+	RateTmpl = `multiIf(row_number() OVER rate_window = 1, nan, (per_series_value - lagInFrame(per_series_value, 1) OVER rate_window) < 0, per_series_value / (ts - lagInFrame(ts, 1) OVER rate_window), (per_series_value - lagInFrame(per_series_value, 1) OVER rate_window) / (ts - lagInFrame(ts, 1) OVER rate_window))`
```
namanv3, line 20: > (ts - lagInFrame(ts, 1) OVER rate_window)
Can we just divide by step interval here btw?

pkg/telemetrymetrics/statement_builder.go
namanv3, line 1: I was checking how PromQL handles restarts (to check for usage of start ts, their logic of not using the previous timestamp at all does seem like a better way in general to go about it. [https://github.com/prometheus/prometheus/blob/4fb6ce4d9af6b5871f454efbd7b36dcac92f9819/promql/functions.go#L224](https://github.com/prometheus/prometheus/blob/4fb6ce4d9af6b5871f454efbd7b36dcac92f9819/promql/functions.go#L224)

namanv3: Let's discuss on this once

pkg/telemetrymetrics/statement_builder.go
srikanthccv: Not really, consider the case where data is missing b/w two points by 2x of the step interval for some reason, then right thing to do is to rate of change from the last known point than the step interval.


step = 1m

12:00 pm - 2002 - 1st point
12:01 pm - 2005 - 2nd point
12:05 pm - 2013 - 3rd point
12:06 pm - 2015 - 4th point

consider this was not a reset but the values were failed to be exported (or scraped). So it should not be using the step interval for the rate calculation.

pkg/telemetrymetrics/statement_builder.go
srikanthccv: Sure please go through this https://docs.google.com/document/d/1kakDVn8aP1JerimLeazuMfy2jaB14W2kZDHMRokPcv4/edit?tab=t.0#heading=h.cmqufn5ktjuv and https://github.com/prometheus/prometheus/pull/12733 once

srikanthccv: Notes from the discussion:
- create a doc to the potential solutions for the problems mentioned in attached notion doc
- start ts has nothing do with the non `rate/increase` aggregations.

pkg/telemetrymetrics/condition_builder.go
```diff
@@ -29,13 +29,7 @@ func (c *conditionBuilder) conditionFor(
...
 ) (string, error) {
 
-	switch operator {
-	case qbtypes.FilterOperatorContains,
-		qbtypes.FilterOperatorNotContains,
-		qbtypes.FilterOperatorILike,
-		qbtypes.FilterOperatorNotILike,
-		qbtypes.FilterOperatorLike,
-		qbtypes.FilterOperatorNotLike:
+	if operator.IsStringSearchOperator() {
```
namanv3, line 32: Not sure why this was done, can you explain?

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -5,67 +5,27 @@ import (
...
+
+	IncreaseTmpl = `multiIf(row_number() OVER rate_window = 1, nan, (per_series_value - lagInFrame(per_series_value, 1) OVER rate_window) < 0, per_series_value, per_series_value - lagInFrame(per_series_value, 1) OVER rate_window)`
+
+	RateWithoutNegativeMultiTemporality = `IF(LOWER(temporality) LIKE LOWER('delta'), %s, multiIf(row_number() OVER rate_window = 1, nan, (%s - lagInFrame(%s, 1) OVER rate_window) < 0, %s / (ts - lagInFrame(ts, 1) OVER rate_window), (%s - lagInFrame(%s, 1) OVER rate_window) / (ts - lagInFrame(ts, 1) OVER rate_window))) AS per_series_value`
```
namanv3: Can name this RateMultiTemporalityTmpl. I had added "WithoutNegative" cuz that's what the OG template was named like. Same for Increase

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -307,11 +221,15 @@ func (b *MetricQueryStatementBuilder) buildTemporalAggDeltaFastPath(
...
-	aggCol := AggregationColumnForSamplesTable(
+	aggCol, err := AggregationColumnForSamplesTable(
 		start, end, query.Aggregations[0].Type, query.Aggregations[0].Temporality,
 		query.Aggregations[0].TimeAggregation, query.Aggregations[0].TableHints,
 	)
+	if err != nil {
+		return "", nil, err
+	}
 	if query.Aggregations[0].TimeAggregation == metrictypes.TimeAggregationRate {
+		// TODO(srikanthccv): should it be step interval or use [start_time_unix_nano](https://github.com/open-telemetry/opentelemetry-proto/blob/d3fb76d70deb0874692bd0ebe03148580d85f3bb/opentelemetry/proto/metrics/v1/metrics.proto#L400C11-L400C31)?
```
namanv3, line 232: `sum(value)` will give the correct sum irrespective of restarts anyway so I doubt start time will be in any way useful.
Also, start time for every delta point is the collection time of previous time stamp. If we were to use it, we can take the star time of the earliest data point in the range, but it will not give results as per the step interval at all, cuz what if that start time is many times before the interval?

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -437,8 +355,12 @@ func (b *MetricQueryStatementBuilder) buildTemporalAggDelta(
 		sb.SelectMore(fmt.Sprintf("`%s`", g.TelemetryFieldKey.Name))
 	}
 
-	aggCol := AggregationColumnForSamplesTable(start, end, query.Aggregations[0].Type, query.Aggregations[0].Temporality, query.Aggregations[0].TimeAggregation, query.Aggregations[0].TableHints)
+	aggCol, err := AggregationColumnForSamplesTable(start, end, query.Aggregations[0].Type, query.Aggregations[0].Temporality, query.Aggregations[0].TimeAggregation, query.Aggregations[0].TableHints)
+	if err != nil {
+		return "", nil, err
+	}
 	if query.Aggregations[0].TimeAggregation == metrictypes.TimeAggregationRate {
+		// TODO(srikanthccv): should it be step interval or use [start_time_unix_nano](https://github.com/open-telemetry/opentelemetry-proto/blob/d3fb76d70deb0874692bd0ebe03148580d85f3bb/opentelemetry/proto/metrics/v1/metrics.proto#L400C11-L400C31)?
```
namanv3, line 363: same as above

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -659,6 +591,8 @@ func (b *MetricQueryStatementBuilder) BuildFinalSelect(
 			sb.Where(rewrittenExpr)
 		}
 	}
+	sb.OrderBy(querybuilder.GroupByKeys(query.GroupBy)...)
+	sb.OrderBy("ts")
```
namanv3, line 595: is ordering necessary? Won't it add extra compute?

namanv3: Saw the attached bug, makes sense

pkg/telemetrymetrics/condition_builder.go
srikanthccv: When the operator is one of the string search kind, the numeric value should be converted to string to work with ClickHouse functions. This change refactors the code to make use of the `IsStringSearchOperator` on the operator.

For example, if someone does `http.status_code REGEXP 2`, the value here is number but `match` function on regex requires the string value so we convert it here before passing the arg to query binding. Does that clarify?

pkg/telemetrymetrics/statement_builder.go
srikanthccv: sure

pkg/telemetrymetrics/statement_builder.go
srikanthccv: Imagine for whatever reason, may be CPU contention or mutex contention, SDK had a gap in export between two that's again larger than the step interval, should it be step interval as the denominator or time since last report?

Same example as I mentioned above but delta (no resets).

```
step = 1m

12:00 pm - 2002 - 1st point
12:01 pm - 2005 - 2nd point
12:05 pm - 2013 - 3rd point
12:06 pm - 2015 - 4th point
```

namanv3: Fair enough. I'll write down this problem and solution in https://www.notion.so/signoz/State-of-Cumulative-Counters-306fcc6bcd19803b8ebee0efbf1c2faa?source=copy_link?

namanv3 (approved):
Statement builder changes look good, other changes also look good based on my understanding

abhishekhugetech:
@srikanthccv the changes done in `BuildFinalSelect` for ordering should fix the MatchType last. 
Thanks for updating the test cases.

pkg/types/querybuildertypes/querybuildertypesv5/builder_query.go
```diff
@@ -174,3 +175,54 @@ func (q *QueryBuilderQuery[T]) Normalize() {
...
+//   - min of mins is same as min of all values
+//
+// 4. time aggregation = max and space aggregation = max
+//   - max of maxs is same as max of all values
+//
+// 5. special case exphist, there is no need for per series/fingerprint aggregation
+// we can directly use the quantilesDDMerge function
+//
+// all of this is true only for delta metrics
+func CanShortCircuitDelta(metricAgg MetricAggregation) bool {
```
tushar-signoz, line 201: sum of counts can also be short circuited. 
Time => count and Space => sum

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
```diff
@@ -10,54 +10,78 @@ import (
...
+		return spec.Name
+	case QueryBuilderJoin:
+		return spec.Name
+	case PromQuery:
+		return spec.Name
+	case ClickHouseQuery:
+		return spec.Name
+	}
+	return ""
+}
```
tushar-signoz, line 34: I was thinking of adding this, that would save complexity in QB for spec handling. 
Consider creating a new file entirely of getters and setters. 
SetLimit, GetLimit, GetQueryName etc  (This file claude can manage)

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
```diff
@@ -10,54 +10,78 @@ import (
...
-			return fmt.Sprintf("formula '%s'", spec.Name)
-		}
-		return fmt.Sprintf("formula at position %d", index+1)
+		typeLabel = "formula"
 	case QueryTypeTraceOperator:
-		if spec, ok := envelope.Spec.(QueryBuilderTraceOperator); ok && spec.Name != "" {
-			return fmt.Sprintf("trace operator '%s'", spec.Name)
-		}
-		return fmt.Sprintf("trace operator at position %d", index+1)
+		typeLabel = "trace operator"
```
tushar-signoz, line 70: Lets add a method to the specs themselves to return a label ? Or a map of TypeToLabel

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
```diff
@@ -94,32 +119,23 @@ func (q *QueryBuilderQuery[T]) Validate(requestType RequestType) error {
...
 
 	return nil
 }
 
-func (q *QueryBuilderQuery[T]) validateSelectFields() error {
+func (q *QueryBuilderQuery[T]) validateSelectFields(requestType RequestType) error {
+	// selectFields don't apply to aggregation queries, skip validation
+	if requestType.IsAggregation() {
+		return nil
+	}
```
tushar-signoz, line 137: Should we return a BAD_REQUEST error if user is specifying select fields ?

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
```diff
@@ -132,6 +148,21 @@ func (q *QueryBuilderQuery[T]) validateSelectFields() error {
 	return nil
 }
 
+func (q *QueryBuilderQuery[T]) validateGroupBy(requestType RequestType) error {
+	// groupBy doesn't apply to non-aggregation queries, skip validation
+	if !requestType.IsAggregation() {
+		return nil
+	}
```
tushar-signoz, line 155: Should we throw BAD_REQUEST ?

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
```diff
@@ -697,12 +495,29 @@ func (c *CompositeQuery) Validate(requestType RequestType) error {
...
+	queryNames := make(map[string]bool)
+
 	for i, envelope := range c.Queries {
 		if err := validateQueryEnvelope(envelope, requestType); err != nil {
 			queryId := getQueryIdentifier(envelope, i)
 			return wrapValidationError(err, queryId, "invalid %s: %s")
 		}
+
+		// Check name uniqueness for builder queries
+		if envelope.Type == QueryTypeBuilder || envelope.Type == QueryTypeSubQuery {
```
tushar-signoz, line 508: Why only limit it for there types ? uniqueness should be enforced regardless of type.

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -5,67 +5,27 @@ import (
...
-			WHEN (per_series_value - lagInFrame(per_series_value, 1) OVER rate_window) < 0 THEN
-				-- Counter reset detected: the increase is the current value
-				per_series_value
-			ELSE 
-				-- Normal case: calculate increase
-				(per_series_value - lagInFrame(per_series_value, 1) OVER rate_window)
-		END`
+	RateTmpl = `multiIf(row_number() OVER rate_window = 1, nan, (per_series_value - lagInFrame(per_series_value, 1) OVER rate_window) < 0, per_series_value / (ts - lagInFrame(ts, 1) OVER rate_window), (per_series_value - lagInFrame(per_series_value, 1) OVER rate_window) / (ts - lagInFrame(ts, 1) OVER rate_window))`
+
+	IncreaseTmpl = `multiIf(row_number() OVER rate_window = 1, nan, (per_series_value - lagInFrame(per_series_value, 1) OVER rate_window) < 0, per_series_value, per_series_value - lagInFrame(per_series_value, 1) OVER rate_window)`
```
tushar-signoz, line 22: Can you break this into, multiple strings
For eg:
HandleFirstRecord := " the actual query "
IncreaseTmpl = multIf(HandleFirstRecord, HandleCounterReset, HandleDefault)

This would help with readability of code.

tushar-signoz (commented):
Added some nits, otherwise Looks good

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
srikanthccv: It's going to break a lot of things unnecessarily. The rule of thumb is no-op for any field if it's not related to the request type.

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
srikanthccv: https://github.com/SigNoz/signoz/pull/10299#discussion_r2816232086

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
srikanthccv: Why would that be?

pkg/telemetrymetrics/statement_builder.go
srikanthccv: I don't see how it changes anything. You don't read every time you read the code. It's a template that you know the shape and doesn't change.

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
srikanthccv: If we find ourselves needing to do the same or similar elsewhere where then we can add `TypeToLabel`. We should be conservative about not adding public methods that don't assist with main functionality.

pkg/telemetrymetrics/statement_builder.go
tushar-signoz: Im coming from perspective of adding comments in the code explaining each breakdown of multi Ifs.
Only for new contributors or people ramping up on this module.

srikanthccv: I see, that won't help because there is more context behind what goes here so they need a separate onboarding than a code comment.

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
srikanthccv: Merging now, let's continue the discussion separately.

tushar-signoz: Lets say, I have range request like
Case: 1
"A" -> builder_query
"B" -> builder_query
"B" -> trace_operator_query which uses only "A"

This gives result for only trace_operator, one of the results "B" get overwritten

Case: 2
"A" -> builder_query
"B" -> builder_query
"C" -> trace_operator_query which uses only "A"

This gives result of both A and C

PR #10304, opened 2026-02-15
Feat: Improve error messaging and UI edge cases in infra hosts monitoring 

nikhilmantri0902:
## Pull Request

---

Today infra structure monitoring does not show proper message to users when they do not see hosts metrics' data. Mentioned below what messaging improvement is added by this PR. 

1. We check if user ever sent host metrics data to us, if not - we show the onboarding guide. 
2. If the user has sent us host metrics data but his requested end time is earlier than the earliest time of retention of host metrics - convey the same to the user, so that they can adjust start and end time accordingly. 
3. If still no data returned and no other errors, that means specifically this time range and set of filters do not contain host metrics, convey the same to the users. 
4. Also we should use `signoz_metrics.distributed_metadata` to extract this info, as this table due to its aggregating rows' engine underlying, has lesser number of rows to scan. 

1. We check if user ever sent host metrics data to us, if not - we show the onboarding guide. 

NoHostDataYet

2. If the user has sent us host metrics data but his requested end time is earlier than the earliest time of retention of host metrics - convey the same to the user, so that they can adjust start and end time accordingly.  

EndTimeBeforeRetention

3. If still no data returned and no other errors, that means specifically this time range and set of filters do not contain host metrics, convey the same to the users.

NoDataInTimeRange

Closes https://github.com/SigNoz/engineering-pod/issues/3603

---

### ✅ Change Type
_Select all that apply_

- [x] 🐛 Bug fix
- [x] ♻️ Refactor

---

Today infra structure monitoring does not show proper message to users when they do not see hosts metrics' data. Refer summary section above for detailed context. 

Edge cases

Introduces proper messaging to end user so that they can check themselves any issues. 

---

- Tests added/updated for frontend. 
- Manual verification. 

---

- Blast radius: InfraStructure Hosts Monitoring, /api/v1/hosts/list API 

---

| Field | Value |
|------|-------|
| Deployment Type | Cloud + OSS + Enterprise |
| Description | Introduce proper messaging to end user |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

Please check the messaging in the screenshots and suggest improvements. 

---

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -131,7 +131,7 @@
 		"react-dom": "18.2.0",
 		"react-drag-listview": "2.0.0",
 		"react-error-boundary": "4.0.11",
-		"react-force-graph": "^1.43.0",
+		"react-force-graph-2d": "^1.29.1",
 		"react-full-screen": "1.1.1",
 		"react-grid-layout": "^1.3.4",
 		"react-helmet-async": "1.3.0",
--- a/frontend/src/modules/Servicemap/Map.tsx
+++ b/frontend/src/modules/Servicemap/Map.tsx
@@ -2,7 +2,7 @@
 //@ts-nocheck
 import { useIsDarkMode } from 'hooks/useDarkMode';
 import { memo } from 'react';
-import { ForceGraph2D } from 'react-force-graph';
+import ForceGraph2D from 'react-force-graph-2d';
 
 import { getGraphData, getTooltip, transformLabel } from './utils';
 
--- a/frontend/yarn.lock
+++ b/frontend/yarn.lock
@@ -2,37 +2,6 @@
 # yarn lockfile v1
 
 
-"3d-force-graph-ar@1":
-  version "1.8.3"
-  resolved "https://registry.npmjs.org/3d-force-graph-ar/-/3d-force-graph-ar-1.8.3.tgz"
-  integrity sha512-irj1Kk2qsRGSMsddSXdiVzzwtsNMZRbsJYBg+0ypkk5t5arZpo3h/PoN9Gfusvlu5fCAOk1wKVPUWN9Vkwrn7g==
-  dependencies:
-    aframe-forcegraph-component "3"
-    kapsule "1"
-
-"3d-force-graph-vr@2":
-  version "2.2.2"
-  resolved "https://registry.npmjs.org/3d-force-graph-vr/-/3d-force-graph-vr-2.2.2.tgz"
-  integrity sha512-LSI24ugjT7SUE8JCusV5pmFHaFZ51EgGH1B7KMOK4JEgmUvogFwkJcXg7Go2mHI28WXDzFyBvrckWiwv5yLUeA==
-  dependencies:
-    accessor-fn "1"
-    aframe "^1.4"
-    aframe-extras "^6.1"
-    aframe-forcegraph-component "3"
-    kapsule "1"
-    polished "4"
-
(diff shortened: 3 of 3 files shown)


frontend/src/api/infraMonitoring/getHostLists.ts
```diff
@@ -50,6 +50,7 @@ export interface HostListResponse {
 		total: number;
 		sentAnyHostMetricsData: boolean;
 		isSendingK8SAgentMetrics: boolean;
+		endTimeBeforeRetention?: boolean;
```
amlannandy: Are we sure it's an optional field?

frontend/src/container/InfraMonitoringHosts/HostsListTable.tsx
```diff
@@ -20,6 +20,93 @@ import {
 	HostsListTableProps,
 } from './utils';
 
+function getEmptyOrLoadingView(viewState: {
```
amlannandy: Let's keep it as a component as it's returning JSX
By that I mean it's naming structure.
- Should be starting uppercase and `get` prefix is not required.
- So like `EmptyOrLoadingView)

frontend/src/container/InfraMonitoringHosts/HostsListTable.tsx
```diff
@@ -20,6 +20,93 @@ import {
...
+	isError: boolean;
+	errorMessage: string;
+	showHostsEmptyState: boolean;
+	sentAnyHostMetricsData: boolean;
+	isSendingIncorrectK8SAgentMetrics: boolean;
+	showEndTimeBeforeRetentionMessage: boolean;
+	showNoRecordsInSelectedTimeRangeMessage: boolean;
+	showTableLoadingState: boolean;
```
amlannandy: Declare the props typing in a separate interface and use that instead.
Similar to `HostsListTable` component using `HostsListTableProps` interface

frontend/src/container/InfraMonitoringHosts/HostsListTable.tsx
```diff
@@ -20,6 +20,93 @@ import {
...
+}): React.ReactNode {
+	const { isError, errorMessage } = viewState;
+	if (isError) {
+		return <Typography>{errorMessage}</Typography>;
```
amlannandy: Should we still have a static fallback error message like "Something went wrong"? In case `errorMessage` is empty?

frontend/src/api/infraMonitoring/getHostLists.ts
nikhilmantri0902: Yes, good catch. its not optional, made it mandatory now.

frontend/src/container/InfraMonitoringHosts/HostsListTable.tsx
nikhilmantri0902: done

frontend/src/container/InfraMonitoringHosts/HostsListTable.tsx
nikhilmantri0902: done

frontend/src/container/InfraMonitoringHosts/HostsListTable.tsx
nikhilmantri0902: done

srikanthccv (approved):

PR #10313, opened 2026-02-16
feat: add root user support

therealpandey:
## Summary

- Adds root user support with environment-based provisioning, protection guards, and automatic reconciliation. A root user is a special admin user that is provisioned via configuration (environment variables) rather than the UI, designed for automated/headless deployments.

## Key Features
- Environment-based provisioning: Configure root user via user.root.enabled, user.root.email, user.root.password, and user.root.org_name settings

- Automatic reconciliation: A background service runs on startup that:
    - Looks up the organization by configured org_name
    - If no matching org exists, creates the organization and root user via CreateFirstUser
    - If the org exists, reconciles the root user (creates, promotes existing user, or updates email/password to match config)
    - Retries every 10 seconds until successful

- Protection guards: Root users cannot be:
    - Updated or deleted through the API
    - Invited or have their password changed through the UI
    - Authenticated via SSO/SAML (password-only authentication enforced)

- Self-registration disabled: When root user provisioning is enabled, the self-registration endpoint (/register) is blocked to prevent creating duplicate organizations

- Idempotent password sync: On every reconciliation, the root user's password is synced with the configured value — if it differs, it's updated; if it matches, no-op

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -55,6 +55,7 @@
 		"@signozhq/icons": "0.1.0",
 		"@signozhq/input": "0.0.2",
 		"@signozhq/popover": "0.0.0",
+		"@signozhq/radio-group": "0.0.2",
 		"@signozhq/resizable": "0.0.0",
 		"@signozhq/sonner": "0.1.0",
 		"@signozhq/table": "0.3.7",
--- a/frontend/src/auto-import-registry.d.ts
+++ b/frontend/src/auto-import-registry.d.ts
@@ -21,6 +21,7 @@ import '@signozhq/design-tokens';
 import '@signozhq/icons';
 import '@signozhq/input';
 import '@signozhq/popover';
+import '@signozhq/radio-group';
 import '@signozhq/resizable';
 import '@signozhq/sonner';
 import '@signozhq/table';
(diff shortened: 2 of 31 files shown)


therealpandey:
Closes https://github.com/SigNoz/platform-pod/issues/1076

karanbalani:
Final report before approving these changes: https://github.com/SigNoz/platform-pod/issues/1076#issuecomment-3913355688

pkg/modules/user/impluser/service.go
```diff
@@ -0,0 +1,187 @@
...
+		store:     store,
+		module:    module,
+		orgGetter: orgGetter,
+		authz:     authz,
+		config:    config,
+		stopC:     make(chan struct{}),
+	}
+}
+
+func (s *service) Start(ctx context.Context) error {
```
vikrantgupta25, line 47:
```suggestion
func (service *service) Start(ctx context.Context) error {
```

pkg/modules/user/impluser/service.go
```diff
@@ -0,0 +1,187 @@
...
+			); err != nil {
+				return err
+			}
+		}
+
+		return s.setPassword(ctx, existingUser.ID)
+	}
+
+	// Create new root user
+	newUser, err := types.NewRootUser(s.config.Email.String(), s.config.Email, orgID)
```
vikrantgupta25, line 137:
```suggestion
	newRootUser, err := types.NewRootUser(s.config.Email.String(), s.config.Email, orgID)
```

PR #10322, opened 2026-02-16
chore: integation tests to use the root user functionality

karanbalani:
### 📄 Summary

Update all integration tests to authenticate using the new environment-configured root user instead of dynamically registering an admin user via the `/api/v1/register` endpoint.

Changes:

- Add ROOT_USER_EMAIL/ROOT_USER_PASSWORD in fixtures/signoz.py and configure root user env vars (SIGNOZ_USER_ROOT_*) on the test server
- Replace all USER_ADMIN_* references with ROOT_USER_* across ~20 test files
- Remove create_user_admin fixture and its dependency from test signatures
- Comment out test_register / test_register_with_invalid_input (register flow changes with root user)

### TEMP: this is for now a separate PR to a diff branch from main, will be fixed in a while, after testing.


karanbalani:
Closing this for now. We will be taking this up later.
cc: @therealpandey

PR #10324, opened 2026-02-17
fix: prevent users from deleting themselves

karanbalani:
### 📄 Summary

Adds a check in the DeleteUser API to return a 403 Forbidden error when a user attempts to delete their own account.

This is part of the fix for [https://github.com/SigNoz/platform-pod/issues/1076]. The complete solution is a combination of:
1. The root user PR (which ensures there's always an admin who can't be removed)
2. This PR -> a small safeguard that prevents a user from deleting themselves via the API

Closes: https://github.com/SigNoz/platform-pod/issues/528

### DIFF
--- a/pkg/modules/user/impluser/handler.go
+++ b/pkg/modules/user/impluser/handler.go
@@ -259,6 +259,11 @@ func (h *handler) DeleteUser(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if claims.UserID == id {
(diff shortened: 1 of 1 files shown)


karanbalani:
Closes: https://github.com/SigNoz/platform-pod/issues/528

karanbalani:
Once root user PR is complete:
- remove last admin removal check from API
- there is one more check in update api that if the user is last admin, we don't allow that

karanbalani:
Closing in favor of: https://github.com/SigNoz/signoz/pull/10688

Issue #9802, opened 2025-12-12
Can two json bodies be merged at the query time


srikanthccv:
> We wanted to check if it's possible to do the merging of two JSON bodies in ClickHouse instead of this code block here. 

 _Originally posted by @srikanthccv in [#9381](https://github.com/SigNoz/signoz/pull/9381/changes#r2613007438)_

piyushsingariya:
```
select jsonMergePatch(toJSONString(body_json), toJSONString(body_json_promoted)) from distributed_logs_v2 
SETTINGS output_format_native_write_json_as_string=1, min_bytes_to_use_direct_io = 1, max_threads = 5, use_query_cache = 0, enable_filesystem_cache = 0, use_query_condition_cache = 0

Showed 1000 out of 639918444 rows.

639918444 rows in set. Elapsed: 35.342 sec. Processed 639.92 million rows, 10.24 GB (18.11 million rows/s., 289.70 MB/s.)
Peak memory usage: 57.38 MiB.
```
```
select jsonMergePatch(body_json::String, body_json_promoted::String) from distributed_logs_v2 
SETTINGS output_format_native_write_json_as_string=1, min_bytes_to_use_direct_io = 1, max_threads = 5, use_query_cache = 0, enable_filesystem_cache = 0, use_query_condition_cache = 0

Showed 1000 out of 639918444 rows.

639918444 rows in set. Elapsed: 34.946 sec. Processed 639.92 million rows, 10.24 GB (18.31 million rows/s., 292.98 MB/s.)
Peak memory usage: 55.73 MiB.
```
```
select body_json, body_json_promoted from distributed_logs_v2 
SETTINGS output_format_native_write_json_as_string=1, min_bytes_to_use_direct_io = 1, max_threads = 5, use_query_cache = 0, enable_filesystem_cache = 0, use_query_condition_cache = 0

639918444 rows in set. Elapsed: 346.043 sec. Processed 639.92 million rows, 10.24 GB (1.85 million rows/s., 29.59 MB/s.)
Peak memory usage: 58.36 MiB.
```

piyushsingariya:
Unblocked on 16th Feb

PR #10325, opened 2026-02-17
refactor: consolidate body column for JSON logs

piyushsingariya:
## Pull Request

---

### 📄 Summary
This PR aims to resolve the issue with body being split up into three columns, stitching logic, as well as mapping of body logical field to body_v2.message

Closes signoz/signoz#9802
Closes signoz/engineering-pod#3737

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Yes
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes how log `body` fields are selected, filtered, and typed (including new warning behavior), which can alter query results and ClickHouse SQL generation across logs dashboards.
> 
> **Overview**
> Fixes the split `body` representation by treating the logical `body` field as `body_v2.message` when JSON-body querying is enabled, including updating intrinsic field mappings, default full-text column selection, and emitting a warning when users query bare `body`.
> 
> Refactors log SQL generation to alias `body_v2` as `body` (instead of selecting/merging `body`, `body_v2`, and `body_promoted`), updates promote/metadata/field-mapping codepaths from `BodyJSON` naming to `BodyV2`, and adjusts `EXISTS`/default-exists-filter behavior for JSON fields.
> 
> Removes `sonic`-based JSON post-processing in the querier (JSON columns are now returned as strings), adds `FieldDataTypeJSON` handling in filter building, and updates/extends tests to cover the new `body` semantics and warning propagation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ab0852bbfb4d3d750a31866c546ec8792cad79d9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/telemetrylogs/json_stmt_builder_test.go
+++ b/pkg/telemetrylogs/json_stmt_builder_test.go
@@ -2,8 +2,6 @@ package telemetrylogs
 
 import (
 	"context"
-	"slices"
-	"strings"
 	"testing"
 	"time"
 
@@ -796,17 +794,17 @@ func TestStatementBuilderListQueryBodyMessage(t *testing.T) {
 	}
 }
 
-func buildTestTelemetryMetadataStore(t *testing.T, promotedPaths ...string) *telemetrytypestest.MockMetadataStore {
+func buildTestTelemetryMetadataStore(t *testing.T) *telemetrytypestest.MockMetadataStore {
 	mockMetadataStore := telemetrytypestest.NewMockMetadataStore()
 	mockMetadataStore.SetStaticFields(IntrinsicFields)
 	types, _ := telemetrytypes.TestJSONTypeSet()
 	for path, jsonTypes := range types {
-		promoted := false
+		// promoted := false
 
-		split := strings.Split(path, telemetrytypes.ArraySep)
-		if slices.Contains(promotedPaths, split[0]) {
-			promoted = true
-		}
+		// split := strings.Split(path, telemetrytypes.ArraySep)
+		// if slices.Contains(promotedPaths, split[0]) {
+		// 	promoted = true
+		// }
 		// Create a TelemetryFieldKey for each JSONDataType for this path
 		// Since a path can have multiple types, we create one key per type
 		for _, jsonType := range jsonTypes {
@@ -816,7 +814,7 @@ func buildTestTelemetryMetadataStore(t *testing.T, promotedPaths ...string) *tel
 				FieldContext:  telemetrytypes.FieldContextBody,
 				FieldDataType: telemetrytypes.MappingJSONDataTypeToFieldDataType[jsonType],
 				JSONDataType:  &jsonType,
-				Materialized:  promoted,
+				// Materialized:  promoted,
 			}
 			err := key.SetJSONAccessPlan(telemetrytypes.JSONColumnMetadata{
 				BaseColumn:     LogsV2BodyV2Column,
@@ -830,8 +828,8 @@ func buildTestTelemetryMetadataStore(t *testing.T, promotedPaths ...string) *tel
 	return mockMetadataStore
 }
 
-func buildJSONTestStatementBuilder(t *testing.T, promotedPaths ...string) *logQueryStatementBuilder {
-	mockMetadataStore := buildTestTelemetryMetadataStore(t, promotedPaths...)
+func buildJSONTestStatementBuilder(t *testing.T) *logQueryStatementBuilder {
+	mockMetadataStore := buildTestTelemetryMetadataStore(t)
 	fm := NewFieldMapper()
 	cb := NewConditionBuilder(fm)
 


pkg/prometheus/clickhouseprometheus/client.go
```diff
@@ -87,6 +87,24 @@ func (client *client) Read(ctx context.Context, query *prompb.Query, sortSeries
 	return remote.FromQueryResult(sortSeries, res), nil
 }
 
+func (c *client) ReadMultiple(ctx context.Context, queries []*prompb.Query, sortSeries bool) (storage.SeriesSet, error) {
```
nityanandagohain, line 94: Where is this used ?

pkg/prometheus/utils.go
```diff
@@ -15,30 +16,24 @@ func RemoveExtraLabels(res *promql.Result, labelsToRemove ...string) error {
 		toRemove[l] = struct{}{}
 	}
 
+	dropLabels := func(metric labels.Labels) labels.Labels {
```
nityanandagohain: why these changes ?

pkg/querier/promql_query.go
```diff
@@ -240,13 +241,13 @@ func (q *promqlQuery) Execute(ctx context.Context) (*qbv5.Result, error) {
...
 	for _, v := range matrix {
 		var s qbv5.TimeSeries
-		lbls := make([]*qbv5.Label, 0, len(v.Metric))
-		for name, value := range v.Metric.Copy().Map() {
+		lbls := make([]*qbv5.Label, 0, v.Metric.Len())
+		v.Metric.Range(func(l labels.Label) {
 			lbls = append(lbls, &qbv5.Label{
-				Key:   telemetrytypes.TelemetryFieldKey{Name: name},
-				Value: value,
+				Key:   telemetrytypes.TelemetryFieldKey{Name: l.Name},
```
nityanandagohain, line 264: these as well

pkg/types/telemetrytypes/json_access_plan_test.go
```diff
@@ -1,12 +1,19 @@
...
+	"fmt"
 	"testing"
 
+	otelconstants "github.com/SigNoz/signoz-otel-collector/constants"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/yaml.v3"
 )
 
+const (
+	bodyV2Column       = otelconstants.BodyV2Column
```
nityanandagohain, line 13: nit: any reason of redefinding the constansts ?

pkg/types/telemetrytypes/field.go
```diff
@@ -29,6 +29,8 @@ const (
 	ArrayAnyIndexSuffix = "[*]"
 )
 
+type TelemetryFieldKeys map[string][]*TelemetryFieldKey
```
nityanandagohain: why is it here ?

pkg/types/telemetrytypes/field.go
```diff
@@ -169,6 +176,10 @@ func (f *TelemetryFieldKey) Normalize() {
 
 }
 
+func (f *TelemetryFieldKey) QualifiedName() string {
+	return f.Name + ";" + f.FieldContext.StringValue() + ";" + f.FieldDataType.StringValue()
```
nityanandagohain: should we add signal as well to make it generic ?

pkg/telemetrylogs/condition_builder.go
```diff
@@ -35,13 +35,19 @@ func (c *conditionBuilder) conditionFor(
...
 	}
 
-	if column.IsJSONColumn() && querybuilder.BodyJSONQueryEnabled {
-		valueType, value := InferDataType(value, operator, key)
-		cond, err := NewJSONConditionBuilder(key, valueType).buildJSONCondition(operator, value, sb)
-		if err != nil {
-			return "", err
+	if column.Type.GetType() == schema.ColumnTypeEnumJSON {
+		// If field data is Not JSON Column Itself, then we need to build a JSON condition
+		if querybuilder.BodyJSONQueryEnabled && key.MustBuildJSONCondition() {
```
nityanandagohain: please explain this better

pkg/telemetrylogs/const.go
```diff
@@ -34,11 +37,23 @@ const (
 	LogsV2ResourcesStringColumn  = "resources_string"
 	LogsV2ScopeStringColumn      = "scope_string"
 
-	BodyJSONColumnPrefix     = constants.BodyJSONColumnPrefix
+	BodyV2ColumnPrefix       = constants.BodyV2ColumnPrefix
 	BodyPromotedColumnPrefix = constants.BodyPromotedColumnPrefix
+	MessageSubColumn         = "message"
+	bodySearchDefaultWarning = "When you search on `body` (full text or by field), Query Builder uses body.message:string by default. Check that this matches what you want to search."
```
nityanandagohain: `body searches default to body.message:string. Use body.<key> to search a different field inside body`

Later we can also add about the operator for searching across all keys

pkg/telemetrylogs/field_mapper.go
```diff
@@ -147,6 +147,9 @@ func (m *fieldMapper) FieldFor(ctx context.Context, key *telemetrytypes.Telemetr
 			}
 
 			return m.buildFieldForJSON(key)
+		case telemetrytypes.FieldContextLog:
+			// return the column name as is for log context fields
```
nityanandagohain: can you point me to a test which can help me understand this ?

pkg/types/querybuildertypes/querybuildertypesv5/builder_elements.go
```diff
@@ -197,6 +197,15 @@ func (f FilterOperator) IsArrayOperator() bool {
 	}
 }
 
+func (f FilterOperator) IsOpValidForJSON() bool {
```
nityanandagohain: please add a comment on this, where and how will it be used

nityanandagohain (commented):
Added some comments, and I see some promql changes added here , don't think it's has to do something with this PR.

pkg/types/telemetrytypes/json_access_plan_test.go
piyushsingariya: for tests I created them, I couldn't import from telemetrylogs const.go cuz it would be import cycle

pkg/types/telemetrytypes/field.go
piyushsingariya: I picked up the same logic that's used in GetMultiKey, it can evolve as needed in future

pkg/telemetrylogs/field_mapper.go
piyushsingariya: This is not in use now, I can remove it from here and add into the FTS (full text search) PR later.

pkg/types/querybuildertypes/querybuildertypesv5/builder_elements.go
piyushsingariya: This too is not in action now, I'll remove this and reintroduce in FTS pr

pkg/telemetrylogs/const.go
```diff
@@ -118,3 +133,29 @@ var (
...
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	if querybuilder.BodyJSONQueryEnabled {
+		DefaultFullTextColumn = BodyLogicalFieldJSONMapping
+		IntrinsicFields["body"] = *BodyLogicalFieldJSONMapping
+	}
+}
```
tushar-signoz: Where is this being used ?

pkg/telemetrymetadata/metadata.go
```diff
@@ -343,9 +343,10 @@ func (t *telemetryMetaStore) getLogsKeys(ctx context.Context, fieldKeySelectors
 	if err != nil {
 		return nil, false, err
 	}
-	mapOfKeys := make(map[string]*telemetrytypes.TelemetryFieldKey)
+	// setOfKeys to reuse the same key object for qualified names
+	setOfKeys := make(map[string]*telemetrytypes.TelemetryFieldKey)
 	for _, key := range matKeys {
-		mapOfKeys[key.Name+";"+key.FieldContext.StringValue()+";"+key.FieldDataType.StringValue()] = key
+		setOfKeys[key.QualifiedName()] = key
```
tushar-signoz: There is Text() method which should serve the purpose

pkg/types/telemetrytypes/field.go
tushar-signoz: `TelemetryFieldKeys` not a good type, `s` can easily be missed and name doesn't suggests that it is a map.

pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
```diff
@@ -176,8 +177,9 @@ func matchesKey(selector *telemetrytypes.FieldKeySelector, key *telemetrytypes.T
 		return true
 	}
 
+	matchNameExceptions := []string{"body"}
 	// Check name (already checked in matchesName, but double-check here)
-	if selector.Name != "" && !matchesName(selector, key.Name) {
+	if selector.Name != "" && !matchesName(selector, key.Name) && slices.Contains(matchNameExceptions, key.Name) {
 		return false
 	}
```
tushar-signoz: Is this needed, since body is an intrinsic field and should always be included in the mapOfKeys

pkg/querier/consume.go
```diff
@@ -394,17 +393,11 @@ func readAsRaw(rows driver.Rows, queryName string) (*qbtypes.RawData, error) {
...
 			if strings.HasPrefix(strings.ToUpper(colTypes[i].DatabaseTypeName()), "JSON") {
 				switch x := val.(type) {
 				case []byte:
-					if len(x) > 0 {
-						var v any
-						if err := sonic.Unmarshal(x, &v); err == nil {
-							val = v
-						}
-					}
+					val = string(x)
```
tushar-signoz, line 400: Why is this not needed anymore ?

pkg/types/telemetrytypes/field.go
```diff
@@ -41,6 +43,7 @@ type TelemetryFieldKey struct {
 	JSONPlan     JSONAccessPlan      `json:"-"`
 	Indexes      []JSONDataTypeIndex `json:"-"`
 	Materialized bool                `json:"-"` // refers to promoted in case of body.... fields
+	Warnings     []string            `json:"-"` // warnings attached to using a particular field key
```
tushar-signoz: imo it should not be part of the key struct.

piyushsingariya: if warnings are attached(specific) to a key, the relationship works both ways, the key is effectively attached to the warnings.

Check this other consecutive PR(just a draft not final changes): https://github.com/SigNoz/signoz/pull/10413

piyushsingariya: Current changes felt right in the above mentioned sense, I tried this earlier 

https://github.com/SigNoz/signoz/blob/15cfccad74588c8028486ab012ae19c9b3e58582/pkg/querybuilder/where_clause_visitor.go#L912-L918

And this same if logic would've repeated in all places where I am referencing key's warnings in current state, and the same for coming PR I shared above

pkg/querier/consume.go
piyushsingariya: Earlier we have three columns representing same body logical field, so I was going some checks, and replacing stuff, and that time it was a must to change JSON into map and perform those checks.

Now not needed so simply accepting as String

pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
piyushsingariya: This is mock metadata store, this one works differently so I had to do this hack for tests

pkg/telemetrymetadata/metadata.go
piyushsingariya: Will check

pkg/telemetrylogs/const.go
piyushsingariya: init functions are invoked during starting up, so they trigger and updates the Intrinsic Fields inplace, I had to go around this cuz IntrinsicFields (map) is directly referenced at multiple places.

@nityanandagohain

pkg/prometheus/clickhouseprometheus/client.go
srikanthccv: I see the collector bumped the prometheus version which make the our adapter to implement `ReadMultiple` this needs some testing.

pkg/types/telemetrytypes/field.go
srikanthccv: It doesn't make sense for a key to have a warnings. If the goal is to warn user for a specific reason, let's do it somewhere else that makes sense. If I understand the goal is a generic warning when a `body` field is used in search expression with new JSON experience; repetition or other reason doesn't justify introducing warnings as a property of the key. 

In the statement builder, run the lexer to get you the keys and attach a warning before responding with query statement.

The `getKeySelectors` uses `QueryStringToKeysSelectors` which in turn uses lexer to get the keys present in the expression. If the `body` key present, then you can return the warning.

pkg/types/telemetrytypes/field.go
```diff
@@ -62,6 +63,10 @@ func (f *TelemetryFieldKey) ArrayParentPaths() []string {
 	return paths
 }
 
+func (f *TelemetryFieldKey) MustBuildJSONCondition() bool {
```
srikanthccv: This is not used anywhere

pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
srikanthccv: Can you please share more details, what is the reason for this change? The name is misleading, it's [fake](https://abseil.io/resources/swe-book/html/ch13.html) and should be functionally same as the production so shouldn't work differently for an outsider (if it does, it should be fixed) but the lightweight implementation exists to ease the testing and remains same as the real implementation for the consumer.

pkg/telemetrymetadata/metadata.go
srikanthccv, line 1: There are bunch of changes in this file. I was not able to follow the changes by reading the code. Please help me with details. What is the objective and why does it require change.

srikanthccv: Discussed on huddle, this is done to accommodate `message` as the `Name` but `body` as the map key. @srikanthccv to get back on this.

pkg/querybuilder/fallback_expr.go
```diff
@@ -204,7 +204,10 @@ func DataTypeCollisionHandledFieldName(key *telemetrytypes.TelemetryFieldKey, va
 	// While we expect user not to send the mixed data types, it inevitably happens
 	// So we handle the data type collisions here
 	switch key.FieldDataType {
-	case telemetrytypes.FieldDataTypeString, telemetrytypes.FieldDataTypeArrayString:
+	case telemetrytypes.FieldDataTypeString, telemetrytypes.FieldDataTypeArrayString, telemetrytypes.FieldDataTypeJSON:
+		if key.FieldDataType == telemetrytypes.FieldDataTypeJSON {
+			tblFieldName = fmt.Sprintf("toString(%s)", tblFieldName)
+		}
```
srikanthccv: Let's move `telemetrytypes.FieldDataTypeJSON` to separate case branch. The other two are already string and `toString` is redundant, however, for JSON, `toString` seem like a need so let's keep them separate to avoid any confusion why three of them share the same branch.

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -313,37 +313,30 @@ func (v *filterExpressionVisitor) VisitPrimary(ctx *grammar.PrimaryContext) any
...
-				return ""
-			}
-			return cond
 		}
+		cond, err := v.conditionBuilder.ConditionFor(context.Background(), v.fullTextColumn, qbtypes.FilterOperatorRegexp, FormatFullTextSearch(searchText), v.builder, v.startNs, v.endNs)
+		if err != nil {
+			v.errors = append(v.errors, fmt.Sprintf("failed to build full text search condition: %s", err.Error()))
+			return ""
+		}
+		return cond
```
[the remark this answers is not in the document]
srikanthccv: As the comment at line "// Handle standalone key/value as a full text search term" says the code handles the standalone key/value. When does it happen that it doesn't match either key/value context, given the expression is valid?

pkg/telemetrylogs/const.go
```diff
@@ -118,3 +133,23 @@ var (
...
+func bodyAliasExpression() string {
+	if !querybuilder.BodyJSONQueryEnabled {
+		return LogsV2BodyColumn
+	}
+
+	return fmt.Sprintf("%s as body", LogsV2BodyV2Column)
+}
+
+func enrichMapsForJSONBodyEnabled() {
+	if querybuilder.BodyJSONQueryEnabled {
```
srikanthccv: Right now, throughout the code base there is simple bool flag based switch b/w the old body and new json. It would be very helpful to chart out the path for roll out for old users. It will bring some problems that either needs to be handled at the application level or some other means of migration. Please give it some thought.

srikanthccv: The reason I bring up is because, there are broadly two-three options,
1. migrate them from old data to new data (including those who have the year+ retention); this will be operationally involving and time taking process
2. roll out and wait for the old data to flushed away with retention; not a viable option for users with longer retention
3. use the same path as the resource json


It would be good idea to write down the pro/con of approaches and take a one path so we can embed the application logic to behave accordingly.

pkg/telemetrylogs/field_mapper.go
```diff
@@ -147,6 +148,8 @@ func (m *fieldMapper) FieldFor(ctx context.Context, key *telemetrytypes.Telemetr
 			}
 
 			return m.buildFieldForJSON(key)
+		case telemetrytypes.FieldContextLog:
+			return column.Name, nil
```
srikanthccv: This is not needed anymore as the datatype is no longer json but instead string?

pkg/telemetrylogs/statement_builder.go
```diff
@@ -76,20 +76,32 @@ func (b *logQueryStatementBuilder) Build(
...
+	default:
+		return nil, errors.NewInvalidInputf(errors.CodeInvalidInput, "unsupported request type: %s", requestType)
 	}
 
-	return nil, errors.NewInvalidInputf(errors.CodeInvalidInput, "unsupported request type: %s", requestType)
+	if err != nil {
+		return nil, err
+	}
+
+	if stmt != nil && len(warnings) > 0 {
```
srikanthccv: Should you want to send warning regardless of the presence of other warnings?

pkg/telemetrylogs/const.go
```diff
@@ -118,3 +133,23 @@ var (
...
+		DefaultFullTextColumn = BodyLogicalFieldJSONMapping
+		IntrinsicFields["body"] = *BodyLogicalFieldJSONMapping
```
srikanthccv: Here is the alternative to consider, make the FieldMapper return the logical mapping instead of directly replacing it. Would that help us avoid the custom changes we did in `metadata.go` and it's fake implementation?

pkg/telemetrymetadata/metadata.go
srikanthccv: https://github.com/SigNoz/signoz/pull/10325/changes#r2917541040

pkg/telemetrylogs/condition_builder.go
```diff
@@ -35,7 +33,7 @@ func (c *conditionBuilder) conditionFor(
 		return "", err
 	}
 
-	if column.Type.GetType() == schema.ColumnTypeEnumJSON && querybuilder.BodyJSONQueryEnabled {
+	if column.Type.GetType() == schema.ColumnTypeEnumJSON && querybuilder.BodyJSONQueryEnabled && key.FieldContext != telemetrytypes.FieldContextLog {
```
srikanthccv: This is not needed I believe given the column type is not JSON?

pkg/telemetrylogs/condition_builder.go
```diff
@@ -247,19 +251,32 @@ func (c *conditionBuilder) ConditionFor(
...
-		if slices.Contains(maps.Keys(IntrinsicFields), field) && key.FieldContext != telemetrytypes.FieldContextBody {
+	// Skip adding exists filter for intrinsic fields i.e. Table level log context fields
+	buildExistCondition := operator.AddDefaultExistsFilter()
+	switch key.FieldContext {
+	case telemetrytypes.FieldContextLog, telemetrytypes.FieldContextScope:
+		// pass; No need to build exist condition for top level columns
+		// immidiately return
+		return condition, nil
+	case telemetrytypes.FieldContextResource, telemetrytypes.FieldContextAttribute:
+		buildExistCondition = buildExistCondition && true
```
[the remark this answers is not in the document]
srikanthccv: relevant

pkg/telemetrylogs/json_stmt_builder_test.go
srikanthccv, line 1: No logical changes here, just the body alias change ✅

pkg/querybuilder/where_clause_visitor.go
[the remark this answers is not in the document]
piyushsingariya: > given the expression is valid?

Is expression actually valid if no key/value is passed? We should be throwing error that such and such case is not handled?

pkg/telemetrylogs/const.go
piyushsingariya: @srikanthccv DefaultFullTextColumn, IntrinsicFields are directly referenced externally from metadata.go and others. I am not able to imagine as of now, I'll look into it and then connect over a call.

piyushsingariya:
```
func (b *logQueryStatementBuilder) adjustKey(key *telemetrytypes.TelemetryFieldKey, keys map[string][]*telemetrytypes.TelemetryFieldKey) []string {
	// First check if it matches with any intrinsic fields
	if base, ok := IntrinsicFields[key.Name]; ok {
		// Route through the field mapper so feature-flag-driven redirects (e.g.
		// "body" → "body_v2.message" when JSON body is enabled) are applied at
		// query time without mutating shared state.
		canonical := b.fm.LogicalKeyFor(context.Background(), &base)
		return querybuilder.AdjustKey(key, keys, canonical)
	}

	return querybuilder.AdjustKey(key, keys, nil)
}
```

if I remove inplace replacement of keys, AdjustKey needs to take on the responsibility to fix/update field keys.

pkg/telemetrylogs/const.go
piyushsingariya: I'm relying on field evolution to help here, based on timestamps we would switch columns.

pkg/querybuilder/where_clause_visitor.go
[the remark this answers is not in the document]
srikanthccv: Oh, sorry, this question was meant for cursorbot.

pkg/telemetrylogs/const.go
srikanthccv: I am assuming that change will come in follow-up PR(s)?

piyushsingariya: Yes @srikanthccv

pkg/querybuilder/fallback_expr.go
nityanandagohain: This still doesn't seem to be addressed

pkg/telemetrylogs/condition_builder.go
```diff
@@ -247,19 +251,30 @@ func (c *conditionBuilder) ConditionFor(
...
-		// with an exception for body json search
-		field, _ := c.fm.FieldFor(ctx, key)
-		if slices.Contains(maps.Keys(IntrinsicFields), field) && key.FieldContext != telemetrytypes.FieldContextBody {
+	// Skip adding exists filter for intrinsic fields i.e. Table level log context fields
+	buildExistCondition := operator.AddDefaultExistsFilter()
+	switch key.FieldContext {
+	case telemetrytypes.FieldContextLog, telemetrytypes.FieldContextScope:
+		// pass; No need to build exist condition for top level columns
+		// immidiately return
+		return condition, nil
```
nityanandagohain, line 252: Why is Telemetrytypes.FieldContextScope in the case case as FieldContextLog instead of FieldContextResource and FieldContextAttribute ?

pkg/telemetrylogs/condition_builder.go
```diff
@@ -176,9 +173,16 @@ func (c *conditionBuilder) conditionFor(
 		var value any
 		switch column.Type.GetType() {
 		case schema.ColumnTypeEnumJSON:
-			if operator == qbtypes.FilterOperatorExists {
-				return sb.IsNotNull(tblFieldName), nil
-			} else {
+			switch key.FieldDataType {
+			case telemetrytypes.FieldDataTypeJSON:
+				if operator == qbtypes.FilterOperatorExists {
```
nityanandagohain: Let's add some unit tests in the condition_builder_test.go . This is a bit confusion as we are doing more based on a similar JSON column

pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
nityanandagohain: Yeah `matchesKey` should not have this logic , it's misleading. Lets find a better way of doing this.

pkg/querybuilder/fallback_expr.go
piyushsingariya: my bad, I missed committing the change

pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
piyushsingariya: It has to exist (or the check `!matchesName(selector, key.Name)` needs to go away on field keys), else mock metadata store will not let pass the alias based key, and WhereClauseVisitor will fail with not finding any key.

pkg/telemetrylogs/condition_builder.go
piyushsingariya: due to these I had added Scope to be in same bracket as Log context https://github.com/SigNoz/signoz/blob/cab4a56694b702d533063189e93be9ea32ce1ec7/pkg/telemetrylogs/const.go#L87-L99

pkg/telemetrylogs/condition_builder.go
piyushsingariya: I am removing these, since they're not in action yet.

pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
nityanandagohain: I did a simple test by chainging line of code i.e `matchNameExceptions := []string{}` and ran the entire test suite and everything passed so what are we missing ?

pkg/telemetrymetadata/metadata.go
```diff
@@ -882,12 +884,20 @@ func (t *telemetryMetaStore) GetKeys(ctx context.Context, fieldKeySelector *tele
...
 	switch fieldKeySelector.Signal {
 	case telemetrytypes.SignalTraces:
 		keys, complete, err = t.getTracesKeys(ctx, selectors)
 	case telemetrytypes.SignalLogs:
-		keys, complete, err = t.getLogsKeys(ctx, selectors)
+		mapOfLogKeys, logsComplete, err := t.getLogsKeys(ctx, selectors)
+		if err != nil {
+			return nil, false, err
+		}
+		for keyName, keys := range mapOfLogKeys {
```
nityanandagohain: can you help me with what we are doing here ? I see that this same logic 

```

	for _, key := range keys {
		mapOfKeys[key.Name] = append(mapOfKeys[key.Name], key)
	}
```

Is repetead again towards the end of the function so my question is is this really needed if we just stored in keys and we perform the same logic towards the end.

piyushsingariya: `getLogsKeys` is actually changed to retain mapping of body.message with `body` searchedText. If I collect everything as it was previously in an array, the mapping relational is lost in process.

pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
piyushsingariya: 🫨  Yes, damn I solved it the other way, StaticFields solved it, removing this fake implementation

srikanthccv:
Following https://github.com/SigNoz/signoz/blob/main/docs/contributing/query-range.md#summary-of-inviolable-rules


> "Field-to-column translation only happens in FieldMapper"
> "The four-layer abstraction stack must not be bypassed or flattened"

The `message` detail leaks across 4 layers:

1. `const.go`: `BodyFieldMessageMapping` rewrites `IntrinsicFields["body"]` from `{FieldContext: Log}` to `{Name: "message", FieldContext: Body, Materialized: true}`. The problem here is that `body` is a logical intrinsic field which shouldn't be re-written with the storage layer specific field.
2. `condition_builder.go`: `!key.Materialized` check was out of the place, that it shouldn't do branching for condition based on the materialized for field. 
3. `metadata.go`: Alias handling (`key != field.Name`) leaks the details into metadata layer
4. `mock metadata_store.go`: Same alias handling in the fake



`body` is an intrinsic log field. When `BodyJSONQueryEnabled` is true, its storage location changes from `body` to `body_v2.message`. This is a field-to-column translation concern and it belongs exclusively in the FieldMapper.

Please take a look at the https://github.com/SigNoz/signoz/pull/10606/changes/c2c2678125de6e4ec358111ece14c78759b24dd1 which keeps the `message` detail to logs field mapper only and leaves the rest unchanged.

PR #10327, opened 2026-02-17
feat: improve root user provisioning and restore users unique index

karanbalani:
### 📄 Summary

- Fix dropped unique index on users(email, org_id): Migration 064 (add_root_user) adds a non-nullable `is_root` column, which on SQLite triggers a full table recreation. The recreation logic only preserves constraints embedded in the CREATE TABLE definition, but the (email, org_id) unique index was a standalone CREATE UNIQUE INDEX (from migration 042), so it was silently dropped. Added migration 065 to restore it.
- Restructure root user config: Replaced flat org_name field with a nested Org.Name config to better organize root org settings.
- Consolidate NewOrganization: Merged NewOrganization and NewOrganizationWithName into a single NewOrganization(displayName, name) function, removing the duplicate.
- Prevent password reset for root user: Added a guard in GetOrCreateResetPasswordToken to block password resets for the root user.

#### Issues closed by this PR
Closes: https://github.com/SigNoz/platform-pod/issues/1076

### DIFF
--- a/pkg/modules/user/config.go
+++ b/pkg/modules/user/config.go
@@ -15,13 +15,13 @@ type Config struct {
 }
 
 type RootConfig struct {
-	Enabled  bool          `mapstructure:"enabled"`
-	Email    valuer.Email  `mapstructure:"email"`
-	Password string        `mapstructure:"password"`
-	Org      RootOrgConfig `mapstructure:"org"`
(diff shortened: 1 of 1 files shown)


karanbalani:
Closes: https://github.com/SigNoz/platform-pod/issues/1076

karanbalani:
Index check:
image

PR #10337, opened 2026-02-18
chore: metrics explorer v2 api migration in summary page

amlannandy:
## Pull Request
---

### 📄 Summary

- Update the Metrics Explorer Summary section to use the V2 APIs for displaying the metrics list table and treemap
- Replace the search bar to use the new one
- Display search by metric type as the V2 api doesn't support it yet
- Minor UX updates

https://github.com/user-attachments/assets/c1dd64fc-2aa2-4cb4-9c4a-7812118424f8

Closes https://github.com/SigNoz/engineering-pod/issues/3497

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Updated the tests
- Manual testing

---

We should check the following to ensure nothing is breaking:
- Loading and displaying list and treemap
- Filters are applied correctly
- Order by is working
- No uneccesary refetching

---

### 📋 Checklist
- [x] Manually tested

---


frontend/src/container/MetricsExplorer/Summary/MetricNameSearch.tsx
```diff
@@ -67,9 +68,12 @@ function MetricNameSearch({
 
 	const handleSelect = useCallback(
 		(selectedMetricName: string): void => {
+			const queryFilters = convertExpressionToFilters(
+				queryFilterExpression?.expression,
```
AshwinBhatkal: Is the undefined check required here?

frontend/src/container/MetricsExplorer/Summary/utils.tsx
```diff
@@ -3,14 +3,16 @@ import { Color } from '@signozhq/design-tokens';
...
+} from 'api/generated/services/sigNoz.schemas';
+import {
 	MetricsListPayload,
 	MetricType,
 } from 'api/metricsExplorer/getMetricsList';
-import {
-	SamplesData,
-	TimeseriesData,
-} from 'api/metricsExplorer/getMetricsTreeMap';
+import { Filter } from 'api/v5/v5';
```
AshwinBhatkal, line 10: No type in generated types?

frontend/src/container/MetricsExplorer/Summary/utils.tsx
```diff
@@ -143,6 +151,60 @@ export function MetricTypeRenderer({
 	);
 }
 
+export function MetricTypeRendererV2({
```
AshwinBhatkal: can you changes this files to utils.ts and move these components to different files?

frontend/src/container/MetricsExplorer/Summary/Summary.tsx
```diff
@@ -254,71 +268,81 @@ function Summary(): JSX.Element {
...
 		logEvent(MetricsExplorerEvents.OrderByApplied, {
 			[MetricsExplorerEventKeys.Tab]: 'summary',
-			[MetricsExplorerEventKeys.ColumnName]: orderBy.columnName,
-			[MetricsExplorerEventKeys.Order]: orderBy.order,
+			[MetricsExplorerEventKeys.ColumnName]: orderBy.key?.name,
+			[MetricsExplorerEventKeys.Order]: orderBy.direction,
 		});
 	};
 
 	const isMetricsListDataEmpty = useMemo(
```
AshwinBhatkal: useMemo's in this file are an overkill. Since all these are primitives, we can stay without useMemo

frontend/src/container/MetricsExplorer/Summary/MetricTypeSearch.tsx
```diff
@@ -26,50 +22,45 @@ function MetricTypeSearch({
...
 		(selectedMetricType: string): void => {
+			let newFilters;
 			if (selectedMetricType !== 'all') {
-				const newFilters = {
+				newFilters = {
 					items: [
 						...queryFilters.items,
 						{
-							id: 'metric_type',
+							id: 'ttype',
```
AshwinBhatkal: is this correct?

frontend/src/container/MetricsExplorer/Summary/MetricsTreemap.tsx
```diff
@@ -54,62 +55,36 @@ function MetricsTreemap({
 	const xMax = treemapWidth - TREEMAP_MARGINS.LEFT - TREEMAP_MARGINS.RIGHT;
 	const yMax = TREEMAP_HEIGHT - TREEMAP_MARGINS.TOP - TREEMAP_MARGINS.BOTTOM;
 
-	if (isLoading) {
-		return (
-			<div data-testid="metrics-treemap-loading-state">
-				<Skeleton
-					style={{ width: treemapWidth, height: TREEMAP_HEIGHT + 55 }}
-					active
+	const treemapContent = useMemo(() => {
```
AshwinBhatkal: overkill of useMemo here. This is not required

AshwinBhatkal: render it inline or create a new component. useMemo is not the solution here

frontend/src/container/MetricsExplorer/Summary/MetricsSearch.tsx
```diff
@@ -1,27 +1,59 @@
...
+import { DataSource } from 'types/common/queryBuilder';
 
 import { MetricsSearchProps } from './types';
 
 function MetricsSearch({ query, onChange }: MetricsSearchProps): JSX.Element {
+	const [currentExpression, setCurrentExpression] = useState<string>(
+		query?.filter?.expression || '',
+	);
+
+	useEffect(() => {
```
AshwinBhatkal: This is not right. If we need to do this, then something else is wrong. Either UX or our code

AshwinBhatkal: Think about how this useEffect should not exist. If not, we can discuss

frontend/src/container/MetricsExplorer/Summary/utils.tsx
amlannandy: So, this is the type used by the exisitng query search bar. So using that only

frontend/src/container/MetricsExplorer/Summary/utils.tsx
amlannandy: Sure can do that

frontend/src/container/MetricsExplorer/Summary/MetricTypeSearch.tsx
amlannandy: Ah yeah this should be `type`. We have disabled/commented out this component as the new API doesn't support it for now. So didn't catch this during testing. Would have to revisit this component when we re-enable it, but will fix this type now itself

frontend/src/container/MetricsExplorer/Summary/MetricsSearch.tsx
amlannandy: Hmm yeah that makes sense. I think I added this useEffect very late on to solve a specific issue so I missed it. But yeah, just moving the useState one level up should have done the trick. Lemme clean this up.

frontend/src/container/MetricsExplorer/Summary/utils.tsx
amlannandy: Moved the 2 MetricTypeRenderer compoents to different files. Some of the other util functions in the file still return JSX, so had to keep the extension as .tsx

frontend/src/container/MetricsExplorer/Summary/MetricsSearch.tsx
```diff
@@ -1,27 +1,50 @@
...
 
-function MetricsSearch({ query, onChange }: MetricsSearchProps): JSX.Element {
+function MetricsSearch({
+	query,
+	onChange,
+	currentQueryFilterExpression,
+	setCurrentQueryFilterExpression,
+	isLoading,
+}: MetricsSearchProps): JSX.Element {
+	const handleOnChange = (expression: string): void => {
```
AshwinBhatkal: Few common things, if a function is being passed as a prop, use useCallback. This is a must

Same goes for below

Do not pass style object to a component. Either use className or memoize the style object. Almost in similar lines to the above

frontend/src/container/MetricsExplorer/Summary/MetricsTreemap.tsx
```diff
@@ -12,21 +13,20 @@ import {
...
 import {
 	getTreemapTileStyle,
 	getTreemapTileTextStyle,
 	transformTreemapData,
 } from './utils';
 
-function MetricsTreemap({
-	viewType,
-	data,
+function TreemapContent({
```
AshwinBhatkal: So this component would either be MetricsTreemap or MetricsTreemapInternal

When MetricsTreemap, parent is MetricsTreemapWrapper
When MetricsTreemapInternal, parent is MetricsTreemap

frontend/src/container/MetricsExplorer/Summary/MetricTypeViewRenderer.tsx
```diff
@@ -0,0 +1,79 @@
...
+	const metricTypeRendererStyle = useMemo(
+		() => ({
+			backgroundColor: `${color}33`,
+			border: `1px solid ${color}`,
+			color,
+		}),
+		[color],
+	);
+
+	const metricTypeRendererTextStyle = useMemo(
```
AshwinBhatkal: perf-improvement: you could return an object from useMemo for the styles instead of 2 different useMemos

frontend/src/container/MetricsExplorer/Summary/utils.tsx
```diff
@@ -55,7 +47,11 @@ export const getMetricsTableColumns = (
 		title: (
 			<div className="metric-type-column-header">
 				<span className="metric-type-column-header-text">TYPE</span>
-				<MetricTypeSearch queryFilters={queryFilters} />
+				{/* TODO: @amlannandy: Re-enable once API supports metric type filtering */}
```
AshwinBhatkal: THese things to re enable need not be in code unless it's fast follow up

frontend/src/container/QueryBuilder/components/RunQueryBtn/RunQueryBtn.tsx
```diff
@@ -28,6 +29,7 @@ function RunQueryBtn({
 	handleCancelQuery,
 	onStageRunQuery,
 	queryRangeKey,
+	disabled,
```
AshwinBhatkal: If we are passing this in, we should pass the other 2 values `isLoading` and  `onStageRunQuery` from the parent and have only the disabled prop control the disabled value and the other 2 values that can be seen below

frontend/src/container/MetricsExplorer/Summary/utils.tsx
amlannandy: Okay will remove it for now then

amlannandy: Don't think backend is picking this up immediately

frontend/src/container/QueryBuilder/components/RunQueryBtn/RunQueryBtn.tsx
amlannandy: Hmm that makes sense. I was wondering if we should do this in a separate PR as the component is being used at multiple places, so we need to review and test all the usages

AshwinBhatkal: Then, let us not do any change in this PR

frontend/src/container/MetricsExplorer/MetricDetails/Metadata.tsx
```diff
@@ -24,7 +24,7 @@ import { useNotifications } from 'hooks/useNotifications';
 import { Edit2, Save, X } from 'lucide-react';
 
 import { MetricsExplorerEventKeys, MetricsExplorerEvents } from '../events';
-import { MetricTypeViewRenderer } from '../Summary/utils';
+import MetricTypeRendererV2 from '../Summary/MetricTypeViewRenderer';
```
AshwinBhatkal, line 27: V2?

frontend/src/container/MetricsExplorer/Summary/utils.tsx
```diff
@@ -56,7 +47,10 @@ export const getMetricsTableColumns = (
 		title: (
 			<div className="metric-type-column-header">
 				<span className="metric-type-column-header-text">TYPE</span>
-				<MetricTypeSearch queryFilters={queryFilters} />
+				{/* <MetricTypeSearch
```
AshwinBhatkal, line 50: I meant remove this piece entirely 😄 not just the comment

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
```diff
@@ -85,6 +85,7 @@ interface QuerySearchProps {
 	signalSource?: string;
 	hardcodedAttributeKeys?: QueryKeyDataSuggestionsProps[];
 	onRun?: (query: string) => void;
+	showFilterSuggestionsWithoutMetric?: boolean;
```
AshwinBhatkal, line 88: info - This change is there in this PR as well - https://github.com/SigNoz/signoz/pull/10342/changes

PR #10341, opened 2026-02-18
feat: Migrate Pipelines Filter from V3 -> V5

piyushsingariya:
## Pull Request

---

### 📄 Summary
This PR aims to enhance user experience with Logs Pipelines with replacing Filter Input from QB V3 to QB V5
- Migrate v3 filters to v5 filters
- Store v3 filters in separate column for safe migration
- AgentConfig is not updated when migration is done, so editing pipelines is definitely required
- Includes Parser Util for v5 expression
- This PR **doesn't change** anything about usage of `attributes` in FROM/TO fields

Closes signoz/engineering-pod#3834
Closes signoz/engineering-pod#4263

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered:

---

- Blast radius: All Tenants impacted
- Potential regressions:
- Rollback plan: N/A (yet)

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/pkg/sqlmigration/066_migrate_pipeline_filters_v5.go
+++ b/pkg/sqlmigration/066_migrate_pipeline_filters_v5.go
@@ -10,7 +10,7 @@ import (
 	"github.com/SigNoz/signoz/pkg/factory"
 	v3 "github.com/SigNoz/signoz/pkg/query-service/model/v3"
 	"github.com/SigNoz/signoz/pkg/query-service/queryBuilderToExpr"
-	"github.com/SigNoz/signoz/pkg/sqlstore"
+	"github.com/SigNoz/signoz/pkg/sqlschema"
 	"github.com/SigNoz/signoz/pkg/transition"
 	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
 	"github.com/uptrace/bun"
@@ -18,17 +18,17 @@ import (
 )
 
 type migratePipelineFiltersV5 struct {
-	store  sqlstore.SQLStore
-	logger *slog.Logger
+	sqlschema sqlschema.SQLSchema
+	logger    *slog.Logger
 }
 
 func NewMigratePipelineFiltersV5Factory(
-	store sqlstore.SQLStore,
+	sqlschema sqlschema.SQLSchema,
 ) factory.ProviderFactory[SQLMigration, Config] {
 	return factory.NewProviderFactory(
 		factory.MustNewName("migrate_pipeline_filters_v5"),
 		func(ctx context.Context, ps factory.ProviderSettings, c Config) (SQLMigration, error) {
-			return newMigratePipelineFiltersV5(ctx, c, store, ps.Logger)
+			return newMigratePipelineFiltersV5(ctx, c, sqlschema, ps.Logger)
 		},
 	)
 }
@@ -36,7 +36,7 @@ func NewMigratePipelineFiltersV5Factory(
 func newMigratePipelineFiltersV5(
 	_ context.Context,
 	_ Config,
-	store sqlstore.SQLStore,
+	sqlschema sqlschema.SQLSchema,
 	logger *slog.Logger,
 ) (SQLMigration, error) {
 	if logger == nil {
@@ -44,8 +44,8 @@ func newMigratePipelineFiltersV5(
 	}
 
 	return &migratePipelineFiltersV5{
-		store:  store,
-		logger: logger,
+		sqlschema: sqlschema,
+		logger:    logger,
 	}, nil
 }
 
@@ -64,6 +64,11 @@ type pipelineFilterRow struct {
 }
 
 func (migration *migratePipelineFiltersV5) Up(ctx context.Context, db *bun.DB) error {
+	table, uniqueConstraints, err := migration.sqlschema.GetTable(ctx, "pipelines")
+	if err != nil {
+		return err
+	}
+
 	tx, err := db.BeginTx(ctx, nil)
 	if err != nil {
 		return err
@@ -74,13 +79,21 @@ func (migration *migratePipelineFiltersV5) Up(ctx context.Context, db *bun.DB) e
 	}()
 
 	// 1. Rename existing filter column to filter_deprecated
-	if _, err := migration.store.Dialect().RenameColumn(ctx, tx, "pipelines", "filter", "filter_deprecated"); err != nil {
-		return err
+	for _, sql := range migration.sqlschema.Operator().RenameColumn(table, &sqlschema.Column{Name: "filter"}, "filter_deprecated") {
+		if _, err := tx.ExecContext(ctx, string(sql)); err != nil {
+			return err
+		}
 	}
 
 	// 2. Add new filter column (v5); existing rows get default ''
-	if err := migration.store.Dialect().AddColumn(ctx, tx, "pipelines", "filter", "TEXT NOT NULL DEFAULT ''"); err != nil {
-		return err
+	for _, sql := range migration.sqlschema.Operator().AddColumn(table, uniqueConstraints, &sqlschema.Column{
+		Name:     "filter",
+		DataType: sqlschema.DataTypeText,
+		Nullable: false,
+	}, "") {
+		if _, err := tx.ExecContext(ctx, string(sql)); err != nil {
+			return err
+		}
 	}
 
 	// 3. Copy v5 filter data: read from filter_deprecated, migrate to v5 expression, write to filter
@@ -133,6 +146,11 @@ func (migration *migratePipelineFiltersV5) Up(ctx context.Context, db *bun.DB) e
 }
 
 func (migration *migratePipelineFiltersV5) Down(ctx context.Context, db *bun.DB) error {
+	table, _, err := migration.sqlschema.GetTable(ctx, "pipelines")
+	if err != nil {
+		return err
+	}
+
 	tx, err := db.BeginTx(ctx, nil)
 	if err != nil {
 		return err
@@ -143,13 +161,17 @@ func (migration *migratePipelineFiltersV5) Down(ctx context.Context, db *bun.DB)
 	}()
 
 	// 1. Drop the new filter column
-	if err := migration.store.Dialect().DropColumn(ctx, tx, "pipelines", "filter"); err != nil {
-		return err
+	for _, sql := range migration.sqlschema.Operator().DropColumn(table, &sqlschema.Column{Name: "filter"}) {
+		if _, err := tx.ExecContext(ctx, string(sql)); err != nil {
+			return err
+		}
 	}
 
 	// 2. Rename filter_deprecated back to filter
-	if _, err := migration.store.Dialect().RenameColumn(ctx, tx, "pipelines", "filter_deprecated", "filter"); err != nil {
-		return err
+	for _, sql := range migration.sqlschema.Operator().RenameColumn(table, &sqlschema.Column{Name: "filter_deprecated"}, "filter") {
+		if _, err := tx.ExecContext(ctx, string(sql)); err != nil {
+			return err
+		}
 	}
 
 	if err := tx.Commit(); err != nil {
--- a/pkg/signoz/provider.go
+++ b/pkg/signoz/provider.go
@@ -169,7 +169,7 @@ func NewSQLMigrationProviderFactories(
 		sqlmigration.NewAddAnonymousPublicDashboardTransactionFactory(sqlstore),
 		sqlmigration.NewAddRootUserFactory(sqlstore, sqlschema),
 		sqlmigration.NewAddUserEmailOrgIDIndexFactory(sqlstore, sqlschema),
-		sqlmigration.NewMigratePipelineFiltersV5Factory(sqlstore),
+		sqlmigration.NewMigratePipelineFiltersV5Factory(sqlschema),
 	)
 }
 
--- a/pkg/sqlschema/column.go
+++ b/pkg/sqlschema/column.go
@@ -118,6 +118,19 @@ func (column *Column) ToUpdateSQL(fmter SQLFormatter, tableName TableName, value
 	return sql
 }
 
+func (column *Column) ToRenameSQL(fmter SQLFormatter, tableName TableName, newName ColumnName) []byte {
+	sql := []byte{}
+
+	sql = append(sql, "ALTER TABLE "...)
+	sql = fmter.AppendIdent(sql, string(tableName))
+	sql = append(sql, " RENAME COLUMN "...)
+	sql = fmter.AppendIdent(sql, string(column.Name))
+	sql = append(sql, " TO "...)
+	sql = fmter.AppendIdent(sql, string(newName))
+
+	return sql
+}
+
 func (column *Column) ToSetNotNullSQL(fmter SQLFormatter, tableName TableName) []byte {
 	sql := []byte{}
 
--- a/pkg/sqlschema/operator.go
+++ b/pkg/sqlschema/operator.go
@@ -95,6 +95,19 @@ func (operator *Operator) DropColumn(table *Table, column *Column) [][]byte {
 	return [][]byte{column.ToDropSQL(operator.fmter, table.Name, operator.support.ColumnIfNotExistsExists)}
 }
 
+func (operator *Operator) RenameColumn(table *Table, column *Column, newName ColumnName) [][]byte {
+	index := operator.findColumnByName(table, column.Name)
+	// If the column does not exist, we do not need to rename it.
+	if index == -1 {
+		return [][]byte{}
+	}
+
+	sql := column.ToRenameSQL(operator.fmter, table.Name, newName)
+	table.Columns[index].Name = newName
+
+	return [][]byte{sql}
+}
+
 func (operator *Operator) DropConstraint(table *Table, uniqueConstraints []*UniqueConstraint, constraint Constraint) [][]byte {
 	// The name of the input constraint is not guaranteed to be the same as the name of the constraint in the database.
 	// So we need to find the constraint in the database and drop it.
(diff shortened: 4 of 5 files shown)


pkg/querybuilder/filter_expr_parse.go
```diff
@@ -0,0 +1,390 @@
...
+)
+
+// ParseFilterExpr parses a v5 filter expression and returns a logical
+// tree that preserves full boolean structure (AND/OR, parentheses, NOT) as
+// well as the individual conditions (keys, operator, values).
+//
+// This can be reused by callers that need deeper introspection into the logic
+// of the filter expression without constructing any SQL or query-engine
+// specific structures.
+func ParseFilterExpr(expr string) (*qbtypes.FilterExprNode, error) {
```
nityanandagohain: @srikanthccv  some suggestion here will be great, here we are building a parser that parses into a different structure to convert it to expr language. Wanted your thoughts on if this is the correct way to go.

pkg/sqlmigration/066_migrate_pipeline_filters_v5.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+
+		var filterSet v3.FilterSet
+		if err := json.Unmarshal([]byte(raw), &filterSet); err != nil {
+			return err
+		}
+
+		// Attempt to treat the existing JSON as a v3 FilterSet and build a v5
+		// expression from it, using the normalized payload.
+		expr, migrated, err := transition.BuildFilterExpressionFromFilterSet(ctx, migration.logger, "logs", &filterSet)
```
nityanandagohain, line 118: will be better to send it to some telemetry with before and after states.

pkg/types/querybuildertypes/querybuildertypesv5/filter_expr.go
```diff
@@ -0,0 +1,66 @@
+package querybuildertypesv5
+
+import "github.com/SigNoz/signoz/pkg/types/telemetrytypes"
+
+// LogicalOp represents how child expressions are combined.
```
nityanandagohain: this comment  feels misleading, along with "LEAF" representing a operator.

pkg/types/querybuildertypes/querybuildertypesv5/filter_expr.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (f *FilterExprNode) Flatten() []FilterCondition {
+	var conditions []FilterCondition
+	var walk func(node *FilterExprNode)
+
+	walk = func(node *FilterExprNode) {
+		if node == nil {
+			return
+		}
+		if node.Op == LogicalOpLeaf {
```
nityanandagohain: empty children representing a leaf will be better I guess.

pkg/types/pipelinetypes/postable_pipeline_test.go
```diff
@@ -70,19 +59,21 @@ func TestIsValidPostablePipeline(t *testing.T) {
...
-							Key: v3.AttributeKey{
-								Key:      "method",
-								DataType: v3.AttributeKeyDataTypeString,
-								Type:     v3.AttributeKeyTypeUnspecified,
-							},
-							Operator: "regex",
-							Value:    "[0-9A-Z*",
-						},
-					},
+				Filter: &qbtypes.Filter{
```
nityanandagohain, line 62: so I see that you added a change to `p.Filter.Expression == ""` . But the aim of this test was to check what happens if I actually send a wrong filter. So maybe add one more tests with an actual invalid filter expression ?

pkg/query-service/app/logparsingpipeline/controller.go
nityanandagohain, line 134: I lost the context here, I remember discussing about doing this from the charts repo as it is a global config change.

one of the reason was that we can easily modify the config for a instance if required, what decision did we take there ?

pkg/query-service/queryBuilderToExpr/queryBuilderToExpr.go
```diff
@@ -15,133 +20,246 @@ var (
...
+	qbtypes.FilterOperatorGreaterThanOrEq: ">=",
+	qbtypes.FilterOperatorContains:        "contains",
+	qbtypes.FilterOperatorNotContains:     "not contains",
+	qbtypes.FilterOperatorRegexp:          "matches",
+	qbtypes.FilterOperatorNotRegexp:       "not matches",
+	qbtypes.FilterOperatorIn:              "in",
+	qbtypes.FilterOperatorNotIn:           "not in",
+	qbtypes.FilterOperatorExists:          "in",
+	qbtypes.FilterOperatorNotExists:       "not in",
+	// nlike and like are not supported yet
```
nityanandagohain: Comment need to be updated	
```
FilterOperatorLike
FilterOperatorNotLike
FilterOperatorILike
FilterOperatorNotILike

Between
```
we added ilike as well in new qbv5

pkg/query-service/queryBuilderToExpr/queryBuilderToExpr.go
```diff
@@ -15,133 +20,246 @@ var (
...
 }
 
-func getName(v v3.AttributeKey) string {
-	if v.Type == v3.AttributeKeyTypeTag {
-		return fmt.Sprintf(`attributes["%s"]`, v.Key)
-	} else if v.Type == v3.AttributeKeyTypeResource {
-		return fmt.Sprintf(`resource["%s"]`, v.Key)
+func getName(key *telemetrytypes.TelemetryFieldKey) string {
+	if key == nil {
```
nityanandagohain, line 45: with qbv5 will we ever reach this case ?

pkg/query-service/queryBuilderToExpr/queryBuilderToExpr.go
```diff
@@ -15,133 +20,246 @@ var (
...
+	}
 
-		case v3.FilterOperatorExists, v3.FilterOperatorNotExists:
-			// accustom log filters like `body.log.message EXISTS` into EXPR language
-			// where User is attempting to check for keys present in JSON log body
-			if strings.HasPrefix(v.Key.Key, "body.") {
-				// if body is a string and is a valid JSON, then check if the key exists in the JSON
-				filter = fmt.Sprintf(`((type(body) == "string" && isJSON(body)) && %s %s %s)`, exprFormattedValue(strings.TrimPrefix(v.Key.Key, "body.")), logOperatorsToExpr[v.Operator], "fromJSON(body)")
+	for _, condition := range node.Flatten() {
+		if condition.Key == nil {
```
nityanandagohain: Seems impossible to reach here.

pkg/query-service/queryBuilderToExpr/queryBuilderToExpr.go
```diff
@@ -15,133 +20,246 @@ var (
...
-			} else if typ := getTypeName(v.Key.Type); typ != "" {
-				filter = fmt.Sprintf("%s %s %s", exprFormattedValue(v.Key.Key), logOperatorsToExpr[v.Operator], typ)
-			} else {
-				// if type of key is not available; is considered as TOP LEVEL key in OTEL Log Data model hence
-				// switch Exist and Not Exists operators with NOT EQUAL and EQUAL respectively
-				operator := v3.FilterOperatorNotEqual
-				if v.Operator == v3.FilterOperatorNotExists {
-					operator = v3.FilterOperatorEqual
+		_, found := telemetrylogs.IntrinsicFields[condition.Key.Name]
+		if condition.Key.FieldContext == telemetrytypes.FieldContextUnspecified && !found {
```
nityanandagohain: my ref:- this should be job of the QB to expr parser ?

pkg/query-service/queryBuilderToExpr/queryBuilderToExpr.go
```diff
@@ -15,133 +20,246 @@ var (
...
+			}
+		}
+	}
+
+	return nodeToExpr(node)
+}
 
-				filter = fmt.Sprintf("%s %s nil", v.Key.Key, logOperatorsToExpr[operator])
+func nodeToExpr(node *qbtypes.FilterExprNode) (string, error) {
+	if node == nil {
```
nityanandagohain: not possible.

pkg/querybuilder/filter_expr_parse.go
srikanthccv: sorry missed this @mention, let me go through it.

pkg/querybuilder/filter_expr_parse.go
```diff
@@ -0,0 +1,390 @@
...
+	if strings.TrimSpace(expr) == "" {
+		return qbtypes.NewEmptyFilterExprNode(), nil
```
nityanandagohain: nit: If we treat empty as invalid input, shouldn't this itself return error of invalid input ?

Basically no error -> atleast one conditon.

Let me know your thoughts.

pkg/query-service/queryBuilderToExpr/queryBuilderToExpr_test.go
```diff
@@ -267,248 +275,241 @@ func TestExpressionVSEntry(t *testing.T) {
...
-				{Key: v3.AttributeKey{Key: "env", DataType: v3.AttributeKeyDataTypeString, Type: v3.AttributeKeyTypeResource}, Value: "never", Operator: "="},
-			}},
+			Name: "no matches (resource value not present)",
+			Query: &qbtypes.Filter{
+				Expression: "resource.env = 'never'",
+			},
 			ExpectedMatches: []int{},
 		},
 		{
 			Name: "attribute equal and trace_id exists",
```
nityanandagohain, line 509: seems ` Multi-filter variations` tests are not really adding any value ? I don't think they will help us catch something extra which the first ones will miss.

let me know your thoughts here

pkg/querybuilder/filter_expr_parse_test.go
```diff
@@ -0,0 +1,355 @@
...
+						Key:   fk(t, "attributes.host.name"),
+						Op:    qbtypes.FilterOperatorEqual,
+						Value: "frontend",
+					},
+				},
+			},
+		},
+		{
+			name: "AND with BETWEEN numeric range and NOT boolean comparison",
+			expr: "attributes.status_code BETWEEN 500 AND 599 AND NOT attributes.is_error = true",
```
nityanandagohain: not really helpful, we don't support between

pkg/querybuilder/filter_expr_parse.go
piyushsingariya: in Pipelines V2, We'll support No filters too

nityanandagohain: Okay then you will add a filter to select all logs maybe if the filter is empty right ?

But what I meant here was that throwing error and adding support for it can be done here itself, instead of handling empty node again in the functions which is calling this. But again I am nitpicking here .

pkg/sqlmigration/066_migrate_pipeline_filters_v5.go
```diff
@@ -0,0 +1,135 @@
...
+
+	if err := tx.Commit(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (migration *migratePipelineFiltersV5) Down(ctx context.Context, db *bun.DB) error {
+	// Not reversible: v3 FilterSet structure is lost once we convert to a v5 expression.
```
nityanandagohain: suggestion: create a new column, add the new modified filter. rename the old column to something that is not used but will help if something goes wrong and will avoid the guesswork needed.

We can drop the old column after few releases.

pkg/query-service/queryBuilderToExpr/queryBuilderToExpr.go
piyushsingariya: True, but a function should always assess their parameters

pkg/query-service/queryBuilderToExpr/queryBuilderToExpr.go
piyushsingariya: True, but a function should always assess their parameters

pkg/query-service/queryBuilderToExpr/queryBuilderToExpr.go
piyushsingariya: Nope, it can not be a job of "parser", Validation is job of the user of parser cuz they only know why are they using it

pkg/query-service/queryBuilderToExpr/queryBuilderToExpr.go
piyushsingariya: True, but a function should always assess their parameters

pkg/query-service/queryBuilderToExpr/queryBuilderToExpr.go
nityanandagohain: but then it's handled incorrectly right where ever getName is used ?

piyushsingariya:
> someone actually editing the pipeline will generate it ?

Yes, editing is required

nityanandagohain:
let's add these important details in the PR description, Unless someone actually goes though all the changes it's not really clear if it will actually update the existing pipelines.

It this case the migration code doesn't update the collector pipeline ( which I thought was the case previously), just the pipeline DB unless user edit's or adds new pipeline

pkg/query-service/queryBuilderToExpr/queryBuilderToExpr.go
piyushsingariya: Plus this is needed when Filter is actually empty

pkg/sqlmigration/066_migrate_pipeline_filters_v5.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		_ = tx.Rollback()
+	}()
+
+	// 1. Rename existing filter column to filter_deprecated
+	if _, err := migration.store.Dialect().RenameColumn(ctx, tx, "pipelines", "filter", "filter_deprecated"); err != nil {
```
vikrantgupta25: please check the recent migrations and use `sqlschema.Operator()`, using Dialect is deprecated.

pkg/types/pipelinetypes/pipeline.go
```diff
@@ -218,6 +218,14 @@ func (p *PostablePipeline) IsValid() error {
 	}
 
 	// check the filter
+	if p.Filter == nil || strings.TrimSpace(p.Filter.Expression) == "" {
```
vikrantgupta25, line 221: Can we move all of this to `Unmarshal` ?

vikrantgupta25 (changes requested):

pkg/sqlmigration/066_migrate_pipeline_filters_v5.go
piyushsingariya: I'll address these after I've received review from @tushar-signoz

pkg/query-service/app/logparsingpipeline/controller.go
piyushsingariya: We decided to go ahead with this temporarily, we'll make all these configurable in pipelinesv2

pkg/querybuilder/filter_expr_parse.go
srikanthccv, line 1: At first look, it looked appealing and seemed reasonable. After going through it in detail, I believe we shouldn't add this. Here is why.

If I understand the flow correctly, it is

- query string
- `querybuilder.ParseFilterExpr`
- `qbtypes.FilterExprNode` tree
- `nodeToExpr(...)`
- expr language string.

Several issues
 
1. At the core, `FilterExprNode` is a degraded AST and poor copy of the ANTLR parse tree (covered below with example and additionally in other points). It's an unnecessary intermediate representation that creates more problems than it solves. The ANTLR grammar already produces a structured AST that supports everything (comparisons, full text, functions, all operators, nested boolean logic). The `FilterExprNode` silently skips some nodes by returning nil, which then the parent AND/OR nodes silently skip.

    - `attribute.method = 'GET' AND has(...)` would silently become just `attribute.method = 'GET'`, which is incorrect.
    - Any fulltext search conditions skipped silently.
    - etc...
2. Let's look at the `Flatten()`, this used before expression preparation to validate that there is an explicit context prefixes and valid regex patterns. But it doesn't have any boolean structure (AND/OR/NOT) to it, so it only works for validation that doesn't care about logic. All the validation done with `Flatten` should just be inline. Unless I missing something here behind it's introduction.
    - "can be reused by callers that need deeper introspection into the logic", this doesn't hold because the only consumer here is `queryBuilderToExpr` and everything is designed around the `queryBuilderToExpr`
3. The types
    - Leaf nodes use `Conditions []FilterCondition`, but every leaf produced by the parser has exactly one condition. The slice is never populated with more than one item.
    - Negated is a flag on the ANTLR node rather than being modeled as a separate NOT node, which means `nodeToExpr` has to handle it as a special case at every level
4. The need for a second visitor to do anything useful
    - Now, the question is, what value does it add? I don't see anything that this new addition solves that otherwise couldn't be done inline with the visitor for the `expr` string.

So in summary, the `FilterExprNode` is a leaky abstraction (it poorly mirrors the ANTLR AST, loses information, and then requires another tree walk, adds an indirection with no benefit (unless I misunderstood something)). The current state of `FilterExprNode` doesn't belong under `pkg/querybuilder` because it lossy and cannot be relied upon for any consumer other than pipelines. We could attempt to make it reliable but then we just end up doing exactly what the ANTLR does, probably done poorly and prone to bugs. Now, does it make sense to move the `FilterExprNode` to pipelines codebase? No, I don't see any benefit it adds. We should write a visitor over the ANTLR parse tree that produces the target output for `expr` (or error based on it's support for expr capabilities) directly.

pkg/types/pipelinetypes/pipeline.go
piyushsingariya: It's not related to unmarshaling exactly. I feel it's at correct location considering it's validation function

PR #10350, opened 2026-02-18
chore: support cmd click on all clickable items

YounixM:
## Pull Request

---

Users expect Cmd+Click (Mac) / Ctrl+Click (Windows/Linux) on clickable items to open the target in a new browser tab - a universal web convention. Currently, SPA navigation via `safeNavigate` always replaces the current page, ignoring modifier keys. Several components (e.g. `DashboardsList`, `ListAlert`) had ad-hoc `event.metaKey || event.ctrlKey` checks with `window.open`, leading to duplicated logic and inconsistent coverage.

This PR centralizes modifier-key detection in `useSafeNavigate` by accepting an optional `event` parameter. When a click event is provided and Cmd/Ctrl is held, the hook automatically opens the target in a new tab instead of performing SPA navigation. A shared `utils/navigation.ts` module provides the low-level helpers (`isModifierKeyPressed`, `openInNewTab`, `navigateToPage`) for cases outside the hook.

50 files are updated to thread the mouse event from click handlers into `safeNavigate`, covering: homepage service tables, infra monitoring K8s lists (Pods, Nodes, Clusters, Deployments, etc.), hosts list, alert rules, dashboards list, metrics explorer, side nav, onboarding flows, messaging queues, breadcrumbs, and more. Two files (`DashboardsList`, `ListAlert`) that had duplicated manual Cmd/Ctrl detection were refactored to use the centralized approach.

N/A - behavioral change: hold Cmd/Ctrl while clicking any navigable row, link, or button to open in a new tab.

#### Issues closed by this PR

---

### Change Type
_Select all that apply_

- [x] Feature
- [ ] Bug fix
- [x] Refactor
- [ ] Infra / Tooling
- [x] Test-only

---

### Bug Context
N/A - this is a new feature / UX enhancement, not a bug fix.

---

- Tests added/updated:
  - `src/utils/__tests__/navigation.test.ts` - 15 tests covering `isModifierKeyPressed`, `openInNewTab`, and `navigateToPage` with various modifier key combinations, edge cases (both metaKey and ctrlKey), and fallback to SPA navigation.
  - `src/hooks/__tests__/useSafeNavigate.test.ts` - 9 tests verifying the mock contract accepts `event` and `newTab` options, and that the `shouldOpenNewTab` decision logic correctly combines explicit `newTab` with modifier-key detection.
- Manual verification: Cmd+Click on service rows, dashboard list items, alert rule links, K8s resource tables, side nav items, and breadcrumbs all open in a new tab. Normal clicks continue to use SPA navigation.
- Edge cases covered: Both `metaKey` (Mac) and `ctrlKey` (Windows/Linux), `newTab: true` without event, event without modifier keys, both `newTab` and modifier keys simultaneously.

---

- Blast radius: Frontend-only. Affects click behavior across ~50 components. No backend changes.
- Potential regressions:
  - A click handler that previously ignored the event parameter now passes it through - if any component has an `onClick` that provides a non-standard event shape, `isModifierKeyPressed` will safely return `false` (checking `metaKey`/`ctrlKey` on undefined returns falsy).
  - Factory functions in `MetricsApplication/Tabs/util.ts` changed return type from `VoidFunction` to `(e?: React.MouseEvent) => void` - callers passing these as `onClick` handlers are compatible since React's `onClick` provides the event as the first argument.
- Rollback plan: Revert the single commit. No data migration or API changes involved.

---

### Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Cmd/Ctrl+Click on navigable items (table rows, links, buttons) now opens the target in a new browser tab instead of replacing the current page. |

---

### Checklist
- [x] Manually tested

---

## Notes for Reviewers

**Core changes (read first):**
- `src/utils/navigation.ts` - new shared helpers for modifier-key detection
- `src/hooks/useSafeNavigate.ts` - accepts `event` option, auto-detects Cmd/Ctrl
- `__mocks__/useSafeNavigate.ts` - mock updated to match new interface

**Pattern across all 50 files:**
1. Add `React` import (for `React.MouseEvent` type)
2. Thread mouse event from `onClick` handler into `safeNavigate(..., { event })`
3. For table `onRow` handlers: change callback signature to accept `(record, event)` and pass `event` through

**Refactored duplicated logic:**
- `DashboardsList.tsx` - removed manual `event.metaKey || event.ctrlKey` + `window.open` (6 lines to 1 line)
- `ListAlert.tsx` - replaced `openInNewTab: boolean` parameter with `{ event?, newTab? }` options object, removed manual `window.open` branch

**Intentionally not changed** - `safeNavigate` calls inside `useEffect`, `onSuccess` mutation callbacks, `setTimeout`, drag handlers, and pagination `onChange` handlers where no `MouseEvent` is available. These are programmatic navigations where Cmd+Click does not apply.

### DIFF
--- a/frontend/src/container/SideNav/SideNav.tsx
+++ b/frontend/src/container/SideNav/SideNav.tsx
@@ -62,8 +62,8 @@ import { AppState } from 'store/reducers';
 import AppReducer from 'types/reducer/app';
 import { USER_ROLES } from 'types/roles';
 import { checkVersionState } from 'utils/app';
-import { showErrorNotification } from 'utils/error';
 import { isModifierKeyPressed } from 'utils/app';
+import { showErrorNotification } from 'utils/error';
 import { openInNewTab } from 'utils/navigation';
 
 import { useCmdK } from '../../providers/cmdKProvider';


frontend/src/components/LogDetail/index.tsx
```diff
@@ -227,7 +227,9 @@ function LogDetailInner({
 				),
 			),
 		};
-		safeNavigate(`${ROUTES.LOGS_EXPLORER}?${createQueryParams(queryParams)}`);
+		safeNavigate(`${ROUTES.LOGS_EXPLORER}?${createQueryParams(queryParams)}`, {
+			event: e,
+		});
```
manika-signoz, line 240: safeNavigate has a newTab parameter in options, passing event might not be needed

YounixM: This helps to evaluate the event at one single place rather than validating at every consumption

frontend/src/utils/navigation.ts
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+
+/**
+ * Returns true if the user is holding Cmd (Mac) or Ctrl (Windows/Linux)
+ * during a click event — the universal "open in new tab" modifier.
+ */
+export const isModifierKeyPressed = (
+	event: MouseEvent | React.MouseEvent,
+): boolean => event.metaKey || event.ctrlKey;
```
H4ad: Should add support for middle click with e.button === 1

frontend/src/container/InfraMonitoringK8s/Clusters/K8sClustersList.tsx
```diff
@@ -450,7 +453,19 @@ function K8sClustersList({
...
+		record: K8sClustersRowData,
+		event: React.MouseEvent,
+	): void => {
+		if (event && isModifierKeyPressed(event)) {
+			const newParams = new URLSearchParams(searchParams);
+			newParams.set(
+				INFRA_MONITORING_K8S_PARAMS_KEYS.CLUSTER_NAME,
+				record.clusterUID,
+			);
+			openInNewTab(`${window.location.pathname}?${newParams.toString()}`);
```
H4ad: With a couple repetitions, maybe introducing another helper like `openInNewTabWith(window.location.pathname, searchParams, { INFRA...: record.clusterUId })`

Not blocking

frontend/src/hooks/useSafeNavigate.ts
```diff
@@ -1,11 +1,13 @@
...
+import React, { useCallback } from 'react';
 import { useLocation, useNavigate } from 'react-router-dom-v5-compat';
 import { cloneDeep, isEqual } from 'lodash-es';
+import { isModifierKeyPressed } from 'utils/navigation';
 
 interface NavigateOptions {
 	replace?: boolean;
 	state?: any;
 	newTab?: boolean;
+	event?: MouseEvent | React.MouseEvent;
```
AshwinBhatkal: I don't think we should be adding this event just for this. We already have an option for newTab. So we should populate that option with isModifierKeyPressed function when passing the newTab option.

YounixM: This helps to evaluate the event at one single place rather than validating at every consumption

AshwinBhatkal: The navigation function's job is to do navigation only. It does not need to care about which event got it here. This violates the Single Responsibility Priniciple. 

When navigating, that option should be passed to open in new tab for which the function already has a param. It is okay if we validate at every consumption because we already have a helper for it.

Ideal scenario would be to have a Link component (with some styles based on usage) with path filled in. That should solve it. Do we plan to get there at some point?

Regarding single place validation, if the goal was to ensure that people don't miss it, then it does not solve that as well as the developer still needs to pass in the event object manually. If they miss it, we go back to the previous state where command click does not work. If we did want to make something required, it would be better to make `openInNewTab` required. This makes it explicit and we avoid this issue overall in the future. It is better than passing in the entire event object which is not relevant to the navigation function.

YounixM: Agree @AshwinBhatkal .

H4ad (approved):

frontend/src/container/InfraMonitoringK8s/Nodes/K8sNodesList.tsx
```diff
@@ -437,7 +438,20 @@ function K8sNodesList({
 		return nodesData.find((node) => node.nodeUID === selectedNodeUID) || null;
 	}, [selectedNodeUID, groupBy.length, nodesData, nestedNodesData]);
 
-	const handleRowClick = (record: K8sNodesRowData): void => {
+	const openNodeInNewTab = (record: K8sNodesRowData): void => {
+		const newParams = new URLSearchParams(searchParams);
+		newParams.set(INFRA_MONITORING_K8S_PARAMS_KEYS.NODE_UID, record.nodeUID);
+		openInNewTab(`${window.location.pathname}?${newParams.toString()}`);
```
AshwinBhatkal: There is a function for this - buildAbsolutePath. All instances same case

frontend/src/container/MetricsApplication/Tabs/util.ts
```diff
@@ -93,8 +93,8 @@ export function onViewTracePopupClick({
 	isViewLogsClicked,
 	stepInterval,
 	safeNavigate,
-}: OnViewTracePopupClickProps): VoidFunction {
-	return (): void => {
+}: OnViewTracePopupClickProps): (e?: React.MouseEvent) => void {
+	return (e?: React.MouseEvent): void => {
```
AshwinBhatkal, line 97: can this event actually be undefined?

frontend/src/hooks/useSafeNavigate.ts
```diff
@@ -122,8 +123,9 @@ export const useSafeNavigate = (
 				);
 			}
 
-			// If newTab is true, open in new tab and return early
-			if (options?.newTab) {
+			const shouldOpenNewTab = options?.newTab;
```
AshwinBhatkal: nit: shouldOpenInNewTab

frontend/src/utils/navigation.ts
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+/**
+ * Returns true if the user is holding Cmd (Mac) or Ctrl (Windows/Linux)
+ * during a click event, or if the middle mouse button is used —
+ * the universal "open in new tab" modifiers.
+ */
+export const isModifierKeyPressed = (
```
AshwinBhatkal: Is this file the right place for this?

frontend/src/components/LogDetail/index.tsx
```diff
@@ -1,6 +1,7 @@
-import { useCallback, useEffect, useMemo, useState } from 'react';
+/* eslint-disable sonarjs/cognitive-complexity */
```
AshwinBhatkal: can we move this comment to the function level?

YounixM: done

frontend/src/utils/navigation.ts
YounixM: moved to app.ts util

frontend/src/hooks/useSafeNavigate.ts
YounixM: done

frontend/src/container/InfraMonitoringK8s/Nodes/K8sNodesList.tsx
YounixM: done

PR #10353, opened 2026-02-18
fix: added a fix for tooltip height when legend is too big

ahrefabhi:
### Summary

This change fixes tooltip layout when legend content makes items taller than the assumed item height (e.g. long labels that wrap). The tooltip now uses the real measured height from Virtuoso instead of `itemCount × TOOLTIP_ITEM_HEIGHT`, so it stays within the max height and shows the full legend correctly.

### Screenshots

- Before
image

- After
image

### What changed

- **Dynamic height calculation**: Use Virtuoso’s `totalListHeightChanged` to get the actual content height instead of assuming a fixed 38px per item.
- **State-based sizing**: Store the measured list height in state and compute the tooltip height as `min(listHeight + 10, TOOLTIP_LIST_MAX_HEIGHT)` when available.
- **Layout updates**: Switch from a fixed `height` to `maxHeight` on the wrapper and set Virtuoso’s `height` explicitly based on the computed `virtuosoHeight`.

### Root cause

The tooltip height was based on `tooltipContent.length * TOOLTIP_ITEM_HEIGHT` (38px). When legend labels wrap, items grow taller than 38px, so the tooltip could overflow or misalign.

### Fix approach

Use Virtuoso’s measured height to drive the tooltip height. On first render we fall back to the length-based estimate; once Virtuoso reports the real height via `totalListHeightChanged`, we update the height to match the content.

### Testing

- Manually tested with long legend labels that wrap.
- Tooltip respects `TOOLTIP_LIST_MAX_HEIGHT` (330px) and shows all items when content fits.

closes https://github.com/SigNoz/signoz/issues/10354


frontend/src/lib/uPlotV2/components/Tooltip/Tooltip.tsx
```diff
@@ -41,34 +42,45 @@ export default function Tooltip({
...
 				</div>
 			)}
 			<div
 				style={{
-					height: Math.min(
-						tooltipContent.length * TOOLTIP_ITEM_HEIGHT,
-						TOOLTIP_LIST_MAX_HEIGHT,
-					),
-					minHeight: 0,
+					maxHeight: TOOLTIP_LIST_MAX_HEIGHT,
```
AshwinBhatkal: can we make this a class?

frontend/src/lib/uPlotV2/components/Tooltip/Tooltip.tsx
```diff
@@ -41,34 +42,45 @@ export default function Tooltip({
...
 				}}
+				data-testid="uplot-tooltip-list"
 			>
 				{tooltipContent.length > 0 ? (
 					<Virtuoso
 						className="uplot-tooltip-list"
 						data={tooltipContent}
-						defaultItemHeight={TOOLTIP_ITEM_HEIGHT}
+						style={{
+							height: virtuosoHeight,
```
AshwinBhatkal: memoize the whole style object instead of just height?

frontend/src/lib/uPlotV2/components/Tooltip/Tooltip.tsx
```diff
@@ -41,34 +42,45 @@ export default function Tooltip({
...
 			<div
 				style={{
-					height: Math.min(
-						tooltipContent.length * TOOLTIP_ITEM_HEIGHT,
-						TOOLTIP_LIST_MAX_HEIGHT,
-					),
-					minHeight: 0,
+					maxHeight: TOOLTIP_LIST_MAX_HEIGHT,
 				}}
+				data-testid="uplot-tooltip-list"
```
AshwinBhatkal: This is the test id but the class is the child element?

frontend/src/lib/uPlotV2/components/Tooltip/__tests__/Tooltip.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+		expect(list).toBeNull();
+	});
+
+	it('sets tooltip list height based on content length, height returned by Virtuoso', () => {
+		const uPlotInstance = createUPlotInstance(null);
+		const content = [createTooltipContent(), createTooltipContent()];
+
+		renderTooltip({ uPlotInstance, content });
+
+		const list = document.querySelector('.uplot-tooltip-list') as HTMLElement;
```
AshwinBhatkal: we have a test id for this

frontend/src/lib/uPlotV2/components/Tooltip/__tests__/Tooltip.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+		renderTooltip({ uPlotInstance, content });
+
+		const list = document.querySelector(
+			'.uplot-tooltip-list',
+		) as HTMLElement | null;
+
+		expect(list).not.toBeNull();
+
+		const marker = document.querySelector(
+			'.uplot-tooltip-item-marker',
```
AshwinBhatkal: can this be selected with test ids and parent/child references? instead of querySelector. Slows down the tests

frontend/src/lib/uPlotV2/components/Tooltip/__tests__/Tooltip.test.tsx
ahrefabhi: Done

frontend/src/lib/uPlotV2/components/Tooltip/__tests__/Tooltip.test.tsx
ahrefabhi: Done

frontend/src/lib/uPlotV2/components/Tooltip/Tooltip.tsx
ahrefabhi: Done

frontend/src/lib/uPlotV2/components/Tooltip/Tooltip.tsx
ahrefabhi: Done

frontend/src/lib/uPlotV2/components/Tooltip/Tooltip.tsx
ahrefabhi: Done

AshwinBhatkal (approved):

Issue #6703, opened 2024-12-24
Count of requests with latency > X ms is not possible for histogram metric using query builder


srikanthccv:
The only options users get are P50, 75, 90, 95, 99. This makes it impossible to build certain percentage metrics (one popular use case is apdex) where the user is not interested in percentile but rather counts.

namanv3:
@srikanthccv given that there is a separate UI ticket now, can this issue be marked as done?

srikanthccv:
The UI support to be added separately for metric query builder.

PR #10355, opened 2026-02-19
feat: add support for count based aggregation in histogram metrics

namanv3:
## Pull Request

---

### 📄 Summary
Query: https://www.notion.so/signoz/Count-based-aggregation-in-Histograms-30afcc6bcd1980cc85b1f0e86e2fd0d3?source=copy_link

#### Issues closed by this PR
Backend change for SigNoz/signoz#6703 

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/frontend/src/container/AppLayout/index.tsx
+++ b/frontend/src/container/AppLayout/index.tsx
@@ -117,7 +117,7 @@ function AppLayout(props: AppLayoutProps): JSX.Element {
 	const [showSlowApiWarning, setShowSlowApiWarning] = useState(false);
 	const [slowApiWarningShown, setSlowApiWarningShown] = useState(false);
 
-	const { latestVersion } = useSelector<AppState, AppReducer>(
+	const { currentVersion } = useSelector<AppState, AppReducer>(
 		(state) => state.app,
 	);
 
@@ -213,9 +213,9 @@ function AppLayout(props: AppLayoutProps): JSX.Element {
 		},
 		{
 			queryFn: (): Promise<SuccessResponse<ChangelogSchema> | ErrorResponse> =>
-				getChangelogByVersion(latestVersion, changelogForTenant),
-			queryKey: ['getChangelogByVersion', latestVersion, changelogForTenant],
-			enabled: isLoggedIn && Boolean(latestVersion),
+				getChangelogByVersion(currentVersion, changelogForTenant),
+			queryKey: ['getChangelogByVersion', currentVersion, changelogForTenant],
+			enabled: isLoggedIn && Boolean(currentVersion),
 		},
 	]);
 
@@ -226,7 +226,7 @@ function AppLayout(props: AppLayoutProps): JSX.Element {
 			!changelog &&
 			!getChangelogByVersionResponse.isLoading &&
 			isLoggedIn &&
-			Boolean(latestVersion)
+			Boolean(currentVersion)
 		) {
 			getChangelogByVersionResponse.refetch();
 		}
@@ -237,9 +237,9 @@ function AppLayout(props: AppLayoutProps): JSX.Element {
 		let timer: ReturnType<typeof setTimeout>;
 		if (
 			isCloudUserVal &&
-			Boolean(latestVersion) &&
+			Boolean(currentVersion) &&
 			seenChangelogVersion != null &&
-			latestVersion !== seenChangelogVersion &&
+			currentVersion !== seenChangelogVersion &&
 			daysSinceAccountCreation > MIN_ACCOUNT_AGE_FOR_CHANGELOG && // Show to only users older than 2 weeks
 			!isWorkspaceAccessRestricted
 		) {
@@ -255,7 +255,7 @@ function AppLayout(props: AppLayoutProps): JSX.Element {
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [
 		isCloudUserVal,
-		latestVersion,
+		currentVersion,
 		seenChangelogVersion,
 		toggleChangelogModal,
 		isWorkspaceAccessRestricted,


pkg/telemetrymetrics/statement_builder.go
```diff
@@ -123,7 +123,7 @@ func (b *MetricQueryStatementBuilder) buildPipelineStatement(
 	origTimeAgg := query.Aggregations[0].TimeAggregation
 	origGroupBy := slices.Clone(query.GroupBy)
 
-	if query.Aggregations[0].SpaceAggregation.IsPercentile() &&
+	if (query.Aggregations[0].SpaceAggregation.IsPercentile() || query.Aggregations[0].SpaceAggregation == metrictypes.SpaceAggregationHistogramCount) &&
```
srikanthccv: Should we get rid of this logic and rely on the metric type metadata instead (similar to what we do with the temporality)?

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -154,7 +154,11 @@ func (b *MetricQueryStatementBuilder) buildPipelineStatement(
 		}
 
 		// make the time aggregation rate and space aggregation sum
-		query.Aggregations[0].TimeAggregation = metrictypes.TimeAggregationRate
+		if query.Aggregations[0].SpaceAggregation.IsPercentile() {
+			query.Aggregations[0].TimeAggregation = metrictypes.TimeAggregationRate
+		} else {
+			query.Aggregations[0].TimeAggregation = metrictypes.TimeAggregationIncrease
+		}
```
srikanthccv, line 160: This can be increase all the times; both rate and increase should work fine for histogram quantile?

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -524,7 +528,7 @@ func (b *MetricQueryStatementBuilder) buildSpatialAggregationCTE(
 		return "", nil, errors.Newf(
 			errors.TypeInvalidInput,
 			errors.CodeInvalidInput,
-			"invalid space aggregation, should be one of the following: [`sum`, `avg`, `min`, `max`, `count`, `p50`, `p75`, `p90`, `p95`, `p99`]",
+			"invalid space aggregation, should be one of the following: [`sum`, `avg`, `min`, `max`, `count`, `p50`, `p75`, `p90`, `p95`, `p99`, `histogram_count`]",
```
srikanthccv: Do we need to introduce another new `histogram_count`? Can we use the `count` instead?

pkg/types/metrictypes/metrictypes.go
```diff
@@ -256,3 +259,22 @@ type MetricTableHints struct {
...
+
+type NoSpaceAggregationParam struct{}
+
+func (_ NoSpaceAggregationParam) StringValue() string {
+	return "{}"
+}
+
+type ComparisonSpaceAggregationParam struct {
+	Operater string  `json:"operator"`
+	Limit    float64 `json:"limit"`
```
srikanthccv: Let's call it a `threshold` instead

pkg/types/querybuildertypes/querybuildertypesv5/builder_elements.go
```diff
@@ -446,6 +446,8 @@ type MetricAggregation struct {
 	TimeAggregation metrictypes.TimeAggregation `json:"timeAggregation"`
 	// space aggregation to apply to the query
 	SpaceAggregation metrictypes.SpaceAggregation `json:"spaceAggregation"`
+	// param for space aggregation if needed
+	SpaceAggregationParam metrictypes.SpaceAggregationParam `json:"spaceAggregationParam"`
```
srikanthccv: I will think about this and get back on this.

srikanthccv (commented):
Let's add an integration test for this

pkg/telemetrymetrics/statement_builder.go
namanv3: I introduced this because a param is required for histogram count but not for normal count

pkg/telemetrymetrics/statement_builder.go
namanv3: Yes they should work, I just didn't want to change it for histogram quantile cuz increase will be greater than rate and it may unexpectedly cross alert limits.

pkg/telemetrymetrics/statement_builder.go
namanv3: I didn't get how this would work.

pkg/types/querybuildertypes/querybuildertypesv5/builder_elements.go
namanv3: Do you mean requiring this SpaceAggregationParam, or how I have implemented it?

pkg/telemetrymetrics/statement_builder.go
srikanthccv: The following logic adds `le` to group by clause. The condition here is checking if it's percentile or histogram_count, a better way would be to use the metric type information than relying on the space aggregation now that we see growing logic here.

pkg/telemetrymetrics/statement_builder.go
srikanthccv: Right now, we only support percentile operations on histogram, given that end result is percentile, rate/increase should result in same, did you have something in mind or was it just safety precaution that don't change what's working philosophy.

pkg/types/querybuildertypes/querybuildertypesv5/builder_elements.go
srikanthccv: the way it's implemented.

pkg/telemetrymetrics/statement_builder.go
namanv3: But the actual value of a point will be more if we consider increase right? Won't that be a problem? Like if the P90 for rate is 2, then for increase it will be 120 (assuming 60 seconds).

pkg/types/querybuildertypes/querybuildertypesv5/builder_elements.go
namanv3: Any particular concern in the current way?

pkg/telemetrymetrics/statement_builder.go
namanv3: Done

pkg/types/metrictypes/metrictypes.go
namanv3: Done

pkg/telemetrymetrics/statement_builder.go
srikanthccv: It's bucket boundaries that dictate the percentage value. The count of the values in each bucket scaling up/down (by same factor across all buckets) should not influence the percentages. 

we can leave it as rate, not a strong opinion.

namanv3: Oh ya makes sense. Got confused. Let’s keep it rate though just to keep existing behaviour the same for now

pkg/types/querybuildertypes/querybuildertypesv5/builder_elements.go
srikanthccv: 1. A new interface is introduced. Introducing new abstractions isn't a problem in itself, but you would want to be careful when defining new abstractions (poorly thought-out abstractions make the codebase hard to work with, real quick, and it will take a lot of time to untangle things).
   - Why specifically this interface a problem?
      - it doesn't provide any polymorphism. There is only one real implementation, and even that implementation means nothing for the consumers (of the interface, i.e statement_builder) because they don't call the method on it but simply unwrap it by type switch. So the interface is not really helping in any way
2. Forces custom unmarshal at the wrong level
    - The Unmarshal on `MetricAggregation` exists solely because Go can't unmarshal into an interface field. This is the second of sign of wrong abstraction

Here is what I would do instead, just use the concrete type.

```go
...
  SpaceAggregationParam *metrictypes.ComparisonSpaceAggregationParam `json:"spaceAggregationParam,omitempty"`
...
```

Since it's a concrete type and is a pointer, the "no param" is simply nil. Now, you may ask a question, what if we need different parameter shapes for different space aggregations _in future_

We will get there when we get there, but there is only one variant now. When we get there, we still have idiomatic ways to solve the problem.

1. introduce a discriminated union type (like we did with the queryenvlope)
2. or introduce additional optional flat struct


At this point, let's just stick to the concrete type `ComparisonSpaceAggregationParam`.

pkg/telemetrymetrics/statement_builder.go
srikanthccv: What does it mean to do `"count"` aggregate on a histogram metric?

pkg/telemetrymetrics/statement_builder.go
namanv3: @srikanthccv I'm reverting back to space aggregation check cuz metric type is currently not being filled by our UI's API call. Will add a todo

pkg/types/querybuildertypes/querybuildertypesv5/builder_elements.go
namanv3: Got it. Good way to look at this. Will take note. ✌️
Made the change as well

pkg/telemetrymetrics/statement_builder.go
namanv3: What is required is a threshold based aggregate that is count based instead of percentile based. We can name it "count_threshold" as well.

srikanthccv: We don't have support for the `distribution` and `heatmap` yet, however, when we add them, the support for Histogram and ExponentialHistogram will have to have space aggregation that't not percentile. The relevant space aggregation is "count" as in `bucket_counts` of the Histogram.
- Count (without group by) shows the distribution of values across all points
- Count (with group by) shows the distribution of values across all points per-group, much like a bar chart with group except there won't be any gap.

Same goes for the heatmap.

Now, for the issue this PR trying to address. It's both valid to ask the count with threshold and simple count that returns the count of all points.

So, I think we should continue with just `count` as the space aggregation and accept the optional threshold param.

srikanthccv: Also let me know if you have any other thoughts/points against the `count` here.

namanv3: I'm thinking if we can somehow combine the `having` filter here instead of introducing a new optional parameter, will look into it

srikanthccv: Ok, a note to keep in mind regarding `having`.

- It's original intent is to mimic to what having does in sql
- The UX should be kept in mind for anything. How easily will it be discoverable when user want to achieve their goal. The space aggregation is the go to place for users now.

namanv3: Having won't make sense cuz it's supposed to work on the resulting count, not on the value for which the count is. I'm keeping the space agg param for now

namanv3: @srikanthccv this is done now. Had to fetch metric type data as well cuz now just knowing the space aggregation won't be enough to figure out if histogram logic is required

pkg/telemetrymetadata/metadata.go
```diff
@@ -1678,15 +1692,15 @@ func (t *telemetryMetaStore) fetchMetricsTemporality(ctx context.Context, queryT
...
 	}
 	defer rows.Close()
 
 	// Process results
 	for rows.Next() {
-		var metricName, temporalityStr string
-		if err := rows.Scan(&metricName, &temporalityStr); err != nil {
-			return nil, errors.Wrapf(err, errors.TypeInternal, errors.CodeInternal, "failed to scan temporality result")
+		var metricName, temporalityStr, metricTypeStr string
+		if err := rows.Scan(&metricName, &temporalityStr, &metricTypeStr); err != nil {
```
srikanthccv: Since the `metricstypes.Temporality` and `metrictypes.Type` already implement the driver interface `Scan` and `Value`, we can declare them as their original types instead of `str` and then scanning into them?

pkg/types/metrictypes/metrictypes.go
```diff
@@ -256,3 +257,12 @@ type MetricTableHints struct {
...
 	Value float64
 }
+
+type ComparisonSpaceAggregationParam struct {
+	Operater  string  `json:"operator"`
+	Threshold float64 `json:"threshold"`
+}
+
+func (param ComparisonSpaceAggregationParam) StringValue() string {
+	return fmt.Sprintf("{\"operator\": \"%s\", \"limit\": \"%f\"}", param.Operater, param.Threshold)
```
srikanthccv: I see this is used in the cache key, let's use the same key format i.e `operator=%s:limit=%f`

pkg/types/metrictypes/metrictypes.go
```diff
@@ -256,3 +257,12 @@ type MetricTableHints struct {
 type MetricValueFilter struct {
 	Value float64
 }
+
+type ComparisonSpaceAggregationParam struct {
+	Operater  string  `json:"operator"`
```
srikanthccv: Mark them as required because they are expected to present if the `ComparisonSpaceAggregationParam` is non-nil https://github.com/SigNoz/signoz/blob/main/docs/contributing/go/handler.md#required-tag

pkg/telemetrymetadata/metadata.go
```diff
@@ -1616,40 +1616,53 @@ func (t *telemetryMetaStore) FetchTemporality(ctx context.Context, queryTimeRang
...
 			if len(temporality) > 1 {
-				result[metricName] = metrictypes.Multiple
+				temporalities[metricName] = metrictypes.Multiple
 			} else {
-				result[metricName] = temporality[0]
+				temporalities[metricName] = temporality[0]
 			}
+		} else if temporality, exists := meterMetricsTemporality[metricName]; exists {
+			temporalities[metricName] = temporality
 			continue
```
srikanthccv: Should this continue be there?

pkg/telemetrymetadata/metadata.go
```diff
@@ -1660,6 +1673,7 @@ func (t *telemetryMetaStore) fetchMetricsTemporality(ctx context.Context, queryT
 	sb := sqlbuilder.Select(
 		"metric_name",
 		"temporality",
+		"any(type) AS type",
```
srikanthccv, line 1675: A note about this, we have seen people send both `Gauge` and `Sum` type (with monotonic false) for the same metric. How does this change affect those users?

pkg/telemetrymetadata/metadata.go
```diff
@@ -1733,15 +1755,15 @@ func (t *telemetryMetaStore) fetchMeterSourceMetricsTemporality(ctx context.Cont
...
+		return nil, nil, errors.Wrapf(err, errors.TypeInternal, errors.CodeInternal, "failed to fetch meter metric temporality")
 	}
 	defer rows.Close()
 
 	// Process results
 	for rows.Next() {
-		var metricName, temporalityStr string
-		if err := rows.Scan(&metricName, &temporalityStr); err != nil {
-			return nil, errors.Wrapf(err, errors.TypeInternal, errors.CodeInternal, "failed to scan temporality result")
+		var metricName, temporalityStr, meterMetricTypeStr string
```
srikanthccv: Same as above about the Scan/Value

tests/integration/testdata/gauge_data_1h.jsonl
srikanthccv, line 1: "is_monotonic":true doesn't make sense for this data. Let's correct it.

tests/integration/src/querier/08_metrics_histogram.py
```diff
@@ -0,0 +1,322 @@
+"""
+Look at the multi_temporality_counters_1h.jsonl file for the relevant data
```
srikanthccv:
```suggestion
Look at the histogram_data_1h.jsonl.jsonl file for the relevant data
```

srikanthccv (commented):
Some final comments

tests/integration/testdata/histogram_data_1h.jsonl
srikanthccv, line 1: Let's add one test for delta as well.

pkg/telemetrymetadata/metadata.go
namanv3: Currently type check is only on histogram. This does not break anything, but in future we may have to handle this if we ever want to add a check on type == Sum or Gauge

namanv3: For now I have added a monotonicity check here

pkg/telemetrymetadata/metadata.go
namanv3: Removed

pkg/telemetrymetadata/metadata.go
namanv3: Done

pkg/types/metrictypes/metrictypes.go
namanv3: Done

pkg/telemetrymetadata/metadata.go
namanv3: Done

tests/integration/testdata/gauge_data_1h.jsonl
namanv3: Done

tests/integration/src/querier/08_metrics_histogram.py
namanv3: Done

tests/integration/testdata/histogram_data_1h.jsonl
namanv3: Done

srikanthccv: Sorry, should have been more clear. I meant the delta temporality for the histogram data 1h.

namanv3: That's already there right? In this file one service is cumulative and one is delta. There's also a test for it `test_histogram_count_for_delta_service`

srikanthccv: Oh I didn't notice it. Let me look check.

namanv3:
@srikanthccv 
1. It won't be an integer if the user enters a custom value. Imo if we can convey in the UI that this is an interpolated value and hence it is not an integer, that's a better user experience along with showing a more likely exact value. Also it's not like people aren't used to seeing non-integers in some places (like the expected value of a role of a die is 3.5)
2. This can only happen if the same le bucket appears twice, which cannot happen

srikanthccv:
> Also it's not like people aren't used to seeing non-integers in some places (like the expected value of a role of a die is 3.5)

One good rule of thumb, assume your users to be dumb (read as they are busy and distracted), not that they are inherently dumb but they wouldn't have to think or guess work as that leads to more time fighting the product than using the product

 > Imo if we can convey in the UI that this is an interpolated value and hence it is not an integer, that's a better user experience along with showing a more likely exact value.

Alright, we can also do this. Please create a ticket for this and ensure this will be implemented when the frontend work is going to be picked up.

namanv3:
@srikanthccv created this ticket for UI handling in general and mentioned this there: https://github.com/SigNoz/engineering-pod/issues/4084

srikanthccv:
Please fix the CI.

PR #10360, opened 2026-02-19
fix: added fix for jerky chart change on panel switch

ahrefabhi:
### 📄 Summary

This PR smooths out the dashboard experience when switching between panel types by centralizing loading behavior in `PanelWrapper` and removing redundant data guards from `WidgetGraphContainer`. It ensures that charts show a consistent loading state instead of briefly flashing stale or empty content, which previously caused a jerky visual change during panel switches.

### Screenshots

- Before

https://github.com/user-attachments/assets/3a71cc76-0527-4e17-b426-6a65b34ad6c2

- After

https://github.com/user-attachments/assets/ab9242e4-69f3-44f0-8acc-81fe51446e3e

### 🐛 Bug Context

- **Root Cause**:  
  - `WidgetGraphContainer` implemented its own early-return logic for “No Data” based on the selected panel type and query results, while `PanelWrapper` did not handle the intermediate loading / re-fetch state on panel switch.  
  - When users changed the panel type (e.g., between LIST / VALUE and other panel types), the panel content could flicker between prior content, “No Data”, and then the new chart, leading to a jerky transition.

- **Fix Strategy**:  
  - **PanelWrapper**  
    - Import and use `Spinner` to render a full-height loading state whenever `queryResponse.isFetching` or `queryResponse.isLoading` is true:
      - This guarantees a stable, predictable loading UI during panel switches.
  - **WidgetGraphContainer**  
    - Remove panel-type-specific “No Data” early returns that depended on both `selectedGraph` and different shapes of `queryResponse.data` (`result` vs `newResult`):
      - Avoids conflicting or premature “No Data” rendering while data is still being fetched or refetched on panel change.
  - **Net effect**: Panel switching now shows a consistent spinner until the new data is ready, eliminating the jerkiness.

### 📝 Changelog

- **Change Type**: Bug Fix  
- **Description**:  
  - Improved dashboard panel switching UX by:
    - Showing a centralized loading spinner from `PanelWrapper` while queries are loading or refetching.
    - Removing redundant, panel-type-specific “No Data” guards from `WidgetGraphContainer` that caused visual jerkiness during panel transitions.

### 🔗 Issues

Closes https://github.com/SigNoz/engineering-pod/issues/3973


frontend/src/container/PanelWrapper/PanelWrapper.tsx
```diff
@@ -32,6 +33,11 @@ function PanelWrapper({
 		// eslint-disable-next-line react/jsx-no-useless-fragment
 		return <></>;
 	}
+
+	if (queryResponse.isFetching || queryResponse.isLoading) {
```
AshwinBhatkal, line 37: If the component is right here, what about an overlay instead?

PR #10367, opened 2026-02-19
chore: authz helpers

H4ad:
## Pull Request

This PR contains the base foundation to implement the new AuthZ authentication.

---

### 📄 Summary

This PR introduces:
- `useAuthZ` to check permissions.
- `GuardAuthZ` component.
- `createGuardedRoute` to guard routes.
- CI Step to ensure we have the correct permissions stored in the UI with strong types.

More about each utility at: https://www.notion.so/signoz/Auth-Z-on-Frontend-30cfcc6bcd19801cbaf0e22db6273266?d=30dfcc6bcd1980c29290001c3acf576c#30cfcc6bcd1980c0b48dd8934bc822d9

#### Issues closed by this PR

Related to [#1693](https://github.com/SigNoz/platform-pod/issues/1693)

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Add base structure to start add support for authentication via custom roles |

---

### 📋 Checklist

- [ ] Manually tested

### DIFF
--- a/frontend/src/lib/uPlotV2/plugins/TooltipPlugin/TooltipPlugin.tsx
+++ b/frontend/src/lib/uPlotV2/plugins/TooltipPlugin/TooltipPlugin.tsx
@@ -8,6 +8,7 @@ import {
 	createSetCursorHandler,
 	createSetLegendHandler,
 	createSetSeriesHandler,
+	getPlot,
 	isScrollEventInPlot,
 	updatePlotVisibility,
 	updateWindowSize,
@@ -53,7 +54,7 @@ export default function TooltipPlugin({
 	const [viewState, setState] = useState<TooltipViewState>(
 		createInitialViewState,
 	);
-	const { plot, isHovering, isPinned, contents, style } = viewState;
+	const { hasPlot, isHovering, isPinned, contents, style } = viewState;
 
 	/**
 	 * Merge a partial view update into the current React state.
@@ -72,11 +73,24 @@ export default function TooltipPlugin({
 		layoutRef.current?.observer.disconnect();
 		layoutRef.current = createLayoutObserver(layoutRef);
 
+		/**
+		 * Plot lifecycle and GC: viewState uses hasPlot (boolean), not the plot
+		 * reference; clearPlotReferences runs in cleanup so
+		 * detached canvases can be garbage collected.
+		 */
 		// Controller holds the mutable interaction state for this tooltip
 		// instance. It is intentionally *not* React state so uPlot hooks
 		// and DOM listeners can update it freely without triggering a
 		// render on every mouse move.
 		const controller: TooltipControllerState = createInitialControllerState();
+
+		/**
+		 * Clear plot references so detached canvases can be garbage collected.
+		 */
+		const clearPlotReferences = (): void => {
+			controller.plot = null;
+			updateState({ hasPlot: false });
+		};
 
 		const syncTooltipWithDashboard = syncMode === DashboardCursorSync.Tooltip;
 
@@ -110,9 +124,10 @@ export default function TooltipPlugin({
 		// Lock uPlot's internal cursor when the tooltip is pinned so
 		// subsequent mouse moves do not move the crosshair.
 		function updateCursorLock(): void {
-			if (controller.plot) {
+			const plot = getPlot(controller);
+			if (plot) {
 				// @ts-ignore uPlot cursor lock is not working as expected
-				controller.plot.cursor._lock = controller.pinned;
+				plot.cursor._lock = controller.pinned;
 			}
 		}
 
@@ -142,8 +157,9 @@ export default function TooltipPlugin({
 			const isPinnedBeforeDismiss = controller.pinned;
 			controller.pinned = false;
 			controller.hoverActive = false;
-			if (controller.plot) {
-				controller.plot.setCursor({ left: -10, top: -10 });
+			const plot = getPlot(controller);
+			if (plot) {
+				plot.setCursor({ left: -10, top: -10 });
 			}
 			scheduleRender(isPinnedBeforeDismiss);
 		}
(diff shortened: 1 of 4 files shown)


frontend/scripts/generate-permissions-type.js
```diff
@@ -0,0 +1,279 @@
+#!/usr/bin/env node
```
therealpandey, line 1: Is it absolutely necessary to spin up the entire cluster to get a list of permissions?

What are some alternatives that have been considered?

therealpandey: While I understand AI has made it super easy to produce code, this CI is not maintainable. Moreover, it's a duplication of the integration test suite setup.

If an environment is needed, please reuse the tests/integration framework whose job is spin up ephemeral signoz environments.

frontend/src/components/createGuardedRoute/createGuardedRoute.styles.scss
```diff
@@ -0,0 +1,41 @@
+.guard-authz-error-no-authz {
```
YounixM, line 1: move createGuardedRoute files to a folder

frontend/src/hooks/useAuthZ/permissions.type.ts
```diff
@@ -0,0 +1,23 @@
+// AUTO GENERATED FILE - DO NOT EDIT - GENERATED BY scripts/generate-permissions-type
+export default {
+	status: 'success',
+	data: {
+		resources: [
+			{
+				name: 'dashboard',
+				type: 'metaresource',
```
YounixM, line 8: can we use camel Casing for type ?

frontend/src/hooks/useAuthZ/useAuthZ.tsx
```diff
@@ -0,0 +1,144 @@
...
+	AuthtypesObjectDTO,
+	AuthtypesTransactionDTO,
+} from '../../api/generated/services/sigNoz.schemas';
+import {
+	BrandedPermission,
+	gettableTransactionToPermission,
+	permissionToTransactionDto,
+} from './utils';
+
+export type UseAuthZPermissionResult = {
```
YounixM: move to a types file

frontend/scripts/generate-permissions-type.js
H4ad: > What are some alternatives that have been considered?

Golang is not my speciality, I don't know if there's alternatives to get the data from authz/resources other than spinning the Docker to fetch from the API directly. @vikrantgupta25 alternatives?

> Moreover, it's a duplication of the integration test suite setup.

This does not run to ensure the route is working, it runs to ensure the frontend is in sync with the backend.

The test is to not forget to update the permissions type, and not let any change in the backend to break the expectations on frontend. And this happened, one of the PRs changed the way the response was mapped, I spotted this bug by using this script.

> If an environment is needed, please reuse the tests/integration framework whose job is spin up ephemeral signoz environments.

And moving the code to test integration framework, from what I read from the triggers inside `integrationci.yaml`, will only catch regressions if we add `safe-to-integrate` label.

frontend/src/hooks/useAuthZ/permissions.type.ts
H4ad: This is a backend decision, this came from the API `/authz/resources`, cc @vikrantgupta25

frontend/scripts/generate-permissions-type.js
therealpandey: So we can use the setup framework to spin up the containers.

Rudimentarily call 'pytest run tests.... setup.py' to spin up an ephemeral environment.

Then all we have to do is call the authz endpoint.

Does this make sense?

frontend/src/components/createGuardedRoute/createGuardedRoute.styles.scss
H4ad: It's not a hook, neither a component, where should I put this thing?

H4ad: It should work together with GuardAuthZ, that's why I kept in the same folder

frontend/scripts/generate-permissions-type.js
H4ad: Done, but let's wait for @vikrantgupta25 to see if he has any other suggestion.

frontend/src/components/createGuardedRoute/createGuardedRoute.styles.scss
H4ad: Kept as a component

frontend/scripts/generate-permissions-type.js
vikrantgupta25: I would still suggest if we just fetch it in context and use it from there it would work. But if we want to move this to compile time from get go we can go via the route @therealpandey mentioned.

vikrantgupta25: and given if the response type changes for some reason the openapi spec is your source of gurrantee since it would break types and the counter PR anyways will be made aware of breaking changes.

H4ad: I want to keep this at compile time, in this case, the suggestion by Pandey was implemented, please take a moment to review.

.github/workflows/generate-permissions-type.yml
```diff
@@ -0,0 +1,58 @@
+name: Check Permissions Type Generation
```
therealpandey: I would like to jeep this a mandatory check @H4ad.

Can you make it a part of `jsci` and rename it to `authz` so the github context becomes `jsci/authz`?

therealpandey (approved):

vikrantgupta25 (approved):

therealpandey (approved):

PR #10368, opened 2026-02-19
fix: add required tags where needed

karanbalani:
### 📄 Summary
- Add `required` tag where needed in the following packages:
    - zeustypes
    - gatewaytypes

### DIFF
--- a/.github/workflows/integrationci.yaml
+++ b/.github/workflows/integrationci.yaml
@@ -53,9 +53,9 @@ jobs:
           - sqlite
         clickhouse-version:
           - 25.5.6
-          - 25.10.1
+          - 25.10.5
         schema-migrator-version:
-          - v0.129.7
+          - v0.142.0
         postgres-version:
           - 15
     if: |
--- a/frontend/src/components/LogDetail/LogDetails.styles.scss
+++ b/frontend/src/components/LogDetail/LogDetails.styles.scss
@@ -42,6 +42,7 @@
 		display: flex;
 		flex-direction: column;
 		padding: 16px;
+		padding-bottom: 0;
 	}
 
 	.title {
@@ -190,7 +191,7 @@
 		padding: 0px;
 	}
 	.log-detail-drawer__footer-hint {
-		position: absolute;
+		position: sticky;
 		bottom: 0;
 		left: 0;
 		right: 0;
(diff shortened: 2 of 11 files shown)


pkg/types/zeustypes/types.go
```diff
@@ -11,15 +11,15 @@ type PostableHost struct {
...
-	TimelineForMigratingToSigNoz string   `json:"timeline_for_migrating_to_signoz"`
+	UsesOtel                     bool     `json:"uses_otel" required:"true"`
+	HasExistingObservabilityTool bool     `json:"has_existing_observability_tool" required:"true"`
+	ExistingObservabilityTool    string   `json:"existing_observability_tool" required:"true"`
+	ReasonsForInterestInSigNoz   []string `json:"reasons_for_interest_in_signoz" required:"true"`
+	LogsScalePerDayInGB          int64    `json:"logs_scale_per_day_in_gb" required:"true"`
+	NumberOfServices             int64    `json:"number_of_services" required:"true"`
+	NumberOfHosts                int64    `json:"number_of_hosts" required:"true"`
+	WhereDidYouDiscoverSigNoz    string   `json:"where_did_you_discover_signoz" required:"true"`
+	TimelineForMigratingToSigNoz string   `json:"timeline_for_migrating_to_signoz" required:"true"`
```
karanbalani, line 22: @SagarRajput-7 - can you confirm if all of these can be marked as required, as frontend might be allowing to skip some of the questions to the users.
just double check once.

SagarRajput-7: Among this, only skipping the following fields is allowed to move to the next step, but in these cases, we also send `0` as the value, so in a way that is also a value (not null), hence yes, all of them can be required.

- LogsScalePerDayInGB         
- NumberOfServices            
- NumberOfHosts

karanbalani: Thanks :D

pkg/types/zeustypes/types.go
```diff
@@ -30,9 +30,9 @@ type GettableHost struct {
 }
 
 type Host struct {
-	Name      string `json:"name"`
-	IsDefault bool   `json:"is_default"`
-	URL       string `json:"url"`
+	Name      string `json:"name" required:"true"`
```
therealpandey, line 33: Please add required true in gettablehost

karanbalani:
Related: https://github.com/SigNoz/platform-pod/issues/1722

vikrantgupta25 (approved):

Issue #3903, opened 2023-11-06
[FE] Cmd+click to open sidebar/dashboard in new tab


makeavish:
Loom video: https://www.loom.com/share/d1cadec1a5fb4f20bdcce5ed350f2dc9?sid=cc7a0513-b649-45d0-b36d-ba3d59823c0e

Mihir-Karbelkar:
Hi @makeavish / @Rajat-Dabade ,

Let me know if I can take this up, looks like a good first issue as well.

makeavish:
@Mihir-Karbelkar Assigned to you

Mihir-Karbelkar:
I have set up the local env, I have two ways to go about this problem.

I wanted to know if we want to just have ctrl/cmd+click to open the link in a new tab or the context menu should also have an 'open in new tab' option. 

If we are looking for the second option (which I feel provides a better UX) then I think we should do away with the on click handler logic in the SideNav.tsx file and use the good old <Link tag and let the browser handle all the nuances of ctrl click or context menu option.

makeavish:
> I have set up the local env, I have two ways to go about this problem.
> 
> I wanted to know if we want to just have ctrl/cmd+click to open the link in a new tab or the context menu should also have an 'open in new tab' option.
> 
> If we are looking for the second option (which I feel provides a better UX) then I think we should do away with the on click handler logic in the SideNav.tsx file and use the good old <Link tag and let the browser handle all the nuances of ctrl click or context menu option.

2nd way is fine. We already have cmd + click option on trace filter page for clicking trace table rows.

AshwinBhatkal:
This has been fixed now

PR #10373, opened 2026-02-20
feat: text selection block

ishan621:
## Pull Request
This PR handles a bug, in logs explorer when we used to select a text , it would automatically open log detail which was a bad UX as user was just trying to select and copy a value. 

Fix - We have introduced a hook , which tells us if text is selected , if yes then we don't open log detail else it will behave as normal.

---

To improve UX of the user, user should be able to select and copy and only open log details when clicked

Before :

https://github.com/user-attachments/assets/042e04ae-5915-4df9-8a43-5436df8e2c17

After :

https://github.com/user-attachments/assets/f97fd6c7-534c-426e-9ecd-899d4d42f839

#### Issues closed by this PR
#3903

---

### ✅ Change Type
_Select all that apply_

- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

🐛 Bug Context
Required if this PR fixes a bug

Root Cause
What caused the issue?
Faulty assumption

Fix Strategy
Root cause: we were not distinguishing between text selection and click events, which led to the issue. Many libraries handle this behavior inherently, but react-virtuoso does not provide it out of the box, so we needed to implement the handling manually.

🧪 Testing Strategy
How was this change validated?
1. open app.us.staging/cloud
2. Goto logs-explorer route , drag select any log row 
3. You should be able to select any content of the row without opening the drawer

Tests added/updated: Yes
Manual verification: Yes
Edge cases covered: Yes

📋 Checklist
 Tests added

### DIFF
--- a/frontend/src/hooks/logs/__tests__/useLogDetailHandlers.test.ts
+++ b/frontend/src/hooks/logs/__tests__/useLogDetailHandlers.test.ts
@@ -1,0 +1,94 @@
+import { act, renderHook } from '@testing-library/react';
+import { useActiveLog } from 'hooks/logs/useActiveLog';
+import { useIsTextSelected } from 'hooks/useIsTextSelected';
+import { ILog } from 'types/api/logs/log';
+
+import useLogDetailHandlers from '../useLogDetailHandlers';
+
+jest.mock('hooks/logs/useActiveLog');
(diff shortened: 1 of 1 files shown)


YounixM:
@cursor review

PR #10375, opened 2026-02-20
fix: fixed the invite member validation error with empty detail rows

SagarRajput-7:
## Pull Request
We found a bug where as with the new UI revamp we are by default we started showing 3 invite team member rows, but with that if user only enters details for one, then we are restricting them with an error - "Please enter valid emails and select roles for all team members", mean the UI is validation on the empty rows also, hence user have to delete the empty rows and then move ahead, which is not a good experience.

Fix: validation would only be triggered on the rows touched by users, where touched means user have added either email or role in those fields. 
- also updated error message
- added test cases for the invitemember file, including the different cases of validations and errors.

---

#### Before 
-----------
image

#### After
------------

https://github.com/user-attachments/assets/bdea929b-0d71-4e86-a888-497b4b748b85

https://github.com/user-attachments/assets/2f027bca-e950-4f39-a826-2715945f6575

- reference - https://signoz-team.slack.com/archives/C02TJ466H8U/p1771574982451329
- https://github.com/SigNoz/engineering-pod/issues/3484

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated: Added test cases
- Manual verification: Done the manual testing
- Edge cases covered: tested multiple cases around the different validation scenerios.

---

- Blast radius: Invite Team member step in onboarding
- Potential regressions: Validation cases missing
- Rollback plan: NA

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [X] Tests added or explicitly not required
- [X] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/frontend/orval.config.ts
+++ b/frontend/orval.config.ts
@@ -25,13 +25,13 @@ export default defineConfig({
 					useMutation: true,
 					useInvalidate: true,
 					signal: true,
-					useOperationIdAsQueryKey: true,
+					useOperationIdAsQueryKey: false,
 				},
 				useDates: true,
 				useNamedParameters: true,
 				enumGenerationType: 'enum',
 				mutator: {
-					path: './src/api/index.ts',
+					path: './src/api/generatedAPIInstance.ts',
 					name: 'GeneratedAPIInstance',
 				},
 
--- a/frontend/src/api/ErrorResponseHandlerForGeneratedAPIs.ts
+++ b/frontend/src/api/ErrorResponseHandlerForGeneratedAPIs.ts
@@ -1,0 +1,48 @@
+import { RenderErrorResponseDTO } from 'api/generated/services/sigNoz.schemas';
+import { AxiosError } from 'axios';
+import APIError from 'types/api/error';
+
+// Handles errors from generated API hooks (which use RenderErrorResponseDTO)
+export function ErrorResponseHandlerForGeneratedAPIs(
+	error: AxiosError<RenderErrorResponseDTO>,
+): never {
+	const { response, request } = error;
+	// The request was made and the server responded with a status code
+	// that falls out of the range of 2xx
+	if (response) {
+		throw new APIError({
+			httpStatusCode: response.status || 500,
+			error: {
+				code: response.data.error.code,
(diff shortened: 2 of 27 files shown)


SagarRajput-7:
As per the discussion with @makeavish, we decided to change the `Complete` button to `Send Invites` (previous behaviour). 

Reason - 

We had two actions - "Complete" and "I'll do this later" - which had 2 separate functions - 
- "Complete" - sends invite API with the user inputs, but in this case users might tend to treat this a submit button i.e. Complete the flow - "Complete" and when invite API is fired with [] input it throughs error, means eventually to get away the user should use the "I'll do this later" button, so rather we should move to the "Send Invites" which clears its intentions
- "I'll do this later" - means skips invite flow and completes the onboarding, users should use this if they want to skip.

#### Older feature before the revamp :
------------------------

Screenshot 2026-02-22 at 6 14 52 PM

#### Current Behaviour:
----------

image

SagarRajput-7:
#### RCA 

Issue - 
We have by default 3 empty rows for user to invite team member. If users doesn't enter any detail and hit complete, they are validated and shown errors, even if they enter a valid entry in one of the row, since the other 2 are empty and being validated hence they are not shown error and not allowed to continue. They now have to either delete the empty rows or hit the `I'll do this later` button.

> Behaviour is same as before, empty rows were validated earlier also but they were by default 1 only. In the last major UI revamp, with a later stage feedback from design we increased the number of rows to 3, but the validation part was kept same. 
   
> How could we have prevented this - Better manual testing and test cases. Should either have tested all the routines again or not have incorporated feedback at a later stage.

> Action taken - Added fix for the validation, added test cases for all the form and submit validation and did the manual testing.
   
Issue/Feedback - "Complete" Button should be "Send Invites" as before.
> This is done as per the design.

> Action Taken - changed button text back to "Send Invites". Refer - [this](https://github.com/SigNoz/signoz/pull/10375#issuecomment-3941801943).

PR #10377, opened 2026-02-20
fix: ui bugs body width and table css

ishan621:
## Pull Request
This PR addresses multiple UI issues:
1. In column view, the table header ordering was incorrect. The first column was taking unexpectedly large space, which disrupted the overall layout.
2. When opening log details, the search fields were not expanded by default.
3. When the number of columns was less than three, the table body was still restricted to 60% width instead of utilizing the available space.

Before :

https://github.com/user-attachments/assets/770eef64-f600-4e06-a16c-16aece9fb961

After :

https://github.com/user-attachments/assets/c306ba7a-2291-464c-a8e1-5b6ae43e51c3

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context
We observed several UI inconsistencies in the column view and log details experience that were impacting usability and layout stability. The table header sizing and ordering were misaligned, default states were not user-friendly (e.g., search fields collapsed), the table body was not utilizing available space for smaller column counts. These issues collectively led to a degraded user experience and visual inconsistency.

Regression, faulty assumption

#### Fix Strategy
This was caused due to css getting overwritten for logs and we also had intentionally made body width to max 60rem which caused the issue

---

Manual

### DIFF
--- a/frontend/src/container/LogDetailedView/Overview.tsx
+++ b/frontend/src/container/LogDetailedView/Overview.tsx
@@ -1,4 +1,4 @@
-import { ReactNode, useCallback, useState } from 'react';
+import { ReactNode, useCallback, useEffect, useRef, useState } from 'react';
 import MEditor, { EditorProps, Monaco } from '@monaco-editor/react';
 import { Color } from '@signozhq/design-tokens';
 import type { InputRef } from 'antd';
@@ -53,11 +53,16 @@ function Overview({
 	);
 	const [fieldSearchInput, setFieldSearchInput] = useState<string>('');
 
+	const focusTimerRef = useRef<ReturnType<typeof setTimeout>>();
+
 	const searchInputRef = useCallback((node: InputRef | null) => {
+		clearTimeout(focusTimerRef.current);
 		if (node) {
-			setTimeout(() => node.focus(), 100);
+			focusTimerRef.current = setTimeout(() => node.focus(), 100);
 		}
 	}, []);
+
+	useEffect(() => (): void => clearTimeout(focusTimerRef.current), []);
 
 	const isDarkMode = useIsDarkMode();
 


frontend/src/container/LogDetailedView/Overview.tsx
```diff
@@ -46,12 +47,18 @@ function Overview({
...
 	const [isAttributesExpanded, setIsAttributesExpanded] = useState<boolean>(
 		true,
 	);
 	const [fieldSearchInput, setFieldSearchInput] = useState<string>('');
 
+	const searchInputRef = useCallback((node: InputRef | null) => {
+		if (node) {
+			setTimeout(() => node.focus(), 100);
+		}
+	}, []);
```
H4ad, line 63: Some issues I can think about this code:
- Every time the search is rendered, after being disposed, this triggers focus.
  - But don't know if this search can be hidden or not in that UI, if it can, it's an issue (or a feature).
- Set timeout not being cleanup, I don't think someone will open and close this tab under 100ms, but if does, node.focus very likely to throw an error.

Maybe callback + setState + useState with cleanup function, this should solve the issue with cleanup and you can add a flag to ensure it's never triggered twice

ishan621: Introduced a useRef to store the setTimeout timer ID.
The callback ref is invoked by React:
* with the node on mount
* with null on unmount
This provides natural lifecycle points to start and cancel the timer.

Chose useRef over useState for storing the timer ID because updating a ref does not trigger re-renders.
Using useState inside a ref callback with Ant Design Input leads to an infinite render loop.
Root cause: Ant Design’s Input uses useImperativeHandle, which creates a new handle object on every render.

PR #10383, opened 2026-02-21
Fix(Infra Monitoring): Remove Double Decoding from URL 

nikhilmantri0902:
## Pull Request

---

When the page loads, we retrieve the filter using searchParams.get('filters'). React Router automatically URL-decodes this value, returning the exact plaintext JSON string that was originally passed (e.g. {"items":[..."value":"open%" ...]}). However, the application was inadvertently decoding them a second time using decodeURIComponent(searchParams.get('filters')). When users entered literal % characters, React Router correctly decoded them into the JSON string, but the second explicit decodeURIComponent threw a URIError on encountering the % without valid hex characters next to it. In the InfraMonitoringK8s views, this error was unhandled and caused the entire page to crash.

we might have done wrapping the JSON.stringify(value) in an explicit encodeURIComponent before calling setSearchParams. However, this approach is incorrect for two reasons:

1. Double Encoding: React Router's setSearchParams automatically URL-encodes values natively. If we manually encodeURIComponent beforehand, the parameters become double-encoded (e.g., %7B becomes %257B). This creates bloated and unreadable URLs.
2. Backward Compatibility: If we only fixed the saving behavior, any previously saved bookmarks, shared links, or manually typed URLs that only single-encoded literal % characters would still cause the application to crash, because the underlying decodeURIComponent bug in the read-path would remain unfixed.

> Include screenshots or screen recordings that clearly show the behavior before the change and the result after the change. This helps reviewers quickly understand the impact and verify the update.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

Refer to 

>  Why does this change exist?  
above. 

refer to 
>  Why does this change exist?  
above. 

#### Fix Strategy

> How does this PR address the root cause?

We will remove the redundant decodeURIComponent calls and replace them with a robust parsing block that correctly prefers JSON.parse directly, while also try/catch just in case the URL was externally created or double-encoded to avoid a crash but log the error ( and possibly report to sentry ). This PR fixes this issue across all subsections ( Pods, Nodes, Deployments, Statefulsets) in `kubernetes` monitoring. 

This PR also fixes the issue for Host List also where sharing a URL and opening it in a new tab makes you lose all the filters. 

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---


frontend/src/container/InfraMonitoringK8s/commonUtils.tsx
```diff
@@ -260,6 +261,19 @@ export const filterDuplicateFilters = (
...
 
+export const safeParseJSON = <T,>(value: string): T | null => {
+	if (!value) {
+		return null;
+	}
+	try {
+		return JSON.parse(value) as T;
+	} catch (e) {
+		console.error('Error parsing JSON from URL parameter:', e);
+		// TODO: Should we capture this error in Sentry?
```
nikhilmantri0902, line 272: Should we capture this error in sentry?

srikanthccv:
This is not the right fix. @H4ad will get it fixed.

PR #10385, opened 2026-02-23
feat(authz): update authz response to prevent pre-compute 

vikrantgupta25:
### 📄 Summary
- update get/patch objects request response to avoid pre-compute on the client 

contributes to: https://github.com/SigNoz/platform-pod/issues/1693

### DIFF
--- a/frontend/src/api/generated/services/sigNoz.schemas.ts
+++ b/frontend/src/api/generated/services/sigNoz.schemas.ts
@@ -657,14 +657,14 @@ export interface GatewaytypesLimitMetricValueDTO {
 export interface GatewaytypesLimitValueDTO {
 	/**
 	 * @type integer
-	 * @format int64
+	 * @nullable true
 	 */
-	count?: number;
+	count?: number | null;
 	/**
 	 * @type integer
-	 * @format int64
+	 * @nullable true
 	 */
-	size?: number;
+	size?: number | null;
 }
 
 export interface GatewaytypesPaginationDTO {
@@ -3138,7 +3138,7 @@ export type GetIngestionKeys200 = {
 	status: string;
 };
 
-export type CreateIngestionKey200 = {
+export type CreateIngestionKey201 = {
 	data: GatewaytypesGettableCreatedIngestionKeyDTO;
 	/**
 	 * @type string
--- a/.github/workflows/integrationci.yaml
+++ b/.github/workflows/integrationci.yaml
@@ -48,6 +48,7 @@ jobs:
           - role
           - ttl
           - alerts
+          - ingestionkeys
         sqlstore-provider:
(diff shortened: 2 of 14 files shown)


pkg/apiserver/signozapiserver/authz.go
```diff
@@ -26,5 +26,22 @@ func (provider *provider) addAuthzRoutes(router *mux.Router) error {
 		return err
 	}
 
+	if err := router.Handle("/api/v1/authz/resources", handler.New(provider.authZ.AdminAccess(provider.authzHandler.GetResources), handler.OpenAPIDef{
```
H4ad: > provider.authZ.AdminAccess

Should be public

pkg/types/authtypes/transaction.go
```diff
@@ -3,24 +3,13 @@ package authtypes
...
-}
-
-type Object struct {
-	Resource Resource `json:"resource" required:"true"`
-	Selector Selector `json:"selector" required:"true"`
-}
-
 type Transaction struct {
-	ID       valuer.UUID `json:"id"`
+	ID       valuer.UUID `json:"-"`
```
H4ad, line 12: Why "-"?

frontend/src/api/generated/services/sigNoz.schemas.ts
```diff
@@ -108,6 +108,35 @@ export interface AuthtypesGettableAuthDomainDTO {
 	updatedAt?: Date;
 }
 
+export interface AuthtypesGettableObjectsDTO {
+	resource: AuthtypesResourceDTO;
+	/**
+	 * @type array
+	 * @nullable true
+	 */
+	selectors: AuthtypesSelectorDTO[] | null;
```
H4ad: Why this is generated as array when it is a string?

pkg/types/authtypes/transaction.go
vikrantgupta25: this ID is for internally mapping the correct authorisation. We aren't generating client side IDs here.

frontend/src/api/generated/services/sigNoz.schemas.ts
vikrantgupta25: it isn't a string. this is an array for role patch APIs

H4ad: The check route expects this to be stirng, but it's being generated as string.

vikrantgupta25: please check now

AshwinBhatkal (dismissed):

vikrantgupta25:
contributes to: https://github.com/SigNoz/platform-pod/issues/1748

PR #10390, opened 2026-02-23
fix(openapi): make the data and status required in success responses

vikrantgupta25:
### 📄 Summary

- make the data and status required in success responses
- wanted to make the error and status required as well but the current `RenderErrorResponseDTO` is wrapped by AxiosError which makes the types go crazy! Will take this separately. 

contributes to: https://github.com/SigNoz/platform-pod/issues/1693

### DIFF
--- a/pkg/http/render/render.go
+++ b/pkg/http/render/render.go
@@ -16,8 +16,8 @@ const (
 var json = jsoniter.ConfigCompatibleWithStandardLibrary
 
 type SuccessResponse struct {
-	Status string      `json:"status"`
-	Data   interface{} `json:"data,omitempty"`
+	Status string      `json:"status" required:"true"`
+	Data   interface{} `json:"data,omitempty" required:"true"`
 }
 
 type ErrorResponse struct {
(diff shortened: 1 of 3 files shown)


pkg/http/render/render.go
```diff
@@ -16,8 +16,8 @@ const (
 var json = jsoniter.ConfigCompatibleWithStandardLibrary
 
 type SuccessResponse struct {
-	Status string      `json:"status"`
-	Data   interface{} `json:"data,omitempty"`
+	Status string      `json:"status" required:"true"`
+	Data   interface{} `json:"data,omitempty" required:"true"`
```
vikrantgupta25, line 20: ahhhh we need something else here. for `204` it won't be present

AshwinBhatkal: It can still be an empty array or string

AshwinBhatkal: omitEmpty might need to be thought about

vikrantgupta25: the more i think about this, for 204 there won't be any response type in openapi as well so doesn't matter i guess

vikrantgupta25:
contributes to: https://github.com/SigNoz/platform-pod/issues/1748

PR #10392, opened 2026-02-23
feat(vite): migrate away from webpack

H4ad:
## Pull Request

This PR migrates from Webpack to Vite.

Build time: 83s -> 10s (7.3x faster)
Hot reload support -> Yes, instant feedback.

---

### 📄 Summary

More details about why at: https://www.notion.so/signoz/Building-30cfcc6bcd1980628418fc7264fb7a0d#30dfcc6bcd1980cd9f82d3609f04d6c6

---

### ✅ Change Type

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Manual verification: I tested manually a couple pages.

---

Since this changes the bundler, this could break some users if we don't configure the bundler to almost the same bundling configuration as the past.

- Blast radius: All users
- Potential regressions: Yes
- Rollback plan: -

---

| Field | Value |
|------|-------|
| Deployment Type | All |
| Change Type | Maintenance |
| Description | We migrated from Webpack to Vite to improve build time and have support for Hot Reload. |

---

### 📋 Checklist

- [x] Manually tested

---

## 👀 Notes for Reviewers

I tried to migrate all the plugins available on Webpack, the only one missing is RetryChunk that is not possible to migrate due to Vite usage of `import` instead of a custom loader.

In the Notion Document, I also added examples of the output files, so you can take a look at what exactly changes.

I also had to update the CI to match the prefix of `VITE_`, let me know if I missed something.

I also updated the Typescript & Jest to work well with the Vite, I didn't see any regressions locally but still need to run the entire test suite on CI to ensure everything is working fine.

## Breaking changes

The following things you should pay more attention:
- Typescript upgraded from 4.9 to 5.9
- Jest from v27 to v30
- Jest Environment Dom from v26 to v29
- Envs for Vite now has the prefix `VITE_`
  - Review the CI carefully.

This did broke some tests that I had to manually update, it required me to be more explicit when dealing with our components package since it's configured correctly for dual package publish.

But almost all these changes should not affect our users, these are most breaking changes in our current workflow/development.


.github/workflows/build-community.yaml
```diff
@@ -56,6 +56,7 @@ jobs:
       JS_OUTPUT_ARTIFACT_PATH: frontend/build
       DOCKER_BUILD: false
       DOCKER_MANIFEST: false
+      NODE_VERSION: '24'
```
AshwinBhatkal: Update this in package json as well
Update it in nvmrc as well

Can you make this change in https://github.com/SigNoz/primus.workflows instead of here. You can avoid this line change in the other build yamls

.github/workflows/build-enterprise.yaml
```diff
@@ -89,6 +89,7 @@ jobs:
       JS_OUTPUT_ARTIFACT_PATH: frontend/build
       DOCKER_BUILD: false
       DOCKER_MANIFEST: false
+      NODE_VERSION: '24'
```
AshwinBhatkal: will be moved to primus

.github/workflows/build-staging.yaml
```diff
@@ -88,6 +88,7 @@ jobs:
       JS_OUTPUT_ARTIFACT_PATH: frontend/build
       DOCKER_BUILD: false
       DOCKER_MANIFEST: false
+      NODE_VERSION: '24'
```
AshwinBhatkal: will be moved to primus

frontend/scripts/update-registry.cjs
```diff
@@ -1,33 +1,27 @@
-/* eslint-disable @typescript-eslint/no-var-requires, import/no-dynamic-require, simple-import-sort/imports, simple-import-sort/exports */
 const fs = require('fs');
 const path = require('path');
 
-// 1. Define paths
```
AshwinBhatkal, line 5: let these comments be there

frontend/src/AppRoutes/index.tsx
```diff
@@ -220,11 +220,11 @@ function App(): JSX.Element {
 		) {
 			// eslint-disable-next-line @typescript-eslint/ban-ts-comment
 			// @ts-ignore
-			window.Pylon('hideChatBubble');
+			window.Pylon?.('hideChatBubble');
```
AshwinBhatkal, line 221: why did we need this change? Is there a change in order of execution?

frontend/.eslintrc.cjs
```diff
@@ -33,7 +33,7 @@ module.exports = {
 		'simple-import-sort', // Auto-sort imports
 		'react-hooks', // React Hooks rules
 		'prettier', // Code formatting
-		'jest', // Jest test rules
+		// 'jest', // Disabled due to old version of eslint
```
AshwinBhatkal: What is the solution for this?

frontend/i18-generate-hash.cjs
```diff
@@ -1,3 +1,4 @@
+// oxlint-disable typescript/no-require-imports
```
AshwinBhatkal: we don't have oxlint yet?

frontend/example.env
```diff
@@ -1,8 +1,8 @@
 NODE_ENV="development"
 BUNDLE_ANALYSER="true"
```
AshwinBhatkal, line 2: Will BUNDLE_ANALYSER still be supported?

frontend/package.json
```diff
@@ -147,21 +143,18 @@
...
 		"rrule": "2.8.1",
 		"stream": "^0.0.2",
-		"style-loader": "1.3.0",
 		"styled-components": "^5.3.11",
-		"terser-webpack-plugin": "^5.2.5",
 		"timestamp-nano": "^1.0.0",
 		"ts-node": "^10.2.1",
-		"tsconfig-paths-webpack-plugin": "^3.5.1",
-		"typescript": "^4.0.5",
+		"typescript": "^5.9.3",
```
AshwinBhatkal: use exact version

frontend/package.json
```diff
@@ -88,31 +90,25 @@
...
 		"history": "4.10.1",
-		"html-webpack-plugin": "5.5.0",
 		"http-proxy-middleware": "3.0.5",
 		"http-status-codes": "2.3.0",
 		"i18next": "^21.6.12",
 		"i18next-browser-languagedetector": "^6.1.3",
 		"i18next-http-backend": "^1.3.2",
 		"immer": "11.1.3",
-		"jest": "^27.5.1",
+		"jest": "^30.2.0",
```
AshwinBhatkal: use exact version

frontend/package.json
```diff
@@ -216,51 +207,48 @@
...
+		"jest-environment-jsdom": "29.7.0",
+		"jest-environment-node": "29.7.0",
+		"jest-styled-components": "^7.2.0",
 		"lint-staged": "^12.5.0",
 		"msw": "1.3.2",
 		"npm-run-all": "latest",
 		"orval": "7.18.0",
 		"portfinder-sync": "^0.0.2",
-		"postcss": "8.4.38",
+		"postcss": "8.5.6",
```
AshwinBhatkal: is see minor version updates for quite a few of them? How confident do we feel of them?

frontend/vite.config.ts
```diff
@@ -0,0 +1,129 @@
...
+			},
+			css: {
+				preprocessorOptions: {
+					less: {
+						javascriptEnabled: true,
+					},
+				},
+			},
+			define: {
+				'process.env': JSON.stringify({
```
AshwinBhatkal, line 105: Probably have a doc added or updated for when someone needs to add a new env var. It touches quite a few files

.github/workflows/build-community.yaml
AshwinBhatkal: I see the changes for package.json and nvmrc are done. So only primus remains

AshwinBhatkal:
So, even with comments resolved, you may not be able to merge it because of checks that are stuck. You may have to consider separating out these changes. Let's try and see though

frontend/.eslintrc.cjs
H4ad: Biome

frontend/i18-generate-hash.cjs
H4ad: No, I probably started this from the oxlint branch, I will remove

frontend/example.env
H4ad: Yes, it's on vite.config.ts

frontend/package.json
H4ad: I updated most of them to work better with Vite, so I don't think it's an issue, we can review everything by testing the Demo environment.

frontend/vite.config.ts
H4ad: The new env vars should be added using "import.meta.env", this is just a way to migrate to Vite without having to update all the code. Added a TODO to fix remove this in next PRs.

.github/workflows/build-community.yaml
H4ad: Let's merge this PR first, then we go another interation to remove this and move to primus. I don't know what will break if we update primus directly.

frontend/src/AppRoutes/index.tsx
H4ad: This breaks when you don't set the value of pylon, I don't know why this worked in the past.

It's basically missing the checks if pylon was initialized (like in other places)

AshwinBhatkal: I see if checks with window.pylon, may be we can unify it that way. I'm not sure why it is required as well right now

frontend/vite.config.ts
```diff
@@ -0,0 +1,137 @@
...
+					data: {
+						PYLON_APP_ID: env.VITE_PYLON_APP_ID || '',
+						APPCUES_APP_ID: env.VITE_APPCUES_APP_ID || '',
+					},
+				},
+			}),
+			vitePluginChecker({
+				typescript: true,
+				// this doubles the build tim
+				// disabled to use Biome/tsgo (in the future) as alternative
```
AshwinBhatkal, line 49: why would in the future matter? The separate type check we run during CI should be good enough

frontend/.eslintrc.cjs
AshwinBhatkal: Cool. Let's keep track of this though. Add comment mentioning 'biome' if required

frontend/vite.config.ts
H4ad: That works as well, we could also could drop that step and leave it via build time/while running in dev, both strategies are good for me.

.github/workflows/build-enterprise.yaml
H4ad: Added TODO to migrate/update primus

H4ad:
~Anyone has any idea why the CI is failling due to node version when I'm explicitly passing v24?~
Rollback the node version to v22, we can update the version on primus workflow later.

frontend/vite.config.ts
AshwinBhatkal: Yep. This can always be false

vikrantgupta25:
@YounixM have disabled the auto merge for the PR, let's have the integration tests run once then we can merge this. Since they aren't required hence had to disable the auto merge

frontend/package.json
```diff
@@ -1,293 +1,283 @@
...
+    "rollup-plugin-visualizer": "7.0.0",
+    "rrule": "2.8.1",
+    "stream": "^0.0.2",
+    "styled-components": "^5.3.11",
+    "timestamp-nano": "^1.0.0",
+    "ts-node": "^10.2.1",
+    "typescript": "5.9.3",
+    "uplot": "1.6.31",
+    "uuid": "^8.3.2",
+    "vite": "npm:rolldown-vite@7.3.1",
```
TheAlexLichter, line 157: Note: You want to use vite@beta now

PR #10394, opened 2026-02-23
fix: thresholds working correctly with number panel

ahrefabhi:
## Summary

**Bug:**  
Thresholds for the number/value panel (Value panel) were not evaluated using the same unit-aware logic as the table panel, causing incorrect threshold matching when units differed (e.g. panel `yAxisUnit` vs threshold `thresholdUnit`) or when values required conversion.

**Fix:**  
Reused the shared, unit-conversion-aware threshold evaluation logic from the table panel in the number panel, and passed `yAxisUnit` through so single-value visualizations determine the active threshold consistently with tables.

---

### Root cause

- The table panel used `evaluateThresholdWithConvertedValue` (with proper unit conversion) via `findMatchingThreshold`.
- The number/value panel had its own comparison logic that worked directly on `rawValue` plus some formatting/parsing, and did **not** use the same conversion path.
- As a result, thresholds could be evaluated differently between table and number panels for the same metric and configuration — especially when threshold units differed from the displayed axis unit.

### User-visible impact

- The number panel could show the wrong threshold color/format (or no threshold) even when the table for the same query and thresholds showed a match.

---

### 1. Unify threshold evaluation logic

- Exported `evaluateThresholdWithConvertedValue` from `GridTableComponent/utils` so it can be reused outside the table.
- Updated `ValueGraph` threshold utilities to call `evaluateThresholdWithConvertedValue` instead of maintaining a separate comparison implementation.

### 2. Make number panel unit-aware in the same way as tables

- Threaded `yAxisUnit` through:
  - `GridValueComponent`
  - → `ValueGraph`
  - → `getBackgroundColorAndThresholdCheck`
- Threshold checks for the number panel now compare `rawValue` to the threshold value using the same unit conversion path as the table (`columnUnit` / `yAxisUnit` → `thresholdUnit`).

This ensures consistent behavior between panels.

### 3. Preserve existing behavior where appropriate

- Table panel behavior is unchanged except that its threshold evaluator is now exported for reuse.
- Threshold precedence (first matching threshold in configured order) and conflict detection (multiple matches) remain the same.
- These now operate on correctly converted values.

---

## Issues Closed by This PR

-  Closes https://github.com/SigNoz/engineering-pod/issues/3917 
---

### Blast radius

Frontend only; limited to:

- Number/value panel (`GridValueComponent` + `ValueGraph`)
- Shared threshold utility used by both table and number panels

### Potential regressions

- Misconfigured thresholds where units were previously (incorrectly) interpreted might now evaluate differently — but correctly.

### Rollback plan

Revert this PR if regressions are observed.

Changes are localized to:

- `GridTableComponent/utils.ts`
- `components/ValueGraph/utils.ts`
- `components/ValueGraph/index.tsx`
- `container/GridValueComponent/index.tsx`

---

## Changelog

**Deployment type:** Cloud / OSS / Enterprise  
**Change type:** Bug fix  

**Description:**  
Fixed inconsistent threshold behavior in the number/value panel by reusing the table panel’s unit-aware threshold evaluation logic and wiring `yAxisUnit` through the value panel stack.

### DIFF
--- a/frontend/src/components/YAxisUnitSelector/__tests__/YAxisUnitSelector.test.tsx
+++ b/frontend/src/components/YAxisUnitSelector/__tests__/YAxisUnitSelector.test.tsx
@@ -1,6 +1,7 @@
 import { fireEvent, render, screen } from '@testing-library/react';
 
-import { YAxisSource } from '../types';
+import { YAxisCategoryNames } from '../constants';
+import { UniversalYAxisUnit, YAxisSource } from '../types';
 import YAxisUnitSelector from '../YAxisUnitSelector';
 
 describe('YAxisUnitSelector', () => {
@@ -122,5 +123,35 @@ describe('YAxisUnitSelector', () => {
 		);
 		const warningIcon = screen.queryByLabelText('warning');
 		expect(warningIcon).not.toBeInTheDocument();
+	});
+
+	it('uses categories override to render custom units', () => {
+		const customCategories = [
+			{
+				name: YAxisCategoryNames.Data,
+				units: [
+					{
+						id: UniversalYAxisUnit.BYTES,
+						name: 'Custom Bytes (B)',
+					},
+				],
+			},
+		];
+
+		render(
+			<YAxisUnitSelector
+				value=""
+				onChange={mockOnChange}
+				source={YAxisSource.ALERTS}
+				categoriesOverride={customCategories}
+			/>,
+		);
+
+		const select = screen.getByRole('combobox');
+
+		fireEvent.mouseDown(select);
+
+		expect(screen.getByText('Custom Bytes (B)')).toBeInTheDocument();
+		expect(screen.queryByText('Bytes (B)')).not.toBeInTheDocument();
 	});
 });


frontend/src/components/YAxisUnitSelector/YAxisUnitSelector.tsx
```diff
@@ -66,10 +68,14 @@ function YAxisUnitSelector({
 		return aliases.some((alias) => alias.toLowerCase().includes(search));
 	};
 
-	const categories = getYAxisCategories(source);
+	const categoriesToRender = useMemo(() => {
```
AshwinBhatkal, line 71: How about calculating it and then passing the categories? Seems like source is passed just for the OR condition?

ahrefabhi: this is when someone wants to render a custom set of categories, otherwise source will the source of truth.
Like in case of threshold we just want to render a set of unit categories that's why added this.

AshwinBhatkal: nit: call it categoriesOverride?

PR #10399, opened 2026-02-23
feat(waterfall): return all span for small traces and nested level for large

krsoninikhil:
## Pull Request

---

- Add support for configurable pre unfurled levels while expanding a span
- Add support for returning all spans in one go upto a configured limit if client requets
- Ensure backward compatibility

https://github.com/user-attachments/assets/d8036300-e481-4071-973a-6795226dbb40

https://github.com/SigNoz/engineering-pod/issues/3845

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification: ✅ 
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/frontend/src/container/AlertHistory/Statistics/AverageResolutionCard/AverageResolutionCard.tsx
+++ b/frontend/src/container/AlertHistory/Statistics/AverageResolutionCard/AverageResolutionCard.tsx
@@ -16,9 +16,9 @@ function AverageResolutionCard({
 }: TotalTriggeredCardProps): JSX.Element {
 	return (
 		<StatsCard
-			displayValue={formatTime(currentAvgResolutionTime)}
-			totalCurrentCount={currentAvgResolutionTime}
-			totalPastCount={pastAvgResolutionTime}
+			displayValue={formatTime(+currentAvgResolutionTime)}
+			totalCurrentCount={+currentAvgResolutionTime}
+			totalPastCount={+pastAvgResolutionTime}
 			title="Avg. Resolution Time"
 			timeSeries={timeSeries}
 		/>
--- a/frontend/src/types/api/alerts/def.ts
+++ b/frontend/src/types/api/alerts/def.ts
@@ -56,8 +56,8 @@ export interface AlertRuleStats {
 	totalPastTriggers: number;
 	currentTriggersSeries: CurrentTriggersSeries;
 	pastTriggersSeries: CurrentTriggersSeries | null;
-	currentAvgResolutionTime: number;
-	pastAvgResolutionTime: number;
+	currentAvgResolutionTime: string;
+	pastAvgResolutionTime: string;
 	currentAvgResolutionTimeSeries: CurrentTriggersSeries;
 	pastAvgResolutionTimeSeries: any | null;
 }
--- a/frontend/src/utils/timeUtils.ts
+++ b/frontend/src/utils/timeUtils.ts
@@ -112,6 +112,12 @@ export function formatEpochTimestamp(epoch: number): string {
  */
 
 export function formatTime(seconds: number): string {
+	seconds = +seconds;
+
+	if (Number.isNaN(seconds)) {
+		return '-';
+	}
+
 	const days = seconds / 86400;
 
 	if (days >= 1) {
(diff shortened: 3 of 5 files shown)


pkg/query-service/app/traces/tracedetail/waterfall.go
```diff
@@ -126,15 +132,40 @@ func traverseTrace(span *model.Span, uncollapsedSpans []string, level uint64, is
 		preOrderTraversal = append(preOrderTraversal, &nodeWithoutChildren)
 	}
 
+	nextDepthFromSelectedSpan := -1
```
nityanandagohain: can you add some test cases ? It will help understand and verify this better

pkg/query-service/app/traces/tracedetail/waterfall.go
```diff
@@ -251,15 +260,13 @@ func GetSelectedSpans(uncollapsedSpans []string, selectedSpanID string, traceRoo
 }
 
 func GetAllSpans(traceRoots []*model.Span) (spans []*model.Span, rootServiceName, rootEntryPoint string) {
+	if len(traceRoots) > 0 {
+		rootServiceName = traceRoots[0].ServiceName
+		rootEntryPoint = traceRoots[0].Name
+	}
```
krsoninikhil, line 281: Can there be cases where a traceRoot might have empty service name or name? May be when multiple roots exists due to missing spans. This assignment was in the loop to account this possibility.

pkg/query-service/app/traces/tracedetail/waterfall.go
```diff
@@ -10,10 +10,11 @@ import (
 var (
 	SPAN_LIMIT_PER_REQUEST_FOR_WATERFALL float64 = 500
 
-	maxDepthForSelectedSpanChildren int  = 5
-	MaxLimitToSelectAllSpans        uint = 10_000
+	MAX_DEPTH_FOR_SELECTED_SPAN_CHILDREN int  = 5
+	MAX_LIMIT_TO_SELECT_ALL_SPANS        uint = 10_000
```
krsoninikhil: Are we not following Go style guide? https://google.github.io/styleguide/go/decisions.html#constant-names

pkg/query-service/app/traces/tracedetail/waterfall.go
nityanandagohain: If it's empty then it will be empty for other spans as well from the same resource, which means something is wrong with the instrumentation or maybe a bigger issue to look at ?

pkg/query-service/app/traces/tracedetail/waterfall.go
```diff
@@ -131,16 +145,36 @@ func traverseTrace(span *model.Span, opts traverseOpts, level uint64, isPartOfPr
...
 	_, isAlreadyUncollapsed := opts.uncollapsedSpans[span.SpanID]
 	for index, child := range span.Children {
-		_childTraversal := traverseTrace(child, opts, level+1, isPartOfPreOrder && isAlreadyUncollapsed, index != (len(span.Children)-1))
+		// A child is included in the pre-order output if its parent is uncollapsed
+		// OR if the child falls within MAX_DEPTH_FOR_SELECTED_SPAN_CHILDREN levels
+		// below the selected span.
+		isChildWithinMaxDepth := remainingAutoExpandDepth > 0
+		childIsPartOfPreOrder := opts.selectAll || (isPartOfPreOrder && (isAlreadyUncollapsed || isChildWithinMaxDepth))
+
+		if isPartOfPreOrder && isChildWithinMaxDepth && !isAlreadyUncollapsed {
```
nityanandagohain, line 163: If I change 161 and 162 to this i.e remove isPartOfPreOrder from checks, all tests still passed, any behavior we are missing to test  or it's unnecessary
```suggestion
				childIsPartOfPreOrder := opts.selectAll || isAlreadyUncollapsed || isChildWithinMaxDepth

		if isChildWithinMaxDepth && !isAlreadyUncollapsed {
```

krsoninikhil: After line 144 check, the method is not returning. So incase a node is not a part of traversal, the child nodes are still traversed. If we are removing the check from here, then after line 144, we can directly return.

krsoninikhil: Case where a parent is collapsed and child are present in uncollapsedIDs, then child spans will get added in the list which shouldn't be.

PR #10412, opened 2026-02-25
fix: outside click bug fix

ishan621:
## Pull Request
We encountered a bug where, when the log detail body contained valid JSON, clicking on body items (which provide filter-in/filter-out options) did not behave correctly. In the previous feature, we introduced outside-click handling for the drawer but missed covering this JSON body interaction case.

This PR fixes the issue and also strengthens the outside-click handling to prevent similar drawer-related bugs.

---

Fixing outside click bug on JSON body
we handled outside click using data-log-detail-ignore="true" attribute but JSON body introduces a tooltip dropdown which creates a new portal outside DOM. This made data-log-detail-ignore fail as it works in current DOM. We have handled popover and dropdown separately which causes new portal to be opened.

change. This helps reviewers quickly understand the impact and verify the update.

Before: 

https://github.com/user-attachments/assets/f4ff759a-71da-41d6-ae05-41ca2a3a3a6c

After:

https://github.com/user-attachments/assets/a1dc63c7-0267-4b0b-b7e5-e5a7c2c63058

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

Faulty assumption - We missed a case where log details can open a popover/dropdown outside DOM. We handled for DOM elements but missed case outside of DOM

We introduced popover class which we have added in log detail to handle outside click and same for dropdown also
---

- Blast radius: All users
- Potential regressions:
- Rollback plan: Commit revert

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/container/LogDetailedView/Overview.tsx
+++ b/frontend/src/container/LogDetailedView/Overview.tsx
@@ -1,7 +1,6 @@
-import { ReactNode, useCallback, useEffect, useRef, useState } from 'react';
+import { ReactNode, useState } from 'react';
 import MEditor, { EditorProps, Monaco } from '@monaco-editor/react';
 import { Color } from '@signozhq/design-tokens';
-import type { InputRef } from 'antd';
 import {
 	Button,
 	Collapse,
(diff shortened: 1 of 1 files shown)


YounixM:
@ishan621 - Should we be opening the drawer after filter / filter out ?

ishan621:
> @ishan621 - Should we be opening the drawer after filter / filter out ?

@YounixM  we have explicitly closed the drawer after apply any filters from log details

frontend/src/components/LogDetail/index.tsx
```diff
@@ -86,8 +86,13 @@ function LogDetailInner({
 		const handleClickOutside = (e: MouseEvent): void => {
 			const target = e.target as HTMLElement;
 
-			// Don't close if clicking on explicitly ignored regions
-			if (target.closest('[data-log-detail-ignore="true"]')) {
+			// Don't close if clicking on drawer content, overlays, or portal elements
+			if (
```
aks07, line 90: This approach does not seem extensible. While introducing new such elements we have to always add to this conditional check. We should think about a better approach later. For now its okay.

aks07 (approved):

PR #10415, opened 2026-02-25
chore: add initial version of query range design principles doc

srikanthccv:
## Pull Request

Adds the doc to describe the choices, and thought process behind the current query range implementation.


docs/contributing/query-range.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Invariant: The text format is `context.name:datatype`
+
+Parsing rules (implemented in `GetFieldKeyFromKeyText` and `Normalize`):
+
+1. Data type is extracted from the right, after the last `:`.
+2. Field context is extracted from the left, before the first `.`, if it matches a known context prefix.
+3. Everything remaining is the name.
+
+Special case: `log.body.X` normalizes to `{FieldContext: body, Name: X}` — the `log.body.` prefix collapses because body fields under log are a nested context.
```
tushar-signoz, line 47: **Note:** Using dot (.) as a separator introduces ambiguity about whether a segment belongs to the context or is part of the key name itself. For example, span.count would be normalized as context: span and name: count. However, the user might have intended attribute.span.count and simply omitted the attribute prefix. This is left to downstreams to handle.

docs/contributing/query-range.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Principle: Fields can be ambiguous until resolved
+
+The same name can map to multiple `TelemetryFieldKey` variants (different contexts, different types). The metadata store returns *all* variants. Resolution to a single field happens during query building, using the query's signal and any explicit context/type hints from the user.
+
+**Consequence:** Code that calls `GetKey` or `GetKeys` must handle multiple results. Do not assume a name maps to a single field.
+
+### Principle: Materialized fields are a performance optimization, not a semantic distinction
+
+A materialized field and its non-materialized equivalent represent the same logical field. The `Materialized` flag tells the FieldMapper to generate a simpler column expression. The user should never need to know whether a field is materialized.
```
tushar-signoz, line 195: Does it now make sense to leverage `Materialized` flag for this. FieldMapper should self contain this  information ?

srikanthccv: I didn't get it, it's already contained within the field mapper? Did you mean to say nobody should know about the `Materialized` nature of the field?

docs/contributing/query-range.md
srikanthccv: I will add the note.

docs/contributing/query-range.md
tushar-signoz: Yes, currently metadata store sets the materialised flag. Field Mapper should leverage evolutions to figure out what is the best/optimal way to fetch a field.

docs/contributing/query-range.md
```diff
@@ -0,0 +1,216 @@
...
+
+A field key is identified by three dimensions:
+
+- **Name** — the field name as the user knows it (`service.name`, `http.method`, `trace_id`)
+- **FieldContext** — where the field lives in the OTel model (`resource`, `attribute`, `span`, `log`, `body`, `scope`, `event`, `metric`)
+- **FieldDataType** — the data type (`string`, `bool`, `number`/`float64`/`int64`, array variants)
+
+### Invariant: Same name does not mean same field
+
+`status` as an attribute, `status` as a body JSON key, and `status` as a span-level field are **three different fields**. The context disambiguates. Code that resolves or compares field keys must always consider all three dimensions, never just the name.
```
nityanandagohain, line 31: Should we also add details about same key and different datatype here ?

nityanandagohain (commented):
Thanks for making the PR. I have one minor suggestion in this  and few improvements we can add in a new PR.

* Since we are adding about metadata layer should we add one for resource fingerprint layer ?
* Also suggestion on when to use hardcoded clickhouse settings and when to not like restricting max_memory of a certain query.
* Something that is must :- always use resource fingerprint table when there is a resource filter. 
* What kind of queries to avoid/be careful :- eg:- global join.

docs/contributing/query-range.md
nityanandagohain: How do we handle this currently ?

srikanthccv:
@nityanandagohain, the scope of this doc is different. It's for the authors who are going to touch the querier code to enable some feature. Your comments are about the end users who are going to write queries to consume querier. They are separate things. We will have a separate doc for the things you are talking about. Imagine an agent is tasked to add a feature to querier, it needs to get these things right and then code it writes will be feature specific so there is no generic guide because it's context dependent. Let me know if this is clear.

docs/contributing/query-range.md
srikanthccv: `materialised` is the property of the key, I am not following why it should be moved from metadata to field mapper.

nityanandagohain:
Thanks for the clarification 👍  it's clear now.

PR #10421, opened 2026-02-25
fix: throttled legend color picker in dashboard + memory leak fix due to tooltip persistance

ahrefabhi:
## Summary

This PR fixes a memory leak and reduces unnecessary work in the widgets/visualization area by:

- **Throttling legend color updates** in `LegendColors` so rapid color changes from the color picker do not spam React state updates.
- **Cleaning up the throttled handler** on unmount so no pending callbacks retain references after the component is gone.
- **Hardening the uPlot tooltip plugin** by:
  - Ensuring the layout observer is disconnected before re‑observing the container.
  - Keeping the tooltip portal mounted but visually hidden when not hovering, instead of mounting/unmounting it on every hover.
  - Using CSS classes and ARIA attributes to control visibility, pointer events, and accessibility without destroying the DOM node.

Closes https://github.com/SigNoz/engineering-pod/issues/4083

---

### Root Cause

- **Legend color picker**
  - The color picker can emit many `onChange` events while the user drags the selector.
  - Each event directly updated `customLegendColors`, causing excessive state updates and potential memory/CPU pressure when multiple charts are present.

- **Tooltip plugin**
  - The tooltip layout observer could be reused without an explicit `disconnect()` call before starting a new observation, risking lingering observations.
  - The tooltip portal was tied directly to the hover state, so frequent hover transitions caused mount/unmount cycles, increasing churn and the chance of observers or references not being fully released over time.

### Fix Strategy

- **LegendColors**
  - Wraps `handleColorChange` in a `useMemo` + `lodash-es/throttle` with a 200 ms window so rapid color changes coalesce into fewer state updates.
  - Adds a `useEffect` cleanup that calls `handleColorChange.cancel()` when the component unmounts (or the memoized handler changes), preventing pending throttled callbacks from firing after unmount.

- **TooltipPlugin**
  - In the layout effect, calls `layout.observer.disconnect()` before `layout.observer.observe(containerRef.current)` so existing observations are always cleaned up before a new one is registered.
  - Changes the render logic to:
    - Return `null` only when `plot` is not available, not when the user is simply not hovering.
    - Always render the tooltip container when `plot` exists, but:
      - Toggle a `visible` class based on `isHovering`.
      - Use CSS (`opacity: 0`, `pointer-events: none`) to hide and disable interaction when not hovering.
      - Use `aria-hidden={!isHovering}` and render `contents` only when `isHovering` is true to reduce work when the tooltip is idle.
  - This avoids unnecessary mount/unmount cycles, ensures observers are properly managed, and reduces the risk of leaks during long-lived dashboard sessions.

---

## Testing Strategy

- **Automated tests**
  - Adds `LegendColors.test.tsx` with scenarios:
    - **Renders legend colors panel and items** using a mocked query builder, asserting that the legend text (e.g. `{service.name}`) appears after expanding the panel.
    - **Calls `setCustomLegendColors` when color changes**:
      - Mocks `antd`’s `ColorPicker` as a simple button that triggers `onChange` with a fixed color.
      - Confirms `setCustomLegendColors` is called on click.
    - **Throttles rapid color changes**:
      - Uses fake timers and `userEvent` configured with `advanceTimers`.
      - Simulates multiple rapid clicks on the mocked color picker.
      - Advances timers and asserts `setCustomLegendColors` is called fewer times than the number of clicks, confirming the throttling behavior.

- **Manual verification (expected)**
  - Open a dashboard or widget using `LegendColors`:
    - Drag the legend color picker quickly and verify that the UI remains responsive while reducing excessive renders.
  - Interact with charts using the tooltip:
    - Hover/unhover rapidly; ensure the tooltip appears and disappears correctly, remains stable when pinned, and does not jitter.

---

## Risk & Impact

- **Blast radius**
  - Limited to:
    - The legend color customization UI in the new widget right-hand panel.
    - The uPlot tooltip plugin used for chart tooltips.

- **Potential regressions**
  - Legend color updates might feel slightly less instantaneous under extremely rapid drag interactions due to the 200 ms throttling window, though normal usage should be unaffected.
  - Tooltip behavior now relies on CSS visibility instead of mount/unmount semantics; any logic that assumed unmount on hover-change could behave differently, although this is not expected in typical usage.

- **Rollback plan**
  - Revert this PR if issues are observed; changes are localized to `LegendColors` and the tooltip plugin component/styles.

---

## Changelog

- **Change Type**: Bug Fix / Performance  
- **Description**: Fixes a memory leak and excessive updates caused by unthrottled legend color changes and tooltip observer handling, improving stability and performance for charts and dashboards.

### DIFF
--- a/frontend/src/container/MetricsExplorer/Explorer/Explorer.tsx
+++ b/frontend/src/container/MetricsExplorer/Explorer/Explorer.tsx
@@ -349,7 +349,7 @@ function Explorer(): JSX.Element {
 				isOneChartPerQuery={showOneChartPerQuery}
 				splitedQueries={splitedQueries}
 			/>
-			{isMetricDetailsOpen && (
+			{isMetricDetailsOpen && selectedMetricName && (
 				<MetricDetails
 					metricName={selectedMetricName}
 					isOpen={isMetricDetailsOpen}
--- a/frontend/src/container/MetricsExplorer/Explorer/RelatedMetricsCard.tsx
+++ b/frontend/src/container/MetricsExplorer/Explorer/RelatedMetricsCard.tsx
@@ -39,10 +39,7 @@ function RelatedMetricsCard({ metric }: RelatedMetricsCardProps): JSX.Element {
 					dataSource={DataSource.METRICS}
 				/>
 			)}
-			<DashboardsAndAlertsPopover
-				dashboards={metric.dashboards}
-				alerts={metric.alerts}
-			/>
+			<DashboardsAndAlertsPopover metricName={metric.name} />
 		</div>
 	);
 }
--- a/frontend/src/container/MetricsExplorer/Inspect/utils.tsx
+++ b/frontend/src/container/MetricsExplorer/Inspect/utils.tsx
@@ -2,8 +2,8 @@
 import { useMemo, useState } from 'react';
 import { Card, Input, Select, Typography } from 'antd';
 import logEvent from 'api/common/logEvent';
+import { MetrictypesTypeDTO } from 'api/generated/services/sigNoz.schemas';
 import { InspectMetricsSeries } from 'api/metricsExplorer/getInspectMetricsDetails';
-import { MetricType } from 'api/metricsExplorer/getMetricsList';
 import classNames from 'classnames';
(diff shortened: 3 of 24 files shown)


srikanthccv:
Please update the PR title.

PR #10428, opened 2026-02-26
Feat(Infra-Monitoring): Fetch Metadata using all metrics in use in the list API

nikhilmantri0902:
## Pull Request

This PR will be ready for review once we merge https://github.com/SigNoz/signoz/pull/10304 this PR

---

In all the /list APIs in infra-monitoring, the metadata like host.name, os.type, pod.name, node.name, pod.uid etc all these are fetched from one specific predefined metric. So if that one metric is absent we do not have metadata for the list API and it does not return host.name, pod.name, node.name etc. , which is of less use. It also breaks the sidedrawers charts because they require the identifiers to be passed in the query range payload. 

In this PR, we fetch metadata using the whole bunch of metrics that are in use in the API to get oob numbers for the list. 
We select all distinct sets of metadata from all these metrics combined, thus making the logic more full proof and lower the chance of 500 errors in the side-drawer 

We will fully be eradicating these 500 side-drawer errors by disabling the sidedrawer if identifier is still absent. But first we should be improving the logic to fetch metadata with this. 

Part of https://github.com/SigNoz/engineering-pod/issues/3354

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Manual verification:

---

- Blast radius: Infra Monitoring all list APIs and UI sections

---

| Field | Value |
|------|-------|
| Deployment Type | Cloud + OSS + Enterprise |
| Change Type | Feature |
| Description | Improvement in metadata fetching logic on UI |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---


srikanthccv:
outdated with the ongoing changes.

PR #10429, opened 2026-02-26
fix: add validity check for spatial aggregation

namanv3:
## Pull Request

---

Currently if someone enters a wrong spatial aggregation (for example `count_distinct`), then signoz returns a generic error message saying that something went wrong.
This change adds a validity check on the spatial aggregation provided, along with a minor check on temporal aggregation that was missing.

Closes https://github.com/SigNoz/engineering-pod/issues/4074

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/frontend/src/AppRoutes/routes.ts
+++ b/frontend/src/AppRoutes/routes.ts
@@ -1,7 +1,6 @@
 import { RouteProps } from 'react-router-dom';
 import ROUTES from 'constants/routes';
 
-import { createGuardedRoute } from '../components/createGuardedRoute/createGuardedRoute';
 import {
 	AlertHistory,
 	AlertOverview,
@@ -85,11 +84,7 @@ const routes: AppRoutes[] = [
 	{
 		path: ROUTES.HOME,
 		exact: true,
-		component: createGuardedRoute(
-			Home,
-			'delete',
-			'dashboard:e6dbc08b-976d-4c41-8572-93c990c3b297',
-		),
+		component: Home,
 		isPrivate: true,
 		key: 'HOME',
 	},


pkg/telemetrymetrics/statement_builder.go
```diff
@@ -159,6 +159,23 @@ func (b *MetricQueryStatementBuilder) buildPipelineStatement(
 			query.Aggregations[0].TimeAggregation = metrictypes.TimeAggregationIncrease
 		}
 		query.Aggregations[0].SpaceAggregation = metrictypes.SpaceAggregationSum
+
+		// check for origTimeAgg's and origSpaceAgg's validity
+		if origTimeAgg.IsZero() || !origTimeAgg.IsValid() {
```
srikanthccv: 1. we don't need time aggregation to be specified for the histogram percentile requests
2. we don't need it to be `rate/increase` only as we always use the time aggregation of `rate`/`increase` ourselves, and the original time aggregation is not used at all in query preparation, so it doesn't make sense to reject the request with bad input.

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -159,6 +159,23 @@ func (b *MetricQueryStatementBuilder) buildPipelineStatement(
...
+			return nil, errors.Newf(
+				errors.TypeInvalidInput,
+				errors.CodeInvalidInput,
+				"invalid time aggregation, should be one of the following: [`rate`, `increase`]",
```
srikanthccv: Irrelevant point give the above point, but the validation does something different and message says something else.

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -159,6 +159,23 @@ func (b *MetricQueryStatementBuilder) buildPipelineStatement(
...
+			)
+		}
+
+		if origSpaceAgg.IsZero() || !origSpaceAgg.IsValid() {
+			return nil, errors.Newf(
+				errors.TypeInvalidInput,
+				errors.CodeInvalidInput,
+				"invalid space aggregation, should be one of the following: [`count`, `p50`, `p75`, `p90`, `p95`, `p99`]",
+			)
```
srikanthccv: If the goal is to enforce a generic validation on spatial aggregation (instead of contextual validation based on metric type), the validation should go here https://github.com/SigNoz/signoz/blob/95fc905448c26123d34f8083c983bba23a2cb18f/pkg/types/querybuildertypes/querybuildertypesv5/validation.go#L206

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -523,11 +540,26 @@ func (b *MetricQueryStatementBuilder) buildSpatialAggregationCTE(
 	query qbtypes.QueryBuilderQuery[qbtypes.MetricAggregation],
 	_ map[string][]*telemetrytypes.TelemetryFieldKey,
 ) (string, []any, error) {
-	if query.Aggregations[0].SpaceAggregation.IsZero() {
+	if query.Aggregations[0].SpaceAggregation.IsZero() || !query.Aggregations[0].SpaceAggregation.IsValid() {
+		if query.Aggregations[0].Type.IsPercentileSpaceAggregationAllowed() {
+			return "", nil, errors.Newf(
+				errors.TypeInvalidInput,
+				errors.CodeInvalidInput,
+				"invalid space aggregation, should be one of the following: [`sum`, `avg`, `min`, `max`, `count`, `p50`, `p75`, `p90`, `p95`, `p99`]",
```
srikanthccv: What does it mean to allow `sum`, `avg`, `min`, `max` on histogram / exp histogram / summary metric?

pkg/telemetrymetrics/statement_builder.go
namanv3: done, kept just a generic check for now cuz even in time agg we allow avg/sum/etc for counter metrics

PR #10436, opened 2026-02-26
feat(service-account): introduce service account 

vikrantgupta25:
### 📄 Summary
-  introduce service account 

contributes to: https://github.com/SigNoz/platform-pod/issues/1696


vikrantgupta25:
will be removing the sqlmigrations from this PR before merge, will add them with the integrations PR ( authz middleware and identity changes )

frontend/src/api/generated/services/sigNoz.schemas.ts
```diff
@@ -2090,6 +2090,154 @@ export interface RoletypesRoleDTO {
 	updatedAt?: Date;
 }
 
+export interface ServiceaccounttypesFactorAPIKeyDTO {
+	/**
+	 * @type string
+	 * @format date-time
+	 */
+	createdAt?: Date;
```
AshwinBhatkal, line 2098: should this always be required? Not be optional

docs/api/openapi.yml
```diff
@@ -1763,6 +1763,134 @@ components:
...
+          format: date-time
+          type: string
+        name:
+          type: string
+        service_account_id:
+          type: string
+        updatedAt:
+          format: date-time
+          type: string
+      required:
```
AshwinBhatkal, line 1788: createdAt and updatedAt to be required?

vikrantgupta25: this requires a root level change, not picking up with this

AshwinBhatkal: Same as success response?

PR #10445, opened 2026-02-27
feat: deprecate user invite table and add user status lifecycle

karanbalani:
### 📄 Summary
- Deprecates the `user_invite` table in favor of a `status` lifecycle on the `users` table (`pending_invite`, `active`, `deleted`), unifying invite and user management into a single table
- Adds a `deleted_at` timestamp column and replaces the `(email, org_id)` unique index with a composite `(org_id, email, deleted_at)` unique index — active/pending users share `deleted_at=zero` (enforcing uniqueness), while each soft-deleted row gets a unique deletion timestamp (allowing re-invitation)
- New invite flow: `admin invites user` → `user row created with status=pending_invite` → `reset password email sent` → `user resets password / SSO login` → `user activated with authz grants`
- Soft delete: sets `status=deleted` and `deleted_at=now()`, revokes authz, and cleans up credentials/tokens/preferences
- All existing invite APIs kept with backward compatibility, now backed by the new user-status flow

### ⁉️ What changed?

**User status lifecycle:**
- `pending_invite` → user has been invited but hasn't set their password yet
- `active` → user has completed onboarding (via password reset or SSO)
- `deleted` → user has been soft deleted

**Invite flow (CreateBulkInvite):**
- Creates user with `pending_invite` status using `createUserWithoutGrant` (**no authz until active**)
- Sends reset password email via existing `GetOrCreateResetPasswordToken` (which also creates `factor_password` if missing)
- Duplicate-email validation added to the bulk invite request payload
- Re-inviting a soft-deleted user creates a new row in DB as `pending_invite`

**Activation paths:**
- Password reset: `UpdatePasswordByResetPasswordToken` grants authz and sets status to `active` when status is `pending_invite` (covers both invite accept and expired-token forgot-password flows)
- SSO login: `GetOrCreateUser` grants authz and activates `pending_invite` users or creates new ones

**Soft delete:**
- `SoftDeleteUser` sets `status=deleted` and `deleted_at=time.Now()`, revokes authz, and deletes credentials, API keys, preferences, and tokens
- Authentication and user listings filter out deleted and pending users

**Migrations:**
- 067: Adds `status TEXT NOT NULL DEFAULT 'active'` and `deleted_at TIMESTAMP NOT NULL DEFAULT zero-time` columns to `users` table; drops old `(email, org_id)` unique index; creates composite unique index `(org_id, email, deleted_at)`
- 068 (idempotent): Moves existing pending invites from `user_invite` into `users` as `pending_invite` rows, then drops the `user_invite` table; skips gracefully if `user_invite` no longer exists

**Auth/session changes:**
- Login and session validation now require `status=active`; pending and deleted users are rejected at authentication
- `GetNonDeletedUserByEmailAndOrgID` enforces at most one non-deleted user per `(email, org_id)`

**Store query changes:**
- `ActiveCountByOrgID` (renamed) now filters by `status=active`
- `GetActiveUsersByRoleAndOrgID` filters by `status=active`
- `GetUsersByEmailsOrgIDAndStatuses` used for duplicate-invite checking

---
### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

> [!NOTE]
> **High Risk**
> Touches authentication, user provisioning, and deletion semantics, and ships schema migrations (new columns, new unique index, dropping `user_invite`). Errors here could block logins/invites or corrupt user state, so rollout and migration testing are critical.
> 
> **Overview**
> Unifies invites and user lifecycle by adding `status`/`deleted_at` to `users` (`pending_invite`/`active`/`deleted`) and migrating away from the `user_invite` table. The invite flow now creates a `pending_invite` user and uses a reset-password token/link (including email template updates) to activate and grant access; legacy invite APIs remain but are backed by this new mechanism.
> 
> Introduces soft-delete semantics for users (status flip + cleanup of passwords/tokens/preferences/API keys, and admin/last-admin safeguards) and updates authentication/session/user listing logic to exclude non-`active` users. Updates the OpenAPI/TS client for `POST /api/v1/invite/bulk` to accept `{ invites: [...] }`, adds `status` to user responses, and adds new SQL migrations (status columns + composite unique index `(org_id,email,deleted_at)` + invite-table deprecation) plus expanded integration tests for invite/SSO/deleted-user edge cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 23cbc43a09017295d41d921397071029fac3873e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -2113,7 +2113,6 @@ components:
         invites:
           items:
             $ref: '#/components/schemas/TypesPostableInvite'
-          nullable: true
           type: array
       required:
       - invites
--- a/pkg/types/invite.go
+++ b/pkg/types/invite.go
@@ -54,7 +54,7 @@ type PostableInvite struct {
 }
 
 type PostableBulkInviteRequest struct {
-	Invites []PostableInvite `json:"invites" required:"true"`
+	Invites []PostableInvite `json:"invites" required:"true" nullable:"false"`
 }
 
 func (request *PostableBulkInviteRequest) UnmarshalJSON(data []byte) error {


karanbalani:
Related Issue: https://github.com/SigNoz/platform-pod/issues/1724

pkg/modules/user/impluser/store.go
```diff
@@ -574,6 +504,7 @@ func (store *store) CountByOrgID(ctx context.Context, orgID valuer.UUID) (int64,
 		NewSelect().
 		Model(user).
 		Where("org_id = ?", orgID).
+		Where("status = ?", types.UserStatusActive.StringValue()).
```
karanbalani: @therealpandey - need a confirmation here, we can either filter here or delicate this logic to the caller, currently this is used for stats.

This is also used to control the `setup_complete` variable, there also we target at least one active user.

pkg/authn/authnstore/sqlauthnstore/store.go
```diff
@@ -28,6 +28,7 @@ func (store *store) GetUserAndFactorPasswordByEmailAndOrgID(ctx context.Context,
 		Model(user).
 		Where("email = ?", email).
 		Where("org_id = ?", orgID).
+		Where("status = ?", types.UserStatusActive).
```
karanbalani: @therealpandey - we need this check here, but this does not align with the name of the function, shall we change the name to `GetActiveUserAndFactorPasswordByEmailAndOrgID` ?

karanbalani: Took a call here. Just let me know if this needs to be reverted.

pkg/modules/user/impluser/store.go
karanbalani: Moving this to `ActiveCountByOrgID` from `CountByOrgID`.

pkg/sqlmigration/068_deprecate_user_invite.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (migration *deprecateUserInvite) Register(migrations *migrate.Migrations) error {
+	if err := migrations.Register(migration.Up, migration.Down); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+type userInviteRow struct {
```
vikrantgupta25, line 21: move types to the top

pkg/sqlmigration/068_deprecate_user_invite.go
```diff
@@ -0,0 +1,141 @@
...
+		if err != nil {
+			return err
+		}
+
+		if existingCount > 0 {
+			continue
+		}
+
+		user := &pendingInviteUser{
+			ID:          invite.ID,
```
vikrantgupta25, line 114: why not create new uuid here ?

pkg/sqlmigration/068_deprecate_user_invite.go
```diff
@@ -0,0 +1,141 @@
...
+			Email:       invite.Email,
+			Role:        invite.Role,
+			OrgID:       invite.OrgID,
+			IsRoot:      false,
+			Status:      "pending_invite",
+			CreatedAt:   invite.CreatedAt,
+			UpdatedAt:   time.Now(),
+		}
+
+		if _, err = tx.NewInsert().Model(user).Exec(ctx); err != nil {
```
vikrantgupta25: please convert this to a bulk insert

pkg/query-service/app/http_handler.go
```diff
@@ -253,7 +253,7 @@ func NewAPIHandler(opts APIHandlerOpts, config signoz.Config) (*APIHandler, erro
 	}
 	// if the first org with the first user is created then the setup is complete.
 	if len(orgs) == 1 {
-		count, err := opts.Signoz.Modules.UserGetter.CountByOrgID(context.Background(), orgs[0].ID)
+		count, err := opts.Signoz.Modules.UserGetter.ActiveCountByOrgID(context.Background(), orgs[0].ID)
```
vikrantgupta25: if there is only one user, the application checks shouldn't let it be marked as inactive. Like we prevent deletion of last user. Having an inactive user doesn't warrant a new org creation

karanbalani: Right, I'll make a separate method to handle this as `CountByOrgID` just like it is right now
 in prod, which will not care about status.

pkg/modules/user/impluser/module.go
```diff
@@ -95,80 +127,174 @@ func (m *Module) CreateBulkInvite(ctx context.Context, orgID valuer.UUID, userID
...
 	}
 
-	invites := make([]*types.Invite, 0, len(bulkInvites.Invites))
+	invitedUsers := make([]*types.User, 0, len(bulkInvites.Invites))
+	var invites []*types.Invite // TODO(balanikaran) remove this and more to types.User as return
 
 	for _, invite := range bulkInvites.Invites {
-		// check if user exists
-		existingUser, err := m.store.GetUserByEmailAndOrgID(ctx, invite.Email, orgID)
+		// check and active user already exists with this email
```
vikrantgupta25:
```suggestion
		// check if an active user already exists with this email
```

pkg/sqlmigration/068_deprecate_user_invite.go
karanbalani: No specific reason, just recycled the ID.

pkg/modules/user/impluser/module.go
```diff
@@ -598,3 +750,49 @@ func (module *Module) createUserWithoutGrant(ctx context.Context, input *types.U
...
+
+// this function restricts that only one non-deleted user email can exist for an org ID, if found more, it throws an error
+func (module *Module) GetNonDeletedUserByEmailAndOrgID(ctx context.Context, email valuer.Email, orgID valuer.UUID) (*types.User, error) {
+	existingUsers, err := module.store.GetUsersByEmailAndOrgID(ctx, email, orgID)
+	if err != nil {
+		return nil, err
+	}
+
+	// filter out the deleted users
+	existingUsers = slices.DeleteFunc(existingUsers, func(user *types.User) bool { return user.Status == types.UserStatusDeleted })
```
vikrantgupta25: what is the reason for not having this filtering on store layer instead ?

pkg/modules/user/impluser/module.go
```diff
@@ -95,80 +127,174 @@ func (m *Module) CreateBulkInvite(ctx context.Context, orgID valuer.UUID, userID
 		return nil, err
 	}
 
-	invites := make([]*types.Invite, 0, len(bulkInvites.Invites))
+	invitedUsers := make([]*types.User, 0, len(bulkInvites.Invites))
```
vikrantgupta25: `CreateBulkInvite` used to transactionally either create all the invites or none. Now it's split into multiple DB calls none of which are wrapped in a transaction and the createResetPasswordToken generation is skipped for an invite if frontendBaseURL is absent ? 


Please correct this

pkg/modules/user/impluser/module.go
```diff
@@ -598,3 +750,49 @@ func (module *Module) createUserWithoutGrant(ctx context.Context, input *types.U
 
 	return nil
 }
+
+func (module *Module) resetLink(frontendBaseUrl string, token string) string {
```
vikrantgupta25: why is this a module function and not in types ?

pkg/modules/user/impluser/module.go
```diff
@@ -51,40 +51,72 @@ func NewModule(store types.UserStore, tokenizer tokenizer.Tokenizer, emailing em
 	}
 }
 
+// TODO(balanikaran): deprecate this one frontend changes are live with new invitation flow
 func (m *Module) AcceptInvite(ctx context.Context, token string, password string) (*types.User, error) {
-	invite, err := m.store.GetInviteByToken(ctx, token)
+	// token in this case is the reset password token
+	resetPasswordToken, err := m.store.GetResetPasswordToken(ctx, token)
```
vikrantgupta25: why is there double work in this function ? when UpdatePasswordByResetPasswordToken does all of this ?

pkg/modules/user/impluser/module.go
```diff
@@ -443,6 +580,13 @@ func (module *Module) UpdatePasswordByResetPasswordToken(ctx context.Context, to
 		return err
 	}
 
+	// update the status of user if this a newly invited user and also grant authz
+	if user.Status == types.UserStatusPendingInvite {
+		if err = module.activatePendingUser(ctx, user); err != nil {
```
vikrantgupta25: this is again not happening in a transaction

pkg/sqlmigration/068_deprecate_user_invite.go
karanbalani: Done

pkg/modules/user/impluser/module.go
```diff
@@ -95,80 +127,174 @@ func (m *Module) CreateBulkInvite(ctx context.Context, orgID valuer.UUID, userID
...
+
+	// get the user
+	user, err := m.store.GetUser(ctx, valuer.MustNewUUID(factorPassword.UserID))
+	if err != nil {
+		return err
+	}
+
+	// revoke grants
+	// since revoke is idempotant multiple calls to revoke won't cause issues in case of retries
+	err = m.authz.Revoke(ctx, user.OrgID, []string{roletypes.MustGetSigNozManagedRoleFromExistingRole(user.Role)}, authtypes.MustNewSubject(authtypes.TypeableUser, user.ID.StringValue(), user.OrgID, nil))
```
vikrantgupta25: there is no check here that the user was in pending_invite state. and why do we need authz revoke here when we are assigning them grants when the move to active ?

pkg/modules/user/impluser/module.go
karanbalani: 1. That would need us to rename all the functions to denote that there is a filter active status
2. It would be better to keep the logic in App layer, how the caller wants to handle users with different statuses

pkg/modules/user/impluser/module.go
karanbalani: I will wrap things into a transaction, but not having a frontendBaseURL currently also skips the user creation.
That's the expected flow also.

pkg/modules/user/impluser/module.go
karanbalani: Due to the return type of this function, I need to somehow figure out the user.

pkg/modules/user/impluser/module.go
karanbalani: Will move

pkg/modules/user/impluser/module.go
karanbalani: Just followed the same flow how delete user flows. I'll remove this.
And hard delete the user only if it is in pending state.

pkg/modules/user/impluser/module.go
```diff
@@ -224,7 +352,7 @@ func (m *Module) UpdateUser(ctx context.Context, orgID valuer.UUID, id string, u
 
 	// Make sure that the request is not demoting the last admin user.
 	if user.Role != "" && user.Role != existingUser.Role && existingUser.Role == types.RoleAdmin {
-		adminUsers, err := m.store.GetUsersByRoleAndOrgID(ctx, types.RoleAdmin, orgID)
+		adminUsers, err := m.store.GetActiveUsersByRoleAndOrgID(ctx, types.RoleAdmin, orgID)
```
vikrantgupta25, line 323: there are no domain checks to not update the user when in pending_invite or deleted state.

pkg/modules/user/impluser/module.go
vikrantgupta25: add another function in the store layer to fetch the user directly rather than 4 round trip DB calls

pkg/modules/user/impluser/module.go
```diff
@@ -598,3 +754,49 @@ func (module *Module) createUserWithoutGrant(ctx context.Context, input *types.U
 
 	return nil
 }
+
+func (module *Module) resetLink(frontendBaseUrl string, token string) string {
+	return fmt.Sprintf("%s/password-reset?token=%s", frontendBaseUrl, token)
+}
+
+func (module *Module) activatePendingUser(ctx context.Context, user *types.User) error {
+	err := module.authz.Grant(
```
vikrantgupta25, line 793: the authz checks won't work if run in a transaction

pkg/modules/user/impluser/module.go
vikrantgupta25: https://github.com/SigNoz/signoz/blob/main/pkg/modules/user/impluser/module.go#L145 it doesn't, It just doesn't send the mail

pkg/modules/user/impluser/module.go
vikrantgupta25: the caller can always fetch based on status they are interested in. Why to fetch unnecessary data if its gonna get filtered anyways ?

pkg/modules/user/impluser/module.go
karanbalani: Didn't understand this, please elaborate a little.

vikrantgupta25: A user cannot be updated when their state is `deleted`

pkg/modules/user/impluser/module.go
karanbalani: Fixed.

pkg/query-service/app/http_handler.go
karanbalani: Add `total`, `active`, `pending_status` and `deleted` count in stats.

pkg/modules/user/impluser/module.go
karanbalani: Moved to factor_password.go

pkg/modules/user/impluser/module.go
karanbalani: Move this to soft deleted status.

pkg/modules/user/impluser/module.go
karanbalani: Yep I meant, we are also not sending email, but creation does happen. Cause the creation already happened before in the TXN.

pkg/modules/user/impluser/module.go
karanbalani: Created a `GetUserByResetPasswordToken` method which uses joins to query data.

pkg/types/user.go
```diff
@@ -67,6 +75,7 @@ func NewUser(displayName string, email valuer.Email, role Role, orgID valuer.UUI
 		Role:        role,
 		OrgID:       orgID,
 		IsRoot:      false,
+		Status:      status,
```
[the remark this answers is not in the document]
karanbalani: Added already

pkg/modules/user/impluser/module.go
```diff
@@ -300,7 +427,9 @@ func (module *Module) DeleteUser(ctx context.Context, orgID valuer.UUID, id stri
 		return err
 	}
 
-	if err := module.store.DeleteUser(ctx, orgID.String(), user.ID.StringValue()); err != nil {
+	// for now we are only soft deleting users
```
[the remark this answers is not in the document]
karanbalani: Revoke is idempotent and for DeleteUser, we will be doing all the deeds.

pkg/query-service/app/http_handler.go
karanbalani: This is done, please check the naming conventions if that needs any update.

pkg/modules/user/impluser/module.go
karanbalani: This is done.

pkg/modules/user/impluser/module.go
```diff
@@ -51,38 +50,57 @@ func NewModule(store types.UserStore, tokenizer tokenizer.Tokenizer, emailing em
...
 
-	user, err := types.NewUser(invite.Name, invite.Email, invite.Role, invite.OrgID)
+	// update the password and delete the token
+	err = m.UpdatePasswordByResetPasswordToken(ctx, token, password)
 	if err != nil {
 		return nil, err
 	}
 
-	factorPassword, err := types.NewFactorPassword(password, user.ID.StringValue())
+	return user, nil
```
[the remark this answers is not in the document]
karanbalani: Frontend is not using this field, and when it will do, this API will be deprecated.

frontend/src/api/generated/services/sigNoz.schemas.ts
```diff
@@ -2665,6 +2665,10 @@ export interface TypesUserDTO {
 	 * @type string
 	 */
 	role?: string;
+	/**
+	 * @type string
+	 */
+	status?: string;
```
vikrantgupta25, line 2678: why is status not required ?

pkg/apiserver/signozapiserver/user.go
```diff
@@ -43,6 +43,7 @@ func (provider *provider) addUserRoutes(router *mux.Router) error {
 		return err
 	}
 
+	// TODO(balanikaran): deprecate this one frontend changes are live with new invitation flow
```
vikrantgupta25: please do not add TODOs in code, create a separate ticket for cleaning this tech-debt

cmd/enterprise/Dockerfile.with-web.integration
```diff
@@ -1,4 +1,4 @@
-FROM node:18-bullseye AS build
+FROM node:22-bookworm AS build
```
vikrantgupta25, line 1: this is required because of node version frontend upgrade ?

pkg/modules/user/impluser/getter.go
```diff
@@ -86,6 +86,15 @@ func (module *getter) CountByOrgID(ctx context.Context, orgID valuer.UUID) (int6
 	return count, nil
 }
 
+func (module *getter) CountByOrgIDGroupedByStatus(ctx context.Context, orgID valuer.UUID, statuses []string) (map[string]int64, error) {
```
vikrantgupta25:
```suggestion
func (module *getter) CountByOrgIDAndStatuses(ctx context.Context, orgID valuer.UUID, statuses []string) (map[valuer.String]int64, error) {
```

pkg/modules/user/impluser/handler.go
```diff
@@ -137,28 +140,24 @@ func (h *handler) ListInvite(w http.ResponseWriter, r *http.Request) {
 	render.Success(w, http.StatusOK, invites)
 }
 
+// TODO(balanikaran): deprecate this one frontend changes are live with new invitation flow
 func (h *handler) DeleteInvite(w http.ResponseWriter, r *http.Request) {
 	ctx, cancel := context.WithTimeout(r.Context(), 10*time.Second)
 	defer cancel()
 
-	id := mux.Vars(r)["id"]
+	id := mux.Vars(r)["id"] // this is now the pending user id
```
vikrantgupta25:
```suggestion
	id := mux.Vars(r)["id"] 
```

a pending_user is no different from an active user let's not differentiate please

pkg/modules/user/impluser/store.go
```diff
@@ -638,3 +663,22 @@ func (store *store) ListUsersByEmailAndOrgIDs(ctx context.Context, email valuer.
...
+
+func (store *store) GetUserByResetPasswordToken(ctx context.Context, token string) (*types.User, error) {
+	user := new(types.User)
+
+	err := store.
+		sqlstore.
+		BunDBCtx(ctx).
+		NewSelect().
+		Model(user).
+		Join("JOIN factor_password ON factor_password.user_id = \"user\".id").
```
vikrantgupta25:
```suggestion
		Join("JOIN factor_password ON factor_password.user_id = user.id").
```

pkg/modules/user/user.go
```diff
@@ -75,20 +77,23 @@ type Getter interface {
 	// List users by email and org ids.
 	ListUsersByEmailAndOrgIDs(context.Context, valuer.Email, []valuer.UUID) ([]*types.User, error)
 
-	// Count users by org id.
+	// Count users by org id irrespective of the user status.
```
vikrantgupta25:
```suggestion
	// Count users by org id 
```

vikrantgupta25: since we are already prefixing when status is involved

pkg/types/factor_password.go
```diff
@@ -220,3 +221,7 @@ func comparePassword(hashedPassword string, password string) bool {
 func (r *ResetPasswordToken) IsExpired() bool {
 	return r.ExpiresAt.Before(time.Now())
 }
+
+func FactorPasswordResetLink(frontendBaseUrl string, token string) string {
```
vikrantgupta25: why a floating function? Can this be a method on `ResetPasswordToken`

pkg/types/user.go
```diff
@@ -111,6 +127,11 @@ func (u *User) Update(displayName string, role Role) {
 	u.UpdatedAt = time.Now()
 }
 
+func (u *User) UpdateStatus(status valuer.String) {
```
vikrantgupta25: check here if the status is getting moved back to active from deleted ?

pkg/modules/user/impluser/module.go
```diff
@@ -95,80 +113,172 @@ func (m *Module) CreateBulkInvite(ctx context.Context, orgID valuer.UUID, userID
...
-	for _, invite := range bulkInvites.Invites {
-		// check if user exists
-		existingUser, err := m.store.GetUserByEmailAndOrgID(ctx, invite.Email, orgID)
-		if err != nil && !errors.Ast(err, errors.TypeNotFound) {
-			return nil, err
-		}
+	if err := m.store.RunInTx(ctx, func(ctx context.Context) error {
+		for _, invite := range bulkInvites.Invites {
+			// check if an active user already exists with this email
+			existingUser, err := m.GetNonDeletedUserByEmailAndOrgID(ctx, invite.Email, orgID)
```
vikrantgupta25: please move the validation checks out of transaction. We cannot hold the transaction for this long. 

Also when you create a user why not create a password there itself in the transaction ?

vikrantgupta25 (changes requested):

pkg/modules/user/impluser/module.go
```diff
@@ -51,38 +50,57 @@ func NewModule(store types.UserStore, tokenizer tokenizer.Tokenizer, emailing em
...
 	if err != nil {
 		return nil, err
 	}
 
-	factorPassword, err := types.NewFactorPassword(password, user.ID.StringValue())
+	return user, nil
+}
+
+// TODO(balanikaran): deprecate this one frontend changes are live with new invitation flow
+func (m *Module) GetInviteByToken(ctx context.Context, token string) (*types.Invite, error) {
```
vikrantgupta25, line 75: please merge the DB calls here to one. `GetUserByResetPasswordTOken`

pkg/modules/user/impluser/module.go
```diff
@@ -443,7 +571,36 @@ func (module *Module) UpdatePasswordByResetPasswordToken(ctx context.Context, to
...
+			[]string{roletypes.MustGetSigNozManagedRoleFromExistingRole(user.Role)},
+			authtypes.MustNewSubject(authtypes.TypeableUser, user.ID.StringValue(), user.OrgID, nil),
+		); err != nil {
+			return err
+		}
+	}
+
+	return module.store.RunInTx(ctx, func(ctx context.Context) error {
+		if user.Status == types.UserStatusPendingInvite {
+			user.UpdateStatus(types.UserStatusActive)
```
vikrantgupta25: move this out of transaction please

pkg/modules/user/impluser/module.go
```diff
@@ -224,7 +338,7 @@ func (m *Module) UpdateUser(ctx context.Context, orgID valuer.UUID, id string, u
 
 	// Make sure that the request is not demoting the last admin user.
 	if user.Role != "" && user.Role != existingUser.Role && existingUser.Role == types.RoleAdmin {
-		adminUsers, err := m.store.GetUsersByRoleAndOrgID(ctx, types.RoleAdmin, orgID)
+		adminUsers, err := m.store.GetActiveUsersByRoleAndOrgID(ctx, types.RoleAdmin, orgID)
```
vikrantgupta25, line 323: authz grants are updated in this function. for user in pending_invite there are no grants currently present so please do not update them

pkg/modules/user/impluser/module.go
```diff
@@ -435,6 +558,11 @@ func (module *Module) UpdatePasswordByResetPasswordToken(ctx context.Context, to
 		return err
 	}
 
+	// handle deleted user
+	if err := user.ErrIfDeleted(); err != nil {
```
vikrantgupta25, line 544: `UpdatePassword` shouldn't be called by `pending_invite` user as well

pkg/modules/user/impluser/store.go
```diff
@@ -443,20 +452,6 @@ func (store *store) UpdatePassword(ctx context.Context, factorPassword *types.Fa
 		return store.sqlstore.WrapNotFoundErrf(err, types.ErrPasswordNotFound, "password for user %s does not exist", factorPassword.UserID)
 	}
 
-	_, err = tx.
```
vikrantgupta25, line 446: why are we not cleaning this up anymore ?

vikrantgupta25: i see this has been moved to app layer

pkg/types/user.go
vikrantgupta25: this will give you more confidence if anywhere it's missed on the app layer.

pkg/modules/user/impluser/module.go
```diff
@@ -469,22 +630,42 @@ func (module *Module) UpdatePassword(ctx context.Context, userID valuer.UUID, ol
...
 		if !errors.Ast(err, errors.TypeNotFound) {
 			return nil, err
 		}
 	}
 
 	if existingUser != nil {
+		// for users logging through SSO flow but are having status as pending_invite
+		if existingUser.Status == types.UserStatusPendingInvite {
+			// respect the role coming from the SSO
+			existingUser.Role = user.Role
```
vikrantgupta25:
```
func (user *user) UpdateRole(role string) {
   user.Role = role 
}
```

tests/integration/src/role/02_user.py
```diff
@@ -34,19 +34,15 @@ def test_user_invite_accept_role_grant(
...
 
-    # accept the invite for editor
-    accept_payload = {
-        "token": invite_token,
-        "password": "password123Z$",
-    }
-    accept_response = requests.post(
-        signoz.self.host_configs["8080"].get("/api/v1/invite/accept"),
-        json=accept_payload,
+    response = requests.post(
```
vikrantgupta25, line 40: the existing test should also work as is. add a new one with new APIs but the exisiting should also pass

karanbalani: We have added a new test case to check the old invite flow.

pkg/modules/user/impluser/module.go
karanbalani: Throwing error in pending state too now

pkg/modules/user/impluser/module.go
```diff
@@ -95,80 +102,159 @@ func (m *Module) CreateBulkInvite(ctx context.Context, orgID valuer.UUID, userID
...
+		if users[0].Status == types.UserStatusPendingInvite {
+			return nil, errors.Newf(errors.TypeAlreadyExists, errors.CodeAlreadyExists, "An invite already exists for this email: %s", users[0].Email.StringValue())
 		}
 
-		newInvite, err := types.NewInvite(invite.Name, role, orgID, invite.Email)
-		if err != nil {
-			return nil, err
-		}
+		return nil, errors.Newf(errors.TypeAlreadyExists, errors.CodeAlreadyExists, "User already exists with this email: %s", users[0].Email.StringValue())
+	}
```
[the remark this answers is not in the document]
karanbalani: Addressed using deduplicating email ids, and also a partial index in db in (org_id, email) in users table

pkg/sqlmigration/067_add_status_user.go
```diff
@@ -0,0 +1,96 @@
...
+	sqls = append(sqls, indexSqls...)
+
+	for _, sql := range sqls {
+		if _, err := tx.ExecContext(ctx, string(sql)); err != nil {
+			return err
+		}
+	}
+
+	// add a partial unique index to enforce uniqueness of (email, org_id) for non-deleted users
+	// replaces the full unique index dropped above, while still allowing multiple deleted
```
vikrantgupta25: let's not add partial indexes without sqlschema support.

pkg/modules/user/impluser/module.go
```diff
@@ -95,80 +102,165 @@ func (m *Module) CreateBulkInvite(ctx context.Context, orgID valuer.UUID, userID
...
-			return nil, err
-		}
-
-		if existingUser != nil {
-			if err := existingUser.ErrIfRoot(); err != nil {
-				return nil, errors.WithAdditionalf(err, "cannot send invite to root user")
-			}
+		email := invite.Email.StringValue()
+		if _, exists := seen[email]; exists {
+			return nil, errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "Duplicate email in request: %s", email)
```
vikrantgupta25: move this to `UnmarshalJSON` for PostableBulkInviteRequest

pkg/modules/user/impluser/module.go
```diff
@@ -95,80 +102,165 @@ func (m *Module) CreateBulkInvite(ctx context.Context, orgID valuer.UUID, userID
...
+		return nil, err
+	}
+
+	pendingUsers := slices.DeleteFunc(users, func(user *types.User) bool { return user.Status != types.UserStatusPendingInvite })
+
+	var invites []*types.Invite
+
+	for _, pUser := range pendingUsers {
+		// get the reset password token
+		resetPasswordToken, err := m.GetOrCreateResetPasswordToken(ctx, pUser.ID)
```
vikrantgupta25, line 235: we are generating a new reset password token even it expired, how does the flow look like now? 

A user gets invited ( a token is sent for max_lifetime ) and then it expires. Now someone needs to go to the page and generate a new token for them ? And now we already created one ?

pkg/modules/user/impluser/module.go
```diff
@@ -52,39 +51,47 @@ func NewModule(store types.UserStore, tokenizer tokenizer.Tokenizer, emailing em
...
-	err = m.CreateUser(ctx, user, root.WithFactorPassword(factorPassword))
-	if err != nil {
-		return nil, err
-	}
-
-	if err := m.DeleteInvite(ctx, invite.OrgID.String(), invite.ID); err != nil {
-		return nil, err
-	}
+	// mark the user as active
+	user.Status = types.UserStatusActive
```
vikrantgupta25: this isn't the right approach, updating the existing inmemory struct post the update. Fetch from DB the user based on the ID you ll get above and then return the same

pkg/modules/user/impluser/module.go
```diff
@@ -52,39 +51,47 @@ func NewModule(store types.UserStore, tokenizer tokenizer.Tokenizer, emailing em
...
 
 	return user, nil
 }
 
 func (m *Module) GetInviteByToken(ctx context.Context, token string) (*types.Invite, error) {
-	invite, err := m.store.GetInviteByToken(ctx, token)
+	// get the user
+	user, err := m.store.GetUserByResetPasswordToken(ctx, token)
```
vikrantgupta25, line 77: what if the token is expired here ?

pkg/sqlmigration/067_add_status_user.go
vikrantgupta25: add another column `deleted_at` with zero value for `pending_invite` and `active` users which will be your unique constraint until we add partial index support

pkg/modules/user/impluser/module.go
karanbalani: Then the user always has an option to go to the "forgot password" flow if enabled, else the user will have to connect with the admin.

pkg/modules/user/impluser/module.go
karanbalani: No we only create a new one if it is expired, else use the same.

pkg/sqlmigration/067_add_status_user.go
karanbalani: Alright, I'll make the changes.

pkg/modules/user/impluser/module.go
karanbalani: Just wanted to save on DB query, I'll add the query here.

tests/integration/src/passwordauthn/01_register.py
```diff
@@ -104,68 +104,66 @@ def test_register(signoz: types.SigNoz, get_token: Callable[[str, str], str]) ->
...
-        None,
-    )
-
-    # Register the editor user using the invite token
+    invited_user = response.json()["data"]
+    assert invited_user["email"] == "editor@integration.test"
+    assert invited_user["role"] == "EDITOR"
+
+    # Verify the user user appears in the users list but as pending_invite status
+    # response = requests.get(
```
vikrantgupta25: let's remove the commented code and add back post frontend integration

frontend/src/api/generated/services/sigNoz.schemas.ts
```diff
@@ -2525,6 +2525,14 @@ export interface TypesPostableAcceptInviteDTO {
 	token?: string;
 }
 
+export interface TypesPostableBulkInviteRequestDTO {
+	/**
+	 * @type array
+	 * @nullable true
+	 */
+	invites: TypesPostableInviteDTO[] | null;
```
AshwinBhatkal: can you remove this null from here?

AshwinBhatkal: basically it should not be nullable

frontend/src/api/generated/services/sigNoz.schemas.ts
```diff
@@ -2665,6 +2673,10 @@ export interface TypesUserDTO {
 	 * @type string
 	 */
 	role?: string;
+	/**
+	 * @type string
+	 */
+	status?: string;
```
AshwinBhatkal, line 2678: Good with making this required before integrating it on the frontend.

frontend/src/api/generated/services/sigNoz.schemas.ts
karanbalani: Done 👍

frontend/src/api/generated/services/sigNoz.schemas.ts
karanbalani: Will take this up in a separate PR @AshwinBhatkal

PR #10446, opened 2026-02-27
chore: enrich clickhouse log_comment

nityanandagohain:
Fixes:- https://github.com/SigNoz/engineering-pod/issues/4071

We are adding more context to the log_comment.

New keys we have added

**ALL** clickhouse queries will have

`trace_id, span_id, module_name, function_name`

Query range V5 will additionaly contains
`panel_type, query_type, duration`

`sinal` is added everywhere it is possible to identify the specific signal.

Before 
```
"auth_type":"tokenizer","dashboard_id":"0xxxxx","http_path":"/dashboard/xxxx/new","module_name":"dashboard","org_id":"xxxxxx","tokenizer_provider":"jwt","user_id":"xxxx","widget_id":"xxxx"}
```

After
actual queries
```
{"auth_type":"tokenizer","dashboard_id":"xxxx","duration":"<1h","http_path":"/dashboard/0xxxxx","module_name":"dashboard","org_id":"xxxx","panel_type":"time_series","query_type":"builder_query","signal":"logs","span_id":"dd3903b3b244597a","tokenizer_provider":"jwt","trace_id":"ea4db6637b18972ff09d7a3912c25091","user_id":"xxxx"
```
metadata
```
{"auth_type":"tokenizer","dashboard_id":"019c55c0-2c89-73b9-bc18-2a49b6d4076b","duration":"\u003c1h","function_name":"executeWithContext","http_path":"/dashboard/019c55c0-2c89-73b9-bc18-2a49b6d4076b","module_name":"builder-query","org_id":"xxxxxx","panel_type":"time_series","query_type":"builder_query","signal":"traces","span_id":"4ba8e215e2980100","tokenizer_provider":"jwt","trace_id":"0d42ee53a9f580037b94d370aae03d50","user_id":"xxxxxx"
```

If  there is a better place to keep `DurationBucket` function or any suggestion, happy to change it.

---

> [!NOTE]
> **Medium Risk**
> Broad, cross-cutting changes add/merge `log_comment` metadata across many query paths; while mostly additive, incorrect context propagation or key collisions could affect ClickHouse query execution/observability in critical request flows.
> 
> **Overview**
> **Enriches ClickHouse `log_comment` across the app for better query observability.** Most query entry points now attach standard comment keys like `code.namespace`, `code.function`, and (where known) `telemetry.signal`.
> 
> Adds a shared `ctxtypes.NewContextWithCommentVals` helper plus new `instrumentationtypes` keys and a `DurationBucket` utility, and uses them to annotate QueryRange execution with `panel.type`, `query.type`, and time-range `query.duration` buckets.
> 
> Updates the telemetry store instrumentation hook to inject `trace_id`/`span_id` into the `log_comment` for *all* ClickHouse queries, and reorders telemetry store hook factories so instrumentation runs before settings are applied.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 93bf591759d31a796b3e8efed8202abb76796526. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/deploy/docker-swarm/docker-compose.ha.yaml
+++ b/deploy/docker-swarm/docker-compose.ha.yaml
@@ -190,7 +190,7 @@ services:
       # - ../common/clickhouse/storage.xml:/etc/clickhouse-server/config.d/storage.xml
   signoz:
     !!merge <<: *db-depend
-    image: signoz/signoz:v0.114.0
+    image: signoz/signoz:v0.114.1
     ports:
       - "8080:8080" # signoz port
     #   - "6060:6060"     # pprof port
--- a/deploy/docker-swarm/docker-compose.yaml
+++ b/deploy/docker-swarm/docker-compose.yaml
@@ -117,7 +117,7 @@ services:
       # - ../common/clickhouse/storage.xml:/etc/clickhouse-server/config.d/storage.xml
   signoz:
     !!merge <<: *db-depend
-    image: signoz/signoz:v0.114.0
+    image: signoz/signoz:v0.114.1
     ports:
       - "8080:8080" # signoz port
     volumes:
--- a/deploy/docker/docker-compose.ha.yaml
+++ b/deploy/docker/docker-compose.ha.yaml
@@ -181,7 +181,7 @@ services:
       # - ../common/clickhouse/storage.xml:/etc/clickhouse-server/config.d/storage.xml
   signoz:
     !!merge <<: *db-depend
-    image: signoz/signoz:${VERSION:-v0.114.0}
+    image: signoz/signoz:${VERSION:-v0.114.1}
     container_name: signoz
     ports:
       - "8080:8080" # signoz port
--- a/deploy/docker/docker-compose.yaml
+++ b/deploy/docker/docker-compose.yaml
@@ -109,7 +109,7 @@ services:
       # - ../common/clickhouse/storage.xml:/etc/clickhouse-server/config.d/storage.xml
   signoz:
     !!merge <<: *db-depend
-    image: signoz/signoz:${VERSION:-v0.114.0}
+    image: signoz/signoz:${VERSION:-v0.114.1}
     container_name: signoz
     ports:
       - "8080:8080" # signoz port
--- a/frontend/src/container/GeneralSettings/GeneralSettings.tsx
+++ b/frontend/src/container/GeneralSettings/GeneralSettings.tsx
@@ -674,7 +674,7 @@ function GeneralSettings({
 	return (
 		<div className="general-settings-page">
 			<div className="general-settings-header">
-				<span className="general-settings-title">General</span>
+				<span className="general-settings-title">Workspace</span>
 				<span className="general-settings-subtitle">
 					Manage your workspace settings.
 				</span>
--- a/frontend/src/container/MySettings/index.tsx
+++ b/frontend/src/container/MySettings/index.tsx
@@ -161,7 +161,7 @@ function MySettings(): JSX.Element {
 		<div className="my-settings-container">
 			<div className="user-info-section">
 				<div className="user-info-section-header">
-					<div className="user-info-section-title">General </div>
+					<div className="user-info-section-title">Account </div>
 
 					<div className="user-info-section-subtitle">
 						Manage your account settings.
--- a/frontend/src/container/SideNav/SideNav.styles.scss
+++ b/frontend/src/container/SideNav/SideNav.styles.scss
@@ -1057,21 +1057,20 @@
 		gap: 8px;
 
 		.user-settings-dropdown-label-text {
-			color: var(--bg-slate-50, #62687c);
+			color: var(--l3-foreground);
 			font-family: Inter;
-			font-size: 10px;
-			font-family: Inter;
-			font-weight: 600;
+			font-size: var(--uppercase-small-500-font-size);
+			font-weight: var(--uppercase-small-500-font-weight);
 			font-style: normal;
-			line-height: 18px; /* 163.636% */
+			line-height: 18px;
 			letter-spacing: 0.88px;
 			text-transform: uppercase;
 		}
 
 		.user-settings-dropdown-label-email {
-			color: var(--bg-vanilla-400, #c0c1c3);
+			color: var(--l1-foreground);
 			font-family: Inter;
-			font-size: 12px;
+			font-size: var(--font-size-xs);
 			font-style: normal;
 			line-height: normal;
 			letter-spacing: 0.14px;
@@ -1079,7 +1078,7 @@
 	}
 
 	.ant-dropdown-menu-item-divider {
-		background-color: var(--bg-slate-500, #161922) !important;
+		background-color: var(--secondary) !important;
 	}
 
 	.ant-dropdown-menu-item-disabled {
@@ -1095,6 +1094,14 @@
 .help-support-dropdown {
 	.ant-dropdown-menu-item {
 		min-height: 32px;
+
+		.ant-dropdown-menu-title-content {
+			color: var(--l1-foreground) !important;
+		}
+
+		.user-settings-dropdown-logout-section {
+			color: var(--danger-background);
+		}
 	}
 }
 
@@ -1271,7 +1278,7 @@
 }
 
 .help-support-dropdown li.ant-dropdown-menu-item-divider {
-	background-color: var(--bg-slate-500, #161922) !important;
+	background-color: var(--secondary) !important;
 }
 
 .lightMode {
@@ -1431,22 +1438,6 @@
 		}
 	}
 
-	.settings-dropdown {
-		.user-settings-dropdown-logged-in-section {
-			.user-settings-dropdown-label-text {
-				color: var(--bg-ink-400);
-			}
-
-			.user-settings-dropdown-label-email {
-				color: var(--bg-ink-300);
-			}
-		}
-
-		.ant-dropdown-menu-item-divider {
-			background-color: var(--bg-vanilla-300) !important;
-		}
-	}
-
 	.reorder-shortcut-nav-items-modal {
 		.ant-modal-content {
 			border: 1px solid var(--bg-vanilla-300);
@@ -1502,10 +1493,6 @@
 		.title {
 			color: var(--bg-ink-400);
 		}
-	}
-
-	.help-support-dropdown li.ant-dropdown-menu-item-divider {
-		background-color: var(--bg-vanilla-300) !important;
 	}
 }
 
--- a/frontend/src/container/SideNav/SideNav.tsx
+++ b/frontend/src/container/SideNav/SideNav.tsx
@@ -69,6 +69,7 @@ import { routeConfig } from './config';
 import { getQueryString } from './helper';
 import {
 	defaultMoreMenuItems,
+	getUserSettingsDropdownMenuItems,
 	helpSupportDropdownMenuItems as DefaultHelpSupportDropdownMenuItems,
 	helpSupportMenuItem,
 	primaryMenuItems,
@@ -485,48 +486,12 @@ function SideNav({ isPinned }: { isPinned: boolean }): JSX.Element {
 
 	const userSettingsDropdownMenuItems: MenuProps['items'] = useMemo(
 		() =>
-			[
-				{
-					key: 'label',
-					label: (
-						<div className="user-settings-dropdown-logged-in-section">
-							<span className="user-settings-dropdown-label-text">LOGGED IN AS</span>
-							<span className="user-settings-dropdown-label-email">{user.email}</span>
-						</div>
-					),
-					disabled: true,
-					dataTestId: 'logged-in-as-nav-item',
-				},
-				{ type: 'divider' as const },
-				{
-					key: 'account',
-					label: 'Account Settings',
-					dataTestId: 'account-settings-nav-item',
-				},
-				{
-					key: 'workspace',
-					label: 'Workspace Settings',
-					disabled: isWorkspaceBlocked,
-					dataTestId: 'workspace-settings-nav-item',
-				},
-				...(isEnterpriseSelfHostedUser || isCommunityEnterpriseUser
-					? [
-							{
-								key: 'license',
-								label: 'Manage License',
-								dataTestId: 'manage-license-nav-item',
-							},
-					  ]
-					: []),
-				{ type: 'divider' as const },
-				{
-					key: 'logout',
-					label: (
-						<span className="user-settings-dropdown-logout-section">Sign out</span>
-					),
-					dataTestId: 'logout-nav-item',
-				},
-			].filter(Boolean),
+			getUserSettingsDropdownMenuItems({
+				userEmail: user.email,
+				isWorkspaceBlocked,
+				isEnterpriseSelfHostedUser,
+				isCommunityEnterpriseUser,
+			}),
 		[
 			isEnterpriseSelfHostedUser,
 			isCommunityEnterpriseUser,
@@ -856,9 +821,6 @@ function SideNav({ isPinned }: { isPinned: boolean }): JSX.Element {
 			});
 
 			switch (item.key) {
-				case ROUTES.SHORTCUTS:
-					history.push(ROUTES.SHORTCUTS);
-					break;
 				case 'invite-collaborators':
 					history.push(`${ROUTES.ORG_SETTINGS}#invite-team-members`);
 					break;
@@ -878,7 +840,7 @@ function SideNav({ isPinned }: { isPinned: boolean }): JSX.Element {
 	};
 
 	const handleSettingsMenuItemClick = (info: SidebarItem): void => {
-		const item = userSettingsDropdownMenuItems.find(
+		const item = (userSettingsDropdownMenuItems ?? []).find(
 			(item) => item?.key === info.key,
 		);
 		let menuLabel = '';
@@ -903,6 +865,9 @@ function SideNav({ isPinned }: { isPinned: boolean }): JSX.Element {
 				break;
 			case 'license':
 				history.push(ROUTES.LIST_LICENSES);
+				break;
+			case 'keyboard-shortcuts':
+				history.push(ROUTES.SHORTCUTS);
 				break;
 			case 'logout':
 				Logout();
--- a/frontend/src/container/SideNav/__tests__/menuItems.test.tsx
+++ b/frontend/src/container/SideNav/__tests__/menuItems.test.tsx
@@ -1,0 +1,74 @@
+import { getUserSettingsDropdownMenuItems } from 'container/SideNav/menuItems';
+
+const BASE_PARAMS = {
+	userEmail: 'test@signoz.io',
+	isWorkspaceBlocked: false,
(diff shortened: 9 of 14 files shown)


pkg/modules/metricsexplorer/implmetricsexplorer/module.go
```diff
@@ -217,6 +217,12 @@ func (m *module) GetTreemap(ctx context.Context, orgID valuer.UUID, req *metrics
 }
 
 func (m *module) GetMetricMetadataMulti(ctx context.Context, orgID valuer.UUID, metricNames []string) (map[string]*metricsexplorertypes.MetricMetadata, error) {
+
+	comment := ctxtypes.CommentFromContext(ctx)
+	comment.Set("signal", telemetrytypes.SignalMetrics.StringValue())
+	comment.Set("is_metadata_query", "true")
```
tushar-signoz: Can we resuse `"query_type"`: `metadata` ?

pkg/telemetrystore/telemetrystorehook/instrumentation.go
```diff
@@ -35,7 +36,14 @@ func NewInstrumentation(ctx context.Context, providerSettings factory.ProviderSe
...
 
 func (hook *instrumentation) BeforeQuery(ctx context.Context, event *telemetrystore.QueryEvent) context.Context {
-	ctx, _ = hook.tracer.Start(ctx, "", trace.WithSpanKind(trace.SpanKindClient))
+	ctx, span := hook.tracer.Start(ctx, "", trace.WithSpanKind(trace.SpanKindClient))
+
+	// add trace_id and span_id to the log_comment
+	comment := ctxtypes.CommentFromContext(ctx)
+	comment.Set("trace_id", span.SpanContext().TraceID().String())
+	comment.Set("span_id", span.SpanContext().SpanID().String())
+	ctx = ctxtypes.NewContextWithComment(ctx, comment)
```
tushar-signoz, line 45: Can we add an helper method something like ctx.AddComment("key", "value") ?

srikanthccv:
@nityanandagohain can you do small exercise to list down all the places a query could come from so that we can have them instrumented as well. The query range covers a large surface area but leaves behind numerous other parts that should be covered as well.

nityanandagohain:
sure, let me get back.

pkg/modules/metricsexplorer/implmetricsexplorer/module.go
nityanandagohain: I have moved to `module-name` instead of this and let the query type be reserved to `builer/clickhouse/promql`

pkg/telemetrystore/telemetrystorehook/instrumentation.go
nityanandagohain: Yep added a helper function, thanks for the suggestion.

nityanandagohain:
@srikanthccv , I have added the comments to all the function which calls clickhouse using db.Exec / db.Query...

Let me know if anything else will help you review this.

pkg/querier/builder_query.go
```diff
@@ -212,6 +213,13 @@ func (q *builderQuery[T]) Execute(ctx context.Context) (*qbtypes.Result, error)
 
 // executeWithContext executes the query with query window and step context for partial value detection
 func (q *builderQuery[T]) executeWithContext(ctx context.Context, query string, args []any) (*qbtypes.Result, error) {
+	ctx = ctxtypes.AddCommentsToContext(ctx, map[string]string{
+		"signal":        q.spec.Signal.StringValue(),
+		"module_name":   "builder-query",
```
srikanthccv: This is going to be misleading? Because the querier gets called from so many actual modules that if we set "module_name" the original module name will be replaced by `builder-query`?

pkg/querier/clickhouse_query.go
```diff
@@ -98,6 +99,11 @@ func (q *chSQLQuery) renderVars(query string, vars map[string]qbtypes.VariableIt
 }
 
 func (q *chSQLQuery) Execute(ctx context.Context) (*qbtypes.Result, error) {
+	ctx = ctxtypes.AddCommentsToContext(ctx, map[string]string{
+		"module_name":   "clickhouse-query",
```
srikanthccv: Same point as above here

pkg/querier/trace_operator_query.go
```diff
@@ -52,6 +54,13 @@ func (q *traceOperatorQuery) Execute(ctx context.Context) (*qbtypes.Result, erro
 }
 
 func (q *traceOperatorQuery) executeWithContext(ctx context.Context, query string, args []any) (*qbtypes.Result, error) {
+	ctx = ctxtypes.AddCommentsToContext(ctx, map[string]string{
+		"signal":        telemetrytypes.SignalTraces.StringValue(),
+		"module_name":   "querier",
```
srikanthccv: Same

pkg/query-service/app/clickhouseReader/reader.go
srikanthccv, line 1: This is wonderful! thanks.

One major point, we need to get the module name right, otherwise we can't quickly decide which page or module the slow queries are. Example every module name here is `clickhouse-reader`, which doesn't give anything any forces us to further drilling even for basic high level module info.

pkg/types/querybuildertypes/querybuildertypesv5/builder_query.go
```diff
@@ -226,3 +227,28 @@ func CanShortCircuitDelta(metricAgg MetricAggregation) bool {
 
 	return false
 }
+
+func DurationBucket(fromMS, toMS uint64) string {
```
srikanthccv: Should we moe this to instrumentation package?

pkg/querier/builder_query.go
nityanandagohain: My reason for this was that, `module, function_name`  will identify where is the actual query getting executed from? Now there will be other attributes in this for all the query range  queries i.e `signal, panel_type, query_type, duration` with `trace and span` id, which will have trace down the source. Also for query_range v5 this is the only place where module and function is added for v5 API's.

Let me know what you think, else we will have to add the function_name and module name in all the functions which doesn't call clickhouse but they use query range in some way or another.

pkg/types/querybuildertypes/querybuildertypesv5/builder_query.go
nityanandagohain: Yess, thanks for this I was a bit confused where to put this.

pkg/query-service/app/clickhouseReader/reader.go
nityanandagohain: But function name will help you to identify the exact one ?

pkg/querier/builder_query.go
srikanthccv: I think it would be nice to have a which user facing module the query originated from otherwise we are blind to which module it comes so can't get an stats by log comment themselves to make sense of the data.

nityanandagohain: Let's discuss this once over call, will be easier to freeze the decision.

nityanandagohain: Discussion:- For query range, we want to know module and function who is actually call querying range. So this should be present in the latest function which calls it instead of the actuall db call.

nityanandagohain:
I have added the changes as discussed, do have a look and let me know if you have any concerns. Also I have skipped the querier v2 internals as don't think it is used much, but the handler functions have the function name in that wherever it is used.

pkg/types/ctxtypes/comment.go
```diff
@@ -104,6 +104,21 @@ func CommentFromHTTPRequest(req *http.Request) map[string]string {
 	return comments
 }
 
+// AddCommentsToContext returns a new context with all key-value pairs from comments merged into the Comment in the context.
+func AddCommentsToContext(ctx context.Context, comments map[string]string) context.Context {
```
srikanthccv: Do we really need this? There is `Merge` on `Comment` already

pkg/query-service/rules/threshold_rule.go
```diff
@@ -433,7 +435,10 @@ func (r *ThresholdRule) buildAndRunQuery(ctx context.Context, orgID valuer.UUID,
 
 	var results []*v3.Result
 	var queryErrors map[string]error
-
+	ctx = ctxtypes.AddCommentsToContext(ctx, map[string]string{
+		instrumentation.CodeNamespace:    "rules",
+		instrumentation.CodeFunctionName: "buildAndRunQuery",
+	})
```
srikanthccv: Please move this to `buildAndRunQueryV5` because `buildAndRunQuery` is no longer active.

pkg/querier/clickhouse_query.go
```diff
@@ -98,6 +100,10 @@ func (q *chSQLQuery) renderVars(query string, vars map[string]qbtypes.VariableIt
 }
 
 func (q *chSQLQuery) Execute(ctx context.Context) (*qbtypes.Result, error) {
+	ctx = ctxtypes.AddCommentsToContext(ctx, map[string]string{
+		"module_name":                 "clickhouse-query",
```
srikanthccv: Let's not add any module name here, let the upstream caller in the call stack set it.

srikanthccv: I see the other one `code.namespace`. Do you want to keep `module_name` as a separate key?

pkg/querier/api.go
```diff
@@ -33,6 +34,10 @@ func NewHandler(set factory.ProviderSettings, querier Querier, analytics analyti
 
 func (handler *handler) QueryRange(rw http.ResponseWriter, req *http.Request) {
 	ctx := req.Context()
+	ctx = ctxtypes.AddCommentsToContext(ctx, map[string]string{
+		instrumentation.CodeNamespace:    "querier",
```
srikanthccv: So there is `CommentFromHTTPRequest` which is setup as middleware to handle the request level things that add number of things. Do we want to keep both `module_name` and `code.namespace`?

srikanthccv (commented):
Overall LGTM, couple of questions.

pkg/querier/clickhouse_query.go
nityanandagohain: I removed `module_name` from all my changes, this was a miss.

pkg/querier/api.go
nityanandagohain: Let's starting with keeping both and then deprate things in the `CommentFromHTTPRequest` as we start noticing the non helpful onces. Also if there is something not correct we will notice it.

Also that is the reason I moved away from `module_name` in the current changes to not touch the existing ones.

pkg/types/ctxtypes/comment.go
nityanandagohain: I see what you are suggesting, Updatded the function now and have removed my own logic of merging and used the merge function inside this, but this helper function is helpful since I have to always get from the context add and update. The get and update part is removed with this helper function.

pkg/types/instrumentationtypes/duration.go
```diff
@@ -0,0 +1,54 @@
...
+	if epoch == 0 {
+		count = 1
+	} else {
+		for epoch != 0 {
+			epoch /= 10
+			count++
+		}
+	}
+	return temp * uint64(math.Pow(10, float64(19-count)))
+}
```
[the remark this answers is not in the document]
nityanandagohain: Don't want to import querybuilder in instrumentatoin package.

srikanthccv (dismissed):
thanks

pkg/types/ctxtypes/comment.go
```diff
@@ -104,6 +104,16 @@ func CommentFromHTTPRequest(req *http.Request) map[string]string {
...
```
vikrantgupta25:
```suggestion
func NewContextWithCommentVals(ctx context.Context, vals map[string]string) context.Context {
```

pkg/types/instrumentationtypes/duration.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"math"
+	"time"
+)
+
+// DurationBucket returns a human-readable bucket label for the duration between fromMS and toMS.
+// fromMS and toMS are Unix timestamps (same unit as used by time.Unix).
+// Returns labels like "<1h", "<6h", "<24h", "<3D", "<1W", "<2W", "<1M", or ">=1M".
+func DurationBucket(from, to uint64) string {
```
vikrantgupta25, line 11: lets add these to instrumentationtypes not in the root pkg please

srikanthccv: This is arbitrary. Please help me understand what makes you think this should be in the types package. I would imagine any supporting function that helps with instrumentation goes into instrumentation package.

srikanthccv: Essentially

`instrumentation.DurationBucket(...)` vs `instrumentationtypes.DurationBucket(...)`

I see the former to be the idiomatic one.

vikrantgupta25: because now we have defined the explicitly defined our semconv here and blocked these attributes. For someone else consuming this outside signoz repository doesn't need to import the entire instrumentation package when they would already be importing the types

vikrantgupta25: by the same logic alertmanager and authz should have their own functions rather than types ? 😅

srikanthccv: I am talking about the `DurationBucket`. It's a supporting function and has nothing to do with the types.

srikanthccv: >by the same logic alertmanager and authz should have their own functions rather than types ? 😅

Yes, indeed. What I am looking for is an objective criteria what goes into types.

srikanthccv: This is important because @nityanandagohain called out where to put. Based on my understanding, I suggested `instrumentation`. Now he is being asked to put into `instrumentationtypes`. This are not the somethin we should be reapeating at the PR multiple times.

vikrantgupta25: duration.go can be here, the comment was for keys.go majorly. That shouldn't be here.

vikrantgupta25: `What I am looking for is an objective criteria what goes into types.`

- The broader idea behind was that consumers should worry just about the interface of the package and if they need to be aware of anything else then it's the types responsibility. 

- Also the things that should/can be imported outside the context of the repository for interacting with the system is the types candidate. 


the first one being more of a broader idea and the second one definitely being a criteria.

vikrantgupta25: the feel that i get from this is we are unsure of whether to put certain convertors/ functions in the pkg itself or in the types

srikanthccv: Correct, One unwritten rule I follow (based on my interpretation) is that if the function has nothing to do with the types, then it goes in the package itself. But if does something on type, or returns the types or just something around the type, then it should be in the types package.

nityanandagohain: I am moving the keys to types package, what should we do about the durationbucket, should I keep it here or move to types.

nityanandagohain: Since I moved the keys to types, moved the function as well.

pkg/modules/metricsexplorer/implmetricsexplorer/module.go
```diff
@@ -1216,3 +1244,12 @@ func (m *module) fetchMetricAttributes(ctx context.Context, metricName string, s
 
 	return attributes, nil
 }
+
+func withMetricsExplorerQuery(ctx context.Context, functionName string) context.Context {
```
vikrantgupta25:
```suggestion
func (m *module) withMetricsExplorerQuery(ctx context.Context, functionName string) context.Context {
```

vikrantgupta25: let's have these defined as module functions

pkg/modules/services/implservices/module.go
```diff
@@ -124,6 +140,10 @@ func (m *module) GetTopOperations(ctx context.Context, orgUUID valuer.UUID, req
 
 // GetEntryPointOperations implements services.Module for QBV5 based entry point ops
 func (m *module) GetEntryPointOperations(ctx context.Context, orgUUID valuer.UUID, req *servicetypesv1.OperationsRequest) ([]servicetypesv1.OperationItem, error) {
+	ctx = ctxtypes.NewContextWithCommentVals(ctx, map[string]string{
```
vikrantgupta25: let's do similar to metricsexplorer ?

pkg/types/ctxtypes/comment.go
```diff
@@ -104,6 +104,13 @@ func CommentFromHTTPRequest(req *http.Request) map[string]string {
 	return comments
 }
 
+// AddCommentsToContext returns a new context with all key-value pairs from comments merged into the Comment in the context.
```
vikrantgupta25: need to update the code comment as well

pkg/modules/services/implservices/module.go
nityanandagohain: sure done , added in prometheus client as well.

pkg/querier/querier.go
```diff
@@ -526,6 +528,11 @@ func (q *querier) run(
 	steps map[string]qbtypes.Step,
 	qbEvent *qbtypes.QBEvent,
 ) (*qbtypes.QueryRangeResponse, error) {
+	ctx = ctxtypes.NewContextWithCommentVals(ctx, map[string]string{
+		instrumentationtypes.PanelType: qbEvent.PanelType,
+		instrumentationtypes.QueryType: qbEvent.QueryType,
+	})
```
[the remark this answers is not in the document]
nityanandagohain: These two cannot be empty, if empty then the query itself cannot be built.

pkg/types/instrumentationtypes/duration.go
```diff
@@ -0,0 +1,53 @@
...
+	if epoch == 0 {
+		count = 1
+	} else {
+		for epoch != 0 {
+			epoch /= 10
+			count++
+		}
+	}
+	return temp * uint64(math.Pow(10, float64(19-count)))
+}
```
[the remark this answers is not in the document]
nityanandagohain: Won't happen.

pkg/query-service/app/inframetrics/deployments.go
```diff
@@ -230,6 +232,10 @@ func (d *DeploymentsRepo) getTopDeploymentGroups(ctx context.Context, orgID valu
 		topDeploymentGroupsQueryRangeParams.CompositeQuery.BuilderQueries[queryName] = query
 	}
 
+	ctx = ctxtypes.NewContextWithCommentVals(ctx, map[string]string{
+		instrumentationtypes.CodeNamespace:    "inframetrics",
+		instrumentationtypes.CodeFunctionName: "getTopDeploymentGroups",
+	})
```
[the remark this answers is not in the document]
nityanandagohain: Will address in the next pr when we add updates

PR #10455, opened 2026-02-28
fix: fixed memory leak because of canvas persistance over mount

ahrefabhi:
### Summary

The TooltipPlugin was causing a memory leak when dashboards or charts unmounted. Strong references to the uPlot instance were retained in two ways: (1) React state held the plot via `viewState.plot`, and (2) event handlers like `handleUPlotOverClick(u, event)` closed over the plot reference. When charts unmounted, these references kept the uPlot instance and its canvases alive, leading to memory growth after repeated mount/unmount (e.g. switching dashboards or navigating away).

This PR fixes it by:

- Replacing `viewState.plot` with `viewState.hasPlot: boolean` so React state never holds the plot
- Changing `handleUPlotOverClick` to use `getPlot(controller)` instead of closing over the plot
- Adding `clearPlotReferences()` in the destroy/cleanup hook to set `controller.plot = null` and `hasPlot: false` when the plugin unmounts

#### Before

https://github.com/user-attachments/assets/4762427b-321f-4a0d-9958-f4ea94bd5aca

#### After 

https://github.com/user-attachments/assets/dfd012f6-6012-4d15-83bc-805646394e2d

### Change Type

- [x] 🐛 Bug fix

### Bug Context

**Root cause:** The plot reference was stored in React state (`viewState.plot`) and event handlers closed over it (`handleUPlotOverClick(u, event)`). Both kept the uPlot instance and its canvases in memory after unmount.

**Fix strategy:**
1. Use `hasPlot: boolean` in view state instead of `plot: uPlot | null`
2. Add `getPlot(controller)` and use it everywhere instead of closing over the plot
3. Update `handleUPlotOverClick` to take only `event` and resolve the plot via `getPlot(controller)`
4. Add `clearPlotReferences()` in the cleanup hook to clear the controller reference when the plugin unmounts

### Testing Strategy

- Existing tests pass
- Manual verification: Navigate between dashboards/views and confirm memory stays stable instead of growing

### Risk & Impact Assessment

- **Blast radius:** Limited to TooltipPlugin
- **Potential regressions:** Low; tooltip behavior is unchanged
- **Rollback plan:** Revert the PR if issues appear

### Changelog

| Field | Value |
|-------|-------|
| Change Type | Bug Fix |
| Description | Fixed memory leak in time series charts caused by the tooltip plugin retaining references to destroyed uPlot instances and canvases |

### Checklist

- [ ] Manually tested

Closes https://github.com/SigNoz/signoz/issues/10456

### DIFF
--- a/frontend/src/lib/uPlotV2/plugins/TooltipPlugin/TooltipPlugin.tsx
+++ b/frontend/src/lib/uPlotV2/plugins/TooltipPlugin/TooltipPlugin.tsx
@@ -74,9 +74,9 @@ export default function TooltipPlugin({
 		layoutRef.current = createLayoutObserver(layoutRef);
 
 		/**
-		 * Plot lifecycle and GC: controller.plot uses WeakRef, viewState uses
-		 * hasPlot (boolean), and the destroy hook clears refs when uPlot.destroy()
-		 * runs—so detached canvases can be garbage collected.
+		 * Plot lifecycle and GC: viewState uses hasPlot (boolean), not the plot
+		 * reference; clearPlotReferences runs in cleanup so
+		 * detached canvases can be garbage collected.
 		 */
 		// Controller holds the mutable interaction state for this tooltip
 		// instance. It is intentionally *not* React state so uPlot hooks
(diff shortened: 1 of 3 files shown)


frontend/src/lib/uPlotV2/plugins/TooltipPlugin/TooltipPlugin.tsx
```diff
@@ -72,12 +73,25 @@ export default function TooltipPlugin({
 		layoutRef.current?.observer.disconnect();
 		layoutRef.current = createLayoutObserver(layoutRef);
 
+		/**
```
AshwinBhatkal, line 76: Is this comment at the right place?

PR #10459, opened 2026-03-01
enh: updated tooltip pinning structure to be used across different charts

ahrefabhi:
### 📄 Summary

This PR adds **pinned tooltips** to uPlot-based charts so users can click on a point/bar/bucket and keep the tooltip visible while they inspect values, compare across series, or interact with other parts of the UI. The feature is implemented in a reusable way in the uPlot v2 tooltip plugin and tooltip component, then wired into the TimeSeries, Bar, and Histogram dashboard charts so all chart types can benefit from consistent pinned-tooltip behaviour.

### ✅ Change Type

- ✨ **Feature**
- ♻️ **Refactor** (tooltip plumbing)
- 🧪 **Tests updated/added**

### 🧩 What’s Changed

- **Pinned tooltip flow in charts**
  - `BarChart`, `Histogram`, and `TimeSeries` charts now enable `canPinTooltip` on the uPlot tooltip plugin.
  - Each chart passes a `pinnedTooltipElement` renderer so pinned tooltips can display richer, chart-specific content when the user clicks.
  - Tooltip render callbacks now receive click metadata, allowing pinned tooltips to show precise timestamp/value/series information.

- **uPlot tooltip plugin enhancements**
  - `TooltipPlugin`:
    - Imports `getFocusedSeriesAtPosition` from `onClickPlugin` to reuse the existing logic for resolving the focused series at the click location.
    - On overlay click (when pinning is enabled and a series is focused), computes:
      - `xValue` / `yValue` using `u.posToVal`.
      - The focused series (index, label, value, colour) using `getFocusedSeriesAtPosition`.
      - An accurate `clickedDataTimestamp` from the x-axis data where possible.
      - Mouse coordinates (offset and absolute) for downstream consumers.
    - Stores this in a new `controller.clickData` field and marks the tooltip as `pinned`, keeping it visible and interactive.
    - Clears `clickData` when dismissing or when data changes.
  - `TooltipPlugin` render path passes `clickData` into the chart’s tooltip render function via `TooltipRenderArgs`.
  - `TooltipPlugin` tests updated to:
    - Verify pinning behaviour when `canPinTooltip` is enabled.
    - Ensure pinned tooltips can be dismissed (via callback, outside click, outside keydown) and that hover/pin state transitions are correct.
    - Use a more complete fake uPlot instance and mock `onClickPlugin` to focus tests on pinning logic rather than full uPlot behaviour.

- **Shared tooltip types & state**
  - New `TooltipClickData` type in `TooltipPlugin/types.ts`:
    - Includes `xValue`, `yValue`, `focusedSeries` info, `clickedDataTimestamp`, and both relative and absolute mouse coordinates.
  - `TooltipControllerState` extended with a `clickData: TooltipClickData | null` field, documenting it as the data associated with the click that triggered the pinned tooltip.

- **Tooltip component integration**
  - `Tooltip` component props extended (`TooltipProps`) to support:
    - `isPinned` flag (from plugin/view state).
    - `pinnedTooltipElement?: (clickData: TooltipClickData) => ReactNode`.
    - `clickData: TooltipClickData | null`.
  - Render logic updated:
    - When `isPinned` and `clickData` are both present, the component renders `pinnedTooltipElement(clickData)` instead of the standard header + scrolling series list.
    - Falls back to the existing header + Virtuoso-based content list when not pinned.
  - Tooltip tests updated to:
    - Import `TooltipClickData`.
    - Assert that when `isPinned` and `clickData` are provided, the pinned content is rendered and the normal header/list are hidden.

- **Chart visualization types**
  - Dashboard chart `types.ts` updated so chart configs/props understand and propagate the new pinned-tooltip capabilities where needed.

### 🧪 Testing Strategy

- **Automated tests**
  - Updated `TooltipPlugin` tests to cover:
    - Pinning when the overlay is clicked with `canPinTooltip` enabled.
    - Dismissing pinned tooltips via the provided `dismiss` callback.
    - Dropping pinned tooltips on data changes.
    - Unpinning on outside mousedown/keydown.
  - Updated `Tooltip` unit tests to verify:
    - Pinned rendering path using `pinnedTooltipElement` and `clickData`.
    - That the standard tooltip UI is suppressed when pinned content is shown.
- **Manual verification**
  - Validated TimeSeries, Bar, and Histogram dashboards to ensure:
    - Clicking a point/bar/bucket pins the tooltip and keeps it visible.
    - Pinned tooltip content reflects the clicked data (timestamp, value, series).
    - Pinned tooltips can be dismissed and do not interfere with normal hover behaviour.

### 📝 Changelog

- **Change Type**: Feature / Maintenance (visualization UX)  
- **Description**: Add pinned tooltip support to uPlot-based dashboard charts, exposing click metadata to tooltip renderers and allowing users to keep tooltips visible while inspecting chart data.

---

> [!NOTE]
> **Medium Risk**
> Changes core tooltip interaction/pinning logic and introduces new click-driven rendering paths, which may affect hover/pin behavior across multiple chart types.
> 
> **Overview**
> Adds a reusable pinned-tooltip pathway for uPlot charts: `TooltipPlugin` now captures click metadata (`TooltipClickData`) when pinning, keeps the tooltip visible while pinned, and can render an optional `pinnedTooltipElement` instead of the hover tooltip.
> 
> Plumbs this through dashboard charts by renaming `renderTooltip` to `customTooltip`, adding `pinnedTooltipElement` to chart props, and passing it through `ChartWrapper` to the plugin; also makes `TimeSeriesPanel`’s `onClick` handler optional and updates tooltip/plugin tests for the new pinned behavior and click-data state.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fd3fdd3d318b1c74d5cf90bef75b43bbc497b7db. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/src/components/CustomTimePicker/CustomTimePicker.test.tsx
+++ b/frontend/src/components/CustomTimePicker/CustomTimePicker.test.tsx
@@ -1,0 +1,279 @@
+import { useState } from 'react';
+import { fireEvent, render, screen } from '@testing-library/react';
+import dayjs from 'dayjs';
+import * as timeUtils from 'utils/timeUtils';
+
+import CustomTimePicker from './CustomTimePicker';
+
+jest.mock('react-router-dom', () => {
+	const actual = jest.requireActual('react-router-dom');
+
+	return {
+		...actual,
+		useLocation: jest.fn().mockReturnValue({
+			pathname: '/test-path',
+		}),
+	};
+});
+
+jest.mock('providers/Timezone', () => {
+	const actual = jest.requireActual('providers/Timezone');
+
+	return {
+		...actual,
+		useTimezone: jest.fn().mockReturnValue({
+			timezone: {
+				value: 'UTC',
+				offset: '+00:00',
+				name: 'UTC',
+			},
+			browserTimezone: {
+				value: 'UTC',
+				offset: '+00:00',
+				name: 'UTC',
+			},
+		}),
+	};
+});
+
+interface WrapperProps {
+	initialValue?: string;
+	showLiveLogs?: boolean;
+	onValidCustomDateChange?: () => void;
+	onError?: () => void;
+	onSelect?: (value: string) => void;
+	onCustomDateHandler?: () => void;
+	onCustomTimeStatusUpdate?: () => void;
+}
+
+function Wrapper({
+	initialValue = '2024-01-01 00:00:00 - 2024-01-01 01:00:00',
+	showLiveLogs = false,
+	onValidCustomDateChange = (): void => {},
+	onError = (): void => {},
+	onSelect = (): void => {},
+	onCustomDateHandler = (): void => {},
+	onCustomTimeStatusUpdate = (): void => {},
+}: WrapperProps): JSX.Element {
+	const [open, setOpen] = useState(false);
+	const [selectedTime, setSelectedTime] = useState('custom');
+	const [selectedValue, setSelectedValue] = useState(initialValue);
+
+	const handleSelect = (value: string): void => {
+		setSelectedTime(value);
+		onSelect(value);
+	};
+
+	return (
+		<CustomTimePicker
+			open={open}
+			setOpen={setOpen}
+			onSelect={handleSelect}
+			onError={onError}
+			selectedTime={selectedTime}
+			selectedValue={selectedValue}
+			onValidCustomDateChange={({ timeStr }): void => {
+				setSelectedValue(timeStr);
+				onValidCustomDateChange();
+			}}
+			onCustomDateHandler={(): void => {
+				onCustomDateHandler();
+			}}
+			onCustomTimeStatusUpdate={(): void => {
+				onCustomTimeStatusUpdate();
+			}}
+			items={[
+				{ label: 'Last 5 minutes', value: '5m' },
+				{ label: 'Custom', value: 'custom' },
+			]}
+			minTime={dayjs('2024-01-01 00:00:00').valueOf() * 1000_000}
+			maxTime={dayjs('2024-01-01 01:00:00').valueOf() * 1000_000}
+			showLiveLogs={showLiveLogs}
+		/>
+	);
+}
+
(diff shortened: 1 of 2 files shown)


frontend/src/lib/uPlotV2/components/Tooltip/Tooltip.tsx
```diff
@@ -18,7 +18,10 @@ export default function Tooltip({
 	uPlotInstance,
 	timezone,
 	content,
+	isPinned,
```
AshwinBhatkal: We were rethinking this right?

frontend/src/container/DashboardContainer/visualization/charts/BarChart/BarChart.tsx
```diff
@@ -10,7 +10,15 @@ import { useBarChartStacking } from '../../hooks/useBarChartStacking';
 import { BarChartProps } from '../types';
 
 export default function BarChart(props: BarChartProps): JSX.Element {
-	const { children, isStackedBarChart, config, data, ...rest } = props;
+	const {
+		children,
+		isStackedBarChart,
+		renderTooltip: customRenderTooltip,
```
AshwinBhatkal: I don't like particularly like this pattern. If something is being overridden, then it should be named that way

This renderTooltip prop is an option to render custom tooltip. So it should have been named that way. I see it's usages in other files but that's how you avoid name conflicts. Like the one you have with renderTooltip in this file. When this collision happens, we end up adding random suffix/prefix just to make things fit in

ankitnayan:
a video recording in the PR with linked issue that the PR solves would be great

ahrefabhi:
> a video recording in the PR with linked issue that the PR solves would be great

This PR doesn't have any visual change @ankitnayan, I've just changed a few structural things that makes pinning easy to implement in code.

PR #10462, opened 2026-03-01
chore: address gaps in summary tab

srikanthccv:
## Pull Request

- fix issues with cmd+enter not working
- remove two tooltips because they explain the same, and instead an placeholder relevant
- change the default order for proportion view to samples
- add small spinner for metric details instead of big skeleton (bad UX for drawer kind of elements)
- add ability to expand all values for key instead of "Show more" with each time 5 because that's not going to be helpful
- change on click options for key/value from direction action to "copy" and "open in explorer" so user can choose relevant action instead of assumed action.
- remove the double-scrolls on page summary (another bad UX)
- the metric details entire section makes it looks everything clickable, address it
- the search contents from other page are propagated to summary page, address this

### DIFF
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -1763,6 +1763,134 @@ components:
       - type
       - orgId
       type: object
+    ServiceaccounttypesFactorAPIKey:
+      properties:
+        createdAt:
+          format: date-time
+          type: string
+        expires_at:
+          minimum: 0
+          type: integer
+        id:
+          type: string
+        key:
+          type: string
+        last_used:
+          format: date-time
+          type: string
+        name:
+          type: string
+        service_account_id:
+          type: string
+        updatedAt:
+          format: date-time
+          type: string
+      required:
+      - id
+      - key
+      - expires_at
+      - last_used
+      - service_account_id
+      type: object
+    ServiceaccounttypesGettableFactorAPIKeyWithKey:
+      properties:
+        id:
+          type: string
+        key:
+          type: string
+      required:
+      - id
+      - key
+      type: object
+    ServiceaccounttypesPostableFactorAPIKey:
+      properties:
+        expires_at:
+          minimum: 0
+          type: integer
+        name:
+          type: string
+      required:
+      - name
+      - expires_at
+      type: object
+    ServiceaccounttypesPostableServiceAccount:
+      properties:
+        email:
+          type: string
+        name:
+          type: string
+        roles:
+          items:
+            type: string
+          type: array
+      required:
+      - name
+      - email
+      - roles
+      type: object
+    ServiceaccounttypesServiceAccount:
+      properties:
+        createdAt:
+          format: date-time
+          type: string
+        email:
+          type: string
+        id:
+          type: string
+        name:
+          type: string
+        orgID:
+          type: string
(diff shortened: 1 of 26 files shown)


frontend/src/container/MetricsExplorer/Summary/Summary.tsx
```diff
@@ -87,7 +84,14 @@ function Summary(): JSX.Element {
 	const [
 		currentQueryFilterExpression,
 		setCurrentQueryFilterExpression,
-	] = useState<string>(query?.filter?.expression || '');
+	] = useState<string>('');
+
+	const [appliedFilterExpression, setAppliedFilterExpression] = useState('');
```
AshwinBhatkal: should the default state here be query.filter.expression? What would happen on reload?

frontend/src/container/MetricsExplorer/MetricDetails/MetricDetails.tsx
```diff
@@ -100,6 +107,21 @@ function MetricDetails({
 	const isActionButtonDisabled =
 		!metricName || isLoadingMetricMetadata || isErrorMetricMetadata;
 
+	const handleDrawerClose = useCallback(
+		(e: React.MouseEvent | React.KeyboardEvent): void => {
+			if ('key' in e && e.key === 'Escape') {
+				const openPopover = document.querySelector(
+					'.ant-popover:not(.ant-popover-hidden)',
```
AshwinBhatkal: We should avoid the usage of an antd class here. If we can use an id or custom classname, that would be better

frontend/src/container/MetricsExplorer/MetricDetails/__tests__/Metadata.test.tsx
```diff
@@ -324,6 +324,21 @@ describe('Metadata', () => {
 		expect(editButton2).toBeInTheDocument();
 	});
 
+	it('should show section header but not edit controls while loading', () => {
```
AshwinBhatkal: should we instead just disable the edit instead of hiding it?

frontend/src/container/MetricsExplorer/MetricDetails/AllAttributes.tsx
AshwinBhatkal: We should move this to a separate file

frontend/src/container/MetricsExplorer/Summary/Summary.tsx
srikanthccv: Updated, it was picking up the search expression from other pages. The way to address this is to start with initial qb state using `useShareBuilderUrl`

frontend/src/container/MetricsExplorer/MetricDetails/__tests__/Metadata.test.tsx
```diff
@@ -336,7 +336,8 @@ describe('Metadata', () => {
 		);
 
 		expect(screen.getByText('Metadata')).toBeInTheDocument();
-		expect(screen.queryByText('Edit')).not.toBeInTheDocument();
+		const editButton = screen.getByText('Edit').closest('button');
```
AshwinBhatkal, line 339: nit: for next time, getByRole is a better option here

PR #10465, opened 2026-03-02
fix: exclude internal attributes from promQL results

### LINKED ISSUE #8189: Exclude private internal attributes from the result in promql queries
Labels that start with `__` are private and provide no value to user. Exclude them from the result.

namanv3:
## Pull Request

---

### 📄 Summary

Currently, internal attributes like `__name__`, or `fingerprint` are shown in dashboards built using promql query. These need to be removed as they are not useful for the user in any way. 

Screenshot 2026-03-02 at 08 53 28

#### Issues closed by this PR

Closes #8189 

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context

For dashboards built using promQL, users see internal keys like __scope.name__ which they shouldn't see

Screenshot 2026-03-02 at 08 56 21

### DIFF
--- a/deploy/docker-swarm/docker-compose.ha.yaml
+++ b/deploy/docker-swarm/docker-compose.ha.yaml
@@ -190,7 +190,7 @@ services:
       # - ../common/clickhouse/storage.xml:/etc/clickhouse-server/config.d/storage.xml
   signoz:
     !!merge <<: *db-depend
-    image: signoz/signoz:v0.113.0
+    image: signoz/signoz:v0.114.0
     ports:
       - "8080:8080" # signoz port
     #   - "6060:6060"     # pprof port
@@ -213,7 +213,7 @@ services:
       retries: 3
   otel-collector:
     !!merge <<: *db-depend
-    image: signoz/signoz-otel-collector:v0.144.1
+    image: signoz/signoz-otel-collector:v0.144.2
     entrypoint:
       - /bin/sh
     command:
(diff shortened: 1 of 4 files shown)


srikanthccv:
The change looks good, just unsure if people are relying on the returned data and have anything built around based on internal attributes.

pkg/querier/promql_query.go
```diff
@@ -242,6 +242,9 @@ func (q *promqlQuery) Execute(ctx context.Context) (*qbv5.Result, error) {
 		var s qbv5.TimeSeries
 		lbls := make([]*qbv5.Label, 0, len(v.Metric))
 		for name, value := range v.Metric.Copy().Map() {
+			if strings.HasPrefix(name, "__") || name == "fingerprint" {
+				continue
+			}
```
srikanthccv: Let's make an exception for `__name__` and include it. The rest of them and `fingerprint`, drop it.

namanv3: Is this cuz you think `__name__` may be useful enough for people to do something with? I'll add it back.
Also how about we add some UI flag to include or exclude internal fields in the results? If after this change goes in and we get a bug saying that they need some internal flag, we can add the toggle

srikanthccv: > Is this cuz you think __name__ may be useful enough for people to do something with?

Yes

> Also how about we add some UI flag to include or exclude internal fields in the results? If after this change goes in and we get a bug saying that they need some internal flag, we can add the toggle

Every option on UI is tax to the most of the users that don't use it. It's not clear how they use other internal attributes. The remaining internal attributes are `__temporality__` and `__scope.***`. I can't think of any way how they can be useful. Let's rollout with just `__name__` and see how it goes.

namanv3: Yeah yeah, I'm assuming no user will actually complain about a missing internal label

PR #10467, opened 2026-03-02
Upgrade prometheus/common and prometheus/prometheus  to latest available version

namanv3:
## Pull Request

---

### 📄 Summary

Needed to unblock https://github.com/SigNoz/signoz/pull/10325

#### Issues closed by this PR

Closes https://github.com/SigNoz/engineering-pod/issues/4102

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

### 📝 Changelog

N/A

### 📋 Checklist
- [ ] Manually tested

---

> [!NOTE]
> **Medium Risk**
> Large dependency and toolchain upgrades (Prometheus/OTel + Go 1.25) can introduce subtle build/runtime or API-compatibility regressions. Frontend template change may alter PagerDuty payload shape and could impact downstream integrations.
> 
> **Overview**
> Upgrades the build/runtime toolchain to Go `1.25` by updating `go.mod` and switching enterprise integration Docker images from `golang:1.24-bullseye` to `golang:1.25-bookworm`.
> 
> Refreshes a wide set of Go module dependencies, notably bumping `prometheus/common` and `prometheus/prometheus` (plus related Alertmanager/OTel collector packages), and adjusts default PagerDuty payload templates in the frontend (and mock data) to emit firing/resolved alert lists via `toJson` instead of the previous `pagerduty.default.instances` template.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8e40060e1b9ba66bc9b037d6017c9b5ed0bef425. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/src/container/CreateAlertChannels/defaults.ts
+++ b/frontend/src/container/CreateAlertChannels/defaults.ts
@@ -16,8 +16,8 @@ export const PagerInitialConfig: Partial<PagerChannel> = {
 	client: 'SigNoz Alert Manager',
 	client_url: 'https://enter-signoz-host-n-port-here/alerts',
 	details: JSON.stringify({
-		firing: `{{ template "pagerduty.default.instances" .Alerts.Firing }}`,
-		resolved: `{{ template "pagerduty.default.instances" .Alerts.Resolved }}`,
+		firing: `{{ .Alerts.Firing | toJson }}`,
+		resolved: `{{ .Alerts.Resolved | toJson }}`,
 		num_firing: '{{ .Alerts.Firing | len }}',
 		num_resolved: '{{ .Alerts.Resolved | len }}',
 	}),
--- a/frontend/src/mocks-server/__mockdata__/alerts.ts
+++ b/frontend/src/mocks-server/__mockdata__/alerts.ts
@@ -32,8 +32,8 @@ export const editSlackDescriptionDefaultValue = `{{ range .Alerts -}} *Alert:* {
 export const pagerDutyDescriptionDefaultVaule = `{{ if gt (len .Alerts.Firing) 0 -}} Alerts Firing: {{ range .Alerts.Firing }} - Message: {{ .Annotations.description }} Labels: {{ range .Labels.SortedPairs }} - {{ .Name }} = {{ .Value }} {{ end }} Annotations: {{ range .Annotations.SortedPairs }} - {{ .Name }} = {{ .Value }} {{ end }} Source: {{ .GeneratorURL }} {{ end }} {{- end }} {{ if gt (len .Alerts.Resolved) 0 -}} Alerts Resolved: {{ range .Alerts.Resolved }} - Message: {{ .Annotations.description }} Labels: {{ range .Labels.SortedPairs }} - {{ .Name }} = {{ .Value }} {{ end }} Annotations: {{ range .Annotations.SortedPairs }} - {{ .Name }} = {{ .Value }} {{ end }} Source: {{ .GeneratorURL }} {{ end }} {{- end }}`;
 
 export const pagerDutyAdditionalDetailsDefaultValue = JSON.stringify({
-	firing: `{{ template "pagerduty.default.instances" .Alerts.Firing }}`,
-	resolved: `{{ template "pagerduty.default.instances" .Alerts.Resolved }}`,
+	firing: `{{ .Alerts.Firing | toJson }}`,
+	resolved: `{{ .Alerts.Resolved | toJson }}`,
 	num_firing: '{{ .Alerts.Firing | len }}',
 	num_resolved: '{{ .Alerts.Resolved | len }}',
 });


pkg/alertmanager/alertmanagerserver/server.go
```diff
@@ -190,7 +190,7 @@ func New(ctx context.Context, logger *slog.Logger, registry prometheus.Registere
 		})
 	}()
 
-	server.alerts, err = mem.NewAlerts(ctx, server.marker, server.srvConfig.Alerts.GCInterval, nil, server.logger, signozRegisterer)
+	server.alerts, err = mem.NewAlerts(ctx, server.marker, server.srvConfig.Alerts.GCInterval, 0, alertmanagertypes.AlertStoreCallback{}, server.logger, signozRegisterer, nil)
```
srikanthccv: 1. Can you look at the upstream changes and share why was this call back introduced?
2. What are these arguments changed to `0` and `nil`??

pkg/prometheus/clickhouseprometheus/client.go
```diff
@@ -87,6 +87,24 @@ func (client *client) Read(ctx context.Context, query *prompb.Query, sortSeries
...
+		return c.Read(ctx, queries[0], sortSeries)
+	}
+	sets := make([]storage.SeriesSet, 0, len(queries))
+	for _, q := range queries {
+		ss, err := c.Read(ctx, q, sortSeries)
+		if err != nil {
+			return nil, err
+		}
+		sets = append(sets, ss)
+	}
```
srikanthccv: One way to do it but naive implementation? Could we combine all of them into one query? Or do in parallel?

Additionally can you share when will this be called vs `c.Read` multiple times by prometheus engine?

pkg/prometheus/utils.go
```diff
@@ -19,26 +20,18 @@ func RemoveExtraLabels(res *promql.Result, labelsToRemove ...string) error {
...
-			for _, lbl := range series.Metric {
-				if _, drop := toRemove[lbl.Name]; !drop {
-					dst = append(dst, lbl)
-				}
-			}
-			series.Metric = dst
+			b := labels.NewBuilder(value[i].Metric)
+			b.Del(labelsToRemove...)
+			newLabels := b.Labels()
+			value[i].Metric = newLabels
```
srikanthccv, line 21: Interesting, do you need all of this? Is there way to remove labels in-place for `value[i].Metric`?

pkg/types/alertmanagertypes/channel_test.go
```diff
@@ -106,11 +109,12 @@ func TestNewConfigFromChannels(t *testing.T) {
 						"client_url":    "{{ template \"pagerduty.default.clientURL\" . }}",
 						"description":   "{{ template \"pagerduty.default.description\" .}}",
 						"source":        "{{ template \"pagerduty.default.client\" . }}",
+						"timeout":       float64(0),
 						"details": map[string]any{
-							"firing":       "{{ template \"pagerduty.default.instances\" .Alerts.Firing }}",
+							"firing":       "{{ .Alerts.Firing | toJson }}",
```
srikanthccv, line 114: Is this change needed? We have saved templates for PD with this as part of it. So if this change is needed we need migraiton.

pkg/prometheus/utils.go
namanv3: yeah, the Labels object has only one edit function DropReserve that only drops `__` labels

pkg/alertmanager/alertmanagerserver/server.go
namanv3:
```
func NewAlerts(
	ctx context.Context,
	m types.AlertMarker,
	intervalGC time.Duration,
	perAlertNameLimit int,
	alertCallback AlertStoreCallback,
	l *slog.Logger,
	r prometheus.Registerer,
	flagger featurecontrol.Flagger,
)
```
`0` is for `perAlertNameLimit int,` and 0 equates to having no limit. Cuz we didn't have any before, so I didnt introduce a limit here
The `nil` is for  `flagger featurecontrol.Flagger`,  and the library code adds a noop impl that always returns false
```
        if flagger == nil {
		flagger = featurecontrol.NoopFlags{}
	}
```
I didn't find any of the flags mentioned in this Flagger in our code base hence I set the value to nil

namanv3: On the Alert Store Callback, it was already there in the previous version of the library, I missed it somehow. We were sending `nil` earlier and we can do that now as well. I've removed the custom thing I added.

pkg/types/alertmanagertypes/channel_test.go
namanv3: Yeah, this change upstream is the reason: https://github.com/prometheus/alertmanager/commit/e3c74f565bd4d18c87112a8d11a7a8855bab6ba5

srikanthccv: Some context if you didn't follow my earlier. When users create a pager duty "notification channel" we save the default template to DB, they may optionally edit as well. This saved tempted is used by alertmanager. If it is backward compatible change we are good. If it's a breaking change we need migration.

Is this a backward incompatible change? Does it require migration?

namanv3: What has changed is the default template for details, and that details are now `map[string]any` instead of `map[string]string`. If someone sends a custom template, or the old default, the prometheus/alertmanager code will still handle that, but now what were strings earlier can render as objects.

The string->object shift happened in the above PR because of this change in it:
Screenshot 2026-03-03 at 22 16 38

Alert sent by main branch for old default Additional Information: https://signoz-1.pagerduty.com/alerts/Q3XB60WYFKTO5K/log_entries/R9FEPWDZN08MV2OBVX6AOZZMRV
Alert sent by this branch for  old default Additional Information: https://signoz-1.pagerduty.com/alerts/Q3423U76R81Q38/log_entries/R2MD14SL7NRS3CZZER61O2I2WQ

If someone has Additional Information as empty in their notification channel, then because over here we fill it with the default, the new default will show up: https://github.com/SigNoz/signoz/blob/ccb88ab03fd4c25c7a0c3aa33bfbf24d1e980a56/pkg/types/alertmanagertypes/receiver.go#L46, we fill in the default if it's empty.
Alert sent by main branch when Additional Information is empty: https://signoz-1.pagerduty.com/alerts/Q3WIKREY6D68AJ/log_entries/R2U6KAXICNESC9OSTJ4HWK6BJU
Alert sent by this branch: https://signoz-1.pagerduty.com/alerts/Q1I147SUMP5KBP/log_entries/R6QIQ1FD2937LQ8L6BU1SBCV8K

namanv3: All in all, while the behaviour does change for the user at PagerDuty level, it doesn't break anything unless the user has some script that is consuming that PagerDuty alert. For such scripts that would be expecting strings, this is indeed a breaking change.

namanv3: Is it that big of a use case, or can we assume that there are no such scripts involved??

pkg/prometheus/clickhouseprometheus/client.go
namanv3: The underlying library code never calls ReadMultiple. Their own implementation's Read calls ReadMultiple, that's it.
Screenshot 2026-03-03 at 22 31 09

As for combining the queries, don't think that we can do that cuz each query will have its own time range and hence its own table (main vs some agg table), so I've added parallelisation instead.

srikanthccv: That's strange, can you please check the original intent behind introducing behind this only to not use in it's own codebase.

pkg/types/alertmanagertypes/channel_test.go
srikanthccv: We never know how people are using it. And we broke alerting in the past with false assumptions so let's at least communicate the change. We have some analytics for the cloud users, let's collect how many are using PD and reach out to them beforehand.

pkg/prometheus/clickhouseprometheus/client.go
namanv3: In the PR that introduced this method, the change seems to be done only to make use of the protobuf properly, and cuz this person wants to use read multiple themselves. https://github.com/prometheus/prometheus/pull/16742/changes

This is where they're using the proto definition of read request properly: https://github.com/prometheus/prometheus/pull/16742/changes#diff-a9748b2f65f05f64cd2108b2db1e717a1f940e9ab859838f011c822fd1e39973R366

srikanthccv: Thanks for the details.

pkg/types/alertmanagertypes/channel_test.go
```diff
@@ -106,11 +109,12 @@ func TestNewConfigFromChannels(t *testing.T) {
...
 							"num_firing":   "{{ .Alerts.Firing | len }}",
 							"num_resolved": "{{ .Alerts.Resolved | len }}",
-							"resolved":     "{{ template \"pagerduty.default.instances\" .Alerts.Resolved }}",
+							"resolved":     "{{ .Alerts.Resolved | toJson }}",
```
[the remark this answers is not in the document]
namanv3: @srikanthccv should we make this change? The current template does work so there is no real reason to change this right away.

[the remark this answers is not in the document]
srikanthccv: Let's change it. Once channel is created the template is already save the current channel won't have any change but all new future channels created will have a better formatted message/payload.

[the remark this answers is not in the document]
namanv3: @cursor push 98db18f3ec

pkg/types/alertmanagertypes/channel_test.go
namanv3: This is done now

pkg/types/alertmanagertypes/alert.go
```diff
@@ -90,8 +91,8 @@ func NewDeprecatedGettableAlertsFromGettableAlerts(gettableAlerts GettableAlerts
 }
 
 // Converts a slice of PostableAlert to a slice of Alert.
-func NewAlertsFromPostableAlerts(postableAlerts PostableAlerts, resolveTimeout time.Duration, now time.Time) ([]*types.Alert, []error) {
-	alerts := v2.OpenAPIAlertsToAlerts(postableAlerts)
+func NewAlertsFromPostableAlerts(postableAlerts PostableAlerts, resolveTimeout time.Duration, now time.Time, ctx context.Context) ([]*types.Alert, []error) {
```
srikanthccv: Make context as the first argument.

pkg/prometheus/clickhouseprometheus/client.go
```diff
@@ -87,6 +88,41 @@ func (client *client) Read(ctx context.Context, query *prompb.Query, sortSeries
...
+	wg.Wait()
+
+	sets := make([]storage.SeriesSet, 0, len(queries))
+	for _, r := range results {
+		if r.err != nil {
+			return nil, r.err
+		}
+		sets = append(sets, r.ss)
+	}
```
srikanthccv, line 122: Should the order of items in the `sets` be same as the order of the `queries`? What happens otherwise?

srikanthccv (commented):
Final comments

srikanthccv:
@abhishekhugetech FYI

pkg/prometheus/clickhouseprometheus/client.go
namanv3: this is from the upstream impl of this method:
```
// Returns a single SeriesSet with interleaved series from all queries.
func (c *Client) ReadMultiple(ctx context.Context, queries []*prompb.Query, sortSeries bool) (storage.SeriesSet, error) {
```
So the expected behaviour is to combine all the series being returned for all the queries, and that is what NewMergeSeriesSet does, and the library also uses it . 
Note:  it only uses NewMergeSeriesSet  when `sortSeries` is true, and when `sortSeries` is false it uses the internal `concreteSeriesSet` impl. However, there is no harm in using NewMergeSeriesSet even if `sortSeries` is false cuz its not like sorting is a problem when it wasn't asked for (which is something the impl also mentions by saying that the server can send sorted response anyway), so its alright to use it always in our case anyway.

pkg/types/alertmanagertypes/alert.go
namanv3: Done

PR #10477, opened 2026-03-03
chore: updated config builder types

ahrefabhi:
## Pull Request

---

### 📄 Summary
- Replaced widgetId with Id and made it mandatory in uplotconfigbuilder


frontend/src/lib/uPlotV2/config/UPlotConfigBuilder.ts
```diff
@@ -380,8 +378,8 @@ export class UPlotConfigBuilder extends ConfigBuilder<
 	/**
 	 * Get the widget id
```
AshwinBhatkal: comment change

Issue #10478, opened 2026-03-03
dynamic variables don't load in imported dashboards with empty textbox variable


AshwinBhatkal:
## Bug description

*Please describe.*  
*If this affects the front-end, screenshots would be of great help.*  

Dynamic variables don't load in imported dashboards with empty textbox variable

## Expected behavior

Dynamic variables should load independent of the textbox having any value or not

## How to reproduce

1. Import this JSON - [dynTextVar.json](https://github.com/user-attachments/files/25710566/dynTextVar.json)
2. Dynamic variables will wrongly say it is waiting on query vars even when they don't exist

PR #10480, opened 2026-03-03
chore(frontend): dynamic vars load with empty textbox variables

AshwinBhatkal:
### 📄 Summary

Dynamic variables don't load in imported dashboards with empty textbox variable. This PR fixes that

##### Before

https://github.com/user-attachments/assets/675d7513-a35e-4d54-827f-ead5942f689a

##### After

https://github.com/user-attachments/assets/576589ac-a120-428c-915b-a12d6db044d1

#### Issues closed by this PR
Closes https://github.com/SigNoz/signoz/issues/10478

---

### ✅ Change Type

- [x] 🐛 Bug fix

---

#### Root Cause

Faulty Assumption

Fixes the assumption

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered: Yes

---

- Blast radius: Loading of dynamic variables
- Potential regressions: None
- Rollback plan: Fix forward

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Imported dashboards with empty textbox variables made the dynamic variable to wait for options forever |

---

### 📋 Checklist
- [x] Manually tested


frontend/src/providers/Dashboard/store/dashboardVariables/dashboardVariablesStore.ts
```diff
@@ -69,27 +69,21 @@ export function updateDashboardVariablesStore({
...
+	// Dynamic variables should only wait on query variables having values,
+	// not on CUSTOM, TEXTBOX, or other types.
+	const doAllQueryVariablesHaveValuesSelected = Object.values(
+		state.variables,
+	).every((variable) => {
+		if (variable.type !== 'QUERY') {
+			return true;
+		}
+		return (
+			!isUndefined(variable.selectedValue) && !isEmpty(variable.selectedValue)
```
ahrefabhi: !isEmpty(variable.selectedValue) will take care of both the things here, we don't need to do !isUndefined(variable.selectedValue), because !isEmpty(undefined) = false

and also this check will fail if the selectedValue is 0 or false??

AshwinBhatkal: Thanks for the comment. Resolved

PR #10486, opened 2026-03-03
feat(authz): migrate authorization to authz instead of user.role

H4ad:
## Pull Request

Instead of using `user.role` to check for user's permission, migrate to use new AuthZ system.

---

### 📄 Summary

closes: https://github.com/SigNoz/platform-pod/issues/1785

To migrate, we will map/use the following roles:
- ADMIN: signoz-admin
- EDITOR: signoz-editor
- VIEWER: signoz-viewer

These roles should map 1:1 the functionality of the current system auth system.

I tried to be minimal in the code changes, the new AuthZ system will change how we show permission errors in the UI.

## Screenshots

User is admin:
https://github.com/user-attachments/assets/078480e6-eea6-4b3a-a1cf-4402deaa72a1

User is Editor (email is viewer but the role is editor):
https://github.com/user-attachments/assets/637d90d2-8b49-4845-b94f-24447cdb0fea

Logout from admin to viewer:
https://github.com/user-attachments/assets/3aa2c1e6-997b-4309-8462-e190302b59ec

New Un-Authorized page for Viewer/Editor/Admin:
image

New Un-Authorized page for Anonymous:
image

When anonymous, the user can't access any page other than support/un-authorized/not found.

Public Dashboard:
image

### ✅ Change Type

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🧪 Testing Strategy

- Manual verification: Only manual testing

---

- Blast radius: Areas dependent on the user role from the `user/me` response. 
- Potential regressions: `ADMIN` not able to view admin action items because the role check failed ? 
- Rollback plan: rollback the PR as there are no changes in sync with backend here. 

---

### 📝 Changelog
NA

---

### 📋 Checklist
- [x] Manually tested

## Notes to the Reviewer/Tester

I didn't clear the permissions after logout because this could trigger a loop of 401 errors.
Checkz should not trigger logout due to possible loops, this can also break the signup page due to redirect to login after open the signup link.
Just for these roles, we use `isFetching` instead of `isLoading` to not show cached results to the user.

### DIFF
--- a/frontend/src/pages/UnAuthorized/index.styles.scss
+++ b/frontend/src/pages/UnAuthorized/index.styles.scss
@@ -1,5 +1,5 @@
 .unauthorized-page {
-  &__description {
-    text-align: center;
-  }
+	&__description {
+		text-align: center;
+	}
 }


frontend/src/api/index.ts
```diff
@@ -102,9 +103,7 @@ export const interceptorRejected = async (
 									...value.config.headers,
 									Authorization: `Bearer ${response.data.accessToken}`,
 								},
-								data: {
-									...JSON.parse(value.config.data || '{}'),
-								},
+								data: value.config.data ? JSON.parse(value.config.data) : undefined,
```
H4ad: Without this, when someone has their role updated, after refresh the page, it has the behavior of:
- Get 401 due to jwt be staled
- Resend the request with the previous data
- Due to data being an array, turns this to `{"0": { bla }}`
- Get error 400

frontend/src/providers/App/__tests__/App.test.tsx
```diff
@@ -0,0 +1,268 @@
...
+	AuthtypesTransactionDTO,
+} from 'api/generated/services/sigNoz.schemas';
+import { LOCALSTORAGE } from 'constants/localStorage';
+import { server } from 'mocks-server/server';
+import { rest } from 'msw';
+import { USER_ROLES } from 'types/roles';
+
+import { AppProvider, useAppContext } from '../App';
+
+const AUTHZ_CHECK_URL = 'http://localhost/api/v1/authz/check';
```
SagarRajput-7, line 17: above you mentioned  `/authz/check2`

frontend/src/providers/App/__tests__/App.test.tsx
```diff
@@ -0,0 +1,268 @@
...
+					ctx.status(200),
+					ctx.json(authzMockResponse(payload, [true, false, false])),
+				);
+			}),
+		);
+
+		const wrapper = createWrapper();
+		const { result } = renderHook(() => useAppContext(), { wrapper });
+
+		await new Promise((r) => setTimeout(r, SINGLE_FLIGHT_WAIT_MS));
```
SagarRajput-7: why setTimout here, and we have timeout also in the waitfor below?

frontend/src/providers/App/App.tsx
```diff
@@ -43,7 +49,7 @@ export const AppContext = createContext<IAppContext | undefined>(undefined);
 
 export function AppProvider({ children }: PropsWithChildren): JSX.Element {
 	// on load of the provider set the user defaults with access token , refresh token from local storage
-	const [user, setUser] = useState<IUser>(() => getUserDefaults());
+	const [defaultUser, setUser] = useState<IUser>(() => getUserDefaults());
```
SagarRajput-7: little confusing on the naming here the setter is still called setUser

frontend/src/providers/App/App.tsx
```diff
@@ -70,14 +76,47 @@ export function AppProvider({ children }: PropsWithChildren): JSX.Element {
...
+		// if none of the permissions, so anonymous
+		return USER_ROLES.ANONYMOUS;
+	}, [permissionsResult]);
+
+	const user: IUser = useMemo(() => {
+		return {
+			...defaultUser,
+			role: userRole as ROLES,
+		};
+	}, [defaultUser, userRole]);
```
SagarRajput-7, line 118: Minor concern: since now the default is changed from Viewer to Anonymous and we rely on the permissionsResult to override the access once resolved, so for the time inbetween and in case of any failure the default fallback changed to Anonymous from Viewer, should we check if any of the flow or ui will get affected or break from this, as we do not handle anonymous as of now.

frontend/src/providers/App/__tests__/App.test.tsx
H4ad: Great catch, I changed to test but I forgot to rollback to just `check`.

frontend/src/providers/App/__tests__/App.test.tsx
H4ad: Due to the useAuthZ single flight support, we have concurrency issues in the test due to mock in server.use conflicting with other mocks in the other tests.

This timeout was introduced to wait for the single flight to resolve.

H4ad: Updated to be more clear what is the purpouse of this call

frontend/src/providers/App/App.tsx
H4ad: I forgot to test this case, and from what I saw, I only needed to correct the `routePermission` to allow anonymous access in `un-authorized`.

Basically, what happens now is that the user doesn't have permission to navigate to any page.

I'm not sure if this is the best behavior; maybe we should use the viewer as an alternative.

@YounixM, any thoughts on this?

H4ad: After talking to Yunus, the un-authorized page was updated to better handle the anonymous role.

frontend/src/pages/UnAuthorized/index.tsx
```diff
@@ -1,20 +1,51 @@
...
-				<UnAuthorized />
-				<Typography.Title level={3}>
-					Oops.. you don&apos;t have permission to view this page
-				</Typography.Title>
-
-				<Button to={ROUTES.HOME} tabIndex={0} className="periscope-btn primary">
-					Return To Home
-				</Button>
+				<UnAuthorized width={64} height={64} />
+				<Typography.Title level={3}>Access Restricted</Typography.Title>
```
YounixM, line 34: Avoid using Antd components

PR #10487, opened 2026-03-04
fix: bug identified signalsource was passed as meter

ishan621:
## Pull Request
We identified a bug where signalSource was passed as meter in value suggestions api but it lead to incomplete results in specific infrastructure-monitoring route. We should pass it as empty string rather than passing hard-code meter here. Value already passes signal correctly which identifies as logs/traces etc

---

We identified a bug where signalSource was being passed as "meter" in the Value Suggestions API. This caused incomplete results on the infrastructure-monitoring route.
What problem does it solve, and why is this the right approach?
The Value Suggestions API already receives the correct signal (logs, traces, etc.) through the value parameter. Passing a hard-coded "meter" overrides the intended behavior and restricts the results.
By sending an empty string instead of "meter", we allow the backend to correctly infer the signal from the existing parameter, restoring complete and accurate suggestions across routes.

Before :
Screenshot 2026-03-04 at 8 13 56 PM

After: 

Screenshot 2026-03-04 at 8 14 58 PM

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context

Incorrect param passed
Faulty assumption

Manually

- Blast radius:
- Potential regressions:
- Rollback plan:

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/components/CeleryOverview/CeleryOverviewTable/CeleryOverviewTable.tsx
+++ b/frontend/src/components/CeleryOverview/CeleryOverviewTable/CeleryOverviewTable.tsx
@@ -16,7 +16,7 @@ import {
 	Tooltip,
 	Typography,
 } from 'antd';
-import { FilterDropdownProps } from 'antd/lib/table/interface';
+import type { FilterDropdownProps } from 'antd/lib/table/interface';
 import logEvent from 'api/common/logEvent';
 import {
 	getQueueOverview,
--- a/frontend/src/components/CeleryTask/CeleryTaskConfigOptions/useGetCeleryFilters.ts
+++ b/frontend/src/components/CeleryTask/CeleryTaskConfigOptions/useGetCeleryFilters.ts
@@ -1,7 +1,7 @@
 import { useQuery } from 'react-query';
 // eslint-disable-next-line no-restricted-imports
 import { useSelector } from 'react-redux';
-import { DefaultOptionType } from 'antd/es/select';
+import type { DefaultOptionType } from 'antd/es/select';
 import { getAttributesValues } from 'api/queryBuilder/getAttributesValues';
 import { DATA_TYPE_VS_ATTRIBUTE_VALUES_KEY } from 'constants/queryBuilder';
 import { REACT_QUERY_KEY } from 'constants/reactQueryKeys';
--- a/frontend/src/components/CeleryTask/useCeleryFilterOptions.ts
+++ b/frontend/src/components/CeleryTask/useCeleryFilterOptions.ts
@@ -1,5 +1,5 @@
 import { useState } from 'react';
-import { DefaultOptionType } from 'antd/es/select';
+import type { DefaultOptionType } from 'antd/es/select';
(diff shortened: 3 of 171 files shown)


frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.tsx
```diff
@@ -127,7 +127,7 @@ export default function CheckboxFilter(props: ICheckboxProps): JSX.Element {
 	} = useGetQueryKeyValueSuggestions({
 		key: filter.attributeKey.key,
 		signal: filter.dataSource || DataSource.LOGS,
-		signalSource: 'meter',
+		signalSource: filter.dataSource === DataSource.METRICS ? 'meter' : '',
```
srikanthccv: Even if the `filter.dataSource` is metrics, it shouldn't pass the `signalSource`. The `signalSource` should only be passed if the quick filters are used in meter explorer.

srikanthccv: Additionally why is the `signal` sent as logs in infra monitoring page?
```
https://app.us.staging.signoz.cloud/api/v1/fields/values?signal=logs&name=deployment.environment&searchText=&metricName=&source=meter&startUnixMilli=1772634952639&endUnixMilli=1772636752639
```

srikanthccv: Regarding the source, look at this

https://github.com/SigNoz/signoz/blob/e8e85a3790e823a80d0296f50663ea94a4126ec1/pkg/telemetrymetadata/metadata.go#L1562-L1567

If the source is specified as meter, then the meter fields are returned, however, the value should be from metrics source.

ishan621: > Additionally why is the `signal` sent as logs in infra monitoring page?
> 
> ```
> https://app.us.staging.signoz.cloud/api/v1/fields/values?signal=logs&name=deployment.environment&searchText=&metricName=&source=meter&startUnixMilli=1772634952639&endUnixMilli=1772636752639
> ```

@srikanthccv  datasource as metrics was added in infra utils files earlier but was removed in this PR - https://github.com/SigNoz/signoz/pull/8357 , not sure why though

@amlannandy  what were the reasons to remove it from utils ?

ishan621: > Even if the `filter.dataSource` is metrics, it shouldn't pass the `signalSource`. The `signalSource` should only be passed if the quick filters are used in meter explorer.

@srikanthccv  as only 'meter-explorer' needs signalsource , we have made the respective changes

PR #10488, opened 2026-03-04
chore: add rule state history module

srikanthccv:
### 📄 Summary

This change introduce a new module with /v2 APIs for the rule state history, the only notable change is the change of `filter` from the list of items to filter expression

#### Issues closed by this PR

Part of https://github.com/SigNoz/engineering-pod/issues/3199

### ✅ Change Type

- [x] ♻️ Refactor
### 🧪 Testing Strategy

> How was this change validated?

- Tests added/updated:for any meaningful, this needs a nightly kind of setup than integration setup
- Manual verification:yes
- Edge cases covered:yes

- Blast radius: rule history page
- Potential regressions: rule history not recorded
- Rollback plan: no rollback

### 📋 Checklist
- [x] Manually tested

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 21bab3ba72bbfce7b1583968ab073753f968d319. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/__mocks__/useSafeNavigate.ts
+++ b/frontend/__mocks__/useSafeNavigate.ts
@@ -2,6 +2,7 @@
 interface SafeNavigateOptions {
 	replace?: boolean;
 	state?: unknown;
+	newTab?: boolean;
 }
 
 interface SafeNavigateTo {
@@ -20,9 +21,7 @@ interface UseSafeNavigateReturn {
 
 export const useSafeNavigate = (): UseSafeNavigateReturn => ({
 	safeNavigate: jest.fn(
-		(to: SafeNavigateToType, options?: SafeNavigateOptions) => {
-			console.log(`Mock safeNavigate called with:`, to, options);
-		},
+		(_to: SafeNavigateToType, _options?: SafeNavigateOptions) => {},
 	) as jest.MockedFunction<
 		(to: SafeNavigateToType, options?: SafeNavigateOptions) => void
 	>,
--- a/frontend/src/components/LogDetail/index.tsx
+++ b/frontend/src/components/LogDetail/index.tsx
@@ -1,6 +1,6 @@
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 // eslint-disable-next-line no-restricted-imports
-import { useSelector } from 'react-redux';
+import { useSelector } from 'react-redux'; // old code, TODO: fix this correctly
 import { useCopyToClipboard, useLocation } from 'react-use';
 import { Color, Spacing } from '@signozhq/design-tokens';
 import { Button, Divider, Drawer, Radio, Tooltip, Typography } from 'antd';
@@ -49,6 +49,7 @@ import { AppState } from 'store/reducers';
 import { Query, TagFilter } from 'types/api/queryBuilder/queryBuilderData';
 import { DataSource, StringOperators } from 'types/common/queryBuilder';
 import { GlobalReducer } from 'types/reducer/globalTime';
+import { isModifierKeyPressed } from 'utils/app';
 
 import { RESOURCE_KEYS, VIEW_TYPES, VIEWS } from './constants';
 import { LogDetailInnerProps, LogDetailProps } from './LogDetail.interfaces';
@@ -221,7 +222,7 @@ function LogDetailInner({
 	};
 
 	// Go to logs explorer page with the log data
-	const handleOpenInExplorer = (): void => {
+	const handleOpenInExplorer = (e?: React.MouseEvent): void => {
 		const queryParams = {
 			[QueryParams.activeLogId]: `"${log?.id}"`,
 			[QueryParams.startTime]: minTime?.toString() || '',
@@ -234,7 +235,9 @@ function LogDetailInner({
 				),
 			),
 		};
-		safeNavigate(`${ROUTES.LOGS_EXPLORER}?${createQueryParams(queryParams)}`);
+		safeNavigate(`${ROUTES.LOGS_EXPLORER}?${createQueryParams(queryParams)}`, {
+			newTab: !!e && isModifierKeyPressed(e),
+		});
 	};
 
 	const handleQueryExpressionChange = useCallback(
--- a/frontend/src/container/CreateAlertRule/SelectAlertType/index.tsx
+++ b/frontend/src/container/CreateAlertRule/SelectAlertType/index.tsx
@@ -6,6 +6,7 @@ import { ALERTS_DATA_SOURCE_MAP } from 'constants/alerts';
 import { FeatureKeys } from 'constants/features';
 import { useAppContext } from 'providers/App/App';
 import { AlertTypes } from 'types/api/alerts/alertTypes';
+import { isModifierKeyPressed } from 'utils/app';
 
 import { getOptionList } from './config';
 import { AlertTypeCard, SelectTypeContainer } from './styles';
@@ -70,8 +71,8 @@ function SelectAlertType({ onSelect }: SelectAlertTypeProps): JSX.Element {
 								</Tag>
 							) : undefined
 						}
-						onClick={(): void => {
-							onSelect(option.selection);
+						onClick={(e): void => {
+							onSelect(option.selection, isModifierKeyPressed(e));
 						}}
 						data-testid={`alert-type-card-${option.selection}`}
 					>
@@ -108,7 +109,7 @@ function SelectAlertType({ onSelect }: SelectAlertTypeProps): JSX.Element {
 }
 
 interface SelectAlertTypeProps {
-	onSelect: (typ: AlertTypes) => void;
+	onSelect: (type: AlertTypes, newTab?: boolean) => void;
 }
 
 export default SelectAlertType;
--- a/frontend/src/container/CreateAlertV2/Footer/Footer.tsx
+++ b/frontend/src/container/CreateAlertV2/Footer/Footer.tsx
@@ -4,6 +4,7 @@ import { Button, Tooltip, Typography } from 'antd';
 import { useQueryBuilder } from 'hooks/queryBuilder/useQueryBuilder';
 import { useSafeNavigate } from 'hooks/useSafeNavigate';
 import { Check, Loader, Send, X } from 'lucide-react';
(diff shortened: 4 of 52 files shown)


srikanthccv:
>i see we have introduced Response structs. We do follow Gettable / Listable nomenclature. Any reason they aren't on the same nomenclature ?

1. there is no "postable" items here. the state is recorded internally, and sent back in to user
2. there is no "gettable" item in the same sense we have gettable regular CRUD endpoints

>certain handler functions have unwanted cursor properties, is it because they eventually will handle cursors ? If not then lets have the same as defined in apiserver.

Avoid the vagueness in review comments. Please be explicit. I reviewed the API and handler before requesting review, there is only one place I see with cursor, `GetRuleHistoryTimeline`, which is the expected. Please share what handler functions do you see with unwanted cursor properties.

>add comments in the module interface with what they are meant for. Didn't understand the difference between RecordRuleState and AddRuleState. Code comments help

Will add

>func toQBLabels(raw rulestatehistorytypes.LabelsString) []*qbtypes.Label { why not make a function on LabelsString type to convert into a different type ? Why a danging handler function ?

I prefer writing private methods over the public methods until things become clear (even when they already look clear and there is nothing to worry about). This servers well in my experience and you probably have seen this being the point of discussion. You will continue to see the same pattern. Unless there is something wrong about any specific case of implementation, I would prefer to follow the same approach, scoped and private at first, then public.

>couple of places the arrays are required but are nullable is it intentional ?

Yes, that's intentional. If the number of items can be zero, nullable. Here is my reasoning. The client should always rely on `total` field for some of its logical purposes. It's redundant to send empty list, and total zero. When I send empty, it gives a false promise that the list will always be non-null leads to sort of logic that disregards the `total` and uses `list.length`. The nullable is false only when something must always exist.

pkg/types/rulestatehistorytypes/response.go
```diff
@@ -0,0 +1,49 @@
...
+	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
+)
+
+type RuleStateTimelineResponse struct {
+	Items      []RuleStateHistoryResponseItem `json:"items" required:"true"`
+	Total      uint64                         `json:"total" required:"true"`
+	NextCursor string                         `json:"nextCursor,omitempty"`
+}
+
+type RuleStateHistoryResponseItem struct {
```
vikrantgupta25:
```suggestion
type GettableRuleStateHistory struct {
```

pkg/types/rulestatehistorytypes/response.go
```diff
@@ -0,0 +1,49 @@
+package rulestatehistorytypes
+
+import (
+	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
+)
+
+type RuleStateTimelineResponse struct {
```
vikrantgupta25:
```suggestion
type GettableRuleStateTimeline struct {
```

pkg/types/rulestatehistorytypes/response.go
```diff
@@ -0,0 +1,49 @@
...
+	OverallStateChanged bool             `json:"overallStateChanged" required:"true"`
+	State               AlertState       `json:"state" required:"true"`
+	StateChanged        bool             `json:"stateChanged" required:"true"`
+	UnixMilli           int64            `json:"unixMilli" required:"true"`
+	Labels              []*qbtypes.Label `json:"labels" required:"true"`
+	Fingerprint         uint64           `json:"fingerprint" required:"true"`
+	Value               float64          `json:"value" required:"true"`
+}
+
+type RuleStateHistoryContributorResponse struct {
```
vikrantgupta25:
```suggestion
type GettableRuleStateHistoryContributor struct {
```

pkg/types/rulestatehistorytypes/response.go
vikrantgupta25: client doesn't care how the rule state has been recorded, what client cares about is that it's getting something based out of rule state. Hence the Gettable type.

pkg/modules/rulestatehistory/implrulestatehistory/handler.go
```diff
@@ -0,0 +1,351 @@
...
+	if err := query.Validate(); err != nil {
+		render.Error(w, errors.WrapInvalidInputf(err, errors.CodeInvalidInput, "invalid query parameters"))
+		return rulestatehistorytypes.Query{}, "", "", 0, false
+	}
+
+	limit := normalizeFilterLimit(int64(raw.Limit))
+	return query, key, strings.TrimSpace(raw.SearchText), limit, true
+}
+
+func parseV2BaseQueryFromURL(r *http.Request) (*ruleHistoryRequest, error) {
```
vikrantgupta25: the handlers are concerned here with the V2HistoryBaseQueryParams as in openapi def, doe we need the conversion to ruleHistoryRequest here with empty cursor ?

pkg/modules/rulestatehistory/implrulestatehistory/handler.go
```diff
@@ -0,0 +1,351 @@
...
+	if limit <= 0 {
+		return 50
+	}
+	if limit > 200 {
+		return 200
+	}
+	return limit
+}
+
+func toQBLabels(raw rulestatehistorytypes.LabelsString) []*qbtypes.Label {
```
vikrantgupta25: the encapsulation of rulestatehistorytypes makes you define such functions there, Trying to understand what is the upside of keeping this private ?

pkg/types/rulestatehistorytypes/response.go
srikanthccv: Please write down the doc first then we come back to discuss this.

pkg/modules/rulestatehistory/implrulestatehistory/handler.go
srikanthccv: The main question is what is the downside?

pkg/modules/rulestatehistory/implrulestatehistory/handler.go
srikanthccv: that's handler internal type struct, please look at the handler API surface not the internal structure.

vikrantgupta25: why is it dealing with an internal struct when it's not even using the cursor out of it ?

pkg/modules/rulestatehistory/implrulestatehistory/handler.go
vikrantgupta25: we cannot have it both ways where we define what goes in types and then have such cases where we question the downside of it.

pkg/modules/rulestatehistory/implrulestatehistory/handler.go
srikanthccv: How does that matter? It's internal for a reason.

pkg/modules/rulestatehistory/implrulestatehistory/handler.go
srikanthccv: I don't get it. What's the point you are trying to make here?

vikrantgupta25: https://github.com/SigNoz/signoz/pull/10446#discussion_r2894986428

srikanthccv: I don't see how that's conflicting. My main point here is the public method, I want to keep it private. When it becomes a public method, the right place for it is the type not the module.

pkg/types/rulestatehistorytypes/response.go
vikrantgupta25: This has been followed even in querier at places like field keys and values. It's practically not feasible to have docs for every other thing. I don't see why we need this discussion here.

pkg/modules/rulestatehistory/implrulestatehistory/handler.go
vikrantgupta25: It matters for the reviewer and the reader why the difference, If there is a genuine reason for the difference please make the case.

pkg/modules/rulestatehistory/implrulestatehistory/handler.go
vikrantgupta25: Why is it private for that matter ?

pkg/types/rulestatehistorytypes/response.go
srikanthccv: It's not followed everywhere, the original change of telemetry field keys didn't have the `Gettable` and `Postable`. There are a lots of places where there is no `Gettable`. I think the change in the PR is completely fine as-is unless otherwise shown how it's a problem or some style guide that says everything should be `Gettable`/`Postable`/`Puttable`. That doc should look at all the place where it's not followed and write why it makes sense to change everything to this nomenclature or write how to objectively decide when to follow.

The lack of doc and discussion is what makes such thing show up in PR reviews. The very existence of this thread is why it requires a doc. I have had enough of this style discussion over the time. Either you write the doc and we discuss or you resolve the comment and accept the change as-is.

pkg/modules/rulestatehistory/implrulestatehistory/handler.go
srikanthccv: This is such a poor nitpicking that it's not a productive discussion. I am not going to change anything here.

pkg/modules/rulestatehistory/implrulestatehistory/handler.go
srikanthccv: author's discretion

pkg/query-service/rules/base_rule.go
```diff
@@ -400,100 +410,59 @@ func (r *BaseRule) ForEachActiveAlert(f func(*ruletypes.Alert)) {
...
-		for _, item := range itemsToAdd {
-			l[item.Fingerprint] = item
-		}
-
-		shouldSkip := map[uint64]bool{}
-
-		for _, item := range lastSavedState {
-			// for the last saved item with fingerprint, check if there is a corresponding entry in the current state
-			currentState, ok := l[item.Fingerprint]
-			if !ok {
```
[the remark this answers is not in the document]
srikanthccv: updated behaviour is correct.

PR #10492, opened 2026-03-04
fix(planned-downtime): notification breaking the page due to invalid description

H4ad:
## Pull Request

This PR fixes the issue on Planned Downtime, any misconfiguration in the screen breaks the entire page.

---

The issue was caused by the ErrorResponse expecting `string` but got `{code:string, message: string}` instead.

Before:

https://github.com/user-attachments/assets/6e5ebc9a-4164-468a-8be3-5daff9600b37

After:

https://github.com/user-attachments/assets/b53cd60f-aa09-4dda-9ed9-d4b57e774d14

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

The issue was passing bad data to useNotification.

Updated types to pass the correct data to useNotification.

---

- Tests added/updated: Added tests for notification & Alerts test manually.

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Fixed the Planned Downtime Page breaking when the information was invalid |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/container/PlannedDowntime/PlannedDowntimeForm.tsx
+++ b/frontend/src/container/PlannedDowntime/PlannedDowntimeForm.tsx
@@ -33,6 +33,8 @@ import { ALL_TIME_ZONES } from 'utils/timeZoneUtil';
 
 import 'dayjs/locale/en';
 
+import { SOMETHING_WENT_WRONG } from '../../constants/api';
+import { showErrorNotification } from '../../utils/error';
 import { AlertRuleTags } from './PlannedDowntimeList';
 import {
 	createEditDowntimeSchedule,
@@ -175,14 +177,14 @@ export function PlannedDowntimeForm(
 				} else {
 					notifications.error({
 						message: 'Error',
-						description: response.error || 'unexpected_error',
+						description:
+							typeof response.error === 'string'
+								? response.error
+								: response.error?.message || SOMETHING_WENT_WRONG,
 					});
 				}
-			} catch (e) {
-				notifications.error({
-					message: 'Error',
-					description: 'unexpected_error',
-				});
+			} catch (e: unknown) {
+				showErrorNotification(notifications, e as Error);
 			}
 			setSaveLoading(false);
 		},
--- a/frontend/src/container/PlannedDowntime/PlannedDowntimeList.tsx
+++ b/frontend/src/container/PlannedDowntime/PlannedDowntimeList.tsx
@@ -25,6 +25,7 @@ import { CalendarClock, PenLine, Trash2 } from 'lucide-react';
 import { useAppContext } from 'providers/App/App';
 import { USER_ROLES } from 'types/roles';
 
+import { showErrorNotification } from '../../utils/error';
 import {
 	formatDateTime,
 	getAlertOptionsFromIds,
@@ -359,7 +360,7 @@ export function PlannedDowntimeList({
 
 	useEffect(() => {
 		if (downtimeSchedules.isError) {
-			notifications.error(downtimeSchedules.error);
+			showErrorNotification(notifications, downtimeSchedules.error);
 		}
 	}, [downtimeSchedules.error, downtimeSchedules.isError, notifications]);
 
--- a/frontend/src/container/PlannedDowntime/PlannedDowntimeutils.ts
+++ b/frontend/src/container/PlannedDowntime/PlannedDowntimeutils.ts
@@ -137,7 +137,10 @@ export const deleteDowntimeHandler = ({
 
 export const createEditDowntimeSchedule = async (
 	props: DowntimeScheduleUpdatePayload,
-): Promise<SuccessResponse<PayloadProps> | ErrorResponse> => {
+): Promise<
+	| SuccessResponse<PayloadProps>
+	| ErrorResponse<{ code: string; message: string } | string>
+> => {
 	if (props.id) {
 		return updateDowntimeSchedule({ ...props });
 	}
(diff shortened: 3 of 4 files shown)


frontend/src/container/PlannedDowntime/PlannedDowntimeForm.tsx
```diff
@@ -175,7 +175,10 @@ export function PlannedDowntimeForm(
 				} else {
 					notifications.error({
 						message: 'Error',
-						description: response.error || 'unexpected_error',
+						description:
+							typeof response.error === 'string'
+								? response.error
+								: response.error?.message || 'unexpected_error',
```
SagarRajput-7: Better to not show `unexpected_error` to the user, See if we can use `showErrorNotification` from `src/components/ExplorerCard/utils.ts`

frontend/src/container/PlannedDowntime/PlannedDowntimeList.tsx
```diff
@@ -359,7 +359,10 @@ export function PlannedDowntimeList({
 
 	useEffect(() => {
 		if (downtimeSchedules.isError) {
-			notifications.error(downtimeSchedules.error);
+			notifications.error({
+				message:
+					downtimeSchedules.error?.message?.toString() || 'Something went wrong',
+			});
 		}
```
SagarRajput-7, line 364: See if we can use `showErrorNotification` from `src/components/ExplorerCard/utils.ts`

frontend/src/hooks/useNotifications.tsx
```diff
@@ -1,30 +1,75 @@
...
+			typeof notificationArgs !== 'object' ||
+			(notificationArgs?.message !== undefined &&
+				typeof notificationArgs.message !== 'string' &&
+				!isValidElement(notificationArgs.message)) ||
+			(notificationArgs?.icon !== undefined &&
+				typeof notificationArgs.icon !== 'string' &&
+				!isValidElement(notificationArgs.icon)) ||
+			(notificationArgs?.description !== undefined &&
+				typeof notificationArgs.description !== 'string' &&
+				!isValidElement(notificationArgs.description))
```
SagarRajput-7: NIT: `null` will satisfy all the cases, so would incorrectly trigger the Sentry capture and show a fallback

frontend/src/hooks/useNotifications.tsx
```diff
@@ -1,30 +1,75 @@
...
+
+			return notificationApi[method]({
+				message: 'Error',
+				description: 'Something went wrong',
+			});
+		}
+
+		return notificationApi[method](notificationArgs);
+	};
+};
```
SagarRajput-7: Think out loud here, I feel guardNotificationPayload is making this little tighter/restrictive which is good in a sense, so considering the input which as per ant will be `message / description / icon: React.ReactNode  `

ReactNode = ReactElement | string | number | boolean | null | undefined | ReactPortal | ReactNodeArray

in the above logic `null`, `number`, `boolean` and array like this `[<span/>, <span/>]` will be skipped and trigger the Sentry.

among all the above i feel `null` could be the only valid case, thinking in case of the valid input to a notification toast. 

for e.g. 

``` 
notifications.error({
      message: 'Upload failed',
      description: uploadError?.message || null,   // null if no message
  });
```

Hence let me know if you feel the same and include null here.

frontend/src/hooks/useNotifications.tsx
```diff
@@ -33,17 +78,21 @@ export function NotificationProvider({
...
 }): JSX.Element {
 	const [notificationApi, NotificationElement] = notification.useNotification();
-	const notifications = useMemo(() => ({ notifications: notificationApi }), [
-		notificationApi,
-	]);
+	const setNotifications = useNotificationStore((s) => s.setNotifications);
+
+	useEffect(() => {
+		setNotifications(notificationApi);
+	}, [notificationApi, setNotifications]);
```
SagarRajput-7: This setup with the `zustland` changes the initialisation flow

How: `notificationApi` from antd is available synchronously at render time inside `NotificationProvider`. But the PR pushes it into the Zustand store via useEffect, which runs after the browser paint. This changes the  flow and might swallow toast in a few cases (mounts) 

this should not effect the codebase in general because of how we use this currently, most of them depends on network responses.

but wanted to discuss if this will be a concern. 

a mock case I reproduced below (video) -

https://github.com/user-attachments/assets/9e0c77b4-ac13-45bb-9f74-317db73bd025

SagarRajput-7: @H4ad  - This is mostly not an issue. So let me know your thoughts around this. 

`Not a blocking comment`

frontend/src/container/PlannedDowntime/PlannedDowntimeForm.tsx
H4ad: showErrorNotification will not identify the object as axios since we transform it in the `createEditDowntime` via `ErrorResponseHandler`.

https://github.com/axios/axios/blob/84285c8f638815460340288dedc71bc06479d023/lib/helpers/isAxiosError.js#L12-L14

But I added this function in the catch path.

frontend/src/hooks/useNotifications.tsx
H4ad: Removed from this PR but I will apply your suggestion, thanks!

H4ad:
After talking with Sagar, make sense to just remove the notification guard and in the future start migrating from antd to sonner.

Issue #6204, opened 2024-10-17
Efficient top list for metrics data source


srikanthccv:
Today, the query fetches all the data, caches it and does the in-process calculation to apply the limit or not. While it is best for many cases, it is not the right tool where there can be several thousand series in the result. Implement better top-list support.

namanv3:
@srikanthccv this issue is about not having order by and limit application in this file right?
https://github.com/SigNoz/signoz/blob/main/pkg/types/querybuildertypes/querybuildertypesv5/series_limit.go
And instead to include it in the clickhouse query built by `statement_builder.go` correct?

srikanthccv:
Yes, that's correct. We don't need to fetch tens of thousands of pods usage into application memory if user is just interested in top 10 pods.

srikanthccv:
Some context why we do this. For metric queries, regardless of the limit, we end up reading all the data, since we do that, we went bit greedy and cache all of that. 1. What do we gain by moving the limit to clickhouse? 2. and what do we lose 3. How will formulas work?

namanv3:
But if the cache fingerprint contains the limit, then how does the cache even help. If the first time with limit 10, the cache has the 1000 rows, and then the user searches for top 20, it won't even hit the cache.

srikanthccv:
You are right, the limit wasn't part of the metrics queries in the initial implementation plan but I see that it has been part of fingerprint as long as the qbv5 existed so the assumed caching benefit never actually helped.

namanv3:
@srikanthccv I have the change ready, but I was wondering if we should instead fix the fingerprint issue and make the cache actually work. If a user has put a limit, they are more likely to change it to other values, and that's where the cached data will really help.

srikanthccv:
> If a user has put a limit, they are more likely to change it to other values, and that's where the cached data will really help.

That was the reason v4/query_range doesn't add the limit to cache fingerprint. However, we missed it for v5, so it's neither helping nor resource efficient.
 https://github.com/SigNoz/signoz/blob/48ccdfcf64b743757ebbb02fc9bb145309beaaf9/pkg/query-service/app/queryBuilder/query_builder.go#L400-L438
[...]

PR #10495, opened 2026-03-05
chore: apply metric limit at query layer

namanv3:
## Pull Request

---

#### Issues closed by this PR

Closes https://github.com/SigNoz/signoz/issues/6204

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---


srikanthccv:
Please add integration tests

namanv3:
@srikanthccv added integration tests

srikanthccv:
Please add tests for order by clause with aggregation value, I don't see any targeted tests for that.

pkg/querier/postprocess.go
srikanthccv, line 172: Should this be removed?

pkg/querybuilder/fallback_expr.go
```diff
@@ -132,6 +132,14 @@ func GroupByKeys(keys []qbtypes.GroupByKey) []string {
 	return k
 }
 
+func OrderByKeys(keys []qbtypes.OrderBy) []string {
```
srikanthccv: I don't see it being used anywhere.

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -555,64 +566,143 @@ func (b *MetricQueryStatementBuilder) BuildFinalSelect(
...
+	} else if hasOrder && hasLimit {
+		labelSelectorSubQueryBuilder := sqlbuilder.NewSelectBuilder()
+		labelSelectorSubQueryBuilder.Select(groupByKeys...)
+		labelSelectorSubQueryBuilder.From(finalCTE)
+		labelSelectorOrderClauses := []string{}
+		orderedKeys := map[string]struct{}{} // this will be used to add the remaining keys as tie breakers in the end
+		for _, o := range query.Order {
+			key := o.Key.Name
+			var clause string
+			if strings.Contains(key, query.Aggregations[0].MetricName) {
```
srikanthccv: Let's use exact match instead of `Contains`. If there is group by key which has metric name as substring, this evals incorrectly (the chances of such scenario is small but never zero, because we don't know how people create their metrics names and attributes in the wild)

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -555,64 +566,143 @@ func (b *MetricQueryStatementBuilder) BuildFinalSelect(
...
+			labelSelectorOrderClauses = append(labelSelectorOrderClauses, clause)
+		}
+		for _, gk := range groupByKeys { // keys that haven't been added via order by keys will be added at the end as tie breakers
+			if _, ok := orderedKeys[gk]; !ok {
+				labelSelectorOrderClauses = append(labelSelectorOrderClauses, fmt.Sprintf("%s ASC", gk))
+			}
+		}
+		labelSelectorSubQueryBuilder.GroupBy(groupByKeys...)
+		labelSelectorSubQueryBuilder.OrderBy(labelSelectorOrderClauses...)
+		labelSelectorSubQuery, _ := labelSelectorSubQueryBuilder.BuildWithFlavor(sqlbuilder.ClickHouse)
```
srikanthccv: potential error silently discarded here

pkg/telemetrymetrics/statement_builder.go
srikanthccv, line 1: We need to check the overall performance, if anything, it should be better given the work is offloaded to ClickHouse servers.

tests/integration/src/querier/08_metrics_histogram.py
```diff
@@ -521,4 +526,568 @@ def test_histogram_percentile_for_delta_service(
...
+
+    response = make_query_request(signoz, token, start_ms, end_ms, [query_p75])
+    assert response.status_code == HTTPStatus.OK
+
+    data = response.json()
+    p75_series = get_all_series(data, "A")
+
+    for series in p75_series:
+        endpoint = series.get("labels", [{}])[0].get("value", "unknown")
+        values = series.get("values", [])
```
srikanthccv: Where is this data used?

tests/integration/src/querier/08_metrics_histogram.py
```diff
@@ -521,4 +526,568 @@ def test_histogram_percentile_for_delta_service(
...
+    # /health (cumulative, service=api)
+    if "/health" in p75_values:
+        vals = p75_values["/health"]
+        assert vals[0]["value"] == 6000, f"Expected /health p75 first=8000, got {vals[0]['value']}"
+        assert vals[-1]["value"] == 6000, f"Expected /health p75 last=991.304, got {vals[-1]['value']}"
+
+    # /orders (cumulative, service=api): same distribution as /health
+    if "/orders" in p75_values:
+        vals = p75_values["/orders"]
+        assert vals[0]["value"] == 4500, f"Expected /orders p75 first=6400, got {vals[0]['value']}"
```
srikanthccv: Number of assertions here have the incorrect values where the asserted value doesn't match the value in assertion failure comment.

pkg/querier/postprocess.go
namanv3: `postProcessBuilderQuery` is called for logs and traces, while `postProcessMetricQuery` is called for metrics. Because this change is only for metrics, I've only removed the apply series limit from the metrics function.
```
for _, query := range req.CompositeQuery.Queries {
		switch spec := query.Spec.(type) {
		case qbtypes.QueryBuilderQuery[qbtypes.TraceAggregation]:
			if result, ok := typedResults[spec.Name]; ok {
				result = postProcessBuilderQuery(q, result, spec, req)
				typedResults[spec.Name] = result
			}
		case qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]:
			if result, ok := typedResults[spec.Name]; ok {
				result = postProcessBuilderQuery(q, result, spec, req)
				typedResults[spec.Name] = result
			}
			if result, ok := typedResults[spec.Name]; ok {
				result = postProcessMetricQuery(q, result, spec, req)
				typedResults[spec.Name] = result
			}
		case qbtypes.QueryBuilderTraceOperator:
			if result, ok := typedResults[spec.Name]; ok {
				result = postProcessTraceOperator(q, result, spec, req)
				typedResults[spec.Name] = result
			}
		}
	}
```

pkg/querybuilder/fallback_expr.go
namanv3: Removed, was using it earlier but usage was removed

pkg/telemetrymetrics/statement_builder.go
namanv3: Should I get my own staging env to do this? How do I do that?

pkg/telemetrymetrics/statement_builder.go
namanv3: its not an error, the second return parameter are the args for the query. Cuz this subquery does not have args, I've ignored the return param

tests/integration/src/querier/08_metrics_histogram.py
namanv3: Fixed

tests/integration/src/querier/08_metrics_histogram.py
namanv3: Nowhere, removed

pkg/telemetrymetrics/statement_builder.go
namanv3: Done

namanv3:
@srikanthccv 
I've pushed a new query structure. This one adds a column for the avg value for each row's group, and then another column for the rank of the group based on the expected order (handles the no order by case as well). If limits are there, the group rank is used as a cut off.

It reads the same number of rows irrespective of the scenario. The query explanation also has `ReadFromRemote (Read from remote replica)` only once.

The time taken by the query on the testing account's clickhouse is similar to the time taken by the API on cloud currently, and sometimes equal to the clickhouse time reported by the API as well.

The unit tests will fail for this change, I'll fix them once you've reviewed the PR.

pkg/querier/postprocess.go
srikanthccv: It should not be removed and should indeed be added to metrics as well. It's the layer for working with cached results. Please test against the ephemeral workloads such as pods to understand what I am referring to better.

namanv3:
Closing this PR because of an edge case that's described here: https://www.notion.so/signoz/Limit-in-CH-breaks-because-of-cache-32dfcc6bcd1980a99b37f1cbe3a4b9c2?source=copy_link

PR #10497, opened 2026-03-05
Fix - Handling for resource. prefix in quick filters

ishan621:
## Pull Request

Quick Filters weren’t matching filters when the query used keys like resource.service.name or span.name, even though the config uses service.name.
Root Cause
Filter keys in the query builder can come in as:
service.name (no prefix)
resource.service.name (resource prefix)
Quick Filters were comparing keys literally, so resource.service.name did not match service.name in the config.
One place to fix it – Only isKeyMatch needs to know about prefixes, not every caller.
Handles all variants – Works for service.name, resource.service.name
	'metric',
	'log',
	'span',
	'trace',
	'resource',
	'scope',
	'attribute',
	'event',
	'body',
No config changes – Quick Filters config stays with unprefixed keys.
Matches existing semantics – Treats resource. as scoping prefixes, not part of the attribute name.
Backward compatible – Plain keys like service.name keep working as before.

Before: 
Screenshot 2026-03-05 at 2 44 43 PM

After: 
Screenshot 2026-03-05 at 2 45 18 PM

#### Issues closed by this PR
https://github.com/SigNoz/engineering-pod/issues/3623

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context
Faulty assumption

Root cause: Quick Filters compared keys literally. Query keys like resource.service.name did not match config keys like service.name.
Fix: Normalize both sides before comparing by stripping  resource. prefixes. That makes resource.service.name and service.name compare equal, so the Checkbox can correctly show and update the filter state.
---

- Blast radius: all users
- Potential regressions: 
- Rollback plan: commit rollback

---

### 📋 Checklist
- [x] Manually tested

---

### DIFF
--- a/frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.tsx
+++ b/frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.tsx
@@ -26,6 +26,7 @@ import { DataSource } from 'types/common/queryBuilder';
 import { v4 as uuid } from 'uuid';
 
 import LogsQuickFilterEmptyState from './LogsQuickFilterEmptyState';
+import { isKeyMatch } from './utils';
 
 import './Checkbox.styles.scss';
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.tsx
```diff
@@ -34,6 +34,48 @@ const NON_SELECTED_OPERATORS = [OPERATORS['!='], 'not in'];
 
 const SOURCES_WITH_EMPTY_STATE_ENABLED = [QuickFiltersSource.LOGS_EXPLORER];
 
+/**
```
aks07: Move this logic to utils.

frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.tsx
```diff
@@ -34,6 +34,48 @@ const NON_SELECTED_OPERATORS = [OPERATORS['!='], 'not in'];
...
+ * These prefixes are added to attribute keys based on their context.
+ */
+const FIELD_CONTEXT_PREFIXES = [
```
aks07: I am assuming this list has been reviewed. cc. @nityanandagohain

ishan621: Yes , we got this from pkg/types/telemetrytypes/field_context.go , @nityanandagohain  has confirmed this

frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.tsx
ishan621: Done

PR #10506, opened 2026-03-05
fix: revert quick filters

ishan621:
## Pull Request
This PR is to revert quick filter changes to old commits. 
https://github.com/SigNoz/signoz/pull/10110 - PR 
https://github.com/SigNoz/signoz/issues/8931 - Ticket

Issue - We saw in production that some of the filters are not working as expected then we came to know that backend is not deployed for every page , so we need to revert the changes now as they were common for all pages.
---
To revert to old commits - https://github.com/SigNoz/signoz/pull/10110 
Code revert is the only option

Before : 

Screenshot 2026-03-05 at 11 38 12 PM

After: 

Screenshot 2026-03-05 at 11 38 25 PM

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

Faulty assumption

Issue: Blank/empty values appearing in quick filters on Traces pages after PR #10110 was merged.
https://github.com/SigNoz/signoz/pull/10110 - PR 
https://github.com/SigNoz/signoz/issues/8931 - Ticket

Root Cause:
PR #10110 migrated all quick filters across the application to use the new /fields/keys and /fields/values API endpoints (adaptive filters experience). However, this migration was implemented for all pages simultaneously (Logs, Traces, Metrics, Infrastructure Monitoring, etc.), when the backend's new API was only fully ready and tested for Logs.
The new API returned incomplete or empty data for Traces signals, causing blank filter values to appear in the UI.

Why Full Revert Instead of Targeted Fix:
The implementation in PR #10110 was tightly coupled across components:
Checkbox.tsx (filter renderer) was modified to use useGetQueryKeyValueSuggestions hook
QuerySearch.tsx (autocomplete) shared the same new API integration
Mock data and tests were updated to the new response format
Due to this coupling, isolating the fix for Traces while keeping the new behaviour for Logs would have required significant refactoring. A full revert was the fastest path to restore functionality.

**_Note: In the original ticket it was mentioned to migrate quick filters for all respective pages._**

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered: Yes

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/mocks-server/__mockdata__/customQuickFilters.ts
+++ b/frontend/src/mocks-server/__mockdata__/customQuickFilters.ts
@@ -118,15 +118,13 @@ export const otherFiltersResponse = {
 export const quickFiltersAttributeValuesResponse = {
 	status: 'success',
 	data: {
-		data: {
-			stringAttributeValues: [
-				'mq-kafka',
-				'otel-demo',
-				'otlp-python',
-				'sample-flask',
-			],
-			numberAttributeValues: null,
-			boolAttributeValues: null,
-		},
+		stringAttributeValues: [
+			'mq-kafka',
+			'otel-demo',
+			'otlp-python',
+			'sample-flask',
+		],
+		numberAttributeValues: null,
+		boolAttributeValues: null,
 	},
 };


frontend/src/components/QuickFilters/FilterRenderers/Checkbox/Checkbox.styles.scss
```diff
@@ -144,92 +144,92 @@
...
-			&__title {
-				color: var(--bg-amber-800);
-			}
-
-			&__subtitle {
-				color: var(--bg-amber-800);
-			}
-		}
-	}
+	// .search-prompt {
```
aks07, line 147: lets remove these comments.

PR #10509, opened 2026-03-05
fix: add destroy hook to prevent memory leak in anomaly tooltip plugin

amareshsm:
## Pull Request

---

Fix memory leak in AnomalyAlertEvaluationView tooltip plugin by adding proper cleanup.

Problem: The tooltip plugin was missing a destroy hook, causing event listeners to accumulate each time the chart re-renders (theme changes, timezone changes, or component updates). This led to tooltip flicker, stale data display, and growing memory usage.

Solution: Added a destroy hook that properly removes all event listeners and cleans up the tooltip DOM element when the uPlot instance is destroyed.

Changes made [Address as part of pr review comments]

1. import uPlot from 'uplot' — added, consistent with every other uPlot file in the codebase.

2. u → uPlotInstance — renamed in init, updateTooltip, and the return type annotation.

3. any eliminated everywhere:

  | Before | After |
  |--------|-------|
  | `init: (u: any) => void` in return type | `init: (uPlotInstance: uPlot) => void` |
  | `let plotInstance: any` | `let plotInstance: uPlot \| null = null` |
  | `function updateTooltip(u: any, ...)` | `function updateTooltip(uPlotInstance: uPlot, ...)` |
  | `function init(u: any)` | `function init(uPlotInstance: uPlot)` |

4.  series.label is string | undefined in uPlot types → extracted to const seriesLabel = series.label ?? ''
data[i][idx] is number | null | undefined → yVal ?? 'N/A' fallback before passing to formatValue

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

Missing destroy hook in tooltip plugin caused event listeners to accumulate without proper cleanup when the chart re-renders due to prop changes (isDarkMode, timezone).

- Extract anonymous event handlers to named functions for proper removal
- Add destroy hook that removes all event listeners and tooltip DOM element
- Store element reference for cleanup access
---

Manual verification

- Tests added/updated:
- Manual verification:
 - Verified tooltip functionality remains intact after theme/timezone changes
 - Confirmed no event listener accumulation in DevTools
 - Tested tooltip shows correct data (main series, upper/lower bands)
 
- Edge cases covered: Multiple theme toggles, timezone changes, component remounts

---

- Blast radius:  only affects anomaly alert chart tooltips
- Potential regressions: follows same pattern as main tooltipPlugin
- Rollback plan: revert of changes if issues arise

---

| Field | Value |
|------|-------|
| Change Type |  Bug Fix  |
| Description | Fixed memory leak in anomaly alert chart tooltips |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> **Low Risk**
> Small, localized UI change that only affects tooltip lifecycle/cleanup; low chance of regression beyond tooltip behavior.
> 
> **Overview**
> Prevents tooltip-related leaks/flicker in `AnomalyAlertEvaluationView` by extending the uPlot `tooltipPlugin` to support teardown.
> 
> The plugin now uses named mouse event handlers, stores references to `u.over` and the plot instance, and implements a `destroy` hook that removes listeners and the tooltip element when the chart is destroyed/recreated (e.g., on theme/timezone changes).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bc231227eb8f08b66a36f5142a5d26c598814efc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


frontend/src/container/AnomalyAlertEvaluationView/tooltipPlugin.ts
```diff
@@ -6,11 +6,13 @@ import { generateColor } from 'lib/uPlotLib/utils/generateColor';
 const tooltipPlugin = (
 	isDarkMode: boolean,
 	timezone: string,
-): { hooks: { init: (u: any) => void } } => {
+): { hooks: { init: (u: any) => void; destroy: () => void } } => {
```
AshwinBhatkal: since we are already touching code here, can we name the variables better across the file? At least the current usages. u can be made uPlotInstance for example

Also, avoid the type any and use the uPlot type here

AshwinBhatkal: Overall, the change in this PR looks good to me

amareshsm: done. Updated pr description with also.

PR #10530, opened 2026-03-09
feat: update url with y-axis unit

YounixM:
This change adds a dedicated `yAxisUnit` URL query param and wires Logs, Meter, and Traces explorers to read/write the y-axis unit from/to the URL instead of keeping it only in local state or composite query. This ensures the selected unit is preserved across reloads, browser navigation, and shared links, and unifies behavior across explorers. Using a single URL param is the right approach because it makes the unit selection explicit, shareable, and easier to reason about than implicit local state.

---

- Before: y-axis unit selection not preserved across reload/share in Logs/Meter/Traces explorers, inconsistent defaults.
- After: y-axis unit selection is synced via URL, persists across reload/back/forward, and is restored when opening shared links.

---

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/3673

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context
N/A

---

- **Tests added/updated**:  
  - (Planned/Optional) Add coverage around `yAxisUnit` initialization and updates in the relevant containers to ensure URL ↔ state sync remains stable.
- **Manual verification**:
  - Logs Explorer:
    - Select a non-default y-axis unit, reload the page, and verify the unit is preserved.
    - Use browser back/forward after changing the unit and verify it follows the URL.
    - Copy the URL and open it in a new tab; verify the selected unit matches the URL.
  - Meter Explorer:
    - Repeat the above flows and ensure unit persistence and correctness of defaults.
  - Traces Explorer:
    - Verify default unit still respects `ms` vs `short` logic when no `yAxisUnit` is present.
    - Verify `yAxisUnit` from URL overrides the computed default.
- **Edge cases covered**:
  - No `yAxisUnit` in URL → falls back to existing defaults (e.g., `ms` vs `short` in Traces).
  - Clearing the unit selection removes the `yAxisUnit` param from the URL.
  - Directly editing/removing `yAxisUnit` in the address bar updates the UI accordingly.

---

- **Blast radius**:
  - Limited to time series views in:
    - Logs Explorer (`LogsExplorerViewsContainer`)
    - Meter Explorer (`TimeSeries` for meters)
    - Traces Explorer (`TimeSeriesViewContainer`)
- **Potential regressions**:
  - Incorrect default units if URL/query param handling is wrong.
  - Y-axis unit getting out of sync if URL and local state diverge.
  - Unexpected URL churn if `URLSearchParams` handling is incorrect.
- **Rollback plan**:
  - Revert this PR to restore prior behavior (no `yAxisUnit` URL param and local-only state).
  - If needed, follow up with a smaller change that re-introduces the param behind a feature flag.

---

N/A

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- Focus on the URL ↔ state sync logic for `yAxisUnit` in all three explorers (`LogsExplorerViewsContainer`, `MeterExplorer` `TimeSeries`, and Traces `TimeSeriesViewContainer`).  
- Pay particular attention to:
  - Defaulting behavior when `yAxisUnit` is absent (especially in Traces where `ms` vs `short` is derived).
  - The `URLSearchParams` manipulation and `useSafeNavigate` usage to ensure we don’t introduce unnecessary navigation or history spam.

### DIFF
--- a/frontend/src/tests/test-utils.tsx
+++ b/frontend/src/tests/test-utils.tsx
@@ -7,6 +7,7 @@ import { render, RenderOptions, RenderResult } from '@testing-library/react';
 import { FeatureKeys } from 'constants/features';
 import { ORG_PREFERENCES } from 'constants/orgPreferences';
 import { ResourceProvider } from 'hooks/useResourceAttribute';
+import { NuqsAdapter } from 'nuqs/adapters/react';
 import { AppContext } from 'providers/App/App';
 import { IAppContext } from 'providers/App/types';
 import { ErrorModalProvider } from 'providers/ErrorModalProvider';
@@ -280,23 +281,25 @@ export function AllTheProviders({
 
 	return (
 		<MemoryRouter initialEntries={[initialRouteValue]}>
-			<QueryClientProvider client={queryClient}>
-				<Provider store={mockStored(roleValue)}>
-					<AppContext.Provider
-						value={getAppContextMock(roleValue, appContextOverridesValue)}
-					>
-						<ResourceProvider>
-							<ErrorModalProvider>
-								<TimezoneProvider>
-									<PreferenceContextProvider>
-										{queryBuilderContent}
-									</PreferenceContextProvider>
-								</TimezoneProvider>
-							</ErrorModalProvider>
-						</ResourceProvider>
-					</AppContext.Provider>
-				</Provider>
-			</QueryClientProvider>
+			<NuqsAdapter>
+				<QueryClientProvider client={queryClient}>
+					<Provider store={mockStored(roleValue)}>
+						<AppContext.Provider
+							value={getAppContextMock(roleValue, appContextOverridesValue)}
+						>
+							<ResourceProvider>
+								<ErrorModalProvider>
+									<TimezoneProvider>
+										<PreferenceContextProvider>
+											{queryBuilderContent}
+										</PreferenceContextProvider>
+									</TimezoneProvider>
+								</ErrorModalProvider>
+							</ResourceProvider>
+						</AppContext.Provider>
+					</Provider>
+				</QueryClientProvider>
+			</NuqsAdapter>
 		</MemoryRouter>
 	);
 }


frontend/src/pages/TracesExplorer/TimeSeriesView/index.tsx
```diff
@@ -52,13 +60,37 @@ function TimeSeriesViewContainer({
...
-	);
+	// yAxisUnit from URL query param only (no compositeQuery.unit)
+	const defaultUnit = isValidToConvertToMs ? 'ms' : 'short';
+	const yAxisUnitFromUrl = urlQuery.get(QueryParams.yAxisUnit) ?? defaultUnit;
+	const [yAxisUnit, setYAxisUnit] = useState<string>(yAxisUnitFromUrl);
 
-	const onUnitChangeHandler = (value: string): void => {
-		setYAxisUnit(value);
-	};
+	// Sync yAxisUnit when URL param changes (e.g. browser back or shared link)
```
ahrefabhi: can we move this entire logic in a hook, and call in each explorer?

frontend/src/pages/TracesExplorer/TimeSeriesView/index.tsx
```diff
@@ -52,13 +60,37 @@ function TimeSeriesViewContainer({
...
+			} else {
+				params.delete(QueryParams.yAxisUnit);
+			}
+			safeNavigate({
+				pathname: location.pathname,
+				search: `?${params.toString()}`,
+			});
+		},
+		[location.pathname, safeNavigate, urlQuery],
+	);
```
H4ad: nuqs?

YounixM: We can handle it separately @H4ad . Want to validate the setup correctly and give it some time on staging.

H4ad: Hmm... I think this PR could be much shorter with Nuqs, since all the hook code is only used to parse/handle the logic for updating/reading the query parameters.

What better opportunity to test Nuqs than this?

frontend/src/hooks/useUrlYAxisUnit.ts
```diff
@@ -0,0 +1,59 @@
...
+ * - Initializes from `QueryParams.yAxisUnit` or a provided default
+ * - Keeps local state in sync when the URL changes (e.g. back/forward, shared links)
+ * - Writes updates back to the URL while preserving existing query params
+ */
+function useUrlYAxisUnit(defaultUnit = ''): UseUrlYAxisUnitResult {
+	const urlQuery = useUrlQuery();
+	const location = useLocation();
+	const { safeNavigate } = useSafeNavigate();
+
+	const initialUnit = urlQuery.get(QueryParams.yAxisUnit) ?? defaultUnit;
```
aks07: I think the local yAxisUnit state might not be necessary here. Since the value is already derived from the URL -> `urlQuery.get(QueryParams.yAxisUnit)`, it should reflect the latest state whenever the router triggers a re-render (via useLocation). Maintaining a separate React state plus a syncing useEffect duplicates the source of truth and adds extra complexity. We could potentially simplify this by deriving yAxisUnit directly from the URL query param and only updating the URL in onUnitChange.

Current Flow: URL → useEffect → state → UI
Suggested Flow:  URL → UI

aks07:
Shouldn't we add test case as well?

YounixM:
Shouldn't we add test case as well?

- This is a temporary fix. We need to fix it correctly across the product.

Details added in the issue

H4ad (approved):

Issue #10531, opened 2026-03-09
Subsequent calls after rotate don't work for APIs generated from openAPI


AshwinBhatkal:
Subsequent calls after rotate don't work for APIs generated from openAPI

https://github.com/user-attachments/assets/cf0e9104-3e81-47f8-938b-d7cf346327cb

PR #10532, opened 2026-03-09
fix(frontend): refresh generated api on 401 errors

AshwinBhatkal:
When the auth token expires and calls are made to any API, the expectation is that it first fails and then a new call is made with the rotated token. This was not happening for APIs generated from openAPI. This PR fixes that

##### Before

https://github.com/user-attachments/assets/73706bd1-93e9-447c-a338-86608a8eb65c

##### After (Page using generated API)

https://github.com/user-attachments/assets/4b8007a2-d8cf-4217-9f8d-277343d1ea3b

##### After (Page not using generated API)

https://github.com/user-attachments/assets/58fec8e1-7c58-4d3d-beda-6fc9a6f9eef9

##### After (Page not using generated API but with POST)

https://github.com/user-attachments/assets/fd79ea75-ae5b-4b9b-ae4a-cbde22152770

Closes https://github.com/SigNoz/signoz/issues/10531

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

### DIFF
--- a/frontend/src/api/index.ts
+++ b/frontend/src/api/index.ts
@@ -94,19 +94,13 @@ export const interceptorRejected = async (
 					afterLogin(response.data.accessToken, response.data.refreshToken, true);
 
 					try {
-						const reResponse = await axios(
-							`${value.config.baseURL}${value.config.url?.substring(1)}`,
-							{
-								method: value.config.method,
-								headers: {
-									...value.config.headers,
-									Authorization: `Bearer ${response.data.accessToken}`,
-								},
-								data: {
-									...JSON.parse(value.config.data || '{}'),
-								},
+						const reResponse = await axios({
+							...value.config,
+							headers: {
+								...value.config.headers,
+								Authorization: `Bearer ${response.data.accessToken}`,
 							},
-						);
+						});
 
 						return await Promise.resolve(reResponse);
 					} catch (error) {


frontend/src/api/index.ts
```diff
@@ -94,19 +94,13 @@ export const interceptorRejected = async (
...
-									...JSON.parse(value.config.data || '{}'),
-								},
+						const reResponse = await axios({
+							...value.config,
+							headers: {
+								...value.config.headers,
+								Authorization: `Bearer ${response.data.accessToken}`,
 							},
-						);
+						});
```
H4ad, line 103: Did you tested any place that sends the body? I had to do a similar fix at [10486](https://github.com/SigNoz/signoz/pull/10486/changes#diff-78e67ba0fa050b9199f57338094e632691b99938951765f142ad749f5ef59675R106), but the issue was caused by the JSON.parse spread. I was expecting the axios only accepting JS Object as data instead of string from original configuration.

AshwinBhatkal: @H4ad added screen recording for the same

H4ad (approved):

PR #10545, opened 2026-03-10
feat: message as type hint in body_v2

piyushsingariya:
## Pull Request

---

### 📄 Summary

This PR aims to add message as type hint in body_v2 JSON column. Since we're streamlining Inserts into `message` field to be `String` always [here](https://github.com/SigNoz/signoz-otel-collector/pull/768/changes#diff-2e4430eb3ea3772d2496ab90bda3b4aed09f0861f3c39485c1be4caa1053d245) hence we're upgrading `message` as type hint. This will also help in compression of text based logs specifically.

Closes signoz/engineering-pod#4181

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/pkg/telemetrylogs/condition_builder.go
+++ b/pkg/telemetrylogs/condition_builder.go
@@ -3,14 +3,12 @@ package telemetrylogs
 import (
 	"context"
 	"fmt"
-	"slices"
 
 	schema "github.com/SigNoz/signoz-otel-collector/cmd/signozschemamigrator/schema_migrator"
 	"github.com/SigNoz/signoz/pkg/errors"
 	"github.com/SigNoz/signoz/pkg/querybuilder"
 	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
 	"github.com/SigNoz/signoz/pkg/types/telemetrytypes"
-	"golang.org/x/exp/maps"
 
 	"github.com/huandu/go-sqlbuilder"
 )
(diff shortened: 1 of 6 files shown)


srikanthccv:
@piyushsingariya, is this already hint on the schema level? I was going to suggest that this is a hint on the promoted column at the schema level and the `fieldmapper` always returns the promoted message when working with `body`.

piyushsingariya:
> @piyushsingariya, is this already hint on the schema level? I was going to suggest that this is a hint on the promoted column at the schema level and the `fieldmapper` always returns the promoted message when working with `body`.

Yes this is hint at schema level, I discussed with @nityanandagohain and figured, Earlier, when we had three columns, having a default `message` value created ambiguity because we couldn’t reliably tell whether the message actually existed in the payload or not.

Now that we’re transforming every text log into `message`, it makes more sense for it to act as the type hint. In this model, a missing message effectively becomes equivalent to an empty string.

There is an open question: if we should make it Nullable(string) so message nullability is still accounted.

piyushsingariya:
>There is an open question: if we should make it Nullable(string) so message nullability is still accounted.

No nullable, discussed with @srikanthccv

PR #10551, opened 2026-03-10
chore: remove usage of updatedTimeRef in dashboard provider

AshwinBhatkal:
Removes updatedTimeRef from global store. Relies on dashboard data itself to find the differences

### DIFF
--- a/frontend/src/pages/DashboardPage/index.tsx
+++ b/frontend/src/pages/DashboardPage/index.tsx
@@ -1,194 +1,37 @@
-import { useEffect, useRef } from 'react';
-import { useTranslation } from 'react-i18next';
-import { useQuery } from 'react-query';
-// eslint-disable-next-line no-restricted-imports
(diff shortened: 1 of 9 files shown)


AshwinBhatkal:
This change is on hold till removal of provider from Dashboard widget

frontend/src/pages/DashboardPage/index.tsx
```diff
@@ -178,61 +154,31 @@ function DashboardPage(): JSX.Element {
...
 							);
-
 							setPanelMap(defaultTo(updatedDashboardData?.data.panelMap, {}));
 						},
 					});
-
 					modalRef.current = modal;
-				} else {
-					// normal flow
-					updatedTimeRef.current = dayjs(updatedDashboardData?.updatedAt);
```
AshwinBhatkal, line 195: This was redundant. We were setting/updating dashboard even when data was the same

PR #10555, opened 2026-03-11
test: integration tests for percentile aggregation

namanv3:
## Pull Request

---

### 📄 Summary
This makes the custom function `histogramQuantile` used in calculating percentile spatial aggregations for histograms work for integration tests env and also for dev env.
The new integration tests added make use of this to finally be able to properly test this functionality.

#### Issues closed by this PR

Closes https://github.com/SigNoz/engineering-pod/issues/3927

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [x] 🧪 Test-only

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

### DIFF
--- a/tests/integration/fixtures/clickhouse.py
+++ b/tests/integration/fixtures/clickhouse.py
@@ -183,9 +183,8 @@ def clickhouse(
             ],
         )
         if exit_code != 0:
-            logger.warning(
-                "Failed to install histogramQuantile binary: %s",
-                output.decode(),
+            raise RuntimeError(
+                f"Failed to install histogramQuantile binary: {output.decode()}"
             )
 
         connection = clickhouse_connect.get_client(


.devenv/docker/clickhouse/fs/etc/clickhouse-server/config.d/config.xml
```diff
@@ -44,4 +44,5 @@
         <shard>01</shard>
         <replica>01</replica>
     </macros>
+    <user_defined_executable_functions_config>*function.xml</user_defined_executable_functions_config>
```
vikrantgupta25, line 47: NIT: IMO can we specify `user_scripts_path` even if we are using the default value ? helps in more clarity that how the paths are mapped ( avoids the magic behind the scenes )

tests/integration/fixtures/clickhouse.py
```diff
@@ -117,17 +119,74 @@ def create() -> types.TestContainerClickhouse:
...
+        container.with_volume_mapping(
+            custom_function_file_path,
+            "/etc/clickhouse-server/custom-function.xml",
+        )
         container.with_network(network)
         container.start()
 
+        # Download and install the histogramQuantile binary
+        wrapped = container.get_wrapped_container()
+        exit_code, output = wrapped.exec_run(
```
vikrantgupta25, line 168: clarifying that is it not necessary to run this before the clickhouse server starts ? Or it's just necessary to run it before the query ?

namanv3: as such it is only necessary to have it before the query runs, and in integration tests its not a problem to start clickhouse before installing cuz the tests will run once the whole setup is ready
(but for devenv the queries can run right after ch is up so better to have a separate init there)

.devenv/docker/clickhouse/fs/etc/clickhouse-server/config.d/config.xml
namanv3: Done

PR #10558, opened 2026-03-11
feat: removed plan name and added copiable license info in custom domain card

SagarRajput-7:
## Pull Request

Earlier as per design ([here](https://www.figma.com/design/QPOedFb96RtjeWlfWk3K9k?node-id=3284-22380&m=dev#1647617321)) we have added the Plan info in the custom domain but this caused confusion among the user - ticket - https://github.com/SigNoz/platform-pod/issues/1826

- So now as per the ticket requirement - 
   - Removed the plan name info section
   - Fixed the empty space when custom domain is not set
   - added license info in the footer instead, with copiable license key
   
- Also, added test case for the new license component

Conditions and Permission:
- Custom domain card is allowed to admins with either the cloud or the enterprise plan.
- License Info (refering from the account setting page where we also the similar license details) - is available to everyone, regardless of role and deployment type.

```
Note: Regarding E2E, the flow here, around the custom domain uses API and is different 
and can only be triggered once in 24h, hence we couldn't right E2E here, and without this
rest of the parts are UI elements and component which are covered in JEST
```

Design - https://www.figma.com/design/QPOedFb96RtjeWlfWk3K9k/Settings-Revamp?node-id=3405-600&m=dev

---

#### Before - 
--------------

image

#### After - 
-------

image

##### Non-admin user
----------------------------------

image

##### Invalid License key 
--------------------------

image

- Closes - https://github.com/SigNoz/platform-pod/issues/1826

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated: added test cases
- Manual verification: tested manually
- Edge cases covered: Yes

---

- Blast radius: Workspace settings
- Potential regressions: NA
- Rollback plan: NA

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/frontend/src/container/MySettings/__tests__/MySettings.test.tsx
+++ b/frontend/src/container/MySettings/__tests__/MySettings.test.tsx
@@ -271,7 +271,7 @@ describe('MySettings Flows', () => {
 				},
 			});
 
-			expect(within(container).getByText('ab********cd')).toBeInTheDocument();
+			expect(within(container).getByText('ab·······cd')).toBeInTheDocument();
 		});
 
 		it('Should not mask license key if it is too short', () => {


frontend/src/container/GeneralSettings/GeneralSettings.tsx
```diff
@@ -680,7 +681,13 @@ function GeneralSettings({
...
 			</div>
 
-			{showCustomDomainSettings && <CustomDomainSettings />}
+			<div className="custom-domain-card">
+				{showCustomDomainSettings && <CustomDomainSettings />}
+				{showCustomDomainSettings && activeLicense?.key && (
+					<div className="custom-domain-card-divider" />
+				)}
+				{activeLicense?.key && <LicenseKeyRow />}
+			</div>
```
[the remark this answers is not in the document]
SagarRajput-7: done

frontend/src/container/GeneralSettings/LicenseKeyRow/LicenseKeyRow.tsx
```diff
@@ -0,0 +1,58 @@
...
+		return `${key.substring(0, 2)}·······${key.slice(-2).trim()}`;
+	};
+
+	const handleCopyLicenseKey = (text: string): void => {
+		copyToClipboard(text);
+		if (!navigator.clipboard) {
+			toast.error('Failed to copy license key.', { richColors: true });
+			return;
+		}
+		toast.success('License key copied to clipboard.', { richColors: true });
```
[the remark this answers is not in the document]
SagarRajput-7: No, 
The copyToClipboard function returned by react-use's hook intentionally swallows all errors and does not return a Promise or a boolean. Its execution is strictly synchronous and it returns void (undefined).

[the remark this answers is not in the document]
SagarRajput-7: same as above

frontend/src/AppRoutes/Private.tsx
```diff
@@ -128,6 +128,7 @@ function PrivateRoute({ children }: PrivateRouteProps): JSX.Element {
 			isAdmin &&
 			(path === ROUTES.SETTINGS ||
 				path === ROUTES.ORG_SETTINGS ||
+				path === ROUTES.MEMBERS_SETTINGS ||
 				path === ROUTES.BILLING ||
 				path === ROUTES.MY_SETTINGS);
```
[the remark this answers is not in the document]
SagarRajput-7: older behaviour, not checking in this one

frontend/src/container/CustomDomainSettings/CustomDomainSettings.tsx
```diff
@@ -175,7 +176,8 @@ export default function CustomDomainSettings(): JSX.Element {
 		[hosts, activeHost],
 	);
 
-	const planName = activeLicense?.plan?.name;
+	const workspaceName =
+		org?.[0]?.displayName || customDomainSubdomain || activeHost?.name;
```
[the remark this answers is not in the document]
SagarRajput-7: not required

frontend/src/container/CustomDomainSettings/CustomDomainSettings.tsx
```diff
@@ -191,106 +193,94 @@ export default function CustomDomainSettings(): JSX.Element {
...
 
+			{isPollingEnabled && (
+				<Callout
+					type="info"
+					showIcon
+					className="custom-domain-callout"
+					size="small"
+					icon={<SolidAlertCircle size={13} color="primary" />}
+					message={`Updating your URL to ⎯ ${customDomainSubdomain}.${dnsSuffix}. This may take a few mins.`}
+				/>
```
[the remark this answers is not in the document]
SagarRajput-7: already covered via - `isPollingEnabled`

frontend/src/container/GeneralSettings/LicenseKeyRow/LicenseKeyRow.tsx
```diff
@@ -0,0 +1,54 @@
+import { useCopyToClipboard } from 'react-use';
+import { Button } from '@signozhq/button';
+import { Copy, KeyRound } from '@signozhq/icons';
+import { toast } from '@signozhq/sonner';
+import { useAppContext } from 'providers/App/App';
+
+import './LicenseKeyRow.styles.scss';
+
+function LicenseKeyRow(): JSX.Element | null {
+	const { activeLicense } = useAppContext();
```
SagarRajput-7, line 11: this is referred from already existing - https://github.com/SigNoz/signoz/blob/557451ed815b2d2fe3d1a42721986d41d29e7437/frontend/src/container/MySettings/LicenseSection/LicenseSection.tsx#L10

frontend/src/pages/Settings/utils.ts
```diff
@@ -36,6 +36,7 @@ export const getRoutes = (
 	if (isWorkspaceBlocked && isAdmin) {
 		settings.push(
 			...organizationSettings(t),
+			...membersSettings(t),
```
SagarRajput-7, line 39: these are some places missing for membersetting that i found

frontend/src/AppRoutes/Private.tsx
```diff
@@ -128,6 +128,7 @@ function PrivateRoute({ children }: PrivateRouteProps): JSX.Element {
 			isAdmin &&
 			(path === ROUTES.SETTINGS ||
 				path === ROUTES.ORG_SETTINGS ||
+				path === ROUTES.MEMBERS_SETTINGS ||
```
H4ad, line 131: What's the reason for this change?

frontend/src/components/LicenseKeyRow/LicenseKeyRow.tsx
```diff
@@ -0,0 +1,54 @@
...
+	if (!activeLicense?.key) {
+		return null;
+	}
+
+	const getMaskedKey = (key: string): string => {
+		if (!key || key.length < 4) {
+			return key || 'N/A';
+		}
+		return `${key.substring(0, 2)}·······${key.slice(-2).trim()}`;
+	};
```
H4ad: This can be exported from AppContext, and you can use the same logic for both places

H4ad: And the implementation of mask for this function is different than the reference

SagarRajput-7: I don't think loading the masked key or function to AppContext would be very helpful here. I see only these 2 places using these, and for the license key, we anyway get from AppContext. And on the other hand, I create this as a common component in the `src/components` in hopes to be reused in future. 

let me know if i am missing something here?

frontend/src/AppRoutes/Private.tsx
SagarRajput-7: components in `ROUTES.MEMBERS_SETTINGS` were extracted from the `ROUTES.ORG_SETTINGS`, so the condition for it will also be the same as `ROUTES.ORG_SETTINGS`. While checking the codebase found this place where it was missing, hence added it.

This is like the very rare case where a user's Workspace is blocked. So mostly just a parity change

frontend/src/components/LicenseKeyRow/LicenseKeyRow.tsx
H4ad: This component looks very specific for custom domains page.

If we don't have plans to use this component in another places, I don't think we should put at the components folder.

Do we have plans to reuse this component on Billing? If not, I still think it's better to add mask inside the AppContext.

SagarRajput-7: As discussed, rather than populating the AppContext, since the masked key function can be useful and reusable and should be consistent hence, decided to have it as a common utility and use at all places.

- also moved licenserow component from `src/components` to local of GeneralSettings.

frontend/src/utils/maskedKey.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Masks a key string, showing only the first 2 and last 2 characters.
+ */
+export function getMaskedKey(key: string): string {
+	if (!key || key.length < 4) {
+		return key || 'N/A';
+	}
+	return `${key.substring(0, 2)}·······${key.slice(-2).trim()}`;
```
H4ad, line 8: Non-blocking: This star icon is different, original: ********

SagarRajput-7: As per the new design, I also asked the same - point - 
- "we should use the unicode dots I added in the figma. Stars are making it too noisy"

Issue #9523, opened 2025-11-08
Do not allow saving alert with invalid payload


srikanthccv:
Ensure the setup is correct with necessary validations before allowing user to save to prevent user from creating broken alert.

tombgh:
It seems to me that the query may be correct at the alert creation stage, but may change to incorrect in the future.
A good solution would also be to send notification and mark the alert as incorrect; currently, you may be unaware of the incorrect query.

srikanthccv:
Hi @tombgh, thanks for the suggestion. Yes, you are indeed right that it can become incorrect in the future for various reasons. We have a plan to address https://github.com/SigNoz/signoz/issues/9276. Is that what you had in mind, or were you thinking of something else other than the "dead" alerts that happen from "telemetry drift" or alert channels that no longer exist?

tombgh:
Hi @srikanthccv, yes, it looks like https://github.com/SigNoz/signoz/issues/9276 should cover that situation.
Recently, we encountered an issue where we weren't receiving notifications due to an incorrect query. We received feedback from end users that something wasn't working in our app. To see and fix the query error, it was necessary to open the details of each alert, one by one. The problem wasn't visible in the alert list.

srikanthccv:
Thank you, I appreciate you taking the time to bring up your case. We plan to build a solution in the product that addresses the class of problems mentioned in https://github.com/SigNoz/signoz/issues/9276

Issue #9276, opened 2025-10-05
Handling dead alerts


srikanthccv:
because 

- the data is no longer flowing
- the service names have been updated
- the endpoints have been migrated
- etc...

PR #10560, opened 2026-03-11
fix: return not-found error with diagnostic info for absent metrics

### LINKED ISSUE #10559: If a histogram metric hasn't seen data in a while, it gives a "something went wrong" error
This happens because the query that fetches for the metric type only checks the query's time range, and if no type is found, the statement builder does not change the space aggregation to sum, which leads to a p75(per_series_value) aggregation that leads to the something went wrong error

namanv3:
## Pull Request

---

### 📄 Summary
If a metric is not returned from the time series table with a temporality and type, that means it is not present with any data in the given time range, and hence we should skip it.

Screenshot 2026-03-12 at 14 14 21

Screenshot 2026-03-12 at 14 14 53

Screenshot 2026-03-12 at 14 16 56

Screenshot 2026-03-12 at 14 18 26

#### Issues closed by this PR
Closes #10559 
Closes https://github.com/SigNoz/engineering-pod/issues/4189

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context
If a histogram metric hasn't seen data in a while, it gives a "something went wrong" error.

This happens because the query that fetches for the metric type only checks the query's time range, and if no type is found, the statement builder does not change the space aggregation to sum, which leads to a p75(per_series_value) aggregation that leads to the something went wrong error.

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/tests/integration/src/querier/03_metrics.py
+++ b/tests/integration/src/querier/03_metrics.py
@@ -19,6 +19,7 @@ from fixtures.utils import get_testdata_file_path
 
 FILL_GAPS = "fillGaps"
 FILL_ZERO = "fillZero"
+HISTOGRAM_FILE = get_testdata_file_path("histogram_data_1h.jsonl")
 
 
 def _build_format_options(fill_mode: str) -> Dict[str, Any]:
@@ -590,7 +591,6 @@ def test_histogram_p90_returns_404_outside_data_window(
     get_token: Callable[[str, str], str],
     insert_metrics: Callable[[List[Metrics]], None],
 ) -> None:
-    HISTOGRAM_FILE = get_testdata_file_path("histogram_data_1h.jsonl")
     
     now = datetime.now(tz=timezone.utc).replace(second=0, microsecond=0)
     metric_name = "test_p90_last_seen_bucket"


pkg/querier/querier.go
```diff
@@ -382,7 +382,7 @@ func (q *querier) QueryRange(ctx context.Context, orgID valuer.UUID, req *qbtype
 					}
 					// TODO(srikanthccv): warn when the metric is missing
 					if spec.Aggregations[i].Temporality == metrictypes.Unknown {
-						spec.Aggregations[i].Temporality = metrictypes.Unspecified
+						spec.Aggregations[i].Temporality = metrictypes.Unspecified // is this needed?
```
srikanthccv: I don't think it's relevant anymore given we now handle the multiple temporality.

pkg/telemetrymetrics/statement_builder.go
```diff
@@ -299,6 +299,7 @@ func (b *MetricQueryStatementBuilder) buildTimeSeriesCTE(
 		sb.LTE("unix_milli", end),
 	)
 
+	// is this needed?
```
srikanthccv: What do we do if a metric goes from `Cumulative` + Monotonic to `Gauge`?

pkg/telemetrymetadata/metadata.go
```diff
@@ -1774,16 +1772,23 @@ func (t *telemetryMetaStore) fetchMetricsTemporalityAndType(ctx context.Context,
...
-		if err := rows.Scan(&metricName, &temporality, &metricType, &isMonotonic); err != nil {
+		var lastSeenAt int64
+		var firstSeenAt int64
+		if err := rows.Scan(&metricName, &temporality, &metricType, &isMonotonic, &lastSeenAt, &firstSeenAt); err != nil {
 			return nil, nil, errors.Wrapf(err, errors.TypeInternal, errors.CodeInternal, "failed to scan temporality result")
 		}
-		if temporality != metrictypes.Unknown {
-			temporalities[metricName] = append(temporalities[metricName], temporality)
-		}
+		// always pick the metric type, regardless of when it was seen
```
srikanthccv: What's the behaviour if the metric is completely missing in the range?

namanv3: the query above has no time range check whatsoever, so we're actually checking the full table. If it doesn't come up even then, then the metric just doesn't exist.

the fact that this change queries the whole table is my biggest concern over here

pkg/telemetrymetrics/statement_builder.go
namanv3: (I'm guessing this comment is for the other file)
Currently the only check is on histogram, if we need a check on counter/gauge then we need to change the above query. This we'd discussed when this change was first made as well

pkg/telemetrymetadata/metadata.go
srikanthccv: https://github.com/SigNoz/signoz/pull/10560#discussion_r2917619831

Here there is a TODO left to warn the users right away about the missing metric rather than attempt to query. We shouldn't let the code reach the point of the query preparation if we know that metric is missing in give period?

namanv3: yes that also makes sense. Let me do this instead

namanv3: this is done

srikanthccv:
Reject the request with invalid input saying metric doesn't exist. It doesn't make sense to silently skip them without notifying user.

namanv3:
@srikanthccv the metric may exist though, just that it has not seen any data recently. And this PR doesn't silently skip, it returns an empty response for that metric, which is the correct response.
The customer had an alert rule for a metric that isn't sending data anymore, wouldn't it make more sense for that customer to see no data instead of seeing an error?

srikanthccv:
Empty response masks the underlying problem. The telemetry drift is a real problem and we deliberately take steps to show such issues to the user as early as possible otherwise it leads to dead alerts/dashboards (or any other assets) that are only discovered later when something goes bad. So the right choice is to let them know details, and about the _metric may exist_, there error should include the detail when did we see last and first see the metric, any other such diagnosis information.

See example https://github.com/SigNoz/signoz/issues/9523#issuecomment-3743836913

namanv3:
@srikanthccv wanted to confirm this approach before adding the diagnostic information about the metric. 
One concern I see is that alerts will still silently fail with this change. Is that something to be handled in #9276 and not this PR?

srikanthccv:
Yes, dead alerts is an independent problem.

namanv3:
@srikanthccv added last seen info for missing metrics

pkg/types/telemetrytypes/store.go
srikanthccv, line 1: We will combine some of the functions into one function to fetch, type, temporality, first seen, last seen, etc... At this point, it would be premature.

namanv3: Yes makes sense. I've fixed the test and added a new one. Please review, will send it over to Vikrant after your approval is there

PR #10563, opened 2026-03-11
fix: newServer uses the stored config hash for mismatch

srikanthccv:
### 📄 Summary

Address the issue introduced with prometheus upgrade in https://github.com/SigNoz/signoz/pull/10467

#### Issues closed by this PR

Partially addresses https://github.com/SigNoz/engineering-pod/issues/4193

### ✅ Change Type

- [x] 🐛 Bug fix

### 🐛 Bug Context

Users who didn't have any Slack channel created prior to this change encountered and error **_no global Slack App URL set_**, as the existing global config didn't have the `slack_app_url` _and_ the code doesn't sync the config

#### Root Cause

Regression, edge case.

#### Fix Strategy

Update getConfig to return the stored hash (before overlays) so newServer compares against the actual DB state. This ensures overlay changes trigger a DB write on the next boot.

- Tests added/updated: no, it's tricky to add meaningful tests for this as it requires a prev DB with specific setup.
- Manual verification: yes
- Edge cases covered: yes

### 📝 Changelog

Fix **_no global Slack App URL set_** error when creating new Slack alert channel. This affected only those users who signed up prior to 11th March and didn't have any existing Slack channel.

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Fix **_no global Slack App URL set_** error when creating new Slack alert channel |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/pkg/alertmanager/service.go
+++ b/pkg/alertmanager/service.go
@@ -72,7 +72,7 @@ func (service *Service) SyncServers(ctx context.Context) error {
 
 	service.serversMtx.Lock()
 	for _, org := range orgs {
-		config, err := service.getConfig(ctx, org.ID.StringValue())
+		config, _, err := service.getConfig(ctx, org.ID.StringValue())
 		if err != nil {
 			service.settings.Logger().ErrorContext(ctx, "failed to get alertmanager config for org", "org_id", org.ID.StringValue(), "error", err)
 			continue
@@ -171,7 +171,7 @@ func (service *Service) Stop(ctx context.Context) error {
 }
 
 func (service *Service) newServer(ctx context.Context, orgID string) (*alertmanagerserver.Server, error) {
-	config, err := service.getConfig(ctx, orgID)
+	config, storedHash, err := service.getConfig(ctx, orgID)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


pkg/alertmanager/service.go
```diff
@@ -200,27 +203,33 @@ func (service *Service) newServer(ctx context.Context, orgID string) (*alertmana
...
 		config, err = alertmanagertypes.NewDefaultConfig(service.config.Global, service.config.Route, orgID)
 		if err != nil {
-			return nil, err
+			return nil, "", err
 		}
+	} else {
+		storedHash = config.StoreableConfig().Hash
 	}
 
 	if err := config.SetGlobalConfig(service.config.Global); err != nil {
```
namanv3, line 225: how can we be sure that the global config being set here is the latest from the package?

pkg/alertmanager/service.go
```diff
@@ -200,27 +203,33 @@ func (service *Service) newServer(ctx context.Context, orgID string) (*alertmana
 	return server, nil
 }
 
-func (service *Service) getConfig(ctx context.Context, orgID string) (*alertmanagertypes.Config, error) {
+// getConfig returns the config for the given orgID with overlays applied, along
+// with the hash that was stored in the DB before overlays. When no config exists
+// in the store yet the stored hash is empty.
+func (service *Service) getConfig(ctx context.Context, orgID string) (*alertmanagertypes.Config, string, error) {
```
namanv3, line 209: can we change the method name to convey that overlays are applied?

pkg/alertmanager/service.go
srikanthccv: Please look at the `SetGlobalConfig`

pkg/alertmanager/service.go
srikanthccv: It would make sense if there was another function which doesn't apply the overlays, but there is only one function and it documents the behaviour so I think the name is fine.

PR #10564, opened 2026-03-12
feat(pnpm): migrate away from yarn

H4ad:
## Pull Request

Migrate to pnpm, reduce install time from [128s](https://github.com/SigNoz/signoz/actions/runs/22976709115/job/66706978385#step:4:95) to [~16s](https://github.com/SigNoz/signoz/actions/runs/22982343129/job/66724768515?pr=10564#step:5:264), ~7x faster.

---

This PR is the first of three PRs, the migration will be in the following way:
- Add pnpm-lock.yaml to allow `pnpm install`.
- Update `primus` repository to use pnpm.
- Update `primus.workflows` to use pnpm.
- Remove `yarn.lock` from this repo.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

The changes in the package.json was caused by some packages that we are using are not correctly declared at the package.json (eg: many types packages such as types for jest-dom).

I tried to minimize and keep the compatibilty between yarn and pnpm as close as possible, the conversion was made with `pnpm import`.

---

- Blast radius: N/A
- Potential regressions: If we forget to update some place, it should still work due to yarn.
- Rollback plan: Or revert this PR, or open a PR to fix the bug.

---

### 📋 Checklist
- [x] Manually tested


frontend/playwright.config.ts
H4ad, line 1: This change fixes a bug caused by the migration of Vite in the Playwright CI Job.

frontend/yarn.lock
```diff
@@ -18802,6 +19005,34 @@ tailwindcss-animate@^1.0.7:
   resolved "https://registry.yarnpkg.com/tailwindcss-animate/-/tailwindcss-animate-1.0.7.tgz#318b692c4c42676cc9e67b19b78775742388bef4"
   integrity sha512-bl6mpH3T7I3UFxuvDEXLxy/VuFxBk5bbzplh7tXI68mwMokNYd1t9qPBHlnyTwfa4JGC4zP516I1hYYtQ/vspA==
 
+tailwindcss@3:
```
H4ad, line 19017: Required because of our packages from signozhq/<component>

frontend/package.json
```diff
@@ -97,7 +98,7 @@
     "color-alpha": "2.0.0",
     "cross-env": "^7.0.3",
     "crypto-js": "4.2.0",
-    "d3-hierarchy": "3.1.2",
+    "d3-hierarchy": "1.1.9",
```
H4ad, line 101: We need to be careful when testing this, by types, looks like we added 3.1.2 while we were using the 1.1.9 behind the scenes.

frontend/package.json
```diff
@@ -33,14 +33,15 @@
     "@ant-design/icons": "4.8.0",
     "@codemirror/autocomplete": "6.18.6",
     "@codemirror/lang-javascript": "6.2.3",
+    "@codemirror/state": "6.5.4",
```
H4ad, line 36: Be careful when testing this one as well.

H4ad:
Closed in favor of https://github.com/SigNoz/signoz/pull/11158

PR #10577, opened 2026-03-13
fix: message field key search in JSON Logs

piyushsingariya:
## Pull Request

---

### 📄 Summary
This PR aims to fix searches on `message` field which has to skip lambda expressions and has to be accessed directly as a column of logs v2 table
- Materialized for promoted sign has been changed if it's a typehint, more similar to how actually fields are materialized as directly accessible columns for Attributes and Resources.
- Promotion later will be recognized by field evolution

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
+++ b/pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
@@ -22,9 +22,6 @@ type MockMetadataStore struct {
 	LogsJSONIndexesMap map[string][]schemamigrator.Index
 	LookupKeysMap      map[telemetrytypes.MetricMetadataLookupKey]int64
 	// StaticFields holds signal-specific intrinsic field definitions (e.g. telemetrylogs.IntrinsicFields).
-	// It is injected into GetKeys / GetKey results, mirroring what the real metadata store does when
-	// it reads telemetrylogs.IntrinsicFields directly. Callers pass their package's IntrinsicFields
-	// map at construction time to avoid a circular import.
 	StaticFields map[string]telemetrytypes.TelemetryFieldKey
 }
 
@@ -75,8 +72,6 @@ func (m *MockMetadataStore) GetKeys(ctx context.Context, fieldKeySelector *telem
 	}
 
 	// StaticFields (e.g. IntrinsicFields), mirroring the real metadata store.
-	// Each intrisic key always gets its own entry (so "message" and "body" both
-	// resolve to the IntrinsicField independently).
 	for key, field := range m.StaticFields {
 		if !matchesName(fieldKeySelector, key) {
 			continue


pkg/telemetrymetadata/body_json_metadata.go
```diff
@@ -184,6 +185,9 @@ func buildGetBodyJSONPathsQuery(fieldKeySelectors []*telemetrytypes.FieldKeySele
 		limit += fieldKeySelector.Limit
 	}
 	sb.Where(sb.Or(orClauses...))
+	// mesasge field is skipped; since it is a type hint and is handled by the field mapper
+	// TODO(Piyush): If typehints increases in future, use aftership parser to skip type hints here
```
nityanandagohain: what is aftership parser ?

Do you mean not writing it from jsontypeexporter ?

pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
```diff
@@ -21,10 +21,20 @@ type MockMetadataStore struct {
...
+	// map at construction time to avoid a circular import.
+	StaticFields map[string]telemetrytypes.TelemetryFieldKey
 }
 
-// NewMockMetadataStore creates a new instance of MockMetadataStore with initialized maps
-func NewMockMetadataStore() *MockMetadataStore {
+// NewMockMetadataStore creates a new instance of MockMetadataStore with initialized maps.
+// Pass the signal-specific intrinsic fields (e.g. telemetrylogs.IntrinsicFields) so the mock
+// mirrors what the real metadata store does when injecting those definitions into key results.
+func NewMockMetadataStore(intrinsicFields map[string]telemetrytypes.TelemetryFieldKey) *MockMetadataStore {
```
nityanandagohain:
```suggestion
func(m *metastore) SetStaticFields(intrinsicFields map[string]telemetrytypes.TelemetryFieldKey) ...
```

pkg/telemetrymetadata/metadata.go
```diff
@@ -566,12 +566,15 @@ func (t *telemetryMetaStore) getLogsKeys(ctx context.Context, fieldKeySelectors
...
 		if found {
 			if field, exists := telemetrylogs.IntrinsicFields[key]; exists {
-				if _, added := setOfKeys[field.Name+";"+field.FieldContext.StringValue()+";"+field.FieldDataType.StringValue()]; !added {
-					mapOfKeys[field.Name] = append(mapOfKeys[field.Name], &field)
-					// intrinsic field can be a logical field mapped to a different physical location
-					mapOfKeys[key] = append(mapOfKeys[key], &field)
+				// Register by physical name only once and only when it differs from the
+				// logical key — if they are the same, the always-register below covers it.
+				if _, added := setOfKeys[field.Text()]; !added {
+					mapOfKeys[field.Text()] = append(mapOfKeys[field.Text()], &field)
```
nityanandagohain: help me understand this, I see that nowhere mapOfKeys store the entire field.text(), it always has a name and array of keys with similar name but different type.
I can see that somewhere later it will help you use with bodyJSONPaths , but not able to wrap around my head what we are doing here.

pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
```diff
@@ -34,35 +44,56 @@ func NewMockMetadataStore() *MockMetadataStore {
...
+	// only once to avoid duplicate entries in that slot.
+	for key, field := range m.StaticFields {
+		if !matchesName(fieldKeySelector, key) {
+			continue
+		}
+
+		// Register by physical name only once and only when it differs from the
+		// logical key — if they are the same, the always-register below covers it.
+		if _, exists := setOfKeys[field.Text()]; !exists {
+			result[field.Text()] = append(result[field.Text()], &field)
```
nityanandagohain: Same question as metadata store. Where is `result[field.Text()]` getting used ?


Also would recommend to revisit the comments that getting added to the code. Also physical and logical is something confusing I feel, better use just name and name with context and type.

eg:- I am not sure what `Each logical key always gets its own entry ` , `The physical name is registered only once to avoid duplicate entries in that slot` and `Always register the logical key so that every alias in IntrinsicFields` means.

pkg/telemetrylogs/const.go
```diff
@@ -39,20 +40,22 @@ const (
...
+	// mapper to access it as a direct sub-column (body_v2.message) rather than via
+	// dynamicElement(). Its name is the bare field name ("message"), not the column path.
 	BodyLogicalFieldJSONMapping = &telemetrytypes.TelemetryFieldKey{
-		Name:          MessageSubColumn,
+		Name:          MessageBodyField,
 		Signal:        telemetrytypes.SignalLogs,
-		FieldContext:  telemetrytypes.FieldContextLog,
+		FieldContext:  telemetrytypes.FieldContextBody,
 		FieldDataType: telemetrytypes.FieldDataTypeString,
+		Materialized:  true,
```
nityanandagohain, line 56: Don't you think bringing materialized logic to something else will be confusing, I would recommend, just use if name==MessageBodyField do anything specific for body and not mix it with materialized. We treat all intrinsic fields as materialized regardless.

pkg/telemetrymetadata/body_json_metadata.go
piyushsingariya: aftership parser -> https://github.com/AfterShip/clickhouse-sql-parser. 

> Do you mean not writing it from jsontypeexporter ?

Yes I've thought about it as well.

pkg/telemetrylogs/const.go
piyushsingariya: `Materialized` signifies keys that are to be accessed from table level, without a field expression, so it fit right for this case.

Currently, we rely on a number of body-specific checks, which makes the component feel tightly coupled to QB like an external extension of it. I want to reduce these checks hence used Materialized field.

pkg/telemetrymetadata/body_json_metadata.go
piyushsingariya: I've pushed some changes to collector to account for not exporting `message`, but I think this check should stay.

pkg/types/telemetrytypes/telemetrytypestest/metadata_store.go
piyushsingariya: I understand, the comments are definitely confusing, fixing these.

pkg/telemetrymetadata/metadata.go
piyushsingariya: This is a mistake, setOfKeys is only needs to work with `field.Text()`

PR #10579, opened 2026-03-13
fix: variables initial url state

AshwinBhatkal:
### 📄 Summary
Dashboard variables were not persisting their selected values when navigating from the widget editor back to the dashboard. This happened because:
1. Variable URL state initialization was running against stale (pre-transform) variable data, causing values to be reset to defaults on load.
2. Navigation from the widget editor (`NewWidget`) discarded the `variables` URL param when returning to the dashboard, losing all current variable selections.

---

### ✅ Change Type

- [x] 🐛 Bug fix

---

#### Root Cause
Two separate issues:

1. In `Dashboard.tsx`, `initializeDefaultVariables` was called with `data?.data?.data?.variables` (raw API response) before `transformDashboardVariables` ran. This meant variables were initialized from untransformed data, potentially overwriting URL-encoded variable state with stale defaults.

2. In `NewWidget/index.tsx` (and the duplicate call in `DashboardVariableSelection.tsx`), navigation back to the dashboard used a plain `generatePath(ROUTES.DASHBOARD, { dashboardId })` URL that stripped all query params — including the `variables` param that holds variable selections.

#### Fix Strategy
- Reorder `Dashboard.tsx` so `transformDashboardVariables` runs first, then pass the transformed variables to `initializeDefaultVariables`.
- Remove the duplicate `initializeDefaultVariables` call from `DashboardVariableSelection` (it was redundant and racing with the dashboard-level initialization).
- Extract a `navigateToDashboardPage` helper in `NewWidget` that forwards the `variables` URL param when navigating back to the dashboard. Apply it to all three navigation call sites (save, discard, switch-to-view-mode).

---

### 🧪 Testing Strategy

- Tests added/updated: None (behaviour change only). State transition will be handled when local storage handling of variables is done.
- Manual verification: Open a dashboard with variables, set non-default values, open a widget for editing, save or discard, confirm variables are preserved on return.
- Edge cases covered: No variables present (no `variables` param → no-op); navigate via save, discard, and view-mode switch.

---

- Blast radius: Dashboard variable state on navigation — limited to the dashboard page and widget editor.
- Potential regressions: Variables that relied on being reset to defaults when re-entering the dashboard could behave differently; however, preserving user-selected values is the intended behavior.
- Rollback plan: Fix forward

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | NA |
| Change Type | Bug Fix |
| Description | This issue did not go to prod |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/pages/DashboardWidget/__tests__/DashboardWidget.test.tsx
+++ b/frontend/src/pages/DashboardWidget/__tests__/DashboardWidget.test.tsx
@@ -1,0 +1,146 @@
+import { Route } from 'react-router-dom';
+import { PANEL_TYPES } from 'constants/queryBuilder';
+import { rest, server } from 'mocks-server/server';
+import { render, screen, waitFor } from 'tests/test-utils';
+
+import DashboardWidget from '../index';
+
+const DASHBOARD_ID = 'dash-1';
+const WIDGET_ID = 'widget-abc';
+
+const mockDashboardResponse = {
+	status: 'success',
+	data: {
+		id: DASHBOARD_ID,
+		createdAt: '2024-01-01T00:00:00Z',
+		createdBy: 'test',
+		updatedAt: '2024-01-01T00:00:00Z',
+		updatedBy: 'test',
+		isLocked: false,
+		data: {
+			collapsableRowsMigrated: true,
+			description: '',
+			name: '',
+			panelMap: {},
+			tags: [],
+			title: 'Test Dashboard',
+			uploadedGrafana: false,
+			uuid: '',
+			version: '',
+			variables: {},
+			widgets: [],
+			layout: [],
+		},
+	},
+};
+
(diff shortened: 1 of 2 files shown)


frontend/src/container/DashboardContainer/DashboardVariablesSelection/DashboardVariableSelection.tsx
```diff
@@ -50,15 +49,6 @@ function DashboardVariableSelection(): JSX.Element | null {
 		(state) => state.globalTime,
 	);
 
-	useEffect(() => {
```
AshwinBhatkal, line 53: Already happening on successful fetch of dashboard. Was required earlier when dashboard provider was at the root as the widget page just navigated to dashboard page without any query params. In this PR, the dashboard page is navigated to with the right variable query params

PR #10580, opened 2026-03-13
chore: add slack notification on dequeue from merge queue

AshwinBhatkal:
- If any PR is dequeued from the merge queue for any reason, a slack notification needs to be sent across notifying the  owners of it. 
- Currently not able to link slack id with github id. 

- Devs could also subscribe to PR workflows with the Github app on slack as an alternative

### DIFF
--- a/.github/workflows/mergequeueci.yaml
+++ b/.github/workflows/mergequeueci.yaml
@@ -1,4 +1,4 @@
-name: Notify on Dequeue from Merge Queue
+name: mergequeueci
 
 on:
   pull_request:
@@ -9,7 +9,7 @@ jobs:
   notify:
     runs-on: ubuntu-latest
     steps:
-      - name: Send Slack alert
+      - name: alert
         uses: slackapi/slack-github-action@v2.1.1
         with:
           webhook: ${{ secrets.SLACK_MERGE_QUEUE_WEBHOOK }}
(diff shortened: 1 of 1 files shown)


.github/workflows/notify-on-dequeue.yaml
```diff
@@ -0,0 +1,59 @@
+name: Notify on Dequeue from Merge Queue
```
therealpandey: `mergequeueci`.

If you look at the other pipelines, we add the suffix `ci` after the name of the pipeline.

Please also rename the file to `mergequeueci`

.github/workflows/notify-on-dequeue.yaml
```diff
@@ -0,0 +1,59 @@
...
+on:
+  pull_request:
+    types:
+      - dequeued
+
+jobs:
+  notify:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send Slack alert
```
therealpandey:
```suggestion
      - name: alert
```

.github/workflows/notify-on-dequeue.yaml
```diff
@@ -0,0 +1,59 @@
...
+                  "fields": [
+                    {
+                      "type": "mrkdwn",
+                      "text": "*Author*\n@${{ github.event.pull_request.user.login }}"
+                    }
+                  ]
+                }
+              ]
+            }
+      - name: Post PR comment
```
therealpandey:
```suggestion
      - name: comment
```

PR #10583, opened 2026-03-13
feat: send all data for trace list api

nityanandagohain:
## Pull Request

---

### 📄 Summary
To support new JSON view.
We want to send all details by default so that the sidepanel can be loaded very fast.

---

This won't break existing frontend as we always send selectFields form the UI and there is no case where selectFields will be empty from the UI

---

#### Issues closed by this PR

Fixes https://github.com/SigNoz/engineering-pod/issues/4203
Fixes https://github.com/SigNoz/engineering-pod/issues/3821

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

* Unit tests and integration tests

---

- Blast radius: Trace list view
- Potential regressions: None
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---


tests/integration/src/querier/04_traces.py
```diff
@@ -707,6 +701,108 @@ def test_traces_list_with_corrupt_data(
...
+        "spec": {
+            "name": "A",
+            "signal": "traces",
+            "selectFields": [],
+            "order": [{"key": {"name": "timestamp"}, "direction": "desc"}],
+            "limit": 1,
+        },
+    }
+
+    tests = [
```
tushar-signoz: Checkout @pytest.mark.parametrize if it makes sense to be used here

tests/integration/src/querier/04_traces.py
```diff
@@ -707,6 +701,108 @@ def test_traces_list_with_corrupt_data(
...
+                * 1000
+            ),
+            end_ms=int(datetime.now(tz=timezone.utc).timestamp() * 1000),
+            request_type="raw",
+            queries=[payload],
+        )
+        assert response.status_code == HTTPStatus.OK
+        data = response.json()
+        print(response.json())
+        print(data["data"]["data"]["results"][0]["rows"][0]["data"].keys())
```
tushar-signoz:
```suggestion
```

tushar-signoz (changes requested):
Also update go unit tests

pkg/telemetrytraces/statement_builder.go
```diff
@@ -311,13 +274,53 @@ func (b *traceQueryStatementBuilder) buildListQuery(
...
+		sb.SelectMore(SpanIsRemoteColumn)
+
+		// select all contextual columns
+		sb.SelectMore(SpanAttributesStringColumn)
+		sb.SelectMore(SpanAttributesNumberColumn)
+		sb.SelectMore(SpanAttributesBoolColumn)
+		sb.SelectMore(SpanResourcesStringColumn)
+	} else {
+		for _, field := range query.SelectFields {
+			if field.Name == SpanTimestampColumn || field.Name == SpanTraceIDColumn || field.Name == SpanSpanIDColumn {
```
tushar-signoz:
```suggestion
            // TODO(tvats): If a user specifies attribute.timestamp in the select fields, this loop will basically ignore it, as we already added a field by default. This can be fixed once we close https://github.com/SigNoz/engineering-pod/issues/3693
			if field.Name == SpanTimestampColumn || field.Name == SpanTraceIDColumn || field.Name == SpanSpanIDColumn {
```

tushar-signoz (commented):
Looks good, can you add the TODO comment

tushar-signoz (dismissed):
LGTM

pkg/telemetrytraces/statement_builder.go
srikanthccv: The comment says they are added by default, but I don't see or recall where they are added by default? Can you please share the ref?

tests/integration/src/querier/04_traces.py
```diff
@@ -1,3 +1,4 @@
+from asyncio.unix_events import SelectorEventLoop
```
srikanthccv: not used

tests/integration/src/querier/04_traces.py
```diff
@@ -490,25 +491,24 @@ def test_traces_list(
...
                 x[3].span_id,
                 format_timestamp(x[3].timestamp),
                 x[3].trace_id,
             ],  # type: Callable[[List[Traces]], List[Any]]
         ),
         # Case 2: order by attribute timestamp field which is there in attributes as well
-        # This should break but it doesn't because attribute.timestamp gets adjusted to timestamp
-        # because of default trace.timestamp gets added by default and bug in field mapper picks
-        # instrinsic field
+        # This breaks as (timestamp >= '1773738262411000000') AND (timestamp < '1773738562411000000') order by attributes_string['timestamp'] as timestamp is invalid query
```
srikanthccv: The behaviour is changing here but it shouldn't change? because we put in a fix to ensure that it doesn't break?

pkg/telemetrytraces/statement_builder.go
srikanthccv: By that I meant I don't recall where they are added other than the code that got removed.

tushar-signoz: `Line 277: sb.SelectMore(SpanTimestampColumn)`, I should have phrased it better.

tests/integration/src/querier/04_traces.py
nityanandagohain: Yeah this is a behaviour change, where now it won't return any rows. It won't result in error though.

Do you have a suggestion if we should still keep it as it is i.e override it to the top level timestamp column like prevously

pkg/telemetrytraces/field_mapper.go
```diff
@@ -247,6 +247,15 @@ func (m *defaultFieldMapper) FieldFor(
...
 	}
 
+	// special handling for contextual map fields
+	if key.FieldContext == telemetrytypes.FieldContextSpan &&
+		(key.Name == SpanAttributesStringColumn ||
+			key.Name == SpanAttributesNumberColumn ||
+			key.Name == SpanAttributesBoolColumn ||
+			key.Name == SpanResourcesStringColumn) {
+		return key.Name, nil
+	}
```
srikanthccv: This doesn't belong here. A  `SpanAttributesStringColumn` / `SpanAttributesNumberColumn` is not a `TelemetryFieldKey`.

srikanthccv (commented):
What should be the response format when we send all the data? There are number of items which are directly the table schema names.

```
                "attributes_number",
                "attributes_string",
                "attributes_bool",
                "resources_string",
```

@tushar-signoz you had some point here

pkg/telemetrytraces/field_mapper.go
nityanandagohain: So there was two options, 
* to keep it here or 
* when I iterate over the selectFields, then I will have to put the if clause to check the key names and override.

I went with this as felt feld_mapper should do it. Let me know if the other option is right or any different suggestion.

srikanthccv: @tushar-signoz had some input and working on here, can you please check with him once.

nityanandagohain: sure will check with him.

nityanandagohain:
* We will wait for @tushar-signoz  to add his changes for aliasing, as right now not possible to add the fields in the sql builder itself, without adding a behaviour change. This will make sure that we don't use `attribute_string` as a telemetry field.
* The response payload shouldn't be connected to `attributes_string,attributes_number` intead it should be just attributes or resources.
* How we select the attributes or resources as a single map
   * Select them individually and postporcess
   * OR if we can merge them at clickhouse itslef by casting to JSON. ( check what is cleaner) ( also check if using json will bring up nested structures)
* How should resource use the column evolution should it use `multiif( resource, resources_string)` should it use the field_mapper/evolutions to find the right column.

----
Change schema version in response since we are changing the structure.

pkg/telemetrytraces/field_mapper.go
nityanandagohain: removed it from field_mapper

nityanandagohain:
* So I have made sure that we convert attributes_string, attributes_number ... to attributes and resources_string to resource. using postprocess as converting to json and then post processing again didn't make sense unless we have nested keys handling in the UI.
* Since we are adding support for this for the first time in trace list endpoint don't need to put it under a new API etc as the existing functionality remains the same only when selectfields is empty the new changes are seen.

pkg/querier/consume.go
```diff
@@ -431,6 +431,48 @@ func readAsRaw(rows driver.Rows, queryName string) (*qbtypes.RawData, error) {
...
+func mergeSpanAttributeColumns(data map[string]any) {
+	attrStr, hasStr := data["attributes_string"]
+	attrNum, hasNum := data["attributes_number"]
+	attrBool, hasBool := data["attributes_bool"]
+	// todo(nitya): move to resource json
+	resStr, hasRes := data["resources_string"]
+
+	if !hasStr && !hasNum && !hasBool && !hasRes {
+		return
+	}
```
tushar-signoz: If none of the map exists. We would still want to return `{"attributes": {}, "resource":{}}`

pkg/telemetrytraces/statement_builder.go
```diff
@@ -841,3 +818,52 @@ func (b *traceQueryStatementBuilder) buildResourceFilterCTE(
...
+	}
+	if wasEmpty {
+		// Select all intrinsic columns
+		selectFields = append(selectFields, telemetrytypes.TelemetryFieldKey{Name: SpanTraceStateColumn, FieldContext: telemetrytypes.FieldContextSpan})
+		selectFields = append(selectFields, telemetrytypes.TelemetryFieldKey{Name: SpanParentSpanIDColumn, FieldContext: telemetrytypes.FieldContextSpan})
+		selectFields = append(selectFields, telemetrytypes.TelemetryFieldKey{Name: SpanFlagsColumn, FieldContext: telemetrytypes.FieldContextSpan})
+		selectFields = append(selectFields, telemetrytypes.TelemetryFieldKey{Name: SpanNameColumn, FieldContext: telemetrytypes.FieldContextSpan})
+		selectFields = append(selectFields, telemetrytypes.TelemetryFieldKey{Name: SpanKindColumn, FieldContext: telemetrytypes.FieldContextSpan})
+		selectFields = append(selectFields, telemetrytypes.TelemetryFieldKey{Name: SpanKindStringColumn, FieldContext: telemetrytypes.FieldContextSpan})
+		selectFields = append(selectFields, telemetrytypes.TelemetryFieldKey{Name: SpanDurationNanoColumn, FieldContext: telemetrytypes.FieldContextSpan})
```
tushar-signoz: Instead of this, can we create a Maps in const.go ?
InstrinsicFields 
CalculatedFields
ContextualFields This would greatly improve readability of this section.

pkg/telemetrytraces/statement_builder.go
```diff
@@ -86,40 +84,12 @@ func (b *traceQueryStatementBuilder) Build(
...
-		}
-
-		for _, x := range []string{"timestamp", "span_id", "trace_id"} {
-			if !slices.Contains(selectFieldKeys, x) {
-				query.SelectFields = append(query.SelectFields, DefaultFields[x])
-			}
-		}
+		// we are expanding here to ensure that all the conflicts are taken care in adjustKeys
+		// i.e if there is a conflict we strip away context of the key in adjustKeys
+		query, isSelectFieldsEmpty = b.expandRawSelectFields(query)
```
tushar-signoz: nit: does b.expandRawSelectFields(query) need to return isSelectFieldsEmpty ? can we not check the length before calling it ?

tests/integration/tests/querier/04_traces.py
```diff
@@ -481,42 +481,44 @@ def test_traces_list(
                     "name": "A",
                     "signal": "traces",
                     "disabled": False,
+                    "selectFields": [
+                        {"name": "span_id"},
+                        {"name": "span.timestamp"},
+                        {"name": "trace_id"},
+                    ],
```
tushar-signoz: Hey @nityanandagohain 
Lets not modify the existing integration test, you may add 1 more to extend the coverage. Makes it easier to spot behaviour change.

tests/integration/tests/querier/04_traces.py
```diff
@@ -481,42 +481,44 @@ def test_traces_list(
...
                     "disabled": False,
-                    "order": [{"key": {"name": "attribute.timestamp"}, "direction": "desc"}],
+                    "selectFields": [
+                        {"name": "span_id"},
+                        {"name": "span.timestamp"},
+                        {"name": "trace_id"},
+                    ],
+                    "order": [
+                        {"key": {"name": "attribute.timestamp"}, "direction": "desc"}
+                    ],
```
tushar-signoz: Same

tests/integration/tests/querier/04_traces.py
```diff
@@ -693,6 +699,112 @@ def test_traces_list_with_corrupt_data(
...
+                "status_message",
+                "status_code_string",
+                "events",
+                "links",
+                # all calculated columns
+                "response_status_code",
+                "external_http_url",
+                "http_url",
+                "external_http_method",
+                "http_method",
```
tushar-signoz: Lets create a constant 'all_select_fields' or something better named. Would be useable for other tests as well

pkg/telemetrytraces/field_mapper_test.go
```diff
@@ -78,6 +78,16 @@ func TestGetFieldKeyName(t *testing.T) {
...
+		{
+			name: "Contextual map column - attributes_string without span context does not short-circuit",
+			key: telemetrytypes.TelemetryFieldKey{
+				Name:          SpanAttributesStringColumn,
+				FieldContext:  telemetrytypes.FieldContextAttribute,
+				FieldDataType: telemetrytypes.FieldDataTypeString,
+			},
+			expectedResult: "attributes_string['attributes_string']",
+			expectedError:  nil,
+		},
```
tushar-signoz, line 95: Hey @nityanandagohain 
Just a nit that the test case name is mis-leading. Maybe add more comments to elaborate that what this test shows is 'attribute.attribute_string:string' is valid telemetry field key enough that the name collides with the contextual map column (IF that was the intent of this test-case)

tushar-signoz (changes requested):

tests/integration/tests/querier/04_traces.py
nityanandagohain: you mean to remove selectFields and let it return all ?

pkg/telemetrytraces/field_mapper_test.go
nityanandagohain: Thanks have updated it

tests/integration/tests/querier/04_traces.py
nityanandagohain: Done 👍

tests/integration/tests/querier/04_traces.py
nityanandagohain: Removed it.

pkg/querier/consume.go
nityanandagohain: @tushar-signoz I had to revert this as otherwise, it also gets added to the case where you are specifying select fields and then `{"attributes": {}, "resource":{}}` appears in the response so revmoed it,

Also when we actually select it there will be rarely a case where these two won't be present.

tushar-signoz (approved):
LGTM !

pkg/querier/trace_operator_query.go
```diff
@@ -85,6 +85,13 @@ func (q *traceOperatorQuery) executeWithContext(ctx context.Context, query strin
 		return nil, err
 	}
 
+	// TODO: This should move to readAsRaw function in consume.go but for now we can keep it here since it's only relevant for traces
```
srikanthccv, line 88: Can you please create an issue in engineering-pod and assign it to Tushar? and link it here.

pkg/telemetrytraces/const.go
```diff
@@ -334,6 +378,51 @@ var (
 	SpanSearchScopeRoot       = "isroot"
 	SpanSearchScopeEntryPoint = "isentrypoint"
 
+	// IntrinsicSpanFields lists the intrinsic span columns, in the order they
+	// should appear when a raw query expands its SelectFields.
+	IntrinsicSpanFields = []telemetrytypes.TelemetryFieldKey{
```
srikanthccv, line 383: This is a duplicate for `IntrinsicFields` but I see why it's needed, instead of modifying the `IntrinsicFields` so we don't return the `events` / `links` / `timestamp` in the keys suggestions.  Can you please create todo ticket to maintain a single list of intrinsic and calculated fields?

pkg/telemetrytraces/statement_builder.go
```diff
@@ -844,3 +821,30 @@ func (b *traceQueryStatementBuilder) buildResourceFilterCTE(
 		variables,
 	)
 }
+
+// expandRawSelectFields populates SelectFields for raw (list view) queries.
+// It must be called before adjustKeys so that normalization runs over the full set.
+func (b *traceQueryStatementBuilder) expandRawSelectFields(query qbtypes.QueryBuilderQuery[qbtypes.TraceAggregation]) qbtypes.QueryBuilderQuery[qbtypes.TraceAggregation] {
+	if len(query.SelectFields) == 0 {
+		selectFields := make([]telemetrytypes.TelemetryFieldKey, 0, len(IntrinsicSpanFields)+len(CalculatedSpanFields))
+		selectFields = append(selectFields, IntrinsicSpanFields...)
```
srikanthccv, line 837: What is the response shape for `events` and `links`? It doesn't make sense to return the stringfied json of events in the response (which I believe is the db state) because that's not useable and user will have to deser themselves to make use of it.

Additionally, it would be the very same problem we are trying to address here https://www.notion.so/signoz/PRD-Semantic-Log-Response-Stop-Leaking-Internal-Schema-WIP-338fcc6bcd19806986cece5b030da782, i.e leaking the schema shape into user responses.

tests/integration/tests/querier/04_traces.py
```diff
@@ -556,7 +641,11 @@ def test_traces_list(
                 },
             },
             HTTPStatus.OK,
-            lambda x: [],  # type: Callable[[List[Traces]], List[Any]]
+            lambda x: [
+                x[0].span_id,
+                format_timestamp(x[0].timestamp),
+                x[0].trace_id,
+            ],  # type: Callable[[List[Traces]], List[Any]]
```
srikanthccv, line 621: This is behaviour change, can you help me understand why it returned no data earlier but returns data now?

nityanandagohain: Prev
```
SELECT
    attributes_string['timestamp'] AS timestamp,
    span_id AS span_id,
    trace_id AS trace_id
WHERE (mapContains(attributes_string, 'timestamp') = 1) AND (timestamp >= '1778241100000000000') AND (timestamp < '1778262700000000000') AND (ts_bucket_start >= 1778239300) AND (ts_bucket_start <= 1778262700)
LIMIT 1
```

Now
```
 SELECT
    timestamp AS timestamp,
    trace_id AS trace_id,
    span_id AS span_id
WHERE (mapContains(attributes_string, 'timestamp') = 1) AND (timestamp >= '1778241100000000000') AND (timestamp < '1778262700000000000') AND (ts_bucket_start >= 1778239300) AND (ts_bucket_start <= 1778262700)
LIMIT 1
```

So previously we don't add `log.timestamp` if a `timestamp` is present in attribute and thus with aliasing issue there is no result, but now we first add the top level timestamp and it get's the higher priority.

pkg/telemetrytraces/const.go
nityanandagohain: sure https://github.com/SigNoz/engineering-pod/issues/4861

pkg/querier/trace_operator_query.go
nityanandagohain: https://github.com/SigNoz/engineering-pod/issues/4862

pkg/telemetrytraces/statement_builder.go
nityanandagohain: I guess you are suggesting to process them and send them as structured one like we are doing for attributes and resource attributes in after consuming right ?

cc @tushar-signoz

pkg/telemetrytraces/statement_builder.go
```diff
@@ -844,3 +821,30 @@ func (b *traceQueryStatementBuilder) buildResourceFilterCTE(
...
+		return query
+	}
+
+	selectFields := []telemetrytypes.TelemetryFieldKey{
+		{Name: SpanTimestampColumn, FieldContext: telemetrytypes.FieldContextSpan},
+		{Name: SpanTraceIDColumn, FieldContext: telemetrytypes.FieldContextSpan},
+		{Name: SpanSpanIDColumn, FieldContext: telemetrytypes.FieldContextSpan},
+	}
+	for _, field := range query.SelectFields {
+		// TODO(tvats): If a user specifies attribute.timestamp in the select fields, this loop will basically ignore it, as we already added a field by default. This can be fixed once we close https://github.com/SigNoz/engineering-pod/issues/3693
```
tushar-signoz, line 849: @srikanthccv Because of this, we're seeing the behaviour change, and its a minor improvement, we should be good with it.

pkg/telemetrytraces/statement_builder.go
nityanandagohain: Yes as of now it is stringified and sent in the response as something like this
```
{
                                "events": [
                                    "{\"name\":\"request_received\",\"timeUnixNano\":1778489782759245000,\"attributeMap\":{\"http.method\":\"GET\",\"http.route\":\"/api/chat\"}}",
                                    "{\"name\":\"cache_lookup\",\"timeUnixNano\":1778489782811697000,\"attributeMap\":{\"cache.hit\":\"true\",\"cache.key\":\"user:123:prompt\"}}",
                                    "{\"name\":\"model_invoked\",\"timeUnixNano\":1778489782863268000,\"attributeMap\":{\"gen_ai.request.model\":\"gpt-4\",\"gen_ai.usage.input_tokens\":\"100\"}}",
                                    "{\"name\":\"response_sent\",\"timeUnixNano\":1778489782863273000,\"attributeMap\":{\"http.status_code\":\"200\",\"response.size_bytes\":\"2048\"}}"
                                ],
                                "links": "[{\"traceId\":\"1a97c5846f205d8cb2e819f7bcf80bcd\",\"refType\":\"CHILD_OF\"},{\"traceId\":\"1ff8b4c80f98b9e12420e27955c77e2d\",\"spanId\":\"3dc5e2497b5f7ba5\",\"refType\":\"FOLLOWS_FROM\"}]"
}
```


While the other parts of the product like waterfall endpoint sends it as a json

eg:
```
{
"references": [
                {
                    "traceId": "1a97c5846f205d8cb2e819f7bcf80bcd",
                    "refType": "CHILD_OF"
                },
                {
                    "traceId": "1ff8b4c80f98b9e12420e27955c77e2d",
                    "spanId": "3dc5e2497b5f7ba5",
                    "refType": "FOLLOWS_FROM"
                }
            ],
 "event": [
                {
                    "name": "request_received",
                    "timeUnixNano": 1778489782759245000,
                    "attributeMap": {
                        "http.method": "GET",
                        "http.route": "/api/chat"
                    }
                },
                {
                    "name": "cache_lookup",
                    "timeUnixNano": 1778489782811697000,
                    "attributeMap": {
                        "cache.hit": "true",
                        "cache.key": "user:123:prompt"
                    }
                },
                {
                    "name": "model_invoked",
                    "timeUnixNano": 1778489782863268000,
                    "attributeMap": {
                        "gen_ai.request.model": "gpt-4",
                        "gen_ai.usage.input_tokens": "100"
                    }
                },
                {
                    "name": "response_sent",
                    "timeUnixNano": 1778489782863273000,
                    "attributeMap": {
                        "http.status_code": "200",
                        "response.size_bytes": "2048"
                    }
                }
            ]
}
```
Also it renames links to references.


Since it makes sense to not send them as it is stored in the db, I am thinking to modify/parse them at the consume layer where we are merging the attributes right now i.e attributes_string, attributes_number etc.

Also need suggestions on the structure of events and links, if you see the above response from waterfall it's kind of different from what the actual event looks since we have keys like `attributeMap` in the response. So thinking of something like
```
 type Event struct {
      Name                   string         `json:"name"`
      TimeUnixNano           uint64         `json:"timeUnixNano"`
      Attributes             map[string]any `json:"attributes,omitempty"`
  }
  
type Link struct {
	TraceId string `json:"traceId,omitempty"`
	SpanId  string `json:"spanId,omitempty"`
	RefType string `json:"refType,omitempty"`
}
```

let me know your thoughts here @srikanthccv @tushar-signoz

srikanthccv: @nityanandagohain 

>Since it makes sense to not send them as it is stored in the db, I am thinking to modify/parse them at the consume layer where we are merging the attributes right now i.e attributes_string, attributes_number etc.


ok, this sounds fine based on what we do already and given the db state is stringified json

> Also need suggestions on the structure of events and links

>

```
type Event struct {
      Name                   string         `json:"name"`
      TimeUnixNano           uint64         `json:"timeUnixNano"`
      Attributes             map[string]any `json:"attributes,omitempty"`
  }
```

+1, please add the list of `events`

```
type Link struct {
	TraceId string `json:"traceId,omitempty"`
	SpanId  string `json:"spanId,omitempty"`
	RefType string `json:"refType,omitempty"`
}
```

The `refType` comes from the jaeger era. Otel doesn't have it. Please remove it and send the span and trace id only. It can always be added later if needed.

nityanandagohain: Since we are already updating this, returning them as `events` and `links`. Kept the structs as you suggested.

nityanandagohain:
One thing I missed was updating trace operator as well to return everything, have added the changes now with tests. @srikanthccv  @tushar-signoz

nityanandagohain:
@tushar-signoz please reivew

tushar-signoz (approved):
LGTM

pkg/types/spantypes/event.go
```diff
@@ -0,0 +1,54 @@
...
+	IsError      bool           `json:"isError,omitempty"`
+}
+
+// Link is the response shape for a span link.
+// The refType field is intentionally not decoded; it's a Jaeger-era
+// concept that OTel doesn't model, so we drop it on the way out.
+type Link struct {
+	TraceID string `json:"traceId,omitempty"`
+	SpanID  string `json:"spanId,omitempty"`
+}
```
krsoninikhil, line 19: These types (Event and Link) are defined [here](https://github.com/SigNoz/signoz/blob/6f79d6b18dedf3bbdf638a7d8836fd3dde295e48/pkg/types/spantypes/waterfall_span.go#L42) if they can be used.

nityanandagohain: Yeah I have updated them here on what we want to send to the UI, so let it be now >

EventV2/Link is different from event/Link in few ways eg:- we call it attribues here, RefType is removed.

Here is the discussion https://github.com/SigNoz/signoz/pull/10583#discussion_r3253180894

pkg/querier/consume.go
```diff
@@ -452,6 +453,53 @@ func readAsRaw(rows driver.Rows, queryName string) (*qbtypes.RawData, error) {
...
 
+// mergeSpanAttributeColumns merges (attributes_string, attributes_number, attributes_bool, resources_string) into
+// unified "attributes" and "resource" keys, and parses the stringified `events`
+// and `links` columns into structured slices. Raw DB columns are removed.
+func mergeSpanAttributeColumns(data map[string]any) {
+	attrStr, hasStr := data["attributes_string"]
+	attrNum, hasNum := data["attributes_number"]
+	attrBool, hasBool := data["attributes_bool"]
+	// todo(nitya): move to resource json
+	resStr, hasRes := data["resources_string"]
```
krsoninikhil, line 464: @nityanandagohain why are we not doing this in this PR itself? Is there any challenge in using resource json field?

nityanandagohain: * Evolutions has to be taken care off here.
* The json has a nested data instead of flattened keys.

Because of these two, kep it out of scope for this PR.

srikanthccv (approved):

therealpandey (approved):

PR #10603, opened 2026-03-16
refactor: reorganize user and role types into dedicated packages

karanbalani:
## Summary

- **Move user-related types into `pkg/types/usertypes`** — `User`, `FactorPassword`, `Invite`, `StorableAPIKey`, `ResetPasswordToken`, and all associated request/response types, error codes, and status constants previously scattered in `pkg/types` now live in a dedicated `usertypes` package, consistent with how other domain types (`dashboardtypes`, `tracefunneltypes`, etc.) are organized.

- **Consolidate role types into `pkg/types/authtypes`** — `Role`, `StorableRole`, `RoleStore` (previously `Store`), managed role constants, and helper functions from `pkg/types/roletypes` are merged into `authtypes`, since roles are fundamentally an auth concern. The `roletypes` package is removed.

- **Rename `types.Role` to `authtypes.LegacyRole`** — The legacy uppercase string role type (`"ADMIN"`, `"EDITOR"`, `"VIEWER"`) is moved from `pkg/types/role.go` to `pkg/types/authtypes/legacy_role.go` and renamed to `LegacyRole` to clearly distinguish it from the newer RBAC `Role` type and signal that it should not be used for new code.

- **Move `GetActiveUserAndFactorPasswordByEmailAndOrgID` from `AuthNStore` to `user.Getter`** — This user+password lookup was the only user-data method on the authn store interface. It now lives behind `user.Getter`, and `emailpasswordauthn` depends on the user module's read interface rather than reaching into the authn store for user data. `AuthNStore` is left with only auth-domain concerns (`GetAuthDomainFromID`).

- **Update OpenAPI spec and frontend generated types** — Regenerated `docs/api/openapi.yml` and the frontend generated API client to reflect the `roletypes` → `authtypes` rename (e.g. `RoletypesRoleDTO` → `AuthtypesRoleDTO`, `RoletypesPostableRoleDTO` → `AuthtypesPostableRoleDTO`). Updated manual frontend references in `CreateRoleModal`, `RolesListingTable`, and mock data accordingly.

## Notes

This is a **pure refactoring** with no behavioral changes. All type definitions, SQL queries, error codes, validation logic, and business rules remain identical — only package locations, import paths, and the dependency wiring for email/password authentication have changed.

## Test plan

- [x] `go build ./...` compiles successfully
- [x] All existing unit and integration tests pass
- [x] Verify login flow (email/password authentication) works end-to-end
- [x] Verify invite creation and acceptance flow works
- [x] Verify API key CRUD operations work
- [x] Verify role assignment/modification works

### DIFF
--- a/pkg/identn/apikeyidentn/identn.go
+++ b/pkg/identn/apikeyidentn/identn.go
@@ -9,8 +9,8 @@ import (
 	"github.com/SigNoz/signoz/pkg/factory"
 	"github.com/SigNoz/signoz/pkg/identn"
 	"github.com/SigNoz/signoz/pkg/sqlstore"
-	"github.com/SigNoz/signoz/pkg/types"
 	"github.com/SigNoz/signoz/pkg/types/authtypes"
+	"github.com/SigNoz/signoz/pkg/types/usertypes"
 	"golang.org/x/sync/singleflight"
 )
 
@@ -64,7 +64,7 @@ func (r *resolver) GetIdentity(req *http.Request) (*authtypes.Identity, error) {
 		return nil, errors.New(errors.TypeUnauthenticated, errors.CodeUnauthenticated, "missing api key")
 	}
 
-	var apiKey types.StorableAPIKey
+	var apiKey usertypes.StorableAPIKey
 	err := r.store.
 		BunDB().
 		NewSelect().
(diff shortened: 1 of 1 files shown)


karanbalani:
AI (Opus 4.6) Assisted in verifying the code changes and ensuring no behaviour changes are pushed.
Purely Organizational PR.

karanbalani:
Part of: https://github.com/SigNoz/platform-pod/issues/1439

pkg/types/usertypes/user.go
```diff
@@ -1,4 +1,4 @@
-package types
```
therealpandey, line 1: Please let user and org be at the root as we had discussed.

karanbalani: That will create a import cycle, that is why this was moved.

`authtypes` import `types` for Identifiable and TimeAuditable, then `user.go` (which is being asked to move to  types) also imports `authtypes` for role.

PR #10607, opened 2026-03-17
fix: adding migration to fix wrong index on cloud integration table

swapnil-signoz:
## Pull Request

Fixes: https://github.com/SigNoz/engineering-pod/issues/4224

---

### 📄 Summary
Added migration to fix the wrong unique index on `cloud_integration` table. More info in the [issue](https://github.com/SigNoz/engineering-pod/issues/4224).

### Approach (AI)
Migration steps (within a single transaction):

  1. Drop wrong index unique_cloud_integration on (id, provider, org_id) using sqlschema.Operator().DropIndex() with Named()
  2. Normalize empty account_id to NULL — UPDATE cloud_integration SET account_id = NULL WHERE account_id = '' (older table structure could store "" instead of
  NULL; empty strings would violate the partial unique index since '' = '' unlike NULL != NULL)
  3. Fetch active rows with non-null account_id, ordered for grouping by (account_id, provider, org_id, updated_at DESC)
  4. Group duplicates in Go — keeper = most recently updated, rest = losers
  5. Merge config — union of EnabledRegions from all duplicates into keeper's config
  6. Reassign services — delete conflicting services from losers (same type as keeper), reassign remaining to keeper
  7. Soft-delete losers — set removed_at and updated_at
  8. Create partial unique index on (account_id, provider, org_id) WHERE removed_at IS NULL using sqlschema.PartialUniqueIndex

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/4224

### DIFF
--- a/pkg/sqlmigration/069_update_cloud_integration_index.go
+++ b/pkg/sqlmigration/069_update_cloud_integration_index.go
@@ -14,16 +14,16 @@ import (
 	"github.com/uptrace/bun/migrate"
 )
 
-type fixCloudIntegrationUniqueIndex struct {
+type updateCloudIntegrationUniqueIndex struct {
 	sqlstore  sqlstore.SQLStore
 	sqlschema sqlschema.SQLSchema
 }
 
-func NewFixCloudIntegrationUniqueIndexFactory(sqlstore sqlstore.SQLStore, sqlschema sqlschema.SQLSchema) factory.ProviderFactory[SQLMigration, Config] {
+func NewUpdateCloudIntegrationUniqueIndexFactory(sqlstore sqlstore.SQLStore, sqlschema sqlschema.SQLSchema) factory.ProviderFactory[SQLMigration, Config] {
 	return factory.NewProviderFactory(
-		factory.MustNewName("fix_cloud_integration_index"),
+		factory.MustNewName("update_cloud_integration_index"),
 		func(ctx context.Context, ps factory.ProviderSettings, c Config) (SQLMigration, error) {
-			return &fixCloudIntegrationUniqueIndex{
+			return &updateCloudIntegrationUniqueIndex{
 				sqlstore:  sqlstore,
 				sqlschema: sqlschema,
 			}, nil
@@ -31,7 +31,7 @@ func NewFixCloudIntegrationUniqueIndexFactory(sqlstore sqlstore.SQLStore, sqlsch
 	)
 }
 
-func (migration *fixCloudIntegrationUniqueIndex) Register(migrations *migrate.Migrations) error {
+func (migration *updateCloudIntegrationUniqueIndex) Register(migrations *migrate.Migrations) error {
 	if err := migrations.Register(migration.Up, migration.Down); err != nil {
 		return err
 	}
@@ -60,7 +60,7 @@ type duplicateGroup struct {
 	losers []*cloudIntegrationRow
 }
 
-func (migration *fixCloudIntegrationUniqueIndex) Up(ctx context.Context, db *bun.DB) error {
+func (migration *updateCloudIntegrationUniqueIndex) Up(ctx context.Context, db *bun.DB) error {
 	tx, err := db.BeginTx(ctx, nil)
 	if err != nil {
 		return err
@@ -69,6 +69,8 @@ func (migration *fixCloudIntegrationUniqueIndex) Up(ctx context.Context, db *bun
 		_ = tx.Rollback()
 	}()
 
+	sqls := [][]byte{}
+
 	// Step 1: Drop the wrong index on (id, provider, org_id)
 	dropSqls := migration.sqlschema.Operator().DropIndex(
 		(&sqlschema.UniqueIndex{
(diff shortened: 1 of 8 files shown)


swapnil-signoz:
@vikrantgupta25 my tenant ran the migration successfully.
```
CREATE TABLE IF NOT EXISTS "cloud_integration" ("id" text NOT NULL, "created_at" TIMESTAMP, "updated_at" TIMESTAMP, "provider" text, "config" text, "account_id" text, "last_agent_report" text, "removed_at" timestamp, "org_id" text, PRIMARY KEY ("id"), FOREIGN KEY ("org_id") REFERENCES "organizations" ("id"));
CREATE UNIQUE INDEX "puq_cloud_integration_account_id_provider_org_id_67f5b2fc" ON "cloud_integration" ("account_id", "provider", "org_id") WHERE removed_at IS NULL;
```

It did not have duplicate scenario per say, but I have tested it locally for sqlite and postgres both. It works fine.

pkg/sqlmigration/069_fix_cloud_integration_index.go
```diff
@@ -0,0 +1,264 @@
...
+	"github.com/uptrace/bun"
+	"github.com/uptrace/bun/migrate"
+)
+
+type fixCloudIntegrationUniqueIndex struct {
+	sqlstore  sqlstore.SQLStore
+	sqlschema sqlschema.SQLSchema
+}
+
+func NewFixCloudIntegrationUniqueIndexFactory(sqlstore sqlstore.SQLStore, sqlschema sqlschema.SQLSchema) factory.ProviderFactory[SQLMigration, Config] {
```
vikrantgupta25:
```suggestion
func NewUpdateCloudIntegrationUniqueIndexFactory(sqlstore sqlstore.SQLStore, sqlschema sqlschema.SQLSchema) factory.ProviderFactory[SQLMigration, Config] {
```

pkg/sqlmigration/069_fix_cloud_integration_index.go
```diff
@@ -0,0 +1,264 @@
...
+
+		// Step 4: Merge config from losers into keeper
+		if err = mergeCloudIntegrationConfigs(ctx, tx, group); err != nil {
+			return err
+		}
+
+		// Step 5: Reassign orphaned cloud_integration_service rows
+		for _, loser := range group.losers {
+			// Delete services from loser that would conflict with keeper's services (same type)
+			_, err = tx.NewDelete().
```
vikrantgupta25: for this if we just update the cloud integration id in services and just add `ON CONFLICT DO NOTHING` it would work right ? 

```
CONSTRAINT "cloud_integration_id_type" UNIQUE ("type", "cloud_integration_id"),
```
since there is already a unique constraint

pkg/sqlmigration/069_update_cloud_integration_index.go
```diff
@@ -0,0 +1,264 @@
...
+				return err
+			}
+
+			loserIDs = append(loserIDs, loser.ID)
+		}
+	}
+
+	// Step 6: Soft-delete all loser rows
+	if len(loserIDs) > 0 {
+		_, err = tx.NewUpdate().
```
vikrantgupta25, line 146: the pending rows for which we couldn't update the cloud integration id because of the above conflict will be deleted here because we have 

```
REFERENCES "cloud_integration" ("id") ON DELETE CASCADE);
```

pkg/sqlmigration/069_update_cloud_integration_index.go
```diff
@@ -0,0 +1,264 @@
...
+	tx, err := db.BeginTx(ctx, nil)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = tx.Rollback()
+	}()
+
+	// Step 1: Drop the wrong index on (id, provider, org_id)
+	dropSqls := migration.sqlschema.Operator().DropIndex(
```
vikrantgupta25, line 75: can we club the exec for this index with the exec for the new create index ?

pkg/sqlmigration/069_update_cloud_integration_index.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	merged := cloudIntegrationAccountConfig{Regions: mergedRegions}
+	mergedJSON, err := json.Marshal(merged)
+	if err != nil {
+		return err
+	}
+
+	// Update keeper's config
+	_, err = tx.NewUpdate().
```
vikrantgupta25, line 233: let's also do a bulk update for this as well please

pkg/sqlmigration/069_fix_cloud_integration_index.go
swapnil-signoz: Which essentially means, we wont delete the conflicting rows.
But, `ON CONFLICT DO NOTHING` is supported on INSERT and not UPDATE queries.
So current approach seems right to me

pkg/sqlmigration/069_update_cloud_integration_index.go
swapnil-signoz: I think it will happen when there is hard delete, and not the soft delete. So cloud integration service rows will stay. Also the conflicting cloud integration service rows gets hard deleted and the non conflicting gets updated, so no rows from older cloud integration account are present anyways.

pkg/sqlmigration/069_update_cloud_integration_index.go
swapnil-signoz: You mean alter the sequence for the operations?
1. Drop older index
2. Add new index
.... rest of operations
If yes, I think we shouldn't as it will disturb the mental model for this flow of operations.
Also, if there are duplicate rows operation will fail.

pkg/sqlmigration/069_update_cloud_integration_index.go
swapnil-signoz: Lets discuss this offline.

PR #10609, opened 2026-03-17
docs: perses schema for dashboards

namanv3:
## Pull Request

---

### 📄 Summary
Perses schema for Signoz Dashboards. Doc: https://www.notion.so/signoz/RFC-Perses-Native-Dashboard-Schema-for-SigNoz-30afcc6bcd198086bc44e65b56303689?source=copy_link

#### Issues closed by this PR

Closes https://github.com/SigNoz/engineering-pod/issues/4172
Closes https://github.com/SigNoz/engineering-pod/issues/4309

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 📝 Changelog

**N/A**

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

go minor version update is because of the perses package


namanv3:
From https://github.com/perses/plugins:
Screenshot 2026-03-19 at 00 27 10

`BarChart`, `PieChart`, `HistogramChart`, `HeatMapChart`, `StatChart`, `Table`, and ofc `TimeSeries`: they only support `TimeSeriesQuery`
`LogsTable` supports only `LogQuery`, and `TraceTable` supports only `TraceQuery`

perses/schemas/panels/signoz-time-series-chart/signoz-time-series-chart.cue
```diff
@@ -0,0 +1,54 @@
+package model
+
+import (
+    tsModel "github.com/perses/plugins/timeserieschart/schemas@v0:model"
+)
+
+kind: "SigNozTimeSeriesChart"
+spec: close({
+    upstream:              tsModel.spec
```
namanv3: even if we use the upstream spec, we still have enough differences that so many fields were required to be added separately. A lot of them are because the field upstream was not enough. I think not using the upstream spec at all is a good idea so that we are not bound by it.
The value of perses can be equal to it being a validator that we use for our schemas.

srikanthccv: > I think not using the upstream spec at all is a good idea so that we are not bound by it.

Sounds good to me.

perses/schemas/panels/signoz-time-series-chart/signoz-time-series-chart.cue
```diff
@@ -0,0 +1,54 @@
...
+    fillMode?:             #FillMode
+    fillSpans?:            bool
+    decimalPrecision?:     number & >=0 & <=6
+    mergeAllActiveQueries?: bool
+    contextLinks?:         #ContextLinks
+    customLegendColors?:   [string]: =~"^#(?:[0-9a-fA-F]{3}){1,2}$"
+    // soft limits — axis can extend beyond these if data exceeds,
+    // unlike upstream yAxis.min/max which are hard limits
+    softMin?:              number
+    softMax?:              number
```
srikanthccv: A lot of these are randomly populated by fronted on ad-hoc basis

- some of the don't make sense for the time series panel type
- some of the them don't apply and relevant to end user
- organization is flat rather than cohesive grouping based on sub-part


Let's write down the spec for each of `SigNozX` panel types.

srikanthccv (commented):
Alright, let's not use upstream schema. Please take some time to define our own chart types schema.

perses/schemas/panels/signoz-time-series-chart/signoz-time-series-chart.cue
namanv3: Done

srikanthccv (commented):
First pass


- Add a schemaVersion field
  - When these schemas evolve (which they will), consumers will have no way to know which schema version a given dashboard targets.
- `spec: close({...})`
  - If I understand this correctly, this disallows any unknown fields?, what does this mean in terms of adding new fields to the Go structs? Is there a potential place for unknown new fields or should the schema be updated along with the go types? Can they be synced automatically somehow?
- How does one go about adding a new panel type? Can you please write down the steps for it?
- Rename `yAxisUnit` to `unit` on panels without axes, or just use `unit` uniformly
- Normalize enum casing (we do this on the backend)
  - #VariableSortOrder: "DISABLED" | "ASC" | "DESC"
  - #TimePreference: "GLOBAL_TIME" | "LAST_5_MIN"
  - #ReduceTo: "sum" | "count" | "avg"
  - #JoinType: "inner" | "left" | "right"
  - #LegendPosition: "bottom" | "right"
  - thresholdFormat: "Text" | "Background"
  - thresholdColor: "Red" | "Orange"
- Drop redundant `threshold` prefix from threshold fields
  - The prefix looks like noise carried over from the previous all things flat widget format
- Mixed use of `Chart` and `Panel`
- Let's extract shared definitions into common packages (ex, `#TimePreference`, `#PrecisionOption`, `#Formatting`, `#ContextLinks`, `#Threshold`). This is true for queries also, let's use the cue module system (which you already did partially in some defs).
- Is it deliberate to leave the nested structures open?
- `{"contextLinks": {"linksData": [...]}}` vs `contextLinks?: [...#ContextLink]`
  - Will there be any property (now or in future) that justifies the wrapper?
- `selectFields` is untyped
  - It should `[]telemetrytypes.TelemetryFieldKey`. The list panel already has a `#TelemetryFieldKey` definition for the same type. Please use it.

```
  spec: close({
      dynamicVariablesAttribute: string
      dynamicVariablesSource:    string
  })
```

- Let's just use `name` and `source`, because they're already scoped inside. Avoids the verbosity and plural

- Threshold definitions are inconsistent across panels, should it rather be one definition with optionaliy?
- Formula is missing functions
- Let's remove the `aggregations` and `expressionAggregations`, just have discriminated union
- Let's not include the feature that we don't support in the backend, example join, sub-query etc...
- The name field is missing in `Function`
- `Step` is custom type in go but number in CUE, how to address such things?

namanv3:
@srikanthccv 

> When these schemas evolve (which they will), consumers will have no way to know which schema version a given dashboard targets.

Earlier each folder (panels/variables/queries) had their own version. Each folder must have its own package.json, so I combined all the folders so that there is only one package.json with one version.

> If I understand this correctly, this disallows any unknown fields?, what does this mean in terms of adding new fields to the Go structs? Is there a potential place for unknown new fields or should the schema be updated along with the go types? Can they be synced automatically somehow?

If the `close` is removed then unknown fields can be added. There is a way to generate go structs from the schema, I haven't looked at how that works but that would be the way to sync schema and go. Should I look at it as a part of this task itself or is that a future ticket?

> How does one go about adding a new panel type? Can you please write down the steps for it?

Currently it is as simple as just defining the cue file. I think the full steps will be required (and actually make sense to note down) once we work on the frontend/backend consumption of this schema

> Normalize enum casing

Didn't get what you mean by normalise here

> {"contextLinks": {"linksData": [...]}} vs contextLinks?: [...#ContextLink]

I had kept it similar to what UI currently has. Removed the nesting now

> Let's just use name and source, because they're already scoped inside. Avoids the verbosity and plural

Using `attributeName` cuz it should be clear that we're using the name of an attribute here

> Threshold definitions are inconsistent across panels, should it rather be one definition with optionaliy?

But this will lead to the definition being too loose and chances of fields coming where they shouldn't. This hold true for other types in panels like Legend that I haven't moved to commons

> Step is custom type in go but number in CUE, how to address such things?

Updated the type, I had missed that it can be a duration string.

srikanthccv:
> Should I look at it as a part of this task itself or is that a future ticket?


Let's keep it as future ticket.

>Currently it is as simple as just defining the cue file. I think the full steps will be required (and actually make sense to note down) once we work on the frontend/backend consumption of this schema


Right, let's how does one go about writing this for heatmap? We should have the guidelines so the people who implement it will have some concrete reference guide.

>Didn't get what you mean by normalise here

There is mixed use of the capital case and small case, example `#VariableSortOrder: *"DISABLED" | "ASC" | "DESC"`.

> Using attributeName cuz it should be clear that we're using the name of an attribute here

While currently we only support the attributes, It should not limited to only the attributes of the span/log/metric record. One should be able to create variable from the intrinsic fields of span/log/metric, example span name, kind etc...

>But this will lead to the definition being too loose and chances of fields coming where they shouldn't. This hold true for other types in panels like Legend that I haven't moved to commons

Would it make sense to define types based on where they applicable. For instance, what applies for line/bar chart don't apply to table/number, then should there be a category of the threshold that can be plugged in depending on the chart type in question?

srikanthccv:
@AshwinBhatkal bump ^ please review

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,541 @@
...
+	VariableKindDynamic VariablePluginKind = "SigNozDynamicVariable"
+	VariableKindQuery   VariablePluginKind = "SigNozQueryVariable"
+	VariableKindCustom  VariablePluginKind = "SigNozCustomVariable"
+	VariableKindTextbox VariablePluginKind = "SigNozTextboxVariable"
+)
+
+type DynamicVariableSpec struct {
+	// Name is the name of the attribute being fetched dynamically from the
+	// source. This could be extended to a richer selector in the future.
+	Name   string `json:"name" validate:"required"`
```
AshwinBhatkal: can we also have required:"true" here?

AshwinBhatkal: type generation from open api requires them

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,541 @@
...
+	"github.com/SigNoz/signoz/pkg/types/telemetrytypes"
+)
+
+// ══════════════════════════════════════════════
+// SigNoz variable plugin specs
+// ══════════════════════════════════════════════
+
+type VariablePluginKind = string
+
+const (
```
AshwinBhatkal, line 19: these variables must be extending some common struct right? Which one is it?

namanv3: this is their type, `type VariablePluginKind = string`, not extending anything.

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: done

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,541 @@
...
+	"github.com/SigNoz/signoz/pkg/errors"
+	qb "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
+	"github.com/SigNoz/signoz/pkg/types/telemetrytypes"
+)
+
+// ══════════════════════════════════════════════
+// SigNoz variable plugin specs
+// ══════════════════════════════════════════════
+
+type VariablePluginKind = string
```
srikanthccv: use `valuer.String` instead for enums.

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,541 @@
...
+	VariableKindQuery   VariablePluginKind = "SigNozQueryVariable"
+	VariableKindCustom  VariablePluginKind = "SigNozCustomVariable"
+	VariableKindTextbox VariablePluginKind = "SigNozTextboxVariable"
+)
+
+type DynamicVariableSpec struct {
+	// Name is the name of the attribute being fetched dynamically from the
+	// source. This could be extended to a richer selector in the future.
+	Name   string `json:"name" validate:"required" required:"true"`
+	Source string `json:"source" validate:"required" required:"true"`
```
srikanthccv:
```suggestion
	Signal string `json:"signal"`
```

Let's use signal to refer to this 1. it will remain consistent with our naming 2. maps exactly how it's used in the dependent APIs.. And it's not mandatory, when not provided it uses the all the signal to return the data.

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,541 @@
...
+	CustomValue string `json:"customValue" validate:"required" required:"true"`
+}
+
+type TextboxVariableSpec struct{}
+
+// ══════════════════════════════════════════════
+// SigNoz query plugin specs — aliased from querybuildertypesv5
+// ══════════════════════════════════════════════
+
+type QueryPluginKind = string
```
srikanthccv: same. use `valuer.String` for enums, if you are looking for example, see how enums are defined in querybuildertypesv5

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,541 @@
...
+	MetricBuilderQuerySpec = qb.QueryBuilderQuery[qb.MetricAggregation]
+	LogBuilderQuerySpec    = qb.QueryBuilderQuery[qb.LogAggregation]
+	TraceBuilderQuerySpec  = qb.QueryBuilderQuery[qb.TraceAggregation]
+)
+
+type BuilderQuerySpec struct {
+	Spec any
+}
+
+func (b *BuilderQuerySpec) UnmarshalJSON(data []byte) error {
```
srikanthccv, line 81: Is there a reason to define this here instead of relying/using https://github.com/SigNoz/signoz/blob/7279c5f770131074d3ea7a2bbe85e88130a8be86/pkg/types/querybuildertypes/querybuildertypesv5/builder_query.go#L130-L177

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,541 @@
...
+		return nil
+	}
+	var n int
+	if err := json.Unmarshal(data, &n); err == nil {
+		switch n {
+		case 0, 1, 2, 3, 4:
+			p.value = n
+			return nil
+		default:
+			return errors.NewInvalidInputf(ErrCodeDashboardInvalidInput, "invalid precision option %d: must be 0, 1, 2, 3, or 4", n)
```
srikanthccv: Are they not allowed to define precision option as string value?

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,541 @@
...
+	}
+	var n float64
+	if err := json.Unmarshal(data, &n); err == nil {
+		if n < 0 {
+			return errors.NewInvalidInputf(ErrCodeDashboardInvalidInput, "invalid spanGaps %v: numeric value must be non-negative", n)
+		}
+		sg.value = n
+		return nil
+	}
+	return errors.NewInvalidInputf(ErrCodeDashboardInvalidInput, "invalid spanGaps: must be a bool or a non-negative number")
```
srikanthccv: I think it should allow go duration values `1h`, `5m` etc...

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,541 @@
...
+// ComparisonOperator: ">" | "<" | ">=" | "<=" | "=".
+type ComparisonOperator string
+
+const (
+	ComparisonOperatorGT  ComparisonOperator = ">"
+	ComparisonOperatorLT  ComparisonOperator = "<"
+	ComparisonOperatorGTE ComparisonOperator = ">="
+	ComparisonOperatorLTE ComparisonOperator = "<="
+	ComparisonOperatorEQ  ComparisonOperator = "="
+)
```
srikanthccv: Let's support literals as the enums while also supporting the symbols where possible, similar to this https://github.com/SigNoz/signoz/blob/7279c5f770131074d3ea7a2bbe85e88130a8be86/pkg/types/ruletypes/compare.go#L14

Along the same lines as the sort option upstream spec.

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,541 @@
...
+	case LegendPositionBottom, LegendPositionRight:
+		*l = LegendPosition(v)
+		return nil
+	default:
+		return errors.NewInvalidInputf(ErrCodeDashboardInvalidInput, "invalid legend position %q: must be bottom or right", v)
+	}
+}
+
+// ThresholdFormat: "Text" | "Background".
+type ThresholdFormat string
```
srikanthccv: valuer.String, and seems there is mix of small case and capital cast start letter for enum?

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,541 @@
...
+	DecimalPrecision PrecisionOption   `json:"decimalPrecision"`
+}
+
+type Legend struct {
+	Position     LegendPosition    `json:"position"`
+	CustomColors map[string]string `json:"customColors"`
+}
+
+type ThresholdWithLabel struct {
+	Value float64 `json:"value"`
```
srikanthccv: value should also be required here?

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,541 @@
...
+	Unit  string  `json:"unit"`
+	Color string  `json:"color" validate:"required" required:"true"`
+	Label string  `json:"label" validate:"required" required:"true"`
+}
+
+type ComparisonThreshold struct {
+	Value    float64            `json:"value"`
+	Operator ComparisonOperator `json:"operator"`
```
srikanthccv: value and operator should be required?

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,541 @@
...
+	DatasourceKindSigNoz DatasourcePluginKind = "SigNozDatasource"
+)
+
+type TimeSeriesPanelSpec struct {
+	Visualization   TimeSeriesVisualization   `json:"visualization"`
+	Formatting      PanelFormatting           `json:"formatting"`
+	ChartAppearance TimeSeriesChartAppearance `json:"chartAppearance"`
+	Axes            Axes                      `json:"axes"`
+	Legend          Legend                    `json:"legend"`
+	Thresholds      []ThresholdWithLabel      `json:"thresholds" validate:"dive"`
```
srikanthccv, line 126: Links is missing

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,541 @@
...
+
+type HistogramBuckets struct {
+	BucketCount           *float64 `json:"bucketCount"`
+	BucketWidth           *float64 `json:"bucketWidth"`
+	MergeAllActiveQueries bool     `json:"mergeAllActiveQueries"`
+}
+
+type ListPanelSpec struct {
+	SelectedLogFields    []LogField                         `json:"selectedLogFields" validate:"dive"`
+	SelectedTracesFields []telemetrytypes.TelemetryFieldKey `json:"selectedTracesFields"`
```
srikanthccv: Let use this https://github.com/SigNoz/signoz/blob/7279c5f770131074d3ea7a2bbe85e88130a8be86/pkg/types/querybuildertypes/querybuildertypesv5/builder_query.go#L40

pkg/types/dashboardtypes/dashboard_v2.go
srikanthccv, line 1: should we disallow all unknown fields https://brandur.org/disallow-unknown-fields?

pkg/types/dashboardtypes/dashboard_v2.go
```diff
@@ -0,0 +1,210 @@
...
+	data, err := json.Marshal(v)
+	if err != nil {
+		return nil, err
+	}
+	var raw struct {
+		Spec struct {
+			Plugin *common.Plugin `json:"plugin,omitempty"`
+		} `json:"spec"`
+	}
+	if err := json.Unmarshal(data, &raw); err != nil {
```
srikanthccv: Why do we have to do this roundtrip and local struct?

pkg/types/dashboardtypes/dashboard_v2.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// Validate variable plugins (only ListVariables have plugins; TextVariables do not).
+	for i, v := range d.Variables {
+		plugin, err := extractPluginFromVariable(v)
+		if err != nil {
+			return errors.WrapInvalidInputf(err, ErrCodeDashboardInvalidInput, "spec.variables[%d]", i)
+		}
+		if plugin == nil {
+			continue
+		}
```
srikanthccv: Can `plugin` be nil in any valid scenario?

pkg/types/dashboardtypes/dashboard_v2.go
```diff
@@ -0,0 +1,210 @@
...
+	if plugin.Kind == "" {
+		return errors.NewInvalidInputf(ErrCodeDashboardInvalidInput, "%s: plugin kind is required", path)
+	}
+	factory, ok := specs[plugin.Kind]
+	if !ok {
+		return errors.NewInvalidInputf(ErrCodeDashboardInvalidInput, "%s: unknown plugin kind %q", path, plugin.Kind)
+	}
+	if plugin.Spec == nil {
+		return nil
+	}
```
srikanthccv, line 206: This is generic and leads to early return and invalid json in many cases. example it wouldn't make sense to have nil plugin for dynamic variable.

pkg/types/dashboardtypes/dashboard_v2.go
```diff
@@ -0,0 +1,210 @@
...
+	return nil
+}
+
+func validatePlugin(plugin common.Plugin, specs map[string]func() any, path string) error {
+	if plugin.Kind == "" {
+		return errors.NewInvalidInputf(ErrCodeDashboardInvalidInput, "%s: plugin kind is required", path)
+	}
+	factory, ok := specs[plugin.Kind]
+	if !ok {
+		return errors.NewInvalidInputf(ErrCodeDashboardInvalidInput, "%s: unknown plugin kind %q", path, plugin.Kind)
```
srikanthccv: include what values are allowed when a field is validated, in this case, what pugin kinds are allowed.

pkg/types/dashboardtypes/dashboard_v2.go
```diff
@@ -0,0 +1,210 @@
...
+		return errors.WrapInvalidInputf(err, ErrCodeDashboardInvalidInput, "%s.spec", path)
+	}
+	target := factory()
+	if err := json.Unmarshal(specJSON, target); err != nil {
+		return errors.WrapInvalidInputf(err, ErrCodeDashboardInvalidInput, "%s.spec", path)
+	}
+	if err := validator.New().Struct(target); err != nil {
+		return errors.WrapInvalidInputf(err, ErrCodeDashboardInvalidInput, "%s.spec", path)
+	}
+	return nil
```
srikanthccv, line 223: There needs to be additional validations. Where are they covered.  For instance, the `spec.plugin` of `Vairable` has the `signal` field and only allowed values are [`logs`, `metrics`, `traces`]. Such validations should also be performed. This can be said about any other fields where values are expected to in known set. Are such validations already happening? I couldn't see that.

pkg/types/dashboardtypes/dashboard_v2.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	datasourcePluginSpecs = map[string]func() any{
+		DatasourceKindSigNoz: func() any { return new(struct{}) },
+	}
+
+	// allowedQueryKinds maps each panel plugin kind to the query plugin
+	// kinds it supports. Composite sub-query types are mapped to these
+	// same kind strings via compositeSubQueryTypeToPluginKind.
+	allowedQueryKinds = map[string][]string{
+		PanelKindTimeSeries: {QueryKindBuilder, QueryKindComposite, QueryKindFormula, QueryKindTraceOperator, QueryKindPromQL, QueryKindClickHouseSQL},
```
srikanthccv, line 83: Does this mean a panel queries could have just a formula kind and be valid?

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: context links are not needed in the plugin cuz perses panels have a field for them.

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: done

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: changed it to string for simplicity, should we support both number and string?

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: `ruletypes.CompareOperator` does not throw an error when something other than the defined values is passed. So I didn't use it but instead made a copy. Added the literal options however.

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: done

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: done

pkg/types/dashboardtypes/dashboard_v2.go
namanv3: done

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: done for this enum and other plugin kinds

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: done for this enum and other plugin kinds

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: Done

pkg/types/dashboardtypes/dashboard_v2.go
namanv3: yes this should be an error actually. fixed

pkg/types/dashboardtypes/dashboard_v2.go
namanv3: text variables have no plugin so it is nil there. fixed the logic to make this explicit

pkg/types/dashboardtypes/dashboard_v2.go
namanv3: simplified to not require this anymore

pkg/types/dashboardtypes/dashboard_v2.go
namanv3: Schema wise yes, I thought this check would actually be too deep for just schema so I did not add it. Do you think it should be added here?

pkg/types/dashboardtypes/dashboard_v2.go
namanv3: the validation for enums is covered in their custom unmarshalls. Signal was a plain string so specifically that wasn't being validated, so I am using an enum for it now

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: fixed the capital letter. I am not using valuer.String here and in other enums because it does not support default values

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: the existing unmarshall will be called by this method right? This new method only figures out which type T to use

namanv3: this code was copied from https://github.com/SigNoz/signoz/blob/7279c5f770131074d3ea7a2bbe85e88130a8be86/pkg/types/querybuildertypes/querybuildertypesv5/req.go#L101 
made a change to extract this logic out into a helper method

pkg/types/dashboardtypes/dashboard_v2.go
namanv3: the unknown fields won't actually be stored in the db (because they will go away after the deserialisation) so I thought they don't harm in any way, and hence didn't disallow.

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: makes sense. also I changed the type to a proper struct

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,553 @@
...
+// SigNoz variable plugin specs
+// ══════════════════════════════════════════════
+
+type VariablePluginKind struct{ valuer.String }
+
+var (
+	VariableKindDynamic = VariablePluginKind{valuer.NewString("signozdynamicvariable")}
+	VariableKindQuery   = VariablePluginKind{valuer.NewString("signozqueryvariable")}
+	VariableKindCustom  = VariablePluginKind{valuer.NewString("signozcustomvariable")}
+	VariableKindTextbox = VariablePluginKind{valuer.NewString("signoztextboxvariable")}
```
srikanthccv: sorry for back and forth here. some kinds are coming from the upstream which are pascal case and we have small case long values which are hard to read, makes it bit worse with mixed style. some suggestions.

- `signoz/DynamicVariable`
- `signoz/QueryVariable`
- ...

follows the perses and k8s style `Kind: Deployment, ConfigMap`

...

or

- `signoz.DynamicVariable`
- `signoz.QueryVariable`

This means we can't use `valuer.String` and that's fine. Same for other enums, let's pick something and use it for all enums

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,553 @@
...
+	var v string
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
+	switch v {
+	case LegendPositionBottom, LegendPositionRight:
+		l.value = v
+		return nil
+	default:
+		return errors.NewInvalidInputf(ErrCodeDashboardInvalidInput, "invalid legend position %q: must be bottom or right", v)
```
srikanthccv:
```suggestion
		return errors.NewInvalidInputf(ErrCodeDashboardInvalidInput, "invalid legend position %q: must be `bottom` or `right`", v)
```

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,553 @@
...
+}
+
+type LegendPosition struct {
+	value string
+}
+
+const (
+	LegendPositionBottom = "bottom" // default
+	LegendPositionRight  = "right"
+)
```
srikanthccv: This could be valuer.String because it's enum value and not a `kind` value?

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,553 @@
...
+type LineInterpolation struct {
+	value string
+}
+
+const (
+	LineInterpolationLinear     = "linear"
+	LineInterpolationSpline     = "spline" // default
+	LineInterpolationStepAfter  = "stepAfter"
+	LineInterpolationStepBefore = "stepBefore"
+)
```
srikanthccv: Similar question here, these are the value for `lineInterpolation`. The values could be enum and use the `linear`, `spline`, `step_after` and `step_before` instead?

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,553 @@
...
+		return PrecisionOption2
+	}
+	return p.value
+}
+
+func (p *PrecisionOption) UnmarshalJSON(data []byte) error {
+	var v string
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
```
srikanthccv, line 602: I think you asked it some where, let's also support the number version of them as fallback before throwing error.

pkg/types/dashboardtypes/dashboard_v2.go
srikanthccv: No, the check can happen later.

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,553 @@
...
+	BucketCount           *float64 `json:"bucketCount"`
+	BucketWidth           *float64 `json:"bucketWidth"`
+	MergeAllActiveQueries bool     `json:"mergeAllActiveQueries"`
+}
+
+type ListPanelSpec struct {
+	SelectFields []telemetrytypes.TelemetryFieldKey `json:"selectFields,omitempty" validate:"dive"`
+}
+
+type LogField struct {
```
srikanthccv: This is not used

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,553 @@
...
+	TimePreferenceGlobalTime = "globalTime" // default
+	TimePreferenceLast5Min   = "last5Min"
+	TimePreferenceLast15Min  = "last15Min"
+	TimePreferenceLast30Min  = "last30Min"
+	TimePreferenceLast1Hr    = "last1Hr"
+	TimePreferenceLast6Hr    = "last6Hr"
+	TimePreferenceLast1Day   = "last1Day"
+	TimePreferenceLast3Days  = "last3Days"
+	TimePreferenceLast1Week  = "last1Week"
+	TimePreferenceLast1Month = "last1Month"
```
srikanthccv: use the snake case for all values of enums.

pkg/types/dashboardtypes/perses_plugin_types.go
srikanthccv: main concern was having default value, which can be done during the unmarshal so valuer can be used here.

pkg/types/dashboardtypes/dashboard_v2.go
srikanthccv: - typos
- fronted could also end up adding unnecessary fields

let's reject with special handling of typos, give suggestions (not as part of this PR) like "did you mean visualization?"

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,553 @@
...
+		return ThresholdFormatText
+	}
+	return f.value
+}
+
+func (f *ThresholdFormat) UnmarshalJSON(data []byte) error {
+	var v string
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
```
srikanthccv, line 358: `if err != nil {return err}` this is naked error from json, and not signoz typed error so user sees unknown in stead of invalid.

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: Done

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: Done

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: Done

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: Done

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: Done

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: Done

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: Done

pkg/types/dashboardtypes/dashboard_v2.go
namanv3: this is done for all plugins, but for perses' own things this won't work cuz they have their custom unmarshall function that don't have this check

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: this is done, it did require Value() and MarshallJson functions though

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,567 @@
...
+	TimePreferenceLast1Week  = "last_1_week"
+	TimePreferenceLast1Month = "last_1_month"
+)
+
+func (t TimePreference) Value() string {
+	if t.IsZero() {
+		return TimePreferenceGlobalTime
+	}
+	return t.StringValue()
+}
```
srikanthccv: 1. do we need the different `Value` method here?
2. it also changes the valuer interface where `Value` is used for completing db `Value` and `Scan` interface.

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,567 @@
...
+
+// ══════════════════════════════════════════════
+// Constrained scalar types — with default value
+// ══════════════════════════════════════════════
+
+type TimePreference struct{ valuer.String }
+
+const (
+	TimePreferenceGlobalTime = "global_time" // default
```
srikanthccv:
```suggestion
	TimePreferenceGlobalTime = TimePreference{valuer.NewString("global_time")} // default
```

and the rest.

pkg/types/dashboardtypes/perses_plugin_types.go
```diff
@@ -0,0 +1,567 @@
...
+type LegendPosition struct{ valuer.String }
+
+const (
+	LegendPositionBottom = "bottom" // default
+	LegendPositionRight  = "right"
+)
+
+func (l LegendPosition) Value() string {
+	if l.IsZero() {
+		return LegendPositionBottom
```
srikanthccv: Similar here

srikanthccv (commented):
Thanks, LGTM. One point about the usage of valuer.String and overriding `Value` method.

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: it doesn't have to be called Value, renamed to ValueOrDefault.

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: done

pkg/types/dashboardtypes/perses_plugin_types.go
namanv3: done

srikanthccv (dismissed):
Thanks! There are other minor things I would like to see changed but they can be addressed in sub-sequent PRs since this is still under development.

pkg/types/dashboardtypes/testdata/perses.json
```diff
@@ -0,0 +1,861 @@
...
+                        }
+                    }
+                ]
+            }
+        }
+    },
+    "layouts": [
+        {
+            "kind": "Grid",
+            "spec": {
```
AshwinBhatkal, line 747: need section name here

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/SigNoz/signoz
 
-go 1.25.0
+go 1.25.7
```
therealpandey, line 3: @namanv3 please revert to the root version. I think this coincided with the version you have locally.

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/go-co-op/gocron v1.30.1
 	github.com/go-openapi/runtime v0.29.2
 	github.com/go-openapi/strfmt v0.25.0
+	github.com/go-playground/validator/v10 v10.27.0
```
therealpandey, line 22: Nice!

go.mod
```diff
@@ -27,8 +28,8 @@ require (
 	github.com/gorilla/handlers v1.5.1
 	github.com/gorilla/mux v1.8.1
 	github.com/gorilla/websocket v1.5.4-0.20250319132907-e064f32e3674
-	github.com/huandu/go-sqlbuilder v1.35.0
-	github.com/jackc/pgx/v5 v5.7.6
+	github.com/huandu/go-sqlbuilder v1.39.1
+	github.com/jackc/pgx/v5 v5.8.0
```
therealpandey, line 32: this should be okay for us

therealpandey (changes requested):

go.mod
namanv3: the perses package actually needs 1.25.7, that's why the upgrade. will this be a problem?

therealpandey: (1.25.0 should work for 1.25.7?) 

Can you try testing with 1.25.0?

pkg/types/dashboardtypes/testdata/perses.json
namanv3: perses actually allows empty names for sections. I didn't add a custom check on their structs cuz that may cause issues if we're importing perses dashboards from anywhere else in the future

namanv3: maybe we can raise an issue in their repo and see what they say? anyway have to do it for required fields in general for the openapi spec

pkg/types/dashboardtypes/testdata/perses_with_sections.json
```diff
@@ -0,0 +1,160 @@
...
+                ]
+            }
+        }
+    },
+    "layouts": [
+        {
+            "kind": "Grid",
+            "spec": {
+                "display": {
+                    "title": "Bravo",
```
AshwinBhatkal: we can remove this from here. It can be a user preference only

pkg/types/dashboardtypes/testdata/perses_with_sections.json
```diff
@@ -0,0 +1,160 @@
...
+                    "collapse": {
```
AshwinBhatkal: we can remove this from here. It can be a user preference only

pkg/types/dashboardtypes/testdata/perses_with_sections.json
namanv3: done

pkg/types/dashboardtypes/testdata/perses_with_sections.json
namanv3: done

pkg/types/dashboardtypes/testdata/perses.json
```diff
@@ -0,0 +1,861 @@
...
+    "variables": [
+        {
+            "kind": "ListVariable",
+            "spec": {
+                "name": "serviceName",
+                "display": {
+                    "name": "serviceName"
+                },
+                "allowAllValue": true,
+                "allowMultiple": false,
```
AshwinBhatkal, line 25: Will this be caught in validation? allAllValue is true while allowMultiple is false

pkg/types/dashboardtypes/testdata/perses.json
```diff
@@ -0,0 +1,861 @@
...
+                "sort": "none",
+                "plugin": {
+                    "kind": "signoz/DynamicVariable",
+                    "spec": {
+                        "name": "service.name",
+                        "signal": "metrics"
```
AshwinBhatkal, line 31: does this take in 'All'?

AshwinBhatkal (commented):
I'm good overall. 2 questions

1. Can the go files be broken down based on type of entity? Like variables, panels, queries, etc?
2. Till we take it live fully, any change should be easy to make. Correct?

go.mod
namanv3: If I keep go version at `1.25.0`, `go build ./...` gives
```
go: updates to go.mod needed; to update it:
        go mod tidy
```
And `go mod tidy` just bumps it to `1.25.7`

pkg/types/dashboardtypes/testdata/perses.json
namanv3: if signal is empty that is equivalent to All

pkg/types/dashboardtypes/testdata/perses.json
namanv3: no this won't be caught in schema checks at least, but if allowMultiple is false then allowAllValue can be ignored as such.
This can be added as additional validations later on though if needed

namanv3:
1. I have smaller examples in the go test file, I wanted to keep one big file for complete validation
2. yes
@AshwinBhatkal

go.mod
namanv3: @therealpandey can this be resolved?

therealpandey (approved):

srikanthccv (approved):

AshwinBhatkal (approved):

PR #10618, opened 2026-03-17
fix(user): add config for user invite token expiry 

vikrantgupta25:
### 📄 Summary
- add config for user invite token expiry

contributes to: https://github.com/SigNoz/platform-pod/issues/1724

### DIFF
--- a/conf/example.yaml
+++ b/conf/example.yaml
@@ -309,7 +309,7 @@ user:
       # The duration within which a user can reset their password.
       max_token_lifetime: 6h
     invite:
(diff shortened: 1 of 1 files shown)


conf/example.yaml
```diff
@@ -308,6 +308,9 @@ user:
       allow_self: true
       # The duration within which a user can reset their password.
       max_token_lifetime: 6h
+    invite:
+      # The duration within which a user can reset their password.
```
therealpandey:
```suggestion
      # The duration within which a user can accept their invite.
```

PR #10620, opened 2026-03-17
fix(devenv): fix otel-collector startup failure

amareshsm:
## Pull Request

---

The `make devenv-signoz-otel-collector` setup step fails on all Docker runtimes (OrbStack, Rancher Desktop, Docker Desktop etc) because the OTel Collector container enters an infinite retry loop and never becomes healthy, causing curl `http://localhost:13133` to return Empty reply from server.

Two bugs were introduced by PR #10334, which moved schema migration into the OTel Collector image and added a startup migration sync-check that requires resolving the clickhouse hostname:

Bug 1 — DNS isolation (affects ALL runtimes):
Each docker compose stack creates its own isolated bridge network (clickhouse_default, signoz-otel-collector_default). Containers on different networks cannot resolve each other by hostname. The OTel Collector's migrate sync check tries tcp://clickhouse:9000 but Docker DNS returns:

```
dial tcp: lookup clickhouse on 127.0.0.11:53: no such host
```

The container loops on this forever and never starts the collector process, so the health check endpoint on :13133 is never bound.

image

Bug 2 — host.docker.internal not portable (affects Rancher Desktop):
otel-collector-config.yaml exporters used host.docker.internal:9000. OrbStack uniquely routes this to 127.0.0.1 on the Mac (where ClickHouse is bound), but Rancher Desktop and Docker Desktop route it to the VM's internal gateway IP, where ClickHouse is not listening:

```
dial tcp 172.17.0.1:9000: connect: connection refused
```

1. Add a shared named network signoz-devenv declared in both compose files using name: signoz-devenv (without external: true). Docker creates it on first use and reuses it on second — fully startup-order independent. This gives the OTel Collector DNS visibility of the clickhouse container.

2. Change all three exporters in `otel-collector-config.yaml` from `host.docker.internal:9000` to `clickhouse:9000` — pure container-to-container routing over the shared network, no VM boundary, no runtime-specific behavior.

 `host.docker.internal` is a runtime-specific convenience hostname (not part of the Docker networking spec), so its behavior varies across environments.

Before:

```
$ curl http://localhost:13133
curl: (52) Empty reply from server

$ docker logs signoz-otel-collector-dev --tail 5
{"msg":"Error occurred while checking for sync migrations to complete, retrying",
 "error":"dial tcp: lookup clickhouse on 127.0.0.11:53: no such host"}
 ```
 
 After:
 
 ```
 $ curl http://localhost:13133
{"status":"Server available","upSince":"2026-03-18T...","uptime":"..."}
```

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

PR #10334  introduced a migrate sync check step in the OTel Collector's entrypoint that requires the clickhouse hostname to be resolvable at container startup. However, make devenv-clickhouse and make devenv-signoz-otel-collector run as separate docker compose stacks in separate directories. Docker Compose auto-creates an isolated <directory>_default network per stack, so clickhouse is unreachable by hostname from the OTel Collector's network — on every runtime.

Additionally, the same PR left the otel-collector-config.yaml exporters using host.docker.internal:9000, which is a host-bridging mechanism that only works correctly on OrbStack (routes to 127.0.0.1). On Rancher Desktop it routes to the VM gateway IP, where ClickHouse is not bound.

- Rather than patching the network post-startup (e.g. docker network connect), both compose files declare the same named network signoz-devenv using name: signoz-devenv. Docker creates the network on whichever stack starts first and reuses it for the second — making the fix startup-order independent with no external: true fragility.

- Changing the exporters to `clickhouse:9000` removes the host.docker.internal dependency entirely. The connection travels purely container-to-container inside Docker Engine's networking layer, which is identical across all runtimes.
---

- Tests added/updated: N/A — devenv infra change, no unit tests applicable
- Manual verification:
     - Verified on OrbStack & rancher desktop (Mac): make devenv-clickhouse && make devenv-signoz-otel-collector → curl http://localhost:13133 returns {"status":"Server available",...} ✅
     - Verified both startup orderings:
            - ClickHouse first, then OTel Collector ✅
            -  OTel Collector first, then ClickHouse ✅ (network created by OTel stack, ClickHouse joins it)
     - Verified container-to-container connectivity: docker run --rm --network signoz-devenv alpine nc -zv clickhouse 9000 → open / reachable ✅

- Edge cases covered:
      - Startup order independence: name: signoz-devenv without external: true means Docker creates-or-reuses, no crash if either stack starts first
      - Runtime portability: clickhouse:9000 works on OrbStack, Rancher Desktop, and Docker Desktop — no VM boundary crossed
---

- Blast radius:  Dev environment only (.devenv directory). No production configs, no application code changed.
- Potential regressions: Nothing much - The signoz-devenv network is additive — both services retain their default network and gain signoz-devenv. Existing port bindings, volumes, and environment variables are unchanged.
- Rollback plan: Revert the pr. The previous behavior (broken on runtimes since PR #10334) is the only regression risk.

---

| Field | Value |
|------|-------|
| Deployment Type | N/A |
| Change Type |  Bug Fix |
| Description |  N/A — dev environment setup only, no user-facing change  |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

Three files changed:
compose.yaml — adds signoz-devenv network to clickhouse and telemetrystore-migrator services + top-level network declaration
compose.yaml — adds signoz-devenv network to the OTel Collector service + top-level network declaration
otel-collector-config.yaml — changes 3 exporter DSNs from host.docker.internal:9000 to clickhouse:9000

I hope I’ve done everything correctly. If there’s anything missing, I’m happy to address it. Thanks.

---


Nageshbansal:
Hey @amareshsm, thanks for the PR. The changes look good to me.

PR #10621, opened 2026-03-17
feat: separate models for user and user_role table with roles

karanbalani:
## Pull Request

---

### 📄 Summary
- Separated out Domains models for users, into `StorableUser` and `User`
- Introduces a dedicated user_role table with migrations
- Backward compatibility for the Role field is maintained during transition.

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

### DIFF
--- a/pkg/modules/user/impluser/getter.go
+++ b/pkg/modules/user/impluser/getter.go
@@ -2,18 +2,42 @@ package impluser
 
 import (
 	"context"
+	"slices"
 
+	"github.com/SigNoz/signoz/pkg/authz"
+	"github.com/SigNoz/signoz/pkg/flagger"
 	"github.com/SigNoz/signoz/pkg/modules/user"
 	"github.com/SigNoz/signoz/pkg/types"
+	"github.com/SigNoz/signoz/pkg/types/authtypes"
+	"github.com/SigNoz/signoz/pkg/types/featuretypes"
 	"github.com/SigNoz/signoz/pkg/valuer"
 )
 
 type getter struct {
-	store types.UserStore
+	store         types.UserStore
+	authz         authz.AuthZ
+	userRoleStore authtypes.UserRoleStore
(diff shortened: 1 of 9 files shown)


pkg/types/user.go
```diff
@@ -33,16 +33,26 @@ var (
...
+
+type StorableUser struct {
 	bun.BaseModel `bun:"table:users"`
 
 	Identifiable
 	DisplayName string        `bun:"display_name" json:"displayName"`
-	Email       valuer.Email  `bun:"email" json:"email"`
-	Role        Role          `bun:"role" json:"role"`
-	OrgID       valuer.UUID   `bun:"org_id" json:"orgId"`
+	Email       string        `bun:"email" json:"email"`
```
therealpandey: Why?

pkg/types/user.go
```diff
@@ -33,16 +33,26 @@ var (
...
+	Role        Role          `bun:"role" json:"role"` // this will be removed as column from here
+	OrgID       string        `bun:"org_id" json:"orgId"`
```
therealpandey: Why?

pkg/types/user.go
karanbalani: https://github.com/SigNoz/signoz/blob/5bafdeb373be45019260ab49648a66b2be38751b/pkg/types/serviceaccounttypes/service_acccount.go#L35-L45

pkg/types/user.go
karanbalani: used `valuer.Email` for email and `valuer.UUID` for org id.

therealpandey:
Closed in favour of https://github.com/SigNoz/signoz/pull/10664

Issue #5088, opened 2024-05-27
LOGS[FE/UI]: Allow resizing of columns 


nityanandagohain:
## Issue:

Allow resizing of columns in logs list page.

## Current Behaviour:
https://github.com/user-attachments/assets/944065bc-261d-47e0-a388-46d32b0498d3

## Expected Behaviour;
We should be able to resize the columns in the column view of logs explorer

Figma Design: https://www.figma.com/design/L0vX330RkWn7ZpelMpkPHL/Logs?node-id=2283-6624&t=zrQjNyq3aBC64ptt-1

GeekBoySupreme:
<img width="915" alt="image" src="https://github.com/SigNoz/signoz/assets/15321738/ae79381c-763c-40f7-81c0-3c0d983a40a4">
Figma is [here](https://www.figma.com/design/L0vX330RkWn7ZpelMpkPHL/Logs?node-id=2283-6624&t=zrQjNyq3aBC64ptt-1)

YounixM:
@vikrantgupta25 : Have already updated the component for this. Need to clean up and adjust for having responsive behaviour. Will share the branch with you and you can take it from there.

nityanandagohain:
Bump https://github.com/SigNoz/signoz/issues/4900#event-15825314237

sawhil:
We will support this with the new table component that we are going to consume

sawhil:
Closing this for now, will reference this in the new table component consumption ticket.

ankitnayan:
part of a broader epic https://www.notion.so/signoz/Table-Component-TDD-1f8fcc6bcd1980f08bd1ef5a8d666c4c?source=copy_link

PR #9952, opened 2026-01-08
Fix duplicate entries in add column options

aks07:
## 📄 Summary

Introduces attribute variant awareness and deduping in the options UI.

- Uses composite attribute keys via buildAttributeKey and parsing helpers to uniquely identify columns and prevent duplicates in AddColumn and logs options menu
- New FieldVariantBadges component and styles to display fieldDataType and fieldContext tags for attributes with the same name in option lists and selected columns
- Updates OptionsMenu/LogsFormatOptionsMenu logic, styling, and popover focus/blur handling; widens nested menu and adds truncation wrappers
- Fixes duration_nano type to number in defaults
- Adds unit/UI tests covering badge rendering and variant display in both Logs and Traces views

---

## ✅ Changes

- [x] Feature: Brief description
- [x] Bug fix: Brief description

---

## 🔍 Related Issues

Closes #
https://github.com/SigNoz/engineering-pod/issues/3345

---

## 📸 Screenshots / Screen Recording (if applicable / mandatory for UI related changes)

Before: 

Attribute keys with same name and different dataTypes breaking the options in trace explorer

https://github.com/user-attachments/assets/7fda86ba-34df-4d56-a432-74e24546f4f4

After:

Show badges for multi-variant attribute keys in traces and logs column options
Traces

https://github.com/user-attachments/assets/f838c63c-8407-4f3d-895e-8951ff233071

Logs Options
image

---

## 📋 Checklist

- [x] Dev Review
- [x] Test cases added (Unit/ Integration / E2E)
- [x] Manually tested the changes

---

> [!NOTE]
> Introduces attribute-variant awareness and deduping in options UIs.
> 
> - Add `FieldVariantBadges` (with styles) to show `fieldDataType` and `fieldContext` for attributes with identical names
> - Use composite attribute keys via `buildAttributeKey`/`parseAttributeKey` and `getOptionLabelCounts` to uniquely select/remove columns and detect multi-variant labels in `OptionsMenu` and `LogsFormatOptionsMenu`
> - Update add-column lists and selected columns to render badges; widen nested menu and add truncation wrappers; manage popover focus/blur
> - Fix default `duration_nano` type to `number`
> - Add unit/UI tests covering variant badge rendering in Logs and Traces views
> 
[...]

PR #10625, opened 2026-03-18
feat: table migration

ishan621:
## Pull Request
Table migration to TansStack table

This PR migrates the Logs table implementation to TanStack Table (headless) to unlock missing table capabilities (notably column resizing) and make the table layer easier to extend and maintain.
The previous table implementation had limitations around table interactions (for example, column resizing) and was harder to evolve without accumulating complexity.
Enables a more capable table foundation via TanStack (headless, lightweight, and maintainable).

Before :
Screenshot 2026-03-23 at 10 35 36 AM

After:

Screenshot 2026-03-26 at 9 51 24 AM

https://github.com/user-attachments/assets/2dba4f20-0120-4971-b267-e20000fba780

#### Issues closed by this PR
https://github.com/SigNoz/engineering-pod/issues/3791
https://github.com/SigNoz/signoz/issues/5088 
https://github.com/SigNoz/engineering-pod/issues/3554

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

### 📋 Checklist
- [x] Manually tested

---


frontend/src/container/LogsExplorerList/TanStackTableView/index.tsx
H4ad, line 1: This should be inside the `components` folder instead of inside the `Container/LogExplorerList` if we want to reuse it across multiple places.

ishan621: for now it is ok to be here, once we create a common component in signoz/ui then we can migrate this as well

frontend/src/container/LogsExplorerList/TanStackTableView/TanStackHeaderRow.styles.scss
```diff
@@ -129,16 +133,20 @@
...
+
+		.tanstack-remove-column-action-icon {
+			color: var(--text-ink-600);
+		}
 	}
 }
 
 .lightMode .tanstack-remove-column-action-icon {
-	color: var(--bg-cherry-600);
+	color: var(--text-ink-400);
```
H4ad: Use semantic tokens

ishan621: updated

H4ad (approved):
Looks good for the initial implementation, some points we can address in the following PRs:

- Extract/Remove specific from logs from the table component, move it to common components folder to be reusable by all pages
- Investigate performance issues of Raw Logs (not related but it's caused by LogActionsLineButtons, we can migrate to our component library instead of antd
- Maybe add some safe-guard to avoid rendering duplicated columns on "add modal" due to issues on backend.

aks07:
> Looks good for the initial implementation, some points we can address in the following PRs:
> 
> * Extract/Remove specific from logs from the table component, move it to common components folder to be reusable by all pages
> * Investigate performance issues of Raw Logs (not related but it's caused by LogActionsLineButtons, we can migrate to our component library instead of antd
> * Maybe add some safe-guard to avoid rendering duplicated columns on "add modal" due to issues on backend.

- We will handle extraction of common component in upcoming enhancements. Will take a look at raw logs performance as well
- Regarding the duplicate col fix. there is an open discussion on the backend which is delaying its frontend fix. check out discussion here: https://github.com/SigNoz/signoz/pull/9952

frontend/src/container/LogsExplorerList/InfinityTableView/config.ts
```diff
@@ -4,18 +4,11 @@ import { FontSize } from 'container/OptionsMenu/types';
 export const infinityDefaultStyles: CSSProperties = {
 	width: '100%',
 	overflowX: 'scroll',
-	marginTop: '15px',
 };
 
-export function getInfinityDefaultStyles(fontSize: FontSize): CSSProperties {
+export function getInfinityDefaultStyles(_fontSize: FontSize): CSSProperties {
```
YounixM, line 9: we can remove the function parameter as it's not being used

YounixM:
@aks07 - As discussed, approving this for the release. We have to clean this up when we convert this to a reusable table component.

YounixM (approved):

PR #10630, opened 2026-03-18
fix: added fix for isolated points render with null both side

ahrefabhi:
## Pull Request

---

### 📄 Summary

When `spanGaps` is configured in threshold (numeric) mode, uPlot injects gap-nulls around data points that exceed the threshold. This causes isolated points — those with no adjacent data on either side — to become invisible because no line segment connects to them and the default `show: false` suppresses dots.

This PR introduces `isolatedPointFilter`, a `Series.Points.filter` function that detects these isolated points by inspecting uPlot's computed gap pixel array and returns their indices so they render as visible dots. The filter is automatically applied in `UPlotSeriesBuilder` when `spanGaps` is numeric, `drawStyle` is `Line`, and no custom filter or explicit `showPoints` is set.

image

> _After: those points render as dots._
image

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/4252

---

### ✅ Change Type

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
When `spanGaps` threshold mode is active, uPlot injects `null` values around segments that exceed the gap threshold, breaking the line. Data points surrounded by these injected nulls on both sides become isolated — no line draws through them and `show: false` (the default for non-point draw styles) means no dot is drawn either, making the data invisible.

#### Fix Strategy
- Added `seriesPointsFilter.ts` with `isolatedPointFilter`: a `Series.Points.filter` callback that receives uPlot's gap pixel array, finds adjacent gap boundaries that share a pixel (indicating a sandwiched isolated point), and returns the data indices of those points.
- Added helper `findSandwichedIndices` (uses gap pixel positions + `posToIdx`) and `findNearestNonNull` (handles pixel-density misalignment).
- In `UPlotSeriesBuilder.buildPointsConfig`, refactored point `show` resolution into `resolvePointsShow()` and added `shouldApplyIsolatedPointFilter()` — the filter is applied only when `drawStyle === Line`, `spanGaps` is numeric, no explicit `pointsFilter` is set, and `show` is falsy. This avoids interfering with the `show=true` fast-path in uPlot that would bypass the filter entirely.

---

### 🧪 Testing Strategy

- **Tests added:**
  - `seriesPointsFilter.test.ts` — 251 lines covering `findNearestNonNull`, `findSandwichedIndices`, and `isolatedPointFilter` across edge cases (empty gaps, first/last point isolation, sandwiched clusters, pixel-density misalignment, `show=true` short-circuit).
  - `UPlotSeriesBuilder.test.ts` — 47 lines covering `shouldApplyIsolatedPointFilter` guard conditions (only activates under correct `drawStyle`, `spanGaps`, and `show` combinations).
- **Manual verification:** Confirmed isolated points become visible dots in charts with numeric `spanGaps`.
- **Edge cases covered:** First/last point isolated, multiple isolated clusters, `show=true` short-circuit leaves filter untouched, no gaps returns `null`.

---

- **Blast radius:** Scoped to `UPlotSeriesBuilder.buildPointsConfig`; only activates when `drawStyle === Line && typeof spanGaps === 'number' && !pointsFilter && !show`. No change to bar, area, or explicit points-draw-style series.
- **Potential regressions:** None expected for series not using numeric `spanGaps`. The `resolvePointsShow` refactor is a pure extraction with identical logic.
- **Rollback plan:** Revert the single commit `e7dc4e1`; no schema or API changes involved.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Isolated data points surrounded by gap-nulls in threshold spanGaps mode now render as visible dots instead of disappearing. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- The key invariant: `isolatedPointFilter` must only be set when `show` is falsy. If `show=true`, uPlot short-circuits the filter call (passes `show=true` to the callback), and the sandwiched-index logic would incorrectly return an empty array instead of `null`, hiding all points. `shouldApplyIsolatedPointFilter` enforces this.
- `findNearestNonNull` has a hardcoded scan radius of 100 — sufficient for any realistic pixel-to-data ratio but worth noting if someone adds extremely sparse data.

---

### DIFF
--- a/frontend/src/lib/uPlotV2/config/UPlotSeriesBuilder.ts
+++ b/frontend/src/lib/uPlotV2/config/UPlotSeriesBuilder.ts
@@ -186,13 +186,8 @@ export class UPlotSeriesBuilder extends ConfigBuilder<SeriesProps, Series> {
 	}
 
 	private shouldApplyIsolatedPointFilter(show: Series.Points['show']): boolean {
-		const { drawStyle, spanGaps, pointsFilter } = this.props;
-		return (
-			drawStyle === DrawStyle.Line &&
-			typeof spanGaps === 'number' &&
-			!pointsFilter &&
-			!show
-		);
+		const { drawStyle, pointsFilter } = this.props;
+		return drawStyle === DrawStyle.Line && !pointsFilter && !show;
 	}
 
 	private getLineColor(): string {


frontend/src/lib/uPlotV2/config/UPlotSeriesBuilder.ts
```diff
@@ -168,19 +157,44 @@ export class UPlotSeriesBuilder extends ConfigBuilder<SeriesProps, Series> {
...
+			return true;
+		}
+		return !!showPoints;
+	}
+
+	private shouldApplyIsolatedPointFilter(show: Series.Points['show']): boolean {
+		const { drawStyle, spanGaps, pointsFilter } = this.props;
+		return (
+			drawStyle === DrawStyle.Line &&
+			typeof spanGaps === 'number' &&
```
AshwinBhatkal: what happens to isolated points when it is boolean?

Issue #9238, opened 2025-10-02
Foreign constrain failed error on deleting rule


srikanthccv:
Trying to delete rule throws "Foreign constrain failed error"

vikrantgupta25:
this has been taken care here: https://github.com/SigNoz/signoz/pull/10632 please reopen if something else is also expected to be done with this.

PR #10632, opened 2026-03-18
fix: handle foreign key constraint on rule and planned maintenance deletion

therealpandey:
## Summary

- Adds migration that recreates the `planned_maintenance_rule` table without `ON DELETE CASCADE ON UPDATE CASCADE` on its foreign keys, aligning with project conventions
- Extends `WrapAlreadyExistsErrf` in SQLite and PostgreSQL providers to also match foreign key constraint violations (`SQLITE_CONSTRAINT_FOREIGNKEY` / PG error code `23503`)
- Wraps FK errors in `DeleteRule` and `DeletePlannedMaintenance` with descriptive messages instead of surfacing raw constraint errors


therealpandey:
Closes https://github.com/SigNoz/signoz/issues/9238

PR #10637, opened 2026-03-18
fix: prevent duplicate and incorrect results from trace_summary timerange override in list view

nityanandagohain:
## Pull Request

---

### 📄 Summary
The previous implementation of updating the start and end time of the query based on trace_summary was wrong with the bucket implementation of list view. Each bucket timerange was updated and it ran the same query resulting in duplicate data and wrong data as well. 

1) Duplicate data with higher timerange :- here you can see the same span is repeated with same timestamp.
image

2) Wrong result from a different timerange - here I am querying 1-2 March, but it is giving me results of 18th march
image

#### Issues closed by this PR
Fixes https://github.com/SigNoz/engineering-pod/issues/4250

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
Directly overriding the start and endtimestamp without checking overlapping timerange context.

#### Fix Strategy
It makes sure we are only overriding if the timrange overlaps

---

- Blast radius: List view
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Correct results while filtering with trace_id |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/tests/integration/src/querier/04_traces.py
+++ b/tests/integration/src/querier/04_traces.py
@@ -2131,9 +2131,9 @@ def test_traces_list_filter_by_trace_id(
     narrow_start_ms = int((now - timedelta(minutes=5)).timestamp() * 1000)
     narrow_rows = _query(narrow_start_ms, now_ms)
 
-    assert len(narrow_rows) == 1, (
-        f"Expected 1 span for trace_id filter (narrow window), got {len(narrow_rows)}"
-    )
+    assert (
+        len(narrow_rows) == 1
+    ), f"Expected 1 span for trace_id filter (narrow window), got {len(narrow_rows)}"
     assert narrow_rows[0]["data"]["span_id"] == span_id_root
     assert narrow_rows[0]["data"]["trace_id"] == target_trace_id
 


pkg/types/querybuildertypes/querybuildertypesv5/qb.go
```diff
@@ -47,6 +47,7 @@ type Statement struct {
 	Args           []any
 	Warnings       []string
 	WarningsDocURL string
+	Skip           bool // don't run this query
```
srikanthccv: The `Skip` field on `Statement` is not good way to go about it. `Statement` represents a built SQL query. Adding a boolean that means "don't run this query" is a control-flow decision leaking through a type. This is a sign of code smell. Every caller of `Build()` must now remember to check `Skip`. 

The main underlying issue is separation of concerns. The statement builder shouldn't be deciding whether a bucket should be skipped. The only responsibility of the statement builder is to is to prepare a `Statement` with `Query` and `Args` that can be executed. The time range optimization should happen before the bucket loop, at the querier orchestration layer.

Here is what I suggest instead. resolve the trace time range once in `executeWindowList`, clamp/filter the buckets before iterating, then pass correct start/end to `Build()`. This keeps `Build()` a pure SQL builder and avoids the Skip hack entirely.

pkg/querier/builder_query.go
```diff
@@ -305,6 +310,45 @@ func (q *builderQuery[T]) executeWindowList(ctx context.Context) (*qbtypes.Resul
...
+			traceStartMS := uint64(traceStart) / 1_000_000
+			traceEndMS := uint64(traceEnd) / 1_000_000
+			if !ok {
+				q.logger.DebugContext(ctx, "failed to get trace time range", slog.Any("trace_ids", traceIDs))
+			} else if traceStartMS > 0 && traceEndMS > 0 {
+				// no overlap — nothing to return
+				if uint64(traceStartMS) > toMS || uint64(traceEndMS) < fromMS {
+					return &qbtypes.Result{
+						Type: qbtypes.RequestTypeRaw,
+						Value: &qbtypes.RawData{
```
srikanthccv, line 331: There is a `Rows` on `RawData`. Where there is no data, is it nil? Or is it empty list?

nityanandagohain: It's nil. This is what I get when my I search for something that is not present, eg: `service.name=abcd`

```
{
    "status": "success",
    "data": {
        "type": "raw",
        "meta": {
            "rowsScanned": 77094,
            "bytesScanned": 65548243,
            "durationMs": 111,
            "stepIntervals": {
                "A": 10
            }
        },
        "data": {
            "results": [
                {
                    "queryName": "A",
                    "nextCursor": "",
                    "rows": null
                }
            ]
        }
    }
}
```

nityanandagohain: And here is the result when I search for the trace_id in wrong interval
```
{
    "status": "success",
    "data": {
        "type": "raw",
        "meta": {
            "rowsScanned": 0,
            "bytesScanned": 0,
            "durationMs": 17,
            "stepIntervals": {
                "A": 575
            }
        },
        "data": {
            "results": [
                {
                    "queryName": "A",
                    "nextCursor": "",
                    "rows": null
                }
            ]
        }
    }
}
```

nityanandagohain:
I feel the tests still adds value and makes sure we are getting what we expect, so keeping them as it is after the new change.

srikanthccv:
@nityanandagohain, do the integration tests fail, if you comment out the change and then run integration tests? Can you do that locally and share the output once?

nityanandagohain:
@srikanthccv 
```
=================================================================================================== short test summary info ====================================================================================================
FAILED src/querier/04_traces.py::test_traces_list_filter_by_trace_id - AssertionError: Expected 0 spans for trace_id filter outside time window, got 1
assert 1 == 0
 +  where 1 = len([{'data': {'name': 'root-span', 'span_id': '1453d8fdd6a9abaa', 'timestamp': '2026-03-24T05:01:50Z', 'trace_id': 'f3ae433f8f6ac470fe32346085418f40'}, 'timestamp': '2026-03-24T05:01:50Z'}])
====================================================================================================== 1 failed in 1.31s =======================================================================================================
```
 test_traces_list_filter_by_trace_id fails while 
 
 test_traces_list_filter_by_trace_id_cross bucket passes
 
 This is what I commented
 
image

vikrantgupta25 (approved):

PR #10640, opened 2026-03-18
refactor(infrastructure-monitoring): use nuqs hooks

H4ad:
## Pull Request

This PR refactors all the Infrastructure Monitoring to use Nuqs instead of useSearchParam.

This also fixes the following bugs:
- Sort not being preserved after page refresh (fixed on K8sPodLists only, next steps is to fix on all pages)
- Selected Row Data could crash the page due to circular references when trying to serialize react nodes
- Filters not supporting % in the query param.

Important changes:
- Added zod as validation library.
  - Nuqs has a good support for it and zod is a small and reliable way of starting adding validation for few places of our codebase, specially for Nuqs + JSON.
---

This PRs superseeds the #10383 that fixed a bug related to decoding, that PR was good to fix the bug but I decided to clean all the components to use Nuqs that also solved this problem and helps clean the code of the components.

Fixes https://github.com/SigNoz/engineering-pod/issues/3935

Before:

https://github.com/user-attachments/assets/7044a39b-3441-4f3a-951e-4bc3b9783040

> The little issue with data not refresh after clean the query can be also reproduced on main, it's caused when you update the query (delete) without hit enter after cleaning the input.

After:

https://github.com/user-attachments/assets/4d6c8a64-0e5c-4ca1-92e6-4a2549666261

Before URL:
image

After URL:
image

Issue: https://github.com/SigNoz/engineering-pod/issues/3935

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

The bug this PR also fixes is related to search filter, when you add a filter like: `k8s.node.name like %%`, it was breaking the UI.

The bug was caused by the way the query params was deserialized using `decodeURIComponent` and the `%` not being correctly formatted to that function.

Replaced all occurrences to use Nuqs that automatically store the data correctly.

---

- Tests added/updated: Added regression test.
- Manual verification: Yes

---

- Blast radius: Infrastructure Monitoring Page
- Potential regressions: If we missed some place or updated the logic incorrectly to the new hooks

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Infrastructure Monitoring now supports filters with % in the query |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/container/InfraMonitoringK8s/Namespaces/K8sNamespacesList.tsx
+++ b/frontend/src/container/InfraMonitoringK8s/Namespaces/K8sNamespacesList.tsx
@@ -1,7 +1,6 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 // eslint-disable-next-line no-restricted-imports
 import { useSelector } from 'react-redux';
-import { useSearchParams } from 'react-router-dom-v5-compat';
 import { LoadingOutlined } from '@ant-design/icons';
 import {
 	Button,
@@ -72,7 +71,6 @@ function K8sNamespacesList({
 		(state) => state.globalTime,
 	);
 
-	const [searchParams] = useSearchParams();
 	const [expandedRowKeys, setExpandedRowKeys] = useState<string[]>([]);
 
 	const [currentPage, setCurrentPage] = useInfraMonitoringCurrentPage();
@@ -435,7 +433,7 @@ function K8sNamespacesList({
 	]);
 
 	const openNamespaceInNewTab = (record: K8sNamespacesRowData): void => {
-		const newParams = new URLSearchParams(searchParams);
+		const newParams = new URLSearchParams(document.location.search);
 		newParams.set(
 			INFRA_MONITORING_K8S_PARAMS_KEYS.NAMESPACE_UID,
 			record.namespaceUID,
--- a/frontend/src/container/InfraMonitoringK8s/StatefulSets/K8sStatefulSetsList.tsx
+++ b/frontend/src/container/InfraMonitoringK8s/StatefulSets/K8sStatefulSetsList.tsx
@@ -1,7 +1,6 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 // eslint-disable-next-line no-restricted-imports
 import { useSelector } from 'react-redux';
-import { useSearchParams } from 'react-router-dom-v5-compat';
 import { LoadingOutlined } from '@ant-design/icons';
 import {
 	Button,
@@ -72,8 +71,6 @@ function K8sStatefulSetsList({
 	const { maxTime, minTime } = useSelector<AppState, GlobalReducer>(
 		(state) => state.globalTime,
 	);
-
-	const [searchParams] = useSearchParams();
 
 	const [currentPage, setCurrentPage] = useInfraMonitoringCurrentPage();
 	const [filtersInitialised, setFiltersInitialised] = useState(false);
@@ -437,7 +434,7 @@ function K8sStatefulSetsList({
 	]);
 
 	const openStatefulSetInNewTab = (record: K8sStatefulSetsRowData): void => {
-		const newParams = new URLSearchParams(searchParams);
+		const newParams = new URLSearchParams(document.location.search);
 		newParams.set(
 			INFRA_MONITORING_K8S_PARAMS_KEYS.STATEFULSET_UID,
 			record.statefulsetUID,
--- a/frontend/src/container/InfraMonitoringK8s/__tests__/StatefulSets/StatefulSetDetails/StatefulSetDetails.test.tsx
+++ b/frontend/src/container/InfraMonitoringK8s/__tests__/StatefulSets/StatefulSetDetails/StatefulSetDetails.test.tsx
@@ -111,7 +111,7 @@ describe('StatefulSetDetails', () => {
 
 		const eventsTab = screen.getByRole('radio', { name: 'Events' });
 		expect(eventsTab).not.toBeChecked();
-		await userEvent.click(eventsTab);
+		await userEvent.click(eventsTab, { pointerEventsCheck: 0 });
 		expect(eventsTab).toBeChecked();
 	});
 
--- a/frontend/src/container/InfraMonitoringK8s/Clusters/K8sClustersList.tsx
+++ b/frontend/src/container/InfraMonitoringK8s/Clusters/K8sClustersList.tsx
@@ -1,7 +1,6 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 // eslint-disable-next-line no-restricted-imports
 import { useSelector } from 'react-redux';
-import { useSearchParams } from 'react-router-dom-v5-compat';
 import { LoadingOutlined } from '@ant-design/icons';
 import {
 	Button,
@@ -67,8 +66,6 @@ function K8sClustersList({
 	const { maxTime, minTime } = useSelector<AppState, GlobalReducer>(
 		(state) => state.globalTime,
 	);
-
-	const [searchParams] = useSearchParams();
 
 	const [currentPage, setCurrentPage] = useInfraMonitoringCurrentPage();
 	const [groupBy, setGroupBy] = useInfraMonitoringGroupBy();
@@ -417,7 +414,7 @@ function K8sClustersList({
 	}, [selectedClusterName, groupBy.length, clustersData, nestedClustersData]);
 
 	const openClusterInNewTab = (record: K8sClustersRowData): void => {
-		const newParams = new URLSearchParams(searchParams);
+		const newParams = new URLSearchParams(document.location.search);
 		newParams.set(
 			INFRA_MONITORING_K8S_PARAMS_KEYS.CLUSTER_NAME,
 			record.clusterUID,
--- a/frontend/src/container/InfraMonitoringK8s/DaemonSets/K8sDaemonSetsList.tsx
+++ b/frontend/src/container/InfraMonitoringK8s/DaemonSets/K8sDaemonSetsList.tsx
@@ -1,7 +1,6 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 // eslint-disable-next-line no-restricted-imports
 import { useSelector } from 'react-redux';
-import { useSearchParams } from 'react-router-dom-v5-compat';
 import { LoadingOutlined } from '@ant-design/icons';
 import {
 	Button,
@@ -72,8 +71,6 @@ function K8sDaemonSetsList({
 	const { maxTime, minTime } = useSelector<AppState, GlobalReducer>(
 		(state) => state.globalTime,
 	);
-
-	const [searchParams] = useSearchParams();
 
 	const [currentPage, setCurrentPage] = useInfraMonitoringCurrentPage();
 	const [filtersInitialised, setFiltersInitialised] = useState(false);
@@ -433,7 +430,7 @@ function K8sDaemonSetsList({
 	]);
 
 	const openDaemonSetInNewTab = (record: K8sDaemonSetsRowData): void => {
-		const newParams = new URLSearchParams(searchParams);
+		const newParams = new URLSearchParams(document.location.search);
 		newParams.set(
 			INFRA_MONITORING_K8S_PARAMS_KEYS.DAEMONSET_UID,
 			record.daemonsetUID,
--- a/frontend/src/container/InfraMonitoringK8s/Deployments/K8sDeploymentsList.tsx
+++ b/frontend/src/container/InfraMonitoringK8s/Deployments/K8sDeploymentsList.tsx
@@ -1,7 +1,6 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 // eslint-disable-next-line no-restricted-imports
 import { useSelector } from 'react-redux';
-import { useSearchParams } from 'react-router-dom-v5-compat';
 import { LoadingOutlined } from '@ant-design/icons';
 import {
 	Button,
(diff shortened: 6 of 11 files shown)


frontend/src/container/InfraMonitoringK8s/utils.tsx
```diff
@@ -293,7 +294,14 @@ export const getK8sPodsListColumns = (
 		return filteredColumns as ColumnType<K8sPodsRowData>[];
 	}
 
-	return updatedColumnsConfig as ColumnType<K8sPodsRowData>[];
+	for (const column of updatedColumnsConfig) {
+		if (column.sorter && column.key === defaultOrderBy?.columnName) {
+			column.defaultSortOrder =
+				defaultOrderBy?.order === 'asc' ? 'ascend' : 'descend';
+		}
+	}
```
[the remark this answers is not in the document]
H4ad: Default Sort will not be used after initial render, so that's the idea to mutate this and force the default sort on page refresh.

frontend/src/container/InfraMonitoringK8s/hooks.ts
```diff
@@ -0,0 +1,164 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
```
AshwinBhatkal: please remove this comment

AshwinBhatkal: below one as well
Let's add the right type

frontend/src/container/InfraMonitoringK8s/InfraMonitoringK8s.tsx
```diff
@@ -317,10 +305,8 @@ export default function InfraMonitoringK8s(): JSX.Element {
 	const handleCategoryChange = (key: string | string[]): void => {
 		if (Array.isArray(key) && key.length > 0) {
 			setSelectedCategory(key[0] as string);
-			setSearchParams({
-				[INFRA_MONITORING_K8S_PARAMS_KEYS.CATEGORY]: key[0] as string,
-			});
 			// Reset filters
+			setFilters(null);
```
AshwinBhatkal, line 317: Is this the correct equivalent of the previous setSearchParams?

frontend/src/container/InfraMonitoringK8s/utils.tsx
```diff
@@ -148,7 +148,7 @@ export const dummyColumnConfig = {
 	className: 'column column-dummy',
 };
 
-const columnsConfig = [
+const columnsConfig: ColumnType<any>[] = [
```
AshwinBhatkal: Let's type them correctly instead of any

frontend/src/container/InfraMonitoringK8s/StatefulSets/K8sStatefulSetsList.tsx
```diff
@@ -625,18 +574,14 @@ function K8sStatefulSetsList({
 			setCurrentPage(1);
 			setGroupBy(groupBy);
 			setExpandedRowKeys([]);
-			setSearchParams({
```
AshwinBhatkal, line 657: I see this setSearchParams removed from everywhere. What is the newer implementation for this?

AshwinBhatkal: Is it that the individual setState's are now just enough to update query and we don't need a separate state?

frontend/src/container/InfraMonitoringK8s/InfraMonitoringK8s.tsx
AshwinBhatkal: Got this answer

frontend/src/container/InfraMonitoringK8s/StatefulSets/K8sStatefulSetsList.tsx
```diff
@@ -61,55 +66,26 @@ function K8sStatefulSetsList({
...
-		);
-		if (statefulSetUID) {
-			return statefulSetUID;
-		}
-		return null;
-	});
+	const [
+		selectedStatefulSetUID,
+		setselectedStatefulSetUID,
+	] = useInfraMonitoringStatefulSetUID();
```
AshwinBhatkal, line 85: Is this taken care of correctly?

frontend/src/container/InfraMonitoringK8s/Namespaces/K8sNamespacesList.tsx
```diff
@@ -62,53 +66,25 @@ function K8sNamespacesList({
...
-	const [groupBy, setGroupBy] = useState<IBuilderQuery['groupBy']>(() => {
-		const groupBy = searchParams.get(INFRA_MONITORING_K8S_PARAMS_KEYS.GROUP_BY);
-		if (groupBy) {
-			const decoded = decodeURIComponent(groupBy);
-			const parsed = JSON.parse(decoded);
-			return parsed as IBuilderQuery['groupBy'];
-		}
-		return [];
-	});
+	const [groupBy, setGroupBy] = useInfraMonitoringGroupBy();
```
AshwinBhatkal, line 88: Feels like this piece code can be put into a hook. See similar usages

frontend/src/container/InfraMonitoringK8s/Jobs/JobDetails/JobDetails.tsx
```diff
@@ -90,23 +92,21 @@ function JobDetails({
...
 
-	const [searchParams, setSearchParams] = useSearchParams();
-	const [selectedView, setSelectedView] = useState<VIEWS>(() => {
-		const view = searchParams.get(INFRA_MONITORING_K8S_PARAMS_KEYS.VIEW);
-		if (view) {
-			return view as VIEWS;
-		}
-		return VIEWS.METRICS;
-	});
+	const [selectedView, setSelectedView] = useInfraMonitoringView();
```
AshwinBhatkal, line 95: Here, this common code can be put inside a common hook

At least these new lines. Maybe the whole component can be made common with some extending props

frontend/src/container/InfraMonitoringK8s/__tests__/K8sHeader.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+	};
+
+	const renderComponent = (
+		searchParams?: string | Record<string, string>,
+	): ReturnType<typeof render> => {
+		const Wrapper = withNuqsTestingAdapter({ searchParams: searchParams ?? {} });
+		return render(
+			<Wrapper>
+				<QueryClientProvider client={queryClient}>
+					<Provider store={store}>
```
AshwinBhatkal: What is this provider required for here?

frontend/src/container/InfraMonitoringK8s/__tests__/StatefulSets/StatefulSetDetails/StatefulSetDetails.test.tsx
```diff
@@ -23,17 +26,19 @@ describe('StatefulSetDetails', () => {
...
-							statefulSet={mockStatefulSet}
-							isModalTimeSelection
-							onClose={mockOnClose}
-						/>
-					</MemoryRouter>
-				</Provider>
-			</QueryClientProvider>,
+			<Wrapper>
+				<QueryClientProvider client={queryClient}>
+					<Provider store={store}>
```
AshwinBhatkal: do we need provider in any of these?

frontend/src/container/InfraMonitoringHosts/utils.tsx
```diff
@@ -127,6 +130,17 @@ export const getHostListsQuery = (): HostListPayload => ({
 	orderBy: { columnName: 'cpu', order: 'desc' },
 });
 
+function mapOrderByToSortOrder(
+	column: string,
+	orderBy: OrderBySchemaType,
+): SortOrder | undefined {
+	return orderBy?.columnName === column
```
AshwinBhatkal, line 137: can this be done with zod?

frontend/src/container/InfraMonitoringHosts/HostsList.tsx
```diff
@@ -242,7 +216,7 @@ function HostsList(): JSX.Element {
 							</div>
 						)}
 						<HostsListControls
-							filters={filters}
+							filters={filters ?? undefined}
```
AshwinBhatkal: what is expected here? 

If it is null, we send in undefined? What about updating prop instead?

frontend/src/container/InfraMonitoringK8s/InfraMonitoringK8s.tsx
H4ad: Yes, it's the same

frontend/src/container/InfraMonitoringK8s/StatefulSets/K8sStatefulSetsList.tsx
H4ad: In the past we had useState + setSearchParams, this change removes all the useState and use the nuqs directly, so one hook is enough to store data in URL and also persist the state in memory to be used by the components.

That's why I removed this setSearch everywhere and almost no change regarging the setState, because I kept the same name and just change the implementation.

frontend/src/container/InfraMonitoringK8s/Jobs/JobDetails/JobDetails.tsx
H4ad: I wanted to do that but I'm not sure if those pages will remain 100% similar in the future. 

But I will try to push to standardize all these components to use one common component since they are almost the same.

I need to upgrade the query builder version, so I will likely update all these component to just one to remove the amount of duplicated code.

frontend/src/container/InfraMonitoringK8s/Namespaces/K8sNamespacesList.tsx
H4ad: I will do a refactor like this when I combine all these components into a single one.

H4ad: Right know, there's no usage on combining them into a hook since it will produce the same amount of lines.

frontend/src/container/InfraMonitoringK8s/StatefulSets/K8sStatefulSetsList.tsx
H4ad: What do you mean exactly? The previous was just verbose, this code is doing the same with less lines, it already gets the value from URL by using nuqs hook.

frontend/src/container/InfraMonitoringHosts/HostsList.tsx
H4ad: filters has the type of `TagFilter | undefined`, I didn't want to change the type of the `IBuilderQuery['filters']`, so I just coalesce to `undefined` to not have issues with type.

But I updated the `HostsListControls` to accept null.

frontend/src/container/InfraMonitoringHosts/utils.tsx
H4ad: We can introduce a custom serializer to read ascend/asc and return only ascend, and then on set, only store in the URL asc.

But we will likely replace this table component to another lib, so I didn't bother to try and just introduced a function to keep the old contract `asc/desc` on URL and also map the default order for Antd table.

frontend/src/container/InfraMonitoringK8s/__tests__/StatefulSets/StatefulSetDetails/StatefulSetDetails.test.tsx
H4ad: Some of the components use `min/max` from the GlobalTime.

But for some of them that doesn't need, I removed it.

frontend/src/container/InfraMonitoringK8s/__tests__/K8sHeader.test.tsx
H4ad: Kept the same behavior of the other tests, but for this component is not needed, removing.

frontend/src/container/InfraMonitoringK8s/Namespaces/K8sNamespacesList.tsx
AshwinBhatkal: Yep. Can be done in a different PR later

frontend/src/container/InfraMonitoringK8s/StatefulSets/K8sStatefulSetsList.tsx
AshwinBhatkal: I meant the same. Previous was indeed verbose. If this is same, this is good

image

frontend/src/container/InfraMonitoringK8s/Jobs/JobDetails/JobDetails.tsx
AshwinBhatkal: Works. Even if pages are not similar later, this seems more like a config that can be set when page is mounted and anything on the page is updated. So, we can think about it in those lines when we do the query builder version update

frontend/src/container/InfraMonitoringHosts/utils.tsx
AshwinBhatkal: out of context of this PR. Can we make a generic serializer that can be used at other places as well?

H4ad: Sure, maybe we can have a folder for custom schemas/serializers, but the creation is quite simple: https://nuqs.dev/docs/parsers/making-your-own

Issue #8997, opened 2025-09-03
[authn] add option to completely disable authentication for internal deployments


taylorsmithgg:
SigNoz's authentication system causes operational pain in internal environments where auth is unnecessary. We need a simple option to completely disable authentication.

## Related Issues

- **#6197** - Trusted Header Authentication (still requires auth headers)
- **#5137** - Disable password login after SSO (enforces auth, opposite direction)
- **#1188** - External SSO support (closed, moved to enterprise)

This request is different: **no authentication needed at all**.

## Current Pain Points

1. **Constant session timeouts** despite 24-hour JWT expiration settings
2. **User sync failures** between components causing workspace issues  
3. **Complex configuration** requiring multiple overlapping variables:
   ```yaml
   SIGNOZ_JWT_SECRET, JWT_SECRET, SIGNOZ_JWT_EXPIRATION_TIME, 
   JWT_EXPIRATION_TIME, SIGNOZ_SESSION_TIMEOUT, SIGNOZ_TOKEN_EXPIRY
   ```
4. **Operational overhead** - custom deployment scripts, validation checks, browser cache clearing

## Proposed Solution

Add a simple environment variable:
```yaml
SIGNOZ_AUTH_ENABLED: "false"
```

**Expected behavior**: Direct access to SigNoz without login screens, JWT tokens, or session management.

## Use Cases

- Internal corporate networks (secure perimeter)
- Development/testing environments
- CI/CD monitoring tools
- Embedded deployments

## Benefits

- Eliminates session timeout issues
- Removes user sync complexity
- Simplifies deployment configuration
- Enables zero-config observability for secure environments

**Environment**: SigNoz v0.93.0, Kubernetes/Helm, internal network

Kotnstantin:
+

danbopes:
I ran into this pain point today. We have a cloudflare managed authentication system, and wish to disable authentication all together.

taylorsmithgg:
> I ran into this pain point today. We have a cloudflare managed authentication system, and wish to disable authentication all together.

I have been using my fork in our environment since I published this and it has been working well. I have not seen any issues so far.

danbopes:
> > I ran into this pain point today. We have a cloudflare managed authentication system, and wish to disable authentication all together.
> 
> I have been using my fork in our environment since I published this and it has been working well. I have not seen any issues so far.

You should submit a PR. It'll be much more likely to get in.

taylorsmithgg:
[...]

PR #10641, opened 2026-03-18
feat(identn): implement an impersonation identn

therealpandey:
### 📄 Summary

- Add an impersonation IdentN provider that bypasses authentication by impersonating the root user on every request

- Expose IdentN config (enabled/disabled booleans only, no secrets) via /api/v1/global/config and change that endpoint to OpenAccess so frontends can read it pre-auth

- Refactor IdentN resolver to explicitly instantiate only enabled providers (removing the now-redundant Enabled() method from the IdentN interface)


therealpandey:
Part of https://github.com/SigNoz/signoz/issues/8997

therealpandey:
@claude review

pkg/identn/resolver.go
```diff
@@ -1,29 +1,67 @@
...
 
-func NewIdentNResolver(providerSettings factory.ProviderSettings, identNs ...IdentN) IdentNResolver {
-	enabledIdentNs := []IdentN{}
+func NewIdentNResolver(ctx context.Context, providerSettings factory.ProviderSettings, identNConfig Config, identNFactories factory.NamedMap[factory.ProviderFactory[IdentN, Config]]) (IdentNResolver, error) {
+	identNs := []IdentN{}
 
-	for _, identN := range identNs {
-		if identN.Enabled() {
-			enabledIdentNs = append(enabledIdentNs, identN)
+	if identNConfig.Impersonation.Enabled {
```
vikrantgupta25, line 19: if impersonation is enabled we don't need to instantiate any other provider factory

therealpandey: There's a validation for this at the config layer which takes care of the correct configuration to be present before this is hit.

The idea is to rely on config because a config will always be validated before it comes here. What do you think?

PR #10643, opened 2026-03-18
feat: adding handlers with OpenAPI specs

swapnil-signoz:
### 📄 Summary
- Adding handler skeleton and unimplemented methods
- OpenAPI is attached with handlers
- Made few changes for consistent types and domain structs

### DIFF
--- a/pkg/querier/builder_query.go
+++ b/pkg/querier/builder_query.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"log/slog"
 	"strconv"
 	"strings"
 	"time"
@@ -11,6 +12,7 @@ import (
 	"github.com/ClickHouse/clickhouse-go/v2"
 	"github.com/SigNoz/signoz/pkg/errors"
 	"github.com/SigNoz/signoz/pkg/telemetrystore"
+	"github.com/SigNoz/signoz/pkg/telemetrytraces"
 	"github.com/SigNoz/signoz/pkg/types/ctxtypes"
 	"github.com/SigNoz/signoz/pkg/types/instrumentationtypes"
 	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
@@ -18,6 +20,7 @@ import (
 )
 
 type builderQuery[T any] struct {
+	logger         *slog.Logger
 	telemetryStore telemetrystore.TelemetryStore
 	stmtBuilder    qbtypes.StatementBuilder[T]
 	spec           qbtypes.QueryBuilderQuery[T]
@@ -31,6 +34,7 @@ type builderQuery[T any] struct {
 var _ qbtypes.Query = (*builderQuery[any])(nil)
 
 func newBuilderQuery[T any](
+	logger *slog.Logger,
 	telemetryStore telemetrystore.TelemetryStore,
 	stmtBuilder qbtypes.StatementBuilder[T],
 	spec qbtypes.QueryBuilderQuery[T],
@@ -39,6 +43,7 @@ func newBuilderQuery[T any](
 	variables map[string]qbtypes.VariableItem,
 ) *builderQuery[T] {
 	return &builderQuery[T]{
+		logger:         logger,
 		telemetryStore: telemetryStore,
 		stmtBuilder:    stmtBuilder,
 		spec:           spec,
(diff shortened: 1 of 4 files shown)


pkg/apiserver/signozapiserver/cloudintegration.go
```diff
@@ -0,0 +1,216 @@
...
+	"net/http"
+
+	"github.com/SigNoz/signoz/pkg/http/handler"
+	"github.com/SigNoz/signoz/pkg/types"
+	citypes "github.com/SigNoz/signoz/pkg/types/cloudintegrationtypes"
+	"github.com/gorilla/mux"
+)
+
+func (provider *provider) addCloudIntegrationRoutes(router *mux.Router) error {
+	if err := router.Handle("/api/v1/cloud_integrations/{cloud_provider}/accounts/connection_artifact", handler.New(
```
vikrantgupta25:
```suggestion
	if err := router.Handle("/api/v1/cloud_integrations/{cloud_provider}/accounts", handler.New(
```

pkg/apiserver/signozapiserver/cloudintegration.go
```diff
@@ -0,0 +1,216 @@
...
+			Deprecated:          false,
+			SecuritySchemes:     newSecuritySchemes(types.RoleAdmin),
+		},
+	)).Methods(http.MethodPut).GetError(); err != nil {
+		return err
+	}
+
+	// Agent check-in endpoint is kept same as older one to maintain backward compatibility with already deployed agents.
+	// In the future, this endpoint will be deprecated and a new endpoint will be introduced for consistency with above endpoints.
+	if err := router.Handle("/api/v1/cloud-integrations/{cloud_provider}/agent-check-in", handler.New(
```
vikrantgupta25, line 175: is this endpoint still in use ? your new handler returns panic, this will fail existing cloud integrations ?

pkg/apiserver/signozapiserver/cloudintegration.go
swapnil-signoz: I wanted to be explicit about what resource will be returned. If we make this change, we should return the account along with artifact then.

pkg/apiserver/signozapiserver/cloudintegration.go
swapnil-signoz: I saw the behaviour by our router, if we register multiple exact same routes it only respects the first route and its handler. In this particular case the older cloud integration routes get registered first, so it is okay to have this endpoint registered here. Once we remove the older code, we will use this endpoint and keep the same response till we deprecate it.

pkg/modules/cloudintegration/implcloudintegration/handler.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/SigNoz/signoz/pkg/modules/cloudintegration"
+)
+
+type handler struct{}
+
+func NewHandler() cloudintegration.Handler {
+	return &handler{}
+}
+
+func (h handler) GetConnectionArtifact(writer http.ResponseWriter, request *http.Request) {
```
vikrantgupta25:
```suggestion
func (hander *handler) GetConnectionArtifact(writer http.ResponseWriter, request *http.Request) {
```

pkg/types/cloudintegrationtypes/account.go
```diff
@@ -10,34 +10,35 @@ import (
...
-	RemovedAt         *time.Time        `json:"removedAt,omitempty"`
-	AgentReport       *AgentReport      `json:"agentReport,omitempty"`
-	OrgID             valuer.UUID       `json:"orgID"`
-	Config            *AccountConfig    `json:"config,omitempty"`
+	ProviderAccountID *string           `json:"providerAccountId" required:"true" nullable:"true"`
+	Provider          CloudProviderType `json:"provider" required:"true"`
+	RemovedAt         *time.Time        `json:"removedAt,omitempty" required:"true" nullable:"true"`
+	AgentReport       *AgentReport      `json:"agentReport,omitempty" required:"true" nullable:"true"`
+	OrgID             valuer.UUID       `json:"orgId" required:"true"`
+	Config            *AccountConfig    `json:"config,omitempty" required:"true" nullable:"false"`
```
vikrantgupta25: i am confused here, omitempty yet required ?

pkg/apiserver/signozapiserver/cloudintegration.go
vikrantgupta25: It's not about being explicit, your POST call can return response which is derived entity for the created resource. Imagine a flow being 

```
-> /accounts -> create account -> /accounts/{id}/connection-artifact -> gets the artifact 
```

now this becomes clear separation that you first create an account and then use the ID to fetch the connection artifact. Now eitheryou make two step flow or just return the derived artifact in the create call itself

vikrantgupta25 (changes requested):

pkg/types/cloudintegrationtypes/account.go
swapnil-signoz: fixed

PR #10646, opened 2026-03-19
fix: allow gateway apis on editor access

karanbalani:
## Pull Request

---

### 📄 Summary
- allow gateway (ingestion keys) apis access to editors and admins both
- fix integration tests to make sure editors are allowed to do crud on ingestion keys and their limits

Closes: https://github.com/SigNoz/platform-pod/issues/1862
---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---


karanbalani:
Related: https://github.com/SigNoz/platform-pod/issues/1862

tests/integration/src/ingestionkeys/01_ingestion_keys.py
```diff
@@ -32,6 +35,31 @@ def test_apply_license(
     add_license(signoz, make_http_mocks, get_token)
 
 
+def test_create_editor_user(
```
vikrantgupta25, line 38: what do you think, should it be here or just create an editor user fixture which everyone can use ?

PR #10647, opened 2026-03-19
fix: replace v3 filter_suggestions source table

piyushsingariya:
## Pull Request

---

### 📄 Summary

This PR aims to replace the source table for suggestions in v3 api to fix an issue with logs pipelines filter

#### Issues closed by this PR

Closes SigNoz/engineering-pod#4248

---

### ✅ Change Type
_Select all that apply_

- [x] 🐛 Bug fix
- [x] ♻️ Refactor

---

### 🐛 Bug Context

Logs Pipelines uses /v3/filter_suggestion; User has been trying to add a filter for `ddtags` but this is not appearing in suggestions but user forced entered the filter as `ddtags Exists` this result into FE not knowing `ddtags` to be part of attributes and sent ddtags as column -> resulting into faulty IF expression -> resulting into REGEX Parser Failing (from user's pov) 

Suggestions wasn't working because the api is sourcing it's suggestions from table `distributed_tag_attributes_v2` which skips ingesting this attribute specifically because the associated value exceeds defined limits.

```
SELECT *
FROM distributed_tag_attributes_v2
WHERE lower(tag_key) LIKE '%ddtags%'

Query id: e21d3a78-54c9-4a04-8c23-7d57bc6ba950

Ok.

0 rows in set. Elapsed: 0.007 sec.

SELECT *
FROM distributed_logs_attribute_keys
WHERE lower(name) LIKE '%ddtags%'

Query id: a49b4a44-f761-4e14-bcc5-e7d87bdc31a9

   ┌─name───┬─datatype─┐
1. │ ddtags │ string   │
   └────────┴──────────┘
```

#### Fix Strategy

Use `distributed_logs_attribute_keys` as source table for suggestions similar to how v5 does

---

- Tests added/updated: N/A
- Manual verification: Manually tested
- Edge cases covered: N/A

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

I am bit unsure about the backward compatibility of v3 api hence would like a thought from reviewers on this.

---


nityanandagohain:
Discussed over call to fix the pipeline save endpoint to use enrich query instead for fixing this here and later move to QB v5 for solving the suggestion issues.

PR #10657, opened 2026-03-19
fix(infra-monitoring): request loop when click to visualize volume

H4ad:
## Pull Request

This PR fixes the loop bug on select volumes caused by the min/max time being updates being included in the query to fetch data for the volume details charts.

---

This PR fixes the bug by not adding the min/max time into the query keys.

Before:

https://github.com/user-attachments/assets/9a8d5300-ccbf-4277-b071-b474c480e34b

After:

https://github.com/user-attachments/assets/375686d8-9fdc-4bad-864f-56e61a10bf50

Closes #8677

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

The bug caused the page to enter in a loop of refetching the queries to get the volume data.

This was caused by min/max being updated for some reason after the queries for chart run.

Remove min/max from the query key.

---

- Tests added/updated: Yes
- Manual verification: Yes

---

| Field | Value |
|------|-------|
| Deployment Type | Cloud / OSS / Enterprise |
| Change Type | Bug Fix |
| Description | Fixed the loop bug when opening volumes inside Infrastructure Monitoring |

---

### 📋 Checklist
- [x] Tests added or explicitly not required
- [x] Manually tested
- [ ] Breaking changes documented
- [ ] Backward compatibility considered

PR #10658, opened 2026-03-19
fix(infra-monitoring): volume details charts rendering undefined as legend

H4ad:
## Pull Request
---

The labels for these charts were all undefined due to legend using variables instead of hardcoded values like the other options.

Before:

image

After:

image

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

The legends were not being rendered.

The backend query does not populate the legend using variables that are not in groupBy (not 100% sure if this is correct).

Use static values like the other charts inside Infrastructure monitoring.

---

- Tests added/updated:
- Manual verification: Yes
- Edge cases covered:

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Volume Details page of Infrastructure Monitoring now shows correctly the legend for the charts. |

---

### 📋 Checklist
- [x] Manually tested


H4ad:
Due to #10657 not being merged yet, there's no way to test locally without applying that fix first.

PR #10665, opened 2026-03-21
feat(instrumentation): add OTel exception semantic convention log handler

therealpandey:
## Summary
- Add `errors.Attr()` helper that returns `slog.Any("exception", err)` for standardized error logging
- Add `loghandler.NewException()` wrapper that enriches error log records with OTel exception semantic convention attributes as flat keys (`exception.type`, `exception.code`, `exception.message`, `exception.stacktrace`)
- Wire exception handler into `NewLogger` so all loggers (SDK and CLI) get it uniformly
- Remove `LogValue()` from `errors.base` as the handler now owns structured error serialization
- Replace all `"error", err` slog key-value pairs with `errors.Attr(err)` across the codebase
- Change sloglint from `kv-only` to `attr-only`, enforcing typed attributes (`slog.String`, `slog.Any`, etc.) to prevent accidental `!BADKEY` errors in logs caused by mismatched key-value pairs
- Tighten `slog.Any` usages to specific types (`slog.String`, `slog.Uint64`) where the value type is known
- Use `Unwrapb` message (`m`) for `exception.message` instead of `foundErr.Error()`, which resolves to the inner cause for wrapped errors
- Capture stacktrace at error creation time using `runtime.Callers` (Thanos-style `[]uintptr`), formatted lazily on demand — the exception handler reads the stacktrace from the error showing where it originated, not where it was logged

## Follow-up
- Remove convenience wrappers (`NewInternalf`, `WrapNotFoundf`, etc.) that add an extra frame to the stacktrace — 661 usages across 116 files, to be done separately

## Output format
```json
{
  "level": "ERROR",
  "msg": "failed to get user",
  "exception.type": "not-found",
  "exception.code": "user_not_found",
  "exception.message": "user with id 123 not found",
  "exception.stacktrace": "github.com/SigNoz/signoz/pkg/modules/user/impluser.(*module).GetUser\n\t/path/to/module.go:42\n..."
}
```

## Test plan
- [x] `go test -race ./pkg/instrumentation/loghandler/...` — new exception handler tests
- [x] `go test -race ./pkg/errors/...` — errors.Attr and stacktrace tests
- [x] `go build ./...` — full build compiles with all replacements
- [x] `make go-lint` — 0 lint issues with attr-only enforcement

### DIFF
--- a/pkg/instrumentation/loghandler/exception.go
+++ b/pkg/instrumentation/loghandler/exception.go
@@ -34,21 +34,19 @@ func (h *exception) Wrap(next LogHandler) LogHandler {
 
 		t, c, m, _, _, _ := errors.Unwrapb(foundErr)
 
-		attrs := []any{
-			slog.String("type", t.String()),
-			slog.String("code", c.String()),
-			slog.String("message", m),
-		}
+		newRecord.AddAttrs(
+			slog.String("exception.type", t.String()),
+			slog.String("exception.code", c.String()),
+			slog.String("exception.message", m),
+		)
 
 		// Use the stacktrace captured at error creation time if available.
 		type stacktracer interface {
 			Stacktrace() string
 		}
 		if st, ok := foundErr.(stacktracer); ok && st.Stacktrace() != "" {
-			attrs = append(attrs, slog.String("stacktrace", st.Stacktrace()))
+			newRecord.AddAttrs(slog.String("exception.stacktrace", st.Stacktrace()))
 		}
-
-		newRecord.AddAttrs(slog.Group("exception", attrs...))
 
 		return next.Handle(ctx, newRecord)
 	})
--- a/pkg/instrumentation/loghandler/exception_test.go
+++ b/pkg/instrumentation/loghandler/exception_test.go
@@ -83,17 +83,17 @@ func TestException(t *testing.T) {
 			assert.Equal(t, "operation failed", m["msg"])
 
 			if tc.hasException {
-				exc, ok := m["exception"].(map[string]any)
-				require.True(t, ok, "exception should be a nested object")
-				assert.Equal(t, tc.exceptionType, exc["type"])
-				assert.Equal(t, tc.exceptionCode, exc["code"])
-				assert.Equal(t, tc.exceptionMessage, exc["message"])
-				stacktrace, ok := exc["stacktrace"].(string)
+				assert.Equal(t, tc.exceptionType, m["exception.type"])
+				assert.Equal(t, tc.exceptionCode, m["exception.code"])
+				assert.Equal(t, tc.exceptionMessage, m["exception.message"])
+				stacktrace, ok := m["exception.stacktrace"].(string)
 				require.True(t, ok)
 				assert.Contains(t, stacktrace, "exception_test.go:")
 			} else {
-				_, isMap := m["exception"].(map[string]any)
-				assert.False(t, isMap, "exception should not be a structured object")
+				assert.Nil(t, m["exception.type"])
+				assert.Nil(t, m["exception.code"])
+				assert.Nil(t, m["exception.message"])
+				assert.Nil(t, m["exception.stacktrace"])
 			}
 		})
 	}


therealpandey:
Part of https://github.com/SigNoz/platform-pod/issues/1383

pkg/alertmanager/alertmanagerserver/dispatcher.go
```diff
@@ -111,24 +111,24 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
 			if !ok || alertWrapper == nil {
 				// Iterator exhausted for some reason.
 				if err := it.Err(); err != nil {
-					d.logger.ErrorContext(d.ctx, "Error on alert update", "err", err)
+					d.logger.ErrorContext(d.ctx, "Error on alert update", errors.Attr(err))
```
therealpandey, line 114: @srikanthccv this was not correct earlier as this had "err" instead of the "error" used across the codebase.

.golangci.yml
```diff
@@ -35,7 +35,7 @@ linters:
         - identical
     sloglint:
       no-mixed-args: true
-      kv-only: true
+      attr-only: true
```
srikanthccv, line 38: 👍

pkg/telemetrytraces/trace_operator_statement_builder.go
```diff
@@ -69,8 +69,8 @@ func (b *traceOperatorStatementBuilder) Build(
 	}
 
 	b.logger.DebugContext(ctx, "Building trace operator query",
-		"expression", query.Expression,
-		"request_type", requestType)
+		slog.String("expression", query.Expression),
+		slog.Any("request_type", requestType))
```
srikanthccv, line 73: question, for the custom types, example new enum kinds, do we need to use slog.Any , can we use slog.String and since it implements the stringer interface, it should work? or does it actually need string value?

pkg/instrumentation/loghandler/exception.go
```diff
@@ -0,0 +1,55 @@
...
+
+		// Use the stacktrace captured at error creation time if available.
+		type stacktracer interface {
+			Stacktrace() string
+		}
+		if st, ok := foundErr.(stacktracer); ok && st.Stacktrace() != "" {
+			attrs = append(attrs, slog.String("stacktrace", st.Stacktrace()))
+		}
+
+		newRecord.AddAttrs(slog.Group("exception", attrs...))
```
srikanthccv: is it going to be nested json like `{"exception":{"type":...` or they going to be flat key like `{"exception.type":...`

therealpandey: It's going to be a nested json not a flat key. Do we want flat keys?

pkg/telemetrytraces/trace_operator_statement_builder.go
therealpandey: Since it implements fmt.Stringer, Any should resolve to Stringer and call `String()` of the valuers. I had the same doubt regarding valuers. You cannot directly put the Struct as `slog.String` requires a string var.

therealpandey:
> can you put in some lint check to discourage from using `slog.Any("exception"` or `slog.Any("error"` and instead use `errors.Attr` Or is it deliberate to not add it?

Let me try adding a lint rule.

.golangci.yml
therealpandey: I noticed a couple of BADKEY errors in our control plane codebase. So this is pre-emptive.

> Change sloglint from kv-only to attr-only, enforcing typed attributes (slog.String, slog.Any, etc.) to prevent accidental !BADKEY errors in logs caused by mismatched key-value pairs

pkg/instrumentation/loghandler/exception.go
therealpandey: The specification lays out flat keys I believe, but we would parse it as JSON keys correct?

srikanthccv: do we already parse on NW, if so we should be good otherwise I would suggest flat keys to avoid additional parsing step.

therealpandey:
@srikanthccv let me try and address the lint rule in a follow-up PR. The problem is with `forbidgo` which is not that straightforward.

For example:
```
        - pattern: slog\.Any\("error"
          msg: Use errors.Attr(err) instead of slog.Any("error", err).
        - pattern: slog\.Any\("exception"
          msg: Use errors.Attr(err) instead of slog.Any("exception", err).
        - pattern: slog\.String\("error"
          msg: Use errors.Attr(err) instead of slog.String("error", err).
```

does not work.

pkg/instrumentation/loghandler/exception.go
therealpandey: Yes we already parse.

therealpandey: Discussed offline. Converted to flat keys.

PR #10667, opened 2026-03-21
refactor(instrumentation): flatten code source into flat keys

therealpandey:
## Summary
- Add a `Source` log handler wrapper that extracts source info from `record.PC` and adds flat `code.filepath`, `code.function`, `code.lineno` keys to log output
- Remove `AddSource: true` from the JSON handler since the wrapper handles source resolution directly
- Produces flat keys matching OTel semantic conventions, consistent with `exception.type`, `exception.message` etc.

## Test plan
- [x] Unit test verifying flat keys are present and nested `source`/`code` group is absent
- [x] All existing instrumentation tests pass


therealpandey:
Part of https://github.com/SigNoz/platform-pod/issues/1383

PR #10670, opened 2026-03-22
fix: allow deleting pending admin invite (fixes #10623)

### LINKED ISSUE #10623: Cannot delete pending admin invite, last admin guard false positive
## Bug description

Deleting a pending invite via **DELETE /api/v1/invite/{id}** returns 403 Forbidden when the invited user has the Admin role, even when the caller is the founding admin of the instance.

The `DeleteInvite` handler routes through `DeleteUser` (`pkg/modules/user/impluser/module.go`). The last-admin guard at line 393 counts only active admins via `GetActiveUsersByRoleAndOrgID` (correctly returning 1, the founding admin), but then checks `user.Role == types.RoleAdmin` on the target user without considering its `pending_invite` status. The pending user was never counted in the active admin query, so deleting it cannot reduce the active admin count. The guard fires incorrectly.

Inviting as Viewer or Editor and then deleting works fine. The guard only triggers for Admin role.

## Expected behavior

Deleting a pending admin invite should succeed. A `pending_invite` user is not an active admin and removing them cannot make the caller the "last admin."

## How to reproduce

1. Deploy SigNoz, log in as the founding admin (only admin in the org)
2. Invite a new user with **Admin** role
3. Attempt to cancel/delete the pending invite via the UI
4. Observe 403 Forbidden

## Version information
* **Signoz version**: v0.115.0
* **Browser version**: Brave 1.87.191
* **Your OS and version**: Kubernetes (Helm chart), Talos Linux nodes
* **Your CPU Architecture**: ARM (arm64)

## Additional context

The `DeleteUser` code path is reasonable for invite deletion since invites create real user records. The last-admin guard just needs to exclude `pending_invite` status users from the check, since they are not yet active admins.

Jatin123lodhi:
## Pull Request

---

### 📄 Summary
-The last-admin guard in DeleteUser incorrectly blocked deletion of pending admin invites.

image

---

### ✅ Change Type

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
The last-admin guard in DeleteUser (around line 296) checked user.Role == types.RoleAdmin without considering user.Status. Pending invites create user records with Admin role, but they are not counted by GetActiveUsersByRoleAndOrgID. Deleting a pending admin invite therefore cannot remove the last active admin, yet the guard blocked the operation.

Skip the last-admin guard when user.Status == types.UserStatusPendingInvite. These users are never included in the active admin count, so deleting them cannot make the caller the last admin.
---

- Tests added/updated: N/A
- Manual verification: Invite a user as Admin, cancel the pending invite via UI, confirm 200/204 instead of 403
- Edge cases covered: Viewer/Editor invites and active admin deletion still protected; guard only skipped for pending invites

---

- Blast radius: Limited to DeleteUser when deleting users with UserStatusPendingInvite
- Potential regressions: None; pending invite users were never considered active admins
- Rollback plan: Revert the change to restore the previous behavior

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Fix deletion of pending admin invites being incorrectly blocked (403) when only one active admin exists |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers
The guard still runs for non-pending users. The only change is wrapping it in if user.Status != types.UserStatusPendingInvite { ... } so pending invite deletions bypass the check, since pending invite users are never included in the active admin count.

---

> [!NOTE]
> **Low Risk**
> Low risk, small conditional change to `DeleteUser` that only affects users in `pending_invite` status; active admin deletion protection remains unchanged.
> 
> **Overview**
> Fixes `DeleteUser` incorrectly blocking deletion of *pending admin invites* under the “last admin” safeguard.
> 
> The last-admin check now runs only for non-`pending_invite` users, allowing invite revocation/soft-delete for pending users while keeping the existing guard for active admins.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 056af999be0eaef52c0d813a70562f79cc3b3f5f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/modules/user/impluser/module.go
+++ b/pkg/modules/user/impluser/module.go
@@ -294,13 +294,16 @@ func (module *Module) DeleteUser(ctx context.Context, orgID valuer.UUID, id stri
 	}
 
 	// don't allow to delete the last admin user
-	adminUsers, err := module.store.GetActiveUsersByRoleAndOrgID(ctx, types.RoleAdmin, orgID)
-	if err != nil {
-		return err
-	}
+	// Skip for pending_invite: they're not active admins, so revoking their invite cannot reduce the active admin count
+	if user.Status != types.UserStatusPendingInvite {
+		adminUsers, err := module.store.GetActiveUsersByRoleAndOrgID(ctx, types.RoleAdmin, orgID)
+		if err != nil {
+			return err
+		}
 
-	if len(adminUsers) == 1 && user.Role == types.RoleAdmin {
-		return errors.New(errors.TypeForbidden, errors.CodeForbidden, "cannot delete the last admin")
+		if len(adminUsers) == 1 && user.Role == types.RoleAdmin {
+			return errors.New(errors.TypeForbidden, errors.CodeForbidden, "cannot delete the last admin")
+		}
 	}
 
 	// since revoke is idempotant multiple calls to revoke won't cause issues in case of retries


karanbalani:
Hi @Jatin123lodhi 

Thanks for identifying and raising a fix PR, highly appreciate this. 🙌

While this is a good solution, we are already planning to introduce some changes which is gonna remove this block entirely. Hopefully that would be merged before next release.

I am gonna close this PR since upcoming release is already gonna handle this differently.

PR #10672, opened 2026-03-22
test: fix flaky public dashboard query range integration test

therealpandey:
## Summary
- widen the fixed public dashboard default time range in the integration test from 10s to 10m
- align the test window with the seeded metrics inserted 1 to 5 minutes before the query
- keep the test validating the public widget query_range path after the querier started returning 404 for missing metric data in range (In https://github.com/SigNoz/signoz/pull/10560)


therealpandey:
Flaky test run here: https://github.com/SigNoz/signoz/actions/runs/23408932454

PR #10673, opened 2026-03-22
refactor(pprof): extract infrastructure provider

therealpandey:
## Summary
- extract pprof into a dedicated infrastructure provider with http and noop implementations
- wire pprof through top-level config and the shared service registry
- remove ad hoc pprof startup from community and ee query-service servers

## Testing
- `go test ./pkg/pprof/... ./pkg/signoz ./pkg/query-service/app ./ee/query-service/app ./cmd ./cmd/community ./cmd/enterprise`

### DIFF
--- a/pkg/pprof/httppprof/provider.go
+++ b/pkg/pprof/httppprof/provider.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"log/slog"
 	"net/http"
-	httppprof "net/http/pprof"
+	nethttppprof "net/http/pprof"
 	runtimepprof "runtime/pprof"
 
 	"github.com/SigNoz/signoz/pkg/factory"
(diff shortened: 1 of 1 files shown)


pkg/pprof/httppprof/provider.go
```diff
@@ -0,0 +1,58 @@
...
+	return provider.server.Start(ctx)
+}
+
+func (provider *provider) Stop(ctx context.Context) error {
+	return provider.server.Stop(ctx)
+}
+
+func newHandler() http.Handler {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/debug/pprof/", httppprof.Index)
```
vikrantgupta25: add reference for https://cs.opensource.google/go/go/+/refs/tags/go1.26.1:src/net/http/pprof/pprof.go;l=101 here ?

pkg/pprof/httppprof/provider.go
```diff
@@ -0,0 +1,58 @@
+package httppprof
+
+import (
+	"context"
+	"log/slog"
+	"net/http"
+	httppprof "net/http/pprof"
```
vikrantgupta25: same name of package and import

therealpandey: Did not understand this (insert sweat emoji). Is there an issue?

pkg/pprof/httppprof/provider.go
therealpandey: I've added a reference to `runtime/pprof` below so that you can jump to the exact go version file.

PR #10675, opened 2026-03-23
fix: bugs in dashboard filter expression

### LINKED ISSUE #10674: [bug] adding a dashboard variable via drilldown to APM metrics dashboard breaks the panels
## Bug: Dashboard panel filter expressions corrupted when managing dynamic variables

Two bugs in dashboard panel filter expressions when using dynamic variables.

### Bug 1: Dangling `AND` after variable deletion

**Steps to reproduce:**
1. Create a dashboard with a panel whose filter expression is:
   `(deployment.environment = $deployment.environment AND service.name = $service.name AND operation IN $top_level_operation)`
2. Delete the `top_level_operation` variable from the dashboard variables.

**Expected:** `(deployment.environment = $deployment.environment AND service.name = $service.name)`  
**Actual:** `(deployment.environment = $deployment.environment AND service.name = $service.name AND )`

> This happens only if the variable to add is in the query of the panel

### Bug 2: Duplicate variable names silently accepted

**Steps to reproduce:**
1. Create a dashboard variable named `env`.
2. Create another variable also named `env`.

**Expected:** An error is shown and the second variable is not created.  
**Actual:** The second variable is created silently, resulting in a duplicate.

> There is no failure check

### Environment
Frontend / Dashboard Variables

AshwinBhatkal:
### 📄 Summary

Two bugs in dashboard panel filter expression management when using dynamic variables.

**Bug 1 — Dangling `AND` after variable deletion:**  
When a dynamic variable whose key appeared last inside a parenthesised filter group was deleted, `removeKeysFromExpression` only absorbed a *following* conjunction. Since the last pair is followed by `)` (not a conjunction), nothing was absorbed and a trailing `AND ` was left inside the group.

Fixed by also absorbing the *preceding* `AND`/`OR` when no following conjunction is found.

**Bug 2 — Duplicate variable names:**  
`createVariable` had no guard against reusing an existing variable name. Added a check that shows an error toast (`notifications.error`) and returns early if the name is already taken.

##### Before

https://github.com/user-attachments/assets/e0a692c9-8e57-43f1-978f-280f3c1a06da

https://github.com/user-attachments/assets/637c9ac3-5ad0-44d5-9f88-535285d36c64

##### After

https://github.com/user-attachments/assets/a0353a93-2f49-4357-bf14-6f69b2cd0c22

---

### ✅ Change Type

- [x] 🐛 Bug fix

---

#### Root Cause

**Bug 1:** `removeKeysFromExpression` searched for a following `AND|OR` conjunction to absorb alongside the removed pair. When the removed pair was the last token before a closing `)`, no following conjunction existed and the cleanup regex `/(AND|OR)\s*$/` did not fire because the string ends with `)`.

**Bug 2:** `createVariable` had no duplicate-name validation before inserting into the variables map.

#### Fix Strategy

**Bug 1:** After the existing following-conjunction check, added a fallback that inspects the substring *before* the pair and absorbs a preceding `\s+(AND|OR)\s+` when no following conjunction was found.

**Bug 2:** Added an `Object.values(currentVariables).some(v => v.name === name)` guard at the top of `createVariable`. On match, `notifications.error` is called and the function returns early.

---

### 🧪 Testing Strategy

- Tests added/updated: Added 4 tests to `removeKeysFromExpression` covering parenthesised expressions (remove last, first, middle, only filter).
- Manual verification: Confirmed dangling `AND` no longer appears; confirmed duplicate name toast fires.
- Edge cases covered: Single filter inside parens, middle filter removal, first filter removal.

---

- Blast radius: Dashboard variables and panel filter expressions only.
- Potential regressions: None expected — the preceding-conjunction fallback only fires when no following conjunction exists (i.e. the existing path is unchanged for non-terminal pairs).
- Rollback plan: Revert the two commits on this branch.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Fixed dangling `AND` in panel filter expressions after variable deletion; duplicate dashboard variable names now blocked with an error toast. |

---

### 📋 Checklist
- [x] Manually tested


aks07 (approved):

YounixM (approved):

srikanthccv (changes requested):
This is band-aid fixing. The regex based approach is fragile and prone to ton of bugs, the case mentioned here is one of many such cases that has been discovered.
Additionally the `extractQueryPairs` is broken too. Here are the some cases to how they are broken.



Example 1:

Let's say the expression is `a = '1' AND b = '2' OR c = '3'`, this is `(a AND b) OR c` (see the grammar/FilterQuery.g4. For `removeKeysFromExpression(expression, ['b'])` The replaced expression is `a = '1' AND c = '3'`, but it should have been `a = '1' OR c = '3'`. Semantics are broken.

Here are the additional cases where semantics are broken

- `a = '1' AND b = '2' OR c = '3' AND d = '4'` with `removeKeysFromExpression(expression, ['b'])`
  - expected should be `a = '1' OR c = '3' AND d = '4'` but it returns `a = '1' AND c = '3' AND d = '4'`
- `a = '1' OR b = '2' AND c = '3' OR d = '4'` with `removeKeysFromExpression(expression, ['c'])`
  - expected should be `a = '1' OR b = '2' OR d = '4'` but it returns `a = '1' OR b = '2' AND d = '4'`
- Take `(a = '1') OR (b = '2')` for example, removing `a` should return `b = '2'` not `() OR (b = '2')`

Some tests that fails because of the broken implementation

- `a BETWEEN 1 AND 10 AND b = '2'` removing `a` should return `b = '2'`, however the return expo is `10 AND b = '2'`
- `status = '200' OR status = '500' AND region = 'us-east'` with remove `status` should return is `status = '200' AND region = 'us-east'`
- broken exists because only complete pairs are removed
  - `a = '1' AND b EXISTS`, remove `b` => `a = '1' AND b EXISTS` but should be `a = '1'`

These are not exhaustive but you get the gist of how many ways it could fail. We should fix the root cause not the symptoms.

PR #10685, opened 2026-03-23
fix(query-builder): filter with nil values should disable query

### LINKED ISSUE #9392: [Bug] Infrastructure Kubernetes dashboard filters show "No values found" and generate invalid SQL (v0.97.1)
# Bug Description

In a separate custom dashboard that I created myself, the same metrics display correctly.

The Infrastructure -> Kubernetes dashboard filter dropdowns (Pod, Namespace, Node, Cluster, Deployment) show "No values found" and generate invalid SQL queries with null filter values, resulting in 500 errors.

Image

## Environment

- **SigNoz Version**: v0.97.1 (Chart: 0.97.0, App: v0.97.1)
- **Deployment**: Kubernetes (EKS)
- **Data Source**: OpenTelemetry Collector with `kubeletstats` receiver
- **ClickHouse Version**: 25.5.6

## Steps to Reproduce

1. Install SigNoz v0.97.1
2. Configure OpenTelemetry Collector with `kubeletstats` receiver to collect Kubernetes metrics
3. Navigate to Infrastructure -> Kubernetes in SigNoz UI
4. Try to select any filter (Pod, Namespace, Node, etc.)
5. Observe "No values found" in all filter dropdowns
6. Dashboard displays error: `API responded with 500 - Something went wrong status: error in builder queries`

## Expected Behavior

- Filter dropdowns should populate with available values (pod names, namespaces, etc.)
- Selecting a filter should generate valid SQL queries
- Dashboard should display Kubernetes metrics

## Actual Behavior

- All filter dropdowns show "No values found"
- Invalid SQL queries are generated with empty filter values
- 500 error from query-service

### Query Service Error:
```json
{
  "level": "error",
  "timestamp": "2025-10-22T05:19:34.857Z",
  "caller": "utils/format.go:229",
  "msg": "invalid type for formatted value",
  "type": null
}
```

### Invalid SQL Generated:
```sql
SELECT DISTINCT fingerprint 
FROM signoz_metrics.time_series_v4 
WHERE metric_name IN ['k8s.pod.memory_request_utilization'] 
  AND JSONExtractString(labels, 'k8s.namespace.name') =   -- Empty value!
  AND JSONExtractString(labels, 'k8s.pod.name') =          -- Empty value!
```

H4ad:
## Pull Request

---

I was investigating the issue #9392, it was caused because the Pods existed but some of the keys used to filter the values were undefined/null. So the POD existed but the metadata didn't exist.

I can fix this on UI but I was thinking that this fix could be on backend instead, to disable queries that have filters with `nil` values.

Before:
image

It will just render No data now:

image

Fixes #9392

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

The bug was causing the query builder to fail to run, rendering status message 500 when click to visualize pods that does not have specific filter values.

The issue was caused because the item didn't had one of the filter values, eg: `k8s.namespace.name`.

Disable the query if receives `nil` as value, including expressions that depends on it.

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered: Not sure

---

- Blast radius: Query Service
- Potential regressions: Because I disable queries that could fail, we should not see any regression caused by this change.

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Queries using filters that does not have values should not throw errors anymore. |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/pkg/query-service/app/parser.go
+++ b/pkg/query-service/app/parser.go
@@ -16,10 +16,11 @@ import (
 
 	"github.com/SigNoz/signoz/pkg/types/thirdpartyapitypes"
 
+	"log/slog"
+
 	"github.com/SigNoz/govaluate"
 	"github.com/SigNoz/signoz/pkg/query-service/app/integrations/messagingQueues/kafka"
 	queues2 "github.com/SigNoz/signoz/pkg/query-service/app/integrations/messagingQueues/queues"
-	"log/slog"
 
 	"github.com/gorilla/mux"
 	promModel "github.com/prometheus/common/model"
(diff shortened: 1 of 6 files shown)


H4ad:
After talking to @srikanthccv, the fix will be moved to UI.

PR #10686, opened 2026-03-24
fix: enrich unspecified fields in logs pipelines filters

piyushsingariya:
## Pull Request

---

### 📄 Summary

This PR aims to add a step to enrich the tagtype to fix an issue with logs pipelines filter

#### Issues closed by this PR

Closes SigNoz/engineering-pod#4248

---

### ✅ Change Type
_Select all that apply_

- [x] 🐛 Bug fix
- [x] ♻️ Refactor

---

### 🐛 Bug Context

Logs Pipelines uses /v3/filter_suggestion; User has been trying to add a filter for `ddtags` but this is not appearing in suggestions but user forced entered the filter as `ddtags Exists` this result into FE not knowing `ddtags` to be part of attributes and sent ddtags as column -> resulting into faulty IF expression -> resulting into REGEX Parser Failing (from user's pov) 

#### Root Cause

/v3/filter_suggestion is the main cause but we're deciding to not fix it there but here

#### Fix Strategy

Enrich the filter's unspecified keys

---

- Tests added/updated: N/A
- Manual verification: Manually tested ✅ 
- Edge cases covered: N/A

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

N?A
---

### DIFF
--- a/pkg/query-service/app/logparsingpipeline/controller.go
+++ b/pkg/query-service/app/logparsingpipeline/controller.go
@@ -63,7 +63,7 @@ func (pc *LogParsingPipelineController) enrichPipelinesFilters(
 	// Collect names of non-static keys that are missing type info.
 	// Static fields (body, trace_id, etc.) are intentionally Unspecified and map
 	// to top-level OTEL fields — they do not need enrichment.
-	unspecifiedNames := map[string]bool{}
+	unspecifiedNames := map[string]struct{}{}
 	for _, p := range pipelines {
 		if p.Filter != nil {
 			for _, item := range p.Filter.Items {
@@ -76,7 +76,7 @@ func (pc *LogParsingPipelineController) enrichPipelinesFilters(
 					if strings.HasPrefix(item.Key.Key, "body.") {
 						continue
 					}
-					unspecifiedNames[item.Key.Key] = true
+					unspecifiedNames[item.Key.Key] = struct{}{}
 				}
 			}
 		}


pkg/query-service/app/logparsingpipeline/controller.go
```diff
@@ -34,19 +36,101 @@ type LogParsingPipelineController struct {
...
+	for _, p := range pipelines {
+		if p.Filter != nil {
+			for _, item := range p.Filter.Items {
+				if item.Key.Type == v3.AttributeKeyTypeUnspecified {
+					// Skip static fields
+					if _, isStatic := constants.StaticFieldsLogsV3[item.Key.Key]; isStatic {
+						continue
+					}
+					// Skip enrich body.* fields
+					if strings.HasPrefix(item.Key.Key, "body.") {
```
nityanandagohain, line 76: we don't support body. in the pipeline filter as of now

Image

Image

pkg/query-service/app/logparsingpipeline/controller.go
```diff
@@ -34,19 +36,101 @@ type LogParsingPipelineController struct {
...
+// missing type info, by looking them up in the store.
+//
+// TODO(Piyush): remove with qbv5 migration
+func (pc *LogParsingPipelineController) enrichPipelinesFilters(
+	ctx context.Context, pipelines []pipelinetypes.GettablePipeline,
+) ([]pipelinetypes.GettablePipeline, error) {
+	// Collect names of non-static keys that are missing type info.
+	// Static fields (body, trace_id, etc.) are intentionally Unspecified and map
+	// to top-level OTEL fields — they do not need enrichment.
+	unspecifiedNames := map[string]bool{}
```
nityanandagohain: nit: 
```suggestion
	unspecifiedNames := map[string]struct{}{}
```

pkg/query-service/app/logparsingpipeline/controller.go
piyushsingariya: image

I was able to run a different filter here

nityanandagohain: I think it is working for some case. Create a note for yourself so that this can be visited. Since it doesn't affect this PR unblocking it.

nityanandagohain (dismissed):

PR #10689, opened 2026-03-24
fix(querier): return proper HTTP status for PromQL timeout errors

therealpandey:
## Summary
- PromQL queries exceeding the context deadline were returning `400 Bad Request` with `invalid_input` instead of `504 Gateway Timeout`. This happened because `enhancePromQLError` unconditionally wrapped all `NewRangeQuery` errors as `TypeInvalidInput`, including `context.DeadlineExceeded`.
- Extracted `tryEnhancePromQLExecError` to properly classify timeout, cancellation, and storage errors. Used in both the `NewRangeQuery` and `Exec` error paths.
- Made the PromQL engine timeout configurable via `prometheus.timeout` (default `2m`) instead of hardcoding it.

## Test plan
- [x] `go test -race ./pkg/querier/...` passes
- [x] `go test -race ./pkg/prometheus/...` passes
- [ ] Manual: send a slow PromQL query to `/api/v5/query_range` and verify the response is 504 instead of 400


pkg/querier/promql_query.go
```diff
@@ -213,27 +235,20 @@ func (q *promqlQuery) Execute(ctx context.Context) (*qbv5.Result, error) {
 		time.Unix(0, end),
 		q.query.Step.Duration,
 	)
-
 	if err != nil {
+		// NewRangeQuery can fail with execution errors (e.g. context deadline exceeded)
+		// during the query queue/scheduling stage, not just parse errors.
+		if err := tryEnhancePromQLExecError(err); err != nil {
+			return nil, err
+		}
```
srikanthccv, line 243: This is where the actual invalid query for the deadline exceed came from? because below it's either cancelle/timeout/internal.

therealpandey: Yes, it came from here.

therealpandey: I have put a comment explaining why this function needs `tryEnhancePromQLExecError`

PR #10690, opened 2026-03-24
chore: add tests for trace waterfall

nityanandagohain:
## Pull Request

---

### 📄 Summary

* This PR is a pre-requisite to https://github.com/SigNoz/signoz/pull/10399 pr, so that we can keep the functionality/behavior in check
* It also fixes a bug where the selected span id appears twice in uncollapsed span.

image

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [x] 🧪 Test-only

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

No risk
---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/pkg/query-service/app/traces/tracedetail/waterfall_test.go
+++ b/pkg/query-service/app/traces/tracedetail/waterfall_test.go
@@ -120,21 +120,6 @@ func TestGetSelectedSpans_MultipleUncollapsed(t *testing.T) {
 	assert.Equal(t, []string{"root", "childA", "grandchildA", "childB", "grandchildB"}, spanIDs(spans))
 }
 
-// The selected span is always present in the output window.
-func TestGetSelectedSpans_SelectedSpanAlwaysInWindow(t *testing.T) {
-	root := mkSpan("root", "svc",
-		mkSpan("A", "svc",
-			mkSpan("B", "svc",
-				mkSpan("selected", "svc"),
-			),
-		),
-	)
-	spanMap := buildSpanMap(root)
-	spans, _, _, _ := GetSelectedSpans([]string{}, "selected", []*model.Span{root}, spanMap, false)
-
-	assert.Equal(t, []string{"root", "A", "B", "selected"}, spanIDs(spans))
-}
-
 // Collapsing a span with other uncollapsed spans
 //
 // root


pkg/query-service/app/traces/tracedetail/waterfall.go
```diff
@@ -88,7 +88,15 @@ func getPathFromRootToSelectedSpanId(node *model.Span, selectedSpanId string, un
 	return isPresentInSubtreeForTheNode, spansFromRootToNode
 }
```
tushar-signoz, line 89: Hey @nityanandagohain
I know this is existing code. But the name getPathFromRootToSelectedSpanId function suggests that it returns path nodes, however it takes additional args uncollapsedSpans for appending the nodes and isSelectedSpanIDUnCollapsed for adding or skipping selectedSpan.
And since path nodes can already be part of uncollapsedSpans, hence de dup check is required.
Ideally getPathFromRootToSelectedSpanId should return all the path node array and caller function should do the dedup once, same with isSelectedSpanIDUnCollapsed.
Or `GetSelectedSpans` can just have a uncollapsedSpansMap map[string]bool and forget about duplicates ?
Another optimisation that can be done is, adding a break once isPresentInSubtreeForTheNode is found to be True.

pkg/query-service/app/traces/tracedetail/waterfall_test.go
```diff
@@ -0,0 +1,370 @@
...
+			mkSpan("B", "svc",
+				mkSpan("selected", "svc"),
+			),
+		),
+	)
+	spanMap := buildSpanMap(root)
+	spans, _, _, _ := GetSelectedSpans([]string{}, "selected", []*model.Span{root}, spanMap, false)
+
+	assert.Equal(t, []string{"root", "A", "B", "selected"}, spanIDs(spans))
+}
```
tushar-signoz: This should at atleast create more than SPAN_LIMIT_PER_REQUEST_FOR_WATERFALL spans, then only it would be an actual test. Or can be removed all together since TestGetSelectedSpans_WindowCentredOnSelected is there

nityanandagohain: Correct, `TestGetSelectedSpans_WindowCentredOnSelected` does it in a better way. Removing this.

pkg/query-service/app/traces/tracedetail/waterfall.go
nityanandagohain: Great thanks for the suggestion, let me create a small PR addressing this 👍  . Don't want to change the scope in this PR.

nityanandagohain: https://github.com/SigNoz/signoz/pull/10706 , will open it once we merge this .

pkg/query-service/app/traces/tracedetail/waterfall_test.go
nityanandagohain: Removed

tushar-signoz (approved):
LGTM

PR #10691, opened 2026-03-24
fix: guard label.toLowerCase() against non-string values in NewSelect

AshwinBhatkal:
## Pull Request

---

### 📄 Summary
Dashboard variable options can have non-string labels (e.g. numbers), causing `label.toLowerCase()` to throw a `TypeError` when filtering options by search text in `CustomSelect`.

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/4122

---

### ✅ Change Type

- [x] 🐛 Bug fix

---

#### Root Cause
`filterOptionsBySearch` in `NewSelect/utils.ts` calls `.toLowerCase()` directly on `option.label` and `subOption.label`, assuming they are always strings. Dashboard variable values can be non-string types (numbers, etc.), causing the crash.

#### Fix Strategy
Wrap each label access with `String()` before calling `.toLowerCase()`, making the coercion explicit and safe for any label type.

---

### 🧪 Testing Strategy

- Tests added/updated: N/A
- Manual verification: Reproduced by passing numeric labels to CustomSelect — no longer crashes
- Edge cases covered: `null`, `undefined`, numbers, React nodes all handled via `String()` coercion

---

- Blast radius: Only affects `filterOptionsBySearch` in `NewSelect/utils.ts`
- Potential regressions: None — `String(str)` on a string is a no-op
- Rollback plan: Revert the two-line change

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Fixed crash when searching dashboard variable options with non-string labels |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

Two-line change — `String(option.label).toLowerCase()` instead of `option.label.toLowerCase()` at lines 130 and 139 in `utils.ts`.

---

### DIFF
--- a/frontend/src/components/NewSelect/__test__/utils.test.ts
+++ b/frontend/src/components/NewSelect/__test__/utils.test.ts
@@ -1,0 +1,100 @@
+import { filterOptionsBySearch } from '../utils';
+
+describe('filterOptionsBySearch', () => {
+	it('returns all options when searchText is empty', () => {
+		const options = [
+			{ label: 'Option 1', value: 'option1' },
+			{ label: 'Option 2', value: 'option2' },
+		];
+		expect(filterOptionsBySearch(options, '')).toEqual(options);
+		expect(filterOptionsBySearch(options, '   ')).toEqual(options);
+	});
+
+	it('filters top-level options by string label', () => {
+		const options = [
+			{ label: 'Apple', value: 'apple' },
+			{ label: 'Banana', value: 'banana' },
+			{ label: 'Apricot', value: 'apricot' },
+		];
+		const result = filterOptionsBySearch(options, 'ap');
+		expect(result).toEqual([
+			{ label: 'Apple', value: 'apple' },
+			{ label: 'Apricot', value: 'apricot' },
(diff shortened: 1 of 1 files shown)


frontend/src/components/NewSelect/utils.ts
```diff
@@ -127,7 +127,7 @@ export const filterOptionsBySearch = (
 			if ('options' in option && Array.isArray(option.options)) {
 				// Filter nested options
 				const filteredSubOptions = option.options.filter((subOption) =>
-					subOption.label.toLowerCase().includes(lowerSearchText),
+					String(subOption.label).toLowerCase().includes(lowerSearchText),
```
H4ad, line 130: Why subOption.label is not typed as `number`? 

`CustomSelect` and `CustomMultiSelect` also needs to be updated, if you change the type to `number`, those places break due to `toLowerCase`.

AshwinBhatkal: Done

PR #10701, opened 2026-03-24
fix(infra-monitoring-hosts): auto-refresh not waiting for query to finish & loading feedback

H4ad:
## Pull Request

---

This PR fixes two bugs on Hosts List page:
- Auto-Refresh not waiting for previous query: https://github.com/SigNoz/engineering-pod/issues/3994
- No loading feedback on auto-refresh: https://github.com/SigNoz/engineering-pod/issues/3989

Before:

https://github.com/user-attachments/assets/61eff3cd-2cd2-45b0-a915-e882ad8c9b63
> No loading state on auto-refresh

After:

https://github.com/user-attachments/assets/7af1c6dc-e992-4da8-a0b4-7e2a07695574
> Auto-Refresh now shows loading state, and also wait for the previous query to finish.

https://github.com/user-attachments/assets/52d4887b-717d-46a5-9be0-5940cc30feb5
> Each query now stores the cache 

Fixes https://github.com/SigNoz/engineering-pod/issues/3994
Fixes https://github.com/SigNoz/engineering-pod/issues/3989
Fixes https://github.com/SigNoz/engineering-pod/issues/3993

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

The auto-refresh is a feature that is not related to any component, it's basically independent, since the implementation that uses this value does not know about who is using it, there's no way to wait or skip the query. That's why I changed to add support with React Query.

About the loading feedback, the loading feedback happens when the query isLoading or isFetching, but if the query takes more time than the refreshInterval, it will trigger the query again to refetch and causes the isLoading/isFetching to be always on. The only thing that stops the UI of having the loading feedback forever is `formattedHostMetricsData.length === 0`, but this also prevents the loading state to be presented when we actually want to show the feedback.

The bug is caused by the way we store min/max, and how we present the loading feedback.

To fix that, I had to:
- Introduce a new Store to start migration from Redux to Zustand
- Be able to convert/expose the duration
- Customize DateTimeSelectionV2 to support show refresh loading state for specific query keys.
  - Now if you want to show the loading state on refresh button, you just need to append `REACT_QUERY_KEYS.AUTO_REFRESH_QUERY`.

With the new Zustand store, I was able to use it in the new hook for HostsList to add support to avoid recreating the query every time min/max was updated.

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered:

---

- Blast radius: Hosts List Page
- Potential regressions: Any bug related to query time since now it's a little bit unrelated to each other.

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Hosts page of Infrastructure Monitoring now shows loading feedback & handle the auto-refresh correctly when query is slower than the refresh interval. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

I tried to simplify even more this PR but in the end we cause different bugs because min/max as part of react query keys is not a really good behavior.

I hope this PR starts the base fundation to start slowly migration away from Redux & opt-in for the query duration in the key instead of hardcoded values for min/max during refresh.


H4ad:
Closed in favor of splitting this PR.

PR #10703, opened 2026-03-24
fix: cancel not working in metrics summary page

AshwinBhatkal:
When metrics data are being fetched in the summary page, cancel button was not working.
This PR fixes that. It also introduces a cancelled state that the page goes to when query is cancelled

https://github.com/user-attachments/assets/281fddfa-07e7-4268-b19d-45fd45be1b1d


srikanthccv:
It doesn't work elsewhere as well https://github.com/SigNoz/engineering-pod/issues/4128, example alerts. Isn't there a fix that doesn't involve page-by-page fix?

AshwinBhatkal:
@srikanthccv The way it has currently been written. It only handles cancel when the props are passed for it. This should have been made required and the relevant props passed but this is not happening. 

I can see 2 ways it can cancel today but not sure of the context around them fully. Let me know, I can discuss this and take it forward if required

image
image

srikanthccv:
Understood. As you mentioned, we need to enforce the cancel prop everywhere and ensure cancellation actually works wherever `Run Query` is supported. Can you check with the team and get this sorted? We shouldn't have a case where the Cancel button is visible but does nothing on click; that's a poor user experience.

frontend/src/container/MeterExplorer/Explorer/QuerySection.tsx
```diff
@@ -1,43 +0,0 @@
-import { Button } from 'antd';
```
AshwinBhatkal, line 1: uses RightToolbarActions internally

PR #10715, opened 2026-03-25
feat: enabled service account and deprecated API Keys

SagarRajput-7:
## Pull Request

Under this - 
- we have enabled the service account page
- deprecated the API_keys page
- shown the banner for the above change
- added redirection of api key urls
- spec changed from BE hence updated the ui accordingly

- Major change was the roles management, now we dont have a batch or a overall save call, every role need to have its own post or delete call, hence hook to manage that and have relevant error and retry management for that.

---

image

Screenshot 2026-03-31 at 1 42 12 AM

https://github.com/user-attachments/assets/5376af48-6a2c-460d-8a23-5e568af10702

https://github.com/user-attachments/assets/44e943d2-6796-414a-8a83-24e56c2da912

- closes - [(task) deprecate API keys from the frontend](https://github.com/SigNoz/platform-pod/issues/1828)
---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated: added and updated
- Manual verification: manually tested
- Edge cases covered: NA

---

- Blast radius: Service account page
- Potential regressions: NA
- Rollback plan: NA

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/frontend/src/AppRoutes/__tests__/Private.test.tsx
+++ b/frontend/src/AppRoutes/__tests__/Private.test.tsx
@@ -306,11 +306,19 @@ describe('PrivateRoute', () => {
 			);
 		});
 
-		it('should redirect /settings/access-tokens to /settings/api-keys', () => {
+		it('should redirect /settings/access-tokens to /settings/service-accounts', () => {
 			renderPrivateRoute({ initialRoute: '/settings/access-tokens' });
 
 			expect(screen.getByTestId('location-display')).toHaveTextContent(
-				'/settings/api-keys',
+				'/settings/service-accounts',
+			);
+		});
+
+		it('should redirect /settings/api-keys to /settings/service-accounts', () => {
+			renderPrivateRoute({ initialRoute: '/settings/api-keys' });
+
+			expect(screen.getByTestId('location-display')).toHaveTextContent(
+				'/settings/service-accounts',
 			);
 		});
 


frontend/src/hooks/serviceAccount/useServiceAccountRoleManager.ts
```diff
@@ -0,0 +1,121 @@
...
+						},
+					});
+				}
+			});
+
+			return failures;
+		},
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+		[accountId, currentRoles, createRole, deleteRole, invalidateRoles],
+	);
```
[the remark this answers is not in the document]
SagarRajput-7: done

frontend/src/components/ServiceAccountDrawer/ServiceAccountDrawer.tsx
```diff
@@ -133,51 +154,152 @@ function ServiceAccountDrawer({
...
+			}
+
+			const roleFailures: RoleUpdateFailure[] =
+				rolesResult.status === 'fulfilled' ? rolesResult.value : [];
+			for (const failure of roleFailures) {
+				const context = `Role '${failure.roleName}'`;
+				errors.push({
+					context,
+					apiError: toSaveApiError(failure.error),
+					onRetry: makeRoleRetry(context, failure.onRetry),
```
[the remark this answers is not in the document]
SagarRajput-7: done

frontend/src/components/ServiceAccountDrawer/ServiceAccountDrawer.tsx
```diff
@@ -133,51 +154,152 @@ function ServiceAccountDrawer({
...
+		setSaveErrors([]);
+		setIsSaving(true);
+		try {
+			const [nameResult, rolesResult] = await Promise.allSettled([
+				updateMutateAsync({
+					pathParams: { id: account.id },
+					data: { name: localName },
+				}),
+				applyDiff(localRoles, availableRoles),
+			]);
```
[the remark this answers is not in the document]
SagarRajput-7: done

frontend/src/components/ServiceAccountDrawer/SaveErrorItem.tsx
```diff
@@ -0,0 +1,72 @@
...
+							try {
+								await onRetry();
+							} finally {
+								setIsRetrying(false);
+							}
+						}}
+					>
+						<RotateCw size={12} color={Color.BG_CHERRY_400} />
+					</Button>
+				)}
```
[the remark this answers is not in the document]
SagarRajput-7: done

frontend/src/components/ServiceAccountDrawer/ServiceAccountDrawer.tsx
```diff
@@ -133,51 +154,152 @@ function ServiceAccountDrawer({
...
+				setSaveErrors(errors);
+			} else {
+				toast.success('Service account updated successfully', {
+					richColors: true,
+				});
+			}
+
+			refetchAccount();
+			onSuccess({ closeDrawer: false });
+			queryClient.invalidateQueries(getListServiceAccountsQueryKey());
```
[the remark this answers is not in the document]
SagarRajput-7: done

frontend/src/components/ServiceAccountDrawer/OverviewTab.tsx
```diff
@@ -143,6 +155,19 @@ function OverviewTab({
...
+				<div className="sa-drawer__save-errors">
+					{saveErrors.map(({ context, apiError, onRetry }) => (
+						<SaveErrorItem
+							key={context}
+							context={context}
+							apiError={apiError}
+							onRetry={onRetry}
+						/>
+					))}
+				</div>
```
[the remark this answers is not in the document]
SagarRajput-7: done

frontend/src/hooks/serviceAccount/useServiceAccountRoleManager.ts
```diff
@@ -0,0 +1,120 @@
...
+					role,
+					run: (): ReturnType<typeof createRole> =>
+						createRole({ pathParams: { id: accountId }, data: { id: role.id } }),
+				})),
+				...removedRoles.map((role) => ({
+					role,
+					run: (): ReturnType<typeof deleteRole> =>
+						deleteRole({ pathParams: { id: accountId, rid: role.id } }),
+				})),
+			];
```
H4ad, line 80: Eventually we can ask team to move this to backend, to rollback everything if something fails.

frontend/src/hooks/serviceAccount/useServiceAccountRoleManager.ts
```diff
@@ -0,0 +1,120 @@
...
+
+	const { data, isLoading } = useGetServiceAccountRoles({ id: accountId });
+
+	const currentRoles = useMemo<AuthtypesRoleDTO[]>(() => data?.data ?? [], [
+		data?.data,
+	]);
+
+	const { mutateAsync: createRole } = useCreateServiceAccountRole({
+		mutation: {
+			retry: 2,
```
H4ad: This is the kind of thing we don't want to have retry, otherwise it could create duplicated items.

H4ad: Unless this have idempotency

frontend/src/components/ServiceAccountDrawer/RevokeKeyModal.tsx
```diff
@@ -64,9 +64,9 @@ function RevokeKeyModal(): JSX.Element {
 	const open = !!revokeKeyId && !!accountId;
 
 	const cachedKeys = accountId
-		? queryClient.getQueryData<{ data: ServiceaccounttypesFactorAPIKeyDTO[] }>(
-				getListServiceAccountKeysQueryKey({ id: accountId }),
-		  )
+		? queryClient.getQueryData<{
+				data: ServiceaccounttypesGettableFactorAPIKeyDTO[];
+		  }>(getListServiceAccountKeysQueryKey({ id: accountId }))
 		: null;
```
H4ad, line 70: No route from backend yet?

frontend/src/components/ServiceAccountDrawer/ServiceAccountDrawer.tsx
```diff
@@ -93,21 +103,31 @@ function ServiceAccountDrawer({
...
+
 	useEffect(() => {
 		if (account) {
 			setLocalName(account.name ?? '');
-			setLocalRoles(account.roles ?? []);
 			setKeysPage(1);
 		}
+		setSaveErrors([]);
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [account?.id]);
```
H4ad: Not related to this PR but I would suggest to fix and remove this eslint disable

frontend/src/components/ServiceAccountDrawer/ServiceAccountDrawer.tsx
```diff
@@ -133,51 +153,192 @@ function ServiceAccountDrawer({
...
+			return;
+		}
+		try {
+			await updateMutateAsync({
+				pathParams: { id: account.id },
+				data: { name: localName },
+			});
+			setSaveErrors((prev) => prev.filter((e) => e.context !== 'Name update'));
+			refetchAccount();
+			queryClient.invalidateQueries(getListServiceAccountsQueryKey());
```
H4ad, line 176: You can move at this to onSuccess of the mutation.

frontend/src/hooks/serviceAccount/useServiceAccountRoleManager.ts
SagarRajput-7: intention behind this is add retires in case of any issues at the BE which can be solved via retry, since in this new management way, multiple operation happens at once.

Also I found this at our queryClient as the app level - 

image


@H4ad should i have this here instead of retry: 2 ?

frontend/src/components/ServiceAccountDrawer/RevokeKeyModal.tsx
SagarRajput-7: No new addition here.

frontend/src/components/ServiceAccountDrawer/ServiceAccountDrawer.tsx
SagarRajput-7: this is intentional, since the updateMutateAsync is used at 2 places and adding `onsuccess` would trigger this at both cases, but the other case doesn't only depend on the name update, it depends on the roles update success also, we want all of the promises to fulfill then decide.

frontend/src/hooks/serviceAccount/useServiceAccountRoleManager.ts
H4ad: Maybe we should not add retry at this component and just have the default option for it, it will try 5xx errors and skip 4xx

SagarRajput-7: no thats what i am saying we currently dont do this for mutations, we only do this for `queries`  right now in our app, please check - https://github.com/SigNoz/signoz/blob/a4c63945420d5605b987181ccdc73b932c5cbc33/frontend/src/index.tsx#L19

frontend/src/hooks/serviceAccount/useServiceAccountRoleManager.ts
vikrantgupta25: it's not very straightforward to have bulk operations for these as of today because of certain underlying restrictions from our authz server. Eventually we would ? Maybe yes if we make the upstream contribution

vikrantgupta25: until we have that singular operations are the way to go and it's not very uncommon to have these orchestrations on the client ?

frontend/src/hooks/serviceAccount/useServiceAccountRoleManager.ts
vikrantgupta25: the requests are idempotent here

frontend/src/hooks/serviceAccount/useServiceAccountRoleManager.ts
H4ad: Authz not supporting does not mean we can't move this to backend, at least on backend, when authz eventually supports it, we don't need to update the UI.

I don't mind keeping in this way now, but I don't think this kind of logic should be on Client, ideally, just send roles to delete/create and the backend handles all of them.

If fails one should fail all, or just the one that fail, should be something the API should decide.

frontend/src/hooks/serviceAccount/useServiceAccountRoleManager.ts
H4ad: I thought it was for mutation as well, in this case adding global to mutations will have side effects we don't know, just add a comment saying the retry for that mutation is safe due to being idempotent on backend, this will cover future reviews.

frontend/src/hooks/serviceAccount/useServiceAccountRoleManager.ts
vikrantgupta25: the bulk endpoints in general are not very clean / straight forward to get out. there are certain shenanigans if taken at backend leaves the user unaware of what happened where.

frontend/src/hooks/serviceAccount/useServiceAccountRoleManager.ts
SagarRajput-7: Removed the retries here, and depending on UI retry options only. Also added comments around idempotent majorly since we are still allowing retry via user action

frontend/src/components/ServiceAccountDrawer/ServiceAccountDrawer.tsx
SagarRajput-7: done, refactored.

PR #10717, opened 2026-03-25
build(vite): migrate from rolldown-vite bridge (temporary pkg) to vite 8 stable

amareshsm:
## Pull Request

---

The bridge package ([rolldown-vite@7.3.1](https://www.npmjs.com/package/rolldown-vite)) is replaced with the official Vite 8 release. rolldown-vite was always a temporary  — it is now abandoned. Vite 8 is the stable, supported version.
image

The bridge package has been abandoned. It's recommended to migrate to the stable release. Vite 8, now officially released, offers full Rollup integration and provides a stable, modern setup.

Changes done part of vite 8 migration:

package.json
- vite: npm:rolldown-vite@7.3.1 → ^8.0.2
- @vitejs/plugin-react: 5.1.4 → ^6.0.1 (Babel removed, Oxc-based)
- @sentry/vite-plugin: 2.22.6 → ^5.1.1 (Vite 8 plugin API compat)
- vite-tsconfig-paths: removed (native resolve.tsconfigPaths: true - vite 8 natively supports this)

vite.config.ts
- remove manual alias map, add tsconfigPaths: true,
- add moduleType: 'js' to rawMarkdownPlugin, 
- the CSS minifier switched from `esbuild` to `lightningcss` 

frontend/src/container/LogsSearchFilter/SearchFields/utils.ts: 
- remove invalid `?` + default value coexistence (Oxc strict)

frontend/src/container/NewWidget/LeftContainer/ExplorerColumnsRenderer.styles.scss: 
- remove invalid var(rgba(...)) syntax (LightningCSS strict)

*.test.snap 
-  @ant-design/cssinjs was bumped from 1.17.2 → 1.24.0. This is a transitive dependency — antd@5.11.0 itself didn't change, but when yarn install ran after updating vite/@vitejs/plugin-react, Yarn resolved a newer compatible version of @ant-design/cssinjs (since antd specifies it as ^1.17.2, any 1.x is allowed).

           
           
           
     

This PR introduces no behavioral changes. It simply replaces a deprecated package with the stable version, with no impact on the user experience.

 
N/A

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

**N/A** — This is a tooling upgrade, not a bug fix.

The two code fixes (`utils.ts`, `.scss`) are pre-existing bugs that were silently tolerated by the old bundler and surfaced by Vite 8's stricter toolchain. They are fixed as part of this PR since they are required for the migration to compile.

1.   utils.ts (`LogsSearchFilter/SearchFields/utils.ts`) 
In TypeScript, ? means “this argument is optional,” while = null gives a default value if it’s missing. Using both together is contradictory, so Vite 8’s new TypeScript transformer (Oxc) now rejects it, unlike the old bundler which ignored it. Simply: just use = null to make an argument optional with a default.

2. `ExplorerColumnsRenderer.styles.scss` — SCSS syntax fix

```
/* BEFORE — invalid CSS */
color: var(rgba(255, 255, 255, 0.85));

/* AFTER — correct CSS */
color: rgba(255, 255, 255, 0.85);
```

var() is only for CSS custom properties like var(--my-color). Wrapping rgba() inside var() is invalid. The old esbuild CSS pipeline passed it through silently. Vite 8 uses LightningCSS which is fully spec-compliant and throws on it.

 N/A

---

- Tests added/updated:
      - 11 snapshot files updated to reflect the new `@ant-design/cssinjs@1.24.0` CSS hash (`2i2tap` → `1l6e01l`). No component logic changed — the hash is derived from the cssinjs library version itself.
      
- Manual verification:
      
      - `yarn install` — clean install, no peer conflicts
      - `yarn dev` — dev server starts in ~437ms, page loads correctly, all path aliases resolve. Manually verified the routes to ensure they work correctly.
      - `yarn build` — production build completes in ~10.4s avg (3 clean runs), zero errors
      - `yarn test` — **3888 tests pass**, 2 pre-existing failures (unrelated to migration), 23 skipped
      
- Edge cases covered:

    - Path alias resolution: 15+ import prefixes (`types`, `constants`, `container`, `hooks`, `api`, `components`, etc.) verified working via native `tsconfigPaths`
    - Env variable injection: all 13 `process.env.*` defines verified with 0 unresolved refs in build
    - Markdown imports: `rawMarkdownPlugin` with `moduleType: 'js'` verified
    - CSS pipeline: LESS (`javascriptEnabled`), SCSS, CSS Modules all working

---

- Blast radius: Build tooling only. Zero changes to application logic, routing, or API layer.
- Potential regressions:
        - HMR behavior: Rolldown's dev bundling is functionally equivalent but may have subtle differences in edge cases. Monitor for any "module not found" errors in dev.
        - CSS specificity: LightningCSS is more spec-compliant than esbuild's CSS pipeline. Any other invalid CSS syntax would now be caught. No issues found in this codebase as of now.
        - Chunk splitting: Rolldown produces 416 JS chunks vs 292 before (+42%). This is expected — Rolldown's code splitting is more granular. Total JS size is actually slightly smaller (−1.11%).

- Rollback plan:
     - revert the pr incase of any issues
---

| Field | Value |
|------|-------|
| Change Type |  Maintenance |
| Description | N/A — internal build tooling change, no user-facing behavior change |

---

### 📋 Checklist
- [x] Manually tested

---

## Build Metrics: Before vs After

> Measured on macOS (Apple Silicon), Node v22.20.0, Yarn 1.22.22. Build times are averages of clean runs (`rm -rf build`) without Sentry plugin. One outlier run (19.56s — thermal throttle) excluded from average.

### Bundle Size Summary

| Metric | Before (rolldown-vite 7.3.1) | After (Vite 8.0.2) | Delta | Change |
|--------|------|-------|-------|--------|
| **JS (raw, minified)** | 14,309,106 bytes (13.65 MB) | 14,150,327 bytes (13.50 MB) | −158,779 bytes | **−1.11%** |
| **CSS (raw, minified)** | 1,409,667 bytes (1.34 MB) | 1,405,581 bytes (1.34 MB) | −4,086 bytes | **−0.29%** |
| **Total build dir** | 67 MB | 67 MB | — | **≈0%** | |

### Build Performance

| Metric | Before | After | Delta |
|--------|--------|-------|-------|
| **Build time (avg)** | ~12.4s (11.80 / 12.18 / 13.14) | ~10.4s (10.36 / 10.43 / 9.56) | **−2.0s (−16%)** |

> **Note:** Removing `vite-tsconfig-paths` and using native `resolve.tsconfigPaths: true` reduced build time from ~15.9s to ~10.4s — faster than the pre-migration baseline. The plugin added ~5s overhead by running its own resolver in JS; the native resolver runs in Rust via Rolldown.

### Chunk Analysis

| Metric | Before | After | Delta |
|--------|--------|-------|-------|
| **JS chunk count** | 292 | 416 | **+124 (+42%)** |
| **CSS chunk count** | 85 | 88 | +3 |
| **Gzip file count** | 256 | 331 | +75 |
| **Sourcemap count** | 287 | 410 | +123 |

### Main Entry Bundle

| Bundle | Before | After | Delta |
|--------|--------|-------|-------|
| **Main JS (`index-*.js`)** | 2,628 KB | 1,206 KB | **−1,422 KB (−54%)** |
| **Main CSS (`index-*.css`)** | 144 KB | 137 KB | −7 KB (−5%) |
| **index.html** | 4.0 KB | 16 KB | +12 KB (\*) |

> \* The `index.html` size increase is due to Vite 8 injecting more `<link rel="modulepreload">` tags for better chunk preloading, which improves initial page load performance.

### Top 10 Largest JS Chunks (After) - 

| Chunk | Size |
|-------|------|
| `index.esm-BL9l3aW2.js` (antd) | 1,449 KB |
| `OnboardingPage-DDnfgiVM.js` | 1,356 KB |
| `index-g-W9iCWY.js` (main entry) | 1,206 KB |
| `utils-CFSb6i69.js` (shared utils) | 1,090 KB |
| `utils-CYpTnGQ4.js` (shared utils #2) | 560 KB |
| `IntegrationsModulePage-DHYsgaQa.js` | 532 KB |
| `Settings-CGyXqrYM.js` | 441 KB |
| `InfrastructureMonitoring-Dvq1aVT5.js` | 422 KB |
| `MarkdownRenderer-CO16rM2L.js` | 262 KB |
| `jsx-dev-runtime-B0Zs7J26.js` | 240 KB |

### Key Observations:

- **JS is slightly smaller** (−1.11%) — Oxc minification produces marginally smaller output than esbuild at this scale.
- **CSS is slightly samller** (−0.29%) — Lightning CSS minifies more aggressively than esbuild.
- **Better code splitting** — Vite 8 splits 292 → 416 chunks. The main entry bundle dropped from 2,628 KB → 1,206 KB (−54%), which means faster initial page load since the browser downloads less for the first paint.
- **Build time is faster** (12.4s → ~10.4s, −16%) — after removing vite-tsconfig-paths plugin, the native Rust-based path resolver in Rolldown is significantly faster. Initial Vite 8 builds with the plugin were ~15.9s (+28%), but removing it brought the total below the original baseline.
- **index.html is quite large** — Vite 8 injects modulepreload hints for critical chunks, which is a net positive for load performance.

---

### Future Cleanup Opportunities:

- Update `vite-plugin-compression` — Last published 4+ years ago. Consider alternatives like `vite-plugin-compression2`.
- Remove unused @constants/* tsconfig path — tsconfig.json has "@constants/*": ["./container/OnboardingContainer/constants/*"] but 0 imports use this alias in the codebase.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -46,7 +46,7 @@
     "@radix-ui/react-tabs": "1.0.4",
     "@radix-ui/react-tooltip": "1.0.7",
     "@sentry/react": "8.41.0",
-    "@sentry/vite-plugin": "2.22.6",
+    "@sentry/vite-plugin": "^5.1.1",
     "@signozhq/badge": "0.0.2",
     "@signozhq/button": "0.0.2",
     "@signozhq/calendar": "0.0.0",
@@ -78,7 +78,7 @@
     "@visx/hierarchy": "3.12.0",
     "@visx/shape": "3.5.0",
     "@visx/tooltip": "3.3.0",
-    "@vitejs/plugin-react": "5.1.4",
+    "@vitejs/plugin-react": "^6.0.1",
     "@xstate/react": "^3.0.0",
     "ansi-to-html": "0.7.2",
     "antd": "5.11.0",
@@ -160,7 +160,7 @@
     "typescript": "5.9.3",
     "uplot": "1.6.31",
     "uuid": "^8.3.2",
-    "vite": "npm:rolldown-vite@7.3.1",
+    "vite": "^8.0.2",
     "vite-plugin-html": "3.2.2",
     "web-vitals": "^0.2.4",
     "xstate": "^4.31.0",
@@ -257,8 +257,7 @@
     "typescript-plugin-css-modules": "5.2.0",
     "vite-plugin-checker": "0.12.0",
     "vite-plugin-compression": "0.5.1",
-    "vite-plugin-image-optimizer": "2.0.3",
-    "vite-tsconfig-paths": "6.1.1"
+    "vite-plugin-image-optimizer": "2.0.3"
   },
   "lint-staged": {
     "*.(js|jsx|ts|tsx)": [
@@ -283,7 +282,6 @@
     "form-data": "4.0.4",
     "brace-expansion": "^2.0.2",
     "on-headers": "^1.1.0",
-    "tmp": "0.2.4",
-    "vite": "npm:rolldown-vite@7.3.1"
+    "tmp": "0.2.4"
   }
 }
--- a/frontend/src/components/DraggableTableRow/tests/__snapshots__/DraggableTableRow.test.tsx.snap
+++ b/frontend/src/components/DraggableTableRow/tests/__snapshots__/DraggableTableRow.test.tsx.snap
@@ -3,10 +3,10 @@
 exports[`DraggableTableRow Snapshot test should render DraggableTableRow 1`] = `
 <DocumentFragment>
   <div
-    class="ant-table-wrapper css-dev-only-do-not-override-2i2tap"
+    class="ant-table-wrapper css-dev-only-do-not-override-1l6e01l"
   >
     <div
-      class="ant-spin-nested-loading css-dev-only-do-not-override-2i2tap"
+      class="ant-spin-nested-loading css-dev-only-do-not-override-1l6e01l"
     >
       <div
         class="ant-spin-container"
@@ -43,7 +43,7 @@ exports[`DraggableTableRow Snapshot test should render DraggableTableRow 1`] = `
                       class="ant-table-cell"
                     >
                       <div
-                        class="css-dev-only-do-not-override-2i2tap ant-empty ant-empty-normal"
+                        class="css-dev-only-do-not-override-1l6e01l ant-empty ant-empty-normal"
                       >
                         <div
                           class="ant-empty-image"
--- a/frontend/src/components/MessageTip/__snapshots__/MessageTip.test.tsx.snap
+++ b/frontend/src/components/MessageTip/__snapshots__/MessageTip.test.tsx.snap
@@ -9,7 +9,7 @@ exports[`MessageTip custom action 1`] = `
 }
 
 <div
-  class="ant-alert ant-alert-info ant-alert-with-description c0 css-dev-only-do-not-override-2i2tap"
+  class="ant-alert ant-alert-info ant-alert-with-description c0 css-dev-only-do-not-override-1l6e01l"
   data-show="true"
   role="alert"
 >
--- a/frontend/src/container/LogsSearchFilter/SearchFields/utils.ts
+++ b/frontend/src/container/LogsSearchFilter/SearchFields/utils.ts
@@ -53,9 +53,9 @@ export const queryKOVPair = (): QueryFields[] => [
 ];
 
 export const initQueryKOVPair = (
-	name?: string = null,
-	op?: string = null,
-	value?: string | string[] = null,
+	name: string = null,
+	op: string = null,
+	value: string | string[] = null,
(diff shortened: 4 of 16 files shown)


H4ad:
The only issue I noticed with this Vite version, because it generates more chunks, the initial loading is a little bit slower due to amount of requests it needs to do, instead of X is 2X or 3X.

H4ad:
@amareshsm Would you mind taking a look at the CI failling?

Thanks for the PR, I will try review it tomorrow!

amareshsm:
> @amareshsm Would you mind taking a look at the CI failling?
> 
> Thanks for the PR, I will try review it tomorrow!

sure thanks. I will take a look.

PR #10718, opened 2026-03-25
refactor: cloud integration modules implementation

swapnil-signoz:
### Summary

- Adding cloud integration module implementation for AWS
- Azure will be added after initial set of reviews.


ee/modules/cloudintegration/implcloudintegration/module.go
```diff
@@ -0,0 +1,262 @@
...
+		return cloudProviderModule, nil
+	}
+
+	return nil, errors.New(errors.TypeUnsupported, cloudintegrationtypes.ErrCodeUnsupported, "cloud provider is not supported")
+}
+
+func (module *module) getOrCreateIngestionKey(ctx context.Context, orgID valuer.UUID, provider cloudintegrationtypes.CloudProviderType) (string, error) {
+	keyName := cloudintegrationtypes.NewIngestionKeyName(provider)
+
+	result, err := module.gateway.SearchIngestionKeysByName(ctx, orgID, keyName, 1, 10)
```
nityanandagohain, line 450: For one provider how many different ingestion keys can be there for a tenant ?

ee/modules/cloudintegration/implcloudintegration/module.go
```diff
@@ -0,0 +1,262 @@
...
+	return created.Value, nil
+}
+
+func (module *module) getOrCreateAPIKey(ctx context.Context, orgID valuer.UUID, provider cloudintegrationtypes.CloudProviderType) (string, error) {
+	integrationUser, err := module.getOrCreateIntegrationUser(ctx, orgID, provider)
+	if err != nil {
+		return "", err
+	}
+
+	existingKeys, err := module.userSetter.ListAPIKeys(ctx, orgID)
```
nityanandagohain: should we add getApiKey by user in the user module ? @vikrantgupta25

pkg/modules/cloudintegration/implcloudintegration/module.go
```diff
@@ -0,0 +1,73 @@
...
+func (m *module) GetDashboardByID(ctx context.Context, orgID valuer.UUID, id string) (*dashboardtypes.Dashboard, error) {
+	return nil, errors.New(errors.TypeUnsupported, cloudintegrationtypes.ErrCodeUnsupported, "get dashboard by ID is not supported")
+}
+
+func (m *module) ListDashboards(ctx context.Context, orgID valuer.UUID) ([]*dashboardtypes.Dashboard, error) {
+	return nil, errors.New(errors.TypeUnsupported, cloudintegrationtypes.ErrCodeUnsupported, "list dashboards is not supported")
+}
+
+func (m *module) GetCloudProvider(provider cloudintegrationtypes.CloudProviderType) (cloudintegration.CloudProviderModule, error) {
+	panic("unimplemented")
```
nityanandagohain: error

pkg/signoz/module.go
```diff
@@ -117,3 +119,12 @@ func NewModules(
 		ServiceAccount:  implserviceaccount.NewModule(implserviceaccount.NewStore(sqlstore), authz, emailing, providerSettings),
 	}
 }
+
+// SetCloudIntegrationModule sets cloud integration module in Modules
+// TODO: find a better way to set the module,
+// why this was done: cloud integration depends on few modules like userSetter which are initialized in NewModules and other deps like zeus/gateway is not present
+// cloud integration is initialized via callback depending on flavor of Signoz ee/community
+func (modules Modules) SetCloudIntegrationModule(module cloudintegration.Module) Modules {
```
nityanandagohain: This is discussed with @vikrantgupta25 ?

nityanandagohain (commented):
some comments, looks good apart from that

ee/modules/cloudintegration/implcloudintegration/module.go
swapnil-signoz: If I understand the question correctly,
there will be `one` ingestion key per cloud provider. It will be created for first cloud integration for a given cloud provider and then subsequent integrations will use same key.

pkg/signoz/module.go
swapnil-signoz: Let me sync

ee/modules/cloudintegration/implcloudintegration/module.go
nityanandagohain: My bad, updated the comment, yeah that is what I wanted to ask.

So it will be  `result, err := module.gateway.SearchIngestionKeysByName(ctx, orgID, keyName, 1, 10)`?

and then just select the first result if exists.

swapnil-signoz: Yes thats correct, updated the code as well to only respect the first entry

pkg/modules/cloudintegration/cloudintegration.go
```diff
@@ -50,6 +53,39 @@ type Module interface {
 	// ListDashboards returns list of dashboards across all connected cloud integration accounts
 	// for enabled services in the org. This list gets added to dashboard list page
 	ListDashboards(ctx context.Context, orgID valuer.UUID) ([]*dashboardtypes.Dashboard, error)
+
+	// GetCloudProvider returns cloud provider specific module
+	GetCloudProvider(provider citypes.CloudProviderType) (CloudProviderModule, error)
+}
+
+type CloudProviderModule interface {
```
nityanandagohain, line 66: This can be just cloudProvider @vikrantgupta25  ?

nityanandagohain (dismissed):
looks good to me, one question for vikrant

ee/modules/cloudintegration/implcloudintegration/module.go
```diff
@@ -0,0 +1,533 @@
...
+	"github.com/SigNoz/signoz/pkg/valuer"
+	"github.com/SigNoz/signoz/pkg/zeus"
+)
+
+type module struct {
+	store             cloudintegrationtypes.Store
+	gateway           gateway.Gateway
+	zeus              zeus.Zeus
+	licensing         licensing.Licensing
+	globalConfig      global.Config
```
vikrantgupta25: do not inject the global config here, inject the global package and use the module methods from the same

ee/modules/cloudintegration/implcloudintegration/module.go
```diff
@@ -0,0 +1,533 @@
...
+	if err != nil {
+		return err
+	}
+
+	return module.store.CreateAccount(ctx, storableCloudIntegration)
+}
+
+func (module *module) GetConnectionArtifact(ctx context.Context, account *cloudintegrationtypes.Account, req *cloudintegrationtypes.ConnectionArtifactRequest) (*cloudintegrationtypes.ConnectionArtifact, error) {
+	// TODO: evaluate if this check is really required and remove if the deployment promises to always have this configured.
+	if module.globalConfig.IngestionURL == nil {
```
vikrantgupta25: and you shouldn't validate this, In case the ingestion URL is not set then request it from the user on the UI modal

ee/modules/cloudintegration/implcloudintegration/module.go
```diff
@@ -0,0 +1,533 @@
...
+	}
+
+	// get license to get the deployment details
+	license, err := module.licensing.GetActive(ctx, account.OrgID)
+	if err != nil {
+		return nil, err
+	}
+
+	// get deployment details from zeus
+	respBytes, err := module.zeus.GetDeployment(ctx, license.Key)
```
vikrantgupta25: this will fail cloud integrations for all self hosted deployments. Please check how this is handled before. If you aren't able to find something then request it from the user. Give them the best information you can deduce, and request the remaining.

ee/modules/cloudintegration/implcloudintegration/module.go
```diff
@@ -0,0 +1,533 @@
...
+	gateway gateway.Gateway,
+	licensing licensing.Licensing,
+	serviceAccount serviceaccount.Module,
+) (cloudintegration.Module, error) {
+	defStore := pkgimpl.NewServiceDefinitionStore()
+	awsCloudProviderModule, err := implcloudprovider.NewAWSCloudProvider(defStore)
+	if err != nil {
+		return nil, err
+	}
+	azureCloudProviderModule := implcloudprovider.NewAzureCloudProvider()
```
vikrantgupta25: do not instantiate here, Accept if from the root layer, In case you decide to do AWS in both community and enterprise and azure just in enterprise, You can control these better at the signoz layer not at the module DI layer.

ee/modules/cloudintegration/implcloudintegration/module.go
```diff
@@ -0,0 +1,533 @@
...
+func (module *module) ListDashboards(ctx context.Context, orgID valuer.UUID) ([]*dashboardtypes.Dashboard, error) {
+	_, err := module.licensing.GetActive(ctx, orgID)
+	if err != nil {
+		return nil, errors.New(errors.TypeLicenseUnavailable, errors.CodeLicenseUnavailable, "a valid license is not available").WithAdditional("this feature requires a valid license").WithAdditional(err.Error())
+	}
+
+	return module.listDashboards(ctx, orgID)
+}
+
+func (module *module) GetCloudProvider(provider cloudintegrationtypes.CloudProviderType) (cloudintegration.CloudProviderModule, error) {
```
vikrantgupta25: this shouldn't be exported function

vikrantgupta25 (changes requested):

ee/modules/cloudintegration/implcloudintegration/module.go
```diff
@@ -0,0 +1,533 @@
...
+		if d.ID == id {
+			return d, nil
+		}
+	}
+
+	return nil, errors.New(errors.TypeNotFound, cloudintegrationtypes.ErrCodeCloudIntegrationNotFound, "cloud integration dashboard not found")
+}
+
+// TODO: use the function in dashboard APIs during removal of older cloud integration code.
+func (module *module) ListDashboards(ctx context.Context, orgID valuer.UUID) ([]*dashboardtypes.Dashboard, error) {
```
vikrantgupta25, line 414: please make sure that if you are returning licensing error here then the dashboard API should ignore it. The dashboards list call shouldn't fail because of this.

ee/modules/cloudintegration/implcloudintegration/module.go
```diff
@@ -0,0 +1,533 @@
...
+		return err
+	}
+
+	storableService := cloudintegrationtypes.NewStorableCloudIntegrationService(integrationService, configJSON)
+
+	return module.store.UpdateService(ctx, storableService)
+}
+
+// TODO: use the function in dashboard APIs during removal of older cloud integration code.
+func (module *module) listDashboards(ctx context.Context, orgID valuer.UUID) ([]*dashboardtypes.Dashboard, error) {
```
vikrantgupta25: unexported functions always at the bottom of the file.

ee/modules/cloudintegration/implcloudintegration/module.go
swapnil-signoz: This was the reason I think the older version had `get connection params` as separate API, which is not there in this version and consumed by CreateAccount flow itself. This was done assuming the server will always have this information.

vikrantgupta25 (changes requested):

ee/modules/cloudintegration/implcloudintegration/implcloudprovider/awscloudprovider.go
```diff
@@ -0,0 +1,184 @@
...
+
+func NewAWSCloudProvider(defStore cloudintegrationtypes.ServiceDefinitionStore) (cloudintegration.CloudProviderModule, error) {
+	return &awscloudprovider{serviceDefinitions: defStore}, nil
+}
+
+func (provider *awscloudprovider) GetConnectionArtifact(ctx context.Context, creds *cloudintegrationtypes.SignozCredentials, account *cloudintegrationtypes.Account, req *cloudintegrationtypes.ConnectionArtifactRequest) (*cloudintegrationtypes.ConnectionArtifact, error) {
+	// TODO: get this from config
+	agentVersion := "v0.0.8"
+
+	baseURL := fmt.Sprintf("https://%s.console.aws.amazon.com/cloudformation/home", req.Aws.DeploymentRegion)
```
vikrantgupta25: why not from types ?

ee/modules/cloudintegration/implcloudintegration/implcloudprovider/awscloudprovider.go
```diff
@@ -0,0 +1,184 @@
...
+
+	q := u.Query()
+	q.Set("region", req.Aws.DeploymentRegion)
+	u.Fragment = "/stacks/quickcreate"
+
+	u.RawQuery = q.Encode()
+
+	q = u.Query()
+	q.Set("stackName", "signoz-integration")
+	q.Set("templateURL", fmt.Sprintf("https://signoz-integrations.s3.us-east-1.amazonaws.com/aws-quickcreate-template-%s.json", agentVersion))
```
vikrantgupta25:
```
types.NewAWSConnectionArtifact() 
```

ee/modules/cloudintegration/implcloudintegration/implcloudprovider/awscloudprovider.go
```diff
@@ -0,0 +1,184 @@
...
+
+	var awsCfg cloudintegrationtypes.AWSServiceConfig
+	if err := json.Unmarshal([]byte(config), &awsCfg); err != nil {
+		return nil, err
+	}
+
+	return &cloudintegrationtypes.ServiceConfig{AWS: &awsCfg}, nil
+}
+
+func (provider *awscloudprovider) IsServiceEnabled(ctx context.Context, config *cloudintegrationtypes.ServiceConfig) bool {
```
vikrantgupta25: why not again a types function ?

vikrantgupta25 (changes requested):

ee/modules/cloudintegration/implcloudintegration/implcloudprovider/awscloudprovider.go
swapnil-signoz: Idea was to have Cloud Provider specific operations to come from CloudProviderModule.
And same for other comments.

vikrantgupta25: Think about where it belongs ? Domain or Usecase ?

swapnil-signoz: I think we are mixing types with domain and small implementation details around it. I understand the NewAWSConectionArtifact CAN come from types, but domain need not know we are using cloudformation stack URL, how to create it, I feel that is an implementation detail and should belong to modules in this case.

Same goes for Azure when it will come, azure requires cli commands, and not the URL, domain types need not know that, it can know the possible structure Artifact can have.

Same goes for If the service is enabled or not, domain can know what service looks like, but its more closer to the usecase/module what enabled service exactly means.

I am sure there is no right or wrong approach here, but this feels more intuitive to me. I am keen to hear your thoughts on this.

nityanandagohain: Yeah unless it's actually getting reused somewhere do we really need this as this is more of a usecase, so unless we have to build this again and again somewhere not able to understand how it helps.

ee/modules/cloudintegration/implcloudintegration/implcloudprovider/awscloudprovider.go
```diff
@@ -0,0 +1,184 @@
...
+	return provider.serviceDefinitions.List(ctx, cloudintegrationtypes.CloudProviderTypeAWS)
+}
+
+func (provider *awscloudprovider) GetServiceDefinition(ctx context.Context, serviceID cloudintegrationtypes.ServiceID) (*cloudintegrationtypes.ServiceDefinition, error) {
+	return provider.serviceDefinitions.Get(ctx, cloudintegrationtypes.CloudProviderTypeAWS, serviceID)
+}
+
+func (provider *awscloudprovider) StorableConfigFromServiceConfig(ctx context.Context, cfg *cloudintegrationtypes.ServiceConfig, supported cloudintegrationtypes.SupportedSignals) (string, error) {
+	if cfg == nil || cfg.AWS == nil {
+		return "", nil
```
nityanandagohain: we should return an error with empty config ?

ee/modules/cloudintegration/implcloudintegration/implcloudprovider/awscloudprovider.go
```diff
@@ -0,0 +1,184 @@
...
+	if config == nil || config.AWS == nil {
+		return false
+	}
+	logsEnabled := config.AWS.Logs != nil && config.AWS.Logs.Enabled
+	metricsEnabled := config.AWS.Metrics != nil && config.AWS.Metrics.Enabled
+	return logsEnabled || metricsEnabled
+}
+
+func (provider *awscloudprovider) IsMetricsEnabled(ctx context.Context, config *cloudintegrationtypes.ServiceConfig) bool {
+	if config == nil || config.AWS == nil {
```
nityanandagohain: as vikrant suggested above, I think this can also go into types  and the function can be generic for all config.

ee/modules/cloudintegration/implcloudintegration/implcloudprovider/awscloudprovider.go
```diff
@@ -0,0 +1,184 @@
...
+		return services[i].Type.StringValue() < services[j].Type.StringValue()
+	})
+
+	compiledMetrics := &cloudintegrationtypes.AWSMetricsStrategy{}
+	compiledLogs := &cloudintegrationtypes.AWSLogsStrategy{}
+	var compiledS3Buckets map[string][]string
+
+	for _, storedSvc := range services {
+		svcCfg, err := provider.ServiceConfigFromStorableServiceConfig(ctx, storedSvc.Config)
+		if err != nil || svcCfg == nil || svcCfg.AWS == nil {
```
nityanandagohain: when is svcCfg and svcCfg.AWS nil ? should we prevent it from happening and not leave this cases ?

ee/modules/cloudintegration/implcloudintegration/implcloudprovider/awscloudprovider.go
```diff
@@ -0,0 +1,184 @@
...
+	return &cloudintegrationtypes.ProviderIntegrationConfig{
+		AWS: &cloudintegrationtypes.AWSIntegrationConfig{
+			EnabledRegions: enabledRegions,
+			Telemetry:      awsTelemetry,
+		},
+	}, nil
+}
+
+// awsLogsEnabled returns true if the AWS service config has logs explicitly enabled.
+func awsLogsEnabled(cfg *cloudintegrationtypes.AWSServiceConfig) bool {
```
nityanandagohain: cango to types as well. cfg.LogsEnabled() ?

ee/modules/cloudintegration/implcloudintegration/implcloudprovider/awscloudprovider.go
```diff
@@ -0,0 +1,184 @@
...
+	if len(compiledLogs.Subscriptions) > 0 {
+		awsTelemetry.Logs = compiledLogs
+	}
+	if compiledS3Buckets != nil {
+		awsTelemetry.S3Buckets = compiledS3Buckets
+	}
+
+	enabledRegions := []string{}
+	if account.Config != nil && account.Config.AWS != nil && account.Config.AWS.Regions != nil {
+		enabledRegions = account.Config.AWS.Regions
```
nityanandagohain: question: What happens with if the regions are empty?

ee/modules/cloudintegration/implcloudintegration/module.go
nityanandagohain: In this case if the licesense is not present, should this just return empty list then?

nityanandagohain (commented):
It's was easier to go through this now, thanks for moving out the service definations

ee/modules/cloudintegration/implcloudintegration/implcloudprovider/awscloudprovider.go
nityanandagohain: I see that the regions cannot be empty and you thow a validation error

ee/modules/cloudintegration/implcloudintegration/module.go
swapnil-signoz: Added cloud integration API for getting credentials and pass it to create account API, please review

ee/modules/cloudintegration/implcloudintegration/module.go
swapnil-signoz: We should let the caller handle this so that the module behaviour does not change.

ee/modules/cloudintegration/implcloudintegration/implcloudprovider/awscloudprovider.go
swapnil-signoz: done

ee/modules/cloudintegration/implcloudintegration/implcloudprovider/awscloudprovider.go
swapnil-signoz: Resolved

ee/modules/cloudintegration/implcloudintegration/implcloudprovider/awscloudprovider.go
swapnil-signoz: Fair, I have removed such checks

nityanandagohain (commented):
overall lgtm, let the other pr get meged and once rebased will approve.

ee/modules/cloudintegration/implcloudintegration/implcloudprovider/awscloudprovider.go
swapnil-signoz: Resolving this discussion for now
cc: @vikrantgupta25

nityanandagohain (dismissed):

pkg/types/zeustypes/types.go
```diff
@@ -56,3 +58,24 @@ func NewGettableHost(data []byte) *GettableHost {
 		Hosts: hosts,
 	}
 }
+
+// GettableDeployment represents the parsed deployment info from zeus.GetDeployment.
+type GettableDeployment struct {
+	Name         string
+	SignozAPIUrl string
```
vikrantgupta25: Do not return a custom thing from here, Get the root fields in the same heirarchy and construct the APIURL in the caller

pkg/signoz/signoz.go
vikrantgupta25: update here as well

pkg/modules/cloudintegration/implcloudintegration/handler.go
```diff
@@ -1,62 +1,493 @@
...
+	defer cancel()
+
+	claims, err := authtypes.ClaimsFromContext(ctx)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	providerString := mux.Vars(r)["cloud_provider"]
+	provider, err := cloudintegrationtypes.NewCloudProvider(providerString)
```
vikrantgupta25:
```suggestion
	provider, err := cloudintegrationtypes.NewCloudProvider(mux.Vars(r)["cloud_provider"])
```

pkg/modules/cloudintegration/implcloudintegration/handler.go
```diff
@@ -1,62 +1,493 @@
...
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	postableAccount := new(cloudintegrationtypes.PostableAccount)
+
```
vikrantgupta25: empty space

pkg/modules/cloudintegration/implcloudintegration/handler.go
```diff
@@ -1,62 +1,493 @@
...
+	err = binding.JSON.BindBody(r.Body, postableAccount)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	if err := postableAccount.Validate(provider); err != nil {
```
vikrantgupta25: in place of validate you can just grab postable account ( validate whatever you can in UnmarshalJSON), then add a method for 

```
cloudintegrationtypes.NewAccount() ( *Account,error )
```

and do the remaining validation here.

pkg/modules/cloudintegration/implcloudintegration/handler.go
```diff
@@ -1,62 +1,493 @@
...
+		return
+	}
+
+	accountConfig, err := cloudintegrationtypes.NewAccountConfigFromPostable(provider, postableAccount.Config)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	account := cloudintegrationtypes.NewAccount(valuer.MustNewUUID(claims.OrgID), provider, accountConfig)
```
vikrantgupta25, line 81: account config should have been generated in the NewAccount ?

pkg/modules/cloudintegration/implcloudintegration/handler.go
```diff
@@ -1,62 +1,493 @@
...
+		return
+	}
+
+	connectionArtifact, err := handler.module.GetConnectionArtifact(ctx, account, postableAccount)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	render.Success(rw, http.StatusOK, &cloudintegrationtypes.GettableAccountWithConnectionArtifact{
```
vikrantgupta25: have a new function for `GettableAccountWithConnectionArtifact` do not use types directly

vikrantgupta25:
```suggestion
	render.Success(rw, http.StatusCreated, &cloudintegrationtypes.GettableAccountWithConnectionArtifact{
```

pkg/modules/cloudintegration/implcloudintegration/handler.go
```diff
@@ -1,62 +1,493 @@
...
+
+	providerString := mux.Vars(r)["cloud_provider"]
+	provider, err := cloudintegrationtypes.NewCloudProvider(providerString)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	accountIDString := mux.Vars(r)["id"]
+	accountID, err := valuer.NewUUID(accountIDString)
```
vikrantgupta25:
```suggestion
	accountID, err := valuer.NewUUID(mux.Vars(r)["id"])
```

pkg/modules/cloudintegration/implcloudintegration/handler.go
```diff
@@ -1,62 +1,493 @@
...
+		return
+	}
+
+	accounts, err := handler.module.ListAccounts(ctx, valuer.MustNewUUID(claims.OrgID), provider)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	render.Success(rw, http.StatusOK, &cloudintegrationtypes.GettableAccounts{
```
vikrantgupta25: always a new function, the downside with this is chances of missing on some fields

pkg/modules/cloudintegration/implcloudintegration/handler.go
```diff
@@ -1,62 +1,493 @@
...
+		return
+	}
+
+	req := new(cloudintegrationtypes.UpdatableAccount)
+	if err := binding.JSON.BindBody(r.Body, req); err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	if err := req.Validate(provider); err != nil {
```
vikrantgupta25: same as create can all of this validation be done in account.update() ?

vikrantgupta25 (changes requested):

pkg/modules/cloudintegration/implcloudintegration/handler.go
```diff
@@ -1,62 +1,493 @@
...
+
+	providerString := mux.Vars(r)["cloud_provider"]
+	provider, err := cloudintegrationtypes.NewCloudProvider(providerString)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	var integrationID *valuer.UUID
+	if idStr := r.URL.Query().Get("cloud_integration_id"); idStr != "" {
```
vikrantgupta25: this handling is bad, its prone to errors and isn't very restful IMO.

```
/accounts/

/services/

/accounts/{id}/services/
```

should be the way to go ? WDYT

vikrantgupta25 (changes requested):

ee/modules/cloudintegration/implcloudintegration/module.go
```diff
@@ -0,0 +1,513 @@
...
+	// get license to get the deployment details
+	license, err := module.licensing.GetActive(ctx, orgID)
+	if err != nil {
+		return nil, errors.New(errors.TypeLicenseUnavailable, errors.CodeLicenseUnavailable, "a valid license is not available").WithAdditional("this feature requires a valid license").WithAdditional(err.Error())
+	}
+
+	var ingestionURL string
+
+	globalConfig := module.global.GetConfig(ctx)
+	if globalConfig != nil {
```
vikrantgupta25: why would this be nil ?

ee/modules/cloudintegration/implcloudintegration/module.go
```diff
@@ -0,0 +1,513 @@
...
+	return module.store.CreateAccount(ctx, storableCloudIntegration)
+}
+
+func (module *module) GetConnectionArtifact(ctx context.Context, account *cloudintegrationtypes.Account, req *cloudintegrationtypes.GetConnectionArtifactRequest) (*cloudintegrationtypes.ConnectionArtifact, error) {
+	cloudProviderModule, err := module.getCloudProvider(account.Provider)
+	if err != nil {
+		return nil, err
+	}
+
+	req.Config.AddAgentVersion(module.config.Agent.Version)
```
vikrantgupta25:
```suggestion
	req.Config.SetAgentVersion(module.config.Agent.Version)
```

pkg/types/zeustypes/types.go
swapnil-signoz: Currently the API is returning slice of bytes, just wanted to return concrete struct similar to `GettableHost`

vikrantgupta25: make a new struct no worries with that but do maintain the same hierarchy you get in response. 

```
deployment: {

name: "",

region: {

dns: ""

}
```

pkg/modules/cloudintegration/implcloudintegration/handler.go
vikrantgupta25: as discussed pick up in next iteration. Do create a ticket for this

pkg/modules/cloudintegration/implcloudintegration/handler.go
swapnil-signoz: I think its better we have separate `New` function AccountConfig as it is a complicated struct and its creation depends on the caller's context.

vikrantgupta25: Can do that `NewWithCOnfig`

swapnil-signoz: Adding two `NewAccountConfigFromPostable` and `NewAccountConfigFromUpdatable` functions.
NewAccount stays as is and no validations are required.

pkg/modules/cloudintegration/implcloudintegration/handler.go
swapnil-signoz: https://github.com/SigNoz/engineering-pod/issues/4616

pkg/modules/cloudintegration/implcloudintegration/handler.go
vikrantgupta25: Why not have just one `NewAccountConfig` which accepts agent version and the respective cloud provider configs. the types doesn't care if it's from Postable or Updatable ? 

```
NewAccountConfig(agentVersion string, awsConfig *AWSConfig, azureConfig *AzureConfig ) ( *AccountConfig,error ) 
```

pkg/signoz/signoz.go
```diff
@@ -426,11 +433,19 @@ func New(
 		return nil, err
 	}
 
+	serviceAccount := implserviceaccount.NewModule(implserviceaccount.NewStore(sqlstore), authz, cache, analytics, providerSettings, config.ServiceAccount)
+
+	cloudIntegrationStore := pkgimplcloudintegration.NewStore(sqlstore)
+	cloudIntegrationModule, err := cloudIntegrationCallback(cloudIntegrationStore, global, zeus, gateway, licensing, serviceAccount, config.CloudIntegration)
```
vikrantgupta25: why is cloudIntegrationStore part of the callback ?

pkg/modules/cloudintegration/implcloudintegration/handler.go
swapnil-signoz: Yes, I thought about that, just that directly adding provider's config (which will grow) did not feel right.
We will try and improve how we are handling configs right now.

pkg/types/zeustypes/types.go
```diff
@@ -56,3 +59,53 @@ func NewGettableHost(data []byte) *GettableHost {
 		Hosts: hosts,
 	}
 }
+
+// GettableDeployment represents the parsed deployment info from zeus.GetDeployment.
+// NOTE: break down deployment into multiple structs if needed.
+type GettableDeployment struct {
```
vikrantgupta25, line 64: please keep this bare minimum for now around how much you need it because the return type of the GetDeployment needs to change from bytes. This will be taken up with that

pkg/modules/cloudintegration/implcloudintegration/handler.go
```diff
@@ -1,62 +1,457 @@
...
+		return
+	}
+
+	account, err := handler.module.GetAccount(ctx, valuer.MustNewUUID(claims.OrgID), cloudIntegrationID, provider)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	if account.IsRemoved() {
```
vikrantgupta25: rather than this have 

```
handler.module.GetActiveAccount
```

vikrantgupta25: or in the `Update` method for account just check if its removed then return error

vikrantgupta25 (changes requested):

ee/modules/cloudintegration/implcloudintegration/module.go
```diff
@@ -0,0 +1,521 @@
...
+	// ignore error
+	apiKey, _ := module.getOrCreateAPIKey(ctx, orgID, provider)
+
+	// ignore error
+	ingestionKey, _ := module.getOrCreateIngestionKey(ctx, orgID, provider)
+
+	return &cloudintegrationtypes.Credentials{
+		SigNozAPIURL: signozAPIURL,
+		SigNozAPIKey: apiKey,
+		IngestionURL: module.global.GetConfig(ctx).IngestionURL,
```
vikrantgupta25: make sure the frontend logic of when to enable/disable field is in sync with this

ee/modules/cloudintegration/implcloudintegration/module.go
```diff
@@ -0,0 +1,521 @@
...
+	storableCloudIntegration, err := cloudintegrationtypes.NewStorableCloudIntegration(account)
+	if err != nil {
+		return err
+	}
+
+	return module.store.CreateAccount(ctx, storableCloudIntegration)
+}
+
+func (module *module) GetConnectionArtifact(ctx context.Context, account *cloudintegrationtypes.Account, req *cloudintegrationtypes.GetConnectionArtifactRequest) (*cloudintegrationtypes.ConnectionArtifact, error) {
+	cloudProviderModule, err := module.getCloudProvider(account.Provider)
```
vikrantgupta25, line 113: license guard please

vikrantgupta25 (changes requested):

pkg/modules/cloudintegration/implcloudintegration/handler.go
swapnil-signoz: Went ahead with first one

pkg/signoz/signoz.go
swapnil-signoz: Removed

vikrantgupta25 (approved):

PR #10720, opened 2026-03-26
chore: reject empty aggregations in payload regardless of disabled st…

srikanthccv:
…atus

Reject empty aggregation payloads.


pkg/types/querybuildertypes/querybuildertypesv5/validation_test.go
```diff
@@ -366,15 +398,25 @@ func TestQueryRangeRequest_ValidateCompositeQuery(t *testing.T) {
 							Spec: QueryBuilderQuery[LogAggregation]{
 								Name:     "A",
 								Disabled: true,
-								Signal:   telemetrytypes.SignalLogs,
+								Aggregations: []LogAggregation{
+									{
+										Expression: "count",
```
nityanandagohain: correct will be to use count() instead ?

PR #10725, opened 2026-03-26
chore: don't run functions if the series is empty

nityanandagohain:
Fixes https://github.com/SigNoz/engineering-pod/issues/4312


pkg/types/querybuildertypes/querybuildertypesv5/functions.go
```diff
@@ -326,6 +326,10 @@ func funcAbsolute(result *TimeSeries) *TimeSeries {
 
 // funcRunningDiff returns the running difference of each point
 func funcRunningDiff(result *TimeSeries) *TimeSeries {
+	if len(result.Values) == 0 {
```
srikanthccv: Should we have this check at the beginngin of `ApplyFunction`?

PR #10726, opened 2026-03-26
Alert Rules crashing due to invalid data

H4ad:
## Pull Request

---

Update the method to expects invalid data, but also prevents the query builder to emit invalid data.

To reproduce, just parse the `compositeQuery` param and remove the `queryFormulas` or `queryTraceOperator`.

Before:
image

After:

image

Fixes https://github.com/SigNoz/engineering-pod/issues/4314

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

The method was expecting valid array but received undefined.

Query Builder sometimes could emit invalid data, and if the user updates the URL, it also breaks the UI.

Prevent the query builder to fallback to undefined, and also be defensive on method that was using filter.

---

- Tests added/updated: No
- Manual verification: Yes
- Edge cases covered: Yes, by updating Query Builder, we should prevent errors like this in future.

---

- Blast radius: Alert Rules & Query Builder
- Potential regressions: If some code relied on those properties being explicitly undefined.
- Rollback plan: Revert the commit of query builder.

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Prevent the Alert Rule Details to crash when query has missing properties. |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/providers/QueryBuilder.tsx
+++ b/frontend/src/providers/QueryBuilder.tsx
@@ -165,20 +165,23 @@ export function QueryBuilderProvider({
 	const prepareQueryBuilderData = useCallback(
 		(query: Query): Query => {
 			const builder: QueryBuilderData = {
-				queryData: query.builder.queryData?.map((item) => ({
-					...initialQueryBuilderFormValuesMap[
-						initialDataSource || DataSource.METRICS
-					],
-					...item,
-				})),
-				queryFormulas: query.builder.queryFormulas?.map((item) => ({
-					...initialFormulaBuilderFormValues,
-					...item,
-				})),
-				queryTraceOperator: query.builder.queryTraceOperator?.map((item) => ({
-					...initialQueryBuilderFormTraceOperatorValues,
-					...item,
-				})),
+				queryData:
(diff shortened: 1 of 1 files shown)


AshwinBhatkal:
We can close this github ticket from sentry and this PR as well. Also, archive the sentry issue for 3 months time. 

None of this operators will ever be undefined unless it is created by an LLM today. We will handle this properly with the json schema migration. The reason to avoid this right now is to catch all such cases later when we actually do the migration and write them properly

H4ad:
The fix is really small, I don't think worth to let the UI break for few weeks/months just because we still working  in the JSON migration.

> The reason to avoid this right now is to catch all such cases later when we actually do the migration and write them properly

Do you have docs about this migration and what exactly it will change/help?

H4ad:
After some talk, we decided to postpone the fix to when we migrate the dashboard to new schema because these bugs are caused by malformed queries likely created by AI.

cc @srikanthccv

Issue #10611, opened 2026-03-17
[web] unable to copy invite link in browser over http


darestov:
## Bug description
Hello!
I want to invite a user by sending him an invite link. I created the invitation (pending) and when I hit the "Copy invite link" button, nothing was copied and an alert "Failed to copy invite link" was shown. No web requests, no errors in console.

<img width="1712" height="1347" alt="Image" src="https://github.com/user-attachments/assets/8f7f14d6-8e05-4e03-90a9-60a77c3e050d" />

## Version information 
* **Signoz version**: 0.115
* **Browser version**:
* **Your OS and version**:
* **Your CPU Architecture**(ARM/Intel): 

## Additional context
I was able to copy invite links in SigNoz v0.110.

SagarRajput-7:
@darestov - to further investigate and reproduce this issue, I would require a few pieces of information from your end, as this is working fine on multiple other platforms with the same Signoz version you mentioned.

- Browser & OS:  info & version
- How are you running Signoz? - local or deployed 

Why do I need these? - Since for this copy we are using Native Web API - `navigator.clipboard` provided by browsers.

Lastly, are you able to use other copy functions in the app? 
- like one example you can find at the
   - Password Reset Link: `Active Member --> click on --> Generate Password Reset Link`

darestov:
@SagarRajput-7 
Browser: Yandex Browser 26.3.0.2166 (Chrome based) , also reproduces in vanilla Chrome 146.0.7680.80.
OS: Windows 11 25H2 (26200) 64bit

I run Signoz in Docker (free license).

> Active Member --> click on --> Generate Password Reset Link
Ends up with this (failure): 

<img width="2066" height="1242" alt="Image" src="https://github.com/user-attachments/assets/9250dd1a-1810-4afa-bcce-b528df281ab8" />

At the same time, Copy Field Value (from span) works well, I'm able to paste it elsewhere.

SagarRajput-7:
@darestov ,

[...]

PR #10729, opened 2026-03-26
fix(member-drawer): use hook to copy to support more browsers

H4ad:
## Pull Request

---

Use hook to copy to clipboard instead of navigator.clipboard to support more browsers.

I also added a new Eslint rule to prevent the usage of this navigator.clipboard in the future.

After:

https://github.com/user-attachments/assets/43a3c4e1-05aa-42f7-89e5-3a36a44e777d
> Success case

https://github.com/user-attachments/assets/9495028a-09f7-45c8-a456-c32259f2305b
> When copy fails

Fixes https://github.com/SigNoz/signoz/issues/10611
---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

The navigator.clipboard can fail for specific browsers, the new implementation uses copy-to-clipboard library behind react-use.

Reference code:
- https://github.com/streamich/react-use/blob/master/src/useCopyToClipboard.ts#L26-L57
- https://github.com/sudodoki/copy-to-clipboard/blob/main/index.js

The navigator.clipboard was not available.

By using a lib, it supports multiple ways to allow the user to copy the string.

---

- Tests added/updated: No
- Manual verification: Yes
- Edge cases covered:

---

- Blast radius: Password Reset Copy & Add Key Modal
- Potential regressions: -

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Use multiple methods to copy to clipboard on copy password reset link to support more browsers. |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/.eslintrc.cjs
+++ b/frontend/.eslintrc.cjs
@@ -219,7 +219,7 @@ module.exports = {
 						object: 'navigator',
 						property: 'clipboard',
 						message:
-							'Do not use navigator.clipboard directly since it does work well with specific browsers. Use hook useCopyToClipboard from react-use library. https://streamich.github.io/react-use/?path=/story/side-effects-usecopytoclipboard--docs',
+							'Do not use navigator.clipboard directly since it does not work well with specific browsers. Use hook useCopyToClipboard from react-use library. https://streamich.github.io/react-use/?path=/story/side-effects-usecopytoclipboard--docs',
 					},
 				],
 			},


frontend/.eslintrc.cjs
```diff
@@ -205,6 +205,25 @@ module.exports = {
...
+				'**/__tests__/**/*.{js,jsx,ts,tsx}',
+			],
+			rules: {
+				'no-restricted-properties': [
+					'error',
+					{
+						object: 'navigator',
+						property: 'clipboard',
+						message:
+							'Do not use navigator.clipboard directly since it does work well with specific browsers. Use hook useCopyToClipboard from react-use library. https://streamich.github.io/react-use/?path=/story/side-effects-usecopytoclipboard--docs',
```
AshwinBhatkal: does not work well?

H4ad: For specific browsers, and under http for specific cases, it does not allow to copy.

So the lib uses a lot of fallbacks to support this behavior.

AshwinBhatkal: Text change
The text is "navigator.clipboard does work well"
It should be "navigator.clipboard does not work well"

Issue #10722, opened 2026-03-26
Throw a warning instead of an error when metric is present but data is not seen in the query range


namanv3:
Users can have dashboards with metrics that sometimes emit data but sometimes don't cuz there is nothing to report (for example some vendor's SDK may emit failure metrics only when there is a failure). This can be especially true for delta metrics.

For such metrics, if there is a dashboard, we should not fail the query API if there is no data in the time range, only give a warning (like we do when the step interval is too low).

However, we should continue throwing an error if the metric name is something that we have never seen.

PR #10737, opened 2026-03-27
fix: warning instead of error for dormant metrics in query range API

namanv3:
## Pull Request

---

### 📄 Summary
Users can have dashboards with metrics that sometimes emit data but sometimes don't cuz there is nothing to report (for example some vendor's SDK may emit failure metrics only when there is a failure). This can be especially true for delta metrics.

For such metrics, if there is a dashboard, we should not fail the query API if there is no data in the time range, only give a warning (like we do when the step interval is too low).

When a metric is dormant
Screenshot 2026-03-30 at 13 07 26
When a query has 2 metrics but one of them is dormant
Screenshot 2026-03-30 at 13 07 37
When a metric in a query does not exist
Screenshot 2026-03-30 at 13 07 43
If the user selects fill zero, the warning is shown and a zero line is also shown
Screenshot 2026-03-30 at 21 52 58

#### Issues closed by this PR

Closes https://github.com/SigNoz/signoz/issues/10722

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🧪 Testing Strategy

Integration tests added

---

| Field | Value |
|------|-------|
| Deployment Type | All |
| Change Type |  Bug Fix |
| Description | If a metric is dormant, the dashboard panels that refer to them should throw a warning instead of an error |

---

### 📋 Checklist
- [x] Manually tested

---


srikanthccv:
What does applying `fillZero` function on dormant metric do? In other words, how does user perform _"What I'd like to do is coerce these metrics to 0 if not found"_?

namanv3:
@srikanthccv fillZero option shows the zero line for that metric, and the warning is still visible.

PR #10751, opened 2026-03-29
fix: prevent index out-of-bounds panics in inframetrics sort, query range, ClickHouse reader, and session email parsing

corazzione:
## Summary

This PR fixes several index out-of-bounds panics that can occur in production when data series have no points, table names are malformed, or emails lack a domain.

### Changes

**`pkg/query-service/app/inframetrics/` (11 files — clusters, pods, hosts, nodes, namespaces, daemonsets, deployments, jobs, statefulsets, processes, pvcs)**

Each file's `sort.Slice` comparator accessed `Series[i].Points[0]` without first checking whether the `Points` slice was non-empty. If a series returns with zero data points (e.g., a freshly-created resource with no metrics yet), the comparator panics with an index out-of-bounds error.

**Fix:** added a `len(Points) == 0` guard inside each comparator that returns `false` (sorts empty-points series last) instead of panicking.

**`pkg/query-service/common/query_range.go`**

`series.Points[0]` and `series.Points[len(series.Points)-1]` were accessed unconditionally after copying the points slice. An empty `Points` slice (possible when all points are filtered out upstream) causes a panic.

**Fix:** wrapped both boundary-trim operations in a `len(points) > 0` guard.

**`pkg/query-service/app/clickhouseReader/reader.go`**

`getLocalTableName` split on `"."` and `"distributed_"` and immediately indexed `[1]` without bounds checking. If the table name doesn't contain a dot or the second segment doesn't contain `"distributed_"`, this panics.

`getLocalTableNameArray` duplicated the same unsafe pattern inline.

**Fix:** added explicit length checks with safe fallbacks in `getLocalTableName`; refactored `getLocalTableNameArray` to delegate to `getLocalTableName`, removing the duplicated unsafe code.

**`pkg/modules/session/implsession/module.go`**

`strings.Split(email.String(), "@")[1]` assumed the email always contains `"@"`. While the `email` value is typed, a defensive check prevents a panic if an unexpected value reaches this code path.

**Fix:** replaced the direct index with a guarded split that returns a typed `InvalidInput` error when no domain is present.

## Test plan

- [ ] Verify infra metrics pages (Clusters, Pods, Hosts, Nodes, Namespaces, DaemonSets, Deployments, Jobs, StatefulSets, Processes, PVCs) render correctly when some resources have no data points in the selected time range
- [ ] Verify query range caching still works correctly for normal (non-empty) series
- [ ] Verify TTL operations work correctly for logs, traces, and metrics
- [ ] Verify login/session context endpoint works for valid emails

---

> [!NOTE]
> **Low Risk**
> Low risk defensive changes that add bounds checks and safe fallbacks to prevent panics when inputs/series are empty or malformed; behavior should only change for previously-crashing edge cases.
> 
> **Overview**
> Prevents several production panics by adding **bounds checks** before indexing into split strings and metric point arrays.
> 
> Session context email parsing now validates the presence of an `@` domain segment and returns an `InvalidInput` error instead of panicking. In the query-service, `inframetrics` sorting now handles series with empty `Points` (sorting empties last), `query_range` guards boundary trimming when `points` is empty, and ClickHouse `getLocalTableName`/`getLocalTableNameArray` safely handle unexpected table name formats and remove duplicated unsafe parsing logic.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e7cd6f188452fe3b76b07884e6aa1e62273c9fb3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/query-service/app/inframetrics/clusters.go
+++ b/pkg/query-service/app/inframetrics/clusters.go
@@ -184,15 +184,19 @@ func (p *ClustersRepo) getTopClusterGroups(ctx context.Context, orgID valuer.UUI
 
 	if req.OrderBy.Order == v3.DirectionDesc {
 		sort.Slice(formattedResponse[0].Series, func(i, j int) bool {
-			if len(formattedResponse[0].Series[i].Points) == 0 || len(formattedResponse[0].Series[j].Points) == 0 {
-				return false
+			iEmpty := len(formattedResponse[0].Series[i].Points) == 0
+			jEmpty := len(formattedResponse[0].Series[j].Points) == 0
+			if iEmpty || jEmpty {
+				return !iEmpty && jEmpty
 			}
 			return formattedResponse[0].Series[i].Points[0].Value > formattedResponse[0].Series[j].Points[0].Value
 		})
 	} else {
 		sort.Slice(formattedResponse[0].Series, func(i, j int) bool {
-			if len(formattedResponse[0].Series[i].Points) == 0 || len(formattedResponse[0].Series[j].Points) == 0 {
-				return false
+			iEmpty := len(formattedResponse[0].Series[i].Points) == 0
+			jEmpty := len(formattedResponse[0].Series[j].Points) == 0
+			if iEmpty || jEmpty {
+				return !iEmpty && jEmpty
 			}
 			return formattedResponse[0].Series[i].Points[0].Value < formattedResponse[0].Series[j].Points[0].Value
 		})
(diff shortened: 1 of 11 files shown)


corazzione:
**Update:** Fixed a strict weak ordering violation flagged in code review.

The original guard returned `false` for both `less(empty, nonEmpty)` and `less(nonEmpty, empty)`, making empty-points series \"equal\" to every non-empty series. This violates `sort.Slice`'s transitivity-of-incomparability requirement and could place empty-points series anywhere in paginated results instead of consistently at the end.

**Corrected comparator:**
\`\`\`go
iEmpty := len(formattedResponse[0].Series[i].Points) == 0
jEmpty := len(formattedResponse[0].Series[j].Points) == 0
if iEmpty || jEmpty {
    return !iEmpty && jEmpty
}
\`\`\`

This gives strict weak ordering: `less(nonEmpty, empty) = true`, `less(empty, nonEmpty) = false`, `less(empty, empty) = false`. Applied to all 11 inframetrics files.

PR #10755, opened 2026-03-30
chore: added make py-fmt and make py-lint in pre-commit hook

tushar-signoz:
## Pull Request
chore: added make py-fmt and make py-lint in pre-commit hook

---

### Summary
Added Python formatting and linting checks to the Husky pre-commit hook
Checks only run when Python files (.py) are staged, avoiding unnecessary overhead on commits with no Python changes
If make py-fmt modifies any files, those changes are automatically staged so the commit proceeds without interruption
make py-lint runs after formatting to catch any remaining lint issues

### Motivation
Ensures consistent Python code style and quality at commit time without requiring developers to manually run formatters/linters before committing.

### Behavior
Detects staged .py files via git diff --cached
Runs make py-fmt — auto-stages any files it modifies
Runs make py-lint — fails the commit if lint errors are found
Skips both steps entirely if no Python files are staged

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---


nityanandagohain (approved):

AshwinBhatkal (changes requested):
This is not the right place to do this. We can do other things to achieve what we want to here. 

Check out updating vs code settings json instead of this. The broader solution is to have pre-commit checks at the repo level. cc: @therealpandey 

We should not have BE checks trickle into frontend code. This will also break those commits in machines which have not run yarn at all. For this to work, running yarn is a pre requisite.

PR #10758, opened 2026-03-30
chore: align usage of negative operators in JSON Body

piyushsingariya:
## Pull Request

---

### 📄 Summary
JSON body filtering in ClickHouse logs (via body_v2) involves navigating dynamically typed nested structures that can contain both Array(JSON) and Array(Dynamic) at any intermediate hop. Previously, the condition builder had issues:
- Negative operators with array paths were not handled correctly — NOT was being pushed inside the arrayExists lambda, which produced semantically wrong SQL (e.g., `arrayExists(x -> x != "foo", arr)` returns true if any element doesn't match, rather than if no element matches).
- Negative operators with Value comparison on top level paths was also applying default Exists check which does not align with QBV5 experience.

This PR fixes:

- Extracting a top-level NOT wrapper via applyNotCondition, so negative operators are inverted before entering the arrayExists lambda and then negated at the outermost level.
- AssumingNotNull when Negative operators are applied onto top level paths (non-nested)

Closes SigNoz/engineering-pod#4340

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor

---

- Tests added/updated: json_stmt_builder_test.go updated with ~876 lines covering primitive conditions, array membership (both Array(JSON) and Array(Dynamic) branches), negative operators, indexed path conditions, and nested array hops.
- Manual verification: Query output validated against expected ClickHouse SQL for all operator/type combinations.
- Edge cases covered: top-level array paths with negative operators, mixed Array(JSON) + Array(Dynamic) branches at the same hop, non-nested path index guard, missing ColumnExpression on fresh indexes.

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- The key behavioral change is in recurseArrayHops: it now strips the NOT at the top, passes the inverted operator down into buildAccessNodeBranches/buildTerminalCondition, and re-applies NOT at the outermost expression. This is correct because `NOT arrayExists(x -> cond(x), arr) ≡ arrayAll(x -> NOT cond(x), arr)` — but we express it as the former since ClickHouse optimizes it equivalently.
- `applyNotCondition` is called twice in the tree — once at recurseArrayHops (for intermediate array hops) and once inside buildTerminalCondition (for terminal array nodes where the top-level path itself is an array). The second call only fires when the terminal is the root of traversal (not reached via buildAccessNodeBranches, which already inverted the operator).
- `FilterOperator.Inverse()` returns the zero value (`FilterOperatorEqual`) for any unmapped operator — callers should only invoke it after confirming `IsNegativeOperator()`.
---

### DIFF
--- a/frontend/src/AppRoutes/Private.tsx
+++ b/frontend/src/AppRoutes/Private.tsx
@@ -1,4 +1,4 @@
-import { ReactChild, useCallback, useEffect, useMemo, useState } from 'react';
+import { ReactChild, useCallback, useMemo } from 'react';
 import { matchPath, Redirect, useLocation } from 'react-router-dom';
 import getLocalStorageApi from 'api/browser/localstorage/get';
 import setLocalStorageApi from 'api/browser/localstorage/set';
@@ -8,12 +8,10 @@ import { LOCALSTORAGE } from 'constants/localStorage';
 import { ORG_PREFERENCES } from 'constants/orgPreferences';
 import ROUTES from 'constants/routes';
 import { useGetTenantLicense } from 'hooks/useGetTenantLicense';
-import history from 'lib/history';
 import { isEmpty } from 'lodash-es';
 import { useAppContext } from 'providers/App/App';
 import { LicensePlatform, LicenseState } from 'types/api/licensesV3/getActive';
 import { OrgPreference } from 'types/api/preferences/preference';
-import { Organization } from 'types/api/user/getOrganization';
 import { USER_ROLES } from 'types/roles';
 import { routePermission } from 'utils/permission';
 
@@ -25,6 +23,7 @@ import routes, {
 	SUPPORT_ROUTE,
 } from './routes';
 
+// eslint-disable-next-line sonarjs/cognitive-complexity
 function PrivateRoute({ children }: PrivateRouteProps): JSX.Element {
 	const location = useLocation();
 	const { pathname } = location;
@@ -57,7 +56,12 @@ function PrivateRoute({ children }: PrivateRouteProps): JSX.Element {
 	const currentRoute = mapRoutes.get('current');
 	const { isCloudUser: isCloudUserVal } = useGetTenantLicense();
 
-	const [orgData, setOrgData] = useState<Organization | undefined>(undefined);
+	const orgData = useMemo(() => {
+		if (org && org.length > 0 && org[0].id !== undefined) {
+			return org[0];
+		}
+		return undefined;
+	}, [org]);
 
 	const { data: usersData, isFetching: isFetchingUsers } = useListUsers({
 		query: {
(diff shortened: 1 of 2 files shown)


pkg/types/querybuildertypes/querybuildertypesv5/builder_elements.go
```diff
@@ -113,6 +113,29 @@ const (
 	FilterOperatorNotContains
 )
 
+var operatorInverseMapping = map[FilterOperator]FilterOperator{
+	FilterOperatorEqual:           FilterOperatorNotEqual,
+	FilterOperatorNotEqual:        FilterOperatorEqual,
+	FilterOperatorGreaterThan:     FilterOperatorLessThanOrEq,
```
nityanandagohain, line 119:
```suggestion
	FilterOperatorGreaterThan:     FilterOperatorLessThan,
```

pkg/querybuilder/fallback_expr.go
```diff
@@ -277,13 +277,29 @@ func DataTypeCollisionHandledFieldName(key *telemetrytypes.TelemetryFieldKey, va
...
+		}
 	}
 	return tblFieldName, value
 }
 
 func castFloat(col string) string     { return fmt.Sprintf("toFloat64OrNull(%s)", col) }
 func castFloatHack(col string) string { return fmt.Sprintf("toFloat64(%s)", col) }
 func castString(col string) string    { return fmt.Sprintf("toString(%s)", col) }
+func castBool(col string) string      { return fmt.Sprintf("accurateCastOrNull(%s, 'Bool')", col) }
+func accurateCastFloat(col string) string {
```
nityanandagohain, line 302: Its it redundant to castFloat because of this case ?

 when I run
select toFloat64OrNull(100) , it gives me an error

while select accurateCastOrNull(100, 'Float64') doesn't give me error

pkg/querybuilder/fallback_expr.go
```diff
@@ -277,13 +277,29 @@ func DataTypeCollisionHandledFieldName(key *telemetrytypes.TelemetryFieldKey, va
...
+		switch v := value.(type) {
+		case string:
+			tblFieldName = castString(tblFieldName)
+		case float32, float64, int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64:
+			tblFieldName = accurateCastFloat(tblFieldName)
+		case bool:
+			tblFieldName = castBool(tblFieldName)
+		case []any:
+			// dynamic array elements will be default casted to string
+			tblFieldName, value = castString(tblFieldName), toStrings(v)
```
nityanandagohain, line 292: this feels weired, Is this case possible right now from the quiery builder?

pkg/querybuilder/fallback_expr.go
piyushsingariya: Yes, due to this case added new function

pkg/querybuilder/fallback_expr.go
piyushsingariya: Added to cover possibility of variation with value param input

pkg/telemetrylogs/json_condition_builder.go
```diff
@@ -33,133 +34,228 @@ func NewJSONConditionBuilder(key *telemetrytypes.TelemetryFieldKey, valueType te
...
+	conditions := []string{}
+	for _, node := range c.key.JSONPlan {
+		condition, err := c.recurseArrayHops(node, operator, value, sb)
+		if err != nil {
+			return "", err
+		}
+		conditions = append(conditions, condition)
+	}
+
+	return sb.Or(conditions...), nil
```
tushar-signoz, line 62: Is this situation possible ? `abc` has multiple types, lets say string and int
`a.b.c != 25` => `NOT(a.b.c = 25) OR NOT(a.b.c = '25')` -> always true

piyushsingariya: This is a Top Level responsibility by WhereClauseVisitor. We can check `body.education[].type != '10001'` testcase.

I looked it up into WhereClauseVisitor
```
if op.IsNegativeOperator() {
	return v.builder.And(conds...)
}
```

PR #10760, opened 2026-03-30
chore: remove deprecated v3/v4 support in rules

srikanthccv:
## Pull Request

Part of https://github.com/SigNoz/engineering-pod/issues/3199


ee/anomaly/params.go
```diff
@@ -32,7 +32,7 @@ func (s Seasonality) IsValid() bool {
 }
 
 type AnomaliesRequest struct {
-	Params      qbtypes.QueryRangeRequest
```
jatinderjit, line 35: Any reason to make this a pointer?
I didn't see any read access with a nil check.

ee/query-service/rules/anomaly.go
```diff
@@ -5,58 +5,39 @@ import (
...
-	anomalyV2 "github.com/SigNoz/signoz/ee/anomaly"
+	"github.com/SigNoz/signoz/ee/anomaly"
 
 	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
 )
 
-const (
-	RuleTypeAnomaly = "anomaly_rule"
+var (
+	RuleTypeAnomaly = ruletypes.RuleType{String: valuer.NewString("anomaly_rule")}
```
jatinderjit:
```suggestion
	RuleTypeAnomaly = ruletypes.RuleTypeAnomaly
```

ee/query-service/rules/anomaly.go
```diff
@@ -5,58 +5,39 @@ import (
...
 	*baserules.BaseRule
 
 	mtx sync.Mutex
 
-	reader interfaces.Reader
-
-	// querierV2 is used for alerts created after the introduction of new metrics query builder
-	querierV2 interfaces.Querier
-
 	// querierV5 is used for alerts migrated after the introduction of new query builder
```
jatinderjit:
```suggestion
	// querier is used for alerts migrated after the introduction of new query builder
```

ee/query-service/rules/anomaly.go
```diff
@@ -158,36 +110,9 @@ func (r *AnomalyRule) Type() ruletypes.RuleType {
...
-		End:            end,
-		Step:           int64(math.Max(float64(common.MinAllowedStepInterval(start, end)), 60)),
-		CompositeQuery: compositeQuery,
-		Variables:      make(map[string]interface{}, 0),
-		NoCache:        false,
-	}, nil
-}
-
-func (r *AnomalyRule) prepareQueryRangeV5(ctx context.Context, ts time.Time) (*qbtypes.QueryRangeRequest, error) {
+func (r *AnomalyRule) prepareQueryRange(ctx context.Context, ts time.Time) (*qbtypes.QueryRangeRequest, error) {
```
jatinderjit: This method always returns a nil error.
```suggestion
func (r *AnomalyRule) prepareQueryRange(ctx context.Context, ts time.Time) qbtypes.QueryRangeRequest {
```

ee/query-service/rules/anomaly.go
```diff
@@ -70,18 +51,17 @@ func NewAnomalyRule(
 	id string,
 	orgID valuer.UUID,
 	p *ruletypes.PostableRule,
-	reader interfaces.Reader,
-	querierV5 querierV5.Querier,
+	querier querier.Querier,
 	logger *slog.Logger,
 	cache cache.Cache,
```
jatinderjit: `cache` isn't used anymore. This too can be deleted.

ee/query-service/rules/anomaly.go
```diff
@@ -341,13 +222,9 @@ func (r *AnomalyRule) Eval(ctx context.Context, ts time.Time) (int, error) {
...
 	var err error
 
-	if r.version == "v5" {
-		r.logger.InfoContext(ctx, "running v5 query")
-		res, err = r.buildAndRunQueryV5(ctx, r.OrgID(), ts)
-	} else {
-		r.logger.InfoContext(ctx, "running v4 query")
-		res, err = r.buildAndRunQuery(ctx, r.OrgID(), ts)
-	}
+	r.logger.InfoContext(ctx, "running v5 query", slog.String("rule.id", r.ID()))
```
jatinderjit: Do we need to specify "v5" now?

frontend/src/api/generated/services/sigNoz.schemas.ts
```diff
@@ -2729,20 +2721,20 @@ export interface RulestatehistorytypesGettableRuleStateHistoryDTO {
...
+	overallState: RuletypesAlertStateDTO;
 	/**
 	 * @type boolean
 	 */
 	overallStateChanged: boolean;
 	/**
 	 * @type string
 	 */
-	ruleID: string;
+	ruleId: string;
```
jatinderjit, line 2732: This is a breaking change, right?
Would it matter anywhere?

pkg/types/rulestatehistorytypes/response.go
```diff
@@ -11,16 +12,16 @@ type GettableRuleStateTimeline struct {
...
-	RuleName            string           `json:"ruleName" required:"true"`
-	OverallState        AlertState       `json:"overallState" required:"true"`
-	OverallStateChanged bool             `json:"overallStateChanged" required:"true"`
-	State               AlertState       `json:"state" required:"true"`
-	StateChanged        bool             `json:"stateChanged" required:"true"`
-	UnixMilli           int64            `json:"unixMilli" required:"true"`
-	Labels              []*qbtypes.Label `json:"labels" required:"true"`
-	Fingerprint         uint64           `json:"fingerprint" required:"true"`
-	Value               float64          `json:"value" required:"true"`
+	RuleID              string               `json:"ruleId" required:"true"`
```
jatinderjit, line 15: Is this intentional?
(related to a prev comment on breaking change).

pkg/ruler/rulestore/sqlrulestore/rule.go
```diff
@@ -139,8 +139,7 @@ func (r *rule) GetStoredRulesByMetricName(ctx context.Context, orgID string, met
 
 		// Check conditions: must be metric-based alert with valid composite query
 		if ruleData.AlertType != ruletypes.AlertTypeMetric ||
-			ruleData.RuleCondition == nil ||
-			ruleData.RuleCondition.CompositeQuery == nil {
```
jatinderjit, line 143: Does this mean `CompositeQuery` won't ever be nil?
Secondary question: so do we still need it to be a pointer?

pkg/query-service/rules/prom_rule.go
```diff
@@ -55,97 +50,70 @@ func NewPromRule(
...
-		// If the user has explicitly set the selected query, we return that.
-		if r.ruleCondition.SelectedQuery != "" {
-			return r.ruleCondition.SelectedQuery
-		}
-		// Historically, we used to have only one query in the alerts for promql.
-		// So, if there is only one query, we return that.
-		// This is to maintain backward compatibility.
-		// For new rules, we will have to explicitly set the selected query.
-		return "A"
+func (r *PromRule) GetSelectedQuery(ctx context.Context) string {
```
jatinderjit: This function is now duplicated across the three `Rule` implementations (except that anomaly's method is `SelectedQuery`). Maybe we can move this to `BaseRule`?

ee/anomaly/params.go
srikanthccv: following the big structs to be pointer to avoid copying data (change some other ref also to the pointers) and it should never be nil

pkg/ruler/rulestore/sqlrulestore/rule.go
srikanthccv: Yes, it should not be nil. It should be rejected if it's invalid. Keeping the non-trivial structs and primitives to be pointers.

frontend/src/api/generated/services/sigNoz.schemas.ts
srikanthccv: yes, these are new APIs so not used anywhere at the moment.

PR #10761, opened 2026-03-30
feat(audit): add auditortypes package

therealpandey:
## Summary
- Add `pkg/types/auditortypes/` — foundational type package for the audit log system
- `AuditEvent` struct with all fields: principal, action, resource, error, transport context, body, event name, timestamps, trace correlation
- `Store` interface with single `Emit(ctx, event) error` method
- `Config` struct for OTLP HTTP endpoint, channel buffer size (default 1000), flush interval (default 1s), max batch size (default 100)
- Constants for `Action`, `Outcome`, `PrincipalType`, `ActionCategory` with `Enum()` methods
- `EventName()` derivation helper implementing `{resource_name}.{pastTense(action)}` rule

## Test plan
- [x] `go test -race ./pkg/types/auditortypes/...` passes
- [x] `EventName()` produces correct results (`dashboard.created`, `session.login`, `user.role.updated`)
- [x] `PastTense()` maps all actions correctly
- [x] All `Enum()` methods return expected values
- [x] JSON round-trip serialization works
- [x] Zero dependencies on other audit packages
- [x] Linter passes with 0 issues

closes SigNoz/platform-pod#1929


pkg/types/auditortypes/action.go
```diff
@@ -0,0 +1,52 @@
...
+
+var (
+	ActionCreate = Action{valuer.NewString("create")}
+	ActionUpdate = Action{valuer.NewString("update")}
+	ActionDelete = Action{valuer.NewString("delete")}
+	ActionLogin  = Action{valuer.NewString("login")}
+	ActionLogout = Action{valuer.NewString("logout")}
+	ActionLock   = Action{valuer.NewString("lock")}
+	ActionUnlock = Action{valuer.NewString("unlock")}
+	ActionRevoke = Action{valuer.NewString("revoke")}
```
therealpandey:
```suggestion
```

Remove all 4 of these. We will try to reuse cud as much as possible.

pkg/types/auditortypes/action.go
```diff
@@ -0,0 +1,52 @@
...
+		ActionLogout,
+		ActionLock,
+		ActionUnlock,
+		ActionRevoke,
+	}
+}
+
+// pastTenseMap maps present-tense action strings to their past-tense form
+// used in EventName derivation.
+var pastTenseMap = map[string]string{
```
therealpandey:
```suggestion
```

Do not create a separate map. Instead extend the action struct to have a property (something called 'paste-tense' and define it in the vars of Action struct.

pkg/types/auditortypes/category.go
```diff
@@ -0,0 +1,22 @@
+package auditortypes
+
+import "github.com/SigNoz/signoz/pkg/valuer"
+
+// ActionCategory classifies the audit event per IEC 62443.
```
therealpandey, line 5:
```suggestion
```

Add a web link to the direct rfc so that I can open and see whether all categories have been added.

pkg/types/auditortypes/category.go
```diff
@@ -0,0 +1,22 @@
...
+type ActionCategory struct{ valuer.String }
+
+var (
+	CategoryAccessControl       = ActionCategory{valuer.NewString("access_control")}
```
therealpandey:
```suggestion
	ActionCategoryAccessControl       = ActionCategory{valuer.NewString("access_control")}
```

Always remember, enums start with the name `ActionCategory` in this case and then add the type also.

pkg/types/auditortypes/config.go
```diff
@@ -0,0 +1,33 @@
+package auditortypes
```
therealpandey: This is not correct and doesnt follow the module pattern. This config needs to be when we introduce the auditor service. Remove this scope from the ticket also. When removing, strikethrough instead of deleting and mention the rationale.

pkg/types/auditortypes/event_name.go
```diff
@@ -0,0 +1,13 @@
...
+
+// EventName derives the audit event name from a resource name and action.
+// Format: {resource_name}.{pastTense(action)}
+//
+// Examples:
+//
+//	EventName("dashboard", ActionCreate)  → "dashboard.created"
+//	EventName("session", ActionLogin)     → "session.login"
+//	EventName("user.role", ActionUpdate)  → "user.role.updated" (sub-resource)
+func EventName(resourceName string, action Action) string {
```
therealpandey:
```suggestion
func NewEventName(resourceName string, action Action) string {}
```

create a struct with `type EventName struct { s string}`. This way, not every string will qualify as an event name.

pkg/types/auditortypes/principal.go
```diff
@@ -0,0 +1,22 @@
+package auditortypes
+
+import "github.com/SigNoz/signoz/pkg/valuer"
+
+// PrincipalType identifies the kind of actor that performed the action.
+type PrincipalType struct{ valuer.String }
```
therealpandey, line 7:
```suggestion
//
type PrincipalType struct{ valuer.String }
```

Make a TODO here to use this from `coretypes` once the coretypes package is available.

pkg/types/auditortypes/event.go
```diff
@@ -0,0 +1,48 @@
+package auditortypes
```
therealpandey, line 1: Name this final event.go.

Also move the store interface inside this instead of having a separate file.

Start this struct in the order that OTel describes a log record.

pkg/types/auditortypes/action.go
therealpandey: Removed login/logout/lock/unlock/revoke — only create/update/delete remain.

pkg/types/auditortypes/action.go
therealpandey: Replaced pastTenseMap with a `pastTense string` field on the Action struct, defined inline in each var declaration.

pkg/types/auditortypes/category.go
therealpandey: Added IEC 62443 reference URL to the ActionCategory comment.

pkg/types/auditortypes/category.go
therealpandey: Renamed all vars to `ActionCategory*` prefix (e.g. `ActionCategoryAccessControl`).

pkg/types/auditortypes/config.go
therealpandey: Deleted config.go — config belongs with the auditor service. Issue updated with strikethrough and rationale.

pkg/types/auditortypes/event_name.go
therealpandey: Created `type EventName struct { s string }` with `NewEventName()` constructor and `String()` method.

pkg/types/auditortypes/principal.go
therealpandey: Added TODO: `// TODO: use PrincipalType from coretypes once the coretypes package is available.`

pkg/types/auditortypes/event.go
therealpandey: Renamed to event.go, merged Store interface into it, deleted store.go. Reordered AuditEvent fields to follow OTel LogRecord structure (timestamps → trace/span → body → event name → attributes).

pkg/types/auditortypes/action_test.go
```diff
@@ -0,0 +1,57 @@
+package auditortypes
```
therealpandey: This entire test file is full of redundant tests. Please remove them.

pkg/types/auditortypes/event_name_test.go
```diff
@@ -0,0 +1,48 @@
+package auditortypes
```
therealpandey: All these tests are useless. Please remove.

pkg/types/auditortypes/action_test.go
therealpandey: Removed both test files.

pkg/types/auditortypes/event_name_test.go
therealpandey: Removed both test files.

Issue #8796, opened 2025-08-13
Return warning if the timestamps of the values of the queries in formula are not aligned


srikanthccv:
Assume there are two queries A and B; both are the same except the B is time-shifted by 30 seconds. If I use the formula A-B to plot the change in the window, it returns no data because the timestamps are not aligned. Help the user by suggesting the change in time shift value.

tushar-signoz:
Hey @srikanthccv 
Problem: Bucket misalignment when comparing time-shifted series

When using timeShift to compare current data against a previous period, ClickHouse's toStartOfInterval aligns buckets to the epoch. If shift is not a multiple of step, the shifted series' buckets land at different positions than the original series' buckets. After the app-level forward-shift, timestamps don't line up and the two series can't be compared point-to-point.

Possible solutions (combinations would also work):

1. Reject queries where shift < step -- Return an error if the user requests a shift smaller than the step interval. Simple and explicit, but overly restrictive: it blocks valid use cases where shift >= step but shift % step != 0 (e.g. shift=1h, step=7min), and it doesn't actually fix the misalignment for those cases.

2. Round shift to the nearest multiple of step -- Snap the requested shift to `round(shift / step) * step` (with a floor to step if the result would be zero). This eliminates misalignment entirely without SQL changes, since shift % step == 0 guarantees epoch-aligned buckets already line up. However, it silently alters the user's requested shift, and the rounding error can be significant for small shift/step ratios (e.g. shift=400s, step=300s rounds to 300s -- a 25% change in the comparison window).

3. Align bucket boundaries in SQL -- Use `toStartOfInterval(col + shift, step) - shift` so that bucket boundaries, after the app-level forward-shift, land exactly on the same positions as the original series. This is the only solution that fixes misalignment for all shift/step combinations, preserves the user's exact requested shift, and does so without rejecting or altering the input.

ethanmontgomery148-hub:
Hey @srikanthccv — good catch. Aligning bucket boundaries in SQL
(toStartOfInterval(col + shift, step) - shift) feels like the cleanest fix
since it preserves the exact shift and works for all cases. Rounding or
rejecting seems a bit limiting or lossy.

srikanthccv:
discussed offline, @tushar-signoz to research the product experience side once.

PR #10763, opened 2026-03-30
fix: timestamp shift

### LINKED ISSUE #9653: Time shift + formula doesn't work when the aggregation interval > timeshift.
> Using the query builder with timeshift limits us to 1 day worth of data. We have a panel that looks like this. It works fine when requesting data for < 1 day. However as soon as it goes >= 1 day, the chart appears as it does in the second screenshot. Is there anything we can do here?

Related to https://github.com/SigNoz/signoz/issues/8796

tushar-signoz:
### Summary
Fix timestamp misalignment in time-shifted ("compare to previous period") queries by aligning ClickHouse bucket boundaries to the shift offset. Without this, when shift % step != 0, the shifted series' buckets land at different positions than the original series, making point-to-point comparison incorrect.

Introduce querybuilder.TimeIntervalExpr which generates toStartOfInterval(col + shift, step) - shift when a shift is present, ensuring bucket boundaries in the shifted series map exactly onto the original series' boundaries after the app-level forward-shift.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---


ankitnayan:
Does this PR fix both of below? Please maintain a habit of mentioning issues in the PR which is solved
1. https://github.com/SigNoz/signoz/issues/9653
2. https://github.com/SigNoz/signoz/issues/8796

tushar-signoz:
> Does this PR fix both of below? Please maintain a habit of mentioning issues in the PR which is solved
> 
> 1. [Time shift + formula doesn't work when the aggregation interval > timeshift. #9653](https://github.com/SigNoz/signoz/issues/9653)
> 2. [Return warning if the timestamps of the values of the queries in formula are not aligned #8796](https://github.com/SigNoz/signoz/issues/8796)

Yes, if agreed upon, fixes both
Need sign-off from @srikanthccv for this https://www.notion.so/signoz/Timeshift-341fcc6bcd19808eaae6d99872c66918

ankitnayan:
try this out and let's reason if the results make sense
Screenshot 2026-05-06 at 2 51 45 PM

tushar-signoz:
Hey @ankitnayan, using your suggested query, A -> step interval 7200 , B -> step interval 7200, shift: 3600, F -> A-B

Since this is fake data, the metric aggregate value remains same more or less. 

[Pre fix] Because of the timestamp mis alignment, data points generate are 100, -100, 100, -100 
Screenshot 2026-05-06 at 3 35 37 PM

[Post fix] Because of the timestamp align correctly, data points generate are (100 - 100) , (100 - 100) , (100 -100)
Screenshot 2026-05-06 at 3 35 13 PM

ankitnayan:
Screenshot 2026-05-06 at 6 01 57 PM

let's forget the formula for once. The above query does not produce a correct timeshifted chart

Before the fix
Screenshot 2026-05-06 at 6 02 57 PM


After the fix
Screenshot 2026-05-06 at 6 03 19 PM

ankitnayan:
It makes sense to restrict timeshift to a multiple of the aggregation interval.

ankitnayan:
Current state: If QueryA and QueryB have different aggregation intervals, formula/charting causes the queries to have values at all the possible timestamps for both the queries which is also a bit weird as the real datapoints exist only at LCM of intervals but Formula uses GCD of intervals. Rest is charting or interpolation issue

srikanthccv:
> formula/charting causes the queries to have values at all the possible timestamps for both the queries which is also a bit weird as the real datapoints exist only at LCM of intervals but Formula uses GCD of intervals

I guess you came to this conclusion based on the data shown. We don't use GCD for formula, in fact, we don't use any intervals for formula calculation at the moment. And in one specific function (fill zero for formula) where we use intervals, we use LCM of intervals so there is no GCD of intervals in formula evaluation https://github.com/SigNoz/signoz/blob/cf3fbb06c62698e0b00bcf9a1ffd8b908a9e6545/pkg/types/querybuildertypes/querybuildertypesv5/req.go#L342-L357

current behaviour in system.

- If you use non-count queries you would notice that points only exist at the LCM of intervals
  - for example, use `system.memory.usage`, or any gauge metric
  - use anything other than `count`, count_distinct`, `rate` for logs/traces
- what do we do for count like queries in formula
  - for count like queries, when the value is missing for we implicitly assume that missing value could be assumed to be zero for the expression evaluation. Why, otherwise, the common cases like error rate don't produce any data. so evaluate it like binary arithmetic instead of being aware of the step intervals. Example; Query A - 300s step, Query B - 120s step, When we take the formula `A/B`, for each timestamp, we check if A and B value are present, if they don't we implicitly assume 0 for count queries and evaluate. This leads to points at intervals other than LCM. In sql model, this is `FULL OUTER JOIN` with with `COALESCE(..., 0)` and no-step awareness.


So, as of now, given a certain query intervals which have asymmetry, the output series cadence becomes data-dependent and non-uniform. Should the behaviour be fixed for counter-like formula with asymmetry? seems yes; the can this no value be interpreted as zero is independent of the decision of union of intervals vs LCM list.

PR #10766, opened 2026-03-30
feat(audit): add Auditor interface and rename auditortypes to audittypes

therealpandey:
## Summary
- Add `Auditor` interface in `pkg/auditor/auditor.go` with fire-and-forget `Audit(ctx, event)` method
- Add `Config` in `pkg/auditor/config.go` with provider pattern (buffer size, batch size, flush interval) and validation
- Add `OTLPHTTPConfig` provider-specific config with full OTLP HTTP options (endpoint, url_path, insecure, compression, timeout, headers, retry)
- Add `pkg/auditor/auditorserver/` — buffered event server with `ExportFunc` callback, batch-size and interval flushing, fail-open drop, drain on stop
- Add OTel metrics via MeterProvider: `signoz.audit.events.emitted`, `signoz.audit.store.write_errors`, `signoz.audit.events.dropped`, `signoz.audit.events.buffer_size`
- Add OTel tracing on Add (event_name, dropped attributes) and Export (batch_size attribute) paths
- Implement `factory.ServiceWithHealthy` on Server with healthyC for lifecycle coordination
- Add 8 unit tests with race detection covering construction, lifecycle, batch/interval flush, drop, drain, export failure, and concurrency
- Rename `pkg/types/auditortypes/` to `pkg/types/audittypes/`

closes SigNoz/platform-pod#1930

## Test plan
- [x] `go build ./pkg/auditor/...` compiles
- [x] `go vet ./pkg/auditor/...` passes
- [x] `go test -race ./pkg/auditor/auditorserver/...` — 8 tests pass, 15/15 runs with no flakes
- [x] `make -f $PRIMUS_HOME/src/make/main.mk go-lint` — 0 issues

### DIFF
--- a/pkg/auditor/auditor.go
+++ b/pkg/auditor/auditor.go
@@ -6,11 +6,7 @@ import (
 	"github.com/SigNoz/signoz/pkg/types/audittypes"
 )
 
-// Auditor is the contract between the HTTP handler layer and the audit
-// implementation. Community uses a noop; enterprise uses a buffered
-// OTLP HTTP exporter gated by licensing.
-//
-// Audit is fire-and-forget: callers never block on audit outcomes.
 type Auditor interface {
+	// Audit emits an audit event. It is fire-and-forget: callers never block on audit outcomes.
 	Audit(ctx context.Context, event audittypes.AuditEvent)
 }
--- a/pkg/auditor/config.go
+++ b/pkg/auditor/config.go
@@ -91,16 +91,16 @@ func NewConfigFactory() factory.ConfigFactory {
 
 func (c Config) Validate() error {
 	if c.BufferSize <= 0 {
-		return errors.New(errors.TypeInvalidInput, errors.CodeInvalidInput, "auditor buffer_size must be greater than 0")
+		return errors.New(errors.TypeInvalidInput, errors.CodeInvalidInput, "auditor::buffer_size must be greater than 0")
 	}
 	if c.BatchSize <= 0 {
-		return errors.New(errors.TypeInvalidInput, errors.CodeInvalidInput, "auditor batch_size must be greater than 0")
+		return errors.New(errors.TypeInvalidInput, errors.CodeInvalidInput, "auditor::batch_size must be greater than 0")
 	}
 	if c.FlushInterval <= 0 {
-		return errors.New(errors.TypeInvalidInput, errors.CodeInvalidInput, "auditor flush_interval must be greater than 0")
+		return errors.New(errors.TypeInvalidInput, errors.CodeInvalidInput, "auditor::flush_interval must be greater than 0")
 	}
 	if c.BatchSize > c.BufferSize {
-		return errors.New(errors.TypeInvalidInput, errors.CodeInvalidInput, "auditor batch_size must not exceed buffer_size")
+		return errors.New(errors.TypeInvalidInput, errors.CodeInvalidInput, "auditor::batch_size must not exceed auditor::buffer_size")
 	}
 	return nil
 }


pkg/auditor/config.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+var _ factory.Config = (*Config)(nil)
+
+type Config struct {
+	Provider      string        `mapstructure:"provider"`
+	Endpoint      string        `mapstructure:"endpoint"`
+	BufferSize    int           `mapstructure:"buffer_size"`
+	BatchSize     int           `mapstructure:"batch_size"`
+	FlushInterval time.Duration `mapstructure:"flush_interval"`
```
therealpandey:
```suggestion
type OTLPHTTP Provider {
	Endpoint      string        `mapstructure:"endpoint"`
}
```

All of this needs to be in the `otlphttp` provider. Can you look at the otelgo-sdk and see what all can be provided as config options here?

therealpandey:
```suggestion
// Common across providers
	BufferSize    int           `mapstructure:"buffer_size"`
	BatchSize     int           `mapstructure:"batch_size"`
	FlushInterval time.Duration `mapstructure:"flush_interval"`
```

These 3 settings will be common across providers so they can stay at the top level. Document the meaning of all 3 settings.

therealpandey: Done — moved `Endpoint` into `OTLPHTTPConfig` provider struct with full OTLP HTTP options from the OTel Go SDK: `URLPath`, `Insecure`, `Compression`, `Timeout`, `Headers`, and `RetryConfig` (with `Enabled`, `InitialInterval`, `MaxInterval`, `MaxElapsedTime`). Defaults match the SDK defaults.

therealpandey: Done — kept `BufferSize`, `BatchSize`, `FlushInterval` at top-level `Config`. Added doc comments for all three explaining purpose and default values.

pkg/auditor/config.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+}
+
+func NewConfigFactory() factory.ConfigFactory {
+	return factory.NewConfigFactory(factory.MustNewName("auditor"), newConfig)
+}
+
+func (c Config) Validate() error {
+	if c.BufferSize <= 0 {
+		return errors.New(errors.TypeInvalidInput, errors.CodeInvalidInput, "auditor buffer_size must be greater than 0")
```
therealpandey:
```suggestion
		return errors.New(errors.TypeInvalidInput, errors.CodeInvalidInput, "auditor::buffer_size must be greater than 0")
```

Use this pattern in all validation errors.

pkg/auditor/auditor.go
```diff
@@ -0,0 +1,16 @@
...
+	"context"
+
+	"github.com/SigNoz/signoz/pkg/types/audittypes"
+)
+
+// Auditor is the contract between the HTTP handler layer and the audit
+// implementation. Community uses a noop; enterprise uses a buffered
+// OTLP HTTP exporter gated by licensing.
+//
+// Audit is fire-and-forget: callers never block on audit outcomes.
```
therealpandey:
```suggestion
```

Move this comment to just above the Audit method

Also there is no need to write a comment on top of the Auditor interface.

pkg/auditor/config.go
therealpandey: Done — all 4 validation errors now use `auditor::` prefix pattern.

pkg/auditor/auditor.go
therealpandey: Done — moved fire-and-forget comment to the `Audit` method, removed interface-level comment.

pkg/auditor/auditorserver/config.go
```diff
@@ -0,0 +1,17 @@
+package auditorserver
+
+import "time"
+
+type Config struct {
+	// BufferSize is the maximum number of events that can be buffered.
+	// When full, new events are dropped (fail-open).
+	BufferSize int
```
vikrantgupta25, line 8: batch processor :p

pkg/auditor/auditorserver/server.go
```diff
@@ -0,0 +1,211 @@
...
+// Add enqueues an audit event for batched export.
+// If the buffer is full the event is dropped and a warning is logged.
+func (server *Server) Add(ctx context.Context, event audittypes.AuditEvent) {
+	ctx, span := server.settings.Tracer().Start(ctx, "auditorserver.Add", trace.WithAttributes(attribute.String("audit.event_name", event.EventName.String())))
+	defer span.End()
+
+	server.queueMtx.Lock()
+	defer server.queueMtx.Unlock()
+
+	if len(server.queue) >= server.config.BufferSize {
```
vikrantgupta25, line 127: why not trigger a force export here when adding another either exceeds the buffer size or buffer count ?

therealpandey: Because the reason is queue is blocked is because the exporter is slow. ForceExport will exacerbate the exporter's problems.

vikrantgupta25: size limits can be reached before count limits as well

vikrantgupta25: the idea being closer to just flush whatever reaches first and forget

PR #10769, opened 2026-03-30
feat(audit): add noop auditor for community edition

therealpandey:
## Summary
- Embed `factory.ServiceWithHealthy` in the `auditor.Auditor` interface so all providers share a uniform lifecycle contract (`Start`/`Stop`/`Healthy`)
- Add `pkg/auditor/noopauditor/` — the community edition auditor that silently discards all events with zero allocations
- Unit tests covering interface compliance, zero-value and populated event handling, Start/Stop lifecycle, and factory registration

## Test plan
- [x] `go build -mod=mod ./pkg/auditor/...` compiles
- [x] `go test -race -mod=mod ./pkg/auditor/...` all tests pass (noopauditor + auditorserver)
- [x] `go vet -mod=mod ./pkg/auditor/...` no issues

closes SigNoz/platform-pod#1931


pkg/auditor/noopauditor/provider_test.go
```diff
@@ -0,0 +1,78 @@
+package noopauditor
```
therealpandey: There is no need for this test file.

therealpandey: Removed in eafcbdbcc.

PR #10779, opened 2026-03-31
feat(/fields/*) : introduce a new param `metricNamespace` in the APIs for prefix match 

nikhilmantri0902:
## Pull Request

A new param `metricNamespace` is added to /fields/keys and /fields/values APIs. This param enforces a prefix match over the metricNames with the provided prefix as its value. So if metricNamespace=k8s.pod. is passed then the keys will be suggested from across the metric_names with k8s.pod. as their prefix. Same is true for the values API also. 

In infra-monitoring, we identified that there are usecases where the frontend needs this API to show a list of attributes belonging to a family of metric names ( k8s.pod.*, system.cpu.*, etc. ). Today the `/fields/keys` and `/fields/values` API only allows specifying one metricName. With this change, you can specify a prefix which will cause lookup to be over all the metricNames with this prefix. 

We considered introducing a new field to accept a list of metricnames but extending the metricName covered the same usecase in a more expressive way for metric naming patterns which themselves have meaning. For examples all kubernetes related metric names start with k8s., system metrics start with system. etc. Thus adding support for metricNamespace to do prefix match proves useful.

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/4383

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

We have now replaced equality operator with Like operator

---


pkg/telemetrymetadata/metadata.go
```diff
@@ -651,7 +651,12 @@ func (t *telemetryMetaStore) getMetricsKeys(ctx context.Context, fieldKeySelecto
 		// }
 
 		if fieldKeySelector.MetricContext != nil {
-			fieldConds = append(fieldConds, sb.E("metric_name", fieldKeySelector.MetricContext.MetricName))
+			if fieldKeySelector.MetricContext.MetricName != "" {
+				fieldConds = append(fieldConds, sb.E("metric_name", fieldKeySelector.MetricContext.MetricName))
+			}
+			if fieldKeySelector.MetricContext.MetricNamespace != "" {
```
namanv3, line 895: confirming if this is the intended api behaviour, if someone sends both a metric name and namespace in the query, we will AND them both, correct?

pkg/telemetrymetadata/metadata.go
```diff
@@ -1383,7 +1393,7 @@ func (t *telemetryMetaStore) getMetricFieldValues(ctx context.Context, fieldValu
 		sb.Where(sb.E("attr_datatype", fieldValueSelector.FieldDataType.TagDataType()))
 	}
 
-	if fieldValueSelector.MetricContext != nil {
+	if fieldValueSelector.MetricContext != nil && fieldValueSelector.MetricContext.MetricName != "" {
```
namanv3, line 1741: no namespace check needed here?

nikhilmantri0902: Yup Because we are adding `where equal to metric_name` on the metric_name here. Earlier because metricContext contained only metricName hence nil check was fine ...

pkg/telemetrymetadata/metadata.go
nikhilmantri0902: Yes,

pkg/telemetrymetadata/metadata.go
srikanthccv: Why does the keys have metric namespace filter check but not the values?

Note: as a general rule of thumb, do not resolve the comments by yourself, let the reviewer resolve if their question is answered.

nikhilmantri0902: Sure. @srikanthccv @namanv3 I am adding the support in /keys/values API as well for metricNamespace field supporting prefix match. 
Also, will take care not to resolve comments.

pkg/telemetrymetadata/metadata_test.go
```diff
@@ -386,3 +400,106 @@ func TestGetMetricFieldValuesIntrinsicBoolReturnsEmpty(t *testing.T) {
...
+	mock := mockTelemetryStore.Mock()
+
+	metadata := newTestTelemetryMetaStoreTestHelper(mockTelemetryStore)
+
+	valueRows := cmock.NewRows([]cmock.ColumnType{
+		{Name: "metric_name", Type: "String"},
+	}, [][]any{{"system.cpu.utilization"}})
+
+	mock.ExpectQuery(regexp.QuoteMeta("SELECT metric_name FROM signoz_metrics.distributed_time_series_v4_1week WHERE metric_name LIKE ? GROUP BY metric_name LIMIT ?")).
+		WithArgs("system.cpu%", 51).
```
namanv3, line 453: why 51 is the limit when request is for 50?

pkg/telemetrymetadata/metadata.go
namanv3: namespace check added, resolving this

pkg/telemetrymetadata/metadata_test.go
nikhilmantri0902: The limit + 1 pattern is used throughout the metadata store to detect whether results are truncated. We fetch one extra row beyond the requested limit — if it comes back, we know there are more results and set complete = false in the response. The extra row is never returned to the caller. So 51 = the requested limit of 50 + 1 for the completeness check.

namanv3 (approved):

srikanthccv (approved):

vikrantgupta25 (approved):

PR #10780, opened 2026-03-31
feat(global-time-adapter): start migration away from redux

H4ad:
## Pull Request

The code started at #10701 but then splitted to this PR.

---

This PR introduces a adapter for the globalTime in the Redux store.

The main problem we have today with min/max time is it being used by the query keys of React Query, this gives us the ability to refresh + single time for all queries but causes all the queries to loose identity when refreshing.

This PR converts the time duration to an interval when make sense, or time range when specific time range is requested.

The react query now will have:

- 1d/1m/etc...
- Date||-||Date

I covered this issue in more details at https://www.notion.so/Min-Max-on-React-Query-Key-32dfcc6bcd1980caa6b5fd92b7a0b797

You can read the #10701 to learn how to adopt, but essentially, you do something like this:

```tsx
const selectedTime = useGlobalTimeStore((store) => store.selectedTime);
const isRefreshEnabled = useGlobalTimeStore((s) => s.isRefreshEnabled);
const refreshInterval = useGlobalTimeStore((s) => s.refreshInterval);
const getMinMaxTime = useGlobalTimeStore((s) => s.getMinMaxTime);

const queryKey = useMemo(
  () => [
  REACT_QUERY_KEY.AUTO_REFRESH_QUERY,
  // your query keys
  selectedTime,
  ], []
);

const { data, isFetching, isLoading, isError } = useQuery({
  queryKey,
  queryFn: ({ signal }) => {
    const { minTime, maxTime } = getMinMaxTime();
    // your query
  },
  refetchInterval: isRefreshEnabled ? refreshInterval : false,
});
```

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered:

---

- Blast radius: None
- Potential regressions: None

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/container/TopNav/DateTimeSelectionV2/index.tsx
+++ b/frontend/src/container/TopNav/DateTimeSelectionV2/index.tsx
@@ -14,6 +14,10 @@ import { QueryParams } from 'constants/query';
 import ROUTES from 'constants/routes';
 import NewExplorerCTA from 'container/NewExplorerCTA';
 import dayjs, { Dayjs } from 'dayjs';
+import {
+	useGlobalTimeQueryInvalidate,
+	useIsGlobalTimeQueryRefreshing,
+} from 'hooks/globalTime';
 import { useQueryBuilder } from 'hooks/queryBuilder/useQueryBuilder';
 import { useSafeNavigate } from 'hooks/useSafeNavigate';
 import useUrlQuery from 'hooks/useUrlQuery';
@@ -26,10 +30,6 @@ import { useTimezone } from 'providers/Timezone';
 import { bindActionCreators, Dispatch } from 'redux';
 import { ThunkDispatch } from 'redux-thunk';
 import { GlobalTimeLoading, UpdateTimeInterval } from 'store/actions';
-import {
-	useGlobalTimeQueryInvalidate,
-	useIsGlobalTimeQueryRefreshing,
-} from 'store/globalTime/hooks';
 import { AppState } from 'store/reducers';
 import AppActions from 'types/actions';
 import { GlobalReducer } from 'types/reducer/globalTime';
--- a/frontend/src/hooks/globalTime/useGlobalTimeQueryInvalidate.ts
+++ b/frontend/src/hooks/globalTime/useGlobalTimeQueryInvalidate.ts
@@ -1,17 +1,6 @@
 import { useCallback } from 'react';
-import { useIsFetching, useQueryClient } from 'react-query';
+import { useQueryClient } from 'react-query';
 import { REACT_QUERY_KEY } from 'constants/reactQueryKeys';
-
-/**
- * Use when you want to know if any query tracked by {@link REACT_QUERY_KEY.AUTO_REFRESH_QUERY} is refreshing
- */
-export function useIsGlobalTimeQueryRefreshing(): boolean {
-	return (
-		useIsFetching({
-			queryKey: [REACT_QUERY_KEY.AUTO_REFRESH_QUERY],
-		}) > 0
-	);
-}
 
 /**
  * Use when you want to invalida any query tracked by {@link REACT_QUERY_KEY.AUTO_REFRESH_QUERY}
--- a/frontend/src/hooks/globalTime/index.ts
+++ b/frontend/src/hooks/globalTime/index.ts
@@ -1,0 +1,2 @@
+export { useGlobalTimeQueryInvalidate } from './useGlobalTimeQueryInvalidate';
+export { useIsGlobalTimeQueryRefreshing } from './useIsGlobalTimeQueryRefreshing';
--- a/frontend/src/hooks/globalTime/useIsGlobalTimeQueryRefreshing.ts
+++ b/frontend/src/hooks/globalTime/useIsGlobalTimeQueryRefreshing.ts
@@ -1,0 +1,13 @@
+import { useIsFetching } from 'react-query';
+import { REACT_QUERY_KEY } from 'constants/reactQueryKeys';
+
+/**
+ * Use when you want to know if any query tracked by {@link REACT_QUERY_KEY.AUTO_REFRESH_QUERY} is refreshing
+ */
+export function useIsGlobalTimeQueryRefreshing(): boolean {
+	return (
+		useIsFetching({
+			queryKey: [REACT_QUERY_KEY.AUTO_REFRESH_QUERY],
+		}) > 0
+	);
+}


frontend/src/store/globalTime/types.ts
```diff
@@ -0,0 +1,43 @@
+export interface IGlobalTimeStoreState {
+	/**
+	 * The selected time range, can be:
+	 * - Relative duration: '1m', '5m', '15m', '1h', '1d', etc.
+	 * - Custom range: '<minTimeUnixNano>||_||<maxTimeUnixNano>' format
+	 */
+	selectedTime: string;
```
AshwinBhatkal: we can type this differently

type RelativeTime =
  | '1m'
  | '5m'
  | '15m'
  | '1h'
  | '6h'
  | '1d';
I believe the above already exists in some form

type CustomRange = `${number}${CUSTOM_TIME_SEPARATOR}${number}`;

selectedTime: RelativeTime | CustomRange

frontend/src/store/globalTime/types.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export interface IGlobalTimeStoreActions {
+	/**
+	 * Set the selected time and optionally the refresh interval.
+	 * isRefreshEnabled is automatically computed:
+	 * - Custom time ranges: always false
+	 * - Duration times with refreshInterval > 0: true
+	 * - Duration times with refreshInterval = 0: false
+	 */
+	setSelectedTime: (selectedTime: string, refreshInterval?: number) => void;
```
AshwinBhatkal: same type change here

frontend/src/store/globalTime/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+	const maxTime = parseInt(maxStr, 10);
+
+	if (Number.isNaN(minTime) || Number.isNaN(maxTime)) {
+		return null;
+	}
+
+	return { minTime, maxTime };
+}
+
+const fallbackDuration = 30 * 1000 * 1000000; // 30s
```
AshwinBhatkal: nit: fallbackDurationInNanoSeconds

frontend/src/store/globalTime/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+ * For custom times, extracts the stored min/max values.
+ */
+export function parseSelectedTime(selectedTime: string): ParsedTimeRange {
+	if (isCustomTimeRange(selectedTime)) {
+		const parsed = parseCustomTimeRange(selectedTime);
+		if (parsed) {
+			return parsed;
+		}
+		// Fallback to current time if parsing fails
+		const now = Date.now() * 1000000;
```
AshwinBhatkal: can you make this 1000000 a constant and reuse it? Maybe NANO_SECOND_MULTIPLIER We can avoid the rename to nano seconds

frontend/src/store/globalTime/globalTimeStore.ts
```diff
@@ -0,0 +1,32 @@
+import { create } from 'zustand';
+
+import { IGlobalTimeStoreActions, IGlobalTimeStoreState } from './types';
+import { isCustomTimeRange, parseSelectedTime } from './utils';
+
+export type IGlobalTimeStore = IGlobalTimeStoreState & IGlobalTimeStoreActions;
+
+export const useGlobalTimeStore = create<IGlobalTimeStore>((set, get) => ({
+	// Initial state
+	selectedTime: '30s',
```
AshwinBhatkal: Isn't the default 30m?

frontend/src/store/globalTime/globalTimeStore.ts
```diff
@@ -0,0 +1,32 @@
...
+
+			return {
+				selectedTime,
+				refreshInterval: newRefreshInterval,
+				isRefreshEnabled: !isCustom && newRefreshInterval > 0,
+			};
+		});
+	},
+
+	getMinMaxTime: (): { minTime: number; maxTime: number } => {
```
AshwinBhatkal: return type is parsedTimeRange

frontend/src/store/globalTime/__tests__/utils.test.ts
```diff
@@ -0,0 +1,138 @@
...
+			const fallbackDuration = 30 * 1000 * 1000000; // 30s in nanoseconds
+			expect(result.maxTime).toBe(now);
+			expect(result.minTime).toBe(now - fallbackDuration);
+		});
+
+		it('should parse relative time strings using getMinMaxForSelectedTime', () => {
+			const result = parseSelectedTime('15m');
+			const now = Date.now() * 1000000;
+			// 15 minutes in nanoseconds
+			const fifteenMinutesNs = 15 * 60 * 1000 * 1000000;
```
AshwinBhatkal: we could use the nano second multiplier here

AshwinBhatkal: at all places in the test

frontend/src/store/globalTime/__tests__/globalTimeStore.test.ts
```diff
@@ -0,0 +1,193 @@
...
+
+			// Advance time by 1 second
+			act(() => {
+				jest.advanceTimersByTime(1000);
+			});
+
+			const second = result.current.getMinMaxTime();
+
+			// maxTime should be different (1 second later)
+			expect(second.maxTime).toBe(first.maxTime + 1000 * 1000000);
```
AshwinBhatkal: nano second multiplier

AshwinBhatkal: across this test as well

frontend/src/container/TopNav/DateTimeSelectionV2/index.tsx
```diff
@@ -352,7 +354,14 @@ function DateTimeSelection({
 		],
 	);
 
+	const queryClient = useQueryClient();
+	const isRefreshingQueries = useIsFetching({
```
AshwinBhatkal: could be made a blind hook

AshwinBhatkal: we could add lint rule maybe

frontend/src/container/TopNav/DateTimeSelectionV2/index.tsx
```diff
@@ -732,7 +741,13 @@ function DateTimeSelection({
 					{showAutoRefresh && selectedTime !== 'custom' && (
 						<div className="refresh-actions">
 							<FormItem hidden={refreshButtonHidden} className="refresh-btn">
-								<Button icon={<SyncOutlined />} onClick={onRefreshHandler} />
+								<Button
+									icon={<SyncOutlined />}
+									loading={!!isRefreshingQueries}
+									onClick={(): void => {
```
AshwinBhatkal: nit: use function directly

frontend/src/components/GlobalTimeStoreAdapter/__tests__/GlobalTimeStoreAdapter.test.tsx
```diff
@@ -0,0 +1,227 @@
...
+		const { result } = renderHook(() => useGlobalTimeStore());
+		expect(result.current.selectedTime).toBe('15m');
+		expect(result.current.refreshInterval).toBe(5000); // 5s = 5000ms
+		expect(result.current.isRefreshEnabled).toBe(true);
+	});
+
+	it('should set refreshInterval to 0 when auto refresh is disabled', () => {
+		store = mockStore({
+			globalTime: createGlobalTimeState({
+				selectedTime: '15m',
```
AshwinBhatkal, line 122: use consts for numbers in these files. I believe they are already defined

frontend/src/container/TopNav/DateTimeSelectionV2/index.tsx
H4ad: Added helper hooks only

frontend/src/store/globalTime/globalTimeStore.ts
H4ad: For alerts, it was default to 30s, but updated to use our global const of 30m

frontend/src/store/globalTime/types.ts
H4ad: Used the `Time` from date time selector.

frontend/src/store/globalTime/types.ts
```diff
@@ -0,0 +1,52 @@
+import { Time } from 'container/TopNav/DateTimeSelectionV2/types';
+
+export type CustomTimeRangeSeparator = '||_||';
+export type CustomTimeRange = `${number}${CustomTimeRangeSeparator}${number}`;
+export type GlobalTimeSelectedTime = Time | CustomTimeRange;
```
AshwinBhatkal, line 5: So we will have 3 different types of time type? Will we phase something out? Let's mark them deprecated now itself, if so

frontend/src/store/globalTime/__tests__/globalTimeStore.test.ts
```diff
@@ -0,0 +1,204 @@
...
+			});
+
+			expect(result.current.refreshInterval).toBe(5000);
+		});
+
+		it('should enable refresh for relative time with refreshInterval > 0', () => {
+			const { result } = renderHook(() => useGlobalTimeStore());
+
+			act(() => {
+				result.current.setSelectedTime('15m', 5000);
```
AshwinBhatkal, line 72: would these values be flagged by ts if not for 15m or 30m or in those lines

frontend/src/store/globalTime/__tests__/utils.test.ts
AshwinBhatkal: I don't see it used here. Can you verify closely other places as well once?

frontend/src/hooks/globalTime/useGlobalTimeQueryInvalidate.ts
```diff
@@ -0,0 +1,27 @@
+import { useCallback } from 'react';
```
AshwinBhatkal, line 1: There is a folder called hooks that has all hooks in it. No opinion, but that is another place to store it

frontend/src/store/globalTime/__tests__/globalTimeStore.test.ts
H4ad: Yes, since we are using `Time`, it way more strict.

frontend/src/hooks/globalTime/useGlobalTimeQueryInvalidate.ts
H4ad: I prefer to have code that is similar into same place, is hard to navigate between a codebase where everything it so far from the same context.

frontend/src/store/globalTime/types.ts
H4ad: We have `Time`, `CustomTImeRange` and `CustomTimeType`.

I didn't want to deprecate something at `DateTimeSelectionV2` because I don't know for now what types I will keep or remove due to backwards compatibility.

frontend/src/hooks/globalTime/useGlobalTimeQueryInvalidate.ts
H4ad: Moved to hooks to keep consistency with the codebase.

frontend/src/store/globalTime/types.ts
AshwinBhatkal: Cool. But that would be the plan, correct?

I'm just itchy of Time | CustomTimeRange

H4ad: Yes

H4ad: The DateTimeSelection we need to do two refactors, one to remove redux and then use zustand, and another to remove react-router in favor of nuqs.

PR #10794, opened 2026-04-01
refactor: setup types and interface for waterfall v3

krsoninikhil:
## Pull Request

---

v3 is required for udpating the response json of
the waterfall api. There wont' be any logical change. Using this requirement as an opportunity to move
waterfall api to provider codebase architecture from older query-service

Update: follow up PR https://github.com/SigNoz/signoz/pull/10797

https://github.com/SigNoz/engineering-pod/issues/4204

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---


pkg/types/tracedetailtypes/waterfall.go
```diff
@@ -0,0 +1,246 @@
...
+// Event represents a span event.
+type Event struct {
+	Name         string                 `json:"name,omitempty"`
+	TimeUnixNano uint64                 `json:"timeUnixNano,omitempty"`
+	AttributeMap map[string]interface{} `json:"attributeMap,omitempty"`
+	IsError      bool                   `json:"isError,omitempty"`
+}
+
+// Span is the v3 span format with separated attributes/resources and snake_case keys.
+type Span struct {
```
nityanandagohain: Add a comment that, this is an exception where we have to use snake_case as the query is directly dependent, and the response represent the keys present that can be use as is in query 

Also a note that not to copy this format anywhere else, specially for AI.

pkg/types/tracedetailtypes/waterfall.go
```diff
@@ -0,0 +1,246 @@
...
+	StatusCode         int32             `json:"status_code"`
+	StatusCodeString   string            `json:"status_code_string"`
+	StatusMessage      string            `json:"status_message"`
+	Timestamp          string            `json:"timestamp"`
+	TraceID            string            `json:"trace_id"`
+	TraceState         string            `json:"trace_state"`
+	// Tree structure fields
+	Children         []*Span `json:"children"`
+	SubTreeNodeCount uint64  `json:"sub_tree_node_count"`
+	HasChildren      bool    `json:"has_children"`
```
nityanandagohain: @vikrantgupta25  any suggestion here, these are api fields and doesn't represent data so this should follow the camel case right ?

Also I think we can create multiple structure and then encapsulate them with new structure for this kind of case as we are repeating again in SpanModel

krsoninikhil: >  encapsulate them with new structure for this kind of case as we are repeating again in SpanModel

Fields are repeating but using same struct for json and db field representation might not be a good idea. For example, if a new field is added in db, it should not automatically reach json response which can be change the contract. Let me know if you want to discuss it.

pkg/types/tracedetailtypes/waterfall.go
krsoninikhil: Fixed in [ed5c300](https://github.com/SigNoz/signoz/pull/10794/commits/ed5c30012d797655654a35a365a8e8ffc4dff9b4)

pkg/types/tracedetailtypes/waterfall.go
```diff
@@ -0,0 +1,249 @@
...
+	AttributeMap map[string]any `json:"attributeMap,omitempty"`
+	IsError      bool           `json:"isError,omitempty"`
+}
+
+// Span represents the span in waterfall response,
+// this uses snake_case keys for response as a special case since these
+// keys can be directly used to query spans and client need to know the actual fields.
+// This pattern should not be copied elsewhere.
+type Span struct {
+	Attributes         map[string]string `json:"attributes"`
```
nityanandagohain: This should be map[string]any

pkg/types/tracedetailtypes/waterfall.go
nityanandagohain: >new field is added in db, it should not automatically reach json response 
Yeah correct in that sense.

let vikrant reply on the other part, added one more comment, apart from that looks good.

pkg/types/tracedetailtypes/waterfall.go
krsoninikhil: Wouldn't all types of attributes be converted to string in the response? For bool and numbers will remain as bool and number type in json response?

nityanandagohain: >bool and numbers will remain as bool and number type in json response
yes

In future will we support true map[string]any as otel where will support maps ans slices

krsoninikhil: Updated in [4e9b3f3](https://github.com/SigNoz/signoz/pull/10794/commits/4e9b3f3b0fc38d761749a6b1e130378fd1ed5dbf)

pkg/types/tracedetailtypes/waterfall.go
```diff
@@ -0,0 +1,249 @@
...
+	StatusCodeString   string            `json:"status_code_string"`
+	StatusMessage      string            `json:"status_message"`
+	Timestamp          string            `json:"timestamp"`
+	TraceID            string            `json:"trace_id"`
+	TraceState         string            `json:"trace_state"`
+	// Tree structure fields
+	Children         []*Span `json:"children"`
+	SubTreeNodeCount uint64  `json:"sub_tree_node_count"`
+	HasChildren      bool    `json:"has_children"`
+	HasSiblings      bool    `json:"has_siblings"`
```
tushar-signoz: Is there a better name possible for this `HasSiblings`, since afaik it means does it have siblings on its right ? Not sure if this can be changed now or not.
Suggestion: `HasSiblingsAfter` or `IsNotLastSibling`

pkg/types/tracedetailtypes/waterfall.go
```diff
@@ -0,0 +1,249 @@
...
+	"time"
+
+	"github.com/SigNoz/signoz/pkg/types/cachetypes"
+)
+
+// WaterfallRequest is the request body for the v3 waterfall API.
+type WaterfallRequest struct {
+	SelectedSpanID              string   `json:"selectedSpanId"`
+	IsSelectedSpanIDUnCollapsed bool     `json:"isSelectedSpanIDUnCollapsed"`
+	UncollapsedSpans            []string `json:"uncollapsedSpans"`
```
tushar-signoz: Does `IsSelectedSpanIDUnCollapsed` have a special meaning ? Wouldn't adding `SelectedSpanID` in `UncollapsedSpans` list convey the same meaning ?

pkg/types/tracedetailtypes/waterfall.go
```diff
@@ -0,0 +1,249 @@
...
+		HTTPHost:           item.HTTPHost,
+		HTTPMethod:         item.HTTPMethod,
+		HTTPURL:            item.HTTPURL,
+		IsRemote:           item.IsRemote,
+		Kind:               int32(item.Kind),
+		KindString:         item.SpanKind,
+		Links:              item.References,
+		Name:               item.Name,
+		ParentSpanID:       item.ParentSpanID,
+		Resources:          resources,
```
tushar-signoz: Take an informed decision on naming it `Resource` or `Resources`. By otel standards it's `Resource`. 
Edit: Wont be applicable if we name the stuct something else

pkg/types/tracedetailtypes/waterfall.go
```diff
@@ -0,0 +1,249 @@
...
+	TimeUnixNano uint64         `json:"timeUnixNano,omitempty"`
+	AttributeMap map[string]any `json:"attributeMap,omitempty"`
+	IsError      bool           `json:"isError,omitempty"`
+}
+
+// Span represents the span in waterfall response,
+// this uses snake_case keys for response as a special case since these
+// keys can be directly used to query spans and client need to know the actual fields.
+// This pattern should not be copied elsewhere.
+type Span struct {
```
tushar-signoz: Not very keen on naming it Span as it doesn't represent otel span. and is a derived object for the waterfall response api so better name it `WaterfallResponseSpanObject`, `WaterfallSpan`, `WaterfallNode`

Lets not use keywords  which have a otel specific meaning in general as name for structs.

pkg/types/tracedetailtypes/waterfall.go
```diff
@@ -0,0 +1,249 @@
...
+	StatusCode         int32              `ch:"status_code"`
+	DBName             string             `ch:"db_name"`
+	DBOperation        string             `ch:"db_operation"`
+	HTTPMethod         string             `ch:"http_method"`
+	HTTPURL            string             `ch:"http_url"`
+	HTTPHost           string             `ch:"http_host"`
+	ExternalHTTPMethod string             `ch:"external_http_method"`
+	ExternalHTTPURL    string             `ch:"external_http_url"`
+	ResponseStatusCode string             `ch:"response_status_code"`
+}
```
tushar-signoz, line 129: nit: (I'm aware other places have this issue as well). Take an informed decision on this approach of using structs as columns `resources_string`, `resource_string_service$$name"` can have evolutions (their name can change along with type ) and how it will support evolutions by default.

pkg/types/tracedetailtypes/waterfall.go
krsoninikhil: This field is not being used anywhere, only getting calculated. Let me remove it altogether.

pkg/types/tracedetailtypes/waterfall.go
krsoninikhil: Right but our doc has `resources` key for logs https://signoz.io/blog/otel-resource-attributes/

pkg/types/tracedetailtypes/waterfall.go
krsoninikhil: Removed in [45590a3](https://github.com/SigNoz/signoz/pull/10794/commits/45590a398dd02076886e40167b99603e58c20c22)

pkg/types/tracedetailtypes/waterfall.go
krsoninikhil: Right, didn't want to change much from existing API except to move it in modules. But sure let me remove it. It should be simpler. 

Edit: removed in [45590a3](https://github.com/SigNoz/signoz/pull/10794/commits/45590a398dd02076886e40167b99603e58c20c22)

pkg/types/tracedetailtypes/waterfall.go
krsoninikhil: Renamed in [45590a3](https://github.com/SigNoz/signoz/pull/10794/commits/45590a398dd02076886e40167b99603e58c20c22)

pkg/types/tracedetailtypes/waterfall.go
krsoninikhil: Not clear on the concern, let's huddle on this.

pkg/types/tracedetailtypes/waterfall.go
nityanandagohain: Yeah we should move to resource, this will require a change in logs as well. But since we are building a new version of this, let's use resource.

pkg/types/tracedetailtypes/waterfall.go
tushar-signoz: Now that we have moved to 'WaterfallSpan'. It doesn't make sense to use snake_case here.
Let client or frontend figure out, how they can/want to use WaterfallAPIResponse to query spans
@nityanandagohain what do you think ?

nityanandagohain: I didn't get you , do you mean changing `external_http_method` to `externalHttpMethod` ?

tushar-signoz: Yes
Earlier we were returning an object with name `Span`, there we were returning snake_case because that's how we have defined fields in Span. 

But now we renamed it to `WaterfallSpan`, now does it still make sense to continue using snake_case ?
IMO we should stay consistent with casing ie camelCase.

nityanandagohain: I would say we should keep snake_case as it represents the actual key in the data. Which is an exception

pkg/types/tracedetailtypes/waterfall.go
nityanandagohain: This is discussed in call we will be as it is here and address this in separate issue https://github.com/SigNoz/engineering-pod/issues/4523

tushar-signoz (approved):
LGTM

PR #10797, opened 2026-04-01
refactor: move waterfall api to module structure with updated response

krsoninikhil:
## Pull Request

---

Followup PR on https://github.com/SigNoz/signoz/pull/10794
v3 is required for udpating the response json of the waterfall api. There wont' be any logical change. Using this requirement as an opportunity to move waterfall api to provider codebase architecture from older query-service

Difference from older version:
- Response contains `attributes` field merging all the attributes instead of `attributes_string` etc. 
- Response contains `resource` field instead of `resource_string`
- Request does not access `isSelectedSpanUncollapsed` field anymore, the selected span id should be included in the `uncollapsedSpans` if it is collapsed. 
- Storing the spans in cache after sorting by timestamp to avoid sorting them on every traversal.
- Break down some long methods for readability.
- Returning the response without event incase of malformed event data instead of responding with error.
- Using parent_span_id to build the span tree instead of links (reference CHILD_OF) to avoid unmarshalling links field. (NEED REVIEW)
- Fix the bug where cache was always getting rejected - https://github.com/SigNoz/engineering-pod/issues/4648

New version endpoint: `/api/v3/traces/:traceID/waterfall`
New response format:
```json
{
  "status": "success",
  "data": {
    "startTimestampMillis": 1772540711280,
    "endTimestampMillis": 1772540716490,
    "durationNano": 5209728334,
    "rootServiceName": "go-trace-domain",
    "rootServiceEntryPoint": "request-1",
    "totalSpansCount": 21,
    "totalErrorSpansCount": 0,
    "serviceNameToTotalDurationMap": { "go-trace-domain": 5209 },
    "spans": [
      {
        "attributes": {
          "request.id": "req-1",
          "request.number": 1,
          "request.spans_count": 20
        },
        "db_name": "",
        "db_operation": "",
        "duration_nano": 5209728334,
        "events": [],
        "external_http_method": "",
        "external_http_url": "",
        "flags": 257,
        "has_error": false,
        "http_host": "",
        "http_method": "",
        "http_url": "",
        "is_remote": "no",
        "kind": 1,
        "kind_string": "Internal",
        "links": "[{\"traceId\":\"1948a35b11a6ecb6e9e44cfc314c9fff\",\"refType\":\"CHILD_OF\"}]",
        "name": "request-1",
        "parent_span_id": "",
        "resource": {
          "service.name": "go-trace-domain",
          "service.version": "1.0.0",
          "signoz.collector.id": "2e98f7f5-3dfd-4391-b2b1-25b1da4393b1"
        },
        "response_status_code": "",
        "span_id": "c2420281b63bdd4c",
        "status_code": 0,
        "status_code_string": "Unset",
        "status_message": "",
        "timestamp": "2026-03-03T12:25:11.280504Z",
        "trace_id": "1948a35b11a6ecb6e9e44cfc314c9fff",
        "trace_state": "",
        "sub_tree_node_count": 21,
        "has_children": true,
        "has_siblings": false,
        "level": 0
      }
    ],
    "hasMissingSpans": false,
    "uncollapsedSpans": null
  }
}

```

https://github.com/SigNoz/engineering-pod/issues/4204

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---


pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -0,0 +1,272 @@
...
+				serviceNameIntervalMap[span.ServiceName],
+				tracedetailv2.Interval{StartTime: startTimeUnixNano, Duration: span.DurationNano, Service: span.ServiceName},
+			)
+
+			spanIDToSpanNodeMap[span.SpanID] = span
+		}
+
+		// Build tree: parent-child relationships and missing spans
+		for _, spanNode := range spanIDToSpanNodeMap {
+			if spanNode.ParentSpanID != "" {
```
krsoninikhil: Parent child calculation is being done based on `spanNode.References` with RefType as CHILD_OF in existing version. https://github.com/SigNoz/signoz/blob/main/pkg/query-service/app/clickhouseReader/reader.go#L1029

pkg/modules/tracedetail/impltracedetail/handler.go
```diff
@@ -0,0 +1,56 @@
...
+type handler struct {
+	module tracedetail.Module
+}
+
+func NewHandler(module tracedetail.Module) tracedetail.Handler {
+	return &handler{module: module}
+}
+
+func (h *handler) GetWaterfall(rw http.ResponseWriter, r *http.Request) {
+	claims, err := authtypes.ClaimsFromContext(r.Context())
```
nityanandagohain: we add the timeout on each handler functions

```
	ctx, cancel := context.WithTimeout(r.Context(), 10*time.Second)
	defer cancel()
```

nityanandagohain: @vikrantgupta25  something to add in docs ?

pkg/modules/tracedetail/impltracedetail/handler.go
```diff
@@ -0,0 +1,56 @@
...
+		return
+	}
+
+	orgID, err := valuer.NewUUID(claims.OrgID)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	traceID := mux.Vars(r)["traceId"]
```
nityanandagohain: should we move to trace_id instead as we moved from traceId to trace_id, but we do support traceId in quey.

pkg/modules/tracedetail/impltracedetail/module.go
nityanandagohain: If possible let's keep it as for scope of this PR and create a separate one where we want to change this

aks07:
V3 waterfall API response issues:                                                                                                                                                                                   
                                                                                                                                                                                                                      
  1. serviceName is missing from the span objects                     
  2. Extra data wrapper — the response is wrapped in { status: "success", data: { ... } } instead of returning the payload directly like v2. The frontend currently handles this with a workaround but ideally the    
  response format should be consistent.                                                                                                                                                                               
  3. [MINOR ]`attributes_string` and `resources_string` instead of `attributes` and `resources`- same naming as the logs structure
  
 As earlier discussed, The overall span format should remain the same as v2, just with `attributes_string` and `resources_string` added as new fields.

cc. @nityanandagohain @krsoninikhil

aks07:
> V3 waterfall API response issues:
> 
> 1. serviceName is missing from the span objects
> 2. Extra data wrapper — the response is wrapped in { status: "success", data: { ... } } instead of returning the payload directly like v2. The frontend currently handles this with a workaround but ideally the
>    response format should be consistent.
> 3. [MINOR ]`attributes_string` and `resources_string` instead of `attributes` and `resources`- same naming as the logs structure
> 
> As earlier discussed, The overall span format should remain the same as v2, just with `attributes_string` and `resources_string` added as new fields.
> 
> cc. @nityanandagohain @krsoninikhil

If there's a rationale behind the current format happy to discuss  Just flagging early since the frontend is already built against the expected response shape, and it's easier to align now than   
  iterate on it multiple times. Let me know!

pkg/modules/tracedetail/impltracedetail/handler.go
krsoninikhil: This is path variable in the route and not a query param, so it's not really exposed to user as key value pair. Don't think this should be changed. But let's discuss if there is a good reason to do so.

nityanandagohain: No strong opinion, just that using `trace_id` everywhere will remove the overhead of someone needing to remeber and maintain `traceId`. Thinking of pov where the API is used by users.

pkg/modules/tracedetail/impltracedetail/module.go
krsoninikhil: Fair, should be taken up separately. We have picked up few improvements in last PR also though since we will be retesting the whole module again anyway. https://github.com/SigNoz/signoz/pull/10794#discussion_r3024537409

Since it's a new endpoint and doesn't impact existing modules anyway, fixing now compare to after frontend starts using this seems to have less impact area.

pkg/modules/tracedetail/impltracedetail/handler.go
krsoninikhil: Even API users don't need to remember right? It's path variable so API url would be like `/api/v3/trace/26c934402c45eb8f4ed1f3a5d3a31177/waterfall`?

nityanandagohain: right my bad

pkg/modules/tracedetail/impltracedetail/handler.go
krsoninikhil: Fixed in [9c056f8](https://github.com/SigNoz/signoz/pull/10797/commits/9c056f809ae66f3c1db4335b8b5295ec75e8ef59)

nityanandagohain (dismissed):
let's try to avoid force push.

pkg/modules/tracedetail/impltracedetail/handler.go
krsoninikhil: Removed again in [276e098](https://github.com/SigNoz/signoz/pull/10797/commits/276e09853e1029cb4f284161276c01b6b818deb5) based on this discussion https://signoz-team.slack.com/archives/C08P77ZDHV0/p1775808188079339?thread_ts=1775128264.504469&cid=C08P77ZDHV0

nityanandagohain (dismissed):

aks07 (dismissed):

pkg/apiserver/signozapiserver/tracedetail.go
```diff
@@ -0,0 +1,33 @@
...
+	"net/http"
+
+	"github.com/SigNoz/signoz/pkg/http/handler"
+	"github.com/SigNoz/signoz/pkg/types"
+	"github.com/SigNoz/signoz/pkg/types/tracedetailtypes"
+	"github.com/gorilla/mux"
+)
+
+func (provider *provider) addTraceDetailRoutes(router *mux.Router) error {
+	if err := router.Handle("/api/v3/traces/{traceID}/waterfall", handler.New(
```
swapnil-signoz, line 13:
```suggestion
	if err := router.Handle("/api/v3/traces/{trace_id}/waterfall", handler.New(
```

pkg/modules/tracedetail/impltracedetail/handler.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	orgID, err := valuer.NewUUID(claims.OrgID)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	traceID := mux.Vars(r)["traceID"]
+	if traceID == "" {
```
swapnil-signoz: Will this ever be the case as path will not match?
`/api/v3/traces//waterfall`

pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -0,0 +1,272 @@
...
+
+const (
+	traceDB           = "signoz_traces"
+	traceTable        = "distributed_signoz_index_v3"
+	traceSummaryTable = "distributed_trace_summary"
+	cacheTTL          = 5 * time.Minute
+	fluxInterval      = 2 * time.Minute
+)
+
+var errTraceNotFound = errors.NewNotFoundf(errors.CodeNotFound, "trace not found")
```
swapnil-signoz: Can be part of the types?

pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -0,0 +1,272 @@
...
+			resource_string_service$$name, name, links as references,
+			attributes_string, attributes_number, attributes_bool, resources_string,
+			events, status_message, status_code_string, kind_string, parent_span_id,
+			flags, is_remote, trace_state, status_code,
+			db_name, db_operation, http_method, http_url, http_host,
+			external_http_method, external_http_url, response_status_code
+		FROM %s.%s
+		WHERE trace_id=$1 AND ts_bucket_start>=$2 AND ts_bucket_start<=$3
+		ORDER BY timestamp ASC, name ASC`,
+		traceDB, traceTable,
```
swapnil-signoz: There is no abstraction store for querying to ClickHouse?

pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -0,0 +1,272 @@
...
+		StartTime:                     startTime,
+		EndTime:                       endTime,
+		DurationNano:                  durationNano,
+		TotalSpans:                    uint64(len(spanItems)),
+		TotalErrorSpans:               totalErrorSpans,
+		SpanIDToSpanNodeMap:           spanIDToSpanNodeMap,
+		ServiceNameToTotalDurationMap: tracedetailv2.CalculateServiceTime(serviceNameIntervalMap),
+		TraceRoots:                    traceRoots,
+		HasMissingSpans:               hasMissingSpans,
+	}, nil
```
swapnil-signoz: Create a new function in types for such things across the change set.
`func NewWaterfallTrace(...)`, do not use types directly.

pkg/modules/tracedetail/impltracedetail/waterfall.go
```diff
@@ -0,0 +1,191 @@
...
+	"sort"
+
+	"github.com/SigNoz/signoz/pkg/types/tracedetailtypes"
+)
+
+var (
+	spanLimitPerRequest         float64 = 500
+	maxDepthForSelectedChildren int     = 5
+	MaxLimitToSelectAllSpans    uint    = 10_000
+)
```
swapnil-signoz: Can this be part of types as well?

pkg/modules/tracedetail/impltracedetail/waterfall.go
```diff
@@ -0,0 +1,191 @@
+package impltracedetail
```
swapnil-signoz: Is there any specific reason to have separate file for this?

pkg/modules/tracedetail/impltracedetail/waterfall.go
```diff
@@ -0,0 +1,191 @@
...
+			preOrderTraversal = append(preOrderTraversal, traversal...)
+
+			if rootServiceName == "" {
+				rootServiceName = rootNode.ServiceName
+			}
+			if rootServiceEntryPoint == "" {
+				rootServiceEntryPoint = rootNode.Name
+			}
+		}
+	}
```
swapnil-signoz: Check if this can be broken into smaller function which is easier to read.
As it is going 5 branches deep.

pkg/modules/tracedetail/impltracedetail/waterfall.go
```diff
@@ -0,0 +1,191 @@
...
+	MaxLimitToSelectAllSpans    uint    = 10_000
+)
+
+type traverseOpts struct {
+	uncollapsedSpans map[string]struct{}
+	selectedSpanID   string
+	selectAll        bool
+}
+
+func traverseTrace(
```
swapnil-signoz: keep public functions/methods above private functions/methods 
```go
func GetSomething()

// then 

func someFunc()
```

pkg/apiserver/signozapiserver/tracedetail.go
krsoninikhil: This is internal variable and not exposed to user, why should it be snake case?

pkg/apiserver/signozapiserver/tracedetail.go
```diff
@@ -0,0 +1,33 @@
...
+		provider.authZ.ViewAccess(provider.traceDetailHandler.GetWaterfall),
+		handler.OpenAPIDef{
+			ID:                  "GetWaterfall",
+			Tags:                []string{"tracedetail"},
+			Summary:             "Get waterfall view for a trace",
+			Description:         "Returns the waterfall view of spans for a given trace ID with tree structure, metadata, and windowed pagination",
+			Request:             new(tracedetailtypes.WaterfallRequest),
```
swapnil-signoz: Can we have `PostableWaterfall` type?
It can be `type PostableWaterfall = WaterfallRequest` as well

pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -0,0 +1,272 @@
...
+	response.UncollapsedSpans = uncollapsedSpans
+	response.StartTimestampMillis = traceData.StartTime / 1000000
+	response.EndTimestampMillis = traceData.EndTime / 1000000
+	response.TotalSpansCount = traceData.TotalSpans
+	response.TotalErrorSpansCount = traceData.TotalErrorSpans
+	response.RootServiceName = rootServiceName
+	response.RootServiceEntryPoint = rootServiceEntryPoint
+	response.ServiceNameToTotalDurationMap = traceData.ServiceNameToTotalDurationMap
+	response.HasMissingSpans = traceData.HasMissingSpans
+	response.HasMore = !selectAllSpans
```
swapnil-signoz: Have a `NewWaterfallResponse` function in types to handle this.

pkg/modules/tracedetail/impltracedetail/handler.go
krsoninikhil: Right, this check is not required.

pkg/modules/tracedetail/impltracedetail/waterfall.go
krsoninikhil: This is waterfall specific logic and in future we might want to have flamegraph in tracedetail as well.

pkg/apiserver/signozapiserver/tracedetail.go
swapnil-signoz: I was not able to find any reference for existing API endpoints for whether these path params should be snake case or camel case. Cloud Integration APIs are using snake cases.
cc: @vikrantgupta25

pkg/apiserver/signozapiserver/tracedetail.go
krsoninikhil: Do we need to define both postable and request? Should I just rename WaterfallRequest to PostableWaterfall? But WaterfallRequest seems more readable then PostableWaterfall.

swapnil-signoz: https://github.com/SigNoz/signoz/blob/main/docs/contributing/go/handler.md
This should answer your question.

We want to have separation of concern among types for handlers (Postables, Gettables), domain and store (Storables). They can be aliases of domain types if required.

To answer your question, both PostableWaterfall and WaterfallRequest will be required, one for handler and one for module.

pkg/modules/tracedetail/impltracedetail/waterfall.go
krsoninikhil: Fixed in [2a3258e](https://github.com/SigNoz/signoz/pull/10797/commits/2a3258e7165b78bf95316b4aa039eee1ce89b02a)

pkg/modules/tracedetail/impltracedetail/module.go
krsoninikhil: Fixed in [864fd37](https://github.com/SigNoz/signoz/pull/10797/commits/864fd3785cb6713fcfea308acf8ce6f80cc76618)

pkg/modules/tracedetail/impltracedetail/module.go
krsoninikhil: Fixed in [7fb1265](https://github.com/SigNoz/signoz/pull/10797/commits/7fb126596e40aba0b01e98584e933114a2ff197d)

pkg/modules/tracedetail/impltracedetail/waterfall.go
krsoninikhil: [7fb1265](https://github.com/SigNoz/signoz/pull/10797/commits/7fb126596e40aba0b01e98584e933114a2ff197d)

pkg/modules/tracedetail/impltracedetail/waterfall.go
krsoninikhil: Fixed in [7fb1265](https://github.com/SigNoz/signoz/pull/10797/commits/7fb126596e40aba0b01e98584e933114a2ff197d)

pkg/modules/tracedetail/impltracedetail/module.go
krsoninikhil: Fixed in [fc684e3](https://github.com/SigNoz/signoz/pull/10797/commits/fc684e3b5db01b0b6d64356145129c3578621e6b)

pkg/modules/tracedetail/impltracedetail/store.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/telemetrystore"
+	"github.com/SigNoz/signoz/pkg/types/tracedetailtypes"
+)
+
+type traceStore interface {
+	GetTraceSummary(ctx context.Context, traceID string) (*tracedetailtypes.TraceSummary, error)
+	GetTraceSpans(ctx context.Context, traceID string, summary *tracedetailtypes.TraceSummary) ([]tracedetailtypes.SpanModel, error)
+}
```
swapnil-signoz: nit: we usually keep stores in types `types/tracedetailtypes/store.go`
You can take reference for other modules and check as well.

pkg/modules/tracedetail/impltracedetail/handler.go
```diff
@@ -0,0 +1,51 @@
...
+
+	traceID := mux.Vars(r)["traceID"] // can't be empty or route won't match
+
+	var req tracedetailtypes.WaterfallRequest
+	if err := binding.JSON.BindBody(r.Body, &req); err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	result, err := h.module.GetWaterfall(r.Context(), orgID, traceID, &req)
```
swapnil-signoz:
```suggestion
	result, err := h.module.GetWaterfall(r.Context(), orgID, mux.Vars(r)["traceID"], &req)
```

nit: variable traceID can be avoided as it is being used only at a single place

pkg/modules/tracedetail/impltracedetail/handler.go
```diff
@@ -0,0 +1,51 @@
...
+
+	orgID, err := valuer.NewUUID(claims.OrgID)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	traceID := mux.Vars(r)["traceID"] // can't be empty or route won't match
+
+	var req tracedetailtypes.WaterfallRequest
```
swapnil-signoz: If pointer is required later while passing to module.GetWaterfall(), we can initialise a pointer itself?
```suggestion
	req := new(tracedetailtypes.WaterfallRequest)
```

pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -0,0 +1,209 @@
...
+		uint64(len(spanItems)),
+		totalErrorSpans,
+		spanIDToSpanNodeMap,
+		tracedetailv2.CalculateServiceTime(serviceNameIntervalMap),
+		traceRoots,
+		hasMissingSpans,
+	)
+}
+
+func (m *module) getFromCache(ctx context.Context, orgID valuer.UUID, traceID string) (*tracedetailtypes.WaterfallTrace, error) {
```
swapnil-signoz: nit: Lets try to keep module methods together, its easier to understand the flow that way.

pkg/modules/tracedetail/impltracedetail/store.go
```diff
@@ -0,0 +1,71 @@
...
+	query := fmt.Sprintf(
+		"SELECT trace_id, min(start) AS start, max(end) AS end, sum(num_spans) AS num_spans FROM %s.%s WHERE trace_id=$1 GROUP BY trace_id",
+		tracedetailtypes.TraceDB, tracedetailtypes.TraceSummaryTable,
+	)
+	var summary tracedetailtypes.TraceSummary
+	err := s.telemetryStore.ClickhouseDB().QueryRow(ctx, query, traceID).Scan(
+		&summary.TraceID, &summary.Start, &summary.End, &summary.NumSpans,
+	)
+	if err != nil {
+		if err == sql.ErrNoRows {
```
swapnil-signoz:
```suggestion
		if errors.Is(err, sql.ErrNoRows) {
```

Check other places as well (if any)

pkg/modules/tracedetail/impltracedetail/store.go
```diff
@@ -0,0 +1,71 @@
...
+			return nil, tracedetailtypes.ErrTraceNotFound
+		}
+		return nil, errors.Newf(errors.TypeInternal, errors.CodeInternal, "error querying trace summary: %v", err)
```
swapnil-signoz:
```suggestion
		return nil, errors.WrapInternalf(err, errors.CodeInternal, "error querying trace summary")
```

We can wrap errors if we are explicitly passing `err` in message

swapnil-signoz:
```suggestion
		return nil, errors.WrapInternalf(err, errors.CodeInternal, "error querying trace summary")
```

We can wrap errors if we are explicitly passing `err` in message. Please check other places if any

pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -0,0 +1,209 @@
...
+	"time"
+
+	"github.com/SigNoz/signoz/pkg/cache"
+	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/factory"
+	"github.com/SigNoz/signoz/pkg/modules/tracedetail"
+	"github.com/SigNoz/signoz/pkg/telemetrystore"
+	"github.com/SigNoz/signoz/pkg/types/tracedetailtypes"
+	"github.com/SigNoz/signoz/pkg/valuer"
+
```
swapnil-signoz: nit: space can be removed as all are repository packages.

There are opinionated import sortings
```
std

3rd party

repository
```

We dont have any strong opinion on this one I think, but following SHOULD be followed, it looks neat.
```
std

all others
```

They are alphabetically sorted (should go without saying) :))

pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -0,0 +1,209 @@
...
+		cache:  cache,
+		logger: providerSettings.Logger,
+	}
+}
+
+func (m *module) GetWaterfall(ctx context.Context, orgID valuer.UUID, traceID string, req *tracedetailtypes.WaterfallRequest) (*tracedetailtypes.WaterfallResponse, error) {
+	traceData, err := m.getTraceData(ctx, orgID, traceID)
+	if err != nil {
+		if errors.Is(err, tracedetailtypes.ErrTraceNotFound) {
+			return new(tracedetailtypes.WaterfallResponse), nil
```
swapnil-signoz: nit: is there any reason we return zero value pointer and not error? Please add a comment so that reasoning is not lost.

pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -0,0 +1,209 @@
...
+					SpanID:        spanNode.ParentSpanID,
+					TraceID:       spanNode.TraceID,
+					Name:          "Missing Span",
+					TimeUnixMilli: spanNode.TimeUnixMilli,
+					DurationNano:  spanNode.DurationNano,
+					Events:        make([]tracedetailtypes.Event, 0),
+					Children:      make([]*tracedetailtypes.WaterfallSpan, 0),
+					Attributes:    make(map[string]any),
+					Resource:      make(map[string]string),
+				}
```
swapnil-signoz: nit: we can have a NewWaterfallSpan for this as well. Goal is to not use types directly.

pkg/modules/tracedetail/impltracedetail/waterfall.go
```diff
@@ -0,0 +1,193 @@
...
+	if len(traceRoots) > 0 {
+		rootServiceName = traceRoots[0].ServiceName
+		rootServiceEntryPoint = traceRoots[0].Name
+	}
+
+	return preOrderTraversal[startIndex:endIndex], slices.Collect(maps.Keys(uncollapsedSpanMap)), rootServiceName, rootServiceEntryPoint
+}
+
+// getUncollapsedSpanMap creates a map from uncollapsed spans ids and root to selected span path.
+func getUncollapsedSpanMap(uncollapsedSpans []string, selectedSpanID string, traceRoots []*tracedetailtypes.WaterfallSpan, spanIDToSpanNodeMap map[string]*tracedetailtypes.WaterfallSpan) map[string]struct{} {
```
swapnil-signoz: I think these were missed to be moved below public functions

pkg/modules/tracedetail/impltracedetail/waterfall.go
```diff
@@ -0,0 +1,193 @@
...
+				}
+			}
+			break
+		}
+	}
+	return uncollapsedSpanMap
+}
+
+// windowAroundIndex returns start/end indices for a window of SpanLimitPerRequest spans.
+func windowAroundIndex(selectedIndex, total int) (start, end int) {
```
swapnil-signoz: This one as well

pkg/modules/tracedetail/impltracedetail/waterfall.go
```diff
@@ -0,0 +1,193 @@
...
+	for _, child := range span.Children {
+		SortSpanChildren(child)
+	}
+}
+
+type traverseOpts struct {
+	uncollapsedSpans map[string]struct{}
+	selectedSpanID   string
+	selectAll        bool
+}
```
swapnil-signoz: nit: we can have types at the top of the file

pkg/types/tracedetailtypes/waterfall.go
```diff
@@ -238,10 +256,58 @@ func (c *WaterfallCache) Clone() cachetypes.Cacheable {
...
+		EndTimestampMillis:            traceData.EndTime / 1_000_000,
+		TotalSpansCount:               traceData.TotalSpans,
+		TotalErrorSpansCount:          traceData.TotalErrorSpans,
+		RootServiceName:               rootServiceName,
+		RootServiceEntryPoint:         rootServiceEntryPoint,
+		ServiceNameToTotalDurationMap: serviceDurationsMillis,
+		HasMissingSpans:               traceData.HasMissingSpans,
+		HasMore:                       !selectAllSpans,
+	}
+}
```
swapnil-signoz: nit: we can move these above the methods on WaterfallTrace, having New functions near the types makes it easier to read how types are initialized (new functions) before reading its behaviour (methods)

You can check `serviceaccounttypes` for reference

swapnil-signoz (commented):
About the `pkg/modules/tracedetail/impltracedetail/store.go`, please get it reviewed from some one relevant.
I am not sure how ClickHouse stores are constructed and used across our repo, but I hope you are already referencing other modules/packages which are doing something similar.
If this is the first time we are doing it this way, please involve right stakeholders :))

pkg/modules/tracedetail/impltracedetail/handler.go
krsoninikhil: Sure, but I don't see any difference.

pkg/modules/tracedetail/impltracedetail/module.go
krsoninikhil: Was trying to preserve old behavior, but let me change this to 404.

pkg/modules/tracedetail/impltracedetail/handler.go
krsoninikhil: Sure but having it separate allows adding comment or any validation in future without touching the usage line.

swapnil-signoz: 100%, just that I could not see any at the moment so suggested the change. Once we have any validations feel free to take it out.

pkg/modules/tracedetail/impltracedetail/handler.go
```diff
@@ -0,0 +1,49 @@
...
+		return
+	}
+
+	orgID, err := valuer.NewUUID(claims.OrgID)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	req := new(tracedetailtypes.WaterfallRequest)
```
swapnil-signoz: Please take a call for using `PostableWaterfall` @vikrantgupta25

swapnil-signoz (dismissed):
Approving with a comment

pkg/modules/tracedetail/impltracedetail/handler.go
```diff
@@ -0,0 +1,49 @@
...
+	if err := binding.JSON.BindBody(r.Body, req); err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	result, err := h.module.GetWaterfall(r.Context(), orgID, mux.Vars(r)["traceID"], req)
```
vikrantgupta25: `mux.Vars(r)["traceID"]` validate that this isn't empty in handler rather than making a ch query ?

pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -0,0 +1,176 @@
...
+	store  tracedetailtypes.TraceStore
+	cache  cache.Cache
+	logger *slog.Logger
+}
+
+func NewModule(telemetryStore telemetrystore.TelemetryStore, cache cache.Cache, providerSettings factory.ProviderSettings) tracedetail.Module {
+	return &module{
+		store:  newClickhouseTraceStore(telemetryStore),
+		cache:  cache,
+		logger: providerSettings.Logger,
```
vikrantgupta25: make this a scoped logger

pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil, errors.Newf(errors.TypeInternal, errors.CodeInternal, "trace end time within flux interval, traceID: %s", traceID)
+	}
+
+	m.logger.InfoContext(ctx, "cache hit for v3 waterfall", slog.String("trace_id", traceID))
+	return cachedData, nil
+}
+
+// computeWaterfallTrace builds a WaterfallTrace from raw span rows by constructing
+// the parent-child tree, inserting missing span placeholders, and calculating service times.
+func computeWaterfallTrace(spanItems []tracedetailtypes.SpanModel) *tracedetailtypes.WaterfallTrace {
```
vikrantgupta25: this should all happen in types.

pkg/modules/tracedetail/impltracedetail/store.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+		return nil, errors.WrapInternalf(err, errors.CodeInternal, "error querying trace summary")
+	}
+	return &summary, nil
+}
+
+func (s *clickhouseTraceStore) GetTraceSpans(ctx context.Context, traceID string, summary *tracedetailtypes.TraceSummary) ([]tracedetailtypes.SpanModel, error) {
+	// DISTINCT ON (span_id) is ClickHouse-specific syntax not supported by sqlbuilder
+	query := fmt.Sprintf(`
+		SELECT DISTINCT ON (span_id)
```
vikrantgupta25, line 47: sqlbuilder please

pkg/modules/tracedetail/impltracedetail/waterfall.go
```diff
@@ -0,0 +1,207 @@
...
+
+func GetSelectedSpans(uncollapsedSpans []string, selectedSpanID string, traceRoots []*tracedetailtypes.WaterfallSpan, spanIDToSpanNodeMap map[string]*tracedetailtypes.WaterfallSpan) ([]*tracedetailtypes.WaterfallSpan, []string, string, string) {
+	uncollapsedSpanMap := getUncollapsedSpanMap(uncollapsedSpans, selectedSpanID, traceRoots, spanIDToSpanNodeMap)
+
+	var preOrderTraversal = make([]*tracedetailtypes.WaterfallSpan, 0)
```
vikrantgupta25: this entire file feels like it belongs to types ? 

```
type Trace struct { 
  spans []*Span
}

func (trace *trace) GetPreOrder() []spans {}
```

vikrantgupta25 (changes requested):
please go through this once as well: https://github.com/SigNoz/signoz/blob/main/docs/contributing/go/types.md

pkg/modules/tracedetail/impltracedetail/handler.go
krsoninikhil: Empty traceID should not match this route -- `/traces//waterfall`, so removed it since it was unnecessary. We can add other validations like a valid id if required.

pkg/modules/tracedetail/impltracedetail/store.go
krsoninikhil: There is no direct support for distinct on in sqlbuilder, we'll have to either change the query to use group by instead of use string interpolation https://github.com/huandu/go-sqlbuilder#freestyle-builder. 

Current version seems simpler than both of these options.

pkg/modules/tracedetail/impltracedetail/waterfall.go
krsoninikhil: Broken down and moved the whole traverse logic to types. Types seems to have a lot of business logic now. Let's discuss if I should move back some of the methods. Checkout `pkg/types/tracedetailtypes/waterfall_trace.go` especially.

pkg/types/tracedetailtypes/waterfall_span.go
```diff
@@ -0,0 +1,282 @@
...
+
+// ErrTraceNotFound is returned when a trace ID has no matching spans in ClickHouse.
+var ErrTraceNotFound = errors.NewNotFoundf(errors.CodeNotFound, "trace not found")
+
+// WaterfallRequest is the request body for the v3 waterfall API.
+type WaterfallRequest struct {
+	SelectedSpanID   string   `json:"selectedSpanId"`
+	UncollapsedSpans []string `json:"uncollapsedSpans"`
+	Limit            uint     `json:"limit"`
+}
```
krsoninikhil: This represent the query parameters to get the waterfall trace and not really the waterfall trace itself, so Gettable/Postable might not fit.  Since we are using post request here instead of get, so kept the `Request` suffix instead of `Query`. Let me know if there is better way to name this.

pkg/modules/tracedetail/tracedetail.go
```diff
@@ -15,5 +15,5 @@ type Handler interface {
 
 // Module defines the business logic for trace detail operations.
 type Module interface {
-	GetWaterfall(ctx context.Context, orgID valuer.UUID, traceID string, req *tracedetailtypes.WaterfallRequest) (*tracedetailtypes.WaterfallResponse, error)
+	GetWaterfall(ctx context.Context, orgID valuer.UUID, traceID string, req *tracedetailtypes.WaterfallRequest) (*tracedetailtypes.GettableWaterfallTrace, error)
```
vikrantgupta25:
```suggestion
	GetWaterfall(ctx context.Context, orgID valuer.UUID, traceID string, req *tracedetailtypes.WaterfallRequest) (*tracedetailtypes.WaterfallTrace, error)
```

pkg/modules/tracedetail/impltracedetail/handler.go
vikrantgupta25: get the `WaterfallTrace` here and then do the conversion for ns to ms and return Gettable

pkg/types/tracedetailtypes/waterfall_trace.go
```diff
@@ -0,0 +1,316 @@
...
+func (c *WaterfallTrace) MarshalBinary() (data []byte, err error) {
+	return json.Marshal(c)
+}
+
+func (c *WaterfallTrace) UnmarshalBinary(data []byte) error {
+	return json.Unmarshal(data, c)
+}
+
+// GettableWaterfallTrace is the response for the v3 waterfall API.
+type GettableWaterfallTrace struct {
```
vikrantgupta25: let's have all the types on the top ?

vikrantgupta25 (changes requested):

pkg/types/tracedetailtypes/waterfall_span.go
```diff
@@ -0,0 +1,282 @@
...
+)
+
+const (
+	// ClickHouse database and table names for trace queries.
+	TraceDB           = "signoz_traces"
+	TraceTable        = "distributed_signoz_index_v3"
+	TraceSummaryTable = "distributed_trace_summary"
+
+	// Cache and freshness thresholds.
+	WaterfallCacheTTL = 5 * time.Minute
```
vikrantgupta25: these should all be part of config for the module

vikrantgupta25 (changes requested):

pkg/types/tracedetailtypes/waterfall_span.go
```diff
@@ -0,0 +1,282 @@
...
+	TimeUnixNano uint64         `json:"timeUnixNano,omitempty"`
+	AttributeMap map[string]any `json:"attributeMap,omitempty"`
+	IsError      bool           `json:"isError,omitempty"`
+}
+
+// WaterfallSpan represents the span in waterfall response,
+// this uses snake_case keys for response as a special case since these
+// keys can be directly used to query spans and client need to know the actual fields.
+// This pattern should not be copied elsewhere.
+type WaterfallSpan struct {
```
vikrantgupta25, line 41: is this type gonna be the same for flamegraph as well ? So basically `Span` instead of `WaterfallSpan`

pkg/types/tracedetailtypes/waterfall_span.go
```diff
@@ -0,0 +1,282 @@
...
+		return s.Children[i].TimeUnixNano < s.Children[j].TimeUnixNano
+	})
+	for _, child := range s.Children {
+		child.SortChildren()
+	}
+}
+
+// CopyWithoutChildren creates a shallow copy and resets tree-structure fields.
+// SubTreeNodeCount is preserved (must be pre-computed via computeSubTreeNodeCount).
+func (s *WaterfallSpan) CopyWithoutChildren(level uint64) *WaterfallSpan {
```
vikrantgupta25:
```suggestion
func (s *WaterfallSpan) GetWithoutChildren(level uint64) *WaterfallSpan {
```

copy gives a notion that it will accept a dest variable which will have the final result ?

pkg/types/tracedetailtypes/waterfall_span.go
```diff
@@ -0,0 +1,282 @@
...
+	cp := *s
+	cp.Level = level
+	cp.HasChildren = len(s.Children) > 0
+	cp.Children = make([]*WaterfallSpan, 0)
+	return &cp
+}
+
+// ComputeSubTreeNodeCount recursively sets SubTreeNodeCount on every span in the subtree.
+// SubTreeNodeCount = 1 (self) + total number of descendants.
+func (s *WaterfallSpan) ComputeSubTreeNodeCount() uint64 {
```
vikrantgupta25:
```suggestion
func (span *WaterfallSpan) GetSubtreeNodeCount() uint64 {
```

vikrantgupta25 (changes requested):

pkg/types/tracedetailtypes/waterfall_span.go
krsoninikhil: Might not be, there was already a FlamegraphSpan for flamegraph API. It has some waterfall tree related fields as well. And using Span was confusing with otel Span to kept it like this.

pkg/modules/tracedetail/impltracedetail/handler.go
krsoninikhil: It's not just the time unit conversion, `WaterfallTrace` contains all spans, while `GettableWaterfallTrace` contains only spans that are requested using `WaterfallRequest`. We'll have to pass the selection logic to handler as well. This is the requirement:
```
tracedetailtypes.NewGettableWaterfallTrace(waterfallTrace, selectedSpans, uncollapsedSpans, selectAllSpans)
```

If unit conversion in module is the reason, would returning `GettableWaterfallTrace` itself without unit change from module and then converting it in handler to what client needs (millisecond) be better?

pkg/types/tracedetailtypes/waterfall_trace.go
krsoninikhil: Sure, I was trying to keep typedef, NewX and it's receiver methods together. Are keeping all types then all constructors and then type specific methods?

pkg/types/tracedetailtypes/waterfall_span.go
krsoninikhil: Fixed in [0faeedf](https://github.com/SigNoz/signoz/pull/10797/commits/0faeedf5a283b9606da456f0f46d0807463f39a6)

pkg/types/tracedetailtypes/waterfall_span.go
krsoninikhil: Fixed in [0faeedf](https://github.com/SigNoz/signoz/pull/10797/commits/0faeedf5a283b9606da456f0f46d0807463f39a6)

pkg/types/tracedetailtypes/waterfall_span.go
krsoninikhil: Removed the cache itself in [8bec044](https://github.com/SigNoz/signoz/pull/10797/commits/8bec0449096eb9bedb5071fadc8a00df77a9f599)

pkg/modules/tracedetail/impltracedetail/handler.go
krsoninikhil: Or to keep the `GettableX` limited to handler layer, I can define another type for module response, something like below and create `GettableWaterfallTrace` in handler from this.

```go
type WaterfallSelectionResult struct {
    TraceData        *WaterfallTrace
    SelectedSpans    []*WaterfallSpan
    UncollapsedSpans []string
    SelectedAllSpans   bool
}
```

pkg/types/tracedetailtypes/waterfall_trace.go
krsoninikhil: Moved in [2648984](https://github.com/SigNoz/signoz/pull/10797/commits/264898486fd83ac7ad1bfbaa5619ff19dcf08862)

vikrantgupta25 (dismissed):

therealpandey (dismissed):

therealpandey (approved):

aks07 (approved):

PR #10798, opened 2026-04-01
feat: adding aws service definitions in types

swapnil-signoz:
### Summary
- Adding AWS service definition JSON files to cloud integration types


vikrantgupta25:
Also evaluate the binary size bump due to this duplication

PR #10800, opened 2026-04-01
feat(infrastructure-monitoring-k8s-and-hosts): add shared component for list/details

H4ad:
## Pull Request

---

This PR introduces a shared compoent to be used by all containers inside the Infrastructure Monitoring K8s, all those containers had/look the same and the code was just copied multiple times inside those containers.

Now we have the `K8sBaseList` and `K8sBaseDetails` that represents the list and the details modal.

They are now separated to not have many props being passed from List to Details just because Details needs specific data.

Also closes https://github.com/SigNoz/engineering-pod/issues/4002 and closes https://github.com/SigNoz/engineering-pod/issues/4000

Before:

https://github.com/user-attachments/assets/6dcc4716-cbf0-4cdf-8a65-cefb6e539b49

After:

https://github.com/user-attachments/assets/5a62bcce-dd5d-4910-a993-138cd73545f3

About visual changes, the only thing I changed was the column add/remove drawer:

Before:

image

After:

image

> Removed reorder to add later when we change the implementation of table.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Infraestructure Monitoring - Pods
- Potential regressions: Something is missing/some query was wrongly migrated.

---

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | We are migrating the sub-pages of Infrastructure monitoring to a shared component to be make the maintainance of these modules easier in the future. |

---

### 📋 Checklist
- [x] Manually tested


frontend/src/container/InfraMonitoringK8s/InfraMonitoringK8s.tsx
```diff
@@ -321,6 +322,25 @@ export default function InfraMonitoringK8s(): JSX.Element {
 		}
 	};
 
+	const showFiltersComp = useMemo(() => {
```
AshwinBhatkal, line 309: nit: showFiltersComponent. Why not name it controlListPrefix? 

Also, wouldn't it be better inside the component itself instead of in this component?

frontend/src/container/InfraMonitoringK8s/Pods/table.module.scss
```diff
@@ -0,0 +1,11 @@
+.entityGroupHeader {
```
AshwinBhatkal, line 1: personal preference here would be to use kebab case as normal css and use camelCase in js when referencing. Keeps the originality of css. We could even have BEM when required for easy differentation

AshwinBhatkal: We should fix on one though and then enforce that

frontend/src/container/InfraMonitoringK8s/Pods/table.config.tsx
```diff
@@ -0,0 +1,328 @@
+import React from 'react';
```
AshwinBhatkal, line 1: This component does not really return a table? Only the configs?

Question is around the naming mainly. See that it is moved from utils.tsx to here

frontend/src/container/InfraMonitoringK8s/Pods/api.ts
```diff
@@ -1,21 +1,35 @@
...
-export interface K8sPodsListPayload {
-	filters: TagFilter;
-	groupBy?: BaseAutocompleteData[];
-	offset?: number;
-	limit?: number;
-	orderBy?: {
-		columnName: string;
-		order: 'asc' | 'desc';
-	};
+export const podsMetaMap = [
```
AshwinBhatkal, line 10: can this be typed better?

AshwinBhatkal: Probably with this itself - dotToUnder const inside frontend/src/container/InfraMonitoringK8s/Base/utils.tsx

frontend/src/container/InfraMonitoringK8s/Base/utils.module.scss
```diff
@@ -0,0 +1,18 @@
+.itemDataGroup {
```
AshwinBhatkal, line 1: same comment as below

frontend/src/container/InfraMonitoringK8s/Base/K8sBaseList.tsx
```diff
@@ -0,0 +1,650 @@
...
+
+export type K8sExpandedRowProps<T> = {
+	record: K8sRenderedRowData;
+	entity: K8sCategory;
+	tableColumns: ColumnType<K8sRenderedRowData>[];
+	fetchListData: K8sBaseListProps<T>['fetchListData'];
+	renderRowData: K8sBaseListProps<T>['renderRowData'];
+};
+
+function K8sExpandedRow<T>({
```
AshwinBhatkal: can you move this to different file?

frontend/src/container/InfraMonitoringK8s/InfraMonitoringK8s.tsx
H4ad: This code renders the quick filter button expand:
image

> Why not name it controlListPrefix?

`controlListPrefix` is the name where it will be, the code itself is just to render filter button, the name seems appropriate.

> component itself instead of in this component?

These components works without quick filters, I don't want to include properties that does not matter for the component functionality directly.

It the same idea behind extracting the modal to another component instead of having it inside the list.

frontend/src/container/InfraMonitoringK8s/Base/K8sBaseList.tsx
```diff
@@ -0,0 +1,650 @@
...
+							</Button>
+						</div>
+					)}
+				</div>
+			)}
+		</div>
+	);
+}
+
+export function K8sBaseList<T>({
```
AshwinBhatkal, line 80: can this component be simplified? Hooks, callbacks, memos?

frontend/src/container/InfraMonitoringK8s/Base/K8sBaseList.tsx
```diff
@@ -0,0 +1,650 @@
...
+									openRecordInNewTab(rowRecord);
+									return;
+								}
+								setSelectedItem(rowRecord.itemKey);
+							},
+							className: styles.expandedClickableRow,
+						})}
+					/>
+
+					{data?.total && data?.total > 10 && (
```
AshwinBhatkal: make 10 a const

frontend/src/container/InfraMonitoringK8s/Pods/table.config.tsx
H4ad: This returns the configs for the table, I wanted to have it as just `table.ts` but we have some jsx to render.

frontend/src/container/InfraMonitoringK8s/Pods/api.ts
H4ad: I will refactor all these APIs endpoints in following PRs, we need to migrate to OpenAPI generated files.

frontend/src/container/InfraMonitoringK8s/Base/K8sBaseList.tsx
H4ad: Can we do this in another PR? I didn't want to break all the +8 PRs already created, to move this out of this file, I also need to move types to another file due to dependency of K8sBaseList (has K8sRenderedRowData) -> K8sExpandedRow -> K8sRenderedRowData

frontend/src/container/InfraMonitoringK8s/Base/K8sBaseList.tsx
H4ad: Don't think so, I expect to have more improvements after migrating to the new table component & generated api code.

frontend/src/container/InfraMonitoringK8s/Base/K8sBaseList.tsx
H4ad: I also want to remove all the CSS and port to css modules, so I can cleanup in that moment.

H4ad:
TODO:
- [x] Standardize css classes names (delayed until we standardize css modules)
- [x] Extract K8sExpandedRow to own file
- [x] table.tsx -> table.config.tsx
- [x] Investigate type for `props.filters?.items.reduce<typeof props.filters.items>`
- [x] Cleanup dead-code
- [x] Test on Staging

AshwinBhatkal (dismissed):

H4ad:
Found two minor bugs while testing not related to the current changes:
- https://github.com/SigNoz/engineering-pod/issues/4681
- https://github.com/SigNoz/engineering-pod/issues/4680

AshwinBhatkal (approved):

PR #10805, opened 2026-04-01
feat(infra-monitoring): v2 hosts list api

nikhilmantri0902:
## Pull Request

This PR builds a v2 hosts list API using the new modular architecture (pkg/modules/inframonitoring/) with query builder v5 types, proper request validation, and adds disk usage (filesystem) as a new sortable column.

- Migrates hosts list to the pkg/modules/ architecture with clean separation: types, handler, module, query building
  - Uses query builder v5 (querybuildertypesv5) and the v5 querier's QueryRange.
  - Adds `disk_usage` as a new metric column and order-by option (v1 only supports cpu, memory, wait, load15)
  - Adds strict input validation (time range, limit bounds, order-by key validation) with proper error types
  - Adds OpenAPI spec generation and auto-generated frontend API types
  - Supports custom groupBy keys with a grouped_list response type. 
  - Ability to filter on status also using the key `filterByStatus` in the payload. 

---

Closes https://github.com/SigNoz/engineering-pod/issues/4528

- Tests added/updated:
- Unit tests for HostsListRequest.Validate() covering all validation branches — pkg/types/inframonitoringtypes/hosts_test.go
  - Unit tests for helper functions
  - Manual verification:
  - API endpoint tested against running ClickHouse cluster — verified query results match v1 for overlapping columns, and reasoned for values that may not be matching and added in below thread for this PR. 
  - Verified pagination, ordering, filtering by expression, and filtering by active/inactive status
 
- Edge cases covered:

  - Missing/never-reported metrics → returns requiredMetricsCheck with missing metric names
  - End time before retention boundary → returns endTimeBeforeRetention: true
  - Empty filter expressions, zero-result queries
  - Pagination backfill: groups present in metadata but missing from metric results are appended alphabetically after metric-ranked groups
  - Metric value unavailable for a host/group → field set to -1
  - Querier error propagation to end user if primary attribue or the group-by attributes not set in the metrics
---

 Blast radius: New v2 endpoint only — v1 API is untouched.
 Potential regressions: None — no existing code paths are modified. Changes to the querier interface/provider are additive.
 Rollback plan: Revert the PR; v1 API continues to serve all traffic.

---

| Field | Value |
|------|-------|
| Change Type | Feature  |
| Description | New v2 hosts list API ( POST /api/v2/infra-monitoring/hosts/list ) with disk usage column, input validation, query builder v5, and custom group-by support |

---

### 📋 Checklist
- [x] Manually tested
- [x] Breaking changes documented - N/A, no breaking changes
- [x] Backward compatibility considered - v1 API is unchanged

---

## 👀 Notes for Reviewers

- if payload key `filterByStatus` is not empty, then those hosts ( active/ inactive ) are appended to the filter expression by adding the expression `(filterExpr) AND host.name IN ('host.1', 'host.2')`
- Core query construction: `newHostsTableListQuery()` in hosts.go builds a composite query with 10 metric queries (A-I) and 4 formulas (F1-F4) for `CPU%, Memory%, Wait%, DiskUsage%, CPU Load`
- Metadata (e.g., os.type) is fetched separately from the time-series + samples table using argMax(tuple(...), unix_milli) to get latest attribute values per group. The total number of groups are length of the metadata result only, because metadata is essentially consuming user passed group by, user passed filters and fingerprint shortlisted from samples table in that time range, fetching all such groups in the time range.
  - Two-pass query strategy: first pass queries only the ordering metric to rank+paginate groups, second pass queries all metrics for just the page's groups
  - Pagination backfill: groups present in metadata but missing from the orderBy-metric's results are appended alphabetically after orderBy-metric-ranked groups
  - Active host detection: hosts with metrics reported in last 10 minutes are considered active (queries `distributed_metadata` table). Also getActiveHosts ignores custom groupBy — status is per-host, not per-group. Added a TODO that acknowledges this. When groupBy is something other than host.name (e.g., os.type), the active/inactive status in records is still per-host. For grouped_list responses where HostName is empty, status will always be HostStatusNone, making the status field meaningless. Will be taken up once correctly planned


nikhilmantri0902:
If a metric was never received, the querier.QueryRange throws an error for all the queries combined. 
Based on this, should the `getMetricsExistenceAndEarliestTime` [function](https://github.com/SigNoz/signoz/pull/10805/changes#diff-25de84ea10aa59a6f787a9b7ba365d246ab94a4989a1b3779c208dd97231556cR68) be made better such that we find out counts ( existence ) of all metrics required in the List APIs individually ? Today we find out about them collectively. Doing this makes sense, because if some metric is missing or its count is 0 we can clearly return early from the API without waiting for the querier to throw error in this scenario. 

Basically, evolve it to do the following: 
```
-- Proposed Query.

SELECT
    metric_name,
    count(*) AS cnt,
    min(first_reported_unix_milli) AS min_first_reported
FROM signoz_metrics.distributed_metadata
WHERE metric_name IN ('system.cpu.time', 'system.memory.usage', 'system.cpu.load_average.15m', 'system.filesystem.usage')
GROUP BY metric_name;

-- Result

   ┌─metric_name───┬───cnt─┬─min_first_reported─┐
1. │ system.filesystem.usage     │   123 │      1775047220928 │
2. │ system.cpu.load_average.15m │ 31137 │      1773014400288 │
3. │ system.memory.usage         │    21 │      1775047220928 │
4. │ system.cpu.time             │    60 │      1775047220848 │
   └────┴───────┴────────────────────┘
  ````
  
If a metric is missing above ( or count 0 )
The API will return early after this check. 
The response of the hosts list v2 api will change like this : 

```
// replace this
"sentAnyMetricsData": false
// with
"requiredMetricsCheck" : {
   "missingMetrics" : ["system.cpu.time"]
}
```
  
  
CC: @srikanthccv @namanv3 let me know of ur thots. 

This prevents the heavy querier queries execution if clearly one of the required metrics was never sent.

**Comment by namanv3** [comment]
Actually even the query range API return "early" by not actually forming the statement builder queries. With your change it does avoid an a few extra trips into the time series tables, and also gives a proper structure to the response with a list of missing metrics. The proper response structure can be a good enough reason in itself I think

**Comment by nikhilmantri0902** [comment]
Tested v1 and v2 hosts list APIs side by side with a 7-day window. Ratio metrics like [memory (C/D) and CPU (A/B)](https://github.com/SigNoz/signoz/blob/e6900dabc89233355a5b23451e73c22cdac8e835/pkg/modules/inframonitoring/implinframonitoring/hosts.go#L46) match closely. 
However, [load15 (no ratio formula)](https://github.com/SigNoz/signoz/blob/e6900dabc89233355a5b23451e73c22cdac8e835/pkg/modules/inframonitoring/implinframonitoring/hosts.go#L50) diverged significantly for 3 hosts in testing that had `non-uniform metric reporting` over the 7 day period. These were my localhosts. deltas of `20-34%` were seen for them with v2 compared to v1.

Both APIs use the same aggregation pattern: 
temporal average per bucket per fingerprint 
→ spatial sum across fingerprints 
→ reduce via average across all buckets. 
The difference comes down to step interval. 
  
v1 computes step explicitly in `GetHostList()` using [MinAllowedStepInterval](https://github.com/SigNoz/signoz/blob/e6900dabc89233355a5b23451e73c22cdac8e835/pkg/query-service/app/inframetrics/hosts.go#L429), which gives (end - start) / 300 / 1000 rounded to nearest 60 = 1980s for this case. 

In v2 we do not set an explicit step; the v5 querier auto-fills it using [RecommendedStepIntervalForMetric](https://github.com/SigNoz/signoz/blob/e6900dabc89233355a5b23451e73c22cdac8e835/pkg/querybuilder/time.go#L109) , which starts with the same formula and gets 1980s, but then applies a 30-minute alignment rule for time ranges >= 7 days, capping this to the nearest multiple of 1800, in this case, 1800 itself. 

To confirm this difference was because of stepInterval difference, Running v2's exact SQL query with only the step changed from 1800 to 1980 ( as decided by v1 for this time range) reproduced v1's values exactly, confirming step interval as the sole cause.

CC: @namanv3 @srikanthccv wanted to call this difference out. Let me know if this makes sense.

**Comment by namanv3** [comment]
If the user runs that metric on their own in metrics explorer or in a dashboard of their own, because v5's 1800s step will be used, we should show the same result in the infra monitoring page as well I feel. So even if v4 used something else, it should be fine?

**Comment by nikhilmantri0902** [comment]
Yes, @namanv3. The experience and the result should be the same in infra-monitoring and dashboard places. so letting v5 decide stepInterval internally.

**Comment by nikhilmantri0902** [comment]
Calling out one more point, the `active/inactive` status in records is still per-host. For grouped_list responses where HostName is empty ( or `host.name` not part of custom group-bys), status will be "" (empty string). This behaviour is same as v1 API. 

We had discussed doing planning that in such custom group-by scenarios, the api should actually return count of active and inactive hosts in that grouping. This can be taken up in a follow up PR, added questions to be first answered here: https://github.com/SigNoz/engineering-pod/issues/4609

File: pkg/apiserver/signozapiserver/inframonitoring.go [hunk 1]
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func (provider *provider) addInfraMonitoringRoutes(router *mux.Router) error {
+	if err := router.Handle("/api/v2/infra_monitoring/hosts", handler.New(
+		provider.authZ.ViewAccess(provider.infraMonitoringHandler.HostsList),
+		handler.OpenAPIDef{
+			ID:                  "HostsList",
+			Tags:                []string{"infra-monitoring"},
+			Summary:             "List Hosts for Infra Monitoring",
+			Description:         "This endpoint returns a list of hosts along with other information for each of them",
```
namanv3: nit: should tell what all info is covered

pkg/modules/inframonitoring/config.go
```diff
@@ -0,0 +1,33 @@
+package inframonitoring
+
+import (
+	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/factory"
+)
+
+type Config struct {
```
namanv3, line 8: this is the same as metrics explorer config right? As of now do we need a separate count of threads for metrics explorer and this page? Or have you duplicated it here so that if in future we need more things here we can add them easily?

pkg/types/inframonitoringtypes/hosts.go
```diff
@@ -0,0 +1,119 @@
...
+
+	if !req.FilterByStatus.IsZero() && req.FilterByStatus != HostStatusActive && req.FilterByStatus != HostStatusInactive {
+		return errors.NewInvalidInputf(errors.CodeInvalidInput, "invalid filter by status: %s", req.FilterByStatus)
+	}
+
+	return nil
+}
+
+// UnmarshalJSON validates input immediately after decoding.
+func (req *HostsListRequest) UnmarshalJSON(data []byte) error {
```
namanv3: this is not used in the handler

pkg/modules/inframonitoring/implinframonitoring/module.go
```diff
@@ -0,0 +1,147 @@
...
+		querier:                querier,
+		fieldMapper:            fieldMapper,
+		condBuilder:            condBuilder,
+		logger:                 providerSettings.Logger,
+		config:                 cfg,
+	}
+}
+
+func (m *module) HostsList(ctx context.Context, orgID valuer.UUID, req *inframonitoringtypes.HostsListRequest) (*inframonitoringtypes.HostsListResponse, error) {
+	if err := req.Validate(); err != nil {
```
namanv3, line 51: shouldn't this be in the handler?

pkg/modules/inframonitoring/implinframonitoring/module.go
```diff
@@ -0,0 +1,147 @@
...
+			Key: qbtypes.OrderByKey{
+				TelemetryFieldKey: telemetrytypes.TelemetryFieldKey{
+					Name: "cpu",
+				},
+			},
+			Direction: qbtypes.OrderDirectionDesc,
+		}
+	}
+
+	if err := m.validateOrderBy(req.OrderBy, orderByToHostsQueryNames); err != nil {
```
namanv3: this should be a part of the validation right?

pkg/modules/inframonitoring/implinframonitoring/module.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil, err
+	}
+
+	// default to host name group by
+	if len(req.GroupBy) == 0 {
+		req.GroupBy = []qbtypes.GroupByKey{hostNameGroupByKey}
```
namanv3, line 71: I feel this kind of defaults are better resolved in a custom unmarshall function, keeps the module layer clean, what do you think?

pkg/modules/inframonitoring/implinframonitoring/commonmethods.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+// getMetricsExistenceAndEarliestTime checks which of the given metric names have been
+// reported. It returns a list of missing metrics (those not found or with zero count)
+// and the earliest first-reported timestamp across all present metrics.
+// When all metrics are missing, minFirstReportedUnixMilli is 0.
+// TODO(nikhilmantri0902, srikanthccv): This method was designed this way because querier errors if any of the metrics
+// in the querier list was never sent, the QueryRange call throws not found error. Modify this method, if QueryRange
+// behaviour changes towards this.
+func (m *module) getMetricsExistenceAndEarliestTime(ctx context.Context, metricNames []string) ([]string, uint64, error) {
```
namanv3: given that you've written this method as a generic one and not tied to infra monitoring, this can go to telemetryMetaStore I think

pkg/modules/inframonitoring/implinframonitoring/module.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil, err
+	}
+	if len(missingMetrics) > 0 {
+		resp.RequiredMetricsCheck = inframonitoringtypes.RequiredMetricsCheck{MissingMetrics: missingMetrics}
+		resp.Records = []inframonitoringtypes.HostRecord{}
+		resp.Total = 0
+		return resp, nil
+	}
+	if req.End < int64(minFirstReportedUnixMilli) {
+		resp.EndTimeBeforeRetention = true
```
namanv3, line 91: I'm not aware of this, is the first_time_reported_unix_milli updated periodically based on retention period? Like if we start collecting a metric on 1Apr, retention is 1 week, and if I check on 25Apr, then will I see first_time_reported_unix_milli as 1Apr or 18Apr?

pkg/modules/inframonitoring/implinframonitoring/hosts.go
```diff
@@ -0,0 +1,492 @@
...
+	defer rows.Close()
+
+	activeHosts := make(map[string]bool)
+	for rows.Next() {
+		var hostName string
+		if err := rows.Scan(&hostName); err != nil {
+			return nil, err
+		}
+		if hostName != "" {
+			activeHosts[hostName] = true
```
namanv3, line 345: nit: if the value is always true why return a map, we can keep it a list

namanv3: actually not a nit cuz later on you do convert the map to a list

namanv3 (commented):
first pass, mostly related to the structuring of the code and maybe not the actual logic itself, will check up a bit more on how infra monitoring works to do that

pkg/modules/inframonitoring/implinframonitoring/hosts.go
nikhilmantri0902: No we later use this map to assign status to the current hostname in [buildHostRecords](https://github.com/SigNoz/signoz/blob/e2b1b73e875397a7e186be4c1f97c0e370a0b636/pkg/modules/inframonitoring/implinframonitoring/hosts.go#L402)

pkg/modules/inframonitoring/config.go
nikhilmantri0902: Kept it seperate from metrics explorer config for independent updates now and later.

pkg/modules/inframonitoring/implinframonitoring/module.go
nikhilmantri0902: It is in the handler — binding.JSON.BindBody calls json.Decode which triggers HostsListRequest.UnmarshalJSON, and that calls Validate() at the end. So any validation error
  (including the orderBy check) is returned as an invalid_input error from the handler itself before we ever reach the module. The req.Validate() in the module is just a defensive
  check for any caller that bypasses the handler

pkg/modules/inframonitoring/implinframonitoring/module.go
nikhilmantri0902: Yes but used the qbTypes.OrderBy struct so how do I push into valid was unclear to me. Probably I will need to shift the orderbymap here in this package. Will do that now.

nikhilmantri0902: Done

pkg/apiserver/signozapiserver/inframonitoring.go
nikhilmantri0902: done

pkg/types/inframonitoringtypes/hosts.go
nikhilmantri0902: It is used — implicitly. binding.JSON.BindBody calls json.NewDecoder(...).Decode(&parsedReq), and since HostsListRequest implements the json.Unmarshaler interface, Go's
  encoding/json automatically calls UnmarshalJSON during decode. This is where Validate() runs, so all validation errors are caught at the handler level before reaching the module.

pkg/modules/inframonitoring/implinframonitoring/commonmethods.go
nikhilmantri0902: i discussed this one with srikanth, decided that let one more usecase come from other modules, then we can think of moving this.

pkg/modules/inframonitoring/implinframonitoring/module.go
nikhilmantri0902: Yes so we decided that instead of getting into if in a replacingAggregatingMergeTree, if the rows collapse first and how do their individual TTLs get applied to the collapsed row, we will keep the frontend messaging according to this data that we see.

pkg/modules/inframonitoring/implinframonitoring/module.go
nikhilmantri0902: oh can you elaborate a little?

namanv3: actually ditch for now, this pattern I am adding in a PR of mine, if approved I can share it here.

pkg/modules/inframonitoring/implinframonitoring/commonmethods.go
namanv3: In that case, can you add a comment saying this so that when we get a use case in the future, we don't have to test anything to figure out that this method is nothing specific to infra monitoring?

pkg/modules/inframonitoring/implinframonitoring/module.go
namanv3: didn't get this 😅

namanv3: discussed on call

pkg/types/inframonitoringtypes/hosts.go
```diff
@@ -0,0 +1,145 @@
...
+	HostsOrderByMemory,
+	HostsOrderByWait,
+	HostsOrderByDiskUsage,
+	HostsOrderByLoad15,
+}
+
+type HostsListRequest struct {
+	Start          int64                `json:"start"`
+	End            int64                `json:"end"`
+	Filter         *qbtypes.Filter      `json:"filter"`
```
namanv3: Filter can be
```
type HostFilter struct {
    *qbtypes.Filter
    FilterByStatus HostStatus 
}
```
Just check if this can work

pkg/modules/inframonitoring/implinframonitoring/module.go
```diff
@@ -0,0 +1,143 @@
...
+		resp.Records = []inframonitoringtypes.HostRecord{}
+		resp.Total = 0
+		return resp, nil
+	}
+
+	// Determine active hosts: those with metrics reported in the last 10 minutes.
+	activeHostsMap, err := m.getActiveHosts(ctx, hostsTableMetricNamesList, hostNameAttrKey)
```
namanv3: discussed on call that this separate CH query is needed cuz we want to use query builder query later on, and that does not support sub queries yet. Let's add a todo for this, for when that sub query support is added

nikhilmantri0902: done added

pkg/types/inframonitoringtypes/hosts.go
nikhilmantri0902: Yes it did work, replaced in the struct and also in the API contracts doc.

pkg/modules/inframonitoring/implinframonitoring/commonmethods.go
nikhilmantri0902: done

pkg/modules/inframonitoring/implinframonitoring/commonmethods.go
```diff
@@ -0,0 +1,277 @@
...
+		)
+	}
+
+	innerSB.Select(innerSelectCols...)
+	innerSB.From(fmt.Sprintf("%s.%s", telemetrymetrics.DBName, distributedTableName))
+	innerSB.Where(
+		innerSB.In("metric_name", sqlbuilder.List(metricNames)),
+		innerSB.GE("unix_milli", adjustedStart),
+		innerSB.L("unix_milli", adjustedEnd),
+		fmt.Sprintf("fingerprint IN (%s)", innerSB.Var(fpSB)),
```
namanv3: one thing here, the query will be something like
```
SELECT ...
FROM distributed samples table
WHERE
  ...
  ...
  AND fingerprint IN (SELECT ... FROM distributed time series table)
...
```
ideally, the inner query should have a local table. The main query fans out the query to all CH nodes because its the distributed samples table, and then if the inner query is also a distributed table, then that means each CH node again calls each CH node for the inner query.

Because we can assume that a sample will have its fingerprint's time series data in the same node, we don't need to fan out for the inner query

@srikanthccv I'm almost sure that this is the case but let me know if you think there's anything missing in what I'm saying

pkg/modules/inframonitoring/implinframonitoring/module.go
```diff
@@ -0,0 +1,145 @@
...
+
+	metadataMap, err := m.getHostsTableMetadata(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	if metadataMap == nil {
+		metadataMap = make(map[string]map[string]string)
+	}
+
+	resp.Total = len(metadataMap)
```
namanv3, line 120: total is supposed to be total groups and not total hosts? Just confirming, either way is fine as such

pkg/modules/inframonitoring/implinframonitoring/commonmethods.go
srikanthccv: @namanv3 you are correct.

nikhilmantri0902: https://github.com/SigNoz/signoz/pull/10805/changes/BASE..82d236742fd83c2f2e8c60494339a31664cfa3a4#diff-25de84ea10aa59a6f787a9b7ba365d246ab94a4989a1b3779c208dd97231556cR158 

Yes Check above its local table only

pkg/modules/inframonitoring/implinframonitoring/module.go
nikhilmantri0902: Yes and the metadataMap is `map[string]map[string]string`. The outermost string is a compositeKey formed by joining all group by columns values.

nikhilmantri0902: So total groups.

namanv3:
LGTM as such, is there anything else pending to be added here?

pkg/modules/inframonitoring/implinframonitoring/commonmethods.go
nikhilmantri0902: https://www.notion.so/signoz/Backend-Details-31ffcc6bcd1980e6b70cc6c23cf44bbb?source=copy_link#31ffcc6bcd1980bc982ceb38160e786e

pkg/modules/inframonitoring/implinframonitoring/hosts.go
```diff
@@ -0,0 +1,494 @@
...
+}
+
+var hostsTableMetricNamesList = []string{
+	"system.cpu.time",
+	"system.memory.usage",
+	"system.cpu.load_average.15m",
+	"system.filesystem.usage",
+}
+
+var hostAttrKeysForMetadata = []string{
```
namanv3: how is this list formed?

nikhilmantri0902: This list is nothing but the metric names being used in the list api to power the various columns

pkg/types/inframonitoringtypes/hosts_constants.go
```diff
@@ -0,0 +1,37 @@
...
+
+func (HostStatus) Enum() []any {
+	return []any{
+		HostStatusActive,
+		HostStatusInactive,
+		HostStatusNone,
+	}
+}
+
+const (
```
namanv3, line 23: why not use a type for this?

pkg/modules/inframonitoring/implinframonitoring/hosts_constants.go
```diff
@@ -0,0 +1,279 @@
...
+				// Query A: CPU usage logic (non-idle)
+				{
+					Type: qbtypes.QueryTypeBuilder,
+					Spec: qbtypes.QueryBuilderQuery[qbtypes.MetricAggregation]{
+						Name:   "A",
+						Signal: telemetrytypes.SignalMetrics,
+						Aggregations: []qbtypes.MetricAggregation{
+							{
+								MetricName:       "system.cpu.time",
+								Temporality:      metrictypes.Cumulative,
```
namanv3: I'd suggest not sending temporality cuz querier will fetch it anyway. Also it is likely that a customer is using delta I think

pkg/modules/inframonitoring/implinframonitoring/module.go
```diff
@@ -0,0 +1,163 @@
...
+	// Determine active hosts: those with metrics reported in the last 10 minutes.
+	// Compute the cutoff once so every downstream query/subquery agrees on what "active" means.
+	sinceUnixMilli := time.Now().Add(-10 * time.Minute).UTC().UnixMilli()
+	activeHostsMap, err := m.getActiveHosts(ctx, hostsTableMetricNamesList, hostNameAttrKey, sinceUnixMilli)
+	if err != nil {
+		return nil, err
+	}
+
+	// this check below modifies req.Filter by adding `AND active hosts filter` if req.FilterByStatus is set.
+	if m.applyHostsActiveStatusFilter(req, activeHostsMap) {
```
namanv3, line 109: what's the reason for early return here when true?

pkg/modules/inframonitoring/implinframonitoring/hosts.go
```diff
@@ -0,0 +1,354 @@
...
+
+	activeHosts := make([]string, 0, len(activeHostsMap))
+	for host := range activeHostsMap {
+		activeHosts = append(activeHosts, fmt.Sprintf("'%s'", host))
+	}
+
+	if len(activeHosts) == 0 {
+		return req.Filter.FilterByStatus == inframonitoringtypes.HostStatusActive
+	}
+
```
namanv3, line 279: what about Len active hosts == Len total and req.Filter.FilterByStatus == inframonitoringtypes.HostStatusInactive?

pkg/modules/inframonitoring/implinframonitoring/module.go
namanv3: got it from the method definition but its not clear reading the module method

nikhilmantri0902: Ya so this method returns a boolean called shouldShortCircuit, which is true if the active hosts map is empty and the user has requested active hosts in `filterByStatus`. And so no active hosts present in the data and yet the request is for active hosts, thus return early. Makes sense @namanv3 ?

pkg/modules/inframonitoring/implinframonitoring/hosts.go
nikhilmantri0902: we dont have len (total hosts) in all the cases with us, specially when user uses custom group by attributes not having host.name in them. And so letting the query result take the right course of action after constructing the filter.

nikhilmantri0902: Sounds right @namanv3 ?

namanv3: okay yeah that happens later

pkg/modules/inframonitoring/implinframonitoring/hosts_constants.go
nikhilmantri0902: Done.

pkg/types/inframonitoringtypes/hosts_constants.go
nikhilmantri0902: leaving it for now as no enum listing needed on this as well as this is internal use-case constant only, not a part of request or response also. If we create a type here, we should create a type of metricNames also. If the usecase becomes clear with subsequent PRs of other sections of infra monitoring, I ll make a type for these. Sounds good @namanv3 ?

namanv3 (dismissed):

pkg/modules/inframonitoring/implinframonitoring/handler.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	orgID := valuer.MustNewUUID(claims.OrgID)
+
+	var parsedReq inframonitoringtypes.PostableHosts
+	if err := binding.JSON.BindBody(req.Body, &parsedReq); err != nil {
+		render.Error(rw, errors.WrapInvalidInputf(err, errors.CodeInvalidInput, "failed to parse request body"))
```
therealpandey:
```suggestion
		render.Error(rw, err)
```

No need to wrap anything. BindBody will take care of everything.

pkg/apiserver/signozapiserver/inframonitoring.go
```diff
@@ -0,0 +1,33 @@
...
+			Summary:             "List Hosts for Infra Monitoring",
+			Description:         "Returns a paginated list of hosts with key infrastructure metrics: CPU usage (%), memory usage (%), I/O wait (%), disk usage (%), and 15-minute load average. Each host includes its current status (active/inactive based on metrics reported in the last 10 minutes) and metadata attributes (e.g., os.type). Supports filtering via a filter expression, filtering by host status, custom groupBy to aggregate hosts by any attribute, ordering by any of the five metrics, and pagination via offset/limit. The response type is 'list' for the default host.name grouping or 'grouped_list' for custom groupBy keys. Also reports missing required metrics and whether the requested time range falls before the data retention boundary.",
+			Request:             new(inframonitoringtypes.PostableHosts),
+			RequestContentType:  "application/json",
+			Response:            new(inframonitoringtypes.Hosts),
+			ResponseContentType: "application/json",
+			SuccessStatusCode:   http.StatusOK,
+			ErrorStatusCodes:    []int{http.StatusBadRequest, http.StatusUnauthorized, http.StatusInternalServerError},
```
therealpandey:
```suggestion
			ErrorStatusCodes:    []int{http.StatusBadRequest},
```

Both the other status codes are already added by the openapi handler to everyone.

pkg/apiserver/signozapiserver/inframonitoring.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/SigNoz/signoz/pkg/types/inframonitoringtypes"
+	"github.com/gorilla/mux"
+)
+
+func (provider *provider) addInfraMonitoringRoutes(router *mux.Router) error {
+	if err := router.Handle("/api/v2/infra_monitoring/hosts", handler.New(
+		provider.authZ.ViewAccess(provider.infraMonitoringHandler.ListHosts),
+		handler.OpenAPIDef{
+			ID:                  "ListHosts",
+			Tags:                []string{"infra-monitoring"},
```
therealpandey:
```suggestion
			Tags:                []string{"inframonitoring"},
```

This affects how the directy is created in the frontend.

pkg/types/inframonitoringtypes/hosts.go
```diff
@@ -0,0 +1,117 @@
...
+	GroupBy []qbtypes.GroupByKey `json:"groupBy"`
+	OrderBy *qbtypes.OrderBy     `json:"orderBy"`
+	Offset  int                  `json:"offset"`
+	Limit   int                  `json:"limit"`
+}
+
+type HostFilter struct {
+	qbtypes.Filter `json:",inline"`
+	FilterByStatus HostStatus `json:"filterByStatus"`
+}
```
therealpandey, line 50: For all these, add `required: true` wherever necessary. This helps the spec generate more type safe code for the frontend.

therealpandey (changes requested):

pkg/modules/inframonitoring/implinframonitoring/handler.go
nikhilmantri0902: done

pkg/apiserver/signozapiserver/inframonitoring.go
nikhilmantri0902: done

pkg/apiserver/signozapiserver/inframonitoring.go
nikhilmantri0902: done

pkg/types/inframonitoringtypes/hosts.go
nikhilmantri0902: done

therealpandey (dismissed):

nikhilmantri0902:
Hi @namanv3, I pushed a commit above that makes `sorting of the groups` stable in case of tie in the values by ordering by their composite key in ascending. Can you check it once? thanks.

namanv3 (dismissed):

srikanthccv (dismissed):

therealpandey (dismissed):

therealpandey (approved):

namanv3 (approved):

srikanthccv (approved):

PR #10807, opened 2026-04-02
fix: Body Search warning with FTS in JSON Logs 

piyushsingariya:
## Pull Request

---

### 📄 Summary

This change targets to address Warning missed when direct text search is fired using `hello` or `'some text'`

image

Closes Signoz/engineering-pod#4497

---

### ✅ Change Type
_Select all that apply_

- [x] 🐛 Bug fix

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered:

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---


tushar-signoz:
How can user do full text search ?
Where Im coming from is that, the warning is loaded with terms, can we think of something more layman terms ?

If I search for `body contains 'xyz'`, then it makes sense to say body defaults to body.message
If I search for `xyz`, then warning doesn't make sense, what `body.message:string` ? IMO better would be to convey that this is not FTS.

nityanandagohain (dismissed):

nityanandagohain (approved):

therealpandey (approved):

PR #10812, opened 2026-04-02
fix(host-list): not showing refresh status & refresh interval queries overlaps

H4ad:
## Pull Request

---

This uses the new global time adapter to fix the issue with:
- feedback of loading
- query overlap when refresh interval is lower than time query
- query cancellation causing the data to never be loaded

Before:

https://github.com/user-attachments/assets/c22982ef-5517-4db1-87b6-e3ee82c70540

After:

> Feedback of loading on query refresh

https://github.com/user-attachments/assets/467d58eb-e2e5-4257-ab83-190effc724e3

> Handle query with duration larger than the refresh interval

https://github.com/user-attachments/assets/4bbdaeb5-3c2b-4d85-aa95-6e38fbdd730e

> Query invalidation when click on refresh

https://github.com/user-attachments/assets/63b750ba-6de9-4315-b3eb-a8d302e5681e

> When query fails

https://github.com/user-attachments/assets/80d8b6db-aa65-4ee2-a6f0-ef0ce223258d

Fixes https://github.com/SigNoz/engineering-pod/issues/3994
Fixes https://github.com/SigNoz/engineering-pod/issues/3989
Fixes https://github.com/SigNoz/engineering-pod/issues/3993

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

There's no feedback on loading of data and queries with larger duration than refresh time caused the query to never finish.

Due to the way we handle min/max in query identity, caused the query to be invalidated and signal was triggered to cancel previous query and start new one

Use new GlobalTimeAdapter.

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Host Metrics
- Potential regressions: -

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Now the Host Metrics on Infrastructure Monitoring show feedback on loading & do not keep loading forever when query duration is larger than refresh interval. |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/container/InfraMonitoringHosts/HostsListTable.tsx
+++ b/frontend/src/container/InfraMonitoringHosts/HostsListTable.tsx
@@ -198,7 +198,8 @@ export default function HostsListTable({
 		!isLoading &&
 		formattedHostMetricsData.length === 0 &&
 		(!sentAnyHostMetricsData || isSendingIncorrectK8SAgentMetrics) &&
-		!filters.items.length;
+		!filters.items.length &&
+		!endTimeBeforeRetention;
 
 	const showEndTimeBeforeRetentionMessage =
 		!isFetching &&


frontend/src/container/InfraMonitoringHosts/HostsList.tsx
```diff
@@ -62,56 +70,48 @@ function HostsList(): JSX.Element {
...
-			end: Math.floor(maxTime / 1000000),
-			orderBy,
-		};
-	}, [pageSize, currentPage, filters, minTime, maxTime, orderBy]);
-
-	const queryKey = useMemo(() => {
-		if (selectedHostName) {
-			return [
-				'hostList',
+	const selectedTime = useGlobalTimeStore((store) => store.selectedTime);
```
AshwinBhatkal: nit: use s or use store below. Just to have it uniform

frontend/src/container/InfraMonitoringHosts/HostsList.tsx
```diff
@@ -62,56 +70,48 @@ function HostsList(): JSX.Element {
...
+		[pageSize, currentPage, filters, orderBy, selectedTime],
+	);
 
-	const { data, isFetching, isLoading, isError } = useGetHostList(
-		query as HostListPayload,
-		{
-			queryKey,
-			enabled: !!query,
-			keepPreviousData: true,
+	const { data, isFetching, isLoading, isError } = useQuery<
```
AshwinBhatkal, line 94: can you see if we can use the auto generated hook for this in api/generated? and then use the auto generated query key function along with auto refresh query key

frontend/src/container/InfraMonitoringHosts/HostsList.tsx
```diff
@@ -62,56 +70,48 @@ function HostsList(): JSX.Element {
...
+		queryFn: ({ signal }) => {
+			const { minTime, maxTime } = getMinMaxTime();
+
+			const payload: HostListPayload = {
+				...baseQuery,
+				limit: pageSize,
+				offset: (currentPage - 1) * pageSize,
+				filters: filters ?? defaultFilters,
+				orderBy,
+				start: Math.floor(minTime / 1000000),
```
AshwinBhatkal: is this nano second multiplier?

AshwinBhatkal:
Can you add screen recordings for?
1. New query when existing query is being fetched
2. Failure of the query

frontend/src/container/InfraMonitoringHosts/HostsList.tsx
H4ad: There's no generated hook yet, these routes are not mapped by openapi

H4ad:
1. New query when existing query is being fetched

This one is the video of "Handle query with duration larger than the refresh interval". There's no way to queries to overlaps each other now because we use refresh interval & add load state on refresh button.

2. Added video and improved the UI (the old one was not good)

AshwinBhatkal:
can we have the failure case be something like a card in between the screen? Similar to when there is no data from the query

PR #10742, opened 2026-03-27
chore: update all absolute tokens to semantic tokens

YounixM:
## Pull Request

Migrate SCSS stylesheets across the entire frontend from hardcoded hex color values and absolute CSS design tokens (e.g., var(--bg-slate-500, #161922), var(--bg-ink-500), #fff) to semantic design tokens (e.g., var(--l1-border), var(--l1-background), var(--l1-foreground)).

Semantic tokens are theme-aware and carry intent (e.g., "layer 1 background") rather than raw color values. This ensures consistent theming across light/dark mode and makes future design system updates easier — changing a semantic token propagates everywhere without file-by-file edits.

### ✅ Change Type
_Select all that apply_

- [x] ♻️ Refactor

---
🧪 Testing Strategy
Tests added/updated: N/A (no logic changes)
Manual verification: Visual inspection across affected pages to confirm no regressions in light/dark mode
Edge cases covered: Pre-login pages (auth flows), trace details, and general app layout all covered across 4 commits
⚠️ Risk & Impact Assessment
Blast radius: ~361 SCSS files across components, containers, and pages
Potential regressions: If a semantic token is missing or misconfigured for a theme variant, the affected element would appear unstyled or use the wrong color. All tokens used here are part of the established design system.
Rollback plan: Revert the 4 commits (ae51a49, 73fcd45, 8a373bc, 365a8ae)

### 📋 Checklist
- [ ] Tests added or explicitly not required
- [x] Manually tested
- [ ] Breaking changes documented
- [ ] Backward compatibility considered

---

👀 Notes for Reviewers

The changes are purely token substitutions — no layout, spacing, or behavior changes.
The pattern applied: var(--old-absolute-token, #hexfallback) → var(--semantic-token) and bare hex values → var(--semantic-token).
Split across 4 commits for review clarity: general app, pre-login pages, trace details, and remaining components.

The views might look slightly different after this change. We will incrementally solve for this by collaborating with the design team.

---

PR #10814, opened 2026-04-02
fix: update border colors in External APIs for light mode

vincent067:
Hi team! 👋

I noticed the border colors in the External APIs page don't look quite right in light mode. The borders are using bg-slate-400 which works well for dark mode, but in light mode it makes the borders look a bit too harsh.

## Changes
- Updated .api-quick-filters-header border colors from bg-slate-400 to bg-slate-200 in light mode
- Updated .toolbar border-bottom color from bg-slate-400 to bg-slate-200 in light mode
- Updated .round-metric-tag border and background colors for better visibility in light mode

## Testing
- [x] Verified changes in light mode
- [x] Verified no impact on dark mode
- [x] Checked External APIs page renders correctly

The changes are minimal and only affect the .lightMode CSS class, so there's no risk of breaking anything in dark mode.

Let me know if you'd like any adjustments! 😊

Thanks for maintaining this awesome project! 🙏

### DIFF
--- a/frontend/src/container/ApiMonitoring/Explorer/Explorer.styles.scss
+++ b/frontend/src/container/ApiMonitoring/Explorer/Explorer.styles.scss
@@ -220,6 +220,21 @@
 }
 
 .lightMode {
+	.api-monitoring-page {
+		.api-quick-filter-left-section {
+			.api-quick-filters-header {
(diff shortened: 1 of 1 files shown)


H4ad:
Thanks for the PR but we are already working on a fix for that at #10742

PR #10817, opened 2026-04-03
fix: several panic errors

srikanthccv:
## Pull Request

Fixes https://github.com/SigNoz/engineering-pod/issues/4384
Fixes SigNoz/engineering-pod#4380
Fixes SigNoz/engineering-pod#4379
Fixes SigNoz/engineering-pod#4377
Fixes https://github.com/SigNoz/engineering-pod/issues/4206

### 1. StepIntervalForQuery panic for trace operator queries

`StepIntervalForQuery` didn't handle `QueryBuilderTraceOperator` in its type switch. When fill-gaps post-processing called it for a trace operator result, the name wasn't found in `stepsMap`, fell through to
  the formula path, `govaluate.NewEvaluableExpressionWithFunctions("Trace Operator")` returned nil (error silently discarded), and `expression.Vars()` panicked.

  Fix:
  - Added `QueryBuilderTraceOperator` case to the type switch in `StepIntervalForQuery`
  - Changed `StepIntervalForQuery` signature to return (int64, error) so govaluate parse failures are propagated instead of silently swallowed
     - Although this should never receive an invalid expression, since it's an independent path, rejecting instead of panicking
  - Added govaluate expression parsing in `QueryBuilderFormula.Validate()` so invalid expressions are rejected at validation time
  - Made `validateQueryEnvelope` call `spec.Validate()` for formulas instead of duplicating an inline empty-string check

### ApplyHavingClause panic for non-builder queries

  A ClickHouse SQL query (`/api/v3/query_range`) with `builderQueries: {}` (empty map) caused `ApplyHavingClause` to panic. `builderQueries != nil` passed (initialized empty map), then
  `builderQueries[result.QueryName]` returned nil for the missing key, and dereferencing `.DataSource` panicked.

  Fix: Extract the map lookup into a local variable and nil-check it before accessing fields.

### GetKeySelectors panic for metric queries in raw requests

  A raw request with disabled metric builder queries (no aggregations) and a formula referencing them caused `GetKeySelectors` to panic on `query.Aggregations[0]` (empty slice). Validation skipped the
  aggregation check for raw requests, but metric queries always need aggregations since there is no raw form for metrics.

  Fix: Reject metric signal queries in raw/trace request types at validation time in `QueryRangeRequest.Validate()` with a clear error message: _"raw request type is not supported for metric queries"_.

### 📝 Changelog

N/A


pkg/queryparser/queryparserimpl.go
```diff
@@ -23,7 +24,15 @@ func New(settings factory.ProviderSettings) QueryParser {
 	}
 }
 
-func (p *queryParserImpl) AnalyzeQueryFilter(ctx context.Context, queryType qbtypes.QueryType, query string) (*queryfilterextractor.FilterResult, error) {
+func (p *queryParserImpl) AnalyzeQueryFilter(ctx context.Context, queryType qbtypes.QueryType, query string) (result *queryfilterextractor.FilterResult, err error) {
+	// the third-party clickhouse sql parser can panic on certain inputs, recover gracefully
+	defer func() {
+		if r := recover(); r != nil {
+			result = nil
+			err = errors.NewInternalf(errors.CodeInternal, "failed to analyze query filter: %s", fmt.Sprint(r))
```
nityanandagohain, line 32: should we monitor this with custom error code ?

srikanthccv: This error is not surfaced to the user at the moment. `QueryParser` component returns error but the consumer decides when to throw further or handle it. They are logged here https://github.com/SigNoz/signoz/blob/a71006b662edfd7b860d7b47253a725d01b3d8e1/pkg/modules/dashboard/impldashboard/module.go#L307-L313

Is that something similar to what you were suggesting?

srikanthccv: Let me know otherwise, will incorporate in follow-up PRs.

PR #10837, opened 2026-04-04
test(integration): add test cases for new user APIs

vikrantgupta25:
### 📄 Summary

- added test cases for `/v2/users` APIs
- added test case to mimic upstream provisioning flow 
- only run integration ci with `wal` mode for sqlite as it's the default now. 

closes: https://github.com/SigNoz/platform-pod/issues/1439
closes: https://github.com/SigNoz/platform-pod/issues/528

### DIFF
--- a/tests/integration/fixtures/authutils.py
+++ b/tests/integration/fixtures/authutils.py
@@ -1,7 +1,7 @@
 """Reusable helpers for user API tests."""
 
 from http import HTTPStatus
-from typing import Dict, Optional
+from typing import Dict
 
 import requests
 
@@ -10,7 +10,7 @@ from fixtures import types
 USERS_BASE = "/api/v2/users"
 
 
-def invite_and_activate_user(
+def create_active_user(
     signoz: types.SigNoz,
     admin_token: str,
     email: str,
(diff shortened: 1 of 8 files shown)


tests/integration/fixtures/authutils.py
```diff
@@ -0,0 +1,117 @@
...
+from typing import Dict, Optional
+
+import requests
+
+from fixtures import types
+
+USERS_BASE = "/api/v2/users"
+
+
+def invite_and_activate_user(
```
therealpandey:
```suggestion
def create_active_user(
```

tests/integration/fixtures/authutils.py
```diff
@@ -0,0 +1,117 @@
...
+def find_user_with_roles_by_email(
+    signoz: types.SigNoz, token: str, email: str
+) -> Optional[Dict]:
+    """Find a user by email and return UserWithRoles (user fields + userRoles).
+
+    Returns None if the user is not found.
+    """
+    user = find_user_by_email(signoz, token, email)
+    if user is None:
+        return None
```
therealpandey: ```suggestion```

Fail this if user was not found. Or change the name of this function. Conflicting with `find_user_by_email`

tests/integration/src/passwordauthn/01_register.py
```diff
@@ -216,30 +174,77 @@ def test_revoke_invite_and_register(
     assert response.status_code in (HTTPStatus.BAD_REQUEST, HTTPStatus.NOT_FOUND)
 
 
-def test_self_access(
+def test_provision_user(
```
therealpandey, line 174:
```suggestion
```

Not needed, the `create_active_user` fixture is doing exactly the same steps.

tests/integration/src/passwordauthn/04_role.py
```diff
@@ -55,32 +67,335 @@ def test_change_role(
...
+    response = requests.get(
+        signoz.self.host_configs["8080"].get(f"/api/v2/users/{user_id}/roles"),
+        headers={"Authorization": f"Bearer {admin_token}"},
+        timeout=5,
+    )
+    assert response.status_code == HTTPStatus.OK
+    roles = response.json()["data"]
+    names = {r["name"] for r in roles}
+    assert "signoz-admin" in names
+    assert "signoz-editor" in names
```
therealpandey, line 170:
```suggestion
```

There is a fixture which does exactly this? Let's reuse the fixture.

tests/integration/src/passwordauthn/01_register.py
vikrantgupta25: nope it's not, the test_provision user is using `/v1/user` APIs to list the users which is followed by upstream.

tests/integration/src/passwordauthn/04_role.py
vikrantgupta25: that is on userWithRole struct, this is on just role struct.

PR #10842, opened 2026-04-06
feat: adding API for getting connection credentials

swapnil-signoz:
### 📄 Summary
- Adding API handler and implementation for fetching Signoz Credentials
- Why: to integrate with the cloud provider agent requires Signoz API and Ingestion creds, which depending on the deployment may (cloud) or may not (enterprise self-hosted) deduce these credentials. So the idea is the try and deduce the credentials first. If the creds are deduced use it otherwise ask user to provide missing creds


swapnil-signoz:
Converting this to draft, until base branch of this PR is merged to main

PR #10860, opened 2026-04-07
fix: add canonical name aliasing for select fields

tushar-signoz:
## Summary
- Alias SELECT columns using canonical names (`context;name;datatype`) so the querier can correctly map result columns back to their telemetry field keys, even when multiple fields share the same short name.
- Handle column-name collisions in raw queries by falling back to `key.Text()` when two columns resolve to the same simple name.
- Add `MaybeAddDefaultSelectFields` helper to prepend default fields (id, timestamp) to list queries only when not already present.
- Propagate non-`ErrColumnNotFound` errors from `FieldFor` in `CollisionHandledFinalExpr` instead of silently swallowing them.

## Test plan
- [x] Updated existing unit tests in `json_stmt_builder_test.go` and `stmt_builder_test.go` to match new aliased column names
- [ ] Verify logs list view, time-series, and scalar queries return correct column names end-to-end


pkg/querier/consume.go
```diff
@@ -121,28 +121,30 @@ func readAsTimeSeries(rows driver.Rows, queryWindow *qbtypes.TimeRange, step qbt
 		)
 
 		for idx, ptr := range slots {
-			name := colNames[idx]
+			key := telemetrytypes.GetFieldKeyFromCanonicalName(colNames[idx])
```
nityanandagohain, line 124: This is running (rows * column) number of times we should move it out and make it run just column number of times.

pkg/querybuilder/collision.go
```diff
@@ -298,3 +298,42 @@ func AdjustKeysForAliasExpressions[T any](query *qbtypes.QueryBuilderQuery[T], r
...
 }
+
+// maybeAddDefaultSelectFields prepends default fields (id, timestamp) to query.SelectFields
+// if they are not already covered by an existing entry.
+// A select field covers a default field when the names match and the select field's
+// context is either unspecified or equal to the default field's context.
+// 3 cases:
+// 1. log.timestamp:number in select fields; we skip
+// 2. timestamp is present in select fields; we skip
+// 3. attribute.timestamp in select fields; we add it.
```
nityanandagohain, line 309: but in the UI we don't have support for this right as of now? so with this change it will return timestamp as well as timestamp in attributes. Did you check if the existing UI breaks and needs some handling before we merge this ?

pkg/querybuilder/collision.go
```diff
@@ -298,3 +298,42 @@ func AdjustKeysForAliasExpressions[T any](query *qbtypes.QueryBuilderQuery[T], r
...
+func MaybeAddDefaultSelectFields[T any](
+	query *qbtypes.QueryBuilderQuery[T],
+	defaultSelectFields []telemetrytypes.TelemetryFieldKey,
+) {
+	var toAdd []telemetrytypes.TelemetryFieldKey
+	for i := range defaultSelectFields {
+		if !isDefaultSelectFieldCovered(&defaultSelectFields[i], query.SelectFields) {
```
nityanandagohain, line 316: nit: create a selectFields map will be better ?

pkg/querybuilder/fallback_expr.go
```diff
@@ -60,7 +60,12 @@ func CollisionHandledFinalExpr(
 	}
 
 	fieldExpression, fieldForErr := fm.FieldFor(ctx, startNs, endNs, field)
-	if errors.Is(fieldForErr, qbtypes.ErrColumnNotFound) {
+	if fieldForErr != nil {
```
nityanandagohain, line 63: so the previous change of running the else case if there was any other error was incorrect ?

pkg/telemetrylogs/statement_builder.go
```diff
@@ -266,33 +266,35 @@ func (b *logQueryStatementBuilder) buildListQuery(
...
+		// Select all columns by default with dummy canonical aliases
+		sb.SelectMore(fmt.Sprintf("`%s` AS `;%s;`", LogsV2TraceIDColumn, LogsV2TraceIDColumn))
+		sb.SelectMore(fmt.Sprintf("`%s` AS `;%s;`", LogsV2SpanIDColumn, LogsV2SpanIDColumn))
+		sb.SelectMore(fmt.Sprintf("`%s` AS `;%s;`", LogsV2TraceFlagsColumn, LogsV2TraceFlagsColumn))
+		sb.SelectMore(fmt.Sprintf("`%s` AS `;%s;`", LogsV2SeverityTextColumn, LogsV2SeverityTextColumn))
+		sb.SelectMore(fmt.Sprintf("`%s` AS `;%s;`", LogsV2SeverityNumberColumn, LogsV2SeverityNumberColumn))
+		sb.SelectMore(fmt.Sprintf("`%s` AS `;%s;`", LogsV2ScopeNameColumn, LogsV2ScopeNameColumn))
+		sb.SelectMore(fmt.Sprintf("`%s` AS `;%s;`", LogsV2ScopeVersionColumn, LogsV2ScopeVersionColumn))
+		sb.SelectMore(fmt.Sprintf("`%s` AS `;%s;`", LogsV2BodyColumn, LogsV2BodyColumn))
+		if querybuilder.BodyJSONQueryEnabled {
```
nityanandagohain, line 287: in this case I don't think we should select both, we should select old or new. Else handling might be required somewhere else ?

nityanandagohain (commented):
Thanks for the PR, have some comments and  we should definately add integration test in this PR

pkg/querybuilder/collision.go
```diff
@@ -298,3 +298,42 @@ func AdjustKeysForAliasExpressions[T any](query *qbtypes.QueryBuilderQuery[T], r
 	}
 	return actions
 }
+
+// maybeAddDefaultSelectFields prepends default fields (id, timestamp) to query.SelectFields
+// if they are not already covered by an existing entry.
+// A select field covers a default field when the names match and the select field's
+// context is either unspecified or equal to the default field's context.
+// 3 cases:
+// 1. log.timestamp:number in select fields; we skip
```
nityanandagohain, line 307: If by mistake someone add log.timestamp:string it will be converted to number right, can you add that we dont consider the type and why.

Also add 1/2 unit tests.

nityanandagohain (commented):
* What happens if there is a field `field;name` . Didn't see a testcase for that

pkg/querybuilder/collision.go
tushar-signoz: This is specifically for api users. signoz UI doesn't specify selectFields

nityanandagohain: In trace ui it does specify select fields right ?

tushar-signoz:
Fixed it in these 2 PRs for logs and traces
#12091 
#12134

PR #10861, opened 2026-04-07
feat: improve perf for queries with empty resource filter

nityanandagohain:
### 📄 Summary
* PrepareWhereClause returns `nil, nil` when there is no actual filter i.e for  filter where all condition return to empty string.
* `addConditions` of resource statement builder handles this and returns empty resource filter.

---

#### Issues closed by this PR
Fixes https://github.com/SigNoz/engineering-pod/issues/4033

---

### 🧪 Testing Strategy
* Unit tests updated

---

- Blast radius: All queries with no resource filters
- Potential regressions: None

---

### 📋 Checklist
- [x] Manually tested
---


pkg/querybuilder/where_clause_visitor.go
```diff
@@ -166,11 +166,10 @@ func PrepareWhereClause(query string, opts FilterExprVisitorOpts) (*PreparedWher
...
 	}
 
-	// The visitor returns exactly SkipConditionLiteral (never a substring) when
-	// there are no evaluable conditions; replace it with the no-op SQL literal.
-	// TODO(nitya): In this case we can choose to ignore resource_filter_cte
+	// Return nil so callers can skip the
+	// entire CTE/subquery rather than emitting WHERE clause that select all the rows
 	if cond == "" || cond == SkipConditionLiteral {
-		cond = TrueConditionLiteral
+		return nil, nil //nolint:nilnil
```
nityanandagohain, line 172: I had to go ahead with this where the condition is nil and there is no error. We handle this wherever we call `PrepareWhereClause`. 

Though the linter throwed this error 
```
pkg/querybuilder/where_clause_visitor.go:172:3: return both a `nil` error and an invalid value: use a sentinel error instead (nilnil)
``` 
so I had to add the nolint. Let me know what you think. @tushar-signoz  @srikanthccv

tushar-signoz: Hmm. Doesn't look elegant. 
Either raise an error "ErrNoOPCondition" 
or 
Check at the upstream if `statement_generated == "sb.Where(TrueConditionLiteral)"`

nityanandagohain: errorNoOpCondition looks clean but not sure if it's the right way to go ahead with this since it's not actually an error.

srikanthccv: I don't have any strong opinion here.

That being said, one thing to note here is, there is a tendency to do `if err != nil {return err}`, which means err could potentially reach the user if the users of the `PrepareWhereClause` don't handle it. So if we go ahead with return error, we should ensure (I imagine there is no automated or lint stage thing we can do), so through reviews or tests that the `ErrNoOPCondition` never reaches user.

nityanandagohain: Yeah correct, then I am keen on keeping this as it is i.e return nil, nil. Which means the query is not there and no error as well.

tushar-signoz:
Upto you to update `auditlogs` module @nityanandagohain. Otherwise looks good.  

Edit: Also do a sanity check where 'PrepareWhereClause' function is used for eg: agg_rewriter. telemetrymeter etc

tushar-signoz (commented):
LGTM. Lets do a sanity check.

nityanandagohain:
@tushar-signoz  I have done the changes for auditlogs and handled the nil where caluse in few other places. All the other places have handled the case where the where clause can be nil

tushar-signoz:
Reason for querier integration tests being flaky is that. The test itself inserts some data and then immediately queries it. Problem is that insertion is async and possibly the query is made before insertion.

srikanthccv (dismissed):

srikanthccv (approved):

PR #10863, opened 2026-04-07
CI: JSON QB E2E Integration Suite

piyushsingariya:
## Pull Request

---

### 📄 Summary

This PR adds end-to-end integration tests for the JSON Query Builder (QB) — the `body_v2` / `body_promoted` column path introduced behind the `BODY_JSON_QUERY_ENABLED` feature flag. 

**Why a separate test package (`querier_json_body/`) instead of adding to `querier/`?**

The JSON QB requires two non-default environment configurations that the regular `querier` tests do not:

1. **SigNoz must start with `BODY_JSON_QUERY_ENABLED=true`** — this flag gates all `body_v2` query paths. Without it, `body.user.name = "x"` would not route through `json_condition_builder.go` at all.
2. **Schema migrations must run with `ENABLE_LOGS_MIGRATIONS_V2=1`** — this creates the `body_v2` and `body_promoted` columns in ClickHouse. Without it, the columns don't exist and every query would fail.

Both `signoz` and `migrator` are **package-scoped fixtures** managed by `dev.wrap()`. A package-scoped fixture is created once per Python package (directory) and shared across all tests in that package. This means:

- The `signoz` fixture in `src/querier/` starts SigNoz **without** `BODY_JSON_QUERY_ENABLED` — and that's correct for the existing querier tests.
- We cannot conditionally override a package-scoped fixture per-file within the same package. Pytest resolves fixture names at the package level, not the module level.
- A separate package with its own `conftest.py` is the **only** mechanism pytest provides to override package-scoped fixtures with different configuration.

This is the same pattern used by `src/rootuser/`, which overrides `signoz` with `SIGNOZ_USER_ROOT_ENABLED=True` for root user tests.

**Why skip on ClickHouse 25.5.6?**

The `body_v2` column uses ClickHouse's `JSON` data type with `dynamicElement()` access. This functionality requires ClickHouse versions newer than 25.5.6. In CI, we test against two ClickHouse versions (currently 25.5.6 and 25.12.5). Rather than hardcoding the "good" version (which can change), the `conftest.py` uses a `pytest_collection_modifyitems` hook to skip all tests when the `--clickhouse-version` is in the `UNSUPPORTED_CLICKHOUSE_VERSIONS` set. This means:

- On 25.5.6: all 7 test functions are SKIPPED (no container startup, no wasted CI time).
- On 25.12.5 (or any future version): tests run normally.
- Adding a new unsupported version only requires adding it to the set.

#### Issues closed by this PR

Closes https://github.com/SigNoz/engineering-pod/issues/3474

---

### ✅ Change Type

- [x] 🧪 Test-only
- [x] ♻️ Refactor (migrator factory extraction)

---

#### Test architecture

```
tests/integration/src/querier_json_body/
├── __init__.py
├── conftest.py                          # Overrides signoz + migrator fixtures, version skip hook
└── 01_logs_json_body_new_qb.py          # 46 test cases across 5 active functions
```

```
tests/integration/fixtures/migrator.py   # Extracted create_migrator() factory (mirrors create_signoz() pattern)
```

#### Fixture changes

**`fixtures/migrator.py`** — extracted `create_migrator()` factory function, mirroring the existing `create_signoz()` pattern in `fixtures/signoz.py`. The default `migrator` fixture now delegates to it. This enables `querier_json_body/conftest.py` to create a migrator variant with `ENABLE_LOGS_MIGRATIONS_V2=1` and a separate cache key (`migrator-json-body`), so `--reuse` / `--teardown` work correctly without colliding with the default migrator.

#### CI changes

Added `querier_json_body` to the matrix `src` list in `.github/workflows/integrationci.yaml`. On ClickHouse 25.5.6, all tests are skipped via the `pytest_collection_modifyitems` hook (no container startup, no wasted CI time). On 25.12.5+, tests run normally.

- Tests added/updated: 46 new E2E test cases in `src/querier_json_body/01_logs_json_body_new_qb.py`
- Manual verification: `pytest --collect-only` confirms 7 functions collected; `--clickhouse-version 25.5.6` confirms all SKIPPED; compilation verified
- Edge cases covered: type ambiguity, sparse fields, missing paths, structurally different logs, 6-hop deep nesting, non-array intermediate hops, hyphen keys, dual Array(JSON)+Array(Dynamic) branches

---

- Blast radius: **Zero** — test-only change. The `create_migrator()` refactor preserves the existing fixture's behavior byte-for-byte (the default fixture delegates to the factory with no overrides).
- Potential regressions: None. No production code changed. The migrator refactor is additive — the existing `migrator` fixture signature and behavior are unchanged.
- Rollback plan: Revert the commit.

---

### 📝 Changelog

N/A — internal test infrastructure only.

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

1. **The separate package is not optional.** Pytest package-scoped fixtures can only be overridden at the package boundary (`conftest.py` in a directory). There is no way to have two different `signoz` fixture configurations within the same `src/querier/` package. This is the same pattern as `src/rootuser/`.

2. **The `create_migrator()` extraction follows `create_signoz()` exactly.** Compare `fixtures/migrator.py` with `fixtures/signoz.py` — same factory pattern, same `cache_key` + `env_overrides` parameters, same `dev.wrap()` lifecycle. The default `migrator` fixture delegates to `create_migrator()` with no overrides, so existing tests are unaffected.

3. **ClickHouse version gating uses a skip hook, not a CI matrix exclude.** This means CI still _runs_ the test suite on 25.5.6 (it just skips instantly). The advantage: if someone accidentally removes the skip logic, CI will catch the failure rather than silently never running the tests.


srikanthccv:
Hi @piyushsingariya , please ensure we have the integration tests added for the failures mentioned in the https://github.com/SigNoz/engineering-pod/issues/3472. They will fail till the bugs are addressed and we will only get this merged once those cases are covered.

piyushsingariya:
@srikanthccv Can we merge this first so these can come in action with my coming PR? I can add additional tests there?

tests/integration/fixtures/jsontypeexporter.py
```diff
@@ -0,0 +1,437 @@
...
+    Union,
+)
+
+import numpy as np
+import pytest
+
+from fixtures import types
+
+if TYPE_CHECKING:
+    pass
```
tushar-signoz: What does this do ?

tests/integration/src/querier_json_body/01_logs_json_body_new_qb.py
```diff
@@ -0,0 +1,1181 @@
...
+# ============================================================================
+#
+# Data landscape (6 logs, mixed shapes):
+#   4 logs with user.name + user.age (repeated alice for count > 1)
+#   1 log with user.name but NO user.age (sparse)
+#   1 log with no user at all (different service)
+# ============================================================================
+
+
+def test_groupby_timeseries(
```
tushar-signoz: nit: Instead of timeseries, better to use scalar queries. Its easier to validate the result. Because we're interested in the handling of body which is strongly. coupled with the request_type.

tests/integration/src/querier_json_body/01_logs_json_body_new_qb.py
```diff
@@ -0,0 +1,1181 @@
...
+
+
+# ============================================================================
+# GroupBy time series
+# ============================================================================
+#
+# Data landscape (6 logs, mixed shapes):
+#   4 logs with user.name + user.age (repeated alice for count > 1)
+#   1 log with user.name but NO user.age (sparse)
+#   1 log with no user at all (different service)
```
tushar-signoz: Can we also, add some corruption in these logs to closely mimic the production. 
Ex: body.user.name is present in attributes. 
user.name is present in attributes.

tushar-signoz (changes requested):
Hey @piyushsingariya 
All these tests are Happy Paths. 
Can you add more tests that cover 
- edge cases, 
- tests that included corrupt data (ie. multiple colliding attributes with body )
- Features that are not allowed at the moment.
- Known bugs

tests/integration/fixtures/jsontypeexporter.py
piyushsingariya: I've forgotten probably earlier it was in use when the tests were running adhoc now fixtures are inplace so Removed.

pkg/telemetrymetadata/body_json_metadata.go
```diff
@@ -173,14 +149,27 @@ func buildListLogsJSONIndexesQuery(cluster string, filters ...string) (string, [
...
+// system.data_skipping_indices (e.g. "user.raw-data.name" → "user.`raw-data`.name").
+func escapedPathForLike(path string) string {
+	parts := strings.Split(path, ".")
+	for i, part := range parts {
+		if keycheck.IsBacktickRequired(part) {
+			parts[i] = "`" + part + "`"
+		}
+	}
+	return strings.Join(parts, ".")
+}
```
tushar-signoz: Does IsBacktickRequired cover all scenarios, or are there potential gaps that may require future updates? And is there any downside to always applying backticks ?

tests/integration/src/querier_json_body/01_logs_json_body_new_qb.py
```diff
@@ -0,0 +1,1691 @@
...
+
+    # Debug: show which body_v2 data-skipping indexes ClickHouse has at this point.
+    idx_result = signoz.telemetrystore.conn.query(
+        "SELECT name, expr FROM system.data_skipping_indices"
+        " WHERE database = 'signoz_logs' AND table = 'logs_v2'"
+        " AND (position(expr, 'body_v2') > 0 OR position(expr, 'body_promoted') > 0)"
+    )
+    print(f"\n[DEBUG] body_v2 data-skipping indexes on logs_v2 ({len(idx_result.result_rows)} total):")
+    for row in idx_result.result_rows:
+        print(f"  name={row[0]}  expr={row[1]}")
```
tushar-signoz: print statements

tests/integration/src/querier_json_body/01_logs_json_body_new_qb.py
```diff
@@ -0,0 +1,1691 @@
...
+    cases = [
+        # ── EXISTS: !isExistsCheck guard → indexed path skipped → plain IS NOT NULL ──────
+        # String ngrambf index on body.user.name: EXISTS skips the indexed path.
+        # IS NOT NULL("") = true → log6 (name="") IS found.
+        # log4 (no user) is the only exclusion → 5 results.
+        # query_log check: dynamicElement(...) IS NOT NULL — no assumeNotNull.
+        {
+            "name": "indexed.string_exists_skips_index_finds_empty",
+            "requestType": "raw",
+            "expression": "body.user.raw-data.name EXISTS",
```
tushar-signoz: Hey @piyushsingariya 
User using full `body.user.raw-data.name` is a best case scenario. Most of the users will use `user.raw-data.name`. Integration tests should also target this. 

This is the case for all the integration tests.

tushar-signoz (changes requested):

pkg/telemetrymetadata/body_json_metadata.go
piyushsingariya: > are there potential gaps that may require future updates?

Yes, more symbols needs to be appended to BackticksRequired Slice in collector.

> is there any downside to always applying backticks 

Nope. But will require changes in both collector and signoz for this.

tests/integration/src/querier_json_body/01_logs_json_body_new_qb.py
piyushsingariya: mybad, removing them

pkg/telemetrymetadata/body_json_metadata.go
piyushsingariya: I'll do a reverse matching removing backticks within the query and matching

piyushsingariya: > potential gaps

We can cover later.

tushar-signoz (dismissed):
LGTM !

tests/integration/tests/querier_json_body/01_logs_json_body_new_qb.py
```diff
@@ -0,0 +1,1716 @@
...
+from fixtures.logs import Logs
+from fixtures.querier import (
+    build_logs_aggregation,
+    build_order_by,
+    build_scalar_query,
+    make_query_request,
+)
+
+
+def _check_query_log(
```
therealpandey: Please move to fixtures/clickhouse.py or fixtures/telemetrystore.go as a shared utility so that others can also use this.

tests/integration/tests/querier_json_body/01_logs_json_body_new_qb.py
```diff
@@ -0,0 +1,1716 @@
...
+    return [row["data"].get(field) for row in _get_rows(response)]
+
+
+def _get_scalar_rows(response: requests.Response) -> List[List[Any]]:
+    """Return table rows from a scalar GroupBy response.
+
+    Scalar response shape: results[0]["data"] = [[group_key..., agg_value], ...]
+    """
+    results = response.json().get("data", {}).get("data", {}).get("results", [])
+    return results[0].get("data", []) if results else []
```
therealpandey: Folks have already created utilies for this in `fixtures/querier.py`. Please re-use.

tests/integration/tests/querier_json_body/conftest.py
```diff
@@ -0,0 +1,70 @@
...
+def pytest_collection_modifyitems(
+    config: pytest.Config, items: list[pytest.Item]
+) -> None:
+    version = config.getoption("--clickhouse-version")
+    if version in UNSUPPORTED_CLICKHOUSE_VERSIONS:
+        skip = pytest.mark.skip(
+            reason=f"JSON body QB tests require ClickHouse > {version}"
+        )
+        for item in items:
+            item.add_marker(skip)
```
therealpandey:
```suggestion
```

Please skip this suite in the matrix of integrationci. There is no reason to create the entire test suite for 10m only to not run the tests.

Alternatively if this doesn't even build the fixtures, we should be okay.

tests/fixtures/jsontypeexporter.py
```diff
@@ -0,0 +1,473 @@
+"""
+Simpler version of jsontypeexporter for test fixtures.
+This exports JSON type metadata to the path_types table by parsing JSON bodies
+and extracting all paths with their types, similar to how the real jsontypeexporter works.
```
therealpandey: This exporter no longer exists and can be misleading. You can keep these utilies in logs.py or maybe just rename this.

therealpandey (changes requested):

tests/integration/tests/querier_json_body/conftest.py
piyushsingariya: https://github.com/SigNoz/signoz/actions/runs/24989673395/job/73171794360?pr=10863

This passes in 20-30 second, so we can keep it right?

tests/fixtures/jsontypeexporter.py
piyushsingariya: These will be reused anywhere JSON is used so I'll rename it to jsontypes.py

tests/fixtures/telemetrystore.py
```diff
@@ -0,0 +1,79 @@
+from collections.abc import Callable
+from datetime import datetime
+
+import pytest
+
+from fixtures import types
+
+
+@pytest.fixture(name="check_query_log")
+def check_query_log_fixture(
```
therealpandey:
```suggestion
def check_query_log(
```

Can we move this to clickhouse.py? I had earlier commented to keep it in clickhouse.py or telemetrystore.py whichever existed.

tests/fixtures/jsontypes.py
```diff
@@ -0,0 +1,448 @@
...
+            '{"user": {"name": "alice", "age": 25}}',
+            '{"user": {"name": "bob", "age": 30}}',
+        ])
+
+        # Auto-extract from Logs objects
+        export_json_types(logs_list)
+    """
+    # Ensure migrator has run to create the table
+    try:
+        request.getfixturevalue("migrator")
```
therealpandey:
```suggestion
```

This becomes problematic because there can be different classes of migrators. Do we really need this?

tests/fixtures/jsontypes.py
```diff
@@ -0,0 +1,448 @@
...
+            raw = path["path"].removeprefix("body.")
+            if raw not in created_paths:
+                created_paths.append(raw)
+
+    yield _create_json_body_index
+
+    if not created_paths:
+        return
+
+    cluster = signoz.telemetrystore.env["SIGNOZ_TELEMETRYSTORE_CLICKHOUSE_CLUSTER"]
```
therealpandey, line 437: I'm assuming there is no deleting the promoted path that is why you have to manually run this to clean it up?

therealpandey (changes requested):

tests/fixtures/jsontypes.py
piyushsingariya: Yes.

therealpandey (approved):

PR #10874, opened 2026-04-07
feat(infra-monitoring-hosts): migrate to use shared component

H4ad:
## Pull Request

---

Using the shared component introduced at #10800, migrated the Hosts Container.

The UI changed a lot, specially the hostname background due to standardize with the rest of the other UI.

I also fixed/migrate to css modules and fixed some issues introduced ealier on pagination/scroll and borders of group header.

Before:

https://github.com/user-attachments/assets/ed9893ba-32a4-4785-ae2e-fe053e3ee9c6

After:

https://github.com/user-attachments/assets/9ba932d0-c80c-44b6-8af6-346ee9825798

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

- Tests added/updated: No
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Infrastructure Monitoring - Hosts Page
- Potential regressions: Missing filters or incorrect migration of the base filters

---

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | The Infrastructure Monitoring Hosts page were migrated to use a shared component for an easier maintainance |

---

### 📋 Checklist
- [x] Manually tested


frontend/src/container/InfraMonitoringK8s/Base/utils.tsx
```diff
@@ -5,6 +5,9 @@ import { IBuilderQuery } from 'types/api/queryBuilder/queryBuilderData';
 import styles from './utils.module.scss';
 
 const dotToUnder: Record<string, string> = {
```
AshwinBhatkal, line 7: just confirming, this conversion will be removed in much future PRs right?

frontend/src/container/InfraMonitoringK8s/Base/K8sBaseList.tsx
```diff
@@ -355,6 +370,37 @@ export function K8sBaseList<T>({
 
 	const showTableLoadingState = isLoading;
 
+	const defaultEmptyState = (
```
AshwinBhatkal: this is better off as it's own component

frontend/src/container/InfraMonitoringHosts/constants.tsx
```diff
@@ -0,0 +1,241 @@
...
+	if (percent >= 90) {
+		return Color.BG_CHERRY_500;
+	}
+	if (percent >= 60) {
+		return Color.BG_AMBER_500;
+	}
+	return Color.BG_FOREST_500;
+}
+
+export const hostDetailsMetadataConfig: K8sDetailsMetadataConfig<HostData>[] = [
```
AshwinBhatkal, line 52: should this be inside table.config.tsx? Like we have in other places

constants.ts is what I would expect in most cases rather than tsx

frontend/src/container/InfraMonitoringK8s/Base/utils.tsx
H4ad: Yes, I talked with Srikanth and every conversion can be removed.

frontend/src/container/InfraMonitoringHosts/constants.tsx
H4ad: Constants is related to the configuration for modal, I had to turn this to `tsx` because the modal requires a progressBar in the metadata config:

image

That's why I had to turn to tsx, to be able to render progress bar while keeping the component generic.

frontend/src/container/InfraMonitoringK8s/Base/K8sBaseList.tsx
H4ad: Good catch.

AshwinBhatkal (approved):

PR #10907, opened 2026-04-12
fix(logs): guard missing resources in log detail initial query

aiSynergy37:
## Summary
- guard against missing resources in the log detail initial query path
- prevent row-click crashes when resource attributes are absent
- add regression coverage for the missing-resources case

## Testing
- frontend unit test updated: useInitialQuery.test.ts


aks07:
resources is not supposed to be `null/undefined`, adding this check can:
- hide bugs
- weaken type guarantees
- make debugging harder because invalid state silently becomes []

Closing this as is not needed.

PR #10913, opened 2026-04-13
chore(sqlstore): added connection max lifetime support

vikrantgupta25:
### 📄 Summary
- added connection max lifetime param support

contributes to: https://github.com/SigNoz/platform-pod/issues/1803

### DIFF
--- a/conf/example.yaml
+++ b/conf/example.yaml
@@ -83,8 +83,8 @@ sqlstore:
   # The maximum number of open connections to the database.
   max_open_conns: 100
   # The maximum amount of time a connection may be reused.
-  # If conn_max_lifetime == 0, connections are not closed due to a connection's age.
-  conn_max_lifetime: 0
+  # If max_conn_lifetime == 0, connections are not closed due to a connection's age.
+  max_conn_lifetime: 0
   sqlite:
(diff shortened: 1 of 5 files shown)


conf/example.yaml
```diff
@@ -82,6 +82,9 @@ sqlstore:
   provider: sqlite
   # The maximum number of open connections to the database.
   max_open_conns: 100
+  # The maximum amount of time a connection may be reused.
+  # If conn_max_lifetime == 0, connections are not closed due to a connection's age.
+  conn_max_lifetime: 0
```
therealpandey:
```suggestion
 max_conn_lifetime: 0
```

therealpandey (changes requested):

conf/example.yaml
vikrantgupta25: this is the naming from database/sql : 

```
conn_max_lifetime:
conn_max_idle_time:
```

everything is nested under the conn for the connections config

therealpandey: See pg config.

vikrantgupta25: donee

therealpandey (approved):

Issue #5319, opened 2024-06-24
(task) add instrumentation scope to metrics and traces


nityanandagohain:
Add instrumentation scope to metrics and traces.

Related issue https://github.com/SigNoz/signoz/issues/2705 (PR https://github.com/SigNoz/signoz/pull/6339)

AB027PS:
Is there an update on this? Having instrumentation scope for traces would be useful.

krsoninikhil:
Marking blocked, [context](https://github.com/SigNoz/signoz/pull/10920#issuecomment-4932717451).

PR #10920, opened 2026-04-14
feat: add support for quering scope fields in traces

krsoninikhil:
### 📄 Summary

Add support for instrumentation scope for traces.  PR on collector - https://github.com/SigNoz/signoz-otel-collector/pull/811

image

#### Issues closed by this PR

https://github.com/SigNoz/signoz/issues/5319


nityanandagohain:
add some statementbuilder tests and integration tests as well @krsoninikhil

nityanandagohain (dismissed):
looks good to me

pkg/telemetrytraces/field_mapper.go
```diff
@@ -261,21 +262,29 @@ func (m *defaultFieldMapper) FieldFor(
...
-		} else {
-			return fmt.Sprintf("multiIf(%s.`%s` IS NOT NULL, %s.`%s`::String, mapContains(%s, '%s'), %s, NULL)", column.Name, key.Name, column.Name, key.Name, oldColumn.Name, key.Name, oldKeyName), nil
+		switch key.FieldContext {
+		case telemetrytypes.FieldContextScope:
+			scopeKey, _ := strings.CutPrefix(key.Name, "scope.")
+			switch scopeKey {
+			case "name", "version":
+				return fmt.Sprintf("%s.%s::String", column.Name, scopeKey), nil
+			default:
+				return fmt.Sprintf("%s.attributes.`%s`::String", column.Name, scopeKey), nil
```
krsoninikhil: This needs to be fixed to handle number and boolean.

krsoninikhil: Reverted other data type support from collector now, so this is okay.

pkg/telemetrytraces/field_mapper.go
```diff
@@ -261,21 +262,29 @@ func (m *defaultFieldMapper) FieldFor(
...
-		// once clickHouse dependency is updated, we need to check if we can remove it.
-		if key.Materialized {
-			oldKeyName = telemetrytypes.FieldKeyToMaterializedColumnName(key)
-			oldKeyNameExists := telemetrytypes.FieldKeyToMaterializedColumnNameForExists(key)
-			return fmt.Sprintf("multiIf(%s.`%s` IS NOT NULL, %s.`%s`::String, %s==true, %s, NULL)", column.Name, key.Name, column.Name, key.Name, oldKeyNameExists, oldKeyName), nil
-		} else {
-			return fmt.Sprintf("multiIf(%s.`%s` IS NOT NULL, %s.`%s`::String, mapContains(%s, '%s'), %s, NULL)", column.Name, key.Name, column.Name, key.Name, oldColumn.Name, key.Name, oldKeyName), nil
+		switch key.FieldContext {
+		case telemetrytypes.FieldContextScope:
+			scopeKey, _ := strings.CutPrefix(key.Name, "scope.")
```
tushar-signoz: Hey @krsoninikhil 
There is a TelemetryFieldKey Normalisation step in upstream which basically Normalises user provided telemetryInput.
`{ FieldContext: "", Name: "scope.name"}` gets converted to `{ FieldContext: "scope", Name: "name" }`
`{ FieldContext: "scope", Name: "name" }` stays as is. 

`{ FieldContext: "span", Name: "scope.name" }` This however is not valid in the sense that we have to stay consistent and here you're forcing it work by cutting the prefix. Try to draw parallel b/w attribute and scope `{ FieldContext: 'span', Name: 'attribute.name' }` is not valid.

tushar-signoz (changes requested):

pkg/telemetrytraces/field_mapper.go
krsoninikhil: Thanks for point this out. Using "name" as `key.Name` instead of `scope.name` highlight another issue that query builder will alias the `scope.name::String` to name [here](https://github.com/SigNoz/signoz/pull/10920/changes#diff-ef8b99bacf586621f377ce8914375b61fe47304ef1fae12f072e03a131b6d672R378), which could be confusing with intrinsic filed span name. 

Can this a problem?

krsoninikhil: The query generated for filters on `name` will add checks for both `name` and `scope.name`.

Filter: `name = 'auth.validate/d5' `

Query: 
```sql
((scope.name::String = 'auth.validate/d5' AND scope.name::String IS NOT NULL) \
OR name = 'auth.validate/d5')
```

Is this expected behavior?


And filter on `name` and `scope.name` both, checks both values in both attributes. 
Filter: `scope.name = 'go-trace-map' and name = 'auth.validate/d5' `

Query: 
```sql
((scope.name::String = 'go-trace-map' AND scope.name::String IS NOT NULL) AND \
((scope.name::String = 'auth.validate/d5' AND scope.name::String IS NOT NULL) OR \
name = 'auth.validate/d5'))
```

tushar-signoz: Lets address [this](https://github.com/SigNoz/signoz-otel-collector/pull/811#discussion_r3257138947) and then circle back here

pkg/telemetrytraces/field_mapper.go
```diff
@@ -261,21 +262,29 @@ func (m *defaultFieldMapper) FieldFor(
...
-		} else {
-			return fmt.Sprintf("multiIf(%s.`%s` IS NOT NULL, %s.`%s`::String, mapContains(%s, '%s'), %s, NULL)", column.Name, key.Name, column.Name, key.Name, oldColumn.Name, key.Name, oldKeyName), nil
+		switch key.FieldContext {
+		case telemetrytypes.FieldContextScope:
+			scopeKey, _ := strings.CutPrefix(key.Name, "scope.") // required for support current implementation of select fields
+			switch scopeKey {
+			case "name", "version":
+				return fmt.Sprintf("%s.%s::String", column.Name, scopeKey), nil
+			default:
+				return fmt.Sprintf("%s.attributes.`%s`::String", column.Name, scopeKey), nil
```
tushar-signoz: Hey @krsoninikhil 
Lets not cut prefix and do exact matches with `scope.name` and `scope.version`. 
This could cause an issue (very rare) when an attribute has a `scope` prefix. eg: `scope.attributes.scope.env`

pkg/telemetrytraces/field_mapper.go
```diff
@@ -261,21 +262,29 @@ func (m *defaultFieldMapper) FieldFor(
...
+			// once clickHouse dependency is updated, we need to check if we can remove it.
+			if key.Materialized {
+				oldKeyName = telemetrytypes.FieldKeyToMaterializedColumnName(key)
+				oldKeyNameExists := telemetrytypes.FieldKeyToMaterializedColumnNameForExists(key)
+				return fmt.Sprintf("multiIf(%s.`%s` IS NOT NULL, %s.`%s`::String, %s==true, %s, NULL)", column.Name, key.Name, column.Name, key.Name, oldKeyNameExists, oldKeyName), nil
+			} else {
+				return fmt.Sprintf("multiIf(%s.`%s` IS NOT NULL, %s.`%s`::String, mapContains(%s, '%s'), %s, NULL)", column.Name, key.Name, column.Name, key.Name, oldColumn.Name, key.Name, oldKeyName), nil
+			}
 		}
+		return "", errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "json column type only supported for resource and scope context, got %s", key.FieldContext.String)
```
tushar-signoz: I think this should be InternalServer Error and not bad request. User input won't result in this. 
Just double check one.

tushar-signoz (changes requested):

pkg/telemetrytraces/field_mapper.go
krsoninikhil: This case was already returning `TypeInvalidInput` for any context other than resource, if you see the diff [here](https://github.com/SigNoz/signoz/pull/10920/changes#diff-d62ca1ead07e78a2a5543c523acf32ea34212e4544ee81d7546fbedb1ae61dccL266).

Should we change that?

pkg/telemetrytraces/field_mapper.go
krsoninikhil: Then we'll need to handle `name`, `version` along with `scope.name` and `scope.version` as well unless we change things either in normalization logic or key validation. Otherwise query like `scope.version = ?` says `key version not found` even when `scope.version` is present. However, this error doesn't show up if `scope.version` is also included in select fields.

Removing it also won't allow querying on attribute name like `scope.env`, unless user queries like `scope.scope.env`, this should ideally work as long as `scope.env` key exists in our metadata store.

pkg/querybuilder/query_to_keys.go
```diff
@@ -56,6 +56,15 @@ func QueryStringToKeysSelectors(query string) []*telemetrytypes.FieldKeySelector
 					FieldDataType: key.FieldDataType,
 				})
 			}
```
tushar-signoz, line 58: Lets modify this to if fieldContext is not unspecified, we prefix the context to the name and set the fieldContext as unspecified.
@srikanthccv Did you any specific reason to pin it to context attribute ?

tushar-signoz (changes requested):

pkg/querybuilder/query_to_keys.go
krsoninikhil: Right, reverting changes of `pkg/querybuilder/query_to_keys.go` done in https://github.com/SigNoz/signoz/pull/8823 should be enough.

tushar-signoz: We'll handle this later, please add a TODO for me

pkg/querybuilder/query_to_keys.go
```diff
@@ -56,6 +56,15 @@ func QueryStringToKeysSelectors(query string) []*telemetrytypes.FieldKeySelector
...
+
+			if key.FieldContext == telemetrytypes.FieldContextScope {
+				keys = append(keys, &telemetrytypes.FieldKeySelector{
+					Name:          key.FieldContext.StringValue() + "." + key.Name,
+					Signal:        key.Signal,
+					FieldContext:  key.FieldContext,
```
tushar-signoz: Please change FieldContext to Unspecified

pkg/telemetrytraces/field_mapper.go
tushar-signoz: Yes, Im aware it's old code.

pkg/querybuilder/query_to_keys.go
krsoninikhil: https://github.com/SigNoz/signoz/issues/11374

pkg/telemetrytraces/field_mapper.go
krsoninikhil: Fixed in [782de2b](https://github.com/SigNoz/signoz/pull/10920/commits/782de2b2107da77e4f98babb35080fdcb1ef2980)

pkg/querybuilder/query_to_keys.go
krsoninikhil: Fixed in [d3c3869](https://github.com/SigNoz/signoz/pull/10920/commits/d3c38693f3c080b0f5dcf38aa340634c852d5212)

tushar-signoz:
LGTM, Please add integration tests for this as well as add test cases in existing ones.

krsoninikhil:
Currently, scope fields are not returned if not selected field is provided, this will done as a follow to include scope in query range and waterfall API.

Below are the test cases for select fields in query range 

Ingested data for context:
```
   ┌─tagKey────────────┬─tagType─┬─dataType─┬─isColumn─┬───────────timestamp─┐
1. │ scope.name        │ scope   │ string   │ false    │ 2026-07-02 10:13:53 │
2. │ scope.prefixed    │ scope   │ string   │ false    │ 2026-07-02 10:13:53 │
3. │ scope.version     │ scope   │ string   │ false    │ 2026-07-02 10:13:53 │
4. │ testing.arch      │ scope   │ string   │ false    │ 2026-07-02 10:13:53 │
5. │ testing.archfloat │ scope   │ string   │ false    │ 2026-07-02 10:13:53 │
6. │ testing.enabled   │ scope   │ string   │ false    │ 2026-07-02 10:13:53 │
7. │ testing.env       │ scope   │ string   │ false    │ 2026-07-02 10:13:53 │
   └───────────────────┴─────────┴──────────┴──────────┴─────────────────────┘
```

Cases:


selectFields | query | expected? 
-|-|-
 { "name": "scope.testing.env",  "fieldContext": "" } | scope.attributes.`testing.env`::String AS `testing.env` | ✅ 
{ "name": "scope.testing.env",  "fieldContext": "scope" }| scope.attributes.`scope.testing.env`::String AS `scope.testing.env` | ✅ 
{ "name": "testing.env",  "fieldContext": "" } | scope.attributes.`testing.env`::String AS `testing.env` | ✅ 
{ "name": "testing.env",  "fieldContext": "scope" } | scope.attributes.`testing.env`::String AS `testing.env` | ✅ 
{ "name": "scope.version",  "fieldContext": "scope" } | scope.version::String AS `scope.version` | ✅ 
{ "name": "version",  "fieldContext": "scope" } | scope.attributes.`version`::String AS `version` | ✅ 
{ "name": "scope.version",  "fieldContext": "" } | scope.attributes.`version`::String AS `version` | ⛔ 
{ "name": "scope.name",  "fieldContext": "scope" } | scope.name::String AS `scope.name` | ✅ 
{ "name": "scope.name",  "fieldContext": "" } |  name AS `name` | ⛔ 
{ "name": "name",  "fieldContext": "scope" } |  name AS `name` | ⛔

krsoninikhil:
After above fix, below case is resolved

selectFields | query | expected?
-|-|-
{ "name": "scope.version",  "fieldContext": "" } | scope.version::String AS `scope.version` | ✅ 


Problem with `scope.name` is still not resolved. @tushar-signoz

krsoninikhil:
`scope.name` without fieldContext is getting resolved to span `name`, which highlights some gap in our queries and need attention. Marking this as blocked until that is resolved. cc/ @tushar-signoz

krsoninikhil:
Current behavior of QB when a key "att_and_res" exists with attribute and resource both as context

In select expression
```
{att_and_res, ''}          → multiIf(mapContains(resources_string,'att_and_res'), toString(resources_string[…]),
                                     mapContains(attributes_string,'att_and_res'), toString(attributes_string[…]), NULL)
{att_and_res, 'attribute'} → attributes_string['att_and_res']      ✓ explicit context honoured
{att_and_res, 'resource'}  → resources_string['att_and_res']       ✓
```

In filter expression
```
att_and_res = 'x'            → resource sub-query CTE only (attribute side dropped)
  WARN: Key `att_and_res` is ambiguous, found 2 different combinations of field context / data type…
        Using `resource` context by default. To query attributes explicitly, use the fully qualified
        name (e.g., 'attribute.att_and_res')
attribute.att_and_res = 'x'  → attributes_string['att_and_res']    ✓
resource.att_and_res = 'x'   → resource sub-query CTE               ✓
```

pkg/querybuilder/collision.go
```diff
@@ -135,6 +135,18 @@ func AdjustKey(key *telemetrytypes.TelemetryFieldKey, keys map[string][]*telemet
...
+		// that as a match too, so it is not force-overridden to a different-context intrinsic.
+		if !match && key.FieldContext != telemetrytypes.FieldContextUnspecified {
+			for _, mapKey := range keys[key.FieldContext.StringValue()+"."+key.Name] {
+				if (key.FieldDataType == telemetrytypes.FieldDataTypeUnspecified || mapKey.FieldDataType == key.FieldDataType) &&
+					!mapKey.Equal(intrinsicOrCalculatedField) {
+					match = true
+					break
+				}
+			}
+		}
```
tushar-signoz: Hey @krsoninikhil 
This is not the right place to handle this. 
Have a look at `adjustTraceKey` function. It looks up the map `IntrinsicFields` to determine if it is an intrinsic field or not, which is no longer valid as there are 2 intrinsic fields now with same name `name`.

Lets correct this method, may be change it map[string][]*telemetrykeys ?

tushar-signoz (changes requested):

krsoninikhil:
> **Edit:** re-ran every case below through `query_range/preview` against a local server on `ns/scope`, after [`a895ba9`](https://github.com/SigNoz/signoz/commit/a895ba9ab5) (#12666) changed how a scope name resolves: a name now resolves to **every** home it names rather than one. All 16 cases still hold. The only difference from the original tables is cosmetic — the union branches in the second table now carry a `toString(...)` wrapper (both branches are already `::String`, so it is a no-op). Statuses and the notes below are updated to the current state.
>
> The first table is the local instance's natural metadata state (declared `scope.name`/`scope.version` plus a `testing.env` scope attribute, no same-named ones); the second table's rows were exercised by adding `name` and `version` as `scope` keys to `span_attributes_keys`, then removing them again.

Re-verified all cases on the current branch. Each select field is exists-guarded (`multiIf(<guard>, <expr>, NULL)`); the value expressions:

| selectFields | generated expression | status |
|-|-|-|
| `{ "name": "scope.testing.env", "fieldContext": "" }` | `` scope.attributes.`testing.env`::String `` | ✅ |
| `{ "name": "scope.testing.env", "fieldContext": "scope" }` | `` scope.attributes.`scope.testing.env`::String `` | ✅ |
| `{ "name": "testing.env", "fieldContext": "" }` | `` scope.attributes.`testing.env`::String `` | ✅ |
| `{ "name": "testing.env", "fieldContext": "scope" }` | `` scope.attributes.`testing.env`::String `` | ✅ |
| `{ "name": "scope.version", "fieldContext": "scope" }` | `scope.version::String` | ✅ |
| `{ "name": "version", "fieldContext": "scope" }` | `scope.version::String` | 🔄 now binds to declared `scope.version` |
| `{ "name": "scope.version", "fieldContext": "" }` | `scope.version::String` | ✅ fixed (was ⛔) |
| `{ "name": "scope.name", "fieldContext": "scope" }` | `scope.name::String` | ✅ |
| `{ "name": "scope.name", "fieldContext": "" }` | `scope.name::String` | ✅ fixed (was ⛔) |
| `{ "name": "name", "fieldContext": "scope" }` | `scope.name::String` | ✅ fixed (was ⛔) |

All three previously-⛔ cases are fixed. One case shifted from the original table: `{ "name": "version", "fieldContext": "scope" }` now binds to the declared `scope.version` (metadata key `scope.version` matched via the context-prefixed union), symmetric with `{name, scope}` → `scope.name`. This is the intended consequence of unifying `{name, scope}` ≡ `scope.name`: a short name under `scope` context binds to a declared scope path when one exists.

---

**Additional cases — when a scope attribute named `version` / `name` is *also* ingested** (alongside the declared `scope.version` / `scope.name`):

| selectFields | generated expression | note |
|-|-|-|
| `{ "name": "version", "fieldContext": "scope" }` | `` multiIf(scope.attributes.`version` IS NOT NULL, toString(scope.attributes.`version`::String), scope.version::String <> '', toString(scope.version::String), NULL) `` | unions the scope attribute **and** the declared path (attribute first) |
| `{ "name": "scope.version", "fieldContext": "" }` | *(same as above)* | normalizes to `{version, scope}`, same union |
| `{ "name": "scope.version", "fieldContext": "scope" }` | `scope.version::String` | the full `scope.version` name under explicit `scope` context addresses **only** the declared path |
| `{ "name": "name", "fieldContext": "scope" }` | `` multiIf(scope.attributes.`name` IS NOT NULL, toString(scope.attributes.`name`::String), scope.name::String <> '', toString(scope.name::String), NULL) `` | unions the scope attribute **and** the declared path |
| `{ "name": "scope.name", "fieldContext": "" }` | *(same as above)* | normalizes to `{name, scope}`, same union |
| `{ "name": "scope.name", "fieldContext": "scope" }` | `scope.name::String` | full `scope.name` under explicit `scope` context addresses **only** the declared path |

So when both a declared path and a same-named scope attribute exist, a short name under `scope` context (or the bare `scope.<x>` spelling) surfaces **both** (attribute first, declared as fallback), while the explicit `{ "name": "scope.<x>", "fieldContext": "scope" }` spelling binds unambiguously to the declared path. (Each expression is exists-guarded to yield `NULL` when absent, as above.)

The reverse also holds after #12666: when no same-named scope attribute is ingested, the short name binds to the declared path alone — the first table's `{name, scope}` and `{version, scope}` rows. The union appears only when metadata actually knows both homes.

pkg/querybuilder/collision.go
krsoninikhil: Let me check but both of these has different context which should not get overridden.

krsoninikhil: From our last discussion on treating `scope.name` as a field name and not the scope attribute (unless scope attribute named name exists). So query on `name` would not consider the scope.name.

tushar-signoz: What Im trying to say is, `IntrinsicFields` map is not valid anymore, since now we have 2 intrinsic fields with the same name.

krsoninikhil: Treating the scope.name as intrinsic field to avoid conflicting names in [cb05e09](https://github.com/SigNoz/signoz/pull/10920/commits/cb05e09a85f24b70ccbaf71123ec41a504e24d9a)

pkg/querybuilder/query_to_keys.go
```diff
@@ -56,6 +56,17 @@ func QueryStringToKeysSelectors(query string) []*telemetrytypes.FieldKeySelector
...
+			// todo(tushar): consider reverting changes done to this method in below PR to avoid scope specific checks
+			// https://github.com/SigNoz/signoz/issues/11374
+			if key.FieldContext == telemetrytypes.FieldContextScope {
+				keys = append(keys, &telemetrytypes.FieldKeySelector{
+					Name:          key.FieldContext.StringValue() + "." + key.Name,
+					Signal:        key.Signal,
+					FieldContext:  telemetrytypes.FieldContextUnspecified, // this allows 'scope.' prefix for keys with other context as well
+					FieldDataType: key.FieldDataType,
+				})
+			}
```
krsoninikhil, line 69: This is still required to cover the cases like "scope.something" being a scope attribute.

pkg/telemetryschema/tracestelemetryschema/field_mapper.go
```diff
@@ -417,23 +428,38 @@ func (m *fieldMapper) ColumnExpressionFor(
...
+		// same-named scope attribute (attribute-first) alongside the declared path, and
+		// CandidateKeys synthesizes when metadata knows neither.
+		matches := querybuilder.MatchingLogicalFields(ctx, orgID, m.fl, field, keys)
+		candidates, _ = querybuilder.ResolveLogicalFields(field, matches)
+		if len(candidates) == 0 {
+			candidates = querybuilder.WrapAsLogicalFields(field.Name, m.CandidateKeys(ctx, orgID, field, nil, keys))
+		}
+		if len(candidates) == 0 {
+			return "", errors.Wrapf(querybuilder.NewKeyNotFoundError(field.Name), errors.TypeInvalidInput, errors.CodeInvalidInput, "field `%s` not found", field.Name).WithSuggestions(errors.NewSuggestionsOnLevenshteinDistance(field.Name, errors.NounKeys, maps.Keys(keys))...)
 		}
```
tushar-signoz: why is there a special handling of scope context ?

tushar-signoz:
Hey @krsoninikhil 
Can we take a stance that `name` and `version` cannot be part of scope attributes. Handling this is making codebase more complicated ?

tushar-signoz (changes requested):

krsoninikhil:
> Hey @krsoninikhil Can we take a stance that `name` and `version` cannot be part of scope attributes. Handling this is making codebase more complicated ?

It would allow getting rid of scope specific branch in the `ColumnExpressionFor`. But If we support "name", "timestamp", "trace_id" etc in attributes, this stance doesn't seem reasonable. Plus, it removes edge case from code, but introduce it in the product. Let's handle it in similar way in collector normalize operator like others.

pkg/telemetryschema/tracestelemetryschema/field_mapper.go
krsoninikhil: Not seeing a way around this. `err == nil` branch of default case uses `logicalForResolvedColumn` and returns only one key, while filter (ConditionFor) does an union of all possible keys. We want the union in select as well for scope attributes.

krsoninikhil: Problem is name+scope represents both, scope.name and scope.attributes.name after normalization and existing branch doesn't handle two possibilities. This is handled well by default in the `ConditionFor`.

krsoninikhil: For other field names, a telemetry key has a fixed location if context is specified. In case of scope, name with scope context, is not enough to point out the column and since context is not unspecified, it doesn't reach `CandidateKeys` for creating a multiIf query. 

So to make name+scope unambiguous, we either keep this separate scope handling or we distinguish the context for schema fields (scope.name, scope.version) and other scope attributes like `scope.attribute.name` and `scope.attribute.version` where attribute can be omitted for non-conflicting keys. So basically, requiring `scope.attribute` prefix if attribute name and version needs to be queried.

pkg/telemetryschema/tracestelemetryschema/field_mapper.go
```diff
@@ -181,7 +182,10 @@ func (m *fieldMapper) getColumn(
 	case telemetrytypes.FieldContextResource:
 		return []*schema.Column{indexV3Columns["resource"], indexV3Columns["resources_string"]}, nil
 	case telemetrytypes.FieldContextScope:
-		return []*schema.Column{}, qbtypes.ErrColumnNotFound
+		if key.Name == "name" || key.Name == "version" {
```
tushar-signoz: Add a note here as to why we're returning no column here

pkg/telemetryschema/tracestelemetryschema/field_mapper.go
```diff
@@ -292,14 +296,25 @@ func (m *fieldMapper) resolveColumnExprs(
...
+					// declared String paths on the scope column read '' for the missing case
+					exprs = append(exprs, fmt.Sprintf("%s::String", key.Name))
+					existExprs = append(existExprs, fmt.Sprintf("%s <> ''", key.Name))
+				} else {
+					attributeName := strings.TrimPrefix(key.Name, "attribute.") // literal "attribute" prefix in attribute keys needs double prefix
+					exprs = append(exprs, fmt.Sprintf("%s.attributes.`%s`::String", columnName, attributeName))
+					existExprs = append(existExprs, fmt.Sprintf("%s.attributes.`%s` IS NOT NULL", columnName, attributeName))
+				}
+			default:
+				return nil, nil, nil, errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "only resource and scope context fields are supported for json columns, got %s", key.FieldContext.String)
```
tushar-signoz: Please change this to NewInternalServerError. I know it used to be bad input

pkg/telemetryschema/tracestelemetryschema/field_mapper.go
```diff
@@ -599,11 +608,51 @@ func (m *fieldMapper) CandidateKeys(ctx context.Context, _ valuer.UUID, field *t
...
+// synthScopeAttributeKey guesses a scope attribute (scope.attributes.<name>) for a name absent
+// from metadata — the scope analog of querybuilder.SynthesizeKeys.
+func synthScopeAttributeKey(field *telemetrytypes.TelemetryFieldKey) *telemetrytypes.TelemetryFieldKey {
+	return telemetrytypes.NewTelemetryFieldKey(field.Name, telemetrytypes.FieldContextScope, telemetrytypes.FieldDataTypeString)
+}
+
+func isDeclaredScopePath(name string) bool {
+	f, ok := IntrinsicFields[name]
+	return ok && f.FieldContext == telemetrytypes.FieldContextScope
+}
```
tushar-signoz: Please remove these 1 liner functions

pkg/telemetryschema/tracestelemetryschema/field_mapper.go
```diff
@@ -292,14 +296,25 @@ func (m *fieldMapper) resolveColumnExprs(
...
 			// have to add ::string as clickHouse throws an error :- data types Variant/Dynamic are not allowed in GROUP BY
 			// once clickHouse dependency is updated, we need to check if we can remove it.
-			exprs = append(exprs, fmt.Sprintf("%s.`%s`::String", columnName, key.Name))
-			existExprs = append(existExprs, fmt.Sprintf("%s.`%s` IS NOT NULL", columnName, key.Name))
+			switch key.FieldContext {
+			case telemetrytypes.FieldContextResource:
+				exprs = append(exprs, fmt.Sprintf("%s.`%s`::String", columnName, key.Name))
+				existExprs = append(existExprs, fmt.Sprintf("%s.`%s` IS NOT NULL", columnName, key.Name))
+			case telemetrytypes.FieldContextScope:
+				if isDeclaredScopePath(key.Name) {
```
tushar-signoz: Lets hard code the two values here instead of having a function, would make it easier to understand

pkg/telemetryschema/tracestelemetryschema/field_mapper.go
```diff
@@ -292,14 +296,25 @@ func (m *fieldMapper) resolveColumnExprs(
...
+					// declared String paths on the scope column read '' for the missing case
+					exprs = append(exprs, fmt.Sprintf("%s::String", key.Name))
```
tushar-signoz, line 305: Could you check one thing for me, what's the difference in following. 
scope.`name`::String 
vs scope.name::String 
vs scope.name 
vs scope.`name`

Leverage `querybuilder.ClickhouseLiteral` here

tushar-signoz (changes requested):
Looks good sir, very minor things

pkg/telemetryschema/tracestelemetryschema/field_mapper.go
krsoninikhil: `::String` is required if the type is Dynamic otherwise group by breaks, which is the case for all attributes. But scope.name and scope.version are already typed String in the schema, it's noop just kept for consistency.

Backticks are only required for paths with non non-alpha num chars so scope.`foo.bar` would look for a literal key `foo.bar` not take bar nested inside foo. Not required for scope name and version.

tushar-signoz: `foo.bar` will not match with `foo { bar : 'xyz' }` ? I thought json by default splits paths on `.`

krsoninikhil: There is a [setting](https://clickhouse.com/docs/guides/clickhouse/data-formats/json/formats#accessing-nested-json-objects) to disable that but right, backticks doesn't escape the dots, there is separate setting for that. It just allows non alpha num in the keys.

tushar-signoz (approved):
LGTM

therealpandey (approved):

ahrefabhi (approved):

PR #10925, opened 2026-04-14
chore:  Removal of JSONDataType field

piyushsingariya:
## Pull Request

---

### Summary
This PR aims to target a series of fixes in JSON logs such as 
- Removal of separate dedicated query for fetching Parent's Array datatypes 
- Removal of JSON DataType field from TelemetryFieldKey and direct usage in SQL expressions.
- Issue with getLogsKeys logging error when opened Landing page with no data ingested.
- Issue with failing when Selecting or Ordering by fields inside Arrays

#### Issues closed by this PR

Closes signoz/engineering-pod#4572
Closes signoz/engineering-pod#4576
Closes signoz/engineering-pod#4626
Closes signoz/engineering-pod#4611

---

### Change Type
_Select all that apply_

- [x] Bug fix
- [x] Refactor

---

- Tests added/updated:
  - `json_stmt_builder_test.go` — updated all ambiguity warning strings to omit `jsondatatype=...` suffix, confirming `TelemetryFieldKey.String()` no longer prints it.
  - `json_access_plan_test.go` — all `makeKey` calls now pass `FieldDataType` instead of `JSONDataType`; YAML assertions updated to reflect `FieldDataType` string values (e.g., `"[]json"` instead of `"Array(JSON)"`).
  - `stmt_builder_test.go` — removed `require.Equal` assertion for `JSONDataType` in `TestAdjustKey`.
  - `body_json_metadata_test.go` — removed `TestBuildGetBodyJSONPathsQuery` (function deleted).
  - `test_data.go` — `TestJSONTypeSet` returns `map[string][]FieldDataType` instead of `map[string][]JSONDataType`.
- Manual verification: N/A (backend-only refactor, validated through unit tests)
- Edge cases covered: ambiguous multi-type keys, promoted vs non-promoted paths, deeply nested array paths (6-level nesting), parent type cache population for `[]json`/`[]dynamic` branches.

---

- Blast radius: None, JSON Logs are not active yet.
- Potential regressions: N/A
- Rollback plan: Revert the commit. No data migration or schema changes are involved.

---

| Field | Value |
|-------|-------|
| Change Type | Maintenance |
| Description | N/A |

---

### Checklist
- [x] Manually tested

---

## Notes for Reviewers
**Design rationale:**
- One UNION query instead of two separate queries reduces DB round-trips and keeps body-key discovery consistent with attribute/resource key discovery.
- `FieldDataType` as the single stored type simplifies the data model — `JSONDataType` is a ClickHouse implementation detail that only matters when constructing SQL expressions, not when representing keys in memory.
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


pkg/telemetrymetadata/metadata.go
```diff
@@ -397,90 +411,152 @@ func (t *telemetryMetaStore) getLogsKeys(ctx context.Context, fieldKeySelectors
...
+			// unspecified context, query all tables
 			queryAttributeTable = true
 			queryResourceTable = true
+			queryBodyTable = true
 			break
 		} else if selector.FieldContext == telemetrytypes.FieldContextAttribute {
 			queryAttributeTable = true
 		} else if selector.FieldContext == telemetrytypes.FieldContextResource {
 			queryResourceTable = true
+		} else if selector.FieldContext == telemetrytypes.FieldContextBody && querybuilder.BodyJSONQueryEnabled {
```
tushar-signoz: Remove the redundant check for `querybuilder.BodyJSONQueryEnabled` as you later check it at line:433 `queryBodyTable = queryBodyTable && querybuilder.BodyJSONQueryEnabled`

pkg/telemetrymetadata/metadata.go
```diff
@@ -397,90 +411,152 @@ func (t *telemetryMetaStore) getLogsKeys(ctx context.Context, fieldKeySelectors
...
+			}
 		}
+	}
 
-		fieldContext := table.fieldContext
+	// Each UNION arm differs only in: table, data-type column name and SELECT
+	// expression (lower-wrapped for historical mixed-case in attr/resource),
+	// base WHERE filters, the per-selector data-type encoding, and (for body)
+	// an extra OR branch that fetches parent array types for JSON plan building.
+	// All other logic is shared by the loop below.
```
tushar-signoz: IMO the amount of cognitive complexity this added outweighs its advantage. 
I would very much prefer 3 blocks of code working for attributes, resource, and body context. Easier on my weathering eyes.

pkg/telemetrymetadata/metadata.go
```diff
@@ -517,6 +593,7 @@ func (t *telemetryMetaStore) getLogsKeys(ctx context.Context, fieldKeySelectors
 	defer rows.Close()
 
 	keys := []*telemetrytypes.TelemetryFieldKey{}
+	parentTypeCache := make(map[string][]telemetrytypes.FieldDataType)
```
tushar-signoz: lets not name it cache

pkg/querybuilder/where_clause_visitor_test.go
```diff
@@ -318,15 +318,15 @@ func TestVisitKey(t *testing.T) {
 					{
 						Name:          "count",
 						FieldContext:  telemetrytypes.FieldContextAttribute,
-						FieldDataType: telemetrytypes.FieldDataTypeNumber,
```
tushar-signoz, line 321: Was this required ?

pkg/telemetrymetadata/body_json_metadata_test.go
```diff
@@ -7,99 +7,9 @@ import (
 	"github.com/SigNoz/signoz-otel-collector/constants"
 	"github.com/SigNoz/signoz/pkg/querybuilder"
 	"github.com/SigNoz/signoz/pkg/telemetrylogs"
-	"github.com/SigNoz/signoz/pkg/types/telemetrytypes"
 	"github.com/stretchr/testify/require"
 )
 
-func TestBuildGetBodyJSONPathsQuery(t *testing.T) {
```
tushar-signoz, line 14: These tests were removed, don't see their ported counter parts

pkg/telemetrylogs/json_stmt_builder_test.go
```diff
@@ -942,20 +929,171 @@ func TestJSONStmtBuilder_IndexedPaths(t *testing.T) {
...
+
+	cases := []struct {
+		name                string
+		requestType         qbtypes.RequestType
+		query               qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]
+		expected            qbtypes.Statement
+		expectedErrContains string
+	}{
+		{
+			name:        "select_x_education[].awards[].participated[].members",
```
tushar-signoz, line 955: Can you add a test case for 
education[].awards.participated[].members 
education[].awards.participated ?

tushar-signoz (commented):
Some comments

pkg/telemetrymetadata/metadata.go
```diff
@@ -540,6 +617,17 @@ func (t *telemetryMetaStore) getLogsKeys(ctx context.Context, fieldKeySelectors
 		if err != nil {
 			return nil, false, errors.Wrap(err, errors.TypeInternal, errors.CodeInternal, ErrFailedToGetLogsKeys.Error())
 		}
+
+		// body keys with ArrayJSON/ArrayDynamic types are internal container types
+		// used only for JSON access plan building; route to parentTypeCache, not to results
+		switch fieldDataType {
+		case telemetrytypes.FieldDataTypeArrayJSON, telemetrytypes.FieldDataTypeArrayDynamic:
+			if fieldContext == telemetrytypes.FieldContextBody && parentPaths[name] {
```
tushar-signoz, line 617: Claude: **Parent path routing may swallow user-requested fields.**

If a user queries for both `body.education` (the array itself) **and** `body.education[].name`, then `"education"` will appear in `parentPaths` (computed from the child selector's parent). The `ArrayJSON` row for `"education"` gets routed to `parentTypeCache` here and **never appears in `keys`**, even though the user explicitly asked for it.

Likely fine while JSON logs aren't active, but worth adding a guard or comment — e.g. only route to `parentTypeCache` if the name is *not* also matched by a user-facing selector.

pkg/types/telemetrytypes/json_datatype.go
```diff
@@ -63,18 +55,8 @@ var MappingFieldDataTypeToJSONDataType = map[FieldDataType]JSONDataType{
 	FieldDataTypeArrayString:  ArrayString,
 	FieldDataTypeArrayInt64:   ArrayInt64,
 	FieldDataTypeArrayFloat64: ArrayFloat64,
+	FieldDataTypeArrayNumber:  ArrayFloat64,
```
tushar-signoz, line 58: Claude: **`FieldDataTypeNumber` is missing from this map.**

`FieldDataTypeArrayNumber` is covered here, but `FieldDataTypeNumber` (the scalar counterpart) is absent. If `GetJSONDataType()` is ever called on a key with `FieldDataTypeNumber`, it returns a zero-value `JSONDataType`.

This may be intentional (body JSON fields should use specific types like `Int64`/`Float64`), but it's a latent trap. Consider either adding `FieldDataTypeNumber: Float64` for safety, or adding a comment explaining why it's intentionally excluded.

pkg/telemetrymetadata/metadata.go
```diff
@@ -593,13 +681,11 @@ func (t *telemetryMetaStore) getLogsKeys(ctx context.Context, fieldKeySelectors
 		}
 	}
 
+	// enrich body keys with promoted paths, indexes, and JSON access plans
 	if querybuilder.BodyJSONQueryEnabled {
-		bodyJSONPaths, finished, err := t.buildBodyJSONPaths(ctx, fieldKeySelectors) // LIKE for pattern matching
-		if err != nil {
-			t.logger.ErrorContext(ctx, "failed to extract body JSON paths", errors.Attr(err))
+		if err := t.enrichJSONKeys(ctx, fieldKeySelectors, keys, parentTypeCache); err != nil {
+			return nil, false, err
```
tushar-signoz, line 681: Claude: **Behavior change in error handling.**

The old code logged the error and continued (graceful degradation):
```go
if err != nil {
    t.logger.ErrorContext(ctx, "failed to extract body JSON paths", errors.Attr(err))
}
```

The new code returns the error, failing the entire `getLogsKeys` call. This is arguably more correct, but if `GetPromotedPaths` or `getJSONPathIndexes` ever fail (e.g. table doesn't exist during a migration), the entire logs-keys API now fails instead of degrading gracefully.

Fine while JSON logs aren't active — just flagging for when the feature goes live.

tushar-signoz (commented):
Claude: Overall this is a solid refactor — simplifies the data model, reduces DB round-trips, and fixes real bugs in EXISTS/NOT EXISTS handling. A few items worth addressing inline.

pkg/telemetrylogs/json_stmt_builder_test.go
```diff
@@ -870,22 +873,6 @@ func TestJSONStmtBuilder_IndexedPaths(t *testing.T) {
...
-				Filter: &qbtypes.Filter{Expression: "body.user.name NOT EXISTS"},
-				Limit:  10,
-			},
-			expected: TestExpected{
-				WhereClause: "((assumeNotNull(dynamicElement(body_v2.`user.name`, 'String')) = ? AND dynamicElement(body_v2.`user.name`, 'String') IS NULL))",
-				Args:        []any{uint64(1747945619), uint64(1747983448), "", "1747947419000000000", uint64(1747945619), "1747983448000000000", uint64(1747983448), 10},
-			},
-		},
 
 		// ── special-character indexed paths ───────────────────────────────────
```
tushar-signoz, line 885: Claude: **"Indexed String NotExists" test was removed without replacement.**

The old test validated the SQL generated for `NOT EXISTS` on an indexed path. After the refactor, indexed paths are skipped for EXISTS/NOT EXISTS checks — the code falls through to the regular `dynamicElement(...) IS NULL` path. But there's no explicit test validating that `NOT EXISTS` still generates the correct SQL after this change.

Consider adding a test case here to prevent regressions.

pkg/telemetrymetadata/body_json_metadata.go
```diff
@@ -21,185 +21,91 @@ import (
...
-	if err != nil {
-		return err
-	}
-
-	typeCache := make(map[string][]telemetrytypes.JSONDataType)
-	for _, key := range parentKeys {
-		typeCache[key.Name] = append(typeCache[key.Name], *key.JSONDataType)
+// buildJSONPlans builds JSON access plans for the given keys
+// using the provided parent type cache (pre-fetched in the main UNION query).
+func (t *telemetryMetaStore) buildJSONPlans(_ context.Context, keys []*telemetrytypes.TelemetryFieldKey, typeCache map[string][]telemetrytypes.FieldDataType) error {
```
tushar-signoz: Claude: **Nit:** `ctx` is now unused (blanked to `_`). Since `buildJSONPlans` no longer makes DB calls, consider removing the `context.Context` parameter entirely to avoid misleading callers — or keep it if you anticipate needing it again soon.

pkg/querybuilder/where_clause_visitor_test.go
piyushsingariya: Yes, int64 strictly needs to be mapped to FieldDataTypeInt64 else, I would end up querying for float64 whereas data would int64.

pkg/telemetrymetadata/body_json_metadata_test.go
piyushsingariya: Yes the merged getLogsKeys is what it is now

pkg/telemetrymetadata/metadata.go
piyushsingariya: Yes I think the earlier was a bad approach, current one is better now

pkg/telemetrylogs/json_stmt_builder_test.go
piyushsingariya: Yes, Exist testcase is already inplace, so the other testcase was not adding much value

pkg/telemetrymetadata/metadata.go
piyushsingariya: Yes, I had a later thought on this, thanks for pointing

pkg/telemetrylogs/json_stmt_builder_test.go
piyushsingariya: It will be same or redundent, I believe. I'll add another case that;s more useful

pkg/telemetrymetadata/metadata.go
piyushsingariya: Noted, will look into this

tushar-signoz (approved):
LGTM

srikanthccv (approved):

PR #10943, opened 2026-04-15
feat(apiserver): derive HTTP route prefix from global.external_url

therealpandey:
## Summary

- Derive HTTP route prefix from the path component of `global.external_url`, enabling SigNoz to be served behind a reverse proxy at a sub-path (e.g. `https://example.com/signoz/`)
- Apply prefix via `http.StripPrefix` at the outermost handler level — zero changes to route registration code
- Health endpoints (`/api/v1/health`, `/api/v2/healthz`, `/api/v2/readyz`, `/api/v2/livez`) remain accessible without the prefix for container healthchecks
- Remove `web.prefix` config in favor of the unified `global.external_url` approach

## Configuration

```yaml
global:
  external_url: https://example.com/signoz
# All routes (API + web) automatically served under /signoz
```

closes SigNoz/platform-pod#1775


pkg/global/config.go
```diff
@@ -37,5 +39,28 @@ func newConfig() factory.Config {
...
+
+func (c Config) ExternalPath() string {
+	if c.ExternalURL == nil || c.ExternalURL.Path == "" || c.ExternalURL.Path == "/" {
+		return ""
+	}
+	p := path.Clean("/" + c.ExternalURL.Path)
+	if p == "/" {
+		return ""
+	}
+	return p
```
therealpandey, line 63:
```suggestion
	if c.ExternalURL == nil || c.ExternalURL.Path == "" || c.ExternalURL.Path == "/" {
		return ""
	}
	
	p := path.Clean("/" + c.ExternalURL.Path)
	if p == "/" {
		return ""
	}
	
	return p
```

Do you see the formatting difference?

vikrantgupta25 (approved):

PR #10945, opened 2026-04-15
fix: improved key suggestions

tushar-signoz:
## Pull Request

https://github.com/user-attachments/assets/407ff853-67ab-4e8f-9824-c1c15c1cbfbb

---

### 📄 Summary

Teaches the filter-expression key autocomplete about field-context prefixes so suggestions behave correctly when the user scopes to a context. The recognized contexts aren't hardcoded — the autocomplete asks the cache directly: if any cached key has `fieldContext === <head>`, then `<head>.` is a scoped prefix.

**Before (on `main`):**
- Key suggestions were a flat substring match: `label.toLowerCase().includes(input)` against the cached key list.
- No awareness of the `<ctx>.<name>` shape — typing `attribute.` matched nothing because no cached label starts with `attribute.`.
- Each fetch hit the API directly; concurrent/repeat fetches weren't deduped or cached across remounts.
- A fetch rejection bubbled up as an unhandled promise rejection (fatal under Node 22's default).
- `fieldContext` type union on `querySuggestions/types.ts` didn't include `log` or `body`.

**After:**
- Bare input (e.g. `att`) → matches against field names only.
- A `<head>.<rest>` input where `<head>` matches some cached key's `fieldContext` → scoped view: only keys in that context, displayed as `head.<name>`. Further characters after the dot substring-filter within that context.
- The "is this a context?" check is driven entirely by the backend's response — no hardcoded list, no stored set. A new backend context becomes scopeable as soon as one key carrying it lands in the cache.
- One debounced fetch with the raw text — the backend parses the context out of `searchText` itself.
- Fetches route through React Query (in-flight dedup + 5-min `staleTime`).
- Fetch errors are swallowed (with a `console.warn` in non-production) so the editor stays usable on transient backend failures.
- `log` and `body` added to the `fieldContext` union in `querySuggestions/types.ts`.

#### Issues closed by this PR
N/A

---

### ✅ Change Type

- [x] ✨ Feature (context-aware key autocomplete)
- [x] 🐛 Bug fix (unhandled rejection)
- [x] ♻️ Refactor (React Query routing)

---

#### Root Cause
Two issues on `main`:

1. The autocomplete treated the filter expression as opaque text. Typing `attribute.` — the natural way to ask for "keys in the attribute context" — matched zero suggestions because no label in the cache starts with `attribute.`.
2. `fetchKeySuggestions` awaited the API without a `catch`. Since the debouncer invokes it fire-and-forget, any failure became an unhandled rejection. Fatal under Node 22; this crashed the Jest worker when a sibling test (e.g. `QueryV2.test.tsx`) didn't mock the endpoint.

#### Fix Strategy
- In `autoSuggestions`: parse the text before the first `.` as `head`; scoped iff any cached key has `opt.fieldContext === head`. The key branch then splits into two explicit display paths: scoped (`<head>.` → prefixed entries from that context) and bare (substring match on field names).
- Wrapped the fetch `await` in a try/catch; `console.warn`s in dev so real bugs don't disappear silently.
- Routed fetches through `queryClient.fetchQuery` with a stable queryKey and 5-min `staleTime`.
- Extended the key-suggestions `fieldContext` union in `querySuggestions/types.ts` to include `log` and `body`.

---

### 🧪 Testing Strategy

- Existing `QuerySearch.test.tsx` suite passes; `QueryV2.test.tsx` (which previously crashed the worker via unhandled rejection) now runs cleanly.
- Full frontend `yarn jest`: 4240 tests pass across multiple runs.
- Manual verification: typed `att`, `attribute`, `attribute.`, `attribute.foo`, `log.`, `body.`; confirmed the displayed list and that no redundant fetches fired.

---

- **Blast radius:** QueryBuilder V2 filter-expression autocomplete only. No backend or API-shape changes.
- **Potential regressions:**
  - Users who relied on the `main` behavior of substring-matching `attribute` inside labels like `attribute.foo` will no longer see matches for the bare text `attribute` — they now type `attribute.` for the scoped view. This is the intended behavior.
  - Cold-start caveat: on a fresh editor with nothing cached, `attribute.` would momentarily parse as bare because no `fieldContext=attribute` key is cached yet. Mitigated by the mount-effect that fires an empty-searchText fetch before the user can realistically type.
- **Rollback plan:** Pure frontend change — revert the commit and redeploy the bundle.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Filter-expression key autocomplete now understands field-context prefixes: typing `<context>.` (e.g. `attribute.`) scopes suggestions to keys in that context. Recognized contexts come from what the backend actually returns, not a hardcoded list. |

---

### 📋 Checklist
- [x] Manually tested
- [x] Breaking changes documented (behavior change noted above)

---

## 👀 Notes for Reviewers

- No stored "known contexts" set, no parse helper. The detection is four lines inline in `autoSuggestions`: slice off the head, check `cachedKeys.some(opt => opt.fieldContext === head)`. The backend's response is the single source of truth.
- The try/catch around `fetchKeySuggestions` is intentional — comment explains why. In non-production it `console.warn`s, so thrown bugs still surface. Please don't unwrap it without replacing with an equivalent safety net on the debounced call sites.
- `staleTime: 5 * 60 * 1000` on the React Query cache is deliberate. If users ever report "I can't find a field I just started emitting" we can dial this down — probably not worth pre-optimizing now.

---


srikanthccv:
@aks07 / @ahrefabhi please review.

ahrefabhi (commented):
Changes look fine and I've also tested this locally, it's working.
@tushar-signoz can you also add tests for these changes?

PR #10946, opened 2026-04-16
feat(tanstack-table): promote logs table component to /components

H4ad:
## Pull Request

---

This change promotes the table introduces just for Logs to be our generic table component, this implementation adds support for:
- Expanded Rows
- Query Params Support
  - Page/PageSize/Expanded/OrderBy is now persisted on URL
- Resize persistence
- Skeleton Loader when paginated
- Pagination support

And also brings features that already existed:
- Column Resizing
- Column Pinning
- Column Reorder

The headers borders were removed in this change to be more compatible with other tables across the product, I also removed the background of resizer to only appear when mouse hover.

Before:

image

After:

> General usage of the logs table

https://github.com/user-attachments/assets/b339027b-a519-4eee-90e3-84bf4b99aae2

> Show live logs

https://github.com/user-attachments/assets/659176c1-6b3d-4675-b9e8-1c5dd8f92641

Closes https://github.com/SigNoz/engineering-pod/issues/4526

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Logs and Logs Live View
- Potential regressions: A bug introduced during table migration
- Rollback plan: Revert this commit or fix the bug.

---

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | We extracted the table component from logs to be able to use across the product. |

---

### 📋 Checklist
- [x] Manually tested


frontend/src/components/Logs/TableView/useLogsTableColumns.tsx
```diff
@@ -0,0 +1,114 @@
+import type { ReactElement } from 'react';
+import { useMemo } from 'react';
+import TanStackTable from 'components/TanStackTableView';
+import { DATE_TIME_FORMATS } from 'constants/dateTimeFormats';
+import { getSanitizedLogBody } from 'container/LogDetailedView/utils';
+import { FontSize } from 'container/OptionsMenu/types';
```
aks07, line 6: I dont see this file used anywhere

frontend/src/container/LogsExplorerList/ColumnView/ColumnView.tsx
```diff
@@ -221,7 +221,7 @@ function ColumnView({
 		onColumnOrderChange(formattedColumns);
 	};
 
-	const handleRowClick = (row: Row<Record<string, unknown>>): void => {
+	const handleRowClick = (row: Row<Record<string, string>>): void => {
```
aks07: This file is also not used.

aks07:
https://github.com/user-attachments/assets/3b012d47-e2ca-470b-8092-abef69f2aed1

changeing max line, adding/deleting col is polluting browser history

aks07:
since we introduced a new localStorage key: @signoz/table-columns/LOGS_LIST_COLUMNS. Make sure we keep this backward compatible else this would reset for existing users. We can add a temporary migration code for this

aks07:
There is an existing issue which is not a bug but codewise is not correct. we have LOGS_LIST_OPTIONS that shows selectedColumns, LOGS_LIST_COLUMNS which also show selected columns. both are not even synced so may show wrong info

and now we have the following keys in local storage: LOGS_LIST_COLUMN_SIZING and @signoz/table-columns/LOGS_LIST_COLUMNS for column sizing.

aks07:
Body is not resizing 

https://github.com/user-attachments/assets/92b710b6-88c3-4772-8933-72297830c139

aks07:
Also may be i am wrong here but seems the perf dropped a bit.

H4ad:
> changeing max line, adding/deleting col is polluting browser history

This is a bug that already exists in the current component, was not introduced in this change.

> since we introduced a new localStorage key: @signoz/table-columns/LOGS_LIST_COLUMNS. Make sure we keep this backward compatible else this would reset for existing users. We can add a temporary migration code for this

This is a new store for resizing columns and other props, but its synced with the old store. The columns are migrated and we will only loose the sizing configuration.

Font-size and line clamp still preserved and stored by the useOptionsMenu.

> Body is not resizing

Fixed the issue causing the body to not resize.

> Also may be i am wrong here but seems the perf dropped a bit.

With React Scan, it indeed got a little worse, don't know exactly the reason.

Without the React Scan, only profilling by Chrome, the perf is almost the same, still re-rendering more than I wanted, and the most expensive part is the parsing body as HTML.

But it's not in a point that makes the component unusable.

frontend/src/components/Logs/TableView/useLogsTableColumns.tsx
H4ad: useLogsTableColumns is used by frontend/src/container/LiveLogs/LiveLogsList/index.tsx

aks07 (dismissed):

aks07 (dismissed):

aks07 (approved):

PR #10950, opened 2026-04-16
fix: order by ignored in formula query

tushar-signoz:
## Pull Request

---

### Summary
When a formula query produces scalar results (used in table, pie, and value panels), the `Order` and `Limit` specified on the formula were silently ignored. This is because `applyFormulas` calls `applySeriesLimit` only on `*TimeSeriesData` results — scalar formula results bypass that path entirely.

This fix applies `ApplySeriesLimit` to the evaluated formula series **before** converting them back to scalar format in `processScalarFormula`, ensuring ordering and limit are respected for all panel types.

#### Issues closed by this PR

---

### Change Type

- [x] 🐛 Bug fix

---

#### Root Cause
`processScalarFormula` evaluates the formula and then converts the result to `ScalarData`, but it never applied `formula.Order` or `formula.Limit`. The `applySeriesLimit` call in the outer `applyFormulas` function only operates on `*TimeSeriesData`, so scalar formula results were always returned in an arbitrary order with no limit applied.

#### Fix Strategy
Call `qbtypes.ApplySeriesLimit(formulaSeries, formula.Order, formula.Limit)` on the evaluated formula series inside `processScalarFormula`, right before the scalar conversion step. This mirrors the ordering/limiting that already happens for time-series formula results.

---

### Testing Strategy

- Manual verification: confirmed that table/pie/value panels now respect order and limit on formula queries.

---

### Risk & Impact Assessment

- Blast radius: limited to formula queries that produce scalar results (table, pie, value panels).
- Potential regressions: none expected — this adds ordering that was previously missing; no existing behavior is changed.
- Rollback plan: revert the single commit.

---

### Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Formula queries now correctly apply ordering and limit for table, pie, and value panels |

---

### Checklist
- [x] Manually tested


nityanandagohain (approved):

ankitnayan:
>This is because applyFormulas calls applySeriesLimit only on *TimeSeriesData results — scalar formula results bypass that path entirely.

Show me proof. I see `applySeriesLimit` applied for both https://github.com/SigNoz/signoz/blob/main/pkg/querier/postprocess.go#L328-L341

>Call qbtypes.ApplySeriesLimit(formulaSeries, formula.Order, formula.Limit) on the evaluated formula series inside processScalarFormula, right before the scalar conversion step. This mirrors the ordering/limiting that already happens for time-series formula results.

There is no `ApplySeriesLimit` call inside the `processTimeSeriesFormula` function as mentioned in the above comment. What am I missing?

tushar-signoz:
> > This is because applyFormulas calls applySeriesLimit only on *TimeSeriesData results — scalar formula results bypass that path entirely.
> 
> Show me proof. I see `applySeriesLimit` applied for both https://github.com/SigNoz/signoz/blob/main/pkg/querier/postprocess.go#L328-L341
> 
> > Call qbtypes.ApplySeriesLimit(formulaSeries, formula.Order, formula.Limit) on the evaluated formula series inside processScalarFormula, right before the scalar conversion step. This mirrors the ordering/limiting that already happens for time-series formula results.
> 
> There is no `ApplySeriesLimit` call inside the `processTimeSeriesFormula` function as mentioned in the above comment. What am I missing?



> > This is because applyFormulas calls applySeriesLimit only on *TimeSeriesData results — scalar formula results bypass that path entirely.
> 
> Show me proof. I see `applySeriesLimit` applied for both https://github.com/SigNoz/signoz/blob/main/pkg/querier/postprocess.go#L328-L341
> 
> > Call qbtypes.ApplySeriesLimit(formulaSeries, formula.Order, formula.Limit) on the evaluated formula series inside processScalarFormula, right before the scalar conversion step. This mirrors the ordering/limiting that already happens for time-series formula results.
> 
> There is no `ApplySeriesLimit` call inside the `processTimeSeriesFormula` function as mentioned in the above comment. What am I missing?


@ankitnayan  The method itself has type assertion and returns early without any change.
https://github.com/SigNoz/signoz/blob/c6683e075e41be4aaaed9a73ffac5040e965144a/pkg/querier/postprocess.go#L271

Now that you have pointed out, I'll update this section as well.

ankitnayan:
>@ankitnayan The method itself has type assertion and returns early without any change.

Right.. so that is the real problem. Isn't it? 

Infact, if you look deeply into the code applying order by and limit on timeseries is not the same as applying orderby and limit on scalar results. If they differ, we should create a separate function for applying limit on scalar types. Try to understand the logic at https://github.com/SigNoz/signoz/blob/main/pkg/types/querybuildertypes/querybuildertypesv5/series_limit.go

tushar-signoz:
> > @ankitnayan The method itself has type assertion and returns early without any change.
> 
> Right.. so that is the real problem. Isn't it?
> 
> Infact, if you look deeply into the code applying order by and limit on timeseries is not the same as applying orderby and limit on scalar results. If they differ, we should create a separate function for applying limit on scalar types. Try to understand the logic at https://github.com/SigNoz/signoz/blob/main/pkg/types/querybuildertypes/querybuildertypesv5/series_limit.go

The codebase doesn't implement methods for scalar results and only re-uses methods implemented for timeseries data. 
Scalar Data is converted to (1 datapoint with zero timestamp) timeseries data and then uses all the timeseries methods.

tushar-signoz:
@srikanthccv Added an integration test

pkg/querier/postprocess_test.go
```diff
@@ -1,15 +1,158 @@
...
+
+		out := q.applyFormulas(context.Background(), makeInputs(), makeReq(formula))
+		got, ok := out["F1"]
+		require.True(t, ok)
+		scalar, ok := got.Value.(*qbtypes.ScalarData)
+		require.True(t, ok)
+
+		// Without the fix, rows came out in alphabetical label order (a, b, c)
+		// regardless of direction. After the fix we expect descending values:
+		// a=101, b=11, c=2.
```
srikanthccv: This comment is incorrect of the case this tests.

srikanthccv (commented):
LGTM except incorrect comment for second test

pkg/querier/postprocess_test.go
tushar-signoz: Removed this comment

therealpandey (approved):

srikanthccv (approved):

PR #10953, opened 2026-04-16
chore: added changes for pinning tooltip with a shortcut key

### LINKED ISSUE #3898: [FE] Tooltip in charts are not scrollable
For high cardinal charts the list of labels overflow and some of the labels are not visible

image

ahrefabhi:
## Pull Request

---

### 📄 Summary

Replaces click-based tooltip pinning with a keyboard shortcut (`l` by default) and redesigns the tooltip UI into focused sub-components with clearer visual affordances.

Previously, users had to click the chart overlay to pin the tooltip, which conflicted with other click interactions. With this change, hovering over a chart and pressing the pin key locks the tooltip in place without requiring a click.

**Keyboard / pinning mechanics:**
- Adds a configurable `pinKey` prop (`ChartWrapper`, `TooltipPlugin`) defaulting to `'l'`
- Enables `canPinTooltip` on `TimeSeriesPanel` and `BarPanel` so dashboard panels support pinning out of the box
- Fixes a race condition where `setSeries`/`setLegend` hooks could clear `hoverActive` while the tooltip was pinned, causing it to disappear
- When pinned with no custom `pinnedTooltipElement`, the last hover contents are kept visible instead of blanking out
- Adds `onClick` callback prop to `ChartWrapper` / `TooltipPlugin` for consumers that need to react to chart clicks

**Tooltip UI redesign:**
- Extracted monolithic `Tooltip` into composable sub-components: `TooltipHeader`, `TooltipFooter`, `TooltipList`, `Kbd`
- `TooltipHeader` shows the timestamp alongside a **Pinned** (pin icon) status badge
- `TooltipFooter` renders contextual keyboard hints — _"Press L to lock"_ when hovering, _"Press L or Esc to unlock"_ + an **Unpin** button when pinned
- `Kbd` is a reusable styled `<kbd>` element used in footer hints

image
image

https://github.com/user-attachments/assets/fbf560a0-3d81-40c5-9b78-47b5e4e03a41

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/4675
Closes https://github.com/SigNoz/signoz/issues/3898

---

### ✅ Change Type

- [x] ✨ Feature
- [x] ♻️ Refactor

---

### 🐛 Bug Context

N/A

---

### 🧪 Testing Strategy

- **Tests added/updated:**
  - All existing pin-behaviour tests updated to use `KeyboardEvent('keydown')` instead of `MouseEvent('click')`
  - New `Tooltip` component tests covering header/footer rendering in pinned vs live state
  - New edge-case suite for `TooltipPlugin` covering:
    - Negative cursor coords (cursor off-screen guard)
    - Pin key pressed when hover is not active
    - Custom `pinKey` — wrong key does not pin, correct key does
    - `keydown` listener not registered when `canPinTooltip` is `false`
    - `keydown` pin listener removed on component unmount
- **Manual verification:** Hover over a Time Series / Bar panel → tooltip shows _Live_ status + "Press L to lock" hint. Press `l` → tooltip pins and shows _Pinned_ status + Unpin button. Press `Escape` or click the Unpin button to dismiss.
- **Edge cases covered:** Off-screen cursor, no active hover, custom key, listener lifecycle, single-series vs multi-series layout

---

- **Blast radius:** `TooltipPlugin`, `ChartWrapper`, `TimeSeriesPanel`, `BarPanel`, and the new `Tooltip` sub-components (`TooltipHeader`, `TooltipFooter`, `TooltipList`, `Kbd`). No other panel types or public APIs changed.
- **Potential regressions:** Panels that previously relied on click-to-pin will now require the keyboard shortcut. The old click handler on `plot.over` is removed entirely.
- **Rollback plan:** Revert the commits on this branch; no data migrations or flag changes required.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Dashboard chart tooltips can now be pinned by pressing `l` while hovering, replacing the previous click-to-pin interaction. The tooltip now shows a Live/Pinned status badge and contextual keyboard hints. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- The `pinKey` prop accepts any single character and is compared case-insensitively, so `'L'` and `'l'` both work.
- The `handleKeyDown` listener is attached to `document` (capture phase) rather than `plot.over` so it fires even if focus is elsewhere on the page.
- `updateHoverState` bails early when `controller.pinned` is `true` — this prevents the async `setSeries`/`setLegend` hooks from racing and clearing `hoverActive` while a tooltip is pinned.
- `TooltipFooter` renders even when not pinned (to show the "Press L to lock" affordance) — this is intentional as it teaches discoverability without being intrusive.
- The `Kbd` component handles both dark/light mode via the `useIsDarkMode` hook, matching the rest of the tooltip's theming pattern.

---


frontend/src/lib/uPlotV2/plugins/TooltipPlugin/types.ts
```diff
@@ -37,6 +37,8 @@ export interface TooltipLayoutInfo {
 export interface TooltipPluginProps {
 	config: UPlotConfigBuilder;
 	canPinTooltip?: boolean;
+	/** Key that pins the tooltip while hovering. Defaults to DEFAULT_PIN_TOOLTIP_KEY ('l'). */
+	pinKey?: string;
```
AshwinBhatkal, line 48: I'm thinking if this should be enum 🤔

frontend/src/lib/uPlotV2/plugins/TooltipPlugin/TooltipPlugin.tsx
```diff
@@ -259,68 +261,73 @@ export default function TooltipPlugin({
...
+		// synthetic event built from the current uPlot cursor position so the
+		// same series-resolution logic as click-pinning is reused.
+		const handleKeyDown = (event: KeyboardEvent): void => {
+			if (event.key.toLowerCase() !== pinKey.toLowerCase()) {
+				return;
+			}
 			const plot = getPlot(controller);
 			if (
-				plot &&
-				event.target === plot.over &&
```
AshwinBhatkal, line 294: What case was this covering earlier? Should this not be present when is is not pinned?

AshwinBhatkal (commented):
There is a behaviour today that pops up shortcuts on long press of the Shift key. We should probably look at having something like that than a bare key press. What do you think?

frontend/src/lib/uPlotV2/components/Tooltip/components/Kbd/Kbd.module.scss
```diff
@@ -0,0 +1,29 @@
...
+	background: var(--bg-ink-200);
+	border: 1px solid var(--bg-slate-200);
+	border-bottom-width: 2px;
+	color: var(--bg-vanilla-100);
+	font-family: 'Space Mono', monospace;
+	font-size: 10px;
+	font-weight: 600;
+	line-height: 1;
+
+	&.lightMode {
```
AshwinBhatkal: We would want to avoid nested styles with css modules

frontend/src/lib/uPlotV2/components/Tooltip/components/Kbd/Kbd.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { useIsDarkMode } from 'hooks/useDarkMode';
+
+import Styles from './Kbd.module.scss';
+
+interface KbdProps {
+	children: React.ReactNode;
+	isPinned?: boolean;
+}
+
+export default function Kbd({
```
AshwinBhatkal: This is generic. Can you make it KbdInput? We should move it to signoz components if we can. Also, we should be passing classname to override instead of isPinned

frontend/src/lib/uPlotV2/components/Tooltip/components/Kbd/Kbd.tsx
```diff
@@ -0,0 +1,27 @@
...
+	children,
+	isPinned = false,
+}: KbdProps): JSX.Element {
+	const isDarkMode = useIsDarkMode();
+
+	return (
+		<kbd
+			className={cx(Styles.kbd, {
+				[Styles.lightMode]: !isDarkMode,
```
AshwinBhatkal: We may not need this anymore after semantic tokens introduction. Can you check it once?

frontend/src/lib/uPlotV2/components/Tooltip/components/TooltipFooter/TooltipFooter.module.scss
```diff
@@ -0,0 +1,67 @@
+.footer {
```
AshwinBhatkal, line 1: Comments in the previous 2 files apply for everything

frontend/src/lib/uPlotV2/components/Tooltip/components/TooltipFooter/TooltipFooter.tsx
```diff
@@ -0,0 +1,62 @@
...
+					<>
+						<span>Press</span>
+						<Kbd>L</Kbd>
+						<span>to lock</span>
+					</>
+				)}
+			</div>
+
+			{isPinned && (
+				<button
```
AshwinBhatkal: Can we use the one from signoz ui

frontend/src/lib/uPlotV2/plugins/TooltipPlugin/TooltipPlugin.styles.scss
AshwinBhatkal: move this to css modules as well? small file

frontend/src/lib/uPlotV2/components/Tooltip/components/TooltipFooter/TooltipFooter.tsx
ahrefabhi: Done

frontend/src/lib/uPlotV2/components/Tooltip/components/TooltipFooter/TooltipFooter.module.scss
```diff
@@ -0,0 +1,37 @@
...
+	gap: 5px;
+	font-size: 11px;
+	color: var(--bg-vanilla-400);
+}
+
+.hintLightMode {
+	color: var(--bg-ink-400);
+}
+
+.kbdPinned {
```
AshwinBhatkal: Should be in kbd styles

frontend/src/lib/uPlotV2/components/Tooltip/components/TooltipList/TooltipList.tsx
```diff
@@ -0,0 +1,48 @@
...
+				: Math.min(content.length * TOOLTIP_ITEM_HEIGHT, LIST_MAX_HEIGHT),
+		[totalListHeight, content.length],
+	);
+
+	return (
+		<Virtuoso
+			className={cx(Styles.list, !isDarkMode && Styles.listLightMode)}
+			data-testid="uplot-tooltip-list"
+			data={content}
+			style={{ height, width: '100%' }}
```
AshwinBhatkal: nit: can have width in class name

frontend/src/lib/uPlotV2/plugins/TooltipPlugin/TooltipPlugin.tsx
```diff
@@ -159,13 +163,14 @@ export default function TooltipPlugin({
 
 		// Attach / detach global listeners when pin state changes so
 		// we can detect when the user interacts outside the tooltip.
+		// Keyboard unpinning is handled exclusively in handleKeyDown so
+		// that only L (toggle) and Escape (release) can dismiss — not
```
AshwinBhatkal: P?

frontend/src/lib/uPlotV2/components/Tooltip/components/TooltipFooter/TooltipFooter.module.scss
```diff
@@ -0,0 +1,31 @@
+.footer {
+	display: flex;
+	align-items: center;
+	justify-content: space-between;
+	gap: 10px;
+	padding: 7px 12px;
+	border-top: 1px solid var(--bg-slate-400);
+	background: var(--l2-background);
```
AshwinBhatkal, line 8: It was recently pointed out to me to use semantic tokens for these 😂

frontend/src/lib/uPlotV2/components/Tooltip/components/TooltipFooter/TooltipFooter.tsx
```diff
@@ -0,0 +1,66 @@
...
+	isPinned: boolean;
+	dismiss: () => void;
+}
+
+export default function TooltipFooter({
+	pinKey = DEFAULT_PIN_TOOLTIP_KEY,
+	isPinned,
+	dismiss,
+}: TooltipFooterProps): JSX.Element {
+	const isDarkMode = useIsDarkMode();
```
AshwinBhatkal: This will not be required if we use semantic tokens

frontend/src/lib/uPlotV2/components/Tooltip/components/TooltipFooter/TooltipFooter.module.scss
AshwinBhatkal: across all scss files

AshwinBhatkal (dismissed):

AshwinBhatkal (approved):

PR #10954, opened 2026-04-16
feat(member): add v2 reset password token endpoints and show invite expiry status

vikrantgupta25:
## Summary

- Adds v2 REST endpoints for reset password tokens (`GET /api/v2/users/{id}/reset_password_tokens` to read, `PUT` to create/regenerate) with org-scoped queries and UUID validation
- Adds v2 endpoint for changing own password (`PUT /api/v2/users/me/factor_password`) — user ID is now derived from auth context instead of being passed in the request body
- Shows admins whether an invite token is valid or expired before they click — button label changes to "Copy Invite Link", "Regenerate Invite Link", or "Generate Invite Link" accordingly
- Displays token expiry in the reset link dialog so admins know when the link will stop working
- Removes old `POST /api/v1/changePassword/{id}` endpoint and marks `GET /api/v1/getResetPasswordToken/{id}` as deprecated
- Removes dead v1 frontend code (`changeMyPassword.ts`, `getResetPasswordToken.ts`, `getResetPasswordToken` types)
- Standardizes handler receiver naming (`handler` instead of `h`) across all user handlers
- Migrates integration tests to use the new v2 endpoints

## Test plan

- [x] Backend: `make go-build-community` and `make go-test` pass
- [x] Frontend: `yarn build` and `yarn jest` pass
- [x] Python: `make py-fmt` and `make py-lint` pass
- [x] Integration: `tests/integration/src/passwordauthn/03_password.py` updated to use v2 endpoints

Closes: https://github.com/SigNoz/platform-pod/issues/2084


pkg/modules/user/user.go
```diff
@@ -113,6 +116,8 @@ type Handler interface {
 
 	// Reset Password
 	GetResetPasswordToken(http.ResponseWriter, *http.Request)
+	GetResetPasswordTokenV2(http.ResponseWriter, *http.Request)
```
vikrantgupta25: there is nothing as v2, make the older one as deprecated and the new one as default

frontend/src/components/EditMemberDrawer/EditMemberDrawer.tsx
```diff
@@ -121,6 +145,27 @@ function EditMemberDrawer({
 		applyDiff,
 	} = useMemberRoleManager(member?.id ?? '', open && !!member?.id);
 
+	// Token status query for invited users
+	const {
+		data: tokenQueryData,
+		isLoading: isLoadingTokenStatus,
+		isError: tokenNotFound,
```
SagarRajput-7, line 152: NIT: but checking here that any error is considered notfound and (no error is shown) and we are showing - "'Generate Invite Link' text to the user.

frontend/src/components/EditMemberDrawer/EditMemberDrawer.tsx
```diff
@@ -121,6 +145,27 @@ function EditMemberDrawer({
...
+	} = useGetResetPasswordToken(
+		{ id: member?.id ?? '' },
+		{ query: { enabled: open && !!member?.id && isInvited } },
+	);
+
+	const existingToken = tokenQueryData?.data;
+	const isTokenExpired =
+		existingToken != null &&
+		new Date(String(existingToken.expiresAt)) < new Date();
```
SagarRajput-7, line 161: should we add an explicit null guard: existingToken.expiresAt != null && before the date comparison

as for a undefined/null response --> invalid date < new Date() ---> though it turn to false, which works in a way, but please confirm.

frontend/src/components/EditMemberDrawer/EditMemberDrawer.tsx
```diff
@@ -55,6 +56,27 @@ function getDeleteTooltip(
...
+		return 'Copy Invite Link';
+	}
+	if (isExpired) {
+		return 'Regenerate Invite Link';
+	}
+	if (notFound) {
+		return 'Generate Invite Link';
+	}
+	return 'Copy Invite Link';
+}
```
SagarRajput-7, line 78: could we have a case where api succeeds with 200 but the token is null or empty data? if yes then we fall back to "Copy Invite Link" which then would be misleading

frontend/src/components/EditMemberDrawer/EditMemberDrawer.tsx
vikrantgupta25: Yes since this new API is just checkin in. In case we aren't able to then it fallbacks to the original plan

frontend/src/components/EditMemberDrawer/EditMemberDrawer.tsx
vikrantgupta25: This would always be present, the openAPI spec isn't updated because of older reliance hence no check here

frontend/src/components/EditMemberDrawer/EditMemberDrawer.tsx
vikrantgupta25: this should never happen. API is responsible for this

SagarRajput-7 (approved):

therealpandey (approved):

YounixM (approved):

PR #10956, opened 2026-04-16
feat(infra-monitoring): v2 hosts list - return counts of active & inactive hosts for custom group by attributes

nikhilmantri0902:
## Pull Request

---

### 📄 Summary

> Why does this change exist?  

When a user passes custom group-by attributes list not containting host.name ( or pod.uid in case of pods), the hosts/pods list APIs ( even the v2 ) returns status as empty. In this case better would be to dynamically replace the single status column by a column showing count of hosts in active and inactive statuses for each group.

When viewing with custom group by attributes like os.type, or cloud.provider etc., the user can get to see how many hosts from this grouping are actively sending data and how many have become inactive. 

Part of https://github.com/SigNoz/engineering-pod/issues/4609

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Manual verification:

---

- Blast radius: Independent merge into v2 hosts list api, and that api is not used by frontend yet, so blast radius is just this new API. 
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|
| Change Type | Feature Maintenance |
| Description | Once UI is built on top of this user can see counts of hosts in various statuses on custom grouping |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

The approach uses a subquery from active hosts method and a subquery for fingerprints from samples table. Because these were already used in existing methods, I have extracted out the common pieces of this query building into smaller method and used at both places. 

---


pkg/modules/inframonitoring/implinframonitoring/hosts.go
```diff
@@ -384,12 +383,18 @@ func (m *module) getHostsTableMetadata(ctx context.Context, req *inframonitoring
 
 // buildHostRecords constructs the final list of HostRecords for a page.
 // Groups that had no metric data get default values of -1.
+//
+// hostCounts is nil when host.name is in the groupBy — in that case, counts are
+// derived directly from activeHostsMap (1/0 per host). When non-nil (custom groupBy
+// without host.name), counts are looked up from the map.
 func (m *module) buildHostRecords(
```
namanv3: nit: because this does not actually need any field inside `m *module`, maybe we can drop the `(m *module)`? usually when I see a function that's not `<something>.FunctionName` but just `FunctionName` it gives an indication v easily that this function doesn't make any external system calls. what do you think?

pkg/modules/inframonitoring/implinframonitoring/module.go
```diff
@@ -132,13 +135,28 @@ func (m *module) HostsList(ctx context.Context, orgID valuer.UUID, req *inframon
 	if req.Filter != nil {
 		hostsFilterExpr = req.Filter.Expression
 	}
+	pageGroupsFilterExpr := buildPageGroupsFilterExpr(pageGroups)
```
namanv3: nit: did you mean to pass pageGroupsFilterExpr to buildFullQueryRequest earlier but then ended up calling buildPageGroupsFilterExpr in buildFullQueryRequest directly?

namanv3 (commented):
this change looks good as such, just minor nits that you can take into account if you want. once this goes into the other branch, let's just think how this module method can be made simpler though, cuz there's a lot of conditions bouncing around that can be difficult for someone new to follow.

namanv3 (approved):

pkg/modules/inframonitoring/implinframonitoring/hosts.go
nikhilmantri0902: Yes Naman, this actually makes sense, the method is not making any module's depdencies calls also, so we can simply define it as a function and not method of the module. Changing that way.

pkg/modules/inframonitoring/implinframonitoring/module.go
nikhilmantri0902: Oh damn, good catch. THe code structuring makes it unclear, refactoring.

nikhilmantri0902: I have refactored the code @namanv3 .

PR #10967, opened 2026-04-16
fix: cache memory leak

tushar-signoz:
## Pull Request

---

The in-memory (ristretto) cache has no effective eviction: every entry is admitted with a hardcoded cost of `1`, so the configured `MaxCost` (64 MiB by default) is only reached after ~67M entries — effectively never. Under anomaly-rule load the heap was ~94% `bytes.Clone` and grew unbounded (observed: 991 MB → 1205 MB in ~2 min).

This PR makes `MaxCost`-based eviction actually work by passing the real byte cost of each entry into ristretto, and fixes two observability gaps that hid the leak.

_Pprof snapshot from the incident (heap by `bytes.Clone`):_

```
Type: inuse_space
Total: 1205.81MB
      flat  flat%        cum   cum%
 1130.83MB 93.78%  1130.83MB 93.78%  bytes.Clone
```

Call path: `AnomalyRule.Eval → BaseSeasonalProvider.getResults → executeWithCache → bucketCache.Put → memorycache.Set → CachedData.Clone → CachedBucket.Clone → bytes.Clone`.

#### Issues closed by this PR

N/A

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause

`pkg/cache/memorycache/provider.go` called `cc.SetWithTTL(..., toCache, 1, ttl)` on both the cloneable and non-cloneable paths. The cost argument is what ristretto uses for `MaxCost` accounting. With cost fixed at `1`, the cache has no idea how much memory it is holding and admission/eviction by `MaxCost` is a no-op.

The leak surfaced via anomaly rules because each eval cycle runs 6 queries, each of which performs up to 3 deep clones per `executeWithCache` call. Each clone deep-copies the `CachedBucket.Value` (`json.RawMessage` of serialized time-series data). With no eviction, the stored clones accumulate indefinitely.

Two observability bugs compounded the problem:
1. `signoz.cache.total.cost` was publishing `cc.MaxCost()` — the **configured ceiling**, a constant — rather than current retained cost. The actual "how full is the cache" signal was not emitted anywhere.
2. Ristretto's monotonic counters (`Hits`, `Misses`, `CostAdded`, `CostEvicted`, `Keys*`, `Sets*`, `Gets*`) were registered as `Int64ObservableGauge`, so `rate()` / `increase()` in PromQL returned nonsense.

#### Fix Strategy

1. **Add `Size() int64` to `cachetypes.Cloneable`.** Non-cloneable entries already know their marshaled length; cloneable entries skip marshaling by design, so they must self-report size (otherwise we'd force serialization just to measure, defeating the clone-path optimization).
2. **Pass the real cost to `cc.SetWithTTL`** on both branches, clamped to `≥ 1` (ristretto treats cost 0 specially).
3. **Implement `Size()` on every `Cloneable`:**
   - `querybuildertypesv5.CachedData` — the hot path: sums `len(bucket.Value)` across buckets + warnings. This is exactly where pprof showed 94% of the heap.
   - `tracedetailtypes.WaterfallCache`, `model.GetWaterfallSpansForTraceWithMetadataCache`, `model.GetFlamegraphSpansForTraceCache` — approximated as span-count × per-span constant.
   - Test doubles in `memorycache` / `rediscache`.
4. **Metrics fixes:**
   - New gauge `signoz.cache.cost.used` = `CostAdded − CostEvicted` (current retained cost).
   - Kept `signoz.cache.total.cost` but clarified description (configured ceiling).
   - Converted all monotonic fields from `Int64ObservableGauge` to `Int64ObservableCounter`.

---

### 🧪 Testing Strategy

- Tests added/updated:
  - New `TestCloneableCostTriggersEviction` in `pkg/cache/memorycache/provider_test.go`: inserts 32 × 256 KiB entries into a cache with `MaxCost = 1 MiB` and asserts that `KeysEvicted + SetsRejected > 0` and `CostAdded − CostEvicted ≤ MaxCost`. Fails on `main` (no eviction), passes with this PR.
  - Existing cache/querier/model tests pass unchanged.
- Manual verification:
  - `go build ./...` clean.
  - `go test ./pkg/cache/... ./pkg/querier/... ./pkg/types/querybuildertypes/... ./pkg/types/tracedetailtypes/...` passes.
  - Follow-up verification run (reproducing the anomaly-rule load + fresh heap profile) is intended for post-merge in a staging env — I cannot reproduce the anomaly workload locally.
- Edge cases:
  - Zero-sized entries clamp to cost 1 (still counted, never bypass admission).
  - `Size()` is an estimate; overestimates cause slightly more eager eviction, underestimates admit more before eviction. Both are safe.

---

- **Blast radius:** All memory-cache users. `Cloneable` gains a new required method, so any out-of-tree implementer breaks at compile time. All in-tree implementers are updated in this PR.
- **Potential regressions:**
  - **Cache hit rate may drop** once eviction actually works. This is expected, not a regression — the previous "high hit rate" was being paid for in unbounded memory. If hit rate drop is unacceptable, the remedy is to raise `MaxCost` (a config change, out of scope here).
  - **Existing dashboards that use cache metrics as gauges will show wrong values after this PR** because the instruments are now counters. Dashboards must be updated to use `rate()` / `increase()`. The metric names are unchanged.
- **Rollback plan:** Straight revert — no data migration, no config changes required. The `Size()` interface addition is the only breaking surface and is internal.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | In-memory cache now respects its configured `MaxCost` limit; adds `signoz.cache.cost.used` metric for live cache usage. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- **`Size()` accuracy is a trade-off, not a contract.** For `CachedData` it tracks the dominant byte cost (`bucket.Value`) exactly. For waterfall/flamegraph types it uses a per-span constant; tuning can come later if those caches grow enough to matter.
- **`MaxCost` default (64 MiB) may now be too small** for sites with heavy anomaly-rule usage, since eviction will actually kick in. I'd suggest raising it as a follow-up after watching `signoz.cache.cost.used` and eviction counters in production — but that's a config change, not part of this PR.
- **Why `Size()` on `Cloneable` and not `Cacheable`:** non-cloneable entries produce a `[]byte` from `MarshalBinary` whose length is already known at admission time — no point making them implement a second method. Cloneable entries bypass marshaling, so they must self-report.


pkg/cache/memorycache/provider.go
```diff
@@ -112,11 +114,13 @@ func (provider *provider) Set(ctx context.Context, orgID valuer.UUID, cacheKey s
 	}
 
 	if cloneable, ok := data.(cachetypes.Cloneable); ok {
+		cost := cloneable.Size()
+		// Clamp to a minimum of 1: ristretto treats cost 0 specially and we
+		// never want zero-size entries to bypass admission accounting.
 		span.SetAttributes(attribute.Bool("memory.cloneable", true))
-		span.SetAttributes(attribute.Int64("memory.cost", 1))
+		span.SetAttributes(attribute.Int64("memory.cost", cost))
```
therealpandey, line 121:
```suggestion
```

This will mismatch with the actual cost being set (`max(cost,1)`).

pkg/cache/memorycache/provider_test.go
```diff
@@ -165,6 +169,50 @@ func TestSetGetWithDifferentTypes(t *testing.T) {
...
+		item := &LargeCloneable{Key: fmt.Sprintf("key-%d", i), Cost: perEntry}
+		assert.NoError(t, c.Set(context.Background(), orgID, fmt.Sprintf("key-%d", i), item, time.Minute))
+	}
+
+	metrics := c.(*provider).cc.Metrics
+	// Eviction (or admission rejection) must have kicked in: we wrote 32 entries
+	// each costing 256 KiB into a 1 MiB cache.
+	assert.Greater(t, metrics.KeysEvicted()+metrics.SetsRejected(), uint64(0),
+		"expected eviction or admission rejection once total cost exceeds MaxCost; got evicted=%d rejected=%d",
+		metrics.KeysEvicted(), metrics.SetsRejected())
```
therealpandey: We can write a better test here please. The assert semantics are very generic.

Let's add items to the cache and fill it up. Then let's try adding one explicitly and assert what is happening in the cache. Instead of asserting on metrics, see if ristretto itself can give something to assert on.

pkg/cache/memorycache/telemetry.go
```diff
@@ -7,17 +7,18 @@ import (
...
+	cacheMisses  metric.Int64ObservableCounter
+	costAdded    metric.Int64ObservableCounter
+	costEvicted  metric.Int64ObservableCounter
+	keysAdded    metric.Int64ObservableCounter
+	keysEvicted  metric.Int64ObservableCounter
+	keysUpdated  metric.Int64ObservableCounter
+	setsDropped  metric.Int64ObservableCounter
+	setsRejected metric.Int64ObservableCounter
+	getsDropped  metric.Int64ObservableCounter
+	getsKept     metric.Int64ObservableCounter
```
therealpandey, line 20: Have we confirmed all of these are counter metrics?

pkg/types/cachetypes/cacheable.go
```diff
@@ -18,6 +18,9 @@ type Cloneable interface {
 	// Creates a deep copy of the Cacheable. This method is useful for memory caches to avoid the need for serialization/deserialization. It also prevents
 	// race conditions in the memory cache.
 	Clone() Cacheable
+	// Size returns the approximate retained byte cost of this entry. Memory-backed
+	// caches use it as the ristretto cost so MaxCost-based eviction works.
+	Size() int64
```
therealpandey: Size is misleading here as this reflects the Cost we want to set in cache (which we are deriving from Size right now).

therealpandey (changes requested):

pkg/cache/memorycache/telemetry.go
tushar-signoz: https://github.com/wolfmetr/ristretto-prometheus/blob/main/options.go

tushar-signoz:
@therealpandey Addressed your comments

pkg/cache/memorycache/provider_test.go
```diff
@@ -165,6 +169,81 @@ func TestSetGetWithDifferentTypes(t *testing.T) {
...
+			}
+		}
+	}
+
+	// The full set's total cost (1.25 MiB) cannot fit, so at least one entry
+	// must be absent. This assertion reads cache state directly, not metrics.
+	require.Less(t, survivors, tip+1,
+		"all %d entries cannot coexist in a %d-byte cache", tip+1, maxCost)
+	require.LessOrEqual(t, retainedCost, maxCost,
+		"sum of survivor costs must respect MaxCost")
```
therealpandey: Use assert when you are actually asserting what the test should do.

Use require when setting up the test (mocks/ New calls etc)

pkg/cache/memorycache/provider_test.go
```diff
@@ -165,6 +169,81 @@ func TestSetGetWithDifferentTypes(t *testing.T) {
...
+	for i := 0; i <= tip; i++ {
+		if _, ok := cc.Get(ristrettoKey(i)); ok {
+			survivors++
+			if i == tip {
+				retainedCost += big
+			} else {
+				retainedCost += small
+			}
+		}
+	}
```
therealpandey:
```suggestion
```

Instead of this, why not add a new entry with cost > 1MiB so we exactly know this will be rejected? The test is harder to maintain and understand.

therealpandey (changes requested):

therealpandey (approved):

PR #10983, opened 2026-04-17
feat: adding migration AWS cloud integration regions config

swapnil-signoz:
### 📄 Summary (NOT AI)
- When AWS Integration is being configured there is an option to select all regions for monitoring. And backend stores `all`  sentinel in DB and NOT individual regions.
- We want to be explicit about this, ie when all regions are selected all the individual regions are stored in DB. This removes surprise for user when a new region is added it will be monitored by default and without noticing user.

**What problem does it solve, and why is this the right approach?**
- Adding migration will move the `all` sentinel value to currently supported regions

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only


swapnil-signoz:
Will be merged after frontend changes for AWS cloud integration is merged.
cc: @YounixM

pkg/sqlmigration/077_migrate_aws_all_regions.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		_ = tx.Rollback()
+	}()
+
+	rows, err := tx.QueryContext(ctx,
+		`SELECT id, config FROM cloud_integration WHERE provider = ? AND removed_at is NULL`,
```
vikrantgupta25: no raw queries please

vikrantgupta25 (changes requested):

pkg/sqlmigration/077_migrate_aws_all_regions.go
swapnil-signoz: addressed, please check again

pkg/sqlmigration/077_migrate_aws_all_regions.go
```diff
@@ -0,0 +1,115 @@
...
+	tx, err := db.BeginTx(ctx, nil)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		_ = tx.Rollback()
+	}()
+
+	var accounts []*cloudIntegrationAWSMigrationRow
```
vikrantgupta25:
```suggestion
	accounts := make([]*cloudIntegrationAWSMigrationRow)
```

vikrantgupta25 (changes requested):

pkg/sqlmigration/077_migrate_aws_all_regions.go
```diff
@@ -0,0 +1,116 @@
...
+		return err
+	}
+
+	idsToUpdate := make([]string, 0)
+	for _, account := range accounts {
+		var cfg awsConfig
+		if err := json.Unmarshal([]byte(account.Config), &cfg); err != nil {
+			continue
+		}
+		if containsAllRegion(cfg.Regions) {
```
vikrantgupta25:
```suggestion
		if slices.Contains(cfg.Regions, "all") {
```

pkg/sqlmigration/077_migrate_aws_all_regions.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (migration *migrateAWSAllRegions) Down(context.Context, *bun.DB) error {
+	return nil
+}
+
+func containsAllRegion(regions []string) bool {
+	return slices.Contains(regions, "all")
+}
+
+func allValidAWSRegions() []string {
```
vikrantgupta25:
```suggestion
func (migration *migrateAWSAllRegions) getAllRegionsConfig() []bytes {
```

vikrantgupta25 (changes requested):

therealpandey (approved):

vikrantgupta25 (approved):

PR #10984, opened 2026-04-17
fix: api monitoring cancel and run query bugs

AshwinBhatkal:
## Summary
- **listOverview API**: Accept optional `AbortSignal` parameter and pass to axios config
- **useListOverview hook**: Pass `signal` from react-query's queryFn context to API; add ERR_CANCELED retry skip
- **DomainList**: Wrap `handleRunQuery` with explicit `invalidateQueries` so clicking "Run Query" always re-fetches (even if URL params didn't change, e.g. after cancel)

## Bug Context

**Root cause of cancel not working**: `cancelQueries` was called on the react-query cache, but the underlying axios request had no `AbortSignal`, so the HTTP request kept running until the server responded.

**Root cause of run query not working**: `onStageRunQuery` called `handleRunQuery` which redirects with query-builder data to the URL. If the user cancelled and then clicked Run without changing anything, the URL stayed the same → `compositeData` stayed the same → query key stayed the same → no re-fetch. Now the handler explicitly invalidates the query before running.

**Fix strategy**: Follow the same AbortSignal + invalidate pattern used in `useGetQueryRange` and `MetricsExplorer Inspect`.

## Screenshots / Screen Recordings

https://github.com/user-attachments/assets/8c0ed6b3-6f30-44e2-90e7-2027f19c08d5

## Change Type
- [x] Bug fix

## Testing Strategy
- TypeScript compilation passes
- Manual: API Monitoring → run query → cancel → verify cancel button returns to Run → click Run → verify data loads
- Manual: API Monitoring → run query → without changing anything, click Run → verify query re-fetches

## Risk & Impact Assessment
- Low risk: scoped to API Monitoring domain list
- Backward compatible: `signal` param is optional on `listOverview`

### DIFF
--- a/frontend/src/hooks/thirdPartyApis/useListOverview.ts
+++ b/frontend/src/hooks/thirdPartyApis/useListOverview.ts
@@ -1,5 +1,6 @@
 import { useQuery, UseQueryResult } from 'react-query';
 import listOverview from 'api/thirdPartyApis/listOverview';
+import { isAxiosError } from 'axios';
 import { REACT_QUERY_KEY } from 'constants/reactQueryKeys';
 import { SuccessResponseV2 } from 'types/api';
 import APIError from 'types/api/error';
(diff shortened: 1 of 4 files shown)


frontend/src/hooks/thirdPartyApis/useListOverview.ts
```diff
@@ -20,12 +21,21 @@ export const useListOverview = (
...
+					show_ip: showIp,
+					filter,
+				},
+				signal,
+			),
+		retry: (failureCount, error): boolean => {
+			if (isAxiosError(error) && error.code === 'ERR_CANCELED') {
+				return false;
+			}
+			return failureCount < 3;
```
ahrefabhi, line 38: can we make 3 a constant, this has been used at multiple places in other PRs?

AshwinBhatkal: Yep. Another PR

AshwinBhatkal: Done in https://github.com/SigNoz/signoz/pull/11049

PR #10985, opened 2026-04-17
chore: remove unused caching primitives from waterfall v2 

krsoninikhil:
## Pull Request

---

We are caching the spans queries from db in memory to avoid querying again and again for the same trace but the cache was not getting used properly due [this bug](https://github.com/SigNoz/engineering-pod/issues/4648), this PR removes setting cache since it was not getting used anyway and only increasing the memory usage.
Alternate considered - https://github.com/SigNoz/signoz/pull/10986

**Why remove the caching altogether instead of fixing the cache utilization?**

In past we have seen OOMKilled issues due to high memory utilization, so a better solution is needed to cache the span data instead of caching it in memory and since we are not using the cache currently, then there is not reason to set the cache, instead we can directly look into adding redis without risking loosing of in-memory cache.

https://github.com/SigNoz/engineering-pod/issues/4648

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---


krsoninikhil:
Closing in favor of https://github.com/SigNoz/signoz/pull/10986. Related discussion https://signoz-team.slack.com/archives/C08P77ZDHV0/p1776673489408039?thread_ts=1776316814.795509&cid=C08P77ZDHV0

PR #10986, opened 2026-04-17
chore: remove setting of trace cache since it was not getting used

krsoninikhil:
## Pull Request

---

Remove setting of trace cache since it was not getting used and we want to introduce redis instead of relying on in-memory cache. Related [discussion](https://signoz-team.slack.com/archives/C08P77ZDHV0/p1776427442280699?thread_ts=1776316814.795509&cid=C08P77ZDHV0).
Alternate change considered to remove cache https://github.com/SigNoz/signoz/pull/10986

Why remove the caching altogether instead of fixing the cache utilization?

In past we have seen OOMKilled issues due to high memory utilization, so a better solution is needed to cache the span data instead of caching it in memory and since we are not using the cache currently, then there is not reason to set the cache, instead we can directly look into adding redis without risking loosing of in-memory cache.

https://github.com/SigNoz/engineering-pod/issues/4648

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/pkg/query-service/app/clickhouseReader/reader.go
+++ b/pkg/query-service/app/clickhouseReader/reader.go
@@ -947,7 +947,6 @@ func (r *ClickHouseReader) GetWaterfallSpansForTraceWithMetadata(ctx context.Con
 			return response, nil
 		}
 		totalSpans = uint64(len(searchScanResponses))
-		processingBeforeCache := time.Now()
 		for _, item := range searchScanResponses {
 			ref := []model.OtelSpanRef{}
 			err := json.Unmarshal([]byte(item.References), &ref)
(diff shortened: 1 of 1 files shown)


pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -1070,23 +1070,11 @@ func (r *ClickHouseReader) GetWaterfallSpansForTraceWithMetadata(ctx context.Con
...
-			SpanIdToSpanNodeMap:           spanIdToSpanNodeMap,
-			ServiceNameToTotalDurationMap: serviceNameToTotalDurationMap,
-			TraceRoots:                    traceRoots,
-			HasMissingSpans:               hasMissingSpans,
-		}
+		// TODO: set the span data (model.GetWaterfallSpansForTraceWithMetadataCache) in cache here
+		// removed existing cache usage since it was not getting used due to this bug https://github.com/SigNoz/engineering-pod/issues/4648
+		// and was causing out of memory issues https://github.com/SigNoz/engineering-pod/issues/4638
 
 		r.logger.Info("getWaterfallSpansForTraceWithMetadata: processing pre cache", "duration", time.Since(processingBeforeCache), "traceID", traceID)
```
vikrantgupta25: we can remove the unwanted loggers now as well ?

vikrantgupta25 (changes requested):

pkg/query-service/app/clickhouseReader/reader.go
krsoninikhil: Kept the log so processingBeforeCache variable declared at the top doesn't need to be removed. Which is only used in log.

vikrantgupta25 (approved):

PR #10411, opened 2026-02-25
fix: normalize version prefix in checkVersionState to fix false upgrade prompt

saravadeo:
### Problem

The `/status` page incorrectly shows "an upgrade is available" even when the user is running the latest version.

Fixes #7484

### Root Cause

The `checkVersionState` function compares the `currentVersion` (from the backend API, e.g. `"0.76.2"`) with the `latestVersion` (from the GitHub Releases API `tag_name`, e.g. `"v0.76.2"`). The `v` prefix on the GitHub release tag causes a string equality check to fail, falsely indicating a version mismatch.

### Solution

Added a `stripVersionPrefix` helper that removes the leading `v`/`V` from version strings before comparison. The fix is scoped to the comparison logic only — the raw version strings (including the `v` prefix) are preserved for display and for constructing GitHub release URLs.

### Changes

- `frontend/src/utils/app.ts`: Added `stripVersionPrefix()` helper; updated `checkVersionState` to normalize both versions before comparing
- `frontend/src/utils/__tests__/app.test.ts`: Added 9 tests covering `v` prefix handling, build suffix stripping, case insensitivity, and version mismatch scenarios

### Testing

- All 20 unit tests pass (11 existing + 9 new)
- ESLint: ✅ clean
- TypeScript type checking: ✅ clean
- Go backend build: ✅ clean (no backend changes)

### Trade-offs

None. This is a minimal, backward-compatible fix. The normalization only occurs during comparison — the stored version values are unmodified, so GitHub release URLs and display values continue to work correctly.

---

> [!NOTE]
> **Low Risk**
> Small, well-scoped string-normalization change with added tests; minimal impact beyond version-update UI logic.
> 
> **Overview**
> Fixes false “upgrade available” detection by normalizing version strings in `checkVersionState` to ignore a leading `v`/`V` and still strip any `-suffix` from the current version before comparison.
> 
> Adds unit coverage in `app.test.ts` for matching/mismatching scenarios, including prefixed tags, case-insensitive prefixes, and pre-release/build suffix handling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 024a501df00d22f8b9e1a4faa7b088ee97ec9d85. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #11003, opened 2026-04-19
fix: handle v prefix in version comparison on /status page

### LINKED ISSUE #7484: /status page incorrectly shows an upgrade is available

avasis-ai:
## Summary

Fixes the `/status` page incorrectly showing that an upgrade is available when the installed version matches the latest version but differs in `v` prefix formatting.

## Problem

The GitHub releases API returns versions with a `v` prefix (e.g. `v0.76.2`), while the locally installed version typically does not have one (e.g. `0.76.2`). The `checkVersionState` function compared these strings directly without normalizing the prefix, causing `"0.76.2" !== "v0.76.2"` to evaluate as true even for matching versions.

## Changes

Updated `checkVersionState()` in `utils/app.ts` to strip the `v` prefix and pre-release suffix from both versions before comparing.

Closes #7484

## Test Plan

- Install SigNoz via Linux standalone binary
- Navigate to `/status` page
- Verify that when the installed version matches the latest GitHub release, the page shows "You are on the latest version" instead of suggesting an upgrade

---

> [!NOTE]
> **Low Risk**
> Low risk utility change that only affects UI version comparison logic; it may slightly alter when upgrade notifications appear if version strings include prefixes/suffixes.
> 
> **Overview**
> Fixes version checks to treat `v0.x.y` and `0.x.y` as equivalent by stripping an optional leading `v` (and still ignoring pre-release suffixes) before comparing current vs latest versions. This prevents false “upgrade available” indicators on pages that rely on `checkVersionState` (e.g., `/status` and the side nav).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 264fa2d8d61dabc2594539f59d0fa887c3bb7ce0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/frontend/src/utils/app.ts
+++ b/frontend/src/utils/app.ts
@@ -20,8 +20,10 @@ export const checkVersionState = (
 	currentVersion: string,
 	latestVersion: string,
 ): boolean => {
-	const versionCore = currentVersion?.split('-')[0];
-	return versionCore === latestVersion;
+	const stripPrefix = (v: string): string => (v?.startsWith('v') ? v.slice(1) : v) ?? '';
+	const versionCore = stripPrefix(currentVersion?.split('-')[0]);
+	const latest = stripPrefix(latestVersion?.split('-')[0]);
+	return versionCore === latest;
 };
 
 // list of forbidden tags to remove in dompurify


Mrsandeep27 (commented):
Saw this after I opened #11011 which fixes the same bug. Happy to defer to whichever lands first — yours is the older PR. One observation you might want to fold in before merge:

**Missing test coverage.** \`checkVersionState\` doesn't have an existing unit test, and regressions here are silent (users see a phantom upgrade banner, not a crash). My #11011 ships the same fix with \`frontend/src/utils/app.test.ts\` covering exact match, either-side \`v\` prefix, both prefixed, pre-release (\`v0.76.2-rc.1\` vs \`0.76.2\`), and true version mismatch. I'm happy to push the test file over to this branch if the maintainers prefer your patch to go in.

**Implementation nit:** depending on how you wrote the strip, a regex like \`/^v/i\` handles both \`v\` and \`V\`; a \`startsWith('v')\` check only catches lowercase. Minor edge case but worth picking the broader one.

ChinmayMhatre:
PR #10411  also working on the same issue

H4ad:
Hey, sorry for taking so long to review, we ended up approving and merging https://github.com/SigNoz/signoz/pull/11011, so I will close this PR, thanks for contributing!

PR #11005, opened 2026-04-20
feat: adding cloud integration azure types and openapi spec

swapnil-signoz:
### 📄 Summary
- Adding cloud integration Azure types
- Generated OpenAPI specs
- Generated Frontend types from OpenAPI specs

### Closes
https://github.com/SigNoz/engineering-pod/issues/4544

### Note for reviewer
Ignore the switch cases for AWS and Azure types interaction, we will improve this in coming iterations.


pkg/types/cloudintegrationtypes/account.go
```diff
@@ -29,8 +29,13 @@ type AgentReport struct {
...
 
 type AccountConfig struct {
-	// required till new providers are added
-	AWS *AWSAccountConfig `json:"aws" required:"true" nullable:"false"`
+	AWS   *AWSAccountConfig   `json:"aws,omitempty" required:"false" nullable:"false"`
+	Azure *AzureAccountConfig `json:"azure,omitempty" required:"false" nullable:"false"`
+}
+
+type UpdatableAccountConfig struct {
+	AWS   *UpdatableAWSAccountConfig   `json:"aws,omitempty" required:"false" nullable:"false"`
```
nityanandagohain, line 37: one can be null right when one of them is present ?

pkg/types/cloudintegrationtypes/cloudprovider_azure.go
```diff
@@ -0,0 +1,62 @@
...
+
+type AzurePostableAccountConfig = AzureAccountConfig
+
+type AzureConnectionArtifact struct {
+	CLICommand        string `json:"cliCommand" required:"true"`
+	CloudShellCommand string `json:"cloudShellCommand" required:"true"`
+}
+
+type AzureServiceConfig struct {
+	Logs    *AzureServiceLogsConfig    `json:"logs" required:"true"`
```
nityanandagohain, line 20: this can just be bool for now, instead of struct and can evolve later when/if required ?

pkg/types/cloudintegrationtypes/cloudintegration.go
```diff
@@ -78,6 +79,19 @@ type StorableAWSMetricsServiceConfig struct {
 	Enabled bool `json:"enabled"`
 }
 
+type StorableAzureServiceConfig struct {
+	Logs    *StorableAzureLogsServiceConfig    `json:"logs,omitempty"`
```
nityanandagohain, line 83: same here can just be bool for now.

pkg/types/cloudintegrationtypes/account.go
swapnil-signoz: Sending `null` in response does not add anything to the response, as that key will never be used anyways.

pkg/types/cloudintegrationtypes/cloudprovider_azure.go
swapnil-signoz: As this config is being stored in DB as JSON, making it just bool now will make it hard to extend in future as it will require migration for existing rows and so on, current struct can grow easily.

pkg/types/cloudintegrationtypes/account.go
nityanandagohain: right 👍

pkg/types/cloudintegrationtypes/cloudprovider_azure.go
nityanandagohain: we store the entire AzureServiceConfig or individual AzureServiceLogsConfig/ AzureServiceMetricsConfig? if individual then what you said holds true, if not then you will have to take care of any changes regardless right ?

swapnil-signoz: entire `AzureServiceConfig`
Something like 
```json
{"logs": {"enabled":true, "s3buckets": {"ap-south-1": ["bucket1"]}}, "metrics": {"enabled":"true"}}
```

This is an example of AWS to give you more idea, similarly for Azure can come in later in future.

nityanandagohain: cool

nityanandagohain (dismissed):

nityanandagohain (approved):

therealpandey (approved):

YounixM (approved):

PR #11013, opened 2026-04-20
Feat/crosshair series highlight

ahrefabhi:
## Pull Request

---

### 📄 Summary

This PR adds two related improvements to dashboard crosshair sync:

**1. Cross-panel series highlighting (feat/crosshair-series-highlight)**
When multiple dashboard panels are synced via crosshair, hovering over a series in one panel now automatically focuses the matching series in other synced panels — bringing it to the foreground while dimming all others. Matching is based on `metric` labels and requires both panels to share the same `groupBy` configuration.

**2. Tooltip sync with series filtering (#11114)**
When tooltip sync is active and the source panel has a focused series, receiver tooltips now filter their content to show only the series that matches the source panel's highlighted series, hiding entirely when no match exists. Previously, receiver tooltips showed all series regardless of what was highlighted in the source.

Three additional fixes are bundled:
- Clicking inside another pinned tooltip no longer dismisses an unrelated pinned tooltip (outside-click guard was too broad).
- `isTooltipVisible` is now driven solely by whether `tooltipBody` is non-null, removing a stale `isHovering` fallback that caused a brief visible flash.
- `syncDisplayHook.ts` consolidates y-axis crosshair visibility logic and series-focusing logic into a single `setCursor` hook.

https://github.com/user-attachments/assets/2f8cf33c-029a-45d2-9ef4-85c81495e843

#### Issues closed by this PR

---

### ✅ Change Type

- [x] ✨ Feature
- [x] 🐛 Bug fix

---

### 🧪 Testing Strategy

- Tests added/updated: existing `TimeSeriesPanel` utils tests updated; `TooltipPlugin` tests updated to cover `syncedSeriesIndexes` filtering.
- Manual verification:
  - Open a dashboard with two time-series / bar panels sharing the same `groupBy` fields, enable crosshair sync, hover — the matching series highlights in both panels and the receiver tooltip shows only the matching series.
  - Panels with different or empty `groupBy` — no highlighting applied, crosshair still syncs normally.
  - Source panel cursor leaves the plot — receiver panel immediately unfocuses all series.
  - `syncedSeriesIndexes = []`: receiver tooltip fully suppressed when no series match.
  - `syncedSeriesIndexes = null`: receiver shows all series (unfiltered, non-sync path).
  - Pinned tooltips are not dismissed by clicks on sibling pinned tooltips.

---

- Blast radius: `TimeSeriesPanel`, `BarPanel`, `ChartWrapper`, `UPlotSeriesBuilder`, `TooltipPlugin`, `syncCursorRegistry` — limited to dashboard panel rendering and crosshair sync path.
- Non-synced tooltips and panels without crosshair sync are entirely unaffected.
- Panels with crosshair sync but mismatched `groupBy` fall through to the existing no-highlight path.
- Rollback plan: revert the commits; no schema or API changes.

---

### 📝 Changelog

| Field | Value |
|---|---|
| Description | Hovering over a series in a synced dashboard panel now highlights the matching series in all other synced panels. Synced tooltips also filter their content to show only the highlighted series, hiding entirely when no match is found. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- `syncDisplayHook.ts` consolidates the y-crosshair visibility logic that previously lived inline in `TooltipPlugin.tsx` — this is a refactor of existing behaviour, not a change to it.
- `applyReceiverSync` now returns `number[] | null` instead of `void` — `null` means no filtering, `[]` means no match (hide tooltip), `[…]` means allowed 1-based series indexes. The controller stores this as `syncedSeriesIndexes` and resets it to `null` on source-driven cursor events to avoid stale filtering.
- The `groupByMatches` guard is intentionally strict (both arrays must be non-empty and structurally equal) to avoid false positives across unrelated panels.
- The `[data-pinned="true"]` attribute used by the outside-click guard needs to be present on the tooltip container when pinned — verify `TooltipPlugin` sets this attribute correctly.
- `metric` is attached to the series config at build time so it travels with the `uPlot.Series` object and is available inside `setCursor` hooks without extra state.


frontend/src/lib/uPlotV2/plugins/TooltipPlugin/syncDisplayHook.ts
```diff
@@ -0,0 +1,130 @@
...
+	syncMetadata: TooltipSyncMetadata | undefined,
+	controller: TooltipControllerState,
+): (u: uPlot) => void {
+	return (u: uPlot): void => {
+		const yCrosshairEl = u.root.querySelector<HTMLElement>('.u-cursor-y');
+		if (!yCrosshairEl) {
+			return;
+		}
+
+		if (u.cursor.event != null) {
```
AshwinBhatkal, line 152: can you add small 1 line comments for this expression and the else

frontend/src/lib/uPlotV2/plugins/TooltipPlugin/syncDisplayHook.ts
```diff
@@ -0,0 +1,130 @@
...
+
+import type { ExtendedSeries } from '../../config/types';
+import { syncCursorRegistry } from './syncCursorRegistry';
+import type { TooltipControllerState, TooltipSyncMetadata } from './types';
+
+/**
+ * Returns true only when both panels declare a non-empty groupBy that is
+ * structurally identical — the precondition for cross-panel series highlighting.
+ */
+function groupByMatches(
```
AshwinBhatkal: This is different from what we discussed noe?

ahrefabhi: In crosshair we just want to do exact groupBy matches, in the tooltip PR I'll handle other cases as well

AshwinBhatkal (dismissed):

AshwinBhatkal (approved):

PR #11018, opened 2026-04-20
fix: disallow group by timestamp for timeseries request

tushar-signoz:
## Pull Request

---

### 📄 Summary

Disallow `GROUP BY timestamp` in query-builder queries when the request type is `timeseries`. Timeseries responses already bucket results by timestamp via the step interval, so grouping by `timestamp` produces duplicated/incorrect series and confusing downstream rendering.

The validation is wired through a new `WithRequestType` option so `validateGroupBy` knows the request type and can reject `timestamp` as a group-by key specifically for `RequestTypeTimeSeries`. Other request types (raw, scalar, trace) are unaffected.

#### Issues closed by this PR

N/A

---

### ✅ Change Type

- [x] 🐛 Bug fix

---

#### Root Cause
Group-by validation had no awareness of the request type, so a timeseries query could include `timestamp` in `group_by`. Since timeseries queries are already bucketed by timestamp, this led to nonsensical groupings in the response.

#### Fix Strategy
- Add `requestType` to `validationConfig` and expose it via a new `WithRequestType` option.
- In `validateGroupBy`, reject `timestamp` as a group-by key when `requestType == RequestTypeTimeSeries`.
- Plumb `WithRequestType(requestType)` through `GetValidationOptions` for all request types so the config is always populated.

---

### 🧪 Testing Strategy

- Manual verification: submitting a timeseries builder query with `timestamp` in `group_by` now returns `CodeInvalidInput` with a clear message; other request types still accept the same payload.
- Edge cases covered: raw/scalar/trace request types remain unaffected; empty/other group-by keys go through the existing validation path unchanged.

---

- Blast radius: query-builder validation only; no query execution or storage changes.
- Potential regressions: any client that was (incorrectly) sending `timestamp` in `group_by` for a timeseries request will now receive a 400. This is the intended behavior.
- Rollback plan: revert this commit — change is isolated to `pkg/types/querybuildertypes/querybuildertypesv5/validation.go`.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Timeseries queries can no longer group by `timestamp`; the API rejects such requests with a clear validation error. |

---

### 📋 Checklist
- [x] Tests added or explicitly not required (validation-only, covered manually)
- [x] Manually tested
- [x] Breaking changes documented (rejects previously-accepted but incorrect payloads)

---

## 👀 Notes for Reviewers

The new `WithRequestType` option is also passed for raw/scalar/trace/default so `validationConfig.requestType` is consistently populated — only the timeseries branch currently uses it, but this leaves the door open for other request-type-specific rules without further plumbing.

### DIFF
--- a/pkg/types/querybuildertypes/querybuildertypesv5/validation.go
+++ b/pkg/types/querybuildertypes/querybuildertypesv5/validation.go
@@ -47,13 +47,13 @@ const (
 type ValidationOption func(*validationConfig)
 
 type validationConfig struct {
-	skipLimitOffsetValidation bool
-	skipAggregationValidation bool
-	skipHavingValidation      bool
-	skipAggregationOrderBy    bool
-	skipSelectFieldValidation bool
-	skipGroupByValidation     bool
-	requestType               RequestType
+	skipLimitOffsetValidation      bool
+	skipAggregationValidation      bool
+	skipHavingValidation           bool
+	skipAggregationOrderBy         bool
+	skipSelectFieldValidation      bool
+	skipGroupByValidation          bool
+	WithTimestampGroupByValidation bool
 }
 
 func applyValidationOptions(opts []ValidationOption) validationConfig {
@@ -113,9 +113,9 @@ func WithSkipGroupByValidation() ValidationOption {
 }
 
 // WithRequestType sets the request type for validation.
-func WithRequestType(requestType RequestType) ValidationOption {
+func WithTimestampGroupByValidation() ValidationOption {
 	return func(cfg *validationConfig) {
-		cfg.requestType = requestType
+		cfg.WithTimestampGroupByValidation = true
 	}
 }
 
@@ -185,9 +185,9 @@ func (q *QueryBuilderQuery[T]) validateGroupBy(cfg validationConfig) error {
 				errors.CodeInvalidInput, "invalid empty key name for group by at index %d", idx,
 			)
 		}
-		if cfg.requestType == RequestTypeTimeSeries && item.Name == "timestamp" {
+		if cfg.WithTimestampGroupByValidation && item.Name == "timestamp" {
 			return errors.NewInvalidInputf(
-				errors.CodeInvalidInput, "group by on timestamp is not allowed for request type %s", cfg.requestType.StringValue(),
+				errors.CodeInvalidInput, "group by on timestamp is not allowed",
 			).WithAdditional("Timeseries request already accounts for timestamp in the response")
 		}
 	}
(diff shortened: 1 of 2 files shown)


nityanandagohain (dismissed):

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
```diff
@@ -111,6 +112,13 @@ func WithSkipGroupByValidation() ValidationOption {
 	}
 }
 
+// WithRequestType sets the request type for validation.
+func WithRequestType(requestType RequestType) ValidationOption {
+	return func(cfg *validationConfig) {
+		cfg.requestType = requestType
+	}
+}
```
srikanthccv: This is odd. We decided to not use the requestType as the param last time and instead add the specific validations directly targeted. It would be clean to use something like `WithTimestampGroupByValidation` to remain consistent with the rest of the aggregation options and it's added when the request type is time series inside `GetValidationOptions`

tushar-signoz:
Hey @nityanandagohain, @therealpandey 
I have addressed what srikanth commented on

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
```diff
@@ -177,6 +185,11 @@ func (q *QueryBuilderQuery[T]) validateGroupBy(cfg validationConfig) error {
 				errors.CodeInvalidInput, "invalid empty key name for group by at index %d", idx,
 			)
 		}
+		if cfg.withTimestampGroupByValidation && item.Name == "timestamp" {
+			return errors.NewInvalidInputf(
+				errors.CodeInvalidInput, "group by on timestamp is not allowed",
+			).WithAdditional("Timeseries request already accounts for timestamp in the response")
```
ankitnayan: NIT: Not a right place to specify `Timeseries` in the error? What if we decide to add this to scalar too? Also, why is groupby timestamp allowed for scalar?

tushar-signoz: Technically it is not wrong to group by on timestamp in scalar query or time-series, because timestamp could be an attribute that user might have specified as something like start_time of a job. 
We have to reject group by on timestamp (even if its an attribute) for time-series because of name collision in the sql generated. We could fix the sql generated using aliases (which complicates things) but then we have to consider that this could be user mistake. 

I agree its not the right place, it can be removed.

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
```diff
@@ -111,6 +112,13 @@ func WithSkipGroupByValidation() ValidationOption {
 	}
 }
 
+// WithRequestType sets the request type for validation.
```
srikanthccv: The doc string is outdated, please update it.

srikanthccv (approved):

PR #11019, opened 2026-04-20
tests: unify integration + e2e under shared pytest project

therealpandey:
closes SigNoz/platform-pod#2038

## Summary

Unifies integration and e2e under a shared pytest project at `tests/`. One fixture graph (containers, auth, license, telemetry inserts, API seeders) serves both trees; e2e gets a hermetic local backend instead of pointing at staging.

- Pytest rootdir lifted from `tests/integration/` to `tests/`; `fixtures/` is shared infrastructure at `tests/fixtures/` — both trees import from the same flat package.
- Integration suites live under `tests/integration/tests/<suite>/` (was `tests/integration/src/<suite>/`).
- Playwright e2e lives under `tests/e2e/tests/<feature>/`. `tests/e2e/fixtures/auth.ts` exports a per-worker `authedPage` fixture (replaces the project-level `global.setup.ts` + `.auth/` dir scheme).
- The bootstrap writes `.env.local` (dotenv-loaded by `playwright.config.ts`) — no JSON sidecar, no custom globalSetup.
- `tests/seeder/` is a standalone FastAPI service that wraps the ClickHouse insert helpers so specs can POST telemetry at runtime.
- `apply_license` fixture stubs the Zeus `/v2/licenses/me` endpoint, POSTs `/api/v3/licenses`, and dismisses the ENTERPRISE onboarding prompt so specs can navigate directly to the feature under test.

## Why

The old `signoz-e2e` repo pointed every run at remote staging. Local dev was slow, PR-blocking CI was impossible, and specs silently depended on whatever staging happened to have. This PR puts e2e on the hermetic local backend that already exists for integration tests.

## Layout after merge

```
tests/
  conftest.py, pyproject.toml, uv.lock      # shared project, rootdir
  fixtures/                                 # flat shared package
    auth.py (+ apply_license fixture)
    alerts.py, cloudintegrations.py, ...    # merged xxx.py + xxxutils.py
    browser.py (selenium), keycloak.py (IdP container)
    traces.py, logs.py, metrics.py          # direct ClickHouse insert helpers
    seeder.py                               # thin fixture around the service
  seeder/                                   # standalone HTTP service
    Dockerfile, server.py
  integration/
    bootstrap/setup.py                      # stack lifecycle
    tests/<suite>/NN_*.py                   # 15 feature suites
    testdata/<suite>/
  e2e/
    bootstrap/setup.py                       # pytest lifecycle (test_setup / test_teardown)
    fixtures/auth.ts                        # authedPage Playwright fixture
    tests/<feature>/*.spec.ts               # Playwright specs
    playwright.config.ts                    # loads .env + .env.local via dotenv
    artifacts/{html,json,results}/          # reporter + per-test output (gitignored)
```

## Usage modes

| Mode | Command |
|------|---------|
| Warm stack + iterate | `uv run pytest ... --reuse --with-web e2e/bootstrap/setup.py::test_setup`, then `cd e2e && yarn test:ui` |
| Integration | `make py-test-setup && make py-test && make py-test-teardown` |
| Staging fallback for e2e | `cd tests/e2e && yarn test:staging` |

## CI

- **`integrationci.yaml`** — matrix path corrected from `integration/src/<suite>` to `integration/tests/<suite>`; matrix updated to the current 15 suites (added `auditquerier` and `rawexportdata`; dropped stale `bootstrap`).
- **`e2eci.yaml`** (new, replaces the broken frontend/-based `run-e2e.yaml`) — label-gated (`safe-to-test` + `safe-to-e2e`). Brings up the stack via the pytest bootstrap, runs Playwright against it, uploads the full `tests/e2e/artifacts/` dir with 5-day retention.

## Test plan

- [x] `uv run pytest --collect-only integration/tests/` — 500 tests collect cleanly from the new rootdir
- [x] Fresh stack bring-up: containers up, admin registered, ENTERPRISE license applied, onboarding dismissed, seeder running, `.env.local` written
- [x] `npx playwright test --project=chromium` — TC-01 alerts smoke test passes
- [x] `make py-fmt && make py-lint` clean
- [ ] Full integration suite on CI (add `safe-to-test` + `safe-to-integrate`)
- [ ] e2e suite on CI (add `safe-to-test` + `safe-to-e2e`)

## Notes for reviewers

- Contributor docs at `docs/contributing/tests/{integration,e2e}.md`; the previous `docs/contributing/go/integration.md` was relocated here via git rename and updated to match the current layout.
- `tests/integration/src/querier/util.py` was merged into `tests/fixtures/querier.py` — three querier tests updated to consolidated imports.
- `tests/fixtures/signoz.py` now anchors the Docker build context to an absolute path computed from `__file__` — works regardless of pytest cwd.
- `ANOMALY_DETECTION` flips on under the applied license; the legacy 2095 anomaly flow is parked for a follow-up — `tests/e2e/tests/alerts/alerts.spec.ts` is currently a single TC-01 smoke test. The earlier 2095 regression replay lives in git history for reference.
- Testdata stays per-tree (integration → `tests/integration/testdata/`, e2e seeds directly via REST / the seeder). Only helper *code* is shared under `tests/fixtures/`.


vikrantgupta25 (approved):

tests/e2e/bootstrap/setup.py
```diff
@@ -0,0 +1,44 @@
+import os
```
AshwinBhatkal, line 1: This file is run only once on start. Correct?

tests/e2e/fixtures/auth.ts
```diff
@@ -0,0 +1,85 @@
...
+  expect,
+  type Browser,
+  type BrowserContext,
+  type Page,
+} from '@playwright/test';
+
+export type User = { email: string; password: string };
+
+// Default user — admin from the pytest bootstrap (.env.local) or staging .env.
+export const ADMIN: User = {
```
AshwinBhatkal, line 12: Where would the other roles be set up

tests/seeder/server.py
```diff
@@ -0,0 +1,154 @@
...
+
+# Metrics payload carries label dicts at the top level, not a `resources`
+# key — tagging goes on the `resource_attrs` wrapper that Metrics.from_dict
+# unpacks. Same effect, different key.
+def _tag_metrics(item: Dict[str, Any]) -> Dict[str, Any]:
+    resource_attrs = {**(item.get("resource_attrs") or {}), **SEEDER_MARKER}
+    return {**item, "resource_attrs": resource_attrs}
+
+
+@app.post("/telemetry/traces", status_code=status.HTTP_201_CREATED)
```
AshwinBhatkal, line 76: what happens for get?

tests/e2e/bootstrap/setup.py
therealpandey: yes that's correct.

tests/e2e/fixtures/auth.ts
therealpandey: So the frontend will have to create that by calling signoz apis. On startup only the root role will be given.

tests/seeder/server.py
therealpandey: Get is `query_range`. The problem that is most common is how to insert telemetry data at the right points so that I can assert them in the UI.

tests/e2e/.prettierrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "useTabs": false,
+  "tabWidth": 2,
+  "singleQuote": true,
+  "trailingComma": "all"
```
AshwinBhatkal, line 5: add all defaults here. It will be better

AshwinBhatkal (approved):

PR #10176, opened 2026-02-04
chore(oxlint): migrate from eslint

H4ad:
## Pull Request

This PR migrates the Eslint configuration to Oxlint, and start the work to migrate Prettier.

I tried to keep this PR as minimal as possible by changing most of the errors to warn, it will requires us some effort to start fixing those lint issues in the future, and there is very good ones to fix.

The current CI time is [~12s](https://github.com/SigNoz/signoz/actions/runs/24087415363/job/70264096503?pr=10176#step:7:11510). To compare, Eslint is taking [~195s](https://github.com/SigNoz/signoz/actions/runs/22921278397/job/66520006786#step:7:2453), Oxlint is ~15x faster.

## Summary

For now, the Oxfmt will be disabled and I want to enable it in another PR together with the format fixes (it will change a lot of files).

I tried to migrate most of the available rules/plugins, the main differences with Eslint it that we need to declare most of the rules manually (unless we start use the Oxlint categories).

The following plugins were removed/not supported:
- prettier: use oxfmt.
- import: use oxfmt.

About Oxlint:
- Ported most of the rules.
- Enabled many good/default rules
  - We can gradually adopt those new rules by changing to error and then fixing the issues.

About type-aware:
- Only two rules are missing: https://github.com/oxc-project/tsgolint/tree/main?tab=readme-ov-file#implemented-rules

About type-check:
- I disabled for now because it slowdown the lint ci by a few seconds
  - I have plans to replace our `typecheck-staged` to use `tsgo` instead, it will be way faster.

| Category | Count |
|----------|-------|
| Rules from ESLint Config (explicitly configured) | ~180 |
| Rules from `correctness` category (auto-enabled) | ~216 |
| Rules from SonarJS plugin | ~30 |
| Rules Not Supported by Oxlint | 16 |
| **Total Active Rules** | **338** |

Currently we are running 338 rules from 332 eslint.

Fixes https://github.com/SigNoz/engineering-pod/issues/5075

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 📋 Checklist
- [ ] Tests added or explicitly not required
- [x] Manually tested
- [ ] Breaking changes documented
- [ ] Backward compatibility considered

PR #11026, opened 2026-04-21
feat: base path config setup and index.html setup as go template for BE injection

SagarRajput-7:
## Pull Request - Base Path Feature

Runtime base-path support: deploy SigNoz under any URL prefix (e.g. `/signoz/`) without a rebuild.
One Docker image, any prefix via `global.external_url` config.

5-phase rollout. Each phase is safe to merge independently.
Zero behavior change at the default `basePath='/'`.

---

#### Go Template Wiring

The mechanism that makes runtime base-path injection possible:

| File             | Change                                                                                                   |
| ---------------- | -------------------------------------------------------------------------------------------------------- |
| `index.html`     | Added `<base href="[[.BaseHref]]" />` - Go `text/template` injects the prefix at startup                 |
| `index.html`     | `favicon.ico` and `css/uPlot.min.css` made relative - `<base href>` resolves them under any prefix       |
| `vite.config.ts` | `devBasePathPlugin()` replaces `[[.BaseHref]]` → `"/"` during `vite serve` (no Go backend needed in dev) |
| `vite.config.ts` | `base: './'` - Vite outputs relative asset paths in the production bundle                                |

#### Frontend Foundation

- **`basePath.ts`** - new utility: `getBasePath()`, `withBasePath()`, `getAbsoluteUrl()`, `getBaseUrl()`
  - Memoized DOM read at module init (not per-call), idempotent, external-URL-safe
- **Unit tests** - both `basePath="/"` and `basePath="/signoz/"` modes, no-tag fallback, normalisation, nested prefix
- **React Router** `basename` wired to `getBasePath()` via `lib/history.ts`
- **Axios** `interceptorsRequestBasePath` on all 8 instances (v1–v5, alert manager, log event, generated API)
- **`openInNewTab()`** and **`useSafeNavigate`** new-tab branch both call `withBasePath` internally - callers unchanged
- **i18n** locale fetch fixed - `/locales/...` → `getBasePath() + 'locales/...'`
- **`ErrorBoundaryFallback`** "Go Home" uses `history.push(ROUTES.HOME)` instead of `window.location.href`
- **ESLint rule** `no-raw-absolute-path` at `warn` - flags `window.open(path, '_blank')` and `origin + path` patterns
- **`.gitignore`** - stale `index.html.gotmpl` entry removed

> ✅ At `/` - all changes are no-ops, zero behavior change from today  
> ✅ At `/signoz/` - assets, router, APIs, translations, new-tab navigation, and error boundary all work

Notion for more details - https://www.notion.so/signoz/Serving-SigNoz-329fcc6bcd1980a487efc5b056f6e95e?source=copy_link#32cfcc6bcd1980d2a900f119a2c01c46
Action Plan - https://github.com/SigNoz/platform-pod/issues/408#issuecomment-4266595996

---

- closes - https://github.com/SigNoz/platform-pod/issues/1775
- https://github.com/SigNoz/platform-pod/issues/408
---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

NA

---

Manually tested, passing build, tested on a deployment with prefix and one without prefix

- Tests added/updated: added for the utils
- Manual verification: manually tested
- Edge cases covered: Yes

---

- Blast radius: Whole app's routing and serving 
- Potential regressions: tested on different levels
- Rollback plan: NA

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---


SagarRajput-7:
@H4ad - reference your comments here since I had to make all the PR again, due to to many conflict and mis-management on the older one, all the commits are same and cherry-pick - 

comment 1 - https://github.com/SigNoz/signoz/pull/10966#discussion_r3107291014
comment 2 - https://github.com/SigNoz/signoz/pull/10966#pullrequestreview-4136399113

H4ad:
Let's wait https://github.com/SigNoz/signoz/pull/10176 before merging this one.

H4ad (approved):

frontend/vite.config.ts
```diff
@@ -28,10 +41,13 @@ function rawMarkdownPlugin(): Plugin {
 export default defineConfig(
 	({ mode }): UserConfig => {
 		const env = loadEnv(mode, process.cwd(), '');
+		// Base path for serving the app (e.g., '/signoz/'). Defaults to '/'.
+		const basePath = env.VITE_BASE_PATH || '/';
```
AshwinBhatkal, line 45: we should have this in example.env

frontend/src/utils/basePath.ts
```diff
@@ -0,0 +1,50 @@
+// Read once at module init — avoids a DOM query on every axios request.
+const _basePath: string = ((): string => {
```
AshwinBhatkal, line 2: nit: why _ prefix

frontend/src/utils/basePath.ts
```diff
@@ -0,0 +1,50 @@
...
+	const href = document.querySelector('base')?.getAttribute('href') ?? '/';
+	return href.endsWith('/') ? href : `${href}/`;
+})();
+
+/** Returns the runtime base path — always trailing-slashed. e.g. "/" or "/signoz/" */
```
AshwinBhatkal, line 7: is this always trailing only or both leading and trailing?

frontend/src/utils/basePath.ts
```diff
@@ -0,0 +1,50 @@
...
+ *   withBasePath('https://x.com') → 'https://x.com' (external, passthrough)
+ */
+export function withBasePath(path: string): string {
+	if (!path.startsWith('/')) {
+		return path;
+	}
+	if (_basePath === '/') {
+		return path;
+	}
+	if (path.startsWith(_basePath) || path === _basePath.slice(0, -1)) {
```
AshwinBhatkal, line 27: what if the path itself had a trailing slash? We should mention that in the desc

frontend/src/utils/__tests__/base-path.test.ts
```diff
@@ -0,0 +1,118 @@
...
+			`${window.location.origin}/signoz/logs`,
+		);
+	});
+
+	it('getBaseUrl returns origin + prefix without trailing slash', () => {
+		expect(m.getBaseUrl()).toBe(`${window.location.origin}/signoz`);
+	});
+});
+
+describe('no <base> tag', () => {
```
AshwinBhatkal, line 97: This is controlled by us right? Would there ever be a case where we don't have base tag?

frontend/src/utils/__tests__/base-path.test.ts
```diff
@@ -0,0 +1,118 @@
...
+	it('getBasePath falls back to "/"', () => {
+		const m = loadModule();
+		expect(m.getBasePath()).toBe('/');
+	});
+});
+
+describe('href without trailing slash', () => {
```
AshwinBhatkal, line 104: similar question to above, can it be without a trailing slash? I can think of 1 case where we don't have value, then what is the default we use? "/" ?

frontend/.stylelintrc.cjs
```diff
@@ -1,3 +1,4 @@
+// oxlint-disable-next-line typescript/no-require-imports
```
AshwinBhatkal, line 1: can we instead add an exception for cjs files? instead of this comment

frontend/plugins/rules/no-raw-absolute-path.mjs
```diff
@@ -0,0 +1,152 @@
+/**
+ * Rule: no-raw-absolute-path
+ *
+ * Catches patterns that break at runtime when the app is served from a
```
AshwinBhatkal, line 4: Too good. Love this ❤️

frontend/plugins/rules/no-raw-absolute-path.mjs
```diff
@@ -0,0 +1,152 @@
...
+		type: 'suggestion',
+		docs: {
+			description:
+				'Disallow raw window.open and origin-concatenation patterns that miss the runtime base path',
+			category: 'Base Path Safety',
+		},
+		schema: [],
+		messages: {
+			windowOpen:
+				'Use openInNewTab(path) instead of window.open(path, "_blank") — openInNewTab prepends the base path automatically.',
```
AshwinBhatkal, line 80: - We should avoid all uses of window.open and replace it with a function which has relevant options. We should not ideally use disable comments at all. So when the error comes, we can ask them to use the right function. In the case of aws connection url

- There are some places that are not erroring out for open in new tab. About 20 at least. I'm on latest main
Image

PR #11044, opened 2026-04-21
feat(authz): add support for delete role

vikrantgupta25:
### 📄 Summary
- add support for delete role

contributes to: https://github.com/SigNoz/platform-pod/issues/2037

### DIFF
--- a/ee/authz/openfgaauthz/provider.go
+++ b/ee/authz/openfgaauthz/provider.go
@@ -11,8 +11,6 @@ import (
 	"github.com/SigNoz/signoz/pkg/errors"
 	"github.com/SigNoz/signoz/pkg/factory"
 	"github.com/SigNoz/signoz/pkg/licensing"
-	"github.com/SigNoz/signoz/pkg/modules/serviceaccount"
-	"github.com/SigNoz/signoz/pkg/modules/user"
 	"github.com/SigNoz/signoz/pkg/sqlstore"
 	"github.com/SigNoz/signoz/pkg/types/authtypes"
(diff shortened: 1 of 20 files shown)


ee/authz/openfgaauthz/provider.go
```diff
@@ -260,18 +277,43 @@ func (provider *provider) Delete(ctx context.Context, orgID valuer.UUID, id valu
...
+	if err := provider.deleteTuples(ctx, role.Name, orgID); err != nil {
+		return errors.WithAdditionalf(err, "failed to delete transactions for the role: %s", role.Name)
+	}
+
+	err = provider.store.Delete(ctx, orgID, id)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
therealpandey: This should be happening at the user and service account layer respectively not in the authz layer.

therealpandey (changes requested):

ee/authz/openfgaauthz/provider.go
therealpandey: Both the modules already take an injection of authz, they should be able to delete roles similar to how they are able to create roles.

vikrantgupta25: they do not create roles, they grant / revoke the role assignment. they cannot delete the role ? Did you mean something else ? 😅

vikrantgupta25: added the callbacks, couldn't make it variadic since the registry is already one

therealpandey (approved):

PR #11050, opened 2026-04-22
chore: add json enabled as feature flag for FE

piyushsingariya:
## Pull Request

---

### 📄 Summary

JSON FE experience is changing few things in FE and older and newer both interactions are supposed to be active based on a feature flag. Hence this PR aims to remove older ENV var and replace it with flagger

image

Closes SigNoz/engineering-pod#4744

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Chore

---

- Tests added/updated: N/A
- Manual verification: Yes 
- Edge cases covered: N/A

---

- Blast radius: N/A
- Potential regressions: N/A
- Rollback plan: revert PR

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---


srikanthccv:
@piyushsingariya either complete to use the feature flagging system throughout or don't use it at all. This change mixes both the feature flagging system and the specific env we have for json body with bootstrap env. Since you are anyway touching the feature flagging, let's use it throughout. Let me know if there is any confusion here.

ee/query-service/app/api/featureFlags.go
```diff
@@ -71,6 +72,15 @@ func (ah *APIHandler) getFeatureFlags(w http.ResponseWriter, r *http.Request) {
 		Route:      "",
 	})
 
+	bodyJSONQuery := querybuilder.IsBodyJSONEnabled(ctx, ah.Signoz.Flagger)
```
tushar-signoz: querybuilder.IsBodyJSONEnabled. Is not a correct exported method. 
querybuilder is not source of truth, ah.Signoz.Flagger is. Essentially you're sending `ah.Signoz.Flagger` to querybuilder as an arg and using querybuilder.IsBodyJSONEnabled as a utlility function that has nothing to do with querybuilder. I hope I was able to convery my thought process.

tushar-signoz (dismissed):

tushar-signoz (commented):
Hey @piyushsingariya 
Do we have to pass flagger object to statement builders ? Why not import flagger pkg which can have a exported JSONBodyEnabled flag ?

ee/query-service/app/api/featureFlags.go
piyushsingariya: Yes, I am able to understand the thought, I'll change it.

piyushsingariya:
> Do we have to pass flagger object to statement builders ? Why not import flagger pkg which can have a exported JSONBodyEnabled flag ?

Flagger needs to evaluate a featureflag. Other way I could've done it is to have a boolean parameter in methods, but that felt wrong, and error prone.

Flagger is modularized and it should be seeded in flows as other use it. I think.

therealpandey (approved):

pkg/flagger/flaggertest/flaggertest.go
```diff
@@ -0,0 +1,52 @@
...
+	registry := flagger.MustNewRegistry()
+	fl, err := flagger.New(
+		context.Background(),
+		instrumentationtest.New().ToProviderSettings(),
+		flagger.Config{},
+		registry,
+		configflagger.NewFactory(registry),
+	)
+	if err != nil {
+		t.Fatalf("flaggertest.New: %v", err)
```
therealpandey: Can we use require please?

pkg/flagger/flaggertest/flaggertest.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	fl, err := flagger.New(
+		context.Background(),
+		instrumentationtest.New().ToProviderSettings(),
+		cfg,
+		registry,
+		configflagger.NewFactory(registry),
+	)
+	if err != nil {
+		t.Fatalf("flaggertest.WithBodyJSON: %v", err)
```
therealpandey: Same please use require

pkg/flagger/registry.go
```diff
@@ -8,6 +8,7 @@ var (
 	FeatureHideRootUser      = featuretypes.MustNewName("hide_root_user")
 	FeatureGetMetersFromZeus = featuretypes.MustNewName("get_meters_from_zeus")
 	FeaturePutMetersInZeus   = featuretypes.MustNewName("put_meters_in_zeus")
+	FeatureBodyJSONQuery     = featuretypes.MustNewName("body_json_enabled")
```
therealpandey: use_json_body

In line with other feature flag names.

therealpandey (changes requested):

YounixM (approved):

therealpandey (approved):

PR #11065, opened 2026-04-22
feat(authz): add resource-level FGA for service accounts

vikrantgupta25:
## Summary

- add resource-level FGA for service accounts

closes: https://github.com/SigNoz/platform-pod/issues/2036


pkg/modules/serviceaccount/implserviceaccount/module.go
```diff
@@ -430,3 +430,138 @@ func apiKeyCacheKey(apiKey string) string {
 func identityCacheKey(serviceAccountID valuer.UUID) string {
 	return "identity::" + serviceAccountID.String()
 }
+
+// typeableRegistry is a standalone RegisterTypeable implementation that can be
+// used independently of the module lifecycle. This is needed because authz
+// initialization requires RegisterTypeable entries before the service account
+// module is created (the module depends on authz).
+type typeableRegistry struct{}
```
vikrantgupta25: make this a separate file inside implservice account

pkg/modules/serviceaccount/implserviceaccount/module.go
```diff
@@ -430,3 +430,138 @@ func apiKeyCacheKey(apiKey string) string {
...
+						Name: serviceaccounttypes.TypeableMetaResourceServiceAccount.Name(),
+					},
+					authtypes.MustNewSelector(authtypes.TypeMetaResource, "*"),
+				),
+			},
+		},
+	}
+}
+
+func (module *module) MustGetTypeables() []authtypes.Typeable {
```
vikrantgupta25: why make these module functions ? Just make sure to have registry separate and use this for service accounts

pkg/modules/serviceaccount/implserviceaccount/module.go
vikrantgupta25: Done — moved `typeableRegistry` and all its methods to `implserviceaccount/typeable_registry.go`.

pkg/modules/serviceaccount/implserviceaccount/module.go
vikrantgupta25: Done — removed `MustGetTypeables()` and `MustGetManagedRoleTransactions()` from the module struct, and removed `authz.RegisterTypeable` embedding from the `Module` interface. The registry is now purely standalone, passed directly to the authz callback.

Also added a privilege escalation guard in `setRole`: before granting a role to a service account, we now verify the caller holds that role themselves via `CheckWithTupleCreation`.

YounixM (approved):

therealpandey (approved):

PR #11081, opened 2026-04-23
feat(global-time-store): add support to context, url persistence, store persistence, drift handle

H4ad:
## Pull Request

---

This PR is part of a large refactor related to Date Time Selection V2, the goal is to rewrite the time selection component to use our Zustand Store, and also get rid of Redux eventually.

In this part of the migration, I added support for the context+zustand via provider, this will allow us to easily use time selection component inside modals with time being separated from list component, and without have the need to drill-down the props into multiple components.

On `index.ts`, I added some docs related to how to use this store to be easy to instruct on how to migrate from old redux to this new store.

This PR also aims to solve one of the issues introduced by this global time store that is: getMinMax was not deterministic, causing each chart (if used by multiple in same UI) to have different time in the query. 

Now, it calculates the min/max time to be used across the queries while round down to 5s when refresh is enabled. This will ensure the charts always use the same value (like redux) but also handle the query identity being stable by using `selectedTime` as `1h`.

This PR also adds little features such as:
- support to store on query params
- support to store selected time on local storage
- support to isolate query invalidation
  - This means I can invalidate the queries of a modal without causing the queries of list to be invalidate as well by passing name in the query.

https://github.com/user-attachments/assets/ca1d0f90-84e0-444d-b442-2f5790fa4ef3

Closes https://github.com/SigNoz/engineering-pod/issues/4764

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Any page using this component, currently on Infra Monitoring is using this component.
- Potential regressions: Any mismanagement in the calculation of the time.
- Rollback plan: Revert this commit or fix the bug directly

---

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | Improved our Zustand store and create new helpers to facilitate the usage of time filters across the pages. |

---

### 📋 Checklist
- [x] Manually tested


frontend/src/store/globalTime/__tests__/globalTimeStore.test.tsx
```diff
@@ -0,0 +1,789 @@
...
+			act(() => {
+				result.current.setSelectedTime('15m', 0);
+			});
+
+			expect(result.current.isRefreshEnabled).toBe(false);
+		});
+
+		it('should disable refresh for custom time range even with refreshInterval > 0', () => {
+			const { result } = renderHook(() => useGlobalTimeStore());
+			const customTime = createCustomTimeRange(1000000000, 2000000000);
```
AshwinBhatkal, line 123: I believe we had the nano second multiplier for this?

frontend/src/store/globalTime/__tests__/globalTimeStore.test.tsx
```diff
@@ -0,0 +1,789 @@
...
+
+			// Compute and cache values for relative time
+			act(() => {
+				result.current.computeAndStoreMinMax();
+			});
+
+			const relativeMinMax = { ...result.current.lastComputedMinMax };
+
+			// Switch to custom time range
+			const customMinTime = 5000000000;
```
AshwinBhatkal, line 200: same here

AshwinBhatkal: and at all places

frontend/src/store/globalTime/__tests__/globalTimeStore.test.tsx
H4ad: No, this is just a random time.

frontend/src/store/globalTime/__tests__/globalTimeStore.test.tsx
H4ad: This is also a custom random time.

frontend/src/store/globalTime/useComputedMinMaxSync.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function useComputedMinMaxSync(store: GlobalTimeStoreApi): void {
+	useEffect(() => {
+		store.getState().computeAndStoreMinMax();
+	}, [store]);
+
+	useEffect(() => {
+		let previousSelectedTime = store.getState().selectedTime;
+
+		return store.subscribe((state) => {
```
AshwinBhatkal: can you please add comments for what this does and the hook in general? Same for other hooks as well

frontend/src/store/globalTime/types.ts
```diff
@@ -50,3 +50,52 @@ export interface IGlobalTimeStoreActions {
...
+	/** URL sync configuration. When false/omitted, no URL sync. */
+	enableUrlParams?:
+		| boolean
+		| {
+				relativeTimeKey?: string;
+				startTimeKey?: string;
+				endTimeKey?: string;
+		  };
+	removeQueryParamsOnUnmount?: boolean;
+	localStoragePersistKey?: string;
```
AshwinBhatkal, line 75: can this be made required? If not, let's add an if check in the context where the persistence hook is used? Can you verify accordingly for other types as well once please

frontend/src/store/globalTime/useQueryCacheSync.ts
```diff
@@ -0,0 +1,42 @@
...
+			const queryKey = event.query.queryKey;
+			if (
+				!Array.isArray(queryKey) ||
+				queryKey[0] !== REACT_QUERY_KEY.AUTO_REFRESH_QUERY
+			) {
+				return;
+			}
+
+			// Update the refresh timestamp in store
+			store.getState().updateRefreshTimestamp();
```
AshwinBhatkal, line 48: should this be global?

frontend/src/store/globalTime/useUrlSync.ts
```diff
@@ -0,0 +1,137 @@
...
+
+interface UrlSyncConfig {
+	relativeTimeKey: string;
+	startTimeKey: string;
+	endTimeKey: string;
+}
+
+export function useUrlSync(
+	store: GlobalTimeStoreApi,
+	enableUrlParams: GlobalTimeProviderOptions['enableUrlParams'],
```
AshwinBhatkal, line 28: what about only calling this hook when enableUrlParams is true?

AshwinBhatkal: also, comments across file would be good

frontend/src/store/globalTime/useComputedMinMaxSync.ts
H4ad: I can add but these hooks are internal ones used only by the Provider, are not intended to be used by us directly.

frontend/src/store/globalTime/types.ts
H4ad: It can but does not make sense to be, not every place needs to be persisted in local storage and since you can enable via URL (I expect everyone to enable), it will cover most of the usages.

frontend/src/store/globalTime/useQueryCacheSync.ts
H4ad: What do you mean? This is not a hook that will be used by us, just by the provider.

I did this just to cleanup the logic of the provider, that's why I use directly store.getState.

frontend/src/store/globalTime/useUrlSync.ts
H4ad: I don't think it's a good idea to have conditional hooks, go against rule of hooks.

frontend/src/store/globalTime/useQueryCacheSync.ts
```diff
@@ -0,0 +1,40 @@
...
+
+			const action = event.action as { type?: string };
+			if (action?.type !== 'success') {
+				return;
+			}
+
+			const queryKey = event.query.queryKey;
+			if (
+				!Array.isArray(queryKey) ||
+				queryKey[0] !== REACT_QUERY_KEY.AUTO_REFRESH_QUERY
```
AshwinBhatkal: can you add a comment for why this is the key that is being checked? I know the reason, but for someone else, they will need context. Maybe a link to the doc with the current implementation works. 

Is it already in the index file in this folder?

frontend/src/store/globalTime/useUrlSync.ts
AshwinBhatkal: True. But I'm thinking we should call this only when it is required. Quite some if checks can be removed with this. Easier to maintain

frontend/src/store/globalTime/useComputedMinMaxSync.ts
AshwinBhatkal: True. But will help others if they ever have to make a change

frontend/package.json
```diff
@@ -231,6 +231,7 @@
     "ts-jest": "29.4.6",
     "ts-node": "^10.2.1",
     "typescript-plugin-css-modules": "5.2.0",
+    "use-sync-external-store": "1.6.0",
```
AshwinBhatkal, line 234: Is this required?

frontend/src/store/globalTime/useQueryCacheSync.ts
AshwinBhatkal: I was thinking refreshTimestamp will be at global level in addition the individual stores

frontend/package.json
H4ad: Due to `useStoreWithEqualityFn`, it's required. https://zustand.docs.pmnd.rs/reference/hooks/use-store-with-equality-fn

frontend/src/store/globalTime/useQueryCacheSync.ts
H4ad: Hm... you are right, but I think instead of removing, what if we keep and then in future I add a way to isolate the react query just for this store? In this way, invalidating queries for this store will not trigger invalidation for all queries under auto refresh (like today, I'm in a modal and the invalidation refreshes also the list table)

H4ad: The next PRs are the refactor to DateTimeSelection, maybe I can introduce there the refactor for the `getAutoRefreshQueryKey` to get that function from the store.

H4ad: Adressed this problem by moving the get auto refresh key to store.

frontend/src/store/globalTime/useQueryCacheSync.ts
H4ad: Docs about auto refresh yes

frontend/src/store/globalTime/useUrlSync.ts
H4ad: Still don't agree, if for some reason someone decides to have conditional enable this prop, the hook will crash due to change in the amount of hooks called.

AshwinBhatkal: I was suggesting we have it without hooks to make it cleaner. But should be okay for now

frontend/src/store/globalTime/utils.ts
```diff
@@ -79,11 +79,60 @@ export function parseSelectedTime(selectedTime: string): ParsedTimeRange {
 }
 
 /**
- * Use to build your react-query key for auto-refresh queries
+ * @deprecated Use store.getAutoRefreshQueryKey() instead.
```
AshwinBhatkal, line 82: Is this used at a lot more places? Some tests in this PR still reference this function. Would be good if we can clean them up

AshwinBhatkal (approved):

frontend/src/store/globalTime/utils.ts
H4ad: I have some PRs that depends on it, so I will clean after merging them.

PR #11088, opened 2026-04-24
feat(infra-monitoring): v2 pods list apis - phase counts when custom grouping

nikhilmantri0902:
## Pull Request

Part of https://github.com/SigNoz/engineering-pod/issues/4609 

When custom grouping, the counts of pods in the various phases should be returned instead of a single phase string

https://github.com/SigNoz/engineering-pod/issues/4609#issuecomment-4304443051

Part of https://github.com/SigNoz/engineering-pod/issues/4609 

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

Follow this thread https://github.com/SigNoz/engineering-pod/issues/4609#issuecomment-4304443051


pkg/modules/inframonitoring/implinframonitoring/pods.go
```diff
@@ -86,75 +200,173 @@ func (m *module) getPodsTableMetadata(ctx context.Context, req *inframonitoringt
...
+	// Resolve tables. Same convention as hosts (distributed names from helpers).
+	adjustedStart, adjustedEnd, _, localTimeSeriesTable := telemetrymetrics.WhichTSTableToUse(
+		uint64(req.Start), uint64(req.End), nil,
+	)
+	samplesTable := telemetrymetrics.WhichSamplesTableToUse(
+		uint64(req.Start), uint64(req.End),
+		metrictypes.UnspecifiedType, metrictypes.TimeAggregationUnspecified, nil,
+	)
+	// Aggregated samples tables hold the latest value in `last`, not `value`.
+	valueCol := "value"
```
namanv3: there might be some method that gets the value column already, you shouldn't need to do this yourself. like I know statement builder uses AggregationColumnForSamplesTable but that returns sum(value) or sum(last) not just the column.

pkg/modules/inframonitoring/implinframonitoring/pods.go
```diff
@@ -86,75 +200,173 @@ func (m *module) getPodsTableMetadata(ctx context.Context, req *inframonitoringt
...
+		timeSeriesFPsGroupBy = append(timeSeriesFPsGroupBy, quoteIdentifier(key.Name))
+	}
+	timeSeriesFPs.GroupBy(timeSeriesFPsGroupBy...)
+	timeSeriesFPsSQL, timeSeriesFPsArgs := timeSeriesFPs.BuildWithFlavor(sqlbuilder.ClickHouse)
 
-		if attrs, ok := metadataMap[compositeKey]; ok && isKeyInGroupByAttrs(groupBy, podUID) {
-			// the condition above ensures we deduce age only if pod uid is in group by because if
-			// it's not in group by then we might have multiple pod uids in the same group and hence then podAge wont make sense
+	// ----- latestPhasePerPod -----
+	// GLOBAL INNER JOIN samples × timeSeriesFPs, collapsed to the latest phase
```
namanv3: Why GLOBAL in comment
Also 'Replaces a prior 2-CTE chain' this comments seems to talk about the change we discussed but if it was never in code lets not mention it

pkg/modules/inframonitoring/implinframonitoring/pods.go
```diff
@@ -86,75 +200,173 @@ func (m *module) getPodsTableMetadata(ctx context.Context, req *inframonitoringt
...
+	// ----- countPodsPerPhase (outer SELECT) -----
+	countPodsPerPhaseSelectCols := make([]string, 0, len(req.GroupBy)+5)
+	countPodsPerPhaseGroupBy := make([]string, 0, len(req.GroupBy))
+	for _, key := range req.GroupBy {
+		col := quoteIdentifier(key.Name)
+		countPodsPerPhaseSelectCols = append(countPodsPerPhaseSelectCols, col)
+		countPodsPerPhaseGroupBy = append(countPodsPerPhaseGroupBy, col)
+	}
+	countPodsPerPhaseSelectCols = append(countPodsPerPhaseSelectCols,
+		"uniqExactIf(pod_uid, phase_value = 1) AS pending_count",
```
namanv3: can we put some map somewhere in the types package for what value of phase_value maps to which count? Would be a good set of constants to keep and reuse elsewhere without having to think everytime

pkg/modules/inframonitoring/implinframonitoring/pods_constants.go
```diff
@@ -53,142 +54,155 @@ var orderByToPodsQueryNames = map[string][]string{
...
-						},
-						GroupBy:  []qbtypes.GroupByKey{podUIDGroupByKey},
-						Disabled: false,
+				GroupBy:  []qbtypes.GroupByKey{podUIDGroupByKey},
+				Disabled: false,
+			},
+		},
+	}
+
+	if includePhaseQuery {
```
namanv3: why not skip G entirely and always use the custom query you have? That's the correct one anyway

nikhilmantri0902: So my take on this is that we should use querier as much as possible. Plus that query is for counts, do u mean to say in case of k8s.pod.uid is in group by, then we know at most count of a single phase will be 1, and then we map it to string?

pkg/modules/inframonitoring/implinframonitoring/pods.go
nikhilmantri0902: okay, should I put it in `telemetrymetrics` package ? I can add two sets of constants there 
1. `PodPhaseRunningNum` numeric values 
2. `PodPhaseRunning` string values 
3. Mapping function

pkg/modules/inframonitoring/implinframonitoring/pods.go
nikhilmantri0902: removed comment as not needed.

pkg/modules/inframonitoring/implinframonitoring/pods.go
nikhilmantri0902: // ValueColumnForSamplesTable returns the bare column name holding
  // the sample value: "last" for the 5m/30m aggregated tables, "value" otherwise.
  ```
  func ValueColumnForSamplesTable(tableName string) string {
      if tableName == SamplesV4Agg5mTableName || tableName == SamplesV4Agg30mTableName {
          return "last"
      }
      return "value"
  }
  ```
  
  should I add this function in telemetrymetrics package?

nikhilmantri0902: Done

pkg/modules/inframonitoring/implinframonitoring/pods_constants.go
nikhilmantri0902: done.

pkg/modules/inframonitoring/implinframonitoring/pods.go
nikhilmantri0902: done as discussed on call, added to inframonitoringytpes.

pkg/modules/inframonitoring/implinframonitoring/pods_constants.go
```diff
@@ -55,10 +55,9 @@ var orderByToPodsQueryNames = map[string][]string{
 }
 
 // newPodsTableListQuery builds the composite QB v5 request for the pods list.
-// includePhaseQuery controls whether query G (per-pod latest phase) is appended.
-// G is meaningful only when k8s.pod.uid is in groupBy (one row = one pod);
-// under custom groupBy, per-group phase counts come from getPerGroupPodPhaseCounts.
-func (m *module) newPodsTableListQuery(includePhaseQuery bool) *qbtypes.QueryRangeRequest {
+// Pod phase is derived separately via getPerGroupPodPhaseCounts (works for both
+// list and grouped_list modes), so no phase query is included here.
+func (m *module) newPodsTableListQuery() *qbtypes.QueryRangeRequest {
```
namanv3, line 60: can you add a comment saying why G was removed

namanv3 (approved):

pkg/modules/inframonitoring/implinframonitoring/pods_constants.go
namanv3: approved otherwise, you can add the comment directly in the target branch or add here and merge after that

nikhilmantri0902: Ya, will add in target branch.

PR #11092, opened 2026-04-24
Add support for custom field selection for color by feature.

krsoninikhil:
## Pull Request

---

We want to provide an experience for user to change how spans are colored in trace detail. Currently, they are always colored by service.name and aggregation appears as execution time percentage for each service. This PR allows UI to fetch different fields in flamegraph API and use those fields in waterfall API to get the aggregated values for those fields. It also exposes new preferences to save the chosed color by attribute and requested attributes (preview fields) on flamegraph.

Why can't UI just read these attributes from waterfall response?
Waterfall response can contain different span then what are shown flamegraph, since flamegraph is sampled for large span and waterfall is paginated.

More details: https://www.notion.so/signoz/Trace-Detail-Improvements-Custom-Analytics-and-Preview-Fields-34cfcc6bcd198096975eeb3725cfd5a6

Flamegraph requriedFields 

image

Waterfall aggregations 

image

https://github.com/SigNoz/engineering-pod/issues/4694

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---


pkg/query-service/model/queryParams.go
```diff
@@ -337,10 +339,11 @@ type GetWaterfallSpansForTraceWithMetadataParams struct {
...
 type GetFlamegraphSpansForTraceParams struct {
-	SelectedSpanID  string `json:"selectedSpanId"`
-	Limit           uint   `json:"limit"`
-	BoundaryStartTS uint64 `json:"boundaryStartTsMilli"`
-	BoundaryEndTS   uint64 `json:"boundarEndTsMilli"`
+	SelectedSpanID  string                             `json:"selectedSpanId"`
+	Limit           uint                               `json:"limit"`
+	BoundaryStartTS uint64                             `json:"boundaryStartTsMilli"`
+	BoundaryEndTS   uint64                             `json:"boundarEndTsMilli"`
+	RequiredFields  []telemetrytypes.TelemetryFieldKey `json:"requiredFields"`
```
nityanandagohain: let's rename to `selectFields` instead of `requiredFields` as it is something we already use in QB

pkg/query-service/model/response.go
```diff
@@ -314,6 +315,30 @@ type FlamegraphSpan struct {
 	Events       []Event           `json:"event"`
 	References   []OtelSpanRef     `json:"references,omitempty"`
 	Children     []*FlamegraphSpan `json:"children"`
+	Attributes   map[string]any    `json:"attributes,omitempty"`
+	Resource     map[string]string `json:"resource,omitempty"`
+}
+
+// SetRequestedFields extracts the requested attribute/resource fields from item into s.
+func (s *FlamegraphSpan) SetRequestedFields(item SpanItemV2, fields []telemetrytypes.TelemetryFieldKey) {
+	for _, field := range fields {
```
nityanandagohain, line 325: in the UI will will just hide other top level fields like `kind` , `kind_string`, `status_code` etc so that user cant select them here ? since we are restricting fields here to  `FieldContextResource` and `FieldContextAttribute`


Also add a note here that eventually this should not be dependent on the user input and should be able to resolve on it's own regardless if the context is given in the input or not.

pkg/query-service/model/trace.go
```diff
@@ -29,3 +29,17 @@ type TraceSummary struct {
 	End      time.Time `ch:"end"`
 	NumSpans uint64    `ch:"num_spans"`
 }
+
+// AttributeValue looks up an attribute across string, number, and bool maps in priority order.
+func (s SpanItemV2) AttributeValue(name string) any {
+	if v, ok := s.Attributes_string[name]; ok && v != "" {
```
nityanandagohain: why `v != ""`  shoudn't it return regardless?

pkg/types/tracedetailtypes/aggregation.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/types/telemetrytypes"
+)
+
+const maxAggregationItems = 10
+
+var ErrTooManyAggregationItems = errors.NewInvalidInputf(errors.CodeInvalidInput, "aggregations request exceeds maximum of %d items", maxAggregationItems)
+
+// SpanAggregationType defines the aggregation to compute over spans grouped by a field.
+type SpanAggregationType string
```
nityanandagohain:
```suggestion
type SpanAggregationType struct {
    valuer.String
}
```


https://github.com/SigNoz/signoz/blob/main/docs/contributing/go/handler.md#enum-method

pkg/types/tracedetailtypes/aggregation.go
```diff
@@ -0,0 +1,50 @@
...
+func (s SpanAggregationType) Enum() []any {
+	return []any{
+		SpanAggregationSpanCount,
+		SpanAggregationExecutionTimePercentage,
+		SpanAggregationDuration,
+	}
+}
+
+func (s SpanAggregationType) isValid() bool {
+	for _, v := range s.Enum() {
```
nityanandagohain:
```suggestion
return slices.ContainsFunc(s.Enum(), func(v any) bool { return v == s })
```

pkg/types/tracedetailtypes/aggregation_test.go
```diff
@@ -0,0 +1,240 @@
...
+				mkASpan("s3", nil, map[string]any{"db.cached": true}, 20, 8),
+			),
+			field: fieldCached,
+			want:  map[string]uint64{"true": 2, "false": 1},
+		},
+		{
+			name: "spans missing the field are excluded",
+			trace: buildTraceFromSpans(
+				mkASpan("s1", map[string]string{"service.name": "frontend"}, nil, 0, 10),
+				mkASpan("s2", map[string]string{}, nil, 10, 5), // no service.name
```
nityanandagohain, line 95: a test case were "service.name": "", so in the result it will be 
```
	want:  map[string]uint64{"frontend": 1, "backend": 1, "": 1},
```

nityanandagohain (commented):
few comments

pkg/query-service/model/response.go
krsoninikhil: Don't want to restrict it but how do allow top level fields like `kind` to be dynamically selected since that would add new field in the response itself? 

They are not required currently, since this is not intended for end user but the UI to request fields they want to allow color and grouping on, which will be only a few selected fields.

> eventually this should not be dependent on the user input

You mean if they don't provide the value for `fieldContext`, we can try something like - attributes first and the resource?

pkg/query-service/model/trace.go
krsoninikhil: Wanted to check if same attribute can have value in `Attributes_number` or `Attributes_bool`. Removed.

pkg/types/tracedetailtypes/aggregation.go
krsoninikhil: Adding `func (...) Enum()` was enough to get below in openapi specs. Not sure why `valuer.String` field is required instead of directly using alias on `string`.

```
    TracedetailtypesSpanAggregationType:
      enum:
      - spanCount
      - executionTimePercentage
      - duration
      type: string
```

pkg/query-service/model/response.go
nityanandagohain: Right now these are restricted to certain fileds from the UI, so we are good to go.

nityanandagohain (dismissed):

nityanandagohain (approved):

therealpandey (approved):

YounixM (approved):

PR #11094, opened 2026-04-24
feat: adding dashboards for Azure One Click Integration services

swapnil-signoz:
### 📄 Summary
- Adding dashboards for CDN and Blob Storage services

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only


nityanandagohain (commented):
since we are using avg in both spatial and temporal aggreation for all panels in blob storage dashboard it doesn’t seem right.

eg:- 

If we take azure_blobcapacity_average, we are already getting an averaged metric, now as an user the most important thing will be to monitor my max capacity so doing average in temporal space isn’t right and for spatial aggregation it should be sum as average will give wrong result.

therealpandey (approved):

nityanandagohain (approved):

PR #11101, opened 2026-04-25
feat(audit): instrument signozapiserver mutation routes with AuditDef

therealpandey:
## Summary

Instruments mutation routes in `pkg/apiserver/signozapiserver/` with `AuditDef` declarations so the audit middleware emits an OTLP-Logs audit event per request. Read routes are not touched. Refactors the `AuditDef` shape along the way so a single route can declare multiple audit events (e.g. a route that creates a resource AND attaches it to a parent).

## AuditDef shape

`AuditDef` is a sealed interface with three implementations:

- `BasicAuditDef{ Resource, Verb, Category, ResourceID }` — single resource, e.g. `dashboard.created`.
- `AttachAuditDef{ AttachedResource, AttachedResourceID, TargetResource, TargetResourceID, Verb, Category }` — one resource attached to another, e.g. `role.attached` → user.
- `AttachManyAuditDef{ AttachedResource, AttachedResourceIDs, TargetResource, TargetResourceID, Verb, Category }` — fans out one attach event per id, e.g. a rule create with N `preferredChannels` emits N `notification-channel.attached` events.

`Resource` is `coretypes.Resource` (not just `Kind`). The middleware emits both `signoz.audit.resource.kind/id` and `signoz.audit.resource.object` (via `Resource.Object(orgID, id)`) so audit log entries correlate 1:1 with authz `Check()` object strings on the same row. For attach events, the same is also emitted under `signoz.audit.target.*`.

Resource picks follow the convention: `ResourceMetaResources{X}` (plural) for Create, `ResourceMetaResource{X}` (singular) or the direct curated form (`ResourceUser`, `ResourceRole`, `ResourceServiceAccount`, `ResourceOrganization`) for Update/Delete.

## Extractors

Resource ids are pulled via `ResourceIDExtractor func(ExtractorContext) (string, error)`. The middleware pre-buffers the request body and captures the response body, so extractors can read either side:

- `PathParam(name)` — Gorilla mux path variable.
- `BodyJSONPath(path)` — gjson lookup on the pre-buffered request body.
- `ResponseJSONPath(path)` — gjson lookup on the captured response body, used by Create routes where the new resource's id is only known after the handler writes the response.
- `BodyJSONArray(path)` — multi-id extractor for `AttachManyAuditDef`.

## Coretypes registry additions

Six new kinds + matching `ResourceMetaResource(s){X}` constants registered in `pkg/types/coretypes/registry_{kind,resource}.go`:

[...]

PR #11103, opened 2026-04-25
feat(auditor): add file provider for audit logs

therealpandey:
## Summary

Add a file-based audit log provider (`ee/auditor/fileauditor/`) alongside `otlphttpauditor`. Useful for local dev, debugging, and offline deployments where the OTLP collector is not reachable. Audit events are converted to OTLP-JSON via `audittypes.NewPLogsFromAuditEvents` + `plog.JSONMarshaler`, then appended to the configured file as one NDJSON line per export batch — same wire format the OTLP transport sends, just landing on disk instead of an HTTP endpoint.

## Changes

- `pkg/auditor/config.go` — new `FileConfig{ Path string }` nested under `Config`; `Validate()` requires `File.Path` when `Provider == "file"`.
- `ee/auditor/fileauditor/provider.go` — `provider` struct + `NewFactory(licensing, build)`. Mirrors `otlphttpauditor` exactly: pre-auth drop on `PrincipalOrgID.IsZero()`, licensing gate via `licensing.GetActive(ctx, orgID)`, buffering through the shared `auditorserver.Server`. Trades the HTTP client / proto marshaler for `*os.File` + `plog.JSONMarshaler`.
- `ee/auditor/fileauditor/export.go` — `export(ctx, events []AuditEvent) error`. Marshals batch to OTLP-JSON, writes payload + newline, calls `file.Sync()`. Mutex-protected so concurrent flushes (none today, but future-proof) stay safe.
- `cmd/enterprise/server.go` — registers `fileauditor.NewFactory(licensing, version.Info)` in the auditor `NamedMap`, alongside `otlphttpauditor`.

OSS factories registry (`pkg/signoz/provider.go:NewAuditorProviderFactories`) is untouched — file provider is enterprise-only, same gating model as OTLP.

## Notes

- **No rotation / size cap** — single append-only file; rotation is a follow-up if needed.
- **NDJSON one-batch-per-line** — each line is a complete OTLP-Logs JSON object containing all events from one flush. Extract per-record with `tail -f /tmp/audit.log | jq -c '.resourceLogs[].scopeLogs[].logRecords[]'`.
- **Buffering identical to OTLP** — `BufferSize`, `BatchSize`, and `FlushInterval` from the top-level `auditor.Config` apply unchanged via the shared `auditorserver`.
- **Licensing parity** — same `licensing.GetActive` gate as OTLP. Audit logs is an enterprise feature regardless of write target.

## Test plan

- [ ] `make go-build-community` and `make go-build-enterprise`
- [ ] Set `auditor.provider=file` and `auditor.file.path=/tmp/audit.log`; run enterprise backend with a valid license; hit `DELETE /api/v2/sessions`; `tail /tmp/audit.log` shows the OTLP-JSON batch with the `session.deleted` event
- [ ] Without an active license, the same request results in zero file writes (gate confirmed)
- [ ] Stop signal drains buffered events into the file before close

### DIFF
--- a/cmd/enterprise/server.go
+++ b/cmd/enterprise/server.go
@@ -8,6 +8,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/SigNoz/signoz/cmd"
+	"github.com/SigNoz/signoz/ee/auditor/fileauditor"
 	"github.com/SigNoz/signoz/ee/auditor/otlphttpauditor"
 	"github.com/SigNoz/signoz/ee/authn/callbackauthn/oidccallbackauthn"
 	"github.com/SigNoz/signoz/ee/authn/callbackauthn/samlcallbackauthn"
@@ -153,6 +154,9 @@ func runServer(ctx context.Context, config signoz.Config, logger *slog.Logger) e
 		func(licensing licensing.Licensing) factory.NamedMap[factory.ProviderFactory[auditor.Auditor, auditor.Config]] {
 			factories := signoz.NewAuditorProviderFactories()
 			if err := factories.Add(otlphttpauditor.NewFactory(licensing, version.Info)); err != nil {
+				panic(err)
+			}
+			if err := factories.Add(fileauditor.NewFactory(licensing, version.Info)); err != nil {
 				panic(err)
 			}
 			return factories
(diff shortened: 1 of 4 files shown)


therealpandey:
Rolling into #11101 — keeping all audit-logs work on the same branch.

PR #11112, opened 2026-04-27
fix: handle ragged label sets in scalar table conversion

nityanandagohain:
## Pull Request

---

### 📄 Summary
PromQL's `by(...)` keeps only labels that were on the source series, so output series can have different label counts. The converter built its columns based on the first series, but then used each series's own label count when filling rows , so any later series with more labels ran off the end.

#### Issues closed by this PR
Fixes https://github.com/SigNoz/engineering-pod/issues/4785

---

### ✅ Change Type
- [x] 🐛 Bug fix

---

### 🐛 Bug Context
The size of row was considered from first series but that was not accurate as the other series might contain more labels and thus overflowing the variable.

#### Root Cause
It just considered the first series.

#### Fix Strategy
Build the column schema from the union of label keys across all series and fill rows by key lookup instead of positional index.

---

- Tests added/updated: New unit tests are added which fails if the change is reverted and the exact error is reproduced.

---

- Blast radius: None
---

### 📋 Checklist
- [x] Tests added 

---


namanv3 (approved):

tushar-signoz:
Hey @nityanandagohain 
Give me till monday to review this.

srikanthccv:
Some context 

- the panic is controlled and only breaks the particular panel in dashboard with something went wrong error
- we don't show promql for table/pie where there can be more than one aggregation series
- we show promql for the number panel type

---

- it's possible that request could come from both api or UI
  - if it comes from the UI, while it resolves the panic, there is another problem
  - the number panel should return multiple series, well, because it's number
     - returning multiple series and showing first series in number panel is misleading
     - if it's just one series, then panic shouldn't happen

The fix itself can go but certainly not addressing the problem, and there should be follow up to resolve the product gaps. Will let pulse-team take it up in https://github.com/SigNoz/engineering-pod/issues/4387

tushar-signoz (approved):
LGTM

srikanthccv (approved):

PR #11130, opened 2026-04-28
move planned maintenance to alertmanager pipeline

### LINKED ISSUE #9270: Move planned downtime to alertmanager
As of now, the planned downtime is implemented by rule manager. It is also implemented is incorrect way that instead of just silencing the alert, it doesn't run query at all, as a consequence the rule state history is not available for this time.

jatinderjit:
### 📄 Summary

Rules currently call `ShouldSkip()` inside the eval loop and skip `rule.Eval()` entirely during active maintenance windows. This has two problems:
1. Gaps in alert state history. The rule never fires, so there is no record that it would have fired had maintenance not been active.
2. Mixed ownership of notification routing logic (specifically, suppression).

The fix moves suppression downstream to the alertmanager notification pipeline. Rules always evaluate and record state; only notification delivery is suppressed. Now eval owns state, **alertmanager owns delivery**.

#### Changes

1. Remove `ShouldSkip` from eval loops
    - `rule_task.go` / `prom_rule_task.go`: removed `ListPlannedMaintenance` call and `shouldSkip` block; `rule.Eval()` now always runs.
    - `task.go`, `manager.go` (OSS + EE): removed `maintenanceStore` from `newTask()`, `PrepareTaskOptions`, and `PrepareTestRuleOptions`. The manager still holds `maintenanceStore` for the maintenance CRUD API.

2. Add `MaintenanceMuter` + `maintenanceMuteStage` (`maintenance_muter.go`)
    - `MaintenanceMuter` implements [`Muter`](https://github.com/prometheus/alertmanager/blob/52342a91814424a5f5f10b89502575bc3f824e8a/notify/mute.go#L33-L38). Reads `ruleId` from the alert label set and checks `PlannedMaintenance.ShouldSkip()`. Results are [cached](https://github.com/SigNoz/signoz/blob/87bf9257b4fd5de660177983b34d513136a98b4f/pkg/alertmanager/alertmanagerserver/maintenance_muter.go#L52) with a 30s TTL to avoid a DB query on every per-alert check.
    - `maintenanceMuteStage` wraps the muter as a `notify.Stage`.

3. Custom `pipelineBuilder` (`pipeline_builder.go`)
    - Copies [`PipelineBuilder.New`](https://github.com/prometheus/alertmanager/blob/52342a91814424a5f5f10b89502575bc3f824e8a/notify/notify.go#L472-L540) (upstream method is not extensible) and inserts `maintenanceMuteStage` immediately before the receiver stage

N/A

#### Issues closed by this PR

- Closes #9270

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Fix Strategy

N/A

---

### 🧪 Testing Strategy

- Tests added/updated: All existing tests updated for the new signatures. `server_test.go` and `server_e2e_test.go` pass `nil` for the maintenance store (pipeline builder is nil-safe).
- Manual verification: Tested.
- Edge cases covered: DB error on `ListPlannedMaintenance` returns stale cache, and does not suppress alerts in maintenance windows.

---

- Blast radius: Alert notification pipeline for all orgs. Rules that were previously silenced during maintenance will now fire into the pipeline and be suppressed there instead — net observable behavior is the same, but state history is now recorded.
- Potential regressions: If `pipelineBuilder` diverges from the upstream `notify.PipelineBuilder` on an alertmanager version bump, the pipeline order could silently differ. The copy is small and the divergence risk is low.
- Rollback plan: Revert the branch; the old `ShouldSkip` logic is self-contained and easy to restore.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Alert state history is now recorded during planned maintenance windows; only notification delivery is suppressed. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- `pipelineBuilder.muter` is nil-safe: when no maintenance store is configured (tests), the stage is omitted and the pipeline is identical to upstream.
- The 30s TTL cache in `MaintenanceMuter` is intentional. Maintenance windows change rarely and a per-alert DB query at dispatch frequency could be expensive.


pkg/alertmanager/alertmanagerserver/distpatcher_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/prometheus/alertmanager/dispatch"
 	"github.com/prometheus/alertmanager/notify"
 	"github.com/prometheus/alertmanager/provider/mem"
+	promTypes "github.com/prometheus/alertmanager/types"
```
srikanthccv: Do we need this change?

pkg/alertmanager/alertmanagerserver/maintenance_muter.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/prometheus/common/model"
+
+	"github.com/SigNoz/signoz/pkg/types/ruletypes"
+)
+
+// MaintenanceMuter implements types.Muter for maintenance windows.
+// It suppresses alerts whose ruleId label matches an active maintenance schedule.
+// Results are cached for cacheTTL to avoid a DB query on every per-alert check.
+type MaintenanceMuter struct {
+	maintenanceStore ruletypes.MaintenanceStore
```
srikanthccv: As discussed earlier, let's move all things related to maintenance to `pkg/alertmanager` or `pkg/types/alertmanagertypes`.

pkg/alertmanager/alertmanagerserver/server.go
```diff
@@ -290,6 +306,7 @@ func (server *Server) SetConfig(ctx context.Context, alertmanagerConfig *alertma
 		server.silencer,
 		intervener,
 		server.marker,
+		server.muter,
```
srikanthccv, line 308: Can we add tests for this?

pkg/alertmanager/alertmanagerserver/distpatcher_test.go
jatinderjit: Need: no!
But removed it because the type was just an alias (not even a wrapper).

```go
type MemMarker = types.MemMarker
```

srikanthccv: let's keep the it was, otherwise there is import mess.

pkg/alertmanager/alertmanagerserver/maintenance_muter_test.go
srikanthccv, line 1: generate the mocks like we do here instead of hand writing the pkg/alertmanager/alertmanagertest/alertmanager.go

pkg/alertmanager/alertmanagerserver/server_e2e_test.go
srikanthccv, line 1: Let's add tests for e2e cases for alert manager.

pkg/alertmanager/signozalertmanager/provider.go
```diff
@@ -4,18 +4,16 @@ import (
...
 	amConfig "github.com/prometheus/alertmanager/config"
+	"github.com/prometheus/common/model"
 
 	"github.com/SigNoz/signoz/pkg/alertmanager"
 	"github.com/SigNoz/signoz/pkg/alertmanager/alertmanagerstore/sqlalertmanagerstore"
 	"github.com/SigNoz/signoz/pkg/alertmanager/nfmanager"
 	"github.com/SigNoz/signoz/pkg/errors"
 	"github.com/SigNoz/signoz/pkg/factory"
 	"github.com/SigNoz/signoz/pkg/modules/organization"
+	"github.com/SigNoz/signoz/pkg/query-service/utils/labels"
```
srikanthccv: what is this `labels` used for? please remove the import from `pkg/query-service` and use the labels pack in rule types or alertmanagertypes.

pkg/query-service/rules/manager.go
```diff
@@ -237,10 +232,11 @@ func (m *Manager) RuleStore() ruletypes.RuleStore {
 	return m.ruleStore
 }
 
-func (m *Manager) MaintenanceStore() ruletypes.MaintenanceStore {
+func (m *Manager) MaintenanceStore() alertmanagertypes.MaintenanceStore {
```
srikanthccv, line 235: do we still need `MaintenanceStore` on manager here?

srikanthccv (commented):
please resolve the new conflicts.

pkg/alertmanager/signozalertmanager/provider.go
jatinderjit: It was for the "ruleId" constant. Now using it from `pkg/types/ruletypes/constants.go`

pkg/query-service/rules/manager.go
jatinderjit: It's used only [here](https://github.com/SigNoz/signoz/blob/b48851e487eb6705b0aa381dbd3bcaf9574a16f6/pkg/ruler/signozruler/provider.go#L133), and that's used only in [handler](https://github.com/SigNoz/signoz/blob/b48851e487eb6705b0aa381dbd3bcaf9574a16f6/pkg/ruler/signozruler/handler.go#L203).
Should I move it directly to the [handler](https://github.com/SigNoz/signoz/blob/b48851e487eb6705b0aa381dbd3bcaf9574a16f6/pkg/ruler/signozruler/handler.go#L19)?

srikanthccv: let's keep it as it is. we will revisit this otherwise it would scope creep the changes in this PR.

pkg/alertmanager/alertmanagerserver/server_e2e_test.go
srikanthccv: to be precise, what I meant is, add tests that verify the muter behaviour in `server_e2e_test.go`

pkg/alertmanager/alertmanagerserver/maintenance_muter_test.go
jatinderjit: Done

pkg/alertmanager/alertmanagerserver/server_e2e_test.go
jatinderjit: Done

pkg/alertmanager/alertmanagerserver/server.go
```diff
@@ -196,17 +203,25 @@ func New(ctx context.Context, logger *slog.Logger, registry prometheus.Registere
...
-		server.inhibitor.Mutes(ctx, labels)
-		server.silencer.Mutes(ctx, labels)
-	}, params)
+	return alertmanagertypes.NewGettableAlertsFromAlertProvider(
+		server.alerts, server.alertmanagerConfig, server.marker.Status,
+		func(labels model.LabelSet) {
+			server.inhibitor.Mutes(ctx, labels)
+			server.silencer.Mutes(ctx, labels)
+		},
+		func(labels model.LabelSet) []string {
```
srikanthccv, line 220: this is not in effect because of the `NewDeprecatedGettableAlertsFromGettableAlerts` in service `GetAlerts`. we should address this (separately, I don't recall the context)

srikanthccv (approved):

pkg/sqlmigration/027_update_rules.go
```diff
@@ -75,7 +75,7 @@ type newMaintenance struct {
 	types.UserAuditable
 	Name        string              `bun:"name,type:text,notnull"`
 	Description string              `bun:"description,type:text"`
-	Schedule    *ruletypes.Schedule `bun:"schedule,type:text,notnull"`
+	Schedule    *alertmanagertypes.Schedule `bun:"schedule,type:text,notnull"`
```
therealpandey, line 78: The only thing to check here is whether the behaviour has subtly changed.

If both the schedules are exactly the same, please feel free to resolve this comment and merge this.

Otherwise create a shadow copy of the older schedule here so as to not change the contract of this earlier migration.

therealpandey (approved):

pkg/sqlmigration/027_update_rules.go
jatinderjit: Ah. I had an old profile active on GH mobile! 😅 

Again, It's the same struct, moved to the alertmanagertypes package.

PR #11133, opened 2026-04-28
feat: adds overview page.

pradeepitm12:
Added system dashboard API endpoints under /api/v1/system/{source}/dashboard
GET /api/v1/system/ai-o11y-overview/dashboard
PUT /api/v1/system/ai-o11y-overview/dashboard
POST /api/v1/system/ai-o11y-overview/dashboard/reset

seeding at two points,
- at org creation,
- existing org migration.

ref: https://www.notion.so/signoz/TRD-AI-Observability-Overview-34afcc6bcd198087b674dcdff6a5c148?d=34afcc6bcd198036a779001c5523194a#34afcc6bcd19806bb688de4ac296fbb5
## Pull Request

---

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---

> [!NOTE]
> **Medium Risk**
> Introduces a new `dashboard.source` column plus migration/seed logic and changes core dashboard queries to filter on `source = ''`, which could affect existing dashboard reads/writes if data/migration assumptions are wrong.
> 
> **Overview**
> Adds **system dashboards** keyed by `source` (initially `ai-o11y-overview`) with new CRUD-ish API endpoints at `/api/v1/system/{source}/dashboard` (GET/PUT plus DELETE to reset to defaults) and updates OpenAPI accordingly.
> 
> Extends the `dashboard` model/storage with a non-null `source` field and adds store methods to `Get/Update/Delete` dashboards by `source`, while ensuring existing dashboard APIs operate only on non-system dashboards via `source = ''` filtering.
> 
> Seeds default system dashboards for both new org creation and existing orgs via a new SQL migration that adds the `source` column and inserts embedded default dashboard JSON per org/source.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 686f234f66ccd4c1f86d86efc792a8017b058840. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/apiserver/signozapiserver/systemdashboard.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/gorilla/mux"
+
+	"github.com/SigNoz/signoz/pkg/http/handler"
+	"github.com/SigNoz/signoz/pkg/types"
+	"github.com/SigNoz/signoz/pkg/types/dashboardtypes"
+)
+
+func (provider *provider) addSystemDashboardRoutes(router *mux.Router) error {
+	if err := router.Handle("/api/v1/system/{source}/dashboard", handler.New(provider.authZ.ViewAccess(provider.systemDashboardHandler.Get), handler.OpenAPIDef{
```
nityanandagohain: feels like the extra `/dashboard` is unnecessary in these endpoints, `/api/v1/system/{source}` should be enough

pkg/apiserver/signozapiserver/systemdashboard.go
```diff
@@ -0,0 +1,66 @@
...
+		ID:                  "GetSystemDashboard",
+		Tags:                []string{"system-dashboard"},
```
nityanandagohain:
```suggestion
		Tags:                []string{"systemdashboard"},
```

pkg/apiserver/signozapiserver/systemdashboard.go
```diff
@@ -0,0 +1,66 @@
...
+		ID:                  "UpdateSystemDashboard",
+		Tags:                []string{"system-dashboard"},
+		Summary:             "Update system dashboard",
+		Description:         "This endpoint replaces the system dashboard for the callers org with the provided payload.",
+		Request:             new(dashboardtypes.UpdatableSystemDashboard),
+		RequestContentType:  "application/json",
+		Response:            new(dashboardtypes.GettableSystemDashboard),
+		ResponseContentType: "application/json",
+		SuccessStatusCode:   http.StatusOK,
+		ErrorStatusCodes:    []int{},
```
nityanandagohain: http.StatusBadRequest

pkg/modules/dashboard/impldashboard/store.go
```diff
@@ -166,6 +186,27 @@ func (store *store) Update(ctx context.Context, orgID valuer.UUID, storableDashb
 	return nil
 }
 
+// UpdateBySource updates the org's system dashboard for the given source.(org_id, source)
+// Delete drops and re-creates the row with a new id, last id from a prior Get should not cause this update to match zero rows.
+func (store *store) UpdateBySource(ctx context.Context, orgID valuer.UUID, source string, storableDashboard *dashboardtypes.StorableDashboard) error {
```
nityanandagohain: don't think this is needed

pkg/modules/dashboard/impldashboard/store.go
```diff
@@ -197,6 +239,22 @@ func (store *store) Delete(ctx context.Context, orgID valuer.UUID, id valuer.UUI
 	return nil
 }
 
+func (store *store) DeleteBySource(ctx context.Context, orgID valuer.UUID, source string) error {
```
nityanandagohain: same here, not needed unless we have a different table altogether

pkg/modules/systemdashboard/implsystemdashboard/handler.go
```diff
@@ -0,0 +1,138 @@
...
+	ctx, cancel := context.WithTimeout(r.Context(), 10*time.Second)
+	defer cancel()
+
+	claims, err := authtypes.ClaimsFromContext(ctx)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	orgID, err := valuer.NewUUID(claims.OrgID)
```
nityanandagohain: MustNewUUID

pkg/modules/systemdashboard/implsystemdashboard/handler.go
```diff
@@ -0,0 +1,138 @@
...
+		render.Error(rw, err)
+		return
+	}
+
+	render.Success(rw, http.StatusOK, dashboardtypes.NewGettableSystemDashboardFromDashboard(dashboard))
+}
+
+func parseSource(r *http.Request) (dashboardtypes.Source, error) {
+	source := mux.Vars(r)["source"]
+	if source == "" {
```
nityanandagohain: cannot be empty as it is a part of path

pkg/modules/systemdashboard/implsystemdashboard/module.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return dashboardtypes.NewDashboardFromStorableDashboard(storableDashboard), nil
+}
+
+// Update applies the new payload as last-writer-wins. The Get and Update run inside one transaction so a
+// concurrent Reset cannot interleave and leave the response with a stale id from before the reset.
+func (module *module) Update(ctx context.Context, orgID valuer.UUID, source dashboardtypes.Source, updatedBy string, data dashboardtypes.UpdatableDashboard) (*dashboardtypes.Dashboard, error) {
```
nityanandagohain: lets not pass updatedBy , it should be a part of dashboard for updated to be sent by handler.

pkg/modules/systemdashboard/systemdashboard.go
```diff
@@ -0,0 +1,25 @@
...
+	"context"
+	"net/http"
+
+	"github.com/SigNoz/signoz/pkg/types/dashboardtypes"
+	"github.com/SigNoz/signoz/pkg/valuer"
+)
+
+type Module interface {
+	Get(ctx context.Context, orgID valuer.UUID, source dashboardtypes.Source) (*dashboardtypes.Dashboard, error)
+	Update(ctx context.Context, orgID valuer.UUID, source dashboardtypes.Source, updatedBy string, data dashboardtypes.UpdatableDashboard) (*dashboardtypes.Dashboard, error)
```
nityanandagohain: Module should work with just Dashboard and not UpdatableDashboard, 

UpdatableDashboard should be limited to handler.

pkg/sqlmigration/078_add_system_dashboard.go
```diff
@@ -0,0 +1,120 @@
...
+	tx, err := db.BeginTx(ctx, nil)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		_ = tx.Rollback()
+	}()
+
+	exists, err := migration.sqlstore.Dialect().ColumnExists(ctx, tx, "dashboard", "source")
```
nityanandagohain: not needed, AddColumn already takes care

pkg/types/dashboardtypes/dashboard.go
```diff
@@ -108,6 +111,30 @@ func NewDashboard(orgID valuer.UUID, createdBy string, storableDashboardData Sto
 	}, nil
 }
 
+// NewSystemDashboard builds a Dashboard owned by the system identified by source (e.g. "ai-o11y-overview").
+func NewSystemDashboard(orgID valuer.UUID, source string, data StorableDashboardData) (*Dashboard, error) {
```
nityanandagohain: Let's modify NewDashboard to accept new param.

pkg/types/dashboardtypes/source.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/valuer"
+)
+
+type Source struct {
+	valuer.String
+}
+
+var (
+	SourceAIO11yOverview = Source{valuer.NewString("ai-o11y-overview")}
```
nityanandagohain: make it an enum

pkg/types/dashboardtypes/source.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+var AllSources = []Source{
+	SourceAIO11yOverview,
+}
+
+func NewSource(s string) (Source, error) {
```
nityanandagohain: don't think this is needed

pkg/types/dashboardtypes/system_dashboard.go
```diff
@@ -0,0 +1,35 @@
+package dashboardtypes
+
+import (
+	"time"
+
+	"github.com/SigNoz/signoz/pkg/valuer"
+)
+
+type GettableSystemDashboard struct {
```
nityanandagohain: Lets add source to existing GettableDashboard

pkg/types/dashboardtypes/ai_o11y_overview.json
```diff
@@ -0,0 +1,107 @@
+{
+  "title": "AI Observability Overview",
```
pradeepitm12: move this in a diff pr itself.

pradeepitm12:
@namanv3 could you please take a look at the PR 🙏

pkg/modules/dashboard/impldashboard/store.go
```diff
@@ -150,14 +169,16 @@ func (store *store) ListPublic(ctx context.Context, orgID valuer.UUID) ([]*dashb
...
 func (store *store) Update(ctx context.Context, orgID valuer.UUID, storableDashboard *dashboardtypes.StorableDashboard) error {
 	_, err := store.
 		sqlstore.
-		BunDB().
+		BunDBCtx(ctx).
 		NewUpdate().
 		Model(storableDashboard).
 		WherePK().
 		Where("org_id = ?", orgID).
+		Where("source = ?", storableDashboard.Source).
```
nityanandagohain: does it matter because we are using WherePK

pkg/modules/systemdashboard/implsystemdashboard/handler.go
```diff
@@ -0,0 +1,102 @@
...
+		render.Error(rw, errors.Wrapf(err, errors.TypeInvalidInput, errors.CodeInvalidInput, "invalid request body"))
+		return
+	}
+
+	if data == nil {
+		render.Error(rw, errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "data is required"))
+		return
+	}
+
+	dashboard, err := handler.module.Update(ctx, valuer.MustNewUUID(claims.OrgID), parseSource(r), &dashboardtypes.Dashboard{
```
nityanandagohain: Use newdashboardfromupdatable in types

pkg/modules/systemdashboard/implsystemdashboard/module.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, err
+	}
+
+	return dashboardtypes.NewDashboardFromStorableDashboard(storable), nil
+}
+
+// Update applies the new payload as last-writer-wins. The Get and Update run inside one transaction so a
+// concurrent Reset cannot interleave and leave the response with a stale id from before the reset.
+func (module *module) Update(ctx context.Context, orgID valuer.UUID, source dashboardtypes.Source, dashboard *dashboardtypes.Dashboard) (*dashboardtypes.Dashboard, error) {
+	if dashboard == nil {
```
nityanandagohain: not required here, the handler can take care of these validations and I see it already there.

pkg/modules/organization/implorganization/setter.go
```diff
@@ -33,6 +35,10 @@ func (module *setter) Create(ctx context.Context, organization *types.Organizati
 		return err
 	}
 
+	if err := module.systemDashboard.SetDefaultConfig(ctx, organization.ID); err != nil {
```
nityanandagohain: so at this point, any kind of restart will reset the user updated dashboards right ?

pkg/modules/systemdashboard/implsystemdashboard/module.go
```diff
@@ -0,0 +1,138 @@
...
+			}
+			if err := module.store.Create(ctx, storable); err != nil {
+				return err
+			}
+			reset = defaultDashboard
+			return nil
+		}
+
+		existing.Data = defaultDashboard.Data
+		existing.UpdatedBy = "system"
```
nityanandagohain: let's not do these, it might break the system

pkg/sqlmigration/078_add_system_dashboard.go
```diff
@@ -0,0 +1,113 @@
...
+			if count > 0 {
+				continue
+			}
+
+			dashboard, err := dashboardtypes.NewDefaultSystemDashboard(orgID, source)
+			if err != nil {
+				return err
+			}
+
+			storable, err := dashboardtypes.NewStorableDashboardFromDashboard(dashboard)
```
nityanandagohain: redundant as you are already doing in settter.go ?

pkg/types/dashboardtypes/defaults.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/valuer"
+)
+
+type Source string
+
+var SystemSources = []Source{}
+
+func NewDefaultSystemDashboard(orgID valuer.UUID, source Source) (*Dashboard, error) {
```
nityanandagohain: ??

nityanandagohain (commented):
some comments

pkg/modules/dashboard/impldashboard/store.go
pradeepitm12: Rechecked, and you are right with whereePK its not required.

pkg/modules/organization/implorganization/setter.go
pradeepitm12: This part Create runs only inside CreateFirstUser. which helps creates the org using either a generated id or a config specified id.

pkg/sqlmigration/078_add_system_dashboard.go
pradeepitm12: Not redundent, this migration is one timer or existing orgs.
In setter is when new org is created.

pkg/modules/systemdashboard/implsystemdashboard/module.go
pradeepitm12: When reset endpoint is called dont we want to set to default dashboard and set updated by system.?

pkg/modules/organization/implorganization/setter.go
nityanandagohain: got it and the migration will then help in case where we already have the first user.

pkg/modules/systemdashboard/implsystemdashboard/module.go
nityanandagohain: Yeah let's keep it empty, since we don't have an actual uuid for system, it will break things later.

pkg/sqlmigration/078_add_system_dashboard.go
nityanandagohain: Makes sense

nityanandagohain (commented):
@pradeepitm12  NewDefaultSystemDashboard is the part I am confused about, since NewDefaultSystemDashboard will already run inside the migration once this is merged, when you add the json again, it won't call NewDefaultSystemDashboard. So This is something to discuss about. Apart from this other looks good.

Might be good idea to get some suggestion from @therealpandey on this.

pradeepitm12:
Hi @therealpandey can you please take a look at this pr. Let me know if you need any help with the context.

pkg/modules/systemdashboard/implsystemdashboard/module.go
pradeepitm12: keep existing.UpdatedBy = "" empty we might loose who actually invoked reset. 
How about we get the claims and pass updatedBy to this function and 
existing.UpdatedBy = updatedByArg

nityanandagohain: keeping the user id who updated it will be better

pradeepitm12:
> @pradeepitm12 NewDefaultSystemDashboard is the part I am confused about, since NewDefaultSystemDashboard will already run inside the migration once this is merged, when you add the json again, it won't call NewDefaultSystemDashboard. So This is something to discuss about. Apart from this other looks good.
> 
> Might be good idea to get some suggestion from @therealpandey on this.

I get issue, the migration run and it gets marked as applied in migration table so it never runs again.
What I have done is commented out the iteration overs orgs, now we do not mark the migration done in the migration table, we only add the column to the dashboard table.
Once we have the other pr with default values we active this part of the code and columns already exists we set default value and mark applied in migration table.

pkg/modules/systemdashboard/implsystemdashboard/module.go
```diff
@@ -0,0 +1,131 @@
+package implsystemdashboard
```
therealpandey: Club with the existing dashboard module. There is no need to create a separate module.

pkg/apiserver/signozapiserver/systemdashboard.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/gorilla/mux"
+
+	"github.com/SigNoz/signoz/pkg/http/handler"
+	"github.com/SigNoz/signoz/pkg/types"
+	"github.com/SigNoz/signoz/pkg/types/dashboardtypes"
+)
+
+func (provider *provider) addSystemDashboardRoutes(router *mux.Router) error {
+	if err := router.Handle("/api/v1/system/{source}", handler.New(provider.authZ.ViewAccess(provider.systemDashboardHandler.Get), handler.OpenAPIDef{
```
therealpandey:
```suggestion
	if err := router.Handle("/api/v1/dashboards?source=system"
```

pkg/apiserver/signozapiserver/systemdashboard.go
```diff
@@ -0,0 +1,66 @@
...
+		ResponseContentType: "application/json",
+		SuccessStatusCode:   http.StatusOK,
+		ErrorStatusCodes:    []int{http.StatusBadRequest},
+		Deprecated:          false,
+		SecuritySchemes:     newSecuritySchemes(types.RoleViewer),
+	})).Methods(http.MethodGet).GetError(); err != nil {
+		return err
+	}
+
+	if err := router.Handle("/api/v1/system/{source}", handler.New(provider.authZ.EditAccess(provider.systemDashboardHandler.Update), handler.OpenAPIDef{
```
therealpandey: We can edit system dashboards? Then how are they different from normal dashboards? They just can't be deleted?

therealpandey (changes requested):

pkg/apiserver/signozapiserver/systemdashboard.go
nityanandagohain: Yes
* They won't appear in dashboards list page.
* They can't be deleted.
* They are tied to a specific page on the product.

pkg/modules/systemdashboard/implsystemdashboard/module.go
nityanandagohain: So the idea was to support different part of the product with dashboard like facility without mixing with the dashboards. From the storage layer srikanth suggested to start with with existing dashboard table and then if required to create one later. 

But now looking at your comment I guess you are coming from how public dashboards are added i.e in the same dashboards module.

pkg/apiserver/signozapiserver/systemdashboard.go
nityanandagohain: so for a page like ai-o11y.


It would have been `/api/v1/system/ai-o11y` . Now it will be `/api/v1/dashboards?source=ai-o11y` . 

Should we do this as only read and update is possible for these dashboards. Or restricting these at the existing handler level is fine ?

pkg/modules/systemdashboard/implsystemdashboard/module.go
pradeepitm12: It does feel quite similar but address separate concern, it would clutter dashboard.
I also see this similar partten with quickfilters.

pkg/apiserver/signozapiserver/systemdashboard.go
pradeepitm12: we did had dashbaord path in the first commit 
```
GET /api/v1/system/ai-o11y-overview/dashboard
PUT /api/v1/system/ai-o11y-overview/dashboard
POST /api/v1/system/ai-o11y-overview/dashboard/reset
```
source is not actually system, source is "ai-o11y-overview" and in near future can be "service-overview"
I also see there could be conflict in Update. 
How about if we change to /api/v1/dashboards/system/{source}
```
GET    /api/v1/dashboards/system/{source}
PUT    /api/v1/dashboards/system/{source}
POST   /api/v1/dashboards/system/{source}/reset
```

docs/api/openapi.yml
```diff
@@ -10597,6 +10600,186 @@ paths:
       summary: Update a span mapper
       tags:
       - spanmapper
+  /api/v1/system/{source}:
```
namanv3: not system/dashboard/source?

docs/api/openapi.yml
```diff
@@ -10597,6 +10600,186 @@ paths:
...
+          description: Forbidden
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RenderErrorResponse'
+          description: Internal Server Error
+      security:
+      - api_key:
+        - EDITOR
```
namanv3: why not admin? just asking for my understanding about the persona we expect to edit system dashboards

pkg/modules/dashboard/impldashboard/store.go
```diff
@@ -63,6 +64,23 @@ func (store *store) Get(ctx context.Context, orgID valuer.UUID, id valuer.UUID)
...
 
+func (store *store) GetBySource(ctx context.Context, orgID valuer.UUID, source string) (*dashboardtypes.StorableDashboard, error) {
+	storableDashboard := new(dashboardtypes.StorableDashboard)
+	err := store.
+		sqlstore.
+		BunDBCtx(ctx).
+		NewSelect().
+		Model(storableDashboard).
+		Where("org_id = ?", orgID).
+		Where("source = ?", source).
```
namanv3: will we need an orgID+source index? any perf numbers for whether its alright to not have one or not?

namanv3: my guess is that a partial index for source == "" will be required

pkg/types/dashboardtypes/defaults.go
namanv3, line 1: didn't get what this file is doing

pkg/modules/systemdashboard/implsystemdashboard/module.go
```diff
@@ -0,0 +1,131 @@
...
+
+	return dashboardtypes.NewDashboardFromStorableDashboard(storable), nil
+}
+
+// Update applies the new payload as last-writer-wins. The Get and Update run inside one transaction so a
+// concurrent Reset cannot interleave and leave the response with a stale id from before the reset.
+func (module *module) Update(ctx context.Context, orgID valuer.UUID, source dashboardtypes.Source, dashboard *dashboardtypes.Dashboard) (*dashboardtypes.Dashboard, error) {
+	var updated *dashboardtypes.Dashboard
+	err := module.store.RunInTx(ctx, func(ctx context.Context) error {
+		existing, err := module.store.GetBySource(ctx, orgID, string(source))
```
namanv3: why is get in a transaction? it doesn't change anything so doesn't make sense. transaction isn't a lock if that's what the attempt is

pkg/modules/systemdashboard/implsystemdashboard/module.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return dashboardtypes.NewDashboardFromStorableDashboard(storable), nil
+}
+
+// Update applies the new payload as last-writer-wins. The Get and Update run inside one transaction so a
+// concurrent Reset cannot interleave and leave the response with a stale id from before the reset.
+func (module *module) Update(ctx context.Context, orgID valuer.UUID, source dashboardtypes.Source, dashboard *dashboardtypes.Dashboard) (*dashboardtypes.Dashboard, error) {
```
namanv3: looking at this I don't see any real reason why the dashboard module's update can't be used

pkg/modules/systemdashboard/implsystemdashboard/module.go
```diff
@@ -0,0 +1,131 @@
...
+	err := module.store.RunInTx(ctx, func(ctx context.Context) error {
+		existing, err := module.store.GetBySource(ctx, orgID, string(source))
+		if err != nil {
+			return err
+		}
+
+		existing.Data = dashboard.Data
+		existing.UpdatedBy = dashboard.UpdatedBy
+		existing.UpdatedAt = time.Now()
+
```
namanv3: there should be a types method to do all this

pkg/modules/dashboard/impldashboard/module.go
namanv3, line 1: other methods like update and all need to check if the API is being used to accidentally update a source != "" dashboard.

pkg/modules/systemdashboard/implsystemdashboard/module.go
namanv3: maybe even check if we should just use the same APIs

docs/api/openapi.yml
pradeepitm12: @nityanandagohain I will change it to /api/v1/system/dashboard/source

pradeepitm12: Going to use existing path with query params.
- Most of the dashboard related endpoints are registered using http_handler, will add `reset` as well there with query parameters, will not have any docs in the openapi.
cc: @nityanandagohain , @therealpandey  please let me know if any concerns.

ee/modules/dashboard/impldashboard/module.go
```diff
@@ -138,6 +138,11 @@ func (module *module) Delete(ctx context.Context, orgID valuer.UUID, id valuer.U
 		return err
 	}
 
+	// Do not delete system dashboard, we can only reset system dashbard to default.
+	if dashboard.Source != "" {
```
namanv3: reset isn't an alternative to delete really, let's not suggest it

ee/modules/dashboard/impldashboard/module.go
```diff
@@ -138,6 +138,11 @@ func (module *module) Delete(ctx context.Context, orgID valuer.UUID, id valuer.U
...
+		return errors.Newf(errors.TypeInvalidInput, errors.CodeInvalidInput, "cannot delete system dashboard with source %s, use reset instead", dashboard.Source)
+	}
+
 	if dashboard.Locked {
```
namanv3, line 162: better to add a CanDelete method in the types package and add the 2 if conditions there

pkg/modules/dashboard/impldashboard/handler.go
```diff
@@ -185,6 +185,49 @@ func (handler *handler) LockUnlock(rw http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	id, err := valuer.NewUUID(mux.Vars(r)["id"])
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	dashboard, err := handler.module.Get(ctx, orgID, id)
```
namanv3: module call should not be in the handler. ideally the module method for reset should get the dashboard

pkg/modules/dashboard/impldashboard/handler.go
```diff
@@ -185,6 +185,49 @@ func (handler *handler) LockUnlock(rw http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	if dashboard.Source == "" {
```
namanv3: CanReset method in types package that is then used in the module

pkg/modules/dashboard/impldashboard/handler.go
namanv3: why is the API endpoint taking a dashboard ID though? Given that a source for an org can appear only once, and the module method takes a source, seems unnatural to use dashboard ID in the API anyway

pkg/types/dashboardtypes/dashboard.go
```diff
@@ -278,6 +283,12 @@ func (dashboard *Dashboard) Update(ctx context.Context, updatableDashboard Updat
 	return nil
 }
 
+func (dashboard *Dashboard) OverwriteData(updatableDashboard UpdatableDashboard, updatedBy string) {
```
namanv3: dashboard.Update does this already?

pkg/sqlmigration/081_add_system_dashboard.go
namanv3, line 1: might need to bump the number, 78 is already on main at least

pkg/modules/dashboard/dashboard.go
```diff
@@ -42,6 +42,11 @@ type Module interface {
 
 	Update(ctx context.Context, orgID valuer.UUID, id valuer.UUID, updatedBy string, data dashboardtypes.UpdatableDashboard, diff int) (*dashboardtypes.Dashboard, error)
 
+	// Reset puts back default value for system dashboard.
+	Reset(ctx context.Context, orgID valuer.UUID, source dashboardtypes.Source, updatedBy string) (*dashboardtypes.Dashboard, error)
```
namanv3: do you think it makes sense to name it ResetSystemDashboard?

pkg/modules/systemdashboard/implsystemdashboard/module.go
namanv3: @pradeepitm12  what I meant here was that there might be one already, my bad

pkg/sqlmigration/081_add_system_dashboard.go
pradeepitm12: added 3 days ago, will fix.

pkg/modules/dashboard/impldashboard/handler.go
pradeepitm12: tried following existing patterns `	router.HandleFunc("/api/v1/dashboards/{id}/lock", am.EditAccess(aH.Signoz.Handlers.Dashboard.LockUnlock)).Methods(http.MethodPut)
`
But it makes sense, we can get the source as QP and we already extract orgid.

pkg/types/dashboardtypes/dashboard.go
pradeepitm12: @namanv3 actually dashboard.Update is almost similar the difference is the CanUpdate lock check, which would either block Reset for a locked system dashboard or get bypassed from Reset's side. OverwriteData is the bypass any ways happy to change this around (disallow locking system dashboards in LockUnlock, then dashboard.Update works for both and we drop OverwriteData).

namanv3: I think locking should be allowed in system dashboards though. It can prevent both accidental resets and accidental updates from users. All the reasons to have locking in a user dashboard apply to system dashboards

pkg/modules/dashboard/impldashboard/module.go
```diff
@@ -91,14 +99,86 @@ func (module *module) Update(ctx context.Context, orgID valuer.UUID, id valuer.U
...
+	}
+	if err := module.store.Update(ctx, orgID, storable); err != nil {
+		return nil, err
+	}
+
+	return existing, nil
+}
+
+// SetDefaultConfig seeds default values for system dashboards for newly created orgs.
+func (module *module) SetDefaultConfig(ctx context.Context, orgID valuer.UUID) error {
```
namanv3: lot of common logic between reset and this method. worth checking if some common code can be taken out. no worries if not

pkg/modules/dashboard/impldashboard/handler.go
```diff
@@ -185,6 +185,38 @@ func (handler *handler) LockUnlock(rw http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	orgID, err := valuer.NewUUID(claims.OrgID)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	source := r.URL.Query().Get("source")
```
namanv3: query param being required feels weird. why not make this a path param though?

pkg/types/dashboardtypes/dashboard.go
```diff
@@ -96,8 +99,8 @@ func NewDashboard(orgID valuer.UUID, createdBy string, storableDashboardData Sto
 			UpdatedBy: createdBy,
 		},
 		OrgID:  orgID,
-		Data:   storableDashboardData,
-		Locked: false,
```
namanv3: nit: add explicit locked: false setup for easier readability

pkg/modules/dashboard/impldashboard/handler.go
pradeepitm12: actually you are right multiple iteration made me assume we have both
POST /api/v1/dashboards/{id}/reset 
POST /api/v1/dashboards/{source}/reset 
so added QP way.
Will fix this as POST /api/v1/dashboards/{source}/reset

pkg/modules/dashboard/impldashboard/module.go
pradeepitm12: yes we can extract the insert part for sure.

pkg/modules/dashboard/impldashboard/handler.go
```diff
@@ -185,6 +185,38 @@ func (handler *handler) LockUnlock(rw http.ResponseWriter, r *http.Request) {
 
 }
 
+// ResetSystemDashboard only resets system dashboard (source != "") to default values.
+func (handler *handler) ResetSystemDashboard(rw http.ResponseWriter, r *http.Request) {
```
namanv3, line 189: this isn't coming in the open api spec, are you planning to add it later on?

pkg/modules/dashboard/impldashboard/handler.go
namanv3: not needed anymore cuz we check by source anyway, resolving

pkg/modules/dashboard/impldashboard/module.go
```diff
@@ -91,14 +99,87 @@ func (module *module) Update(ctx context.Context, orgID valuer.UUID, id valuer.U
...
+func (module *module) ResetSystemDashboard(ctx context.Context, orgID valuer.UUID, source dashboardtypes.Source, updatedBy string) (*dashboardtypes.Dashboard, error) {
+	existing, err := module.GetBySource(ctx, orgID, source)
+	if err != nil && !errors.Ast(err, errors.TypeNotFound) {
+		return nil, err
+	}
+
+	if existing == nil {
+		return module.insertSystemDashboard(ctx, orgID, source, updatedBy)
+	}
+
```
namanv3, line 121: lock check should be there right?

pkg/modules/dashboard/impldashboard/handler.go
pradeepitm12: Discussed earlier in this thread [here](https://github.com/SigNoz/signoz/pull/11133#discussion_r3207440340]) going to use the existing path under /api/v1/dashboards/* and register via http_handler.go. Most of the dashboard endpoints (Get, List, Update, Delete, Lock) live in pkg/query-service/app/http_handler.go

pkg/modules/dashboard/impldashboard/module.go
pradeepitm12: lock would actually make sense for an existing row.

pkg/types/dashboardtypes/defaults.go
```diff
@@ -0,0 +1,17 @@
+package dashboardtypes
+
+import (
+	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/valuer"
+)
+
+type Source string
```
srikanthccv: Use valuer for defining enum, please see how enum are defined in types package, example https://github.com/SigNoz/signoz/blob/eb95364abad45a020fb1dbb6e5f398306789dc52/pkg/types/metrictypes/metrictypes.go#L13-L25

pkg/types/dashboardtypes/defaults.go
srikanthccv: To remove this as we decided to remove the seeding and it's related code from this PR.

pkg/types/dashboardtypes/store.go
```diff
@@ -13,6 +13,8 @@ type Store interface {
 
 	Get(context.Context, valuer.UUID, valuer.UUID) (*StorableDashboard, error)
 
+	GetBySource(context.Context, valuer.UUID, string) (*StorableDashboard, error)
```
srikanthccv: Use the defined enum type for the 3rd arg.

pkg/types/dashboardtypes/dashboard.go
```diff
@@ -30,6 +30,7 @@ type StorableDashboard struct {
 	Data   StorableDashboardData `bun:"data,type:text,notnull"`
 	Locked bool                  `bun:"locked,notnull,default:false"`
 	OrgID  valuer.UUID           `bun:"org_id,notnull"`
+	Source string                `bun:"source,type:text,notnull"`
```
srikanthccv: same update the type here from string to the defined enumerator type in this file

pkg/types/dashboardtypes/dashboard.go
```diff
@@ -278,6 +284,16 @@ func (dashboard *Dashboard) Update(ctx context.Context, updatableDashboard Updat
 	return nil
 }
 
+func (dashboard *Dashboard) CanDelete() error {
+	if dashboard.Source != "" {
```
srikanthccv: post enum def, use `IsZero` here

pkg/sqlmigration/081_add_system_dashboard.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	sqls := migration.sqlschema.Operator().AddColumn(table, uniqueConstraints, column, "")
+	for _, sql := range sqls {
+		if _, err := tx.ExecContext(ctx, string(sql)); err != nil {
+			return err
+		}
+	}
+
+	// We activate this part of code once we add default value for the overview page.
```
srikanthccv: to remove everything from here except adding the column

pkg/sqlmigration/081_add_system_dashboard.go
```diff
@@ -0,0 +1,124 @@
...
+		return err
+	}
+
+	column := &sqlschema.Column{
+		Name:     sqlschema.ColumnName("source"),
+		DataType: sqlschema.DataTypeText,
+		Nullable: false,
+	}
+
+	sqls := migration.sqlschema.Operator().AddColumn(table, uniqueConstraints, column, "")
```
srikanthccv: set `user` as the default value for `source` and use `system` value for your ai-overview

pkg/modules/dashboard/impldashboard/module.go
```diff
@@ -91,14 +99,87 @@ func (module *module) Update(ctx context.Context, orgID valuer.UUID, id valuer.U
...
 		return nil, err
 	}
+	if err := module.store.Create(ctx, storable); err != nil {
+		return nil, err
+	}
 
 	return dashboard, nil
 }
 
+func (module *module) ResetSystemDashboard(ctx context.Context, orgID valuer.UUID, source dashboardtypes.Source, updatedBy string) (*dashboardtypes.Dashboard, error) {
```
srikanthccv: validate the `source` input value

pkg/modules/dashboard/impldashboard/module.go
```diff
@@ -91,14 +99,87 @@ func (module *module) Update(ctx context.Context, orgID valuer.UUID, id valuer.U
...
+	existing, err := module.GetBySource(ctx, orgID, source)
+	if err != nil && !errors.Ast(err, errors.TypeNotFound) {
+		return nil, err
+	}
+
+	if existing == nil {
```
srikanthccv: Will this ever happen? The default system dashboard would always be present because of the seeding at the bootstrap?

therealpandey:
@pradeepitm12 please rebase with main, let's try to get this merged.

pradeepitm12:
@srikanthccv @therealpandey please take a look at the pr.
as source is enum with strict values  the  system dashboards now use the per-name identifier
(e.g. ai-o11y-overview) is stored inside Data["system_source"].
ref: https://github.com/SigNoz/signoz/pull/11367#discussion_r3273447378  
no seeding code part present in this pr

pradeepitm12:
> @srikanthccv @therealpandey please take a look at the pr. as source is enum with strict values the system dashboards now use the per-name identifier (e.g. ai-o11y-overview) is stored inside Data["system_source"]. ref: [#11367 (comment)](https://github.com/SigNoz/signoz/pull/11367#discussion_r3273447378) no seeding code part present in this pr

@therealpandey  ready for review, and as per discussion with @srikanthccv , I have updated the PR with assumption that we have ai-o11y-overview as system dashboard, and we will be able to handle other system dashboards which may appear in future with one of upcoming prs having unique Name row in Dashboard table.

PR #11147, opened 2026-04-30
fix: trace operator query for clickhouse version 25.12.5

piyushsingariya:
## Pull Request

---

### 📄 Summary

Fixes issue with ClickHouse error in TraceOperator

#### Issues closed by this PR

Closes signoz/engineering-pod#4754

---

### ✅ Change Type
_Select all that apply_

- [x] 🐛 Bug fix

---

### 🐛 Bug Context
```
{"status":"error","error":{"code":"unknown","message":"code: 10, message: Not found column __table1.timestamp in block. There are only columns: __table1.timestamp_0, __global_row_index, ts_bucket_start, attributes_string"}}
```

#### Root Cause
The failure is a ClickHouse 25.12.5 server-side regression: when signoz_index_v3 contains any data, CH's new distributed query analyzer renames the timestamp column to timestamp_0 while constructing the execution plan for this specific CTE shape, making it invisible to the outer SELECT.

It fails on any core field selected in OrderBy

#### Fix Strategy

In Inner Select, select all core fields by alias, and in outer select flip alias back to original field keys

---

- Tests added/updated: ✅ 
- Manual verification: 
- Edge cases covered:

---

- Blast radius: Alerts, Traces Explorer
- Potential regressions:
- Rollback plan: Revert pull request

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---


tushar-signoz (changes requested):
Hey @piyushsingariya 
Lets the rawexportdata integration test remain as is. This helps in picking up behaviour change much easier. 
Add integration tests in the `querier` for TraceOperatorQuery (rawexportdata is not the apt place)

tests/integration/tests/querier/15_trace_operator.py
```diff
@@ -0,0 +1,537 @@
...
+                expression="A -> B",  # indirect descendant
+                return_spans_from="A",
+                limit=100,
+                select_fields=[
+                    TelemetryFieldKey(name="service.name", field_data_type="string", field_context="resource"),
+                ],
+                order=[
+                    # http.method is intentionally absent from select_fields
+                    OrderBy(
+                        key=TelemetryFieldKey(name="http.method", field_data_type="string", field_context="tag"),
```
tushar-signoz: Hey @piyushsingariya 
Im not sure how `tag` as fieldContext is coming here. If it is a legacy fieldContext and `span` would still work here then lets use `span`. 
Same for other occurences.

tushar-signoz (commented):
@piyushsingariya Integration tests not passing

tests/integration/tests/querier/15_trace_operator.py
piyushsingariya: resolved: thanks for catching them

tushar-signoz (approved):
Hey @piyushsingariya 
Honestly this doesn't feel right. However it fixes the problem.
Lets create a ticket to re-visit this.

piyushsingariya:
> Honestly this doesn't feel right.

Agreed, the pattern is weird and only happens to corefields. I've added tests, we can come here later and fix in an organized manner

tushar-signoz:
@srikanthccv If we can get this merged today, it will go out in the release tomorrow

srikanthccv (approved):

srikanthccv:
@tushar-signoz added addl context here https://signoz-team.slack.com/archives/C089D1B5516/p1778577108205259?thread_ts=1778480420.401969&cid=C089D1B5516, please sync with @piyushsingariya once.
tl;dr
- the experience is not broken in the product today for trace operators
- the change here focuses on CH error that only appears when `returnSpansFrom` is non-empty
   - the `returnSpansFrom` has always been broken, past discussions https://signoz-team.slack.com/archives/C089D1B5516/p1755780684195609

piyushsingariya:
Closing this, if this error stays after addressing broken issue with 'returnSpansFrom' then should handled with realiasing

PR #11153, opened 2026-04-30
chore(openapi): clean up generated schema names

SferaDev:
## Summary

Default schema names from `swaggest/jsonschema-go` produce verbose keys for generic instantiations and `*types` packages, e.g.
`Querybuildertypesv5QueryBuilderQueryGithubComSigNozSignozPkgTypesQuerybuildertypesQuerybuildertypesv5LogAggregation`.

The `InterceptDefName` hook in `pkg/signoz/openapi.go` now:

- Collapses generic instantiations to base + leaf-name concatenation, so `QueryBuilderQuery[…LogAggregation]` becomes `QueryBuilderQueryLogAggregation`.
- Strips the `types` / `typesvN` suffix from the package basename and de-stutters when the type name already starts with the prefix (`sigv4.SigV4Config` becomes `SigV4Config`; `types.User` becomes `User`).
- Falls back to a small override map for packages whose all-lowercase Go names lose word boundaries (`alertmanagertypes` becomes `AlertManager`, `inframonitoringtypes` becomes `InfraMonitoring`, etc).

Only schema keys and `$ref` strings change. Property names, JSON shapes, operation IDs, tags, and the existing `RenderSuccessResponse` handling are all unchanged.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit e06c1b73b54d54794284d09be25d9c0439e4d3ab. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -1,6 +1,6 @@
 components:
   schemas:
-    AlertmanagertypesChannel:
+    AlertManagerChannel:
       properties:
         createdAt:
           format: date-time
@@ -25,7 +25,7 @@ components:
       - data
       - orgId
       type: object
-    AlertmanagertypesDeprecatedGettableAlert:
+    AlertManagerDeprecatedGettableAlert:
       properties:
         annotations:
           $ref: '#/components/schemas/ModelLabelSet'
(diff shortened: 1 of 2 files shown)


therealpandey:
Hi @SferaDev,
This will break multiple places in the frontend as these names are used there as well.

SferaDev:
But these names are not usable as they are 😞

PR #11159, opened 2026-05-01
feat: user dashboard preference

### LINKED ISSUE #4958: Link multiple graphs in dashboard with time
Hover over linked charts in dashboard at time T and have all other charts show the same vertical time for their respective time T

ahrefabhi:
## 📄 Summary

Adds a per-dashboard user preference for cross-panel cursor sync (Crosshair / Tooltip / No Sync), exposed in the dashboard General settings panel and persisted locally per dashboard ID.

image

#### Issues closed by this PR

Closes https://github.com/SigNoz/signoz/issues/4958

---

## ✅ Change Type

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

## 🐛 Bug Context

N/A — this is a feature addition, not a bug fix.

---

## 🧪 Testing Strategy

- **Tests added/updated:**
  - `hooks/dashboard/__test__/useDashboardPreference.test.ts` — read/write/default behavior, per-dashboard isolation, cross-tab sync via `storage` event, and resilience to malformed localStorage payloads.
  - `hooks/dashboard/__test__/useDashboardCursorSyncMode.test.ts` — defaulting to `Crosshair` and persisting selected mode under the `cursorSyncMode` key.
  - `lib/uPlotV2/plugins/__tests__/syncDisplayHook.test.ts` — source/receiver branches, groupBy intersection logic, hidden-series handling, and y-crosshair visibility based on `yAxisUnit` match.
- **Manual verification:**
  - Switch between No Sync / Crosshair / Tooltip in Dashboard Settings → General → Cross-Panel Sync and confirm panels update without a page refresh.
  - Reload the dashboard and confirm the selection sticks.
  - Open the same dashboard in two tabs and confirm the cross-tab `storage` event syncs the selection.
  - Switch dashboards and confirm each dashboard remembers its own setting independently.
- **Edge cases covered:**
  - Undefined `dashboardId` (no-op writes, returns default).
  - Malformed/non-object JSON in `DASHBOARD_PREFERENCES` (falls back to default, logs a warning).
  - Switching to "No Sync" — panels remount via `key={cursorSyncMode}` so the previous mode's mutated cursor config is discarded rather than inherited.
  - Non-dashboard panel modes force `DashboardCursorSync.None` regardless of the stored preference.

---

## ⚠️ Risk & Impact Assessment

- **Blast radius:** Limited to dashboard panel rendering (BarPanel, TimeSeriesPanel) and the General settings tab. No backend / API changes. Default behavior matches the prior hardcoded `Crosshair` for users with no stored preference.
- **Potential regressions:**
  - `config` memo dependency now includes `syncMode`, so panels rebuild config and remount on mode change — verified intentional (see inline comment) to avoid carrying mutated cursor config across modes. Worth checking that this does not visibly flicker on dashboards with many panels.
  - localStorage growth is bounded (one entry per dashboard the user has visited), but stale entries for deleted dashboards are not cleaned up.
- **Rollback plan:** Revert the single commit on this branch. No migrations or persisted server state involved; users with stored preferences will simply revert to the previous fixed-Crosshair behavior on next load.

---

## 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Dashboards now support a per-dashboard Cross-Panel Sync preference (No Sync / Crosshair / Tooltip), available in Dashboard Settings → General. The selection is remembered locally per dashboard. |

---

## 📋 Checklist

- [x] Manually tested
- [x] Backward compatibility considered — defaults to existing Crosshair behavior when no preference is stored.

---

## 👀 Notes for Reviewers

- `useDashboardPreference` is intentionally generic (`<K extends keyof DashboardPreferences>`) so future per-dashboard preferences can be added by extending the `DashboardPreferences` type and writing a thin wrapper hook (see `useDashboardCursorSyncMode` as the reference shape).
- The `useSyncExternalStore` + module-level `subscribers` set is what keeps multiple consumers (BarPanel, TimeSeriesPanel, GeneralSettings) in sync within a tab; the `storage` event listener handles the cross-tab case.
- Panels intentionally remount on `syncMode` change via `key={cursorSyncMode}` because the uPlot config object is mutated by `TooltipPlugin` — see inline comments in `BarPanel.tsx` / `TimeSeriesPanel.tsx`.
- `syncDisplayHook.ts` changes are small (y-crosshair visibility timing) and fully covered by the new tests.


frontend/src/container/DashboardContainer/DashboardSettings/General/GeneralSettings.styles.scss
```diff
@@ -1,12 +1,43 @@
 .overview-content {
 	display: flex;
```
AshwinBhatkal: can we replace this with module.scss

frontend/src/container/DashboardContainer/visualization/panels/TimeSeriesPanel/TimeSeriesPanel.tsx
```diff
@@ -109,10 +121,18 @@ function TimeSeriesPanel(props: PanelWrapperProps): JSX.Element {
 		return get(widget, 'query.builder.queryData[0].groupBy', []);
 	}, [widget.query]);
 
+	const cursorSyncMode = useMemo(() => {
```
AshwinBhatkal: can we move this out similar to groupBy

frontend/src/hooks/dashboard/useDashboardPreference.ts
```diff
@@ -0,0 +1,107 @@
...
+
+// Per-dashboard preferences persisted in localStorage. Add new preference
+// fields here as they are introduced.
+export type DashboardPreferences = {
+	cursorSyncMode?: DashboardCursorSync;
+};
+
+type DashboardPreferencesStore = Record<string, DashboardPreferences>;
+
+const subscribers = new Set<() => void>();
```
AshwinBhatkal: We can Zustand persist instead of this. that should solve this

AshwinBhatkal (changes requested):

frontend/src/container/DashboardContainer/visualization/panels/TimeSeriesPanel/TimeSeriesPanel.tsx
ahrefabhi: Discussed over call,

AshwinBhatkal (approved):

PR #11175, opened 2026-05-04
feat: added changes for synced tooltip modes

ahrefabhi:
## Pull Request

---

### 📄 Summary

Adds a "Synced Tooltip Series" preference that controls how the receiver tooltip behaves when cross-panel sync mode is `Tooltip`. Two modes:

- **Filtered** (default, prior behavior): the receiver tooltip only renders series whose `groupBy` values intersect the source panel's focused series, and is hidden entirely when nothing matches.
- **All**: the receiver tooltip renders every series, with the matching ones highlighted (active row style) and non-matching ones dimmed — useful when users want full panel context but still want to see what's correlated with the hovered series.

image

#### Issues closed by this PR
N/A

---

### ✅ Change Type

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor

---

#### Root Cause
`useDashboardCursorSyncMode` early-returned for non-`DASHBOARD_VIEW` panel modes *before* calling `useDashboardPreference`. If `panelMode` ever changed across renders, hook order would change and React would throw. It also didn't handle the `undefined` `panelMode` call site (dashboard settings) coherently.

#### Fix Strategy
Call `useDashboardPreference` unconditionally first, then apply the panel-mode short-circuit on the return value. The check is rephrased as `panelMode && panelMode !== PanelMode.DASHBOARD_VIEW` so an undefined `panelMode` falls through to the stored value (matches the dashboard-settings usage).

Separately, switched `DashboardCursorSync` from numeric to string enum values (`'crosshair' | 'none' | 'tooltip'`) so persisted localStorage values remain stable if the enum is reordered.

---

### 🧪 Testing Strategy

- **Tests added:**
  - New: `useSyncTooltipFilterMode.test.ts` — defaults, store read/write, localStorage persistence, undefined-id no-op, cross-dashboard isolation.
  - Updated: `useDashboardCursorSyncMode.test.ts` — added a "without panelMode" describe block covering the dashboard-settings call site (read/write through to the store).
- **Manual verification:**
  - Toggled Sync Mode between None / Crosshair / Tooltip; "Synced Tooltip Series" row appears only when Tooltip is selected.
  - Verified Filtered vs All on multi-panel dashboards with overlapping and non-overlapping `groupBy` keys.
  - Confirmed the preference persists across reloads and is scoped per dashboard.
  - Verified panels in edit / standalone view do not participate in cross-panel sync (existing behavior preserved).
- **Edge cases covered:**
  - No matching receiver series in Filtered mode → tooltip suppressed.
  - No matching receiver series in All mode → tooltip still renders, no row highlighted.
  - Source-side `activeSeriesMetric` unset → All mode still shows all series; Filtered hides.
  - Y-axis-unit mismatch between source and receiver continues to hide the y-crosshair (unchanged).

---

- **Blast radius:** Only kicks in when `cursorSyncMode === Tooltip`. Crosshair / None paths are unchanged. Default value (`Filtered`) preserves the prior tooltip behavior, so users who never touch the new setting see no change.
- **Potential regressions:**
  - `DashboardCursorSync` is now a string enum — any code that compared against numeric values would break. Audited usages and updated `useDashboardPreference` typing accordingly.
  - `BarPanel` / `TimeSeriesPanel` now key the chart on `${syncMode}-${syncFilterMode}` so the chart remounts when filter mode flips (needed because the sync metadata is captured at mount via the registry).
- **Rollback plan:** Revert the single commit on this branch — the parent `feat/dashboard-preferences` branch is unaffected.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Dashboards now expose a "Synced Tooltip Series" preference (Filtered / All) when cross-panel Tooltip sync is enabled, controlling whether the receiver tooltip shows only matching series or every series with matches highlighted. |

---

### 📋 Checklist
- [x] Manually tested

---


frontend/src/hooks/dashboard/useDashboardCursorSyncMode.ts
```diff
@@ -9,14 +9,18 @@ export function useDashboardCursorSyncMode(
 	dashboardId: string | undefined,
 	panelMode?: PanelMode,
 ): [DashboardCursorSync, (value: DashboardCursorSync) => void] {
-	if (panelMode !== PanelMode.DASHBOARD_VIEW) {
-		// Only allow cursor sync modes to be set in the dashboard view mode, as the other modes are used for editing and standalone panel views where cursor sync is not applicable.
-		return [DEFAULT_CURSOR_SYNC_MODE, () => {}];
-	}
-
-	return useDashboardPreference(
+	const [value, setValue] = useDashboardPreference(
```
AshwinBhatkal, line 14: nit: destructuring is not required here

frontend/src/container/DashboardContainer/DashboardSettings/General/index.tsx
```diff
@@ -162,27 +169,56 @@ function GeneralDashboardSettings(): JSX.Element {
...
+						<Radio.Group
+							value={syncTooltipFilterMode}
+							onChange={(e): void => {
+								setSyncTooltipFilterMode(e.target.value as SyncTooltipFilterMode);
+							}}
+						>
+							<Radio.Button value={SyncTooltipFilterMode.Filtered}>
+								Filtered
+							</Radio.Button>
+							<Radio.Button value={SyncTooltipFilterMode.All}>All</Radio.Button>
```
AshwinBhatkal: Move all before filtered. All should be default. I see right now filtered is across the other files

Also, when sending it for docs, the intersection should be on only one series intersect. We should specifically separate this out

AshwinBhatkal (commented):
Few comments

frontend/src/container/DashboardContainer/DashboardSettings/General/index.tsx
ahrefabhi: Keep filtered as default but move All as the first option, add event for the switch

AshwinBhatkal (approved):

PR #11176, opened 2026-05-04
fix(authdomain): nest config response, rename Updateable→Updatable, return Identifiable on create

therealpandey:
## Summary
Three small API-shape corrections on auth_domain.

- **`GettableAuthDomain` nests `config`.** Previously `AuthDomainConfig` was embedded, which flattened `sso_enabled` / `saml_config` / `oidc_config` / `google_auth_config` / `role_mapping` at the response root and made the response shape diverge from `PostableAuthDomain` (which has all of those under `config`). After this change request and response both carry the same nested object.

- **`UpdateableAuthDomain` → `UpdatableAuthDomain`.** Typo fix; aligns with `UpdatableUser` already in the codebase.

- **`CreateAuthDomain` returns `Identifiable` instead of `GettableAuthDomain`.** The only field clients need from the create response is the new ID; switching tightens the contract and keeps create/read concerns separated. Clients re-Read for the full domain.

`docs/api/openapi.yml` and `frontend/src/api/generated/services/sigNoz.schemas.ts` regenerated.

## Test plan
- [ ] `make test` / unit tests pass
- [ ] Existing auth-domain create / update flows still work end-to-end
- [ ] Frontend auth-domain pages still load (response shape change)


frontend/src/api/generated/services/sigNoz.schemas.ts
```diff
@@ -1641,109 +1641,32 @@ export interface AuthtypesCallbackAuthNSupportDTO {
 	url?: string;
 }
 
-export type AuthtypesGettableAuthDomainDTO =
```
therealpandey, line 1644: @AshwinBhatkal 
Fixed it in this. Not 100% sure how I got here, gimme some more time.

H4ad (approved):

PR #11180, opened 2026-05-04
feat(ruletypes): publish OpenAPI 3 discriminator on RuleThresholdData and EvaluationEnvelope

therealpandey:
## Summary
Both `RuleThresholdData` and `EvaluationEnvelope` model `{kind, spec}` discriminated unions on the wire, but the generated OpenAPI lacked the `discriminator:` keyword. Codegen tools (oapi-codegen, terraform-plugin-codegen-openapi) saw a bare \`oneOf\` and fell back to opaque \`Spec: any\`, forcing consumers to hand-write JSON-bridges instead of using typed Expand/Flatten.

After this PR:

```yaml
RuletypesEvaluationEnvelope:
  discriminator:
    propertyName: kind
    mapping:
      cumulative: '#/components/schemas/RuletypesEvaluationCumulative'
      rolling: '#/components/schemas/RuletypesEvaluationRolling'
  oneOf:
  - \$ref: '#/components/schemas/RuletypesEvaluationRolling'
  - \$ref: '#/components/schemas/RuletypesEvaluationCumulative'
  ...

RuletypesRuleThresholdData:
  discriminator:
    propertyName: kind
    mapping:
      basic: '#/components/schemas/RuletypesThresholdBasic'
  oneOf:
  - \$ref: '#/components/schemas/RuletypesThresholdBasic'
  ...
```

## How
\`jsonschema-go.Schema\` has no Discriminator field of its own (the keyword is OpenAPI-specific, not JSON Schema), and \`openapi-go\` only carries through \`x-\`-prefixed extras unchanged. So the discriminator goes through a marker convention:

1. Each parent type implements \`jsonschema.Preparer\` to set an \`x-signoz-discriminator\` extra property carrying \`propertyName\` and the per-kind \`mapping\`. Preparer runs after \`OneOfExposer\` populates the \`oneOf\` list, so both coexist cleanly.
2. A small \`attachDiscriminators\` pass in \`pkg/signoz/openapi.go\` runs after spec reflection, walks every component schema, promotes the marker into a real \`openapi3.Discriminator\`, and removes the marker so it doesn't leak into the rendered YAML.

The pass is generic — it doesn't know about RuleThresholdData or EvaluationEnvelope by name. Any future type that wants a discriminator just implements \`Preparer\` with the same marker convention.

## Wire shape
Unchanged. \`{kind: "<value>", spec: <variant>}\` is still what's sent and received. Only the published schema metadata changes.

### DIFF
--- a/frontend/src/api/generated/services/sigNoz.schemas.ts
+++ b/frontend/src/api/generated/services/sigNoz.schemas.ts
@@ -6591,14 +6591,8 @@ export interface RuletypesEvaluationCumulativeDTO {
 }
 
 export type RuletypesEvaluationEnvelopeDTO =
-	| (RuletypesEvaluationRollingDTO & {
-			kind: RuletypesEvaluationKindDTO;
-			spec: unknown;
-	  })
-	| (RuletypesEvaluationCumulativeDTO & {
-			kind: RuletypesEvaluationKindDTO;
-			spec: unknown;
-	  });
+	| RuletypesEvaluationRollingDTO
+	| RuletypesEvaluationCumulativeDTO;
 
 export enum RuletypesEvaluationKindDTO {
 	rolling = 'rolling',
@@ -6968,10 +6962,7 @@ export interface RuletypesRuleConditionDTO {
 	thresholds?: RuletypesRuleThresholdDataDTO;
 }
 
-export type RuletypesRuleThresholdDataDTO = RuletypesThresholdBasicDTO & {
-	kind: RuletypesThresholdKindDTO;
-	spec: unknown;
-};
+export type RuletypesRuleThresholdDataDTO = RuletypesThresholdBasicDTO;
 
 export enum RuletypesRuleTypeDTO {
 	threshold_rule = 'threshold_rule',
--- a/frontend/src/api/generated/services/sigNoz.schemas.ts
+++ b/frontend/src/api/generated/services/sigNoz.schemas.ts
@@ -6578,8 +6578,15 @@ export interface RuletypesCumulativeWindowDTO {
 	timezone: string;
 }
 
+export enum RuletypesEvaluationCumulativeDTOKind {
+	cumulative = 'cumulative',
+}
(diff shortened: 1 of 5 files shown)


frontend/src/api/generated/services/sigNoz.schemas.ts
```diff
@@ -6578,28 +6578,36 @@ export interface RuletypesCumulativeWindowDTO {
...
 export type RuletypesEvaluationEnvelopeDTO =
-	| (RuletypesEvaluationRollingDTO & {
-			kind: RuletypesEvaluationKindDTO;
-			spec: unknown;
-	  })
-	| (RuletypesEvaluationCumulativeDTO & {
-			kind: RuletypesEvaluationKindDTO;
-			spec: unknown;
-	  });
+	| RuletypesEvaluationRollingDTO
```
therealpandey, line 6594: And victory @AshwinBhatkal

pkg/types/ruletypes/threshold.go
```diff
@@ -50,6 +53,24 @@ func (RuleThresholdData) JSONSchemaOneOf() []any {
 	}
 }
 
+// PrepareJSONSchema marks the schema with x-signoz-discriminator;
+// signoz.attachDiscriminators promotes it to a real OpenAPI 3
+// discriminator after reflection.
+func (RuleThresholdData) PrepareJSONSchema(schema *jsonschema.Schema) error {
```
AshwinBhatkal, line 59: Do we have any core maintainers who will ensure and not forget this is done for all such cases?

AshwinBhatkal (approved):

pkg/types/ruletypes/threshold.go
therealpandey: I have a couple of other places where I need to push this change. After doing changes there, I will create an abstraction over the existing JSONSchema which will enforce this.

srikanthccv (approved):

PR #10679, opened 2026-03-23
feat(qb): added basic validation in clickhouse queries

tushar-signoz:
## Pull Request
Validate ClickHouse SQL queries and map CH errors to invalid input.
---

### 📄 Summary
Improves ClickHouse SQL queries in the query builder by validating specs before execution, returning client errors (400) for bad SQL and disallowed table references where appropriate, and adds integration coverage plus a small fixture fix for scalar responses.

### Changes
Query builder validation (ClickHouseQuery.Validate)
Requires non-empty name and query.
Rejects SQL that references deprecated tables (e.g. traces V2 / logs V1) or local tables when distributed tables should be used, using word-boundary regexes (case-insensitive) to limit false positives.
validateQueryEnvelope now delegates ClickHouse SQL spec checks to spec.Validate() instead of only checking an empty query string.
Querier: ClickHouse errors → invalid input
On query execution, maps a subset of ClickHouse exception codes (syntax, unknown table/database/identifier/function/type, illegal arguments/columns, etc.) to InvalidInput so callers get a user-facing error class instead of treating everything as an internal failure.
Removes the old TODO about mapping ClickHouse errors.

### Tests & fixtures
Unit tests for ClickHouseQuery validation (including deprecated/local table cases and case-insensitivity)
Integration (tests/integration/src/querier/12_clickhouse_sql.py): happy paths for scalar and raw ClickHouse SQL against distributed_logs_v2; negative cases for deprecated table, local table, and unknown table (exercising validation vs. mapClickHouseError).
get_scalar_value in tests/integration/fixtures/querier.py updated to read scalar results from the actual data[0][0] shape instead of going through series helpers.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|
| Deployment Type | Cloud / OSS / Enterprise |
| Change Type | Feature / Bug Fix / Maintenance |
| Description | User-facing summary |

---

### 📋 Checklist
- [ ] Tests added or explicitly not required
- [ ] Manually tested
- [ ] Breaking changes documented
- [ ] Backward compatibility considered

---

## 👀 Notes for Reviewers

---

PR #11185, opened 2026-05-05
feat: QB support for QueryRangePreview

tushar-signoz:
## Summary

Adds a **dry-run preview** endpoint for v5 composite queries:

```
POST /api/v5/query_range/preview
```

- Accepts the **exact same payload** as `POST /api/v5/query_range`, but **validates and renders** each query's underlying ClickHouse/PromQL statement(s) **without executing** them.
- **Per-query verdict, not fail-fast:** request-level failures (time range, request type, duplicate names, malformed body) abort with a top-level `status:"error"`; per-query structural/semantic errors are attributed to the query under `data.compositeQuery.<name>.error` with `status:"success"` — so one round trip surfaces *every* problem.
- **OTel-aligned field names:** rendered SQL/args use `db.statement.query` / `db.statement.args`.
- Built for agentic / programmatic query construction — check a query before paying for a full scan.

### Query parameter

| Param | Default | Effect |
|-------|---------|--------|
| `verbose` | `true` | Render each statement with its `EXPLAIN ESTIMATE` (per-table `parts`/`rows`/`marks`) and granule index analysis (candidate vs. surviving granules + the per-index pruning funnel). `verbose=false` → lightweight verdict only (valid/error/warnings), **zero ClickHouse round trips**. |

### Response shape

```
data.compositeQuery.<name> = {
  valid:       bool                                                  // true ⇔ error absent
  error?:      { type, code, message, errors?[], suggestions?[] }    // only when valid=false
  warnings?:   string[]                                              // non-fatal notes (step adjusted, missing metric, absent groupBy key, …)
  statements?: [ {                                                   // when verbose (default)
    "db.statement.query": string,
    "db.statement.args"?: any[],
    estimate?: [ { database, table, parts, rows, marks } ],          // parsed EXPLAIN ESTIMATE
    granules?: { initial, selected, skipped,                         // granule-skip summary
                 reads?: [ { table, steps: [ { type, name?, keys?, condition?,
                              initialParts, selectedParts, initialGranules, selectedGranules } ] } ] }
  } ]
}
```

- **Raw counts, no derived scores.** The preview reports granule and row numbers directly — selectivity is `skipped/initial`, magnitude is `estimate.rows` — and does **not** collapse them into a 0–100 score.

### How each query type renders to statement(s)

- **Builder / ClickHouse SQL / trace-operator** → one statement, via a `Statement(ctx)` method on each query type (the exported `qbtypes.StatementProvider`).
- **PromQL** → one statement per metric selector, captured by driving the Prometheus engine with a capturing storage (`StatementCapturer`) that records each selector's SQL and returns empty results — no ClickHouse reads.
- **Raw `clickhouse_sql`** → bind-checked via `EXPLAIN PLAN` (parses + tables/columns/types resolve); a user-facing CH error code → invalid-input (400), anything else downgraded to a warning (infra, not a bad query).
- **Trace operator (`C := A => B`)** → `query_range` never runs `A`/`B` standalone, but the preview renders each on its own (as a `RequestTypeRaw` span selection) so the caller sees *which* sub-query failed, not just an opaque `C`.

### Index-pruning funnel (`granules.reads[]`)

- One entry per `ReadFromMergeTree` node, listing index steps (MinMax / Partition / PrimaryKey / Skip) in execution order with the parts/granules entering vs. surviving each — the diagnostic for *which* index pruned and which did nothing (a `Skip` step where `selected == initial` isn't engaging).

---

### ✅ Change Type

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🧪 Testing Strategy

- **Manual verification** against a live local enterprise server (ClickHouse on `localhost:9001`); JSON responses captured in `query-range-preview-api-responses.md`. Cases covered:
  - **Params:** default (verbose), `verbose=false` (verdict only), invalid `?verbose=` value.
  - **Per-query errors** under `data.compositeQuery.<name>`: field-not-found (+ "did you mean" suggestion), invalid `having` reference, metric-not-found (warning, not error), **mixed valid + invalid in one response**.
  - **Trace operator:** all-sub-queries-valid (A/B/C each rendered) and **bad sub-query pinpointed** (B invalid → C invalid, A still valid).
  - **PromQL** (one statement per selector), **metric builder with `groupBy`** (multi-table `estimate`), raw **`clickhouse_sql`** valid + invalid (bind-check failure).
  - **Granule funnel:** full MinMax → Partition → PrimaryKey trace + missing-index diagnosis (`body LIKE '%…%'` → no `Skip` step → unindexed substring search).
  - **Top-level aborts:** duplicate query name, invalid enum at decode time, malformed JSON body.
- Preview reuses the same statement builders and Prometheus engine as the execution path, so the rendered SQL is the SQL that would execute. Automated per-query verdict matrix is a follow-up.

---

- **Blast radius:** purely additive — new route, new response/options types, new `Statement()` methods, optional `StatementCapturer`. No change to the existing `query_range` execution path. Gated behind `ViewAccess` (viewer role).
- **Shared-code touches:** `buildSamplesQuery` extracted from `querySamples` (pure refactor — same SQL/args) and a `make([]string, 0, cap)` fix in `where_clause_visitor.go`.
- **ClickHouse load:** `EXPLAIN` round trips only on the default `verbose=true` path; `verbose=false` touches ClickHouse zero times.
- **Rollback plan:** revert the PR — no migrations, no schema changes, no persisted state.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | New `POST /api/v5/query_range/preview` endpoint that validates a v5 composite query and renders its underlying ClickHouse/PromQL statement(s) without executing it — returning a per-query verdict (valid / error / warnings) and, by default, each statement's `EXPLAIN ESTIMATE` and granule index-pruning funnel. `verbose=false` returns the verdict only. Intended for agentic / dry-run query construction. |

---

### 📋 Checklist
- [x] Tests added or explicitly not required _(manually verified; automated per-query matrix is a follow-up)_
- [x] Manually tested
- [x] Breaking changes documented _(none — additive)_
- [x] Backward compatibility considered _(`query_range` untouched)_

---

## 👀 Notes for Reviewers

- **Param surface simplified to a single `verbose` toggle** (default `true`): both EXPLAIN analyses run together or not at all. Earlier revisions had separate tri-state `estimate`/`granules` params and two derived scores (`selectivityScore` / `magnitudeScore`) — **both removed**; the preview now returns the raw `estimate` / `granules` counts only.
- **`userFacingClickHouseErrorCodes`** in `preview.go` is a temporary copy of #10679's `userFacingCHCodes`; `TODO(#10679)` to delete it and call the shared `mapClickHouseError` once that lands.
- **Validation split:** `QueryRangeRequest.ValidateRequestScope()` checks request-level invariants and returns the per-request `ValidationOption`s; `QueryEnvelope.Validate()` validates a single query's spec. The preview validates per query; the execution path keeps the fail-fast `Validate()`.
- **Spec-normalization ordering:** phase 1 normalizes all specs (step interval + metric metadata, including trace-op dependencies); phase 2 renders only the queries that execute (plus trace-op dependencies, as `RequestTypeRaw` span selections) — trace-operator queries read their siblings' specs at render time.

### DIFF
--- a/pkg/querier/preview.go
+++ b/pkg/querier/preview.go
@@ -1,0 +1,360 @@
+package querier
+
+import (
+	"context"
+	"log/slog"
+	"strings"
+	"time"
+
+	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/querybuilder"
+	"github.com/SigNoz/signoz/pkg/types/metrictypes"
+	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
+	"github.com/SigNoz/signoz/pkg/types/telemetrytypes"
+	"github.com/SigNoz/signoz/pkg/valuer"
+)
+
+// statementProvider is implemented by query types that can render the
+// underlying SQL/PromQL statement without executing it.
+type statementProvider interface {
+	Statement(ctx context.Context) (*qbtypes.Statement, error)
+}
+
+// clickhouseExplainClause maps a variant to the EXPLAIN clause understood
+// by ClickHouse (i.e. what comes between EXPLAIN and the SELECT).
+func clickhouseExplainClause(v qbtypes.ExplainVariant) (string, bool) {
+	switch v {
+	case qbtypes.ExplainVariantPlan:
+		return "PLAN", true
+	case qbtypes.ExplainVariantAST:
+		return "AST", true
+	case qbtypes.ExplainVariantSyntax:
+		return "SYNTAX", true
+	case qbtypes.ExplainVariantPipeline:
+		return "PIPELINE", true
+	case qbtypes.ExplainVariantEstimate:
+		return "ESTIMATE", true
+	case qbtypes.ExplainVariantQueryTree:
+		return "QUERY TREE", true
+	default:
+		return "", false
+	}
+}
+
+// ParseExplainVariant parses the ?explain= query parameter. An empty value
+// (or "false") returns ExplainVariantNone. The literal "true" maps to PLAN
+// for back-compat with simple ?explain=true. Otherwise the value must
+// match one of the named variants.
+func ParseExplainVariant(value string) (qbtypes.ExplainVariant, error) {
+	token := strings.ToLower(strings.TrimSpace(value))
+	switch token {
+	case "", "false":
+		return qbtypes.ExplainVariantNone, nil
+	case "true":
+		return qbtypes.ExplainVariantPlan, nil
+	}
+	v := qbtypes.ExplainVariant(token)
+	if _, ok := clickhouseExplainClause(v); !ok {
+		return qbtypes.ExplainVariantNone, errors.NewInvalidInputf(errors.CodeInvalidInput, "unsupported explain variant %q (allowed: plan, ast, syntax, pipeline, estimate, query_tree)", token)
+	}
+	return v, nil
+}
+
+// QueryRangePreview validates the request and renders the underlying SQL
+// (or PromQL) for each query in the composite query without executing it.
+// When opts.Explain is non-empty, EXPLAIN <variant> is run against the
+// telemetry store for each rendered SQL statement and attached to the
+// response.
+func (q *querier) QueryRangePreview(
+	ctx context.Context,
+	_ valuer.UUID,
+	req *qbtypes.QueryRangeRequest,
+	opts qbtypes.QueryRangePreviewOptions,
+) (*qbtypes.QueryRangePreviewResponse, error) {
+	tmplVars := req.Variables
+	if tmplVars == nil {
+		tmplVars = make(map[string]qbtypes.VariableItem)
+	}
+
+	dependencyQueries := make(map[string]bool)
+	for _, query := range req.CompositeQuery.Queries {
+		if query.Type == qbtypes.QueryTypeTraceOperator {
+			if spec, ok := query.Spec.(qbtypes.QueryBuilderTraceOperator); ok {
+				if err := spec.ParseExpression(); err != nil {
+					return nil, err
+				}
+				for _, dep := range spec.CollectReferencedQueries(spec.ParsedExpression) {
+					dependencyQueries[dep] = true
+				}
+			}
+		}
+	}
+
+	// First pass: normalize step intervals and collect metric names that
+	// need temporality/type lookup.
+	metricNames := make([]string, 0)
+	for idx, query := range req.CompositeQuery.Queries {
+		switch query.Type {
+		case qbtypes.QueryTypeBuilder:
+			if spec, ok := query.Spec.(qbtypes.QueryBuilderQuery[qbtypes.MetricAggregation]); ok {
+				for _, agg := range spec.Aggregations {
+					if agg.MetricName != "" {
+						metricNames = append(metricNames, agg.MetricName)
+					}
+				}
+			}
+			switch spec := query.Spec.(type) {
+			case qbtypes.QueryBuilderQuery[qbtypes.TraceAggregation]:
+				if spec.StepInterval.Seconds() == 0 {
+					spec.StepInterval = qbtypes.Step{Duration: time.Second * time.Duration(querybuilder.RecommendedStepInterval(req.Start, req.End))}
+				}
+				if spec.StepInterval.Seconds() < float64(querybuilder.MinAllowedStepInterval(req.Start, req.End)) {
+					spec.StepInterval = qbtypes.Step{Duration: time.Second * time.Duration(querybuilder.MinAllowedStepInterval(req.Start, req.End))}
+				}
+				req.CompositeQuery.Queries[idx].Spec = spec
+			case qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]:
+				if spec.StepInterval.Seconds() == 0 {
+					spec.StepInterval = qbtypes.Step{Duration: time.Second * time.Duration(querybuilder.RecommendedStepInterval(req.Start, req.End))}
+				}
+				if spec.StepInterval.Seconds() < float64(querybuilder.MinAllowedStepInterval(req.Start, req.End)) {
+					spec.StepInterval = qbtypes.Step{Duration: time.Second * time.Duration(querybuilder.MinAllowedStepInterval(req.Start, req.End))}
+				}
+				req.CompositeQuery.Queries[idx].Spec = spec
+			case qbtypes.QueryBuilderQuery[qbtypes.MetricAggregation]:
+				if spec.Source == telemetrytypes.SourceMeter {
+					if spec.StepInterval.Seconds() == 0 {
+						spec.StepInterval = qbtypes.Step{Duration: time.Second * time.Duration(querybuilder.RecommendedStepIntervalForMeter(req.Start, req.End))}
+					}
+					if spec.StepInterval.Seconds() < float64(querybuilder.MinAllowedStepIntervalForMeter(req.Start, req.End)) {
+						spec.StepInterval = qbtypes.Step{Duration: time.Second * time.Duration(querybuilder.MinAllowedStepIntervalForMeter(req.Start, req.End))}
+					}
+				} else {
+					if spec.StepInterval.Seconds() == 0 {
+						spec.StepInterval = qbtypes.Step{Duration: time.Second * time.Duration(querybuilder.RecommendedStepIntervalForMetric(req.Start, req.End))}
+					}
+					if spec.StepInterval.Seconds() < float64(querybuilder.MinAllowedStepIntervalForMetric(req.Start, req.End)) {
+						spec.StepInterval = qbtypes.Step{Duration: time.Second * time.Duration(querybuilder.MinAllowedStepIntervalForMetric(req.Start, req.End))}
+					}
+				}
+				req.CompositeQuery.Queries[idx].Spec = spec
+			}
+		case qbtypes.QueryTypePromQL:
+			if spec, ok := query.Spec.(qbtypes.PromQuery); ok {
+				if spec.Step.Seconds() == 0 {
+					spec.Step = qbtypes.Step{Duration: time.Second * time.Duration(querybuilder.RecommendedStepIntervalForMetric(req.Start, req.End))}
+				}
+				req.CompositeQuery.Queries[idx].Spec = spec
+			}
+		case qbtypes.QueryTypeTraceOperator:
+			if spec, ok := query.Spec.(qbtypes.QueryBuilderTraceOperator); ok {
+				if spec.StepInterval.Seconds() == 0 {
+					spec.StepInterval = qbtypes.Step{Duration: time.Second * time.Duration(querybuilder.RecommendedStepInterval(req.Start, req.End))}
+				}
+				if spec.StepInterval.Seconds() < float64(querybuilder.MinAllowedStepInterval(req.Start, req.End)) {
+					spec.StepInterval = qbtypes.Step{Duration: time.Second * time.Duration(querybuilder.MinAllowedStepInterval(req.Start, req.End))}
+				}
+				req.CompositeQuery.Queries[idx].Spec = spec
+			}
+		}
+	}
+
+	// Fetch metric temporality/type for any builder metric queries.
+	var metricTemporality map[string]metrictypes.Temporality
+	var metricTypes map[string]metrictypes.Type
+	if len(metricNames) > 0 {
+		var err error
+		metricTemporality, metricTypes, err = q.metadataStore.FetchTemporalityAndTypeMulti(ctx, req.Start, req.End, metricNames...)
+		if err != nil {
+			q.logger.WarnContext(ctx, "failed to fetch metric temporality during preview", errors.Attr(err), slog.Any("metrics", metricNames))
+			return nil, errors.NewInternalf(errors.CodeInternal, "failed to fetch metrics temporality")
+		}
+	}
+
+	type queryEntry struct {
+		name      string
+		queryType qbtypes.QueryType
+		query     qbtypes.Query
+	}
+	entries := make([]queryEntry, 0, len(req.CompositeQuery.Queries))
+
+	for _, query := range req.CompositeQuery.Queries {
+		var queryName string
+		isTraceOperator := query.Type == qbtypes.QueryTypeTraceOperator
+
+		switch query.Type {
+		case qbtypes.QueryTypeTraceOperator:
+			if spec, ok := query.Spec.(qbtypes.QueryBuilderTraceOperator); ok {
+				queryName = spec.Name
+			}
+		case qbtypes.QueryTypePromQL:
+			if spec, ok := query.Spec.(qbtypes.PromQuery); ok {
+				queryName = spec.Name
+			}
+		case qbtypes.QueryTypeClickHouseSQL:
+			if spec, ok := query.Spec.(qbtypes.ClickHouseQuery); ok {
+				queryName = spec.Name
+			}
+		case qbtypes.QueryTypeBuilder:
+			switch spec := query.Spec.(type) {
+			case qbtypes.QueryBuilderQuery[qbtypes.TraceAggregation]:
+				queryName = spec.Name
+			case qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]:
+				queryName = spec.Name
+			case qbtypes.QueryBuilderQuery[qbtypes.MetricAggregation]:
+				queryName = spec.Name
+			}
+		}
+
+		if !isTraceOperator && dependencyQueries[queryName] {
+			continue
+		}
+
+		switch query.Type {
+		case qbtypes.QueryTypePromQL:
+			promQuery, ok := query.Spec.(qbtypes.PromQuery)
+			if !ok {
+				return nil, errors.NewInvalidInputf(errors.CodeInvalidInput, "invalid promql query spec %T", query.Spec)
+			}
+			pq := newPromqlQuery(q.logger, q.promEngine, promQuery, qbtypes.TimeRange{From: req.Start, To: req.End}, req.RequestType, tmplVars)
+			entries = append(entries, queryEntry{name: promQuery.Name, queryType: query.Type, query: pq})
+		case qbtypes.QueryTypeClickHouseSQL:
+			chQuery, ok := query.Spec.(qbtypes.ClickHouseQuery)
+			if !ok {
+				return nil, errors.NewInvalidInputf(errors.CodeInvalidInput, "invalid clickhouse query spec %T", query.Spec)
+			}
+			cq := newchSQLQuery(q.logger, q.telemetryStore, chQuery, nil, qbtypes.TimeRange{From: req.Start, To: req.End}, req.RequestType, tmplVars)
+			entries = append(entries, queryEntry{name: chQuery.Name, queryType: query.Type, query: cq})
+		case qbtypes.QueryTypeTraceOperator:
+			traceOpQuery, ok := query.Spec.(qbtypes.QueryBuilderTraceOperator)
+			if !ok {
+				return nil, errors.NewInvalidInputf(errors.CodeInvalidInput, "invalid trace operator query spec %T", query.Spec)
+			}
+			toq := &traceOperatorQuery{
+				telemetryStore: q.telemetryStore,
+				stmtBuilder:    q.traceOperatorStmtBuilder,
+				spec:           traceOpQuery,
+				compositeQuery: &req.CompositeQuery,
+				fromMS:         uint64(req.Start),
+				toMS:           uint64(req.End),
+				kind:           req.RequestType,
+			}
+			entries = append(entries, queryEntry{name: traceOpQuery.Name, queryType: query.Type, query: toq})
+		case qbtypes.QueryTypeBuilder:
+			switch spec := query.Spec.(type) {
+			case qbtypes.QueryBuilderQuery[qbtypes.TraceAggregation]:
+				spec.ShiftBy = extractShiftFromBuilderQuery(spec)
+				timeRange := adjustTimeRangeForShift(spec, qbtypes.TimeRange{From: req.Start, To: req.End}, req.RequestType)
+				bq := newBuilderQuery(q.logger, q.telemetryStore, q.traceStmtBuilder, spec, timeRange, req.RequestType, tmplVars)
+				entries = append(entries, queryEntry{name: spec.Name, queryType: query.Type, query: bq})
+			case qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]:
+				spec.ShiftBy = extractShiftFromBuilderQuery(spec)
+				timeRange := adjustTimeRangeForShift(spec, qbtypes.TimeRange{From: req.Start, To: req.End}, req.RequestType)
+				stmtBuilder := q.logStmtBuilder
+				if spec.Source == telemetrytypes.SourceAudit {
+					stmtBuilder = q.auditStmtBuilder
+				}
+				bq := newBuilderQuery(q.logger, q.telemetryStore, stmtBuilder, spec, timeRange, req.RequestType, tmplVars)
+				entries = append(entries, queryEntry{name: spec.Name, queryType: query.Type, query: bq})
+			case qbtypes.QueryBuilderQuery[qbtypes.MetricAggregation]:
+				presentAggregations := []qbtypes.MetricAggregation{}
+				for i := range spec.Aggregations {
+					if spec.Aggregations[i].MetricName != "" && spec.Aggregations[i].Temporality == metrictypes.Unknown {
+						if temp, ok := metricTemporality[spec.Aggregations[i].MetricName]; ok && temp != metrictypes.Unknown {
+							spec.Aggregations[i].Temporality = temp
+						}
+					}
+					if spec.Aggregations[i].MetricName != "" && spec.Aggregations[i].Type == metrictypes.UnspecifiedType {
+						if foundMetricType, ok := metricTypes[spec.Aggregations[i].MetricName]; ok && foundMetricType != metrictypes.UnspecifiedType {
+							spec.Aggregations[i].Type = foundMetricType
+						}
+					}
+					if spec.Aggregations[i].Type == metrictypes.UnspecifiedType {
+						continue
+					}
+					presentAggregations = append(presentAggregations, spec.Aggregations[i])
+				}
+				if len(presentAggregations) == 0 {
+					// nothing renderable for this query — skip
+					continue
+				}
+				spec.Aggregations = presentAggregations
+				spec.ShiftBy = extractShiftFromBuilderQuery(spec)
+				timeRange := adjustTimeRangeForShift(spec, qbtypes.TimeRange{From: req.Start, To: req.End}, req.RequestType)
+				var bq *builderQuery[qbtypes.MetricAggregation]
+				if spec.Source == telemetrytypes.SourceMeter {
+					bq = newBuilderQuery(q.logger, q.telemetryStore, q.meterStmtBuilder, spec, timeRange, req.RequestType, tmplVars)
+				} else {
+					bq = newBuilderQuery(q.logger, q.telemetryStore, q.metricStmtBuilder, spec, timeRange, req.RequestType, tmplVars)
+				}
+				entries = append(entries, queryEntry{name: spec.Name, queryType: query.Type, query: bq})
+			default:
+				return nil, errors.NewInvalidInputf(errors.CodeInvalidInput, "unsupported builder spec type %T", query.Spec)
+			}
+		}
+	}
+
+	statements := make([]*qbtypes.PreviewStatement, 0, len(entries))
+	for _, entry := range entries {
+		provider, ok := entry.query.(statementProvider)
+		if !ok {
+			return nil, errors.NewInternalf(errors.CodeInternal, "query %s does not support preview", entry.name)
+		}
+		stmt, err := provider.Statement(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		ps := &qbtypes.PreviewStatement{
+			QueryName: entry.name,
+			QueryType: entry.queryType.StringValue(),
+			Query:     stmt.Query,
+			Args:      stmt.Args,
+			Warnings:  stmt.Warnings,
+		}
+
+		if opts.Explain != qbtypes.ExplainVariantNone && entry.queryType != qbtypes.QueryTypePromQL {
+			out, err := q.runExplain(ctx, opts.Explain, stmt.Query, stmt.Args)
+			if err != nil {
+				return nil, err
+			}
+			ps.ExplainVariant = string(opts.Explain)
+			ps.Explain = out
+		}
+
+		statements = append(statements, ps)
+	}
+
+	return &qbtypes.QueryRangePreviewResponse{
+		Type:       req.RequestType,
+		Statements: statements,
+	}, nil
+}
+
+// runExplain runs `EXPLAIN <variant> <stmt>` against the telemetry store
+// and returns the formatted output as a single string with one row per
+// line.
+func (q *querier) runExplain(ctx context.Context, variant qbtypes.ExplainVariant, stmt string, args []any) (string, error) {
+	clause, ok := clickhouseExplainClause(variant)
+	if !ok {
+		return "", errors.NewInvalidInputf(errors.CodeInvalidInput, "unsupported explain variant %q", string(variant))
+	}
+	explainQuery := "EXPLAIN " + clause + " " + stmt
+	rows, err := q.telemetryStore.ClickhouseDB().Query(ctx, explainQuery, args...)
+	if err != nil {
+		return "", errors.WrapInternalf(err, errors.CodeInternal, "failed to run EXPLAIN %s", clause)
+	}
+	defer rows.Close()
+
+	var lines []string
+	for rows.Next() {
+		var line string
+		if err := rows.Scan(&line); err != nil {
+			return "", errors.WrapInternalf(err, errors.CodeInternal, "failed to scan EXPLAIN row")
+		}
+		lines = append(lines, line)
+	}
+	if err := rows.Err(); err != nil {
+		return "", errors.WrapInternalf(err, errors.CodeInternal, "EXPLAIN row iteration failed")
+	}
+	return strings.Join(lines, "\n"), nil
+}
--- a/pkg/querier/preview.go
+++ b/pkg/querier/preview.go
@@ -8,8 +8,8 @@ import (
 	"sync"
 
 	"github.com/SigNoz/signoz/pkg/errors"
-	"github.com/SigNoz/signoz/pkg/telemetrystore"
 	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
+	"github.com/SigNoz/signoz/pkg/types/telemetrystoretypes"
 	"github.com/SigNoz/signoz/pkg/valuer"
 )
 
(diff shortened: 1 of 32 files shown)


tushar-signoz:
# Query range preview API — example responses

`POST /api/v5/query_range/preview`

The preview endpoint accepts the **same payload** as `POST /api/v5/query_range`, but instead of
executing each query it validates it and renders the underlying ClickHouse/PromQL statement(s) —
all **without executing** the query (a dry run). Its purpose is a per-query verdict: every query in
the composite query is checked independently and its result reported under its own name, so one
round trip surfaces *every* problem rather than failing fast on the first.

The per-query results are returned under `data.compositeQuery`, keyed by the same query name the
request used in `compositeQuery.queries[]` — the response mirrors the request. Each rendered
statement reports its SQL and bound args under the OpenTelemetry-style `db.statement.query` /
`db.statement.args` keys, so an agent sees the same field names it would on a database span.

Responses below were captured against a live local enterprise server (`SIGNOZ-API-KEY` header,
ClickHouse on `localhost:9001`). Successful responses are wrapped as
`{"status":"success","data": <object>}`; errors as `{"status":"error","error": <object>}`.

## Query parameters

| Param | Default | Effect |
|-------|---------|--------|
| `verbose` | `true` | The **default** is the full preview: each rendered statement carries its `EXPLAIN ESTIMATE` (per-table `parts`/`rows`/`marks`) and granule index analysis (candidate vs. surviving granules and the per-index pruning funnel). Pass `verbose=false` for the lightweight per-query verdict (valid/error/warnings) with no rendered SQL and **no ClickHouse round trips**. |

The two EXPLAIN analyses each cost one ClickHouse `EXPLAIN` round trip per statement, so they run
only on the default (`verbose=true`) path; `verbose=false` touches ClickHouse zero times.

## Response shape

```
data.compositeQuery.<name> = {
  valid:       bool          // headline verdict; true ⇔ error absent
  error?:      { type, code, message, errors?[], suggestions?[] }   // structured; present only when valid=false
  warnings?:   string[]      // non-fatal notes (step adjusted, dormant/missing metric, groupBy key absent, …)
  statements?: [ {                 // when verbose (the default)
    "db.statement.query": string,
    "db.statement.args"?: any[],
    estimate?: [ { database, table, parts, rows, marks } ],   // parsed EXPLAIN ESTIMATE
    granules?: {              // parsed granule-skip breakdown (when the statement reads a MergeTree table)
      initial:   number,      // candidate granules before pruning
      selected:  number,      // granules surviving pruning (what would be read)
      skipped:   number,      // initial - selected
      reads:    [ {           // raw per-read index funnel — which index pruned what (see §2)
        table: string,        // e.g. signoz_logs.logs_v2
        steps: [ {            // index steps in execution order; each step's initial* = prev step's selected*
          type: string,       // MinMax | Partition | PrimaryKey | Skip
          name?: string,      // skip-index name
          keys?: string[], condition?: string,
          initialParts, selectedParts, initialGranules, selectedGranules: number
        } ]
      } ]
    }
  } ]
}
```

**Raw counts, no derived scores.** The preview reports the granule and row numbers directly — it
does *not* fold them into a 0–100 score. You read the two cost axes off the raw output yourself, and
they stay separate so you can see which one is the problem:
- **Selectivity** — how much the indexes prune. Compare `granules.skipped` to `granules.initial`
  (e.g. 1850 / 1860 ≈ 99.5% of granules skipped). The per-index `reads[]` funnel (§2) shows *which*
  index did the pruning.
- **Magnitude** — the absolute work. `estimate[].rows` is the rows ClickHouse expects to scan. A
  query can prune ~99% of granules (high selectivity) yet still scan tens of thousands of rows.

Each statement's `granules` also carries `reads[]` — the raw per-index pruning trace. It is shown in
full only in §2 (where it's the point); the other examples elide it for brevity, keeping just the
`initial`/`selected`/`skipped` summary.

**Two error tiers.** Request-level / decode-time failures (malformed JSON, bad enum/signal,
duplicate query names, invalid query param) abort the whole request and return a **top-level**
`{"status":"error", …}`. Per-query structural/semantic failures (field-not-found, bad `having`
reference, unknown table in raw SQL) are reported **per query** under
`data.compositeQuery.<name>.error` with `"status":"success"` — that is the whole point of the
preview.

---

## 1. Default (`verbose=true`) — full preview

With no params, every query is validated and rendered, and each statement carries its `estimate`
and `granules`. Here the indexes prune 1850 of 1860 granules (~99.5% selectivity) while `estimate`
shows the surviving work is still ~26k rows to scan. (`granules.reads[]` elided — see §2.)

```json
{
  "status": "success",
  "data": {
    "compositeQuery": {
      "A": {
        "valid": true,
        "statements": [
          {
            "db.statement.query": "SELECT count() AS __result_0 FROM signoz_logs.distributed_logs_v2 WHERE timestamp >= ? AND ts_bucket_start >= ? AND timestamp < ? AND ts_bucket_start <= ? ORDER BY __result_0 DESC",
            "db.statement.args": ["1782364504000000000", 1782362704, "1782368104000000000", 1782368104],
            "estimate": [
              { "database": "signoz_logs", "table": "logs_v2", "parts": 2, "rows": 25979, "marks": 8 }
            ],
            "granules": { "initial": 1860, "selected": 10, "skipped": 1850 }
          }
        ]
      }
    }
  }
}
```

## 2. The granule index funnel (`granules.reads[]`)

Each statement's `granules` carries the **raw per-index pruning trace**: one entry per
`ReadFromMergeTree` node, its `steps[]` the indexes ClickHouse applies in order, each with the
parts/granules entering (`initial*`) and surviving (`selected*`) it. Reading the funnel tells you
which index did the pruning and which did nothing.

### 2a. The full funnel

The `granules` block from §1, expanded. `MinMax(timestamp)` does the bulk (1860 → 16 granules);
`Partition` prunes nothing here (16 → 16); `PrimaryKey` trims 16 → 10.

```json
"granules": {
  "initial": 1860, "selected": 10, "skipped": 1850,
  "reads": [
    {
      "table": "signoz_logs.logs_v2",
      "steps": [
        { "type": "MinMax", "keys": ["timestamp"], "condition": "and((timestamp in (-Inf, 1782368103999999999]), (timestamp in [1782364504000000000, +Inf)))", "initialParts": 53, "selectedParts": 4, "initialGranules": 1860, "selectedGranules": 16 },
        { "type": "Partition", "keys": ["toDate(divide(timestamp, 1000000000))"], "condition": "and((toDate(divide(timestamp, 1000000000)) in (-Inf, 20629]), (toDate(divide(timestamp, 1000000000)) in [20629, +Inf)))", "initialParts": 4, "selectedParts": 4, "initialGranules": 16, "selectedGranules": 16 },
        { "type": "PrimaryKey", "keys": ["ts_bucket_start", "timestamp"], "condition": "and((ts_bucket_start in (-Inf, 1782368104]), and((timestamp in (-Inf, 1782368103999999999]), and((ts_bucket_start in [1782362704, +Inf)), (timestamp in [1782364504000000000, +Inf)))))", "initialParts": 4, "selectedParts": 4, "initialGranules": 16, "selectedGranules": 10 }
      ]
    }
  ]
}
```

### 2b. Diagnosing a missing index

Add a `body CONTAINS 'error'` filter (rendered `LOWER(body) LIKE LOWER('%error%')`). The funnel
shows **no `Skip` step** — every granule was pruned by `timestamp`/`ts_bucket_start` alone, and the
`body` predicate engaged no index. That's the signal: a substring match on `body` is unindexed (the
token bloom-filter index doesn't serve arbitrary `%…%` substrings), so this is where a skip index
would help. (Conversely, the trace statements in §14 *do* carry `Skip` steps — there `idx_name`
prunes 91 → 9 while `idx_timestamp` prunes nothing, 91 → 91: an index present but doing no work.)

```json
"granules": {
  "initial": 1862, "selected": 12, "skipped": 1850,
  "reads": [
    {
      "table": "signoz_logs.logs_v2",
      "steps": [
        { "type": "MinMax", "keys": ["timestamp"], "condition": "and((timestamp in (-Inf, 1782368103999999999]), (timestamp in [1782364504000000000, +Inf)))", "initialParts": 55, "selectedParts": 6, "initialGranules": 1862, "selectedGranules": 18 },
        { "type": "Partition", "keys": ["toDate(divide(timestamp, 1000000000))"], "condition": "and((toDate(divide(timestamp, 1000000000)) in (-Inf, 20629]), (toDate(divide(timestamp, 1000000000)) in [20629, +Inf)))", "initialParts": 6, "selectedParts": 6, "initialGranules": 18, "selectedGranules": 18 },
        { "type": "PrimaryKey", "keys": ["ts_bucket_start", "timestamp"], "condition": "and((ts_bucket_start in (-Inf, 1782368104]), and((timestamp in (-Inf, 1782368103999999999]), and((ts_bucket_start in [1782362704, +Inf)), (timestamp in [1782364504000000000, +Inf)))))", "initialParts": 6, "selectedParts": 6, "initialGranules": 18, "selectedGranules": 12 }
      ]
    }
  ]
}
```

## 3. `?verbose=false` — verdict only

The lightweight opt-out: every query is still fully validated (its statement is built behind the
scenes), but nothing is rendered or EXPLAIN-ed — no ClickHouse round trips.

```json
{
  "status": "success",
  "data": {
    "compositeQuery": {
      "A": { "valid": true }
    }
  }
}
```

---

## 4. Per-query error — field not found

A `groupBy` (or aggregation) on a non-existent field. Reported per query; `valid:false` with a
structured `error` (and no rendered statements, since rendering is what failed). The nested
`errors[].suggestions` lists the valid references the field could have been.

```json
{
  "status": "success",
  "data": {
    "compositeQuery": {
      "A": {
        "valid": false,
        "error": {
          "type": "not-found",
          "code": "not_found",
          "message": "field not found",
          "errors": [
            {
              "message": "field `nonexistent_field_xyz` not found",
              "suggestions": ["valid references: `id`, `timestamp`"]
            }
          ]
        }
      }
    }
  }
}
```

## 5. Per-query error — with a "did you mean" suggestion

A near-miss field name (`service.nam` on traces) yields a `did you mean` entry in the nested
`errors[].suggestions`, ahead of the valid-references list (truncated by the server with `(+N more)`).

```json
{
  "status": "success",
  "data": {
    "compositeQuery": {
      "A": {
        "valid": false,
        "error": {
          "type": "not-found",
          "code": "not_found",
          "message": "field not found",
          "errors": [
            {
              "message": "field `service.nam` not found",
              "suggestions": [
                "did you mean: `serviceName`",
                "valid references: `durationNano`, `kind`, `dbName`, `externalHttpUrl`, `isRemote`, `httpMethod`, `responseStatusCode`, `timestamp`, `rpcMethod`, `spanKind`, `externalHttpMethod`, `dbSystem`, `httpUrl`, `httpRoute`, `statusCodeString`, `spanID`, `traceID`, `msgSystem`, `httpHost`, `parentSpanID` (+9 more)"
              ]
            }
          ]
        }
      }
    }
  }
}
```

## 6. Mixed — one valid, one invalid (per-query reporting)

The headline feature: query `A` validates (full output) while `B` fails, both reported in a single
response. (`A`'s `granules.reads[]` elided — see §2.)

```json
{
  "status": "success",
  "data": {
    "compositeQuery": {
      "A": {
        "valid": true,
        "statements": [
          {
            "db.statement.query": "SELECT count() AS __result_0 FROM signoz_logs.distributed_logs_v2 WHERE timestamp >= ? AND ts_bucket_start >= ? AND timestamp < ? AND ts_bucket_start <= ? ORDER BY __result_0 DESC",
            "db.statement.args": ["1782364504000000000", 1782362704, "1782368104000000000", 1782368104],
            "estimate": [
              { "database": "signoz_logs", "table": "logs_v2", "parts": 5, "rows": 26046, "marks": 11 }
            ],
            "granules": { "initial": 1864, "selected": 13, "skipped": 1851 }
          }
        ]
      },
      "B": {
        "valid": false,
        "error": {
          "type": "not-found",
          "code": "not_found",
          "message": "field not found",
          "errors": [
            {
              "message": "field `nonexistent.field` not found",
              "suggestions": ["valid references: `id`, `timestamp`"]
            }
          ]
        }
      }
    }
  }
}
```

## 7. Per-query error — invalid `having` reference

A `having` expression referencing an aggregation that isn't defined. The error lists the valid
references both in the legacy `errors[]` message and the structured `suggestions`.

```json
{
  "status": "success",
  "data": {
    "compositeQuery": {
      "A": {
        "valid": false,
        "error": {
          "type": "invalid-input",
          "code": "invalid_input",
          "message": "Invalid references in `Having` expression: [undefined_agg]",
          "errors": [ { "message": "Valid references are: [__result, __result0, count()]" } ],
          "suggestions": ["valid references: `__result`, `__result0`, `count()`"]
        }
      }
    }
  }
}
```

---

## 8. Valid query with a warning (step interval adjusted)

A `time_series` request over a 7-day range with `stepInterval:1` (1s). The query is valid; the step
is clamped to the minimum and the adjustment is reported as a warning. Shown with `?verbose=false`
to isolate the warning — the default full preview would also carry statements + analyses.

```json
{
  "status": "success",
  "data": {
    "compositeQuery": {
      "A": {
        "valid": true,
        "warnings": [
          "Query A is requesting aggregation interval 1 seconds, which is smaller than the minimum allowed interval of 400 seconds for selected time range. Using the minimum instead"
        ]
      }
    }
  }
}
```

## 9. Metric not found (warning, not error)

A metric builder query whose `metricName` has no metadata. This does **not** fail the query — it is
reported as a non-fatal warning, so `valid` stays `true` (the statement still builds; it just won't
match any samples).

```json
{
  "status": "success",
  "data": {
    "compositeQuery": {
      "A": {
        "valid": true,
        "warnings": [
          "metric nonexistent_metric_xyz_123 has never been received. Check the metric name and instrumentation"
        ]
      }
    }
  }
}
```

---

## 10. PromQL query

PromQL is not SQL — the Prometheus engine renders **one ClickHouse statement per metric selector**,
each captured (without executing) in `statements[]`. The selector matched no samples in the window,
so `selected` is 0 and the `EXPLAIN ESTIMATE` is all zeros. (`granules.reads[]` elided.)

```json
{
  "status": "success",
  "data": {
    "compositeQuery": {
      "A": {
        "valid": true,
        "statements": [
          {
            "db.statement.query": "SELECT metric_name, fingerprint, unix_milli, value, flags\n\t\t\tFROM signoz_metrics.distributed_samples_v4\n\t\t\tWHERE metric_name = $1 AND fingerprint GLOBAL IN (SELECT fingerprint FROM signoz_metrics.distributed_time_series_v4 WHERE metric_name = $2 AND temporality IN ['Cumulative', 'Unspecified'] AND unix_milli >= 1782363600000 AND unix_milli < 1782368104000 AND __normalized = false AND JSONExtractString(labels, $3) = $4 AND JSONExtractString(labels, $5) = $6 GROUP BY fingerprint) AND unix_milli >= $7 AND unix_milli <= $8 ORDER BY fingerprint, unix_milli;",
            "db.statement.args": ["signoz_calls_total", "signoz_calls_total", "service_name", "signoz_calls_total", "__name__", "signoz_calls_total", 1782364204001, 1782368104000],
            "estimate": [
              { "database": "signoz_metrics", "table": "samples_v4", "parts": 0, "rows": 0, "marks": 0 }
            ],
            "granules": { "initial": 143924, "selected": 0, "skipped": 143924 }
          }
        ]
      }
    }
  }
}
```

## 11. Metric builder query with `groupBy`

A rate→sum metric query renders to a CTE statement reading **two tables** (the time-series index
plus the samples table) — so `estimate` has two entries. This metric is delta-temporality, so the
rate is computed as `sum(value)/interval` in a single spatial-aggregation CTE; a cumulative metric
would also carry a temporal rate-window CTE. The `groupBy` label is absent on this metric, which is
surfaced as a non-fatal warning (the query still renders). (`granules.reads[]` elided.)

```json
{
  "status": "success",
  "data": {
    "compositeQuery": {
      "A": {
        "valid": true,
        "warnings": ["key `service_name` not found on metric signoz_calls_total"],
        "statements": [
          {
            "db.statement.query": "WITH __spatial_aggregation_cte AS (SELECT toStartOfInterval(toDateTime(intDiv(unix_milli, 1000)), toIntervalSecond(60)) AS ts, `service_name`, sum(value)/60 AS value FROM signoz_metrics.distributed_samples_v4 AS points INNER JOIN (SELECT fingerprint, JSONExtractString(labels, 'service_name') AS `service_name` FROM signoz_metrics.time_series_v4 WHERE metric_name IN (?) AND unix_milli >= ? AND unix_milli <= ? AND LOWER(temporality) LIKE LOWER(?) AND __normalized = ? GROUP BY fingerprint, `service_name`) AS filtered_time_series ON points.fingerprint = filtered_time_series.fingerprint WHERE metric_name IN (?) AND unix_milli >= ? AND unix_milli < ? GROUP BY ts, `service_name`) SELECT * FROM __spatial_aggregation_cte ORDER BY `service_name`, ts",
            "db.statement.args": ["signoz_calls_total", 1782363600000, 1782368100000, "delta", false, "signoz_calls_total", 1782364500000, 1782368100000],
            "estimate": [
              { "database": "signoz_metrics", "table": "time_series_v4", "parts": 6, "rows": 155495, "marks": 26 },
              { "database": "signoz_metrics", "table": "samples_v4", "parts": 4, "rows": 288609, "marks": 36 }
            ],
            "granules": { "initial": 149304, "selected": 62, "skipped": 149242 }
          }
        ]
      }
    }
  }
}
```

---

## 12. Raw `clickhouse_sql` — valid

Raw SQL is validated by an `EXPLAIN PLAN` bind check (it parses and its tables/columns/types
resolve) without executing. No `estimate` or `granules` here: the query reads a `distributed_*`
table, whose plan exposes no local MergeTree analysis, so both come back empty and are omitted.

```json
{
  "status": "success",
  "data": {
    "compositeQuery": {
      "A": {
        "valid": true,
        "statements": [
          { "db.statement.query": "SELECT count() AS total FROM signoz_logs.distributed_logs_v2" }
        ]
      }
    }
  }
}
```

## 13. Raw `clickhouse_sql` — invalid (bind-check failure)

An unknown table is caught by the `EXPLAIN PLAN` bind check and mapped to invalid input.

```json
{
  "status": "success",
  "data": {
    "compositeQuery": {
      "A": {
        "valid": false,
        "error": {
          "type": "invalid-input",
          "code": "invalid_input",
          "message": "invalid ClickHouse SQL: Unknown table expression identifier 'signoz_logs.no_such_table' in scope SELECT count() FROM signoz_logs.no_such_table"
        }
      }
    }
  }
}
```

---

## 14. Trace operator — per-sub-query reporting

A trace operator query (e.g. `C := A => B`) combines sibling trace queries. `QueryRange` never
executes the dependencies (`A`, `B`) standalone, but the preview **renders each one on its own** so
the caller can see *which* sub-query is the problem rather than getting a single opaque failure on
`C`. A dependency is rendered the way the operator actually consumes it — a `RequestTypeRaw` span
selection (its filter only; aggregation/group-by/order run later on the combined result) — so its
statement and analyses reflect that span-selection read, not the operator's final shape.

### 14a. All sub-queries valid

`A` and `B` each render their own span-selection statement with estimate + granules; `C` renders the
combined CTE query that joins them. (`granules.reads[]` elided — the trace `reads[]` carry the
`Skip` steps noted in §2b.)

```json
{
  "status": "success",
  "data": {
    "compositeQuery": {
      "A": {
        "valid": true,
        "statements": [
          {
            "db.statement.query": "SELECT duration_nano AS `duration_nano`, name AS `name`, response_status_code AS `response_status_code`, resource.`service.name`::String AS `service.name`, span_id AS `span_id`, timestamp AS `timestamp`, trace_id AS `trace_id` FROM signoz_traces.distributed_signoz_index_v3 WHERE name = ? AND timestamp >= ? AND timestamp < ? AND ts_bucket_start >= ? AND ts_bucket_start <= ? LIMIT ?",
            "db.statement.args": ["HTTP GET", "1782364504000000000", "1782368104000000000", 1782362704, 1782368104, 100],
            "estimate": [
              { "database": "signoz_traces", "table": "signoz_index_v3", "parts": 5, "rows": 25580, "marks": 9 }
            ],
            "granules": { "initial": 87075, "selected": 9, "skipped": 87066 }
          }
        ]
      },
      "B": {
        "valid": true,
        "statements": [
          {
            "db.statement.query": "SELECT duration_nano AS `duration_nano`, name AS `name`, response_status_code AS `response_status_code`, resource.`service.name`::String AS `service.name`, span_id AS `span_id`, timestamp AS `timestamp`, trace_id AS `trace_id` FROM signoz_traces.distributed_signoz_index_v3 WHERE name = ? AND timestamp >= ? AND timestamp < ? AND ts_bucket_start >= ? AND ts_bucket_start <= ? LIMIT ?",
            "db.statement.args": ["SELECT", "1782364504000000000", "1782368104000000000", 1782362704, 1782368104, 100],
            "estimate": [
              { "database": "signoz_traces", "table": "signoz_index_v3", "parts": 0, "rows": 0, "marks": 0 }
            ],
            "granules": { "initial": 87075, "selected": 0, "skipped": 87075 }
          }
        ]
      },
      "C": {
        "valid": true,
        "statements": [
          {
            "db.statement.query": "WITH toDateTime64(1782364504000000000, 9) AS t_from, toDateTime64(1782368104000000000, 9) AS t_to, 1782362704 AS bucket_from, 1782368104 AS bucket_to, all_spans AS (SELECT *, resource_string_service$$name AS `service.name` FROM signoz_traces.distributed_signoz_index_v3 WHERE timestamp >= ? AND timestamp < ? AND ts_bucket_start >= ? AND ts_bucket_start <= ?), A AS (SELECT * FROM signoz_traces.distributed_signoz_index_v3 WHERE timestamp >= ? AND timestamp < ? AND ts_bucket_start >= ? AND ts_bucket_start <= ? AND name = ?), B AS (SELECT * FROM signoz_traces.distributed_signoz_index_v3 WHERE timestamp >= ? AND timestamp < ? AND ts_bucket_start >= ? AND ts_bucket_start <= ? AND name = ?), A_DIR_DESC_B AS (SELECT p.* FROM A AS p INNER JOIN B AS c ON p.trace_id = c.trace_id AND p.span_id = c.parent_span_id) SELECT timestamp, trace_id, span_id, name, duration_nano, parent_span_id FROM A_DIR_DESC_B ORDER BY timestamp DESC LIMIT ? SETTINGS distributed_product_mode='allow', max_memory_usage=10000000000",
            "db.statement.args": ["1782364504000000000", "1782368104000000000", 1782362704, 1782368104, "1782364504000000000", "1782368104000000000", 1782362704, 1782368104, "HTTP GET", "1782364504000000000", "1782368104000000000", 1782362704, 1782368104, "SELECT", 100],
            "estimate": [
              { "database": "signoz_traces", "table": "signoz_index_v3", "parts": 5, "rows": 25580, "marks": 9 }
            ],
            "granules": { "initial": 174150, "selected": 9, "skipped": 174141 }
          }
        ]
      }
    }
  }
}
```

### 14b. A bad sub-query is pinpointed

Sub-query `B` filters on a non-existent field. `B` is reported invalid with the precise error, `A`
stays valid, and the operator `C` fails too — so the caller sees the root cause is `B`, not just
that `C` broke. (Before per-sub-query previewing, `A` and `B` were omitted entirely and only `C`'s
opaque failure was returned. `A`'s statements elided here for brevity.)

```json
{
  "status": "success",
  "data": {
    "compositeQuery": {
      "A": { "valid": true, "statements": [ /* … */ ] },
      "B": {
        "valid": false,
        "error": {
          "type": "invalid-input",
          "code": "invalid_input",
          "message": "Found 1 errors while parsing the search expression.",
          "url": "https://signoz.io/docs/userguide/search-troubleshooting/#key-fieldname-not-found",
          "errors": [ { "message": "key `nonexistent_field_xyz` not found" } ]
        }
      },
      "C": {
        "valid": false,
        "error": {
          "type": "invalid-input",
          "code": "invalid_input",
          "message": "Found 1 errors while parsing the search expression.",
          "url": "https://signoz.io/docs/userguide/search-troubleshooting/#key-fieldname-not-found",
          "errors": [ { "message": "key `nonexistent_field_xyz` not found" } ]
        }
      }
    }
  }
}
```

---

## Top-level (request-level) errors

These abort the whole request before per-query previewing, so they return `{"status":"error", …}`
rather than a `data.compositeQuery` map.

### 15. Duplicate query name

```json
{
  "status": "error",
  "error": {
    "type": "invalid-input",
    "code": "invalid_input",
    "message": "duplicate query name 'A'"
  }
}
```

### 16. Invalid `?verbose=` value

A non-boolean value for `verbose` is rejected before previewing.

```json
{
  "status": "error",
  "error": {
    "type": "invalid-input",
    "code": "invalid_input",
    "message": "invalid verbose value \"bogus\" (allowed: true, false)"
  }
}
```

### 17. Invalid enum at decode time (bad `signal`)

An invalid enum value (`signal:"bogus_signal"`) fails during request decode, so it aborts at the top
level instead of being reported under the query. The valid values come back in `suggestions`.

```json
{
  "status": "error",
  "error": {
    "type": "invalid-input",
    "code": "invalid_input",
    "message": "invalid signal \"bogus_signal\"",
    "suggestions": ["valid references: `traces`, `logs`, `metrics`"]
  }
}
```

### 18. Malformed JSON body

A syntactically broken body is an opaque decoder error (wrapped as internal/unknown).

```json
{
  "status": "error",
  "error": {
    "type": "internal",
    "code": "unknown",
    "message": "invalid character 'n' looking for beginning of object key string"
  }
}
```

pkg/querier/preview.go
```diff
@@ -0,0 +1,734 @@
...
+	return false, errors.NewInvalidInputf(errors.CodeInvalidInput, "invalid verbose value %q (allowed: true, false)", value)
+}
+
+// QueryRangePreview validates each query in the composite query without
+// executing it. With opts.Verbose=false it returns a lightweight per-query
+// verdict (valid/error/warnings). With opts.Verbose=true it also renders the
+// underlying ClickHouse statement(s) each query would run and attaches, per
+// statement, the EXPLAIN ESTIMATE and granule index analysis, deriving the
+// top-level SelectivityScore and MagnitudeScore.
+func (q *querier) QueryRangePreview(
```
therealpandey, line 19: Looks like the entire querier has been copy pasted and duplicated here.

therealpandey (changes requested):

pkg/querier/preview.go
tushar-signoz: Done

pkg/querier/preview.go
```diff
@@ -0,0 +1,606 @@
...
+	chproto.ErrIllegalColumn:                true,
+	chproto.ErrNumberOfArgumentsDoesntMatch: true,
+	chproto.ErrTooManyArgumentsForFunction:  true,
+	chproto.ErrTooLessArgumentsForFunction:  true,
+}
+
+// statementProvider renders a query's underlying statement without executing it.
+type statementProvider interface {
+	Statement(ctx context.Context) (*qbtypes.Statement, error)
+}
```
therealpandey: This interface needs to be present along with the statement builder interface. Either in the same package or the same interface.

pkg/querier/preview.go
```diff
@@ -0,0 +1,606 @@
...
+	chproto.ErrUnknownType:                  true,
+	chproto.ErrUnknownStorage:               true,
+	chproto.ErrUnknownElementInAst:          true,
+	chproto.ErrUnknownTypeOfQuery:           true,
+	chproto.ErrIllegalTypeOfArgument:        true,
+	chproto.ErrIllegalColumn:                true,
+	chproto.ErrNumberOfArgumentsDoesntMatch: true,
+	chproto.ErrTooManyArgumentsForFunction:  true,
+	chproto.ErrTooLessArgumentsForFunction:  true,
+}
```
therealpandey: This map needs to be pushed down in telemetry store.

therealpandey (changes requested):

pkg/querier/preview.go
tushar-signoz: We can't push it to telemetrystore, reason being signoz is building the sql and any of the above errors is signoz's fault and user cannot do anything about it. 

Only in case of clickhouse_sql request type where sql is given by the user, we can return a 404 bad request.

tushar-signoz:
# Claude Review

**What it does:** Adds `POST /api/v5/query_range/preview`, a dry-run endpoint that takes the same payload as `query_range` but validates and *renders* each query to its underlying ClickHouse/PromQL statement(s) without executing them — returning a per-query verdict plus, on the default `verbose=true` path, each statement's `EXPLAIN ESTIMATE` and granule index-pruning funnel. It's purely additive (new route gated behind `ViewAccess`), splits validation into request-scope vs per-query, adds `Statement()` providers per query type, and a PromQL `StatementCapturer` that drives the Prom engine with an empty-result storage to record SQL.

The capture path faithfully mirrors `client.Read`, the `buildSamplesQuery` extraction is byte-identical, and the `make([]string, 0, cap)` change in `where_clause_visitor.go` is a genuine fix (the old `make([]string, len)` + `append` produced leading empty strings). The findings below survived verification.

## Findings (most severe first)

**1. `pkg/querier/preview.go:135` — invalid `clickhouse_sql` is reported `valid:true` instead of as an error.**
`explainBindCheck` (preview.go:389) returns the *raw* ClickHouse driver error, and `errors.Ast(bindErr, errors.TypeInvalidInput)` is always false for a non-`*base` error — `Unwrapb` returns `TypeInternal` for it (errors.go:269). So the invalid-input branch is dead: a user previewing `clickhouse_sql` with a syntax/semantic error (e.g. `SELEC count() FROM signoz_traces`) gets the failure downgraded to a warning with `ps.Error` nil → `valid:true`. The endpoint's core promise — catch invalid queries — silently fails for every CH-SQL error. **Fix:** wrap/classify the error *inside* `explainBindCheck` (e.g. `errors.WrapInvalidInputf`) so the caller's `Ast` check can match.

**2. `pkg/querier/preview.go:20` — preview never normalizes `req.Start`/`req.End` to milliseconds.**
`QueryRange` coerces the window up front (`querier.go:92-93`) "so every downstream consumer can assume ms regardless of the resolution the caller sent." The preview feeds the raw window straight into `adjustStepInterval` (:52), `resolveMetricMetadata` (:54), and the statement builders. A caller sending seconds or nanoseconds gets rendered SQL, time-bound args, EXPLAIN estimates, and step recommendations for a window off by ~1000× from what `query_range` would actually run — defeating the dry-run for any non-ms caller. **Fix:** add the same `ToMilliSecs` coercion at the top of `QueryRangePreview`.

**3. `pkg/querier/preview.go:297` — unbounded goroutine fan-out of ClickHouse `EXPLAIN`s.**
`runPreviewTasks` spawns one goroutine per statement with no concurrency cap, each running 2 blocking `EXPLAIN` round trips (ESTIMATE + `json=1,indexes=1`), on top of a 3rd `EXPLAIN PLAN` bind-check for CH-SQL queries. A verbose preview of a many-query / many-selector composite issues up to ~3N concurrent EXPLAINs against the pool shared with production traffic; a few concurrent VIEWER preview requests can saturate `MaxOpenConns` and reproduce the known `acquire conn timeout` pool exhaustion. **Fix:** bound the fan-out (e.g. `errgroup.SetLimit`) and/or cap statements per request.

**4. `pkg/querier/preview.go:54` — per-query metric-metadata fan-out (regression from execution path).**
`resolveMetricMetadata` and `adjustStepInterval` are called once *per query* inside the loop on a single-element slice, whereas `QueryRange` calls them once with the whole `CompositeQuery.Queries` slice (`querier.go:112-114`), batching all metric names into a single metadata fetch. A request with N metric queries issues N metadata round trips where 1 suffices. **Fix:** hoist both calls out of the loop and attribute warnings back by query name.

**5. `pkg/types/querybuildertypes/querybuildertypesv5/resp.go:92` — `QueryPreview.Valid` is a dead field.**
No producer ever assigns it; the verdict is derived solely in `MarshalJSON` (`out.Valid = p.Error == nil`, resp.go:160). Any Go caller reading `preview.Valid` before marshaling gets the zero value `false` regardless of `Error` — a latent correctness trap in a public struct. **Fix:** drop the field and rely on the derived value.

**6. `pkg/querier/preview.go:343` — reflection scan buys nothing and duplicates `consume.go`.**
`runExplainEstimate` allocates fully-typed scan slots via `reflect.New(ct.ScanType())` then collapses every value through `fmt.Sprintf("%v")` / `toInt64`, discarding the typing; `toInt64` (:375) also re-implements the numeric-coercion switch already in `consume.go:504` (`numericAsFloat`), and the slot pattern is the same one at `consume.go:65/323/382`. EXPLAIN ESTIMATE has a fixed schema (`database/table String, parts/rows/marks UInt64`). **Fix:** scan into typed locals directly and drop the reflection block plus `toInt64`, or share the existing helper.

**7. `pkg/types/querybuildertypes/querybuildertypesv5/validation.go:582` — `ValidateRequestScope` duplicates request-level checks.**
It re-implements every request-scope check already in `Validate` / `CompositeQuery.Validate` (time range, request-type enum, raw-vs-metric guard, empty-queries, builder/sub-query name uniqueness, all-disabled) as a parallel copy with no shared helper. As request-level validation evolves, the two will drift and `query_range` vs `query_range/preview` will come to disagree about what is a valid request. **Fix:** extract a shared helper both paths call.

## Considered and dropped
- **Name-collision overwrite** for same-named non-builder queries — the uniqueness gap is identical in old and new validation (pre-existing, not introduced here).
- **Trace-operator `RequestType` "inconsistency"** — the operator correctly renders under the request type, matching execution, while dependencies are intentionally previewed as raw.
- **Unrecovered-goroutine-panic** concern — `reflect.New(ScanType())` is the same pattern `consume.go` already relies on in production for fixed EXPLAIN columns, so nil-ScanType isn't a realistic trigger here.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

pkg/querier/preview.go
```diff
@@ -0,0 +1,497 @@
...
+			}
+		}
+		ps.Warnings = append(ps.Warnings, outcomes[i].warnings...)
+		previews[tasks[i].Name] = ps
+	}
+}
+
+func (q *querier) runExplainEstimate(ctx context.Context, stmt string, args []any) ([]qbtypes.EstimateEntry, error) {
+	rows, err := q.telemetryStore.ClickhouseDB().Query(ctx, "EXPLAIN ESTIMATE "+stmt, args...)
+	if err != nil {
```
therealpandey:
```suggestion
	rows, err := q.telemetryStore.Explain(ctx, "EXPLAIN ESTIMATE "+stmt, args...)
```

Can we make these a part of the telemetrystore interface? The types can be in `telemetrystore/`.

While moving it there, please ensure there is no sql injection possible via `stmt`

pkg/querier/preview.go
```diff
@@ -0,0 +1,497 @@
...
+	rows, err := q.telemetryStore.ClickhouseDB().Query(ctx, "EXPLAIN PLAN "+stmt, args...)
+	if err != nil {
+		return err
+	}
+	rows.Close()
+	return nil
+}
+
+func (q *querier) computeGranuleStats(ctx context.Context, stmt string, args []any) (qbtypes.Granules, bool, error) {
+	rows, err := q.telemetryStore.ClickhouseDB().Query(ctx, "EXPLAIN json = 1, indexes = 1 "+stmt, args...)
```
therealpandey:
```suggestion
	rows, err := q.telemetryStore.Explain(ctx, "EXPLAIN json = 1, indexes = 1 "+stmt, args...)
```

Same as explain estimate

therealpandey (changes requested):

pkg/telemetrystore/clickhousetelemetrystore/explain.go
```diff
@@ -0,0 +1,218 @@
...
+	Keys             []string `json:"Keys"`
+	Condition        string   `json:"Condition"`
+	InitialParts     *int64   `json:"Initial Parts"`
+	SelectedParts    *int64   `json:"Selected Parts"`
+	InitialGranules  *int64   `json:"Initial Granules"`
+	SelectedGranules *int64   `json:"Selected Granules"`
+}
+
+// RunExplainEstimate backs TelemetryStore.Estimate.
+func RunExplainEstimate(ctx context.Context, conn clickhouse.Conn, stmt string, args ...any) ([]telemetrystoretypes.EstimateEntry, error) {
```
therealpandey, line 38: All this in clickhousetelemterystore now.

therealpandey (changes requested):

therealpandey (approved):

PR #11190, opened 2026-05-05
test: dashboards list spec with new e2e framework

AshwinBhatkal:
## Pull Request

---

### 📄 Summary

Adds E2E coverage for the dashboards list page, plus the supporting harness around it: helper utilities, suite-level seeding, a real-world JSON test fixture, and three Claude agents that codify the planner / generator / healer workflow this repo will use to write more specs.

The spec runs 33 scenarios (TC-01–TC-33) against the pytest-bootstrapped backend — page load, search, sorting, row actions, the New-dashboard dropdown, the Import JSON dialog, delete flows, sidebar navigation, URL deep-linking, and header actions. Every test is self-contained (seeds via API, cleans up via one `afterAll`) and runs green in chromium.

This is the first feature spec written against the new harness in [`docs/contributing/tests/e2e.md`](https://github.com/SigNoz/signoz/blob/main/docs/contributing/tests/e2e.md), so the PR also doubles as the canonical example future specs will copy from.

N/A — adds tests + tooling only, no app code change.

#### Issues closed by this PR

N/A

---

### ✅ Change Type

- [x] 🧪 Test-only
- [x] 🛠️ Infra / Tooling
- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor

---

### 🐛 Bug Context

N/A

---

### 🧪 Testing Strategy

- **Tests added:** 33 scenarios in `tests/e2e/tests/dashboards/dashboards-list.spec.ts` covering page load, search (title / tag / case-insensitive / clear / empty-state / deeplink), sorting (default / Last updated / Last created), row actions (View / Open in New Tab / Copy Link / Export JSON / dismiss), New-dashboard dropdown (3 options / Create / Import JSON dialog open / Escape close / X close), delete flows (dialog / Cancel / Confirm), row-click navigation, sidebar nav, URL state, and header buttons.
- **Manual verification:** ran 3 consecutive full-suite runs locally — 33/33 green each time. Spot-checked CI failure paths after fixes (Ant Popover viewport, Monaco Escape, sticky-header scroll, React Query mutation race).
- **Edge cases covered:** zero-state vs populated workspace, search-after-pagination, popover positioning under sticky table headers, Monaco editor keystroke interception, the inline "Enter dashboard name…" form being a feedback widget rather than a create flow.

---

- **Blast radius:** test-only — no production code changes. The one repo-level change is `.dockerignore` (the community Dockerfile expects `frontend/build/` in the build context; previously excluded). E2E build is unaffected because `Dockerfile.with-web.integration` builds the frontend inside the container.
- **Potential regressions:** none for production. Spec accuracy depends on stable `data-testid` attributes in `frontend/src/container/ListOfDashboard/`; future renames there will surface as test failures.
- **Rollback plan:** revert the PR. No data migrations, no infra changes.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | N/A |
| Change Type | Maintenance |
| Description | N/A |

---

### 📋 Checklist

- [x] Tests added
- [x] Manually tested (3 full-suite runs locally + targeted re-runs)
- [x] Backward compatibility considered (test-only)

---

## 👀 Notes for Reviewers

**Layout introduced under `tests/e2e/`:**

- `fixtures/auth.ts` — only true Playwright `test.extend` fixture (`authedPage`).
- `helpers/auth.ts` — `newAdminContext(browser)` for `beforeAll` / `afterAll` hooks where the test-scoped `authedPage` isn't reachable.
- `helpers/dashboards.ts` — function helpers (`gotoDashboardsList`, `createDashboardViaApi`, `importApmMetricsDashboardViaUI`, `deleteDashboardViaApi`, `findDashboardIdByTitle`, `openDashboardActionMenu`) plus the constants both helpers and specs share (`SEARCH_PLACEHOLDER`, `LIST_HEADING`, `DEFAULT_DASHBOARD_TITLE`, `APM_METRICS_TITLE`).
- `testdata/apm-metrics.json` — real APM Metrics dashboard payload, uploaded through the actual Import JSON UI flow during `beforeAll`.
- `specs/dashboards/checklists/` — manual QA checklist (companion to the automated spec).

**Patterns established that future specs should copy:**

1. **Suite-level seed registry:** `seedIds: Set<string>` populated by every `seed()` call, drained by one `afterAll`. Replaces per-test `try / finally` boilerplate in larger spec files.
2. **API-driven seeding via `page.request.*` with explicit bearer token** (the JWT lives in `localStorage.AUTH_TOKEN`; the React Axios interceptor that auto-attaches it doesn't run for `page.request` calls — `authToken(page)` reads it manually).
3. **`importApmMetricsDashboardViaUI` exercises the real Import JSON path** (file upload via Ant Upload's hidden `<input type="file">` + Monaco editor + Import and Next button) rather than POSTing the JSON directly. So the suite seed itself is end-to-end tested every run.

**Documented gotchas baked into the helpers (and the healer agent's runbook):**

- **Ant Popover viewport.** `tooltip.getByText('Delete dashboard').click({ force: true })` *still* fails with "Element is outside of the viewport" in CI even though `force` skips actionability checks — Playwright requires click coordinates inside the viewport. Fix: `dispatchEvent('click')` (no coordinate calculation, React's onClick still runs).
- **Sticky header.** Action icons can land below the table's sticky header; `openDashboardActionMenu` does `scrollIntoViewIfNeeded` before clicking.
- **React Query mutations race navigation.** UI delete fires async; navigating away cancels the in-flight `DELETE`. Pair the click with `page.waitForResponse((r) => r.request().method() === 'DELETE' && /\/dashboards\//.test(r.url()))` and `await expect(dialog).not.toBeVisible()` before the next `goto()`.
- **Monaco swallows Escape.** Inside the Import JSON dialog, click the modal title first to blur Monaco, then press Escape.
- **The inline "Enter dashboard name…" field is a `RequestDashboardBtn` (template-feedback form), not a create flow.** Don't use it to seed.

**`.dockerignore` change** — `frontend/build/` was excluded; the community Dockerfile (`COPY frontend/build/ /etc/signoz/web/`) was failing in CI with `failed to compute cache key`. Removed from the exclusions. E2E build is unaffected — `Dockerfile.with-web.integration` runs `yarn build` inside the container, which overwrites whatever stale host artifacts get copied in.

**Claude agents added under `.claude/agents/`** — `playwright-test-planner`, `playwright-test-generator`, `playwright-test-healer`. They reference the helpers, the fixture, the gotchas, and the test-plan / checklist convention so future feature coverage can be driven through them.


docs/contributing/tests/e2e.md
```diff
@@ -81,8 +81,17 @@ tests/
     ├── .env.local             # generated by bootstrap/setup.py (gitignored)
     ├── bootstrap/
     │   └── setup.py           # test_setup / test_teardown — pytest lifecycle
-    ├── fixtures/
+    ├── fixtures/              # Playwright test fixtures (test.extend) only
     │   └── auth.ts            # authedPage Playwright fixture + per-worker storageState cache
+    ├── helpers/               # function helpers + the constants they share with specs
+    │   ├── auth.ts            # newAdminContext(browser) for beforeAll/afterAll
+    │   └── dashboards.ts      # gotoDashboardsList, createDashboardViaApi, openDashboardActionMenu, plus SEARCH_PLACEHOLDER, APM_METRICS_TITLE…
```
therealpandey:
```suggestion
    │   ├── auth.ts 
    │   └── dashboards.ts 
```

Let's not put exact descriptions as this file might go out of sync as the codebase evolves.

docs/contributing/tests/e2e.md
```diff
@@ -81,8 +81,17 @@ tests/
...
+    ├── testdata/              # static data files (JSON) used by helpers and specs
+    │   └── apm-metrics.json   # real APM dashboard payload — seed via importApmMetricsDashboardViaUI
```
therealpandey:
```suggestion
    │   └── apm-metrics.json 
```

Just mention (example) in the bracket.

docs/contributing/tests/e2e.md
```diff
@@ -81,8 +81,17 @@ tests/
...
+    ├── specs/                 # human-readable test plans + QA checklists per feature
+    │   └── dashboards/
+    │       ├── dashboards-list-test-plan.md
+    │       └── checklists/
```
therealpandey:
```suggestion
```

Just need to update this. Are we keeping specs also? If there is the slightest chance it will go out of sync with the actual tests, no need to keep them.

therealpandey (changes requested):

docs/contributing/tests/e2e.md
AshwinBhatkal: It won't be there in the actual files pushed. specs is already in the gitignore. It is just directory structure. Have updated the actual source of truth once tests have been written

therealpandey (approved):

PR #11196, opened 2026-05-05
fix: use ms in prepareFillZeroArgsWithStep

nityanandagohain:
## Pull Request

---

### 📄 Summary
This makes sure that it uses ms regardless of whether the input is ms or ns in prepareFillZeroArgsWithStep

#### Issues closed by this PR
Fixes https://github.com/SigNoz/platform-pod/issues/2205
---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context
If the timestamp was in ns, then this function created too many iterations as the step is in ms.  So now the interval is converted to ms

#### Fix Strategy
The start and end is converted to ms

---

- Manual verification:

---

- Blast radius: None
- Potential regressions:
- Rollback plan:

---

### DIFF
--- a/pkg/querier/postprocess.go
+++ b/pkg/querier/postprocess.go
@@ -12,6 +12,7 @@ import (
 	"github.com/SigNoz/govaluate"
 
 	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/querybuilder"
 	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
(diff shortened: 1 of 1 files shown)


tushar-signoz (approved):
Good catch !!!

nityanandagohain:
Thanks for the third eye @tushar-signoz

therealpandey (approved):

PR #11208, opened 2026-05-06
fix: added fix for traceoperator not getting saved in alerts

ahrefabhi:
## Pull Request

---

### 📄 Summary
Trace operator queries were not being persisted when creating threshold alerts in Alerts V2, and trace operator results returned by the V5 query-range API were not being correctly mapped back into the query builder state. This PR wires `queryTraceOperator` into the alert payload, fixes the round-trip mapping for `builder_trace_operator` specs, and prevents the V5 payload builder from emitting a metrics-shaped aggregation for trace operators (which the backend rejects).

N/A

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/4817

---

### ✅ Change Type
- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
Three issues combined to break trace-operator persistence in alerts:

1. **`buildCreateThresholdAlertRulePayload`** only mapped `query.builder.queryData` and `query.builder.queryFormulas` into the outgoing `builderQueries` payload. `query.builder.queryTraceOperator` was dropped on the floor, so trace operators were never sent to the alerts API on save.
2. **`mapQueryFromV5`** read `builder_trace_operator` envelopes back from the API by casting the raw `spec` straight to `IBuilderTraceOperator`. The shape returned by the V5 backend is the wire format, not the query-builder shape, so fields like `filters`, `aggregations`, ordering, etc. were not normalized when re-hydrating the form.
3. **`convertTraceOperatorToV5`** invoked `createAggregation(traceOperatorData, ...)` directly. Because `IBuilderTraceOperator` has no `dataSource`, the function's branching was fragile: any change that pushed it down the metrics branch would emit a `metricName` field, which the backend rejects for `builder_trace_operator` envelopes.

#### Fix Strategy
- `Footer/utils.tsx`: include `mapQueryDataToApi(query.builder.queryTraceOperator, 'queryName').data` in the `builderQueries` spread so trace operators are part of the saved alert payload.
- `mapQueryDataFromApi.ts`: route `builder_trace_operator` specs through `convertBuilderQueryToIBuilderQuery` (then cast to `IBuilderTraceOperator`) so the API response is normalized into the query-builder shape on load.
- `prepareQueryRangePayloadV5.ts`: spread `{ ...traceOperatorData, dataSource: DataSource.TRACES }` into the `createAggregation` call so the helper always takes the traces branch and produces a `TraceAggregation[]` the backend accepts. Comment updated to capture the reasoning.

---

### 🧪 Testing Strategy

- Tests added/updated: None — existing alert + query-builder coverage exercises these paths.
- Manual verification:
  - Created a threshold alert with a trace-operator query (e.g. `A => B`), saved, reopened — operator now persists with its expression, filters, and aggregations.
  - Edited and re-saved the same alert; round-trip preserved.
  - Verified non-trace-operator alerts (metrics, logs, formulas, PromQL, ClickHouse) still save and reload unchanged.
- Edge cases covered: empty/whitespace `expression` (still filtered out before payload prep), `raw` request type (aggregation still skipped), trace operators with custom `groupBy` / `having` / `orderBy`.

---

- **Blast radius:** Alerts V2 create/edit flow and the V5 query-range payload pipeline. The `mapQueryDataFromApi` change touches the load path for any query that comes back as `builder_trace_operator`.
- **Potential regressions:** If a previously saved alert had a `builder_trace_operator` spec relying on the old raw cast, its hydrated form may now look different (it should look *more* correct, but worth eyeballing). Non-trace-operator queries are not touched.
- **Rollback plan:** Revert this PR — the three changes are isolated to alert payload assembly and trace-operator handling, and contain no schema or data migrations.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Fixes trace-operator queries not being saved when creating threshold alerts and not being correctly restored when editing them. |

---

### 📋 Checklist
- [x] Manually tested

---


H4ad (approved):
While testing, foudn a bug that the cache, after saving, is persisting the old alert instead of the updated one.

piyushsingariya:
@ahrefabhi any update here?

PR #11211, opened 2026-05-06
chore: breakdown query range function

tushar-signoz:
## Pull Request

---

### 📄 Summary

Refactor `querier.QueryRange` by breaking the ~250-line function into focused helpers. No behavior change — same inputs produce the same outputs and warnings; the responsibilities are just easier to read and test.

Extracted helpers in `pkg/querier/querier.go`:
- `constructTraceOperatorDependencyMap` — parses `QueryBuilderTraceOperator` expressions and returns the set of dependency query names.
- `adjustStepInterval` (+ `clampStep`, `secondsStep`) — normalizes step intervals per signal (trace/log, meter, metric, PromQL, trace-operator), computes per-signal recommended/min bounds once per call, and returns clamp warnings. Replaces five near-identical step-clamp blocks.
- `populateQBEvent` — fills the `QBEvent` telemetry struct (signal flags, filter/group-by applied, ClickHouse source detection).
- `resolveMetricMetadata` — fetches metric temporality and type via `FetchTemporalityAndTypeMulti`, patches each metric-aggregation spec in place, returns the set of fully-missing-metric query names, and surfaces dormant-metric warnings / `NotFound` errors.

The main `QueryRange` body now reads as a small pipeline: build the QB event → resolve trace-operator deps → adjust step intervals → resolve metric metadata → run queries → assemble response. The per-query type switch also uses the new `QueryEnvelope` accessors (`GetQueryName`, `GetType`, `GetStepInterval` / `SetStepInterval`, `GetSource`, `GetFilter`, `GetGroupBy`) instead of repeating the same type assertions inline.

#### Issues closed by this PR

Fixes https://github.com/SigNoz/engineering-pod/issues/4374

---

### ✅ Change Type

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🧪 Testing Strategy

- Tests added/updated: none — refactor preserves existing behavior; covered by existing querier tests.
- Manual verification: ran query-range requests across builder (traces/logs/metrics/meter), PromQL, ClickHouse SQL, and trace-operator query types; verified step-interval clamp warnings, missing-metric `NotFound` errors, and dormant-metric warning strings still match pre-refactor output.
- Edge cases covered: zero step interval, sub-minimum step interval, mixed signal composite queries, metric query whose every aggregation is missing (preseeded empty result path), trace-operator dependency skipping.

---

- Blast radius: every query-range request goes through this code path, so any behavioral drift would surface broadly.
- Potential regressions: subtle differences in warning ordering or `QBEvent` flags if a helper is called in a different order than before. Verified the call order matches the original pass.
- Rollback plan: revert this single commit — the refactor is contained to `pkg/querier/querier.go`.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | N/A |
| Change Type | Maintenance |
| Description | N/A (internal refactor, no user-facing change) |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- Best reviewed with whitespace ignored — most of the diff is code moving into helpers.
- The new `QueryEnvelope` accessors (`GetQueryName`, `GetStepInterval`/`SetStepInterval`, `GetSource`, `GetFilter`, `GetGroupBy`, `GetType`) are assumed to exist on the envelope type — please confirm these are the intended public API and that `SetStepInterval` mutates the underlying spec correctly for every query variant.
- `adjustStepInterval` precomputes all six recommended/min bounds even when only one signal is in the request. Cheap, but flagging in case that's worth lazifying.


tushar-signoz:
⏺ Code Review — PR #11211: chore: breakdown query range function

  Overview

  Refactors querier.QueryRange (~250 lines) into focused helpers: constructTraceOperatorDependencyMap, adjustStepInterval (+clampStep/secondsStep), populateQBEvent, resolveMetricMetadata. The split is clean and
  readable, and the main pipeline is much easier to follow. However, despite the PR claiming "no behavior change," I found three behavioral drifts that should be acknowledged or fixed.

  Behavior changes (the PR description says there are none)

  1. FilterApplied / GroupByApplied semantics changed — = → ||=

  Original (querier.go:182-183, :199-200, :216-217): each builder query overwrites the flags. The result depends on the order of queries in CompositeQuery.Queries and the last builder query's filter/groupby state
  wins.

  New (populateQBEvent):
  event.FilterApplied = event.FilterApplied || (filter != nil && filter.Expression != "")
  event.GroupByApplied = event.GroupByApplied || len(query.GetGroupBy()) > 0
  Now any query with a filter or group-by sets the flag. This is almost certainly more correct, but it is a real change to the telemetry signal. Either call it out in the PR body, or restore the original last-wins
  assignment if downstream metrics depend on it.

  2. TracesUsed is now set for QueryTypeTraceOperator

  Original never set TracesUsed for trace-operator queries. New populateQBEvent does:
  case qbtypes.QueryTypeTraceOperator:
      event.TracesUsed = true
  Likely correct, but it's a new flag in the QBEvent telemetry. Worth a line in the PR.

  3. Meter step-clamp now emits a warning to users — intervalWarnings drift

  This one I'd flag as a potential bug, not just a telemetry tweak.

  Original meter branch (querier.go:219-229):
  if spec.Source == telemetrytypes.SourceMeter {
      ...
      if spec.StepInterval.Seconds() < float64(...MinAllowedStepIntervalForMeter...) {
          newStep := ...
          spec.StepInterval = newStep   // NO intervalWarnings append
      }
  }
  The trace/log/metric branches all append(intervalWarnings, ...), but the meter branch deliberately does not.

  New code routes meter through the same clampStep helper:
      if qe.GetSource() == telemetrytypes.SourceMeter {
          clampStep(qe, meterRecommended, meterMin, &warnings)  // always warns
  So users running meter queries with sub-minimum step intervals will now see a warning string they did not see before. If the omission was intentional (meter clamps may be expected/silent), this is a regression.
  If it was an oversight in the original, this is a (silent) bug fix — but either way, it should be acknowledged.

  Correctness — other items

  - QueryEnvelope accessor assumption. The PR self-flags this. The biggest landmine is SetStepInterval on qbtypes.PromQuery, since the original wrote spec.Step (not spec.StepInterval). Confirm the accessor maps to
  the correct field; otherwise PromQL recommended-step inflation silently breaks. Same for GetSource() on MetricAggregation — original reads spec.Source directly and switches on SourceMeter.
  - Pointer-receiver mutation through the slice. adjustStepInterval takes qe := &queries[idx] and calls qe.SetStepInterval(...). The original explicitly wrote back req.CompositeQuery.Queries[idx].Spec = spec after
  each clamp because Go type assertions yield a copy. This refactor relies entirely on SetStepInterval correctly updating queries[idx].Spec for every spec variant — easy to get wrong silently for one variant. A
  unit test per query type asserting "after adjustStepInterval, the slice's spec actually reflects the new step" would protect against this.
  - steps map and missing-metric queries. In the new code, queries flagged in missingMetricQuerySet are continued before being added to queries/steps. Behavior matches the original. ✅
  - Dead code removed. traceOperatorQueries map in the original was populated but never read. Fine to drop.
  - event.QueryType last-wins behavior preserved. ✅
  - dormantMetricsWarningMsg plumbing. Verified it's still read at querier.go:359 and threaded into the response. ✅

  Code quality

  - Helper names and responsibilities are well-chosen. The new QueryRange body reads as a clear pipeline.
  - clampStep is a nice deduplication of five near-identical blocks.
  - resolveMetricMetadata doc comment is excellent — clearly states pre/postconditions, returns, and side effects on input.
  - secondsStep is a tiny helper used 5 times — fine, but consider just inlining if you want to keep helper count down.
  - Style nit: populateQBEvent is on *querier but does not use the receiver. Free function would suffice. Same for constructTraceOperatorDependencyMap. (Method form is fine if you anticipate future use of
  q.logger.)

  Tests

  The PR adds none. For a 250→257 line refactor of a hot path with ≥3 detected behavior drifts, that's the main risk. Minimum I'd want before merging:

  - clampStep boundaries: step=0 → recommended; step<min → clamped+warning; step==min → no change; step>min → no change.
  - adjustStepInterval per signal: trace, log, metric, meter, PromQL (no clamp), trace-operator. Verify the slice's spec is actually mutated.
  - resolveMetricMetadata: all aggregations missing → query name in returned slice; partial missing → spec patched, query absent from missing set; dormant vs. never-seen classification → correct error vs. warning.
  - populateQBEvent: composite query with one filtered + one unfiltered builder query → assert FilterApplied=true (locks in the new accumulating semantics).

  Performance

  adjustStepInterval precomputing six bounds per call is a non-issue — they're cheap arithmetic, not store calls. Don't lazify.

  Summary

  The refactor is a clear readability win. Before merging:
  1. Decide on the meter-warning drift (item 3) — bug fix or regression? Update PR description either way.
  2. Acknowledge the FilterApplied/GroupByApplied and TracesUsed semantic changes in the PR body, or revert them.
  3. Verify QueryEnvelope.SetStepInterval mutates PromQuery.Step specifically — this is the one place the field name differs and it's easy to get wrong.
  4. Add at least the clampStep and resolveMetricMetadata unit tests — this is the right moment, since the helpers are now isolated.

pkg/querier/querier.go
```diff
@@ -155,169 +142,117 @@ func (q *querier) QueryRange(ctx context.Context, orgID valuer.UUID, req *qbtype
...
 
-					if spec.StepInterval.Seconds() < float64(querybuilder.MinAllowedStepIntervalForMeter(req.Start, req.End)) {
-						newStep := qbtypes.Step{
-							Duration: time.Second * time.Duration(querybuilder.MinAllowedStepIntervalForMeter(req.Start, req.End)),
-						}
-						spec.StepInterval = newStep
-					}
+// clampStep sets the step to recommended when zero and clamps to min when below it.
+// When clamped and warn is true, a warning is appended for the user.
+func clampStep(qe *qbtypes.QueryEnvelope, recommended, min uint64, warnings *[]string) {
```
nityanandagohain: This is what we have started following in organising a file

```
var/const

types

exported functions/methods

private functions/methods
```

nityanandagohain (commented):
looks good apart from one suggestion on organising the functions

pkg/querier/querier.go
tushar-signoz: Done

pkg/querier/querier.go
```diff
@@ -492,6 +242,166 @@ func (q *querier) QueryRange(ctx context.Context, orgID valuer.UUID, req *qbtype
...
+// Returns:
+//   - missingMetricQueries: names of queries whose every aggregation was
+//     missing. Used downstream to preseed empty result placeholders so the
+//     response still has an entry per requested query name.
+//   - dormantWarning: a human-readable warning describing metrics that exist in
+//     the store but produced no data within the query window. Empty when no
+//     such metrics are present.
+//   - err: NotFound when one or more referenced metrics have never been seen,
+//     or Internal when a metadata fetch fails.
+func (q *querier) resolveMetricMetadata(ctx context.Context, queries []qbtypes.QueryEnvelope, start, end uint64) (missingMetricQueries []string, dormantWarning string, err error) {
```
tushar-signoz, line 304: Hey @namanv3 
You need to review this part. Nothing new here, it is all lift and shift, ensure that the doc string capturing the function is correct and function is supposed to do things mentioned in the doc string.

pkg/querier/querier.go
```diff
@@ -492,6 +242,166 @@ func (q *querier) QueryRange(ctx context.Context, orgID valuer.UUID, req *qbtype
...
+		if len(presentAggregations) == 0 {
+			missingMetricQueries = append(missingMetricQueries, spec.Name)
+			continue
+		}
+		spec.Aggregations = presentAggregations
+		queries[idx].Spec = spec
+	}
+
+	if len(missingMetrics) == 0 {
+		return missingMetricQueries, "", nil
```
namanv3, line 369: nit: `nil` instead of `missingMetricQueries`, makes it easier to read and immediately know that nil is being sent.

namanv3 (approved):

tushar-signoz:
@srikanthccv Need your review

pkg/querier/querier.go
```diff
@@ -492,6 +242,166 @@ func (q *querier) QueryRange(ctx context.Context, orgID valuer.UUID, req *qbtype
...
+				event.MetricsUsed = strings.Contains(sql, "signoz_metrics")
+				event.LogsUsed = strings.Contains(sql, "signoz_logs")
+				event.TracesUsed = strings.Contains(sql, "signoz_traces")
+			}
+		}
+	}
+}
+
+// resolveMetricMetadata fetches metadata for every metric referenced by builder
+// metric-aggregation queries, patches each query's aggregations in place with
```
srikanthccv, line 283: @namanv3 I don't think this "patches each query's aggregations in place" claim is true. Any change done is local and wouldn't be seen by the caller.

namanv3: @srikanthccv the change is indeed observed by the caller. Tested the relevant flows to confirm and had breakpoints to confirm as well
@tushar-signoz maybe we should add some unit tests for this method to ensure that this doesn't break in the future.

srikanthccv: can you help me understand how the change is not local and visible to the caller?


let's not add unit tests, if we add unit test, it should survive the refactoring, i.e test should not be required to updated because of internal refactoring. Are there any existing integration tests that test this? If not, let's add them otherwise please highlight which existing integration tests are covering this.

namanv3: all updates in the function are via an index so that's why. And yes, there are integration tests: `test_histogram_p90_returns_warning_outside_data_window`, `test_non_existent_metrics_returns_404`

srikanthccv (approved):

therealpandey (approved):

tushar-signoz:
Will merge this 1 day after release

PR #10808, opened 2026-04-02
chore: add schema version specific validations

srikanthccv:
## Pull Request

follow up to https://github.com/SigNoz/signoz/pull/9872
Fixes https://github.com/SigNoz/engineering-pod/issues/4378
Fixes https://github.com/SigNoz/engineering-pod/issues/4299
Fixes https://github.com/SigNoz/engineering-pod/issues/4215

PR #11217, opened 2026-05-07
fix(rules): handle legacy LOG_BASED_ALERT value when patching a rule

krsoninikhil:
## Pull Request

---

### 📄 Summary
Fix the error encountered in disabling the alert which might have alert type value already stored that doesn't fit the currently used enum. See Bug Context heading below for more details.

#### Issues closed by this PR
https://github.com/SigNoz/engineering-pod/issues/4814

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
PR #10808 added `storedRule.Validate()` in `PatchRule`. The stored rule for some customers has `alertType: "LOG_BASED_ALERT"` (no 'S') — a value found in some customer tenants. `Validate()` rejects it, so any PATCH (including toggling `disabled`) fails.

#### Fix Strategy
After unmarshalling the stored rule from the DB, call `NormalizeAlertType()` which rewrites `LOG_BASED_ALERT → LOGS_BASED_ALERT`. This happens before the patch is merged, so if the caller explicitly sends the wrong value it still gets validated and rejected. The normalization is a one-liner on `PostableRule` in `api_params.go` with no impact on any other code path.

---

### 🧪 Testing Strategy

- Tests added/updated: no
- Manual verification: reproduced the error by patching a rule with the legacy alertType in the DB; confirmed the patch now succeeds and the corrected value is persisted

---

- Blast radius: only `PatchRule` (the PATCH `/api/v1|v2/rules/{id}` endpoint); `CreateRule` and `EditRule` are untouched
- Potential regressions: none
- Rollback plan: no schema or DB side effects

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Disabling or enabling an alert rule no longer fails for rules that have a legacy `LOG_BASED_ALERT` alert type stored in the database |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

`NormalizeAlertType` is intentionally placed **before** `json.Unmarshal(ruleStr, &storedRule)` so the patch JSON wins: if someone explicitly sends a bad `alertType` in the patch body, it overrides the normalization and `Validate()` rejects it as expected.

### DIFF
--- a/pkg/query-service/rules/manager.go
+++ b/pkg/query-service/rules/manager.go
@@ -952,6 +952,7 @@ func (m *Manager) PatchRule(ctx context.Context, ruleStr string, id valuer.UUID)
 		m.logger.ErrorContext(ctx, "failed to unmarshal rule from db", slog.String("rule.id", id.StringValue()), errors.Attr(err))
 		return nil, err
 	}
+	storedRule.NormalizeAlertType()
 
 	if err := json.Unmarshal([]byte(ruleStr), &storedRule); err != nil {
 		m.logger.ErrorContext(ctx, "failed to unmarshal patched rule with given id", slog.String("rule.id", id.StringValue()), errors.Attr(err))
--- a/pkg/types/ruletypes/api_params.go
+++ b/pkg/types/ruletypes/api_params.go
@@ -36,6 +36,15 @@ func (AlertType) Enum() []any {
 	}
 }
 
+// NormalizeAlertType corrects known legacy alert type values that were stored
+// before strict validation was introduced. The corrected value is persisted on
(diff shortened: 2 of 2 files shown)


nityanandagohain (commented):
We should fix it at the source instead of handling it here. So that `LOG_BASED_ALERT` is not even generated, or if generated they are migrated to the new `LOGS_BASED_ALERT`

PR #11222, opened 2026-05-07
feat: migrate from lucide react and antD icons to signoz icons

tewarig:
##   Why does this change exist?                                                                                                                                         
  
  The frontend currently pulls icons from three different packages — @ant-design/icons, lucide-react, and @signozhq/icons — which means three different icon styles,  
  three different sizing/prop conventions, and three sets of bundle weight for what should be one design-system primitive. This PR consolidates everything onto     
  @signozhq/icons, so SigNoz's own icon package becomes the single source of truth across the app.                                                                    
                                                                                                                                                                    
  This is the right approach because @signozhq/icons is the package the design system already publishes for SigNoz; standardizing on it keeps icon visuals consistent 
  with the rest of the design tokens, removes two third-party dependencies, and gives us one place to add/version new glyphs going forward.
                                                                                                                                                                      
                                           
                                                                                                                                                                    
 ## Bundle Size - 
 
 
 before - 
 
 
<img width="1728" height="1083" alt="image" src="https://github.com/user-attachments/assets/9c4c57c4-81e8-428b-a0b3-5059b94a01c9" />

  
 after - 
 
<img width="641" height="346" alt="image" src="https://github.com/user-attachments/assets/0c4f87b1-ecfd-46da-8c27-066065fda0ae" />
                                                                                                 
                                                                                                                                                                    
                                                                                                                                       
  ---
##  🧪 Testing Strategy                                                                                                                                                 
                                                                                                                                                                      
  Tests added/updated:
  - Added a Jest module mock at frontend/__mocks__/signozhqIconsMock.tsx and remapped ^@signozhq/icons$ to it in jest.config.ts (replacing the old mapping that       
  pointed directly at the package's ESM build, which doesn't play nicely with Jest).                                                                                  
  - Updated several component tests to assert on accessible labels (aria-label, role) and data-testid instead of brittle CSS classnames or AntD icon-specific
  selectors — these now survive the icon-package swap.                                                                                                                
                                                                                                                                                                      
 ## Manual verification:
  - Walked through the surfaces touched by the migration (dashboards, pipelines page incl. change history, traces/logs filters, query builder, header right section,  
  workspace-locked page, copy-to-clipboard, tabs) and confirmed icons render at the expected size/weight and click handlers still fire.                               
  - Verified focus behavior on LabelsInput after the a11y tweak.                                                                       
                                                                                                                                                                      
  Edge cases covered:                                                                                                                                                 
  - Icon size prop differences between packages 
  - Buttons whose labels needed to be wrapped in Flex to keep alignment after the icon swap.                                                                          
  - Tests that previously matched on lucide/AntD class names now match on accessible labels.                                                                        
                                                                                                                                                                      
  ---                                                                                                                                                                 
 ##  ⚠️  Risk & Impact Assessment                                                                                                                                         
                                                                                                                                                                      
  Blast radius: Wide but shallow — touches ~450 files across the frontend, but each change is a like-for-like icon swap . 
  
  Could be the case that icon is not rendered / looks different. 
  
  
   No   
  backend behavior, no API contract, no data-model change.                                                                                                            
                                                                                                                                                                    
 ##  Potential regressions:                                                                                                                                              
  - Visual: a missing/misnamed icon mapping, or a glyph that renders at a slightly different size/stroke weight than the lucide or AntD original.                   
  - Functional: subtle prop/type drift between packages — refs, click event targets, or size semantics — in a hot path that's only reachable manually.                
  - Test stability: any test still relying on lucide/AntD-specific selectors will break; we converted the ones we found, but stragglers may surface in CI.
                                                                                                                                                                      
  Rollback plan: Revert the PR. The migration is self-contained on the frontend; no migrations, feature flags, or persisted state to unwind.                          
                                                                                                                                                                      
  ---                                                                                                                                                                 
  
  ---                                                                                                                                                                 
##   📋 Checklist                                                                                                                                                      
                                                                                                                                                                      
  - Tests added or explicitly not required
  - Manually tested                                                                                                                                                   
  - Breaking changes documented (none — internal-only)                                                                                                                
  - Backward compatibility considered (N/A — no external surface)                                                                                                     
                                                                                                                                                                      
  ---                                                                                                                                                                 
 ## 👀 Notes for Reviewers                                                                                                                                            
                                                                                                                                                                      
  ▎ Medium Risk — Broad UI refactor across many components to replace lucide-react / @ant-design/icons with @signozhq/icons, plus a dependency bump. Risk is mainly 
  ▎ visual/icon regressions and subtle prop/type differences (sizes, refs, event targets) across widely-used views.                                                   
                                                                                                                                                                    
##   Overview                                                                                                                                                            
                                                                                                                                                                    
  Migrates the frontend icon system by replacing lucide-react and @ant-design/icons usage across many components/containers with @signozhq/icons (including updated   
  icon names, sizing props, and small layout tweaks like wrapping button labels with Flex).                                                                         
                                                                                                                                                                      
  Updates tooling/tests to support the new icon package:                                                                                                              
  - Bumps @signozhq/icons 0.1.0 → 0.4.0.
  - Drops @ant-design/icons@4.8.0 and lucide-react@0.498.0 from package.json.                                                                                         
  - Adds a Jest module mock (__mocks__/signozhqIconsMock.tsx) and remaps ^@signozhq/icons$ to it (replacing the old mapping to the package's ESM build).            
  - Adjusts several tests to assert via accessible labels / data-testid instead of CSS classnames or AntD icon selectors.                                             
  - Includes small a11y/UX tweaks such as adding aria-labels and ensuring focus behavior in LabelsInput.                                                              
                                                                                                                                                                      
  ## Suggested review order                                                                                                                                              
                                                                                                                                                                      
  1. frontend/package.json + frontend/jest.config.ts + frontend/__mocks__/signozhqIconsMock.tsx — the infra change.                                                   
  2. A representative component swap  to sanity-check the prop-mapping pattern.                                                 
  3. Test changes — confirm new selectors match the user-facing intent rather than the old DOM shape.                                                                 
  4. Skim the rest as bulk swaps.


frontend/src/container/DashboardContainer/DashboardSettings/DashboardVariableSettings/index.tsx
```diff
@@ -58,11 +66,12 @@ function TableRow({ children, ...props }: RowProps): JSX.Element {
 						key: 'name-with-drag',
 						children: (
 							<div className="variable-name-drag">
-								<HolderOutlined
-									ref={setActivatorNodeRef}
+								<GripVertical
+									ref={setActivatorNodeRef as unknown as React.Ref<SVGSVGElement>}
```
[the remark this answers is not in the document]
tewarig: seems to work

frontend/src/components/CeleryOverview/CeleryOverviewTable/CeleryOverviewTable.tsx
```diff
@@ -107,9 +108,11 @@ const getColumnSearchProps = (
 					type="primary"
 					size="small"
 					onClick={(): void => handleSearch(selectedKeys as string[], confirm)}
-					icon={<SearchOutlined />}
 				>
-					Search
+					<Flex align="center" gap={4}>
```
AshwinBhatkal, line 112: nit: avoid the antd component here

frontend/src/components/HeaderRightSection/__tests__/HeaderRightSection.test.tsx
```diff
@@ -210,7 +196,7 @@ describe('HeaderRightSection', () => {
 
 		render(<HeaderRightSection {...defaultProps} />);
 
-		const feedbackButton = document.querySelector('.lucide-square-pen');
+		const feedbackButton = screen.queryByRole('button', { name: /feedback/i });
```
AshwinBhatkal, line 199: suggestion: This usage is mixed up in some places. use queryByRole to assert the absence of an element while use getByRole for asserting presence

frontend/src/components/MessagingQueueHealthCheck/AttributeCheckList.tsx
```diff
@@ -141,6 +148,10 @@ function treeTitleAndKey({
 	};
 }
 
+function renderSwitcherIcon(props: { expanded?: boolean }): JSX.Element {
```
AshwinBhatkal, line 144: question to self: is this used somewhere else

frontend/src/components/Tags/Tags.tsx
```diff
@@ -1,8 +1,7 @@
 import React, { Dispatch, SetStateAction, useState } from 'react';
-import { PlusOutlined } from '@ant-design/icons';
-import { Button, Tag } from 'antd';
+import { Check, Plus, X } from '@signozhq/icons';
+import { Button, Flex, Tag } from 'antd';
```
AshwinBhatkal, line 3: nit: avoid

frontend/src/container/AllError/index.tsx
```diff
@@ -4,10 +4,11 @@ import { useQueries } from 'react-query';
 // eslint-disable-next-line no-restricted-imports
 import { useSelector } from 'react-redux';
 import { Link, useLocation } from 'react-router-dom';
-import { SearchOutlined } from '@ant-design/icons';
+import { Search } from '@signozhq/icons';
 import {
 	Button,
 	Card,
+	Flex,
```
AshwinBhatkal: nit: avoid

frontend/src/container/BillingContainer/BillingContainer.tsx
```diff
@@ -448,10 +448,12 @@ export default function BillingContainer(): JSX.Element {
 							loading={isLoadingBilling || isLoadingManageBilling}
 							disabled={isLoading || isFetchingBillingData}
 							onClick={handleCsvDownload}
-							icon={<CloudDownloadOutlined />}
 							className="periscope-btn"
 						>
-							Download CSV
+							<Flex align="center" justify="center" gap={4}>
```
AshwinBhatkal, line 455: nit: seems like already there. avoid if possible

frontend/src/container/CreateAlertV2/CreateAlertHeader/LabelsInput.tsx
```diff
@@ -154,14 +162,16 @@ function LabelsInput({
 			) : (
 				<div className="labels-input__input-container">
 					<input
+						ref={inputRef}
+						// oxlint-disable-next-line jsx_a11y/no-autofocus
```
AshwinBhatkal: we can remove/ turn off this rule.

frontend/src/container/DashboardContainer/DashboardSettings/DashboardVariableSettings/index.tsx
```diff
@@ -11,14 +11,22 @@ import {
 import { restrictToVerticalAxis } from '@dnd-kit/modifiers';
 import { arrayMove, SortableContext, useSortable } from '@dnd-kit/sortable';
 import { CSS } from '@dnd-kit/utilities';
-import { Button, Modal, Row, RowProps, Space, Table, Typography } from 'antd';
+import {
+	Button,
+	Flex,
```
AshwinBhatkal: nit: avoid

frontend/src/container/FormAlertRules/BasicInfo.tsx
```diff
@@ -254,11 +254,13 @@ function BasicInfo({
 					>
 						<Button
```
AshwinBhatkal, line 255: suggestion: we should add an IconButton component to our library

frontend/src/container/MetricsExplorer/Inspect/constants.ts
```diff
@@ -26,18 +18,14 @@ export const METRIC_TYPE_TO_COLOR_MAP: Record<MetrictypesTypeDTO, string> = {
...
-		Omit<LucideProps, 'ref'> & RefAttributes<SVGSVGElement>
-	>
-> = {
-	[MetrictypesTypeDTO.gauge]: Gauge,
-	[MetrictypesTypeDTO.histogram]: BarChart2,
-	[MetrictypesTypeDTO.sum]: Diff,
-	[MetrictypesTypeDTO.summary]: BarChartHorizontal,
-	[MetrictypesTypeDTO.exponentialhistogram]: BarChart,
-};
+export const METRIC_TYPE_TO_ICON_MAP: Record<MetrictypesTypeDTO, typeof Gauge> =
```
AshwinBhatkal, line 23: is there not type for icon being exported from the library? We should use that one here ideally instead of Gauge

frontend/src/pages/Services/Metrics.test.tsx
```diff
@@ -12,7 +12,7 @@ describe('Services', () => {
 		expect(screen.getByTestId('resource-environment-filter')).toBeInTheDocument();
 
 		const application = screen.getByRole('columnheader', {
-			name: /application search/i,
+			name: /application/i,
```
AshwinBhatkal, line 15: we changed the name here?

H4ad (dismissed):
We can address the suggestions by Ashwin but it's in a good state that we can ship and adress them in follow-up PRs to not block this for too long since this conflicts with any PR merged

frontend/src/container/MetricsExplorer/Inspect/constants.ts
tewarig: we don't export type as of now. will add a todo and create an issue.

frontend/src/components/CeleryOverview/CeleryOverviewTable/CeleryOverviewTable.tsx
tewarig: we can have our own flex / box component later which will be a drop in replacement

frontend/src/pages/Services/Metrics.test.tsx
tewarig: So , SearchOutlined from @ant-design/icons renders with aria-label="search" (antd icons set role="img" + an aria-label derived from the icon   
  name). That aria-label became part of the column header's accessible name, so the header's name was effectively "Application search" —  
  which is why the original test matched /application search/i.                                                                             
                                                                                                                                          
  @signozhq/icons' Search doesn't expose that aria-label, so the column header's accessible name is now just "Application".

H4ad (approved):

makeavish (approved):

YounixM (approved):

AshwinBhatkal (approved):

PR #11224, opened 2026-05-07
test(span-login): split huge file tests in smaller ones

H4ad:
## Pull Request

---

Working in progress...

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---


frontend/src/container/Login/__tests__/Login.test-utils.ts
```diff
@@ -0,0 +1,172 @@
+/**
```
therealpandey, line 1: @H4ad would you be able to migrate this to use the openapi spec? A lot of these files might vanish if the openapi generated service is integrated.

It also helps the org move faster as someone is going to integrate openapi eventually.

H4ad: Do we have a epic issue to migrate to openapi? I can create a task there and pick for myself.

This whole split of files is just an experimentation to see if runs faster on our CI but looks like not :/ This only helps to run faster locally with multiple cores.

PR #11228, opened 2026-05-08
feat: extend error fields

tushar-signoz:
## Summary
Extend `pkg/errors` with structured retry strategy and attribute metadata, and wire both into the HTTP error response. Existing constructors (`NewInvalidInputf`, `NewTimeoutf`, …) now declare appropriate retry policies automatically; `render.Error` emits a standard `Retry-After` header when the error carries an explicit delay.

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -4,12 +4,12 @@ import (
 	"errors" //nolint:depguard
 	"fmt"
 	"log/slog"
+	"time"
 
 	"go.opentelemetry.io/otel/attribute"
 )
 
 // base is the fundamental struct that implements the error interface.
-// The order of the struct is 'TCMEUAS'.
 type base struct {
 	// t denotes the custom type of the error.
 	t typ
@@ -22,9 +22,16 @@ type base struct {
 	// u denotes the url for the documentation (if present) for the error.
 	u string
 	// a denotes any additional error messages (if present).
+	// NOTE: use attrs['suggestions'] for additional structured suggestions instead of this field.
 	a []string
 	// s contains the stacktrace captured at error creation time.
 	s fmt.Stringer
+	// r is the retry strategy for the error, if applicable.
+	r *retry
+	// suggestions is a list of user-facing suggestions related to the error, if present.
+	suggestions []string
+	// invalidReferences is a list of references that were invalid and contributed to the error, if present.
+	invalidReferences []string
 }
 
 // Stacktrace returns the stacktrace captured at error creation time, formatted as a string.
@@ -39,13 +46,16 @@ func (b *base) Stacktrace() string {
 // and returns a new base error.
 func (b *base) WithStacktrace(s string) *base {
 	return &base{
-		t: b.t,
-		c: b.c,
-		m: b.m,
-		e: b.e,
-		u: b.u,
-		a: b.a,
-		s: rawStacktrace(s),
+		t:                 b.t,
+		c:                 b.c,
+		m:                 b.m,
+		e:                 b.e,
+		u:                 b.u,
+		a:                 b.a,
+		s:                 rawStacktrace(s),
+		r:                 b.r,
+		suggestions:       b.suggestions,
+		invalidReferences: b.invalidReferences,
 	}
 }
 
@@ -113,13 +123,16 @@ func WithAdditionalf(cause error, format string, args ...any) *base {
 		s = original.s
 	}
 	b := &base{
-		t: t,
-		c: c,
-		m: m,
-		e: e,
-		u: u,
-		a: a,
-		s: s,
+		t:                 t,
+		c:                 c,
+		m:                 m,
+		e:                 e,
+		u:                 u,
+		a:                 a,
+		s:                 s,
+		r:                 retryOf(cause),
+		suggestions:       suggestionsOf(cause),
+		invalidReferences: invalidReferencesOf(cause),
 	}
 
 	return b.WithAdditional(append(a, fmt.Sprintf(format, args...))...)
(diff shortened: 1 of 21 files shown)


therealpandey:
@tushar-signoz you need to extend the extending errors pacakge with additional information.

You can't create a v2 errors altogether. Add whatever fields you need to the existing errors field and then add support for them in the frontend.

signoz-ai-assistant and zeus are using the same errors package so it can only be extended.

pkg/errors/errors.go
```diff
@@ -22,9 +23,15 @@ type base struct {
...
 	// a denotes any additional error messages (if present).
+	// NOTE: use attrs['suggestions'] for additional structured suggestions instead of this field.
 	a []string
 	// s contains the stacktrace captured at error creation time.
 	s fmt.Stringer
+	// r is the retry strategy for the error, if applicable.
+	r *retry
+	// attrs contains any additional attributes for the error, if present.
+	// attrs is declared to be map[string]any, however map[string][]string is expected to be used for values.
+	attrs map[string]any
```
therealpandey:
```suggestion
	suggestions map[string][]string
```

No need to bundle suggestions here in attrs. If there is a `WithSuggestions`, suggestions should come at the top level as `[]string`.

pkg/errors/errors.go
```diff
@@ -128,55 +139,199 @@ func WithAdditionalf(cause error, format string, args ...any) *base {
...
 
+// WithSuggestions replaces the list of suggestions on the base error.
+func (b *base) WithSuggestions(suggestions ...string) *base {
+	return b.setAttrs("suggestions", suggestions...)
+}
+
+// AddSuggestions appends to the existing list of suggestions on the base error.
+func (b *base) AddSuggestions(suggestions ...string) *base {
+	return b.addAttrs("suggestions", suggestions...)
+}
```
therealpandey:
```suggestion
func (b *base) WithSuggestions(suggestions ...string) *base {
	return b.setAttrs("suggestions", suggestions...)
}
```

Append like we do for Additional.

pkg/errors/errors.go
```diff
@@ -128,55 +139,199 @@ func WithAdditionalf(cause error, format string, args ...any) *base {
...
+
+// AddInvalidReference appends a single invalid reference to the existing list on the base error.
+func (b *base) AddInvalidReference(invalidReference string) *base {
+	return b.addAttrs("invalidReferences", invalidReference)
+}
+
+// AddInvalidReferences appends to the existing list of invalid references on the base error.
+func (b *base) AddInvalidReferences(invalidReferences ...string) *base {
+	return b.addAttrs("invalidReferences", invalidReferences...)
+}
```
therealpandey:
```suggestion
```

Can we have one example here where we would use `invalidReferences`?

pkg/errors/errors.go
```diff
@@ -128,55 +139,199 @@ func WithAdditionalf(cause error, format string, args ...any) *base {
...
+
+// WithWarnings replaces the list of warnings on the base error.
+func (b *base) WithWarnings(warnings ...string) *base {
+	return b.setAttrs("warnings", warnings...)
+}
+
+// AddWarnings appends to the existing list of warnings on the base error.
+func (b *base) AddWarnings(warnings ...string) *base {
+	return b.addAttrs("warnings", warnings...)
+}
```
therealpandey: This should ideally be a property of the success struct like it is for the querier.

"Hey, your request succeeded, but here are a list of warnings."

Otherwise, this is semantically no different from additional.

pkg/errors/errors.go
```diff
@@ -128,55 +139,199 @@ func WithAdditionalf(cause error, format string, args ...any) *base {
...
 // Otherwise, it returns TypeInternal, the original error string
-// and the error itself.
+// and the error itself. For non-base errors the retry pointer defaults to
+// &retry{policy: RetryNever} so callers always observe a "do not retry"
+// stance for arbitrary errors.
 //
-//nolint:staticcheck // ST1008: intentional return order matching struct field order (TCMEUA)
-func Unwrapb(cause error) (typ, Code, string, error, string, []string) {
+//nolint:staticcheck // ST1008: intentional return order matching struct field order (TCMEUARA)
+func Unwrapb(cause error) (typ, Code, string, error, string, []string, *retry, map[string]any) {
```
therealpandey:
```suggestion
func Unwrapb(cause error) (typ, Code, string, error, string, []string) {
```

We can hold off on making changes here as this changes a lot of places. We can do that in the PR which utilizes the new fields for the first time.

therealpandey (changes requested):

pkg/errors/retry.go
```diff
@@ -0,0 +1,31 @@
...
+type RetryPolicy string
+
+const (
+	RetryNever     RetryPolicy = "never"      // retry with the same inputs cannot succeed.
+	RetryImmediate RetryPolicy = "immediate"  // retry without waiting.
+	RetryBackoff   RetryPolicy = "backoff"    // caller picks its own backoff schedule.
+	RetryAfter     RetryPolicy = "after"      // honor Retry.After exactly (the producer knows the wait).
+	RetryAfterFix  RetryPolicy = "after_fix"  // retry pointless until the caller fixes the request.
+	RetryAfterAuth RetryPolicy = "after_auth" // retry pointless until the caller re-authenticates.
+)
```
therealpandey: Can remove this policy string for now.

pkg/errors/errors.go
```diff
@@ -128,29 +140,113 @@ func WithAdditionalf(cause error, format string, args ...any) *base {
...
+		a:                 a,
+		s:                 b.s,
+		r:                 b.r,
+		suggestions:       b.suggestions,
+		invalidReferences: b.invalidReferences,
+	}
+}
+
+// withRetry adds retry metadata to the base error and returns a new base error.
+func (b *base) withRetry(r retry) *base {
```
therealpandey, line 175:
```suggestion
func (b *base) WithRetry(r retry) *base {
```

pkg/errors/errors.go
```diff
@@ -25,6 +25,12 @@ type base struct {
 	a []string
 	// s contains the stacktrace captured at error creation time.
 	s fmt.Stringer
+	// r is the retry strategy for the error, if applicable.
+	r *retry
+	// suggestions is a list of user-facing suggestions related to the error, if present.
+	suggestions []string
+	// invalidReferences is a list of references that were invalid and contributed to the error, if present.
```
therealpandey, line 33: Can we have one example here?

therealpandey (changes requested):

pkg/errors/errors.go
tushar-signoz: There is exported `WithRetryAfter` method, this was meant to be internal

tushar-signoz: If need be, can be modified

therealpandey (approved):

AshwinBhatkal (approved):

PR #11229, opened 2026-05-08
fix(signup): add frontend password validation matching backend rules

### LINKED ISSUE #11111: [Bug]: Inconsistent and Restrictive Password Validation Logic (Backend & Frontend)
## Bug description

The password validation logic in SigNoz contains several technical flaws and UI inconsistencies that impact the onboarding experience:

1.  **Incorrect Length Calculation (Backend):** Uses `len(password)` instead of rune count. This allows multi-byte characters (emojis/non-Latin scripts) to satisfy the 12-character requirement with as few as 3-4 actual characters.
2.  **Over-restrictive Whitelist (Backend):** Rejects valid symbols like `;`, `'`, and ` ` (space) because they aren't in a hardcoded list.
3.  **Internal Generator Conflict (Backend):** The random generator can produce semicolons, which the validator then rejects, potentially causing **panics** in `MustGenerateFactorPassword`.
4.  **Stale Error Messages (Frontend):** Backend validation errors (e.g., "invalid password") are not cleared when the user modifies the password field. 

Image

5.  **Broken Error Formatting:** Error messages display Go-style slice formatting instead of a human-readable list.

Image

## Expected behavior

*   Password length should be calculated based on character count (`utf8.RuneCountInString`), not bytes.
*   The character whitelist should allow all printable characters (including `;` and spaces).
*   The frontend should clear stale backend errors when the user types.
*   Error messages should be properly formatted for end-users.

## How to reproduce

1.  **Reproduce Length Bug:** Enter `🚀🚀🚀Aa1` as a password. Observe that it is accepted despite being only 6 characters long.
2.  **Reproduce Symbol Bug:** Enter a 12+ character password containing a semicolon (e.g., `Admin1234567;`). Observe the "invalid password" rejection.
3.  **Reproduce UI Bug:** 
    *   Submit a short password (`admin`) to trigger a backend error alert.
    *   Correct the password to be 12+ characters long. 
    *   Observe that the red error alert remains visible even after the "Access My Workspace" button becomes enabled.

## Version information 
* **Signoz version**: v0.120.0
* **Browser version**: Brave
* **Your OS and version**: macOS

## Additional context

The core logic is located in `pkg/types/factor_password.go`. The issue with the root user startup loop is also related, where the `query-service` can get stuck in a retry loop if a simple root password is provided in the configuration.

AshutoshSharma-pixel:
## Problem
The signup form had no password validation on the frontend — it only checked 
that the field was non-empty. The backend enforces strict rules via 
`IsPasswordValid()` in `pkg/types/factor_password.go`:
- Minimum 12 characters
- At least one uppercase letter [A-Z]
- At least one lowercase letter [a-z]
- At least one number [0-9]
- At least one symbol from the allowed set

This caused a poor UX where users would fill out the entire form, hit submit, 
and only then receive a cryptic backend error about invalid password.

## Fix
Added `validatePassword()` in `SignUp.tsx` that mirrors the backend rules exactly.
Users now get clear inline feedback on the password field before submitting.

## Changes
- Added `validatePassword()` function matching backend `IsPasswordValid()` logic
- Added `passwordError` derived state shown inline on the password field
- Updated `isValidForm` to require valid password before enabling submit button

Fixes #11111

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes signup form validation and could block some passwords client-side (e.g., non-ASCII letters/numbers) if the frontend rules diverge from backend expectations.
> 
> **Overview**
> Adds inline password policy validation to the signup form via a new `validatePassword()` helper (min length, required character classes, allowed symbol set, and invalid-character rejection), surfacing errors directly on the password field.
> 
> Updates submit gating (`isValidForm`) to require both password match and a passing password-policy check before enabling account creation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 809f9b05bb40f3226bfa66e1409c8c77552555e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/frontend/src/pages/SignUp/SignUp.tsx
+++ b/frontend/src/pages/SignUp/SignUp.tsx
@@ -27,6 +27,39 @@ type FormValues = {
 	isAnonymous: boolean;
 };
 
+const validatePassword = (pwd: string): string | null => {
+	if (pwd.length < 12) {
+		return 'Password must be at least 12 characters long';
+	}
+
+	const symbols = '~!@#$%^&*()_+`-={}|[]\\:"<>?,./';
+	let hasUpperCase = false;
+	let hasLowerCase = false;
+	let hasNumber = false;
+	let hasSymbol = false;
+
+	for (let i = 0; i < pwd.length; i++) {
+		const char = pwd[i];
+		if (char >= 'a' && char <= 'z') {
+			hasLowerCase = true;
+		} else if (char >= 'A' && char <= 'Z') {
+			hasUpperCase = true;
+		} else if (char >= '0' && char <= '9') {
+			hasNumber = true;
+		} else if (symbols.includes(char)) {
+			hasSymbol = true;
+		} else {
+			return 'Password contains invalid characters';
+		}
+	}
+
+	if (!hasUpperCase || !hasLowerCase || !hasNumber || !hasSymbol) {
+		return 'Password must contain at least one uppercase letter, one lowercase letter, one number, and one symbol';
+	}
+
+	return null;
+};
+
 function SignUp(): JSX.Element {
 	const [loading, setLoading] = useState(false);
 	const [confirmPasswordTouched, setConfirmPasswordTouched] = useState(false);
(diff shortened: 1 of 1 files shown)


vikrantgupta25:
@AshutoshSharma-pixel thanks for the contribution but we do not want to couple frontend validation with backend validation. There should only be one source of truth for the validation. Rather than hardcoding the same on frontend it should be config driven via backend.

AshutoshSharma-pixel:
Thanks for the feedback @vikrantgupta25 understood  the validation should be config-driven rather than duplicated on the frontend. Happy to rework this to fetch the password policy from the backend and apply it dynamically if you can point me to where that config should live or if there's an existing endpoint for it.

PR #11234, opened 2026-05-08
feat: [traces] time aware dynamic field mapper

nityanandagohain:
## Pull Request

---

### 📄 Summary
This changes adds support for time aware dynamic fields in traces, the implementation is already merged for logs https://github.com/SigNoz/signoz/pull/9669

Things to note:
1) The evolution code in field_mapper.go is moved to types, which is generic for different signals.
2) Statement builder is updated w.r.t when we fetch the metadata for keys.
3) Trace operator statement builder was behind in logic for adjust keys, updated it just enough for this PR. It needs more work and will be addressed separately.
4) Integration tests is similar to logs.

#### Issues closed by this PR
Fixes https://github.com/SigNoz/engineering-pod/issues/4522
---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: ✅ 
- Manual verification: ✅ 
- Edge cases covered: ✅ 

---

- Blast radius: Traces
- Potential regressions:
- Rollback plan: Revert

---

## 👀 Notes for Reviewers

* Moving out the function to types from logs field_mapper and then adding the same changes to traces to support it.
---


pkg/telemetrytraces/trace_operator_cte_builder.go
```diff
@@ -936,3 +914,16 @@ func (b *traceOperatorCTEBuilder) aggOrderBy(k qbtypes.OrderBy) (int, bool) {
 	}
 	return 0, false
 }
+
+func (b *traceOperatorCTEBuilder) adjustKeys(keys map[string][]*telemetrytypes.TelemetryFieldKey) {
+	// todo: this needs to be updated w.r.t trace statement builder.
```
srikanthccv: please update

pkg/telemetrymetadata/metadata_test.go
srikanthccv, line 1: remove this test file entirely @nityanandagohain , this is of no real use. We have enough covered with other unit tests and integration tests.

tests/fixtures/traces.py
```diff
@@ -575,7 +580,7 @@ def np_arr(self) -> np.array:
                 self.db_operation,
                 self.has_error,
                 self.is_remote,
-                self.resources_string,
+                self.resource_json,
```
srikanthccv, line 583: Why is `self.resources_string` replaced with `self.resource_json` here? Shouldn't this be based on the `resource_write_mode`. And when dual mode is enabled, it should be ingesting both?

pkg/telemetrytraces/statement_builder.go
```diff
@@ -124,6 +117,14 @@ func (b *traceQueryStatementBuilder) Build(
 		-------------------------------- End of tech debt ----------------------------
 	*/
 
+	// since we are modifying the selectFields, they might include keys which need evolutions so we should get keys after that.
```
srikanthccv: depending whether this PR gets merged or https://github.com/SigNoz/signoz/pull/10583 first, this can be removed, because we would expect the evolution for the intrinsic fields `"timestamp", "span_id", "trace_id"`, however, in case of 0 selected fields we are using default fields which includes `service.name` which has the evolution.

tests/integration/tests/querier/13_traces_resource_evolution.py
```diff
@@ -0,0 +1,240 @@
...
+    build_logs_aggregation,
+    index_series_by_label,
+    make_query_request,
+)
+from fixtures.traces import TraceIdGenerator, Traces
+
+
+# we already create the evolution for resource during schema migration
+# since we have to create test data around it, we need to get the evolution time
+def _get_traces_resource_evolution_time_json(signoz: types.SigNoz) -> datetime:
```
srikanthccv: move this to fixtures and update `tests/integration/tests/querier/12_logs_resource_evolution.py` as well.

tests/fixtures/traces.py
srikanthccv: nvm, I see that `self.resources_string` is already there above in the array.

tests/integration/tests/querier/13_traces_resource_evolution.py
```diff
@@ -0,0 +1,240 @@
...
+    results = response.json()["data"]["data"]["results"]
+    assert len(results) == 1
+
+    aggregations = results[0]["aggregations"]
+    assert len(aggregations) == 1
+
+    return index_series_by_label(aggregations[0]["series"], group_by)
+
+
+def _assert_grouped_series(
```
srikanthccv: this one as well, please move the fixtures.

pkg/telemetrytraces/trace_operator_cte_builder.go
nityanandagohain: creating a new PR to address this and will get that merged first.

pkg/telemetrytraces/statement_builder.go
nityanandagohain: Let it be as it is in this pr ?

srikanthccv: 👍

srikanthccv (approved):

therealpandey (approved):

PR #11236, opened 2026-05-08
test(flaky): try reduce flakyness by increasing timeout

H4ad:
## Pull Request

---

These tests depends heavily on userEvent, waitFor, etc... each interaction is a little heavy and the timeout for each one of them is usually 1s (waitFor), so it will be slower if the machine is not stronger enough.

The alternative is to rewrite the test to use fireEvent (less real), or try reduce the test size (not so good as well)

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [x] 🧪 Test-only


YounixM (approved):

H4ad:
Likely fixed by #11237, closing this.

PR #11243, opened 2026-05-09
Fix: dangling and syntax in dashboard filter ( #10674)

### LINKED ISSUE #10674: [bug] adding a dashboard variable via drilldown to APM metrics dashboard breaks the panels

rinkydevi:
## Pull Request

---

### 📄 Summary
  This PR fixes two independent bugs introduced when dashboard variables are created or removed in panels that      
  reference them via filter expressions.                                                                            
                                                                                                                    
  Bug 1 — Dangling AND in parenthesized filter expressions                                                          
  When a dashboard variable is deleted and its corresponding filter clause is removed from a query expression like:
  (deployment.environment = $deployment.environment AND service.name = $service.name AND operation IN               
  $top_level_operation)                                                                                     
  the expression degrades to (... AND ) — a trailing AND with no right-hand operand — breaking the panel query      
  entirely.      

  Bug 2 — Duplicate variable names accepted silently via drilldown                                                  
  Clicking a field value in APM/metrics drilldown to create a dashboard variable had no duplicate-name guard. A
  variable with the same name as an existing one could be created without any error, leading to silent conflicts    
  across panels referencing that name.                                                                      
      

Bug 1 resolved, dangling AND in query


https://github.com/user-attachments/assets/49d3d168-1bee-4c69-92c3-b456e51e1061


Bug 2 resolved , adding duplicate variable

https://github.com/user-attachments/assets/1cfd2d9b-2df5-454e-85cf-2c3c7927398e








#### Issues closed by this PR
Closes #10674

---

### ✅ Change Type
_Select all that apply_

- [x] 🐛 Bug fix

---

### 🐛 Bug Context
Root Cause
                                                                                                                    
  Bug 1: The removeKeysFromExpression utility used string scanning with a regex to consume the conjunction that     
  followed a removed clause. When the removed clause was the last item before a ), no forward conjunction existed — 
  leaving AND ) in the output. Post-processing regexes were not reliable for expressions containing BETWEEN x AND y 
  (whose internal AND is structurally different from a clause separator) or EXISTS predicates.              

  Bug 2: Variable creation via drilldown called createVariable in useDashboardVariableUpdate.ts directly, which had 
  no name-uniqueness check. The Settings UI path (onVariableSaveHandler) had validation, but the drilldown path was
  completely unguarded.                                                                                             
                                                                                                            
  Fix Strategy

  Bug 1: Replaced the regex approach with an ANTLR parse-tree visitor over the existing                             
  FilterQueryParser/FilterQueryLexer infrastructure. Each ComparisonContext node is evaluated atomically — if its
  key is in the removal set, the whole node is dropped. Surviving siblings are rejoined with the correct operator at
   each level of the tree. Empty groups collapse to null and are omitted from parent joins.                 

  Bug 2: Added a duplicate-name guard at the top of createVariable. If Object.values(currentVariables).some(v =>    
  v.name === name) is true, an error notification is shown and the function returns early before writing anything to
   the dashboard state.      

---

### 🧪 Testing Strategy
 - Tests added/updated: QueryBuilderV2/utils.test.ts — replaced regex-era tests with semantically correct ANTLR
  tests covering: AND/OR precedence, BETWEEN x AND y, EXISTS, nested parenthesized groups, and multiple variable    
  references in one expression. All 94 tests pass.
  - Manual verification: Reproduced both bugs locally against an APM metrics dashboard with multiple variables;     
  confirmed both are resolved after the fix.                                                                        
  - Edge cases covered: BETWEEN internal AND not treated as conjunction; EXISTS predicate removed cleanly;
  expression with only one clause inside parens collapses correctly; mixed AND/OR precedence preserved.             
    

---

 - Blast radius: removeKeysFromExpression is called whenever a dashboard variable is removed from a panel filter
  expression. Change is isolated to that utility and its direct callers. The createVariable guard is additive (early
   return) with no side effects on the happy path.
  - Potential regressions: Any caller that previously relied on the regex path producing a particular (incorrect)   
  output for malformed expressions may see different behavior — but the new output is correct per the grammar.      
  - Rollback plan: Revert this PR. No schema or API changes; no migrations required.
     

---
### 📝 Changelog

| Field | Value |
|---|---|
| Change Type | Bug Fix |
| Description | Fixed dangling `AND` operator in panel filter expressions when a dashboard variable is removed, and blocked duplicate variable names from being created via the drilldown path. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers
  - The ANTLR grammar (FilterQuery.g4) and generated parser/lexer were already present in the codebase — this PR
  adds no new grammar or generated files, only a consumer of the existing parse tree.  
  - Bug 2 fix is a single guard (9 lines added) in useDashboardVariableUpdate.ts. No UI changes. 

---

<!-- CURSOR_SUMMARY -->
> [!NOTE]
> **Medium Risk**
> Replaces string/regex-based filter clause removal with ANTLR parse-tree rewriting, which changes behavior for complex/malformed expressions and could affect dashboards relying on prior output. Duplicate-variable validation is low risk but touches variable creation UX.
> 
> **Overview**
> Fixes dashboard panel filter breakage when removing variables by rewriting `removeKeysFromExpression` to drop matching clauses via the existing ANTLR `FilterQuery` parse tree, correctly handling parentheses, AND/OR precedence, and operators like `BETWEEN`/`EXISTS` without leaving dangling conjunctions (returns original expression on parse errors).
> 
> Prevents creating dashboard variables with duplicate names via the drilldown path by adding a name-exists guard and showing a localized error notification; adds the new i18n string and updates `IDashboardVariable` typing (`dynamicVariablesWidgetIds`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea69fe31fe56cdd26a94cae6222a561cedf3d9f9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->


frontend/src/container/DashboardContainer/DashboardVariablesSelection/useDashboardVariableUpdate.ts
AshwinBhatkal: you can remove this instead of adding the type in the interface

frontend/src/container/DashboardContainer/DashboardVariablesSelection/useDashboardVariableUpdate.ts
```diff
@@ -231,7 +245,7 @@ export const useDashboardVariableUpdate =
 				const updatedVariables = convertVariablesToDbFormat(tableRowData);
 				updateVariables(updatedVariables, newVariable.id, [], false);
 			},
-			[dashboardData, updateVariables],
+			[dashboardData, updateVariables, notifications, t],
```
AshwinBhatkal: these 2 need not be added in the dependencies.

frontend/src/components/QueryBuilderV2/__tests__/utils.test.ts
```diff
@@ -1164,10 +1164,8 @@ describe('removeKeysFromExpression', () => {
 				true,
 			);
 
-			// Should remove one occurence as this case in itself is invalid to have multiple variable expressions for the same key
-			expect(result).toBe(
-				"deployment.environment = $env1 deployment.environment = 'default'",
-			);
+			// ANTLR removes every clause whose value contains $, non-variable clause is kept
```
AshwinBhatkal: is this controllable?

AshwinBhatkal: we don't want to always remove clause whose contains $

rinkydevi: Yes — controlled by removeOnlyVariableExpressions = true.

frontend/src/types/api/dashboard/getAll.ts
```diff
@@ -61,6 +61,7 @@ export interface IDashboardVariable {
 	defaultValue?: string;
 	dynamicVariablesAttribute?: string;
 	dynamicVariablesSource?: string;
+	dynamicVariablesWidgetIds?: string[];
```
AshwinBhatkal: we can remove this

frontend/src/container/DashboardContainer/DashboardVariablesSelection/useDashboardVariableUpdate.ts
```diff
@@ -177,6 +181,16 @@ export const useDashboardVariableUpdate =
 				// Get current dashboard variables
 				const currentVariables = dashboardData.data.variables || {};
 
+				const nameExists = Object.values(currentVariables).some(
+					(v) => v.name === name,
+				);
+				if (nameExists) {
+					notifications.error({
```
AshwinBhatkal: Let's use toast here.

frontend/src/components/QueryBuilderV2/__tests__/utils.test.ts
```diff
@@ -1199,6 +1201,107 @@ describe('removeKeysFromExpression', () => {
...
+		});
+
+		it('removes first clause without leaving a dangling AND', () => {
+			expect(
+				removeKeysFromExpression(
+					'(deployment.environment = $deployment.environment AND service.name = $service.name)',
+					['deployment.environment'],
+					true,
+				),
+			).toBe('(service.name = $service.name)');
```
AshwinBhatkal: can we have this without the parens?

frontend/src/components/QueryBuilderV2/__tests__/utils.test.ts
```diff
@@ -1199,6 +1201,107 @@ describe('removeKeysFromExpression', () => {
...
+			).toBe(
+				'(deployment.environment = $deployment.environment AND region = $region)',
+			);
+		});
+
+		it('drops the empty paren group when its only child is removed', () => {
+			// (a) OR (b) — removing a must not leave () OR (b = '2')
+			// The empty group is dropped; the remaining group keeps its parens (semantically equivalent)
+			expect(removeKeysFromExpression("(a = '1') OR (b = '2')", ['a'])).toBe(
+				"(b = '2')",
```
AshwinBhatkal: same as previous comment

frontend/src/components/QueryBuilderV2/__tests__/utils.test.ts
```diff
@@ -1199,6 +1201,107 @@ describe('removeKeysFromExpression', () => {
...
+		});
+
+		it('removes an EXISTS clause (no value token)', () => {
+			expect(removeKeysFromExpression("a = '1' AND b EXISTS", ['b'])).toBe(
+				"a = '1'",
+			);
+		});
+
+		it('removes a NOT EXISTS clause', () => {
+			expect(removeKeysFromExpression("a = '1' AND b NOT EXISTS", ['b'])).toBe(
```
AshwinBhatkal, line 1306: Can you also write tests of IN, NOT IN, NOT BETWEEN, CONTAINS, LIKE, NOT LIKE

frontend/src/components/QueryBuilderV2/__tests__/utils.test.ts
```diff
@@ -1199,6 +1201,107 @@ describe('removeKeysFromExpression', () => {
...
+				removeKeysFromExpression("a = '1' AND b = '2' OR c = '3' AND d = '4'", [
+					'b',
+				]),
+			).toBe("a = '1' OR c = '3' AND d = '4'");
+		});
+
+		it('preserves correct conjunctions when removing from a trailing AND group', () => {
+			// a OR b AND c OR d — removing c collapses second AND group to b
+			expect(
+				removeKeysFromExpression("a = '1' OR b = '2' AND c = '3' OR d = '4'", [
```
AshwinBhatkal, line 1226: Can you also verify - a AND b AND c OR d

frontend/src/components/QueryBuilderV2/__tests__/utils.test.ts
```diff
@@ -1155,7 +1155,7 @@ describe('removeKeysFromExpression', () => {
 	});
 
 	describe('Real-world scenarios', () => {
-		it('should handle multiple variable instances of same key', () => {
+		it('should remove at most one variable expression per key', () => {
 			const expression =
 				"deployment.environment = $env1 deployment.environment = $env2 deployment.environment = 'default'";
```
AshwinBhatkal, line 1160: Is there a test for something like this - 

((deployment.environment = $env1 OR deployment.environment = 'default') AND service.name = $svc1)

Can you verify the output when you remove service.name and deployment.environment separately?

AshwinBhatkal (approved):

rinkydevi:
while adding variable in the dashboard , the variable gets created but is not getting added to queries and include in labels according to the current workflow. as per my understanding this might be cause from the function `addTagFiltersToDashboard` which have condition as `(!applyToAll && widgetIds && !widgetIds.includes(widget.id)) ` and during our call the value `widgetIds = [] ` cause it call `updateAfterRemoval(widget, filter)` directly.

cause the flow to more like:
1. create variable metadata
2. cleanup/remove stale references
3. dashboard state update
4. later query application/rendering

cc: @AshwinBhatkal , this is discovery we made in our discussion

PR #10600, opened 2026-03-16
feat: deprecate old user invite apis

karanbalani:
## Pull Request

---

### 📄 Summary
- remove all the invite APIs used by the old invite flow

#### Issues closed by this PR
Closes: https://github.com/SigNoz/platform-pod/issues/1840
Also related to: https://github.com/SigNoz/platform-pod/issues/1827

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

PR #11249, opened 2026-05-10
fix(authn): redirect to login on invalid invitation token in signup

### LINKED ISSUE #7011: handle invalid invitation token gracefully in signup page

AshutoshSharma-pixel:
## Problem
When a user visits /signup?token=... with an invalid, used, or expired 
invitation token, the signup page is still shown. Any attempt to submit 
the form results in a confusing error with no clear guidance.

## Fix
Added a useEffect in SignUp.tsx that checks for a `token` query param 
on mount. If present, the user is immediately redirected to the login 
page with a clear error message:
"Your invite link is no longer valid. Please contact your admin."

This uses the existing `callbackauthnerr` + `message` query param 
pipeline that the Login page already handles.

## Changes
- `frontend/src/pages/SignUp/SignUp.tsx`: added redirect logic for 
  invalid/stale invite tokens

Fixes #7011


SagarRajput-7:
Hey @AshutoshSharma-pixel, 

Thanks for taking the time to contribute to SigNoz and for investigating issue [#7011](https://github.com/SigNoz/signoz/issues/7011) - we genuinely appreciate the effort! 🚀 
  
Unfortunately, we'll need to close this PR. Some context on why:

The `/signup?token=` flow was fully removed in PR #10600 as part of deprecating the old invite API. The `/signup` page is now solely for first-time admin registration. The current invite system no longer routes users through `/signup` at all - invited users now receive a `/password-reset?token=` link instead. This means the `SignUp.tsx` page is intentionally unaware of any invite token, and issue [#7011](https://github.com/SigNoz/signoz/issues/7011) no longer reflects the actual user journey. 

Additionally we have addressed the issue ticket - [here](https://github.com/SigNoz/signoz/issues/7011#issuecomment-4603148051)

Thanks again for the contribution !!!

PR #11254, opened 2026-05-11
feat: new module for tags

namanv3:
### 📄 Summary

V2 Dashboards will have Tags as their own entity. Various dashboard operations will interact with this module.

This PR defines the module and the type, and all the module methods that will be required by the v2 dashboard APIs. 

#### Issues closed by this PR

Part of https://github.com/SigNoz/pulse-pod/issues/9

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Unit tests added for the new code

---

- Blast radius: Adds two new tables to the database, not a risk though
- Potential regressions: None
- Rollback plan: Delete the tables


pkg/types/tagtypes/store.go
srikanthccv, line 1: Is there any reason some of the reads and deletes methods are not org scoped?

pkg/modules/tag/impltag/module.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	return m.store.CreateRelations(ctx, tagtypes.NewTagRelations(orgID, entityType, entityID, tagIDs))
+}
+
+func (m *module) SyncLinksForEntity(ctx context.Context, orgID valuer.UUID, entityType tagtypes.EntityType, entityID valuer.UUID, tagIDs []valuer.UUID) error {
+	if err := m.store.CreateRelations(ctx, tagtypes.NewTagRelations(orgID, entityType, entityID, tagIDs)); err != nil {
+		return err
+	}
+	return m.store.DeleteRelationsExcept(ctx, entityType, entityID, tagIDs)
+}
```
srikanthccv: Missing txn here

pkg/types/tagtypes/tag_relation.go
```diff
@@ -0,0 +1,38 @@
+package tagtypes
+
+import (
+	"github.com/SigNoz/signoz/pkg/valuer"
+	"github.com/uptrace/bun"
+)
+
+type EntityType struct{ valuer.String }
```
srikanthccv: Should this be named `ResourceType` instead for the uniform nomenclature?

pkg/types/tagtypes/tag_relation.go
```diff
@@ -0,0 +1,38 @@
...
+
+func MustNewEntityType(name string) EntityType {
+	return EntityType{valuer.NewString(name)}
+}
+
+type TagRelation struct {
+	bun.BaseModel `bun:"table:tag_relations,alias:tag_relations"`
+
+	EntityType EntityType  `json:"entityType" required:"true" bun:"entity_type,type:text,notnull"`
+	EntityID   valuer.UUID `json:"entityId" required:"true" bun:"entity_id,pk,type:text,notnull"`
```
srikanthccv: `EntityID` is `valuer.UUID`. I imagine this is the dashboard ID if the entity/resource is "dashboard". How does this work when we move the integration dashboard into the same tables that power user tables. Today they have ID of the form `integration--{integrationId}--{dashboardId}`?

pkg/types/tagtypes/tag.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Resolve canonicalizes a batch of user-supplied (key, value) tag pairs against
+// the existing tags for an org. Lookup is case-insensitive on both key and
+// value (matching the storage uniqueness rule); when an existing row matches,
+// its display casing is reused. Inputs are deduped on (LOWER(key), LOWER(value));
+// the first input's casing wins on collisions. Returns:
+//   - toCreate: new Tag rows the caller should insert (with pre-generated IDs)
+//   - matched: existing rows the caller's input already pointed to. They
+//     already carry authoritative IDs from the store.
+func Resolve(ctx context.Context, store Store, orgID valuer.UUID, entityType EntityType, postable []PostableTag, createdBy string) ([]*Tag, []*Tag, error) {
```
srikanthccv: I suggest pass the list of existing tags as args, and remove the store dependency on `Resolve`. And should this `Resolve` be done in the module instead?

pkg/types/tagtypes/tag.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	existing, err := store.List(ctx, orgID, entityType)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	lowercaseTagsMap := make(map[string]*Tag, len(existing))
+	for _, t := range existing {
+		mapKey := strings.ToLower(t.Key) + "\x00" + strings.ToLower(t.Value)
```
srikanthccv: Is the assumption that tag key and value form the uniqueness always or should it include resource type and org id?

pkg/types/tagtypes/tag.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	if value == "" {
+		return "", "", errors.Newf(errors.TypeInvalidInput, ErrCodeTagInvalidName, "tag value cannot be empty")
+	}
+	if strings.ContainsRune(key, '/') {
+		return "", "", errors.Newf(errors.TypeInvalidInput, ErrCodeTagInvalidName, "tag key %q cannot contain '/'", key)
+	}
+	if strings.ContainsRune(value, '/') {
+		return "", "", errors.Newf(errors.TypeInvalidInput, ErrCodeTagInvalidName, "tag value %q cannot contain '/'", value)
+	}
```
srikanthccv: Is there a reason value can't have `/`?

pkg/types/tagtypes/tag_test.go
```diff
@@ -0,0 +1,166 @@
...
+				return
+			}
+			require.NoError(t, err)
+			assert.Equal(t, tc.wantKey, gotKey)
+			assert.Equal(t, tc.wantValue, gotValue)
+		})
+	}
+}
+
+var testEntityType = MustNewEntityType("dashboard")
```
srikanthccv: Should you define these as `EntityTypeDashboard` / `EntityTypeAlert` in `pkg/types/tagtypes/tag_relation.go` and use here?

pkg/sqlmigration/079_add_tags.go
```diff
@@ -0,0 +1,102 @@
...
+
+	tagRelationsTableSQLs := migration.sqlschema.Operator().CreateTable(&sqlschema.Table{
+		Name: "tag_relations",
+		Columns: []*sqlschema.Column{
+			{Name: "entity_type", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "entity_id", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "tag_id", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "org_id", DataType: sqlschema.DataTypeText, Nullable: false},
+		},
+		PrimaryKeyConstraint: &sqlschema.PrimaryKeyConstraint{ColumnNames: []sqlschema.ColumnName{"entity_type", "entity_id", "tag_id"}},
```
srikanthccv: The `entity_type` is pk here but in the bun model it's not?

pkg/sqlmigration/079_add_tags.go
```diff
@@ -0,0 +1,102 @@
...
+		Columns: []*sqlschema.Column{
+			{Name: "id", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "key", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "value", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "org_id", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "entity_type", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "created_at", DataType: sqlschema.DataTypeTimestamp, Nullable: false},
+			{Name: "created_by", DataType: sqlschema.DataTypeText, Nullable: true},
+			{Name: "updated_at", DataType: sqlschema.DataTypeTimestamp, Nullable: false},
+			{Name: "updated_by", DataType: sqlschema.DataTypeText, Nullable: true},
```
srikanthccv: Should this be nullable? because the new tag always expects the non null `created_by` and `updated_by`

pkg/types/tagtypes/tag.go
```diff
@@ -0,0 +1,149 @@
...
+// collide with the list-query DSL) are the caller's responsibility — perform
+// them before calling into the tag module.
+func validatePostableTag(p PostableTag) (string, string, error) {
+	key := strings.TrimSpace(p.Key)
+	value := strings.TrimSpace(p.Value)
+	if key == "" {
+		return "", "", errors.Newf(errors.TypeInvalidInput, ErrCodeTagInvalidName, "tag key cannot be empty")
+	}
+	if value == "" {
+		return "", "", errors.Newf(errors.TypeInvalidInput, ErrCodeTagInvalidName, "tag value cannot be empty")
```
srikanthccv: The error code used for the value is not ideal? and it should have it's own error code?

srikanthccv (commented):
Some comments

pkg/modules/tag/impltag/module.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	created, err := m.store.Create(ctx, toCreate)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(matched, created...), nil
```
srikanthccv, line 58: Should this preserve the order in which the tags were posted?

therealpandey:
Sorry for coming late to this PR, as I was hoping the schema was already verified.

Can we use the coretypes terminology?

pkg/modules/tag/impltag/module.go
namanv3: the idea was that the caller will put this in a transaction in the caller along with create/update dashboard

pkg/types/tagtypes/tag_relation.go
namanv3: Okay then actually even dashboards table's ID field will need a migration to plain string.

pkg/types/tagtypes/tag.go
namanv3: I did it this way so that the module does not have to know what data is exactly needed to resolve tags. If in the future a filtered list call is needed for example, then the module won't have to change, only the types package will have to.

pkg/types/tagtypes/tag.go
namanv3: this is just for the uniqueness on a particular entity so there the uniqueness has to be on key value pair

pkg/types/tagtypes/tag.go
namanv3: access control uses `/` so we're blocking it here

pkg/types/tagtypes/tag_test.go
namanv3: I kept these definitions in the respective modules actually, cuz the tags module shouldn't actually care about what all entity types exist. It just provides a way to save tags that's all.

pkg/types/tagtypes/tag.go
namanv3: oh yeah forgot to change this after the moving to key/value from simple names. Fixed

pkg/sqlmigration/079_add_tags.go
namanv3: Fixed, the entity type was added to the primary key after a review discussion with Vikrant, forgot to add it in the bun model

pkg/types/tagtypes/tag_test.go
namanv3: actually because access control has defined a Kind type, I'll just use that

pkg/sqlmigration/079_add_tags.go
```diff
@@ -0,0 +1,102 @@
...
+	sqls := [][]byte{}
+
+	tagTableSQLs := migration.sqlschema.Operator().CreateTable(&sqlschema.Table{
+		Name: "tag",
+		Columns: []*sqlschema.Column{
+			{Name: "id", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "key", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "value", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "org_id", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "entity_type", DataType: sqlschema.DataTypeText, Nullable: false},
```
vikrantgupta25:
```suggestion
			{Name: "kind", DataType: sqlschema.DataTypeText, Nullable: false},
```

pkg/sqlmigration/079_add_tags.go
```diff
@@ -0,0 +1,102 @@
...
+	// Postgres and SQLite (modernc 3.50.x) support expression indexes.
+	sqls = append(sqls, []byte(`CREATE UNIQUE INDEX IF NOT EXISTS uq_tag_org_entity_lower_key_lower_value ON tag (org_id, entity_type, LOWER(key), LOWER(value))`))
+
+	tagRelationsTableSQLs := migration.sqlschema.Operator().CreateTable(&sqlschema.Table{
+		Name: "tag_relations",
+		Columns: []*sqlschema.Column{
+			{Name: "entity_type", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "entity_id", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "tag_id", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "org_id", DataType: sqlschema.DataTypeText, Nullable: false},
```
vikrantgupta25: we don't need org_id here in the junction table ?

pkg/sqlmigration/079_add_tags.go
vikrantgupta25: do we need these user auditable fields if not a hard product requirement ?

vikrantgupta25 (changes requested):

pkg/sqlmigration/079_add_tags.go
```diff
@@ -0,0 +1,102 @@
...
+	})
+	sqls = append(sqls, tagTableSQLs...)
+
+	// Functional unique index: case-insensitive uniqueness on (org_id, entity_type, key, value).
+	// sqlschema.UniqueIndex doesn't support expressions, so emit raw SQL — both
+	// Postgres and SQLite (modernc 3.50.x) support expression indexes.
+	sqls = append(sqls, []byte(`CREATE UNIQUE INDEX IF NOT EXISTS uq_tag_org_entity_lower_key_lower_value ON tag (org_id, entity_type, LOWER(key), LOWER(value))`))
+
+	tagRelationsTableSQLs := migration.sqlschema.Operator().CreateTable(&sqlschema.Table{
+		Name: "tag_relations",
```
vikrantgupta25:
```suggestion
		Name: "tag_relation",
```

singular names please

pkg/types/tagtypes/tag_relation.go
```diff
@@ -0,0 +1,38 @@
...
+type EntityType struct{ valuer.String }
+
+func MustNewEntityType(name string) EntityType {
+	return EntityType{valuer.NewString(name)}
+}
+
+type TagRelation struct {
+	bun.BaseModel `bun:"table:tag_relations,alias:tag_relations"`
+
+	EntityType EntityType  `json:"entityType" required:"true" bun:"entity_type,pk,type:text,notnull"`
```
vikrantgupta25: we do not need bun annotations for pk / fk here.

vikrantgupta25:
```suggestion
	Kind Kind  `json:"kind" required:"true" bun:"type:text"`
```

vikrantgupta25 (changes requested):

pkg/types/tagtypes/tag_test.go
namanv3: Done

pkg/sqlmigration/079_add_tags.go
namanv3: Done

pkg/types/tagtypes/tag_relation.go
namanv3: Why not?

namanv3: Done

pkg/sqlmigration/079_add_tags.go
namanv3: Done

pkg/types/tagtypes/tag_relation.go
vikrantgupta25: Because we aren't creating any tables via bun anymore. Bun annotations for read are the only things required.

pkg/sqlmigration/080_add_tags.go
```diff
@@ -0,0 +1,102 @@
...
+	// Functional unique index: case-insensitive uniqueness on (org_id, kind, key, value).
+	// sqlschema.UniqueIndex doesn't support expressions, so emit raw SQL — both
+	// Postgres and SQLite (modernc 3.50.x) support expression indexes.
+	sqls = append(sqls, []byte(`CREATE UNIQUE INDEX IF NOT EXISTS uq_tag_org_kind_lower_key_lower_value ON tag (org_id, kind, LOWER(key), LOWER(value))`))
+
+	tagRelationsTableSQLs := migration.sqlschema.Operator().CreateTable(&sqlschema.Table{
+		Name: "tag_relation",
+		Columns: []*sqlschema.Column{
+			{Name: "kind", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "entity_id", DataType: sqlschema.DataTypeText, Nullable: false},
```
vikrantgupta25:
```suggestion
			{Name: "selector", DataType: sqlschema.DataTypeText, Nullable: false},
```
?

pkg/types/tagtypes/store.go
namanv3: because the entity ID and tag ID will be primary keys in their respective tables, I thought we don't really need org IDs here. Do you have any concern on this?

pkg/types/tagtypes/tag_relation.go
namanv3: This is removed now

pkg/sqlmigration/079_add_tags.go
namanv3: Removed

pkg/types/tagtypes/tag_relation.go
namanv3: Still even if we add it here it's not a concern right? Better to have as much info in the types as possible

pkg/sqlmigration/079_add_tags.go
namanv3: this can be info that we want to show once we have a page for tags. so even if not needed right now, we will probably need it soon enough

pkg/types/tagtypes/store.go
namanv3: @vikrantgupta25 even suggested that because entity ID and tag ID are primary keys in their tables so we shouldn't have org ID in the relations table anyway

pkg/sqlmigration/080_add_tags.go
namanv3: I see its used in core types but honestly the term selector does not make sense to me at all 😅 especially over here where entity_id easily makes more sense

vikrantgupta25: the idea behind this is to keep it in sync with what we convey to the users in audit / authz so that the language remains ubiquitous

pkg/types/tagtypes/store.go
srikanthccv: that is "safe" because entity ID/ tag ID is a UUID? but multi-tenancy correctness should not rely on UUID collision improbability IMO. I think regardless of the pk, the org-scoping should always be the lowest-layer invariant so the systemic conditions are enforced structurally and guaranteed.

pkg/types/tagtypes/tag.go
srikanthccv: Who will use this `Resolve` other than the module?

pkg/types/tagtypes/tag.go
srikanthccv: I can understand for keys but why does it error on value also?

pkg/sqlmigration/080_add_tags.go
srikanthccv: Even if we go by the language, the column isn't really the selector right? It's the ID value of the resource. It would be misleading to say that ID value is a selector when the selector is generally an expression.

pkg/modules/tag/tag.go
```diff
@@ -0,0 +1,29 @@
...
+	SyncTags(ctx context.Context, orgID valuer.UUID, kind coretypes.Kind, entityID valuer.UUID, postable []tagtypes.PostableTag, createdBy string) ([]*tagtypes.Tag, error)
+
+	// Does not link the resolved tags to any entity — call LinkToEntity for that.
+	CreateMany(ctx context.Context, orgID valuer.UUID, kind coretypes.Kind, postable []tagtypes.PostableTag, createdBy string) ([]*tagtypes.Tag, error)
+
+	// Existing rows are left untouched.
+	LinkToEntity(ctx context.Context, kind coretypes.Kind, entityID valuer.UUID, tagIDs []valuer.UUID) error
+
+	// missing links are inserted, obsolete ones removed.
+	SyncLinksForEntity(ctx context.Context, kind coretypes.Kind, entityID valuer.UUID, tagIDs []valuer.UUID) error
```
srikanthccv: Do we still need these? Can they be removed?

pkg/sqlmigration/082_add_tags.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// Functional unique index: case-insensitive uniqueness on (org_id, kind, key, value).
+	// sqlschema.UniqueIndex doesn't support expressions, so emit raw SQL — both
+	// Postgres and SQLite (modernc 3.50.x) support expression indexes.
+	sqls = append(sqls, []byte(`CREATE UNIQUE INDEX IF NOT EXISTS uq_tag_org_kind_lower_key_lower_value ON tag (org_id, kind, LOWER(key), LOWER(value))`))
+
+	tagRelationsTableSQLs := migration.sqlschema.Operator().CreateTable(&sqlschema.Table{
+		Name: "tag_relation",
+		Columns: []*sqlschema.Column{
+			{Name: "kind", DataType: sqlschema.DataTypeText, Nullable: false},
```
vikrantgupta25, line 71: we should have the `id` primary key and the `kind` , `entity_id` , `tag_id` in the unique index ? 

Any reason for keeping this as the primary key ? A composite key of this format would affect the writes

pkg/types/tagtypes/store.go
namanv3: not because its a UUID but because the ID will be a primary key in their table, so its the db's primary key constraint that makes ID unique across orgs, not the fact that it is a UUID

pkg/types/tagtypes/tag.go
namanv3: As of now no one, but that shouldn't be a reason to not ensure that business logic lives inside the types package right?

pkg/types/tagtypes/tag.go
namanv3: I didn't want access control code to keep in mind that value can have `/`, might just lead to some random bugs.

pkg/modules/tag/impltag/module.go
namanv3: I don't think so. We restrict to max 5 anyway so its not really required I think to sort in any way

pkg/modules/tag/tag.go
namanv3: create dashboard will still need it cuz it won't call the sync tags method.
The order in create API should be tags -> dashboard -> tagRelations. So this method can't be called. Update API can use it however.

Even though right now tag relations can be created without the entity ID being actually present in its own db, we shouldn't assume that in the dashboards package I think.

pkg/sqlmigration/082_add_tags.go
namanv3: I did some perf testing in both the models and the gap is around 1-2% for both reads and writes most of the time, and the winner isn't really consistent. I think performance isn't really a decision maker here. 

But let me test with multiple shards as well once just to be sure.

pkg/types/tagtypes/store.go
namanv3: added org id check in the queries

pkg/types/tagtypes/tag_relation.go
namanv3: have commented on the RFC shared by Swapnil to have UUID for integration dashboards as well. Let's see how that goes

pkg/sqlmigration/079_add_tags.go
namanv3: removed UserAuditable from tags

pkg/sqlmigration/080_add_tags.go
namanv3: changed it to resource ID

pkg/sqlmigration/082_add_tags.go
namanv3: shifted to the model you suggested

pkg/types/tagtypes/tag.go
namanv3: @vikrantgupta25 @therealpandey I have removed the check for `/` altogether and haven't added any encoding in the database for now. That's because access control would anyway interact with the tags module and not the database directly to handle tags, so encoding and storing didn't make sense. Once access control starts using tags, we can expose some methods on tags to get the encoded values

pkg/sqlschema/index.go
```diff
@@ -232,3 +235,111 @@ func (index *PartialUniqueIndex) ToDropSQL(fmter SQLFormatter) []byte {
...
+// expressions (e.g. LOWER(key)) rather than plain column names. The strings in
+// Expressions are emitted into the generated SQL exactly as supplied — they are
+// not identifier-quoted by the formatter — so the caller is responsible for
+// well-formed expressions. Auto-generated names are opaque (`fnq_<table>_<hash>`);
+// prefer calling Named(...) to give the index a readable name.
+//
+// ReferencedColumns is informational: it lets consumers (drift detection,
+// "which columns are indexed" lookups) answer that question without having to
+// parse Expressions. It does not affect the emitted SQL or Equals.
+type FunctionalUniqueIndex struct {
```
therealpandey: Please club this with Unique Index

pkg/types/tagtypes/store.go
```diff
@@ -0,0 +1,30 @@
...
+	ListByResource(ctx context.Context, orgID valuer.UUID, kind coretypes.Kind, resourceID valuer.UUID) ([]*Tag, error)
+
+	ListByResources(ctx context.Context, orgID valuer.UUID, kind coretypes.Kind, resourceIDs []valuer.UUID) (map[valuer.UUID][]*Tag, error)
+
+	// Create upserts the given tags and returns them with authoritative IDs.
+	// On conflict on (org_id, kind, LOWER(key), LOWER(value)) — which
+	// happens only when a concurrent insert raced ours, including casing-only
+	// collisions — the returned entry carries the existing row's ID rather
+	// than the pre-generated one in the input.
+	Create(ctx context.Context, tags []*Tag) ([]*Tag, error)
```
therealpandey: CreateOrGet

pkg/types/tagtypes/tag.go
```diff
@@ -0,0 +1,162 @@
...
+
+// Resolve canonicalizes a batch of user-supplied (key, value) tag pairs against
+// the existing tags for an org. Lookup is case-insensitive on both key and
+// value (matching the storage uniqueness rule); when an existing row matches,
+// its display casing is reused. Inputs are deduped on (LOWER(key), LOWER(value));
+// the first input's casing wins on collisions. Returns:
+//   - toCreate: new Tag rows the caller should insert (with pre-generated IDs)
+//   - matched: existing rows the caller's input already pointed to. They
+//     already carry authoritative IDs from the store.
+func Resolve(ctx context.Context, store Store, orgID valuer.UUID, kind coretypes.Kind, postable []PostableTag) ([]*Tag, []*Tag, error) {
```
therealpandey: Move to module

therealpandey (changes requested):

pkg/types/tagtypes/tag_test.go
```diff
@@ -0,0 +1,196 @@
...
+			require.NoError(t, err)
+			assert.Equal(t, tc.wantKey, gotKey)
+			assert.Equal(t, tc.wantValue, gotValue)
+		})
+	}
+}
+
+var testKind = coretypes.KindDashboard
+
+type fakeStore struct {
```
therealpandey: Move this to tagtypestest like other packages in types.

therealpandey (changes requested):

pkg/types/tagtypes/tag.go
namanv3: I did it this way so that the module does not have to know what data is exactly needed to resolve tags. If in the future a filtered list call is needed for example, then the module won't have to change, only the types package will have to.

Let me know if there's anything I'm missing

pkg/types/tagtypes/tag_test.go
namanv3: done

pkg/types/tagtypes/store.go
namanv3: done

pkg/types/tagtypes/tag.go
therealpandey: You cant use store interfaces as arguments in types. Breaks the contract of types and the way store has been modelled. Please see patterns in rest of the types layer for further clarification.

pkg/sqlschema/index.go
namanv3: done

pkg/types/tagtypes/tag.go
namanv3: done

pkg/sqlschema/index.go
namanv3: added unit tests as well

srikanthccv (dismissed):

therealpandey:
@namanv3 is postgres and sqlite able to read such indexes back from their ddl schemas. Please test. Try calling GetIndexes functions of both engines and see if you are able to recreate the index exactly with expressions.

therealpandey (approved):

srikanthccv (dismissed):

vikrantgupta25 (dismissed):

pkg/sqlmigration/082_add_tags.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// TODO (@namanverma): add a unique index for tags: (org_id, kind, (LOWER(key)), (LOWER(value)))
+
+	tagRelationsTableSQLs := migration.sqlschema.Operator().CreateTable(&sqlschema.Table{
+		Name: "tag_relation",
+		Columns: []*sqlschema.Column{
+			{Name: "id", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "kind", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "resource_id", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "tag_id", DataType: sqlschema.DataTypeText, Nullable: false},
```
therealpandey, line 73: @namanv3 let's add a created_at, updated_at here also. If a relation cannot be updated, only deleted and recreated, let's only add created_at.

therealpandey (changes requested):

pkg/sqlmigration/082_add_tags.go
namanv3: done

srikanthccv (approved):

therealpandey (approved):

vikrantgupta25 (approved):

PR #11264, opened 2026-05-12
feat(infra-monitoring): allow order by primary name column in v2 apis

nikhilmantri0902:
### 📄 Summary

  Add ordering by primary name attribute across all v2 infra-monitoring list APIs (10 entities) so callers can browse alphabetically instead of being forced to rank by a numeric metric. Opt-in — existing defaults
  unchanged.

  When `orderBy.key.name` matches the entity's primary name attr, `getTopXGroups` skips the metric ranking query and sorts the already-fetched `metadataMap` by that attr. Validation rejects this key when `groupBy` is
  non-empty (only valid for `ResponseTypeList`).

  **Per-entity sortable name attr keys:**

  | Entity | `orderBy.key.name` |
  |---|---|
  | clusters | `k8s.cluster.name` |
  | daemonsets | `k8s.daemonset.name` |
  | deployments | `k8s.deployment.name` |
  | hosts | `host.name` |
  | jobs | `k8s.job.name` |
  | namespaces | `k8s.namespace.name` |
  | nodes | `k8s.node.name` |
  | pods | `k8s.pod.name` |
  | statefulsets | `k8s.statefulset.name` |
  | volumes | `k8s.persistentvolumeclaim.name` |

  Wire-value uses the actual attr key (not a magic `"name"` token) — self-describing + forward-compat w/ future "order by any groupBy key".

  N/A

  #### Issues closed by this PR
  Closes SigNoz/engineering-pod#4847

  ---

  ### ✅ Change Type
  - [x] ✨ Feature
  - [ ] 🐛 Bug fix
  - [ ] ♻️ Refactor
  - [ ] 🛠️ Infra / Tooling
  - [ ] 🧪 Test-only

  ---

  ### 🐛 Bug Context
  N/A

  ---

  ### 🧪 Testing Strategy

  - **Tests added/updated**: 3 `Validate()` cases per entity × 10 entities (name-asc/empty-groupBy → ok, name-desc/empty-groupBy → ok, name/non-empty-groupBy → rejected).
  - **Manual verification**: `go build ./...` clean, `go test ./pkg/types/inframonitoringtypes/... ./pkg/modules/inframonitoring/...` green.
  - **Edge cases covered**: empty `metadataMap` → empty page; `offset >= total` → nil; composite-key tie-break; pods (UID-as-primary groupBy) sorts by `k8s.pod.name` from enrich attrs, not opaque UID.

  ---

  - **Blast radius**: low — single short-circuit per entity at top of `getTopXGroups`; metric path untouched.
  - **Potential regressions**: none expected — opt-in.
  - **Rollback plan**: revert PR; no data migration / schema / frontend type change.

  ---

  ### 📝 Changelog

  | Field | Value |
  |------|-------|
  | Change Type | Feature |
  | Description | v2 infra-monitoring list APIs now accept the entity's primary name attr (e.g. `k8s.cluster.name`, `host.name`) as `orderBy.key.name` for alphabetical sort. Only valid when `groupBy` is empty
  (ResponseTypeList). |

  ---

  ### 📋 Checklist
  - [x] Manually tested
  - [x] Backward compatibility considered (additive, opt-in)

  ---

  ## 👀 Notes for Reviewers

  - 10 new exported consts `inframonitoringtypes.<Entity>NameAttrKey` (per entity `_constants.go`), appended to each `<Entity>ValidOrderByKeys`. Private dupes in `implinframonitoring/<entity>_constants.go` removed; call
   sites point at exported const.
  - New helper `paginateMetadataByName` in `implinframonitoring/helpers.go` — sorts `metadataMap[k][sortByMetaKey]` w/ composite-key tie-break.
  - Pods orderBy key is `k8s.pod.name` even though primary groupBy is `k8s.pod.uid` — intentional UX.
  - OpenAPI schema unchanged (orderBy free-string); no `pnpm generate:api` needed.
  - Validation rule inlined in all 10 `Validate()` methods — no shared helper.


pkg/types/inframonitoringtypes/clusters_constants.go
```diff
@@ -12,4 +12,5 @@ var ClustersValidOrderByKeys = []string{
 	ClustersOrderByCPUAllocatable,
 	ClustersOrderByMemory,
 	ClustersOrderByMemoryAllocatable,
+	OrderByName,
```
namanv3: Let’s not use “name” everywhere and use the proper key so that it’s clear to the client that it can only be used when there is no group by. Also it will flow in well when we add ordering by any group by key in the future

pkg/types/inframonitoringtypes/hosts_constants.go
```diff
@@ -34,4 +34,5 @@ var HostsValidOrderByKeys = []string{
 	HostsOrderByWait,
 	HostsOrderByDiskUsage,
 	HostsOrderByLoad15,
+	OrderByName,
```
AshwinBhatkal: question: can these constants be reflected in the open api yaml as well?

AshwinBhatkal: we'll need to verify this once during integration

pkg/types/inframonitoringtypes/clusters_constants.go
nikhilmantri0902: On it.

nikhilmantri0902: this was done

nikhilmantri0902: updated PR description as well. Do check, thanks.

pkg/types/inframonitoringtypes/hosts_constants.go
nikhilmantri0902: Hi Ashwin, No these constants are not available in open api yaml because we are using a struct predefined in qbtypes.go . I will do this seperately. Out of scope of this PR.

pkg/modules/inframonitoring/implinframonitoring/helpers.go
```diff
@@ -190,6 +190,53 @@ func paginateWithBackfill(
 	return pageGroups
 }
 
+// paginateMetadataByName returns metadataMap groups sorted by name
+// (lexicographic, asc or desc), paginated by offset/limit, and rebuilt into
+// label maps using groupBy. When sortByMetaKey is non-empty, groups are sorted
+// by metadataMap[k][sortByMetaKey]; otherwise sorted by composite key directly.
+func paginateMetadataByName(
+	metadataMap map[string]map[string]string,
+	groupBy []qbtypes.GroupByKey,
```
namanv3: don't need group by really right? It'll be just the name attribute anyway

nikhilmantri0902: no, group by will be needed in order to extract group by attributes from metadataMap of compositeKey -> attr_name -> attr_key mapping after sorting. Check function once.

nikhilmantri0902: So groupBy IS needed in the function — the returned page label maps must carry the groupBy attr (e.g. `k8s.pod.uid`) to match metadataMap composite keys and downstream IN-filters, which is not always the same as the sort attr.

namanv3 (dismissed):

nikhilmantri0902:
@therealpandey can you review and approve?

pkg/modules/inframonitoring/implinframonitoring/helpers.go
```diff
@@ -190,6 +190,53 @@ func paginateWithBackfill(
 	return pageGroups
 }
 
+// paginateMetadataByName returns metadataMap groups sorted by name
+// (lexicographic, asc or desc), paginated by offset/limit, and rebuilt into
+// label maps using groupBy. When sortByMetaKey is non-empty, groups are sorted
+// by metadataMap[k][sortByMetaKey]; otherwise sorted by composite key directly.
+func paginateMetadataByName(
```
therealpandey: Please move to types package and add detailed unit tests covering array out of bounds and nil pointers.

therealpandey (changes requested):

pkg/modules/inframonitoring/implinframonitoring/helpers.go
nikhilmantri0902: on it.

nikhilmantri0902: This was done.

pkg/types/inframonitoringtypes/pagination.go
```diff
@@ -0,0 +1,54 @@
...
+// by metadataMap[k][sortByMetaKey]; otherwise sorted by composite key directly.
+func PaginateMetadataByName(
+	metadataMap map[string]map[string]string,
+	groupBy []qbtypes.GroupByKey,
+	direction qbtypes.OrderDirection,
+	offset, limit int,
+	sortByMetaKey string,
+) []map[string]string {
+	if offset >= len(metadataMap) {
+		return nil
```
therealpandey: There is a difference between `return nil` and `return []map[string]string`. What do we want here?

Also side note: Please ask claude to leave newlines between logical code blocks.

therealpandey: The pagegroups will always return a `[]map[string]string`

therealpandey (changes requested):

pkg/types/inframonitoringtypes/pagination.go
nikhilmantri0902: okay so to keep it consistent, I ll return empty slice of the map, is that fine?

nikhilmantri0902: We should return nil in case of error. so not applicable for this function.

pkg/types/inframonitoringtypes/pagination_test.go
```diff
@@ -0,0 +1,371 @@
...
+				tt.offset,
+				tt.limit,
+				tt.sortByMetaKey,
+			)
+			if tt.wantNotNil {
+				require.NotNil(t, got)
+				require.Len(t, got, 0)
+				return
+			}
+			require.Equal(t, tt.want, got)
```
therealpandey:
```suggestion
			assert.Equal(t, tt.want, got)
```

pkg/types/inframonitoringtypes/pagination_test.go
```diff
@@ -0,0 +1,371 @@
...
+				tt.metadataMap,
+				tt.groupBy,
+				tt.direction,
+				tt.offset,
+				tt.limit,
+				tt.sortByMetaKey,
+			)
+			if tt.wantNotNil {
+				require.NotNil(t, got)
+				require.Len(t, got, 0)
```
therealpandey:
```suggestion
				assert.NotNil(t, got)
				assert.Len(t, got, 0)
```

pkg/types/inframonitoringtypes/pagination.go
```diff
@@ -0,0 +1,55 @@
...
+				return entries[i].sortVal < entries[j].sortVal
+			}
+			return entries[i].sortVal > entries[j].sortVal
+		}
+		return entries[i].compositeKey < entries[j].compositeKey
+	})
+
+	end := min(offset+limit, len(entries))
+
+	for _, e := range entries[offset:end] {
```
therealpandey, line 46: How are we sure offset <= end here? There is no guard/check in the function which prevents this from happening.

One unit test and fix if required here.

therealpandey (changes requested):

pkg/types/inframonitoringtypes/pagination.go
nikhilmantri0902: if offset >= len(metadataMap) {
	return pageGroups
}
	
check this uptop the function. this guarantees.

nikhilmantri0902: unit test already added:

{
			name:          "offset_way_past_len",
			metadataMap:   fiveHostMap(),
			groupBy:       hostGB,
			direction:     qbtypes.OrderDirectionAsc,
			offset:        100,
			limit:         10,
			sortByMetaKey: "host.name",
			wantNotNil:    true,
		},

therealpandey (approved):

nikhilmantri0902:
@therealpandey the failing CI integration tests are not due to this change.

therealpandey:
@nikhilmantri0902 the alert integration tests are flaky, please re-run them if you encounter a failure. They should pass.

Flag if it is persistent.

PR #11265, opened 2026-05-12
feat: add query length validation

tushar-signoz:
## Pull Request

---

### 📄 Summary

Adds size guards to v5 query validation so requests with pathologically large
query strings are rejected up front instead of being forwarded to ClickHouse /
the PromQL engine. Three caps are added:

- **`Filter.Expression`** on builder queries → 300,000 characters
- **`ClickHouseQuery.Query`** (raw SQL spec) → 300,000 characters
- **`PromQuery.Query`** (PromQL spec) → 300,000 characters

Validation lives on the spec types (`ClickHouseQuery.Validate()`,
`PromQuery.Validate()`) and on `QueryBuilderQuery.validateFilterExpression()`,
keeping length rules colocated with the type they guard.

### NOTE:
This PR is currently a NO-OP
Why ? Current clickhouse default is 261k something, so this guard of 300_000 is more conservative. 
Next step will be to bump the clickhouse default to 350_000. 
Why take this route ?
This ensures none of the currently working bad clickhouse queries break. 
Why 300_000 ? No reason, it the next big round number.

#### Issues closed by this PR
N/A

---

### ✅ Change Type

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context

N/A — not a bug fix.

---

### 🧪 Testing Strategy

- Tests added/updated: None — guard is a straightforward length check on a
  primitive field; behavior is exercised by existing validation paths.
- Manual verification: `go build ./pkg/types/querybuildertypes/...` and
  running the existing v5 unit tests locally.
- Edge cases covered: empty string (already-existing required-field error
  path is preserved), exactly-at-limit (allowed), one over limit (rejected).

---

- **Blast radius:** No Impact
---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | v5 query API now rejects builder filter expressions over 250,000 characters and raw ClickHouse / PromQL queries over 260,000 characters with a 400 `invalid_input` error. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

### DIFF
--- a/conf/example.yaml
+++ b/conf/example.yaml
@@ -152,6 +152,7 @@ telemetrystore:
       max_result_rows: 0
       ignore_data_skipping_indices: ""
       secondary_indices_enable_bulk_filtering: false
+      max_query_size: 350000
 
 ##################### Prometheus #####################
 prometheus:
--- a/tests/fixtures/signoz.py
+++ b/tests/fixtures/signoz.py
@@ -78,6 +78,7 @@ def create_signoz(
                 "SIGNOZ_ALERTMANAGER_SIGNOZ_ROUTE_GROUP__WAIT": "1s",
                 "SIGNOZ_ALERTMANAGER_SIGNOZ_ROUTE_GROUP__INTERVAL": "5s",
                 "SIGNOZ_CLOUDINTEGRATION_AGENT_VERSION": "v0.0.8",
+                "SIGNOZ_TELEMETRYSTORE_CLICKHOUSE_SETTINGS_MAX__QUERY__SIZE": "350000",
             }
             | sqlstore.env
             | clickhouse.env
--- a/pkg/telemetrystore/config.go
+++ b/pkg/telemetrystore/config.go
@@ -47,6 +47,7 @@ type QuerySettings struct {
 	MaxResultRows                       int    `mapstructure:"max_result_rows"`
 	IgnoreDataSkippingIndices           string `mapstructure:"ignore_data_skipping_indices"`
 	SecondaryIndicesEnableBulkFiltering bool   `mapstructure:"secondary_indices_enable_bulk_filtering"`
+	MaxClickHouseQuerySize              int    `mapstructure:"max_query_size"`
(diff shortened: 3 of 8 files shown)


conf/example.yaml
```diff
@@ -152,6 +152,7 @@ telemetrystore:
       max_result_rows: 0
       ignore_data_skipping_indices: ""
       secondary_indices_enable_bulk_filtering: false
+      max_query_size: 350000
```
nityanandagohain, line 155:
```suggestion
      max_query_size: 300000
```

tests/fixtures/signoz.py
```diff
@@ -78,6 +78,7 @@ def create() -> types.SigNoz:
                 "SIGNOZ_ALERTMANAGER_SIGNOZ_ROUTE_GROUP__WAIT": "1s",
                 "SIGNOZ_ALERTMANAGER_SIGNOZ_ROUTE_GROUP__INTERVAL": "5s",
                 "SIGNOZ_CLOUDINTEGRATION_AGENT_VERSION": "v0.0.8",
+                "SIGNOZ_TELEMETRYSTORE_CLICKHOUSE_SETTINGS_MAX__QUERY__SIZE": "350000",
```
nityanandagohain, line 81:
```suggestion
                "SIGNOZ_TELEMETRYSTORE_CLICKHOUSE_SETTINGS_MAX__QUERY__SIZE": "300000",
```

nityanandagohain (commented):
350k should be 300k ?

conf/example.yaml
tushar-signoz: No no, clickhouse threshold has to be greater than what we have in signoz backend

nityanandagohain (commented):
lgtm

PR #11268, opened 2026-05-12
fix: ClickHouse 25.12.5 Trace Operator query analyzer fail due to dangling CTE

piyushsingariya:
## Pull Request

---

### 📄 Summary

Fixes issue with ClickHouse error in TraceOperator

#### Issues closed by this PR

Closes signoz/engineering-pod#4754
Closes signoz/engineering-pod#4867

---

### ✅ Change Type
_Select all that apply_

- [x] 🐛 Bug fix

---

### 🐛 Bug Context
```
{"status":"error","error":{"code":"unknown","message":"code: 10, message: Not found column __table1.timestamp in block. There are only columns: __table1.timestamp_0, __global_row_index, ts_bucket_start, attributes_string"}}
```

#### Root Cause
The failure is a ClickHouse 25.12.5 server-side regression: when signoz_index_v3 contains any data, CH's new distributed query analyzer renames the timestamp column to timestamp_0 while constructing the execution plan for this specific CTE shape, making it invisible to the outer SELECT.

The issue stems from a dangling CTE present in Query when `returnSpansFrom` is populated.

#### Fix Strategy

Acknowledge the dangling CTE, select correct spans from TraceMatching, acknowledge `returnSpansFrom`

---

- Tests added/updated: ✅ 
- Manual verification: 
- Edge cases covered:

---

- Blast radius: Alerts, Traces Explorer
- Potential regressions:
- Rollback plan: Revert pull request

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---


pkg/telemetrytraces/trace_operator_cte_builder.go
```diff
@@ -70,12 +70,39 @@ func (b *traceOperatorCTEBuilder) build(ctx context.Context, requestType qbtypes
...
+		matchingTracedSB := sqlbuilder.NewSelectBuilder()
+		matchingTracedSB.Select("DISTINCT trace_id")
+		matchingTracedSB.From(rootCTEName)
+		matchedTracesSQL, matchedTracesArgs := matchingTracedSB.BuildWithFlavor(sqlbuilder.ClickHouse)
+
+		filteredSB := sqlbuilder.NewSelectBuilder()
+		filteredSB.Select("src.*")
+		filteredSB.From(fmt.Sprintf("%s AS src", sourceQueryCTE))
+		filteredSB.JoinWithOption(
+			sqlbuilder.InnerJoin,
```
krsoninikhil: Will this be a global join or should be use `GLOBAL IN`?

piyushsingariya: We don't need Global Join, since the table is joining with itself. and same traceid data remain in same shard.

krsoninikhil (dismissed):

tests/integration/tests/querier/15_trace_operator.py
```diff
@@ -0,0 +1,537 @@
...
+                attributes={"operation.type": "child"},
+            ),
+        ]
+    )
+
+    token = get_token(USER_ADMIN_EMAIL, USER_ADMIN_PASSWORD)
+    start_ms = int((now - timedelta(minutes=5)).timestamp() * 1000)
+    end_ms = int(now.timestamp() * 1000)
+
+    response = make_query_request(
```
nityanandagohain: Add the following test cases here to making it complete you can use `pytest.param` to build it, check other files.

With and without return from spans for each of
    * `=> , -> , || , &&, NOT`

tests/integration/tests/querier/15_trace_operator.py
```diff
@@ -0,0 +1,537 @@
...
+
+    # Both grandparent spans must be returned
+    assert len(rows) == 2
+
+    # Ordering: POST > GET in DESC — svc-a (POST) must come before svc-b (GET)
+    assert rows[0]["data"]["service.name"] == "svc-a", f"Expected svc-a (POST) first in http.method DESC order, got {rows[0]['data']['service.name']}"
+    assert rows[1]["data"]["service.name"] == "svc-b", f"Expected svc-b (GET) second in http.method DESC order, got {rows[1]['data']['service.name']}"
+
+
+def test_trace_operator_query_order_by_select_field(
```
nityanandagohain: can be part of th previous test itself ?

nityanandagohain (commented):
* Order by tests can be combined to a signle test with multiple cases.
---
I will add the group by and aggreagtion tests in the future pr's, as it is currently wrong right now.

tests/integration/tests/querier/15_trace_operator.py
piyushsingariya: Added

tests/integration/tests/querier/15_trace_operator.py
piyushsingariya: Addressed

pkg/telemetrytraces/trace_operator_cte_builder.go
```diff
@@ -70,12 +70,39 @@ func (b *traceOperatorCTEBuilder) build(ctx context.Context, requestType qbtypes
...
+		// making the join a clean membership test with no fan-out.
+		matchingTracedSB := sqlbuilder.NewSelectBuilder()
+		matchingTracedSB.Select("DISTINCT trace_id")
+		matchingTracedSB.From(rootCTEName)
+		matchedTracesSQL, matchedTracesArgs := matchingTracedSB.BuildWithFlavor(sqlbuilder.ClickHouse)
+
+		filteredSB := sqlbuilder.NewSelectBuilder()
+		filteredSB.Select("src.*")
+		filteredSB.From(fmt.Sprintf("%s AS src", sourceQueryCTE))
+		filteredSB.JoinWithOption(
```
srikanthccv: Why is this a join? This could be simply `IN`?

https://github.com/ClickHouse/ClickHouse/issues/38140#issuecomment-2466457226

tests/integration/tests/querier/15_trace_operator.py
```diff
@@ -0,0 +1,286 @@
...
+            # noise — must not surface in any query below
+            *_chain_trace(now, ("noise-span", "svc-noise", "noise-op", 1)),
+        ]
+    )
+
+    token = get_token(USER_ADMIN_EMAIL, USER_ADMIN_PASSWORD)
+    start_ms = int((now - timedelta(minutes=5)).timestamp() * 1000)
+    end_ms = int(now.timestamp() * 1000)
+
+    def check(case_id, filter_a, filter_b, expression, return_spans_from, expected_names):
```
srikanthccv: with each integration test we are introducing a new way to write a test case and assert it, I am assuming this is assisted by AI which produces some plausible good looking new test case pattern each time. let me go through the other tests ...

piyushsingariya: Yes, testing each signal is different, and each time AI comes up with new pattern, I like how our querier unit tests are written, I tried doing the same here, but that was bloating up a lot.

pkg/telemetrytraces/trace_operator_cte_builder.go
piyushsingariya: Looked into it, replacing it. Thanks

tests/integration/tests/querier/15_trace_operator.py
```diff
@@ -0,0 +1,429 @@
...
+                status_message="",
+                resources={"service.name": "checkout-svc"},
+                attributes={"operation.type": "root"},
+            ),
+            Traces(
+                timestamp=now - timedelta(seconds=10),
+                duration=timedelta(seconds=5),
+                trace_id=t1_trace_id,
+                span_id=t1_checkout_span_id,
+                parent_span_id="",
```
nityanandagohain: this is incorrect example data, where two of them have same trace_id but no parent_span_id. No parent_span_id means that it is a root, but there cannot be two roots.

tests/integration/tests/querier/15_trace_operator.py
```diff
@@ -0,0 +1,429 @@
...
+                span_id=TraceIdGenerator.span_id(),
+                parent_span_id="",
+                name="standalone-server",
+                kind=TracesKind.SPAN_KIND_SERVER,
+                status_code=TracesStatusCode.STATUS_CODE_OK,
+                status_message="",
+                resources={"service.name": "standalone-svc"},
+                attributes={"operation.type": "root", "http.method": "POST"},
+            ),
+            # T4 — isolated-svc: single orphan span with no "root" ancestor in its trace.
```
nityanandagohain: This itself is a correct root span for a difference trace, so the comment is not entirely correct.

tests/integration/tests/querier/15_trace_operator.py
```diff
@@ -0,0 +1,429 @@
...
+            },
+            "formatOptions": {"formatTableResultForUI": False, "fillGaps": False},
+        },
+    )
+    assert response.status_code == HTTPStatus.OK, f"[{case['name']}] HTTP {response.status_code}: {response.text}"
+    assert case["validate"](response), f"[{case['name']}] validation failed: {response.json()}"
+
+
+# ============================================================================
+# Dataset — 4 traces shared across all order-by and expression cases
```
nityanandagohain: this comment can be simpler as it takes a lot of bandwith to go through it, a simpl diragram using ai if it helps. something like this https://github.com/SigNoz/signoz/blob/9ba57d323dad0ca98b8a50aab8141a9c4591862b/pkg/modules/tracedetail/impltracedetail/waterfall_test.go#L150

tests/integration/tests/querier/15_trace_operator.py
```diff
@@ -0,0 +1,429 @@
...
+                attributes={"operation.type": "orphan"},
+            ),
+        ]
+    )
+
+    token = get_token(USER_ADMIN_EMAIL, USER_ADMIN_PASSWORD)
+    start_ms = int((now - timedelta(minutes=5)).timestamp() * 1000)
+    end_ms = int(now.timestamp() * 1000)
+
+    cases = [
```
nityanandagohain: nit: @pytest.mark.parametrize can be used now I guess.

nityanandagohain (commented):
some minor comments regarding the sample spans.

Thanks for cleaning up the test file

tests/integration/tests/querier/15_trace_operator.py
piyushsingariya: I'm still working on it. Will ping over slack once I am done with these

nityanandagohain (dismissed):

tests/integration/tests/querier/15_trace_operator.py
```diff
@@ -0,0 +1,443 @@
...
+from fixtures.auth import USER_ADMIN_EMAIL, USER_ADMIN_PASSWORD
+from fixtures.traces import TraceIdGenerator, Traces, TracesKind, TracesStatusCode
+
+
+def _rows(response: requests.Response) -> list:
+    return response.json()["data"]["data"]["results"][0].get("rows") or []
+
+
+def _names(response: requests.Response) -> set:
+    return {r["data"]["name"] for r in _rows(response)}
```
therealpandey: Reuse utilities from fixtures/querier.

tests/integration/tests/querier/15_trace_operator.py
```diff
@@ -0,0 +1,443 @@
...
+
+
+def _run_case(signoz: types.SigNoz, token: str, start_ms: int, end_ms: int, case: dict) -> None:
```
therealpandey: Redundant. can put it directly in the test below.

therealpandey (changes requested):

srikanthccv (approved):
haven't reviewed the tests, only the query part.

therealpandey (approved):

PR #11273, opened 2026-05-12
feat(planned-downtime): mute alerts from their rules page

### LINKED ISSUE #389: Enable muting of alerts
Currently, SigNoz doesn't support muting of alerts.

This may be useful for teams to manage alert load when there are known issues which are causing alerts and team is already working to fix it

jatinderjit:
### 📄 Summary

-  Adds a self-service mute action directly on the alert detail page and alert list, so users can suppress notifications for a rule without navigating away to Planned Downtimes

#### Mute alerts

- **AlertStateSegmented** replaces the Active/Disabled toggle with a three-state pill (Active / Mute / Disable). Active is non-clickable while muted; unmuting happens via the Planned Downtimes page.
- **MutePopover** offers quick durations (15 m, 1 h, 4 h, 1 d, 1 w, Forever) plus an optional Name field.
- **MuteSchedulerDrawer** exposes the full Planned Downtime form for custom windows and recurrence.
- **MutedBanner / DisabledBanner** render an informative banner under the alert header showing mute end time, countdown pill, and reason.
- **MutedBadge** appears next to the rule name in the alert list with a live countdown.
- **Backend**: `activeMute` / `mutes` field added to `ListRules` and `GetRuleByID` responses so the frontend reads mute state from the rule API directly instead of a separate downtime-schedules fetch.

##### Mute alerts

- List alerts
  image

- Alert page - Active
  image

- Alert page - Muted
  image

- Alert page - Disabled
  image

- Create Mute - quick
  image

- Create Mute - "Custom window"
  image

##### Promote sub-tabs

Before

image

After

image

#### Issues closed by this PR

- Closes: #389

---

### ✅ Change Type

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🧪 Testing Strategy

- Tests added/updated: None — UI-only surface; automated tests not added.
- Manual verification: Tested quick-duration mute, custom window mute, forever mute (null endTime), muted badge countdown in list, banner expiry on the detail page, Escape / outside-click dismissal of the popover, mute pill disabled state when rule is disabled.
- Edge cases covered: Forever mute (endTime=null) shows no countdown and never auto-expires client-side; mute on a disabled rule is blocked at the UI level; banner clears within 30 s of expiry via polling refetch + client-side `onExpire` callback.

---

- Blast radius: List alerts and alert details APIs (new `mutes` / `activeMute` fields on rule DTOs); alert header rendering.
- Potential regressions: Rule list page (new MutedBadge column in Alert Name cell); alert header state logic (three-state instead of two-state).
- Rollback plan: Revert PR

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Alert rules can now be muted directly from the alert detail page or list view. Muting creates a scoped Planned Downtime entry; the rule keeps evaluating but notifications are suppressed for the chosen duration. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- The popover uses `trigger={[]}` (fully controlled) because antd's default `"click"` trigger fired on any click within the anchor element, including the Active and Disable pills.
- `MutedBanner` accepts an `onExpire` callback that fires when the countdown interval detects the mute has ended, triggering a refetch so the banner clears client-side without waiting for the next poll cycle.
- The backend join (`findActiveMuteForRule`) runs on every `ListRules` / `GetRuleByID` call and iterates the in-memory maintenance store — no DB query added.


pkg/apiserver/signozapiserver/ruler.go
```diff
@@ -116,6 +116,22 @@ func (provider *provider) addRulerRoutes(router *mux.Router) error {
 		return err
 	}
 
+	if err := router.Handle("/api/v2/rules/{id}/mute", handler.New(provider.authzMiddleware.EditAccess(provider.rulerHandler.MuteRule), handler.OpenAPIDef{
```
therealpandey: @vikrantgupta25 this endpoint needs attach.

vikrantgupta25: this would need a `create:planned-maintenance` and `attach`

H4ad (commented):
This will need to be rebased after https://github.com/SigNoz/signoz/pull/11277

Also, for every new css code, use css modules, take a look at any scss file with suffix `.module.scss`.

PR #11274, opened 2026-05-12
chore(pnpm): upgrade deps & add minimum release age

H4ad:
## Pull Request

---

Update dependencies to fix critical/high vulnerabilities.

I also used knip to clean unused deps.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: No
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Dependencies
- Potential regressions: A page breaking due to dependency being bad updated.
- Rollback plan: Revert the dependency.

---

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | Cleaned some dependencies in the frontend & added minimum release age to avoid installing dependencies newer than 2d. |

---

### 📋 Checklist
- [x] Manually tested


tewarig (dismissed):

frontend/knip.json
```diff
@@ -1,5 +1,10 @@
 {
   "$schema": "https://unpkg.com/knip@5/schema.json",
   "project": ["src/**/*.ts", "src/**/*.tsx"],
-  "ignore": ["src/api/generated/**/*.ts", "src/typings/*.ts"]
+  "ignore": ["src/api/generated/**/*.ts", "src/typings/*.ts"],
+  "ignoreDependencies": [
+    "http-proxy-middleware",
+    "react-error-boundary",
+    "@typescript/native-preview"
```
AshwinBhatkal, line 7: what is this dep used for?

AshwinBhatkal (dismissed):

frontend/src/api/generated/services/alerts/index.ts
```diff
@@ -3,6 +3,7 @@
  * * The file has been auto-generated using Orval for SigNoz
  * * regenerate with 'pnpm generate:api'
  * SigNoz
+ * OpenAPI spec version: 0.0.1
```
AshwinBhatkal, line 6: where is this version number coming from? openapi.yaml has empty string version. Has orval made this number default with the new version

frontend/knip.json
H4ad: http proxy: for tests
react-error-boundary: I thought this was being used but seems like its safe to remove
typescript native: for tsgo/tsc checks

frontend/src/api/generated/services/alerts/index.ts
H4ad: This was always in the `orval.config.ts`, but orval always had empty value for version, seems like it fallsback to 0.0.1 when we not set on `openapi.yaml`.

AshwinBhatkal: This was there in orval.config.ts, yes. But the info object didn't have version earlier. This is confirmed with the previous absence of the line. info object version falls back to 0.0.1 with the newer version or orval it looks like

AshwinBhatkal (approved):

tewarig (approved):

Issue #9845, opened 2025-12-22
[BUG] Alert Title Rename Shows Success Message but Changes Are Not Saved


qkrqkstm:
## Bug description

When I rename the title of an existing alert and click Save, a popup says it was completed successfully, but the change is not actually applied.

## How to reproduce

1. Go To Alerts
2. Go into the already created Alert.
3. Click Rename in the Alert title, edit, and save everything.

## Version information 
* **Signoz version**: 102.1
* **Browser version**:
* **Your OS and version**:
* **Your CPU Architecture**(ARM/Intel):

ankitnayan:
not able to reproduce. Please share a recording

qkrqkstm:
https://github.com/user-attachments/assets/b5b364a6-03aa-452b-93ce-a0361ba4e50d

Hey @ankitnayan, I'm sharing a screen recording with you.

ankitnayan:
@amlannandy PTAL

DipeshIngle13:
Is this still open, or has someone already taken it? I can work on it.

Jatin123lodhi:
Now as I can see we can directly click on the title and edit, there is no option to click on 3 dots and choose rename. So this is working fine and it can be closed.

<img width="1916" height="496" alt="Image" src="https://github.com/user-attachments/assets/66b0da29-16ef-4e48-969f-947b6fba02e7" />

PR #11295, opened 2026-05-13
fix(alerts): invalidate rule cache after disable/enable toggle

### LINKED ISSUE #11290: Alert Disable Toggle not respected
## Bug description
When editing any alert, the Disable toggle is available, yet IF one were to "disable" the alert, while editing it, then "Save" it, the toggle config doesn't stick, the "Save" action overrides it and renders the alert "Enabled"

## Expected behavior
Alert remains Disabled as per Toggled disabled setting

## How to reproduce

1. Edit alert
2. Disable it
3. Save alert
4. Alert is re-Enabled

## Version information 
* Signoz: 0.113.0
* Firefox: Version 150.0.1 (64-bit)
* MacOS: Tahoe v26.2
* Architecture: ARM

jatinderjit:
### 📄 Summary

The Disable/Enable toggle on alert rules was being reverted whenever a user edited and saved the rule without reloading the page. The toggle updated the disabled state, but didn't invalidate the react-query cache used by the Edit form. When user saved the form, it sent the old values to the server, silently reverting the user's toggle.

The fix invalidates the generated `useGetRuleByID` query key after a successful toggle. This forces a refetch so the next save sends the correct disabled state.

NA

#### Issues closed by this PR

- Closes #11290

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause

The two disable/enable toggle paths (`ToggleAlertState` and `useAlertRuleStatusToggle`) called patchRulePartial but never invalidated the generated useGetRuleByID query. On save, the form serialized the stale flag, overwriting the toggled state on the server.

#### Fix Strategy

After a successful call to enable/disable a rule, invalidate the cache so the Edit form refetches fresh data.

---

### 🧪 Testing Strategy

- Tests added/updated:
- Manual verification: Open an alert page. Enable/disable it. The click the save button. On reload, the disable rule toggle should not have reverted.
- Edge cases covered:

---

- Blast radius: Alert rules enable/disable toggle and the edit form.
- Potential regressions: Not a regression, but there there would be an extra API call after invalidation.
- Rollback plan: Revert the PR

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Description | Enabling/disabling an alert rule would no longer silently revert that status on subsequent save. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/frontend/src/container/ListAlertRules/ToggleAlertState.tsx
+++ b/frontend/src/container/ListAlertRules/ToggleAlertState.tsx
@@ -1,6 +1,8 @@
 import { Dispatch, SetStateAction, useState } from 'react';
+import { useQueryClient } from 'react-query';
 import { patchRulePartial } from 'api/alerts/patchRulePartial';
 import { convertToApiError } from 'api/ErrorResponseHandlerForGeneratedAPIs';
+import { getGetRuleByIDQueryKey } from 'api/generated/services/rules';
 import type {
 	RenderErrorResponseDTO,
 	RuletypesRuleDTO,
@@ -28,6 +30,7 @@ function ToggleAlertState({
 
 	const { notifications } = useNotifications();
 	const { showErrorModal } = useErrorModal();
+	const queryClient = useQueryClient();
 
 	const onToggleHandler = async (
 		id: string,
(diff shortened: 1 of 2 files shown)


H4ad (approved):

H4ad:
We still have the issue with https://github.com/SigNoz/signoz/issues/9845

But this is safe enough to merge.

PR #11300, opened 2026-05-13
fix(infra-monitoring): error due to invalid operators on query builder

H4ad:
## Pull Request

---

Add support for the following operators:
- nin
- nlike
- notilike
- nregex
- nexists
- ncontains

https://github.com/user-attachments/assets/01198b8a-f0a0-42d0-b48d-b0444ebbdc1c

Fixes https://github.com/SigNoz/engineering-pod/issues/4680

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

The API does not handle operators like NOT_IN.

Hide not supported operators and use the ones we actually support.

---

- Tests added/updated: -
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Infrastructure Monitoring and pages using QueryBuilderSearch.
- Potential regressions: An issue related to bad handling of operators.
- Rollback plan: Revert this commit or fix the issue directly.

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Fixed the issue that was not allowing to use operators of not in, ncontains, and other negate operators inside the query builder. |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/constants/queryBuilder.ts
+++ b/frontend/src/constants/queryBuilder.ts
@@ -431,6 +431,31 @@ export const OPERATORS = {
 	NOTILIKE: 'NOT_ILIKE',
 };
 
+/**
+ * Maps short-form InfraMonitoring operators to long-form display labels.
+ * InfraMonitoring backend uses short forms (NIN), UI displays long forms (NOT_IN).
+ */
+export const INFRA_SHORT_TO_LONG_OPERATOR_MAP: Record<string, string> = {
+	NIN: 'NOT_IN',
+	NLIKE: 'NOT_LIKE',
+	NOTILIKE: 'NOT_ILIKE',
+	NREGEX: 'NOT_REGEX',
+	NEXISTS: 'NOT_EXISTS',
+	NCONTAINS: 'NOT_CONTAINS',
+};
+
+/**
+ * Maps long-form operators to short-form for InfraMonitoring API.
+ */
+export const INFRA_LONG_TO_SHORT_OPERATOR_MAP: Record<string, string> = {
+	NOT_IN: 'NIN',
+	NOT_LIKE: 'NLIKE',
+	NOT_ILIKE: 'NOTILIKE',
+	NOT_REGEX: 'NREGEX',
+	NOT_EXISTS: 'NEXISTS',
+	NOT_CONTAINS: 'NCONTAINS',
+};
+
 export const QUERY_BUILDER_OPERATORS_BY_TYPES = {
 	string: [
 		OPERATORS['='],
--- a/frontend/src/hooks/queryBuilder/useOperatorType.ts
+++ b/frontend/src/hooks/queryBuilder/useOperatorType.ts
@@ -27,6 +27,13 @@ export const operatorTypeMapper: Record<string, OperatorType> = {
 	[OPERATORS['!=']]: 'SINGLE_VALUE',
 	[OPERATORS.HAS]: 'SINGLE_VALUE',
 	[OPERATORS.NHAS]: 'SINGLE_VALUE',
+	// Short-form operators for InfraMonitoring
+	NIN: 'MULTIPLY_VALUE',
+	NLIKE: 'SINGLE_VALUE',
+	NOTILIKE: 'SINGLE_VALUE',
+	NREGEX: 'SINGLE_VALUE',
+	NEXISTS: 'NON_VALUE',
+	NCONTAINS: 'SINGLE_VALUE',
 };
 
 export const useOperatorType = (operator: string): OperatorType =>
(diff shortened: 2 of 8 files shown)


srikanthccv:
Hi @H4ad , this is not the right fix i.e add `nin`, `nlike`, ... as the operator. They were already the supported operators. It's just that the operators have a different label and value. The `NOT_IN` is the operator label with value `nin` under the hood. The fix should only include sending the right operator value. This is broken in some recent versions, so please also share the change that broke the behaviour.

H4ad:
@srikanthccv Changed to `NOT_IN`, but I think it's very delicate to keep this since we need to detect when is `infraMonitoring` and then convert back to `nin` internally. We should migrate to expressions soon.

Also, this was breaking since it was introduced likely, the latest version I found this broken was on `v116` (didn't tested earlier versions), the previous behavior was to return `no data` but in the network tab you can see it got error from backend. (tested this on nightwatch). This version was before any change in the infra monitoring.

Another solution could just update the backend to accept/map the operators based in the new ones, this will fix this component and many others that could rely on this not in.

frontend/src/hooks/queryBuilder/useTag.ts
```diff
@@ -16,19 +17,29 @@ import { WhereClauseConfig } from './useAutoComplete';
...
  * @returns {string[]} An array of formatted conditions. Eg: `["service = web", "severity_text = INFO"]`)
  */
-export function queryFilterTags(filter: TagFilter): string[] {
+export function queryFilterTags(
+	filter: TagFilter,
+	isInfraMonitoring?: boolean,
+): string[] {
 	return (filter?.items || []).map((ele) => {
-		if (isInNInOperator(getOperatorFromValue(ele.op))) {
+		const rawOp = getOperatorFromValue(ele.op);
```
AshwinBhatkal, line 28: nit: rawOperator. Same for displayOperator

frontend/src/hooks/queryBuilder/useOperatorType.ts
```diff
@@ -27,6 +27,13 @@ export const operatorTypeMapper: Record<string, OperatorType> = {
 	[OPERATORS['!=']]: 'SINGLE_VALUE',
 	[OPERATORS.HAS]: 'SINGLE_VALUE',
 	[OPERATORS.NHAS]: 'SINGLE_VALUE',
+	// Short-form operators for InfraMonitoring
+	NIN: 'MULTIPLY_VALUE',
```
AshwinBhatkal, line 31: Aren't these the same as the above operators?

frontend/src/constants/queryBuilder.ts
```diff
@@ -431,6 +431,31 @@ export const OPERATORS = {
 	NOTILIKE: 'NOT_ILIKE',
 };
 
+/**
+ * Maps short-form InfraMonitoring operators to long-form display labels.
+ * InfraMonitoring backend uses short forms (NIN), UI displays long forms (NOT_IN).
+ */
+export const INFRA_SHORT_TO_LONG_OPERATOR_MAP: Record<string, string> = {
```
AshwinBhatkal, line 438: nit: INFRA_MONITORING_SHORT_TO_LONG_OPERATOR_MAP

AshwinBhatkal: same below

frontend/src/hooks/queryBuilder/useOperatorType.ts
H4ad: OPERATORS.NIN has the value `NOT_IN` 🫠

AshwinBhatkal: can we have it as OPERATORS.NIN here? Also can the type be changed accordingly
Record<(typeof OPERATORS)[keyof typeof OPERATORS], OperatorType>

H4ad: Created https://github.com/SigNoz/engineering-pod/issues/4937 to remove this fix once we migrate this query search to qbv5.

AshwinBhatkal (approved):

frontend/src/constants/queryBuilder.ts
H4ad: Not changing for now due to https://github.com/SigNoz/signoz/pull/11300#discussion_r3250179867

frontend/src/hooks/queryBuilder/useTag.ts
H4ad: Ignored for now due to https://github.com/SigNoz/signoz/pull/11300#discussion_r3250179867

H4ad:
Created https://github.com/SigNoz/engineering-pod/issues/4937 to revert this fix after migrating to expressions in this page.

PR #11302, opened 2026-05-14
fix: resource tag querybuilding in conditionFor

piyushsingariya:
## Pull Request

---

### 📄 Summary

JSONCondition was getting built for normal resource attributes, causing invalid query building.

Before:
```
SELECT toStartOfInterval(fromUnixTimestamp64Nano(timestamp), INTERVAL 30 SECOND) AS ts, toString(multiIf(, resource.`region`::String, NULL)) AS `region`, countDistinct(multiIf(, resource.`service.name`::String, NULL)) AS __result_0 FROM signoz_logs.distributed_logs_v2 WHERE ((dynamicElement(body_v2.`user.name`, 'String') IS NOT NULL) OR mapContains(attributes_string, 'user.name') = ?) AND timestamp >= ? AND ts_bucket_start >= ? AND timestamp < ? AND ts_bucket_start <= ? GROUP BY ts, `region`
```

After:
```
SELECT toStartOfInterval(fromUnixTimestamp64Nano(timestamp), INTERVAL 30 SECOND) AS ts, toString(multiIf(resource.`region`::String IS NOT NULL, resource.`region`::String, NULL)) AS `region`, countDistinct(multiIf(resource.`service.name`::String IS NOT NULL, resource.`service.name`::String, NULL)) AS __result_0 FROM signoz_logs.distributed_logs_v2 WHERE ((dynamicElement(body_v2.`user.name`, 'String') IS NOT NULL) OR mapContains(attributes_string, 'user.name') = ?) AND timestamp >= ? AND ts_bucket_start >= ? AND timestamp < ? AND ts_bucket_start <= ? GROUP BY ts, `region`
```

#### Issues closed by this PR

related to signoz/platform-pod#2265

---

### ✅ Change Type
_Select all that apply_

- [x] 🐛 Bug fix

---

### 🐛 Bug Context
JSONCondition was getting built for normal resource attributes, causing invalid query building.

#### Root Cause
Check was missing for BodyContext

#### Fix Strategy

Add a check for BodyContext 

---

- Tests added/updated: ✅ 
- Manual verification: ✅ 
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---


srikanthccv (changes requested):
Please add integration tests with JSON enabled, for the missing areas from the following
- search
   - attributes, resource attributes, top-level fields
- group by
   - attributes, resource attributes, top-level fields
- aggregation
- order by

tests/integration/tests/querier_json_body/02_non_body_fields.py
```diff
@@ -0,0 +1,425 @@
...
+    token: str,
+    start_ms: int,
+    end_ms: int,
+    name: str,
+    *,
+    expression: str | None = None,
+    order: list[dict] | None = None,
+    limit: int = 100,
+) -> requests.Response:
+    q = build_scalar_query(
```
nityanandagohain: we should build a raw query here ?

tests/integration/tests/querier_json_body/02_non_body_fields.py
```diff
@@ -0,0 +1,425 @@
...
+        )
+        for i, (svc, method, sev, body) in enumerate(log_data)
+    ]
+    export_json_types(logs_list)
+    insert_logs(logs_list)
+    token = get_token(email=USER_ADMIN_EMAIL, password=USER_ADMIN_PASSWORD)
+
+    cases = [
+        # ── top-level field filters ────────────────────────────────────────
+        {
```
nityanandagohain, line 157: one of the three is enough I feel for `severity_text` , unless it will help us find something different.

tests/integration/tests/querier_json_body/02_non_body_fields.py
```diff
@@ -0,0 +1,425 @@
...
+        # resource attr + log attr + body field (all three contexts)
+        {
+            "name": "filter.all_three",
+            "requestType": "raw",
+            "expression": 'service.name = "auth-svc" AND http.method = "POST" AND status = 500',
+            "validate": lambda r: len(get_rows(r)) == 1 and _body_users(r) == {"bob"},
+        },
+        # ── group by ──────────────────────────────────────────────────────
+        # resource attribute: auth-svc×3, api-gw×1, worker×1
+        {
```
nityanandagohain, line 165: a single query with multiple group by should be enough.

nityanandagohain: Also we can add aggregation to these queries itself instead of writing individual test cases.

nityanandagohain (commented):
Added few comments.

Also I think we should trim down the tests and try to check for all functionality while keeping it lean.

tests/integration/tests/querier_json_body/02_non_body_fields.py
piyushsingariya: the main difference comes from make_query_request, where request_type is added.

nityanandagohain: The create a separate helper function instead, as you are passing aggregation as well here. Check if there is any existing helper function as well.

piyushsingariya: addressed

tests/integration/tests/querier_json_body/02_non_body_fields.py
```diff
@@ -0,0 +1,287 @@
...
+# A single 4-log dataset with varied attributes exercises ORDER BY
+# across resource attribute, and top-level fields.
+#
+# Data landscape:
+#   log1 — svc-a, GET,    INFO, score=80, ts=now-4s
+#   log2 — svc-a, POST,   INFO, score=90, ts=now-3s
+#   log3 — svc-b, GET,    WARN, score=60, ts=now-2s
+#   log4 — svc-b, DELETE, WARN, score=70, ts=now-1s
+#
+# Aggregation cross-reference:
```
nityanandagohain: incorrect, we are not doing aggregation below

piyushsingariya: yes, I had separated the tests earlier, removing the comment

nityanandagohain (approved):

srikanthccv (approved):
there are issues with operations array, but unblocking for now.

piyushsingariya:
> there are issues with operations array, but unblocking for now.

Hi @srikanthccv , can you create a ticket on this, will address in next week's release.

therealpandey (approved):

PR #11307, opened 2026-05-14
feat: add functional unique index

namanv3:
Closes https://github.com/SigNoz/pulse-pod/issues/61


ee/sqlschema/postgressqlschema/provider.go
```diff
@@ -224,19 +239,21 @@ SELECT
...
 FROM
     pg_index i
     LEFT JOIN pg_class ct ON ct.oid = i.indrelid
     LEFT JOIN pg_class ci ON ci.oid = i.indexrelid
-    LEFT JOIN pg_attribute a ON a.attrelid = ct.oid
     LEFT JOIN pg_constraint con ON con.conindid = i.indexrelid
+    CROSS JOIN generate_series(1, i.indnkeyatts) AS n
+    LEFT JOIN pg_attribute a ON a.attrelid = ct.oid AND a.attnum = i.indkey[n - 1]
 WHERE
-    a.attnum = ANY(i.indkey)
```
namanv3, line 236: `a.attnum = ANY(I.indkey)` -> dropped cuz expression keys have `indkey` element 0 (no real column/attnum), so none matched and expression keys were silently dropped

pkg/sqlschema/index.go
```diff
@@ -49,13 +51,29 @@ type Index interface {
...
+//     ColumnNames on read-back and won't compare equal — use ColumnNames instead.
 type UniqueIndex struct {
 	TableName   TableName
 	ColumnNames []ColumnName
+	Expressions []string
 	name        string
 }
 
 func (index *UniqueIndex) Name() string {
+	if len(index.ColumnNames) > 0 && len(index.Expressions) > 0 {
```
namanv3: with this constraint in mind, should functional unique indices be their own type?

therealpandey: Functional has a different connotation altogether. Can we call it UniqueIndexWithExpressions if we're creating another type which satisfies the Index interface?

pkg/sqlschema/index.go
```diff
@@ -49,13 +51,29 @@ type Index interface {
...
+		panic("sqlschema: UniqueIndex sets both ColumnNames and Expressions; they are mutually exclusive — for a functional index put every key (plain columns as bare identifiers) in Expressions and leave ColumnNames empty")
```
therealpandey: I think this justifies creating another index type. Please don't panic in these functions.

therealpandey (changes requested):

pkg/sqlschema/index.go
namanv3: done

pkg/sqlschema/index.go
namanv3: done

therealpandey (approved):

Issue #9909, opened 2025-12-31
Allow creating alert from all panel types


srikanthccv:
The "create alert" is hidden for other panels, users should be able to create alert from any panel type.

nishchay-veer:
hi @srikanthccv would like to work on this, can you please assign this one to me?

ankitnayan:
Does this mean we will support table and number panels in alerts?

srikanthccv:
yes, the "cost meter" alert today is number based alert, the current experience is not good. the "number alert" will still have the time series in the dim background with number at the centre of the chart. similarly, many alerts today users setup are table in the sense that, there is one value for the entire period they alert on, example, if total request count for the window > 10k send notification.

srikanthccv:
removed the label; this requires work on the alerts first.

PR #11308, opened 2026-05-14
chore: added changes to migerate alert chart component to new charts

ahrefabhi:
## Pull Request

---

### 📄 Summary
Migrates the alert chart preview's `TIME_SERIES` and `BAR` rendering from the legacy `GridPanelSwitch` + `getUPlotChartOptions` pipeline to the new `<TimeSeries>` / `<BarChart>` components that the dashboards already use. The alert preview now reaches parity with dashboard charts (tooltip pinning, cursor sync, threshold drawing via `UPlotConfigBuilder`) and a parallel chart pipeline is removed from the alert path.

Structure is reorganised around a small `SUPPORTED_CHARTS` lookup in `ChartContent.tsx`. Supporting a new modern-chart panel type is one map entry plus a case in `buildAlertChartConfig` — no more dual `options` vs `config` memos in the parent.

Each alert's chart id is now suffixed with `alertDef.id`, so the per-chart localStorage bucket used for selection preferences (series visibility, zoom range) is unique per alert. Without this, editing one alert would clobber another's saved preferences.

_N/A — visual output matches the dashboard chart components already in production._

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/4922

---

### ✅ Change Type
- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context
N/A — no user-visible bug being fixed. The localStorage-id scoping is a forward-looking guard added alongside the migration to prevent preference collisions once persistence kicks in.

---

### 🧪 Testing Strategy

- **Tests added/updated:** existing `ChartPreview/utils.test.ts` (10) and `CreateAlertV2/.../ChartPreview.test.tsx` (7) pass — the latter trimmed to drop the obsolete `name` prop assertions that are no longer wired through.
- **Manual verification:** alert chart preview rendered with `TIME_SERIES` (single + multi-series, with thresholds) and `BAR` panel types.
- **Edge cases covered:**
  - Series with only one valid plot point — `hasSingleVisiblePoint` was duplicated between `TimeSeriesPanel/utils.ts` and the alert preview; consolidated into `lib/uPlotV2/utils/dataUtils.ts` next to `isInvalidPlotValue` and decoupled from the API-layer `QueryData` type so the lib stays free of API imports.
  - Panel types outside `TIME_SERIES`/`BAR` (e.g. `VALUE`, `TABLE`, `PIE`) — `ChartContent` returns `null`. Alert chart preview is not exercised with those panel types in any current flow.
  - Create-alert flow (no alert id yet) — falls back to the bare `CHART_ID_PREFIX`; transient prefs only.

---

- **Blast radius:** alert chart preview only. Anomaly-detection alerts (`AnomalyAlertEvaluationView`) and the underlying `useGetQueryRange` data flow are untouched.
- **Potential regressions:**
  - Series-visibility toggle and legend scroll-position memory (previously persisted via `getLocalStorageGraphVisibilityState`) are dropped — the new wrapper handles its own legend. Matches dashboard behaviour today.
  - `baseConfigBuilder` is touched (refactored `persistSelectionPreference` flag). Behaviour for `DASHBOARD_VIEW` is unchanged; `STANDALONE_VIEW` and "no panel mode" paths are unchanged at the call sites in use today.
- **Rollback plan:** revert this PR — no schema, API, or stored-state migrations involved.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | N/A — internal refactor; user-visible chart behaviour matches the dashboard chart components already in use. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- Bulk of the diff is the new `ChartPreview/ChartContent.tsx` dispatcher (a `SUPPORTED_CHARTS` map plus a small `isSupportedPanelType` type-guard) and the new `buildAlertChartConfig` helper in `ChartPreview/utils.ts`. `ChartPreview/index.tsx` shrinks by ~60 lines — the dual config/options memos and the `renderChart` helper are gone, replaced by a single `<ChartContent />`.
- `ChartContent.tsx` derives its chart id as `alert_legend_widget_<alertDef.id>` via `buildChartId(alertId)`. The id is what `UPlotConfigBuilder` keys localStorage on, so this prevents one alert's preferences from clobbering another's.
- `name` prop was removed from `ChartPreviewProps` along with all three call sites — it was never consumed inside the component after the migration. Two test assertions on `name` were dropped accordingly.
- `hasSingleVisiblePoint` was duplicated; moved into `lib/uPlotV2/utils/dataUtils.ts` and made API-type-free (takes the values array directly instead of `QueryData`).
- `baseConfigBuilder.ts` change is a small readability refactor (collapses the two related flags onto a single `persistSelectionPreference` variable). Persistence behaviour is unchanged for existing callers.


frontend/src/container/FormAlertRules/ChartPreview/index.tsx
```diff
@@ -77,7 +77,6 @@ export interface ChartPreviewProps {
 
 // eslint-disable-next-line sonarjs/cognitive-complexity
 function ChartPreview({
-	name,
```
AshwinBhatkal, line 80: what was this used for?

frontend/src/lib/uPlotV2/utils/dataUtils.ts
```diff
@@ -52,6 +52,28 @@ export function normalizePlotValue(
...
 }
 
+/**
+ * Returns true if at most one entry in `values` is a valid plot value.
+ *
+ * Used to decide whether a series should render as a single point (drawStyle:
+ * Points) vs a line — a continuous line with only one visible sample is
+ * invisible to the user.
+ */
+export function hasSingleVisiblePoint(
```
AshwinBhatkal, line 68: we already have a function for this noe?

frontend/src/container/FormAlertRules/ChartPreview/ChartContent.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { LegendPosition } from 'lib/uPlotV2/components/types';
+import { MetricRangePayloadProps } from 'types/api/metrics/getQueryRange';
+import { Query } from 'types/api/queryBuilder/queryBuilderData';
+import uPlot from 'uplot';
+
+import { AlertChartPanelType, buildAlertChartConfig } from './utils';
+
+const CHART_ID_PREFIX = 'alert_legend_widget';
+
+const buildChartId = (alertId?: string): string =>
```
AshwinBhatkal: move these constants to utils and constants files

frontend/src/container/FormAlertRules/ChartPreview/ChartContent.tsx
```diff
@@ -0,0 +1,119 @@
...
+	width: number;
+	height: number;
+	minTimeScale?: number;
+	maxTimeScale?: number;
+	onDragSelect: (start: number, end: number) => void;
+}
+
+export default function ChartContent({
+	panelType,
+	alertId,
```
AshwinBhatkal, line 53: call this id and pass alertId

frontend/src/container/FormAlertRules/ChartPreview/ChartContent.tsx
```diff
@@ -0,0 +1,119 @@
...
+const buildChartId = (alertId?: string): string =>
+	alertId ? `${CHART_ID_PREFIX}_${alertId}` : CHART_ID_PREFIX;
+
+// Panel types that render through the UPlotConfigBuilder pipeline.
+// To support a new modern-chart panel type, add an entry here and extend
+// `AlertChartPanelType` / `buildAlertChartConfig` to handle its series setup.
+const SUPPORTED_CHARTS: Record<
+	AlertChartPanelType,
+	typeof TimeSeries | typeof BarChart
+> = {
```
AshwinBhatkal, line 24: What happens in the case of number type panel? Today we have option to create alert for number type panel as well. Internal conversion to time series?

AshwinBhatkal: Please think about this one as well - https://github.com/SigNoz/signoz/issues/9909
Need not make any change for that as such right now but would be good to evaluate how this solution can extend to the above

AshwinBhatkal (commented):
few comments

frontend/src/lib/uPlotV2/utils/dataUtils.ts
ahrefabhi: yes just moved that function to a utility and using this at both places.

frontend/src/container/FormAlertRules/ChartPreview/ChartContent.tsx
ahrefabhi: We don't support number panel as of now, We want to support all the panels going forward, and that's a separate task altogether.

and based on how things are implemented right now, It'll be easy to extend changes to new panel types going forward.
cc: @srikanthccv

frontend/src/container/FormAlertRules/ChartPreview/index.tsx
ahrefabhi: In the pre-migration code name was forwarded to GridPanelSwitch as name={name || 'Chart Preview'} — the legacy uPlot pipeline used it as the widget id (e.g. for getLocalStorageGraphVisibilityState).

After the migration the chart id is derived from alertDef.id via buildChartId, so nothing inside ChartPreview consumes name any more. Removed it from ChartPreviewProps and the three call sites; two test assertions on it were dropped too.

AshwinBhatkal (approved):

PR #11320, opened 2026-05-15
chore: added changes for sorting tooltip content

### LINKED ISSUE #11321: Sort tooltip content in descending order
## Bug description

*Please describe.*  
*If this affects the front-end, screenshots would be of great help.*  

## How to reproduce

1.
2.
3.

## Version information 
* **Signoz version**:
* **Browser version**:
* **Your OS and version**:

ahrefabhi:
## Pull Request

---

### 📄 Summary

Sort tooltip rows by value in descending order so the largest series appears at the top of the uPlotV2 tooltip. Previously rows were emitted in series-index order, which made it hard to identify the top contributors at a glance — especially with many series. The sort is added once at the end of `buildTooltipContent`, which is already memoized on `dataIndexes` and rendered inside a rAF-throttled cycle by `TooltipPlugin`, so per-frame cost on a 50–100 series tooltip is negligible compared to React reconciliation.

`'No Data'` rows (synced mode) and any genuine zero/negative values land at the bottom because they share `value: 0` or lower.

image

#### Issues closed by this PR

Closes https://github.com/SigNoz/signoz/issues/11321

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

(Closest match: ♻️ Refactor / UX polish — keeping `chore` to match the commit.)

---

### 🐛 Bug Context

_N/A — not a bug fix._

---

### 🧪 Testing Strategy

- Tests added/updated:
  - `frontend/src/lib/uPlotV2/components/Tooltip/__tests__/utils.test.ts` — updated the two tests that previously asserted series-index order (`builds tooltip content…` and `returns items in series-index order`) to assert descending-by-value ordering. Full suite (12 tests) passes locally.
- Manual verification:
  - Hover a time series panel with several visible series and confirm rows are sorted top→bottom by value.
  - Verify stacked bar tooltip still uses the raw (unstacked) value for sort comparison — order matches the stacked segment magnitudes.
  - Verify synced tooltips: `'No Data'` rows sink to the bottom.
- Edge cases covered:
  - Hidden series (skipped before sort).
  - Stacked bar charts (sort uses base value, post-unstacking).
  - Sync `Filtered` / `All` modes.
  - Single-row tooltips (no observable change).

---

- Blast radius: every uPlotV2 tooltip — TimeSeries, BarChart, Histogram — across the dashboards/explorer surfaces that have migrated to uPlotV2.
- Potential regressions:
  - Anything (snapshots, downstream consumers, screenshot-based tests) that relied on series-index ordering. None found in this codebase outside the tests already updated here.
  - Active row is no longer guaranteed to be at a specific position in the list; the header still renders the active item independently, so behavior is unchanged for users.
- Rollback plan: revert the single `items.sort(...)` line in `buildTooltipContent` ([utils.ts](https://github.com/SigNoz/signoz/blob/chore/tooltip-sorting/frontend/src/lib/uPlotV2/components/Tooltip/utils.ts)).

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | uPlotV2 chart tooltips now list series sorted by value (largest first), making top contributors easier to spot at a glance. |

---

### 📋 Checklist
- [x] Manually tested
- [x] Breaking changes documented (none — ordering-only change)

---

## 👀 Notes for Reviewers

- The sort runs inside `buildTooltipContent` which is wrapped in `useMemo` keyed on `dataIndexes` and downstream of a `requestAnimationFrame` in `TooltipPlugin`, so it executes at most once per frame per data-column change — not per pixel of cursor movement. Sort cost on 50–100 items is microseconds and is dominated by the React reconciliation already happening per frame.
- `'No Data'` rows have `value: 0`, so they will be interleaved with any genuine zero values rather than always sinking to the very bottom. Happy to switch to a two-key sort (real values first, `'No Data'` last) if reviewers prefer.

---


srikanthccv:
Discussed on Slack about the behaviour. It wasn't like this earlier. Please take some time to investigate and help me understand when did this behaviour change. It's important to understand how are regressions in functional behaviour are getting introduced in the system. So that we can put things in place to prevent them getting shipped to users.

ahrefabhi:
> Discussed on Slack about the behaviour. It wasn't like this earlier. Please take some time to investigate and help me understand when did this behaviour change. It's important to understand how are regressions in functional behaviour are getting introduced in the system. So that we can put things in place to prevent them getting shipped to users.

This was not regression @srikanthccv, This was a choice made to keep the experience smooth while moving between the panel, There was a miss on my end, I should've communicated this better with the team. Will take care of things like these going forward.

AshwinBhatkal (approved):

PR #11326, opened 2026-05-16
fix: resolve contextLinks variable placeholders for per-row values

### LINKED ISSUE #11325: (task) Dashboard contextLinks URL templates don't substitute per-row groupBy values on table panel aggregate-click
## What happened

On a dashboard Query Builder **table** panel with a `contextLinks` configuration like:

```json
{
  "contextLinks": {
    "linksData": [
      {
        "label": "Open trace",
        "url": "/trace/{{trace_id}}?spanId={{span_id}}&levelUp=0&levelDown=0",
        "id": "..."
      }
    ]
  }
}
```

and a query that groups by `trace_id`, `span_id`, clicking the aggregate-column drilldown menu item navigates to the URL with the template placeholders **literally URL-encoded**, not substituted:

```
/trace/%7B%7Btrace_id%7D%7D?spanId=%7B%7Bspan_id%7D%7D&levelUp=0&levelDown=0
```

instead of (expected):

```
/trace/<actual-trace-id-from-row>?spanId=<actual-span-id-from-row>&levelUp=0&levelDown=0
```

## Where the gap is

The `contextLinks` plumbing is partly wired, but the per-row value injection never happens.

- `frontend/src/container/QueryTable/Drilldown/useBaseAggregateOptions.tsx` calls `useContextVariables({ customVariables: fieldVariables })`, where `fieldVariables` is built from the query's groupBy keys with **empty string values** (e.g. `{ trace_id: '', span_id: '' }`).
- `frontend/src/hooks/dashboard/useContextVariables.tsx` then registers those names with a `_` prefix and `value === ''`. Its `resolveTexts` does support `{{var}}`, `{{.var}}`, `$var` placeholder syntaxes — but since the registered values are empty, every placeholder resolves to `''` (and the bracketed form ends up URL-encoded).
- The clicked row's actual cell values are never threaded into `useContextVariables` from `useTableContextMenu` / `useBaseAggregateOptions`, so per-row substitution is structurally impossible today, not just a config issue.

## Why it matters

Without per-row substitution, table panel `contextLinks` can only point at static URLs — they can't deep-link to a row's specific trace, span, host, pod, etc. That defeats most realistic drilldown use cases (jump-to-trace from an audit table, jump-to-pod from a Kubernetes table, etc.).

The UI exposes the `{{var}}` syntax and the documentation in `useContextVariables.tsx` describes per-row substitution, so today's behavior reads as a bug rather than an unfinished feature.

## Reproduction

1. Create a Query Builder table widget on a dashboard.
2. Query logs/traces with `groupBy: [trace_id, span_id]` and an aggregate (e.g. `count()`).
3. In the widget's contextLinks UI, add a link with URL `/trace/{{trace_id}}?spanId={{span_id}}&levelUp=0&levelDown=0`.
4. Click the aggregate cell on any row → "Open trace" in the drilldown menu.
5. Observe the new tab/route opens with the literal/encoded `{{trace_id}}` and `{{span_id}}` in the path/query.

Tested on SigNoz **v0.113.0** (self-hosted).

## Suggested fix direction (for reference, not prescriptive)

`useTableContextMenu` already has the clicked row in scope when it builds the drilldown menu. Threading the row's groupBy values into the `customVariables` map passed to `useContextVariables` — instead of empty strings derived from key names — should be enough for `resolveTexts` to substitute correctly. The `_` prefix indirection in `useContextVariables` may also need a look, since the registered name diverges from the placeholder syntax used in URLs.

## Related

- #11214 — feature request to extend `contextLinks` to Clickhouse Queries on dashboards. That's a separate gap (query-type coverage); this issue is about per-row substitution on the Query Builder table case that *is* already wired up.

## Environment

- SigNoz: v0.113.0 (self-hosted)
- Browser: Chrome (reproduces independent of browser; substitution happens server-agnostically in the frontend)

lloyd-c137:
## Summary

Fixes #11325.

When a dashboard table panel has `contextLinks` configured with URL templates like:
```json
{
  "url": "/trace/{{trace_id}}?spanId={{span_id}}"
}
```

the `{{trace_id}}` and `{{span_id}}` placeholders were left URL-encoded in the final link instead of being replaced with the actual row values.

## Root cause

`useContextVariables` registers custom (per-row) field variables with a leading `_` prefix to avoid conflicts with dashboard and global variables (e.g. `_trace_id`, `_span_id`). The `resolveText` and `resolveTextWithTruncation` utility functions, however, looked up variable names exactly as extracted from the placeholder syntax (`{{trace_id}}` → `trace_id`), so the `_`-prefixed custom variables were never matched.

## Fix

Added a fallback lookup in both `resolveText` and `resolveTextWithTruncation`: when the exact variable name is not found, also try the `_`-prefixed variant before falling back to leaving the placeholder unresolved.

## Validation

- `resolveText("/trace/{{trace_id}}", { _trace_id: "abc123" })` → `"/trace/abc123"` (was `"/trace/{{trace_id}}"`)
- Existing `$var` and `[[var]]` placeholder syntaxes continue to work unchanged
- Dashboard and global variables without `_` prefix continue to match as before

## Related

- `frontend/src/hooks/dashboard/useContextVariables.tsx` — both `resolveText` and `resolveTextWithTruncation` updated

### DIFF
--- a/frontend/src/hooks/dashboard/useContextVariables.tsx
+++ b/frontend/src/hooks/dashboard/useContextVariables.tsx
@@ -233,7 +233,12 @@ const resolveText = (
 
 	return text.replace(combinedPattern, (match) => {
 		const varName = extractVarName(match, matcher, processedVariables);
-		const value = processedVariables[varName];
+		// Try the variable name as-is, then with the '_' prefix used by useContextVariables
+		// for custom (per-row) variables to avoid conflicts with dashboard/global variables.
+		let value = processedVariables[varName];
+		if (value == null) {
+			value = processedVariables[`_${varName}`];
+		}
 
 		if (value != null) {
 			const parts = value.split('-|-');
@@ -254,7 +259,11 @@ const resolveTextWithTruncation = (
 
 	const result = text.replace(combinedPattern, (match) => {
 		const varName = extractVarName(match, matcher, processedVariables);
-		const value = processedVariables[varName];
+		// Try the variable name as-is, then with the '_' prefix
+		let value = processedVariables[varName];
+		if (value == null) {
+			value = processedVariables[`_${varName}`];
+		}
 
 		if (value != null) {
 			const parts = value.split('-|-');


lloyd-c137:
I have amended the commit with the correct email associated with my GitHub account (zhaoanran666@foxmail.com). Please re-check the CLA status.

lloyd-c137:
Closing this PR as it's been open and inactive for an extended period. Can re-open if needed.

PR #11332, opened 2026-05-17
test: added e2e tests for dashboard create flow

ahrefabhi:
## Pull Request

---

### 📄 Summary

Adds end-to-end coverage for the dashboard creation flow under `tests/e2e/tests/dashboards/create.spec.ts` — 13 Playwright TCs exercising every UI entrypoint into "create a dashboard" plus the post-create configuration journey. Deduplicated from an earlier 20-TC pass: drawer/variable/panel-modal coverage was merged into the round-trip TC-13 or moved to a more appropriate spec file, eliminating overlapping UI walks.

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/4524

---

### ✅ Change Type

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [x] 🧪 Test-only

---

### 🧪 Testing Strategy

**Tests added** — three files, all under `tests/e2e/`:

| File | What it adds |
|------|--------------|
| `tests/dashboards/create.spec.ts` | 13 TCs across blank create, toolbar rename, Import JSON, View Templates, navigation, per-signal panel-save, delete, and a full settings+variables+panel round-trip |
| `helpers/dashboards.ts` | Helpers used by the new spec: `openDashboardSettingsDrawer`, `saveDashboardSettings`, `renameDashboardViaToolbar`, `configureAndSavePanel` (drives the widget editor for metrics / logs / traces using `testdata/queries.json`) |
| `testdata/queries.json` | Per-signal query inputs the panel-add helper consumes (metric name `signoz_calls_total` for the metrics signal; empty queries for logs/traces use Query Builder defaults) |

**TC coverage map (13 TCs):**
- TC-01: blank create → POST contract + onboarding state
- TC-02: rename via toolbar options popover persists to title + list
- TC-03: Import JSON via file upload — POST contract, navigation, widgets/tags persisted, list-view tag chip
- TC-04: Import JSON invalid file shows "Invalid JSON" and never POSTs
- TC-05: Import JSON empty editor shows error and never POSTs
- TC-06: "New Dashboard" dropdown has the three expected entries; "View templates" is an external docs link with proper `target="_blank"` + `rel=noopener` (guards against the dead in-app `DashboardTemplatesModal`)
- TC-07: browser Back from dashboard detail restores list URL + search input
- TC-08: navigating away with the settings drawer open does not crash
- TC-09–11: Add Panel end-to-end per signal (metrics / logs / traces) — Time Series, configure query, Save, panel renders, survives reload
- TC-12: Delete dashboard via list action menu — DELETE request fires, row gone, GET returns 404
- TC-13: Full round-trip — settings (title/description/tag) + Custom variable + metrics panel all persist across a hard reload, verified both in the UI and the server JSON

**Deduplication done in `4b3518c`** — removed seven tests that overlapped with TC-13 (round-trip) or with `details/35-add-panel.spec.ts`:
- Drawer open + pre-fill check, drawer rename/description/tags save, drawer Discard, Custom-variable add, duplicate-variable-name rejection → either subsumed by TC-13's round-trip, or unique-enough to defer
- Two "New Panel modal opens" TCs (empty-state and toolbar entry points) → already covered in `tests/e2e/tests/dashboards/details/35-add-panel.spec.ts`

**Manual verification** — ran the suite headed against the local stack (port from `tests/e2e/.env.local`); all 13 TCs pass on chromium. TC-09 was also run with a `page.pause()` so the metrics chart could be visually verified rendering on the dashboard grid.

**Edge cases covered:**
- Cleanup integrity when a test creates a dashboard via UI (no upfront ID): the POST response body's `data.id` is captured and registered in `seedIds` so `afterAll` cleans it up (TC-01, TC-03).
- TC-04 / TC-05 assert no POST fires by intercepting `/api/v1/dashboards` with `page.route` — guards against an invalid-input regression that silently creates a broken dashboard.
- TC-03 verifies both the POST contract and downstream list-view tag-chip surfacing, so a regression in either is caught.
- TC-12 verifies deletion via both UI state (empty-search message) and the API (`GET /api/v1/dashboards/{id}` returns 404), catching optimistic-update bugs where the row vanishes but the backend keeps the record.

---

- **Blast radius**: zero on production code paths. All changes are confined to `tests/e2e/`.
- **Potential regressions**: only test flakiness — the new TCs rely on existing testids (`add-panel`, `panel-name-input`, `metric-name-selector-0`, `query-data-source-selector-0`, `new-widget-save`, etc.) that are already in the source. The Ant Design dropdown locators (`.ant-select-item-option-content`) are the most fragile — if Ant's class names shift, those locators break in `configureAndSavePanel`.
- **Rollback plan**: revert the commit. Nothing downstream consumes these tests except CI.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | N/A |
| Change Type | N/A |
| Description | N/A — internal test coverage; no user-facing behavior change |

---

### 📋 Checklist
- [x] Manually tested
- [x] Backward compatibility considered (test-only; no API/UI changes)

---

## 👀 Notes for Reviewers

- The spec assumes the stack is up via `tests/e2e/bootstrap/setup.py::test_setup`. See `docs/contributing/tests/e2e.md` for the bring-up command; the suite needs `--with-web` so the SigNoz container has the frontend bundled in.
- The Monaco-paste import path is intentionally not covered — TC-03 (file upload) exercises the same populate-editor-then-import code path, and keyboard-typing large JSON into Monaco is unreliable in headless CI.
- New Panel modal opening + widget-editor navigation lives in `tests/e2e/tests/dashboards/details/35-add-panel.spec.ts`; this spec picks up from there with the per-signal save-and-render TCs (TC-09–11).
- `configureAndSavePanel` in `helpers/dashboards.ts` is the place to extend if you want to cover non-Time-Series panel types (Value, Table) in a follow-up. Today it's hard-coded to `panel-type-graph` since `queries.json` is shaped around the Time Series widget.
- Cleanup: every UI-created dashboard registers its ID into the suite-level `seedIds` set and `afterAll` deletes them via `deleteDashboardViaApi`. TC-12 deletes its dashboard through the UI so it intentionally skips registration. Panels and variables live inside the dashboard document, so deleting the dashboard cascades — no per-entity cleanup needed.


AshwinBhatkal:
Please base this off of - https://github.com/SigNoz/signoz/pull/11279

Also, for changes in specs, please add safe-to-e2e label to trigger the check

AshwinBhatkal:
1. Please update the tests after the rebase. I think some of them can be removed. Some of them covered in details folder tests and some in list.spec.ts
2. For dashboard details, move the tests inside details folder. See if the tests can be broken down into the already existing suites. Create a new suite only if required
3. Let's not use safe-to-integrate if there is only change in e2e folder
4. Check for fmtlint failure

AshwinBhatkal:
1 - Some tests are already covered in list.spec.ts. Please take a look at them once
2 - Move tests inside details folder. Use numbering to place it the right way
3 - Let's not use safe-to-integrate label. Changes to e2e folder can be tested only with safe-to-integrate

ahrefabhi:
> 1 - Some tests are already covered in list.spec.ts. Please take a look at them once 
2 - Move tests inside details folder. Use numbering to place it the right way 
3 - Let's not use safe-to-integrate label. Changes to e2e folder can be tested only with safe-to-integrate


Removed duplicate tests, regarding moving this inside details folder, I feel this tests a very specific case of dashboard creation, which spans from listing to edit, so I don't think we should move this to details folder.

and what tag should I use `safe-to-integrate` or `safe-to-e2e`, you mentioned `safe-to-integrate` twice.

AshwinBhatkal:
> Removed duplicate tests, regarding moving this inside details folder, I feel this tests a very specific case of dashboard creation, which spans from listing to edit, so I don't think we should move this to details folder.

We should separate it out. Creation is part of list. Edit is part of details

> and what tag should I use safe-to-integrate or safe-to-e2e, you mentioned safe-to-integrate twice.

Use [safe-to-e2e](https://github.com/SigNoz/signoz/pull/11332#issuecomment-4482162591). I mentioned *not* to use safe-to-integrate. safe-to-integrate is for BE integration tests

tests/e2e/tests/dashboards/create.spec.ts
```diff
@@ -0,0 +1,382 @@
...
+
+	// ─── 5. Full Round-Trip ──────────────────────────────────────────────────
+	//
+	// Delete-via-list coverage lives in list.spec.ts (TC-21).
+	//
+	// Catches cross-feature regressions: a settings save that nukes variables,
+	// a variable add that strips widgets, a panel save that overwrites tags, etc.
+	// Stress-tests the dashboard PUT contract by writing every editable surface.
+
+	test('TC-13 settings + variable + panel survive a hard reload', async ({
```
AshwinBhatkal: this can be dropped I believe

AshwinBhatkal (commented):
please update test case numbers. Rest looks good

PR #11334, opened 2026-05-18
refactor: only select the required map column in flamegraph query

krsoninikhil:
## Pull Request

---

When we added feature to allow flamegraph client to request certain attributes (https://github.com/SigNoz/signoz/pull/11092/changes?w=1), we started querying all the attributes in the flamegraph query which is not required. In our current UI implementation, we are only requesting some resource attributes, so this change will help in avoiding the unnecesary loading of unused attirbutes.

Part of https://github.com/SigNoz/engineering-pod/issues/4949

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/pkg/query-service/model/queryParams.go
+++ b/pkg/query-service/model/queryParams.go
@@ -346,6 +346,27 @@ type GetFlamegraphSpansForTraceParams struct {
 	SelectFields    []telemetrytypes.TelemetryFieldKey `json:"selectFields"`
 }
 
+func (r *GetFlamegraphSpansForTraceParams) GetSelectedFeildsSourceColumns() []string {
+	needsAttrMap, needsResourceMap := false, false
+	for _, f := range r.SelectFields {
+		switch f.FieldContext {
+		case telemetrytypes.FieldContextAttribute:
+			needsAttrMap = true
+		case telemetrytypes.FieldContextResource:
+			needsResourceMap = true
+		}
+	}
+
+	var cols []string
+	if needsAttrMap {
+		cols = append(cols, "attributes_string", "attributes_number", "attributes_bool")
+	}
+	if needsResourceMap {
+		cols = append(cols, "resources_string")
+	}
+	return cols
+}
+
 type SpanFilterParams struct {
 	TraceID            []string `json:"traceID"`
 	Status             []string `json:"status"`
--- a/pkg/query-service/model/queryParams.go
+++ b/pkg/query-service/model/queryParams.go
@@ -346,20 +346,38 @@ type GetFlamegraphSpansForTraceParams struct {
 	SelectFields    []telemetrytypes.TelemetryFieldKey `json:"selectFields"`
 }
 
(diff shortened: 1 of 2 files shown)


therealpandey:
@krsoninikhil shouldn't this go in the new trace detail module? We're still implementing stuff for the older endpoint here?

krsoninikhil:
> @krsoninikhil shouldn't this go in the new trace detail module? We're still implementing stuff for the older endpoint here?

The flamegraph endpoint has not moved to module yet, so keeping all the improvements in existing API itself. We moved the waterfall API because the changes we required in that needed a version change.

pkg/query-service/model/queryParams.go
```diff
@@ -346,6 +346,45 @@ type GetFlamegraphSpansForTraceParams struct {
...
+				needsAttrBool = true
+			default:
+				// Unknown type: AttributeValue searches all three maps, so we need all.
+				needsAttrString = true
+				needsAttrNumber = true
+				needsAttrBool = true
+			}
+		case telemetrytypes.FieldContextResource:
+			needsResourceMap = true
+		}
```
tushar-signoz, line 369: Add a default if FieldContext is not specified then it adds ResourceMap as well

tushar-signoz (changes requested):
Can you check the api requests coming from the frontend once for this and check the selecteCols values. 
Also on the frontend side that nothing breaks

pkg/query-service/model/queryParams.go
krsoninikhil: Shouldn't we always require the FieldContext? If we add a default case, chances are client side can just skip the context and we'll end always querying everything. 

The UI change which uses these `SelectField` is not yet live, so any mismatch can be caught without breaking anything.

krsoninikhil:
> Can you check the api requests coming from the frontend once for this and check the selecteCols values. Also on the frontend side that nothing breaks

Have verified the request payload and the generated query, here is a sample request payload
image

pkg/query-service/model/queryParams.go
tushar-signoz: @krsoninikhil

nityanandagohain: Nope, we should expect the client to not know stuff and we should always correct or popluate metadata based on what is present in the backed, this seems like a fix that will work but overall the correct way will be to move this to modules and then use `ColumnExpressionFor` for actually fetching it from the db. 

But this fix can go ahead.

And yeah default should check all i.e attributes and resource.

krsoninikhil:
Closing this in favor of https://github.com/SigNoz/signoz/pull/11462

PR #11338, opened 2026-05-18
Fix drilldown on service details page

aks07:
## Summary

Two related fixes for the service detail page (`/services/:servicename`) drilldown experience, surfaced by a customer hitting empty result sets when drilling from charts into Logs / Traces.

Closes [engineering-pod#4658](https://github.com/SigNoz/engineering-pod/issues/4658).

### 1. Drilldown filter expression sanitisation

The drilldown filter inherited legacy SigNoz trace schema keys (`serviceName`, `name`) which don't match downstream explorers in the customer's build:

- `serviceName` works for traces (backend tolerates the alias) but fails for logs.
- `name` is a span-name concept — works for traces, has no logs equivalent.

Extended the existing `KeyValueMapping[]` mechanism in `container/QueryTable/Drilldown/metricsCorrelationUtils.tsx`:

- `KeyValueMapping.newAttribute` widened to `string | null`; `null` means **drop the pair**.
- Two new target-keyed mapping lists: `DRILLDOWN_TO_LOGS_MAPPINGS` and `DRILLDOWN_TO_TRACES_MAPPINGS`.
- New helper `applyMappingsToExpression` that delegates rewrites to `replaceKeysAndValuesInExpression` and drops to `removeKeysFromExpression` (already exists at `components/QueryBuilderV2/utils.ts:545`, ANTLR-parser-based, handles `AND`/`OR`/`NOT`/parens correctly).

In `drilldownUtils.tsx`, the existing source-side `METRIC_TO_LOGS_TRACES_MAPPINGS` block now also runs the target-keyed sanitisation depending on whether the drilldown is going to `view_logs` or `view_traces`.

**Net behaviour:**

| Source → Target | `serviceName` | `name` | `operation` (metric-APM) |
|---|---|---|---|
| any → traces | rewritten to `service.name` | kept | rewritten to `name` (existing) |
| any → logs   | rewritten to `service.name` | dropped | rewritten to `name` then dropped |
| any → metrics | unchanged | unchanged | unchanged |

### 2. Drilldown gating leak + enable on service detail page

The drilldown context menu was firing on every `TimeSeriesPanel` and `BarPanel` regardless of `enableDrillDown` because:

- `usePanelContextMenu` didn't accept the prop — it wired `useGraphContextMenu` unconditionally.
- `TimeSeriesPanel` / `BarPanel` accepted `PanelWrapperProps` (which has `enableDrillDown`) but ignored the prop.

Fixed by:

- `usePanelContextMenu` accepts `enableDrillDown?: boolean` (default `false`). When `false`, returns empty `menuItemsConfig` and a no-op click handler. Hooks (`useCoordinates`, `useGraphContextMenu`) still called unconditionally for Rules-of-Hooks compliance.
- `TimeSeriesPanel` and `BarPanel` destructure `enableDrillDown` from props and pass it to the hook.
- New constant `SERVICE_DETAIL_DRILLDOWN_ENABLED` in `container/MetricsApplication/constant.ts` — single source of truth for the service detail page. All 9 `<Graph>` instances across `MetricsApplication/Tabs/**` pass `enableDrillDown={SERVICE_DETAIL_DRILLDOWN_ENABLED}` so the page-wide setting is toggled in one place.

Other call sites already honour `enableDrillDown`: `QueryTable` (table cells), `GridValueComponent` (value panels), `GridCard/FullView`, `PiePanelWrapper`. These now correctly hide the drilldown menu when the prop is `false`.

`isDrilldownEnabled()` (the dashboard-side flag) left as-is.

SCREENSHOTS
Before

- View in logs/traces broken. 
- serviceName sent in query

https://github.com/user-attachments/assets/59efdc68-0f22-40d9-b043-f131c824a566

After

- Correct query passed in trace/logs
- serviceName change to service.name 
- name dropped in logs query
- breakout works as expected now.

https://github.com/user-attachments/assets/5002a068-23a4-4c7c-bcf0-daf7bb31cf8e

### Manual smoke (reviewer)
- [ ] `/services/<svc>` → click chart series → context menu appears
- [ ] Drill to **View Traces** — URL filter has `service.name='<svc>'` (not `serviceName`); `name='<op>'` retained
- [ ] Drill to **View Logs** — URL filter has `service.name='<svc>'`; no `name`, no `serviceName`
- [ ] Repeat with `use_span_metrics=true` — span-metrics → traces drilldown still rewrites `operation` → `name` correctly
- [ ] Dashboard drilldown unchanged (regression): existing `METRIC_TO_LOGS_TRACES_MAPPINGS` rewrites for `operation` / `span.kind` / `status.code` still produce correct trace filters
- [ ] Table-cell drilldown (`QueryTable`) — works when source widget has drilldown enabled, no menu otherwise


frontend/src/container/QueryTable/Drilldown/metricsCorrelationUtils.tsx
```diff
@@ -40,8 +43,33 @@ export const METRIC_TO_LOGS_TRACES_MAPPINGS: KeyValueMapping[] = [
...
 
+export const DRILLDOWN_TO_LOGS_MAPPINGS: KeyValueMapping[] = [
+	{
+		attribute: 'serviceName',
+		newAttribute: 'service.name',
+		valueMappings: {},
+	},
+	{
+		attribute: 'name',
+		newAttribute: null,
```
ahrefabhi, line 54: is this intentional? name -> null

aks07: yes. if newAttribute is null, we remove it. did it this way so that the current mapper format can be used to remove keys as well, instead of just replacing them. keeps the same convention.

ahrefabhi: okay

ahrefabhi (approved):

frontend/src/container/QueryTable/Drilldown/__tests__/drilldownUtils.test.tsx
```diff
@@ -169,9 +169,10 @@ describe('drilldownUtils', () => {
 
 			// Verify transformations were applied
 			if (filterExpression) {
-				// Rule 2: operation → name
-				expect(filterExpression).toContain(`name = 'GET'`);
+				// `operation` rewrites to `name` via source-side pass, then `name`
```
AshwinBhatkal, line 172: Will this affect things outside of service details page? Do we need a separate test for that?

AshwinBhatkal: same for the below changes

frontend/src/container/QueryTable/Drilldown/metricsCorrelationUtils.tsx
```diff
@@ -40,8 +43,33 @@ export const METRIC_TO_LOGS_TRACES_MAPPINGS: KeyValueMapping[] = [
 	},
 ];
 
+export const DRILLDOWN_TO_LOGS_MAPPINGS: KeyValueMapping[] = [
```
AshwinBhatkal, line 46: can you help me understand what this is for?

AshwinBhatkal (approved):

PR #11348, opened 2026-05-18
fix(alerts): ensure edit alert name is updated correctly and not override after save

### LINKED ISSUE #9845: [BUG] Alert Title Rename Shows Success Message but Changes Are Not Saved

H4ad:
## Pull Request

---

The Alert Rule name had two definitions, now it's synchronized in both places to use the same value, and both values are updated when we change the alert rule name.

Before:

https://github.com/user-attachments/assets/3f899698-92a1-4bc4-9174-30f2f4a1827d

After:

https://github.com/user-attachments/assets/b03e80dc-4921-4d78-9ba8-c8de7d9cfb0c

Fixes https://github.com/SigNoz/signoz/issues/9845

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

The issue was causing the header/modal/input to not sync during updates.

Each place defined their own state.

Make all the states sync using the context.

---

- Tests added/updated: No
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Edit Alert
- Potential regressions: -

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Fixed the issue that caused the renamed alert via actions menu to not be stored when you hit save changes after renaming. |

---

### 📋 Checklist
- [x] Manually tested


AshwinBhatkal (dismissed):
Overall, this looks okay

What I'm thinking though -
- We should remove that menu option in the edit mode. It should happen purely from view mode.
- If this is part of the refactor around alerts, then we should do the above
- Else, the current change is good

AshwinBhatkal (approved):

PR #11349, opened 2026-05-18
fix: add adjustkeys in trace operator cte builder

nityanandagohain:
## Pull Request

---

### 📄 Summary
Adds ajustkeys in trace operator cte builder as it didn't have it.

All the tests and integration tests fails if the changes are reverted.

#### Issues closed by this PR
Fixes https://github.com/SigNoz/engineering-pod/issues/4967

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature

- Tests added/updated:✅ 
- Manual verification:✅ 
- Edge cases covered:

---

- Blast radius: Trace operators
- Potential regressions: none
- Rollback plan:rever


pkg/telemetrytraces/statement_builder.go
```diff
@@ -193,24 +193,30 @@ func getKeySelectors(query qbtypes.QueryBuilderQuery[qbtypes.TraceAggregation])
...
+func mergeDeprecatedTraceKeys(keys map[string][]*telemetrytypes.TelemetryFieldKey) {
 	for fieldKeyName, fieldKey := range IntrinsicFieldsDeprecated {
 		keys[fieldKeyName] = append([]*telemetrytypes.TelemetryFieldKey{&fieldKey}, keys[fieldKeyName]...)
 	}
 	for fieldKeyName, fieldKey := range CalculatedFieldsDeprecated {
 		keys[fieldKeyName] = append([]*telemetrytypes.TelemetryFieldKey{&fieldKey}, keys[fieldKeyName]...)
 	}
+}
+
+func adjustTraceKeys(ctx context.Context, logger *slog.Logger, keys map[string][]*telemetrytypes.TelemetryFieldKey, query *qbtypes.QueryBuilderQuery[qbtypes.TraceAggregation], requestType qbtypes.RequestType) {
```
srikanthccv: remove the logger as param here and return the `actions` and log the actions in the caller instead (each caller already holds the logger). something like.

```
  for _, a := range adjustTraceKeys(keys, &query, requestType) {
      b.logger.InfoContext(ctx, "key adjustment action", slog.String("action", a))
  }
```

pkg/telemetrytraces/trace_operator_cte_builder.go
```diff
@@ -417,21 +421,28 @@ func (b *traceOperatorCTEBuilder) buildNotCTE(leftCTE, rightCTE string) (string,
...
+	keys, _, err := b.stmtBuilder.metadataStore.GetKeysMulti(ctx, keySelectors)
+	if err != nil {
+		return nil, err
+	}
+	b.adjustOperatorKeys(ctx, keys, requestType)
+
 	switch requestType {
 	case qbtypes.RequestTypeRaw:
-		return b.buildListQuery(ctx, selectFromCTE)
+		return b.buildListQuery(ctx, selectFromCTE, keys)
```
srikanthccv, line 436: I noticed variables is not supported here, however, the trace operator as a feature should support it regardless of the frontend uses it or not. Please create a ticket in engineering pod.

pkg/telemetrytraces/trace_operator_cte_builder.go
```diff
@@ -518,6 +513,45 @@ func (b *traceOperatorCTEBuilder) buildListQuery(ctx context.Context, selectFrom
...
+	b.operator.SelectFields = tmp.SelectFields
+	b.operator.GroupBy = tmp.GroupBy
+	b.operator.Order = tmp.Order
+
+	for _, action := range actions {
+		b.stmtBuilder.logger.InfoContext(ctx, "key adjustment action", slog.String("action", action))
+	}
+}
+
 func (b *traceOperatorCTEBuilder) getKeySelectors() []*telemetrytypes.FieldKeySelector {
```
srikanthccv, line 557: Should this `getKeySelectors` take care of fetching keys for the referenced queries as well (right now it's only doing it for `b.operator`)? Otherwise how is the queer preparation working?

pkg/telemetrytraces/trace_operator_cte_builder.go
```diff
@@ -518,6 +513,45 @@ func (b *traceOperatorCTEBuilder) buildListQuery(ctx context.Context, selectFrom
...
+	for idx := range tmp.GroupBy {
+		actions = append(actions, adjustTraceKey(&tmp.GroupBy[idx].TelemetryFieldKey, keys)...)
+	}
+	for idx := range tmp.Order {
+		actions = append(actions, adjustTraceKey(&tmp.Order[idx].Key.TelemetryFieldKey, keys)...)
+	}
+
+	// Copy back the three slices that the helpers above can rewrite
+	// (AdjustDuplicateKeys reconstructs them, adjustTraceKey mutates in place).
+	// Aggregations is only read by the helpers, never reassigned, so no copy-back.
```
srikanthccv: please elaborate, why isn't `Aggregations` updated?

pkg/telemetrytraces/trace_operator_cte_builder_test.go
```diff
@@ -626,3 +595,143 @@ func TestTraceOperatorStatementBuilderErrors(t *testing.T) {
...
+						Signal: telemetrytypes.SignalTraces,
+						Filter: &qbtypes.Filter{Expression: "service.name = 'frontend'"},
+					},
+				},
+			},
+		},
+	)
+
+	require.NoError(t, err)
+	require.Equal(t, 1, strings.Count(q.Query, "AS `http.method`"),
```
srikanthccv: is there a better way to do this? because the sub-string match is very fragile here. There could be still be bug with double appearances in the query and the test would pass if the shape of the prepared query doesn't have the expected sub-sting

srikanthccv: I see there is already case in integration tests, then there is no need for this unit test.

pkg/telemetrytraces/trace_operator_cte_builder.go
nityanandagohain: https://github.com/SigNoz/engineering-pod/issues/5165

pkg/telemetrytraces/trace_operator_cte_builder_test.go
nityanandagohain: yeah kind of duplicate test, but better to keep it I guess ? if someone is developing it will help identify issue faster rahter than waiting for integration test ?

srikanthccv: even then, there should be a better way to identify whether the group by duplicate is taken care, instead of relying on the end query sub-string match.

pkg/telemetrytraces/trace_operator_cte_builder.go
nityanandagohain: For the referenced queries it's is done in line 211  in this file.

pkg/telemetrytraces/trace_operator_cte_builder.go
nityanandagohain: Updating it to keep it simple.

pkg/telemetrytraces/trace_operator_cte_builder_test.go
nityanandagohain: Yeah agree on that, have updated it.

pkg/telemetrytraces/statement_builder.go
```diff
@@ -124,8 +124,10 @@ func (b *traceQueryStatementBuilder) Build(
 		-------------------------------- End of tech debt ----------------------------
 	*/
 
-	query = b.adjustKeys(ctx, keys, query, requestType)
-
+	for _, action := range adjustTraceKeys(ctx, keys, &query, requestType) {
+		// TODO: change to debug level once we are confident about the behavior
+		b.logger.InfoContext(ctx, "key adjustment action", slog.String("action", action))
```
srikanthccv, line 129: @tushar-signoz , please validate and update log level (in separate ticket).

srikanthccv (dismissed):

srikanthccv (approved):

tushar-signoz (approved):
LGTM

therealpandey (approved):

PR #11356, opened 2026-05-19
fix: added fix for widget warning/save modal states

### LINKED ISSUE #9186: Incorrect unsaved changes warning in edit mode
The warning is shown even when I have made no changes.

ahrefabhi:
## Pull Request

---

### 📄 Summary

The unsaved-changes warning on dashboard widgets was firing in two wrong ways and missing one correct way:

1. **False positive on edit-mode mount** — opening an existing widget showed "Unsaved Changes" before the user had touched anything. The dirty check did a deep-equal between `currentQuery` and `stagedQuery`, but the live query carries v4-legacy fields (`aggregateAttribute`, `aggregateOperator`, `timeAggregation`, `spaceAggregation`, `reduceTo`) that are stripped from the persisted widget spec — so the two never matched on first render.
2. **Silent data loss after Stage and Run** — the comparison baseline (`stagedQuery`) was reset to match `currentQuery` on every Stage and Run. After running once, the dirty check could never see a divergence, so discarding lost edits without warning.
3. **Redundant "Save Widget" confirmation modal** — pre-Save modal asked the user to confirm with the same copy the discard flow used; just removed.

The fix compares `currentQuery` against the **saved widget query** (stable across Stage-and-Run cycles) for existing widgets, falls back to `stagedQuery` only for brand-new panels, and normalizes both sides to strip the v4-vs-v5 noise and regenerated `id`s before equality.

Before
image

https://github.com/user-attachments/assets/ce8d5db3-3e3d-45d5-b64d-6f09b0446fab

After
image

#### Issues closed by this PR

Closes https://github.com/SigNoz/signoz/issues/9186
Closes https://github.com/SigNoz/engineering-pod/issues/5002
Closes https://github.com/SigNoz/engineering-pod/issues/4941
Closes https://github.com/SigNoz/engineering-pod/issues/4849

---

### ✅ Change Type

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause

`getIsQueryModified` was doing a deep-equal between `currentQuery` and `stagedQuery` via `omitIdFromQuery`. Two structural issues made that wrong:

- `stagedQuery` is overwritten with `currentQuery` on every Stage and Run, so it isn't a stable baseline for an existing widget — after the first run, any subsequent edits compared dirty=against=clean and looked clean.
- The persisted widget spec is stored in v5 shape (`aggregations[]` only); when loaded into the form, `initialQueryBuilderFormValuesMap` spreads the legacy v4 fields back on top, and `MetricNameSelector` backfills `aggregateAttribute` for edit mode. The freshly-loaded query and the saved spec therefore differ even when the user hasn't touched anything.

#### Fix Strategy

- In [container/NewWidget/index.tsx](frontend/src/container/NewWidget/index.tsx), derive `savedWidgetQuery` from the persisted widget (looked up by `widgetId`) and compare against that. For new panels with no saved baseline, fall back to `stagedQuery` so initial edits still register.
- In [container/NewWidget/utils.ts](frontend/src/container/NewWidget/utils.ts), rewrote `getIsQueryModified` to normalize both sides before deep-equality:
  - Drop the v4 legacy fields (`aggregateAttribute`, `aggregateOperator`, `timeAggregation`, `spaceAggregation`, `reduceTo`, `filters`) from each `queryData`.
  - Drop regenerated identifiers — top-level `id` (new UUID every Stage and Run), `groupBy[].id` (rebuilt each load), and coerce `unit` from `undefined`/`''` to a stable `''`.
  - Strip explicit-`undefined` keys, since `lodash.isEqual` distinguishes `{a: undefined}` from `{}` and initial-values spreads leave those around.
  - `adjustQueryForV5` was deliberately not reused — its migration step resurrects the legacy fields onto each `aggregations[i]`, which is correct on save but bleeds asymmetrically into a comparator.
- Replaced the two inline antd `Modal` instances in `index.tsx` with a single [DiscardChangesModal](frontend/src/container/NewWidget/WidgetModals/DiscardChangesModal.tsx) (built on `@signozhq/ui/dialog`'s `ConfirmDialog`) that branches its copy on `isNewPanel`.
- Removed the pre-Save confirmation modal — Save now calls `onClickSaveHandler` directly.

---

### 🧪 Testing Strategy

- Tests added/updated: 8 new unit tests in [container/NewWidget/__test__/utils.test.ts](frontend/src/container/NewWidget/__test__/utils.test.ts) covering `getIsQueryModified`:
  - returns `false` when baseline is `null`/`undefined`
  - returns `false` when current differs only by auto-backfilled `aggregateAttribute` (the false-positive case)
  - returns `true` for genuine edits: legend change, metric swap, groupBy added
  - returns `true` on the Stage-and-Run silent-loss flow (current diverges from saved while stagedQuery matches current)
  - new-panel paths: clean baseline returns `false`; legend edit returns `true`
- Manual verification:
  - [ ] Open existing widget → no warning shown on mount
  - [ ] Edit widget → click Stage and Run → click Discard → warning fires
  - [ ] Edit widget → click Save → no extra confirmation, change persists
  - [ ] New panel → no warning on mount; edit then back-navigate → warning fires
- Edge cases covered: v4 → v5 mixed-shape queries, group-by id regeneration, `unit` flipping between `''` and `undefined`.

---

- Blast radius: scoped to the widget edit page (`NewWidget` container) and the discard/save warning flow. `getIsQueryModified` is only consumed inside that container.
- Potential regressions: false negatives where a genuine edit fails to mark the widget dirty (e.g. an edit on a field we now strip from the comparator). Manually verified the common-case fields (legend, metric, groupBy, panel type) still register.
- Rollback plan: single commit, easy revert. No data/schema changes.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Fix incorrect "unsaved changes" warning when editing dashboard widgets — no longer fires on mount of an unchanged widget, and now correctly fires after Stage and Run when the user discards edits. |

---

### 📋 Checklist

- [ ] Manually tested

---

## 👀 Notes for Reviewers

- The `getIsQueryModified` rewrite intentionally avoids `adjustQueryForV5` even though it looks like the obvious tool — see the comment in [utils.ts](frontend/src/container/NewWidget/utils.ts) explaining the asymmetry.
- The `savedWidgetQuery` fallback (`?? stagedQuery`) is only there for new panels where no saved widget exists yet — once you've saved the widget once, the baseline becomes stable across Stage-and-Run cycles.
- `DiscardChangesModal` lives under a new `WidgetModals/` folder — keeping room for more modal extractions if the inline modal pattern in `index.tsx` keeps growing.

---

### DIFF
--- a/frontend/src/container/NewWidget/utils.ts
+++ b/frontend/src/container/NewWidget/utils.ts
@@ -1,5 +1,4 @@
 import { Layout } from 'react-grid-layout';
-import { omitIdFromQuery } from 'components/ExplorerCard/utils';
 import { PrecisionOptionsEnum } from 'components/Graph/types';
 import { YAxisCategoryNames } from 'components/YAxisUnitSelector/constants';
 import {
@@ -26,16 +25,84 @@ import { DataSource } from 'types/common/queryBuilder';
 
 import { getCategoryName } from './RightContainer/dataFormatCategories';
 
+// Asks "would saving the current panel change the persisted widget spec?".
+//
+// `adjustQueryForV5` is deliberately not reused here: in addition to stripping
+// the legacy v4 fields, it also resurrects them onto each metric
+// `aggregations[i]`. That migration step is correct on save but bleeds
+// asymmetrically across a comparator — the live query still carries the
+// legacy defaults from `initialQueryBuilderFormValuesMap` while a previously
+// saved widget had them stripped.
+const stripQueryDataForCompare = (
+	queryData: IBuilderQuery,
+): Record<string, unknown> => {
+	const {
+		aggregateAttribute: _aggregateAttribute,
+		aggregateOperator: _aggregateOperator,
+		timeAggregation: _timeAggregation,
+		spaceAggregation: _spaceAggregation,
+		reduceTo: _reduceTo,
+		filters: _filters,
+		...retained
+	} = queryData ?? ({} as IBuilderQuery);
+
+	const groupBy = (retained.groupBy ?? []).map((entry) => {
+		const { id: _id, ...rest } = entry;
+		return rest;
+	});
+
+	return {
+		...retained,
+		groupBy,
+		source: retained.source || '',
+	};
+};
+
+const normalizeForDirtyCheck = (query: Query): Record<string, unknown> => {
+	const { id: _id, unit, builder, ...rest } = query;
+	return {
+		...rest,
+		// `id` is regenerated on every Stage and Run; `unit` flips between ''
+		// and undefined depending on whether the user has touched the selector.
+		unit: unit || '',
+		builder: {
+			...builder,
+			queryData: (builder?.queryData ?? []).map(stripQueryDataForCompare),
+		},
+	};
+};
+
+// `lodash.isEqual` distinguishes `{a: undefined}` from `{}`; for the dirty
+// check those are the same. Initial-values spreads on the live query
+// frequently leave such explicit-undefined keys.
+const stripUndefined = (value: unknown): unknown => {
+	if (Array.isArray(value)) {
+		return value.map(stripUndefined);
+	}
+	if (value && typeof value === 'object') {
+		const out: Record<string, unknown> = {};
+		Object.entries(value as Record<string, unknown>).forEach(([k, v]) => {
+			if (v === undefined) {
+				return;
+			}
+			out[k] = stripUndefined(v);
+		});
+		return out;
+	}
+	return value;
+};
+
 export const getIsQueryModified = (
 	currentQuery: Query,
-	stagedQuery: Query | null,
+	baselineQuery: Query | null | undefined,
 ): boolean => {
-	if (!stagedQuery) {
+	if (!baselineQuery) {
 		return false;
 	}
-	const omitIdFromStageQuery = omitIdFromQuery(stagedQuery);
-	const omitIdFromCurrentQuery = omitIdFromQuery(currentQuery);
-	return !isEqual(omitIdFromStageQuery, omitIdFromCurrentQuery);
+	return !isEqual(
+		stripUndefined(normalizeForDirtyCheck(baselineQuery)),
+		stripUndefined(normalizeForDirtyCheck(currentQuery)),
+	);
 };
 
 export type PartialPanelTypes = {
--- a/frontend/src/container/NewWidget/WidgetModals/DiscardChangesModal.tsx
+++ b/frontend/src/container/NewWidget/WidgetModals/DiscardChangesModal.tsx
@@ -1,0 +1,53 @@
+import { SolidAlertTriangle } from '@signozhq/icons';
+import { ConfirmDialog } from '@signozhq/ui/dialog';
+import { Typography } from '@signozhq/ui/typography';
+
+export interface DiscardChangesModalProps {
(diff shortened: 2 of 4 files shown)


frontend/src/container/NewWidget/utils.ts
```diff
@@ -26,16 +25,84 @@ import { DataSource } from 'types/common/queryBuilder';
...
+		});
+		return out;
+	}
+	return value;
+};
+
 export const getIsQueryModified = (
 	currentQuery: Query,
-	stagedQuery: Query | null,
+	baselineQuery: Query | null | undefined,
```
AshwinBhatkal, line 97: any reason to change the name for this? This is the same name across the app today

frontend/src/container/NewWidget/utils.ts
```diff
@@ -26,16 +25,84 @@ import { DataSource } from 'types/common/queryBuilder';
...
 
+// Asks "would saving the current panel change the persisted widget spec?".
+//
+// `adjustQueryForV5` is deliberately not reused here: in addition to stripping
+// the legacy v4 fields, it also resurrects them onto each metric
+// `aggregations[i]`. That migration step is correct on save but bleeds
+// asymmetrically across a comparator — the live query still carries the
+// legacy defaults from `initialQueryBuilderFormValuesMap` while a previously
+// saved widget had them stripped.
+const stripQueryDataForCompare = (
```
AshwinBhatkal, line 36: how can we verify this? Can you add screen recording in the PR depicting this case?

AshwinBhatkal (approved):

frontend/src/container/NewWidget/utils.ts
ahrefabhi: Because we are not comparing stagedQuery and currentQuery now, we are comparing currentQuery and (panel Query || stagedQuery)

frontend/src/container/NewWidget/utils.ts
ahrefabhi: Added the video above.

The metric-panel false-positive came from the dirty check running `currentQuery` against `stagedQuery` with `omitIdFromQuery` as the only normalizer. Two things made an unedited widget look "modified":

1. **Aggregation rehydration asymmetry.** The live `currentQuery` still carries the v4 defaults from `initialQueryBuilderFormValuesMap` (`aggregateAttribute`, `timeAggregation`, `spaceAggregation`, `reduceTo`). The saved widget loaded from the server has those stripped and the real values live on `aggregations[i]`. Comparing the two raw → `isEqual` returns `false` on a panel the user never touched.
2. **Stage and Run reset.** `stagedQuery` is overwritten to match `currentQuery` after every Stage and Run, so it stops being a stable baseline for the saved widget.

Fix anchors the comparison to the actual saved widget (looked up by `widgetId` from `widgets[]`) and runs both sides through `stripQueryDataForCompare` + `stripUndefined` so the v4 legacy fields, regenerated `groupBy.id`s, and `undefined`-vs-missing keys can't trip `isEqual`. Recording in the PR shows the metric panel opening straight to the normal "Save Widget" confirmation now.

PR #11358, opened 2026-05-19
chore: added changes to auto-run query when disabled, changed groupby and changed orderBy

ahrefabhi:
## Pull Request

---

### 📄 Summary

Three controls on the query builder — toggling a query's visibility/disabled state, editing its **group by** keys, and editing its **order by** keys — used to require the user to click **Stage and Run** afterwards before the panel reflected the change. That extra click felt redundant: the change is intent-explicit (you didn't just type a partial filter, you committed to a new groupBy / hid a query), so the panel should just update.

This PR adds a `runAfterUpdate` option on `handleChangeQueryData` that callers can opt into per call. Those three controls now opt in; everything else (filters, legend, aggregations, etc.) behaves exactly as before — they continue to wait for an explicit Stage and Run, because there the user is mid-edit and auto-running on every keystroke would be wrong.

Before

https://github.com/user-attachments/assets/c70616b2-8e46-49f7-b498-0a158c51cccd

After

https://github.com/user-attachments/assets/86e76ccd-c60d-4682-9f3a-04a7caa12025

#### Issues closed by this PR

Closes https://github.com/SigNoz/pulse-pod/issues/24

---

### ✅ Change Type

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

(Behaviorally a small UX improvement; the underlying change is a type/contract refactor on `handleChangeQueryData`.)

---

### 🐛 Bug Context

_N/A — not a bug fix. UX consistency improvement on three controls whose previous behavior required an extra click._

---

### 🧪 Testing Strategy

- Tests added/updated: extended [useQueryBuilderOperations.test.ts](frontend/src/hooks/queryBuilder/__tests__/useQueryBuilderOperations.test.ts) (+195 lines) with a new `describe` block covering:
  - omitting the option → `handleRunQuery` is **not** called (existing behavior preserved)
  - `runAfterUpdate: true` → calls `handleSetQueryData` and `handleRunQuery` with the freshly-changed query (the override carries the new value at the right index, siblings untouched)
  - non-zero index path: the change lands at `index: 1` without disturbing the query at `index: 0`
  - legend formatting: the value passed to `handleRunQuery` matches what's passed to `handleSetQueryData` (the formatter is applied once, consistently)
  - trace-operator queries (`isForTraceOperator: true`) early-return — `handleRunQuery` is not called even with the option set
- Manual verification:
  - [ ] Toggle visibility on a query → panel refreshes
  - [ ] Change groupBy keys on a metrics query → panel refreshes
  - [ ] Change orderBy on a logs/traces query → panel refreshes
  - [ ] Type into a filter or legend → panel does **not** refresh until Stage and Run (unchanged)
- Edge cases covered: index-targeting (siblings preserved), `isForTraceOperator` short-circuit, formatter parity between set + run paths.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | Query builder now auto-refreshes the panel when you toggle a query's visibility or edit its **group by** / **order by** — no need to click **Stage and Run** for those three actions. Filter/legend/aggregation edits still wait for an explicit run. |

---

### 📋 Checklist

- [ ] Manually tested

---

## 👀 Notes for Reviewers

- The override-query argument on `handleRunQuery` is load-bearing — the React `setState` from `handleSetQueryData` hasn't flushed by the time we call `handleRunQuery`, so we pass the freshly-computed query through explicitly rather than reading stale `currentQuery`. See the comment in [useQueryBuilderOperations.ts](frontend/src/hooks/queryBuilder/useQueryBuilderOperations.ts).
- The opt-in API (`{ runAfterUpdate: true }`) is intentionally narrow — a single boolean option. If the list of "auto-run on" controls grows materially, we can promote it to a named policy enum, but right now four callsites doesn't justify that.
- `handleToggleDisableQuery` in `QueryV2.tsx` also has its dependency array tightened to `[handleChangeQueryData, query.disabled]` (was `query`) — this was an unrelated dep-array correctness fix surfaced while touching the file.

---


AshwinBhatkal:
Please check for the other query types as well. Visibility should be handled there as well. Typing into the input is something we can avoid for this PR. 

However, for input fields we can still have an onBlur handle that executes the query. Also, there is a subtle difference between a few input fields that take in inputs while also providing an option that can replace the entire field. An example would be Group By when writing a Logs query. We can evaluate it before making the change here.

Also, now that there is no separate need to stage, we can rename Stage & Run Query to just Run Query.

ahrefabhi:
> Please check for the other query types as well. Visibility should be handled there as well. Typing into the input is something we can avoid for this PR.
> 
> However, for input fields we can still have an onBlur handle that executes the query. Also, there is a subtle difference between a few input fields that take in inputs while also providing an option that can replace the entire field. An example would be Group By when writing a Logs query. We can evaluate it before making the change here.
> 
> Also, now that there is no separate need to stage, we can rename Stage & Run Query to just Run Query.

Handled the case of Clickhouse and PromQL query. I'm not running the query automatically in case of input fields, only running it in case of toggles + tag based inputs.

AshwinBhatkal:
This change I feel makes sense for visibility. For the Selects, I'm wondering if it can get confusing when to run and when to just select. Also, the current implementation just changes based on selection in the dropdown which can cause wasteful queries. For multi select at least, a blur would help. For single select (Reduce to, for example), it is equivalent. We should do it on blur instead of click. If differentiating between them is difficult, blur can be the option for both. However though, this still would raise the confusion - "When to run and when to select or input". Do you have any ideas around this? 

And say we went with blur, it could apply to input fields as well. At that time, Run Query would just be a button to refresh

Also, what about this 👇🏼 
> Also, now that there is no separate need to stage, we can rename Stage & Run Query to just Run Query.

PR #11361, opened 2026-05-19
feat(ai-assistant): collapse thinking + tool-call steps into one row

makeavish:
## Pull Request

---

### 📄 Summary

Long sequences of `Thinking` / `Searched tools…` / `Loaded skill…` rows in the AI Assistant chat pushed the actual answer below the fold and made every response feel verbose, even when the model only did 2–3 steps. This PR routes **every** activity row (single-item or multi-item) through a single `ActivityGroup` component so all "what the agent did" rows share one visual contract: `✦ <summary>  ›` aligned at the same x, click to expand inline.

The summary adapts to the items:
- 1 thinking item → `✦ Thinking…` (live) / `✦ Thought for a few seconds` (settled)
- 1 tool call → `✦ <tool's display text>` (e.g. `✦ Get Dashboard`)
- N items → `✦ Working… · Xs · N steps` (live) / `✦ Worked through N steps` (settled)

Single-item groups expand directly to the underlying content (no second chevron disclosure); multi-item groups expand to a list of per-step `ThinkingStep` / `ToolCallStep` rows as before. The "Thought for a few seconds" wording is intentionally vague because the API doesn't persist precise timing — matches what we show for reloaded history threads.

Before
Screenshot 2026-05-19 at 22 46 07

After
Screenshot 2026-05-20 at 13 18 58

| Before (9 stacked rows) | After (collapsed) | After (expanded) |
|---|---|---|
| \`> Thinking\` × 5 interleaved with \`> Searched tools…\`, \`> Loaded skill…\`, \`> Searched docs…\` | \`✦ Worked through 9 steps\` on one row | Underlying step rows revealed inline; tool-call \`INPUT\` / \`OUTPUT\` JSON uses the full bubble width |

#### Issues closed by this PR

N/A

---

### ✅ Change Type

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🧪 Testing Strategy

- Tests added/updated: **N/A** — the AI Assistant module has no existing unit/spec coverage; this PR matches that convention. The component is pure render logic against typed input.
- Manual verification:
  - Triggered a multi-step assistant response (`how to instrument?` → 9 thinking + tool steps): verified the live \`Working… · Xs · N steps\` ticker advances once per second, freezes to \`Worked through N steps\` once text streams.
  - Expanded the collapsed row: each underlying \`ThinkingStep\` / \`ToolCallStep\` renders flush with the parent label; tool-call \`INPUT\` / \`OUTPUT\` JSON spans the full bubble width.
  - Reloaded a history thread: thinking rows read \`Thought for a few seconds\` (same wording as fresh streams), confirming no dependency on server-stored timing.
  - Single-step message: renders as \`✦ Thought for a few seconds\` (or the tool's display label) at the same x as multi-item groups — expanding shows the body content directly with no inner chevron.
  - Approval-pause path: while paused, the trailing group label freezes; on resume, the elapsed-time counter restarts from zero rather than rolling the user's wait time into the duration.
- Edge cases covered: empty events array, all-text message (no activity groups), interleaved \`[thinking, text, tool, text]\` (each becomes its own group, no trailing live group), compact panel variant.
- Lint: \`oxlint\` introduces zero new errors/warnings on touched files.
- Typecheck: \`yarn tsc --noEmit\` clean.

---

- **Blast radius**: AI Assistant chat UI only. No store, API, or backend changes — purely a render-layer reshuffle of the existing \`MessageBlock\` / \`StreamingEventItem\` data.
- **Potential regressions**: text/tool/thinking interleaving and approval/clarification pause paths are the highest-risk flows, since they exercise the partition logic and the \`isLive\` propagation. Both are verified manually above. Persisted history threads use the same \`groupBlocks\` path as fresh streams, so a regression would surface immediately on any reopened conversation.
- **Rollback plan**: revert the feature commits on this branch. The PR has no migrations, schema changes, or persisted state — reverting restores the previous flat-list rendering with no data cleanup required.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | AI Assistant chat collapses thinking + tool-call steps behind a single expandable summary row so the assistant's answer stays in view. |

---

### 📋 Checklist

- [x] Tests added or explicitly not required (no existing AI Assistant test coverage to extend)
- [x] Manually tested
- [x] Backward compatibility considered (renders against the existing block schema; opportunistic forward-compat read for any future timing fields)

---

## 👀 Notes for Reviewers

- Every activity row now goes through \`ActivityGroup\` — including single-item ones. The component owns the summary text (\`buildSummary\`) and decides whether the expanded body is the underlying content directly (single-item) or a list of per-step rows (multi-item). This collapses two previously-divergent visual treatments into one.
- \`ThinkingStep\` and \`ToolCallStep\` now expose \`ThinkingContent\` / \`ToolCallContent\` body sub-components and \`thinkingLabel\` / \`getToolDisplayLabel\` helpers, so \`ActivityGroup\` can render single-item bodies without duplicating markup or relying on a "force-expanded headerless" mode on the step components.
- The two partition helpers (\`groupBlocks\` in \`MessageBubble.tsx\`, \`groupStreamingEvents\` in \`StreamingMessage.tsx\`) intentionally duplicate the same contract — \`MessageBlock\` and \`StreamingEventItem\` are separate types and we render into a shared \`ActivityItem\`. If a third call-site lands, factor out into \`types.ts\`; not worth the abstraction yet.
- \`ActivityGroup\`'s elapsed timer relies on each group's React instance surviving across re-renders. The invariant is documented in \`groupStreamingEvents\` (events only ever append; existing groups are never re-keyed). If a future change shrinks or re-orders the streaming events array, the \`wasLive → isLive\` re-stamp logic needs revisiting.
- \`aria-expanded\` + \`aria-controls\` are wired up on the disclosure button; the existing \`ThinkingStep\` is still a \`<div onClick>\` (pre-existing — out of scope).

### DIFF
--- a/frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.module.scss
+++ b/frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.module.scss
@@ -1,0 +1,64 @@
+// Collapsed activity summary — one row that hides the underlying
+// thinking + tool-call steps. Reuses the same quiet treatment as
+// ThinkingStep / ToolCallStep so it sits flush in the assistant bubble.
+.group {
+	width: 100%;
+	font-size: 12px;
+}
+
+.header {
+	display: flex;
+	align-items: center;
+	gap: 8px;
+	padding: 4px 0;
+	background: transparent;
+	border: none;
+	cursor: pointer;
+	color: var(--l3-foreground);
+	user-select: none;
+	transition: color 0.12s ease;
+	width: 100%;
+	text-align: left;
+
+	&:hover {
+		color: var(--l1-foreground);
+	}
+}
+
+.icon {
+	flex-shrink: 0;
+	color: var(--accent-primary);
+
+	&.spin {
+		animation: pulse 1.4s ease-in-out infinite;
+	}
+}
+
+@keyframes pulse {
+	0%,
+	100% {
+		opacity: 0.55;
+	}
+	50% {
+		opacity: 1;
+	}
+}
+
+.label {
+	flex: 1;
+	overflow: hidden;
+	text-overflow: ellipsis;
+	white-space: nowrap;
+	font-weight: 400;
+}
+
+.chevron {
+	flex-shrink: 0;
+	color: inherit;
+}
+
+.body {
+	display: flex;
+	flex-direction: column;
+	padding: 2px 0 4px;
+}
--- a/frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.module.scss
+++ b/frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.module.scss
@@ -1,12 +1,12 @@
 // Collapsed activity summary — one row that hides the underlying
 // thinking + tool-call steps. Reuses the same quiet treatment as
 // ThinkingStep / ToolCallStep so it sits flush in the assistant bubble.
-.group {
+.activityGroup {
 	width: 100%;
 	font-size: 12px;
 }
 
-.header {
+.activityHeader {
 	display: flex;
 	align-items: center;
 	gap: 8px;
@@ -25,11 +25,11 @@
 	}
 }
 
-.icon {
+.sparkleIcon {
 	flex-shrink: 0;
 	color: var(--accent-primary);
 
-	&.spin {
+	&.iconPulsing {
 		animation: activityGroupPulse 1.4s ease-in-out infinite;
 	}
 }
@@ -44,7 +44,7 @@
 	}
 }
 
-.label {
+.activitySummary {
 	flex: 1;
 	overflow: hidden;
 	text-overflow: ellipsis;
@@ -52,12 +52,12 @@
 	font-weight: 400;
 }
 
-.chevron {
+.toggleChevron {
 	flex-shrink: 0;
 	color: inherit;
 }
 
-.body {
+.activityBody {
 	display: flex;
 	flex-direction: column;
 	padding: 2px 0 4px;
--- a/frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.module.scss
+++ b/frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.module.scss
@@ -30,11 +30,11 @@
 	color: var(--accent-primary);
 
 	&.spin {
-		animation: pulse 1.4s ease-in-out infinite;
+		animation: activityGroupPulse 1.4s ease-in-out infinite;
 	}
 }
 
-@keyframes pulse {
+@keyframes activityGroupPulse {
 	0%,
 	100% {
 		opacity: 0.55;
--- a/frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.tsx
+++ b/frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from 'react';
+import { useEffect, useId, useRef, useState } from 'react';
 import cx from 'classnames';
 import { ChevronDown, ChevronRight, Sparkles } from '@signozhq/icons';
 
@@ -23,9 +23,6 @@ interface ActivityGroupProps {
 }
 
 function formatElapsed(ms: number): string {
-	if (ms < 1000) {
-		return '<1s';
-	}
 	const seconds = Math.round(ms / 1000);
 	if (seconds < 60) {
 		return `${seconds}s`;
@@ -44,6 +41,7 @@ export default function ActivityGroup({
 	isLive = false,
 }: ActivityGroupProps): JSX.Element {
 	const [expanded, setExpanded] = useState(false);
+	const bodyId = useId();
 
 	// Captures the moment this live phase started. Re-stamped on every
 	// false→true transition so a stream that pauses on
@@ -64,9 +62,11 @@ export default function ActivityGroup({
 		if (!isLive) {
 			return undefined;
 		}
+		// Tick once per second — the display is integer-second precision, so
+		// faster ticks would just re-render the bubble for no visible change.
 		const id = window.setInterval(() => {
 			setElapsed(Date.now() - startedAtRef.current);
-		}, 500);
+		}, 1000);
 		return (): void => window.clearInterval(id);
 	}, [isLive]);
 
@@ -75,8 +75,10 @@ export default function ActivityGroup({
 
 	let summary: string;
 	if (isLive) {
+		// Suppress the elapsed token until ≥ 1s — the first tick fires after
+		// 1s anyway, and showing "0s" or "<1s" briefly adds noise.
 		summary =
-			elapsed > 0
+			elapsed >= 1000
 				? `Working… · ${formatElapsed(elapsed)} · ${stepLabel}`
 				: `Working… · ${stepLabel}`;
 	} else {
@@ -87,7 +89,13 @@ export default function ActivityGroup({
 
 	return (
 		<div className={styles.group}>
-			<button type="button" className={styles.header} onClick={toggle}>
+			<button
+				type="button"
+				className={styles.header}
+				onClick={toggle}
+				aria-expanded={expanded}
+				aria-controls={bodyId}
+			>
 				<Sparkles
 					size={12}
 					className={cx(styles.icon, { [styles.spin]: isLive })}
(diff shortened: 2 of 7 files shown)


frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.module.scss
```diff
@@ -0,0 +1,64 @@
...
+	white-space: nowrap;
+	font-weight: 400;
+}
+
+.chevron {
+	flex-shrink: 0;
+	color: inherit;
+}
+
+.body {
```
YounixM: Though we are using modules we should avoid common class names.

frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+import styles from './ActivityGroup.module.scss';
+
+export type ActivityItem =
+	| { kind: 'thinking'; content: string }
+	| { kind: 'tool'; toolCall: StreamingToolCall };
+
+interface ActivityGroupProps {
+	items: ActivityItem[];
+	/**
```
YounixM, line 22: Avoid obvious comments

frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.tsx
```diff
@@ -0,0 +1,134 @@
...
+	 * True while this group is the trailing activity in an ongoing stream —
+	 * the underlying steps may still be growing or a tool may still be
+	 * running. Drives the live "Working…" label and elapsed-time tick.
+	 */
+	isLive?: boolean;
+}
+
+function formatElapsed(ms: number): string {
```
YounixM: use existing utils

frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.tsx
```diff
@@ -0,0 +1,134 @@
...
+				: `Working… · ${stepLabel}`;
+	} else {
+		summary = `Worked through ${stepLabel}`;
+	}
+
+	const toggle = (): void => setExpanded((v) => !v);
+
+	return (
+		<div className={styles.group}>
+			<button
```
YounixM: use component

frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.tsx
```diff
@@ -0,0 +1,134 @@
...
+				<span className={styles.label}>{summary}</span>
+				{expanded ? (
+					<ChevronDown size={12} className={styles.chevron} />
+				) : (
+					<ChevronRight size={12} className={styles.chevron} />
+				)}
+			</button>
+
+			{expanded && (
+				<div id={bodyId} className={styles.body}>
```
YounixM: Why are we using id here ?

frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.tsx
```diff
@@ -0,0 +1,134 @@
...
+					{/* eslint-disable react/no-array-index-key */}
+					{items.map((item, i) => {
+						// A thinking step is live only while it's the trailing item in
+						// a trailing live group — once any later event (text or tool)
+						// arrives, the pass is done.
+						const isLastItem = i === items.length - 1;
+						return item.kind === 'thinking' ? (
+							<ThinkingStep
+								key={i}
```
YounixM: do not use index as key

frontend/src/container/AIAssistant/components/ThinkingStep/ThinkingStep.tsx
```diff
@@ -5,16 +5,28 @@ import styles from './ThinkingStep.module.scss';
 
 interface ThinkingStepProps {
 	content: string;
+	/**
```
YounixM, line 8: remove  obvious comments

YounixM (commented):
@makeavish - Please check the comments.

frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.tsx
makeavish: Tightened in 82cabd2e8 — kept the bits about the timer re-stamp on approval/clarification resume since that's the non-obvious invariant a future reader would otherwise have to derive from `useEffect` + `wasLiveRef`.

frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.tsx
makeavish: Done in 82cabd2e8 — switched to `formatTime` from `utils/timeUtils`. The output format changes slightly (`45.0s` instead of `45s` for sub-minute durations), but the consistency win is worth it.

frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.tsx
makeavish: Pushing back on this. I tried swapping to `<Button variant="ghost">` from `@signozhq/ui/button` and it visually regressed — the Ghost button brings its own focus/active outline, base padding, and hover background tuned for action buttons, which fights this quiet full-width disclosure row (it ended up rendering as a filled blue button with a ring). Overriding all of those via `className` is fragile. The raw `<button type="button">` has `aria-expanded` + `aria-controls` wired up and the styling is fully under `.module.scss`, which is the cleaner contract for a disclosure widget. Open to revisiting if/when the design system grows a disclosure-specific primitive.

frontend/src/container/AIAssistant/components/ThinkingStep/ThinkingStep.tsx
makeavish: Tightened in 82cabd2e8 — kept only the rationale that's not derivable from the prop name (the API doesn't persist precise timing, so the copy is vague to stay consistent across fresh vs reloaded threads).

frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.module.scss
makeavish: Done in bcdfba5ab — renamed every common class in `ActivityGroup.module.scss` (`.group → .activityGroup`, `.header → .activityHeader`, `.icon → .sparkleIcon`, `.label → .activitySummary`, `.spin → .iconPulsing`, `.chevron → .toggleChevron`, `.body → .activityBody`). The component-side className refs were updated to match.

frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.tsx
makeavish: Switched to `<div onClick>` in bcdfba5ab — matches the existing `ThinkingStep` / `ToolCallStep` disclosure pattern in this folder. (Keeping the SCSS-driven styling since the `@signozhq/ui/button` Ghost variant brought its own focus ring / hover background that visually regressed the quiet row, as covered earlier in this thread.)

frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.tsx
makeavish: Dropped in bcdfba5ab — removed the `useId` import and the `aria-controls`/`id` wiring. With the disclosure body rendered inline beneath the header, `aria-expanded` already conveys the open/closed state to screen readers, so the id wasn't earning its keep.

frontend/src/container/AIAssistant/components/ActivityGroup/ActivityGroup.tsx
makeavish: Done in bcdfba5ab — `ActivityItem` and `RenderGroup` now carry an explicit `id` field assigned by the partition helpers. Persisted tool blocks key off the server-assigned `toolCallId`; streaming items and thinking blocks key off their position in the append-only source array (events / blocks are never reordered or removed mid-array, so the derived id is stable for the lifetime of the item). List keys now read `key={item.id}` instead of `key={i}`.

YounixM (approved):

PR #11362, opened 2026-05-19
feat(ai-assistant): polish composer UX and add Cmd+K entry

makeavish:
## Pull Request

---

### 📄 Summary

AI Assistant polish bundle. Original work plus a UX/a11y/correctness sweep validated by three review passes (live UX audit, multi-agent review, Codex `xhigh 5.5`), and PR-review follow-ups on top.

**Composer (ChatInput)**

- Textarea auto-grows up to ~200px before falling back to internal scroll; default rows 2 → 3.
- `:focus-within` ring scoped to the inner `.composer` row so the action footer (Add Context / mic / send) doesn't light up.
- Cmd+K palette gets an "Open AI Assistant" entry (gated by `useIsAIAssistantEnabled()`); selecting it starts a new conversation and opens the modal — same surface as Cmd+J — and the row shows the `cmd+j` shortcut hint on the right like every other palette entry.
- ChatInput is now keyed by `conversationId` in `ConversationView` (mirroring `VirtualizedMessages`). Clicking "+ New conversation" — or switching threads — remounts the composer, refocuses the textarea via its mount effect, and clears any leftover text / attachments / selected contexts from the prior conversation.

**Streaming auto-scroll**

- Manual `scrollTo` + Virtuoso `followOutput` now both gate on a live `atBottom` flag (`atBottomStateChange`). User can scroll up mid-stream without being yanked back.
- Override: a fresh user-send re-anchors regardless of scroll position (tracked by latest-message id), so sending while reading history still shows your own message.
- `VirtualizedMessages` keyed by `conversationId` so the scroll-anchor refs and Virtuoso's internal state reset together on conversation switch.

**Browser titles**

- `getRouteKey` now matches dynamic `:param` templates after the exact-match pass — fixes the stale title on `/ai-assistant/:conversationId` and six other routes that previously fell through to `DEFAULT`. Titles added in both `en` and `en-GB` locales. Inline comment notes that `ROUTES` declaration order is load-bearing for overlapping templates.

**Accessibility**

- Tablist context picker (`ChatInput`): `<button role="tab">` inside `role="tablist"`; entities as `<button aria-pressed>` inside a single stable `role="tabpanel"` (id `ai-context-tabpanel`) whose `aria-labelledby` swaps to the active tab. Full keyboard nav: ArrowUp/Down, Home/End, ArrowRight (tabs → entities), ArrowLeft (entities → active tab), roving tabindex with programmatic focus.
- Voice recording controls converted from `<div onClick>` to native `<button>` (WCAG 2.1.1); `.micRecording` container is now `role="status" aria-live="polite"`.
- Message action buttons (Copy / Thumbs / Regenerate) have `aria-label` + `aria-pressed`.
- ClarificationForm: `aria-required` on inputs, visual `*` is `aria-hidden`, multi_select group is now `<fieldset>` + `<legend>` (WCAG 1.3.1).
- Icon-only button sweep: `TooltipSimple` added to the `ApprovalCard` diff toolbar (split / unified / wrap / expand / copy), `RichCodeBlock` copy-code button, and `ChatInput` voice discard / stop / send so hover labels are visible to sighted users instead of relying on the native `title` attribute alone.

**Clarification form correctness**

- Submit disabled until every required field is filled (`isFieldFilled` per type).
- Server-supplied numeric defaults are parsed in `initialAnswerFor` (the DTO types `default` as `string | string[] | null`, so `"5"` was previously stored as a string and Submit stayed disabled).
- Cleared numeric `<Input>` emits `null` instead of `Number('') === 0`.

**PR review follow-ups**

- DS Button `prefix` slot used everywhere for icon-only / icon-prefixed buttons (composer, approval-card, message-feedback, modal/panel headers); icon-as-children pattern removed.
- Context-picker category + entity rows render as DS Button. DS Button defaults to `justify-content: center`; overridden via the exposed `--button-justify-content` CSS variable on the wrapper class so the rows stay left-aligned without specificity wars or `!important`.
- Context-picker inline keyboard handlers extracted into `handleCategoryKeyDown` / `handleEntityKeyDown` callbacks.
- `MessageFeedback` vote vocabulary (`'positive'` / `'negative'` / `'up'` / `'down'`, aria-labels, tooltip text) moved to a typed `as const` map keyed on `FeedbackRatingDTO` per the repo's "no magic strings" convention.
- Approval-diff view-mode toggle stays on the manual `ToggleGroup` + `ToggleGroupItem` composition (vs. `ToggleGroupSimple`) — the items-API renders the icon inside its own trigger, which Radix Tooltip can't anchor on. Manual composition wraps each `ToggleGroupItem` button directly with `TooltipSimple`, preserving the hover label.
- Approval-diff `DialogContent` adds `onOpenAutoFocus={(e) => e.preventDefault()}` — without it, Radix auto-focuses the first focusable child (the "Copy before" Button), and its `TooltipSimple` opens immediately on dialog mount because Radix Tooltip opens on focus too.
- Modal header icons (history / new / expand / minimize / close) had no explicit `color`, so they defaulted to `primary` and rendered blue. Now `color="secondary"`, matching the side-panel header.
- ChatInput re-grabs textarea focus on new-conversation / conversation-switch (see Composer section).

**Misc**

- Analytics `source` values extracted into typed `AIAssistantOpenSource` / `VoiceInputSource` constants in `events.ts` per `CLAUDE.local.md`. Wire values unchanged.

_See thread for live walkthroughs of:_ composer auto-grow, Cmd+K entry, streaming scroll behavior, context picker keyboard nav, voice control keyboard reachability, approval-diff toolbar tooltips, modal header color fix, new-chat refocus.

#### Issues closed by this PR

N/A — internal audit + review passes.

---

### ✅ Change Type

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause

- **Auto-scroll fight**: streaming effect re-anchored on every delta with no awareness of user position; `followOutput` returned `'auto'` even when `atBottom` was false.
- **Stale title on AI Assistant**: `getRouteKey` did exact string comparison, so every `:param` template fell through to `DEFAULT`.
- **Non-operable picker**: clickable `<div>`s; `role="tab"` without `role="tablist"`; `tabIndex={0}` on every tab broke roving; arrow handlers updated state but didn't move DOM focus, so subsequent keydowns fired from the stale tab's closure (couldn't reach the 3rd category). Inactive tabs also pointed at nonexistent panel ids.
- **Voice controls**: `<div onClick>` predates this PR — no keyboard access, no SR announcement.
- **ClarificationForm**: `handleSubmit` had no validation; required `*` was decorative; numeric defaults arrived as strings while `isFieldFilled` required `typeof === 'number'`.
- **Hidden hover labels**: several icon-only buttons in `ApprovalCard`, `RichCodeBlock`, and `ChatInput` had only `aria-label` and the native `title` attribute — no consistent visible tooltip for sighted users.
- **DS Button rows center-aligned**: the migration from raw `<button>` to DS `Button` for the context-picker rows lost the left-aligned layout because DS Button defaults to `justify-content: center`. Fixed by overriding the exposed `--button-justify-content` CSS variable rather than chasing class specificity.
- **`ToggleGroupSimple` + per-item tooltips**: the items-API wraps your `label` ReactNode inside its own trigger, so `TooltipSimple` around the icon can't anchor on the button. Reverted to manual `ToggleGroup` composition; the items-API isn't usable when per-item tooltips matter.
- **Stray tooltip on approval-diff dialog open**: Radix Dialog auto-focuses the first focusable child; that's a `TooltipSimple`-wrapped Copy button, and Radix Tooltip opens on focus.
- **Modal header icons blue**: missing `color="secondary"` on the modal-only header `<Button>`s; the side-panel header was already correct.
- **New-conversation didn't refocus the composer**: ChatInput stayed mounted across conversation switches (only `VirtualizedMessages` was keyed by `conversationId`), so its mount-effect for `textareaRef.current?.focus()` never re-ran when "+ New conversation" updated the active conversation.

#### Fix Strategy

Each fix is local to the symptom — no shared abstractions added.

---

### 🧪 Testing Strategy

Manual verification (live, `localhost:3301`):

- Composer auto-grow + scoped focus ring.
- Cmd+K → "Open AI Assistant" entry visible only when feature flag is on; `cmd+j` hint appears right-aligned; Enter opens the modal (not the drawer) and starts a new conversation; matches the Cmd+J behavior.
- Stream a long response → scroll up mid-stream (position holds) → scroll to bottom (follow resumes) → send a new message while scrolled up (re-anchors).
- Switch conversations → scroll position resets cleanly; composer text / attachments / selected contexts also clear; textarea regains focus.
- Click "+ New conversation" → fresh empty conversation; `document.activeElement === <textarea>`.
- `document.title === 'SigNoz | AI Assistant'` on `/ai-assistant/<conversationId>`.
- Context picker keyboard run-through: ArrowDown across all 3 tabs (wraps), Home/End, ArrowRight into entity list, ArrowDown/Up among entities, ArrowLeft back to active tab, Enter/Space to toggle, Escape to dismiss.
- Context-picker category + entity rows are left-aligned (DS Button + `--button-justify-content` override).
- DOM check: every tab's `aria-controls` resolves to `#ai-context-tabpanel`.
- Voice recording: Tab into recording state, discard / stop reachable via Enter; VoiceOver announces the `role="status"` container.
- ClarificationForm: required fields gate Submit; numeric defaults pre-fill and Submit is enabled.
- Hover over icon-only buttons (diff toolbar, expand-diff, code copy, voice discard / stop, Send) → tooltips appear via `TooltipSimple`.
- Approval-diff dialog: open → no stray "Copy before" tooltip; hover each toolbar toggle (Split view / Unified view / Wrap long lines) → tooltip appears; clicking each toggle switches state correctly.
- Modal header icons render in secondary (neutral gray), not primary blue.

Review passes — multi-agent review (commits `acab71bc8`, `e922598e6`, `9e845d283`, `99833e64e`), Codex `gpt-5.5` `xhigh` (commit `cc23d49c0`), a tooltip sweep (commit `a9cb5d4c7`), and PR-review follow-ups (commits `65db26749`, `ee2969cf5`, `640456bb6`) — all findings addressed.

`tsc --noEmit` clean.

---

- **Blast radius**: AI Assistant feature, plus `getRouteKey` + `titles.json` (used by every page through `AppLayout`), plus the Cmd+K palette (`shortcutActions.tsx` — only the AI Assistant entry is new).
- **Worth a smoke check**: `getRouteKey` is a strict superset (exact-match path unchanged), but the six newly-resolving dynamic routes (`/trace-old/:id`, `/services/:servicename/top-level-operations`, `/settings/roles/:roleId`, `/traces/funnels/:funnelId`, `/integrations/:integrationId`, `/public/dashboard/:dashboardId`) now show their proper title instead of `DEFAULT`.
- **Payload shape**: numeric clarification answers now serialize as `number` instead of string. Backend should accept either; flagging in case a downstream consumer relies on the string form.
- **`VirtualizedMessages` and `ChatInput` both keyed by `conversationId`**: causes a remount on conversation switch. Cheap (Virtuoso is lazy and ChatInput is small), and matches the expected UX of starting at the bottom of the target conversation with a clean composer.
- **Cmd+K AI Assistant entry**: previously called `openAIAssistant()` (drawer); now calls `startNewConversation()` + `openAIAssistantModal()`. The drawer entry point still exists for other call sites — only the palette path changed.
- **Rollback**: revert; no migrations or persisted state.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | Cloud / OSS / Enterprise (wherever AI Assistant is enabled via `ai_assistant_url`) |
| Change Type | Feature + Bug Fix |
| Description | AI Assistant: composer auto-grows for long prompts; focus ring scoped to the textarea row; Cmd+K palette gains an "Open AI Assistant" entry that opens the modal (matching Cmd+J) with the `cmd+j` shortcut hint; streaming auto-scroll respects user position but re-anchors on user sends and conversation switches; browser title resolves on dynamic routes; context picker is a fully-keyboard-navigable tablist; voice recording controls are keyboard-operable and SR-announced; icon-only buttons across the assistant get visible hover tooltips; clarification forms enforce required fields, parse numeric defaults correctly, and use proper ARIA grouping; approval-diff dialog no longer surfaces a stray "Copy" tooltip on open; modal header icons render in the neutral secondary color; the composer auto-focuses on new-conversation / conversation-switch. |

---

### 📋 Checklist

- [x] Manually tested

---

## 👀 Notes for Reviewers

- **`getRouteKey`** runs on every page render — please eyeball the dynamic-template branch in `frontend/src/container/AppLayout/utils.ts` and confirm the new title resolutions look right across the six routes listed under Risk.
- **`VirtualizedMessages` and `ChatInput` are both keyed by `conversationId`** in `ConversationView.tsx`. Two lines, but it's how per-conversation scroll-anchor state resets *and* how the composer refocuses on new chat / switch — please don't drop them during cleanup.
- **Context picker tablist**: single stable `id="ai-context-tabpanel"` shared by every tab's `aria-controls`. The panel's `aria-labelledby` swaps to the active tab so the announced label stays correct. Worth a quick keyboard run-through in dev.
- **Composer cap (200px)**: please eyeball inside `AIAssistantPanel` on a short viewport — happy to lower or make it variant-aware if it crowds the message list.
- **Analytics `source`** now lives as typed enums in `frontend/src/container/AIAssistant/events.ts` (`AIAssistantOpenSource`, `VoiceInputSource`). Wire values unchanged; new emit sites should use the enums.
- **Tooltips**: new icon-only buttons in the AI Assistant should use `TooltipSimple` rather than the native HTML `title` attribute, matching the rest of the surface.
- **DS Button layout override via CSS vars**: the context-picker rows set `--button-justify-content: flex-start` on their wrapper class rather than fighting class specificity. `@signozhq/ui/button` exposes `display`, `justify-content`, `align-items`, `white-space` via `var(--button-*, …)` — please reach for the variable first whenever you need to override DS layout defaults.
- **`ToggleGroupSimple` deliberately not used** for the approval-diff toolbar: the items-API renders the icon inside its own trigger, so `TooltipSimple` around the icon can't anchor on the button (hover surfaces the wrong region). Manual `ToggleGroup` + `ToggleGroupItem` with `TooltipSimple` wrapping each item button is the only shape that keeps per-item tooltips working.
- **Approval-diff `onOpenAutoFocus`**: Radix Dialog auto-focuses the first focusable child on open. Without `e.preventDefault()`, the "Copy before" Button gains focus and its `TooltipSimple` opens immediately because Radix Tooltip opens on focus too. Generally applicable to any `<DialogContent>` whose first focusable child is a tooltip trigger.
- **Modal vs. panel header parity**: both surfaces now share `color="secondary"` + `prefix` slot for the header icon buttons. The modal previously omitted `color`, which is why those icons looked blue.
- **Known pre-existing warning not in scope**: `@signozhq/ui@0.0.19`'s `TooltipContentInner` triggers React's "Function components cannot be given refs" warning from every `TooltipSimple` portal (also affects `AIAssistantTrigger`, `AIAssistantPanel`, `MCPServerSettings/CopyIconButton` on `main`). Needs a dep bump.


makeavish:
CMD + K global search should show AI Assistant with CMD + J

frontend/src/container/AIAssistant/components/ApprovalCard/ApprovalCard.tsx
```diff
@@ -100,16 +101,17 @@ export default function ApprovalCard({
...
-						</Button>
+						<TooltipSimple title="Expand diff">
+							<Button
+								variant="link"
+								size="sm"
+								color="secondary"
+								onClick={(): void => setDiffExpanded(true)}
+								aria-label="Expand diff"
+							>
+								<Maximize2 size={12} />
```
YounixM: prefixIcon

frontend/src/container/AIAssistant/components/ApprovalCard/ApprovalCard.tsx
```diff
@@ -141,25 +143,33 @@ export default function ApprovalCard({
 									}
 								}}
 							>
-								<ToggleGroupItem value="split" aria-label="Split view">
-									<Columns2 size={12} />
-								</ToggleGroupItem>
-								<ToggleGroupItem value="unified" aria-label="Unified view">
-									<List size={12} />
-								</ToggleGroupItem>
+								<TooltipSimple title="Split view">
```
YounixM, line 146: https://periscope.signoz.io/?path=/docs/components-togglegroup-togglegroupsimple--docs

frontend/src/container/AIAssistant/components/ChatInput/ChatInput.tsx
```diff
@@ -877,15 +922,34 @@ export default function ChatInput({
...
+									role="tablist"
+									aria-orientation="vertical"
+									aria-label="Context categories"
+								>
 									{CONTEXT_CATEGORIES.map((category) => {
 										const CategoryIcon = CONTEXT_CATEGORY_ICONS[category];
 										const isActive = activeContextCategory === category;
 										return (
-											<div
+											<button
```
YounixM: use component

frontend/src/container/AIAssistant/components/ChatInput/ChatInput.tsx
```diff
@@ -939,16 +1038,21 @@ export default function ChatInput({
...
+											filteredContextOptions.map((option, index) => {
 												const isSelected = selectedContexts.some(
 													(item) =>
 														item.category === activeContextCategory &&
 														item.entityId === option.id,
 												);
 
 												return (
-													<div
+													<button
```
YounixM: use component

frontend/src/container/AIAssistant/components/ChatInput/ChatInput.tsx
```diff
@@ -959,11 +1063,33 @@ export default function ChatInput({
 																option.value,
 															)
 														}
+														onKeyDown={(e): void => {
+															const count = filteredContextOptions.length;
+															if (e.key === 'ArrowDown') {
+																e.preventDefault();
```
YounixM: improve code

frontend/src/container/AIAssistant/components/ChatInput/ChatInput.tsx
```diff
@@ -977,14 +1103,22 @@ export default function ChatInput({
...
-									<X size={12} />
-								</div>
+							<div
+								className={styles.micRecording}
+								role="status"
+								aria-live="polite"
+								aria-label="Recording voice input"
+							>
+								<TooltipSimple title="Discard recording">
+									<button
```
YounixM: use component

frontend/src/container/AIAssistant/components/ChatInput/ChatInput.tsx
```diff
@@ -1031,16 +1168,18 @@ export default function ChatInput({
...
+						<TooltipSimple title="Send message">
+							<Button
+								variant="solid"
+								size="icon"
+								color="primary"
+								onClick={isListening ? handleStopAndSend : handleSend}
+								disabled={disabled || (!text.trim() && pendingFiles.length === 0)}
+								aria-label="Send message"
+							>
+								<Send size={14} />
```
YounixM: prefixIcon

frontend/src/container/AIAssistant/components/ClarificationForm/ClarificationForm.tsx
```diff
@@ -63,7 +63,18 @@ export default function ClarificationForm({
 		setAnswers((prev) => ({ ...prev, [id]: value }));
 	};
 
+	const isFormValid = fields.every(
+		(f) => !f.required || isFieldFilled(f, answers[f.id]),
+	);
+
 	const handleSubmit = async (): Promise<void> => {
+		// Defensive guard for direct handler invocation. The disabled Submit
```
YounixM: remove comment

frontend/src/container/AIAssistant/components/MessageFeedback/MessageFeedback.tsx
```diff
@@ -172,6 +173,8 @@ export default function MessageFeedback({
 							variant="ghost"
 							color="secondary"
 							onClick={(): void => handleVote('positive')}
+							aria-label="Good response"
+							aria-pressed={vote === 'positive'}
```
YounixM: move to enum

frontend/src/container/AIAssistant/ConversationView/ConversationView.tsx
```diff
@@ -172,6 +172,12 @@ export default function ConversationView({
 	return (
 		<div className={styles.conversation}>
 			<VirtualizedMessages
+				// Remount on conversation switch so the scroll-anchor refs
```
YounixM: remove the comment

YounixM (commented):
Added review comments

YounixM (approved):

PR #11367, opened 2026-05-20
chore: create source field in dashboards

srikanthccv:
## Pull Request

Fixes https://github.com/SigNoz/engineering-pod/issues/5004


pkg/types/dashboardtypes/dashboard.go
```diff
@@ -19,8 +21,66 @@ var (
 	ErrCodeDashboardNotFound           = errors.MustNewCode("dashboard_not_found")
 	ErrCodeDashboardInvalidData        = errors.MustNewCode("dashboard_invalid_data")
 	ErrCodeDashboardInvalidWidgetQuery = errors.MustNewCode("dashboard_invalid_widget_query")
+	ErrCodeDashboardInvalidSource      = errors.MustNewCode("dashboard_invalid_source")
+	ErrCodeDashboardImmutable          = errors.MustNewCode("dashboard_immutable")
 )
 
+type Source struct {
+	valuer.String
+}
```
therealpandey:
```suggestion
type Source struct {
	s valuer.String
}
```

This is a true enum since almost all methods have been overridden. Can you please move to a `source.go` file with the above struct syntax?

Subclassing `valuer.String` will give you access to the valuer methods which can be reused instead of writing your own. Can try directly calling `Scan/Value` on `valuer.String`.

pkg/types/dashboardtypes/dashboard.go
```diff
@@ -238,6 +312,43 @@ func (storableDashboardData *StorableDashboardData) GetWidgetIds() []string {
 	return widgetIds
 }
 
+func (dashboard *Dashboard) ErrIfNotMutable() error {
+	if dashboard.Source == SourceIntegration {
+		return errors.Newf(errors.TypeForbidden, ErrCodeDashboardImmutable, "integration dashboards cannot be modified")
```
therealpandey:
```suggestion
		return errors.Newf(errors.TypeInvalidInput, ErrCodeDashboardImmutable, "integration dashboards cannot be modified")
```

Forbidden is connected to permissions mostly. If a user tried to do an action which is not impossible, the input itself was invalid.

`PUT /dashboards/:id -> can't be edited -> 400 can't be edited`

therealpandey (changes requested):

pkg/types/dashboardtypes/dashboard.go
srikanthccv: Not really almost all methods though. Just the `Scan`/`Value`. If I have to use `s valuer.String` then every method will have to be implemented (that delegate to valuer). Is that the suggestion?

pkg/types/dashboardtypes/dashboard.go
srikanthccv: The intent was "the operation is just disallowed for that resource" for anyone. So went back and forth between choosing `errors.TypeInvalidInput` and `errors.TypeForbidden` and went with `errors.TypeForbidden`. On the other hand, the user resource continue to get `errors.TypeInvalidInput`. I am fine either way, but `errors.TypeForbidden` felt more natural. Will update.

pkg/types/dashboardtypes/source.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/valuer"
+)
+
+type Source struct {
+	s valuer.String
+}
+
+var (
+	SourceUser        = Source{s: valuer.NewString("user")}
```
namanv3, line 16: wondering if we should call it custom instead of user so that service account created dashboards are also included naturally

pkg/types/dashboardtypes/dashboard.go
```diff
@@ -64,6 +68,10 @@ func NewStorableDashboardFromDashboard(dashboard *Dashboard) (*StorableDashboard
 		return nil, errors.Wrapf(err, errors.TypeInvalidInput, errors.CodeInvalidInput, "id is not a valid uuid")
 	}
 
+	if !dashboard.Source.IsValid() {
+		return nil, errors.Newf(errors.TypeInvalidInput, ErrCodeDashboardInvalidSource, "invalid dashboard source %q", dashboard.Source.StringValue())
```
namanv3: should we add allowed values in the err message?

pkg/sqlmigration/085_add_source_to_dashboard.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	sourceColumn := &sqlschema.Column{
+		Name:     sqlschema.ColumnName("source"),
+		DataType: sqlschema.DataTypeText,
+		Nullable: false,
+	}
+
+	// backfill existing rows with 'user' before the NOT NULL flip.
+	sqls := migration.sqlschema.Operator().AddColumn(table, uniqueConstraints, sourceColumn, "user")
```
namanv3, line 58: commenting to remember to change this to "source" in case we decide to change the enum value

frontend/src/api/generated/services/sigNoz.schemas.ts
```diff
@@ -2999,6 +2999,11 @@ export interface CoretypesPatchableObjectsDTO {
 	deletions: CoretypesObjectGroupDTO[] | null;
 }
 
+export enum DashboardtypesSourceDTO {
+	user = 'user',
+	system = 'system',
```
pradeepitm12, line 3004: Looking at the source column with the user/system/integration enum — wanted to bring some context from #11133 (system-dashboard/overview PR).
Here we are susing source as owner of the dashboard. But it does not carry specific identity of dashbaord. Today we just have one system dahsobard `ai-o11y-overview` but discussion in the PRD:https://www.notion.so/signoz/Overview-Page-ai-o11y-33efcc6bcd1980dca23af75ca81d955a says we may have other system dashboards like `service-overview`, `infra-overview` etc.

pkg/types/dashboardtypes/dashboard.go
therealpandey: Yesss

pkg/sqlmigration/085_add_source_to_dashboard.go
pradeepitm12, line 58: This migration runs once, and adds `user` to all the existing rows source, which is right as well we would not have any system dashboard so far in the db.

pkg/types/dashboardtypes/dashboard.go
srikanthccv: done

pkg/types/dashboardtypes/dashboard.go
srikanthccv: done

pkg/types/dashboardtypes/dashboard.go
srikanthccv: will add. I wasn't sure if adds value because all user generated dashboards have constant valid value passed to this but its good idea to generally add allowed values.

pkg/types/dashboardtypes/source.go
srikanthccv: I think "custom" is vague. You'd have to explain it to someone why it's "custom". On the other hand, "user" doesn't need explanation. 

Regarding the service account, isn't service account is also user, just that a non-human one? 

I think user is fine.

namanv3 (approved):

therealpandey (approved):

AshwinBhatkal (approved):

PR #11368, opened 2026-05-20
feat: FTS on logs

piyushsingariya:
## Pull Request

---

### 📄 Summary

This PR introduces a `search()` function for Full Text Search (FTS) across the entire logs table — not just `body`, but also `severity_text`, `trace_id`, `span_id`, all attribute maps (`attributes_string`, `attributes_number`, `attributes_bool`), all resource maps (`resources_string`, `resource`), and the body JSON column.

Two Text Search entry points exist:
1. **Bare / quoted text** (`error`, `"healthcheck"`) — existing free-text grammar (`freeText` rule); still targets only the `body` column (via `freeTextColumn`), unchanged behavior.
2. **`search()` function** — new grammar keyword added to both the ANTLR grammar (`FilterQuery.g4`) and the Go/TypeScript generated parsers. `search('error')`  this fans out across all FTS field contexts.

**Grammar rename:** the old `fullText` grammar rule is renamed to `freeText`; a new `fullText` rule is introduced for the `search(...)` call. A new case-insensitive `SEARCH` lexer token is added.

`search()` is **restricted to a 6-hour time window** (`FTSMaxWindowNs`) to guard against expensive full-table scans. It is enabled only for log queries — traces and metrics leave `ftsConditionFunc` nil, and calling `search()` on them returns an explicit error.

For map columns (`attributes_*`, `resources_*`) the WHERE fragment uses `arrayExists(x -> match(x, ?), mapKeys/mapValues(col))`. For string / low-cardinality columns it uses `match(LOWER(col), LOWER(?))`. JSON body uses `match(LOWER(toString(body)), LOWER(?))`. Non-string map value columns are cast with `arrayMap(x -> toString(x), mapValues(col))` before matching.

A user-facing warning is emitted on every `search()` hit: _"Full text searches across all fields and will be slow and expensive. Consider using specific field..."_

#### Renames / breaking changes in internal API

| Old name | New name | Location |
|---|---|---|
| `fullText` grammar rule | `freeText` | `FilterQuery.g4` + all generated parsers |
| `FullTextColumn` (opts field) | `FreeTextColumn` | `FilterExprVisitorOpts` |
| `fullTextColumn` (struct field) | `freeTextColumn` | `logQueryStatementBuilder`, `filterExpressionVisitor` |
| `SkipFullTextFilter` | `SkipFreeTextFilter` | `FilterExprVisitorOpts` |
| `BodyFullTextSearchDefaultWarning` | `BodyFreeTextSearchWarning` | `pkg/querybuilder/constants.go` |
| `NewLogQueryStatementBuilder` signature | gains `ftsConditionFunc qbtypes.FTSConditionFunc` | `pkg/telemetrylogs/statement_builder.go` |
| `addFilterCondition` signature | gains `enableFTS bool` parameter | internal |

New additions: `FTSConditionFunc` type in `querybuildertypes`, `FullTextSearchDefaultWarning` and `FTSMaxWindowNs` constants in `pkg/querybuilder/constants.go`, `ftsColumns()` helper in `field_mapper.go`, `ConditionForSearch` method on `conditionBuilder`, `LogsV2ResourceJSONColumn` constant, and new `SkipFullTextSearch` flag.

#### Issues closed by this PR

Closes signoz/engineering-pod#3792

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] ♻️ Refactor

---

- Tests added/updated: ✅
  - `pkg/telemetrylogs/stmt_builder_test.go` — covers `search('...')`, `search(unquoted)`, `NOT search(...)`, combined `search() AND field=val`, and the 6-hour window guard.
  - `pkg/querybuilder/where_clause_visitor_test.go` — updated to reflect `FreeTextColumn`/`SkipFreeTextFilter` renames and new `SkipFullTextSearch` flag.
  - `pkg/telemetrylogs/filter_expr_logs_test.go` — updated to use `BodyFreeTextSearchWarning` instead of the old `BodyFullTextSearchDefaultWarning`.
  - `tests/integration/tests/querier/15_fts.py` — new integration test: 10 logs, one unique token per field context (`severity_text`, `trace_id`, `span_id`, body key, body value, `attributes_string` key/value, `attributes_number` key, `resources_string` key/value). Each FTS form (bare, quoted, `search()`) is validated to return exactly the right row with the FTS warning present.
  - `tests/integration/tests/querier_json_body/03_fts.py` — same suite replayed with JSON body enabled.
- Manual verification: ✅
- Edge cases covered:
  - Time window exceeding 6 hours → error returned, no query executed.
  - `search()` called on traces/metrics query → explicit error "search() is only supported for log queries".
  - `search()` with no argument → descriptive parse error.
  - Map columns with non-string value type → values cast via `arrayMap(x -> toString(x), mapValues(col))` before matching.
  - Scalar (aggregation) queries → `enableFTS=false` prevents `search()` from running.

---

- **Blast radius:** Log Stmt Builder and all callers of `NewLogQueryStatementBuilder` (new required `ftsConditionFunc` parameter). The `ftsConditionFunc` is `nil` for all trace/metric query builders, so those code paths are unaffected. Scalar (aggregation) queries pass `enableFTS=false`, preventing `search()` from running there.
- **Potential regressions:**
  - Any caller constructing `FilterExprVisitorOpts` with `FullTextColumn` or `SkipFullTextFilter` must be updated to `FreeTextColumn` / `SkipFreeTextFilter`.
  - Client code keying off `BodyFullTextSearchDefaultWarning` will no longer see it — the new constant is `BodyFreeTextSearchWarning`.
  - Bare text / quoted text behavior is **unchanged** — they still target only the `body` column. Only `search()` fans out.
- **Rollback plan:** Remove the `ftsConditionFunc` assignment in `NewLogQueryStatementBuilder` (pass `nil`) to disable `search()` entirely. No schema or migration changes are involved.

---

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | A new `search()` filter function is available for log queries. It searches across all log fields (`severity_text`, `trace_id`, `span_id`, body JSON, all attribute and resource maps). Searches are limited to a 6-hour time window and emit a performance warning. Bare/quoted free-text search behavior is unchanged (still targets `body` only). |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- **`FTSConditionFunc`** is a function type `func(ctx context.Context, value any, sb *sqlbuilder.SelectBuilder) (string, error)`. Log statement builders wire `conditionBuilder.ConditionForSearch` to it; all other builders leave it `nil`. The visitor checks for `nil` and returns a user-facing error if `search()` is called on a non-log signal.
- **`ftsColumns()`** in `field_mapper.go` is the single source of truth for which physical columns `search()` fans out across, keyed by `FieldContext`. Adjust it to add or remove columns from FTS coverage.
- **`FreeTextColumn`** (formerly `FullTextColumn`) is retained and still used for bare/quoted free-text search in log queries — it points to the `body` column. Only `search()` uses the full fan-out path.
- The grammar split (`freeText` for bare/quoted, `fullText` for `search()`) means the two paths are independently skippable via `SkipFreeTextFilter` and `SkipFullTextSearch` flags respectively.

---


pkg/querybuilder/where_clause_visitor.go
```diff
@@ -721,19 +730,22 @@ func (v *filterExpressionVisitor) VisitFullText(ctx *grammar.FullTextContext) an
...
 	if v.skipFunctionCalls {
 		return SkipConditionLiteral
 	}
 
+	// search() must be handled before visiting params: unquoted tokens like
+	// search(error) are parsed as a key context, and visiting them through VisitKey
+	// would append "key not found" errors before we can treat the text as a search string.
+	if ctx.SEARCH() != nil {
+		return v.visitSearchFunction(ctx)
+	}
```
tushar-signoz: Hey @piyushsingariya 

You're handling search different from other function calls which is fine. However you're leading the user in wrong direction by making search(error) work. IMO this should fail. 
Will following work in the manner user intended or do we return an error which is informative enough so that user self corrects ?
`search(body, "error")` 
`search("err1", "err2")`
`search(attributes, "err")`

piyushsingariya: > `search(body, "error")` `search("err1", "err2")` `search(attributes, "err")`

Scoped search not supported yet. 


> You're handling search different from other function calls which is fine.

Yes, I tried to use our existing FieldFor structure, but it wasn't fitting entirely.

> making search(error) work. IMO this should fail.

The tried to replicate behavior from base FTS, where user can search without quotes too. @srikanthccv  What're your thoughts??

tushar-signoz: @piyushsingariya 
Its not about what is supported or not. Its about sending right errors so that user is aware, what is supported and what is not.

piyushsingariya: Notes: I've to properly throw errors based on argument check. and without quotes will be supported

grammar/FilterQuery.g4
```diff
@@ -107,7 +107,7 @@ fullText
  *    ...
  */
 functionCall
-    : (HASTOKEN | HAS | HASANY | HASALL) LPAREN functionParamList RPAREN
+    : (HASTOKEN | HAS | HASANY | HASALL | SEARCH) LPAREN functionParamList RPAREN
```
tushar-signoz: Lets separate out SEARCH from function call and make it first class citizen. 
This would also help us later if want to add more functionality in search.

tushar-signoz (changes requested):

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -337,10 +346,11 @@ func (v *filterExpressionVisitor) VisitPrimary(ctx *grammar.PrimaryContext) any
 			return SkipConditionLiteral
 		}
 
-		if v.fullTextColumn == nil {
+		if v.freeTextColumn == nil {
 			v.errors = append(v.errors, "full text search is not supported")
```
tushar-signoz: nit: consider renaming the error also to `free text search`

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -714,24 +725,79 @@ func (v *filterExpressionVisitor) VisitFullText(ctx *grammar.FullTextContext) an
...
 		v.warnings = append(v.warnings, BodyFullTextSearchDefaultWarning)
 	}
 
 	return cond
 }
 
-// VisitFunctionCall handles function calls like has(), hasAny(), etc.
+// VisitSearchCall handles the search() function call.
+func (v *filterExpressionVisitor) VisitSearchCall(ctx *grammar.SearchCallContext) any {
+	if v.skipFunctionCalls {
```
tushar-signoz: add another `v.skipSearch`

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -714,24 +725,79 @@ func (v *filterExpressionVisitor) VisitFullText(ctx *grammar.FullTextContext) an
...
+	if paramCtx.Value() != nil {
+		raw := v.Visit(paramCtx.Value())
+		searchText = fmt.Sprintf("%v", raw)
+	} else {
+		v.errors = append(v.errors, "search() parameter must be a quoted string, e.g. search('error')")
+		return ErrorConditionLiteral
+	}
+
+	if v.endNs > 0 && v.startNs > 0 && (v.endNs-v.startNs) > FTSMaxWindowNs {
+		v.errors = append(v.errors, "full text search is restricted to a maximum of 6-hour time window")
```
tushar-signoz: nit: instead of `full text search` lets use `search()`

grammar/FilterQuery.g4
```diff
@@ -110,6 +111,14 @@ functionCall
     : (HASTOKEN | HAS | HASANY | HASALL) LPAREN functionParamList RPAREN
     ;
 
+/*
+ * Search call: search(field) or search("query text")
```
tushar-signoz: nit: search(field) isn't supported right ?

piyushsingariya: It was miss, addressed

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -714,24 +726,71 @@ func (v *filterExpressionVisitor) VisitFullText(ctx *grammar.FullTextContext) an
...
+}
+
+// VisitFullText handles the search() function call.
+func (v *filterExpressionVisitor) VisitFullText(ctx *grammar.FullTextContext) any {
+	if v.skipFunctionCalls || v.skipFullTextSearch {
+		return SkipConditionLiteral
+	}
+	// ftsCondition nil means search() is not enabled for this signal.
+	// Only log statement builders set these; traces/metrics leave them nil.
+	if v.ftsCondition == nil {
```
tushar-signoz, line 753: Why can't we use v.conditionBuilder.ConditionForSearch here and need to pass the function ?

tests/integration/tests/querier_json_body/01_logs_json_body_new_qb.py
```diff
@@ -1217,22 +1217,42 @@ def _body_messages(response: requests.Response) -> list[str]:
...
             "name": "msg.fts_quoted_without_quotes",
             "requestType": "raw",
             "expression": "Payment",
             "aggregation": "count()",
             "validate": lambda r: len(get_rows(r)) == 2 and all("Payment" in b.get("message", "") for b in _get_bodies(r)) and r.json().get("data", {}).get("warning") is not None,
         },
+        # ── Full Text Search: search() explicit function ──────────────────────────────────────
+        # search("Payment") is equivalent to the bare-text '"Payment"' FTS:
+        # both fan out a case-insensitive match across severity_text, trace_id,
+        # span_id, body JSON values, attribute values, and resource values.
```
tushar-signoz: Stale comment

tests/integration/tests/querier_json_body/01_logs_json_body_new_qb.py
```diff
@@ -1217,22 +1217,42 @@ def _body_messages(response: requests.Response) -> list[str]:
...
+        {
+            "name": "msg.search_quoted",
+            "requestType": "raw",
+            "expression": 'search("Payment")',
+            "aggregation": "count()",
+            "validate": lambda r: len(get_rows(r)) == 2 and set(_body_messages(r)) == payment_messages and r.json().get("data", {}).get("warning") is not None,
+        },
+        # NOT search("Payment") — inverted FTS: logs that do NOT have "payment"
```
tushar-signoz: Avoid using FTS term as its loaded now, Free and Full

pkg/querybuilder/where_clause_visitor.go
piyushsingariya: I think we discussed this over call, lmk if I can close this.

tushar-signoz:
There have been many changes and would require a lot of rebasing.

PR #11372, opened 2026-05-20
chore: reduce integration test build time

nityanandagohain:
## Pull Request

---

### 📄 Summary
It take around 10 mins everytime I run `make py-test-setup` after a teardown. This change brings it down to <2 mins.

* Use `docker buildx build` directly; stream output into pytest logs.
* Build image in parallel with ClickHouse migrator.
* Add BuildKit cache mounts for Go module and build caches.
* Use type=gha cache backend instead of actions/cache tarball roundtrip
* Add ghaction-github-runtime so buildx sees ACTIONS_RUNTIME_TOKEN
* Give Zeus a stable network alias so ZEUSURL stops busting cache
* Pin Dockerfile syntax to 1.13
* Add py-docker-clean / py-test-clean Makefile targets

### Results

Local
----
- Cold-ish run: `2:10.79`
- Follow-up run: `1:29.22`

Github
---
image

Previously they use to take > 7mins

#### Issues closed by this PR
Fixes https://github.com/SigNoz/platform-pod/issues/2869

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated:  ✅ 
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

* Do review the github ci files
* `py-clean` remains Python-only to avoid surprising Docker cache deletion. Docker cleanup is available through the explicit `py-docker-clean` and `py-test-clean` targets.
---


tushar-signoz:
Hey @nityanandagohain 
If I make a change in signoz codebase, will it use the cached signoz image or build a new one ?

nityanandagohain:
With this change it will build a new one, but it is way faster now.

.github/workflows/e2eci.yaml
```diff
@@ -45,9 +45,15 @@ jobs:
...
     timeout-minutes: 30
+    env:
+      SIGNOZ_BUILDX_GHA_SCOPE: signoz-e2e
     steps:
       - name: checkout
         uses: actions/checkout@v4
+      - name: setup-buildx
+        uses: docker/setup-buildx-action@v3
+      - name: expose-gha-runtime
+        uses: crazy-max/ghaction-github-runtime@v3
```
H4ad: nit: would be better to pin this to commit hash

nityanandagohain: Didn't get what you mean here @H4ad

H4ad: https://github.com/crazy-max/ghaction-github-runtime/commit/3cb05d89e1f492524af3d41a1c98c83bc3025124

Instead of `@v3`, we can pin via commit hash `@3cb05d89e1f492524af3d41a1c98c83bc3025124`

It's better to avoid supply chain attacks from external github actions, specially the ones we don't control/trust like docker/github.

nityanandagohain: Done, switched to a new version as well. Thanks for the suggestion.

therealpandey:
Closed by https://github.com/SigNoz/signoz/pull/12366

PR #11386, opened 2026-05-21
fix: added fix for dashboard dropdown menus

ahrefabhi:
## Pull Request

---

Continues the antd → Periscope migration started in #11375. Two dashboard overflow menus (the `…` icon on the dashboard list rows and the `…` icon on a dashboard's header) were antd `<Popover>` containers stuffed with a flat list of `<Button>`s — not real menus. This PR replaces both with `DropdownMenuSimple` from `@signozhq/ui`, driven by declarative `MenuItem[]` arrays.

The non-trivial bit is `<DeleteButton>`: it owns its own `Modal.useModal()` confirmation dialog plus a disabled-state tooltip, so it couldn't drop into a `MenuItem.onClick` as-is. Extracted a `useDeleteDashboardDialog(...)` hook that returns `{ openConfirmation, isDisabled, tooltipContent, contextHolder }` — the existing `<DeleteButton>` (still used by the dashboards-list table) consumes the hook internally, and the new menus call the hook at their own scope.

For the row-level menu, the hook needs to be called per-row, which can't happen inside `columns.render`. Extracted a tiny `<DashboardRowActions>` sub-component so each row has its own hook scope.

##### Before
image
image

##### After
image
image

#### Issues closed by this PR
N/A — follow-up to #11375.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: none — pure UI refactor; existing tests cover the underlying delete flow via `DeleteButton`.
- Manual verification:
  - **Dashboard header menu** (`/dashboard/:id`): every item triggers the correct action; Lock/Unlock is hidden for non-author non-admin; Rename / New section hide when locked; Lock tooltip shows for integration-created dashboards.
  - **Dashboard row menu** (`/dashboard`): View / Open in New Tab / Copy Link / Export JSON / Delete — each fires its own action only, no double-navigation; Delete confirmation modal renders and routes correctly.
  - Delete disabled-state tooltip ("…locked dashboard…") still renders on hover when the dashboard is locked.
- Edge cases covered:
  - Clicking a menu item no longer also fires the row-level `onClick` (was navigating to the dashboard on every menu click).
  - `display: contents` wrapper / per-call SCSS overrides have been removed once equivalent fixes land in `@signozhq/ui` (see the periscope-side companion change to `dropdown-menu-content.tsx` and `dropdown-menu.module.scss`).

---

- **Blast radius:** Two dashboard pages (header + list). `useDeleteDashboardDialog` is also consumed by `<DeleteButton>` which is used in the dashboards-list table.
- **Potential regressions:**
  - `<DeleteButton>` was refactored to consume the hook — the rendered output is byte-for-byte the same, but worth eyeballing the dashboards-list table delete flow.
  - The new menus rely on `@signozhq/ui` `DropdownMenuSimple`; visual polish depends on the Periscope library version that ships the `min-width` / portal-click-bubble / `clickable` cursor changes. Until that publishes, the local result will look slightly different from the design (slightly wider menu, no pointer cursor on items). The functional behavior is correct.
- **Rollback plan:** Revert this PR; no schema/data changes.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | N/A — internal refactor, no user-facing behavior change |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- Base branch is `chore/button-migration`, not `main`. Review the diff against that branch.
- This PR depends on the companion Periscope (`@signozhq/ui`) source changes for the best visual result (`min-width: 0` default on items, `clickable` modifier, and `DropdownMenuContent` stopping click propagation). The code here works against the currently-published `0.0.20` — the lib changes just remove the rough edges (extra width, no pointer, click bubbling).
- New file: `frontend/src/container/ListOfDashboard/TableComponents/DashboardRowActions.tsx` — extracted only because hooks can't live in `columns.render`.
- `useDeleteDashboardDialog` is colocated with `DeleteButton.tsx` to keep the delete-flow logic in one place.
- Two unrelated lint cleanups in `DashboardDescription/index.tsx`: `selectedData` is now memoized and `handleLockDashboardToggle` is wrapped in `useCallback` — both were already flagged by `react-hooks/exhaustive-deps` and surfaced when I introduced the menu items `useMemo`.

### DIFF
--- a/frontend/src/container/DashboardContainer/DashboardDescription/index.tsx
+++ b/frontend/src/container/DashboardContainer/DashboardDescription/index.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useRef, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { FullScreenHandle } from 'react-full-screen';
 import { Layout } from 'react-grid-layout';
 import { useTranslation } from 'react-i18next';
@@ -14,15 +14,18 @@ import {
 	LockKeyhole,
 	PenLine,
 	Plus,
+	Trash2,
 	X,
 } from '@signozhq/icons';
-import { Card, Input, Modal, Popover, Tag, Tooltip } from 'antd';
+import { Card, Input, Modal, Tag, Tooltip } from 'antd';
 import { Button } from '@signozhq/ui/button';
+import type { MenuItem } from '@signozhq/ui/dropdown-menu';
+import { DropdownMenuSimple as Dropdown } from '@signozhq/ui/dropdown-menu';
 import { Typography } from '@signozhq/ui/typography';
 import logEvent from 'api/common/logEvent';
 import ConfigureIcon from 'assets/Integrations/ConfigureIcon';
 import { PANEL_GROUP_TYPES, PANEL_TYPES } from 'constants/queryBuilder';
-import { DeleteButton } from 'container/ListOfDashboard/TableComponents/DeleteButton';
+import { useDeleteDashboardDialog } from 'container/ListOfDashboard/TableComponents/DeleteButton';
 import DateTimeSelectionV2 from 'container/TopNav/DateTimeSelectionV2';
 import { useDashboardVariables } from 'hooks/dashboard/useDashboardVariables';
 import { useGetPublicDashboardMeta } from 'hooks/dashboard/useGetPublicDashboardMeta';
@@ -90,12 +93,16 @@ function DashboardDescription(props: DashboardDescriptionProps): JSX.Element {
 
 	const isPublicDashboardEnabled = isCloudUser || isEnterpriseSelfHostedUser;
 
-	const selectedData = dashboardData
-		? {
-				...dashboardData.data,
-				uuid: dashboardData.id,
-			}
-		: ({} as DashboardData);
+	const selectedData = useMemo(
+		() =>
+			dashboardData
+				? {
+						...dashboardData.data,
+						uuid: dashboardData.id,
+					}
+				: ({} as DashboardData),
+		[dashboardData],
+	);
 	const { dashboardVariables } = useDashboardVariables();
 
 	const {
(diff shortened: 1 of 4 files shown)


AshwinBhatkal:
These have been fixed on Tuesday. Can you please rebase with main? The UX should be corrected

ahrefabhi:
> These have been fixed on Tuesday. Can you please rebase with main? The UX should be corrected

This is not the same @AshwinBhatkal, Previous dropdown was made using antd Buttons/Popover. This is a stacked PR on top of antd->@signozhq/ui buttons. 

After the migration, these dropdowns start to break.

frontend/src/container/DashboardContainer/DashboardDescription/index.tsx
```diff
@@ -360,114 +482,7 @@ function DashboardDescription(props: DashboardDescriptionProps): JSX.Element {
...
-										isLocked={isDashboardLocked}
-										routeToListPage
-									/>
-								</section>
-							</div>
-						}
-						trigger="click"
-						placement="bottomRight"
-					>
+					<Dropdown menu={{ items: dashboardMenuItems }} align="end">
```
AshwinBhatkal, line 485: nit: we can memoize the whole menu object instead of just items

frontend/src/container/ListOfDashboard/TableComponents/DeleteButton.tsx
```diff
@@ -15,34 +15,38 @@ import { USER_ROLES } from 'types/roles';
...
 
-export function DeleteButton({
+interface UseDeleteDashboardDialogResult {
+	openConfirmation: () => void;
+	isDisabled: boolean;
+	tooltipContent: string;
+	contextHolder: React.ReactElement;
+}
+
+export function useDeleteDashboardDialog({
```
AshwinBhatkal, line 33: Move this to it's own file

AshwinBhatkal (commented):
Got it. 1 small comment.

Also, I got confused with the before/after. Ideally there is no before after then. We had antd earlier with the right UI. When we migrate to signoz dropdown, we fix even if it breaks something. Ideally the before and after should look the same. One with antd and one with signoz dropdown.

PR #11387, opened 2026-05-21
fix: add check for percentile aggregation for non-histogram metrics

namanv3:
## Pull Request

---

### 📄 Summary
If someone is sending percentile aggregations for Sum metrics, currently they get a something went wrong error. This can happen in a faulty query.

Post this change:
Screenshot 2026-05-21 at 11 58 59

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/4189

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: unit test added
- Manual verification: Via API 
- Edge cases covered: None required

---

- Blast radius:  Query Range V5
- Potential regressions: None
- Rollback plan: Revert PR

---

| Field | Value |
|------|-------|
| Description | If a query range API call accidentally sends a percentile aggregation for a non-histogram metric, then a proper error message is sent |

---

### 📋 Checklist
- [x] Manually tested

---


srikanthccv (commented):
There should be only two places where the validation happens for the `MetricAggregation`.
1. At the request unmarshal validation
2. Post metric metadata resolve step for logical checks that can't be done at the request validation.

Right now, it's sort of all over the place and misses these fixes in meter (regardless of who owns, the fixes should go as it heavily relies the metric builder).

Please consolidate.

1. re-check the `validation.go`;
2. add something like `func (m MetricAggregation) ValidateForType() error { ... }` and keep all the type X aggregation compatibility knowledge lives in one place
3. call it at the one choke point where metric type is known
4. check any other cases
5. the change in this PR says, percentile ops are supported for the summary. No, they are not. The summary metric already sends the quantiles so there is no data to further derive the quantile. Please send a summary metric and check how it works on platform.

namanv3:
@srikanthccv this validation can't go in unmarshal cuz we don't have type info there so I've added it after we get it. 
I tested out summary metrics now, and yes it doesn't make sense for percentile to be allowed there. Earlier I had just copied what was in the existing method, which was wrong actually, so I have fixed that too.

srikanthccv (approved):

PR #11394, opened 2026-05-21
feat: trace based filters for logs, supporting aggregations as well

### LINKED ISSUE #8965: trace-based time range filtering for log queries
Currently, when searching for spans related to a specific trace, the system automatically adjusts the query's time range to match the trace's start and end timestamps. This same optimisation could be applied to log queries with trace_id in filters? Would implementing this trace-based time range filtering for logs provide performance benefits or other improvements?

nityanandagohain:
## Pull Request

---

### 📄 Summary
* If there existing a entry in the trace summary table then the corresponding timerange should be used.
* For traces if the entry is not present in trace summary it means trace doesn't exist but not the case for logs. We can only modify if found
* Changes applied to aggreagtion queries as well for logs and traces. Previously it was just present for trace list query

#### Issues closed by this PR
Fixes https://github.com/SigNoz/signoz/issues/8965

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

- Tests added/updated:✅ 
- Manual verification:✅ 
- Edge cases covered:

---

- Blast radius: Logs/List and aggregation queries with trace_id in filter.
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested


pkg/querier/builder_query.go
```diff
@@ -215,6 +224,81 @@ func (q *builderQuery[T]) Execute(ctx context.Context) (*qbtypes.Result, error)
...
+
+	traceStartMS := uint64(traceStart) / 1_000_000
+	traceEndMS := uint64(traceEnd) / 1_000_000
+	if traceStartMS == 0 || traceEndMS == 0 {
+		return fromMS, toMS, true
+	}
+
+	if traceStartMS > toMS || traceEndMS < fromMS {
+		return fromMS, toMS, false
+	}
```
tushar-signoz, line 284: Comparison signs should be flipped right ?

tushar-signoz (commented):
Haven't reviewed yet, just a quick glace

pkg/querier/builder_query.go
nityanandagohain: This means that the query range is not overlapping with the timerange from tracesummary. Thus overlap=false will make sure that we return empty result.

This logic remains same as previouly we had.

pkg/querier/builder_query.go
```diff
@@ -215,6 +224,81 @@ func (q *builderQuery[T]) Execute(ctx context.Context) (*qbtypes.Result, error)
...
+	if !ok {
+		if q.spec.Signal == telemetrytypes.SignalTraces {
+			q.logger.DebugContext(ctx, "trace_id not found in trace_summary; short-circuiting traces query to empty",
+				slog.Any("trace_ids", traceIDs))
+			return fromMS, toMS, false
+		}
+		q.logger.DebugContext(ctx, "trace_id not found in trace_summary; leaving time range untouched for logs",
+			slog.Any("trace_ids", traceIDs))
+		return fromMS, toMS, true
+	}
```
srikanthccv: `GetTraceTimeRangeMulti` return `ok=false` on error. The prior code defaulted to full time range when `!ok`

pkg/querier/builder_query.go
```diff
@@ -310,42 +394,22 @@ func (q *builderQuery[T]) executeWindowList(ctx context.Context) (*qbtypes.Resul
...
+		if !overlap {
+			return &qbtypes.Result{
+				Type: qbtypes.RequestTypeRaw,
+				Value: &qbtypes.RawData{
+					QueryName: q.spec.Name,
+				},
+				Stats: qbtypes.ExecStats{
+					DurationMS: uint64(time.Since(start).Milliseconds()),
+				},
+			}, nil
```
srikanthccv: this is how it shows up in product when trace falls outside the selected time range

Image

- misleading (all the points), logs are indeed linked; being sent; and associated with trace
- not a good user experience

Distinguish truly no associated logs from the outside time range and inform user accordingly.

nityanandagohain: Ok so the current state of the product right now is that if we search for a trace_id and it is not found then this is the page it shows by default. For the scope of this PR let me add a warning that trace is outside outside the the range of the query. And then on the frontend we can add support for the warning.

nityanandagohain: Image

This is what is looks now. Will get the frontend change as well if this looks good

pkg/querier/builder_query.go
nityanandagohain: Thanks have udated GetTraceTimeRangeMulti to separate between error and not present.

pkg/querier/builder_query.go
srikanthccv: This is better than nothing, please create a ticket on engineering pod to add the change on frontend as well.

pkg/telemetrytraces/trace_time_range.go
```diff
@@ -21,19 +21,19 @@ func NewTraceTimeRangeFinder(telemetryStore telemetrystore.TelemetryStore) *Trac
 	}
 }
 
-func (f *TraceTimeRangeFinder) GetTraceTimeRange(ctx context.Context, traceID string) (startNano, endNano int64, ok bool) {
+func (f *TraceTimeRangeFinder) GetTraceTimeRange(ctx context.Context, traceID string) (startNano, endNano int64, isPresent bool, error error) {
```
srikanthccv:
```suggestion
func (f *TraceTimeRangeFinder) GetTraceTimeRange(ctx context.Context, traceID string) (startNano, endNano int64, exists bool, error error) {
```

pkg/telemetrytraces/trace_time_range.go
```diff
@@ -21,19 +21,19 @@ func NewTraceTimeRangeFinder(telemetryStore telemetrystore.TelemetryStore) *Trac
...
 	traceIDs := []string{traceID}
 	return f.GetTraceTimeRangeMulti(ctx, traceIDs)
 }
 
-func (f *TraceTimeRangeFinder) GetTraceTimeRangeMulti(ctx context.Context, traceIDs []string) (startNano, endNano int64, ok bool) {
+func (f *TraceTimeRangeFinder) GetTraceTimeRangeMulti(ctx context.Context, traceIDs []string) (startNano, endNano int64, isPresent bool, error error) {
```
srikanthccv:
```suggestion
func (f *TraceTimeRangeFinder) GetTraceTimeRangeMulti(ctx context.Context, traceIDs []string) (startNano, endNano int64, exists bool, error error) {
```

pkg/querier/builder_query.go
```diff
@@ -215,6 +231,88 @@ func (q *builderQuery[T]) Execute(ctx context.Context) (*qbtypes.Result, error)
...
+	if !found || len(traceIDs) == 0 {
+		return fromMS, toMS, true, ""
+	}
+
+	finder := telemetrytraces.NewTraceTimeRangeFinder(q.telemetryStore)
+	traceStart, traceEnd, isPresent, err := finder.GetTraceTimeRangeMulti(ctx, traceIDs)
+	if err != nil {
+		return fromMS, toMS, true, ""
+	}
+	if !isPresent {
```
srikanthccv:
```suggestion
	traceStart, traceEnd, exists, err := finder.GetTraceTimeRangeMulti(ctx, traceIDs)
	if err != nil {
		return fromMS, toMS, true, ""
	}
	if !exists {
```

tests/integration/tests/querier/01_logs.py
```diff
@@ -2293,3 +2294,347 @@ def extract_services_and_values(query_name: str) -> tuple[list, list]:
...
+        ]
+    )
+
+    # Insert logs:
+    # - one with the target trace_id, at a timestamp within the trace's
+    #   recorded window (now-10s..now-5s, padded ±1s).
+    # - one with a different trace_id; must never appear in target_trace_id
+    #   results.
+    # - one with an orphan trace_id whose trace was never ingested — used to
+    #   verify the lookup miss does NOT short-circuit logs queries.
```
srikanthccv: `other_{trace_id,span_id}` is not ingested at all, so there is no different b/w the `other_trace_id` and `orphan_trace_id` here.

tests/integration/tests/querier/01_logs.py
```diff
@@ -2293,3 +2294,347 @@ def extract_services_and_values(query_name: str) -> tuple[list, list]:
...
+    # --- Test 1: narrow window (single bucket, <1 h) ---
+    narrow_start_ms = int((now - timedelta(minutes=5)).timestamp() * 1000)
+    narrow_rows, narrow_warnings = _query(narrow_start_ms, now_ms, target_trace_id)
+
+    assert len(narrow_rows) == 1, f"Expected 1 log for trace_id filter (narrow window), got {len(narrow_rows)}"
+    assert narrow_rows[0]["data"]["trace_id"] == target_trace_id
+    assert narrow_rows[0]["data"]["span_id"] == target_root_span_id
+    assert not any(outside_range_msg in m for m in narrow_warnings), f"Did not expect outside-range warning, got {narrow_warnings}"
+
+    # --- Test 2: wide window (>1 h, camp to the timerange from trace_summary) ---
```
srikanthccv:
```suggestion
    # --- Test 2: wide window (>1 h, clamp to the timerange from trace_summary) ---
```

tests/integration/tests/querier/01_logs.py
```diff
@@ -2293,3 +2294,347 @@ def extract_services_and_values(query_name: str) -> tuple[list, list]:
...
+    )
+
+    # Two logs for the target trace_id, both inside the recorded trace window.
+    # One additional log carries an orphan trace_id with no row in
+    # trace_summary — used to verify that the lookup miss does not
+    # short-circuit logs aggregations.
+    insert_logs(
+        [
+            Logs(
+                timestamp=now - timedelta(seconds=9),
```
srikanthccv: this is going to be flaky, please use some more tolerable. may be 7 or something
```suggestion
                timestamp=now - timedelta(seconds=7),
```

srikanthccv (commented):
LGTM, some nits

srikanthccv:
@nityanandagohain , can you cross check the `emptyResultFor` kind is indeed how the empty result is returned if it were truly empty result from regular queries.

nityanandagohain:
Yeah nothing breaks with `emptyResultFor` @srikanthccv

srikanthccv (approved):

therealpandey (approved):

PR #11395, opened 2026-05-21
fix(querybuilder): return PreparedWhereClause by value so warnings propagate when clause is empty

tushar-signoz:
## Summary

`PrepareWhereClause` previously returned `*PreparedWhereClause` and used `nil` as the sentinel for "no WHERE clause to emit" (empty condition or `SkipConditionLiteral`). In that path the visitor's collected `warnings` / `mainWarnURL` were discarded — callers had no way to surface them because the pointer itself was `nil`.

This PR changes the return type to the value type `PreparedWhereClause`:

- An empty clause is now represented by `PreparedWhereClause{WhereClause: nil, Warnings: ..., WarningsDocURL: ...}` so warnings still flow through to the resulting `Statement`.
- Adds `PreparedWhereClause.IsEmpty()` for callers that need to branch on "should I attach the WHERE clause to the builder?".
- Updates every caller in `pkg/telemetrytraces`, `pkg/telemetrylogs`, `pkg/telemetrymetrics`, `pkg/telemetrymeter`, `pkg/telemetryaudit`, `pkg/telemetryresourcefilter`, `pkg/telemetrymetadata`, `pkg/querybuilder`, and the `modules/**` consumers to use the value type and `IsEmpty()` instead of nil checks.

Net effect: warnings produced while parsing a filter expression are no longer silently dropped when the expression resolves to an empty/skipped clause.

### Change Type

- [x] Bug fix
- [x] Refactor

### Bug Context

**Root cause:** The pointer-returning signature couples "no clause" with "no result", so the warnings collected on the visitor were unreachable in the empty-clause path.

**Fix strategy:** Always return a populated value; let callers check `IsEmpty()` for the "skip attaching WHERE" decision while still reading `Warnings` / `WarningsDocURL` from the same struct.

### Testing Strategy

- Updated `where_clause_visitor_test.go` to match the new return type.
- Existing telemetry builder unit tests cover the warning propagation paths.

### Risk & Impact

- **Blast radius:** Internal API; all in-tree callers updated in the same PR.
- **Regressions:** None expected — behavior change is limited to surfacing warnings that were previously dropped.
- **Rollback:** Revert the PR; signature change is mechanical.


nityanandagohain (dismissed):

srikanthccv:
Changes LGTM @tushar-signoz. could you add an integration test before we merge? Without one, the expected behavior could regress in a future refactor.

tushar-signoz:
@srikanthccv 
Added integration test

srikanthccv (approved):

therealpandey (approved):

PR #11397, opened 2026-05-21
fix(infra-monitoring): align v2 custom queries' bounds with QBv5 querier step adjustment

nikhilmantri0902:
## Pull Request

  ---

  ### 📄 Summary

The v2 inframonitoring endpoints (pods/hosts/nodes) build raw ClickHouse queries against `time_series_v4_*` and `samples_v4_*`. The QB v5 metric querier step-floors its window before SQL generation (required so aggregation buckets that overlap the request window are included); the module's hand-built queries used raw `req.Start`/`req.End`. The two paths therefore looked at slightly different pod/host/node sets — 

Adds a package-private `alignedMetricWindow` helper in `helpers.go` that returns step-floored bounds + the TS/samples table names, and routes all 5 raw-query call sites through it.

#### Issues closed by this PR
  
Ref: SigNoz/engineering-pod#5048

  ---

  ### ✅ Change Type

  - [x] 🐛 Bug fix
  - [x] ♻️ Refactor

  ---

  ### 🐛 Bug Context

  #### Root Cause
  QB v5 querier step-floors `req.Start`/`req.End` via `querybuilder.AdjustedMetricTimeRange`. The `pkg/modules/inframonitoring/implinframonitoring` package built its own CH queries and passed raw bounds to
  `WhichTSTableToUse` / `WhichSamplesTableToUse`, so the samples-side WHERE clause looked at a slightly narrower window than the querier did. 

  #### Fix Strategy
  One helper, `alignedMetricWindow(startMs, endMs)`, encapsulates step-floor + table selection. Five call sites updated to use it (`getMetadata`, `buildSamplesTblFingerprintSubQuery`, `getPerGroupPodPhaseCounts`, `getPerGroupNodeConditionCounts`, `getPerGroupHostStatusCounts`). 

  ---

  ### 🧪 Testing Strategy

  - **Manual verification**: Diff'd the querier path (CTEs) and metadata path (subquery) in one ClickHouse query — `NOT IN` set went from N rows to 0 after the change. Re-ran the repro curl with `offset > resp.Total`; response now returns `records: []` as expected, and `resp.Total` matches `totalAll` in DEBUG output.
  - **Tests**: existing module tests pass (`go test ./pkg/modules/inframonitoring/...`).

  ---

  - **Blast radius**: v2 `/api/v2/infra_monitoring/{pods,hosts,nodes,...}` and the per-group phase/condition/status count sub-queries.
  - **Potential regressions**: `metadataMap` is marginally wider since it now covers the same set the ranking sees — groups that previously had `-1` placeholders may now show full metadata. No effect on aggregated values; the floor change is a few seconds at the boundaries.
 - **Rollback**: single-PR revert.

  ---

  ### 📝 Changelog

  | Field | Value |
  |------|-------|
  | Change Type | Bug Fix |
  | Description | Fix pagination boundary in v2 infra monitoring list endpoints — `offset >= total` no longer returns records. |

  ---

  ### 📋 Checklist

  - [x] Tests added or explicitly not required (existing coverage exercises the touched paths)
  - [x] Manually tested

  ---

  ## 👀 Notes for Reviewers

  - `alignedMetricWindow` is intentionally package-private — it captures this module's contract with the metric querier. If a third caller outside this package ever needs it, lift to `pkg/querybuilder` as `StepFlooredMetricTimeRange`.
  - `buildSamplesTblFingerprintSubQuery` signature changed: it now takes the resolved samples table name + floored bounds instead of resolving internally. Only two callers; both updated.
  - Operator parity vs. querier verified — TS uses `LE`, samples uses `L`, both use `GE` on the lower bound. Matches `pkg/telemetrymetrics/statement_builder.go:301-302, :248-249`.


pkg/modules/inframonitoring/implinframonitoring/helpers.go
```diff
@@ -140,6 +140,22 @@ func parseAndSortGroups(
 	return groups
 }
 
+// filterMetricGroupsByMetadata drops any ranked group whose compositeKey is
+// not present in metadataMap. Keeps metricGroups a strict subset of the
+// metadatamap ( which ideally should be the case anyway, but this is a sanity check)
```
namanv3: shouldn't the step interval adjustment mean that this is not required anymore?

pkg/modules/inframonitoring/implinframonitoring/helpers.go
```diff
@@ -307,23 +326,57 @@ func parseFullQueryResponse(
...
-		metrictypes.UnspecifiedType,
-		metrictypes.TimeAggregationUnspecified,
-		nil,
+// alignedMetricWindow returns step-floored time bounds and the metric tables
+// to use for the given window. The floor matches what the QB v5 metric
+// querier does internally (see querybuilder.AdjustedMetricTimeRange).
+// Please use the flooredStartMs with samples table and tsAdjustedStartMs with ts tables.
+// Both can use the same flooredEndMs.
+func alignedMetricWindow(startMs, endMs int64) (
+	flooredStartMS uint64,
```
namanv3: call it samplesStartMs maybe

pkg/modules/inframonitoring/implinframonitoring/helpers.go
nikhilmantri0902: Thats right, can we keep it as a sanity check though?

nikhilmantri0902: although, i think u r right, removing.

nikhilmantri0902: removed

pkg/modules/inframonitoring/implinframonitoring/helpers.go
nikhilmantri0902: this was also done.

pkg/modules/inframonitoring/implinframonitoring/helpers.go
```diff
@@ -307,23 +308,57 @@ func parseFullQueryResponse(
...
+	distributedTSTable string,
+	localTSTable string,
+	distributedSamplesTable string,
+	localSamplesTable string,
+) {
+	samplesAdjustedStartMs = uint64(startMs)
+	flooredEndMs = uint64(endMs)
+	stepSecs := querybuilder.RecommendedStepIntervalForMetric(samplesAdjustedStartMs, flooredEndMs)
+	// note: this is the same flooring logic as in querybuilder.AdjustedMetricTimeRange
+	if stepSecs > 0 {
```
nikhilmantri0902, line 334: @srikanthccv check this block, this code was adopted from [querybuilder.AdjustedTimeRange](https://github.com/SigNoz/signoz/blob/27dc996fd860ec8f727f18dd369c60356f59c32e/pkg/querybuilder/time.go#L157) where start and end are adjusted using stepInterval. 

I could have extracted out start and end adjustment code into a seperate function in querybuilder itself and then used that here, but was not sure to touch querybuilder layer hence duplicated the code. 

This adjustment is needed for custom queries that use start and end times and ts and samples tables to look at observations in same time range as querier sees them. 

Thoughts?

namanv3 (dismissed):

nikhilmantri0902:
Found a bug specifically for pods in integration tests, fixed that in the last commit @namanv3

pkg/modules/inframonitoring/implinframonitoring/pods.go
```diff
@@ -95,14 +95,16 @@ func buildPodRecords(
...
 
-		if attrs, ok := metadataMap[compositeKey]; ok && isPodUIDInGroupBy {
-			// the condition above ensures we deduce age only if pod uid is in group by because if
-			// it's not in group by then we might have multiple pod uids in the same group and hence then podAge wont make sense
-			if startTimeStr, exists := attrs[podStartTimeAttrKey]; exists && startTimeStr != "" {
-				if t, err := time.Parse(time.RFC3339, startTimeStr); err == nil {
-					startTimeMs := t.UnixMilli()
-					if startTimeMs > 0 {
-						record.PodAge = reqEnd - startTimeMs
+		if attrs, ok := metadataMap[compositeKey]; ok {
```
namanv3, line 98: wait what really changed here? what was the bug?

nikhilmantri0902: The isPodUIDInGroupBy is now inside this if block and is applicable only for pod age, not for the population of other attributes in meta map. Earlier, no attrs were getting returned in metamap in case of custom group by, @namanv3

namanv3 (dismissed):

pkg/modules/inframonitoring/implinframonitoring/helpers.go
```diff
@@ -307,23 +312,58 @@ func parseFullQueryResponse(
...
 	)
-	localSamplesTable := strings.TrimPrefix(samplesTableName, "distributed_")
+
+	distributedSamplesTable = telemetrymetrics.WhichSamplesTableToUse(
+		samplesAdjustedStartMs, flooredEndMs,
+		metrictypes.UnspecifiedType, metrictypes.TimeAggregationUnspecified, nil,
+	)
+	localSamplesTable = strings.TrimPrefix(distributedSamplesTable, "distributed_")
+
+	return
```
srikanthccv: no naked returns

pkg/modules/inframonitoring/implinframonitoring/helpers.go
```diff
@@ -307,23 +312,58 @@ func parseFullQueryResponse(
...
+	tsAdjustedStartMs, _, distributedTSTable, localTSTable = telemetrymetrics.WhichTSTableToUse(
+		samplesAdjustedStartMs, flooredEndMs, nil,
 	)
-	localSamplesTable := strings.TrimPrefix(samplesTableName, "distributed_")
+
+	distributedSamplesTable = telemetrymetrics.WhichSamplesTableToUse(
+		samplesAdjustedStartMs, flooredEndMs,
+		metrictypes.UnspecifiedType, metrictypes.TimeAggregationUnspecified, nil,
+	)
+	localSamplesTable = strings.TrimPrefix(distributedSamplesTable, "distributed_")
```
srikanthccv: don't strip like this. return from the `WhichSamplesTableToUse` to also return the local table and use it.

pkg/modules/inframonitoring/implinframonitoring/helpers.go
nikhilmantri0902: this was done.

pkg/modules/inframonitoring/implinframonitoring/helpers.go
nikhilmantri0902: this was done as well.

pkg/telemetrymetrics/tables.go
```diff
@@ -135,44 +135,56 @@ func ValueColumnForSamplesTable(tableName string) string {
...
+		case SamplesV4Agg30mTableName:
+			return SamplesV4Agg30mTableName, SamplesV4Agg30mLocalTableName
+		case ExpHistogramTableName:
+			return ExpHistogramTableName, ExpHistogramLocalTableName
 		}
+		return tableHints.SamplesTableName, tableHints.SamplesTableName
 	}
 
 	// we don't have any aggregated table for sketches (yet)
 	if metricType == metrictypes.ExpHistogramType {
```
nikhilmantri0902, line 172: @srikanthccv , please confirm this line of change. We are now returning distributed and local version of exp_hist tables in that order. Earlier as you can see for this table, local table was being returned and for all other distributed ... was there a reason ?

srikanthccv: no, it was a bug. it should have been distributed table.

nikhilmantri0902: okay, thanks for the confirmation.

srikanthccv (approved):

PR #11399, opened 2026-05-21
fix(date-time-selection-v2): out of sync query params

H4ad:
## Pull Request

---

This PR fixes a bug that was cleaning any query param after changing the time selector.

https://github.com/user-attachments/assets/5d6dddb4-cb7c-4352-b7a8-b488aa78b07f

Fixes https://github.com/SigNoz/engineering-pod/issues/4827

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

The problem with the de-sync between our component and nuqs is caused because internally the data is being stored differently, nuqs has their store and react-router-dom has another one.

We are using the wrong adapter for nuqs, instead of https://nuqs.dev/docs/adapters#react-router-v6 we are using the https://nuqs.dev/docs/adapters#react-spa because our version of react router is too old.

Because the stores are differently, when we updated the URL via nuqs, the `useLocation` hook inside the `useUrlQuery` was not updated.

Instead of relying on `useUrlQuery`, I just created a specific function to parse the current URL params to always use the fresh search params after the re-render.

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Date Time Selector V2
- Potential regressions: Any bug related to time selection or out of sync due to missing re-render.

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | We fixed the bug that causing the filters to be cleaned after changing the date time selection on Infrastructure Monitoring. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

Some alternative fixes for this bug are:
- Using nuqs: I didn't want to make this PR huge in terms of coding change, and this could introduce more problems due to the other usages of `useUrlQuery`.
- Update `useUrlQuery` to detect URL changes: this also is problematic because of the amount of places this hook is being used, we will need to test too many places. Also, this causes some double write on URL due to Nuqs automatic batching, I tested but didn't like the final implementation.


frontend/src/container/TopNav/DateTimeSelectionV2/index.tsx
```diff
@@ -90,10 +90,12 @@ function DateTimeSelection({
 	const [hasSelectedTimeError, setHasSelectedTimeError] = useState(false);
 	const [isOpen, setIsOpen] = useState<boolean>(false);
 
-	const urlQuery = useUrlQuery();
-	const searchStartTime = urlQuery.get('startTime');
-	const searchEndTime = urlQuery.get('endTime');
-	const relativeTimeFromUrl = urlQuery.get(QueryParams.relativeTime);
+	const currentSearchParams = getCurrentSearchParams();
+	const searchStartTime = currentSearchParams.get('startTime');
```
AshwinBhatkal: nit; use QuryParams here and below

frontend/src/container/TopNav/DateTimeSelectionV2/utils/getCurrentSearchParams.ts
```diff
@@ -0,0 +1,17 @@
+type SearchParamsGetter = () => URLSearchParams;
+
+let getter: SearchParamsGetter = (): URLSearchParams =>
```
AshwinBhatkal: Can you please add a TODO as to why this file is currently required and when we upgrade our react-router, we can replace with nuqs?

AshwinBhatkal: Also, do you know of any there other places where we could have this issue with nuqs?

AshwinBhatkal (commented):
minor comments. looks good otherwise

srikanthccv:
Is this a regression introduced in https://github.com/SigNoz/signoz/pull/10780?

frontend/src/container/TopNav/DateTimeSelectionV2/index.tsx
H4ad: `currentSearchParams` is always mutated after re-render since it parses directly the URL and generates another reference of URLSearchParams, it should not be used in use effects/callbacks.

H4ad:
@srikanthccv at https://github.com/SigNoz/signoz/pull/10640

frontend/src/container/TopNav/DateTimeSelectionV2/utils/getCurrentSearchParams.ts
H4ad: Mostly the usages are on infra monitoring + settings

To have this issue, we two components, one using nuqs and the other using useUrlQuery, I don't think this is common.

frontend/src/container/TopNav/DateTimeSelectionV2/index.tsx
AshwinBhatkal: I meant the constants. Like it is used below

AshwinBhatkal (approved):

PR #11411, opened 2026-05-22
chore: added changes to migrate slider component from antd to signozhq/ui

ahrefabhi:
## Pull Request

---

### 📄 Summary
Migrates remaining `Slider` usages from `antd` to `@signozhq/ui/slider` so the
codebase aligns with the in-house design system and removes one more antd
dependency from the surface area we still control.

Touched call sites:
- `OnboardingQuestionaire/OptimiseSignozNeeds` — Logs/Hosts/Services sliders
- `Trace/Filters/Panel/PanelBody/Duration` — duration range slider
- `pages/TracesExplorer/Filter/DurationSection` — duration range slider

Notable adjustments made during the swap:
- `styles.track.background` → `styles.range.backgroundColor` (new API on the
  signoz Slider).
- `onChange` callbacks now receive `number | number[]`; range handlers cast
  accordingly instead of relying on antd's `SliderRangeProps`.
- `TipComponent` no longer needs the `undefined` guard since the new Slider
  always passes a `number`.
- Removed bespoke `.ant-slider-mark` overrides in
  `OnboardingQuestionaire.styles.scss` — marks come styled out of the box now.
- Added small `.duration-input-slider` spacing in
  `QuickFilters/FilterRenderers/Duration/Duration.styles.scss` to preserve the
  vertical rhythm previously contributed by antd's defaults.

_Before/after for the three screens — Onboarding questionnaire sliders, Trace
filter duration slider, and Traces Explorer duration slider — to be attached._

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/4460

---

### ✅ Change Type
- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context
N/A

---

### 🧪 Testing Strategy
- Tests added/updated: none — pure component swap, behaviour preserved.
- Manual verification:
  - Onboarding questionnaire: drag Logs/Hosts/Services sliders, confirm
    tooltip values, marks, and the blue range fill render correctly and the
    submitted values match prior behaviour.
  - Trace filter panel: open the Duration panel, drag both handles, confirm
    the URL updates only on `onAfterChange` and tooltip shows `<n>ms`.
  - Traces Explorer filter: same range/tooltip checks, plus the new
    `.duration-input-slider` spacing visually matches the prior layout.
- Edge cases covered: min/max boundaries, default `[0, 0]` initial state in
  the trace duration filter, and tooltip rendering when only one handle moves.

---

- Blast radius: frontend only — onboarding flow and trace duration filters.
- Potential regressions: tooltip formatting, marks alignment, or range-fill
  colour on the three migrated screens; type narrowing of the new
  `number | number[]` onChange signature at the affected call sites.
- Rollback plan: revert this PR; no schema or API changes involved.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | N/A (internal component migration, no user-visible behaviour change) |

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers
- Worth double-checking the `range.backgroundColor` style maps to the same
  visual as the old `track.background` on the signoz Slider.
- The `value as number[]` / `value as number` casts in the range handlers are
  intentional — the new Slider's union type covers both single and range
  modes via the same callback signature.
- Removed `.ant-slider-mark` SCSS is safe to drop: the new Slider styles
  marks itself, and no other consumer in this file relied on those rules.


frontend/src/container/Trace/Filters/Panel/PanelBody/Duration/index.tsx
```diff
@@ -169,16 +168,15 @@ function Duration(): JSX.Element {
...
-	const TipComponent = useCallback((value: undefined | number) => {
-		if (value === undefined) {
-			return <div />;
-		}
-		return <div>{`${value?.toString()}ms`}</div>;
-	}, []);
+	const TipComponent = useCallback(
+		(value: number) => <div>{`${value.toString()}ms`}</div>,
+		[],
+	);
```
H4ad, line 179: If value can be undefined, this will throw an error, right?

ahrefabhi: value cannot be undefined.

H4ad (approved):

PR #11413, opened 2026-05-22
fix(rules): use alertmanager external URL for related logs/traces and generator URL

jatinderjit:
### 📄 Summary

The host portion of `related_logs` / `related_traces` annotations and the rule generator URL (used as the link-back in notifications) was previously derived by parsing `window.location` captured from the frontend at rule-creation time. This happens to be blank for API or MCP users.

This PR switches the source-of-truth to `SIGNOZ_ALERTMANAGER_SIGNOZ_EXTERNAL__URL`, which is the canonical URL for each tenant.

Plumbed via `signoz.go` → `signozruler.NewFactory` → `rules.ManagerOptions.ExternalURL` → `WithExternalURL` option on `BaseRule`. Covers community + enterprise prepare-task and test-notification paths.

The work is split into two commits so the fallback can be reinstated by reverting the second commit if needed:

#### Issues closed by this PR

- Closes https://github.com/SigNoz/engineering-pod/issues/5055

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause

`PrepareRuleGeneratorURL` and `ThresholdRule.hostFromSource()` parsed `PostableRule.Source`, which is set to `window.location` by the frontend when the rule is created. This could be blank for API or MCP users.

#### Fix Strategy

Use the `alertmanager.signoz.external_url` config (env var `SIGNOZ_ALERTMANAGER_SIGNOZ_EXTERNAL__URL`) as the host. This is the canonical "where SigNoz lives" URL elsewhere in the alertmanager (template `.ExternalURL`, email links, etc.), so the same value now drives related-logs/traces links and generator URLs.

---

### 🧪 Testing Strategy

- **Tests added/updated:**
  - `TestBaseRule_ExternalURLHost`: covers nil URL, default URL, configured URL, port, trailing slash.
  - `TestBaseRule_GeneratorURL`: verifies generator URL composition for configured and default external URL.
- **Manual verification:** Tested these scenarios for both enterprise and community editions:
  - Metric based alert -  query builder
  - Metric based alert - PromQL
  - Log based alert
  - Trace based alert
- **Edge cases covered:** nil URL, default `http://localhost:8080`, trailing slash, host with port, https host.

---

- **Blast radius:** All alert notification channels. The host portion of the `related_logs`, `related_traces`, and `RuleSource` URL changes.
- **Potential regressions:** Self-hosted users who have not set `SIGNOZ_ALERTMANAGER_SIGNOZ_EXTERNAL__URL` will see `http://localhost:8080` in their alert links. This is intentional, it surfaces the need to configure the env var.
- **Rollback plan:** Revert the PR.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | OSS / Enterprise (self-hosted impact); Cloud unaffected |
| Change Type | Bug Fix |
| Description | Alert notification links (`related_logs`, `related_traces`, rule edit link) are now built from `SIGNOZ_ALERTMANAGER_SIGNOZ_EXTERNAL__URL` instead of the rule author's browser URL. Self-hosted deployments should ensure this env var is set to their public SigNoz URL. |

---

### 📋 Checklist
- [x] Manually tested
- [x] Breaking changes documented (changelog entry above)

---

## 👀 Notes for Reviewers

- Review commit-by-commit. The first commit is the additive change with the fallback (host from source) intact; the second is the removal.
- The `Source` field on `PostableRule` is left in place for API compatibility but is no longer consumed when building URLs.
- `ruletypes.PrepareRuleGeneratorURL` is removed (was only called from `BaseRule.GeneratorURL`).


pkg/query-service/rules/base_rule.go
```diff
@@ -138,6 +140,17 @@ func WithRuleStateHistoryModule(module rulestatehistory.Module) RuleOption {
 	}
 }
 
+// WithExternalURL injects the alertmanager external URL
+func WithExternalURL(externalURL *url.URL) RuleOption {
+	return func(r *BaseRule) {
+		if externalURL == nil {
+			r.externalURL = ""
+		} else {
```
srikanthccv: this is redundant

pkg/query-service/rules/base_rule.go
```diff
@@ -241,7 +253,14 @@ func (r *BaseRule) Annotations() ruletypes.Labels       { return r.annotations }
 func (r *BaseRule) PreferredChannels() []string         { return r.preferredChannels }
 
 func (r *BaseRule) GeneratorURL() string {
-	return ruletypes.PrepareRuleGeneratorURL(r.ID(), r.source)
+	return fmt.Sprintf("%s/alerts/edit?ruleId=%s", r.ExternalURLHost(), r.ID())
```
srikanthccv:
```suggestion
	return fmt.Sprintf("%s/alerts/overview?ruleId=%s", r.ExternalURLHost(), r.ID())
```

srikanthccv (dismissed):

cmd/community/server.go
```diff
@@ -118,8 +119,8 @@ func runServer(ctx context.Context, config signoz.Config, logger *slog.Logger) e
 		func(_ sqlstore.SQLStore, _ dashboard.Module, _ global.Global, _ zeus.Zeus, _ gateway.Gateway, _ licensing.Licensing, _ serviceaccount.Module, _ cloudintegration.Config) (cloudintegration.Module, error) {
 			return implcloudintegration.NewModule(), nil
 		},
-		func(c cache.Cache, am alertmanager.Alertmanager, ss sqlstore.SQLStore, ts telemetrystore.TelemetryStore, ms telemetrytypes.MetadataStore, p prometheus.Prometheus, og organization.Getter, rsh rulestatehistory.Module, q querier.Querier, qp queryparser.QueryParser) factory.NamedMap[factory.ProviderFactory[ruler.Ruler, ruler.Config]] {
-			return factory.MustNewNamedMap(signozruler.NewFactory(c, am, ss, ts, ms, p, og, rsh, q, qp, nil, nil))
+		func(c cache.Cache, am alertmanager.Alertmanager, ss sqlstore.SQLStore, ts telemetrystore.TelemetryStore, ms telemetrytypes.MetadataStore, p prometheus.Prometheus, og organization.Getter, rsh rulestatehistory.Module, q querier.Querier, qp queryparser.QueryParser, externalURL *url.URL) factory.NamedMap[factory.ProviderFactory[ruler.Ruler, ruler.Config]] {
```
therealpandey: @jatinderjit for injecting the config, can we expose a config function in alertmanager.Alertmanager and use the config from there?

pkg/query-service/rules/base_rule.go
```diff
@@ -241,7 +251,14 @@ func (r *BaseRule) Annotations() ruletypes.Labels       { return r.annotations }
 func (r *BaseRule) PreferredChannels() []string         { return r.preferredChannels }
 
 func (r *BaseRule) GeneratorURL() string {
-	return ruletypes.PrepareRuleGeneratorURL(r.ID(), r.source)
+	return fmt.Sprintf("%s/alerts/overview?ruleId=%s", r.ExternalURLHost(), r.ID())
```
therealpandey: Please use am.config().ExternalURL.Join here

therealpandey (changes requested):

cmd/community/server.go
jatinderjit: Done in 829ecd01a — added `Config() Config` to the `alertmanager.Alertmanager` interface and consumed `alertmanager.Config().Signoz.ExternalURL` directly inside `signozruler.NewFactory`. The `*url.URL` parameter is no longer threaded through `cmd/community/server.go`, `cmd/enterprise/server.go`, or `pkg/signoz/signoz.go`.

pkg/query-service/rules/base_rule.go
jatinderjit: Done in 829ecd01a — `GeneratorURL` now uses `*url.URL.JoinPath("alerts", "overview")` + `Query().Set("ruleId", ...)` so the URL is composed correctly when the external URL carries a base path (e.g. `https://signoz.example.com/signoz`) and `ruleId` is properly query-encoded. Added a test case for the base-path scenario.

therealpandey (approved):

pkg/contextlinks/links.go
```diff
@@ -312,10 +316,15 @@ func PrepareLinksToLogsV5(start, end time.Time, whereClause string) string {
 	}
 
 	data, _ := json.Marshal(urlData)
-	compositeQuery := url.QueryEscape(url.QueryEscape(string(data)))
+	compositeQuery := url.QueryEscape(string(data))
```
srikanthccv, line 319: one round of escaping is done here and `Encode` does the 2nd escaping, preserving the double escaping needed on frontend.

srikanthccv (approved):

PR #11414, opened 2026-05-22
fix(trace-details): make back button reliably return to previous in-app page

tewarig:
### Steps to recreate 

https://www.loom.com/share/5deaa44e96724cedbf1ddf7d99b6c3fd                         

1. Login into staging 
2. try to add filters  , go open trace/traceid , press back button , everything will work 
3. now open stagin in a new tab.  do step 2 again things will break.

                                                                                                                                                                                                                                                   
  ### What was happening                                                                                                                      
                                                                                                                                              
  The back arrow on the new trace details page was inconsistently sending users to the Traces Explorer even when they had navigated from      
  somewhere else inside SigNoz. The handler tried to detect "did the user come from inside the app?" using `document.referrer` and
  `window.history.length`, but neither signal answers that question reliably in a SPA:                                                        
                                                                                                                                              
  - **`document.referrer` is set once, at document load, and never updates for SPA navigations.** After that, every back-click in the tab gets
   the same answer regardless of where the user actually navigated. So the back button's behavior was determined by how the tab originally    
  arrived at SigNoz, not by the user's real navigation path. It happened to work after a login redirect or a hard refresh (referrer =       
  same-origin), but broke when the user opened SigNoz from Slack, a bookmark, a new tab, or any external link.                                
  - **`window.history.length` counts the entire tab's session history**, including pre-SigNoz entries, and never decreases. It tells you the
  back button is *clickable*, not where it *goes*.                                                                                            
                                                                                                                                            
  ### How this PR fixes it
                                                                                                                                              
  Subscribe to the shared React Router `history` singleton in `lib/history.ts` and count `PUSH` actions (the only action type that adds to the
   back-stack). Expose `hasInAppHistory()` which returns `true` once the SPA router has pushed at least one entry in this page load — i.e.,   
  the precise condition under which `history.goBack()` is guaranteed to land on a SigNoz route.                                             
                                                                                                                                              
  `handlePreviousBtnClick` now reads:         
                                                                                                                                              
  ```ts                                                                                                                                     
  if (hasInAppHistory()) history.goBack();                                                                                                    
  else history.push(ROUTES.TRACES_EXPLORER);

``` 
                                                                                                                                              
🤔  will check the code base and replace with this logic at all places in follow up

### DIFF
--- a/frontend/src/lib/history.ts
+++ b/frontend/src/lib/history.ts
@@ -1,4 +1,15 @@
 import { createBrowserHistory } from 'history';
 import { getBasePath } from 'utils/basePath';
 
-export default createBrowserHistory({ basename: getBasePath() });
+const history = createBrowserHistory({ basename: getBasePath() });
+
+let inAppPushCount = 0;
+history.listen((_, action) => {
+	if (action === 'PUSH') {
+		inAppPushCount += 1;
+	}
+});
+
+export const hasInAppHistory = (): boolean => inAppPushCount > 0;
+
+export default history;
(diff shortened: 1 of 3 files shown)


aks07:
Changes look good. Lets add basic unit test around this

tewarig:
> Changes look good. Lets add basic unit test around this

added

aks07 (approved):

PR #11427, opened 2026-05-22
feat: right dock span details

aks07:
## Pull Request

---

### 📄 Summary

Adds a third **right dock** mode for the Span Details panel on the Trace Details V3 page, alongside the existing **bottom dock** and **floating** modes. The single dock-toggle button in the panel header is replaced by a 3-icon segmented switcher (bottom · right · float), and **right** becomes the default for users without a stored preference.

**Why:** users were finding both existing modes uncomfortable on wide screens. Bottom dock eats vertical space the waterfall should own; floating mode obscures the waterfall and gets in the way of comparing a span with its siblings. Right dock makes use of horizontal space that was previously empty and lets the waterfall keep its natural `flex: 1` vertical sizing.

https://github.com/user-attachments/assets/82867975-3f13-46bc-9221-fea12dd2efb7

#### Issues closed by this PR

Closes SigNoz/engineering-pod#5063

---

### ✅ Change Type

- [x] ✨ Feature

---

### 🧪 Testing Strategy

- Tests added/updated: none — no existing unit/integration/e2e specs cover `SpanDetailsPanel`, `ResizableBox`, or `TraceDetailsV3/index.tsx`, and the `ResizableBox` change is purely additive (`handlePosition` defaults to `'end'`, preserving prior behaviour).
- Manual verification:
  - First-run default (cleared `TRACE_DETAILS_SPAN_DETAILS_POSITION` in localStorage → Span Details opens on the right).
  - Switcher toggles between bottom / right / float; active mode is visually highlighted; tooltips read "Dock at the bottom" / "Dock on the right" / "Open as floating panel".
  - Right-dock left-edge drag clamps to 480–720; waterfall column to the left grows / shrinks inversely. Width is session-only — resets to 480 on reload (intentional).
  - In right mode the waterfall reclaims `flex: 1`. In bottom mode it stays `flex: none` as before.
  - Bottom-dock vertical resize and float-dock drag/close are unchanged.
  - `tsc --noEmit` clean; `oxlint` produces no new findings (all warnings on touched files were pre-existing `any` types and exhaustive-deps).
- Edge cases covered:
  - Existing users with a stored `TRACE_DETAILS_SPAN_DETAILS_POSITION` value keep their preference (string values are unchanged — `'docked'` and `'dialog'` still resolve; only a new `'right'` value was added). No migration needed.
  - Closing the panel and clicking a span re-opens it in whichever mode is selected.

---

- **Blast radius:** Trace Details V3 page only. `ResizableBox` is the only shared primitive touched — it gains a new optional prop with a default that preserves current behaviour, so existing call sites are unaffected.
- **Potential regressions:**
  - Layout reflow around `.layoutRow`/`.content` — bottom-dock and float paths were re-tested but warrant a quick review.
  - `SpanDetailVariant.DOCKED` semantic is "bottom dock" (string value `'docked'` kept stable for localStorage continuity). A future cleanup could rename the enum member to `DOCKED_BOTTOM` for readability; out of scope here.
- **Rollback plan:** revert the PR — no data shape changes, no migrations, no API surface touched.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Trace Details: Span Details panel can now be docked on the right. Switch between bottom / right / float from the icon switcher in the panel header. Right dock is the new default. |

---

### 📋 Checklist

- [x] Tests added or explicitly not required _(none required — additive UI change; no existing specs touched)_
- [x] Manually tested
- [x] Breaking changes documented _(none)_
- [x] Backward compatibility considered _(localStorage values preserved; new variant is additive)_

---

## 👀 Notes for Reviewers

- **`SpanDetailVariant.DOCKED` is "bottom dock"** — kept the enum member name and its string value `'docked'` unchanged so existing users' stored preference still resolves. New right-mode is `DOCKED_RIGHT = 'right'`.
- **`ResizableBox.handlePosition`** is a small additive prop. `'start'` puts the resize handle on the *left* edge (for horizontal) or the *top* edge (for vertical), and inverts the delta sign so dragging *away from* the content grows it. Default `'end'` preserves current behaviour for every existing caller.
- **`DockModeSwitcher`** is scoped locally to `SpanDetailsPanel/`, not lifted to periscope. If we add another trace-side icon-segmented control later we can pull it up then.
- **`<span>` wrap inside `TooltipTrigger asChild`** in `DockModeSwitcher` is load-bearing: `TooltipTrigger asChild` clones its child and merges its own `data-state`, which would otherwise clobber `ToggleGroupItem`'s `data-state="on"|"off"` active styling. Same pattern used in `AuthZTooltip`.
- **Right-dock width persistence** is deliberately session-only — resets to 480px on reload. We can move to localStorage later if users ask for it.


H4ad (approved):
Follow-up: start adding tests to cover these changes/features.

PR #11434, opened 2026-05-23
feat: add ripple effect on dark / light mode toggle

tewarig:
so @YounixM  was showing me his side project. where their is  an effect for dark and light effect which goes from left to right. 
i also though of having that. 

but i think this one is better. 

insipired by https://tweakcn.com/ 
https://github.com/user-attachments/assets/b9d66ab5-7644-417d-a497-17b66706ecb5


tewarig:
updated animation - 


https://github.com/user-attachments/assets/cf9f36d1-1595-40f6-bbca-d91b6a7fb8f1

tewarig:
will try to find time for this in sprint

PR #11445, opened 2026-05-25
feat(web): add support for generating web settings types

vikrantgupta25:
### 📄 Summary
- add support for generating web settings types

contributes to: https://github.com/SigNoz/platform-pod/issues/2144

### DIFF
--- a/cmd/genconfig.go
+++ b/cmd/genconfig.go
@@ -4,32 +4,33 @@ import (
 	"encoding/json"
 	"os"
 	"reflect"
+	"strings"
 
 	"github.com/SigNoz/signoz/pkg/web"
 	"github.com/spf13/cobra"
 	"github.com/swaggest/jsonschema-go"
 )
 
-const webSettingsSchemaPath = "docs/settings/web.json"
+const webSettingsSchemaPath = "docs/config/web-settings.json"
 
-func registerGenerateSettings(parentCmd *cobra.Command) {
-	settingsCmd := &cobra.Command{
-		Use:   "settings",
-		Short: "Generate JSON Schema for settings",
+func registerGenerateConfig(parentCmd *cobra.Command) {
+	configCmd := &cobra.Command{
+		Use:   "config",
+		Short: "Generate JSON Schema for config",
 	}
 
-	settingsCmd.AddCommand(&cobra.Command{
-		Use:   "web",
+	configCmd.AddCommand(&cobra.Command{
+		Use:   "web-settings",
 		Short: "Generate JSON Schema for web settings",
 		RunE: func(currCmd *cobra.Command, args []string) error {
-			return runGenerateWebSettings()
+			return generateWebSettings()
 		},
 	})
 
-	parentCmd.AddCommand(settingsCmd)
+	parentCmd.AddCommand(configCmd)
 }
 
-func runGenerateWebSettings() error {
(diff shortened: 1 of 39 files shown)


cmd/genconfig.go
```diff
@@ -0,0 +1,57 @@
...
+	reflector.DefaultOptions = append(reflector.DefaultOptions,
+		jsonschema.InterceptSchema(func(params jsonschema.InterceptSchemaParams) (bool, error) {
+			if params.Value.Kind() == reflect.Struct {
+				params.Schema.AdditionalProperties = &noAdditional
+			}
+			return false, nil
+		}),
+	)
+
+	schema, err := reflector.Reflect(web.Settings{})
```
therealpandey, line 50: In the json reflector, trim Web prefix @vikrantgupta25

cmd/settings.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/SigNoz/signoz/pkg/web"
+	"github.com/spf13/cobra"
+	"github.com/swaggest/jsonschema-go"
+)
+
+const webSettingsSchemaPath = "docs/settings/web.json"
+
+func registerGenerateSettings(parentCmd *cobra.Command) {
+	settingsCmd := &cobra.Command{
+		Use:   "settings",
```
therealpandey: Call it config maybe?

We can use this to generate schemas out of our config struct with web.settings being one of them?

therealpandey (changes requested):

cmd/settings.go
vikrantgupta25: I started with the same but then it requires collecting multiple structs, hosting them together and then constructing one schema out of it.

vikrantgupta25: The idea behind keeping the settings folder was to have config structs as JSON schema here eventually.

vikrantgupta25: > We can use this to generate schemas out of our config struct with web.settings being one of them?

Hard to say right now because the `web.Settings` and the `web.Config` are decoupled and have their own shape . web settings can have derived fields from other config entries as well.

cmd/genconfig.go
vikrantgupta25: done

therealpandey (approved):

frontend/package.json
```diff
@@ -24,7 +24,8 @@
     "commitlint": "commitlint --edit $1",
     "test": "jest",
     "test:changedsince": "jest --changedSince=main --coverage --silent",
-    "generate:api": "orval --config ./orval.config.ts && sh scripts/post-types-generation.sh"
+    "generate:api": "orval --config ./orval.config.ts && sh scripts/post-types-generation.sh",
+    "generate:config:web-settings": "json2ts ../docs/config/web-settings.json -o src/types/generated/webSettings.ts --style.useTabs --style.tabWidth=1 --style.singleQuote --bannerComment '/* AUTO GENERATED FILE - DO NOT EDIT - GENERATED FROM docs/config/web-settings.json */'"
```
AshwinBhatkal, line 28: Add a comment in the file how to regenerate

AshwinBhatkal (approved):

frontend/package.json
vikrantgupta25: Will add in the next PR when adding for schema

PR #11451, opened 2026-05-25
feat(meterreporter): add jitter to meter collection cycles

karanbalani:
## Summary

When we deploy a new SigNoz version, all cloud deployments call Zeus at the same instant. Then every 6 hours after that, they all call again together. This creates large traffic spikes on Zeus that scale with our fleet size.

This PR spreads those calls evenly across time so Zeus sees a manageable rate instead of synchronized bursts.

## How it works

A single `Jitter` value controls the randomness applied to both the first call after startup and every subsequent cycle:

```
First fire = startup + random(0, Jitter)
Next fire  = previous + (Interval − random(0, Jitter))
```

- The first call after startup is delayed by a random amount in `[0, Jitter)` — different on every pod.
- Each subsequent cycle takes `Interval − random(0, Jitter)`, so the time between any two consecutive reports is between `Interval − Jitter` and `Interval`.
- The strict upper bound is still `Interval`, so every deployment is guaranteed to report at least once in any rolling `Interval`-sized window.
- A failed cycle (Zeus down, network blip, single collector failure) does **not** stop the loop — the next cycle still fires on schedule, and the existing checkpoint + backfill machinery catches up any missed sealed days.

## New config

| Field | Default | Bounds | Purpose |
|---|---|---|---|
| `meterreporter.interval` | `6h` | `[10m, 24h]` | How often each deployment reports |
| `meterreporter.jitter` | `min(Interval, 2h)` | `[10m, Interval]` | Randomness on first run and per cycle |

`Jitter` defaults are derived from `Interval` at use time — you don't need to touch it. Override `Interval` alone and the jitter scales to a sensible value automatically.

### What you get for common configurations

| You set | Resulting `Jitter` |
|---|---|
| (nothing) | 2h (= min(6h, 2h)) |
| `INTERVAL=10m` | 10m (capped at Interval) |
| `INTERVAL=1h` | 1h (capped at Interval) |
| `INTERVAL=24h` | 2h |
| `INTERVAL=6h`, `JITTER=1h` | 1h (explicit) |

### Env var names

`jitter` has no underscores in the field name.

```bash
SIGNOZ_METERREPORTER_INTERVAL=10m
SIGNOZ_METERREPORTER_JITTER=10m
```

## Why this is safe

Correctness is unchanged — only *when* reports are sent changes, never *what* is sent. Three existing mechanisms cover any deferral or repetition:

- **Idempotency key** per (org, day) — Zeus dedupes duplicate calls within the same UTC day
- **Backfill** — any sealed day that gets missed is shipped on the next cycle
- **Checkpoint API** — Zeus tells us what's already been received, so we never re-ship accepted days

## Impact

| | Before | After |
|---|---|---|
| Peak Zeus calls/minute after a rolling deploy | ~1500 | ~12.5/min for 2h |
| Sustained Zeus call rate (after ~1.7 days) | spike + 0 + spike | ~4/min uniform |
| Total daily traffic | unchanged | ~20% more (idempotent, deduped by Zeus) |

That's roughly a **120× reduction** in worst-minute load on Zeus, with no change in correctness. After the fleet has diffused into a uniform distribution across the interval (~1.7 days post-deploy), Zeus sees the same steady rate as the no-spike theoretical ideal.

## Files changed

- `pkg/meterreporter/config.go` — single `Jitter` field, sentinel-based default that scales with `Interval`, `ResolvedJitter` helper
- `ee/meterreporter/httpmeterreporter/provider.go` — jittered `time.Timer` scheduler + observability logs

## Test plan

- [x] Verify defaults: with only `Interval=6h` set, `ResolvedJitter()` returns `2h`; with `Interval=10m` set, it returns `10m`
- [x] Confirm the "scheduling first meter collect" log shows a non-zero `delay` and `delay_ns` on startup
- [x] Confirm the "scheduled next meter collect" log appears after each cycle with delay between `Interval − Jitter` and `Interval`
- [x] In staging with `INTERVAL=10m`, observe at least 3 cycles to confirm per-cycle jitter is varying as expected


karanbalani:
Closes: https://github.com/SigNoz/platform-pod/issues/2348

pkg/meterreporter/config.go
```diff
@@ -15,12 +15,27 @@ type Config struct {
...
+	// (the default) means "derive from Interval" — see ResolvedMaxStartJitter.
+	// Set to 0 explicitly to collect immediately on startup (dev / testing).
+	MaxStartJitter time.Duration `mapstructure:"max_start_jitter"`
+
+	// MaxTickJitter is the maximum amount shaved off Interval for each
+	// subsequent fire. Each cycle = uniform(Interval - MaxTickJitter, Interval].
+	// Negative (the default) means "derive from Interval" — see
+	// ResolvedMaxTickJitter. Must be < Interval so two consecutive fires
+	// cannot collapse to the same instant.
+	MaxTickJitter time.Duration `mapstructure:"max_tick_jitter"`
```
therealpandey:
```suggestion
```

One parameter called jitter.

therealpandey (changes requested):

ee/meterreporter/httpmeterreporter/provider.go
```diff
@@ -170,6 +186,11 @@ func (provider *Provider) collectOrg(ctx context.Context, org *types.Organizatio
 
 		start, end, ok := backfillRange(nextByCollector, todayStart)
 		if ok {
+			provider.settings.Logger().InfoContext(ctx, "backfilling sealed days",
+				slog.String("org_id", org.ID.StringValue()),
+				slog.String("start", start.Format("2006-01-02")),
+				slog.String("end", end.Format("2006-01-02")))
```
therealpandey: please remove.

ee/meterreporter/httpmeterreporter/provider.go
```diff
@@ -257,6 +278,10 @@ func (provider *Provider) report(ctx context.Context, orgID valuer.UUID, license
 		collectedReadings, err := collector.Collect(ctx, orgID, license, window)
 		if err != nil {
 			provider.metrics.collections.Add(ctx, 1, metric.WithAttributes(meterAttr, errors.TypeAttr(err)))
+			provider.settings.Logger().ErrorContext(ctx, "meter collector failed",
+				errors.Attr(err),
+				slog.String("org_id", orgID.StringValue()),
+				slog.String("meter", collector.Name().String()))
```
therealpandey: single line

ee/meterreporter/httpmeterreporter/provider.go
```diff
@@ -283,6 +308,10 @@ func (provider *Provider) report(ctx context.Context, orgID valuer.UUID, license
 
 	provider.metrics.reports.Add(ctx, 1)
 	provider.metrics.meters.Add(ctx, int64(len(meters)))
+	provider.settings.Logger().InfoContext(ctx, "reported meters to zeus",
+		slog.String("org_id", orgID.StringValue()),
+		slog.String("date", date),
+		slog.Int("meters", len(meters)))
```
therealpandey: please remove

ee/meterreporter/httpmeterreporter/provider.go
```diff
@@ -147,6 +162,7 @@ func (provider *Provider) collect(ctx context.Context) {
 }
 
 func (provider *Provider) Stop(ctx context.Context) error {
+	provider.settings.Logger().InfoContext(ctx, "stopping meter reporter")
```
therealpandey:
```suggestion
```

ee/meterreporter/httpmeterreporter/provider.go
```diff
@@ -94,21 +95,35 @@ func newProvider(
...
 	for {
 		select {
 		case <-provider.stopC:
 			return nil
-		case <-ticker.C:
+		case <-timer.C:
 			provider.collect(ctx)
+			next := provider.config.Interval - jitter(jitterCap)
+			timer.Reset(next)
+			provider.settings.Logger().InfoContext(ctx, "scheduled next meter collect", slog.String("delay", next.String()), slog.Int64("delay_ns", next.Nanoseconds()))
```
therealpandey:
```suggestion
```

ee/meterreporter/httpmeterreporter/provider.go
```diff
@@ -94,21 +95,35 @@ func newProvider(
 func (provider *Provider) Start(ctx context.Context) error {
 	close(provider.healthyC)
 
-	provider.collect(ctx)
+	jitterCap := provider.config.ResolvedJitter()
 
-	ticker := time.NewTicker(provider.config.Interval)
-	defer ticker.Stop()
+	startDelay := jitter(jitterCap)
+	provider.settings.Logger().InfoContext(ctx, "scheduling first meter collect", slog.String("delay", startDelay.String()), slog.Int64("delay_ns", startDelay.Nanoseconds()))
```
therealpandey:
```suggestion
```

pkg/meterreporter/config.go
```diff
@@ -29,9 +38,27 @@ func NewConfigFactory() factory.ConfigFactory {
...
+	if c.Jitter >= 0 && (c.Jitter < 10*time.Minute || c.Jitter > c.Interval) {
+		return errors.New(errors.TypeInvalidInput, ErrCodeInvalidInput, "meterreporter::jitter must be between 10m and interval")
 	}
 
 	return nil
 }
+
+// ResolvedJitter returns the configured Jitter, or
+// min(Interval, defaultJitter) when the sentinel default is still in place.
+func (c Config) ResolvedJitter() time.Duration {
```
therealpandey:
```suggestion
func (c Config) Jitter() time.Duration {
```

ee/meterreporter/httpmeterreporter/provider.go
```diff
@@ -94,21 +95,35 @@ func newProvider(
...
 			provider.collect(ctx)
+			next := provider.config.Interval - jitter(jitterCap)
+			timer.Reset(next)
+			provider.settings.Logger().InfoContext(ctx, "scheduled next meter collect", slog.String("delay", next.String()), slog.Int64("delay_ns", next.Nanoseconds()))
 		}
 	}
 }
 
+// jitter returns a uniform random duration in [0, max). Returns 0 if max <= 0.
+func jitter(max time.Duration) time.Duration {
```
therealpandey: make this a function of config. A self-describable name is important.

therealpandey (changes requested):

ee/meterreporter/httpmeterreporter/provider.go
karanbalani: Created NewJitter()

therealpandey (approved):

pkg/meterreporter/config.go
```diff
@@ -15,12 +16,21 @@ type Config struct {
...
 	// Backfill enables sealed-day catch-up from the license creation day.
 	Backfill bool `mapstructure:"backfill"`
+
+	// Jitter is the randomness applied to both the first collect after
+	// Start() and to every subsequent cycle. The first fire happens at a
+	// random time in [0, Jitter); each subsequent cycle takes
+	// Interval - random(0, Jitter). Negative (the default) means "derive
+	// from Interval" via ResolvedJitter, so the value scales with whatever
+	// Interval the user picks.
+	Jitter time.Duration `mapstructure:"jitter"`
```
vikrantgupta25, line 26: When adding things to config lets make sure to update the `example.yaml` as well

karanbalani: Resolved by: https://github.com/SigNoz/signoz/pull/11482

PR #11458, opened 2026-05-26
feat(metrics-explorer): fast path for get stats and get treemap when no filter

nikhilmantri0902:
## Pull Request

  ---

  ### 📄 Summary

  `GetStats` (metrics explorer) narrows the samples-table scan via `fingerprint IN (huge set)` regardless of filter presence. That set is lots of UInt64s. 

  When no user filter is present, `metric_name IN (small set)` is semantically equivalent. 

  This PR adds `fetchMetricsStatsWithSamplesFastPath` and dispatches to it from `GetStats` when `req.Filter` is empty.

  #### Issues closed by this PR

  Part of https://github.com/SigNoz/engineering-pod/issues/5043

  ---

  ### ✅ Change Type

  - [x] ♻️ Refactor (perf optimization, behavior-preserving)

  ---

  ### 🐛 Bug Context

  #### Fix Strategy

  Dispatch upstream from `req.Filter` itself in `GetStats`:

  - `hasFilter := req.Filter != nil && strings.TrimSpace(req.Filter.Expression) != ""`
  - `!hasFilter` → call new `fetchMetricsStatsWithSamplesFastPath` (no filter param, skip `buildFilterClause` entirely)
  - `hasFilter` → existing path unchanged

  Fastpath emits two CTEs (`__time_series_counts`, `__sample_counts`) with samples scan narrowed by `metric_name IN (SELECT DISTINCT metric_name FROM <local_ts> WHERE ...)` via `sb.Var(metricNamesSB)`. Local TS table chosen to avoid distributed broadcast for the inner DISTINCT.

  ---

  ### 🧪 Testing Strategy

  - **Tests added/updated:** none (no existing unit coverage on this function; deferred).
  - **Manual verification:**
    - No-filter call → SQL contains `metric_name IN (SELECT DISTINCT metric_name ...)`, no `__filtered_fingerprints`.
    - With-filter call → existing slowpath SQL unchanged (`__filtered_fingerprints` + `fingerprint IN`).
    - Row counts + ordering identical between fastpath and slowpath outputs for no-filter case against a populated dev ClickHouse.
  - **Edge cases covered:**
    - Empty filter expression (whitespace-only) → fastpath (matches `buildFilterClause` internal check).
    - `req.Filter == nil` → fastpath.
    - Same `orderBy`, `limit`, `offset` semantics as slowpath.
    - `count(*)` vs `sum(count)` table-aware via `CountExpressionForSamplesTable` (unchanged from slowpath).

  ---

  - **Blast radius:** `GetStats` only. `GetTreemap`, `computeTimeseriesTreemap`, `computeSamplesTreemap`, `fetchMetricsStatsWithSamples` untouched.
  - **Potential regressions:**
    - With-filter path unchanged — safest.
    - No-filter path: new SQL shape. Semantic equivalence with slowpath (same row set, same counts) — verified manually. Worth a perf bake before relying on it for downstream automation.
  - **Rollback plan:** revert this PR. Slowpath alone still handles all cases correctly (just slower for no-filter).

  ---

  ### 📝 Changelog

  | Field            | Value                                              |
  | ---------------- | -------------------------------------------------- |
  | Deployment Type  | Cloud / OSS / Enterprise                           |
  | Change Type      | Maintenance (perf)                                 |
  | Description      | Faster metrics-stats endpoint when no filter is applied |

  ---

  ### 📋 Checklist

  - [x] Tests added or explicitly not required (none added — no existing coverage on this function, behavior-preserving change)
  - [x] Manually tested
  - [x] Backward compatibility considered (response shape unchanged)

  ---

  ## 👀 Notes for Reviewers

  - Dispatch logic mirrors `buildFilterClause`'s own emptiness check (`module.go:928-934`) — kept identical so behavior tracks if that ever changes.
  - `fetchMetricsStatsWithSamples` left intact (not renamed to `*SlowPath`) — minimizes diff. 
  - Inner subquery uses **local** TS table (`localTsTable` from `WhichTSTableToUse`), not distributed — matches the existing slowpath's `__filtered_fingerprints` CTE pattern and avoids broadcast overhead for the small DISTINCT result.


pkg/modules/metricsexplorer/implmetricsexplorer/module.go
```diff
@@ -1085,6 +1091,109 @@ func (m *module) fetchMetricsStatsWithSamples(
...
+
+	// Samples counts per metric
+	samplesSB := sqlbuilder.NewSelectBuilder()
+	samplesSB.Select(
+		"metric_name",
+		fmt.Sprintf("%s AS samples", countExp),
+	)
+	samplesSB.From(fmt.Sprintf("%s.%s", telemetrymetrics.DBName, samplesTable))
+	samplesSB.Where(samplesSB.Between("unix_milli", req.Start, req.End))
+	samplesSB.Where("NOT startsWith(metric_name, 'signoz')")
```
namanv3: this not starts with check is already there in the distinct metric name subquery, we won't need it here right?

pkg/modules/metricsexplorer/implmetricsexplorer/module.go
```diff
@@ -1085,6 +1091,109 @@ func (m *module) fetchMetricsStatsWithSamples(
...
+		"COALESCE(ts.metric_name, s.metric_name) AS metric_name",
+		"COALESCE(ts.timeseries, 0) AS timeseries",
+		"COALESCE(s.samples, 0) AS samples",
+		"COUNT(*) OVER() AS total",
+	)
+	finalSB.From("__time_series_counts ts")
+	finalSB.JoinWithOption(sqlbuilder.FullOuterJoin, "__sample_counts s", "ts.metric_name = s.metric_name")
+	finalSB.Where("(COALESCE(ts.timeseries, 0) > 0 OR COALESCE(s.samples, 0) > 0)")
+
+	orderByColumn, orderDirection, err := getStatsOrderByColumn(orderBy)
```
namanv3: lot of logic is common which I think can be extracted out?

namanv3 (commented):
Sorry for checking this a bit late, I can pick these changes up if you don't get to see these comments before you go on leave

pkg/modules/metricsexplorer/implmetricsexplorer/module.go
nikhilmantri0902: Extracted out common logic, please review again

pkg/modules/metricsexplorer/implmetricsexplorer/module.go
nikhilmantri0902: please review after the common logic extraction, I believe we can keep these filters for every CTE for faster lookup.

namanv3 (dismissed):

pkg/modules/metricsexplorer/implmetricsexplorer/module.go
```diff
@@ -974,15 +962,23 @@ func (m *module) buildFilterClause(ctx context.Context, filter *qbtypes.Filter,
 func (m *module) fetchMetricsStatsWithSamples(
 	ctx context.Context,
 	req *metricsexplorertypes.StatsRequest,
-	filterWhereClause *sqlbuilder.WhereClause,
 	normalized bool,
-	orderBy *qbtypes.OrderBy,
 ) ([]metricsexplorertypes.Stat, uint64, error) {
 	ctx = m.withMetricsExplorerContext(ctx, "fetchMetricsStatsWithSamples")
 
+	hasFilter := req.Filter != nil && strings.TrimSpace(req.Filter.Expression) != ""
```
therealpandey, line 964: Unit test this function please.

therealpandey: Please write the unit tests in the a way that you are asserting on actual sql statements like https://github.com/SigNoz/signoz/blob/a0b14e08353c5e3ebb9e35997153f01871f9bee4/pkg/modules/tracedetail/impltracedetail/store_test.go#L46 or any number of examples in the querier codebase.

therealpandey (changes requested):

pkg/modules/metricsexplorer/implmetricsexplorer/module.go
nikhilmantri0902: added tests for get stats and get treemap load bearing methods of this module.

therealpandey (approved):

PR #11462, opened 2026-05-26
feat(trace-detail): add flamegraph v3 with optimized memory consumption (part 1)

krsoninikhil:
## Pull Request

Part of https://github.com/SigNoz/engineering-pod/issues/4787

Closes https://github.com/SigNoz/engineering-pod/issues/4949 (sub-issue)

Objective: Creates a new version for flamegraph that builds the flamegraph tree with minimal fields and only queries the full spans after sampling the supported no. (`selectedAllSpanLImit`) of spans.

Contract change from v2:
- No need for `limit` in request payload (it was added in v2 for backward compatibility only)
- service name is removed from the span 
- references is removed from the span and `parent_span_id` is added instead.

Follow up PR: https://github.com/SigNoz/signoz/pull/11491

TDD: https://www.notion.so/signoz/TDD-Waterfall-Flamegraph-Memory-Allocation-Optimization-364fcc6bcd19804fac09d844e96cb5a2?v=2a7fcc6bcd198049b122000c4ff33372&source=copy_link#364fcc6bcd1980969633f867a5679285

### DIFF
--- a/frontend/src/api/generated/services/sigNoz.schemas.ts
+++ b/frontend/src/api/generated/services/sigNoz.schemas.ts
@@ -7675,20 +7675,34 @@ export enum SpantypesFieldContextDTO {
 	attribute = 'attribute',
 	resource = 'resource',
 }
-export type SpantypesFlamegraphSpanDTOAttributes = { [key: string]: unknown };
+export type SpantypesFlamegraphSpanDTOAttributesAnyOf = {
+	[key: string]: unknown;
+};
 
-export type SpantypesFlamegraphSpanDTOResource = { [key: string]: string };
+/**
+ * @nullable
+ */
+export type SpantypesFlamegraphSpanDTOAttributes =
+	SpantypesFlamegraphSpanDTOAttributesAnyOf | null;
+
+export type SpantypesFlamegraphSpanDTOResourceAnyOf = { [key: string]: string };
+
+/**
+ * @nullable
+ */
+export type SpantypesFlamegraphSpanDTOResource =
+	SpantypesFlamegraphSpanDTOResourceAnyOf | null;
 
 export interface SpantypesFlamegraphSpanDTO {
 	/**
-	 * @type object
+	 * @type object,null
 	 */
-	attributes?: SpantypesFlamegraphSpanDTOAttributes;
+	attributes: SpantypesFlamegraphSpanDTOAttributes;
 	/**
 	 * @type integer
 	 * @minimum 0
 	 */
-	durationNano?: number;
+	durationNano: number;
 	/**
 	 * @type array
 	 */
@@ -7696,33 +7710,33 @@ export interface SpantypesFlamegraphSpanDTO {
 	/**
 	 * @type boolean
 	 */
-	hasError?: boolean;
+	hasError: boolean;
 	/**
 	 * @type integer
 	 * @format int64
 	 */
-	level?: number;
+	level: number;
 	/**
 	 * @type string
 	 */
-	name?: string;
+	name: string;
 	/**
 	 * @type string
 	 */
-	parentSpanId?: string;
+	parentSpanId: string;
 	/**
-	 * @type object
+	 * @type object,null
 	 */
-	resource?: SpantypesFlamegraphSpanDTOResource;
+	resource: SpantypesFlamegraphSpanDTOResource;
 	/**
 	 * @type string
 	 */
-	spanId?: string;
+	spanId: string;
 	/**
 	 * @type integer
 	 * @minimum 0
 	 */
-	timestamp?: number;
+	timestamp: number;
 }
 
 export interface SpantypesGettableFlamegraphTraceDTO {
--- a/frontend/src/api/generated/services/sigNoz.schemas.ts
+++ b/frontend/src/api/generated/services/sigNoz.schemas.ts
@@ -7675,6 +7675,81 @@ export enum SpantypesFieldContextDTO {
 	attribute = 'attribute',
 	resource = 'resource',
 }
+export type SpantypesFlamegraphSpanDTOAttributes = { [key: string]: unknown };
+
+export type SpantypesFlamegraphSpanDTOResource = { [key: string]: string };
+
+export interface SpantypesFlamegraphSpanDTO {
+	/**
+	 * @type object
+	 */
+	attributes?: SpantypesFlamegraphSpanDTOAttributes;
+	/**
+	 * @type integer
+	 * @minimum 0
+	 */
+	durationNano?: number;
+	/**
+	 * @type array,null
+	 */
+	event?: SpantypesEventDTO[] | null;
+	/**
+	 * @type boolean
+	 */
+	hasError?: boolean;
+	/**
+	 * @type integer
+	 * @format int64
+	 */
+	level?: number;
+	/**
+	 * @type string
+	 */
+	name?: string;
+	/**
+	 * @type string
+	 */
+	parentSpanId?: string;
+	/**
+	 * @type object
+	 */
+	resource?: SpantypesFlamegraphSpanDTOResource;
+	/**
+	 * @type string
+	 */
+	serviceName?: string;
+	/**
+	 * @type string
+	 */
+	spanId?: string;
+	/**
+	 * @type integer
+	 * @minimum 0
+	 */
+	timestamp?: number;
+}
+
+export interface SpantypesGettableFlamegraphTraceDTO {
+	/**
+	 * @type integer
+	 * @format int64
+	 */
+	endTimestampMillis?: number;
+	/**
+	 * @type boolean
+	 */
+	hasMore?: boolean;
+	/**
+	 * @type array,null
+	 */
+	spans?: SpantypesFlamegraphSpanDTO[][] | null;
+	/**
+	 * @type integer
+	 * @format int64
+	 */
+	startTimestampMillis?: number;
+}
+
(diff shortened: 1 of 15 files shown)


pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -69,22 +96,6 @@ func (m *module) getTraceData(ctx context.Context, traceID string) (*spantypes.W
...
-	if err != nil {
-		return nil, err
-	}
-	effectiveLimit := min(selectAllLimit, m.config.Waterfall.MaxLimitToSelectAllSpans)
-	if summary.NumSpans > uint64(effectiveLimit) {
-		return m.getWindowedWaterfall(ctx, traceID, selectedSpanID, uncollapsedSpans, summary.Start, summary.End)
-	}
-	return m.getFullWaterfall(ctx, traceID, summary)
-}
-
```
krsoninikhil, line 103: Just moved this to the top.

pkg/types/spantypes/flamegraph_trace.go
```diff
@@ -0,0 +1,118 @@
...
+	return nil
+}
+
+// GetSelectedLevels returns the window of levels around selectedSpanID with sampling applied to dense levels.
+func (t *FlamegraphTrace) GetSelectedLevels(
+	selectedSpanID string,
+	levelLimit, spansPerLevel, topLatencyCount, bucketCount int,
+) []FlamegraphLevel {
+	return nil
+}
```
krsoninikhil, line 50: Working separately on these methods: https://github.com/SigNoz/signoz/pull/11491

pkg/modules/tracedetail/tracedetail.go
```diff
@@ -11,10 +11,12 @@ import (
...
 	GetWaterfall(http.ResponseWriter, *http.Request)
 	GetWaterfallV4(http.ResponseWriter, *http.Request)
+	GetFlamegraph(http.ResponseWriter, *http.Request)
 }
 
 // Module defines the business logic for trace detail operations.
 type Module interface {
 	GetWaterfall(ctx context.Context, traceID string, req *spantypes.PostableWaterfall) (*spantypes.GettableWaterfallTrace, error)
 	GetWaterfallV4(ctx context.Context, traceID string, selectedSpanID string, uncollapsedSpans []string, selectAllLimit uint) (*spantypes.GettableWaterfallTrace, error)
+	GetFlamegraph(ctx context.Context, traceID string, req *spantypes.PostableFlamegraph) (*spantypes.GettableFlamegraphTrace, error)
```
nityanandagohain: let's decompose and not pass req to module

pkg/types/spantypes/flamegraph_span.go
```diff
@@ -0,0 +1,81 @@
...
+	Children     []*FlamegraphSpan `json:"-"` // internal tree use only
+}
+
+// FlamegraphLevel groups span IDs at a single level within the selected window.
+type FlamegraphLevel struct {
+	Level   int64
+	SpanIDs []string
+}
+
+type PostableFlamegraph struct {
```
nityanandagohain, line 27: @therealpandey any suggestions for naming here, as it is more a of query payload and not really a `PostableFlamegraph`

pkg/types/spantypes/flamegraph_trace.go
```diff
@@ -0,0 +1,118 @@
...
+		if t.roots[i].Timestamp == t.roots[j].Timestamp {
+			return t.roots[i].SpanID < t.roots[j].SpanID
+		}
+		return t.roots[i].Timestamp < t.roots[j].Timestamp
+	})
+}
+
+func flamegraphSpanIndex(spans []*FlamegraphSpan, spanID string) int {
+	for i, s := range spans {
+		if s != nil && s.SpanID == spanID {
```
nityanandagohain: s can never be nil

pkg/types/spantypes/flamegraph_span.go
```diff
@@ -0,0 +1,81 @@
...
+type FlamegraphSpan struct {
+	SpanID       string            `json:"spanId"`
+	ParentSpanID string            `json:"parentSpanId"`
+	Timestamp    uint64            `json:"timestamp"`
+	DurationNano uint64            `json:"durationNano"`
+	HasError     bool              `json:"hasError"`
+	ServiceName  string            `json:"serviceName"`
+	Name         string            `json:"name"`
+	Level        int64             `json:"level"`
+	Events       []Event           `json:"event"`
```
nityanandagohain: All of thee are needed for flamegraph span ?

nityanandagohain (commented):
some comments

pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -46,29 +46,6 @@ func (m *module) GetWaterfall(ctx context.Context, traceID string, req *spantype
...
-		return nil, spantypes.ErrTraceNotFound
-	}
-
-	nodes := make([]*spantypes.WaterfallSpan, len(spanItems))
-	for i := range spanItems {
-		nodes[i] = spanItems[i].ToWaterfallSpan(traceID)
-	}
-	return spantypes.NewWaterfallTraceFromSpans(nodes), nil
-}
-
```
krsoninikhil, line 84: Moved the internal methods to the bottom.

pkg/types/spantypes/flamegraph_trace.go
krsoninikhil: It's a independent util method, should we keep this check to avoid breaking for bad arguments?

pkg/types/spantypes/flamegraph_span.go
krsoninikhil: For waterfall v3, we had discussed the request struct names to be `WaterfallQuery` or `WaterfallRequest`. But since this `Postable*` format was already being used for similar cases in other handlers, we settled for this.

pkg/types/spantypes/flamegraph_span.go
krsoninikhil: Yes, we show events and in flamegraph. serviceName should no longer be required after `selectFields` change, will confirm from UI if they are using it.

pkg/types/spantypes/flamegraph_trace.go
nityanandagohain: I mean we should hanlde things in a way that nil never gets added to spans, is there a case where it will be added.

pkg/types/spantypes/flamegraph_span.go
nityanandagohain: ok

pkg/types/spantypes/flamegraph_trace.go
krsoninikhil: There won't be any nil, removed the check [722a6fb](https://github.com/SigNoz/signoz/pull/11462/commits/722a6fb3f597450c430165a262e835f05b49eba6).

pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -164,6 +165,17 @@ func (m *module) GetTraceAggregations(ctx context.Context, traceID string, req *
...
 
+func (m *module) GetFlamegraph(ctx context.Context, traceID string, selectedSpanID string, selectFields []telemetrytypes.TelemetryFieldKey) (*spantypes.GettableFlamegraphTrace, error) {
+	summary, err := m.store.GetTraceSummary(ctx, traceID)
+	if err != nil {
+		return nil, err
+	}
+	if summary.NumSpans <= uint64(m.config.Flamegraph.SelectAllSpansLimit) {
+		return m.getFullFlamegraph(ctx, traceID, summary, selectFields)
+	}
+	return m.getWindowedFlamegraph(ctx, traceID, selectedSpanID, summary, selectFields)
```
krsoninikhil, line 176: Will add a similar metric here like waterfall in next PR.

pkg/types/spantypes/flamegraph_trace.go
```diff
@@ -0,0 +1,111 @@
...
+		t.startTime = timestamp
+	}
+	if end := timestamp + durationNano; end > t.endTime {
+		t.endTime = end
+	}
+}
+
+func (t *FlamegraphTrace) buildSpanTree() {
+	for _, node := range t.nodeByID {
+		if node.ParentSpanID != "" {
```
krsoninikhil, line 83: In existing flamgraph api, we use `links` with `CHILD_OF` to get the parent ([ref](https://github.com/SigNoz/signoz/blob/341fa1900696772a3795c4d5d549939f4e7ec8a3/pkg/query-service/app/clickhouseReader/reader.go#L1228)). Change that to this format to avoid reading unmarshalling the links since it's stored as string.

Let me know if that is not the same thing. @nityanandagohain

frontend/src/api/generated/services/sigNoz.schemas.ts
```diff
@@ -7976,6 +8047,17 @@ export interface SpantypesGettableWaterfallTraceDTO {
...
+export interface SpantypesPostableFlamegraphDTO {
+	/**
+	 * @type array
+	 */
+	selectFields?: TelemetrytypesTelemetryFieldKeyDTO[];
+	/**
+	 * @type string
+	 */
+	selectedSpanId?: string;
+}
```
krsoninikhil, line 8059: These are optional parameters in the request.

therealpandey (approved):

PR #11465, opened 2026-05-26
feat: introduce dashboard views for list dashboards page

### LINKED ISSUE #10735: Saved views for dashboards
A dashboard can have multiple variables, and for various use cases, having different values for those variables can be required. Selecting those values one by one each time can get cumbersome.

A user should be save pre-defined variable selections as a quick way to toggle the view for the different use cases. These selections can be saved as Saved Views/Presets/any other name that makes sense.

namanv3:
## Pull Request

---

### 📄 Summary

Saved views for dashboards list are used to share filter configs within an org. They can be used to easily select a filter configuration without having to go through multiple clicks.

This PR covers all the APIs needed to support them in the UI

#### Issues closed by this PR

Closes https://github.com/SigNoz/signoz/issues/10735

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

- Blast radius: New APIs, New table, no other flow impacted
- Potential regressions: None
- Rollback plan: Revert PR, Delete new table

---

## 👀 Notes for Reviewers

Dashboard Views are not explicitly made their own Kind, nor are they set up to have their own permission set like public dashboard config. The idea is that views should be tied to the resource that they are for, and access level to the resource can decide access to views. 

---


docs/api/openapi.yml
```diff
@@ -13818,6 +13872,236 @@ paths:
       summary: Lock dashboard (v2)
       tags:
       - dashboard
+  /api/v2/dashboards/views:
```
srikanthccv: this should be `api/v2/dashboards-views` or `api/v2/dashboards_views` there are two "collections" nested. The fact that you had to reorder hack in router should be the giveaway to see this as smell and address.

pkg/modules/dashboard/impldashboard/store.go
```diff
@@ -472,3 +472,92 @@ func (store *store) DeletePreferencesForUser(ctx context.Context, orgID valuer.U
...
+		Where("id = ?", id).
+		Where("org_id = ?", orgID).
+		Scan(ctx)
+	if err != nil {
+		return nil, store.sqlstore.WrapNotFoundErrf(err, dashboardtypes.ErrCodeDashboardViewNotFound, "dashboard view with id %s doesn't exist", id)
+	}
+	return view, nil
+}
+
+func (store *store) ListDashboardViews(ctx context.Context, orgID valuer.UUID) ([]*dashboardtypes.DashboardView, error) {
```
srikanthccv, line 505: In the design, we had out-of-box quick to access virtual views for user such as `pinned`, `mine`, is that going to driven by frontend?

pkg/apiserver/signozapiserver/dashboard.go
```diff
@@ -14,6 +14,76 @@ import (
...
+		SecuritySchemes:     newSecuritySchemes(types.RoleEditor),
+	})).Methods(http.MethodPut).GetError(); err != nil {
+		return err
+	}
+
+	if err := router.Handle("/api/v2/dashboards/views/{id}", handler.New(provider.authzMiddleware.EditAccess(provider.dashboardHandler.DeleteView), handler.OpenAPIDef{
+		ID:                  "DeleteDashboardView",
+		Tags:                []string{"dashboard"},
+		Summary:             "Delete dashboard saved view",
+		Description:         "Removes a saved view. Saved views are shared org-wide; any user in the org may delete any view. Deleting a non-existent view returns 404.",
```
srikanthccv: outdated descriptions. please re-check.

pkg/types/dashboardtypes/dashboard_view.go
```diff
@@ -0,0 +1,142 @@
...
+	var tmp alias
+	if err := dec.Decode(&tmp); err != nil {
+		return errors.WrapInvalidInputf(err, ErrCodeDashboardViewInvalidInput, "invalid saved view request body").WithAdditional(err.Error())
+	}
+	*p = PostableDashboardView(tmp)
+	return p.Validate()
+}
+
+func (p *PostableDashboardView) Validate() error {
+	name, err := trimAndValidateDashboardViewName(p.Name)
```
srikanthccv: have some max length restriction on query string as well. perhaps something like 1024 or similar.

pkg/types/dashboardtypes/dashboard_view.go
```diff
@@ -0,0 +1,142 @@
...
+		Name:          p.Name,
+		Data:          p.Data,
+	}
+}
+
+// ════════════════════════════════════════════════════════════════════════
+// Updateable
+// ════════════════════════════════════════════════════════════════════════
+
+type UpdateableDashboardView = PostableDashboardView
```
srikanthccv: `UpdatableX` consistency.

pkg/sqlmigration/095_add_dashboard_view.go
```diff
@@ -0,0 +1,69 @@
...
+
+	sqls := migration.sqlschema.Operator().CreateTable(&sqlschema.Table{
+		Name: "dashboard_view",
+		Columns: []*sqlschema.Column{
+			{Name: "id", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "name", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "data", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "org_id", DataType: sqlschema.DataTypeText, Nullable: false},
+			{Name: "created_at", DataType: sqlschema.DataTypeTimestamp, Nullable: false},
+			{Name: "updated_at", DataType: sqlschema.DataTypeTimestamp, Nullable: false},
```
srikanthccv, line 46: This should have the user auditable fields as well?

srikanthccv (commented):
first pass

pkg/modules/dashboard/impldashboard/store.go
namanv3: yes, storing in backend doesn't offer anything. in fact it adds complexities like having system views, can delete, etc.

docs/api/openapi.yml
namanv3: makes sense. fixed

pkg/apiserver/signozapiserver/dashboard.go
namanv3: fixed

pkg/types/dashboardtypes/dashboard_view.go
namanv3: done, realised that we should have it on list api as well so just extracted out a common type for it

pkg/types/dashboardtypes/dashboard_view.go
namanv3: fixed

pkg/sqlmigration/095_add_dashboard_view.go
namanv3: @vikrantgupta25 had suggested not to at the time of tags because of proper audit trail coming up I think.
https://github.com/SigNoz/signoz/pull/11254#discussion_r3224028428

pkg/types/dashboardtypes/dashboard_view.go
```diff
@@ -0,0 +1,125 @@
...
+type ListableDashboardView struct {
+	Views []*DashboardView `json:"views" required:"true" nullable:"false"`
+}
+
+// ════════════════════════════════════════════════════════════════════════
+// Helpers
+// ════════════════════════════════════════════════════════════════════════
+
+func trimAndValidateDashboardViewName(name string) (string, error) {
+	trimmed := strings.TrimSpace(name)
```
therealpandey:
```suggestion
```

Consider the user sending a request with body:
{
"name": "  ValidNameButLeadingAndTrailingSpace "
}.

You're going to end up storing "ValidNameButLeadingAndTrailingSpace". Magic happened when you mutated the input and stored it.

In reconciliation systems like terraform or k8s, this is going to lead to permanent diff.
```
resource "dashboard_view" "my_view" {
 name = " ValidNameButLeadingAndTrailingSpace "
}
```

When the system calls GET to validate this, GET will return "ValidNameButLeadingAndTrailingSpace" and it will be permanent diff.

Reject the request if your regex is not followed.

therealpandey (changes requested):

pkg/types/dashboardtypes/dashboard_view.go
namanv3: I agree with this concept and that we shouldn't do it for entities that will be a part of such systems. But I don't think saved views will ever be in such a system. These views are purely for our UI so they don't make sense to be backed up through gitops or terraform. With that in mind I think we can make the API more friendly by trimming down the name ourselves.

srikanthccv: On a second thought, I agree with @therealpandey. One immediate thing brings to table is deciding this per entity. If views trim but dashboard doesn't, we end up creating a somewhat arbitrary criteria and someone has to remember and a debate that recurs on every new resource.

In practice, APIs almost always outlive their intended consumers. Let's say our leading consumer is frontend, then frontend can definitely trim itself before sending to API (where the trimmed values is visible to user, pretty common in UI input elements). If the second leading client is API from agent or terraform or something else, the consumer is always going to act on error because they want the resource to be saved.

The mutation on server side is a problem on its own even if we keep the gitops aside. This mutation is not the same thing as the populating default values because defaulting works on the fact that client didn't specify anything. So trimming and rewriting becomes sort of category you can't quite give a clear reason to. What I mean to say is, A missing field has exactly one way to populate (there is only one set of default values) but trimming is not so you need to be clear about it and make it contract level thing at which point it's better not do that in the first place.

Does this mean, we should never trim? That's not a rule either. For example, if you have list query api and the sort order is ` desc  ` with then be liberal and trim it. However, the moment you are going to save something to DB, mutating client sent fields is generally not a good idea and silent operation that client doesn't expect (unless it's contract in itself). 

As a general rule of thumb. Let's not mutate any fields client sends. Either reject with clear error or keep as is.

namanv3: makes sense. I've made the change. I am not introducing any regex though, cuz dashboard view name is more like a display name. let me know if you think we should

srikanthccv: sounds good, regex is not needed given it's a display label and not an identity.

srikanthccv (dismissed):

therealpandey (dismissed):

srikanthccv (approved):

therealpandey (approved):

PR #11467, opened 2026-05-26
chore(oxlint): disable max-params, jsdoc, and more rules for tests

H4ad:
## Pull Request

---

A few changes on oxlint:
- disable max-params: we don't actually use it or care to reduce it, most of the times is just noise
- disable check-tag-names: orval emit some non-standard tags, most of the usages are from them, don't make sense to not allow custom tags on jsdoc.
- for test files:
  - disable explicit return types
  - allow usage of any

This reduces the amount of warnings from ~4,2k to ~3,1k

By rule:

| Rule | Amount of Warnings |
|--------|--------|
| eslint(max-params) | 256 |
| eslint-plugin-jsdoc(check-tag-names) | 286 |
| typescript-eslint(no-explicit-any) | ~500 only on tests, 1k in total |

Part of https://github.com/SigNoz/engineering-pod/issues/5087, we need to first reduce this amount of warnings before trying to add any other rule.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Blast radius: No
- Potential regressions: No

---

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | N/A |

---

### 📋 Checklist
- [ ] Manually tested

### DIFF
--- a/frontend/.oxlintrc.json
+++ b/frontend/.oxlintrc.json
@@ -123,6 +123,7 @@
     "react/require-render-return": "error",
     "react/no-unsafe": "off",
     "no-array-constructor": "error",
+    "jsdoc/check-tag-names": "off",
     "@typescript-eslint/no-duplicate-enum-values": "warn",
     // TODO: Change to error after migration to oxlint
     "@typescript-eslint/no-empty-object-type": "error",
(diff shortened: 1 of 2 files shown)


AshwinBhatkal:
Can we keep max-params as warn itself so that we don't do it for new ones? Old ones can still have some warnings and change can be  made when convenient

H4ad:
@AshwinBhatkal I don't think we will fix them anytime soon, so unless we have a plan to fix them, would be better to just disable and focus to keep errors/warnings as zero. If we don't plan to fix, it's just noise

AshwinBhatkal (approved):

PR #11470, opened 2026-05-27
refactor: explicit docker image registry for podman support

### LINKED ISSUE #11469: using the Compose file with podman does not work out-of-the-box
## Bug description

The `deploy/docker` Compose works fine with Docker. But it does not work immediately with Podman, because podman does not assume that images will be located on the `docker.io` registry (because Podman is not Docker).

## Expected behavior

Running the Compose file with Podman would work without editing the Compose file.

## How to reproduce

1. `cd deploy/docker`
2. `podman compose up`
3. `Error: short-name "signoz/zookeeper:3.7.1" did not resolve to an alias and no unqualified-search registries are defined in "/etc/containers/registries.conf"`

## Version information 
* **Signoz version**: current latest (commit 9d36031d4ef2880a9bc77bda758ab5d6b5a7ab6b, describe `v0.108.0-rc.1-746-g9d36031d4`)
* **Browser version**: N/A
* **Your OS and version**: Ubuntu 24.04
* **Your CPU Architecture**(ARM/Intel): AMD64

## Additional context

The `docker.io` registry could be added as default to podman configuration, but that's not the default.

Also, this changes has no impact on Docker, it handles explicit reference too.

I have opened a Pull Request alongside, if you are interested. But because the lines edited contain also version numbers, which are frequently updated, the PR may rapidly get conflicts.

Lenormju:
## Pull Request

---

### 📄 Summary

#11469 says it all : without explicit `docker.io/` registry prefix, image references will be correctly handled by Docker but not by Podman. I'd like to make them explicit : no downside for Docker users, and will make it work out-of-the-box for Podman users.

 N/A

Closes #11469

---

### ✅ Change Type

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context

Docker automatically assumes images will be available on `docker.io`, Podman don't.

#### Root Cause

Docker is not the official standard, but de-facto standard.

#### Fix Strategy

Actually, this PR does not solve the root cause, which is outside the scope of SigNoz.

---

### 🧪 Testing Strategy

1. I have run `podman compose up` without the `docker.io` registry rpefix, and got the error.
2. Added the prefix (commit), run again and it works (download the images then started).
3. Removed the images from my local registry (cache) and the prefix from the Compose file, error again.

On Docker, no visible change (as expected).

No automatic testing.

---

No runtime risk.

---

### 📝 Changelog

N/A

---

### 📋 Checklist

- [x] Tests ~~added~~ or explicitly not required
- [x] Manually tested
- [x] ~~Breaking changes documented~~

---

## 👀 Notes for Reviewers

I haven't updated the `.devenv/docker/clickhouse/compose.yaml` image references, could not test locally.

---

### DIFF
--- a/deploy/docker-swarm/docker-compose.ha.yaml
+++ b/deploy/docker-swarm/docker-compose.ha.yaml
@@ -11,7 +11,7 @@ x-common: &common
       max-file: "3"
 x-clickhouse-defaults: &clickhouse-defaults
   !!merge <<: *common
-  image: clickhouse/clickhouse-server:25.5.6
+  image: docker.io/clickhouse/clickhouse-server:25.5.6
   tty: true
   deploy:
     labels:
@@ -41,7 +41,7 @@ x-clickhouse-defaults: &clickhouse-defaults
     - CLICKHOUSE_SKIP_USER_SETUP=1
 x-zookeeper-defaults: &zookeeper-defaults
   !!merge <<: *common
-  image: signoz/zookeeper:3.7.1
+  image: docker.io/signoz/zookeeper:3.7.1
   user: root
   deploy:
     labels:
@@ -64,7 +64,7 @@ x-db-depend: &db-depend
 services:
   init-clickhouse:
     !!merge <<: *common
-    image: clickhouse/clickhouse-server:25.5.6
+    image: docker.io/clickhouse/clickhouse-server:25.5.6
     command:
       - bash
       - -c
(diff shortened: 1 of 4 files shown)


therealpandey:
Hi @Lenormju,
Please use https://github.com/SigNoz/foundry. It's the canonical way to install SigNoz now.

Feedback is appreciated on foundry. (PS this is already fixed there).

Lenormju:
> Hi @Lenormju, Please use https://github.com/SigNoz/foundry. It's the canonical way to install SigNoz now.
> 
> Feedback is appreciated on foundry. (PS this is already fixed there).

Then I can close the issue.

PR #11471, opened 2026-05-27
fix: classify numeric columns as aggregation in readAsScalar

### LINKED ISSUE #8844: [Bug] Pie Chart with ClickHouse Query only works if aggregation alias is `__result_[n]`
## Bug description

When creating a Pie Chart panel using a ClickHouse Query, the chart shows "No Data" unless the aggregation column is explicitly named `__result_0`.

Example:
✅ Works
```sql
SELECT
  response_status_code,
  count() AS __result_0
GROUP BY response_status_code
```

❌ Fails
```sql
SELECT
  response_status_code,
  count() AS total_requests
GROUP BY response_status_code
```

The backend returns valid data in all cases, but the frontend only interprets columns with aliases like `__result_0`, `__result_1`, etc. as aggregations.
Any other alias is treated as a group, causing the Pie chart to fail with "No Data".

## Expected behavior

- Pie Chart should accept any alias for the aggregation column (e.g., `AS total_requests`), not only `__result_0`.
- Alternatively, the user should be able to manually select which column is the "label" and which column is the "value".

## How to reproduce

1. Create a new dashboard panel → select Pie Chart.
2. Switch to ClickHouse Query mode.
3. Run a query like:
```sql
SELECT
  response_status_code,
  count() AS total_requests
GROUP BY response_status_code
```
4. Observe that the query response contains results, but the Pie Chart shows "No Data".
5. Change the alias to AS `__result_0` (or `__result_1`, `__result_2`, …). → Pie Chart renders correctly.

## Version information 
* **Signoz version**: v0.92.2
* **Browser version**: Version 139.0.7258.128 (Official Build) (arm64)
* **Your OS and version**: macOS Sequoia Version 15.5 (24F74)

## Additional context

- This behavior makes ClickHouse Query panels fragile and non-intuitive, as users must know about internal alias naming (`__result_[n]`).
- Likely a regression compared to older SigNoz versions, where arbitrary aliases worked.

krsoninikhil:
## Pull Request

---

### 📄 Summary

`readAsScalar` only recognised columns named `__result_N` as aggregations. Any other name — including user-defined aliases like `count() AS total_requests` — defaulted to `columnType: "group"`. The frontend uses this field to split columns into labels vs values for pie charts and scalar panels, so custom aliases caused charts to show "No Data".

The fix mirrors the fallback logic already present in `readAsTimeSeries`: if a column doesn't match an explicit marker (`__result_N` or a legacy alias like `result`/`value`), infer aggregation intent from the DB column's numeric type.

#### Issues closed by this PR

Closes #8844

---

### ✅ Change Type

- [x] 🐛 Bug fix

---

#### Root Cause

`readAsScalar` classified columns purely by name regex (`^__result_(\d+)$`). `readAsTimeSeries` had a fuller fallback chain (single numeric column, legacy aliases) but `readAsScalar` never received the same treatment.

#### Fix Strategy

Add two fallback conditions to `readAsScalar`, in priority order:
1. Name in `legacyReservedColumnTargetAliases` (`result`, `value`, `res`, …) → aggregation
2. Column has a numeric DB type → aggregation

Explicit `__result_N` markers still take priority, so all existing queries are unaffected.

---

### 🧪 Testing Strategy

- Manual verification: pie chart with `count() AS total_requests` now renders correctly; `count() AS __result_0` continues to work.
- Edge cases covered: mixed query (`__result_0` + custom alias), legacy aliases, string-only group columns.

---

- Blast radius: `readAsScalar` only — time-series and raw query paths are untouched.
- Potential regressions: queries that intentionally used a numeric column as a group dimension will now be classified as aggregation. This is an edge case unlikely in practice (scalar queries for pie/table panels).
- Rollback plan: revert the 5-line addition to `consume.go`.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Pie charts and scalar panels now render correctly when ClickHouse queries use custom aggregation column aliases. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

The change is 5 lines in `readAsScalar`. The key insight: `readAsTimeSeries` already had this fallback logic but `readAsScalar` never did. The fix brings them in line with each other.

---

### DIFF
--- a/pkg/querier/consume.go
+++ b/pkg/querier/consume.go
@@ -292,6 +292,11 @@ func readAsScalar(rows driver.Rows, queryName string) (*qbtypes.ScalarData, erro
 		colType := qbtypes.ColumnTypeGroup
 		if aggRe.MatchString(name) {
 			colType = qbtypes.ColumnTypeAggregation
+		} else if slices.Contains(legacyReservedColumnTargetAliases, name) {
+			colType = qbtypes.ColumnTypeAggregation
+		} else if numericKind(colTypes[i].ScanType().Kind()) {
+			// Custom alias or unnamed aggregation: infer from numeric type.
+			colType = qbtypes.ColumnTypeAggregation
 		}
 		cd[i] = &qbtypes.ColumnDescriptor{
 			TelemetryFieldKey: telemetrytypes.TelemetryFieldKey{Name: name},


krsoninikhil:
Closing in favor of https://github.com/SigNoz/signoz/pull/11593

PR #11263, opened 2026-05-12
feat(alertmanager): add support for google chat alert channel

niladrix719:
## Pull Request

---

### 📄 Summary

Adds support for Google Chat as an alert notification channel.

- UI form for Google Chat configuration (`webhook URL, title, text, send_resolved`)
- Create/edit/test API calls for Google Chat channels
- Google Chat notifier implementation with webhook support
- Threading Support, Alerts with the same ruleId are grouped together

Added Docs for it here - https://github.com/SigNoz/signoz.io/pull/3273

#### Screenshots / Screen Recordings (if applicable)

Google chat Alert Channel

Screenshot 2026-05-12 at 1 28 03 PM

Google chat Alert Message

Screenshot 2026-05-12 at 5 21 08 PM

#### Issues closed by this PR

Closes #5095 

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature

---

- Tests added/updated: yes
- Manual verification: yes
- Edge cases covered: yes

---

### 📋 Checklist
- [x] Tests added or explicitly not required
- [x] Manually tested

---

PR #11473, opened 2026-05-27
chore(alertmanager): support custom receiver configs

srikanthccv:
## Pull Request

There were roughly three options and only one reasonable option to start supporting custom receivers.

1. copy the `alertmanager/config` -> this would bring the types to our codebase and easy to add new configs to `Receiver` struct. However, this would mean updating all the references where `config.Receiver` used, which then lead to copying the most of the upstream codebase function by function. Hence, discarded this as first option.
2. fork the alertamanger and use replace directive in go.mod. Discard as an option.
3. extend the `config.Receiver` and adjust it's usage to ensure the roundtrips to DB preserve the configs that upstream doesn't carry. Follow this approach.

Unblocks https://github.com/SigNoz/signoz/pull/11263 and upcoming custom receivers.


pkg/sqlmigration/014_add_alertmanager.go
srikanthccv, line 1: should I being old type copy here instead of updating?

pkg/types/alertmanagertypes/receiver_test.go
```diff
@@ -39,3 +45,27 @@ func TestNewReceiver(t *testing.T) {
 		})
 	}
 }
+
+// Omitted fields fall back to DefaultGoogleChatReceiverConfig.
+func TestNewReceiverGoogleChatAppliesDefaults(t *testing.T) {
```
therealpandey, line 50: This can be merged with the `TestNewReceiver` with a test case `GoogleChatConfig_Defaults`.

pkg/types/alertmanagertypes/receiver_test.go
```diff
@@ -39,3 +45,27 @@ func TestNewReceiver(t *testing.T) {
...
+	require.Len(t, receiver.GoogleChatConfigs, 1)
+
+	got := receiver.GoogleChatConfigs[0]
+	assert.Equal(t, DefaultGoogleChatReceiverConfig.Title, got.Title, "Title should fall back to the default template")
+	assert.Equal(t, DefaultGoogleChatReceiverConfig.Text, got.Text, "Text should fall back to the default template")
+	assert.Equal(t, DefaultGoogleChatReceiverConfig.VSendResolved, got.SendResolved(), "send_resolved should fall back to the default")
+}
+
+// User-specified values override defaults.
+func TestNewReceiverGoogleChatPreservesUserOverrides(t *testing.T) {
```
therealpandey, line 62: This can be merged with the TestNewReceiver with a test case GoogleChatConfig_UserOverride.

pkg/sqlmigration/014_add_alertmanager.go
therealpandey: Yes let's bring the old type here instead of updating.

pkg/types/alertmanagertypes/config.go
```diff
@@ -59,12 +59,43 @@ type Config struct {
 
 	// storeableConfig is the representation of the config in the store
 	storeableConfig *StoreableConfig
+
+	// customConfigs holds the custom notifier configs upstream's
+	// config.Receiver cannot carry, keyed by receiver name.
+	customConfigs map[string]customReceiverConfigs
```
therealpandey, line 65: Instead of having a `customConfigs`, shall we just subclass the entire alertmanager config now?

I'm guessing it's not that easy @srikanthccv?

therealpandey (changes requested):

pkg/types/alertmanagertypes/config.go
srikanthccv: right, I considered that as an option, however, it was touching every surface area and becoming too big of a change.

pkg/types/alertmanagertypes/receiver_test.go
srikanthccv: I would have to add conditional block to access specifically `receiver.GoogleChatConfigs` based on the test case, so it seemed better to be a separate test. Let me know If I misunderstood suggestion.

therealpandey: Can ignore for now, non-blocker.

pkg/sqlmigration/014_add_alertmanager.go
therealpandey: We are good after this.

srikanthccv: done

therealpandey (approved):

PR #11480, opened 2026-05-27
chore: remove cloud integration service cascade delete constraint

swapnil-signoz:
## Pull Request

Removing cascade delete constraint on cloud_integration_service table

---

### 📄 Summary
Clean up

### Closes
https://github.com/SigNoz/engineering-pod/issues/4257


pkg/sqlmigration/091_ci_remove_cascade_delete.go
```diff
@@ -0,0 +1,73 @@
...
+	defer func() {
+		_ = tx.Rollback()
+	}()
+
+	oldTable, uniqueConstraints, err := migration.sqlschema.GetTable(ctx, sqlschema.TableName("cloud_integration_service"))
+	if err != nil {
+		return err
+	}
+
+	newTable := *oldTable
```
therealpandey: I got confused by this. The oldTable should already have foreign key references without the cascade?

therealpandey (changes requested):

pkg/sqlmigration/091_ci_remove_cascade_delete.go
swapnil-signoz: No, old table HAS the cascade delete fk reference which we want to delete. New table creates the new fk reference without the cascade delete.

therealpandey: The operator does not have support for reading cascade constraints.

Can you check what the create table for the oldtable is?

swapnil-signoz:
```sql
CREATE TABLE "cloud_integration_service" ("id" text NOT NULL, "created_at" TIMESTAMP, "updated_at" TIMESTAMP, "type" text NOT NULL, "config" text, "cloud_integration_id" text NOT NULL, PRIMARY KEY ("id"), CONSTRAINT "cloud_integration_id_type" UNIQUE ("type", "cloud_integration_id"), FOREIGN KEY ("cloud_integration_id") REFERENCES "cloud_integration" ("id") ON DELETE CASCADE);
```

This is the current DDL for the table (before migration)

therealpandey (approved):

PR #11483, opened 2026-05-27
fix: ensure timestamp is always in ms

nityanandagohain:
## Pull Request

---

### 📄 Summary
In our `QueryRangeRequest` we say that `Start` and `End` is milliseconds, but we don't enforce it anywhere. If someone sends something different i.e seconds or microseconds we don't reject it nor correct it. We do have something for logs where it converts to nanoseconds to query from clickhouse. But not everywhere.

So the logs view in trace details was sending microseconds and because the logic relied on it to be in milliseconds it broke.

This fix ensures that regardless of whatever the client sends it is always in milliseconds.

#### Issues closed by this PR
Fixes https://github.com/SigNoz/engineering-pod/issues/5123

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
Timestamp from the client is in microseconds

#### Fix Strategy
We are converting start and and to milliseconds

---

- Tests added/updated:
- Manual verification:
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

### 📋 Checklist
- [x] Manually tested

---


tushar-signoz (approved):
LGMT 
We should re-visit this, as to why we haven't enforced for a timestamp format and why downstream has to handle it. Also Why pick millisecs and not nano ?

therealpandey (approved):

PR #11490, opened 2026-05-29
fix: do not close suggestion on typing period

tushar-signoz:
## Pull Request

---

Trailing `.` is treated as a value instead of a key and hence no suggestions.
The suggestion box closes as soon as on types `.`. For example typing `service` will suggest auto-complete but after `serice.` it will close itl   

Before

https://github.com/user-attachments/assets/9144d840-7044-4cac-aad5-7b5a7775acc0

After

https://github.com/user-attachments/assets/34443c04-3a04-4cd8-a40b-f7efbbe48e86

### DIFF
--- a/frontend/src/utils/__tests__/queryContextUtils.test.ts
+++ b/frontend/src/utils/__tests__/queryContextUtils.test.ts
@@ -3,6 +3,7 @@ import {
 	extractQueryPairs,
 	getCurrentQueryPair,
 	getCurrentValueIndexAtCursor,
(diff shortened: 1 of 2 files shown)


H4ad:
Duplicated of #10945?

tushar-signoz:
@H4ad The changes are non-overlapping. Created a separate PR so that its easier to review.

ahrefabhi (approved):

aks07 (approved):

PR #11492, opened 2026-05-29
fix: badge label color in logs

tewarig:
Issue - 

https://signoz-team.slack.com/archives/C02TJ466H8U/p1779970190359879 

image

Root cause - 

  @signozhq/ui/badge defaults to color="primary", which the component's colorMap resolves to robin. Its stylesheet then sets, directly on the  The TagContainer styled wrapper only override border-color, border-radius, font-weight, font-size, line-height — it did not override
  
  background-color or color. So every field pill rendered as a solid robin-blue background with white text inherited by TagLabel (which has poor contrast against robin-blue.)
  
  have added. color: var(--foreground)
  
  which fixes this issue 

image
 
image

### DIFF
--- a/frontend/src/container/LogDetailedView/FieldRenderer.styles.ts
+++ b/frontend/src/container/LogDetailedView/FieldRenderer.styles.ts
@@ -12,6 +12,7 @@ export const TagContainer = styled(Badge)`
 `;
 
 export const TagLabel = styled.span`
+	color: var(--foreground);
 	font-weight: 400;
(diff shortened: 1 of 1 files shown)


H4ad (approved):
Can you create a follow-up issue for this? The whole file can be rewriten to use our semantic tokens

tewarig:
sure

tewarig:
done https://github.com/SigNoz/engineering-pod/issues/5150

PR #11495, opened 2026-05-29
feat(web): add support for sentry and pylon

vikrantgupta25:
### 📄 Summary

- add support for sentry and pylon

contributes to: https://github.com/SigNoz/platform-pod/issues/2144


H4ad (changes requested):
Needs to update `vite.config.ts` and include pylon/sentry, otherwise it will break on dev if we access it.

therealpandey (approved):

SagarRajput-7:
> Needs to update `vite.config.ts` and include pylon/sentry, otherwise it will break on dev if we access it.

added - https://github.com/SigNoz/signoz/pull/11498/changes#diff-97c6e2c429ec483132c584137a18a1ade2ff6c3f4f6485f4a83db604d0323d7c

please check if you were talking about this.

H4ad (approved):

PR #11497, opened 2026-05-29
feat: adding get cloud integration service for account handler

swapnil-signoz:
## Pull Request

Adding endpoint `/api/v1/cloud_integrations/{cloud_provider}/accounts/{id}/services/{service_id}` and respective handler. This endpoint was missing in previous implementation (consciously), this provides better visibility into hierarchy for cloud integrations and it's domain models. 

PS: Older API remains where cloud_integration_id was passed as query param remains as is and no changes in module level as well, both the older and this new endpoint call same module method

---

### 📄 Summary
Following REST API conventions of having right hierarchy visible in the endpoint itself
What problem does it solve, and why is this the right approach?
Clean up task

#### Issues closed by this PR
https://github.com/SigNoz/engineering-pod/issues/4616

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only


therealpandey:
@swapnil-signoz are we planning to remove the older endpoint?

swapnil-signoz:
> @swapnil-signoz are we planning to remove the older endpoint?

No @therealpandey, we will be keeping the older one as it is still required to get service details when no accounts are connected. We might remove the query param from the older API later. Whats our SOP to convey breaking API changes to users (as people use API for different reasons via various ways sometimes like terraform, automation)?

therealpandey (dismissed):

therealpandey (dismissed):

therealpandey (approved):

H4ad (approved):

PR #11500, opened 2026-05-29
refactor(planned-maintenance): remove time bounds from recurrence

jatinderjit:
### 📄 Summary

1. Planned maintenance stored start/end bounds twice: on the `Schedule` and again on its nested `Recurrence`. This makes `Schedule.StartTime` and `Schedule.EndTime` the single source of truth, and removes the corresponding fields from `Recurrence`.
2. Downtime APIs can now handle the embedded timezones in the serialized timestamps.
3. Simplify `IsActive`/`IsUpcoming`

#### Issues closed by this PR

- Closes https://github.com/SigNoz/engineering-pod/issues/5042

---

### ✅ Change Type

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🧪 Testing Strategy

- Reworked `TestShouldSkipMaintenance` onto the new schema; added `TestIsActiveFixedSchedule` and `TestIsActiveRecurringSchedule`.
- Manual: create/edit fixed + recurring maintenances; confirm an existing recurring maintenance still silences alerts after the migration runs.

---

### Risk & Impact Assessment

- **Blast radius:** planned-maintenance silencing only.
- **Regressions:** API payload changed, `recurrence.startTime`/`endTime` removed,
  `schedule.startTime` now required; direct API clients must update.
- **Rollback:** revert the PR.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Description | Planned maintenance now uses a single start/end on the schedule and handles DST correctly for recurring windows. API: `recurrence.startTime`/`endTime` removed, `schedule.startTime` required. |

---

### 📋 Checklist
- [x] Manually tested

---

### 👀 Notes for Reviewers

From https://github.com/SigNoz/engineering-pod/issues/5042

> remove the offset in the request and response payload and use UTC

The offset in the request payload still exists, as removing that would be an extra (and redundant) step.

The backend doesn't ignore the offset in the payload anymore. The timezone/offset won't matter as long as proper timezone conversions were done. The API would now handle both of these in the payload as exactly the same: `2026-06-02T00:00:00.0+0530` and `2026-06-01T18:30:00.0Z`.


frontend/src/container/PlannedDowntime/PlannedDowntimeForm.tsx
```diff
@@ -152,6 +152,11 @@ export function PlannedDowntimeForm(
 
 	const saveHandler = useCallback(
 		async (values: PlannedDowntimeFormData) => {
+			const { startTime, timezone } = values;
+			if (!startTime || !timezone) {
+				// unreachable
```
AshwinBhatkal: This piece of code is unreachable at any time you mean?

AshwinBhatkal: If yes, then we will have to update our types accordingly

jatinderjit: Not at any time, but at form submission. These are the required fields.

jatinderjit: updated the comment with justification

jatinderjit:
From https://github.com/SigNoz/engineering-pod/issues/5042

> remove the offset in the request and response payload and use UTC

I didn't remove the offset in the request payload. Doing that would be an extra (redundant) step.

But the backend doesn't ignore the timezone details anymore. The offset in the payload won't matter as long as proper timezone conversions were done. The API would now handle both of these in the payload as exactly the same: `2026-06-02T00:00:00.0+0530` and `2026-06-01T18:30:00.0Z`.

frontend/src/container/PlannedDowntime/PlannedDowntimeForm.tsx
AshwinBhatkal: If these are required fields, then they should not have been typed as being optional. Can you check if their type can be made required? We can figure it out together if you want. Lmk

frontend/src/container/PlannedDowntime/PlannedDowntimeForm.tsx
```diff
@@ -200,25 +205,17 @@ export function PlannedDowntimeForm(
...
-						repeatOn: recurrence.repeatOn,
-						repeatType: recurrence.repeatType,
-					};
-
-		await saveHandler({
-			...values,
-			recurrence: recurrenceData,
-		});
+		const rec = values.recurrence;
+		const recurrence =
```
AshwinBhatkal, line 209: Somehow liked the previous naming itself better. Hehe

frontend/src/container/PlannedDowntime/PlannedDowntimeForm.tsx
jatinderjit: Those fields are already required in the type used to post the data ([ref](https://github.com/SigNoz/signoz/blob/ce64ca93f78379f83f685a93d975250bf883c258/frontend/src/api/generated/services/sigNoz.schemas.ts#L1718-L1718)).
They're optional in the form type to allow for empty input fields (initial state and when user clears the field). `startTime` is of type dayjs, so can't have a blank string as the default.
Do you think this is fine? Or can this be improved?

frontend/src/container/PlannedDowntime/PlannedDowntimeForm.tsx
jatinderjit: You're right! 😄
Will change it next time I touch this code.

AshwinBhatkal (dismissed):

therealpandey:
@jatinderjit please bump the migration number to 94

pkg/types/alertmanagertypes/schedule.go
```diff
@@ -93,50 +76,26 @@ func (s *Schedule) UnmarshalJSON(data []byte) error {
...
 	}
 
-	var startTime time.Time
-	if aux.StartTime != "" {
-		startTime, err = time.Parse(time.RFC3339, aux.StartTime)
-		if err != nil {
-			return err
-		}
-		s.StartTime = time.Date(startTime.Year(), startTime.Month(), startTime.Day(), startTime.Hour(), startTime.Minute(), startTime.Second(), startTime.Nanosecond(), loc)
+	startTime, err := time.Parse(time.RFC3339, aux.StartTime)
```
srikanthccv, line 83: Not the change in this PR but this time.Parse(time.RFC3339, "") when the start time is absent and returns the raw error. and render.Error maps untyped errors to 500. The "missing start time in the payload" is a dead in a sense that it never reaches because the json marshal fails. Since there is no other way to create planned maintenance than API, let's remove the `Validate` and have the Unmarshal returns the clear errors.

pkg/types/alertmanagertypes/schedule.go
```diff
@@ -93,50 +76,26 @@ func (s *Schedule) UnmarshalJSON(data []byte) error {
 		return err
```
srikanthccv: Similarly here as well. There are number of untyped errors here. Let's make them return `errors.Newf(errors.TypeInvalidInput, ...)` from `UnmarshalJSON` only

pkg/types/alertmanagertypes/schedule.go
```diff
@@ -11,7 +11,7 @@ import (
 
 type Schedule struct {
```
srikanthccv, line 12: There is a migration which references this `alertmanagertypes.Schedule`. Please copy the pre-this PR types into that migration instead of using them.

As a general rule, may be, we should say migrations must not use evolving domain types at all. And must re-declare/define them on the migrations only.

pkg/types/alertmanagertypes/schedule.go
```diff
@@ -11,7 +11,7 @@ import (
...
 	Timezone   string      `json:"timezone" required:"true"`
-	StartTime  time.Time   `json:"startTime,omitempty"`
+	StartTime  time.Time   `json:"startTime" required:"true"`
 	EndTime    time.Time   `json:"endTime,omitzero"`
```
srikanthccv, line 15: Let's make it explicit (in api docs) that no end time means open schedule forever. Does the front-end require end time for fixed schedule now?

pkg/types/alertmanagertypes/maintenance.go
```diff
@@ -304,16 +254,17 @@ func (m *PlannedMaintenance) checkWeekly(currentTime time.Time, rec *Recurrence,
...
+	refDay := startTime.Day()
 	year, month, _ := currentTime.Date()
 	lastDay := time.Date(year, month+1, 0, 0, 0, 0, 0, loc).Day()
 	day := refDay
 	if refDay > lastDay {
 		day = lastDay
 	}
 	candidate := time.Date(year, month, day,
-		rec.StartTime.Hour(), rec.StartTime.Minute(), rec.StartTime.Second(), rec.StartTime.Nanosecond(),
+		startTime.Hour(), startTime.Minute(), startTime.Second(), startTime.Nanosecond(),
```
srikanthccv, line 258: the `startTime.Hour()` in `checkX` assume that the `Schedule.StartTime` location is the schedule timezone. Which is true after deserialization. However, let's `StartTime.In(loc).Hour()` would make it work regardless of how they are constructed. You can see the tests try to do this by hand.

pkg/types/alertmanagertypes/schedule.go
jatinderjit: Done

pkg/types/alertmanagertypes/schedule.go
jatinderjit: Done

pkg/types/alertmanagertypes/schedule.go
jatinderjit: End time is required right now. Those changes are in another PR. I'll add to API docs there.

pkg/types/alertmanagertypes/maintenance.go
jatinderjit: The `checkX` methods are called from one place only, and we're ensure the timezone is correct before calling these methods ([ref](https://github.com/SigNoz/signoz/blob/354c6416afc606897f06010a956a7fd4da313887/pkg/types/alertmanagertypes/maintenance.go#L182)).
Do you want me to move that conversion to these methods?

jatinderjit: Done

srikanthccv (approved):

therealpandey (approved):

PR #11506, opened 2026-05-29
fix(telemetrystore): fix clickhouse connection-pool slot leak (acquire conn timeout)

therealpandey:
## Problem

Since v0.126.0, production hits `clickhouse: acquire conn timeout. you can increase the number of max open conn or the dial timeout`. It surfaces first in **rule state history** (the alerting engine runs those queries continuously with cancellable contexts). This is a connection-pool **slot leak**, not a misconfiguration — once `MaxOpenConns` slots leak, every query blocks for `DialTimeout` then errors.

## Root cause

PR #11475 (OpenFGA upgrade) transitively bumped `clickhouse-go/v2` **v2.40.1 → v2.43.0**. v2.43.0 rewrote the connection pool and shipped two slot-leak bugs, both triggered by **context cancellation**:

- **acquire() slot leak** ([ClickHouse/clickhouse-go#1759](https://github.com/ClickHouse/clickhouse-go/pull/1759)) — after taking an `ch.open` slot, if `ch.idle.Get(ctx)` returns a cancellation error the slot is never released. Affects every query path (`Query`/`QueryRow`/`Select`/`Exec`/`PrepareBatch` all call `acquire()`).
- **batch.Close() leak** ([ClickHouse/clickhouse-go#1795](https://github.com/ClickHouse/clickhouse-go/pull/1795)) — when `closeQuery()` fails on a cancelled context, the connection is never returned. Hits the rule-state-history `PrepareBatch` write path.

Both are fixed in **clickhouse-go v2.44.0**.

## Changes

1. **Upgrade `clickhouse-go/v2` v2.43.0 → v2.44.0** (the fix).
2. v2.44.0 adds `HasData()` to the `driver.Rows` interface, which the old test mock didn't implement. **Swap the mock to `github.com/SigNoz/clickhouse-go-mock` v0.14.0** (SigNoz/clickhouse-go-mock#4), which implements `HasData()` and tracks v2.44.0.
3. **Emit connection-pool metrics** — new OTel observable gauges `signoz.telemetrystore.connection.{open,idle,max_open,max_idle}` from `driver.Stats()`, so pool saturation/leaks are visible in Prometheus (plot `open` vs `max_open`).

## Verification

- `go build ./...` ✅
- `go vet ./pkg/telemetrystore/...` ✅
- `go test ./...` for all mock-consuming packages (telemetrystore, querier, telemetrymetadata, clickhouseprometheus, rules, ee rules) ✅

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -89,8 +89,8 @@ github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
 github.com/ClickHouse/ch-go v0.71.0 h1:bUdZ/EZj/LcVHsMqaRUP2holqygrPWQKeMjc6nZoyRM=
 github.com/ClickHouse/ch-go v0.71.0/go.mod h1:NwbNc+7jaqfY58dmdDUbG4Jl22vThgx1cYjBw0vtgXw=
-github.com/ClickHouse/clickhouse-go/v2 v2.43.0 h1:fUR05TrF1GyvLDa/mAQjkx7KbgwdLRffs2n9O3WobtE=
-github.com/ClickHouse/clickhouse-go/v2 v2.43.0/go.mod h1:o6jf7JM/zveWC/PP277BLxjHy5KjnGX/jfljhM4s34g=
+github.com/ClickHouse/clickhouse-go/v2 v2.44.0 h1:9pxs5pRwIvhni5BDRPn/n5A8DeUod5TnBaeulFBX8EQ=
+github.com/ClickHouse/clickhouse-go/v2 v2.44.0/go.mod h1:giJfUVlMkcfUEPVfRpt51zZaGEx9i17gCos8gBl392c=
 github.com/Code-Hex/go-generics-cache v1.5.1 h1:6vhZGc5M7Y/YD8cIUcY8kcuQLB4cHR7U+0KMqAA0KcU=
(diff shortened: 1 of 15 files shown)


tushar-signoz (approved):
LGTM !

go.sum
```diff
@@ -1062,8 +1064,8 @@ github.com/spf13/pflag v1.0.10/go.mod h1:McXfInJRrz4CZXVZOBLb0bTZqETkiAhM9Iw0y3A
 github.com/spf13/viper v1.13.0/go.mod h1:Icm2xNL3/8uyh/wFuB1jI7TiTNKp8632Nwegu+zgdYw=
 github.com/spf13/viper v1.20.1 h1:ZMi+z/lvLyPSCoNtFCpqjy0S4kPbirhpTMwl8BkW9X4=
 github.com/spf13/viper v1.20.1/go.mod h1:P9Mdzt1zoHIG8m2eZQinpiBjo6kCmZSKBClNNqjJvu4=
-github.com/srikanthccv/ClickHouse-go-mock v0.13.0 h1:/b7DQphGkh29ocNtLh4DGmQxQYA0CfHz65Wy2zAH2GM=
-github.com/srikanthccv/ClickHouse-go-mock v0.13.0/go.mod h1:LiiyBUdXNwB/1DE9rgK/8q9qjVYsTzg6WXQ/3mU3TeY=
+github.com/srikanthccv/ClickHouse-go-mock v0.12.0 h1:KUzaWTwuqMc2uf5FylM/oAcTFdE2DdZjvISm9V0/NAA=
+github.com/srikanthccv/ClickHouse-go-mock v0.12.0/go.mod h1:1oUmLtXEXOyS0EEWVKlKEfLfv9y02agCMAvD3tVnhlo=
```
tushar-signoz, line 1068: Wouldn't this create issues ?

therealpandey: It's imported by signoz-otel-collector which is imported here.

PR #11511, opened 2026-05-29
fix: allow editing of notification channel names

rossigee:
## Summary

- Removes the `disabled={editing}` attribute from the channel name field in `FormAlertChannels`, allowing users to edit the name of an existing notification channel.

## Problem

When editing an existing notification channel, the **Name** field was disabled, preventing users from renaming it. This appears to be unintentional — all other fields remain editable.

## Test plan

- [ ] Create a notification channel
- [ ] Navigate to edit the channel
- [ ] Verify the Name field is now editable
- [ ] Save and confirm the updated name persists


H4ad:
Hey @rossigee, thanks for the PR and sorry for taking so long to review this PR!

We currently don't allow renaming notification channels since it's used by alerts as the reference for the notification channel. By allowing the rename on UI, we could break existing alerts that are assigned to this channel.

I'm closing this PR but feel free to create an issue for a feature request to rename channels if this is something that could be useful for you!

PR #11514, opened 2026-05-31
fix: redirect to login when invite/reset-password token is invalid or expired

### LINKED ISSUE #7011: handle invalid invitation token gracefully in signup page

sridhar-3009:
## Problem

Resolves #7011.

When a user followed an expired, already-used, or tampered invitation link, the reset-password form was displayed unconditionally. The error was only shown **after** the user completed and submitted the form — a confusing experience.

## Fix

Use the existing `/api/v2/reset_password_tokens/verify` endpoint to validate the token on mount, **before** rendering the form:

- If validation succeeds → show the form as normal.
- If validation fails (expired / used / invalid) → show an error notification and redirect to the login page immediately.
- While validation is in progress → render `null` (brief flash, no layout shift).

## Before / After

**Before:** User opens stale invite link → sees the full reset-password form → fills it out → submits → sees "invalid token" error.

**After:** User opens stale invite link → brief loading → error notification "Invalid or expired invitation token. Please contact your admin for a new invite link." → redirected to login page.

## Changes

`frontend/src/container/ResetPassword/index.tsx`:
- Import `verifyResetPasswordToken` from generated API.
- Add `tokenValid` state (`null | boolean`).
- In `useEffect`: call `verifyResetPasswordToken({ token })` after confirming the token is present; redirect to login with an error notification on failure.
- Guard the JSX render behind `tokenValid === true`.

### DIFF
--- a/frontend/src/container/ResetPassword/index.tsx
+++ b/frontend/src/container/ResetPassword/index.tsx
@@ -7,6 +7,7 @@ import { Form, Input as AntdInput } from 'antd';
 import { Typography } from '@signozhq/ui/typography';
 import { Logout } from 'api/utils';
 import resetPasswordApi from 'api/v1/factor_password/resetPassword';
+import { verifyResetPasswordToken } from 'api/generated/services/users';
 import AuthError from 'components/AuthError/AuthError';
 import AuthPageContainer from 'components/AuthPageContainer';
 import ROUTES from 'constants/routes';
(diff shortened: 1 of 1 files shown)


SagarRajput-7:
Hey @sridhar-3009,

Thanks for taking the time to work on this and for the contribution! 🙌

Unfortunately, I'll have to close this PR as we've merged a different implementation for the same issue.

Our implementation also integrates `/api/v2/reset_password_tokens/verify`, but instead of redirecting users to the login page, it shows a dedicated error screen to better communicate to users what happened.

Please take a look: https://github.com/SigNoz/signoz/pull/11522

Really appreciate the effort you put into this and your contribution to SigNoz. Looking forward to seeing more contributions from you! 👏

PR #11426, opened 2026-05-22
feat(logs/traces): streamline column state — selectColumns becomes canonical source

aks07:
## Pull Request

---

### 📄 Summary

Phase 2 of the column-selection revamp ([#4279](https://github.com/SigNoz/engineering-pod/issues/4279), sub-issue [#5061](https://github.com/SigNoz/engineering-pod/issues/5061)). Makes `selectColumns` (in preferences) the single source of truth for column membership AND order in both explorers.

---

### Before this PR — Phase 1 (shipped in [#11360](https://github.com/SigNoz/signoz/pull/11360))

- Promoted `FieldsSettings` (forked from `QuickFilterSetting` during trace-details v3) → shared `frontend/src/components/FieldsSelector/`.
- Migrated to `TelemetryFieldKey` shape + V5 `useGetQueryKeySuggestions` discovery hook.
- Migrated trace-details v3 store + 3 consumers (`TraceFlamegraph`, `useFlamegraphHover`, `SpanHoverCard`) to `TelemetryFieldKey` end-to-end. Added a read-side compatibility shim in `derivePreviewFields` for legacy persisted user prefs.
- 8 unit tests on the legacy-pref migration shim.

---

#### Why

- Column state was split across 2–3 stores per explorer (preferences `selectColumns`, TanStack `hiddenColumnIds` / `columnOrder`, traces' `useDragColumns` localStorage). Picker, table-drag, table-X wrote to different places.
- `body` + `timestamp` could silently stick hidden after a logs table-X click — they're spread from defaults so the auto-cleanup never saw them.
- A version-gated one-time migration tried to enforce body/timestamp presence; the gate let users drift out of compliance after first load.
- Phase 3 (picker UX swap) assumes one canonical source for membership + order.

#### Logs

- `selectColumns` now holds `body` + `timestamp` directly. Always-on `ensureLogsRequiredColumns` invariant at:
  - loader (`usePreferenceLoader.logsPreferencesLoader`)
  - writer (`logsUpdaterConfig.updateColumns`, DIRECT + SAVED_VIEW)
  - saved-view branch (`usePreferenceSync` LOGS path)
- `body` + `timestamp` column defs locked with `canBeHidden: false` + `enableRemove: false` (drag still allowed). `useColumnState`'s internal filter strips any legacy `hiddenColumnIds` entries.
- Drop `useHiddenColumnIds` + `syncedSelectedColumns` filter + reconciliation `useEffect` from `LogsExplorerList` and `LiveLogsList`.
[...]

Issue #4279, opened 2023-12-22
Options to enable disable alerts in bulk


nityanandagohain:
It will be a nice feature to have though which we can select multiple alerts and enable/disable them

cc @pranay01 @YounixM

al-mcintosh:
Would really appreciate this functionality that would allow to mute alerts especially during maintenance, etc.  

Any one of the following would be helpful:

- Disable Action for channels in the `/settings/channels` tab. 
OR
- Disable All Action for all alerts in the `/alerts` tab.

PR #11516, opened 2026-05-31
feat(logs/traces): migrate column picker UI to FieldsSelector

aks07:
## Summary

Wires the shared `FieldsSelector` panel into the user-facing column pickers in **Logs Explorer**, **Trace Explorer**, and **LiveLogs**. Save / Discard, search, drag-reorder, lockable required fields. The legacy `OptionsMenu` / `AddColumnField` picker is no longer reachable from those flows; the code stays compiled (zero consumers) and is cleaned up in a follow-up PR.

**Stacked on [#11426](https://github.com/SigNoz/signoz/pull/11426)** — that PR collapses logs/traces column state into `selectColumns` as the canonical source for membership + order, which this PR builds on. Diff against `feat/field-selector-migration`, not `main`.

Tracks [#4279](https://github.com/SigNoz/engineering-pod/issues/4279).

---

### Picker UX

`FieldsSelector` (already used by trace-details v3 for preview-fields) now drives column selection across all three explorers — same panel, same Save / Discard, same drag-to-reorder.

- New optional `requiredFields?: readonly string[]` prop. When a field's `name` matches, its Remove button is hidden; drag still works. Logs + LiveLogs pass `LOGS_REQUIRED_COLUMNS = ['timestamp', 'body']`. Trace passes nothing (no required fields today).
- Row height pinned to `36px` so the rows look identical whether the Remove button is hidden or not.
- `OtherFields` loader skeletons now stack as full-width rows instead of a wrapping inline grid.

### Logs Explorer

- `LogsActionsContainer` owns the picker-open state and renders `<FieldsSelector>` **outside** the popover host. (antd `destroyTooltipOnHide` unmounts portal content on outside-click; `FieldsSelector`'s `FloatingPanel` portals to `document.body`, so a click inside the panel reads as "outside the popover" and would destroy mid-edit.)
- `LogsFormatOptionsMenu` becomes format-only (Font Size, Format, Max lines). The inline column list + "Add New Column" search container + keyboard nav + scroll effect are removed. An "Edit columns >" row appears only when format = Column; click fires a new `onOpenColumns` callback and closes the popover.
- `useLogsTableColumns` rewritten — fields now render in `selectColumns` order with `body` / `timestamp` inline at their position (the hook previously hardcoded `[stateIndicator, timestamp, ...userFields, body]`, ignoring `selectColumns` order for body/timestamp). The unused `appendTo` prop dropped. `stateIndicator` still pinned at index 0.
- `<TanStackTable respectColumnOrder={false}>` opt-out — fixes a picker-reorder-vs.-table-render divergence: picker writes to `selectColumns`, but TanStack's zustand `columnOrder` slot stayed stale, and `useColumnState.sortedColumns` was re-sorting by that stale slot on every render.

### Trace Explorer

- Drops `<OptionsMenu>` + `<Popover>` entirely. The trace gear menu held nothing else (no font size / format / max lines), so the popover host disappears with it.
- Trigger is now a plain "Options" + gear `<div>`; `<FieldsSelector>` renders as a sibling.
- Legacy `styles.ts` (styled-components) deleted; replaced with `Controls.module.scss`.

### LiveLogs

- Mirrors the logs-explorer picker pattern: open-state + `<FieldsSelector>` outside the popover host + `requiredFields={LOGS_REQUIRED_COLUMNS}` + `onOpenColumns` threaded into `LogsFormatOptionsMenu`.
- `<TanStackTable respectColumnOrder={false}>` for the same reason logs needs it — LiveLogs shares the `LOGS_LIST_COLUMNS` zustand storage key with the logs explorer, so without the opt-out it inherits the same picker/table mismatch.

### Shared TanStack — `respectColumnOrder` prop

New optional prop on `<TanStackTable>` (default `true` → other consumers untouched). When `false`:
- `useColumnState.sortedColumns` skips the zustand `columnOrder` override and returns `columns` directly.
- `useColumnHandlers.handleColumnOrderChange` skips the `storeSetOrder` write.

K8s + every existing consumer keeps current behavior. LiveLogs + Logs Explorer opt in to `false` explicitly.

### Hook plumbing — additive

- `OptionsMenuConfig` gains an additive `fieldsSelector?: { value, onFieldsChange }` slot. The old `addColumn` config + handlers remain for now — table-side X / drag-reorder callers still consume them.

---

## Screenshot

https://github.com/user-attachments/assets/f2c0eb0a-6c8b-4d49-b01b-3280aaecd8f2

https://github.com/user-attachments/assets/eab2f1e8-2fec-4af8-8729-5a194258a598

## Risk & Impact

- **Blast radius:** Logs Explorer + LiveLogs + Trace Explorer list views; saved-view path for both. K8s tables / CeleryOverviewTable unaffected (`respectColumnOrder` defaults to `true`).
- **Potential regressions:**
  - Users with prior column drag history on logs / LiveLogs lose the persisted `columnOrder` (now ignored on read). Cosmetic; first interaction re-establishes order via `selectColumns`. Same regression class accepted in the base PR.
  - Fields with the same `name` but different `fieldContext` (e.g. `service.name` as both a resource attribute and a log attribute) now appear as separate rows in OTHER FIELDS. The old `<antd Select>` picker silently collapsed them and picked one variant on add. Both rows are visible now; addressed properly in a follow-up.
- **Rollback plan:** revert the PR. Legacy persisted prefs continue to load. The `addColumn` config + handlers are still alive on the old hook, just unused.

---

## Testing

**New unit tests (15):**
- `useColumnState`: `respectColumnOrder=false` skips re-sort; `=true` (default) preserves existing behavior.
- `AddedFields`: `requiredFields` hides Remove; non-required keep Remove; names still render for required fields; exact-name match (no substring).
- `LogsFormatOptionsMenu`: "Edit columns" row visibility (format gate + `onOpenColumns` gate); click fires callback + closes popover.
- `useLogsTableColumns`: fields render in array order; body/timestamp inline at their position; `id` skipped; body/timestamp keep `canBeHidden: false` + `enableRemove: false`; empty `fields` renders only stateIndicator.

**Existing suites pass unchanged:** TanStackTable (126 tests), `LogsFormatOptionsMenu` (3), TracesExplorer ListView (6), `ensureLogsRequiredColumns` (7). Aggregate sweep: 12 suites, 157 tests, 0 failures.

**Manual verification:**
- **Logs:** open gear → "Edit columns >" appears only when format = Column → click → FieldsSelector opens outside the popover → add column → Save → column appears, persists across reload, `?options=` URL updated. Remove via table X → gone. Drag column header → picker reflects new order on next open. Reorder in picker → Save → table reflects.
- **Trace:** click "Options" → FieldsSelector opens. Same add/remove/reorder + persistence checks against `TRACES_LIST_OPTIONS`.
- **LiveLogs:** parity with logs.
- **Trace-details v3 preview-fields:** regression check — still works (FieldsSelector internals touched, no consumer-side changes there).

---

## Future work (not in this PR)

- Cleanup PR — drops the now-dead `AddColumnField/` folder, the legacy `<OptionsMenu>` default-export component, the no-longer-called handlers in `useOptionsMenu.ts`, and the dead SCSS in `LogsFormatOptionsMenu.styles.scss`.
- Composite-identity helper to disambiguate same-`name`-different-`fieldContext` fields at the TanStack `column.id` level (frontend-only; no backend changes needed).
- Trace timestamp unification — currently rendered via a hardcoded prepend in `getListColumns`, asymmetric with logs. Decide whether to bring it under `selectColumns` or formally close as won't-do.

---

## Changelog

| Field | Value |
|------|-------|
| Change Type | Feature + Refactor |
| Description | Column picker in Logs Explorer, Trace Explorer, and LiveLogs now uses a unified FieldsSelector panel with Save / Discard, search, and drag-to-reorder. `body` and `timestamp` are locked from removal in logs / LiveLogs. |

---

## Checklist

- [x] Manually tested
- [x] Unit tests added
- [x] Breaking changes documented (none for consumers; internal API is additive)

---

## Notes for Reviewers

- **`respectColumnOrder` is additive on shared TanStack code.** Default `true` preserves every existing consumer's behavior; only Logs Explorer + LiveLogs opt in to `false`.
- **`useLogsTableColumns` rewrite changes column-render order for `body` / `timestamp`.** Previously hardcoded to `[..., timestamp, ...userFields, body]`; now they slot in wherever `selectColumns` puts them. Existing users see body/timestamp at their last-saved positions; new users see them in `LOGS_REQUIRED_COLUMNS` order (`[timestamp, body]` then user-added).
- **No popover wraps `<FieldsSelector>` at any of the three trigger sites.** This is deliberate — antd `Popover` with `destroyTooltipOnHide` unmounts content on outside-click, and `FieldsSelector`'s `FloatingPanel` portals to `document.body`, so the picker would destroy itself on every interaction.
- **Trace timestamp remains asymmetric with logs** — rendered via hardcoded prepend in `getListColumns`, not from `selectColumns`. Decision parked; revisit in the follow-up.

Uploading traces.mov…


frontend/src/components/FieldsSelector/AddedFields.tsx
```diff
@@ -53,15 +55,17 @@ function SortableField({
 				{allowDrag && <GripVertical size={14} />}
 				<span className={styles.fieldKey}>{field.name}</span>
```
H4ad, line 56: Let's add the fieldContext to know the different between two names, eg:

Image

We can add prefix like: 

Image

Or use badges

We can also like to the docs to teach the users the difference between using one or another.

H4ad (commented):
Move the success notification to bottom left or bottom center

Image

Be more explicit when max was reached, I'm trying to add but I'm just guessing the max reached but the items looks clickable on add.

Image

We still have the bug of resize (and we can't open/close again after opened), did you created a ticket for this?

Image

aks07:
> Move the success notification to bottom left or bottom center

> Be more explicit when max was reached, I'm trying to add but I'm just guessing the max reached but the items looks clickable on add.

These are more of design choices which are part of trace details revamp. Feel free to add these up here: https://app.notion.com/p/signoz/Trace-Details-Revamp-VQA-348fcc6bcd1980a4bdf2ec21163a741c

We will be later going through these and fixing in one go. else there is a lot of back and forth on these since these are subjective.

aks07:
> We still have the bug of resize (and we can't open/close again after opened), did you created a ticket for this?

Created a ticket for this: https://github.com/SigNoz/engineering-pod/issues/5197
will prioritise this in next sprint

frontend/src/components/FieldsSelector/AddedFields.tsx
aks07: > Let's add the fieldContext to know the different between two names, eg:
> 
> Image
> We can add prefix like:
> 
> Image
> Or use badges
> 
> We can also like to the docs to teach the users the difference between using one or another.



As already pointed out, this is still open for discussion as to how we will be solving this issue. Here is a closed pr on same: https://github.com/SigNoz/signoz/pull/9952

For now as first steps, we have moved to key as id instead of names as that introduced multiple bugs and not extensible when we later add a fix for this via badges or names.

H4ad (approved):
Please, create a follow-up tickets for:
- How we want to handle duplicated fields on selector (attribute/resource)
  - We can link the old conversation there and go with a final solution, or even a temporary one just to avoid total confusion that this causes today.
- Move the notification position, this blocks the user to do any other operation on those buttons, this is more a bug than a opinion.
- For the issue with resize we have https://github.com/SigNoz/engineering-pod/issues/5197

PR #11523, opened 2026-06-01
feat: show recent queries  to user

tewarig:
## Pull Request





### 📄 Summary
                                                                                                                                        
  Adds a "Recent searches" section inside the V5 Query Builder filter autocomplete so users can replay filter expressions they've  previously run — without retyping or remembering exact syntax.
  Implementation is **frontend-only, localStorage-backed** — no backend API. Recent expressions are partitioned by signal (`logs`,  `traces`, `metrics`) so a logs query the user built in the Logs Explorer surfaces in any other logs context (dashboard panel editor, alert builder, etc.) within the same browser/origin.                                                                                   
                                                                                                                                         
  Save side is centralized in `QueryBuilderProvider` — one `useSaveRecentQuery(stagedQuery)` call covers every surface that uses the QB
  context. One additional explicit call lives in `GridCardLayout/GridCard` for dashboard panel viewing (different data source —          
  `widget.query` instead of `stagedQuery`). Read side is a CodeMirror completion source merged with the existing `autoSuggestions`.    
    

https://github.com/orgs/SigNoz/projects/39/views/11?filterQuery=assignee%3Atewarig&pane=issue&itemId=156341316&issue=SigNoz%7Cengineering-pod%7C3922

https://github.com/SigNoz/engineering-pod/issues/5247#event-26427754919 

UI in light and dark mode - 


https://github.com/user-attachments/assets/46c1bb04-d620-4710-a7df-45a5cc5f7686




in case of long query - 

<img width="1475" height="463" alt="image" src="https://github.com/user-attachments/assets/f9376554-a564-48a3-bc96-205f4870cabc" />
 
UI walkthrough - 

https://github.com/user-attachments/assets/88289d59-26fc-46b1-ba37-c4d982957896



#### Issues closed by this PR


https://github.com/orgs/SigNoz/projects/39/views/11?filterQuery=+5+searches&pane=issue&itemId=156341316&issue=SigNoz%7Cengineering-pod%7C3922 


---

### ✅ Change Type
_Select all that apply_

- [ x ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

### 🧪 Testing Strategy
> How was this change validated?

have tested it manually and added test cases.

---

> What could break? How do we recover?
last radius:                                                                                                                        
  - Scoped to the V5 Query Builder filter autocomplete dropdown. Save calls are added inside `QueryBuilderProvider` and `GridCard`, both
  pre-existing code paths — no new mount points. Display layer is a second source on CodeMirror's `autocompletion.override` that returns
  `null` when no recents match, leaving existing behavior unchanged.

  Potential regressions:                                                                                                                 
  - Filter autocomplete: existing `autoSuggestions` is wrapped by a new `combinedSuggestions`. Behavior preserved (`filter: false` on the
   merged result; positional options pass through untagged).                                                                             
  - `validateQuery` is called in the save effect — same validator already used inside `QuerySearch`, no new evaluation logic.          
  - Every surface using the QB context now writes to localStorage on staged-query commit. Writes are deduped by a signature ref to avoid
  duplicate effect fires.                                                                                                                
                                          
  Rollback plan:                                                                                                                         
  - Feature can be disabled by removing two one-line call sites: `useSaveRecentQuery(stagedQuery)` in `providers/QueryBuilder.tsx` and   
  `useSaveRecentQuery(widget?.query)` in `GridCardLayout/GridCard/index.tsx`.                                                            
  - Read-side rollback: revert `override: [combinedSuggestions]` to `override: [autoSuggestions]` and remove the `addToOptions` block in 
  `QuerySearch.tsx`.                                                                                                                   
  - localStorage keys (`qb_recent_v1:logs|traces|metrics`) are versioned; orphan data is harmless and ignored if the schema changes (`v1`
   → `v2`).                                   
  - No backend, shared infra, or persisted user data touched.    
---

### 📝 Changelog
> Fill only if this affects users, APIs, UI, or documented behavior  
> Use **N/A** for internal or non-user-facing changes

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

**Save sites — two on purpose:**            
  - `providers/QueryBuilder.tsx` (centralized) covers every page that mounts `QueryBuilderProvider`: Logs / Traces / Metrics explorers,  
  NewWidget editor, alert builder.                                                                                                     
  - `container/GridCardLayout/GridCard/index.tsx` is the only intentional second call site — dashboard widgets render with `widget.query`
   (a saved query on the widget), not `useQueryBuilder().stagedQuery`. The comment above the call explains why.
                                                                                                                                         
  **Storage shape:**                                                                                                                   
  - One localStorage key per signal: `qb_recent_v1:logs`, `qb_recent_v1:traces`, `qb_recent_v1:metrics`.                                 
  - Each entry: `{ id, signal, filter, lastUsedAt }` — only the filter expression is persisted. Selecting a recent restores only the   
  filter; the rest of the user's builder query stays as-is.                                                                              
  - Dedup key inside the store: `(signal, normalized filter)`. Normalization lowercases keywords and strips whitespace outside quoted  
  strings.                                                                                                                               
                                                                                                                                         
  **CodeMirror integration:**                                                                                                            
  - The recents read path wraps the existing `autoSuggestions` in a `combinedSuggestions` function. Both option groups carry a `section`
  field so the dropdown renders labelled "Suggestions" and "Recent searches" headers (built-in `Completion.section` feature).            
  - `addToOptions` adds an inline ✕ delete button to recent rows. Click handler stops `pointerdown`/`mousedown`/`click` so it doesn't  
  trigger apply-completion, calls `store.remove`, then `startCompletion(view)` to refresh the open dropdown.
  - Long expressions truncate with `text-overflow: ellipsis`. Full text shown via native `title`, set in a microtask after CM mounts the 
  row (CM doesn't let us set attributes on the parent `<li>` directly).
  - Dropdown `max-height` grows proportionally with the number of recents (SCSS `@for` loop over `:has(li:nth-of-type(N))` selectors).   
  Falls back to the base height in browsers without `:has()` support.                                                                  
  
---


tushar-signoz:
https://github.com/SigNoz/engineering-pod/issues/3922#issuecomment-4619851953

aks07:
image

Can we remove the white border color we have added to `recent searches` ? This would make it similar to below footer...which also does not have border.

tewarig:
> image
> Can we remove the white border color we have added to `recent searches` ? This would make it similar to below footer...which also does not have border.

done - 

image

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
```diff
@@ -1250,6 +1255,40 @@ function QuerySearch({
...
+				from: 0,
+				to: fullDoc.length,
+				options: recentOptions,
+				filter: false,
+			};
+		}
+
+		return {
+			...result,
+			options: [...recentOptions, ...sectionedSuggestions],
```
aks07: Can we name them better? name is confusing
How about `recentOptions` and `suggestionOptions`?

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.styles.scss
```diff
@@ -159,11 +186,78 @@
...
+					display: inline-flex;
+					align-items: center;
+					justify-content: center;
+					width: 20px;
+					height: 20px;
+					padding: 0;
+					border: 0;
+					border-radius: 4px;
+					background: transparent;
+					color: var(--text-vanilla-400);
```
aks07: use semantic token

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.styles.scss
```diff
@@ -159,11 +186,78 @@
...
+					line-height: 1;
+					cursor: pointer;
+					opacity: 0.5;
+					transition:
+						opacity 0.12s ease,
+						background-color 0.12s ease;
+
+					&:hover {
+						opacity: 1;
+						background: color-mix(in srgb, var(--text-vanilla-400) 18%, transparent);
```
aks07: use semantic token

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.styles.scss
```diff
@@ -117,7 +122,16 @@
 				width: 100% !important;
 				max-width: 100% !important;
 				font-family: 'Space Mono', monospace !important;
-				min-height: 200px !important;
+				max-height: $dropdown-base-height !important;
+				overflow-y: auto !important;
+
+				@for $i from 1 through $max-recents-shown {
```
aks07, line 136: why do we need this?

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.styles.scss
```diff
@@ -159,11 +186,78 @@
 						display: none !important;
 					}
 
+					.cm-completionDetail {
+						margin-left: auto;
+						font-style: normal;
+						font-size: 10px;
```
aks07: Lets stick to using periscope font token: `--periscope-font-size-*`

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.styles.scss
```diff
@@ -1,5 +1,10 @@
 // TODO: Improve the styling of the query aggregation container and its components. - @YounixM , @H4ad
 
+$dropdown-base-height: 250px;
+$recents-header-height: 30px;
+$recent-row-height: 36px;
+$max-recents-shown: 5;
```
aks07, line 7: worth noting we are duplicating this with RECENTS_DISPLAY_CAP . Make sure if this is required to comment that it stays in conjunction to RECENTS_DISPLAY_CAP

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/utils.ts
```diff
@@ -38,3 +50,94 @@ export function getUserExpressionFromCombined(
...
+		},
+	}));
+}
+
+export function renderRecentDeleteButton(
+	completion: Completion,
+	_state: unknown,
+	view: EditorView | null,
+): Node {
+	if (completion.type !== 'recent') {
```
aks07: I think we should define 'recent' as CUSTOM_COMPLETION instead of hardcoding this everyplace

frontend/src/lib/recentQueries/normalize.ts
```diff
@@ -0,0 +1,29 @@
+const KEYWORDS_RE =
+	/\b(AND|OR|NOT|IN|LIKE|ILIKE|CONTAINS|EXISTS|BETWEEN|IS|NULL|TRUE|FALSE)\b/gi;
+
+const QUOTED_RE = /'(?:\\.|[^'\\])*'|"(?:\\.|[^"\\])*"/g;
```
aks07, line 25: some comments here would have been helpful :)

ankitnayan:
@aks07 please review thoroughly

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/utils.ts
```diff
@@ -38,3 +50,94 @@ export function getUserExpressionFromCombined(
...
+		}
+	});
+
+	const stop = (e: Event): void => {
+		e.preventDefault();
+		e.stopPropagation();
+	};
+	btn.addEventListener('pointerdown', stop);
+	btn.addEventListener('mousedown', stop);
+	btn.addEventListener('click', (e) => {
```
aks07, line 146: comment on why these event listeners was added could help

frontend/src/lib/recentQueries/store.ts
```diff
@@ -0,0 +1,161 @@
...
+	const next = current.filter((e) => e.id !== id);
+	if (next.length === current.length) {
+		return;
+	}
+	cache.set(signal, next);
+	writeToStorage(signal, next);
+	notify();
+}
+
+export function subscribe(cb: () => void): () => void {
```
aks07: why do we need pub/sub here? Even if we do need to sync states somewhere we should use useSyncExternalStore instead. check examples

aks07 (changes requested):

aks07:
Current implementation shows same queries for both metrics and cost meter. It should not do so. I assume this should not be the case. Thoughts? @tushar-signoz

tushar-signoz:
> Current implementation shows same queries for both metrics and cost meter. It should not do so. I assume this should not be the case. Thoughts? @tushar-signoz

@aks07 Yes, I missed it thanks. 
Yes @tewarig, along with `signal`, there's field called `source` use it for keying.
Metrics can have 2 sources, "meter", ""
Logs can also have 2 sources, "audit", "" . audit you might not find reference in frontend, even I don't know how its going to used in the frontend.

aks07:
> > Current implementation shows same queries for both metrics and cost meter. It should not do so. I assume this should not be the case. Thoughts? @tushar-signoz
> 
> @aks07 Yes, I missed it thanks. Yes @tewarig, along with `signal`, there's field called `source` use it for keying. Metrics can have 2 sources, "meter", "" Logs can also have 2 sources, "audit", "" . audit you might not find reference in frontend, even I don't know how its going to used in the frontend.

I dont think `audit` is used yet. Do confirm @tewarig

tewarig:
> > > Current implementation shows same queries for both metrics and cost meter. It should not do so. I assume this should not be the case. Thoughts? @tushar-signoz
> > 
> > 
> > @aks07 Yes, I missed it thanks. Yes @tewarig, along with `signal`, there's field called `source` use it for keying. Metrics can have 2 sources, "meter", "" Logs can also have 2 sources, "audit", "" . audit you might not find reference in frontend, even I don't know how its going to used in the frontend.
> 
> I dont think `audit` is used yet. Do confirm @tewarig

checked @aks07 we don't use it anywhere in 

> Logs can also have 2 sources, "audit", "" . audit

@aks07  , @tushar-signoz i checked ,  we don't use  "audit", ""  as of now in frontend.

tewarig:
> > Current implementation shows same queries for both metrics and cost meter. It should not do so. I assume this should not be the case. Thoughts? @tushar-signoz
> 
> @aks07 Yes, I missed it thanks. Yes @tewarig, along with `signal`, there's field called `source` use it for keying. Metrics can have 2 sources, "meter", "" Logs can also have 2 sources, "audit", "" . audit you might not find reference in frontend, even I don't know how its going to used in the frontend.

@tushar-signoz , @aks07 we will get source from backend right ?  so maybe i can use them. and show 

> > > > Current implementation shows same queries for both metrics and cost meter. It should not do so. I assume this should not be the case. Thoughts? @tushar-signoz
> > > 
> > > 
> > > @aks07 Yes, I missed it thanks. Yes @tewarig, along with `signal`, there's field called `source` use it for keying. Metrics can have 2 sources, "meter", "" Logs can also have 2 sources, "audit", "" . audit you might not find reference in frontend, even I don't know how its going to used in the frontend.
> > 
> > 
> > I dont think `audit` is used yet. Do confirm @tewarig
> 
> checked @aks07 we don't use it anywhere in
> 
> > Logs can also have 2 sources, "audit", "" . audit
> 
> @aks07 , @tushar-signoz i checked , we don't use "audit", "" as of now in frontend.


so @aks07 I am thinking of using source : (audit || "" ) now in frontend. I will use it to construct localStorage keys dynamically on the frontend. Now the localStorage keys will look like qb_recent_v1:metrics:meter and qb_recent_v1:metrics:

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.styles.scss
tewarig: so i want to have datadog like behaviour where we have dropdown to have a fix height. but it grows with how many recent search options we add dynamically.

tewarig:
hi @aks07 have resolved you comments.

frontend/src/lib/recentQueries/normalize.ts
```diff
@@ -0,0 +1,43 @@
+// Reserved query-builder keywords. Matched case-insensitively so we can canonicalise
+// them to lowercase — `a = 1 AND b = 2` and `a = 1 and b = 2` should dedup to the
+// same entry. \b boundaries prevent partial matches inside identifiers
+// (e.g. `OR` inside `originator`).
+const KEYWORDS_RE =
+	/\b(AND|OR|NOT|IN|LIKE|ILIKE|CONTAINS|EXISTS|BETWEEN|IS|NULL|TRUE|FALSE)\b/gi;
+
```
aks07, line 20: This looks like something that may exist in our code base somewhere please check once.

aks07: Try to see if we can use an existing util: extractQueryPairs to solve this

frontend/src/hooks/recentQueries/useRecents.ts
```diff
@@ -0,0 +1,30 @@
+import * as recentQueriesStore from 'lib/recentQueries/recentQueriesStore';
+import { useRecentQueriesStore } from 'lib/recentQueries/recentQueriesStore';
+import type { RecentQueryEntry } from 'lib/recentQueries/types';
```
aks07: useRecents could be named better.

tewarig: updated it

frontend/src/lib/recentQueries/normalize.ts
tewarig: done

tewarig:
@aks07  resolved your comments . have created a follow up as well for test and that initial loading thing 
https://github.com/SigNoz/engineering-pod/issues/5281

aks07:
image

There is very little room in dashboard to show the suggestions and the recents. I dont know how to solve this. personally we should move the whole qb section to the top. this gives us room to work with qb suggestions and also mimics explorer pages layout cc. @GeekBoySupreme

tewarig:
> image
> There is very little room in dashboard to show the suggestions and the recents. I dont know how to solve this. personally we should move the whole qb section to the top. this gives us room to work with qb suggestions and also mimics explorer pages layout cc. @GeekBoySupreme

Can we do the following things?

1. When the user opens this dropdown, we can still scroll the page, so the dropdown stays open while the content moves. Can we lock scroll when it's open?
2. Can we make this behave like a popover? When there's no space at the bottom, show the options at the top instead.

frontend/src/container/GridCardLayout/GridCard/index.tsx
```diff
@@ -266,6 +267,15 @@ function GridCardGraph({
...
 	);
 
+	/*
+	`widget.query` is the dashboard's saved query for this panel — it
+	does not flow through `useQueryBuilder().stagedQuery`, so the
+	provider-level save effect in `providers/QueryBuilder.tsx` doesn't
+	cover this surface. This explicit call ensures dashboard-viewed
+	queries are surfaced in the user's recents.
+	*/
+	useSaveRecentQuery(widget?.query);
```
aks07: This piece of code causes a major bug.

This hook runs for every grid card on the dashboard whenever a dashboard loads, causing it to save query filters from all panels and override the existing options.

See the attached video below.


https://github.com/user-attachments/assets/0e92749a-2263-4a1a-b7fd-9b5399df86c6

All recent queries from the Logs page disappear as soon as I visit a dashboard containing log panels, because the panels overwrite the existing recent queries

aks07: > This piece of code causes a major bug.
> 
> This hook runs for every grid card on the dashboard whenever a dashboard loads, causing it to save query filters from all panels and override the existing options.
> 
> See the attached video below.
> 
>  qb.bug.mov 
> All recent queries from the Logs page disappear as soon as I visit a dashboard containing log panels, because the panels overwrite the existing recent queries

Solution
drop the `useSaveRecentQuery(widget?.query)` altogether from the code. This does nothing meaningful. the QueryBuilder useSaveRecentQuery already registers for changes inside the panels as well so we should be good to go.

tushar-signoz:
Hey @tewarig 
You can skip dashboards for now if it's turning out to be difficult. People don't create dashboards too often. It the explorers where this recent is needed the most. Lets get this feature out first for the explorers, take feedback, iterate.

aks07:
> This piece of code causes a major bug.
> 
> This hook runs for every grid card on the dashboard whenever a dashboard loads, causing it to save query filters from all panels and override the existing options.
> 
> See the attached video below.
> 
>  qb.bug.mov 
> All recent queries from the Logs page disappear as soon as I visit a dashboard containing log panels, because the panels overwrite the existing recent queries


Along the same class of bugs as the dashboard bug, listening to  `stagedQuery`  change to save to recent queries would pollute the recent searches as `stagedQuery` is set from far more than a run click. i.e. page visit, refresh, navigation, and correlation redirects 
(logs↔traces "go to" clicks all push query params, which re-derive and re-stage)
Even quick filter usages

I assume this is not expected @tushar-signoz 


https://github.com/user-attachments/assets/593a920f-574a-4834-b3ae-d0eb61dde953

aks07:
> > This piece of code causes a major bug.
> > This hook runs for every grid card on the dashboard whenever a dashboard loads, causing it to save query filters from all panels and override the existing options.
> > See the attached video below.
> > qb.bug.mov
> > All recent queries from the Logs page disappear as soon as I visit a dashboard containing log panels, because the panels overwrite the existing recent queries
> 
> Along the same class of bugs as the dashboard bug, listening to `stagedQuery` change to save to recent queries would pollute the recent searches as `stagedQuery` is set from far more than a run click. i.e. page visit, refresh, navigation, and correlation redirects (logs↔traces "go to" clicks all push query params, which re-derive and re-stage)
> 
> I assume this is not expected @tushar-signoz
> 
>  staged.query.issue.mov

The right trigger is the explicit run...saving inside the QB context's `handleRunQuery` rather than reacting to `stagedQuery`. But the point to note is that not all call-site use the `handleRunQuery` from qb context. eg. Metrics explorer and some other places(would have to dig deeper).
If we do not want it to be saved on the above mentioned cases, we would have to dig deeper. cc. @therealpandey

aks07:
> > > This piece of code causes a major bug.
> > > This hook runs for every grid card on the dashboard whenever a dashboard loads, causing it to save query filters from all panels and override the existing options.
> > > See the attached video below.
> > > qb.bug.mov
> > > All recent queries from the Logs page disappear as soon as I visit a dashboard containing log panels, because the panels overwrite the existing recent queries
> > 
> > 
> > Along the same class of bugs as the dashboard bug, listening to `stagedQuery` change to save to recent queries would pollute the recent searches as `stagedQuery` is set from far more than a run click. i.e. page visit, refresh, navigation, and correlation redirects (logs↔traces "go to" clicks all push query params, which re-derive and re-stage)
> > I assume this is not expected @tushar-signoz
> > staged.query.issue.mov
> 
> The right trigger is the explicit run...saving inside the QB context's `handleRunQuery` rather than reacting to `stagedQuery`. But the point to note is that not all call-site use the `handleRunQuery` from qb context. eg. Metrics explorer and some other places(would have to dig deeper). If we do not want it to be saved on the above mentioned cases, we would have to dig deeper. cc. @therealpandey

For starters, we can start storing recent queries on handleRunQuery click. This would cover most of our places. Only one missing should be Metrics Summary, Infra Monitoring and Trace details page. We can pick these up later. cc. @tushar-signoz

tushar-signoz:
> > > This piece of code causes a major bug.
> > > This hook runs for every grid card on the dashboard whenever a dashboard loads, causing it to save query filters from all panels and override the existing options.
> > > See the attached video below.
> > > qb.bug.mov
> > > All recent queries from the Logs page disappear as soon as I visit a dashboard containing log panels, because the panels overwrite the existing recent queries
> > 
> > 
> > Along the same class of bugs as the dashboard bug, listening to `stagedQuery` change to save to recent queries would pollute the recent searches as `stagedQuery` is set from far more than a run click. i.e. page visit, refresh, navigation, and correlation redirects (logs↔traces "go to" clicks all push query params, which re-derive and re-stage)
> > I assume this is not expected @tushar-signoz
> > staged.query.issue.mov
> 
> The right trigger is the explicit run...saving inside the QB context's `handleRunQuery` rather than reacting to `stagedQuery`. But the point to note is that not all call-site use the `handleRunQuery` from qb context. eg. Metrics explorer and some other places(would have to dig deeper). If we do not want it to be saved on the above mentioned cases, we would have to dig deeper. cc. @therealpandey

Yes @aks07, we only want explicit searches by user and not navigations or go-tos.

aks07 (approved):
Changes look good to me

tushar-signoz:
Looks good

PR #11524, opened 2026-06-01
fix: skip resource filter if it exceeds a threshold

nityanandagohain:
## Pull Request

---

### 📄 Summary
Skip resource filter if the count > threshold

#### Issues closed by this PR
Part of https://github.com/SigNoz/engineering-pod/issues/5131

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: ✅ 
- Manual verification: ✅ 
- Edge cases covered:

---

- Blast radius:
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---


therealpandey (approved):

tushar-signoz:
A very noob question. Has an analysis been done which compares time-taken with `resource_cte` and `count query + without_resource_cte` query ? I get that we're reducing the worst case scenarios but we're also worsening best case scenarios too. And is there not a way to run both the queries in clickhouse sequentially ?

ankitnayan:
> A very noob question. Has an analysis been done which compares time-taken with `resource_cte` and `count query + without_resource_cte` query ? I get that we're reducing the worst case scenarios but we're also worsening best case scenarios too. And is there not a way to run both the queries in clickhouse sequentially ?

We are not removing the best case scenario. If the count of resource_fingerprint < 100K, we still use it. A better question would be at what value does the resource_fingperprint filter stop outperforming the one without it and what can be done to optimize it. Currently, 100K is a blant guesswork

pkg/querier/config.go
```diff
@@ -42,6 +54,9 @@ func (c Config) Validate() error {
 	if c.MaxConcurrentQueries <= 0 {
 		return errors.NewInvalidInputf(errors.CodeInvalidInput, "max_concurrent_queries must be positive, got %v", c.MaxConcurrentQueries)
 	}
+	if c.SkipResourceFingerprint.Enabled && c.SkipResourceFingerprint.Threshold == 0 {
```
srikanthccv, line 57:
```suggestion
	if c.SkipResourceFingerprint.Enabled && c.SkipResourceFingerprint.Threshold <= 0 {
```

pkg/telemetrylogs/statement_builder.go
```diff
@@ -707,33 +722,30 @@ func (b *logQueryStatementBuilder) maybeAttachResourceFilter(
...
-) (cteSQL string, cteArgs []any, err error) {
+) (cteSQL string, cteArgs []any, skipResourceFilter bool, err error) {
+
+	if b.skipResourceFingerprintEnabled {
+		decision, err := b.resourceFilterResolver.Resolve(ctx, query, start, end, variables)
+		if err != nil {
+			return "", nil, true, err
+		}
+		switch decision {
+		case qbtypes.ResourceFilterResolveKindNoOp:
```
srikanthccv, line 733: Please add comment on `Resolve` when the cases discussed when the resolved decision is going to be nop and how the result will be correct. Additionally have a dedicated integration tests for such cases.

pkg/telemetryresourcefilter/statement_builder.go
```diff
@@ -127,6 +127,25 @@ func (b *resourceFilterStatementBuilder[T]) Build(
...
+	end uint64,
+	query qbtypes.QueryBuilderQuery[T],
+	variables map[string]qbtypes.VariableItem,
+) (*qbtypes.Statement, error) {
+	inner, err := b.Build(ctx, start, end, qbtypes.RequestTypeRaw, query, variables)
+	if err != nil || inner == nil {
+		return nil, err
+	}
+	return &qbtypes.Statement{
+		Query: fmt.Sprintf("SELECT count() FROM (%s)", inner.Query),
```
srikanthccv, line 144: Check if `uniq` improves the performance here. 

`SELECT uniq(fingerprint) FROM WHERE ...`

srikanthccv (approved):
non-blocking comments, feel free to address them separately.

tushar-signoz:
> > A very noob question. Has an analysis been done which compares time-taken with `resource_cte` and `count query + without_resource_cte` query ? I get that we're reducing the worst case scenarios but we're also worsening best case scenarios too. And is there not a way to run both the queries in clickhouse sequentially ?
> 
> We are not removing the best case scenario. If the count of resource_fingerprint < 100K, we still use it. A better question would be at what value does the resource_fingperprint filter stop outperforming the one without it and what can be done to optimize it. Currently, 100K is a blant guesswork

But we're adding extra `count` query + round trip network call in every case.

nityanandagohain:
The count query for resource attribute is not expensive @tushar-signoz compared to a main table query . The next step is to check if we can further optimise the count query using uniq

PR #11529, opened 2026-06-01
perf: reuse label maps and index series by variable in formulas

namanv3:
## Pull Request

---

### 📄 Summary

For a customer instance, majority (80-90%) of the CPU usage was being taken by the `evaluateForLabelSet` or the `findUniqueLabelSets` method, based on the load. The former was dominated the `isSubset` and the `strings.HasPrefix` method, while the latter is almost all `isSubset`. This lead to high CPU usage alerts on that node, hence the `isSubset` method needs to be optimised and the usage of this and the `hasPrefix` methods have to be reduced.

### Solution

For `evaluateForLabelSet`, the superset sent to `isSubset` was always the same. In `findUniqueLabelSets`, each unique set found would be sent as the superset many times. The `isSubset` method would create a map from these sets (which are slices of Labels) every time, which showed up in the `pprof` analysis with a high amount of map assign cpu time.

To solve this, the `isSubset` method was changed to expect a map for the superset and the callers would construct the maps once and pass it on.

Next thing in `evaluateForLabelSet` was that the prefix check was happening inside a double for loop, which could be easily avoided if we construct a variable to series map once and use that, hence shortening the number of iterations in the double for loop and also removing the prefix check altogether.

### PProf analysis

Setup: 2 metrics that emit data for 10k time series, and a load test that checks for `A + B` for these metrics.

Before this change, a similar setup locally to the customer's would lead to 96s of CPU time in 30s of wall clock time, afterwards it was 28.04s.

Screenshot 2026-06-01 at 19 54 32
Screenshot 2026-06-01 at 19 53 08

#### Issues closed by this PR

After deployment, will review if the load is reduced for the pod, if yes then https://github.com/SigNoz/engineering-pod/issues/5122#issuecomment-4556207054 can be closed.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: None
- Manual verification: Via Load Test, and the existing bench tests
- Edge cases covered: None

---

- Blast radius: Formulas based queries in query builder v5
- Potential regressions: Nothing beyond formulas is possible
- Rollback plan: Revert PR

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---


srikanthccv:
Pending for more perf testing https://signoz-team.slack.com/archives/C08P77ZDHV0/p1780328269704029

srikanthccv:
This is a good speedup on `isSubset`. Can you add the before and after for the 1k / 5k / 10k benchmarks to the PR description? And `-cpuprofile` (more options at go test https://pkg.go.dev/cmd/go/internal/test#pkg-variables you may find useful while running these tests) the 10k case and paste `pprof -top`. Two things you should answer from that, how does runtime grow from 1k to 10k, and what is at the top? is it our code doing formula math?

namanv3:
@srikanthccv 
The cpu time increase from 1k -> 10k is worse than linear in the tests run. Even after the change it is still that, but the growth is much less.
Before: 140.73ms → 11.55s (~82×)
After: 50.80ms → 2.27s (~45×)

As for pprof on 10k, `github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5.(*FormulaEvaluator).isSubset (inline)` has the second highest cumulative time before the change, but after that it is not even in the top 10.

Full dump: https://www.notion.so/signoz/Perf-benchmarks-379fcc6bcd1980dbabcffa65a814f5a5?source=copy_link

namanv3:
Next targets after improving isSubset were reducing the number of allocations, and the high cpu usage by the `semawakeup` <> `pthread_cond_signal` pair. Tried to tackle the first problem by introducing more worker pools but that didn't help in the CPU usage at all. 

For the second problem, given that the per label set workload was drastically reduced, it made sense to not use a go routine per set cuz all that did was add operational overhead. Switching to a simple worker pool removed the second problem altogether.

Full dump: https://app.notion.com/p/signoz/More-experiments-379fcc6bcd19809287bec881a9049371?source=copy_link

Will need to make some logical changes in formula calculation to solve the first problem.

namanv3:
Some more analysis: https://app.notion.com/p/signoz/Reducing-number-of-allocations-37afcc6bcd1980b2a698efc6c74b77fe?source=copy_link

Avenue 3 from this doc has been pushed. Avenue 1 can be pushed if the `govaluate` package's `Evaluate` method can be made to accept `*float` values as well in the input map.

srikanthccv (approved):

PR #11552, opened 2026-06-02
refactor(sentry-config): disable tracing & ensure correct environment

H4ad:
## Pull Request

---

Added function to detect the environment for Sentry, disable traces from being emitted, and set  the correct version when uploading the source-map and emitting errors.

Related https://github.com/SigNoz/platform-pod/issues/2393

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Blast radius: Sentry Alerts
- Potential regressions: -

---

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | N/A |

---

### 📋 Checklist
- [ ] Manually tested


therealpandey (dismissed):

frontend/vite.config.ts
```diff
@@ -87,6 +87,9 @@ export default defineConfig(({ mode }): UserConfig => {
 				authToken: env.VITE_SENTRY_AUTH_TOKEN,
 				org: env.VITE_SENTRY_ORG,
 				project: env.VITE_SENTRY_PROJECT_ID,
+				// Pin the sourcemap-upload release to the same value injected as
+				// process.env.VERSION so uploaded sourcemaps resolve. Ref: platform-pod#2393
+				release: { name: env.VITE_VERSION || 'dev' },
```
AshwinBhatkal: we should avoid this fallback here. For whatever reason if version does not come, we should not be tagging it with dev. We should either not upload and/or error out prior

therealpandey: @AshwinBhatkal it's for local dev testing. The backend also sets dev as the build version if an explicit version is not passed.

therealpandey: Thanks for the input. Updated.

AshwinBhatkal (approved):

therealpandey (approved):

PR #11555, opened 2026-06-02
feat(metrics): add Prometheus-compatible /api/v1/series endpoint

### LINKED ISSUE #11519: Add Prometheus-compatible `/api/v1/series` endpoint (support `match[]` + point PromQL)
**Yes.** I’m trying to connect `prometheus-adapter` to SigNoz and it requires a Prometheus-compatible `/api/v1/series` endpoint (accepting `match[]` with PromQL and optional `start`/`end`) so the adapter can discover series and perform metric queries. Currently Signoz implements:

```http
GET /api/v1/query_range?query={promql}&start={unix_ts}&end={unix_ts}&step={duration}
GET /api/v1/query?query={promql}&time={unix_ts}
! BUT! No POST versions supporting `Content-Type: application/x-www-form-urlencoded` header, unfortunately...
```

but `/api/v1/series` is missing or not Prometheus-compatible, which prevents `prometheus-adapter` from working.

Add a Prometheus-compatible `/api/v1/series` endpoint that:

- Accepts one or more `match[]` query parameters containing PromQL matchers (e.g., `match[]={__name__=~"http_requests_total|cpu_usage"}` or `match[]={job="node"}`).
- Optionally accepts `start` and `end` as Unix timestamps to limit the time range for series discovery.
- Returns a Prometheus-compatible JSON response with **status: "success"** and **data** as an array of label sets (each label set is an object of label→value pairs).
- Supports the same semantics `prometheus-adapter` expects (instant/point semantics for matchers; i.e., treat `match[]` as PromQL series selectors).

### Query parameters
- **match[]** (required, repeatable): PromQL series selector(s).
- **start** (optional): Unix timestamp (seconds) for range start.
- **end** (optional): Unix timestamp (seconds) for range end.

### Response format (Prometheus-compatible)
```json
{
  "status": "success",
  "data": [
    {"__name__":"http_requests_total","job":"api-server","instance":"10.0.0.1"},
    {"__name__":"http_requests_total","job":"api-server","instance":"10.0.0.2"},
    {"__name__":"cpu_usage","host":"host-1"}
  ]
}
```

## Example requests and expected responses

- **Single match, no time range**

```http
GET /api/v1/series?match[]={__name__=~"http_requests_total"}
```

**Response**

```json
{
  "status": "success",
  "data": [
    {"__name__":"http_requests_total","job":"api-server","instance":"10.0.0.1"},
    {"__name__":"http_requests_total","job":"api-server","instance":"10.0.0.2"}
  ]
}
```

- **Multiple matchers and time range**

```http
GET /api/v1/series?match[]={job="node"}&match[]={__name__=~"cpu_.*"}&start=1685600000&end=1685686400
```

**Response**

```json
{
  "status": "success",
  "data": [
    {"__name__":"cpu_usage","job":"node","instance":"node-1"},
    {"__name__":"cpu_idle","job":"node","instance":"node-2"}
  ]
}
```

- Returning a custom Signoz-only endpoint format — rejected because `prometheus-adapter` expects Prometheus semantics and response shape.
- Mapping `prometheus-adapter` to use `/api/v1/query` or `/api/v1/query_range` — not feasible for series discovery because the adapter relies on `series` semantics and `match[]` support.

## Additional context
- https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers
- https://github.com/kubernetes-sigs/prometheus-adapter/blob/01919d0ef11859bc214e0c8a8bd5368afd9d47f7/pkg/client/api.go#L138-L232
- Having `/api/v1/status/buildinfo` too would prevent some issues with other apps.
  ```json
  {
    "status": "success",
    "data": {
      "version": "0.126.1",
      "revision": "cb7cbad5f9a2823a622aaa668833ca04f50a0ea7",
      "branch": "master",
      "buildUser": "prometheus@signoz.io",
      "buildDate": "20260530-16:19:59",
      "goVersion": "go1.25.0"
    }
  }
  ```

#### *Thank you* for your feature request – we love each and every one!

gingeekrishna:
## What does this PR do?

Fixes #11519

Adds the missing Prometheus-compatible endpoints required for `prometheus-adapter` and other tools to work with SigNoz as a Prometheus data source.

### New: `GET|POST /api/v1/series`
Accepts one or more `match[]` PromQL selectors and optional `start`/`end` Unix timestamps. Returns a Prometheus-compatible label-set array:
```json
{"status":"success","data":[{"__name__":"http_requests_total","job":"api-server"}]}
```
- Uses the existing Prometheus storage querier with a `Func: "series"` hint (no sample data fetched)
- Deduplicates series by label string key
- Strips the internal `fingerprint` label from results

### New: `GET|POST /api/v1/labels`
Returns all label names visible in the given time range. Accepts optional `match[]` selectors to filter by series and optional `start`/`end` bounds:
```json
{"status":"success","data":["__name__","job","instance"]}
```

### New: `GET|POST /api/v1/label/{label}/values`
Returns all values for a given label name. Accepts optional `match[]` selectors and `start`/`end` bounds:
```json
{"status":"success","data":["api-server","node-exporter"]}
```

### Updated: `GET|POST /api/v1/query_range` and `GET|POST /api/v1/query`
Added `POST` method support. The existing `r.FormValue()` parsers already handle `application/x-www-form-urlencoded` bodies, so only the route method list needed updating.

### New: `GET /api/v1/status/buildinfo`
Returns SigNoz version, git hash, branch, and Go version in the Prometheus buildinfo format.

## Files changed
- `pkg/query-service/model/queryParams.go` — new `SeriesQueryParams` and `LabelQueryParams` structs
- `pkg/query-service/app/parser.go` — new `parseSeriesRequest` and `parseLabelRequest` functions
- `pkg/query-service/interfaces/interface.go` — added `GetSeries`, `GetLabels`, `GetLabelValues` to `Reader` interface
- `pkg/query-service/app/clickhouseReader/reader.go` — implemented all three using Prometheus storage querier
- `pkg/query-service/app/http_handler.go` — new handlers; updated route registrations

## Test plan
- [ ] `GET /api/v1/series?match[]={__name__=~".+"}` returns a `{"status":"success","data":[...]}` array of label maps
- [ ] `POST /api/v1/series` with form body `match[]={job="my-service"}` returns the same format
- [ ] `GET /api/v1/labels` returns all label names
- [ ] `GET /api/v1/labels?match[]={job="my-service"}` returns labels scoped to that series
- [ ] `GET /api/v1/label/job/values` returns all job label values
- [ ] `GET /api/v1/query_range` and `POST /api/v1/query_range` both work
- [ ] `GET /api/v1/status/buildinfo` returns version info
- [ ] `prometheus-adapter` can connect and discover series

### DIFF
--- a/pkg/query-service/app/http_handler.go
+++ b/pkg/query-service/app/http_handler.go
@@ -385,8 +385,10 @@ func (aH *APIHandler) Respond(w http.ResponseWriter, data interface{}) {
 
 // RegisterRoutes registers routes for this handler on the given router
 func (aH *APIHandler) RegisterRoutes(router *mux.Router, am *middleware.AuthZ) {
-	router.HandleFunc("/api/v1/query_range", am.ViewAccess(aH.queryRangeMetrics)).Methods(http.MethodGet)
-	router.HandleFunc("/api/v1/query", am.ViewAccess(aH.queryMetrics)).Methods(http.MethodGet)
+	router.HandleFunc("/api/v1/query_range", am.ViewAccess(aH.queryRangeMetrics)).Methods(http.MethodGet, http.MethodPost)
+	router.HandleFunc("/api/v1/query", am.ViewAccess(aH.queryMetrics)).Methods(http.MethodGet, http.MethodPost)
+	router.HandleFunc("/api/v1/series", am.ViewAccess(aH.seriesMetrics)).Methods(http.MethodGet, http.MethodPost)
+	router.HandleFunc("/api/v1/status/buildinfo", am.ViewAccess(aH.buildInfo)).Methods(http.MethodGet)
 	router.HandleFunc("/api/v1/rules", am.ViewAccess(aH.listRules)).Methods(http.MethodGet)
 	router.HandleFunc("/api/v1/rules/{id}", am.ViewAccess(aH.getRule)).Methods(http.MethodGet)
 	router.HandleFunc("/api/v1/rules", am.EditAccess(aH.createRule)).Methods(http.MethodPost)
@@ -1108,6 +1110,42 @@ func (aH *APIHandler) queryMetrics(w http.ResponseWriter, r *http.Request) {
 
 	aH.Respond(w, responseData)
 
+}
+
+func (aH *APIHandler) seriesMetrics(w http.ResponseWriter, r *http.Request) {
+	params, apiErr := parseSeriesRequest(r)
+	if apiErr != nil {
+		RespondError(w, apiErr, nil)
+		return
+	}
+
+	ctx := r.Context()
+	if to := r.FormValue("timeout"); to != "" {
+		var cancel context.CancelFunc
+		timeout, err := parseMetricsDuration(to)
+		if aH.HandleError(w, err, http.StatusBadRequest) {
+			return
+		}
+		ctx, cancel = context.WithTimeout(ctx, timeout)
+		defer cancel()
+	}
+
+	result, apiError := aH.reader.GetSeries(ctx, params)
+	if apiError != nil {
+		RespondError(w, apiError, nil)
+		return
+	}
+
+	aH.Respond(w, result)
+}
+
+func (aH *APIHandler) buildInfo(w http.ResponseWriter, r *http.Request) {
+	aH.Respond(w, map[string]string{
+		"version":   version.Info.Version(),
+		"revision":  version.Info.Hash(),
+		"branch":    version.Info.Branch(),
+		"goVersion": version.Info.GoVersion(),
+	})
 }
 
 func (aH *APIHandler) registerEvent(w http.ResponseWriter, r *http.Request) {
--- a/pkg/query-service/app/http_handler.go
+++ b/pkg/query-service/app/http_handler.go
@@ -388,6 +388,8 @@ func (aH *APIHandler) RegisterRoutes(router *mux.Router, am *middleware.AuthZ) {
 	router.HandleFunc("/api/v1/query_range", am.ViewAccess(aH.queryRangeMetrics)).Methods(http.MethodGet, http.MethodPost)
 	router.HandleFunc("/api/v1/query", am.ViewAccess(aH.queryMetrics)).Methods(http.MethodGet, http.MethodPost)
 	router.HandleFunc("/api/v1/series", am.ViewAccess(aH.seriesMetrics)).Methods(http.MethodGet, http.MethodPost)
+	router.HandleFunc("/api/v1/labels", am.ViewAccess(aH.labelsMetrics)).Methods(http.MethodGet, http.MethodPost)
+	router.HandleFunc("/api/v1/label/{label}/values", am.ViewAccess(aH.labelValuesMetrics)).Methods(http.MethodGet, http.MethodPost)
 	router.HandleFunc("/api/v1/status/buildinfo", am.ViewAccess(aH.buildInfo)).Methods(http.MethodGet)
 	router.HandleFunc("/api/v1/rules", am.ViewAccess(aH.listRules)).Methods(http.MethodGet)
 	router.HandleFunc("/api/v1/rules/{id}", am.ViewAccess(aH.getRule)).Methods(http.MethodGet)
@@ -1146,6 +1148,62 @@ func (aH *APIHandler) buildInfo(w http.ResponseWriter, r *http.Request) {
 		"branch":    version.Info.Branch(),
 		"goVersion": version.Info.GoVersion(),
 	})
+}
+
+func (aH *APIHandler) labelsMetrics(w http.ResponseWriter, r *http.Request) {
+	params, apiErr := parseLabelRequest(r)
+	if apiErr != nil {
+		RespondError(w, apiErr, nil)
+		return
+	}
+
+	ctx := r.Context()
+	if to := r.FormValue("timeout"); to != "" {
+		var cancel context.CancelFunc
+		timeout, err := parseMetricsDuration(to)
+		if aH.HandleError(w, err, http.StatusBadRequest) {
+			return
+		}
+		ctx, cancel = context.WithTimeout(ctx, timeout)
+		defer cancel()
+	}
+
+	result, apiError := aH.reader.GetLabels(ctx, params)
+	if apiError != nil {
+		RespondError(w, apiError, nil)
+		return
+	}
+
+	aH.Respond(w, result)
+}
+
+func (aH *APIHandler) labelValuesMetrics(w http.ResponseWriter, r *http.Request) {
+	labelName := mux.Vars(r)["label"]
+
+	params, apiErr := parseLabelRequest(r)
+	if apiErr != nil {
+		RespondError(w, apiErr, nil)
+		return
+	}
+
+	ctx := r.Context()
+	if to := r.FormValue("timeout"); to != "" {
+		var cancel context.CancelFunc
+		timeout, err := parseMetricsDuration(to)
+		if aH.HandleError(w, err, http.StatusBadRequest) {
+			return
+		}
+		ctx, cancel = context.WithTimeout(ctx, timeout)
+		defer cancel()
+	}
+
+	result, apiError := aH.reader.GetLabelValues(ctx, labelName, params)
+	if apiError != nil {
+		RespondError(w, apiError, nil)
+		return
+	}
+
+	aH.Respond(w, result)
 }
 
 func (aH *APIHandler) registerEvent(w http.ResponseWriter, r *http.Request) {
--- a/pkg/prometheus/clickhouseprometheus/client.go
+++ b/pkg/prometheus/clickhouseprometheus/client.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math"
 	"sort"
+	"strings"
 	"sync"
 
 	"github.com/SigNoz/signoz/pkg/errors"
@@ -68,6 +69,17 @@ func (client *client) Read(ctx context.Context, query *prompb.Query, sortSeries 
 	}
 	if len(fingerprints) == 0 {
 		return remote.FromQueryResult(sortSeries, new(prompb.QueryResult)), nil
+	}
+
+	// For series-only requests (e.g. /api/v1/series), return label sets directly
+	// without fetching sample data. This avoids the "metric_name = ''" condition
+	// in querySamples when no __name__ matcher is present.
+	if query.Hints != nil && query.Hints.Func == "series" {
+		res := new(prompb.QueryResult)
+		for _, lbls := range fingerprints {
+			res.Timeseries = append(res.Timeseries, &prompb.TimeSeries{Labels: lbls})
+		}
+		return remote.FromQueryResult(sortSeries, res), nil
 	}
 
 	sub, err := seriesLookupQuery(query, true)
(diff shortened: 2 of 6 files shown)


lucasfcnunes (changes requested):
@gingeekrishna, it's not building...

```console
$ > gh pr checkout 11555
remote: Enumerating objects: 64, done.
remote: Counting objects: 100% (55/55), done.
remote: Compressing objects: 100% (13/13), done.
remote: Total 40 (delta 34), reused 31 (delta 25), pack-reused 0 (from 0)
Unpacking objects: 100% (40/40), 9.36 KiB | 273.00 KiB/s, done.
From github.com:SigNoz/signoz
 * [new ref]             refs/pull/11555/head -> feat/prometheus-series-endpoint
M       Makefile
Switched to branch 'feat/prometheus-series-endpoint'
$ > NAME=signoz ARCHS=amd64 DOCKER_REGISTRY_COMMUNITY=docker.io/lucasfcnunes/signoz-community make docker-build-community
>> building binary /home/lucasfcnunes/dotfiles/lucasfcnunes/signoz/target/linux-amd64/signoz-community
# github.com/SigNoz/signoz/pkg/query-service/app/clickhouseReader
../../pkg/query-service/app/clickhouseReader/reader.go:306:27: undefined: parser.ParseMetricSelector
../../pkg/query-service/app/clickhouseReader/reader.go:363:32: undefined: parser.ParseMetricSelector
../../pkg/query-service/app/clickhouseReader/reader.go:403:32: undefined: parser.ParseMetricSelector
make: *** [Makefile:120: go-build-community-amd64] Error 1
```

gingeekrishna:
@lucasfcnunes Thanks for catching that! The build error was caused by an API change in the version of Prometheus used by this project (v0.311.3). In that version, `ParseMetricSelector` is a method on the `parser.Parser` interface rather than a package-level function — so `parser.ParseMetricSelector()` is undefined.

Fixed by constructing a `parser.NewParser(parser.Options{})` instance in each of the three functions (`GetSeries`, `GetLabels`, `GetLabelValues`) and calling `.ParseMetricSelector()` on it. The branch has been updated — it should build cleanly now.

lucasfcnunes (changes requested):
OK! Now it's really building. Some things are working but most are not:
- POST methods on the old endpoints (`/api/v1/query` and `/api/v1/query_range`) are working as expected.
- On `/api/v1/labels` and e.g. `/api/v1/label/__name__/values`
  - return nothing [1].
- On `/api/v1/series`:
  - only work with one `match[]` not two. With two it explodes [2].
  - e.g. `match[]` = `{__name__="k8s.container.ready"}` or `{__name__="k8s.container.ready", "k8s.container.name"="coredns"}` work as expected but `{"k8s.container.name"="coredns"}` returns nothing [1].



[1] returning nothing means it gives:
```json
{
  "status": "success",
  "data": []
  // or
  "data": null
}
```

[2] exploding means it gives:
```json
{
  "status": "error",
  "error": {
    "type": "fatal",
    "code": "fatal",
    "message": "An unexpected error occurred. Please retry, and if the issue persists, report it at https://github.com/SigNoz/signoz/issues or contact support."
  }
}
```

gingeekrishna:
Hi @lucasfcnunes, all three issues are fixed. Here's what each root cause was and how it was addressed:

**1. `/api/v1/labels` and `/api/v1/label/{name}/values` returning empty**

Root cause: Prometheus v0.311.3's remote-read client (`remote.SampleAndChunkQueryableClient`) does not implement `LabelNames()` or `LabelValues()` on the querier — this is a known upstream limitation (prometheus/prometheus#3351). Calls to these methods silently returned empty results.

Fix: Replaced both `GetLabels` and `GetLabelValues` with direct ClickHouse queries. Labels are enumerated using `arrayJoin(JSONExtractKeys(labels))` on the `distributed_time_series_v4_1day` table (the same approach used by `GetMetricAttributeKeys`). The `__name__` label maps directly to the `metric_name` column for the values query. Any active `match[]` selector is translated to ClickHouse conditions via a new `matchersToClickhouse` helper.

**2. `/api/v1/series` crashing with two `match[]` parameters**

Root cause: The Prometheus remote-read querier batches all `Select()` calls and dispatches them via a single `ReadMultiple` RPC on the first `Next()` call. Once that dispatch happens, the querier is exhausted and cannot be reused — calling `Select()` on the same querier for a second `match[]` caused the fatal error.

Fix: Moved querier creation inside the `match[]` loop in `GetSeries`, so each matcher gets a fresh querier instance. Each querier is explicitly closed after its series set is fully consumed.

**3. `/api/v1/series` returning nothing for `{k8s.container.name="coredns"}` (no `__name__` matcher)**

Root cause: `queryToClickhouseQuery` unconditionally appended `metric_name = $N` even when `metricName` was an empty string, so `metric_name = ''` matched no rows.

Fix: Made the `metric_name` condition conditional — added only when `metricName != ""`. The argument index tracking was also refactored from a fixed offset to a dynamic `nextArgIdx` that correctly adjusts based on whether the `metric_name` arg was included.

lucasfcnunes (changes requested):
Some queries are hitting ClickHouse as `U__k8s_2e_pod_2e_name` (escaping=values) instead of `k8s.pod.name`.
See https://prometheus.io/docs/instrumenting/escaping_schemes/#value-encoding-escaping-values

```sql
SELECT DISTINCT lv FROM (
    SELECT JSONExtractString(labels, 'U__k8s_2e_pod_2e_name') AS lv -- WRONG
    SELECT JSONExtractString(labels, 'k8s.pod.name') AS lv -- CORRECT
    FROM signoz_metrics.distributed_time_series_v4_1day
    WHERE true
     and unix_milli >= 1780987560000 AND unix_milli <= 1781074020000
     AND __normalized = 0
    AND metric_name = 'k8s.container.ready'
) WHERE lv != '' ORDER BY lv
;
```

PS: I really think that it should get from `signoz_metrics.distributed_metadata` in this case. But I'm not sure.

```sql
select metric_name (aka __name__), attr_name (aka label_name), attr_string_value (aka label_value), first_reported_unix_milli (aka aprox start), last_reported_unix_milli (aka aprox end)
from signoz_metrics.distributed_metadata
where ...
```

The above table may be able to produce `api/v1/series`, `api/v1/labels` and `api/v1/label/{label_name}/values` faster, but i don't know if it would be better or not.

---

On `/api/v1/series`:

`match[]` = `{__name__="k8s.container.ready", "k8s.container.name"="coredns"}` returns 1 item in `data` but `{"k8s.container.name"="coredns"}` returns 0.

gingeekrishna:
Thanks for the detailed SQL trace — that made it easy to pinpoint both issues.

Fixed in commit 2e198f1:

1. **Label name unescaping** — `U__k8s_2e_pod_2e_name` (and similar value-encoded names) are now decoded back to `k8s.pod.name` before being used in ClickHouse JSON extraction. The fix is applied in all three code paths: `queryToClickhouseQuery` (remote-read), `matchersToClickhouse` (used by `GetLabels`/`GetLabelValues`), and the `labelName` URL parameter in `GetLabelValues`.

2. **`/api/v1/series` returning 0 without `\_\_name\_\_`** — `Read()` now detects the `series` hint (`query.Hints.Func == "series"`) and returns label sets directly from the outer fingerprint query, skipping `querySamples` entirely. This avoids the `metric_name = ''` condition that was returning no rows.

Regarding `distributed_metadata` — agreed it could be faster for label/series lookups. Happy to explore that as a follow-up once the correctness issues are resolved.

Brunotlps:
Hey @gingeekrishna, sorry for th delay. As I mentioned in the issue, I'm not a SigNoz contributor and can't influence merge, but I gave the diff a focused read from a Prometheus-compatibility angle. I didn't run it locally and I'm not familiar with the SigNoz internals deeply enough to evaluate everything, so the points below are framed as questions/things worth confirm.

1. Regex anchoring in matchersToClickhouse (reader.go) - for MatchRegexp/MatchNotRegexp the SQL uses ClickHouse's match(), which does substring matching. Prometheus regex matchers are fully anchored (prom.* is treated as ^prom.*$). Since GetSeries goes through the Prometheus querier but GetLabels/GetLabelValues build the SQL directly, the same match[] selector could yield different results across the two paths. Was that intentional, or should the regex be wrapped to anchor it on the SQL side?

2. /status/buildinfo shape (http_handler.go: buildInfo) - the response omits buildUser and buildDate compared to the Prometheus shape. More relevant for consumers: prometheus-adapter parses version as semver. Does version.Info.Version() always return a valid semver string across dev/CI/release builds?

3. limit handling in GetSeries (reader.go) - params.Limit is silently clamped to seriesResultHardLimit = 100_000, and reaching the limit returns ErrorExec (422) rather than truncating. The usual expectation from Prometheus consumers is either a truncated response with a warning or the requested limit honored. Returning 422 seems like it could leave the endpoint silently unusable on larger deployments, with no way for the client to tell missing-from-data apart from missing-due-to-cap. Intentional?

4. Description vs. interface change - the issue thread suggested using the existing Prometheus storage querier (LabelNames/LabelValues) without new abstractions. Reading reader.go, only GetSeries actually goes through the querier; GetLabels and GetLabelValues issue direct ClickHouse queries (with a comment pointing to prometheus/prometheus#3351 as the reason). That seems reasonable given the upstream gap, just worth updating the PR description to reflect it, since the three new methods do extend the Reader interface and that's a non-trivial change for downstream implementers/mocks.

5. Validation gaps in parser.go - parseQueryRangeRequest rejects end < start with ErrorBadData, but parseSeriesRequest and parseLabelRequest don't. A request with an inverted range passes through to the reader as-is. Also, the len(matches) == 0 check in parseSeriesRequest doesn't catch ?match[]= (param present, empty value) — r.Form["match[]"] returns [""] in that case, which then fails further down. Both look like minor gaps; was the intent to match parseQueryRangeRequest's checks?

6. Duplicate route registration (http_handler.go: RegisterRoutes) - /api/v1/query_range and /api/v1/query are registered twice: once with GET (pre-existing) and once with GET, POST (new). It looks like the new lines were appended rather than replacing the originals. Is that intended, or just an oversight?

Again, leaving the merge call to the SigNoz core team. Happy to follow up on anything that's not clear.

pkg/query-service/app/http_handler.go
```diff
@@ -482,8 +482,12 @@ func (aH *APIHandler) Respond(w http.ResponseWriter, data interface{}) {
 
 // RegisterRoutes registers routes for this handler on the given router
 func (aH *APIHandler) RegisterRoutes(router *mux.Router, am *middleware.AuthZ) {
-	router.HandleFunc("/api/v1/query_range", am.ViewAccess(aH.queryRangeMetrics)).Methods(http.MethodGet)
-	router.HandleFunc("/api/v1/query", am.ViewAccess(aH.queryMetrics)).Methods(http.MethodGet)
```
lucasfcnunes, line 389: > Duplicate route registration (http_handler.go: RegisterRoutes) - /api/v1/query_range and /api/v1/query are registered twice: once with GET (pre-existing) and once with GET, POST (new). It looks like the new lines were appended rather than replacing the originals. Is that intended, or just an oversight?

@Brunotlps
This point doesn't seem right

lucasfcnunes:
@gingeekrishna 
Everything seems to be working now, but I have to put last 2 days to get results.
I suspect it's caused by the use of the table `signoz_metrics.distributed_time_series_v4_1day` of 1 day size.

EDIT:
Found a bug. (caused by the dot in "container.id")

Request:
```
Request URL: /api/v1/label/__name__/values
Method:  GET
Raw query string: limit=40000&start=1781032560&end=1781205420&match[]={__name__%3D~".*cpu.*"%2Ccontainer.id%3D"4899dcbbb64786df565ef109139a98e1c8855a7d6575118beb4187a01f229599"}
limit: 40000
start: 1781032560
end: 1781205420
match.0: {__name__=~".*cpu.*",container.id="4899dcbbb64786df565ef109139a98e1c8855a7d6575118beb4187a01f229599"}
```

Response:
```json
{
  "status": "error",
  "errorType": "bad_data",
  "error": "1:30: parse error: unexpected character inside braces: '.'"
}
```

lucasfcnunes:
@gingeekrishna, I think you merged other unrelated things in this PR this time...

gingeekrishna:
Hi @lucasfcnunes — thanks for flagging this! The branch had two git merge main commits that pulled unrelated upstream changes into the diff. I've rebased onto main now, so the PR shows only the 6 feature commits.

gingeekrishna:
> Hi @lucasfcnunes — thanks for flagging this! The branch had two git merge main commits that pulled unrelated upstream changes into the diff. I've rebased onto main now, so the PR shows only the 6 feature commits.

Hi @lucasfcnunes 

Could you check this PR and confirm?

PR #11452, opened 2026-05-25
feat(trace-details): Add API endpoint & module for trace aggregations

krsoninikhil:
Part of https://github.com/SigNoz/engineering-pod/issues/4787

Add API endpoint & module for trace aggregations. Will raise a follow up for implementation of store methods.

Query used for calculating the non-overlapping span execution time:

```sql
WITH

-- query minimum span fields
all_spans AS (
	SELECT DISTINCT ON (span_id)
		%s AS field_value,
		toUnixTimestamp64Nano(timestamp) AS start_ns,
		start_ns + duration_nano AS end_ns
	FROM %s.%s
	WHERE trace_id=? AND ts_bucket_start>=? AND ts_bucket_start<=? AND notEmpty(field_value)
	ORDER BY timestamp ASC, name ASC
),

-- calculate max end time of preceding spans
effective_start AS (
	SELECT
		field_value, end_ns,
		greatest(
			start_ns,
			ifNull(
				max(end_ns) OVER (
					PARTITION BY field_value
					ORDER BY start_ns
					ROWS BETWEEN UNBOUNDED PRECEDING AND 1 PRECEDING
				),
				toUInt64(0)
			)
		) AS effective_start_ns
	FROM all_spans
)

-- each span contributes only the tail past its effective start.
SELECT field_value, sum(toUInt64(greatest(end_ns - effective_start_ns, 0))) AS total_ns
FROM effective_start
GROUP BY field_value
```

Another possible option using `arrayFold` and `arraySort` simpler but hard to understand (rejected):
```sql
WITH spans AS (
    SELECT DISTINCT ON (span_id)
        <field_expr> AS grp,
        toUInt64(timestamp)                AS s,
        toUInt64(timestamp) + duration_nano AS e
    FROM signoz_traces.distributed_signoz_index_v3
    WHERE trace_id = ? AND ts_bucket_start >= ? AND ts_bucket_start <= ?
      AND notEmpty(grp)
)
SELECT
    grp,
    arrayFold(
        (acc, x) -> (
            greatest(acc.1, x.2),
            acc.2 + if(x.2 > greatest(x.1, acc.1),
                       x.2 - greatest(x.1, acc.1),
                       toUInt64(0))
        ),
        arraySort(t -> t.1, groupArray((s, e))),
        (toUInt64(0), toUInt64(0))
    ).2 AS value
FROM spans
GROUP BY grp
```

PR #11556, opened 2026-06-02
feat(trace-details): lazy aggregations + waterfall v4 cutover

aks07:
## Pull Request

---

### 📄 Summary

The waterfall API is being broken down to reduce its memory footprint. As part of that, **aggregations no longer ship inside the waterfall response** — there's now a separate `POST /api/v1/traces/{traceID}/aggregations` endpoint, and an OOM-safe `POST /api/v4/traces/{traceID}/waterfall` that drops aggregations vs v3 (backend already shipped in #11452 + the v4 route).

This PR is the **frontend migration** for the new trace-details page (`src/pages/TraceDetailsV3/`). The end state:

- **Waterfall → v4** — the waterfall call moves to `/api/v4/.../waterfall` and no longer requests or reads aggregations.
- **Aggregations are now lazy** — instead of being fetched up-front (all candidate color-by fields × 2 aggregations, bundled into the waterfall request), they're fetched **only when the Analytics panel is opened**, and **only for the currently-selected color-by field**.
- **Color-by availability is decoupled from aggregations** — the color-by selector previously derived its options from the eager aggregations response. It now derives them client-side from the loaded spans (`getAvailableColorByFieldNames`, an any-span `.some()` over `service.name`/`host.name`/`k8s.*`), which mirrors the old server-side "field exists in the trace" signal without the parallel call.

Delivered in three independently-reviewable commits: (1) decouple color-by gating from aggregations, (2) lazy on-demand aggregations in the Analytics panel, (3) v4 waterfall cutover.

Net effect: opening a trace fires **one** waterfall request (v4, no aggregations); the aggregations request fires **only** when the Analytics panel is opened.

https://github.com/user-attachments/assets/40623fe3-4e3f-445b-bc98-a6d8a5d11dfa

#### Issues closed by this PR

Implements SigNoz/engineering-pod#5076

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🧪 Testing Strategy

- **Tests added/updated:**
  - `getAvailableColorByFieldNames.test.ts` — any-span availability, attributes fallback, empty set
  - `traceStore.test.ts` — color-by gating (default-only before load, default + present fields, honor persisted when available, fall back when not, trust persisted while loading)
  - `useGetTraceAggregations.test.tsx` — `enabled` gating (fires only with `enabled` + `traceId` + non-empty aggregations)
  - `AnalyticsPanel.test.tsx` — closed → no fetch, request shape (current field × both aggs), loading-with-tabs-visible, error, empty, rows
- **Manual verification:** trace detail loads on v4 (network shows v4 URL, no aggregations in the waterfall request/response); Analytics panel closed → no aggregations request; open → one request for the current field; switching color-by refetches; loading spinner centered with tabs visible; field with no data → empty state.
- **Edge cases covered:** windowed/large traces (any loaded window yields the full color-by set under the trace-wide-resource premise); cross-trace navigation via linked spans.

---

- **Blast radius:** trace-details V3 page only (`src/pages/TraceDetailsV3/`); waterfall fetch layer + the Analytics panel. No other consumers read the waterfall aggregations.
- **Potential regressions:** color-by selector options now come from loaded spans rather than the aggregations API — same signal in practice (resource attributes are trace-wide). Waterfall endpoint version bump (v3 → v4).
- **Rollback plan:** revert the PR — the v3 waterfall endpoint and the prior eager-aggregations path are unchanged on the backend.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | Trace-detail aggregations (exec-time % / span counts) now load on demand when the Analytics panel is opened, and the waterfall uses the lower-memory v4 endpoint. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- **Gating-source change (worth a close look):** color-by availability is now derived from loaded spans via `getAvailableColorByFieldNames` (`.some()` over candidate fields), not from the aggregations API. Considered anchoring on the root span for determinism, but the v4 waterfall is windowed (root not guaranteed in any window), so `.some()` over loaded spans is both more robust and faithful to the old server-side semantic. No accumulation across windows — under the trace-wide-resource premise any window already carries the full set.
- **Naming:** kept `SpanV3`/`EventV3`/`TraceDetailV3URLProps` and the `getTraceV3.ts` types file name (renaming the file would churn 27 `SpanV3` importers for no value); renamed only the wire-contract types to `GetTraceV4*` and the fetch fn/hook to `getTraceV4`/`useGetTraceV4`. Aggregation types moved to `getTraceAggregations.ts` and renamed `WaterfallAggregation*` → `TraceAggregation*` (no longer waterfall-bound).


frontend/src/types/api/trace/getTraceV3.ts
```diff
@@ -88,7 +71,7 @@ export interface SpanV3 {
 	trace_state: string;
 }
 
-export interface GetTraceV3SuccessResponse {
+export interface GetTraceV4SuccessResponse {
```
tewarig, line 74: nit: it is okay to have v4 types in getTracesV3.ts ?

frontend/src/api/trace/getTraceAggregations.ts
H4ad, line 1: Any reason for not using `getTraceAggregations` from `frontend/src/api/generated/services/tracedetail/index.ts`?

frontend/src/api/trace/getTraceV4.tsx
H4ad, line 1: Same for this one, we already have a generated code for doing this HTTP Request

frontend/src/hooks/trace/useGetTraceAggregations.tsx
H4ad, line 1: Replace by `useGetTraceAggregations` from generated schema.

frontend/src/hooks/trace/useGetTraceV4.tsx
H4ad, line 1: Same for this one, let's use the generated one

frontend/src/pages/TraceDetailsV3/SpanDetailsPanel/AnalyticsPanel/AnalyticsPanel.tsx
```diff
@@ -93,6 +118,33 @@ function AnalyticsPanel({
...
+		}
+		if (rowCount === 0) {
+			return (
+				<div className={styles.state}>
+					<Typography.Text>No data for {colorByFieldName}</Typography.Text>
+				</div>
+			);
+		}
+		return null;
+	};
```
H4ad: nit: this is something you could split/move to a subcomponent, and also move the `null` and replace by the actual content, it would make this component very small and the new one very focused

tewarig (commented):
https://github.com/user-attachments/assets/73012beb-01bf-4ee8-8a62-8dda8906a0ed

Nit :  we need to fix zIndex for this . seems very odd

frontend/src/pages/TraceDetailsV3/SpanDetailsPanel/AnalyticsPanel/AnalyticsPanel.tsx
```diff
@@ -132,65 +184,69 @@ function AnalyticsPanel({
...
-												/>
-											</div>
-											<span className={cx(styles.value, styles.valueWide)}>
-												{row.percentage.toFixed(2)}%
+							{renderState(execTimeRows.length) ?? (
+								<div className={styles.list}>
+									{execTimeRows.map((row) => (
+										<>
+											<div
+												key={`${row.group}-dot`}
```
tewarig: nit: since we are anyway touching this. 
we should change it to 
```
...
  import { Fragment } from 'react';
....
  {execTimeRows.map((row) => (
    <Fragment key={row.group}>
```

frontend/src/api/trace/getTraceAggregations.ts
aks07: Both endpoints are POSTs... so orval generates them as useMutation hooks. This feature needs query semantics ...the aggregations fetch is gated on enabled and is cached.
For now let keep pattern the page already uses (useGetTraceV3 was hand-written too). Also worth noting the waterfall hook stays a thin wrapper either way getTraceV4 does pre/post-processing.

frontend/src/types/api/trace/getTraceV3.ts
aks07: Yes this terminology has become really confusing, we had originally kept everything v3 and now the waterfall api has moved to v4..will get cleaned up when we get rid of old v2 code.

frontend/src/api/trace/getTraceAggregations.ts
H4ad: Orval generates the hook and the function, you can replace your axios call and types to use the generated ones.

H4ad: https://github.com/SigNoz/signoz/blob/main/frontend/src/api/generated/services/tracedetail/index.ts#L29-L45

frontend/src/api/trace/getTraceV4.tsx
aks07: now using the generated version. but we do need this wrapper currently to support both flows

frontend/src/hooks/trace/useGetTraceAggregations.tsx
aks07: changed to using the getter function. still need the hooks

H4ad (approved):
Follow-ups:
- Ensure no custom type is defined for apis, use generated
- for AnalyticsTabContent, extract the css used to its own file instead of sharing with analyticspanel.module.scss

tewarig (approved):

PR #11559, opened 2026-06-03
chore: replace perses/perses with perses/spec and introduce kind enum values

namanv3:
## Pull Request

---

### 📄 Summary

Perses has just deprecated a lot of things in the perses/perses repo right now, and is commenting to use the perses/spec repo instead.
https://github.com/perses/perses/pull/3932/changes#diff-f442f80a9107499b02697a86172a24b635a707b23981fef89c901425657eae1a

This also adds allowed values for the Query and Panel Kinds

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Unit tests updated
- Manual verification: Create/Get flows tested
- Edge cases covered:

---


pkg/types/dashboardtypes/perses_replicas.go
```diff
@@ -27,26 +29,66 @@ type DatasourceSpec struct {
...
 
 type Panel struct {
-	Kind string    `json:"kind"`
+	Kind PanelKind `json:"kind"`
 	Spec PanelSpec `json:"spec"`
 }
 
+// PanelKind is the panel envelope discriminator. Perses leaves it a free
+// string; SigNoz locks it to the single valid value.
+type PanelKind string
```
srikanthccv, line 37: use valuer.String

pkg/types/dashboardtypes/perses_replicas.go
```diff
@@ -27,26 +29,66 @@ type DatasourceSpec struct {
...
 // Query
 // ══════════════════════════════════════════════
 
 type Query struct {
-	Kind string    `json:"kind"`
+	Kind QueryKind `json:"kind"`
 	Spec QuerySpec `json:"spec"`
 }
 
+type QueryKind qb.RequestType
```
srikanthccv: do we really need to create type alias here?

pkg/types/dashboardtypes/perses_replicas.go
namanv3: Need a capital P in Panel to match Perses schema, that's why didn't use valuer string. Same reason as Perses plugin kinds, where we need capital letters
https://github.com/SigNoz/signoz/blob/81eadac3a83ebe6d9321b723ea805845152d4c91/pkg/types/dashboardtypes/perses_signoz_plugins.go#L18

pkg/types/dashboardtypes/perses_replicas.go
namanv3: qb.RequestType doesn't reject unknown values, but dashboards should. That's why a custom unmarshal function was needed, so I created a type alias here

srikanthccv: is that the only reason? If so, we should update the `RequestType` to throw unsupported error. The statement builder already do that today (each have their own supported types, example trace operation can't be for metrics), so that logic will continue to be there.

namanv3: done

therealpandey (approved):

srikanthccv (approved):

PR #11578, opened 2026-06-03
fix(trace-flamegraph): prevent layout hang on very wide traces

aks07:
## Summary

The trace flamegraph hung (~11 min, effectively never rendered) on very **wide** traces — e.g. a single span with tens of thousands of direct children. This fixes the super-linear spiral in the row-placement algorithm so layout completes in **~0.2 s** for such traces.

Full root-cause analysis in the issue: **SigNoz/engineering-pod#5219**.

#### Issues closed by this PR

Closes SigNoz/engineering-pod#5219

---

### ✅ Change Type

- [x] 🐛 Bug fix

---

#### Root Cause
Placement did two jobs per span: overlap avoidance (correctness) and **connector-line avoidance** (cosmetic — keep lines from crossing span boxes). On a very wide sibling group the cosmetic part forms a **positive-feedback loop**: each pushed-down child stamps connector points on intermediate rows, which push later children even higher, which stamp more points… A layout needing ~48 rows inflated to **6,000+**, at **O(N²)+** (never terminates at ~92k siblings). Each check is also a linear scan, compounding the cost. (The worker timeout was also misconfigured to 25 min, so it spun instead of failing fast.)

#### Fix Strategy
Split placement at a sibling-group-size threshold; small/normal groups keep the original path unchanged. For groups above the threshold:
1. **Drop the connector-avoidance checks** — meaningless at that density (the row is a solid wall) and the source of the spiral. This alone makes layout *complete* (rows back to ~48).
2. **Place by overlap only via a segment tree** — siblings are processed in descending start order, so a row is free iff its minimum start ≥ the span's end; the tree answers "lowest free row" in **O(log rows)** instead of a linear scan. This makes it *fast*.

Result: **O(N²)+ → O(N log N)**; same overlap packing (no two time-overlapping spans share a row). Worker timeout corrected to 15 s so any future pathological trace fails fast.

---

### 🧪 Testing Strategy

- **Tests added:** 3 wide-group tests — thousands of sequential siblings → 1 row; thousands of fully-overlapping → N rows; mixed group → subtree adjacency preserved.
- **Existing:** `computeVisualLayout` 16/16, full flamegraph suite 82/82; `tsgo --noEmit` and `oxlint` clean.
- **Edge cases / real data:** verified on a real ~92k-span wide trace — ~11 min → ~0.2 s, rows 6,000+ → 49, **zero** same-row overlaps.

---

- **Blast radius:** trace flamegraph layout only.
- **Potential regressions:** none expected for normal/deep traces — the `≤ threshold` path is the original code unchanged, and all pre-existing tests pass. Behavior changes only for very wide groups (where connector lines may now cross span boxes — invisible at that density).
- **Rollback:** revert this PR.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Trace flamegraph now renders on very large/wide traces instead of hanging. |


frontend/src/pages/TraceDetailsV3/TraceFlamegraph/__tests__/computeVisualLayout.test.ts
```diff
@@ -472,4 +472,98 @@ describe('computeVisualLayout', () => {
...
+		});
+
+		const layout = computeVisualLayout([[root], kids, [grandchild]]);
+
+		const parentRow = layout.spanToVisualRow['k500'];
+		const gcRow = layout.spanToVisualRow['gc'];
+		expect(gcRow - parentRow).toBe(1); // subtree adjacency preserved
+		expect(Object.keys(layout.spanToVisualRow)).toHaveLength(1002);
+		noRowHasOverlap(layout);
+	});
```
H4ad, line 571: These tests does not fail without the fix, they should fail otherwise we can introduce this bug again.

aks07: The algorithm did not fail but timed out so it will pass in any case. We cannot mimic the timeout here but we can compare the new path is faster. added a test for the same.

H4ad (approved):

PR #11588, opened 2026-06-04
fix(authn): include base path in SSO callback and error-redirect URLs

vikrantgupta25:
## Summary

- base path handling for SSO callback and error redirect URLs
- added integration tests for base path 

closes: https://github.com/SigNoz/platform-pod/issues/2413


tests/fixtures/signoz.py
```diff
@@ -86,6 +89,12 @@ def create() -> types.SigNoz:
         if with_web:
             env["SIGNOZ_WEB_ENABLED"] = True
 
+        if base_path:
```
vikrantgupta25: there is a chicken egg problem here. Since only the path is relevant here used localhost

therealpandey (approved):

PR #11592, opened 2026-06-05
chore: add normalize for QueryRangeRequest

nityanandagohain:
## Pull Request

---

### 📄 Summary
Refactors what we did in this PR https://github.com/SigNoz/signoz/pull/11483/changes accroding to the comments here https://github.com/SigNoz/engineering-pod/issues/5123

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/5123

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: ✅
- Manual verification: ✅
- Edge cases covered:

---

- Blast radius:None
- Potential regressions:
- Rollback plan: revert


pkg/querier/querier.go
```diff
@@ -86,11 +86,12 @@ func New(
...
-	// consumer (TimeRange, narrowWindowByTraceID, step interval, etc.) can
-	// safely assume ms regardless of the resolution the caller sent.
-	req.Start = querybuilder.ToMilliSecs(req.Start)
-	req.End = querybuilder.ToMilliSecs(req.End)
+	// Normalize Start/End to ms. UnmarshalJSON covers HTTP requests; callers
+	// that build the request programmatically skip it, so this is the catch-all
+	// (idempotent for the already-normalized path).
+	if err := req.Normalize(); err != nil {
+		return nil, err
+	}
```
tushar-signoz: This can be removed as this is no-op right, cause Unmarshal would have already called req.Normalize()

pkg/querier/querier.go
```diff
@@ -427,10 +428,12 @@ func (q *querier) resolveMetricMetadata(ctx context.Context, queries []qbtypes.Q
...
-	// Coerce the window to epoch milliseconds up front (End may be 0 for the
-	// open-ended stream, which ToMilliSecs leaves untouched).
-	req.Start = querybuilder.ToMilliSecs(req.Start)
-	req.End = querybuilder.ToMilliSecs(req.End)
+	// Catch-all normalization for programmatic callers (see QueryRange). End is
+	// 0 here for the open-ended stream, which Normalize leaves untouched.
+	if err := req.Normalize(); err != nil {
+		client.Error <- err
+		return
+	}
```
tushar-signoz: same

pkg/querier/querier.go
nityanandagohain: Nope, we call QueryRange() from other places, where the request is formed internally so for that this is there.

tushar-signoz: If that is the case, we need to identify and correct the caller Otherwise this would break things 😢 
edit: it won't break stuff, but it shouldn't be QueryRange's responsibility.

nityanandagohain: Why would it break? it will throw an error right if they are sending something wrong, and if it's sending in seconds/ns etc it will convert it to ms

tushar-signoz (dismissed):
LGTM

tushar-signoz:
@nityanandagohain I need the normalise function

PR #11593, opened 2026-06-05
fix: mark numeric columns as aggregation in scalar query response

### LINKED ISSUE #8844: [Bug] Pie Chart with ClickHouse Query only works if aggregation alias is `__result_[n]`

tushar-signoz:
## Pull Request

---

### 📄 Summary

Scalar query responses classify each result column as either an **aggregation** (a value) or a **group** (a dimension/label). Previously `readAsScalar` treated a column as an aggregation **only** when its alias matched the builder's `__result_N` convention.

That assumption holds for builder queries, but **raw ClickHouse SQL queries use arbitrary, user-controlled aliases**. A numeric aggregation column written as `SELECT count() AS value` (or any non-`__result_N` alias) was misclassified as a group-by key. On the UI this shifts aggregation indices and drops the column's real meaning.

This PR additionally classifies a column as an aggregation when its **scan type is numeric**, so raw ClickHouse numeric columns are handled correctly regardless of alias.

#### Issues closed by this PR

N/A

---

### ✅ Change Type

- [x] 🐛 Bug fix
- [x] 🧪 Test-only (integration coverage added)

---

#### Root Cause

`readAsScalar` classified result columns purely by alias (`^__result_(\d+)$`). Raw ClickHouse queries don't follow that aliasing convention, so their numeric aggregation columns fell through to the group-by branch.

A secondary issue: ClickHouse `Nullable(T)` columns surface through the driver with a **pointer scan type** (`*float64`, `*int64`, …). A naive `reflect.Kind` check sees `Ptr` (not numeric) and misclassifies nullable numeric columns even when the numeric gate is added.

#### Fix Strategy

In `pkg/querier/consume.go`:
- Classify a scalar column as an aggregation when `aggRe.MatchString(name) || isNumericKind(scanType)`.
- Refactor the numeric check into `isNumericKind(reflect.Type)`, which **unwraps pointer types** (`reflect.Ptr` / `reflect.UnsafePointer`) before testing the underlying kind, so `Nullable(...)` numeric columns are detected correctly. The same helper is reused in `readAsTimeSeries`.

> Note: this intentionally assumes non-numeric ⇒ group-by for raw ClickHouse queries. A user grouping by an integer column (e.g. `status_code`) would be classified as an aggregation. This is an accepted tradeoff for now; a proper fix would require the frontend to supply a column-name → column-type mapping.

---

### 🧪 Testing Strategy

- **Tests added:** `tests/integration/tests/querier/17_clickhouse_queries.py` — exercises classification directly against a live ClickHouse:
  - A numeric column aliased `__result_0` → classified as `aggregation`.
  - A `Nullable(Float64)` column under a non-`__result_N` alias (`SELECT toNullable(toFloat64(1.5)) AS value`) → classified as `aggregation`. This isolates the pointer-unwrap path: the alias doesn't match the regex, so classification can only come from the numeric scan-type check.
- **Regression verified:** with the pointer-unwrap removed and the backend rebuilt, the nullable case fails with `assert 'group' == 'aggregation'`; with the fix it passes. The `__result_N` case stays green either way, confirming the new assertion pins the pointer behavior specifically.

---

- **Blast radius:** scalar response shaping for raw ClickHouse queries only (`readAsScalar`). Builder queries already matched via the `__result_N` branch and are unaffected.
- **Potential regressions:** a raw query that groups by a numeric column will now be labeled `aggregation` instead of `group` (documented tradeoff above).
- **Rollback plan:** revert this PR; behavior returns to alias-only classification.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Raw ClickHouse scalar queries now correctly classify numeric (including `Nullable`) result columns as aggregations regardless of column alias. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

The core decision is the fallback heuristic for raw ClickHouse queries: alias `__result_N` ⇒ aggregation, otherwise numeric scan type ⇒ aggregation, else group-by. See the inline comment in `readAsScalar` for the reasoning and the known limitation around numeric group-by keys.


pkg/querier/consume.go
```diff
@@ -290,7 +296,13 @@ func readAsScalar(rows driver.Rows, queryName string) (*qbtypes.ScalarData, erro
 	var aggIndex int64
 	for i, name := range colNames {
 		colType := qbtypes.ColumnTypeGroup
-		if aggRe.MatchString(name) {
+		// Builder queries aliases aggregation columns as __result_N (always numeric) and wraps group-by keys with toString (always string);
+		// Raw ClickHouse queries may use any aliases.
+		// Handling Builder queries, If name like __result_N -> aggregation, otherwise group-by column
+		// Handling Raw ClickHouse queries, If type is numeric -> aggregation, otherwise group-by column
+		// NOTE: For clickhouse queries, its wrong to assume that numeric columns are always aggregations, user might be grouping by on integer status_code.
```
nityanandagohain, line 303: this can be a very common case if I am not wrong.

So now if there are numeric group by in a ch query previously it would not render aggregation, but now it may render something as aggregation but actually it's a group by right ?

nityanandagohain: The statement seems wrong here we don't allow group by on scalar queries in logs and traces, if that is the case in ch sql then better return an error ?

but we do allow in metrics so something to dig more ?

Image

nityanandagohain (commented):
qq

pkg/querier/consume.go
tushar-signoz: it depends on the panel type if groupBy is allowed or not. The api call doesn't get the information as to which panel type it will be used for, so backend as to be generic about it.

nityanandagohain (approved):
So now the annotation of clickhouse sql queries will work which has any number of group by string columns, but if there is a number in the group by it won't work. We can go ahead with this and then create a ticket to solve this permanently after deciding the priority. Where it solves for group by on numbers as well.

therealpandey (approved):

PR #11603, opened 2026-06-06
fix(infra): use hostname instead of host.name

### LINKED ISSUE #11314: Inrstructure tab shows no logs as it's using host.name instead of hostname
## Bug description

When clicking on the Infrastructure tab I can see all my hosts, when I click on one host and go to the Logs tab it shows nothing.  When I click on the navigation button at the top right it brings me to the Logs page but shows it's trying to query host.name which is a key that is not found?  I've confirmed under the Logs tab directly that everything is working and logs are showing up.  Seems like it needs to being using the key hostname instead.

Image

Image

## Expected behavior

Logs should be showing up under the Infrastructure tab for each host.

## How to reproduce

1.Click into Infrastructure tab
2. Click on a host
3.Select logs.

## Version information 
* **Signoz version**: 1.22.0
* **Browser version**: Firefox 150.0.1
* **Your OS and version**: Fedora Workstation 44
* **Your CPU Architecture**(ARM/Intel): Intel 64 bit

pandareen:
## Pull Request

---

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

After this change:

 - The Logs tab on a host detail drawer should show logs for that host.
 - The compass button to open Logs Explorer should query `hostname = "<host>"` instead of `host.name`.

---

Fixes #11314

### DIFF
--- a/frontend/src/container/InfraMonitoringHosts/constants.tsx
+++ b/frontend/src/container/InfraMonitoringHosts/constants.tsx
@@ -120,12 +120,8 @@ export function hostGetSelectedItemFilters(
 	};
 }
 
-export function hostInitialLogTracesFilter(
-	host: HostData,
-	dotMetricsEnabled: boolean,
-): TagFilterItem[] {
-	const hostKey = dotMetricsEnabled ? 'host.name' : 'host_name';
-	return [createFilterItem(hostKey, host.hostName || '')];
+export function hostInitialLogTracesFilter(host: HostData): TagFilterItem[] {
+	return [createFilterItem('hostname', host.hostName || '')];
 }
 
 export function hostInitialEventsFilter(_host: HostData): TagFilterItem[] {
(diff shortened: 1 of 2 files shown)


frontend/src/container/InfraMonitoringHosts/constants.tsx
```diff
@@ -120,12 +120,8 @@ export function hostGetSelectedItemFilters(
 	};
 }
 
-export function hostInitialLogTracesFilter(
```
pandareen, line 123: The `dotMetricsEnabled` flag was removed from `hostInitialLogTracesFilter` because it only applies to metrics, not logs or traces.

Logs and traces use `hostname` regardless of that flag (same idea as K8s entities: pod logs use `k8s.pod.name`, not a metrics-style underscore variant). So passing `dotMetricsEnabled` into the `logs/traces` filter was both wrong and unnecessary.

PR #11605, opened 2026-06-07
fix(querier): cap body group-by results to prevent OOM

tushar-signoz:
## Pull Request

---

Grouping by a **body-context column** (a `body.*` JSON field on logs) is effectively unbounded in cardinality — each log line can contribute a distinct group. Such a query returns a huge number of series/rows and can OOM the SigNoz server.

It puts a safety cap on body group-by queries that don't already set an explicit limit, while telling the user it happened:

- For **scalar / time series** queries that group by a body-context column and have **no explicit limit**, the query is run with `LIMIT 1000 + 1`.
- After execution, if a capped query came back with more than 1000 groups, the result is **trimmed to 1000** and a **warning** is attached to the response (`"Query <name> returned too many groups; results have been limited to 1000. Add an explicit limit to the query to override this."`).
- The cap is applied **before** the SQL is built (so the database enforces it), and the trim/warning runs on the **raw results before post-processing** (which can reshape them via table formatting / formulas and make group counts unreliable).

Only body-context group-bys are affected for now — every other query (other group-bys, queries with an explicit limit, raw/live-tail) is left untouched. The mechanism is generic so it can be opened up to more group-bys later by relaxing a single gate.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
Body group-by has very high (effectively unbounded) cardinality. With no default limit, a body group-by query materializes a series/row per distinct body value, which can exhaust memory and OOM the server.

#### Fix Strategy
- `applyGroupByLimit` (in `QueryRange`, before `buildQueries`): for scalar/timeseries queries with no explicit limit that group by a body-context column, set `LIMIT groupByLimit + 1` and record the capped query names.
- `enforceGroupByLimit` / `trimGroupsToLimit` (inside `run`, before post-processing): trim any capped query that returned more than `groupByLimit` groups down to the cap and emit a warning per trimmed query into `resp.Warning`.
- `QueryEnvelope.HasBodyContextGroupBy()` (new helper): reports whether any group-by column has `FieldContext == body`. This is the single gate that keeps the behavior scoped to body group-by for now.

`groupByLimit = 1000`. Using `limit + 1` lets us distinguish "exactly at the cap" from "over the cap" so the warning only fires when results were actually dropped.

---

- Tests added/updated: none yet — pure helpers (`applyGroupByLimit`, `trimGroupsToLimit`) are straightforward to unit test and can be added on request.
- Manual verification: `go build ./...`, `gofmt`, `go vet ./pkg/querier/...`, and existing `pkg/querier` + `pkg/types/querybuildertypes/querybuildertypesv5` tests pass.
- Edge cases covered: explicit-limit queries skipped; non-body group-bys skipped; raw / live-tail path unaffected; both `TimeSeriesData` (per-aggregation `Series`) and `ScalarData` (`Data` rows) trimmed.

---

- Blast radius: scalar/timeseries builder queries that group by a body-context column and don't set a limit. All other queries are untouched.
- Potential regressions: a body group-by previously returning >1000 groups now returns at most 1000 (with a warning) instead of everything. This is the intended behavior change.
- Rollback plan: revert this PR — there is no schema/state change.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Body group-by queries without an explicit limit are now capped at 1000 results, with a warning when the cap is applied, to prevent the server from OOMing on high-cardinality body group-bys. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- The limit mechanism is intentionally generic (`applyGroupByLimit` / `enforceGroupByLimit` / `trimGroupsToLimit`); scope is restricted to body group-by by the `HasBodyContextGroupBy()` gate, so broadening it later is a one-line change.
- The trim runs on raw results **before** post-processing on purpose — post-processing (table formatting, formulas, fill-gaps) can reshape results and make group counts unreliable.

---

### DIFF
--- a/pkg/types/querybuildertypes/querybuildertypesv5/req.go
+++ b/pkg/types/querybuildertypes/querybuildertypesv5/req.go
@@ -466,6 +466,65 @@ func (q *QueryEnvelope) UseDefaultOrderByForListQuery() {
 	}
 }
 
+// DefaultGroupByLimit caps groups for a group by query with no explicit limit. The query runs
+// with one extra row so the querier can detect when the cap was hit and warn.
+const DefaultGroupByLimit = 1000
+
+// Normalize prepares scalar/time series group by queries: it makes the implicit order explicit
+// and caps unbounded body group by queries at DefaultGroupByLimit+1 rows, recording the implicit
+// limit so the querier can trim and warn.
+func (r *QueryRangeRequest) Normalize() {
+	if r.RequestType != RequestTypeScalar && r.RequestType != RequestTypeTimeSeries {
+		return
+	}
+
+	for idx := range r.CompositeQuery.Queries {
+		qe := &r.CompositeQuery.Queries[idx]
+		qe.UseDefaultGroupByOrder()
+
+		// Only body group by is capped for now; relax this gate to cap others.
+		if qe.GetLimit() == 0 && qe.HasBodyContextGroupBy() {
+			qe.SetImplicitLimit(DefaultGroupByLimit)
+			qe.SetLimit(DefaultGroupByLimit + 1)
+		}
+	}
+}
+
+// UseDefaultGroupByOrder defaults a log/trace group by query's order to the first aggregation
+// descending (the SQL default) when none is set. Metrics order themselves.
+func (q *QueryEnvelope) UseDefaultGroupByOrder() {
+
+	if len(q.GetGroupBy()) == 0 {
+		return
+	}
+
+	var aggExpr string
+
+	switch spec := q.Spec.(type) {
+	case QueryBuilderQuery[LogAggregation]:
+		if len(spec.Aggregations) > 0 {
+			aggExpr = spec.Aggregations[0].Expression
+		}
+	case QueryBuilderQuery[TraceAggregation]:
+		if len(spec.Aggregations) > 0 {
+			aggExpr = spec.Aggregations[0].Expression
+		}
+	default:
+		return
+	}
+
+	order := q.GetOrder()
+
+	if len(order) == 0 && aggExpr != "" {
+		order = append(order, OrderBy{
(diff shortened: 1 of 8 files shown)


tushar-signoz:
The lint error is some setup failure of the job

pkg/types/querybuildertypes/querybuildertypesv5/req.go
```diff
@@ -549,6 +549,20 @@ func (r *QueryRangeRequest) SkipFillGaps(name string) bool {
...
 }
 
+func (r *QueryRangeRequest) Normalize() error {
+	// Add a default limit for group by if not provided
+	// TODO(tushar): This is a bad fix, because this is silently adding a default limit to the query without the user's knowledge.
+	// We should remove this once we have a better way to handle this.
+	if r.RequestType == RequestTypeScalar || r.RequestType == RequestTypeTimeSeries {
+		for idx := range r.CompositeQuery.Queries {
+			if r.CompositeQuery.Queries[idx].GetLimit() == 0 && len(r.CompositeQuery.Queries[idx].GetGroupBy()) > 0 {
+				r.CompositeQuery.Queries[idx].SetLimit(1000)
```
nityanandagohain: I am not sure if applying this to all queries is correct, as there can be some queries which are already running with more then 1000 keys in group by, so on backend maybe just start with body ? And on the frontend add default limit of 1000 so that any new dashboard panel getting created/updated will have this limit known from the UI side itself  and existing ones will work as it is

Issue #5051, opened 2024-05-22
Change of env does not update data in the service overview page


ankitnayan:
I'm using the Services option to look at my application, but I noticed the following:
If I change the environment, the panels do not update unless I navigate to a different and return to my original tab. Ideally after an env change all panels should update with the appropriate data

Vikrant2520:
this is fixed under this PR :- https://github.com/SigNoz/signoz/pull/5040

PR #11614, opened 2026-06-08
feat(traces): migrate list + traces views to TanStack table

aks07:
## Pull Request

---

### 📄 Summary

Migrates Trace Explorer's **List view** and **Traces (grouped) view** off antd `ResizeTable` onto the shared `TanStackTable` (`components/TanStackTableView/`) — same renderer logs uses. Both views inherit virtualized infinite scroll, draggable + resizable columns, and the shared field-selector pipeline.

What changed (high level):

- **Renderer swap** — both views now use `TanStackTable`. List view drives column membership/order via `selectColumns` writers (`config?.addColumn?.onReorder` / `onRemove`); grouped view persists sizing + order in the TanStack column-store (`LOCALSTORAGE.TRACES_VIEW_COLUMNS`).
- **Pagination model** — URL `pagination` param dropped for these two views; replaced with `react-virtuoso` infinite scroll via `onEndReached`. The bottom `Controls` toolbar's pagination block is hidden for them.
- **Shared primitives** — extracted to `components/Traces/TableView/`:
  - `useTraceInfiniteQuery` — owns pagination + accumulator + reset/append + queryKey + `useGetQueryRange` wiring.
  - `TracesTable` — wrapper that renders the empty/loading/error states + the table; takes `getRowHref` for link-style nav.
  - `useTracesTableColumns` — column factory; `baseColumns` per view + optional dynamic `fields` from `selectColumns`.
  - `getTraceLink` — lifted out so legacy antd consumers (`TracesTableComponent`, `EntityTraces`) can share it.
- **Page layout fix** — `TracesModulePage` wraps the explorer in antd `Tabs` (Explorer / Funnels / Saved Views) which had no flex-column setup; this caused the **whole page to scroll** when the quick-filter content grew tall. Set up the chain (`.traces-module-container` → `.ant-tabs` → `.ant-tabs-content-holder` → `.ant-tabs-tabpane-active` → `.trace-explorer-page`) as bounded flex columns so the filter and table scroll inside their own panes (matches logs).
- **Filter wrapper** — replaced antd `<Card>` with plain `<section>` for the filter sidebar; the Card's `.ant-card-body` was breaking the height-inheritance chain. Mirrors logs.
- **Row identity fix** — TanStack falls back to positional row IDs when rows lack an `id` field, which made the table skip refreshing cells on orderBy toggles (cached transitions). Inject `id = span_id` (List) / `id = trace_id` (Grouped) into the response transform — TanStack now sees genuine row changes.
- **Composite column ID extended** — `buildCompositeKey(name, context, dataType?)` now takes optional `dataType` to disambiguate same-name fields shipped as both number + string (e.g. `http.status_code`). Logs callers stay on the 2-arg form.

Issues:

- https://github.com/SigNoz/engineering-pod/issues/5052

https://github.com/user-attachments/assets/4bbf93ac-d001-4d28-897a-a805b59e7654

---

### ✅ Change Type

- [x] ✨ Feature — column resize/reorder + virtualized infinite scroll for these two views.
- [x] ♻️ Refactor — renderer swap, primitives extraction, layout chain.

---

### 🧪 Testing Strategy

- Manual: list view + grouped view smoke; orderBy toggle (asc ↔ desc) refreshes rows; column add/remove via field selector; column resize persisted across reload; infinite scroll past page 1; quick-filter scrolls in its own pane on tall content.
- Tests added/updated: existing `ListView/index.test.tsx` updated; unit tests for the shared hook + column adapter are tracked as follow-up in the sprint task.
- Edge cases: empty rows, fetch error, query cancel, view-switch mid-fetch.

---

- **Blast radius:** Trace Explorer's List + Traces views only. TableView and TimeSeriesView untouched. Logs untouched.
- **Potential regressions:** column persistence (new Zustand prefix; greenfield, no migration). Infinite-scroll termination uses a length heuristic (`newRows.length >= PAGE_SIZE`) — known to trail by one empty fetch on exact multiples; non-blocking.
- **Rollback:** revert this PR. No new storage keys are written until users land on the view.

---

### 📝 Changelog

| Field           | Value                                                                                                          |
| --------------- | -------------------------------------------------------------------------------------------------------------- |
| Deployment Type | Cloud / OSS / Enterprise                                                                                       |
| Change Type     | Feature                                                                                                        |
| Description     | Trace Explorer list and traces (grouped) views now use the shared table — resizable and draggable columns, virtualized infinite scroll. |

---

### 📋 Checklist

- [x] Tests added or explicitly not required (existing list-view tests updated; new unit tests for shared primitives tracked as follow-up)
- [x] Manually tested
- [x] Breaking changes documented (none — URL `pagination` param removal for these two views is the only visible behavior change)
- [x] Backward compatibility considered (`buildCompositeKey` `dataType` arg is optional → logs callers unaffected)

---

## 👀 Notes for Reviewers

- The `tableResetKey` approach (forcing `<TanStackTable>` remount) was tried first to fix the orderBy-toggle row-stickiness, then replaced with the cleaner `id` injection in the transform — TanStack's `getRowId` picks up real identity. No remount needed.
- `useTraceInfiniteQuery` uses `keepPreviousData: true` and runs the append effect off `data?.payload` only (pagination state changes drive the queryKey → drive a new fetch → arrives here as a fresh payload). Pagination is intentionally not in the effect deps.
- Legacy `getListColumns` + `BlockLink` in `container/TracesExplorer/ListView/utils.tsx` are retained — `TracesTableComponent` (dashboard widget) and `InfraMonitoring/EntityTraces` are still on antd `ResizeTable` and consume them. New code uses the TanStack `make*Col` factories in the same file.
- `nextCursor` is on the response shape but the hook ignores it for now (uses length heuristic for `hasMore`). Revisit when the span drawer work converges the response envelope (#5051).


frontend/src/components/Traces/TableView/TracesTable.tsx
```diff
@@ -0,0 +1,113 @@
...
+		},
+		[history, getRowHref],
+	);
+
+	const onRowClickNewTab = useCallback(
+		(row: TRow): void => {
+			window.open(
+				getAbsoluteUrl(getRowHref(row)),
+				'_blank',
+				'noopener,noreferrer',
```
H4ad, line 74: noopener here is enough, no need to noreferrer

frontend/src/container/InfraMonitoringK8s/EntityDetailsUtils/EntityTraces/traceListColumns.tsx
```diff
@@ -3,10 +3,8 @@ import { Badge } from '@signozhq/ui/badge';
 import { Typography } from '@signozhq/ui/typography';
 import { DATE_TIME_FORMATS } from 'constants/dateTimeFormats';
 import { getMs } from 'container/Trace/Filters/Panel/PanelBody/Duration/util';
-import {
-	BlockLink,
-	getTraceLink,
-} from 'container/TracesExplorer/ListView/utils';
+import { getTraceLink } from 'components/Traces/TableView/getTraceLink';
+import { BlockLink } from 'container/TracesExplorer/ListView/utils';
```
H4ad, line 7: Eventually, we can move this to some shared place we can consume in multiple places, I don't think is healthy to keep one page importing helpers of another page.

frontend/src/container/TracesExplorer/ListView/utils.tsx
```diff
@@ -136,3 +155,111 @@ export const getListColumns = (
...
+
+export function makeListFieldCol(
+	f: TelemetryFieldKey,
+): TableColumnDef<SpanRow> {
+	return {
+		id: buildCompositeKey(f.name, f.fieldContext, f.fieldDataType),
+		header: f.name,
+		accessorFn: (row): unknown => row[f.name],
+		enableRemove: true,
+		width: { min: 192 },
```
H4ad, line 241: We can add `default: '100%'`, it will look better:

Image

frontend/src/container/TracesExplorer/ListView/utils.tsx
```diff
@@ -41,12 +50,22 @@ export const transformDataWithDate = (
...
 
-export const getTraceLink = (record: RowData): string =>
-	`${ROUTES.TRACE}/${record.traceID || record.trace_id}${formUrlParams({
-		spanId: record.spanID || record.span_id,
-		levelUp: 0,
-		levelDown: 0,
-	})}`;
+// Re-export for legacy antd consumers (TracesTableComponent, EntityTraces) that
+// import from this path. New code should import from
+// `components/Traces/TableView/getTraceLink`.
```
H4ad, line 55: Remove this

frontend/src/container/TracesExplorer/ListView/utils.tsx
```diff
@@ -136,3 +155,111 @@ export const getListColumns = (
...
+	f: TelemetryFieldKey,
+): TableColumnDef<SpanRow> {
+	return {
+		id: buildCompositeKey(f.name, f.fieldContext, f.fieldDataType),
+		header: f.name,
+		accessorFn: (row): unknown => row[f.name],
+		enableRemove: true,
+		width: { min: 192 },
+		cell: ({ value }): JSX.Element => {
+			if (value === '' || value == null) {
```
H4ad, line 243: Replace `==` to `===` and include `undefined`.

https://dorey.github.io/JavaScript-Equality-Table/

frontend/src/container/TracesExplorer/ListView/utils.tsx
```diff
@@ -136,3 +155,111 @@ export const getListColumns = (
...
+		cell: ({ value }): JSX.Element => {
+			const ts = value as string | number;
+			const formatted =
+				typeof ts === 'string'
+					? formatTimezoneAdjustedTimestamp(ts, DATE_TIME_FORMATS.ISO_DATETIME_MS)
+					: formatTimezoneAdjustedTimestamp(
+							ts / 1e6,
+							DATE_TIME_FORMATS.ISO_DATETIME_MS,
+						);
+			return <TanStackTable.Text>{String(formatted)}</TanStackTable.Text>;
```
H4ad, line 228: This can have a testid to be asserted in tests

frontend/src/container/TracesExplorer/TracesView/configs.tsx
```diff
@@ -1,50 +1,72 @@
...
-		title: 'Root Duration (in ms)',
-		dataIndex: 'duration_nano',
-		key: 'durationNano',
-		render: (duration: number): JSX.Element => (
-			<Typography>{getMs(String(duration))}ms</Typography>
+		id: 'duration_nano',
+		header: 'Root Duration (in ms)',
+		accessorFn: (row): unknown => row.duration_nano,
+		cell: ({ value }): JSX.Element => (
+			<TanStackTable.Text>{getMs(String(value))}ms</TanStackTable.Text>
```
H4ad, line 50: Ensure all these columns have testid, to assert on tests

frontend/src/components/Traces/TableView/useTracesTableColumns.tsx
```diff
@@ -0,0 +1,58 @@
...
+	);
+}
+
+function makeUserFieldCol<TRow>(f: TelemetryFieldKey): TableColumnDef<TRow> {
+	const col: TableColumnDef<Record<string, unknown>> = {
+		id: buildCompositeKey(f.name, f.fieldContext, f.fieldDataType),
+		header: f.name,
+		accessorFn: (row): unknown => row[f.name],
+		enableRemove: true,
+		width: { min: 192 },
```
H4ad, line 39: This can also have `default: '100%'``

frontend/src/components/Traces/TableView/useTracesTableColumns.tsx
```diff
@@ -0,0 +1,58 @@
...
+		return '';
+	}
+	if (typeof value === 'string') {
+		return value;
+	}
+	if (typeof value === 'number' || typeof value === 'boolean') {
+		return String(value);
+	}
+	return JSON.stringify(value);
+}
```
H4ad, line 58: This is duplicated with another file, move to a util file (probably outside components)

frontend/src/container/TracesExplorer/ListView/utils.tsx
```diff
@@ -136,3 +155,111 @@ export const getListColumns = (
...
+			}
+			if (DURATION_FIELD_NAMES.has(f.name)) {
+				return (
+					<TanStackTable.Text data-testid={f.name}>
+						{getMs(text)}
+						ms
+					</TanStackTable.Text>
+				);
+			}
+			return <TanStackTable.Text data-testid={f.name}>{text}</TanStackTable.Text>;
```
H4ad, line 262: We can add `title` to be able to see what is the value of 

Image

frontend/src/components/Traces/TableView/useTracesTableColumns.tsx
```diff
@@ -0,0 +1,58 @@
...
+	baseColumns,
+	fields = [],
+}: UseTracesTableColumnsProps<TRow>): TableColumnDef<TRow>[] {
+	return useMemo<TableColumnDef<TRow>[]>(
+		() => [...baseColumns, ...fields.map((f) => makeUserFieldCol<TRow>(f))],
+		[baseColumns, fields],
+	);
+}
+
+function makeUserFieldCol<TRow>(f: TelemetryFieldKey): TableColumnDef<TRow> {
```
H4ad, line 33: Whats the difference between this and https://github.com/SigNoz/signoz/pull/11614/changes#diff-e9f5d756315ec273023ef37cc70eb4c947f31444b2f3fa207220c8575d55b029R233

We have two ways to build a column for the same table?

frontend/src/components/Traces/TableView/TracesTable.tsx
```diff
@@ -0,0 +1,113 @@
...
+	error: APIError | Error | null;
+	isFilterApplied: boolean;
+	panelType: TracesTablePanelType;
+
+	columnStorageKey: string;
+	respectColumnOrder?: boolean;
+	cellTypographySize?: 'small' | 'medium' | 'large';
+
+	onColumnOrderChange?: (cols: TableColumnDef<TRow>[]) => void;
+	onColumnRemove?: (id: string) => void;
```
H4ad, line 33: For these callbacks, you can just Pick<TanStackTableProps, 'onColumnRemove'> to ensure the type is correct.

frontend/src/components/Traces/TableView/useTraceInfiniteQuery.ts
```diff
@@ -0,0 +1,133 @@
...
+
+	if (queryKeyRef) {
+		queryKeyRef.current = queryKey;
+	}
+
+	const { data, isLoading, isFetching, isError, error } = useGetQueryRange(
+		requestParams,
+		entityVersion,
+		{ queryKey, enabled, keepPreviousData: true },
+	);
```
H4ad, line 86: Since this is a data that does not change, maybe you can add some cacheTime or staleTime to see if we can benefit from caching.

frontend/src/container/TracesExplorer/ListView/index.tsx
```diff
@@ -106,140 +90,92 @@ function ListView({
...
+		isError,
+		error,
+		handleEndReached,
+	} = useTraceInfiniteQuery<SpanRow>({
+		queryDeps: [
+			stagedQuery,
+			panelType,
+			globalSelectedTime,
+			maxTime,
+			minTime,
```
H4ad, line 136: If you use the new globalTime adapter that uses Zustand, you could make this query way more cache-friendly.

H4ad (commented):
About order by, do we really need to have the feature to sort by columns that is not on table? Otherwise, we could just enable the sort via column

---

Due to `Link`, the middle click button worked to open a new tab but now it's not opening a new tab. We can create a ticket to fix this on table component, but this is a regression for traces.

---

We need a copy link to have support for deep link, like logs:

Image

Otherwise, there's no way to share what I'm seeing, in the past we could just share the page, but we don't have it anymore.

---

I think we should enable cache from react-query, open a trace and then go back make us lost the entire scroll we did earlier, the ideal behavior should be go back with the same data + in the same scroll position.

With logs, this is not a problem because it opens in a modal.

---

We should try parse the page/offset to show the same initial data as before, to not break any old link and be more backwards compatible as possible

tewarig:
UX comments - 


can we add skeleton loading here ? 🤔  i think it would be better . spinner is quite slow. 

https://github.com/user-attachments/assets/37b2f81c-6e07-4ac9-a5aa-8397f8e14ab7


nit :  should we add black background on header of this table ? 
https://github.com/user-attachments/assets/8c68cc44-f185-42ee-a0b6-98a90ca2dfa6

nit: Should we add some margin-left so the text in cell starts from the text of header ? 
ideal 
image
now
image

PR #11618, opened 2026-06-08
fix(web): app crash when web speech api access throws

### LINKED ISSUE #11617: Home page fails to load when Web Speech API is disabled
## Bug description

The signoz homepage fails to load in managed/hardened browsers that have the Web Speech API disabled. 

All assets, files, etc. load (no 4xx/5xx errors) but the page is blank. A single console error persists:

    Web Speech API is disabled at useSpeechRecognition.ts:46

## Expected behavior

The home page should load.

## How to reproduce

1. Using latest docker image
2. Boot the app
3. Try to load the homepage `/` or `/home`
4. Observe blank page
5. Check network traffic and see no errors
6. Check web console and see Web Speech API

## Version information 
* **Signoz version**: Latest (docker image)
* **Browser version**: [Island](https://www.island.io/enterprise-browser?) (enterprise managed Chrome)
* **Your OS and version**: macOS Tahoe

## Additional context

1. I'm running the docker image inside of k8s locally (but I don't think this matters)
2. I'm running it SSL locally (but I don't think that matters)
3. I'm running a local SSL cert, untrusted (but I don't think that matters)
4. I have a patch [here](https://github.com/SigNoz/signoz/compare/main...jaydorsey:signoz:jay/web-speech-api-error) that I've tested locally and it seems to resolve the issue

jaydorsey:
## Pull Request

---

`useSpeechRecognition.ts` reads `window.SpeechRecognition` / `window.webkitSpeechRecognition` at module-evaluation time. The `??` null guards against the API being absent, but some hardened/enterprise browsers install a getter that throws on access ("Web Speech API is disabled due to your security policy"). 

Because the read is at module top level, the uncaught throw aborts bundle evaluation and the entire app renders a blank page. This wraps the access in try/catch so a throwing getter degrades to "unsupported" (feature-detection best practice) instead of crashing.

Reproduction steps: Open SigNoz in the a hardened browser → blank page + uncaught error:

`Web Speech API is disabled at useSpeechRecognition.ts:46`

image

Closes #11617

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

Browser edge case

Allows the check to fail gracefully when the Web Speech API isn't available due to security requirements.

---

- Blast radius: Web Speech Feature
- Potential regressions: -

---

| Field | Value |
|------|-------|
| Deployment Type | Cloud |
| Change Type | Bug Fix |
| Description | For more hardened browsers, the home would appear blank due to access to a property that could generate an error without proper error handling. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---

### DIFF
--- a/frontend/src/container/AIAssistant/hooks/useSpeechRecognition.ts
+++ b/frontend/src/container/AIAssistant/hooks/useSpeechRecognition.ts
@@ -40,13 +40,31 @@ type SpeechRecognitionConstructor = new () => ISpeechRecognition;
 
 // ── Vendor-prefix shim for Safari / older browsers ────────────────────────────
 
-const SpeechRecognitionAPI: SpeechRecognitionConstructor | null =
-	typeof window !== 'undefined'
-		? // eslint-disable-next-line @typescript-eslint/no-explicit-any
-			((window as any).SpeechRecognition ??
+// Some hardened/enterprise browsers install a getter
+// on window.SpeechRecognition that THROWS on access ("Web Speech API is disabled
+// due to your security policy") instead of leaving the property undefined.
+// Because this resolves at module-evaluation time, an uncaught throw here aborts
+// the entire bundle and the app renders a blank page. Read defensively so a
+// throwing getter degrades to "unsupported" rather than crashing the app.
+function resolveSpeechRecognitionAPI(): SpeechRecognitionConstructor | null {
+	if (typeof window === 'undefined') {
+		return null;
+	}
+	try {
+		return (
+			// eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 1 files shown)


H4ad (dismissed):

H4ad:
Thanks for the contribution!

H4ad (commented):
@jaydorsey would you mind fixing the fmt issue? You can run `pnpm prettify`.

H4ad (approved):

PR #11622, opened 2026-06-09
fix(billing): fix cancel subscription flow failing silently for users without a default mail client

SagarRajput-7:
## Pull Request

---

### 📄 Summary

The cancel subscription flow silently failed for users without a default mail client configured. When `link.click()` was called on a detached anchor element, some browsers would silently do nothing - the user saw the dialog close with no confirmation and no way to reach support.

This PR fixes the reliability of the mailto trigger and adds a **fallback view** so users always have a path forward even if their mail client doesn't open.

**What changed:**
- `openMailto` now appends the anchor to the DOM before clicking and removes it after - this is the cross-browser-reliable pattern for triggering `mailto:` links
- After clicking "Cancel subscription", the dialog transitions to a **fallback view** instead of closing
- The fallback view shows the support email (`cloud-support@signoz.io`), a **Copy email template** button, and a **Reopen email client** button
- Added a `MAX_MAILTO_URI_LENGTH = 1800` guard - if the encoded URI is too long (possible with long org names), it falls back to a shorter body so the link still works
- Added `data-testid` attributes to the confirm input and action buttons for stable test targeting
- Tests updated and expanded to cover: anchor DOM-attach/detach, fallback view rendering, copy-to-clipboard, retry mailto, and close/reset

----

#### Full flow 
----

https://github.com/user-attachments/assets/d78a119e-402b-42ed-b856-c17c0e9be371

----

image

----

#### Fallback view 

----

image

----

#### Issues closed by this PR
Closes: https://github.com/SigNoz/platform-pod/issues/2442

---

### ✅ Change Type

- [x] 🐛 Bug fix

---

#### Root Cause

The old code called `link.click()` on an anchor that was **never appended to the document**. Detached element clicks are ignored by some browsers (notably Safari and certain Chromium builds) when the target is a `mailto:` URI. The dialog closed immediately with no error - users had no way to know the action failed.

A secondary edge case: very long org names could push the `mailto:` URI over browser length limits, silently truncating the pre-filled body.

#### Fix Strategy

1. **DOM-attach pattern** - append → click → remove. This is the standard workaround for reliable `mailto:` / `blob:` link triggering.
2. **Fallback view** - instead of closing the dialog after the mailto fires, switch to a `'fallback'` dialog view. Users who _did_ get their mail client opened see a confirmation; users who didn't have a clear path (copy template, retry).
3. **URI length guard** - if the full URI exceeds 1800 chars, use a short-form body so the link always resolves.

---

### 🧪 Testing Strategy

- **Tests added/updated:** `CancelSubscriptionBanner.test.tsx` - 4 new test cases (fallback view render, copy template, retry mailto, close/reset) plus updated existing tests to use `data-testid` selectors
- **Manual verification:** Tested with and without a default mail client; fallback view appears in both cases after confirming
- **Edge cases covered:** Long org names (URI length guard), copy feedback timer cleanup on unmount, retry fires a second mailto click

---

- **Blast radius:** `CancelSubscriptionBanner` component only - no shared state, no API calls changed
- **Potential regressions:** None; the dialog open/close flow is unchanged for non-cancellation paths
- **Rollback plan:** Revert this commit; the old silent-close behavior is restored

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | Cloud |
| Change Type | Bug Fix |
| Description | Cancel subscription dialog now shows a fallback with copy/retry options when the mail client doesn't open |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- The `openMailto` helper (append → click → remove) is the standard cross-browser workaround - it mirrors what file-download utilities do for blob URLs.
- `DialogView = 'confirm' | 'fallback' | null` replaces the old boolean `open` state. The dialog is open whenever `dialogView !== null`.
- `MAX_MAILTO_URI_LENGTH = 1800` is conservative (most browsers cap at 2083); chosen to leave headroom for URL encoding expansion.


frontend/src/container/BillingContainer/CancelSubscriptionBanner.test.tsx
```diff
@@ -84,38 +108,85 @@ describe('CancelSubscriptionBanner', () => {
...
 			screen.getByRole('button', { name: /cancel subscription/i }),
 		);
+		await user.type(screen.getByTestId('cancel-confirm-input'), 'cancel');
+		await user.click(screen.getByTestId('cancel-subscription-confirm-btn'));
 
-		const input = screen.getByPlaceholderText(/Enter the word cancel/i);
-		await user.type(input, 'cancel');
+		expect(appendSpy).toHaveBeenCalled();
+		expect(mockClick).toHaveBeenCalledTimes(1);
+		expect(removeSpy).toHaveBeenCalled();
```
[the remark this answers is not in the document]
SagarRajput-7: done, Both assertions are now immune to false positives from React's own appendChild/removeChild calls.

frontend/src/container/BillingContainer/CancelSubscriptionBanner.tsx
```diff
@@ -1,71 +1,143 @@
...
+}
+
+function openMailto(uri: string): void {
+	const link = document.createElement('a');
+	link.href = uri;
+	link.style.display = 'none';
+	document.body.appendChild(link);
+	link.click();
+	document.body.removeChild(link);
+}
```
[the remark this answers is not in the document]
SagarRajput-7: click() on an anchor cannot throw and the residual risk is zero

frontend/src/container/BillingContainer/CancelSubscriptionBanner.test.tsx
```diff
@@ -9,7 +9,35 @@ jest.mock('utils/basePath', () => ({
...
+	// Create a real anchor so JSDOM's appendChild/removeChild accept it.
+	// Override its click() so no navigation occurs.
+	jest.spyOn(document, 'createElement').mockImplementation((tag: string) => {
+		if (tag === 'a') {
+			const anchor = realCreateElement('a') as HTMLAnchorElement;
+			anchor.click = mockClick;
+			return anchor;
+		}
+		return realCreateElement(tag);
+	});
```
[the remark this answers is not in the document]
SagarRajput-7: NIT but done it

frontend/src/container/BillingContainer/CancelSubscriptionBanner.tsx
```diff
@@ -111,27 +190,59 @@ function CancelSubscriptionBanner(): JSX.Element {
...
+							</Button>
+							<Button
+								variant="outlined"
+								color="secondary"
+								prefix={<MailOpen size={14} />}
+								onClick={handleRetryMailto}
+								data-testid="retry-mailto-btn"
+							>
+								Reopen email client
+							</Button>
```
H4ad, line 249: THis could be the `a` tag with the link with email, so its unlikely to fail on retry.

SagarRajput-7: What difference/benefit are we targeting here with <a>?

With the DOM-attach fix already providing the same reliability as a user-clicked <a> tag.

So essentially they both opens the same path

--------

https://github.com/user-attachments/assets/010c9aec-62f3-4631-b53c-da5170305c4c

H4ad: If this failed once for whatever the reason, why it would work on retry using the same method? Since we will expect the user interaction, we can just use `a` tag instead.

SagarRajput-7: This is intended to be a retry button that exists for transient failures (focus/gesture chain interruption during dialog transition, browser quirk, timing), as earlier we were just closing the confirmation dialog directly after the "Cancel Subscription" click.

I am mainly struck by the thought - "failure cases where <a> succeeds but openMailto() fails for the retry"

@H4ad, I am open to making the change, but can you help with this thought above?

H4ad: I'm thinking more on, if everything goes wrong, the worst case the user can just click to copy the link and paste on as new URL, avoiding any JS to handle the behavior we have natively by the browser (link + mailto)

makeavish (approved):

frontend/src/container/BillingContainer/CancelSubscriptionBanner.tsx
SagarRajput-7: as discussed, using <a> tag principal for the retry, since it give us the chance to get success if previous method failed for some reason.

H4ad (approved):

PR #11636, opened 2026-06-10
refactor: sso config to support type and sso enabled in openapi spec

karanbalani:
### 📄 Summary
- reshaped the auth domain config structure
  - `ssoEnabled` and `roleMapping` are not root fields
  - provider config collapses into `provider: { type, config }`, where `type` is the discriminator and `config` is the actual provider specific payload
- updates to all the config read sites
- updates to OpenAPI specs
- `092` migration to rewrite the existing `auth_domain.data` rows from the old shape to new shape (idempotent as it skips the already migrated rows)

#### Issues closed by this PR
Closes: https://github.com/SigNoz/platform-pod/issues/2268

---

### ✅ Change Type
- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only


karanbalani:
Part of: https://github.com/SigNoz/platform-pod/issues/2268

PR #11219, opened 2026-05-07
feat: v2 list, delete, pin, unpin dashboards api

namanv3:
## Pull Request

---

### 📄 Summary

V2 List, Delete, Pin, Unpin Dashboard APIs.

#### Issues closed by this PR

Closes https://github.com/SigNoz/signoz/issues/8212
Closes https://github.com/SigNoz/signoz/issues/6943
Closes https://github.com/SigNoz/signoz/issues/8210
Closes https://github.com/SigNoz/signoz/issues/6939

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Integration tests added
- Manual verification: Manual tested all APIs

---

- Blast radius: New APIs, New table for pinned dashboards, no other flow impacted
- Potential regressions: None
- Rollback plan: Revert PR, Delete new table

---

PR #11643, opened 2026-06-10
fix: proper definition of user dashboard preferences

### LINKED ISSUE #8210: Enable pinning / favoriting dashboards
Users should be able to pin / favorite dashboards in dashboards listing page. 

- Pinned dashboards are moved to the pinned section similar to all dashboards section (check screenshot)

Use the preference framework. *

Image

namanv3:
## Pull Request

---

### 📄 Summary

Fixes needed as per comments on https://github.com/SigNoz/signoz/pull/11219
Closes https://github.com/SigNoz/signoz/issues/8210

### DIFF
--- a/pkg/modules/dashboard/impldashboard/store.go
+++ b/pkg/modules/dashboard/impldashboard/store.go
@@ -383,15 +383,16 @@ func (store *store) RunInTx(ctx context.Context, cb func(ctx context.Context) er
 // rows = 0 is the only signal of a real limit hit.
 func (store *store) PinForUser(ctx context.Context, preference *dashboardtypes.UserDashboardPreference) error {
 	res, err := store.sqlstore.BunDBCtx(ctx).NewRaw(`
-		INSERT INTO user_dashboard_preference (user_id, dashboard_id, is_pinned)
-		SELECT ?, ?, true
+		INSERT INTO user_dashboard_preference (id, user_id, dashboard_id, is_pinned, created_at, updated_at)
+		SELECT ?, ?, ?, true, ?, ?
 		WHERE (SELECT COUNT(*) FROM user_dashboard_preference WHERE user_id = ? AND is_pinned = true) < ?
 		   OR EXISTS (SELECT 1 FROM user_dashboard_preference WHERE user_id = ? AND dashboard_id = ? AND is_pinned = true)
-		ON CONFLICT (user_id, dashboard_id) DO UPDATE SET is_pinned = true
+		ON CONFLICT (user_id, dashboard_id) DO UPDATE SET is_pinned = true, updated_at = ?
 	`,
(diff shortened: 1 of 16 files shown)


namanv3:
Not picked up one suggestion as per this response: https://github.com/SigNoz/signoz/pull/11219#discussion_r3391180744

pkg/modules/dashboard/impldashboard/store.go
```diff
@@ -410,36 +411,61 @@ func (store *store) PinForUser(ctx context.Context, preference *dashboardtypes.U
...
 		return errors.WrapInternalf(err, errors.CodeInternal, "couldn't delete dashboard preferences")
 	}
 	return nil
 }
 
-func (store *store) DeletePreferencesForUser(ctx context.Context, userID valuer.UUID) error {
+func (store *store) DeletePreferencesForUser(ctx context.Context, orgID valuer.UUID, userID valuer.UUID) error {
+	// No org_id on the preference table, so scope by org via a subquery on the
+	// parent (DELETE-with-JOIN isn't portable across Postgres/SQLite).
+	userIDsInOrgSubQuery := store.sqlstore.BunDBCtx(ctx).
```
namanv3, line 459: feels icky to build a (sub)query to a table that this store does not really handle. Not sure what else to do though.

therealpandey (approved):

pkg/modules/dashboard/impldashboard/store.go
```diff
@@ -410,36 +411,61 @@ func (store *store) PinForUser(ctx context.Context, preference *dashboardtypes.U
...
+		TableExpr("dashboard").
+		Column("id").
+		Where("org_id = ?", orgID)
+
 	_, err := store.sqlstore.BunDBCtx(ctx).
 		NewDelete().
 		Model((*dashboardtypes.UserDashboardPreference)(nil)).
 		Where("user_id = ?", userID).
 		Where("dashboard_id = ?", dashboardID).
+		Where("dashboard_id IN (?)", dashboardIDsInOrgSubQuery).
```
jatinderjit, line 428: minor nit: prefer `exists` over `in`. This is probably going to be a short (and non-nullable) list, so not a major issue here.

srikanthccv (approved):

pkg/modules/dashboard/impldashboard/store.go
namanv3: Hey, resolving for now as its a minor nit and because yes it is a non-nullable list. Will read up a bit on this to understand a bit more and include this in the next PR. Thanks though!

namanv3: In resulting in NULL in some cases instead of being pure true/false like EXISTS is something I will keep note of from now on

PR #11646, opened 2026-06-11
feat(dashboard-v2): runtime variable selection

AshwinBhatkal:
Stacked on #11645 (variables settings tab) — this PR adds the **runtime variable selection** experience: the selector bar above the panels, per-type value selection, and inter-variable dependency orchestration.

## Design

- Selected values live in a persisted zustand `variableValues` slice + the URL (`?variables=`), seeded **URL → localStorage → default**. They are **never** written to the dashboard spec.
- Inter-variable dependencies are derived from query references (`{{.name}}` / `$name` / `[[name]]`): a topological graph with cycle detection. Orchestration is **declarative** — a Query selector is `enabled` only once its parents resolve and its query key carries the parent values, so it refetches when a parent changes (cycles are starved by the gate).
- Value pickers reuse the shared `NewSelect` (`CustomSelect`/`CustomMultiSelect`) for search, the ALL option and apply-on-close batching.

## What's included

- [x] Foundation: dependency graph + selection value types + persisted store slice.
- [x] Selector bar: `VariablesBar` above the panels; store + URL sync with seeding.
- [x] All four selector types: Custom, Text, Query (`/variables/query`, parent values) and Dynamic (`/fields/values`).
- [x] Dependency orchestration: declarative parent-gating + parent values in the query key.
- [x] Sibling-dynamic option filtering: dynamic variables scope each other via an `existingQuery` (e.g. `pod` narrows to the chosen `namespace`).
- [x] Multi-select / ALL UX via `NewSelect` (apply-on-close batching, no per-toggle cascade).

## Notes

- Dark-shipped/parked like the rest of V2 (renders once #11642 un-parks). Since V2 panels don't render queries yet, selected values populate the store/URL but don't yet drive panels.

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableModel.ts
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableModel.ts
@@ -1,3 +1,4 @@
+import { sortBy } from 'lodash-es';
 import type { VariableDefaultValueDTO } from 'api/generated/services/sigNoz.schemas';
 
 /**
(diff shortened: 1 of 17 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableModel.ts
```diff
@@ -76,3 +77,26 @@ export function emptyVariableFormModel(): VariableFormModel {
 		dynamicSignal: 'traces',
 	};
 }
+
+/** Maps the dynamic-variable signal to the field-values API signal. */
+export function signalForApi(
```
ahrefabhi, line 82: what is the use of this function, this doesn't do anything 😕

AshwinBhatkal: Follow up PR - https://github.com/SigNoz/signoz/pull/11775

ahrefabhi (approved):

PR #11648, opened 2026-06-11
feat(traces): integrate flamegraph v3 API

aks07:
## Pull Request

---

The backend shipped a memory-optimised **v3 flamegraph API** with a changed contract. This PR migrates the **TraceDetailsV3** flamegraph page to it. The legacy trace-details page (`PaginatedTraceFlamegraph`) is untouched and continues to use the v2 endpoint.

Contract changes adopted (v2 → v3):
- **Endpoint**: `POST /traces/flamegraph/:traceId` → `POST /api/v3/traces/:traceID/flamegraph` (same shape as the new waterfall API).
- **Envelope**: response is now nested under `data` (`{ "status": "...", "data": {...} }`).
- **`limit`** request param removed.
- **`serviceName`** no longer sent at the span root — colouring is derived from the selected fields / `resource["service.name"]`.
- **Parent linkage**: a single `parentSpanId` field instead of `references[]`.

Approach: consume the generated OpenAPI client directly. The v3 endpoint is already generated (`getFlamegraph` + `Spantypes…FlamegraphDTO` schemas), so a thin `useQuery` hook wraps the generated util and returns the typed `response.data` as-is — no `SuccessResponse` re-wrap. A custom hook is needed only because the endpoint is a POST (orval generates a mutation for POSTs).

#### Screenshots / Screen Recordings [No new changes UI wise]

https://github.com/user-attachments/assets/febdda2e-68b3-4a49-9415-28c9ba3df454

#### Issues closed by this PR
Closes: https://github.com/SigNoz/engineering-pod/issues/5221

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
After pointing the page at the v3 API, the flamegraph fetched data and computed a valid layout but **rendered no bars** (no console error). v3 returns `span.timestamp` in **nanoseconds**, whereas the render pipeline — and the shared v2 page — treat `timestamp` as **milliseconds** (the same scale as `startTimestampMillis`). Bars were therefore positioned ~1e6× too far right, off-screen.

#### Fix Strategy
Normalise `span.timestamp` ns → ms **once** at the data boundary (the v3 hook's `queryFn`), so the existing render pipeline works unchanged. `durationNano` stays in ns — it's already divided by `1e6` at every use site, matching the pre-existing mixed-unit convention.

---

- **Tests added/updated**: `TraceFlamegraph.test.tsx` (mocks the v3 hook + un-wrapped `data.spans` shape), `computeVisualLayout.test.ts` / `testUtils.ts` (fixtures match the generated DTO — dropped `serviceName`/`traceId`, added required `resource`/`attributes`), `utils.presentation.test.ts` (replaced the obsolete "falls back to top-level serviceName" case with "returns unknown when resource empty").
- **Manual verification**: flamegraph renders for normal and sampled traces; hover tooltips (service name, events), connector lines, and click-to-select verified.
- **Edge cases covered**: empty `spans`; spans without `service.name` in resource; optional `event` fields (guarded with defaults).
- tsgo `--noEmit` clean · oxlint clean · jest 87/87 in the flamegraph suite · production build passes.

---

- **Blast radius**: TraceDetailsV3 flamegraph only. The v2 API util, hook, types, and the legacy `PaginatedTraceFlamegraph` page are untouched.
- **Potential regressions**: span positioning / colouring in the v3 flamegraph if the v3 contract differs further from assumptions (e.g. `startTimestampMillis` scale).
- **Rollback plan**: revert the PR — the v2 path is independent and unaffected.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Description | Trace flamegraph now uses the memory-optimised v3 API. |

---

### 📋 Checklist
- [x] Manually tested
- [x] Breaking changes documented (none — additive; v2 path untouched)

---

## 👀 Notes for Reviewers

- The span type on the v3 path now aliases the generated `SpantypesFlamegraphSpanDTO` per-file (`SpantypesFlamegraphSpanDTO as FlamegraphSpan`) — chosen over a central re-export to avoid an import cycle between `types.ts` and `computeVisualLayout.ts`.
- `hasMore` (new in the v3 response) is intentionally **not** wired up — it was needed by the old *paginated* trace-details view; the v3 flamegraph is single-shot. Left as a follow-up.
- Removing the legacy `PaginatedTraceFlamegraph` page is out of scope for this PR.


frontend/src/pages/TraceDetailsV3/TraceFlamegraph/computeVisualLayout.ts
```diff
@@ -159,24 +158,8 @@ export function computeVisualLayout(spans: FlamegraphSpan[][]): VisualLayout {
...
-			| Array<{ spanId?: string; refType?: string }>
-			| undefined;
-		if (refs) {
-			for (const ref of refs) {
-				if (ref.refType === 'CHILD_OF' && ref.spanId) {
-					return ref.spanId;
-				}
-			}
-		}
-		return '';
```
H4ad, line 179: Is this safe to remove?

frontend/src/pages/TraceDetailsV3/TraceFlamegraph/utils.ts
```diff
@@ -74,34 +74,25 @@ export function getFlamegraphRowMetrics(
...
 	field: TelemetryFieldKey,
 ): string {
-	const fromAttribute = getSpanAttribute(span, field.name);
-	if (fromAttribute) {
-		return fromAttribute;
-	}
-	if (field.name === 'service.name') {
-		return span.serviceName || 'unknown';
-	}
-	return 'unknown';
```
H4ad, line 104: Same for this one, is safe to remove?

frontend/src/pages/TraceDetailsV3/TraceFlamegraph/TraceFlamegraph.tsx
```diff
@@ -70,17 +70,13 @@ function TraceFlamegraph({
 		data,
 		isFetching,
 		error: fetchError,
-	} = useGetTraceFlamegraph({
```
H4ad, line 73: The old hook still being used? Let's add `@deprecated` to avoid using it and prefer v3.

frontend/src/pages/TraceDetailsV3/TraceFlamegraph/utils.ts
H4ad: From the PR description, looks like it's safe: no longer sent at the span root — colouring is derived from the selected fields / resource["service.name"].

frontend/src/pages/TraceDetailsV3/TraceFlamegraph/computeVisualLayout.ts
H4ad: From the PR description, looks like it's safe: a single parentSpanId field instead of references[].

frontend/src/pages/TraceDetailsV3/TraceFlamegraph/TraceFlamegraph.tsx
aks07: The old hook is kept for supporting old trace details. Will be removed once we phase it out.

frontend/src/pages/TraceDetailsV3/TraceFlamegraph/utils.ts
aks07: Yes this is not to be used

frontend/src/pages/TraceDetailsV3/TraceFlamegraph/computeVisualLayout.ts
aks07: references wont be sent from backend now.

frontend/src/pages/TraceDetailsV3/TraceFlamegraph/TraceFlamegraph.tsx
H4ad: Let's do a follow-up to mark everything as deprecated, no need to do in this PR

H4ad (dismissed):

frontend/src/pages/TraceDetailsV3/TraceFlamegraph/TraceFlamegraph.tsx
H4ad: After discussion, we don't need to mark as deprecated now, we can wait when we don't maintain those pages/hooks anymore and then mark as deprecated. Today, these hooks/components are still being used and maintained.

tewarig (dismissed):

H4ad (dismissed):

H4ad (approved):

therealpandey (approved):

PR #11680, opened 2026-06-11
fix(infra-monitoring-k8s-pods): working set memory should use space aggregation sum

H4ad:
## Pull Request

---

Just a one-line fix for an inconsistency found by Nikhil. 

I didn't see any perceptive change in the chart but this can cause subtle differences in the chart in case we have duplicated data being ingested. This could lead the memory of the pod to appear higher since the memory will `sum` instead of `avg`.

Before / After:

image

Fixes https://github.com/SigNoz/engineering-pod/issues/5331

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

One of the queries uses `avg` and the other uses `sum`.

Likely something that was forgot to be updated.

We are prefering to use `sum` instead of `avg`, so we just need to update the code.

---

- Tests added/updated: No
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Infrastructure Monitoring - Pods - Working Memory Chart
- Potential regressions: Higher values when data is being duplicated during ingestion, but this is something expected.
- Rollback plan: None

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Ensure the chart of "Memory by State" on Pods inside Infrastructure Monitoring is consistent on space aggregation, previously the Working Set Memory was usingg `avg`, we changed to use `sum`. |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/container/InfraMonitoringK8s/Pods/constants.ts
+++ b/frontend/src/container/InfraMonitoringK8s/Pods/constants.ts
@@ -1366,7 +1366,7 @@ export const getPodMetricsQueryPayload = (
 							orderBy: [],
 							queryName: 'B',
 							reduceTo: ReduceOperators.AVG,
-							spaceAggregation: 'avg',
+							spaceAggregation: 'sum',
 							stepInterval: 60,
 							timeAggregation: 'avg',
 						},


ahrefabhi (approved):

nikhilmantri0902 (approved):
LGTM

PR #11400, opened 2026-05-21
refactor(frontend): migrate plain antd dropdown to @signozhq/ui/dropdown

tewarig:
### 📄 Summary
Migrates `Dropdown` usages from **antd** to the `@signozhq/ui/dropdown-menu` component (Radix-based) across ~15 production call-sites. Part of the broader effort to remove antd in favor of the SigNoz UI design system. The API differs (`placement="bottomRight"` → `align="end"`, `overlayClassName` → `className`, `MenuProps['items']` → `MenuItem[]`, `dropdownRender` → compositional `<DropdownMenuContent>` children) so call-sites are updated mechanically.

Two adoption patterns are used:
- **`DropdownMenuSimple`** (antd-like declarative `menu={{items}}` API) — used for the common case where the trigger is a button and items are a flat list.
- **Compositional primitives** (`DropdownMenu` + `Trigger` + `Content` + `Item`) — used when the trigger needs to forward refs, when the open state is controlled, or when the menu renders custom content that doesn't fit a flat item list (e.g. `SideNav`, `CustomDomainSettings`, `BodyTitleRenderer`, `ExplorerColumnsRenderer`).

One special case: `VariablesDropdown` is renamed to `VariablesPopover` and now uses `@signozhq/ui/popover` instead of `dropdown-menu`. `DropdownMenuTrigger` preventDefaults `pointerdown`, which breaks the focus-driven input inside that component; `PopoverAnchor` is a passive positioning element and behaves correctly for that flow.

ideally we should have our own autocomplete component for this. 

Also adds **`Dropdown`** to the `signoz/no-antd-components` ban-list (`frontend/plugins/rules/no-antd-components.mjs`) so future regressions surface as lint errors.

#### Screenshots / Screen Recordings (if applicable)

walkthough video - 

https://drive.google.com/file/d/1PTqUxTTq6AlLbcrKIRl1SfOdFBtKVpLu/view?usp=sharing 

#### Issues closed by this PR
N/A

---

### ✅ Change Type
- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context
N/A — not a bug fix.

---

### 🧪 Testing Strategy

- **Tests added/updated:**
[...]

PR #11684, opened 2026-06-11
fix(settings): trial banner causing the scroll of the sub-pages to be broken

SagarRajput-7:
## Summary

Three UI regressions fixed: 

Closes: https://github.com/SigNoz/platform-pod/issues/2501

### 1. **Billing page bottom content clipped on free trial**: 

Settings layout used hardcoded `max-height: 100vh` and `height: calc(100vh - 48px)` which didn't account for the 48px trial banner. Replaced with a flex-based layout (`flex: 1; min-height: 0`) so heights derive from the actual container. Also fixed a dead `margin-bottom` on `BillingContainer` that was silently overridden by a subsequent `margin` shorthand.

#### Before (video) - 

https://github.com/user-attachments/assets/7caee22f-9781-4373-87cc-710d6d507cc6

------

#### After (video) - 

https://github.com/user-attachments/assets/47b7f491-a66f-4e05-96e4-c9d1c3ff0891

-------

### 2. **Workspace callout overflowing its container**:

The `@signozhq/ui` Callout component sets `width: 100%` on its root. Adding horizontal `margin` to the same element caused total width to exceed the parent. Fixed with `width: auto !important` to override the design system default.

| Before | After |
|---|---|
| image | image |

---------

### 3. **Sign-out text lost red color**:

PR #11400 migrated the user settings dropdown from `antd` to `@signozhq/ui/dropdown-menu`. The new component doesn't use `ant-dropdown-menu-item` or `ant-dropdown-menu-title-content` class names, so the `.user-settings-dropdown-logout-section` color rule (nested under those dead selectors) never matched. Lifted the selector out of the stale Ant Design nesting.

| Before | After |
|---|---|
| image | image |

---

### ✅ Change Type

- [x] 🐛 Bug fix

---

### 🧪 Testing Strategy

- Manual verification on free trial account (billing page, workspace page, sign-out menu)

---

- **Blast radius:** Settings layout, workspace callout, sign-out menu item
- **Potential regressions:** None - flex layout change only affects the Settings shell; callout fix is scoped to the license key callout; sign-out fix removes dead selectors
- **Rollback plan:** Revert commit

### DIFF
--- a/frontend/src/pages/Settings/Settings.styles.scss
+++ b/frontend/src/pages/Settings/Settings.styles.scss
@@ -1,8 +1,12 @@
 .settings-page {
-	max-height: 100vh;
+	flex: 1;
+	min-height: 0;
+	display: flex;
+	flex-direction: column;
 	overflow: hidden;
 
 	.settings-page-header {
+		flex-shrink: 0;
 		border-bottom: 1px solid var(--l1-border);
 		background: var(--l1-background);
 		backdrop-filter: blur(20px);
@@ -24,13 +28,14 @@
 	}
 
 	.settings-page-content-container {
+		flex: 1;
+		min-height: 0;
 		display: flex;
 		flex-direction: row;
-		align-items: flex-start;
+		align-items: stretch;
 
 		.settings-page-sidenav {
 			width: 240px;
-			height: calc(100vh - 48px);
 			border-right: 1px solid var(--l1-border);
 			background: var(--l1-background);
 			padding-top: var(--padding-1);
@@ -74,7 +79,6 @@
 
 		.settings-page-content {
 			flex: 1;
-			height: calc(100vh - 48px);
 			background: var(--l1-background);
 			padding: 10px 8px;
 			overflow-y: auto;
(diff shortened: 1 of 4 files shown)


frontend/src/pages/Settings/Settings.styles.scss
H4ad, line 1: This was not enough to fix on my test:

Image

frontend/src/pages/Settings/Settings.styles.scss
```diff
@@ -24,13 +28,14 @@
...
+		flex: 1;
+		min-height: 0;
 		display: flex;
 		flex-direction: row;
-		align-items: flex-start;
+		align-items: stretch;
 
 		.settings-page-sidenav {
 			width: 240px;
-			height: calc(100vh - 48px);
```
H4ad, line 33: Another approach could be introducing a css var with the size, we have more places with this kind of hardcoded value

frontend/src/pages/Settings/Settings.styles.scss
H4ad: I was testing it in a wrong way, after mocking the response, it fixed the issue :)

H4ad (approved):
Follow-up: find another places that could have the same issue and address in the same way

PR #11693, opened 2026-06-12
fix(frontend): add dynamic page titles reflecting content in browser tabs

### LINKED ISSUE #11640: HTML Title tags should reflect the page content

Many of the Signoz title tags are the same and make it difficult to differentiate when looking at multiple tabs.

Specifically Service pages and Metrics pages, it would be very helpful to have the service name in the title tag and have a <h1> tag using service.name.

<title>Signoz Service {service.name}</title>
< h1 >Service {service.name}</ h1 >

None

#### *Thank you* for your feature request – we love each and every one!

sawirricardo:
## Summary

Many SigNoz page titles are identical, making it difficult to differentiate tabs when viewing multiple pages. This PR adds dynamic titles that include contextual information like service names, dashboard names, and alert names.

Closes #11640

### Core: route param extraction (`AppLayout/utils.ts`)
- Added `extractRouteParams()` — extracts dynamic URL params from route templates
- Added `getRouteMatch()` — like existing `getRouteKey()` but returns both the route key and extracted params

### Dynamic titles for service pages (`AppLayout/index.tsx`)
- Service Metrics: `SigNoz | frontend - Service Metrics` (was `SigNoz | Service Metrics`)
- Service Operations: `SigNoz | frontend - Operations` (was `SigNoz | Service Operations`)

### Consolidated title management via Helmet
Replaced direct `document.title` mutations with `<Helmet>` from `react-helmet-async` in:
- **DashboardPage** — now shows `SigNoz | Dashboard Name`
- **DashboardPageV2** — now shows `SigNoz | Dashboard Name`
- **AlertDetails** — now shows `SigNoz | Alert Rule Name`

This ensures all title updates go through the same Helmet provider and follow a consistent `SigNoz | <context>` format.

## Screenshots

| Before | After |
|--------|-------|
| All service tabs show "SigNoz \| Service Metrics" | Each tab shows "SigNoz \| frontend - Service Metrics" |
| Dashboard tab shows only dashboard name (no SigNoz prefix) | Shows "SigNoz \| Dashboard Name" |

## Testing

- TypeScript compilation passes (`tsc --noEmit` — no errors)
- Verified route matching logic covers both static and dynamic routes
- Tested that `extractRouteParams` correctly decodes URL-encoded service names

## Notes

The `getRouteKey` function is preserved for backward compatibility. `getRouteMatch` is a superset that also returns extracted params. Future routes with dynamic segments (e.g. funnel detail, integration detail) can be extended in the same `pageTitle` useMemo block in `AppLayout`.


gman-wa:
Can the proposed title names be switched to have 'Signoz' last.

i.e.

{proposed change} | SigNoz

the page and specific service/trace/metric/dashboard should take priority over SigNoz branding.

H4ad:
Hey, thanks for the contribution!

I'm just checking internally with team if we have more pages or any other feedback about how it should be structured, but this is a really good addition.

I will do a proper review in the following days once we got the product usability defined.

H4ad (commented):
Hey, I'm back to this, we decided to go with `<name> | SigNoz` since we already have the icon to distinct the signoz app.

Also, I want to completely revamp this, instead of using the old code, I want to remove everything and start from scratch.

I'm planning to introduce a hook, like `usePageTitle` that we can customize the page title by the component

The idea is to stack the titles to be able to have multiple titles depending on what you open in the app, eg:

- Open Infrastructure Monitoring -> `Hosts | Infrastructure Monitoring | SigNoz`
- Change to K8s, pods is the first selected -> `Pods | Infrastructure Monitoring | SigNoz`
- Select a pod -> `pod-name | Pods | Infrastructure Monitoring | SigNoz`

Each time the `usePageTitle(value)` is called, we can stack the title and remove when unmount.

In the end, the first value will be on root with `usePageTitle('SigNoz')`.

And then we will need to update every part of the app to use this new hook

Let me know if you want to work in this implementation, otherwise I will close this PR for now.

sawirricardo:
Reworked from scratch per the feedback, the old Helmet/`document.title` code is removed and replaced with a stackable `usePageTitle` hook (zustand-backed)

How it works:
- Each `usePageTitle(segment)` call pushes a segment while the component is mounted, updates in place on change, and pops on unmount. The title renders most-specific-first, SigNoz last: `My Dashboard | All Dashboards | SigNoz`
- `AppLayout` registers the root `SigNoz` segment plus the route-level title from `titles.json` (prefixes stripped there so segments stack cleanly, this also addresses the earlier "SigNoz last" feedback)
- Passing `undefined` contributes nothing, so loading states just keep the parent title

AlertDetails, DashboardPage, DashboardPageV2 (the pages that previously set `document.title` by hand), plus a unit test for the hook and updated e2e title assertions

deeper per-page segments like the `pod-name | Pods | Infrastructure Monitoring | SigNoz` example, every page currently gets its route-level segment from `titles.json`, and richer nesting can be added incrementally by dropping `usePageTitle(...)` into any component. Happy to do that migration here or in follow-ups, whichever you prefer @H4ad

PR #11698, opened 2026-06-12
feat(statsreporter): expose collected stats via GET /api/v1/stats

therealpandey:
## What

Adds `GET /api/v1/stats`, exposing the per-org stats bag that was previously only pushed to analytics on a 6-hour timer.

- **Always-on collection.** Stats collection is extracted out of the analytics reporter into a shared `Aggregator` used by both the reporter and a new HTTP handler, so the endpoint works regardless of whether scheduled reporting is enabled (`statsreporter.enabled`). The endpoint requires viewer access.
- **Telemetry reads moved to the querier.** The trace/log/metric row-count and last-observed ClickHouse queries move out of `statsreporter` into the querier, which now implements `statsreporter.StatsCollector` and is wired in as one of the collectors. The aggregator becomes a pure collector fan-out and no longer depends on `telemetrystore` - - addressing the standing _"move this to the querier layer"_ TODO.

closes SigNoz/platform-pod#2489

### DIFF
--- a/frontend/src/api/generated/services/sigNoz.schemas.ts
+++ b/frontend/src/api/generated/services/sigNoz.schemas.ts
@@ -9746,6 +9746,19 @@ export type UpdateSpanMapperPathParameters = {
 	groupId: string;
 	mapperId: string;
 };
+export type GetStats200Data = { [key: string]: unknown };
+
+export type GetStats200 = {
+	/**
+	 * @type object
+	 */
+	data: GetStats200Data;
+	/**
+	 * @type string
+	 */
+	status: string;
+};
+
 export type GetTraceAggregationsPathParameters = {
 	traceID: string;
 };
--- a/pkg/apiserver/signozapiserver/statsreporter.go
+++ b/pkg/apiserver/signozapiserver/statsreporter.go
@@ -1,0 +1,33 @@
+package signozapiserver
+
+import (
+	"net/http"
+
+	"github.com/SigNoz/signoz/pkg/http/handler"
+	"github.com/SigNoz/signoz/pkg/types"
+	"github.com/gorilla/mux"
+)
+
+func (provider *provider) addStatsReporterRoutes(router *mux.Router) error {
+	if err := router.Handle("/api/v1/stats", handler.New(
+		provider.authzMiddleware.ViewAccess(provider.statsHandler.Get),
+		handler.OpenAPIDef{
+			ID:                  "GetStats",
+			Tags:                []string{"stats"},
+			Summary:             "Get stats",
+			Description:         "This endpoint returns the collected stats for the organization",
+			Request:             nil,
+			RequestContentType:  "",
+			Response:            map[string]any{},
+			ResponseContentType: "application/json",
+			SuccessStatusCode:   http.StatusOK,
+			ErrorStatusCodes:    []int{},
+			Deprecated:          false,
+			SecuritySchemes:     newSecuritySchemes(types.RoleViewer),
+		},
+	)).Methods(http.MethodGet).GetError(); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/flagger/handler.go
+++ b/pkg/flagger/handler.go
@@ -35,11 +35,7 @@ func (handler *handler) GetFeatures(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	orgID, err := valuer.NewUUID(claims.OrgID)
-	if err != nil {
-		render.Error(rw, err)
-		return
-	}
+	orgID := valuer.MustNewUUID(claims.OrgID)
 
 	evalCtx := featuretypes.NewFlaggerEvaluationContext(orgID)
 
--- a/pkg/querier/collect.go
+++ b/pkg/querier/collect.go
@@ -12,10 +12,6 @@ import (
 	"github.com/SigNoz/signoz/pkg/valuer"
 )
 
-// Collect implements statsreporter.StatsCollector. It returns telemetry counts and last-observed
-// times for traces, logs and metrics. orgID is unused because the underlying tables are
-// cluster-global. Collection is best-effort: a failing query is logged and its key omitted rather
-// than failing the whole collection.
 func (q *querier) Collect(ctx context.Context, _ valuer.UUID) (map[string]any, error) {
 	stats := make(map[string]any)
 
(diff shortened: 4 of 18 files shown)


frontend/src/api/generated/services/sigNoz.schemas.ts
```diff
@@ -9746,6 +9746,19 @@ export type UpdateSpanMapperPathParameters = {
 	groupId: string;
 	mapperId: string;
 };
+export type GetStats200Data = { [key: string]: unknown };
```
therealpandey, line 9749: @AshwinBhatkal please excuse this. I have a follow up PR to address this more strongly and will have it implemented before that.

pkg/statsreporter/analyticsstatsreporter/provider.go
```diff
@@ -32,11 +28,8 @@ type provider struct {
 	// config
 	config statsreporter.Config
 
-	// used to get telemetry details. srikanthcvv to move this to the querier layer
-	telemetryStore telemetrystore.TelemetryStore
```
therealpandey, line 36: @srikanthccv i addressed this also.

pkg/flagger/handler.go
```diff
@@ -35,11 +35,7 @@ func (handler *handler) GetFeatures(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	orgID, err := valuer.NewUUID(claims.OrgID)
-	if err != nil {
-		render.Error(rw, err)
-		return
-	}
+	orgID := valuer.MustNewUUID(claims.OrgID)
```
therealpandey, line 38: Wrong precedent fix.

pkg/apiserver/signozapiserver/statsreporter.go
```diff
@@ -0,0 +1,33 @@
...
+	if err := router.Handle("/api/v1/stats", handler.New(
+		provider.authzMiddleware.ViewAccess(provider.statsHandler.Get),
+		handler.OpenAPIDef{
+			ID:                  "GetStats",
+			Tags:                []string{"stats"},
+			Summary:             "Get stats",
+			Description:         "This endpoint returns the collected stats for the organization",
+			Request:             nil,
+			RequestContentType:  "",
+			Response:            map[string]any{},
```
namanv3, line 21: related to this thread right?
https://signoz-team.slack.com/archives/C088SMWD6E8/p1781260536853249

namanv3 (commented):
got a rough idea about what's going on, that much looks good. Will need another pass and or so to understand this area of code more.

pkg/apiserver/signozapiserver/statsreporter.go
therealpandey: Yes, will think of strongly typing these.

These are dynamically created in some places so still thinking of a better plan.

pkg/querier/collect.go
```diff
@@ -0,0 +1,74 @@
...
+
+	var metrics uint64
+	if err := q.telemetryStore.ClickhouseDB().QueryRow(ctx, fmt.Sprintf("SELECT COUNT(*) FROM %s", metricsTable)).Scan(&metrics); err == nil {
+		stats["telemetry.metrics.count"] = metrics
+	} else {
+		q.logger.DebugContext(ctx, "failed to collect metrics count", errors.Attr(err))
+	}
+
+	var tracesLastSeenAt time.Time
+	if err := q.telemetryStore.ClickhouseDB().QueryRow(ctx, fmt.Sprintf("SELECT max(timestamp) FROM %s", tracesTable)).Scan(&tracesLastSeenAt); err == nil {
```
srikanthccv: nit: these queries are already cheap. we could combine the count and max timestamp into one query.

srikanthccv (dismissed):
LGTM

pkg/querier/collect.go
therealpandey: I can do that, hold on.

therealpandey: Done, combined into 1

srikanthccv (approved):

pkg/statsreporter/aggregator.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/factory"
+	"github.com/SigNoz/signoz/pkg/types/ctxtypes"
+	"github.com/SigNoz/signoz/pkg/types/instrumentationtypes"
+	"github.com/SigNoz/signoz/pkg/valuer"
+)
+
+// Aggregator aggregates stats from every registered StatsCollector for a single organization.
+type Aggregator interface {
```
vikrantgupta25, line 15: interface and implementation at the same level

pkg/statsreporter/handler.go
```diff
@@ -0,0 +1,46 @@
...
+	"context"
+	"net/http"
+	"time"
+
+	"github.com/SigNoz/signoz/pkg/http/render"
+	"github.com/SigNoz/signoz/pkg/types/authtypes"
+	"github.com/SigNoz/signoz/pkg/valuer"
+)
+
+type Handler interface {
```
vikrantgupta25, line 13: interfaces in the `statsreporter.go` and have `signozstatsapi` implementing the handler and use DI for `signozstatsagregagtor` 

same for aggregator interface

pkg/apiserver/signozapiserver/statsreporter.go
```diff
@@ -0,0 +1,33 @@
...
+	"net/http"
+
+	"github.com/SigNoz/signoz/pkg/http/handler"
+	"github.com/SigNoz/signoz/pkg/types"
+	"github.com/gorilla/mux"
+)
+
+func (provider *provider) addStatsReporterRoutes(router *mux.Router) error {
+	if err := router.Handle("/api/v1/stats", handler.New(
+		provider.authzMiddleware.ViewAccess(provider.statsHandler.Get),
```
vikrantgupta25, line 13: use the new resource middleware for this

vikrantgupta25 (changes requested):

pkg/statsreporter/aggregator.go
therealpandey: The only reason interface and implementation can't be at the same level if the interface implementation is importing other implementation or dependencies. Otherwise there is no reason it can't stay at the same level.

See `flagger`, `gateway`, `identn`.

pkg/statsreporter/handler.go
therealpandey: Same as above comment.

pkg/apiserver/signozapiserver/statsreporter.go
therealpandey: Let it go GA first.

vikrantgupta25 (approved):

PR #11700, opened 2026-06-12
Feat/qb dual mode base

ahrefabhi:
## Pull Request

---

### 📄 Summary

The query builder has exactly one way to persist its staged query: the `compositeQuery` URL param. Every "Stage & Run" serializes the whole query to the URL and the provider re-reads it from there. That's the right contract for the explorers (shareable, survives refresh, browser back/forward works), but the wrong one for embedded surfaces like the dashboard panel editor — where the URL fills with kilobytes of JSON that goes stale, breaks copy-paste, and gets mutated by unrelated components (e.g. `DateTimeSelectionV2` re-encodes `compositeQuery` and clears filters on every time change).

This PR introduces a **persistence strategy** so the builder can store its staged query either in the URL (`url` mode, the unchanged default) or in React state (`memory` mode), selected by mounting a nested `QueryBuilderProvider mode="memory"`.

**This PR is the infrastructure only — it is behavior-neutral.** Memory mode has no consumer yet; the dashboard panel editor adoption lands in a follow-up. Shipping the foundation separately keeps that diff small and lets this be reviewed on its own.

What it does:
- Extracts the wire-format logic out of the 1,200-line provider into pure, tested helpers (`lib/compositeQuery/`: migrate / serialize / normalize) — one source of truth.
- Adds a `CompositeQueryStore` interface with two implementations (`useUrlCompositeQueryStore`, `useMemoryCompositeQueryStore`).
- The provider now talks to a store and never touches the URL directly.
- Exposes `mode` and `committedQuery` on the query-builder context.
- Makes `useShareBuilderUrl` store-agnostic (seeds via `committedQuery`, so it no longer navigates in memory mode).

N/A — no user-facing change in this PR (infra only; existing `url`-mode behavior is unchanged).

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/5321

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature _(enables memory mode; dormant until a consumer adopts it)_
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor _(URL logic extracted from the provider into stores + pure utils)_
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- **Tests added:**
  - `lib/compositeQuery/__tests__/compositeQuery.test.ts` — legacy→new migration (filters/having/aggregations), serialize↔parse round-trip (incl. `+` handling), normalization defaults + fresh-id.
  - `compositeQueryStore/__tests__/useUrlCompositeQueryStore.test.tsx` — reads query/panelType from URL, commit resets pagination & drops `activeLogId`, `redirectingUrl`/`newTab`/stringified search params.
  - `compositeQueryStore/__tests__/useMemoryCompositeQueryStore.test.ts` — seed (with migration), commit + `onCommit`, ignores URL-only options.
  - `providers/__tests__/QueryBuilderMemoryMode.test.tsx` — real provider in memory mode: staging never navigates, `onStagedQueryChange` fires, `useShareBuilderUrl` seeds without clobbering `initialQuery`.
- **Manual verification:** `tsgo --noEmit` clean and the branch's test suites green with **zero** downstream (editor) changes present, confirming the infra stands alone. `pnpm build` succeeds.
- **Edge cases covered:** empty/garbage URL param → `null`; legacy-format queries seeded into memory mode get the same migration as URL parsing; invalid `queryType` → falls back to `QUERY_BUILDER`.

---

- **Blast radius:** every existing builder surface (explorers, alerts, live logs). All run `url` mode via the global provider, and `useUrlCompositeQueryStore` reproduces the old read/write/sync exactly. Net deletions in the provider (logic moved out).
- **Potential regressions:** the only consumer-visible API change is two new required context fields (`mode`, `committedQuery`); a handful of test mocks of `QueryBuilderContextType` were updated to match. An audit confirmed no component in `components/QueryBuilderV2/**` or `container/QueryBuilder/**` reads `compositeQuery`/`pagination`/`panelTypes` from the URL directly, so the builder UI is safe under either store.
- **Rollback plan:** revert the PR — no migrations, no persisted state, no API changes.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | N/A |
| Change Type | Maintenance |
| Description | N/A — internal, non-user-facing. `url` mode is byte-for-byte unchanged; `memory` mode is dormant until a follow-up adopts it. |

---

### 📋 Checklist
- [x] Manually tested
- [x] Breaking changes documented _(none; behavior-neutral)_
- [x] Backward compatibility considered _(default mode unchanged)_

---

## 👀 Notes for Reviewers

- **Why a new `committedQuery` field instead of reusing `stagedQuery`?** `stagedQuery` is set by a provider *effect*, and React runs child effects before parent effects — so a child like `useShareBuilderUrl` always sees `stagedQuery === null` on mount and would clobber a valid seed. `null` is also ambiguous ("not yet initialized" vs "genuinely empty"). `committedQuery` is the synchronous, unambiguous truth (URL-parse memo / memory state); `stagedQuery` stays the validated derivative.
- **Why nested providers instead of a flag / route-config?** It composes — modals, multiple builders on one page — and needs zero changes to existing pages; adoption is one JSX wrapper.
- **Two store hooks always run** (rules of hooks); only the selected one is used. The unused store is a few cheap memos.
- **Suggested review path:** `lib/compositeQuery/*` (pure, easy to verify) → `compositeQueryStore/types.ts` (the contract) → the two stores → the provider diff (mostly mechanical: URL calls replaced by `store.commit`).
- `useBeforeUnloadWarning` is a standalone reusable hook included here as infra; its first consumer arrives with the editor migration.

### DIFF
--- a/frontend/src/components/QueryBuilderV2/__tests__/QueryBuilder.PrevQuery.test.tsx
+++ b/frontend/src/components/QueryBuilderV2/__tests__/QueryBuilder.PrevQuery.test.tsx
@@ -106,6 +106,8 @@ describe('MetricsSelect - signal source switching (standalone)', () => {
 		};
 
 		(mockedUseQueryBuilder as any).mockReturnValue({
+			mode: 'url',
+			committedQuery: null,
 			currentQuery: currentQueryObj,
 			stagedQuery: null,
 			lastUsedQuery: null,
@@ -263,6 +265,8 @@ describe('DataSource change - Logs to Traces', () => {
 
 		mockedUseQueryBuilder.mockReset();
 		mockedUseQueryBuilder.mockReturnValue({
+			mode: 'url',
+			committedQuery: null,
 			currentQuery: logsCurrentQuery,
 			stagedQuery: null,
 			lastUsedQuery: null,
--- a/frontend/src/container/GridTableComponent/__tests__/utils.test.tsx
+++ b/frontend/src/container/GridTableComponent/__tests__/utils.test.tsx
@@ -2,6 +2,7 @@ import { Query } from 'types/api/queryBuilder/queryBuilderData';
 
 import {
 	createColumnsAndDataSource,
+	evaluateThresholdWithConvertedValue,
 	getQueryLegend,
 	sortFunction,
 } from '../utils';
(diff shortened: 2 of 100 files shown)


ahrefabhi:
Created new PR from main https://github.com/SigNoz/signoz/pull/11701

PR #11707, opened 2026-06-13
feat(planned-downtime): drop configuration tab & cascade downtime when deleting alert

jatinderjit:
### 📄 Summary

UX improvements for planned maintenance

1. Allow deletion of downtime without requiring users to first remove associated rules
2. Promote Planned Downtime and Routing Policies to top level tabs
3. Sort planned downtimes by Active > upcoming > expired
4. Show downtime status badge on the list page

image

#### Issues closed by this PR

- Closes https://github.com/SigNoz/engineering-pod/issues/5426
- Closes https://github.com/SigNoz/engineering-pod/issues/5427

---

### ✅ Change Type

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🧪 Testing Strategy

- Tests added/updated: Added for moving "Planned Downtime" and "Routing Policies" sub-tabs to top level
- Manual verification: Yes
- Edge cases covered:

---

- Blast radius: Planned maintenance
- Potential regressions:

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | UX improvements: easier access to planned downtime page, easier deletion of downtimes, and add badges for downtime status. |

---

### 📋 Checklist

- [x] Manually tested

### DIFF
--- a/pkg/alertmanager/alertmanagerstore/sqlalertmanagerstore/maintenance.go
+++ b/pkg/alertmanager/alertmanagerstore/sqlalertmanagerstore/maintenance.go
@@ -163,17 +163,29 @@ func (r *maintenance) CreatePlannedMaintenance(ctx context.Context, maintenance 
 }
 
 func (r *maintenance) DeletePlannedMaintenance(ctx context.Context, id valuer.UUID) error {
-	_, err := r.sqlstore.
-		BunDB().
-		NewDelete().
-		Model(new(alertmanagertypes.StorablePlannedMaintenance)).
-		Where("id = ?", id.StringValue()).
-		Exec(ctx)
-	if err != nil {
-		return r.sqlstore.WrapAlreadyExistsErrf(err, errors.CodeAlreadyExists, "cannot delete planned maintenance because it is referenced by associated rules, remove the rules from the planned maintenance first")
-	}
+	return r.sqlstore.RunInTxCtx(ctx, nil, func(ctx context.Context) error {
+		_, err := r.sqlstore.
+			BunDBCtx(ctx).
+			NewDelete().
+			Model(new(alertmanagertypes.StorablePlannedMaintenanceRule)).
+			Where("planned_maintenance_id = ?", id.StringValue()).
+			Exec(ctx)
+		if err != nil {
+			return err
+		}
 
-	return nil
+		_, err = r.sqlstore.
+			BunDBCtx(ctx).
+			NewDelete().
+			Model(new(alertmanagertypes.StorablePlannedMaintenance)).
+			Where("id = ?", id.StringValue()).
+			Exec(ctx)
+		if err != nil {
+			return err
+		}
+
+		return nil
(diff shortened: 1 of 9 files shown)


H4ad (commented):
You need to add support to old-links still work on new routes, so redirect from:

`/alerts?tab=Configuration&subTab=planned-downtime` to `/alerts?tab=PlannedDowntime`.

You can update the `Private.tsx` to add support to that, and also add tests for this.

pkg/alertmanager/alertmanagerstore/sqlalertmanagerstore/maintenance.go
```diff
@@ -163,17 +163,29 @@ func (r *maintenance) CreatePlannedMaintenance(ctx context.Context, maintenance
...
+			NewDelete().
+			Model(new(alertmanagertypes.StorablePlannedMaintenance)).
+			Where("id = ?", id.StringValue()).
+			Exec(ctx)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
```
H4ad, line 188: This is not breaking change? We should not create new endpoint or accept a flag to enable this behavior?

Also, split this to another PR, we need to update the UI of delete alert to alert about the deletion of the planned maintainance.

jatinderjit: Breaking change in the sense that it "fixes" the behavior. Added more context here: https://github.com/SigNoz/engineering-pod/issues/5426.
We don't need new endpoint or flag since we don't want the old behavior.

---

> we need to update the UI of delete alert to alert about the deletion of the planned maintainance

We already have that. 
image

jatinderjit:
> You need to add support to old-links still work on new routes, so redirect from:
> `/alerts?tab=Configuration&subTab=planned-downtime` to `/alerts?tab=PlannedDowntime`.
>
> You can update the Private.tsx to add support to that, and also add tests for this.

@H4ad Actually I did that earlier, but removed it later ([commit](https://github.com/SigNoz/signoz/pull/11707/commits/cbb53972b57928afbd4c47f8b499a60dcced5723)).
Reasoning: These aren't the type of pages that are commonly accessed, or users would have bookmarked, or sent as references in alert mails. I thought it's not worth the hassle to keep the extra code around. If we do need to support the old links for some reason, I can revert that commit.

Can't add to `Private.tsx` in its current form, since it accepts just the pathname. The pathname is just `/alerts`, which is shared among the 4 tabs. Will need to modify it to support search params. It'd complicate the current straightforward logic. We can do it, but that again doesn't feel worth the hassle for this specific use case.

H4ad (approved):
The frontend changes looks good, agreed with:
- no need to redirect old links
- the behavior around alerts

PR #11712, opened 2026-06-14
chore: add migration script from current to perses dashboard

namanv3:
## Pull Request

---

### 📄 Summary

Migration of a v1 dashboard to a v2 perses dashboard.
This is not wired into any migration script cuz we actually don't want to run it for dashboards. That will be done as a part of https://github.com/SigNoz/engineering-pod/issues/5502

#### Issues closed by this PR

Closes https://github.com/SigNoz/engineering-pod/issues/5203

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

- Tests added/updated: Unit tests added
- Manual verification: Created many dashboards via MCP and UI and migrated them using a part-time API that runs the migration for all existing dashboards

---

- Blast radius: V2 dashboards
- Potential regressions: none
- Rollback plan: revert PR

---


pkg/types/dashboardtypes/perses_v1_to_v2_queries.go
```diff
@@ -0,0 +1,249 @@
+package dashboardtypes
+
+import (
+	"encoding/json"
+
+	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/transition"
```
therealpandey:
```suggestion
```

This is an infra level import in types. Please evaluate and either move this to transition or move some required transition functions in types.

therealpandey (changes requested):

pkg/types/dashboardtypes/perses_v1_to_v2_queries.go
namanv3: this is done

srikanthccv (commented):
- is this run against https://github.com/SigNoz/dashboards? If not, please test this against the public dashboards as that will reveal the number of strange ways a dashboard would have been shaped in json. 
- does the validator reject the zero values when the required tag is set to true? we should ensure we don't incorrectly fail migration because of it, because it's genuinely possible to have a zero value, threshold at `0`.
- when a field is shaped differently from the expected, what do we do as part of the migration and what does frontend do today when such dashboard is imported?

namanv3:
Comment from @therealpandey on slack:
I'd suggest keeping a cobra command line entry for migration.
```
signoz resource migrate dashboard --input --output
```

srikanthccv:
@namanv3 can you please resolve the conflicts?

srikanthccv:
- The `svc EXISTS` from the earlier getting migrated to `svc EXISTS ''` for some reason, please check
- If there is both `filter: {expression: ...}` and the old `filters: { items: [...]}` what do we do? In other words, what do we do if there is both old and new format fields in the object?

namanv3:
@srikanthccv 
1. EXISTS was not handled in `migrateCommons` so have added handling to change it to `exists` in the migration logic
2. UI currently gives filter the precedence and that's what the migration script does. Have added a unit test for this as well

srikanthccv:
Some final findings

- count with attribute is lost the attribute. please check
- bar panel didn't have the label for thresholds, and thresholds without labels are getting dropped. IIRC, we had different thresholds types to solve this problem. Please check.
- looks like `normalizeOrderByKeys` running before `ensureDefaultAggregation` leaves the #SIGNOZ_VALUE fix not applied
- Formula type should have the order, limit, having should not be dropped. Check all valid fields gets preserved for each query type.
- question on variable `allSelected`, `multiSelect`, `customeValue` mapping, check with @AshwinBhatkal  to ensure incorrect mapping doesn't lead to bugs like the one happened here https://signoz-team.slack.com/archives/C088SMWD6E8/p1784663833894359

namanv3:
@srikanthccv 
1. as discussed, its the v4->v5 migration that drops the attribute, so we aren't fixing it in perses migration
2. there was a bug in the migration that assumed label is required. Fixed that now.
3. thanks for pointing this out, this was indeed the reason why 2 of the remaining signozvalue issues were there. Fixed and verified that no new issues sprung up
4. wrapInV5Envelope is not handling this, so added handling for this in migration
5. the allSelected flag does not have a counterpart in v2 schema so this issue can't happen. However I did add a check that if allowMultiple is false then the allowAllValue flag should also be false.

srikanthccv (commented):
LGTM, let's do the base64 to icon reverse mapping, then we are good to merge.

srikanthccv (approved):

therealpandey (approved):

PR #11716, opened 2026-06-15
feat(infra-monitoring): add default group by namespace name, cluster name for workloads. 

nikhilmantri0902:
### 📄 Summary

  v2 infra-monitoring workload list APIs (deployments, statefulsets, jobs, daemonsets) defaulted their groupBy to **workload name only**. But K8s names are unique only *within a namespace*, and `(name, namespace)` still collides across clusters. Result: two same-named workloads in different namespaces/clusters silently **collapsed into one row** with summed metrics and an arbitrary namespace/cluster in `meta`.

  This changes the default group identity to `(workload-name, namespace, cluster)` for all four workloads, so each real workload is its own row. `k8s.cluster.name` is included tolerantly — it has no canonical source in K8s and is only set when the collector's resource processor injects it (SigNoz Helm presets do), so a workload whose pods lack the label degrades gracefully to a `cluster=""` group rather than being dropped.

  #### Issues closed by this PR
  Part of https://github.com/SigNoz/engineering-pod/issues/5347

  ---

  ### ✅ Change Type

  - [x] ✨ Feature
  - [x] 🧪 Test-only (integration + unit tests for the above)

  ---

  ### 🧪 Testing Strategy

  - **Tests added/updated:** Each `test_*_groupby` (07–10) refactored into a scenario-driven, multi-key form; new `default_disambiguates_ns_and_cluster` scenario per workload backed by new
  `*_same_name_across_ns_and_clusters.jsonl` fixtures (one workload name spread over `(cluster-a,ns-x)`, `(cluster-a,ns-y)`, `(cluster-b,ns-x)`, and an **absent-cluster** `(—,ns-x)` group). Go
  `TestCompositeKeyFromLabels` extended with 3-key + empty-cluster cases.
  - **Manual verification:** Seeded an isolated dataset and hit `/api/v2/infra_monitoring/deployments` — confirmed 4 distinct rows incl. a `meta["k8s.cluster.name"] == ""` row carrying real metrics (not
  dropped, not `-1`).
  - **Edge cases covered:** same name across namespaces; same name across clusters; missing `k8s.cluster.name` label → own `""` group; explicit groupBy unaffected; existing per-workload
  value/filter/pagination/orderby tests unchanged (fixtures are constant-cluster, so group counts don't shift).

  ---

  - **Blast radius:** the four workload list endpoints' default (no-groupBy) response only. Explicit-groupBy callers unaffected (`Validate()` runs before the default is applied).
  Composite-key/metadata/pagination/pod-phase code is already generic over groupBy.
  - **Potential regressions:** clients assuming one row per workload name now get one row per `(name, namespace, cluster)`; `total`/row count can increase for genuinely-colliding workloads. `meta` still
  carries namespace + cluster.
  - **Rollback plan:** revert the 4 one-line `module.go` groupBy edits; pure-additive otherwise (no schema/migration).

  ---

  ### 📝 Changelog

  | Field | Value |
  |------|------|
  | Change Type | Feature |
  | Description | Infra-monitoring workload lists (Deployments/StatefulSets/Jobs/DaemonSets) now group by namespace and cluster in addition to name by default, so same-named workloads across
  namespaces/clusters are shown as distinct rows instead of being merged. |

  ---

  ### 📋 Checklist

  - [x] Manually tested
  - [x] Breaking changes documented (response now multi-row per name; see Risk)

  ---

  ## 👀 Notes for Reviewers

  - Core change is 4 one-liners in `module.go` (append `clusterNameGroupByKey`); everything downstream is already groupBy-generic. `Meta` keeps `k8s.cluster.name` because `getMetadata` returns groupBy cols too.
  - **Empty cluster is intentional & verified:** an absent `k8s.cluster.name` label → `JSONExtractString` → `""` in both the metric and metadata paths, no implicit exclusion, IN-filter `''` matches → a real `cluster=""` row. The `default_disambiguates_ns_and_cluster` cases assert this end-to-end.
  - Out of scope (deferred per #5347): frontend sidedrawer/detail-page filters; cluster groupBy for pods/namespaces/nodes/volumes.


namanv3:
what about other entities? Shouldn't nodes and namespaces also be grouped by cluster name, and volumes by namespace and cluster?

nikhilmantri0902:
So @namanv3 discussed and decided with Srikanth that it will be probabilistically rare that other entities, pod.uid, node.name are colliding across clusters. 

Although I will check for namespaces and persistent volume claims, as to how easily it can happen. Will do that in a seperate PR.

namanv3 (approved):

therealpandey (approved):

PR #11718, opened 2026-06-15
fix: schema fixes based on UI and MCP integration review

namanv3:
## Pull Request

---

### 📄 Summary

All issues found via UI integration and MCP testing (as mentioned in the below ticket, and some more):
* Add description in open api spec for Span Gaps
* Make label optional in DashboardtypesThresholdWithLabel
* Add LegendMode in legend spec with default value as list
* Change default value of fill mode to none
* Invalid panels in layouts aren't caught
* Two variables with same names are being saved
* Same error `{"status":"error","error":{"type":"invalid-input","code":"dashboard_invalid_patch","message":"JSON Patch could not be applied to the target dashboard"}}` in all scenarios of patch application errors (dashboard validations post application come fine)
* Variable name verification, and making name required in all variables
* Variable sort options should show up in the API Spec

Closes https://github.com/SigNoz/pulse-pod/issues/67

---

- Tests added/updated: Unit tests updated

---

- Blast radius: V2 dashboards
- Potential regressions: UI integration for v2 dashboards
- Rollback plan: Not needed, a fix can be put in another PR

---


docs/api/openapi.yml
```diff
@@ -2777,9 +2777,16 @@ components:
...
           type: object
+        mode:
+          $ref: '#/components/schemas/DashboardtypesLegendMode'
         position:
           $ref: '#/components/schemas/DashboardtypesLegendPosition'
       type: object
+    DashboardtypesLegendMode:
+      enum:
+      - list
+      - table
```
srikanthccv: please make a list of option that are part of the spec but not supported yet so we can follow up on them. we can't spec that `table` or other configs are allowed and not support for long time.

srikanthccv (dismissed):

docs/api/openapi.yml
namanv3: as discussed will remove the unsupported fields from the spec in this PR itself
Waiting for confirmation here: https://signoz-team.slack.com/archives/C0AFR4L8R2L/p1781678999886159

namanv3: The new `table` legend mode is the only enum value we don't support. For others support is being added

docs/api/openapi.yml
```diff
@@ -3432,6 +3442,20 @@ components:
       - color
       - columnName
       type: object
+    DashboardtypesTextVariableSpec:
+      properties:
+        constant:
```
AshwinBhatkal, line 3580: ohh yeah. I think missed to tell this. what is this constant field?

namanv3: If a user has some v long string that they are tired of remembering or typing in every panel, they can define a constant variable. It can't be changed when viewing a dashboard.

Sounds like another thing we don't support. I also forgot to bring this up yesterday

docs/api/openapi.yml
```diff
@@ -3432,6 +3442,20 @@ components:
...
+          type: boolean
+        display:
```
AshwinBhatkal, line 3582: can you make display also required here? same for value

namanv3: This is done

docs/api/openapi.yml
AshwinBhatkal: Yeah. True
Also, I have seen it only for text variable btw

namanv3: doesn't make sense to have it for a list variable.

docs/api/openapi.yml
```diff
@@ -2977,19 +2971,34 @@ components:
         customAllValue:
           type: string
         defaultValue:
-          $ref: '#/components/schemas/VariableDefaultValue'
```
therealpandey, line 2996: This `VariableDefaultValue` is no longer referenced anywhere but has a stale entry in components and the frontend generated schemas. Can we delete it?

therealpandey (changes requested):

docs/api/openapi.yml
namanv3: I think because every type referred in ListVariableSpec is walked and added, this type will be added to the spec. Would have to make changes to the way schema generation works I think

therealpandey: There's a standalone hanging entry. Check components/schemas/. You'll find this there

namanv3: That I got 😅 that dangling entry will be there cuz every type inside ListVariableSpec will be added to the spec.
ListVariableSpec's PrepareJSONSchema won't remove that entry from components/schemas/

therealpandey: Pushed 2 commits, please check.

therealpandey:
Blocking Blocking https://github.com/SigNoz/terraform-provider-signoz/pull/113

docs/api/openapi.yml
namanv3: @therealpandey tested this with a dashboard locally and using different values. Looks good. Thanks for picking this up while I was on leave!

Anything else needed here to merge this?

srikanthccv (dismissed):
LGTM

therealpandey (dismissed):

therealpandey (approved):

srikanthccv (approved):

PR #11719, opened 2026-06-15
fix(serviceaccount): return 204 No Content without a body

therealpandey:
## Problem

The service account endpoints that respond `204 No Content` declare an `application/json` string body in the generated OpenAPI spec:

```yaml
"204":
  content:
    application/json:
      schema:
        type: string
  description: No Content
```

This happens because their route definitions set `ResponseContentType: "application/json"` while `Response` is `nil` — and `ServeOpenAPI`'s nil-response branch still passes `WithContentType(ResponseContentType)`, so a content block is emitted for the 204.

A 204 has no body by definition, so generated OpenAPI clients try to unmarshal an empty body and fail. Concretely, the Terraform provider (oapi-codegen) errors on delete:

```
Error: Delete service_account
unexpected end of JSON input
```

(The server is correct — `render.Success(rw, http.StatusNoContent, nil)` — only the spec is wrong.)

## Fix

Set `ResponseContentType: ""` for the six service-account 204 responses — `DeleteServiceAccount`, `DeleteServiceAccountRole`, `UpdateMyServiceAccount`, `UpdateServiceAccount`, `UpdateServiceAccountKey`, `RevokeServiceAccountKey` — matching the existing `route_policy` delete (which already uses `""` and renders no 204 body). Regenerated `docs/api/openapi.yml`.

Result — the 204 now correctly carries no body:

```yaml
"204":
  description: No Content
```

## Notes

- Scope is service-account only, as a first pass. The same `204`-with-`content` pattern exists on other endpoints across the spec (a follow-up could fix `ServeOpenAPI` to never emit a content type when `Response == nil`, fixing the whole class).
- `go vet ./pkg/apiserver/signozapiserver/` clean; spec regenerated via `go run ./cmd/community generate openapi`.


therealpandey:
Closed by https://github.com/SigNoz/signoz/pull/11720

PR #11725, opened 2026-06-15
feat: v2 versions of public dashboard APIs 

namanv3:
## Pull Request

---

### 📄 Summary

The `GetPublicData` and `GetPublicWidgetQueryRange` rely on the dashboard structure in their response, and because the structure has changed in v2 dashboards, these APIs need a v2 version.

Other public dashboard APIs like `CreatePublic` etc don't rely on the dashboard structure so they don't need a v2 version.

#### Issues closed by this PR

Closes https://github.com/SigNoz/pulse-pod/issues/62
Closes https://github.com/SigNoz/pulse-pod/issues/5

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Unit tests and integration tests added
- Manual verification: Done by running the enterprise version 
- Edge cases covered: Manually and unit tests

---

- Blast radius: V2 Dashboards
- Potential regressions: None
- Rollback plan: Revert PR

---


pkg/types/dashboardtypes/perses_public_dashboard.go
```diff
@@ -0,0 +1,181 @@
...
+type GettablePublicDashboardDataV2 struct {
+	Dashboard       *GettableDashboardV2      `json:"dashboard"`
+	PublicDashboard *GettablePublicDasbhboard `json:"publicDashboard"`
+}
+
+// NewPublicDashboardDataFromDashboardV2 builds the anonymous v2 payload: panel queries
+// are redacted, and only the body fields v1 exposed (name, metadata, tags, spec) are set.
+func NewPublicDashboardDataFromDashboardV2(dashboard *DashboardV2, publicDashboard *PublicDashboard) *GettablePublicDashboardDataV2 {
+	spec := dashboard.Spec
+	redactPanelQueries(&spec)
```
srikanthccv, line 24: I noticed the variables query values are not redacted, same in v1. I think we should redact them as well because otherwise the stated claim of no leaking info is not correct.

srikanthccv: oh the variables are not supported.

ee/modules/dashboard/impldashboard/module.go
```diff
@@ -132,6 +134,49 @@ func (module *module) GetPublicWidgetQueryRange(ctx context.Context, id valuer.U
...
+		instrumentationtypes.CodeNamespace:    "dashboard",
+		instrumentationtypes.CodeFunctionName: "GetPublicWidgetQueryRangeV2",
+	})
+
+	dashboard, err := module.GetDashboardByPublicIDV2(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+
+	publicDashboard, err := module.GetPublic(ctx, dashboard.OrgID, dashboard.ID)
```
srikanthccv, line 168: unnecessary number of db round trips I believe. example `tagModule.ListForResource` fetches tags that GetPanelQuery never reads

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
```diff
@@ -341,6 +346,48 @@ func (q *QueryBuilderQuery[T]) validateAggregations(cfg validationConfig) error
...
+// validateSingleExpressionAggregation rejects a logs/traces aggregation whose
+// expression contains more than one function call. The frontend stores multiple
+// aggregations as a single comma-separated string and splits them before querying;
+// anything persisted server-side must already be split into one call per entry.
+func validateSingleExpressionAggregation(expression string) error {
+	// aggregationCallRegexp matches a single function-style aggregation call such as
+	// count() or sum(field). It is used to detect expressions that pack several
+	// comma-separated calls into one aggregation (e.g. "count(), sum(field)"), which
+	// must instead be provided as separate aggregation entries.
+	var aggregationCallRegexp = regexp.MustCompile(`[a-zA-Z0-9_]+\([^)]*\)`)
```
srikanthccv: two things from slack


- Is the frontend also fixed to ensure it doesn't keep invalid expression in the query spec. 
- the regex rejects certain expression countIf(body LIKE '%(done)%') where the query builder supports it, the whether an expression is valid should not deviate/split b/w here and the real query builder. see https://github.com/SigNoz/signoz/blob/94a5e568fd5ef0f5d590b74660bb776e909b4c72/pkg/querybuilder/agg_funcs.go and https://github.com/SigNoz/signoz/blob/94a5e568fd5ef0f5d590b74660bb776e909b4c72/pkg/querybuilder/agg_rewrite.go

srikanthccv (commented):
overall LGTM, some comments

pkg/types/dashboardtypes/perses_public_dashboard.go
namanv3: still good to remove them, and yes v1 still sends them. Remove query from v2

ee/modules/dashboard/impldashboard/module.go
namanv3: Fixed

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
namanv3: 1. yes, frontend is already parsing properly
2. made the validation same as the agg_rewrite parser, also moved it to dashboards so that query range behaviour doesn't change for now

pkg/types/dashboardtypes/testdata/perses.json
```diff
@@ -652,6 +652,7 @@
                                             "type": "builder_trace_operator",
                                             "spec": {
                                                 "name": "T1",
+                                                "expression": "A => B",
                                                 "aggregations": [
                                                     {
                                                         "expression": "count()",
```
srikanthccv, line 658: this made me realise that for the `builder_trace_operator`
- `GetWidgetQuery`'s trace-operator loop still doesn't split aggregations
- `WrapInV5Envelope` doesn't recognize


Please create a tracking issue, and get the trace operators gaps addressed everywhere including client. This is not a blocker for me.

pkg/modules/dashboard/impldashboard/v2_handler.go
```diff
@@ -376,3 +376,53 @@ func (handler *handler) DeleteV2(rw http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	dashboard, err := handler.module.GetDashboardByPublicIDV2(ctx, id)
+	if err != nil {
+		render.Error(rw, err)
+		return
+	}
+
+	publicDashboard, err := handler.module.GetPublic(ctx, dashboard.OrgID, dashboard.ID)
```
srikanthccv, line 396: non-blocker: I see this is a port of v1, but couldn't this be single `GetPublicByPublicID` or something.

tests/integration/tests/dashboard/05_v2_public_dashboard.py
```diff
@@ -0,0 +1,251 @@
...
+                "duration": "1h",
+                "variables": [
+                    {
+                        "kind": "ListVariable",
+                        "spec": {
+                            "name": "ns",
+                            "display": {"name": "ns"},
+                            "plugin": {
+                                "kind": "signoz/QueryVariable",
+                                "spec": {"queryValue": "SELECT DISTINCT service.name FROM signoz_logs"},
```
srikanthccv: ok, this should be corrected, this is not a valid query on SigNoz instance, the test shouldn't have invalid query unless the goal of the test itself is invalid query test.

srikanthccv (commented):
one comment about the queryValue used otherwise LGTM.

tests/integration/tests/dashboard/05_v2_public_dashboard.py
namanv3: done

pkg/modules/dashboard/impldashboard/v2_handler.go
namanv3: I thought that they are separate module methods cuz they return different resource kinds, they should be separate module methods. A combined module method would need to have 2 permissions attached to it.

Honestly though I'm not too sure if this "only one return type per module method" is an actual requirement from an access control POV, so I'll understand this once from Pandey. Given that this is a non blocker, I'm not changing this for now. Can do it in some follow up PR based on my discussion with Pandey

pkg/types/dashboardtypes/testdata/perses.json
namanv3: Didn't get what the exact issue is, created a ticket though: https://github.com/SigNoz/engineering-pod/issues/5598. Will get more clarity from you when I pick this up

tests/integration/tests/dashboard/05_v2_public_dashboard.py
srikanthccv: could you write something meaningful instead, nowhere one would want a query variable name `ns` that's querying the temporalities on the entire table. Please update to something meaningful. You could just take `host.name` example or something from the templates repo.

namanv3: done

srikanthccv (approved):

therealpandey (approved):

PR #11732, opened 2026-06-16
  chore(traces): remove legacy trace view toggle and redirect /trace-old to /trace

tewarig:
## Summary

this fixes https://github.com/SigNoz/engineering-pod/issues/5360

  The new trace detail view (V3) is the default at `/trace`, so the "Switch to legacy
  trace view" escape hatch and the persisted preference behind it are no longer needed.
  This PR removes that toggle, deletes the `TRACE_DETAILS_PREFER_OLD_VIEW` localStorage
  key, and redirects the legacy `/trace-old/:id` route to `/trace/:id`.

  ## Changes

  - **Remove the "Switch to legacy trace view" button** from the V3 trace header
    (`TraceDetailsHeader`), along with its click handler and the `ViewSwitched`
    analytics call it triggered.
  - **Drop the `TRACE_DETAILS_PREFER_OLD_VIEW` localStorage key** and all its readers/writers:
    - the `preferOld` redirect in the trace page entry,
    - the `entryPreferOldView` property on the `Data loaded` analytics event,
    - the now-dead "Try new experience" button in the V2 `TraceMetadata` (it only
      existed to clear this key when landing on the old route).
  - **Redirect `/trace-old/:id` → `/trace/:id`** via a small `TraceDetailOldRedirect`
    component, preserving the query string and hash.
  - **Remove the `TraceDetailV3Page` wrapper** — it had become a passthrough once the
    `preferOld` redirect was gone; the route now lazy-loads `TraceDetailsV3` directly.

  ## Notes

  - The V2 page and `ROUTES.TRACE_DETAIL_OLD` constant are intentionally left in place;
    `/trace-old` now just redirects. 
  -  we also need to remove v2 in a follow up
  ## Testing

  - `pnpm tsgo --noEmit` — clean
  - `pnpm oxlint` on changed files — 0 warnings / 0 errors
  - Updated and ran `TraceDetailsHeader` + `useTraceDetailLogEvent` tests — all passing
  
  
  

https://github.com/user-attachments/assets/a5cb3942-a28b-4317-928d-879d2abb6b30


frontend/src/container/TraceMetadata/TraceMetadata.tsx
aks07, line 1: Let's not make change to old component. This page should be inaccessible anyways.

aks07 (commented):
One minor comment

frontend/src/container/TraceMetadata/TraceMetadata.tsx
tewarig: i wanted to remove TRACE_DETAILS_PREFER_OLD_VIEW from localstorage . that's why these changes exits. will revert these and add key back.

aks07 (approved):
Changes look good. Do test thoroughly.

H4ad (approved):

PR #11542, opened 2026-06-02
feat: extend error responses with new error struct

tushar-signoz:
## Pull Request

---

### 📄 Summary

Enriches API validation errors with **structured, machine-consumable suggestions** so clients (frontend / API consumers) can offer a fix — a typo correction or the list of allowed values — instead of parsing English error strings.

Suggestions now have **two homes**, chosen by how closely the hint is tied to a specific detail:

- **Top-level `suggestions[]`** — relate to the error as a whole (the main `message` already names the bad token), e.g. unknown query type, invalid signal, unknown JSON field, invalid `Having` reference, invalid aggregate `orderBy` key.
Example:
```
{
  "type": "invalid-input",
  "code": "invalid_input",
  "message": "invalid query 'A': invalid order by key 'badkey' for order by clause #1 in query 'A'",
  "suggestions": ["valid references: `0`, `count()`, `severity_text`"]
}
```
- **Per-detail `errors[].suggestions[]`** — bound to a specific entry in the `errors` array (via `WithSuggestiveAdditional`), e.g. a field-not-found detail or a single `Having` syntax error carrying its corrected-expression hint. Example: 
```
{
  "type": "invalid-input",
  "code": "invalid_input",
  "message": "Syntax error in `Having` expression",
  "errors": [
    {
      "message": "line 1:9 expecting one of {'*', '+', '-', (, ), IDENTIFIER, number, string} but got EOF",
      "suggestions": ["did you mean: `count() > 0`"]
    }
  ]
}
```

**Note:** the earlier `invalidReferences` field was **removed** — the message already carried that information. All hints (typo fix + valid-value set) flow through `suggestions`. The field indeed makes sense however all the errors in the code base refer to the invalid token within the message body making it redundant.

**Enterprise querier handler (`ee/querier/handler.go`)**
- The `query_range` handler used to re-wrap request-decode errors with `%v`, which **flattened the structured error and discarded its `suggestions`**. It now preserves structured `invalid-input` errors and only wraps opaque decoder errors (e.g. malformed JSON). Without this, none of the decode-time field/enum suggestions reached the client.

**Moved Unmarshling to bind pkg**
[...]

PR #11740, opened 2026-06-16
chore: refactor error response + normalize ClickHouse telemetrystore errors

tushar-signoz:
## Summary

Follow-up to #11542 on the structured error response, plus a ClickHouse error-normalization layer in the telemetry store.

- **OpenAPI / frontend types** — every error-response field (`type`, `code`, `message`, `url`, `errors[]`, `retry`, `suggestions`) is now `required`; conditionally-empty ones are `nullable` (`T | null`) instead of optional, so the generated TS types drop the `?`. `errors` now serializes as `null` (not `[]`) when empty.
- **`errors` package** — suggestion builders are `New`-prefixed (`NewSuggestionsOnLevenshteinDistance`, `NewValidReferences`, `NewSuggestionsFromFunc`); `NewValidReferences` takes a `noun` and renders `valid <noun> are ...` via exported `errors.Noun*` constants. Added `base.Unwrap()` (so `errors.Is`/`As` reach the cause) and `WrapCanceledf`/`NewCanceledf`. Code regex widened to allow dotted codes.
- **ClickHouse telemetry store** — a new `castError` maps raw driver errors to structured base errors: server exceptions by `chproto.Error` code (→ `InvalidInput`/`NotFound`/`Internal`, each with a `telemetrystore.clickhouse.*` code), context cancel/deadline → `Canceled`/`Timeout`. Returned `Rows`/`Row`/`Batch` are wrapped so later scan/append/close errors normalize too. `AsyncInsert` moved to the non-deprecated `Exec(WithAsync)`; `ch-go` promoted to a direct dep.
- **Organization & docs** — `JSONFieldNames` moved from `pkg/http/binding` to a new `pkg/jsonschema`; `docs/contributing/go/errors.md` documents the response field-by-field.

## Change Type
- [x] ♻️ Refactor
- [x] 🛠️ Infra / Tooling (OpenAPI + frontend client regen)
- [x] ✨ Feature (structured ClickHouse query errors)

## Testing
- Unit tests in `pkg/errors`, `pkg/querybuilder`, `pkg/http/binding` updated and passing.
- Regenerated `docs/api/openapi.yml` + frontend client (`pnpm generate:api`); frontend type-check and `PanelStatus` tests pass.

## Risk & Impact
- Error responses always include the previously-optional fields (empty → `null`); frontend types are stricter.
- **Behavior change**: ClickHouse telemetry-store callers now get structured base errors instead of raw `clickhouse.Exception` — assert via `errors.As` (cause stays reachable through `Unwrap`).

## Changelog

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | N/A (internal error-response + telemetrystore error refactor) |

### DIFF
--- a/pkg/errors/http.go
+++ b/pkg/errors/http.go
@@ -7,22 +7,22 @@ import (
 )
 
 type JSON struct {
-	Type        string                    `json:"type,omitempty"`
+	Type        string                    `json:"type" required:"true"`
 	Code        string                    `json:"code" required:"true"`
 	Message     string                    `json:"message" required:"true"`
-	Url         string                    `json:"url,omitempty"`
-	Errors      []responseerroradditional `json:"errors,omitempty"`
-	Retry       *responseretryjson        `json:"retry,omitempty"`
-	Suggestions []string                  `json:"suggestions,omitempty"`
+	Url         string                    `json:"url" required:"true" nullable:"true"`
+	Errors      []responseerroradditional `json:"errors" required:"true" nullable:"true"`
+	Retry       *responseretryjson        `json:"retry" required:"true" nullable:"true"`
+	Suggestions []string                  `json:"suggestions" required:"true" nullable:"true"`
 }
 
 type responseretryjson struct {
-	Delay time.Duration `json:"delay"`
+	Delay time.Duration `json:"delay" required:"true" nullable:"false"`
 }
 
 type responseerroradditional struct {
-	Message     string   `json:"message,omitempty"`
-	Suggestions []string `json:"suggestions,omitempty"`
+	Message     string   `json:"message" required:"true"`
+	Suggestions []string `json:"suggestions" required:"true" nullable:"true"`
 }
 
 func AsJSON(cause error) *JSON {
--- a/pkg/errors/http.go
+++ b/pkg/errors/http.go
@@ -29,9 +29,12 @@ func AsJSON(cause error) *JSON {
 	// See if this is an instance of the base error or not
 	t, c, m, _, u, a := Unwrapb(cause)
 
-	rea := make([]responseerroradditional, len(a))
-	for k, v := range a {
-		rea[k] = responseerroradditional{Message: v.message, Suggestions: v.suggestions}
+	var rea []responseerroradditional
+	if len(a) > 0 {
+		rea = make([]responseerroradditional, len(a))
+		for k, v := range a {
+			rea[k] = responseerroradditional{Message: v.message, Suggestions: v.suggestions}
+		}
 	}
 
 	var retry *responseretryjson
@@ -54,9 +57,12 @@ func AsURLValues(cause error) url.Values {
 	// See if this is an instance of the base error or not
 	_, c, m, _, u, a := Unwrapb(cause)
 
-	rea := make([]responseerroradditional, len(a))
-	for k, v := range a {
-		rea[k] = responseerroradditional{Message: v.message, Suggestions: v.suggestions}
+	var rea []responseerroradditional
+	if len(a) > 0 {
+		rea = make([]responseerroradditional, len(a))
+		for k, v := range a {
+			rea[k] = responseerroradditional{Message: v.message, Suggestions: v.suggestions}
+		}
 	}
 
 	errors, err := json.Marshal(rea)
--- a/pkg/errors/http.go
+++ b/pkg/errors/http.go
@@ -57,12 +57,12 @@ func AsURLValues(cause error) url.Values {
 	// See if this is an instance of the base error or not
 	_, c, m, _, u, a := Unwrapb(cause)
 
-	var rea []responseerroradditional
-	if len(a) > 0 {
-		rea = make([]responseerroradditional, len(a))
-		for k, v := range a {
-			rea[k] = responseerroradditional{Message: v.message, Suggestions: v.suggestions}
-		}
+	// Unlike AsJSON (whose null `errors` honors the OpenAPI nullable contract),
+	// this goes into a redirect query param that the frontend JSON.parses and
+	// .maps over, so keep a non-nil empty slice -> marshals to "[]", not "null".
+	rea := make([]responseerroradditional, len(a))
+	for k, v := range a {
+		rea[k] = responseerroradditional{Message: v.message, Suggestions: v.suggestions}
 	}
 
 	errors, err := json.Marshal(rea)
--- a/pkg/errors/http.go
+++ b/pkg/errors/http.go
@@ -11,9 +11,9 @@ type JSON struct {
 	Code        string                    `json:"code" required:"true"`
 	Message     string                    `json:"message" required:"true"`
 	Url         string                    `json:"url" required:"true" nullable:"true"`
-	Errors      []responseerroradditional `json:"errors" required:"true" nullable:"true"`
+	Errors      []responseerroradditional `json:"errors" required:"true" nullable:"false"`
 	Retry       *responseretryjson        `json:"retry" required:"true" nullable:"true"`
-	Suggestions []string                  `json:"suggestions" required:"true" nullable:"true"`
+	Suggestions []string                  `json:"suggestions" required:"true" nullable:"false"`
 }
 
 type responseretryjson struct {
@@ -22,20 +22,14 @@ type responseretryjson struct {
 
 type responseerroradditional struct {
 	Message     string   `json:"message" required:"true"`
-	Suggestions []string `json:"suggestions" required:"true" nullable:"true"`
+	Suggestions []string `json:"suggestions" required:"true" nullable:"false"`
 }
 
 func AsJSON(cause error) *JSON {
 	// See if this is an instance of the base error or not
 	t, c, m, _, u, a := Unwrapb(cause)
 
-	var rea []responseerroradditional
-	if len(a) > 0 {
-		rea = make([]responseerroradditional, len(a))
-		for k, v := range a {
-			rea[k] = responseerroradditional{Message: v.message, Suggestions: v.suggestions}
-		}
-	}
+	rea := responseAdditionals(a)
 
 	var retry *responseretryjson
 	if r := retryOf(cause); r != nil {
@@ -49,7 +43,7 @@ func AsJSON(cause error) *JSON {
 		Url:         u,
 		Errors:      rea,
 		Retry:       retry,
-		Suggestions: suggestionsOf(cause),
+		Suggestions: nonNilStrings(suggestionsOf(cause)),
 	}
 }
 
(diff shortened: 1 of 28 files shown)


pkg/reflectutil/reflectutil.go
```diff
@@ -0,0 +1,38 @@
+// Package reflectutil holds small reflection helpers shared across packages.
+package reflectutil
```
therealpandey: @tushar-signoz can we move this to a `jsonschema` package at the root. There are other functions related to `jsonschema` which can go there.

pkg/telemetrystore/clickhousetelemetrystore/provider.go
```diff
@@ -3,13 +3,33 @@ package clickhousetelemetrystore
...
+	chproto.ErrUnknownStorage:               true,
+	chproto.ErrUnknownElementInAst:          true,
+	chproto.ErrUnknownTypeOfQuery:           true,
+	chproto.ErrIllegalTypeOfArgument:        true,
+	chproto.ErrIllegalColumn:                true,
+	chproto.ErrNumberOfArgumentsDoesntMatch: true,
+	chproto.ErrTooManyArgumentsForFunction:  true,
+	chproto.ErrTooLessArgumentsForFunction:  true,
+}
+
```
tushar-signoz: @therealpandey Here is the mapping

pkg/telemetrystore/clickhousetelemetrystore/provider.go
```diff
@@ -3,13 +3,33 @@ package clickhousetelemetrystore
...
+	chproto "github.com/ClickHouse/ch-go/proto"
 	"github.com/ClickHouse/clickhouse-go/v2"
 	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
+	"github.com/SigNoz/signoz/pkg/errors"
 	"github.com/SigNoz/signoz/pkg/factory"
 	"github.com/SigNoz/signoz/pkg/telemetrystore"
 	"go.opentelemetry.io/otel/metric"
 )
 
+var userFacingClickHouseErrorCodes = map[chproto.Error]bool{
```
therealpandey:
```suggestion
```

Can we make this a `map[chproto.Error]func() error` map? Give better wrapped errors better messages according to the chproto.Error encountered.

therealpandey (changes requested):

tushar-signoz:
### Review notes (for code owners)

_Updated for `26a8336`._ `go build ./...`, `go vet`, and the `pkg/errors` / `pkg/http/binding` / `pkg/querybuilder` tests are green. OpenAPI `required`/`nullable` changes line up with the Go struct tags and the regenerated TS types (frontend test fixtures updated to match).

**✅ Resolved — `PrepareBatch` typed-nil** (`provider.go`): the new `if batch == nil { return nil, castError(err) }` guard now returns a true nil batch on error instead of a non-nil `*batchWithCastError` wrapping a nil. (The matching `QueryRow` nil guard is defensive — the driver doesn't return a nil row — but harmless.)

**🔸 Open — `AsURLValues` emits `"null"` instead of `"[]"` — `pkg/errors/http.go:60-68`**
`var rea []responseerroradditional` means that when an error carries no detail entries, `json.Marshal(rea)` produces the string `"null"` rather than `"[]"` in the `errors` query param. The only consumer is the SSO-callback redirect (`implsession/handler.go:165` → frontend `Login/index.tsx` `parseErrors`), which does `JSON.parse(errors).map(...)`. `JSON.parse("null").map` throws, but it's caught and returns `[]`, so **no user-visible breakage** — just a `console.error` on every callback error with no detail entries (the common case). `AsURLValues` isn't bound by the OpenAPI `nullable` contract (that applies to the JSON body via `AsJSON`), so keeping a non-nil empty slice here avoids the noise:
```go
rea := make([]responseerroradditional, len(a)) // non-nil even when empty -> marshals to "[]"
```

**Scope note (not a blocker):** the `clickhousetelemetrystore` `castError` layer + the `AsyncInsert` → `Exec(WithAsync)` migration are correct and do take effect on the live query path (`ClickhouseDB()` returns the provider), but aren't mentioned in the PR description ("refactor error suggestion methods"). Worth a line so reviewers don't skim those files as a no-op rename.

pkg/telemetrystore/clickhousetelemetrystore/provider.go
```diff
@@ -3,13 +3,76 @@ package clickhousetelemetrystore
...
+var (
+	ErrCodeSyntaxError       = errors.MustNewCode("telemetrystore.clickhouse.syntax_error")
+	ErrCodeUnknownTable      = errors.MustNewCode("telemetrystore.clickhouse.unknown_table")
+	ErrCodeUnknownDatabase   = errors.MustNewCode("telemetrystore.clickhouse.unknown_database")
+	ErrCodeUnknownIdentifier = errors.MustNewCode("telemetrystore.clickhouse.unknown_identifier")
+	ErrCodeIllegalArgument   = errors.MustNewCode("telemetrystore.clickhouse.illegal_argument")
+
+	ErrCodeQueryCanceled   = errors.MustNewCode("telemetrystore.clickhouse.query_canceled")
+	ErrCodeQueryTimeout    = errors.MustNewCode("telemetrystore.clickhouse.query_timeout")
+	ErrCodeExecutionFailed = errors.MustNewCode("telemetrystore.clickhouse.execution_failed")
```
therealpandey:
```suggestion
	ErrCodeSyntaxError       = errors.MustNewCode("syntax_error")
	ErrCodeUnknownTable      = errors.MustNewCode("unknown_table")
	ErrCodeUnknownDatabase   = errors.MustNewCode("unknown_database")
	ErrCodeUnknownIdentifier = errors.MustNewCode("unknown_identifier")
	ErrCodeIllegalArgument   = errors.MustNewCode("illegal_argument")

	ErrCodeQueryCanceled   = errors.MustNewCode("query_canceled")
	ErrCodeQueryTimeout    = errors.MustNewCode("query_timeout")
	ErrCodeExecutionFailed = errors.MustNewCode("execution_failed")
```

pkg/telemetrystore/clickhousetelemetrystore/provider.go
```diff
@@ -182,3 +252,27 @@ func (p *provider) ServerVersion() (*driver.ServerVersion, error) {
...
 	return p.clickHouseConn.Contributors()
 }
+
+func castError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	if errors.Is(err, context.Canceled) {
+		return errors.WrapCanceledf(err, ErrCodeQueryCanceled, "ClickHouse query canceled")
```
therealpandey:
```suggestion
		return errors.WrapCanceledf(err, ErrCodeQueryCanceled, "query canceled")
```

pkg/telemetrystore/clickhousetelemetrystore/provider.go
```diff
@@ -182,3 +252,27 @@ func (p *provider) ServerVersion() (*driver.ServerVersion, error) {
...
+	}
+	if errors.Is(err, context.DeadlineExceeded) {
+		return errors.WrapTimeoutf(err, ErrCodeQueryTimeout, "ClickHouse query deadline exceeded")
```
therealpandey:
```suggestion
		return errors.WrapTimeoutf(err, ErrCodeQueryTimeout, "query timed out")
```

pkg/telemetrystore/clickhousetelemetrystore/provider.go
```diff
@@ -182,3 +252,27 @@ func (p *provider) ServerVersion() (*driver.ServerVersion, error) {
...
+	var ex *clickhouse.Exception
+	if !errors.As(err, &ex) {
+		return err
+	}
+
+	if wrap, ok := clickHouseExceptionWrappers[chproto.Error(ex.Code)]; ok {
+		return wrap(err, ex)
+	}
+
+	return errors.WrapInternalf(err, ErrCodeExecutionFailed, "ClickHouse SQL execution failed")
```
therealpandey:
```suggestion
	return err
```

therealpandey (changes requested):

pkg/telemetrystore/clickhousetelemetrystore/provider.go
```diff
@@ -147,20 +215,19 @@ func (p *provider) Exec(ctx context.Context, query string, args ...interface{})
...
+	return castError(err)
 }
 
 func (p *provider) AsyncInsert(ctx context.Context, query string, wait bool, args ...interface{}) error {
 	event := telemetrystore.NewQueryEvent(query, args)
 
 	ctx = telemetrystore.WrapBeforeQuery(p.hooks, ctx, event)
-	// TODO: migrate to WithAsync() — https://github.com/SigNoz/engineering-pod/issues/5093
-	err := p.clickHouseConn.AsyncInsert(ctx, query, wait, args...) //nolint:staticcheck
+	err := p.clickHouseConn.Exec(clickhouse.Context(ctx, clickhouse.WithAsync(wait)), query, args...)
```
therealpandey, line 225: Thank you ❤️

therealpandey (changes requested):

therealpandey (dismissed):

therealpandey (approved):

AshwinBhatkal (approved):

pkg/errors/http.go
```diff
@@ -7,32 +7,29 @@ import (
...
 type JSON struct {
-	Type        string                    `json:"type,omitempty"`
+	Type        string                    `json:"type" required:"true"`
 	Code        string                    `json:"code" required:"true"`
 	Message     string                    `json:"message" required:"true"`
-	Url         string                    `json:"url,omitempty"`
-	Errors      []responseerroradditional `json:"errors,omitempty"`
-	Retry       *responseretryjson        `json:"retry,omitempty"`
-	Suggestions []string                  `json:"suggestions,omitempty"`
+	Url         string                    `json:"url" required:"true" nullable:"true"`
```
therealpandey, line 13:
```suggestion
	Url         string                    `json:"url"`
```

pkg/errors/http.go
```diff
@@ -7,32 +7,29 @@ import (
...
+	Errors      []responseerroradditional `json:"errors" required:"true" nullable:"false"`
+	Retry       *responseretryjson        `json:"retry" required:"true" nullable:"true"`
```
therealpandey, line 15:
```suggestion
	Retry       *responseretryjson        `json:"retry" nullable:"true"`
```

pkg/errors/http.go
tushar-signoz: Will look into this

pkg/errors/http.go
tushar-signoz: Will look into this

PR #11754, opened 2026-06-17
chore: send warning instead of error for unseen metrics and missing (…

srikanthccv:
…metric, key)

Fixes https://github.com/SigNoz/engineering-pod/issues/5378
Fixes https://github.com/SigNoz/engineering-pod/issues/4900


pkg/telemetrymetrics/statement_builder.go
```diff
@@ -101,9 +101,29 @@ func (b *MetricQueryStatementBuilder) Build(
 		return nil, err
 	}
 
+	var pairFallbackWarnings []string
+	for _, sel := range keySelectors {
+		if _, ok := keys[sel.Name]; !ok {
+			keys[sel.Name] = []*telemetrytypes.TelemetryFieldKey{{
```
namanv3, line 107: so where clause visitor expects every key to be present, so in a way we're just building the absent ones here. Seems a bit iffy. Shouldn't the fix be in the visitor? 

Also if we are building the absent keys here, why even make the `keys, _, err := b.metadataStore.GetKeysMulti(ctx, keySelectors)` call? We can just build the whole map without that call

namanv3:
any update required to this doc? it is a part of the error being sent by the API currently
https://signoz.io/docs/userguide/search-troubleshooting/#key-fieldname-not-found

pkg/telemetrymetrics/statement_builder.go
srikanthccv: >so where clause visitor expects every key to be present, so in a way we're just building the absent ones here. Seems a bit iffy. Shouldn't the fix be in the visitor?

correct, the ideal place to make this change is visitor. however, the scope of this PR is just metrics, and traces and logs implementations are yet to be made (and they will be made by logs/traces primary owner @tushar-signoz). so the till they land, the best option is to ensure there won't be any missing keys before visitor gets called.


> Also if we are building the absent keys here, why even make the keys, _, err := b.metadataStore.GetKeysMulti(ctx, keySelectors) call? We can just build the whole map without that call

Right now, and specifically for metrics at the time moment yes because the field mapper maps everything to `labels`, however, given the right field context, they should map to the respective `attrs`, `scope_attrs`, and `resource_attrs` column on CH so it's important to have the metadata call to enrich the context for a given key selector so that's reason to keep the metadata call.

srikanthccv:
Yes, the doc will be updated once the logs and traces are also addressed https://github.com/SigNoz/engineering-pod/issues/5380 and https://github.com/SigNoz/engineering-pod/issues/5379

pkg/telemetrymetrics/statement_builder.go
namanv3: So when the logs+traces fix will be in the visitor, and at that point ideally we remove this change in the statement builder right?

srikanthccv: yes

namanv3 (approved):
non blocker: add a todo in statement builder

therealpandey (approved):

PR #11779, opened 2026-06-18
feat(llm-attribute-mapping): read-only listing [2/5]

tewarig:
## Summary

Second PR in the **LLM Observability — Attribute Mapping** stack. Adds the **read-only listing**: a single, GitHub-style table where mapping groups are collapsible banner rows and each group expands to reveal its mappers, loaded lazily from the API. Enable/disable switches are rendered but **read-only** in this layer; staging (draft), persistence (save/discard) and the create/edit drawers land in PRs 3–4.

## Component structure

```
AttributeMapping/
├── LLMObservabilityAttributeMapping.tsx    — page shell: header + inner tabs (Test disabled)
├── components/AttributeMappingHeader/      — description; Save/Discard render only when dirty
├── AttributeMappingsTab/
│   ├── AttributeMappingsTab.tsx            — page-level error state around the table
│   ├── hooks/useAttributeMappingStore.ts   — read-only store: groups query → MappingGroup[]
│   └── components/MappingsTable/
│       ├── MappingsTable.tsx               — column-header table + groups Collapse + banner skeletons
│       ├── MappingsColgroup/               — shared <colgroup>, single source of column widths
│       ├── GroupHeader/                    — group name + condition count (Collapse label)
│       ├── GroupHeaderActions/             — read-only group enable switch (Collapse extra)
│       ├── GroupMappers/                   — lazy per-group mappers query + error/empty/skeleton rows
│       └── MapperRow/                      — mapper row (+ MapperRowSkeleton)
├── types.ts                                — UI domain types (MappingGroup, Mapping, SourceConfig)
└── utils.ts                                — DTO → domain builders
```

Two deliberate style-sharing exceptions, both cross-component contracts rather than private styling: the column widths live only in `MappingsColgroup.module.scss` (both tables render it), and `MapperRowSkeleton` shares `MapperRow.module.scss`. The fixed-layout `.table` rule is duplicated between the header and panel tables with "must mirror" comments — cross-file `composes` is off the table (the target file gets parsed as plain CSS at build time).

## UX polish

- **Reads as one table** — groups are full-width banner rows inside the table, not a table-in-a-table; mapper rows indent under their group and align to the single shared column header.
- **Expand reveal** — panel rows fade/slide in with a small capped stagger (via `motion`), riding antd Collapse's height animation; state rows (error/empty) fade in step.
[...]

PR #11780, opened 2026-06-18
 feat(llm-attribute-mapping): group + mapper drawers + diff-save [3-4/5]

tewarig:
## Summary

PRs **3 + 4** in the **LLM Observability — Attribute Mapping** stack, combined into one. Adds **group *and* mapper editing** end-to-end: drawers to create/edit/delete mapping groups and their nested mappers, the draft layer in the store, and the diff-based save that persists both.

PR 2 (listing, #11779) has **merged into `main`**, and this branch has been **synced with `main`** — so this PR targets `main` directly and its diff is scoped to the group- and mapper-editing work on top of the merged listing.

### Group editing
- **`GroupFormDrawer`** + styles — name, enabled toggle, and two `ConditionKeyList` editors (span-attribute keys / resource keys).
- **`ConditionKeyList`** + **`KeySearchInput`** — condition-key editor with typeahead suggestions from `/api/v1/fields/keys` (signal=traces).
- **`useGroupFormDrawer`** — owns drawer open/mode/draft (form-local).
- **`MappingsTable`** grows an Actions column — an inline enabled toggle plus a **`GroupActionsMenu`** overflow (`⋮`) menu holding **Edit** / **Delete** — and an "Add a new group" affordance in a toolbar above the table.
- **`ConditionsTooltip`** on the group header surfaces the configured condition keys.

### Mapper editing
- **`MapperFormDrawer`** + styles — pick a target attribute and an ordered list of source attributes, with **drag-to-reorder** (`dnd-kit`) via **`SourceAttributeRow`**, add/remove rows, and `KeySearchInput` typeahead for each key.
- **`useMapperFormDrawer`** — owns the mapper drawer's open/mode/draft.
- **`MapperActionsMenu`** (`⋮`) on each mapper row — **Edit** / **Delete**; `MapperRow` / `GroupMappers` reworked to host inline mapper actions and staging.

### Store + draft layer
- **Store grows** to the full draft pattern: working-copy `draft` + `snapshot`, `isDirty`, `isSaving`, `saveError`, `upsertGroup` / `removeGroup` / `toggleGroup`, `upsertMapper` / `removeMapper` / `toggleMapper`, `save`, `discard`.
- **`DiscardChangesDialog`** guards `discard` when the working copy is dirty.
- Save/discard surface `toast` success/error feedback.

### Diff-based save
- **`saveDraft.ts`** — `persistDraft` diffs staged **groups and mappers** against the snapshot and issues the minimal `createGroup` / `updateGroup` / `deleteGroup` and, per group, `persistMappers` → `createMapper` / `updateMapper` / `deleteMapper` calls. `createGroup` returns the new server id so its mappers are created under it in the same save.

## Save / refetch flow
`save()` runs `persistDraft`, then `await`s the groups query's own `refetch` (destructured as `refetchGroups`), which refreshes the active groups list in place — it stays mounted, so fresh data swaps in without a loading flash. It resolves once the refetch settles; only then is the working copy reset (`setDraft(null)`) so the seeding effect re-seeds it from the fresh server data (new server ids included).

Using the query's own `refetch` — rather than `queryClient.invalidateQueries` with a key predicate — keeps the refresh scoped to this one query by construction: it can't touch the per-group mapper lists, and there's no hardcoded query-key string to drift from codegen. (Trade-off: `refetch` refreshes only this hook's instance of the groups query; the store is the sole consumer today, so if that list ever gains other readers, `invalidateQueries` would be the way to refresh them all.)

Mapper lists use `refetchOnMount: false`, so an edited group's mappers refresh on a hard reload rather than on next expand.

**Ordering:** both `persistDraft` (groups) and `persistMappers` (mappers) apply creates/updates **first** and deletes **last**. Deletes are irreversible — group deletes cascade mappers server-side — so they run only after the additive work has succeeded; a mid-save failure then leaves at worst incomplete additions rather than entities deleted with no replacement persisted.

**Known limitation of this ordering.** Names are unique per scope (unique index on `(org_id, name)` for groups, `(group_id, name)` for mappers). Because deletes run last, deleting an entity and creating a new one with the *same* name in a single save collides with the still-present old row, so the save errors. This is recoverable — split the rename-by-replace across two saves — and we deliberately prefer it to delete-first, which would risk **irreversible data loss** if a create fails after the deletes have already committed (the create/update/delete calls are separate mutations, not one transaction).

## Refactor
- Barrel `index.ts` re-exports under `MappingsTable/` (and the header/drawer folders) are removed in favor of direct component imports, per the repo's no-barrel-files convention.
- **`ModelPricing` polish** (unrelated to the mapping work, carried on this branch): label casing ("Model Costs", "Unpriced Models", "Billing Model ID") and a shared `tableCellStyles` class applied across the model-costs table cells for consistent typography.

## Known follow-ups (tracked for later PRs)
- **Snapshot is derived live from the query.** `save`/`isDirty`/`discard` read a snapshot memoized off `groupsQuery.data`. It's safe today (the save refetch is the only thing that refreshes the groups list, and it runs after `persistDraft`), but once something can trigger a groups-list refetch mid-edit, the baseline could shift under an in-progress edit — the intended fix is to pin the snapshot in state at draft-seed time.
- **Retry safety after a mid-save failure.** Newly `createGroup`-ed ids aren't written back into the working copy until the post-save re-seed, so a retry after a partial failure can re-create already-created groups.
- **Collapse expansion is lost for a just-created group after save.** Expansion is keyed by group `localId`; on the post-save re-seed a new group's `localId` becomes its server id, so its expanded-state entry (the old client localId) no longer matches and the panel collapses. Minor UX — the real fix is the same "feed created server ids back into the draft" work above (avoids the full re-seed that flips the id).
- Source-priority sort order (carried from PR 2) and the `KeySearchInput` keyboard/ARIA gaps.

## Stack
1. foundation ✅ merged
2. listing (#11779) ✅ merged
3. **group drawer ⬅️ this PR**
4. **mapper drawer ⬅️ this PR** (folded in; #11781)
5. tests

## Testing notes
Staging behavior (toggle / delete-from-working-copy / edit callback for both groups and mappers) is covered by the tab integration tests. Direct coverage for the diff-save logic in `saveDraft.ts` (create/update/delete ordering, group + mapper) and the store's `save` + re-seed flow lands with the test PR (5/5).

## Verification
- `pnpm tsgo --noEmit` ✅
- `pnpm oxlint` (feature dir) ✅
- `pnpm build` ✅
- `pnpm jest` (feature suite) ✅
- pre-commit hooks ✅

https://github.com/user-attachments/assets/5322c9f9-0a69-432c-a0b5-26ac0c093262


aks07:
image

Shouldnt this be attribute instead of span attribute?

aks07:
image

This tooltip looks a little too verbose. Should we stick to simple headings?

aks07:
image

I dont think this header needs to stay above the tabs. Does not match other places

frontend/src/container/LLMObservability/AttributeMapping/AttributeMappingsTab/components/MappingsTable/GroupHeader/ConditionsTooltip.tsx
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	return (
+		<div
+			className={styles.conditionsTooltip}
+			data-testid="group-conditions-tooltip"
+		>
+			{attributes.length > 0 && (
+				<div className={styles.section}>
+					<span className={styles.sectionTitle}>
```
aks07: Use Typography.

frontend/src/container/LLMObservability/AttributeMapping/AttributeMappingsTab/components/MappingsTable/GroupHeader/GroupHeader.module.scss
```diff
@@ -17,3 +18,15 @@
 	font-size: var(--font-size-xs);
 	white-space: nowrap;
 }
+
+.conditionCount {
+	composes: groupCount;
```
aks07, line 23: Did we need to use composes here?

aks07:
image

Failed and empty state looking same. Lets atleast add a caution icon or somethin

frontend/src/container/LLMObservability/AttributeMapping/AttributeMappingsTab/components/MappingsTable/GroupMappers/GroupMappers.tsx
```diff
@@ -19,72 +24,125 @@ const STATE_ROW_MOTION = {
...
-function ErrorRow({ groupId }: StateRowProps): JSX.Element {
-	return (
-		<motion.tr className={styles.mapperStateRow} {...STATE_ROW_MOTION}>
+function GroupMappers({
+	group,
+	store,
+	onAddMapper,
+	onEditMapper,
+}: GroupMappersProps): JSX.Element {
+	const { hydrateGroupMappers, removeMapper, toggleMapper } = store;
```
aks07: Why are we prop drilling the store around?

aks07:
As discussed with @nityanandagohain 
The API is intentionally per entity. Each group/mapper change sends its own PATCH call so that every entity keeps its own change history. A bulk endpoint would save the whole config as one blob under a single author (like log pipelines), which the backend wants to avoid. 
The frontend can't persist per action either. Users stage changes in a local draft and test them via the `Test` tab before publishing.

**Issue:** 
`Save Changes` sends everything at once. So a save fans out into multiple sequential requests. If one fails midway, the server holds a partial subset of the intended changes, and the UI can't reliably report what was applied and what wasn't.

frontend/src/container/LLMObservability/AttributeMapping/AttributeMappingsTab/__tests__/AttributeMappingsTab.test.tsx
```diff
@@ -39,6 +51,62 @@ async function expandGroup(
 	await user.click(screen.getByTestId(`group-expand-${groupId}`));
 }
 
+async function openGroupActionsMenu(
+	user: ReturnType<typeof userEvent.setup>,
+	groupId: string,
+): Promise<void> {
+	const row = screen
+		.getByTestId(`group-name-${groupId}`)
+		.closest('.ant-collapse-item') as HTMLElement;
```
aks07: Lets not use antd classes in test. use testId or data-testId

frontend/src/container/LLMObservability/Settings/ModelPricing/ModelCostTabPanel/components/ModelCostsTable/ModelCostsTable.module.scss
```diff
@@ -7,6 +7,11 @@
 	:global(table) tbody tr {
 		cursor: default;
 	}
+
+	// Column headings use the l2 foreground token.
```
aks07: remove unnecessary comments

tewarig:
> image
> Failed and empty state looking same. Lets atleast add a caution icon or somethin

updated

image

tewarig:
> image
> Shouldnt this be attribute instead of span attribute?
update span attribute - 

image

frontend/src/container/LLMObservability/AttributeMapping/AttributeMappingsTab/components/MappingsTable/GroupMappers/GroupMappers.tsx
tewarig: 🤔 i think we should use zustand store here ... the store is given by a hook itself

aks07 (approved):

PR #11745, opened 2026-06-16
feat(metrics-explorer): move metric_name from path param to query param

nikhilmantri0902:
## Pull Request

  ---

  ### 📄 Summary

  Metrics-explorer endpoints took `metric_name` as a URL **path** param. In gorilla/mux an unconstrained `{metric_name}` matches `[^/]+`, so it can't span slashes. Metric names containing `/` (GCP/AWS/Azure, e.g. `run.googleapis.com/request_latencies.bucket`) added extra path segments, failed to match, fell through to the SPA `PathPrefix("/")` fallback, and returned HTTP 200 + `index.html` instead of JSON — a silent failure across the metric details flow for cloud metrics.

  This moves `metric_name` off the path for all six routes: `metricName` becomes a required **query** param on the 5 GETs, and the POST reads `metricName` from the JSON body (it already carried it). Query/body values are URL-encoded by the client, so slashes are handled correctly.

  #### Issues closed by this PR

  Closes https://github.com/SigNoz/engineering-pod/issues/4820
  Refer to [this comment](https://github.com/SigNoz/engineering-pod/issues/4820#issuecomment-4719978714) to know which APIs were refactored. 

  ---

  ### ✅ Change Type

  - [ ] ✨ Feature
  - [x] 🐛 Bug fix
  - [x] ♻️ Refactor
  - [ ] 🛠️ Infra / Tooling
  - [ ] 🧪 Test-only

  ---

  ### 🐛 Bug Context

  #### Root Cause
  gorilla/mux path var `{metric_name}` is `[^/]+` and cannot match slashes. Metric names with `/` produced unmatched routes that silently fell through to the SPA fallback (200 + HTML).

  #### Fix Strategy
  Drop the `{metric_name}` path segment from all six routes. The 5 GETs (`attributes`, `metadata`, `highlights`, `alerts`, `dashboards`) take a required `metricName` query param via a shared `MetricNameQuery` type; POST `/metadata` reads `metricName` from the existing request body. Regenerated the OpenAPI spec.

  ---

  ### 🧪 Testing Strategy

  - Tests added/updated: existing `pkg/modules/metricsexplorer` + `pkg/apiserver` suites pass; OpenAPI spec regenerated and verified.
  - Manual verification: `GET /api/v2/metrics/metadata?metricName=run.googleapis.com/request_latencies.bucket` → 200 + JSON (was HTML).
[...]

PR #11784, opened 2026-06-18
feat: add api to fetch v2 dashboards for a metric name

namanv3:
## Pull Request

---

### 📄 Summary

The current API that fetches dashboards for a metric name cannot handle perses dashboards, that's why another version of that API is needed.

While we're at it, this new API improves on the previous version by actually filtering out most of the dashboards by searching the data column for the metric name. The resultant candidates still need to be filtered in the application code however.

More complex ideas like having a metric -> dashboards relations table were discarded in favour of simplicity.

#### Issues closed by this PR

Closes https://github.com/SigNoz/pulse-pod/issues/63

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Integration tests added
- Manual verification: Manually tested the API
- Edge cases covered: Metric name present but not actually queried, metric name queried multiple times in a dashboard

---

- Blast radius: Just this API
- Potential regressions: None
- Rollback plan: Revert PR

---


docs/api/openapi.yml
```diff
@@ -21476,6 +21476,71 @@ paths:
       summary: Put profile in Zeus for a deployment.
       tags:
       - zeus
+  /api/v3/metrics/{metric_name}/dashboards:
```
namanv3: current API is `/api/v2/metics...`, so this new one becomes `v3`. Other options were keeping v2 and changing `dashboards` to `dashboard_perses`, but that didn't feel like a good idea.

therealpandey: @namanv3 does this have the same problems as the metrics explorer apis are having with path params?

therealpandey (changes requested):

docs/api/openapi.yml
namanv3: Yes actually. Will check with @nikhilmantri0902 on this.

namanv3: Moved to query param as it is being done in https://github.com/SigNoz/signoz/pull/11745

pkg/modules/dashboard/impldashboard/v2_dashboards_by_metric_name.go
```diff
@@ -0,0 +1,138 @@
...
+	"log/slog"
+	"maps"
+
+	"github.com/SigNoz/signoz/pkg/errors"
+	"github.com/SigNoz/signoz/pkg/types/dashboardtypes"
+	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
+	"github.com/SigNoz/signoz/pkg/valuer"
+)
+
+func (m *module) GetByMetricNamesV2(ctx context.Context, orgID valuer.UUID, metricNames []string) (map[string][]map[string]string, error) {
```
srikanthccv: please return a typed value from here instead of `map[string][]map[string]string`. It's hard to follow what does the response contain without looking at the implementation.

pkg/modules/dashboard/impldashboard/store.go
```diff
@@ -213,6 +213,26 @@ func (store *store) sortExprForListV2(sort dashboardtypes.ListSort) (string, err
 		"unsupported sort field %q", sort)
 }
 
+func (store *store) ListByDataContains(ctx context.Context, orgID valuer.UUID, search string) ([]*dashboardtypes.StorableDashboard, error) {
+	// LIKE wildcards (`%`, `_`, `\`) in the search are escaped so it matches literally.
+	escaped := strings.NewReplacer(`\`, `\\`, `%`, `\%`, `_`, `\_`).Replace(search)
```
srikanthccv: is there a escaper part of the `pkg/sql*` packages?

pkg/modules/dashboard/impldashboard/v2_dashboards_by_metric_name.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	return m.selectDashboardsFromCandidates(ctx, metricNamesMap, candidateDashboards), nil
+}
+
+func (m *module) getCandidatesDashboardsForMetricNames(ctx context.Context, orgID valuer.UUID, metricNames []string) ([]*dashboardtypes.DashboardV2, error) {
+	candidateIDs := make(map[valuer.UUID]bool)
+	candidates := make([]*dashboardtypes.DashboardV2, 0)
+	for _, name := range metricNames {
+		storables, err := m.store.ListByDataContains(ctx, orgID, name)
```
srikanthccv: this could be single db call instead of one round trip for each metric name.

pkg/modules/dashboard/impldashboard/v2_dashboards_by_metric_name.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	return found
+}
+
+func (m *module) extractMetricNamesFromRawQuery(ctx context.Context, queryType qbtypes.QueryType, query string) map[string]bool {
+	found := make(map[string]bool)
+	if query == "" {
+		return found
+	}
+	result, err := m.queryParser.AnalyzeQueryFilter(ctx, queryType, query)
```
srikanthccv, line 123: have seen this panic in the past from the clickhouse parser library, so panic guard around this.

namanv3: Panic handling is already there in the impl, do we still need it? No other caller seems to be adding a guard either
https://github.com/SigNoz/signoz/blob/e42159f2574f5d50704841d6b9f3e347cf932de7/pkg/queryparser/queryparserimpl.go#L27-L34

pkg/modules/dashboard/impldashboard/v2_dashboards_by_metric_name.go
namanv3: done

pkg/modules/dashboard/impldashboard/store.go
namanv3: it wasn't, added it now

pkg/modules/dashboard/impldashboard/v2_dashboards_by_metric_name.go
namanv3: Done. Also changed the return type of the API accordingly

pkg/types/dashboardtypes/dashboards_metric_usage.go
```diff
@@ -0,0 +1,13 @@
...
+// "dashboards by metric name" lookup returns these to report each panel that
+// references a given metric.
+type DashboardPanelRef struct {
+	DashboardID   string `json:"dashboardId" required:"true"`
+	DashboardName string `json:"dashboardName" required:"true"`
+	PanelID       string `json:"panelId" required:"true"`
+	PanelName     string `json:"panelName" required:"true"`
+}
+
+type DashboardPanelReferences []DashboardPanelRef
```
srikanthccv: is this aliased type helping with anything? can you directly use `[]DashboardPanelRef` wherever needed instead?

image

you can see how this is reflected in openapi spec and frontend types. The type alias seems unnecessary.

namanv3: felt like it made the go code easier to read, but I'm okay either way. Reverted.

namanv3: I've still kept nullable:true to match the previous API response.

srikanthccv (approved):
LGTM

therealpandey (approved):

PR #11788, opened 2026-06-18
feat(quick-filters): add support for /v1/fields/values

H4ad:
## Pull Request

---

This modify the QuickFilter to add a V2 of Checkbox to use fields/values API instead of autocomplete.

This does not add full support for fields API, the `/v1/fields/key` still missing but it's not used by the Infra Monitoring, so I kept this PR focused for Infra Monitoring.

About the states, we will have:
- no section - when we are listing all values and we have no filter
- related - when the existingQuery is present, we can show the `related` as section
- all values - when the existingQuery is present, we classify `all values` anything that will cause the filter to return no values (with current filter, none of the pods has that namespace/node/environment/etc...)

https://github.com/user-attachments/assets/41e86fd1-566c-47ce-8b0f-79cdf4727a76

Closes https://github.com/SigNoz/engineering-pod/issues/4005

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered: Yes

---

- Blast radius: Quick Filters
- Potential regressions: Filter breaking or wrong suggestions.
- Rollback plan: Find and fix the issue instead of reverting this PR.

---

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | N/A |

---

### 📋 Checklist
- [x] Manually tested


aks07:
> About the states of this, we will have:
> 
> no badge - when we are listing all values and we have no filter
> related - when the existingQuery is present, we can show the related values to the filter
> other - when the existingQuery is present, we classify other anything that will cause the filter to return no values (with current filter, none of the pods has that namespace/node/environment/etc...)


Would it not be better to have separators(All values and related values)? As you suggested here: https://github.com/SigNoz/engineering-pod/issues/4005#issuecomment-4731483994

Or operate with 2 states: related badge or no badge

3 states makes sense from dev perspective but as a user i feel only showing the related badge suffices. user knows this value is related badge shows up next to the value; if no badge its not related

Also, re-using the badge for Toggle/Only is okay but `NotIn` in the same feels confusing again. Do take a designers perspective on this.

aks07:
https://github.com/user-attachments/assets/877069d1-891e-490c-b059-21593215988c

A little hard to understand what is happening here. Why does on load i see the related value but then it goes away?


Also, whats that half tick?

H4ad:
@aks07 

> Would it not be better to have separators(All values and related values)? As you suggested here: https://github.com/SigNoz/engineering-pod/issues/4005#issuecomment-4731483994

Because I wanted to reduce the amount of layout shift caused by the sections, by using the badges, I can keep a flat list without any layout shift (except on search) due to showing related/other by using the badge.

> if no badge its not related

Hm... I need to give it a try but I don't mind changing to be no badge instead of `Other`.

> NotIn in the same feels confusing again

Because of the change to be a flat list, I wanted to have a way to differentiate when the value is `not in` selected in the filter, without the badge, the UI looks weird.

---

About the last comment, this happens because the selectedItem is reflected immediately while the API with the data if the value is related or not still staled

So, it thinks it's related (that's way old value because related) but it will disappear once the loading is finished.

The [/] happens because instead of showing `Checked` or `Unchecked`, I also added support to show `Indeterminated` as a state, when the value is related, not selected, it will show as `Indeterminated` in a way of saying the value is selected but not explicitly.

H4ad:
@aks07 removed the "other" label, and also changed the behaviour of `toggle` the `unchecked` checkbox, instead of going to `not in`, it will turn into `checked` and `in`, same when in the state of `indeterminate`.

H4ad:
@aks07 Updated the implementation to follow suggestions/discussion with Ashwin/Nikhil, see more: https://github.com/SigNoz/engineering-pod/issues/4005#issuecomment-4896560660

frontend/src/components/QuickFilters/FilterRenderers/Checkbox/v2/CheckboxFilterV2ValueRow.tsx
```diff
@@ -0,0 +1,118 @@
...
+					color="primary"
+					testId={`checkbox-${title}-${value}`}
+				/>
+			</div>
+
+			<div
+				role="button"
+				tabIndex={disabled ? -1 : 0}
+				className={cx(styles.valueButton, disabled && styles.isDisabled)}
+				onClick={(): void => {
```
AshwinBhatkal, line 71: Why not Button? 
If check can be avoided

frontend/src/components/QuickFilters/FilterRenderers/Checkbox/v2/CheckboxFilterV2ValueRow.module.scss
```diff
@@ -0,0 +1,166 @@
...
+
+	// Stack badge / only / toggle in a single cell so the crossfade overlaps
+	// instead of laying them side-by-side mid-transition.
+	> * {
+		grid-area: 1 / 1;
+	}
+}
+
+.badge {
+	display: inline-flex;
```
AshwinBhatkal: Is this required?

frontend/src/components/QuickFilters/FilterRenderers/Checkbox/v2/CheckboxFilterV2.tsx
```diff
@@ -171,54 +171,46 @@ export default function CheckboxFilterV2(
 						/>
 					</section>
 
-					{totalCount > 0 && (
+					{(totalCount > 0 || searchText) && (
```
AshwinBhatkal: this condition seems off. Can you test behaviour around this once

frontend/src/components/QuickFilters/FilterRenderers/Checkbox/v2/__tests__/CheckboxFilterV2.interactions.test.tsx
```diff
@@ -157,7 +157,7 @@ describe('CheckboxFilterV2 - interactions', () => {
 			});
```
AshwinBhatkal, line 157: Let's atleast do basic organization. Tests go inisde the tests folder

frontend/src/components/QuickFilters/FilterRenderers/Checkbox/v2/CheckboxFilterV2ValueRow.module.scss
H4ad: No, my initial implementation for this was not using the Badge component

frontend/src/components/QuickFilters/FilterRenderers/Checkbox/v2/CheckboxFilterV2.tsx
H4ad: One of the implementations rendered the searchText section customized with loading/empty states

Since we removed in favor of section (related/all values) and no section for search, then this remained.

But with current logic, totalCount === 0 means sections.length === 0, I can remove to not render this empty div.

frontend/src/components/QuickFilters/FilterRenderers/Checkbox/v2/CheckboxFilterV2ValueRow.tsx
H4ad: This wraps the text and only/all badge, 
image

This is not a button but neither could be a div, having the button here just cause more problems because I need to add CSS to reset the button styling, not worth to just have the semantic html

AshwinBhatkal (approved):

PR #11793, opened 2026-06-19
feat(spanmapper): add preview endpoint for span attribute mapping

nityanandagohain:
## What

Adds `POST /api/v1/span_mapper_groups/preview` — runs a sample span through an ordered list of mapping groups and returns the transformed `attributes` and `resource` maps, so the UI can show a live before/after while editing.

## Payload design

```jsonc
{
  "span": {
    "attributes": { "llm.model": "gpt-4" },
    "resource":   { "service.name": "svc" }
  },
  "groups": [
    { "group": { "name": "llm", "condition": {...}, "enabled": true } },              // mappers omitted -> load saved mappers by name
    { "group": { "name": "http", ... }, "mappers": [ /* edited */ ] }                  // mappers present -> authoritative
  ]
}
```

- **Every group always carries its meta** (`name`, `condition`, `enabled`).
- **`mappers` is optional per group.** Omitted (`null`) means "unchanged" — the API loads that group's saved mappers **by name**. Present (including `[]`) means the user is editing — the payload wins. The `*[]` pointer is what lets us tell "omitted" apart from "emptied".
- **Groups apply in the order they appear** in the request (no group-level priority column exists; source-level `Priority` still governs source selection within a mapper).
- **`span` is typed** (`{attributes, resource}`) instead of an opaque object, so the contract is documented in the spec.

This replaces the earlier tri-modal request (`groups` | `groupId` | implicit "all enabled"), which couldn't be expressed cleanly in OpenAPI.

## Notes / caveats

- Name-keyed backfill is safe **as long as the FE sends a group's mappers whenever anything in that group changes (including a rename)** — so name lookup only ever runs for fully-untouched groups, whose names still match the DB. If `mappers` is omitted but no saved group matches that name, the API returns a clear 400.

## Tests

- Unit tests for `PreviewMapping`: payload mappers used, omitted mappers backfilled by name, unknown-name error, and group apply-order (last writer wins).
- Existing `SimulateMappingForAttributes` tests unchanged.
- `docs/api/openapi.yml` and the generated frontend client regenerated.

### DIFF
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -7240,6 +7240,49 @@ components:
       - operation
       - priority
       type: object
+    SpantypesSpanMappingPreviewGroup:
+      properties:
+        group:
+          $ref: '#/components/schemas/SpantypesPostableSpanMapperGroup'
+        mappers:
+          items:
+            $ref: '#/components/schemas/SpantypesPostableSpanMapper'
+          nullable: true
+          type: array
+      required:
+      - group
+      type: object
+    SpantypesSpanMappingPreviewRequest:
+      properties:
+        groups:
+          items:
+            $ref: '#/components/schemas/SpantypesSpanMappingPreviewGroup'
+          nullable: true
+          type: array
+        span:
+          $ref: '#/components/schemas/SpantypesSpanMappingPreviewSpan'
+      required:
+      - span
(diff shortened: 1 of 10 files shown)


nityanandagohain:
ignore this PR 🙈

PR #11800, opened 2026-06-21
fix(querier): pad clamped time range for trace_id-filtered logs

### LINKED ISSUE #11709: Logs filtered by trace_id are silently dropped when they occur after the trace's recorded end time
## Bug description

Filtering logs by `trace_id` in Logs Explorer returns fewer logs than the trace actually has. Lines that carry the correct `trace_id` but whose `timestamp` is **after the trace's recorded end** are silently dropped, no matter what time range is selected.

This looks like a side effect of the trace-based time-range optimization for log queries (#8965 / #11394): a `trace_id` filter clamps the query window to the trace's recorded `[start, end]` instead of the selected range. Logs emitted after the last span closes (or a reused/long-lived `trace_id`) then fall outside the window.

Example — trace `3137b1a3f67ba94d8575b41b183d080b`: logs run to `16:27:16` (UTC+8) but everything after `16:27:03` is missing. The UI even shows the clamped window in a warning:

> Query A references a trace_id that exists between 2026-06-13T03:59:59Z and **2026-06-13T08:27:03Z** (UTC) but lies outside the selected time range; adjust the time range to see results.

`08:27:03Z` = `16:27:03 (+8)`, ~13s before the trace's actual last log. (Front-end screenshots available.)

## Expected behavior

All logs carrying the given `trace_id` within the selected time range should be returned. The trace-window optimization must never be tighter than the matching data, i.e. must not drop in-range logs that occur after the recorded span end.

## How to reproduce

1. Have a trace with logs after its recorded span end (e.g. completion/response logs after the root span closes, or a reused `trace_id`).
2. Logs Explorer → filter `trace_id = '<id>'`, pick a range that fully covers the trace. List View tops out at the recorded end (newest lines missing); Table `count()` is short.

Image

3. Same filter direct in ClickHouse returns more rows: in our case ClickHouse `count()` = **30145** vs SigNoz **30104** (41 missing). `system.query_log` shows the generated SQL using `timestamp < '1781339223960000000'` (= `16:27:03.960` +8) regardless of the selected end.
```
-- step1: get origin sql
SELECT
    event_time,
    query
FROM system.query_log
WHERE (query LIKE '%3137b1a3f67ba94d8575b41b183d080b%') AND (query LIKE '%distributed_logs_v2%') AND (query NOT LIKE '%system.query_log%') AND (type = 'QueryFinish') AND (event_time > (now() - toIntervalMinute(1)))
ORDER BY event_time DESC
LIMIT 10
FORMAT Vertical

Query id: 798a6908-f3f8-4865-b17d-9ae9cbf70537

Row 1:
──────
event_time: 2026-06-14 13:32:48
query:      SELECT count() AS __result_0 FROM signoz_logs.distributed_logs_v2 WHERE trace_id = '3137b1a3f67ba94d8575b41b183d080b' AND timestamp >= '1781323199585000000' AND ts_bucket_start >= 1781321399 AND timestamp < '1781339223960000000' AND ts_bucket_start <= 1781339223 ORDER BY __result_0 DESC

1 row in set. Elapsed: 0.010 sec. Processed 135.03 thousand rows, 1.80 MB (13.72 million rows/s., 182.92 MB/s.)
Peak memory usage: 4.24 MiB.

-- step2: query origin sql
SELECT count() AS __result_0
FROM signoz_logs.distributed_logs_v2
WHERE (trace_id = '3137b1a3f67ba94d8575b41b183d080b') AND (timestamp >= '1781323199585000000') AND (ts_bucket_start >= 1781321399) AND (timestamp < '1781339223960000000') AND (ts_bucket_start <= 1781339223)
ORDER BY __result_0 DESC

Query id: c5500473-2f1e-477d-9b2f-b0d691a06e46

   ┌─__result_0─┐
1. │      30104 │
   └────────────┘

1 row in set. Elapsed: 0.312 sec. Processed 1.73 million rows, 70.13 MB (5.55 million rows/s., 224.61 MB/s.)
Peak memory usage: 4.57 MiB.


-- step 3: query expect sql

SELECT count()
FROM signoz_logs.distributed_logs_v2
WHERE (trace_id = '3137b1a3f67ba94d8575b41b183d080b') AND (timestamp >= toUnixTimestamp64Nano(toDateTime64('2026-06-12 17:22:38', 8, 'Asia/Shanghai'))) AND (timestamp <= toUnixTimestamp64Nano(toDateTime64('2026-06-13 17:22:38', 8, 'Asia/Shanghai')))

Query id: cd4c06eb-bc07-4cbb-887e-5e3d0ed56f22

   ┌─count()─┐
1. │   30145 │
   └─────────┘

1 row in set. Elapsed: 0.263 sec. Processed 1.73 million rows, 56.38 MB (6.58 million rows/s., 214.18 MB/s.)
Peak memory usage: 4.42 MiB.
```

## Version information

* **Signoz version**: v0.128.x (ClickHouse 25.5.6)
* **Browser version**: edge 149.0.4022.52

## Additional context

The dropped rows carry the intrinsic `trace_id` (not a parse/missing issue). Running SigNoz's own generated SQL directly in ClickHouse reproduces 30104, confirming the clamped upper bound is the sole cause.

Possible fixes: pad the optimized upper bound (or use `max(trace end, max log ts)`); only narrow when no logs exist beyond the trace window, else fall back to the selected range; or add a flag to disable this optimization for log queries.

Related: #8965, #11394.

nityanandagohain:
## Pull Request

---

### 📄 Summary
The time-range for trace_id filtered logs was clamped to the min and max duration recorded in traces, but in reality there can be some delay after which the logs are recorded/exported

#### Issues closed by this PR
Fixes https://github.com/SigNoz/signoz/issues/11709

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context
The time-range for trace_id filtered logs was clamped to the min and max duration recorded in traces, but in reality there can be some delay after which the logs are recorded/exported

#### Fix Strategy
Added a default padding of 5 mins, which can be overriden using config

Why +- 5 mins?
- Large enough to absorb realistic clock skew plus normal post-span logging across real deployments — seconds would be too tight for fleets with imperfect time sync. ( considering network retries as well)
- Small enough to preserve the optimization. The window-list scan buckets in ≥1 h chunks, so the trace_id narrowing only pays off when it keeps the effective window well under an hour. 5 min is far below that floor, so the scan stays tight.

---

- Tests added/updated: ✅ 
- Manual verification:✅ 
- Edge cases covered:

---

- Blast radius: Trace id filtering
- Potential regressions:
- Rollback plan: revert

---

### 📋 Checklist
- [x] Manually tested


pkg/querier/builder_query.go
```diff
@@ -43,16 +45,18 @@ func newBuilderQuery[T any](
...
-		kind:           kind,
+		logger:                    logger,
+		telemetryStore:            telemetryStore,
+		stmtBuilder:               stmtBuilder,
+		spec:                      spec,
+		variables:                 variables,
+		fromMS:                    tr.From,
+		toMS:                      tr.To,
+		kind:                      kind,
+		logTraceIDWindowPaddingMS: logTraceIDWindowPaddingMS,
```
tushar-signoz: nit: Can we create a builderConfig and logTraceIDWindowPaddingMS (this is very specific to a use-case) can be part of it ?

nityanandagohain: thanks, have made the change.

srikanthccv (commented):
Did you consider maintaining summary table for logs independently of traces and use the logs trace summary for the optimisation instead?

nityanandagohain:
yes @srikanthccv , but it felt like an overkill. But if we see issues where this is going way out of the 5 minutes, then maybe something we can consider.

srikanthccv:
Ok, cool. The reason I bring it up is, any padding strategy remains inherently fragile (you never know how people instrument their systems, for example, pass the trace context over the wire and attach it to the service that only emits logs but those logs can much later than 5m). The reliable solution is to generate summaries off of trace_id and span_id present in the log data separately and use.

nityanandagohain:
Yeah makes sense, let me keep an internal issue for that and we can pick them up in the future.

srikanthccv (approved):
More discussion here https://signoz-team.slack.com/archives/C08P77ZDHV0/p1782111433560959

therealpandey (approved):

PR #11801, opened 2026-06-21
chore(querier): route metric reads to buffer/reduced tables behind a …

srikanthccv:
…flag

### 📄 Summary

A metric volume control rule can be created/edited/deleted anytime. This means the data is always going to be split b/w the `samples_v4`, and `{}_reduced_{last_60s/sum_60s}` table. To ensure the complete data is severed, there were two options  at high level 1. single query to fetch all data 2. independent queries, one for each. With the option #2, the whole querier pipeline needs to change, which takes away the flow querier has today. And given that, the rule reduction is not a frequent option, for a give time range the change of query reading from two tables is relatively less compared to single table. With this in mind, the UNION ALL is chosen despite it executing queries sequentially, If it becomes the bottleneck, then we will revisit the querier orchestration.

#### Issues closed by this PR

Fixes https://github.com/SigNoz/engineering-pod/issues/5341

### ✅ Change Type

- [x] ♻️ Refactor

### 🧪 Testing Strategy

- Tests added/updated: yes
- Manual verification: yes
- Edge cases covered: yes

---

- Blast radius: metrics querier affected
- Potential regressions: nan
- Rollback plan: revert

### 📋 Checklist
- [x] Manually tested

### Notes

- will integration tests when collector PR lands


therealpandey (approved):

srikanthccv:
merging, @namanv3 you can continue review when you are back.

PR #11802, opened 2026-06-21
feat(user): accept custom roles in user invite

vikrantgupta25:
### 📄 Summary

- mark the existing bulk invite and invite API as deprecated. 
- added `/v2/users` endpoint with custom role support. 

contributes to: https://github.com/SigNoz/platform-pod/issues/2279


pkg/types/authtypes/user_role.go
```diff
@@ -28,6 +30,38 @@ type UserRole struct {
...
 
+type UserWithRoles struct {
+	*types.User
+	UserRoles []*UserRole `json:"userRoles"`
+}
+
+type PostableUser struct {
+	DisplayName     string              `json:"displayName"`
+	Email           valuer.Email        `json:"email" required:"true"`
+	FrontendBaseUrl string              `json:"frontendBaseUrl"`
```
vikrantgupta25: Wanted to not use this but use `url` from global config but that hasn't been enforced yet.

pkg/types/authtypes/user_role.go
```diff
@@ -28,6 +30,38 @@ type UserRole struct {
...
+type UserWithRoles struct {
+	*types.User
+	UserRoles []*UserRole `json:"userRoles"`
+}
+
+type PostableUser struct {
+	DisplayName     string             `json:"displayName"`
+	Email           valuer.Email       `json:"email" required:"true"`
+	FrontendBaseUrl string             `json:"frontendBaseUrl"`
+	UserRoles       []PostableUserRole `json:"userRoles" required:"true" nullable:"false"`
```
therealpandey: Why not pointer to array here?

therealpandey (changes requested):

pkg/types/authtypes/user_role.go
vikrantgupta25: updated

therealpandey (approved):

PR #11812, opened 2026-06-22
fix(metrics-explorer): resolve pagination overlap in table

### LINKED ISSUE #11807: [UI]: Pagination UI in Metrics Explorer Hides The Last Entry
## Bug description

See this image:

Image

You might not see in the image, but the pagination ui at the bottom is hiding one metric. In the image, the last entry is **otelcol_receiver_failed_metric_points**, but if you see the image below which i took by scrolling more than needed, you will see the last one is otelcol_receiver_refused_metric_points

Image

## Expected behavior

Expected behavior is simple. The pagination UI shouldn't interfere with rows being displayed. 

## How to reproduce

1. Open metrics
2. Try what I said above, you will understand

## Version information 
* **Signoz version**:  Free v0.129.0
* **Browser version**: Zen 1.19.13b (Firefox 150.0.3) (64-bit)
* **Your OS and version**: OpenSUSE Tumbleweed

## Additional context

Same as for the other issue I opened. I can solve it if you don't have time, otherwise all yours :)

ElyesLaaribi:
## Pull Request

---

This PR fixes a UI bug in the Metrics Explorer where the pagination component at the bottom of the List View table would overlap the bottom rows of the table, making them unreadable.

Before:
image

After:
image

Closes #11807

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: N/A (CSS only)
- Manual verification: Ran the frontend locally against the community backend. Generated metrics and verified in the "Metrics Explorer" List View that the bottom rows (e.g., `otelcol_exporter_queue_capacity`) are fully visible and sit above the pagination bar.
- Edge cases covered: Verified that the pagination still stays at the bottom when there are few rows vs many rows.

---

- Blast radius: Very small, only affects the metrics table in the Metrics Explorer Summary tab.
- Potential regressions: Pagination layout breaking on extremely small screens (mitigated by setting width to 100%).
- Rollback plan: Revert the CSS change in `Summary.styles.scss`.

---

| Field | Value |
|------|-------|
| Deployment Type | OSS |
| Change Type | Bug Fix |
| Description | Fixed a UI overlap issue in the Metrics Explorer where the table pagination obscured the bottom rows. |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/container/MetricsExplorer/Summary/Summary.styles.scss
+++ b/frontend/src/container/MetricsExplorer/Summary/Summary.styles.scss
@@ -58,6 +58,7 @@
 	}
 
 	.metrics-table-container {
+		padding-bottom: 48px;
 		.ant-table {
 			margin-left: -16px;
 			margin-right: -16px;
@@ -182,13 +183,15 @@
 		}
 
 		.ant-pagination {
-			position: sticky;
+			position: fixed;
 			bottom: 0;
-			width: 100%;
+			width: calc(100% - 54px);
 			background: var(--l1-background);
 			padding: 16px;
 			margin: 0;
-			z-index: 10;
+
+			// this is to offset chat support icon till we improve the design
+			right: 20px;
 
 			.ant-pagination-item {
 				border-radius: 4px;
--- a/frontend/src/container/MetricsExplorer/Summary/Summary.styles.scss
+++ b/frontend/src/container/MetricsExplorer/Summary/Summary.styles.scss
@@ -182,15 +182,13 @@
 		}
 
 		.ant-pagination {
-			position: fixed;
+			position: sticky;
 			bottom: 0;
-			width: calc(100% - 54px);
+			width: 100%;
 			background: var(--l1-background);
 			padding: 16px;
 			margin: 0;
-
-			// this is to offset chat support icon till we improve the design
-			right: 20px;
+			z-index: 10;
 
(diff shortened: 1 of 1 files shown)


frontend/src/container/MetricsExplorer/Summary/Summary.styles.scss
H4ad, line 60: Instead of changing the pagination component to not be fixed on bottom, you can add `padding-bottom: 48px` here.

It will be one-line fix and will still maintain the pagination in the bottom.

H4ad (commented):
Thanks for the PR, I just left a comment around the fix, we still want to keep the pagination fixed in the bottom.

frontend/src/container/MetricsExplorer/Summary/Summary.styles.scss
ElyesLaaribi: Thanks for the suggestion. I've updated the fix to keep the pagination fixed at the bottom and added the bottom padding to `.metrics-table-container` instead.

H4ad (approved):
Thanks for the bug fix 🚀

YounixM (approved):

PR #11819, opened 2026-06-23
feat(llm-attribute-mapping): add Test tab with sample-span runner

tewarig:
## Pull Request

---

### 📄 Summary

Adds the **Test** tab to the LLM Observability → Attribute Mapping page, replacing the previous `Coming soon` placeholder.

Users can paste a sample JSON span and run it through their configured span mappers to preview the outcome **before saving** — seeing which target attributes get populated and which source key matched. This closes the authoring loop: today you configure mappers blind and only find out the result after saving and waiting for live spans.

Key behaviour:

- **Dry-run against unsaved edits.** The payload sends every group currently in the editor draft. Per-group `mappers` are sent only when they differ from the last-saved server snapshot; unchanged groups send `null`, letting the backend reuse its stored mappers. This lets you test in-progress edits without persisting them, while keeping the request minimal.
- **Monaco-based JSON editor** with a SigNoz light/dark theme, a prefilled sample span, inline JSON validation (Run is disabled while the input is invalid), and a skeleton loader.
- **Result diff view** — each returned span is rendered against the input, labelling attributes as added / changed / unchanged / removed so the mapping effect is obvious at a glance.

https://github.com/user-attachments/assets/f8aae6cc-e8c4-4497-b119-699c047f7b97

with viewer  account - 

https://github.com/user-attachments/assets/0e2a8acf-3bcd-466f-8ced-1fb9f2b4c173

 

#### Issues closed by this PR

Close - https://github.com/orgs/SigNoz/projects/39/views/11?filterQuery=assignee%3Atewarig&pane=issue&itemId=203322763&issue=SigNoz%7Cengineering-pod%7C5456

---

### ✅ Change Type

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🧪 Testing Strategy
 Test will be added in follow up with e2e for test tab
---

- **Blast radius:** Scoped to the LLM Observability → Attribute Mapping page. The only cross-cutting change is exposing `snapshot` from `useAttributeMappingEditor` (additive) and a new `useCanManageAttributeMapping` hook.
- **Potential regressions:** Low. Additive tab; existing Mappings tab behaviour unchanged aside from minor supporting tweaks.
- **Rollback plan:** Revert the PR; the Test tab reverts to the `Coming soon` disabled state.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Test attribute-mapping rules against a sample span directly in the UI before saving. |

---

### 📋 Checklist

- [x] Manually tested
- [x] Backward compatibility considered (additive tab + editor field)

---

## 👀 Notes for Reviewers

- We have added Authz module as well.


frontend/src/container/LLMObservability/AttributeMapping/hooks/useAttributeMappingEditor.ts
```diff
@@ -304,6 +305,7 @@ export function useAttributeMappingEditor(): AttributeMappingEditor {
 
 	return {
 		groups: draft ?? [],
+		snapshot,
```
aks07, line 308: Whats this change?

frontend/src/container/LLMObservability/AttributeMapping/TestTab/testPayload.ts
aks07, line 1: Have we added test for these? it seems this is the core logic and should be tested

frontend/src/container/LLMObservability/AttributeMapping/hooks/useCanManageAttributeMapping.ts
```diff
@@ -0,0 +1,9 @@
+import { IsAdminPermission } from 'lib/authz/hooks/useAuthZ/legacy';
+import { useAuthZ } from 'lib/authz/hooks/useAuthZ/useAuthZ';
+
+const ADMIN_PERMISSION = [IsAdminPermission];
+
+export function useCanManageAttributeMapping(): boolean {
+	const { allowed } = useAuthZ(ADMIN_PERMISSION);
```
aks07, line 7: While authZ is loading allowed is false. Ensure that this is not causing a bad UX. Wherever we have used canManage directly in the flow

frontend/src/container/LLMObservability/AttributeMapping/TestTab/useTestSpanMapper.ts
```diff
@@ -0,0 +1,127 @@
...
+import { useTestSpanMappers } from 'api/generated/services/spanmapper';
+import { AxiosError } from 'axios';
+
+import { buildTestRequest, parseSpanInput } from './testPayload';
+import { DraftGroup } from '../types';
+
+export type TestTabAttributes = Record<string, unknown>;
+export type TestTabResource = Record<string, unknown>;
+
+export const SAMPLE_SPAN_JSON = `{
```
aks07: The SAMPLE_SPAN_JSON assumes these keys exists in users attribute mapping. Because if it does not, and for the most cases if wont. This will be of no use.
Consider creating this sample from user added attribute mappings. so it makes sense to the user.

aks07: Also, Test tab can be kept hidden if there is no mapping. (just thinking out loud here This one you can take a call on)

aks07:
https://github.com/user-attachments/assets/736fbeac-15ac-4316-83c2-67e1e5bccbc2



The editor state should persist if we're expecting users to test their mappings before saving, which isn't happening here.

I assume this isn't the intended behaviour. If a user switches tabs to check their changes, the editor state is lost.

cc. @nityanandagohain

tewarig:
> The editor state should persist if we're expecting users to test their mappings before saving, which isn't happening here.
> 
> I assume this isn't the intended behaviour. If a user switches tabs to check their changes, the editor state is lost.

🤔  it should be persisted. will update this

frontend/src/container/LLMObservability/AttributeMapping/TestTab/testPayload.ts
tewarig: https://github.com/SigNoz/signoz/pull/12235/changes 
Already have created PR for Testing this logic.

frontend/src/container/LLMObservability/AttributeMapping/TestTab/useTestSpanMapper.ts
tewarig: > The SAMPLE_SPAN_JSON assumes these keys exists in users attribute mapping. Because if it does not, and for the most cases if wont. This will be of no use.
> Consider creating this sample from user added attribute mappings. so it makes sense to the user.

No need for now. we need to fetch all data from span_mapper_groups and span_mappers.

tewarig: > Also, Test tab can be kept hidden if there is no mapping. (just thinking out loud here This one you can take a call on)

So we don't have to disable it. There will be mapping in each and every case, so by default, the backend will seed mapping in the users database. If there is no mapping, it is a conscious choice: the user has deleted it. I think this would be allowed to test how the result would look.

frontend/src/container/LLMObservability/AttributeMapping/hooks/useAttributeMappingEditor.ts
tewarig: snapshot was already in the editor hook (for dirty/discard/save) — we just expose it for Test.

Test sends all groups, but mappers: null when they match the snapshot so the backend keeps the stored ones. Needed because mappers are lazy-loaded — unexpanded groups are [] locally, and sending that would wipe the real mappers. Lets us test unsaved edits without saving first.

tewarig:
> test.tab.mov 
> The editor state should persist if we're expecting users to test their mappings before saving, which isn't happening here.
> 
> I assume this isn't the intended behaviour. If a user switches tabs to check their changes, the editor state is lost.
> 
> cc. @nityanandagohain

Updated Flow

https://github.com/user-attachments/assets/6f119049-6860-4ca3-a9e7-c39ee6d42a26

aks07 (approved):
I assume we are adding unit tests for the core test tab logic in an upcoming PR. Thus, Approving.

@tewarig Do add a note regarding the test tab sample data part we discussed

tewarig:
> I assume we are adding unit tests for the core test tab logic in an upcoming PR. Thus, Approving.
> 
> @tewarig Do add a note regarding the test tab sample data part we discussed

Sure, we will update the default JSON in frontend , once backend seeds the default group of span mappers  // @nityanandagohain

PR #11823, opened 2026-06-23
fix(dashboard): resolve contextLinks variable placeholders for per-ro…

### LINKED ISSUE #11325: (task) Dashboard contextLinks URL templates don't substitute per-row groupBy values on table panel aggregate-click

Abdul-Moiz31:
…w values

Custom (per-row) variables such as trace_id/span_id are stored internally with a '_' prefix to avoid colliding with dashboard/global variable names, but {{var}} and [[var]] templates were only looked up by their bare name. This left placeholders like {{trace_id}} unresolved on contextLinks URLs, so they were navigated to literally URL-encoded instead of substituted.

Fixes #11325

## Pull Request

---

### 📄 Summary
A dashboard table panel's `contextLinks` URL templates (e.g. `/trace/{{trace_id}}?spanId={{span_id}}`) were not substituting per-row group-by values like `trace_id`/`span_id`. The placeholders were left literal and got URL-encoded on navigation, producing links like `/trace/%7B%7Btrace_id%7D%7D` instead of a working trace link.

#### Issues closed by this PR

> Closes #11325

---

### Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause

Per-row custom variables (like `trace_id`/`span_id` passed from the table drilldown) are stored internally with a `_` prefix to avoid colliding with dashboard/global variable names:

`name: `_${name}`, // Add '_' prefix to avoid conflicts`

The {{var}} / [[var]] template resolver (resolveText / resolveTextWithTruncation in useContextVariables.tsx) only looked up the bare name, e.g. processedVariables['trace_id'], which never exists — only processedVariables['_trace_id'] does. The lookup returned undefined, so the placeholder was left untouched in the final URL.

Added a getVariableValue helper that tries the bare name first, then falls back to the _-prefixed name, and used it in both resolver functions. Dashboard/global variables (which aren't prefixed) continue to resolve exactly as before — this only adds a fallback lookup, it doesn't change any existing resolution path.

- Tests added: frontend/src/hooks/dashboard/__test__/useContextVariables.test.tsx
- {{var}} and [[var]] resolution against namespaced custom variables (the bug scenario)
- Regression check that bare dashboard/global variable resolution still works
- Unknown variables are left untouched
- Manual verification: pnpm jest useContextVariables (5/5 pass), pnpm jest useGetResolvedText (existing suite sharing the same resolver — 12/12 pass, no regressions)
- Edge cases covered: variable present only under bare name, only under prefixed name, and absent entirely

- Blast radius: limited to useContextVariables.tsx's text/URL template resolution, used by dashboard contextLinks and panel title/description variable substitution.
- Potential regressions: none expected — the change only adds a fallback lookup when the bare-name lookup misses; existing bare-name and dashboard/global variable resolution is untouched.
- Rollback plan: revert the single commit; no data/schema changes involved.

Field	Value
Deployment Type	Cloud / OSS
Change Type	Bug Fix
Description	Fixed dashboard contextLinks URL templates (e.g. {{trace_id}}) not substituting per-row group-by values on table panel drilldown links

### 📋 Checklist
- [ ]  Manually tested
- [ ]  Breaking changes documented (none)

## 👀 Notes for Reviewers

A previous attempt at this fix (#11326) took the same approach but went stale on inactivity, not on a maintainer objection — I picked it back up with the same core idea (fallback lookup with _ prefix) plus test coverage.


ahrefabhi:
Closing this PR as this issue is already fixed as part of [this](https://github.com/SigNoz/signoz/pull/12328) PR @Abdul-Moiz31.
Thanks for the contribution ❤️

PR #11815, opened 2026-06-22
feat(roles): rework the create/update pages

H4ad:
## Pull Request

---

This PR is a complete rework on the pages for create/edit/view role.

This removes the old modal to create/edit a role and replaced by an entire page that supports two modes, interactive via cards/toggle/input and JSON mode to paste the JSON with the permissions you want to add for that specific role.

The edit page uses the same component to edit, we just disable the edit of the name.

When clicking in the role, now we will show a page with a preview of the permissions of that role, this will be generated based on a config on UI to render the cards with the permissions.

When adding a new resource on backend, this will trigger an update on `permissions.type.ts` and it will give a compile time error if we didn't update the record `RESOURCE_PANELS` at `permissions.config.ts`.

Before:

https://github.com/user-attachments/assets/2633dd48-3afd-4ff1-a648-71ddfb2c14e6

---

After:

On create custom role:

https://github.com/user-attachments/assets/2218c112-0838-4606-86ef-45448353fc2b

On click to view a custom role:

https://github.com/user-attachments/assets/6fe3ae59-54ea-4716-bba7-631b7398aa3c

On click to view a managed role:

https://github.com/user-attachments/assets/986ef4f8-7ab5-4ed0-85c4-168837c7aba9

On edit custom role:

https://github.com/user-attachments/assets/d723c7e1-838e-479c-9ba3-29a6bc09f935

On delete custom role:

https://github.com/user-attachments/assets/70f69f4d-16ad-4f05-a0d3-eca9468654b7

On list roles, no permission to create:

screenshot-2026-06-24_14-59-31

On list roles, no permission to list:

screenshot-2026-06-24_14-55-42

On navigate to create role page, no permission to create:

screenshot-2026-06-24_14-58-57

On navigate to view role page, no permission to view:

screenshot-2026-06-24_14-56-03

On view, when no permission to update, but with permission to delete:

screenshot-2026-06-24_14-55-24

On view, when no permission to delete:

screenshot-2026-06-24_14-54-34

On view, when no permission to update:

image

On feature gate is disabled on list:

screenshot-2026-06-24_14-57-53

On feature gate is disabled on create:

screenshot-2026-06-24_14-57-43

On feature gate is disabled on edit:

screenshot-2026-06-24_14-57-27

On feature gate is disabled on view:

screenshot-2026-06-24_14-57-17

On failed to load role:

screenshot-2026-06-24_14-56-47

On failed to delete a role:

image

---

Some unrelated changes around Service Account asked by Vikrant.

[...]

PR #11827, opened 2026-06-23
feat(authz): add transaction groups JSON schema

vikrantgupta25:
### 📄 Summary

- add transaction groups JSON schema for editor consumption 
- add kind enums in openapi spec
- add detailed error messages for transaction groups validations 

contributes to: https://github.com/SigNoz/platform-pod/issues/2529


pkg/types/authtypes/transaction.go
```diff
@@ -156,3 +156,60 @@ func (groups TransactionGroups) selectorSet() map[string]struct{} {
...
+		selectors = append(selectors, selector)
+	}
+
+	return &TransactionGroup{
+		Relation:    Relation{Verb: verb},
+		ObjectGroup: coretypes.ObjectGroup{Resource: resourceRef, Selectors: selectors},
+	}, nil
+}
+
+func wrapValidationError(cause error, contextIdentifier string) error {
```
therealpandey: Seems too complicated. Just appending the error received from upstream into additional should be good enough.

cmd/genconfig.go
```diff
@@ -6,12 +6,15 @@ import (
...
 	"strings"
 
+	"github.com/SigNoz/signoz/pkg/types/authtypes"
 	"github.com/SigNoz/signoz/pkg/web"
 	"github.com/spf13/cobra"
 	"github.com/swaggest/jsonschema-go"
 )
 
-const webSettingsSchemaPath = "docs/config/web-settings.json"
+const webSettingsSchemaPath = "frontend/src/schemas/generated/web-settings.json"
```
therealpandey: How about putting them in generated/schemas so all generated resources are in a one place?

Is the orval cli not allowing it?

Also webSettings.schema.json

therealpandey (changes requested):

pkg/types/authtypes/transaction.go
vikrantgupta25: Now with that if we have additionals docs etc they do not get copied :/ followed the same pattern from querier

H4ad (dismissed):
The schema was tested on #11815, looks good

cmd/genconfig.go
vikrantgupta25: the current one is nested under `/api` inside frontend. Didn't understand by one place ?

therealpandey (approved):

H4ad (approved):

PR #11838, opened 2026-06-24
perf(querier): execute composite sub-queries in parallel behind a flag

kunalrawat425:
## Problem

A single `POST /api/v5/query_range` request can carry several independent sub-queries (builder queries A/B/C, formulas, PromQL). The `run()` loop in `pkg/querier/querier.go` executed them **serially**, so an N-query dashboard panel paid the **sum** of every sub-query's ClickHouse round trip instead of the slowest one.

## Change

Execute the sub-queries concurrently with bounded fan-out using `errgroup`, gated behind a new **default-off** experimental flag `enable_parallel_queries`.

- Per-query work is extracted into an `execOne` closure shared by both paths.
- Parallel path writes each result into its own pre-sized slot — **no shared state mutated concurrently**; the response is folded together sequentially after `g.Wait()`.
- When the flag is off, the original serial loop runs unchanged (zero blast radius).
- `errgroup.WithContext` preserves the existing first-error-wins + sibling-cancellation semantics.

### Concurrency bound / fan-out safety

`maxParallelQueries = 4`. Each sub-query may itself fan missing cache ranges out across up to 4 goroutines inside `executeWithCache`, so worst-case concurrent ClickHouse statements is `4 * 4 = 16` — well under `telemetrystore` `MaxOpenConns` (default 100).

## Benchmarks

Added `pkg/querier/run_bench_test.go` (`fakeQuery` simulating a 20ms ClickHouse round trip, isolating the dispatch loop):

| sub-queries | serial | parallel | speedup |
|---|---|---|---|
| 4 | 83.9 ms/op | 21.5 ms/op | **3.9×** |
| 8 | 168.7 ms/op | 42.3 ms/op | **4.0×** |

8 queries at limit 4 = two waves × 20ms ≈ 42ms, exactly as designed. Allocation overhead is negligible (slots slice + goroutines).

## Tests

- `go test -race ./pkg/querier/` — passes, including new `TestRunParallelMatchesSerial` (the concurrency correctness gate: parallel output matches serial).
- `go vet` + `gofmt` clean.

## Rollout

Default-off experimental flag; ramp per-org via the flagger. Output is identical to the serial path with the flag on or off.


kunalrawat425:
@srikanthccv  @therealpandey need review

kunalrawat425:
@srikanthccv please review ad merge

kunalrawat425:
@therealpandey approve

srikanthccv:
hello, this is the area we are not currently accepting external contributions. thank you.

PR #11839, opened 2026-06-24
fix(querier): honor max_concurrent_queries config for missing-range execution

kunalrawat425:
## Problem

`querier.Config` exposes `max_concurrent_queries` (default 4, validated to be `> 0`), but `executeWithCache` hard-coded its missing-range fan-out semaphore to a literal `4`:

```go
sem := make(chan struct{}, 4)
```

So the documented, validated config knob was **silently ignored**. When a query range request has uncached time ranges, the querier splits them and fetches them concurrently — operators had no way to tune how many of those ClickHouse queries run in parallel per request.

## Change

Thread the configured value through `querier.New` into the semaphore:

- New `maxConcurrentQueries` field on the querier, set from `cfg.MaxConcurrentQueries` in the provider wiring.
- A non-positive value falls back to the historical default (`defaultMaxConcurrentQueries = 4`), so existing deployments and in-package test constructors are unaffected.

No behavioural change at the default; this only makes the existing knob actually take effect.

## Tests

- New `TestNewMaxConcurrentQueries` covers honored value + fallback for `0`/negative.
- `go test -race ./pkg/querier/` passes; `go vet` + `gofmt` clean across the querier, provider, and rules packages.

### DIFF
--- a/pkg/querier/querier.go
+++ b/pkg/querier/querier.go
@@ -34,6 +34,10 @@ var (
 	intervalWarn = "Query %s is requesting aggregation interval %v seconds, which is smaller than the minimum allowed interval of %v seconds for selected time range. Using the minimum instead"
 )
 
+// defaultMaxConcurrentQueries is the fallback bound on goroutines used to fetch
+// the missing cache ranges of a single query when no valid value is configured.
+const defaultMaxConcurrentQueries = 4
+
 type querier struct {
 	logger                   *slog.Logger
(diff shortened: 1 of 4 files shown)


kunalrawat425:
@srikanthccv  @therealpandey need review

srikanthccv:
hello, this is the area we are not currently accepting external contributions. thank you.

PR #11840, opened 2026-06-24
fix(rules): load rules for all orgs when an earlier org has none

kunalrawat425:
## Problem

`Manager.initiate` loads stored alerting/recording rules for every org owned by the instance:

```go
for _, org := range orgs {
    ...
    if len(storedRules) == 0 {
        return nil   // <-- exits initiate entirely
    }
    ...
}
```

The `return nil` exits the **whole function** the moment any org has zero stored rules, so every org ordered after an empty one has its rules **silently skipped at startup** — their alerts never fire. With multiple owned orgs this is order-dependent and easy to miss.

## Fix

Replace the early `return nil` with `continue` so an org with no rules is simply skipped while the rest are still loaded.

## Test

Added `TestInitiateVisitsAllOrgsWhenSomeHaveNoRules`: drives `initiate` with two orgs where the first returns no rules and asserts both orgs are queried. It fails on the old code (only the first org is visited) and passes with the fix. Uses lightweight interface fakes — no DB/sqlmock setup required.


kunalrawat425:
@srikanthccv revioew

PR #11849, opened 2026-06-24
feat(metrics): add per-metric label-reduction rules

srikanthccv:
## Pull Request

Fixes https://github.com/SigNoz/engineering-pod/issues/5340
Fixes https://github.com/SigNoz/engineering-pod/issues/5342


pkg/types/metricreductionruletypes/reduction_rule.go
```diff
@@ -0,0 +1,168 @@
...
+func IsProtectedLabel(label string) bool {
+	_, ok := protectedLabels[label]
+	return ok
+}
+
+func (req *UpdatableReductionRule) Validate() error {
+	if req == nil {
+		return errors.NewInvalidInputf(errors.CodeInvalidInput, "request is nil")
+	}
+	if req.MatchType != MatchTypeDrop && req.MatchType != MatchTypeKeep {
```
namanv3, line 141: matchtype.validate() can be used

pkg/types/metricreductionruletypes/reduction_rule.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+	}
+	return nil
+}
+
+func (req *PostableReductionRule) Validate() error {
+	if req == nil {
+		return errors.NewInvalidInputf(errors.CodeInvalidInput, "request is nil")
+	}
+	if req.MetricName == "" {
```
namanv3, line 164: question: will the actual module impl involve metric existence checks? I'm guessing that will have to be present somewhere

pkg/types/metricreductionruletypes/reduction_rule.go
```diff
@@ -0,0 +1,168 @@
...
+	types.Identifiable
+	types.TimeAuditable
+	types.UserAuditable
+
+	MetricName       string    `json:"metricName" required:"true"`
+	MatchType        MatchType `json:"matchType" required:"true"`
+	Labels           []string  `json:"labels" required:"true" nullable:"true"`
+	EffectiveFrom    time.Time `json:"effectiveFrom" required:"true"`
+	Active           bool      `json:"active" required:"true"`
+	IngestedSeries   uint64    `json:"ingestedSeries" required:"true"`
```
namanv3, line 103: shouldn't the base ReductionRule contain all info about it? So these extra fields that are present here should be there in the base model, and maybe the storable can be a different type

pkg/types/metricreductionruletypes/reduction_rule.go
srikanthccv: yes

pkg/types/metricreductionruletypes/reduction_rule.go
srikanthccv: following  https://github.com/SigNoz/signoz/blob/main/docs/contributing/go/types.md#add-a-flavor-only-when-it-differs

the core type is `ReductionRule`

`ingestedSeries/retainedSeries/reductionPercent/active` are server-computed and never persisted so they warrant a `GettableX` flavour. 


similarly there isn't any specific thing that differs in core type to have a separate `StorableX` flavour.

pkg/types/metricreductionruletypes/reduction_rule.go
srikanthccv: would it be okay if I address this in follow up? I am expecting more comments in the next PR.

namanv3: sure, I'll approve once merge conflict is resolved

namanv3 (dismissed):

therealpandey (dismissed):

therealpandey (approved):

PR #11850, opened 2026-06-25
feat(querybuilder): type untyped value fields in v5 openapi schema

therealpandey:
## Summary

Makes the query-builder v5 OpenAPI schema generator-friendly so the generated backend spec (`docs/api/openapi.yml`) and the orval frontend client carry the real wire contract instead of `{}` / `unknown`. **Go field types are unchanged (`any` stays `any`) — this only shapes the generated schema**, so runtime decode paths and the wire format are untouched.

### Typed `value` fields

Previously-untyped `value` properties are typed via `PrepareJSONSchema`:

- `FunctionArg.value` → `number | string`
- `Label.value` → `string | number | boolean`
- `VariableItem.value` → `string | number | boolean | scalar[] | null` (the `null` covers a dynamic variable with "ALL" selected)

### Discriminated `QueryEnvelope`

`QueryEnvelope` now emits a real OpenAPI `discriminator` instead of a `spec: {}` blob, so code generators (oapi-codegen / orval) can map each variant 1:1. It is a two-level union:

- **`QueryEnvelope`** is discriminated by `type` over 5 variants: `builder_query`, `builder_formula`, `builder_trace_operator`, `promql`, `clickhouse_sql`.
- **`BuilderQuerySpec`** (the `builder_query` spec) is in turn discriminated by `signal` over trace / log / metric aggregations.

The discriminator `type` is marked `required` on every variant so oapi-codegen renders it as a non-pointer field (an optional/pointer discriminator wouldn't compile against the string-literal assignment in the generated `From<Variant>` constructors).

`builder_join` is **deferred** out of the union: its `aggregations` are an undiscriminable `oneOf` (trace ≡ log, and a join carries no `signal` to dispatch on), so generators can't map it yet. It stays `[]any` with a `TODO` to add a discriminator before re-enabling.

### Frontend

Regenerates `frontend/src/api/generated/services/sigNoz.schemas.ts` and adapts the dashboard-v2 `queryV5` consumers to the now-discriminated `QueryEnvelope` union — comparing the discriminator through the generated per-variant `type` enums (no magic strings) and keeping casts localized to the envelope-spec boundary. `tsc --noEmit` passes.

## Comparison with the hand-rolled FE types (`src/types/api/v5/queryRange.ts`)

| Field | Generated (this PR) | Hand-rolled FE | Notes |
|---|---|---|---|
| `FunctionArg.value` | `number \| string` | `string \| number` | exact match |
| `VariableItem.value` | `string \| number \| boolean \| scalar[] \| null` | `any` | generated is stricter; `null` confirmed in FE (dynamic "ALL" → `null`) |
| `Label.value` | `string \| number \| boolean` | `any` | generated is stricter; no wire `null` (FE only coalesces defensively) |
| `QueryEnvelope.spec` | `type`-discriminated `oneOf` of 5 variants | union of 5 spec types | see variant table |

### `QueryEnvelope` variant coverage

| Query type | FE spec union | Generated `oneOf` |
|---|---|---|
| builder_query (trace / log / metric) | ✓ | ✓ (signal-discriminated) |
| builder_query + meter | `MeterBuilderQuery` | ✓ (via metric `source` field) |
| builder_sub_query | `BuilderQuery` | ✓ (decoded as builder) |
| builder_formula | ✓ | ✓ |
| builder_trace_operator | ✗ (missing in FE) | ✓ |
| promql | ✓ | ✓ |
| clickhouse_sql | ✓ | ✓ |
| builder_join | ✓ | ✗ (deferred — needs an aggregation discriminator) |


AshwinBhatkal (dismissed):

therealpandey:
Blocking https://github.com/SigNoz/terraform-provider-signoz/pull/113

pkg/types/querybuildertypes/querybuildertypesv5/req.go
```diff
@@ -276,6 +318,42 @@ type VariableItem struct {
...
+	list.WithItems(items)
+
+	value := jsonschema.Schema{}
+	value.OneOf = []jsonschema.SchemaOrBool{
+		jsonschema.String.ToSchemaOrBool(),
+		jsonschema.Number.ToSchemaOrBool(),
+		jsonschema.Boolean.ToSchemaOrBool(),
+		list.ToSchemaOrBool(),
+	}
+	// nullable: the frontend sends null for a dynamic variable's "ALL" selection.
```
srikanthccv: misleading, it sends special marker value `__all__` for `ALL` selection of dynamic variable. should be non-nullable.

pkg/types/querybuildertypes/querybuildertypesv5/req.go
```diff
@@ -21,71 +21,113 @@ type QueryEnvelope struct {
...
 type queryEnvelopePromQL struct {
-	Type QueryType `json:"type" description:"The type of the query."`
+	Type QueryType `json:"type" required:"true" description:"The type of the query."`
 	Spec PromQuery `json:"spec" description:"The PromQL query specification."`
 }
 
 // queryEnvelopeClickHouseSQL is the OpenAPI schema for a QueryEnvelope with type=clickhouse_sql.
 type queryEnvelopeClickHouseSQL struct {
-	Type QueryType       `json:"type" description:"The type of the query."`
+	Type QueryType       `json:"type" required:"true" description:"The type of the query."`
```
srikanthccv, line 92: question: is this `required` the tag the property used for `Computed: true`?

pkg/types/querybuildertypes/querybuildertypesv5/builder_elements.go
```diff
@@ -621,6 +621,25 @@ func (f FunctionArg) Copy() FunctionArg {
...
+// untyped {}. The Go field stays `any`; this only shapes the generated schema.
+func (FunctionArg) PrepareJSONSchema(s *jsonschema.Schema) error {
+	if _, ok := s.Properties["value"]; !ok {
+		return nil
+	}
+
+	value := jsonschema.Schema{}
+	value.OneOf = []jsonschema.SchemaOrBool{
+		jsonschema.Number.ToSchemaOrBool(),
+		jsonschema.String.ToSchemaOrBool(),
```
srikanthccv, line 636: note: at the moment only numbers are supported for function args. 

and perhaps, this could even be typed by function name similar to other discriminators.

pkg/types/querybuildertypes/querybuildertypesv5/req.go
therealpandey: Will update this.

pkg/types/querybuildertypes/querybuildertypesv5/req.go
therealpandey: Not sure, it's a combination of a number of things.

pkg/types/querybuildertypes/querybuildertypesv5/builder_elements.go
therealpandey: anomaly has string I suppose?

pkg/types/querybuildertypes/querybuildertypesv5/req.go
therealpandey: Fixed

srikanthccv (approved):

PR #11876, opened 2026-06-26
fix(query-service): parameterize log field name lookup in GetLogFieldsFromNames

kunalrawat425:
## Summary

`GetLogFieldsFromNames` (`pkg/query-service/app/clickhouseReader/reader.go`) built its `WHERE name IN (...)` clause by string-joining the supplied field names directly into the SQL:

```go
query := fmt.Sprintf("... where name in ('%s') ...", strings.Join(fieldNames, "','"))
```

These names are request-derived (the builder query keys — `groupBy` / filter / aggregate / `orderBy` — collected by `logsv3.GetFieldNames`), so the raw `IN ('...')` interpolation is a SQL-injection sink. Every other value in these queries already uses bind parameters; the field-name list was the exception.

## Change

- Bind the field names as `?` parameters and pass them as query args (the same pattern already used elsewhere in this file for an `IN` list).
- Skip the lookup entirely when there are no field names (avoids an empty `IN ()`).
- No behavioural change for legitimate field names.

## Notes for maintainers

This is a **security-relevant** fix (input validation / SQL injection hardening on an authenticated request path) — you may want to handle it via your advisory/backport process. I've intentionally kept exploit specifics out of this public description; happy to share full details privately at security@signoz.io.

## Testing

- `go build` / `go vet` / `gofmt` clean on the package.


kunalrawat425:
@srikanthccv review

PR #11880, opened 2026-06-27
fix(frontend): default Traces Explorer to Trace View and root spans

### LINKED ISSUE #11861: Traces Page - default view should be traces root spans not all spans

From the default traces page showing all spans is not very helpful.  showing root spans (the trace id) is what i would expect (coming from DataDog)

When I click into the traces page, for the query the default 'in' should be 'root spans' not 'all spans'

I would also argue that the 'Trace View' tab should be the default vs the 'List View' tab.

And if you happen to click on a span or trace without right clicking and opening in a new tab, when i click back UI should remember my query.

none

## Additional context

Switching from DataDog .

Taranum01:
## Summary
- Default `/traces-explorer` to **Trace View** (`PANEL_TYPES.TRACE`) instead of List View.
- Add `isRoot=true` to the explorer's initial query via `withRootSpanFilter()` so the span scope defaults to **Root Spans** (scoped to Traces Explorer only).
- Persist defaults in the URL on first load (`panelTypes` + `selectedExplorerView`) and sync QueryBuilder `panelType` on mount.

Fixes #11861

## Problem
Users landing on Traces Explorer see **All Spans** in **List View**, which is noisy compared to tools like Datadog that default to trace/root-span oriented views.

## Out of scope
Back-navigation query persistence (third item in the issue) looks like a separate `panelType`/URL sync bug — happy to follow up in another PR if maintainers want it.

## Test plan
- [ ] Visit `/traces-explorer` with no query params → Trace View tab active, span scope shows **Root Spans**, results are root spans only
- [ ] URL includes `panelTypes` and `selectedExplorerView` for trace on first load
- [ ] Switching to List View / All Spans still works manually
- [ ] Dashboards, alerts, and other trace query entry points unchanged (no global `initialQueriesMap.traces` change)
- [x] `explorerUtils.test.ts` — `withRootSpanFilter` adds `isRoot` once, no duplicate

Made with [Cursor](https://cursor.com)

### DIFF
--- a/frontend/src/container/TracesExplorer/QuerySection/index.tsx
+++ b/frontend/src/container/TracesExplorer/QuerySection/index.tsx
@@ -8,7 +8,7 @@ import { useGetPanelTypesQueryParam } from 'hooks/queryBuilder/useGetPanelTypesQ
 import { DataSource } from 'types/common/queryBuilder';
 
 function QuerySection(): JSX.Element {
-	const panelTypes = useGetPanelTypesQueryParam(PANEL_TYPES.LIST);
+	const panelTypes = useGetPanelTypesQueryParam(PANEL_TYPES.TRACE);
 
 	const filterConfigs: QueryBuilderProps['filterConfigs'] = useMemo(() => {
 		const isList = panelTypes === PANEL_TYPES.LIST;
--- a/frontend/src/container/TracesExplorer/explorerUtils.test.ts
+++ b/frontend/src/container/TracesExplorer/explorerUtils.test.ts
@@ -1,0 +1,31 @@
+import { initialQueriesMap } from 'constants/queryBuilder';
+import { cloneDeep } from 'lodash-es';
+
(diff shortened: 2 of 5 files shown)


Taranum01:
Bumping — opened 11 days ago and not yet picked up for review.

This is a small frontend fix for the Traces Explorer: by default
the panel opens on the Trace View filtered to root spans, which is
what most users want when investigating a request. It also fixes a
related state-hydration bug where the persisted preference was being
applied before the trace data finished loading, causing a flash of
the unfiltered list.

CLA is signed, CI checks are green on the latest commit, and the PR
is mergeable. Could a maintainer take a quick look when convenient?
Happy to address any feedback.

Taranum01:
Friendly bump — this PR has been sitting without review attention. Rebased, ready for review.

If anything needs to be addressed (additional tests, naming, scope, etc.), happy to iterate.

PR #11885, opened 2026-06-29
feat(infra-monitoring): pod status and restart counts.

nikhilmantri0902:
### 📄 Summary

Adds **kubectl-style pod display status** and **restart counts** to Infra Monitoring, derived purely from `k8sclusterreceiver` metrics (no extra API calls).

- **Pods list** now returns per-pod `podStatus` (e.g. `CrashLoopBackOff`, `OOMKilled`, `ImagePullBackOff`, `Evicted`, `Completed`, …), per-group `podCountsByStatus`, and `podRestarts` — mirroring `kubectl get pods` STATUS/RESTARTS and DD's `computeStatus`/`ExtractPod`.
- **Status group counts** (`podCountsByStatus`) added to the grouped infra entities that already surface phase counts: nodes, namespaces, clusters, deployments, daemonsets, statefulsets, jobs.
- Status is computed via a priority cascade (container reason > pod-level reason > phase) and is **gated on the 3 required metrics existing** in `distributed_metadata`; if any is missing the query is skipped and a non-blocking warning (with a docs link) is surfaced instead.
- `/checks` now reports the status metrics so users know to enable the optional ones.

This is the right approach because the status/restart values come from the same `containerStatuses[].*` source kubectl and Datadog use, so the UI will match what users already expect — without per-pod API reads.

Full discussion ticket: https://github.com/SigNoz/engineering-pod/issues/5471

#### Issues closed by this PR

Closes SigNoz/engineering-pod#5472

---

### ✅ Change Type
- [x] ✨ Feature
- [x] 🧪 Test-only (integration tests included)

---

### 🧪 Testing Strategy

- **Tests added:** Python integration tests for pods (`podStatus`/`podCountsByStatus`/`podRestarts` — list, grouped, priority cascade, latest-wins/staleness, missing-metric gate warning) and grouped `podCountsByStatus` for clusters & deployments; `/checks` SPECS extended (auto-generates a missing-optional case per entity). Full infra-monitoring suite green (**549 passed**).
- **Manual verification:** ran end-to-end against a local kind cluster with deliberately-broken pods (CrashLoopBackOff / OOMKilled / ImagePullBackOff / CreateContainerConfigError / Pending / Evicted). `podStatus`, `podCountsByStatus`, and `podRestarts` matched `kubectl get pods` and per-cluster `RESTARTS` sums exactly.
- **Edge cases:** crashloop oscillation (waiting vs terminated reason at scrape time), fingerprint staleness across container restarts (argMax-by-latest), restart-count incarnations (no double-counting), and the no-data sentinel (`podStatus: no_data`, `podRestarts: -1`).

---

- **Blast radius:** Infra Monitoring list + checks APIs only; all additions are new, additive response fields.
- **Potential regressions:** low — new fields default to zero/sentinel; the status query is gated and degrades to a warning when metrics are absent.
- **Rollback:** revert the PR; no schema/migration changes.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Infra Monitoring now shows kubectl-style pod **status** (CrashLoopBackOff, OOMKilled, etc.) and **restart counts** on the Pods list, plus per-group pod-status counts across nodes/namespaces/clusters/deployments/daemonsets/statefulsets/jobs. Requires the optional `k8s.pod.status_reason` and `k8s.container.status.reason` metrics (flagged in the Infra Monitoring checks page). |

---

### 📋 Checklist
- [x] Manually tested
- [x] Breaking changes documented (none — additive)

---

## 👀 Notes for Reviewers

- **Known, documented limitations** (receiver gaps, not bugs): `Terminating`, `Init:*`, `Signal/ExitCode:N` aren't derivable; init-container restarts are excluded (matches DD post-init); for multi-container pods we pick the reason by **priority** (waiting > terminated), since metrics carry no container ordering — a deliberate divergence from kubectl's order-based pick.
- The optional metrics are enabled by default in the `k8s-infra` chart via a separate merged PR (SigNoz/charts#886).
- Status/restart wiring reuses one generic per-group query + the shared `podStatusCountsToResponse` helper across all entities (counts-only for the grouped entities; pods additionally has per-pod status + restarts).


pkg/modules/inframonitoring/implinframonitoring/helpers.go
```diff
@@ -440,6 +440,30 @@ func (m *module) buildFilterClause(ctx context.Context, filter *qbtypes.Filter,
 	return whereClause.WhereClause, nil
 }
 
+// mergeQueryWarnings combines any number of query warnings. It is nil-safe and
+// skips nil entries. The first non-nil warning becomes the primary; each
+// subsequent one contributes its message (as an additional warning) and its
+// own additional warnings. Returns nil when all inputs are nil.
+func mergeQueryWarnings(warnings ...*qbtypes.QueryWarnData) *qbtypes.QueryWarnData {
```
namanv3, line 447: didn't get why this was needed

pkg/types/inframonitoringtypes/pods_constants.go
```diff
@@ -26,6 +26,80 @@ func (PodPhase) Enum() []any {
...
+		PodStatusNodeAffinity,
+		PodStatusNodeLost,
+		PodStatusShutdown,
+		PodStatusUnexpectedAdmissionError,
+		PodStatusNoData,
+	}
+}
+
+// Numeric pod status_reason values emitted by the k8s.pod.status_reason metric.
+const (
```
namanv3: these aren't used anywhere

pkg/modules/inframonitoring/implinframonitoring/pods.go
```diff
@@ -334,3 +389,519 @@ func (m *module) getPerGroupPodPhaseCounts(
...
+	ctx context.Context,
+	start, end int64,
+	filter *qbtypes.Filter,
+	groupBy []qbtypes.GroupByKey,
+	pageGroups []map[string]string,
+) (map[string]int64, error) {
+	if len(pageGroups) == 0 || len(groupBy) == 0 {
+		return map[string]int64{}, nil
+	}
+
```
namanv3, line 785: no check for metric being absent by chance?

nikhilmantri0902: Ya we have a check for the metrics used in status calculation being present via the getPerGroupPodStatusCounts function ( only, because it should be calculated only if all 3 metrics are flowing in ). Rest we have checks in the /checks API that will be called when the page loads to inform the user what metrics are flowing in and what are absent.

pkg/modules/inframonitoring/implinframonitoring/helpers.go
nikhilmantri0902: The status counts come from a second, separately-gated query (getPerGroupPodStatusCountsWithReqMetricChecks) that emits its own warning (e.g. "enable the optional status metrics"), independent of the main query's warning (e.g. a utilization metric never received). Both can be non-nil at once, but resp.Warning is a single field — a plain assignment would drop one. mergeQueryWarnings combines them (first non-nil = primary, the rest appended as additional warnings), is nil-safe, and is reused across all 8 list endpoints that run this second query.

pkg/types/inframonitoringtypes/pods_constants.go
nikhilmantri0902: removed

namanv3 (approved):

therealpandey (approved):

PR #11641, opened 2026-06-10
refactor(alerts): move channels to alerts

H4ad:
## Pull Request

---

This PR moves the Notification Channels to be under Alerts, before Configuration Tab.

I still kept the `Notification Channels` under settings, but we can eventually remove it from there after few releases.

https://github.com/user-attachments/assets/d620ef4f-3898-41fc-a45b-c41e04f1512a

When viewer:

image

Fixes https://github.com/SigNoz/engineering-pod/issues/5019

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Notification Channels
- Potential regressions: Bad Redirect / Unknown
- Rollback plan: Find and fix the issue specifically.

---

| Field | Value |
|------|-------|
| Deployment Type | Cloud / OSS / Enterprise |
| Change Type | Feature |
| Description | The Notifications Channels now was moved to be under `Alerts` page. |

---

### 📋 Checklist
- [x] Tests added or explicitly not required
- [x] Manually tested
- [ ] Breaking changes documented
- [ ] Backward compatibility considered

PR #11888, opened 2026-06-29
fix(notification-channels): edit not persisting any information

H4ad:
## Pull Request

---

This fixes a regression introduced at https://github.com/SigNoz/signoz/pull/11641 that broke the edit for notification channels.

After:

https://github.com/user-attachments/assets/9f3f8241-c548-4a3a-b267-4b7bd27ebbb8

Fixes https://github.com/SigNoz/engineering-pod/issues/5509

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

I updated the path to be from `/settings/` to be `/alerts`, and the channels page was using RegExp to extract the `channelId`.

I forgot to update the RegExp to include the updated path.

Pass ID as prop instead of parsing it twice, and use `useMatch` instead of RegExp to extract the ID more reliably.

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered: Yes

---

- Blast radius: Notification Channels - Edit
- Potential regressions: -

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | FIxed the issue on Notification Channels not persisting the changes of the page. |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/container/AllAlertChannels/__tests__/EditAlertChannelSave.test.tsx
+++ b/frontend/src/container/AllAlertChannels/__tests__/EditAlertChannelSave.test.tsx
@@ -1,0 +1,81 @@
+import EditAlertChannels from 'container/EditAlertChannels';
+import { editAlertChannelInitialValue } from 'mocks-server/__mockdata__/alerts';
+import { server } from 'mocks-server/server';
+import { rest } from 'msw';
+import { render, screen, userEvent, waitFor } from 'tests/test-utils';
+
+jest.mock('hooks/useNotifications', () => ({
+	__esModule: true,
+	useNotifications: jest.fn(() => ({
+		notifications: { success: jest.fn(), error: jest.fn() },
+	})),
+}));
+
+jest.mock('components/MarkdownRenderer/MarkdownRenderer', () => ({
+	MarkdownRenderer: jest.fn(() => <div>Mocked MarkdownRenderer</div>),
+}));
+
+interface EditRequest {
+	id: string;
+	body: { name: string; slack_configs: { send_resolved: boolean }[] };
+}
+
+// Captures the PUT /channels/:id request the edit form fires, so assertions can
+// run against the real HTTP payload instead of a hand-mocked api client.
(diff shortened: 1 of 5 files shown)


frontend/src/pages/ChannelsEdit/index.tsx
```diff
@@ -147,6 +148,7 @@ function ChannelsEdit(): JSX.Element {
 			<div className="edit-alert-channels-container">
 				<EditAlertChannels
 					{...{
+						channelId: channelId || '',
```
AshwinBhatkal, line 151: nit: just preference. I like this to be clean given we already are declaring a variable for this

So just { channelId }

AshwinBhatkal (approved):

frontend/src/container/AllAlertChannels/__tests__/EditAlertChannelSave.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+			return res(
+				ctx.status(200),
+				ctx.json({ status: 'success', data: 'channel updated' }),
+			);
+		}),
+	);
+	return result;
+}
+
+describe('EditAlertChannels save', () => {
```
AshwinBhatkal, line 42: are there other parts that still require tests? Can we create a ticket for tracking them now that you have already gone into it? You can assign them to me

H4ad: We can rewrite/add tests all pages/components, don't think the existing tests are good enough.

Ref: https://github.com/SigNoz/engineering-pod/issues/5522

frontend/src/pages/ChannelsEdit/index.tsx
H4ad: There are certainly many things to improve here

PR #11911, opened 2026-06-30
chore: refactor querier to lossen the tight coupling of querybuilder module with other modules

tushar-signoz:
## Summary

Loosens the tight coupling between `querybuilder` and the telemetry modules by making each condition builder own its key-resolution logic instead of the shared filter visitor.

- **`ConditionBuilder.ConditionFor` now returns `([]string, []string, error)`** (conditions + warnings) and takes the matching `fieldKeysForName`. The builder now owns the decisions previously baked into the visitor: resolve ambiguity, fall back to a body-JSON search, emit a "key not found" error, or skip.
- **`VisitKey` only parses** a key now; key matching/ambiguity resolution moved into shared helpers (`ResolveKeys`, `NewKeyNotFoundError`, `matchingFieldKeys`) and the per-builder `ConditionFor`.
- **Body-JSON handling moved out of the visitor.** Removed `BodyJSONEnabled`, `JsonKeyToKey`, and `IgnoreNotFoundKeys` from `FilterExprVisitorOpts`; the logs builder now owns `body.*` resolution, the `body` full-text default warning, and the `has/hasAny/hasAll/hasToken` functions.
- **Added `FilterOperatorHas/HasToken/HasAny/HasAll`** (+ `IsArrayFunctionOperator`) so function calls flow through the standard condition-builder path.
- **`GetFieldKeyFromKeyText` now delegates to `Normalize`** (single parsing path).

## Change Type

♻️ Refactor

## Testing

- Existing condition-builder / filter-expression unit tests updated to the new signature; behavior preserved.

## Risk & Impact

- **Blast radius:** all signals (logs, traces, metrics, metadata, audit, resource filter, rule-state-history) go through the changed `ConditionFor` interface.
- **Backward compatibility:** query semantics unchanged for standard filters; no API/schema changes.

## Changelog

| Field | Value |
|------|-------|
| Deployment Type | OSS / Cloud / Enterprise |
| Change Type | Maintenance |
| Description | N/A (internal refactor) |

## Known Gaps / Follow-ups

- **`has/hasAny/hasAll/hasToken` are only wired through the logs builder.** The grammar emits these for all signals, but other builders don't handle them: traces falls through to `return "", nil` so the filter term is silently dropped (query runs unfiltered, no error); metadata's related-values path builds malformed `if(mapContains(...), , true)` SQL. Metrics errors cleanly. Needs per-builder handling (or upstream gating to logs) plus the missing `len(conds) == 0` guard in `VisitFunctionCall`.
- **gofmt:** `pkg/querybuilder/where_clause_visitor.go` (missing comma spaces in `buildConditions(... op,nil)` calls) and `pkg/telemetryresourcefilter/statement_builder.go` (struct alignment) are not gofmt-clean — CI lint will fail. Run `gofmt -w`.
- **Dropped array-type validation:** logs `conditionForArrayFunction` no longer rejects non-array body fields, so `has(body.scalarField, …)` (json-body on) hits a ClickHouse runtime error instead of a clear validation message.
- **Latent:** `fallback_expr.go` `conds[0]` has no length guard (safe with current callers; panics if a builder that can return empty is wired in).


tushar-signoz:
## SQL-parity verification via `/api/v5/query_range/preview`

Since this is meant to be a mechanical refactor (`ConditionFor` single-condition → multi-key `([]conditions, []warnings, error)`, key resolution moved from the filter visitor into the per-signal condition builders), I verified that generated SQL is unchanged by diffing the dry-run endpoint's output between the merge-base build and this branch.

**Method**
- 64 scenarios covering: key resolution (explicit contexts, `:datatype` suffixes, ambiguous keys, not-found), all operator families (=, IN, BETWEEN, LIKE/ILIKE, CONTAINS, REGEXP, EXISTS/NOT EXISTS), full-text, body-JSON paths, `has`/`hasAny`/`hasAll`/`hasToken` incl. error paths, aggregations/groupBy, traces (string+number ambiguity, function rejection), metrics (rate/gauge).
- Both binaries hit the same tenant with a pinned time window; compared per scenario: statement SQL + args, warnings, error message/URL (EXPLAIN estimates excluded).
- Run twice: `use_json_body` **off** (legacy-schema tenant) and **on** (tenant with `body_v2` + `signoz_metadata.distributed_field_keys`).

**Results**
1. **Flag off: byte-identical.** All 64 scenarios — 55 rendered statements (34,168 SQL chars), 9 expected-error scenarios with matching messages/doc URLs. `diff` exit 0.
2. **Flag on: byte-identical except 7 scenarios, all deliberate error-shape changes** (no SQL differences for any resolvable key):
   - `has()`/`hasToken` misuse now reports a single error instead of key-not-found + function-error double-reporting (builder-owned resolution).
   - On traces/metrics, the function-unsupported error takes precedence over key-not-found (`has(unknown.key, ...)` → ``function `has` supports only body JSON search``).
   - `hasToken(body, x)` in JSON mode targets `body.message` (valid — the SELECT aliases `body_v2 as body`; passes EXPLAIN) instead of the legacy raw `body` column.

Behavior notes confirmed on both sides: unknown `body.<key>` errors `key not found` when `use_json_body` is on (the body-JSON fallback applies only in legacy string-body mode), and known body keys (`body.message CONTAINS`, planned JSON paths) render identical conditions.

tushar-signoz:
## Black-box SQL parity check (dry-run, vs `main`)

Diffed the v5 statement-builder output between this branch (`7e08f757c8`) and `main` (`66f03d5912`), byte-for-byte, using **real query shapes harvested from 9 SaaS tenants'** ClickHouse `system.query_log` (last 3d, filtered to `log_comment.code.namespace = 'querier'` builder queries). Both binaries ran against the **same seeded devenv ClickHouse** and were compared via `POST /api/v5/query_range/preview?verbose=true` (renders SQL, never executes).

**Result: 69 MATCH · 0 MISMATCH** — rendered ClickHouse SQL **and** bind args are **byte-identical** for every reconstructed query. No regressions.

**Covered** (logs + traces filter shapes): resource-filter fingerprint CTEs, materialized→map attribute access, string/number/bool attrs, `EXISTS`/`NOT EXISTS`, `IN`/`LIKE`/`ILIKE`/`REGEXP`/`NOT CONTAINS`, `hasError`, body-JSON (legacy fallback), nested AND/OR, `groupBy`.

**32 SKIPPED** (not failures — both binaries behaved identically): metrics shapes (need proper metric-aggregation payloads, not `count()`), filterless shapes, and raw `clickhouse_sql` panels (not statement-builder output). `wired-tortoise` had no v5 `query_range` traffic in the window.

Corroborates the earlier hand-authored `preview-sql-check` parity (0-byte diff). Validates HEAD = the mechanical refactor (the not-found→warning change is not in this HEAD); metrics parity is a follow-up.

therealpandey (approved):

PR #11918, opened 2026-06-30
feat(dashboards-v2): panel editor — threshold carry, span-gaps fixes, metric unit defaults, live thresholds and pie multi-column fix

ahrefabhi:
## Pull Request

---

### 📄 Summary

Five related improvements to the **Dashboards V2 panel editor / panels**, bringing them closer to V1 parity and fixing reactivity gaps in the config controls. Each is a self-contained commit:

1. **Carry thresholds across a visualization-type switch** — switching a panel's kind mid-edit now preserves thresholds when the target kind supports them, remapping each to the target's variant (`label` / `comparison` / `table`): the shared core (color, value, unit) is kept and any variant-required fields (operator, format, column) are seeded with sensible defaults. Kinds without a Thresholds section drop them.
2. **Span-gaps "Disconnect Values" reactivity + `fillOnlyBelow` flag** — fixes three bugs in the chart-appearance control and makes `fillOnlyBelow` the authoritative mode flag end-to-end.
3. **Auto-seed the panel unit from the metric, with a mismatch warning** — V1 parity for the formatting unit selector; also fixes the warning tooltip never being visible.
4. **Reactive threshold editing** — threshold edits now reflect in the panel preview live, before Save.
5. **Pie panel: multi-column ClickHouse scalar no longer collapses to a single slice** — a pie backed by a query with several value columns (e.g. `count() AS col1, sum() AS col2`) rendered only the first column as one slice; it now renders one slice per value column. This is the V2 counterpart of the V1 fix in #11919 (table collapse family, #11794).

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/5543

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
**Span-gaps "Disconnect Values" control:**
- The threshold default was captured into local state **once at mount**, but the query step interval arrives asynchronously — so it always fell back to `1m`.
- The threshold input committed on **blur even when unchanged**; clicking "Never" blurs the input first and re-commits the threshold, so the mode wouldn't switch (a blur/toggle race).
- Validation only ran on blur, so a random string showed **no error while typing**.
- `fillOnlyBelow` was never written and the renderer keyed only off `fillLessThan`, so the documented mode flag was effectively ignored.

**Unit-mismatch warning tooltip:** antd renders the Select `suffixIcon` inside `.ant-select-arrow`, which has `pointer-events: none`, so hover never reached the icon and the tooltip never appeared.

**Pie panel multi-column collapse:** `preparePieData` resolved the value column with `columns.find(isValueColumn)` — the **first** value column only — so the additional aggregation columns of a scalar (a ClickHouse multi-aggregation query) were dropped and the pie collapsed to a single slice. Same family as the V1 table collapse fixed in #11794 / #11919.

#### Fix Strategy
- Seed the threshold from the **live** step interval at switch time, not at mount.
- Make `commit` a no-op when the value is unchanged (removes the blur/toggle race).
- Validate live as the user types.
- Write `fillOnlyBelow` (`true` for Threshold, `false` for Never, dropping the duration); the editor's selected mode and the renderer both honor it, with a back-compat fallback for panels saved before the flag existed.
- Wrap the warning icon in a `span` with `pointer-events: auto` (also gives antd's Tooltip a DOM node to attach its ref/handlers to).
- **Pie panel:** iterate **every** value column instead of the first — one slice per (row × value column); with multiple value columns the column name distinguishes the slices (prefixed by the group when grouped). Single-value and grouped pies iterate exactly once, so their output is unchanged.

---

- **Tests added/updated:**
  - `getSwitchedPluginSpec.test.ts` — threshold carry + variant remap, drop when unsupported.
  - `ChartAppearanceSection.test.tsx` + `resolvers.test.ts` — step-interval seeding (incl. async-arrival regression), no re-commit on blur, live validation, `fillOnlyBelow` true/false writes, renderer honoring the flag + legacy fallback.
  - `FormattingSection.test.tsx` + `useMetricYAxisUnit.test.ts` — auto-seed on new panel only, no overwrite on edit, mismatch warning shown/hidden.
  - `YAxisUnitSelector.test.tsx` — warning tooltip on hover (converted to `userEvent`).
  - `ThresholdsSection.test.tsx` + `ThresholdsSection.comparison.test.tsx` — live edit reflected before Save, revert-to-snapshot on Discard.
  - `PieChartPanel/prepareData.test.ts` — slice-per-value-column for a multi-column ClickHouse scalar, single grouped value column unchanged, grouped multi-value labels (`group · column`), `customColors` precedence, non-positive/non-numeric filtering.
- **Manual verification:** local app; `pnpm tsgo --noEmit`, `pnpm lint:js`, and `pnpm build` all green. Multi-column ClickHouse pie now renders one slice per column, matching the Table panel for the same query.
- **Edge cases covered:** async step-interval; legacy span-gaps spec without `fillOnlyBelow`; threshold variant transforms across kinds; new-panel vs edit gating for unit auto-seed; pie single-value vs multi-value column paths.

---

- **Blast radius:** V2 dashboards panel editor (thresholds, formatting, chart appearance) and the V2 Pie panel. One **shared** component touched — `YAxisUnitSelector` (also used by V1 dashboards, alerts, metrics explorer); the change is additive (a `span` wrapper + `pointer-events`), and the warning is gated on `initialValue`, which only V2 newly passes.
- **Potential regressions:** TimeSeries span-gaps rendering (mitigated by `fillOnlyBelow` back-compat + resolver tests); the `YAxisUnitSelector` warning icon is now hoverable (previously inert, no behavior removed). Pie: a single value column iterates exactly once, so single-value and grouped pies are byte-identical; only the previously-broken multi-value-column case changes.
- **Rollback plan:** revert the PR; the five commits are independent and per-feature.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Description | Dashboards V2 panel editor: thresholds carry across panel-type switches, span-gaps "Disconnect Values" reactivity fixes, metric-driven unit defaults with a mismatch warning, live threshold preview before save, and pie panels now render one slice per value column for multi-column ClickHouse queries. |

---

### 📋 Checklist
- [x] Manually tested
- [x] Backward compatibility considered (`fillOnlyBelow` legacy fallback; warning is opt-in via `initialValue`; pie single-value path unchanged)

---

## 👀 Notes for Reviewers

- **Shared component:** the `YAxisUnitSelector` tooltip fix (commit "auto-seed panel unit…") touches a V1-shared component. It's additive and folded into the metric-unit commit since that feature surfaced the bug. The V2 work reuses `useGetYAxisUnit` and `YAxisUnitSelector` **read-only** per the V1/V2 split policy.
- **`fillOnlyBelow`:** now authoritative — explicit `false` spans every gap regardless of a stored `fillLessThan`; a duration with no flag still thresholds (legacy panels).
- **Live threshold editing:** because edits now stream to the spec, abandoning an edit by clicking another row's Edit (instead of Discard) keeps the change; only explicit **Discard** reverts (to the snapshot taken on edit entry). Flag if you'd prefer abandon-reverts.
- **Pie multi-column fix** duplicates the slice-building logic into V2 land rather than importing the V1 helper, per the V1/V2 split policy. Grouped + multi-value slices are labelled `group · column` (middle-dot separator) — easy to change if a different convention is preferred.
- Commits are scoped per problem for easier review.

---

### DIFF
--- a/frontend/src/components/YAxisUnitSelector/YAxisUnitSelector.tsx
+++ b/frontend/src/components/YAxisUnitSelector/YAxisUnitSelector.tsx
@@ -2,7 +2,7 @@ import { useMemo } from 'react';
 import { SolidAlertTriangle } from '@signozhq/icons';
 import { Select, Tooltip } from 'antd';
 import type { DefaultOptionType } from 'antd/es/select';
-import classNames from 'classnames';
+import cx from 'classnames';
 
 import { UniversalYAxisUnitMappings } from './constants';
 import { UniversalYAxisUnit, YAxisUnitSelectorProps } from './types';
@@ -72,9 +72,7 @@ function YAxisUnitSelector({
 	}, [categoriesOverride, source]);
 
 	return (
-		<div
-			className={classNames('y-axis-unit-selector-component', containerClassName)}
-		>
+		<div className={cx('y-axis-unit-selector-component', containerClassName)}>
 			<Select
 				showSearch
 				value={universalUnit}
@@ -84,12 +82,17 @@ function YAxisUnitSelector({
 				loading={loading}
 				suffixIcon={
 					incompatibleUnitMessage ? (
-						<Tooltip title={incompatibleUnitMessage}>
-							<SolidAlertTriangle role="img" aria-label="warning" size="md" />
+						<Tooltip
+							title={incompatibleUnitMessage}
+							overlayClassName="y-axis-unit-warning-tooltip"
+						>
+							<span className="y-axis-unit-warning" role="img" aria-label="warning">
+								<SolidAlertTriangle size="md" />
+							</span>
 						</Tooltip>
 					) : undefined
 				}
-				className={classNames({
+				className={cx({
 					'warning-state': incompatibleUnitMessage,
 				})}
 				data-testid={dataTestId}
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/sections/ThresholdsSection/ThresholdsSection.tsx
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/sections/ThresholdsSection/ThresholdsSection.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useRef, useState } from 'react';
 import { Plus } from '@signozhq/icons';
 import { Button } from '@signozhq/ui/button';
 import {
@@ -82,12 +82,26 @@ function ThresholdsSection({
 	// Which row is being edited, and whether it was just added (so Discard removes it).
 	const [editingIndex, setEditingIndex] = useState<number | null>(null);
 	const [unsavedIndex, setUnsavedIndex] = useState<number | null>(null);
+	// The saved threshold captured on edit entry, restored if the edit is discarded
+	// (edits stream into the spec live, so Discard can't just drop a local draft).
+	const editSnapshot = useRef<AnyThreshold | null>(null);
+
+	const updateAt =
+		(index: number) =>
+		(next: AnyThreshold): void => {
+			onChange(thresholds.map((t, i) => (i === index ? next : t)));
+		};
 
 	const addThreshold = (): void => {
 		const nextIndex = thresholds.length;
 		onChange([...thresholds, defaultThreshold(variant, tableColumns)]);
 		setEditingIndex(nextIndex);
 		setUnsavedIndex(nextIndex);
+	};
+
+	const beginEdit = (index: number): void => {
+		editSnapshot.current = thresholds[index] ?? null;
+		setEditingIndex(index);
 	};
 
 	const saveAt =
@@ -105,10 +119,14 @@ function ThresholdsSection({
 	};
 
 	const discardAt = (index: number) => (): void => {
-		// Discarding a row that was never saved removes it; otherwise just exit edit.
+		// A never-saved row is removed; otherwise revert the live edits to the snapshot.
 		if (index === unsavedIndex) {
 			removeAt(index);
 			return;
+		}
+		const original = editSnapshot.current;
+		if (original) {
+			onChange(thresholds.map((t, i) => (i === index ? original : t)));
 		}
 		setEditingIndex(null);
 	};
@@ -120,8 +138,9 @@ function ThresholdsSection({
 			index,
 			yAxisUnit,
 			isEditing: editingIndex === index,
-			onEdit: (): void => setEditingIndex(index),
+			onEdit: (): void => beginEdit(index),
 			onSave: saveAt(index),
+			onLiveChange: updateAt(index),
 			onDiscard: discardAt(index),
 			onRemove: (): void => removeAt(index),
 		};
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/sections/ThresholdsSection/__tests__/ThresholdsSection.test.tsx
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/sections/ThresholdsSection/__tests__/ThresholdsSection.test.tsx
@@ -1,5 +1,5 @@
 import { useState } from 'react';
-import { fireEvent, render, screen } from '@testing-library/react';
+import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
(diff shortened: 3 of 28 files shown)


frontend/src/components/YAxisUnitSelector/YAxisUnitSelector.tsx
```diff
@@ -84,12 +82,17 @@ function YAxisUnitSelector({
 				loading={loading}
 				suffixIcon={
 					incompatibleUnitMessage ? (
-						<Tooltip title={incompatibleUnitMessage}>
-							<SolidAlertTriangle role="img" aria-label="warning" size="md" />
+						<Tooltip
```
AshwinBhatkal, line 85: this v1 component?

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/sections/ThresholdsSection/__tests__/ThresholdsSection.test.tsx
```diff
@@ -70,19 +76,34 @@ describe('ThresholdsSection', () => {
 		]);
 	});
 
-	it('does not commit edits when Discard is clicked', () => {
+	it('reflects edits live (before Save) so the preview can react', () => {
 		const onChange = jest.fn();
 		render(<ThresholdsSection value={THRESHOLDS} onChange={onChange} />);
 
+		fireEvent.click(screen.getByTestId('threshold-edit-0'));
```
AshwinBhatkal: userEvent

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/sections/ThresholdsSection/ThresholdsSection.tsx
```diff
@@ -82,6 +82,15 @@ function ThresholdsSection({
 	// Which row is being edited, and whether it was just added (so Discard removes it).
 	const [editingIndex, setEditingIndex] = useState<number | null>(null);
 	const [unsavedIndex, setUnsavedIndex] = useState<number | null>(null);
+	// The saved threshold captured on edit entry, restored if the edit is discarded
+	// (edits stream into the spec live, so Discard can't just drop a local draft).
+	const editSnapshot = useRef<AnyThreshold | null>(null);
+
+	const updateAt =
```
AshwinBhatkal, line 89: no need of debounce? Perf is good?

frontend/src/components/YAxisUnitSelector/YAxisUnitSelector.tsx
ahrefabhi: yes

ahrefabhi: this is used across multiple places, not just dashboards

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/sections/ThresholdsSection/ThresholdsSection.tsx
ahrefabhi: made the changes, debouced

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/sections/ThresholdsSection/__tests__/ThresholdsSection.test.tsx
ahrefabhi: done

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/sections/ThresholdsSection/rows/shared/useThresholdDraft.ts
```diff
@@ -7,6 +8,8 @@ interface ThresholdDraft<T> {
 	setValue: (raw: string) => void;
 }
 
+const LIVE_PREVIEW_DEBOUNCE_MS = 150;
```
AshwinBhatkal, line 11: nit: may not have be required

AshwinBhatkal (approved):

PR #11923, opened 2026-07-01
fix: opening /ai-assistant opens 2 new conversations instead of one

YounixM:
## Pull Request

---

Navigating to `/ai-assistant` was creating **two** blank conversations instead of one. Users saw an extra throwaway thread in the sidebar — briefly during load, or persistently after startup route churn.

The page effect treated any bare or unknown `conversationId` as “start fresh” and always called `startNewConversation()`. That was wrong when:

1. **Startup route churn** — the page remounts at `/ai-assistant` while an active conversation already exists in the store.
2. **Persisted rehydration** — `onRehydrateStorage` primes the active thread as a hydrating placeholder before `fetchThreads` resolves; the page minted a second blank thread alongside it.
3. **Stale URL params** — e.g. `/ai-assistant/new` with an active conversation already set.

The fix prefers **resuming the active conversation** (including hydrating placeholders) and only calls `startNewConversation()` when there is genuinely nothing to resume. URL sync behavior is unchanged for valid conversation IDs.

N/A — behavior fix; covered by unit tests. Manual check: open `/ai-assistant` and confirm exactly one thread appears in the sidebar (no flash of a second blank thread during load).

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
`AIAssistantPage` resolved the URL param with a binary check: known `conversationId` → activate; otherwise → `startNewConversation()`. It ignored `activeConversationId` and rehydrating placeholders, so remounts and load-time hydration each spawned an extra blank conversation.

#### Fix Strategy
When the URL has no usable conversation id, read `activeConversationId` from the store (via a ref to avoid effect re-fires on store mutation). If that id exists in `conversations`, resume it and `history.replace` to the canonical URL. Only mint a new conversation when there is no active conversation to resume.

---

- Tests added/updated:
  - `frontend/src/pages/AIAssistantPage/__tests__/AIAssistantPage.test.tsx` (new)
    - Opens exactly one conversation at `/ai-assistant`
    - Does not stack a second conversation on page remount (route churn)
    - Activates an existing conversation from the URL without creating a new one
    - Resumes active conversation on `/ai-assistant/new` instead of minting
    - Resumes persisted hydrating conversation during load instead of creating a duplicate
- Manual verification:
  - Navigate to `/ai-assistant` and confirm a single thread in the sidebar
  - Reload / remount and confirm no duplicate blank threads
- Edge cases covered:
  - Bare URL, stale `new` param, rehydrating placeholder, valid URL param

---

- Blast radius: `AIAssistantPage` URL ↔ store sync only; drawer/modal/panel flows unchanged.
- Potential regressions: First visit with an empty store should still create exactly one new conversation; explicit “New chat” via sidebar should still mint a new thread (handled by `handleNewConversation`, not this effect).
- Rollback plan: Revert the commit; no migrations or API changes.

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Fixed AI Assistant opening two blank conversations when navigating to `/ai-assistant`. |

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

- The effect intentionally depends only on `conversationId`; refs hold latest `conversations` and `activeConversationId` to avoid re-firing when the store mutates mid-effect.
- “New chat” from the conversations list still goes through `handleNewConversation` → `startNewConversation()`; this change only affects automatic resolution when the URL has no valid id.


frontend/src/pages/AIAssistantPage/AIAssistantPage.tsx
H4ad, line 23: We should avoid useLocation when possible, this re-renders for even the tiny changes in the URL.

H4ad (approved):

makeavish (approved):

PR #11925, opened 2026-07-01
fix: context not being passed when user click suggestions in empty state

YounixM:
## Pull Request

---

When users open Noz from an embedded surface (panel/modal) on a page such as a dashboard and click an empty-state suggestion chip, the message was sent **without** the auto-derived page context (dashboard ID, service, explorer, etc.). Manually typing the same prompt via `ChatInput` correctly attached that context.

This happened because `onSendSuggestedPrompt` called `handleSend(text)` with only the chip text, while the typed-send path passes `autoContexts` as the third argument.

The fix forwards the same `autoContexts` array used by `ChatInput` when a chip is clicked. On the standalone AI Assistant page (`variant === 'page'`), contexts remain `undefined` since there is no underlying page to attach.

https://github.com/user-attachments/assets/6948e18d-279d-4e9a-904b-1001c5fb0d17

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
`ConversationView` wires empty-state chips through `VirtualizedMessages` → `onSendSuggestedPrompt`, which invoked `handleSend(text)` without the optional `contexts` parameter. `ChatInput` already passed `autoContexts` on typed sends, so the two send paths were inconsistent.

#### Fix Strategy
Update `onSendSuggestedPrompt` to call `handleSend(text, undefined, autoContexts.length > 0 ? autoContexts : undefined)`, matching the typed-send contract and respecting dismissed auto-context entries (filtered via `autoContexts`).

---

- Tests added/updated:
  - `frontend/src/container/AIAssistant/ConversationView/__tests__/ConversationView.test.tsx`
    - Embedded (`panel`) variant: chip click forwards page `autoContexts` to `sendMessage`
    - Standalone (`page`) variant: chip click sends `undefined` contexts
- Manual verification:
  - Open Noz panel on a dashboard → click an empty-state suggestion → confirm request includes dashboard context
  - Open standalone `/ai-assistant` page → click suggestion → confirm no page context is attached
- Edge cases covered:
  - Standalone page variant (no auto-context)
  - User-dismissed auto-contexts (uses filtered `autoContexts`, not raw `allAutoContexts`)

---

- Blast radius: AI Assistant empty-state chip sends only (`ConversationView`)
- Potential regressions: Unlikely — change only adds context to a path that previously omitted it; typed sends and non-chip flows are unchanged
- Rollback plan: Revert commit; chip sends return to context-less behavior

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Fixed Noz empty-state suggestion chips not including page context when opened from dashboards and other embedded views |

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

- The fix is intentionally scoped to the chip send path; `ChatInput` already handled typed sends correctly.
- `autoContexts` respects user dismissals and URL changes — same semantics as manual input.
- Branch name has a typo (`propogation`); no functional impact.


frontend/src/container/AIAssistant/ConversationView/__tests__/ConversationView.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+jest.mock('container/AIAssistant/components/StreamingMessage', () => ({
+	__esModule: true,
+	default: (): null => null,
+}));
+
+jest.mock('react-virtuoso', () => ({
+	__esModule: true,
+	Virtuoso: (): null => null,
+}));
```
H4ad, line 70: react+logEvent I think we already mock globally,

but try create a better way to test these components, I'm seeing a lot of mocks in those tests, we can centralize or think in another way to consolidate them to be reused across the tests.

H4ad (approved):

makeavish (approved):

PR #11929, opened 2026-07-01
feat(dashboards-v2): substitute dashboard variables when creating an alert from a panel

ahrefabhi:
## Pull Request

---

### 📄 Summary
Creating an alert from a V2 dashboard panel shipped the panel's query **verbatim** — any dashboard-variable references (`$service`, `{{.env}}`, dynamic `__all__`) landed in the alert builder unresolved, so the seeded alert didn't match what the panel was actually showing.

This wires the `/substitute_vars` round-trip into the V2 create-alert flow (V1 parity with `useCreateAlerts`), now that V2 has the variable plumbing (`resolvedVariables` store channel from #11909). When the dashboard has variable selections, `useCreateAlertFromPanel` builds a V5 query-range request (panel queries + resolved variables) and POSTs it through the generated `useReplaceVariables` hook; on success the substituted envelopes are translated to the V1 `Query` the alert page reads. When there are no variables to substitute the round-trip would be a no-op, so we keep seeding synchronously — no network dependency and the new tab stays tied to the click (no async popup-blocker surprise).

Supporting refactors (no behavior change to existing callers):
- `persesQueryAdapters`: extracted `envelopesToQuery` from `fromPerses` (the substitute response hands back envelopes, not panel-query wrappers).
- `buildCreateAlertUrl`: exported `buildAlertUrl` + `readPanelUnit` so the sync and substituted paths share URL assembly.

N/A — no visual change; the difference is the query values carried into `/alerts/new`.

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/5555

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context
N/A

---

### 🧪 Testing Strategy

- Tests added/updated:
  - `useCreateAlertFromPanel.test.ts` — no-variable sync path, variable substitution path (request built, mutation invoked, navigation deferred until success), and the error path (`toast.error`, no navigation).
  - `persesQueryAdapters.test.ts` — `envelopesToQuery` (empty list default + envelope → query type derivation); existing `fromPerses` cases still green (output unchanged).
- Manual verification: `tsgo --noEmit`, `oxlint`, and `pnpm build` all clean; full suite for the touched areas passes (create-alert hook, buildCreateAlertUrl, persesQueryAdapters, PanelActionsMenu, ConfigActions).
- Edge cases covered: dashboard with no variables (skip round-trip), substitution API failure, empty/missing `compositeQuery.queries` in the response.

---

- Blast radius: V2 dashboards only — the panel actions menu "Create Alerts" item and the panel-editor "Create alert" action. `fromPerses` output is unchanged, so nothing else that consumes it is affected.
- Potential regressions: the substituted path opens the new tab inside the async `onSuccess` callback (V1 parity) — strict popup blockers could block it. If that surfaces, the fix is to open a placeholder tab synchronously on click and redirect after substitution.
- Rollback plan: revert the commit; the no-variable path is identical to the prior behavior.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Alerts created from a dashboard panel now resolve dashboard variables to their selected values (matching the panel), instead of carrying raw variable references. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- Uses the **generated** `useReplaceVariables` (`api/generated/services/querier`, → `/api/v5/substitute_vars`), not the deprecated hand-written `getSubstituteVars` — per the "V2 uses the generated API contract" convention.
- The skip-when-empty branch is a deliberate optimization (and popup-blocker safeguard), not a divergence from V1's resolved output — with zero variables the backend substitution is a pass-through.
- Redux global time (ns → ms) feeds the request window; it's irrelevant to substitution but the request DTO requires a valid window.

---

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/queryV5/persesQueryAdapters.ts
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/queryV5/persesQueryAdapters.ts
@@ -74,14 +74,14 @@ export function deriveQueryType(
 }
 
 /**
- * Perses panel queries → V1 `Query` (to seed the query builder), via the V5 envelope
- * list + `mapQueryDataFromApi`. An empty panel opens on a fresh metrics builder query.
+ * V5 query-envelope list → V1 `Query`, via `mapQueryDataFromApi`. An empty list opens
+ * on a fresh metrics builder query. Used by `fromPerses` and by the envelopes a
+ * `/substitute_vars` round-trip returns with dashboard variables resolved.
  */
-export function fromPerses(
-	queries: DashboardtypesQueryDTO[],
+export function envelopesToQuery(
+	envelopes: Querybuildertypesv5QueryEnvelopeDTO[],
 	panelType: PANEL_TYPES,
 ): Query {
-	const envelopes = toQueryEnvelopes(queries);
 	if (envelopes.length === 0) {
 		return initialQueriesMap[DataSource.METRICS];
 	}
@@ -97,6 +97,17 @@ export function fromPerses(
 	};
 
 	return mapQueryDataFromApi(composite);
+}
+
+/**
+ * Perses panel queries → V1 `Query` (to seed the query builder), via the V5 envelope
+ * list + `mapQueryDataFromApi`. An empty panel opens on a fresh metrics builder query.
+ */
+export function fromPerses(
+	queries: DashboardtypesQueryDTO[],
+	panelType: PANEL_TYPES,
+): Query {
+	return envelopesToQuery(toQueryEnvelopes(queries), panelType);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/queryV5/persesQueryAdapters.ts
```diff
@@ -74,14 +74,14 @@ export function deriveQueryType(
...
 /**
- * Perses panel queries → V1 `Query` (to seed the query builder), via the V5 envelope
- * list + `mapQueryDataFromApi`. An empty panel opens on a fresh metrics builder query.
+ * V5 query-envelope list → V1 `Query`, via `mapQueryDataFromApi`. An empty list opens
+ * on a fresh metrics builder query. Used by `fromPerses` and by the envelopes a
+ * `/substitute_vars` round-trip returns with dashboard variables resolved.
  */
-export function fromPerses(
-	queries: DashboardtypesQueryDTO[],
+export function envelopesToQuery(
```
AshwinBhatkal, line 81: is this split mainly required for the test?

AshwinBhatkal (approved):

frontend/src/pages/DashboardPageV2/DashboardContainer/queryV5/persesQueryAdapters.ts
ahrefabhi: yes

PR #11836, opened 2026-06-24
Fix dashboard source OpenAPI schema

yukimotochern:
Fixes the generated OpenAPI schema for `dashboardtypes.Source`.

It serializes as a string enum, but the reflector saw the private `valuer.String` field and emitted `type: object`. This exposes the schema explicitly and regenerates `docs/api/openapi.yml`.

Checks:
- `GOTOOLCHAIN=go1.25.7+auto go run cmd/enterprise/*.go generate openapi`
- `npx --yes openapi-typescript@7.8.0 docs/api/openapi.yml -o /tmp/signozApi.d.ts`
- `GOTOOLCHAIN=go1.25.7+auto go test ./pkg/types/dashboardtypes`

PR #11930, opened 2026-07-01
fix(dashboards): expose Source as a string enum in the OpenAPI schema

therealpandey:
## Summary

`dashboardtypes.Source` serializes as a string enum but the reflector saw its unexported `valuer.String` field and emitted `type: object` in the OpenAPI schema. This adds a `JSONSchema()` exposer that pins `type: string`, and regenerates `docs/api/openapi.yml`.

The enum values are derived from the existing `Enum()` method rather than being hardcoded again, so the list of sources lives in exactly one place — adding a new source only requires editing `Enum()`.

### DIFF
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -3571,7 +3571,7 @@ components:
       - user
       - system
       - integration
-      type: object
+      type: string
     DashboardtypesSpanGaps:
       properties:
         fillLessThan:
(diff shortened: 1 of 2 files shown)


therealpandey:
Closes https://github.com/SigNoz/signoz/pull/11836

therealpandey:
Blocking https://github.com/SigNoz/terraform-provider-signoz/pull/113

srikanthccv (approved):

Issue #11822, opened 2026-06-23
Cannot open navbar links in new tab


jvllmr:
## Bug description

When I right-click a link in the navbar I expect "Open link in new tab" to be an option.
I don't know if this is really a bug or rather a feature, but I filed a bug since it is the behavior I'd expect from a link.

<img width="434" height="367" alt="Image" src="https://github.com/user-attachments/assets/c2ab8eae-4dfa-4656-90de-e6d3d048a43d" />

("Link in neuem Tab öffnen" would be the german translation for "Open link in new tab")

## Expected behavior
I can open the link in a new tab

## How to reproduce

1. Right-click on navbar link
2. See that there is no option to open link in new tab

## Version information 
* **Signoz version**: 0.129.0
* **Browser version**: Firefox 152.0.1
* **Your OS and version**: Docker
* **Your CPU Architecture**(ARM/Intel): Intel

## Additional context
I have experience with react + react-router and would like to implement a PR myself if there is interest :)

atharva-again:
Are you saying that in the sidebar, if you for example right click on Infrastructure, in the context menu you want to see "Open Link In New Tab"?

jvllmr:
> Are you saying that in the sidebar, if you for example right click on Infrastructure, in the context menu you want to see "Open Link In New Tab"?

Yes, that is exactly what I mean.

H4ad:
@jvllmr Thanks for opening the issue, the current implementation indeed does not allow to open as a link, but you can `Ctrl + Left Click`, it will open a in new tab.

I marked this as a feature request so we can improve this eventually and make it a link. 

> I have experience with react + react-router and would like to implement a PR myself if there is interest :)

Certainly, I have already reviewed the code for the navigation bar, but I’m not sure if there was a specific reason for keeping it as a button instead of a link. In any case, if you're interested, feel free to explore, contributions are very welcome!

Harsh23Kashyap:
I'd like to work on this. I'll open a PR shortly.

Harsh23Kashyap:
Opened the PR here: https://github.com/SigNoz/signoz/pull/11877

PR #11934, opened 2026-07-01
feat(frontend): allow opening sidebar nav items in new tab

justin212407:
### 📄 Summary

The left sidebar navigation rendered all nav items as `<div>` elements with `onClick` handlers calling `useNavigate()`. This broke standard browser behaviors that technical users rely on daily - right-click → "Open in new tab", middle-click, Ctrl/Cmd+click, and copying link addresses all silently did nothing.

This PR replaces those `<div>` elements with React Router `<Link>` components, which render as real `<a href="...">` tags in the DOM. This is the correct approach because `<Link>` preserves SPA routing (no full page reload on normal click) while restoring all native browser behaviors for free, with no extra event-handling logic needed.

**Scope:** Sidebar navigation only - primary nav items, pinned shortcuts, "More" section items, brand logo, and "Get Started" button. Dashboard cards, Services list, Alerts list, and trace/span links are follow-up PRs.

https://github.com/user-attachments/assets/a6923935-9744-4d0b-80c4-38e1f810ddec

#### Issues closed by this PR
Fixes part of #11822

---
### ✅ Change Type
_Select all that apply_
- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---
#### Root Cause
Sidebar nav items were implemented as `<div>` elements with imperative `onClick` → `history.push()` navigation. The browser has no way to know these are navigable links, so right-click context menus, middle-click, and modifier-key clicks all silently do nothing. This is a structural issue, the element type itself is wrong for the intended behavior.

#### Fix Strategy
- Added an optional `href` prop to `NavItem`. When provided and the item is not disabled, it renders as `<Link to={href}>` instead of `<div>`, giving the DOM a real `<a>` tag with a valid `href`.
- Added `getNavItemHref()` in `SideNav.tsx` to compute the correct URL for each item at render time (reusing the existing `buildNavUrl` + `getQueryString` logic already used for keyboard shortcuts).
- `handleMenuItemClick` is simplified to analytics logging only, navigation is now handled natively by the browser via `<Link>`.
- Non-route items (Quick Search → CmdK modal) receive `href={undefined}` and remain as `<div>` elements intentionally.
- Anchor reset styles added to both SCSS files so `<a>` tags are visually identical to the previous `<div>` styling (no underline,
  inherited color, all states covered including `:visited`).

---
### 🧪 Testing Strategy

- Tests added/updated: None in this PR (can add tests in case needed)
- Manual verification:
  - Right-click → "Open link in new tab" works on all primary nav items 
  - Middle-click opens new tab 
  - Ctrl+Click opens new tab 
  - Normal left-click still does SPA navigation (no full page reload)
  - Active-state highlight follows current route correctly 
  - Quick Search still opens CmdK modal, no href exposed 
  - Keyboard Tab/Enter navigation intact, focus outlines visible 
  - Brand logo right-clickable with correct href 
- Edge cases covered:
  - Disabled items (workspace blocked): fall through to `<div>` path, `event.preventDefault()` guard also present inside Link onClick
  - Pin/unpin icon clicks: `stopPropagation` + `preventDefault` prevent the pin click from triggering Link navigation
  - AI Assistant item: href computed dynamically based on active conversation ID

---

- Blast radius: Sidebar navigation only. No backend changes, no API changes, no data model changes.
- Potential regressions:
  - Visual: `<a>` tags could pick up global anchor styles if any exist outside the reset rules added, mitigated by the explicit resets on`.nav-item`, `.brand-logo`, and `.get-started-btn` covering all pseudo-states (`:hover`, `:focus`, `:active`, `:visited`)
  - Analytics: `logEvent` calls are preserved in the `onClick` prop on `<Link>`, so tracking is unaffected
  - Keyboard shortcuts: `onClickHandler` (used by `registerShortcut`) is unchanged and still calls `history.push()` directly
- Rollback plan: Revert the 4 changed files. No migrations or infrastructure changes involved.

---
### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Sidebar nav items now render as real anchor elements, enabling right-click → "Open in new tab", middle-click, and Ctrl/Cmd+click across all primary nav, shortcuts, and "More" section items |

---
### 📋 Checklist
- [x] Manually tested

---
## 👀 Notes for Reviewers

- Quick Search is intentionally left as a `<div>`, it triggers the CmdK modal and has no route URL to link to. It would not make sense as an anchor element.
- Help & Support and Settings dropdown items are out of scope, they live inside Radix dropdown portals and follow a different interaction
  pattern. They can be addressed separately if desired.
- The `getNavItemHref()` function reuses `buildNavUrl` + `getQueryString` which are the same utilities already used by keyboard shortcuts and `onClickHandler`-  no new URL-building logic was introduced.

### DIFF
--- a/frontend/src/container/SideNav/NavItem/NavItem.styles.scss
+++ b/frontend/src/container/SideNav/NavItem/NavItem.styles.scss
@@ -9,6 +9,18 @@
 	margin-bottom: 4px;
 	cursor: pointer;
 
+	// Reset anchor defaults when rendered as a <Link> (i.e. <a> tag)
+	text-decoration: none;
+	color: inherit;
+
+	&:hover,
+	&:focus,
+	&:active,
+	&:visited {
+		text-decoration: none;
+		color: inherit;
+	}
+
 	&.active {
 		.nav-item-active-marker {
 			background: var(--primary-background);
--- a/frontend/src/container/SideNav/NavItem/NavItem.tsx
+++ b/frontend/src/container/SideNav/NavItem/NavItem.tsx
@@ -2,6 +2,7 @@ import { Tooltip } from 'antd';
 import { Badge } from '@signozhq/ui/badge';
 import cx from 'classnames';
 import { Pin, PinOff } from '@signozhq/icons';
+import { Link } from 'react-router-dom';
 
 import { SidebarItem } from '../sideNav.types';
 
@@ -16,15 +17,18 @@ export default function NavItem({
 	isPinned,
 	showIcon,
 	dataTestId,
+	href,
 }: {
 	item: SidebarItem;
 	isActive: boolean;
-	onClick: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => void;
+	onClick: (event: React.MouseEvent) => void;
 	isDisabled: boolean;
 	onTogglePin?: (item: SidebarItem) => void;
 	isPinned?: boolean;
 	showIcon?: boolean;
 	dataTestId?: string;
+	/** When provided, the nav item renders as a <Link> with a real href. */
+	href?: string;
 }): JSX.Element {
 	const { label, icon, isBeta, isNew, isEarlyAccess, tooltip } = item;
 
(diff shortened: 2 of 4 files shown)


justin212407:
@H4ad can you please review the changes and lmk if there is anything else required that i might need to add or might be good to have in the pr (according to the existing conventions of the codebase)

PR #11937, opened 2026-07-02
fix(dashboards): render logs/traces list panels on public dashboards

AshwinBhatkal:
## Pull Request

> **Stacked on #11935** (public-dashboard query-cache fix). Review/merge that first — this branch is based on it, so its diff shows only the list-rendering change once #11935 lands.

---

### 📄 Summary

On the public dashboard viewer (`/public/dashboard/:id`), **logs/traces list panels rendered nothing** even though the query range API returned data — the issue reporter confirmed the response came back but the UI didn't draw it. It renders fine on the authenticated dashboard.

Root: `ListPanelWrapper` returns an empty fragment when `setRequestData` is `undefined`, and the public `Panel` never forwarded a `setRequestData` to `WidgetGraphComponent`. So the list wrapper bailed before rendering.

Fix: back `requestData` with `useState` and forward the setter, mirroring the authenticated `GridCard`. The per-panel request builder is extracted to `utils.ts`. Because the public payload redacts each widget's `limit`, the list components would otherwise always show a pager that can't page (the public endpoint takes no pagination params) — so a `hidePagination` flag is threaded through to suppress it on public list panels.

#### Issues closed by this PR

Closes https://github.com/SigNoz/engineering-pod/issues/3646

---

### ✅ Change Type

- [x] 🐛 Bug fix
- [x] 🧪 Test-only (added coverage)

---

#### Root Cause

`ListPanelWrapper` guards `if (!setRequestData) return <></>` (the setter drives list pagination). The authenticated `GridCard` always passes it; the public `Panel` computed `requestData` with `useMemo` and passed no setter, so logs/traces list panels never rendered despite valid data.

#### Fix Strategy

- Move `requestData` from `useMemo` to `useState` so there's a real setter, and forward `setRequestData` to `WidgetGraphComponent`.
- Extract the request-data builder into `utils.ts` (`getPublicPanelRequestData`).
- Dropped the public-only `pageSize: 10` query mutation so the initial fetch matches the authenticated view (it never reached the index-based backend anyway).
- Thread an optional `hidePagination` flag (`WidgetGraphComponent` → `PanelWrapper` → `ListPanelWrapper` → `LogsPanelComponent`/`TracesTableComponent`); default off, so authenticated dashboards are unchanged.
- Guard `useLogsData` against a missing `orderBy`: the public payload redacts the query (`orderBy`/`filter`/`limit` stripped), so once the logs panel actually renders, `listQuery?.orderBy.find(...)` threw (`?.` guarded `listQuery` but not `orderBy`). Changed to `listQuery?.orderBy?.find(...)`.

---

### 🧪 Testing Strategy

- **Tests added:** `PublicDashboardContainer/__tests__/Panel.test.tsx` — asserts `setRequestData` is forwarded to `WidgetGraphComponent`.
- **Manual verification:** typecheck (`tsgo --noEmit`), `oxlint`, and `oxfmt --check` clean; full `PublicDashboardContainer` suite green.

---

- **Blast radius:** Public-dashboard viewer, plus an additive optional `hidePagination` prop on the shared list components (defaults off → authenticated path unchanged).
- **Potential regressions:** Low. The pager-hide is gated behind a flag only the public `Panel` sets. Functional public pagination remains a separate backend task (endpoint must accept page params).
- **Rollback plan:** Revert the commits; isolated to `PublicDashboardContainer` and the additive prop.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Public dashboards now render logs/traces list-view panels. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- Stacked on #11935; the first commit here is that PR's fix.

### DIFF
--- a/frontend/src/container/GridCardLayout/GridCard/WidgetGraphComponent.tsx
+++ b/frontend/src/container/GridCardLayout/GridCard/WidgetGraphComponent.tsx
@@ -66,6 +66,7 @@ function WidgetGraphComponent({
 	customOnRowClick,
 	customTimeRangeWindowForCoRelation,
 	enableDrillDown,
+	hidePagination,
 }: WidgetGraphComponentProps): JSX.Element {
 	const { safeNavigate } = useSafeNavigate();
 	const [deleteModal, setDeleteModal] = useState(false);
@@ -430,6 +431,7 @@ function WidgetGraphComponent({
 						customSeries={customSeries}
 						customOnRowClick={customOnRowClick}
 						enableDrillDown={enableDrillDown}
+						hidePagination={hidePagination}
 						onColumnWidthsChange={onColumnWidthsChange}
 					/>
 				</div>
--- a/frontend/src/container/GridCardLayout/GridCard/types.ts
+++ b/frontend/src/container/GridCardLayout/GridCard/types.ts
@@ -42,6 +42,8 @@ export interface WidgetGraphComponentProps {
 	customOnRowClick?: (record: RowData) => void;
 	customTimeRangeWindowForCoRelation?: string | undefined;
 	enableDrillDown?: boolean;
+	/** Hide list-panel pagination controls (e.g. public dashboards, where paging isn't supported). */
+	hidePagination?: boolean;
 }
 
 export interface GridCardGraphProps {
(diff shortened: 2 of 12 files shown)


ahrefabhi (approved):
aren't we doing these changes in dashboard v2?

AshwinBhatkal:
Common component is used in v2 as well @ahrefabhi. Only API changes

tewarig (approved):

PR #11942, opened 2026-07-02
feat(dashboard-v2): variable editor form, add-variable UX & styling

AshwinBhatkal:
## Summary
Variable editor form, add-variable UX, and toolbar/empty-state styling for the V2 dashboard (dark-shipped behind `use_dashboard_v2`). Builds on the variable fetch engine and the panel-linking work.

**Variable editor form**
- Restore the V1 combined field/source row for dynamic variables (field on the left, `from` + telemetry source on the right).
- Cap the "Preview of values" box height.
- Raise the dynamic-variable attribute search debounce to 500 ms.

**Add-variable UX**
- Add-variable button in the variables bar: full-width labelled button when no variables exist; a `+` icon (label on hover) once they do. It opens the settings drawer straight to the Variables tab in add mode (via a transient `settingsRequest` store slice + `destroyOnClose` remount).
- The `+` sits after the `+N`/Less trigger, kept inline and always mounted so the overflow measurement never toggles it.

**Layout & styling**
- The variables bar flows under the floated time selector again (block wrapper, so the strip wraps around the float; expanded rows drop full-width below it).
- Flat `l3` borders on the Actions / Configure / JSON toolbar buttons (no depth/shadow); `l3` dashed border on the empty state.
- Add a "Configure your new dashboard" step to the empty state, opening the settings drawer on the Overview tab.

## Test plan
- Manual (dev, flag on): add/edit variables via the drawer; add-variable button in both empty and populated bars; collapse/expand the bar and confirm it flows under the time selector without layout shift; empty-state Configure button opens the drawer.

Stacked PR — base: `feat/dashboard-v2-link-variables-panels` (merge the two upstream PRs first).

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/VariablesBar/VariablesBar.tsx
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/VariablesBar/VariablesBar.tsx
@@ -1,5 +1,5 @@
 import { useState } from 'react';
-import { ChevronLeft, Plus } from '@signozhq/icons';
+import { ChevronLeft } from '@signozhq/icons';
 import { Button } from '@signozhq/ui/button';
 import { TooltipSimple } from '@signozhq/ui/tooltip';
 import cx from 'classnames';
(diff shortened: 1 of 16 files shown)


ahrefabhi (dismissed):

frontend/src/pages/DashboardPageV2/DashboardContainer/VariablesBar/VariablesBar.tsx
```diff
@@ -74,6 +116,14 @@ function VariablesBar({ dashboard }: VariablesBarProps): JSX.Element | null {
 		</Button>
 	);
 
+	if (variables.length === 0) {
+		return (
+			<div className={styles.bar} data-testid="dashboard-variables-bar">
+				{addVariableFull}
```
AshwinBhatkal: use the whole component instead of this pattern. Same for the other one

ahrefabhi (dismissed):

ahrefabhi (approved):

PR #11895, opened 2026-06-29
feat(dashboards-v2): panel drilldown — View in Logs/Traces

ahrefabhi:
## Pull Request

---

### 📄 Summary

Ports V1's panel **drill-down** to V2 dashboards. Clicking a data point (time-series, bar, pie, value, or table) opens a context menu with **View in Logs / View in Traces**, which open the corresponding explorer in a new tab. Dashboard-variable references (`$var`) in the panel query are **substituted** before the explorer URL is built, matching V1.

> **Scope:** this is the base of the drill-down stack — the context menu + View in Logs/Traces + variable substitution. Table filter-by-value, Breakout, context links, and the dashboard-variables submenu are stacked follow-ups (see _Notes for Reviewers_).

**Approach:** each renderer turns its native click into a single `DrilldownClickPayload` via pure enrichment utils (`enrichChartClick` / `enrichTableClick` / `enrichNumberClick` / `enrichPieClick`); a kind-agnostic `useDrilldown` owns the popover and reuses V1's read-only navigation helpers (`useBaseDrilldownNavigate`, …) — **no V1 files are modified** (V1/V2 split policy). Drill-down is declared per-kind as a capability (`actions.drilldown`, matching V1 coverage — Histogram/List opt out). Variable substitution reuses the V2-native `/substitute_vars` round-trip (`useResolvedDrilldownQuery` → `useReplaceVariables` + `buildQueryRangeRequest`), mirroring `useCreateAlertFromPanel`; while resolving, the menu actions show a spinner and are disabled.

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/5486

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

---

### 🧪 Testing Strategy

- **Tests added/updated:** unit tests for the enrichment utils (`drilldown.test.ts`), `getPanelTimeRange`, and `useDrilldown` orchestration (gating, aggregate-menu render, navigate, disabled-while-resolving). `tsgo --noEmit` clean; drill-down jest suite green.
- **Manual verification:** _pending (author)_.
- **Edge cases covered:** formula / PromQL queries (no drill), panels without a builder query (disabled), multi-query Number panels (drills the *displayed* value's query), no-variable dashboards (skip the substitution round-trip).

---

### ⚠️ Risk & Impact Assessment

[...]

PR #11945, opened 2026-07-02
feat(dashboards-v2): persist the View modal's query edits across refresh

ahrefabhi:
## What

The View modal's in-modal query builder is URL-synced (`compositeQuery`, plus `graphType` when present), but the modal seeded its draft from the **saved panel** on mount — so a page refresh discarded any in-modal edits.

This change seeds the modal draft from the URL when a query is present (falling back to the saved panel otherwise), so in-modal edits are shareable and survive a refresh — matching V1 parity.

## Changes

- **`useViewPanelEditor.ts`** — mount-only read-seed: builds `initialPanel` from `compositeQuery` + `graphType` via `buildViewPanelSpec` when a URL query is present, else the saved panel. Reset still targets the query the view opened with.
- **`buildViewPanelSpec.ts`** (+test) — bakes a V1 query + target panel type into a View-modal spec, switching the panel kind via the editor's `getSwitchedPluginSpec` when the type maps to a different kind.
- **`useViewPanel.ts`** — `openView` now clears leftover `compositeQuery`/`graphType` so a plain View opens on the saved panel, not a stale URL query.
- **`panelKind.ts`** — adds `PANEL_TYPE_TO_PANEL_KIND` (inverse of `PANEL_KIND_TO_PANEL_TYPE`), needed by `buildViewPanelSpec`.

## Context

Extracted from the drilldown branch so the URL-state layer lands as its own change on top of the View-mode PR. The panel drilldown PR (#11895) stacks on top of this.

## Verification

- `tsgo --noEmit` — clean
- `oxlint` (files + project) — clean
- `buildViewPanelSpec.test.ts` + `ViewPanelModal.test.tsx` — pass
- `vite build` — clean

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/drilldown/buildViewPanelSpec.ts
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/drilldown/buildViewPanelSpec.ts
@@ -1,0 +1,52 @@
+import type {
+	DashboardtypesPanelPluginDTO,
+	DashboardtypesPanelSpecDTO,
+	TelemetrytypesSignalDTO,
+} from 'api/generated/services/sigNoz.schemas';
+import type { PANEL_TYPES } from 'constants/queryBuilder';
+import { getSwitchedPluginSpec } from 'pages/DashboardPageV2/DashboardContainer/PanelEditor/getSwitchedPluginSpec';
+import {
+	PANEL_TYPE_TO_PANEL_KIND,
+	type PanelKind,
+} from 'pages/DashboardPageV2/DashboardContainer/Panels/types/panelKind';
+import { toPerses } from 'pages/DashboardPageV2/DashboardContainer/queryV5/persesQueryAdapters';
+import type { Query } from 'types/api/queryBuilder/queryBuilderData';
+
+import { getBuilderQueries } from '../getBuilderQueries';
+
+/**
+ * Bakes a V1 query + target panel type into a View-modal spec (drilldown seed + URL re-hydration).
+ * When `panelType` maps to a different kind than the panel's (e.g. a breakout turns Value → Table),
+ * the kind is switched via the editor's `getSwitchedPluginSpec` so it opens with populated config.
+ */
+export function buildViewPanelSpec({
+	spec,
+	query,
+	panelType,
+}: {
+	spec: DashboardtypesPanelSpecDTO;
+	query: Query;
+	panelType: PANEL_TYPES;
+}): DashboardtypesPanelSpecDTO {
+	const queries = toPerses(query, panelType);
+	const currentKind = spec.plugin.kind as PanelKind;
+	const newKind = PANEL_TYPE_TO_PANEL_KIND[panelType] ?? currentKind;
+
+	if (newKind === currentKind) {
+		return { ...spec, queries };
+	}
+
+	// The plugin cast mirrors the editor's type-switch — a dynamically chosen kind can't be
+	// correlated with its spec statically.
+	const signal = getBuilderQueries(spec.queries ?? [])[0]
+		?.signal as TelemetrytypesSignalDTO;
+	return {
+		...spec,
+		plugin: {
+			...spec.plugin,
+			kind: newKind,
+			spec: getSwitchedPluginSpec(spec, newKind, signal),
+		} as DashboardtypesPanelPluginDTO,
+		queries,
+	};
+}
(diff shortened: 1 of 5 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/drilldown/buildViewPanelSpec.ts
```diff
@@ -0,0 +1,52 @@
...
+	const currentKind = spec.plugin.kind as PanelKind;
+	const newKind = PANEL_TYPE_TO_PANEL_KIND[panelType] ?? currentKind;
+
+	if (newKind === currentKind) {
+		return { ...spec, queries };
+	}
+
+	// The plugin cast mirrors the editor's type-switch — a dynamically chosen kind can't be
+	// correlated with its spec statically.
+	const signal = getBuilderQueries(spec.queries ?? [])[0]
```
AshwinBhatkal: queries is always array right?

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/ViewPanelModal/useViewPanelMode.ts
```diff
@@ -4,14 +4,19 @@ import type {
 	DashboardtypesPanelSpecDTO,
 	TelemetrytypesSignalDTO,
 } from 'api/generated/services/sigNoz.schemas';
+import { QueryParams } from 'constants/query';
+import { PANEL_TYPES } from 'constants/queryBuilder';
+import { useGetCompositeQueryParam } from 'hooks/queryBuilder/useGetCompositeQueryParam';
 import { useQueryBuilder } from 'hooks/queryBuilder/useQueryBuilder';
+import useUrlQuery from 'hooks/useUrlQuery';
```
AshwinBhatkal, line 11: I may have missed this earlier. But let's use nuqs for this

AshwinBhatkal: Check out usage in infra monitoring
https://github.com/SigNoz/signoz/blob/20dd264ac1decce651f2e5bc66101502b2a3acd5/frontend/src/container/InfraMonitoringK8s/hooks.ts

AshwinBhatkal: Could be done in a whole another PR across all files now touched

AshwinBhatkal: But create ticket for it

frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/drilldown/buildViewPanelSpec.ts
ahrefabhi: Done

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/ViewPanelModal/useViewPanelMode.ts
ahrefabhi: I tried migrating these to `nuqs` and it doesn't work cleanly here — sharing why before we track it as follow-up.

The hook logic itself is fine in isolation. But it breaks in the running app — the modal's `expandedWidgetId` never survives a reload because it gets clobbered.

Root cause is the adapter/router split:

- We mount the **generic** `nuqs/adapters/react` adapter (`src/index.tsx`), not the react-router one, and we're on **react-router v5**.
- `nuqs` writes the URL through the **native History API**; RR v5 keeps its own location state and never sees those writes.
- On the dashboard route, `expandedWidgetId` has to coexist in the URL with `compositeQuery` / `graphType`, which are written by the shared query builder via **react-router** (`redirectWithQueryBuilderData`) while the View modal is open.
- So the two systems overwrite each other: `nuqs` sets `expandedWidgetId` → the next RR write (e.g. an in-modal query edit writing `compositeQuery`) rebuilds the URL from RR's state, which never had `expandedWidgetId`, and drops it.

The existing `nuqs` usages (InfraMonitoring, alerts) work because their params live entirely within `nuqs` — nothing on those routes writes the same params through RR. The dashboard is different: `compositeQuery` is RR-managed and shared app-wide.

A proper migration would mean adopting the react-router `nuqs` adapter (`nuqs/adapters/react-router`, v5-compat) **app-wide** so `nuqs` and RR share one history — which touches every existing `nuqs` feature and needs its own regression pass. Keeping the `useUrlQuery` + `safeNavigate` approach here for now.

AshwinBhatkal (approved):

PR #11974, opened 2026-07-04
fix(dashboards-v2): list panel pagination, spanGaps clamp, discard dialog & create-alert fixes

ahrefabhi:
## Pull Request

---

A batch of V2 dashboard panel fixes, mostly around the **List panel's server pagination**, plus a few smaller panel fixes. Each is a self-contained commit:

1. **Show the list panel pager** — a new/edited List panel never showed its pager. The shared V1→V5 mapper folds a list query's internal `pageSize` into the V5 `limit`, so a seeded panel always looked like it had a user limit, and `usePanelQuery` suppresses the pager when a limit is present. Fixed by stripping the V1 explorer's `pageSize`/`offset` in `toPerses` before conversion.
2. **Deterministic pagination order** — offset paging over a non-total order can duplicate/drop same-millisecond rows across pages. Every logs-list request now carries a total order (`timestamp desc, id`), request-only (logs-explorer parity).
3. **List pagination runtime behaviour** — `canNext` is driven by the backend `nextCursor` (not a row-count guess); the table + pager stay mounted on a page change (`keepPreviousData`); a stale cross-type response kept across a panel-kind switch no longer flashes "No data"; and the body shows horizontal **skeleton rows** while the next page loads (dashboard grid + editor preview).
4. **Order on panel-type switch** — switching to a List panel left the builder's Order By empty; it now seeds `timestamp desc` (parity with a fresh list panel).
5. **Time-series `spanGaps`** — a numeric threshold is now floored at the step interval so a sub-step value doesn't break the line at every normal point.
6. **Discard dialog** — "Keep editing" moved to the right and made outlined; "Discard" stays left (destructive).
7. **Create-alert time window** — floored the ns→ms conversion (V5 start/end are int64 ms) and used the named `NANO_SECOND_MULTIPLIER`.
8. **New-panel save on an empty dashboard** — creating the first panel failed with an `items overlap` error: the optimistic patch applier aliased op values by reference, mutating the outgoing ops so the panel was added to the layout twice. Fixed by cloning inserted values in `applyJsonPatch`.

Verified manually: the pager now appears for logs List panels, page changes show skeleton rows (no NoData flash, no stale rows), switching to a list keeps `timestamp desc` in Order By, and the discard dialog shows `[Discard] [Keep editing]` with Keep editing outlined. (Screenshots shared during review.)

Closes https://github.com/SigNoz/pulse-pod/issues/90

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
- **Pager never showed:** the shared V1→V5 mapper (`createBaseSpec`) computes `limit = limit || pageSize` for list/table, so the seed's internal `pageSize: 100` became a V5 `limit`, which the pager gate reads as an explicit user limit.
- **Dup/dropped rows & unreliable "next":** the backend emits `ORDER BY` verbatim (no auto tiebreaker), so `timestamp desc` alone is not a total order across separate offset queries.
- **NoData flash on kind switch:** `keepPreviousData` (added to keep the pager on page change) also retains the *previous* panel type's response across a query-key change; the new renderer can't read it.
- **Empty Order By on switch:** V1's `handleQueryChange` clears `orderBy` when switching to a list.
- **spanGaps:** a numeric threshold below one step interval breaks the line at every normal point.
- **New-panel double-add:** `applyJsonPatch` (run in `onMutate` before the request is sent) inserted each `op.value` by reference, so the cache's `layouts[0].spec.items` aliased the empty `items` array held inside the section-add op; applying the item op then mutated that op's value, and react-query sent a layout that already held the item *plus* the separate item-add op → two overlapping items. New-dashboard-only, since that is the only batch that adds a layout and an item into that same layout together.

#### Fix Strategy
- Strip `pageSize`/`offset` at the V2 conversion boundary (`toPerses`) so `limit` reflects only a real user limit.
- Enforce a total order (`timestamp desc, id`) at request-build time, request-only.
- Drive `canNext` from `nextCursor`; gate the full-panel loader on first-load only; require a retained response's type to match the current request before treating it as data.
- Re-seed the list default order in `usePanelTypeSwitch`.
- Clamp the numeric `spanGaps` to `max(minStepInterval, value)`.
- Clone the inserted value on `add`/`replace` in `applyJsonPatch` so the applied document never shares references with the input ops (restoring the purity the docstring already promises).

---

- **Tests added/updated:** unit tests for `toPerses` (pageSize→limit strip), `buildQueryRangeRequest` (id tiebreaker + defaults), `usePanelQuery` (cursor `canNext`, `keepPreviousData`), `PanelBody` (first-load loader vs mounted refetch, cross-type stale response), `ListPanel/Renderer` (skeleton rows), `usePanelTypeSwitch` (default list order), `buildConfig` (spanGaps clamp), and `applyJsonPatch` (a later op targeting a just-added node must not mutate the input ops).
- **Manual verification:** ran the app and exercised list paging (next/prev, page-size change), panel-type switching, the discard dialog, and creating the first panel on a brand-new dashboard.
- **Edge cases covered:** last page (partial page → no cursor), traces list (no `id` tiebreaker, explorer parity), boolean `spanGaps` (span-all passes through), first fetch with no data (loader, not skeleton).
- `tsgo --noEmit`, `oxlint`, and `pnpm build` all green; pre-commit hook validated each commit.

---

- **Blast radius:** V2 dashboards only. Mostly the List panel; the spanGaps change is Time-series-only; the discard-dialog and create-alert changes are editor-only; the `applyJsonPatch` clone affects every optimistic spec mutation.
- **Potential regressions:** the `PanelBody` loader-gate change (`isFetching && !hasData`) affects **all V2 panels** — a same-key background refetch now keeps the chart mounted instead of showing a full-panel spinner (time-range changes still show the loader). The request-time `id` tiebreaker only affects logs list requests and is not persisted. The `applyJsonPatch` change only deep-clones values already destined for the cache, so it is behaviour-preserving apart from removing the aliasing.
- **Rollback plan:** each fix is an isolated commit and can be reverted independently.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | V2 dashboard List panels now show the pager, page deterministically, and display skeleton rows while loading the next page; switching to a list keeps timestamp ordering; time-series spanGaps respects the step interval; discard-dialog button layout updated; and creating the first panel on a new dashboard no longer fails with an items-overlap error. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- The **list pagination runtime** fixes (cursor `canNext`, keep-pager, no-flash, skeletons) are in a single commit because they share `usePanelQuery.ts` / `PanelBody.tsx`; splitting them by hunk broke the pre-commit hook's project-wide `tsgo` under lint-staged's partial-stage stashing. Every commit compiles in isolation.
- The **create-alert** commit (`useCreateAlertFromPanel.ts`) was a pre-existing working-tree change, included at request.
- `@signozhq/ui/skeleton` isn't in the package's export map, so the skeleton rows use antd's `Skeleton.Input`.
- The **new-panel double-add** fix is in the optimistic applier, not the op builder: `createPanelOps` was already correct (it emits an empty section + a separate item add); the corruption came from `applyJsonPatch` aliasing the op values while writing the optimistic cache.

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/queryV5/v5ResponseData.ts
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/queryV5/v5ResponseData.ts
@@ -2,6 +2,7 @@ import type {
 	Querybuildertypesv5AggregationBucketDTO,
 	Querybuildertypesv5ExecStatsDTO,
 	Querybuildertypesv5RawDataDTO,
+	Querybuildertypesv5RequestTypeDTO,
 	Querybuildertypesv5ScalarDataDTO,
 	Querybuildertypesv5TimeSeriesDataDTO,
 	Querybuildertypesv5TimeSeriesDTO,
@@ -42,6 +43,13 @@ export function getRawResults(
 		return [];
 	}
 	return (data.data?.results ?? []) as Querybuildertypesv5RawDataDTO[];
+}
+
+/** Response request-type discriminator (raw/trace/scalar/time_series); detects a stale cross-type response. */
+export function getResponseType(
+	response: QueryRangeV5200 | undefined,
+): Querybuildertypesv5RequestTypeDTO | undefined {
+	return response?.data?.type;
 }
 
 /** Exec stats (incl. per-query `stepIntervals`) from the response top level. */
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/Header/Header.tsx
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/Header/Header.tsx
@@ -1,7 +1,7 @@
 import { useCallback } from 'react';
 import { SolidAlertTriangle, X } from '@signozhq/icons';
 import { Button } from '@signozhq/ui/button';
-import { ConfirmDialog } from '@signozhq/ui/dialog';
+import { DialogWrapper } from '@signozhq/ui/dialog';
 import { Divider } from '@signozhq/ui/divider';
 import { Typography } from '@signozhq/ui/typography';
 import { useConfirmableAction } from 'hooks/useConfirmableAction';
@@ -61,24 +61,42 @@ function Header({
 				</Button>
 			</div>
 
-			<ConfirmDialog
+			<DialogWrapper
 				open={discard.open}
-				onOpenChange={(next): void => {
+				onOpenChange={(next: boolean): void => {
 					if (!next) {
 						discard.cancel();
 					}
 				}}
 				title="Discard changes?"
 				titleIcon={<SolidAlertTriangle size={14} color="#fdd600" />}
-				confirmText="Discard"
-				confirmColor="destructive"
-				cancelText="Keep editing"
-				onConfirm={discard.confirm}
-				onCancel={discard.cancel}
-				data-testid="panel-editor-v2-discard-modal"
+				testId="panel-editor-v2-discard-modal"
+				footer={
+					<>
+						<Button
+							type="button"
+							variant="solid"
+							color="destructive"
+							data-testid="panel-editor-v2-discard-confirm"
+							loading={discard.isPending}
+							onClick={discard.confirm}
+						>
+							Discard
+						</Button>
+						<Button
+							type="button"
+							variant="outlined"
+							color="secondary"
+							data-testid="panel-editor-v2-discard-cancel"
+							onClick={discard.cancel}
+						>
+							Keep editing
+						</Button>
+					</>
+				}
 			>
 				<Typography>Your unsaved edits to this panel will be lost.</Typography>
-			</ConfirmDialog>
+			</DialogWrapper>
 		</div>
 	);
 }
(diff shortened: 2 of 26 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/queryV5/v5ResponseData.ts
```diff
@@ -44,6 +45,13 @@ export function getRawResults(
 	return (data.data?.results ?? []) as Querybuildertypesv5RawDataDTO[];
 }
 
+/** Response request-type discriminator (raw/trace/scalar/time_series); detects a stale cross-type response. */
+export function getResponseType(
```
AshwinBhatkal, line 49: see if this can be typed better. Feels loose loose

AshwinBhatkal (approved):

frontend/src/pages/DashboardPageV2/DashboardContainer/queryV5/v5ResponseData.ts
ahrefabhi: noted, doing it in the next PR

PR #11977, opened 2026-07-05
feat(authz): make serviceaccount_role and user_role first class citizens

vikrantgupta25:
### 📄 Summary
- make serviceaccount_role and user_role first class citizens

closes: https://github.com/SigNoz/platform-pod/issues/2600


pkg/apiserver/signozapiserver/serviceaccount.go
```diff
@@ -421,3 +504,53 @@ func (provider *provider) roleSelector(ctx context.Context, resource coretypes.R
 		resource.Type().MustSelector(coretypes.WildCardSelectorString),
 	}, nil
 }
+
+func (provider *provider) roleIDExtractor() coretypes.ResourceIDExtractor {
```
vikrantgupta25, line 508: these extractors need a home. module is not the place ( mux http deps ), types is definitely not the place again ( module deps ). Handler ? Maybe ( no issues but the resourceDef lies here, its better to keep it close to the resource def )

pkg/apiserver/signozapiserver/serviceaccount.go
```diff
@@ -398,13 +398,96 @@ func (provider *provider) addServiceAccountRoutes(router *mux.Router) error {
...
+			SuccessStatusCode:   http.StatusNoContent,
+			ErrorStatusCodes:    []int{http.StatusBadRequest, http.StatusNotFound},
+			Deprecated:          false,
+			SecuritySchemes:     newScopedSecuritySchemes([]string{coretypes.ResourceServiceAccount.Scope(coretypes.VerbDetach), coretypes.ResourceRole.Scope(coretypes.VerbDetach)}),
+		},
+		handler.WithResourceDefs(handler.AttachDetachSiblingResourceDef{
+			Verb:           coretypes.VerbDetach,
+			Category:       coretypes.ActionCategoryAccessControl,
+			SourceResource: coretypes.ResourceServiceAccount,
+			SourceIDs:      coretypes.OneID(provider.serviceAccountIDExtractor()),
```
vikrantgupta25, line 478: this API will require 4 DB calls to gather the authz context now

vikrantgupta25: doesn't seem like the way to go for future APIs 🤔

pkg/http/middleware/authz.go
```diff
@@ -216,6 +216,11 @@ func (middleware *AuthZ) CheckResources(next http.HandlerFunc, roles ...string)
 		}
 
 		for _, resource := range resolved {
+			if err := resource.Err(); err != nil {
```
vikrantgupta25, line 219: this is handled here instead of resource middleware so that audit records it.

therealpandey (approved):

H4ad (approved):

PR #11979, opened 2026-07-06
chore: add integration tests for metrics under reduction - query part

srikanthccv:
## Pull Request

querier part of https://github.com/SigNoz/engineering-pod/issues/5535


tests/fixtures/clickhouse.py
```diff
@@ -265,6 +320,334 @@ def restore(cache: dict) -> types.TestContainerClickhouse:
...
+    return create_clickhouse(
+        tmpfs=tmpfs,
+        network=network,
+        keeper=zookeeper,
+        request=request,
+        pytestconfig=pytestconfig,
+    )
+
+
+def local_series_counts(
```
therealpandey: @srikanthccv can we move this out of this file?

tests/fixtures/clickhouse.py
```diff
@@ -265,6 +320,334 @@ def restore(cache: dict) -> types.TestContainerClickhouse:
...
+            conn.query(
+                f"SELECT count(DISTINCT fingerprint) FROM signoz_metrics.{table} WHERE metric_name = %(metric_name)s",
+                parameters={"metric_name": metric_name},
+            ).result_rows[0][0]
+        )
+        for conn in node_conns
+    ]
+
+
+def assert_spans_shards(
```
therealpandey: @srikanthccv same here

tests/fixtures/http.py
```diff
@@ -18,19 +18,22 @@
 logger = setup_logger(__name__)
 
 
-@pytest.fixture(name="zeus", scope="package")
-def zeus(
+ZEUS_NETWORK_ALIAS = "signoz-zeus-it"
```
therealpandey: I understand this was done to speed up local tests but can we remove this unrelated change?

tests/fixtures/clickhouse.py
```diff
@@ -265,6 +320,334 @@ def restore(cache: dict) -> types.TestContainerClickhouse:
...
+                <hostname>localhost</hostname>
+                <port>9234</port>
+            </server>
+        </raft_configuration>
+    </keeper_server>
+</clickhouse>
+"""
+
+
+def create_clickhouse_keeper(
```
therealpandey: Can you move this into a `keeper.py` file and get rid of zookeeper altogether?

therealpandey (changes requested):

tests/fixtures/clickhouse.py
srikanthccv: does the `tests/fixtures/metricreduction.py` work?

tests/fixtures/http.py
srikanthccv: yeah, I can do that. If we could find some way to speed up the local test that would greatly help.

tests/fixtures/clickhouse.py
srikanthccv: sure, I originally did remove the zookeeper but then to avoid scope creep retained it. Will do this.

tests/fixtures/clickhouse.py
srikanthccv: moved to `tests/fixtures/metricreduction.py`

tests/fixtures/clickhouse.py
srikanthccv: done

tests/fixtures/http.py
srikanthccv: reverted

tests/fixtures/clickhouse.py
srikanthccv: updated

therealpandey (approved):

PR #11985, opened 2026-07-06
chore(querier): address impl gap in max_concurrent_queries

srikanthccv:
### 📄 Summary

`max_concurrent_queries` has existed in the querier config since the beginning, but it was never wired into execution. The v5 querier ran a request's queries strictly one at a time, and the only place with any parallelism (bucket-cache missing-range queries) used a hardcoded limit of 4. This PR addresses that implementation gap.

A query range request now runs its queries concurrently, bounded by a per-request semaphore of `max_concurrent_queries` (default 8). The same limit covers the missing-range queries in `executeWithCache`, replacing the hardcoded 4 so one number bounds the total ClickHouse concurrency a request can generate. 

#### Issues closed by this PR

Closes SigNoz/engineering-pod#4960


pkg/querier/querier.go
```diff
@@ -67,25 +69,30 @@ func New(
...
+		auditStmtBuilder:         auditStmtBuilder,
+		metricStmtBuilder:        metricStmtBuilder,
+		meterStmtBuilder:         meterStmtBuilder,
+		traceOperatorStmtBuilder: traceOperatorStmtBuilder,
+		bucketCache:              bucketCache,
+		liveDataRefresh:          5 * time.Second,
 		builderConfig: builderConfig{
 			logTraceIDWindowPaddingMS: uint64(logTraceIDWindowPadding.Milliseconds()),
 		},
+		maxConcurrentQueries: maxConcurrentQueries,
```
tushar-signoz, line 95: nit: Can this be part of builderConfig ?

srikanthccv: not really a builder related config option so no I don't think it should be on builder config. additionally this config option pre-dates all other options just that it was not implemented.

therealpandey (approved):

PR #11989, opened 2026-07-06
fix(dashboard-v2): warn about orphaned panels & surface JSON save errors

AshwinBhatkal:
## Summary

Two robustness fixes for the dashboard JSON editor drawer (dark-shipped behind `use_dashboard_v2`):

1. **Warn about orphaned panels.** Editing the JSON can leave panels in `spec.panels` that no layout references — e.g. deleting a layout entry (or the whole `layouts` array). Those panels stay in the document but render in no section. The drawer now detects them and shows a footer warning (e.g. *"2 panels not placed in any section — they won't be shown after saving."*) so it isn't a silent data trap. Non-blocking: you can still save if that's intended.

2. **Surface save errors.** A failed save (e.g. editing a **locked** dashboard's JSON — the API rejects it) passed the raw generated error to the error modal cast as `APIError`, so no real message rendered. Route it through `toAPIError` so the actual API error is shown.

## Screenshots

image

Closes https://github.com/SigNoz/pulse-pod/issues/93

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/JsonEditorDrawer.tsx
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/JsonEditorDrawer.tsx
@@ -60,6 +60,7 @@ function JsonEditorDrawer({
 
 	const onKeyDown = useCallback(
 		(event: KeyboardEvent<HTMLDivElement>): void => {
+			event.stopPropagation();
 			if ((event.metaKey || event.ctrlKey) && event.key === 'Enter') {
 				event.preventDefault();
 				void apply();
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/JsonEditorDrawer.tsx
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/JsonEditorDrawer.tsx
@@ -1,6 +1,8 @@
 import { KeyboardEvent, useCallback } from 'react';
 import MEditor from '@monaco-editor/react';
+import { TriangleAlert } from '@signozhq/icons';
 import { Button } from '@signozhq/ui/button';
+import { TooltipSimple } from '@signozhq/ui/tooltip';
 import { Typography } from '@signozhq/ui/typography';
 import cx from 'classnames';
 import { Drawer } from 'antd';
@@ -26,8 +28,18 @@ function JsonEditorDrawer({
 }: JsonEditorDrawerProps): JSX.Element {
 	const [, copyToClipboard] = useCopyToClipboard();
 
-	const { draft, setDraft, validity, isDirty, isSaving, format, reset, apply } =
-		useJsonEditor({ dashboard, isOpen, onApplied: onClose });
+	const {
+		draft,
+		setDraft,
+		validity,
+		isDirty,
+		isSaving,
+		danglingPanelIds,
+		missingPanelRefs,
+		format,
+		reset,
+		apply,
+	} = useJsonEditor({ dashboard, isOpen, onApplied: onClose });
 
 	const onCopy = useCallback((): void => {
 		copyToClipboard(draft);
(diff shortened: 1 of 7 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/danglingPanels.ts
```diff
@@ -0,0 +1,41 @@
...
+	content?: { $ref?: string };
+}
+
+interface LayoutEnvelope {
+	kind?: string;
+	spec?: { items?: GridItemRef[] };
+}
+
+interface SpecShape {
+	panels?: Record<string, unknown>;
```
AshwinBhatkal: This will always be required. Use existing type

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/danglingPanels.ts
```diff
@@ -0,0 +1,41 @@
...
+	layouts?: LayoutEnvelope[];
```
AshwinBhatkal: This will always be required as well. You can draw SpecShape from the DashboardType and pick

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/JsonEditorDrawer.tsx
```diff
@@ -60,6 +69,12 @@ function JsonEditorDrawer({
 	const validationText = validity.valid
 		? `Valid JSON · ${validity.lineCount} lines`
 		: `Line ${validity.errorLine ?? '?'} · ${validity.message ?? 'Invalid JSON'}`;
+	const danglingWarning =
+		danglingPanelIds.length > 0
+			? `${danglingPanelIds.length} panel${
+					danglingPanelIds.length === 1 ? '' : 's'
+				} not placed in any section — they won't be shown after saving.`
```
AshwinBhatkal: will not be shown after saving as they are not present in layouts

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/danglingPanels.ts
AshwinBhatkal: Done — now typed against `DashboardtypesDashboardSpecDTO` / `DashboardtypesLayoutDTO` instead of hand-rolled shapes (3cb50873bb).

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/danglingPanels.ts
AshwinBhatkal: Done — `findPanelLayoutIssues` now takes a `DashboardtypesDashboardSpecDTO` (drawn from the dashboard DTO) rather than a local optional shape.

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/JsonEditorDrawer.tsx
AshwinBhatkal: Reworded to "not present in any section — they won't be shown after saving." Also added the reverse check (layout refs to a panel that no longer exists) as its own warning.

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/danglingPanels.ts
```diff
@@ -0,0 +1,47 @@
...
+	// Panel ids a layout item references that no longer exist in `spec.panels`.
+	missingPanelRefs: string[];
+}
+
+const referencedPanelIds = (
+	layouts: DashboardtypesLayoutDTO[],
+): Set<string> => {
+	const referenced = new Set<string>();
+	layouts.forEach((layout) => {
+		if (layout?.kind !== 'Grid') {
```
ahrefabhi, line 20: kind should be available, no need for ?, i think

AshwinBhatkal: It's json editing, so kind can be removed from the layout

ahrefabhi (approved):

PR #11994, opened 2026-07-07
fix(retention): persist TTL error message on failure

### LINKED ISSUE #1822: Store error information in TTL status table
Currently when TTL fails we log the error detail but don't store error detail in the TTL status table. 
Storing error detail in case of TTL failures helps in debugging.

Harsh23Kashyap:
## Pull Request

---

### 📄 Summary

> Why does this change exist? What problem does it solve?

When a TTL or cold-storage apply fails today, the failure reason is
written to the application log but never persisted. Users hitting a
stuck "TTL running" state have no way to see why without SSH access
to the query-service container.

This PR persists the failure message in a new `error_message` column
on `ttl_setting` and surfaces it on the existing
`GET /api/v2/settings/ttl` response, so the UI can show it next to
the failed status row.

#### Issues closed by this PR
- Closes #1822

---

### ✅ Change Type

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause

`pkg/query-service/app/clickhouseReader/reader.go` writes the TTL
status (`pending`, `success`, `failed`) back to the `ttl_setting`
row on every state transition, but the failure paths only call
`r.logger.Error(...)` and never store the underlying message. The
three `setTTL{Logs,Traces,Metrics}` functions each had a copy of the
same 14-line `Update` block, and the custom retention path
(`SetTTLV2`) had its own helper that had the same gap.

#### Fix Strategy

- Add an `error_message` column to `ttl_setting` via migration `099`
  (nullable `text`, idempotent against existing rows).
- Extend the `TTLSetting` Go struct with `ErrorMessage` and the
  `GetCustomRetentionTTLResponse` API response with `error_message`.
- Add a private `updateTTLStatus(ctx, statusItem, status, errMsg...)`
  helper on `ClickHouseReader`. Variadic `errMsg` keeps the success
  callers (`updateTTLStatus(ctx, item, Success)`) unchanged.
- Replace the three repeated Update blocks in
  `setTTLLogs` / `setTTLTraces` / `setTTLMetrics` with calls to the
  helper, passing `err.Error()` on the failure branches.
- Update the existing `updateCustomRetentionTTLStatus` helper the
  same way and pass `err.Error()` from the two failure sites in
  `SetTTLV2`.
- Wire `error_message` through `GetCustomRetentionTTL` so the API
  exposes it.

---

### 🧪 Testing Strategy

- Tests added/updated: none at the clickhouseReader level (no test
  infrastructure exists for it; the file does not yet have a
  `_test.go` sibling). Behaviour is covered by the focused
  build/vet passes below.
- Manual verification:
  - `go build ./pkg/types/retentiontypes/... ./pkg/sqlmigration/...
    ./pkg/query-service/app/clickhouseReader/... ./pkg/signoz/...` →
    pass
  - `go vet` on the same packages → pass
  - `go test ./pkg/types/retentiontypes/... ./pkg/sqlstore/...` →
    pass
- Edge cases covered:
  - Success path: `error_message` is left untouched (variadic
    `errMsg...` is empty so the column is not included in the
    UPDATE). Existing success rows keep their (NULL) value.
  - Failure path: `err.Error()` is written atomically with the
    `status = 'failed'` set.
  - Nil `statusItem`: helper is a no-op, same as the previous inline
    code path.

---

- Blast radius: query-service TTL paths only. No public API
  contract change — the new field is omitempty. The new column is
  nullable so old rows are valid as-is.
- Potential regressions: if a downstream consumer (alert manager,
  webhook, third-party) was relying on the absence of the
  `error_message` key, that is unlikely because the response is
  parsed by SigNoz clients first. The new column adds disk space
  per TTL row (a few hundred bytes per failed apply).
- Rollback plan: migration `099` Down is empty (the existing
  `048` and others follow this pattern). Reverting the code change
  alone leaves the column unused; existing rows remain valid.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | OSS |
| Change Type | Bug Fix |
| Description | TTL failures now persist the error message in `ttl_setting.error_message`, returned by `GET /api/v2/settings/ttl`, so users see why a TTL change failed without grepping logs. |

---

### 📋 Checklist

- [x] Tests added or explicitly not required (no test infra in this package; existing touched packages pass)
- [x] Manually tested (build, vet, focused go test suites)
- [x] Backward compatibility considered (nullable column, omitempty API field)

---

### 👀 Notes for Reviewers

The diff is net `-69` lines because the helper absorbed three
~14-line Update blocks. The `updateCustomRetentionTTLStatus`
signature gains a variadic `errMsg ...string` which is backwards
compatible with the existing two callers in this PR (success paths
still pass nothing).

### DIFF
--- a/pkg/query-service/app/clickhouseReader/reader.go
+++ b/pkg/query-service/app/clickhouseReader/reader.go
@@ -1038,21 +1038,8 @@ func (r *ClickHouseReader) setTTLLogs(ctx context.Context, orgID string, params 
 			err := r.setColdStorage(context.Background(), tableName, params.ColdStorageVolume)
 			if err != nil {
 				r.logger.Error("error in setting cold storage", errorsV2.Attr(err))
-				statusItem, apiErr := r.checkTTLStatusItem(ctx, orgID, tableName)
-				if apiErr == nil {
-					_, dbErr := r.
-						sqlDB.
-						BunDB().
-						NewUpdate().
-						Model(new(retentiontypes.TTLSetting)).
-						Set("updated_at = ?", time.Now()).
-						Set("status = ?", retentiontypes.TTLSettingStatusFailed).
-						Where("id = ?", statusItem.ID.StringValue()).
-						Exec(ctx)
-					if dbErr != nil {
-						r.logger.Error("Error in processing ttl_status update sql query", errorsV2.Attr(dbErr))
-						return
-					}
+				if statusItem, apiErr := r.checkTTLStatusItem(ctx, orgID, tableName); apiErr == nil {
+					r.updateTTLStatus(ctx, statusItem, retentiontypes.TTLSettingStatusFailed, err.Error())
 				}
 				return
 			}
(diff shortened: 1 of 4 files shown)


Harsh23Kashyap:
Can someone take a look when you have a chance?

Harsh23Kashyap:
Can someone take a look when you have a chance?

Harsh23Kashyap:
Can someone take a look when you have a chance?

Harsh23Kashyap:
CI still pending on the merge queue. No new reviewer comments since the last update. (TTL error persist)

Harsh23Kashyap:
TTL error persist fix. CI pending, no new review threads. Standing by.

Harsh23Kashyap:
Synced with upstream. TTL error persist. Synced (8→0).

PR #11997, opened 2026-07-07
fix(docs): correct broken relative links in contributing docs

saicharan62:
Noticed a few doc links pointing to files that had been moved or renamed, so I went through and fixed them. While I was at it, I checked the rest of the repo for the same problem, turned out these were the only three.


saicharan62:
Hi @AshwinBhatkal ,  this is a small docs fix (broken links in the e2e/contributing docs) that's been open since a while without review, could you take a look when you get a chance?

PR #11999, opened 2026-07-07
feat(dashboard-v2): extend time window from empty panel state

ahrefabhi:
## Pull Request

---

When a V2 dashboard panel has **no data in the selected time window**, the empty state offered a **Retry** button. Retrying can't help — the query succeeded, there's simply nothing in range. The useful next step is to **look at a wider window**.

This PR replaces Retry with an **"Extend to last `<preset>`"** action (or **"Widen time range"** for a custom drag-zoomed window):

- **Relative window** → jumps to the next-larger time-picker preset, chosen by the window's *actual* span (so off-ladder values like `3h` work too).
- **Custom window** → grows 3× around its centre (capped at now), keeping the user's historical anchor instead of throwing them to a now-relative range.
- **Context-aware:** on the grid and editor preview it widens the dashboard's global time (dispatch + URL sync, mirroring the time selector). In the **View modal** — which owns a local, in-place time window and never touches global time or the URL — it widens *that* local window. The modal publishes its extender through a transient dashboard-store slice so the deeply-nested `NoData` reads it from the store rather than threading a prop through `PreviewPane → PanelBody → every renderer`.

It also fixes a related empty-state glitch: because `keepPreviousData` retains the previous (empty) response, the `NoData` state **lingered through the refetch and then snapped to the chart** when data arrived. Now a refetch over empty data shows the shared full-panel loader instead.

Before
image

After
image

#### Issues closed by this PR

Closes https://github.com/SigNoz/pulse-pod/issues/98

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
`PanelBody`'s `hasData` is true whenever a type-matching response exists. With react-query `keepPreviousData`, a time change retains the previous **empty** response, so `hasData` stayed true, the first-fetch loader branch was skipped, and the (empty) renderer kept showing `NoData` for the whole refetch — then abruptly swapped to the chart when the new data landed. Emptiness is only known per-renderer, so `PanelBody` alone can't tell an empty retained response from a populated one.

#### Fix Strategy
Move the loader decision to where emptiness is known: `NoData` now takes `isFetching` (threaded from each renderer) and renders the shared `PanelLoader` instead of the empty message while a fetch is in flight. Refetches over *real* data still keep the renderer mounted (no flash), so only the empty case changes.

---

- Tests added/updated:
  - `useExtendTimeWindow.test.tsx` — next-preset selection, dispatch + URL sync, widest-preset no-op, custom-window widening.
  - `NoData.test.tsx` — extend action (not Retry), store-extender-beats-global precedence, no action when not extendable, loader shown while `isFetching`.
  - `useViewPanelTimeWindow.test.ts` — in-place relative + custom widening via `onTimeChange`.
  - `ViewPanelModal.test.tsx` — mock updated for the new `extendWindow`.
- Manual verification: `pnpm tsgo --noEmit`, `pnpm oxlint`, and `pnpm build` all clean. **Not yet driven in a live browser** (no running stack in this environment) — behaviour is verified via the state logic + unit tests.
- Edge cases covered: already-widest window (no action), degenerate/zero-span custom window, off-ladder relative values, custom window not extended past "now".

---

- Blast radius: V2 dashboards only (`DashboardPageV2`). Touches the shared `NoData` empty state + all 7 panel renderers + `PanelBody` + the View modal.
- Potential regressions: the empty-state affordance and the first-fetch/refetch loader behaviour. The `refetch` prop was removed from the renderer contract (it was only wired to the old Retry); `PanelBody` keeps its own `refetch` for the real error-state Retry, which is unchanged.
- Known tradeoff: the View modal's extender lives in a single store slot, so while the modal is open a grid panel behind it that is also empty reads the modal's extender (behind the backdrop, unclickable, cleared on close) — same "one active" semantics as the store's existing `refetch`.
- Rollback plan: revert the PR; no schema, API, or persisted-state changes (the store slice is transient/not persisted).

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | Cloud / OSS |
| Change Type | Feature |
| Description | On V2 dashboards, an empty panel now offers to extend/widen the time window instead of a no-op Retry, and shows a loader while re-fetching a widened range. |

---

### 📋 Checklist
- [ ] Manually tested (unit-tested + build-verified; live-browser check pending a running stack)
- [x] Breaking changes documented (none; internal renderer-prop cleanup only)

---

## 👀 Notes for Reviewers

- **Why Zustand, not a prop / React Context:** Context is lint-banned in this codebase ("migrate to Zustand"), and prop-threading `extendWindow` through `PreviewPane → PanelBody → all 7 renderers` was a lot of plumbing for one deeply-nested consumer. The transient `viewSessionSlice` follows the same pattern as the existing `editContextSlice` (which stores `refetch`).
- **Widening ladder** lives in `Panels/components/NoData/extendWindow.ts` and mirrors the global picker presets so the resulting selection renders as a named range.
- Happy to key the store slot by `panelId` if you'd prefer the modal's extender strictly scoped rather than the current single-slot approach.


frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/components/NoData/useExtendTimeWindow.ts
```diff
@@ -0,0 +1,27 @@
+// eslint-disable-next-line no-restricted-imports -- global time lives in redux; V2 time dispatch is not yet migrated
+import { useSelector } from 'react-redux';
+import { useZoomOut } from 'hooks/useZoomOut';
+import { getNextZoomOutRange } from 'lib/zoomOutUtils';
+import { AppState } from 'store/reducers';
+import { GlobalReducer } from 'types/reducer/globalTime';
+
+import { buildExtendWindow, ExtendTimeWindow } from './extendWindow';
+
+const NS_PER_MS = 1e6;
```
AshwinBhatkal: This constant is defined somewhere else?

frontend/src/pages/DashboardPageV2/DashboardContainer/store/slices/viewSessionSlice.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export const createViewSessionSlice: StateCreator<
+	DashboardStore,
+	[['zustand/persist', unknown]],
+	[],
+	ViewSessionSlice
+> = (set) => ({
+	viewPanelExtendWindow: null,
+	setViewPanelExtendWindow: (viewPanelExtendWindow): void => {
```
AshwinBhatkal: why not derive it from the url? 🤔

AshwinBhatkal (commented):
1 small comment

frontend/src/pages/DashboardPageV2/DashboardContainer/store/slices/viewSessionSlice.ts
ahrefabhi: this is for the view mode, if I derive it from the url, on every time window change I'll end up running the panels behind the view modal as well

frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/components/NoData/useExtendTimeWindow.ts
ahrefabhi: replaced

frontend/src/pages/DashboardPageV2/DashboardContainer/store/slices/viewSessionSlice.ts
AshwinBhatkal: hmmm. If it is view mode, I don't see the need for it in the dashboard store slice. This can be a separate store in itself. Also, how is the time being handled now for the view mode?

ahrefabhi: made it a separate store

frontend/src/pages/DashboardPageV2/DashboardContainer/store/useViewSessionStore.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Kept out of `useDashboardStore` (never persisted): the View modal publishes its
+ * local-window extender here so the deeply-nested `NoData` state reads it without
+ * prop threading. Set while the modal is open, cleared on close.
+ */
+export interface ViewSessionStore {
+	viewPanelExtendWindow: ExtendTimeWindow | null;
+	setViewPanelExtendWindow: (extendWindow: ExtendTimeWindow | null) => void;
+}
+
+export const useViewSessionStore = create<ViewSessionStore>((set) => ({
```
AshwinBhatkal: useViewPanelStore? 🤔

ahrefabhi: Done

AshwinBhatkal (approved):

PR #12000, opened 2026-07-07
feat(infra-monitoring): counts query for counting related entities

### LINKED ISSUE #5469: Alerts based on deviations from baseline

### LINKED ISSUE #5470: Consider `materialize_ttl_after_modify=0` for TTL modification

nikhilmantri0902:
## Pull Request

---

### 📄 Summary

Adds per-cluster and per-namespace counts of related resources to the v2 infra-monitoring list APIs.

New generic helper `getPerGroupDistinctCounts` counts, per groupBy combination, the distinct values of any set of attributes over a metric universe — modeled on `getMetadata` (fingerprints from samples, labels from the timeseries table, raw-only or raw+reduced union under `FeatureEnableMetricsReduction`, user filter merged with page-groups). It runs as one query via `uniqExactIf(JSONExtractString(labels, attr), != '')` per attribute, wired into `ListClusters`/`ListNamespaces` as an errgroup goroutine.

- **Clusters** (`counts`): `nodes`, `namespaces`, `deployments`, `daemonSets`, `jobs`, `statefulSets`
- **Namespaces** (`counts`): `deployments`, `daemonSets`, `jobs`, `statefulSets`

Metric universe is the pod list (owner labels live on pod series); clusters additionally union node metrics to count nodes. Counts reflect entities that reported metrics in the window.

#### Issues closed by this PR
Closes #5469
Closes #5470

---

### ✅ Change Type

- [x] ✨ Feature

---

### 🧪 Testing Strategy

- Tests added/updated: extended `04_namespaces.py` / `05_clusters.py` accuracy integration tests to assert `counts`; enriched existing accuracy datasets with owner-workload + second-namespace labels (no new datasets).
- Manual verification: `go build ./...` green; query semantics simulated against enriched data — expected counts match exactly.
- Edge cases covered: missing attr → `0` (via `!= ''` guard); distinct dedup across raw+reduced union; multiple distinct workloads per group.

---

- Blast radius: additive `counts` field on clusters/namespaces list responses only; one extra concurrent CH query per request.
- Potential regressions: none to existing fields — counts computed in a separate goroutine, existing record building unchanged.
- Rollback plan: revert PR; response field is purely additive.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Cluster and namespace list views now return counts of related resources (nodes/namespaces/workloads). |

---

### 📋 Checklist
- [x] Manually tested
- [x] Breaking changes documented (none — additive)

---

## 👀 Notes for Reviewers

- Counts only include entities with metrics reported in the window (a zero-pod workload is invisible) — accepted trade-off to keep it a single query over the pod universe.
- `getPerGroupDistinctCounts` mirrors `getMetadata` for both reduced/non-reduced paths; the only divergence is selecting `labels` (no `unix_milli`/`argMax`) since it counts distinct values, not latest.
- Integration tests validated statically (compile/JSON/semantics); not executed live (Docker unavailable) — worth a CI run.


pkg/modules/inframonitoring/implinframonitoring/helpers.go
```diff
@@ -762,3 +762,170 @@ func (m *module) getMetadata(
...
+	for _, col := range groupByCols {
+		selectCols = append(selectCols,
+			fmt.Sprintf("JSONExtractString(labels, %s) AS %s", sb.Var(col), quoteIdentifier(col)),
+		)
+	}
+	for _, attr := range attrNames {
+		// Guard on != '' so a series missing the attr isn't counted as one empty value.
+		extract := fmt.Sprintf("JSONExtractString(labels, %s)", sb.Var(attr))
+		selectCols = append(selectCols,
+			fmt.Sprintf("uniqExactIf(%s, %s != '') AS %s", extract, extract, quoteIdentifier(attr)),
```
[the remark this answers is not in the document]
nikhilmantri0902: This is fixed now.

pkg/types/inframonitoringtypes/clusters.go
```diff
@@ -28,7 +28,15 @@ type ClusterRecord struct {
 	NodeCountsByReadiness    NodeCountsByReadiness `json:"nodeCountsByReadiness" required:"true"`
 	PodCountsByPhase         PodCountsByPhase      `json:"podCountsByPhase" required:"true"`
 	PodCountsByStatus        PodCountsByStatus     `json:"podCountsByStatus" required:"true"`
-	Meta                     map[string]string     `json:"meta" required:"true"`
+	Counts                   struct {
```
namanv3, line 31: why not properly defined type? why inline?

nikhilmantri0902: I thought that its not needed to exist as a full type. Plus the counts struct is different for clusters and namespaces for the values we return for each one. For namespaces for example, we do not show nodes.

nikhilmantri0902: @namanv3

pkg/modules/inframonitoring/implinframonitoring/helpers.go
```diff
@@ -762,3 +763,197 @@ func (m *module) getMetadata(
...
+	}
+	for _, attr := range attrNames {
+		// Guard on != '' so a series missing the attr isn't counted as one empty value.
+		extract := fmt.Sprintf("JSONExtractString(labels, %s)", sb.Var(attr))
+
+		// Count on the entity's full identity tuple where one is defined, so
+		// same-named entities in different scopes (e.g. workloads sharing a name
+		// across namespaces) aren't collapsed. Falls back to the bare name.
+		valueExpr := extract
+		if tuple, ok := countAttrIdentityTuples[attr]; ok {
```
namanv3, line 845: what is an example of when it can be !ok in the future?

nikhilmantri0902: Any other attribute, lets say k8s.container.name will be !ok and will be counted as count distinct k8s.container.name directly. 
If we want to we can logically keep on mappings of attributes to their unique identification set.

nikhilmantri0902: For now added for required ones.

namanv3 (approved):

therealpandey (approved):

PR #12002, opened 2026-07-07
feat(dashboard-v2): runtime variable fetch engine & selectors

AshwinBhatkal:
## Summary
Foundation for runtime dashboard variables in the V2 dashboard (dark-shipped behind `use_dashboard_v2`). Adds a fetch-orchestration engine that decides **when** each variable's options load and wires the selector controls to it.

- **Fetch-state engine** (`variableFetchSlice`): a per-variable state machine (`idle`/`waiting`/`loading`/`revalidating`/`error`) with a per-variable `cycleId` request nonce. Query variables load in dependency order; dynamics load together once query values exist; text/custom never fetch. `enqueueFetchAll` on load/time change; `enqueueDescendants` on a single value change.
- **Selectors driven off the engine**: Query/Dynamic/Custom/Text selectors read their fetch state and cycle id from the store, so a parent value change refetches only the dependent variables, in order.
- **Default value read as a string** (`string | string[]`), matching the wire type — fixes a default that could not be cleared once set.
- **Batched auto-selection**: the initial load burst of auto-fills (each selector picks a value as its options resolve, at different times) is coalesced into one store write + one downstream refresh, so dependents re-fetch once instead of once per fill.
- **Collapsed `+N` hover** shows the hidden variables' current values.

## Test plan
- Unit: `variableFetchSlice` state transitions (load order, descendant revalidation, batched refresh).
- Manual (dev, flag on): open a dashboard with dependent query + dynamic variables → options load in order; change a parent → only dependents refetch; reload → defaults apply and clear correctly.

Stacked PR (base: `main`). Follow-ups build on this: link-variables-to-panels, then the variable editor/UX.

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/VariablesBar/useVariableFetchState.ts
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/VariablesBar/useVariableFetchState.ts
@@ -1,0 +1,44 @@
+import {
+	selectVariableCycleId,
+	selectVariableFetchedOnce,
+	selectVariableFetchState,
+	type VariableFetchState,
+} from '../store/slices/variableFetchSlice';
+import { useDashboardStore } from '../store/useDashboardStore';
+
+export interface VariableFetchStateResult {
+	variableFetchState: VariableFetchState;
+	/** Include in the selector's react-query key to auto-cancel stale requests. */
+	variableFetchCycleId: number;
+	/** Actively fetching (first load or revalidating). */
+	isVariableFetching: boolean;
+	/** Stable — the fetch completed (or errored). */
+	isVariableSettled: boolean;
+	/** Blocked on parent dependencies (query order) or query variables (dynamics). */
+	isVariableWaiting: boolean;
+	/** Completed at least one fetch — keeps the query subscribed so a cycle bump refetches. */
+	hasVariableFetchedOnce: boolean;
+}
+
+/**
+ * Per-variable view of the runtime fetch engine (`variableFetchSlice`), consumed
+ * by the Query/Dynamic selectors to gate their fetch and key it by cycle id.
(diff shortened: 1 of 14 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/store/slices/variableFetchSlice.ts
```diff
@@ -0,0 +1,261 @@
...
+
+/**
+ * Runtime fetch orchestration for dashboard variables — native port of V1's
+ * `variableFetchStore`. Decides WHEN each variable's options fetch: query
+ * variables in dependency order, dynamics together once query values exist,
+ * text/custom never. `cycleIds` is a per-variable request nonce keyed into each
+ * selector's react-query key (bump = fresh fetch, auto-cancel stale). Transient.
+ * `enqueueFetchAll` = load/time change; `enqueueDescendants` = one value changed.
+ */
+export interface VariableFetchSlice {
```
AshwinBhatkal, line 24: This is same logic as V1. I tried something different for V2. Going back to this implementation only

frontend/src/pages/DashboardPageV2/DashboardContainer/VariablesBar/useVariableFetchState.ts
```diff
@@ -0,0 +1,44 @@
...
+	const variableFetchCycleId = useDashboardStore(selectVariableCycleId(name));
+	const hasVariableFetchedOnce = useDashboardStore(
+		selectVariableFetchedOnce(name),
+	);
+
+	return {
+		variableFetchState,
+		variableFetchCycleId,
+		isVariableFetching:
+			variableFetchState === 'loading' || variableFetchState === 'revalidating',
```
AshwinBhatkal: add enum

ahrefabhi (approved):

PR #12027, opened 2026-07-08
feat: QB support for llm trace list and span list

nityanandagohain:
## Pull Request

---

### 📄 Summary
Support for ai trace list and span list

#### Issues closed by this PR
Fixes https://github.com/SigNoz/engineering-pod/issues/5601

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only
---

- Tests added/updated: ✅
- Manual verification:
- Edge cases covered:

---

- Blast radius: None
- Potential regressions:
- Rollback plan:

---

| Field | Value |
|------|-------|

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

---


tushar-signoz:
Nice work — the pluggable provider design (`BaseConditionProvider` / `ColumnProvider`), field-mapper-aware resolution (verified by the materialized-column golden), and the precedence care in the filter splitter are all solid. A few structural suggestions and a couple of correctness issues before merge.

### Structure

**1. Split the generic scoped-trace builder out of `telemetryai`.**
The seam is already drawn — the code comment even says *"a new category is a new pair of providers, not new topology"* — so let's make it a package boundary instead of an internal one:

- **`pkg/telemetryscopedtraces`** (generic): `scopedTraceStatementBuilder` + the CTE pipeline, the provider contracts (`BaseConditionProvider`, `ColumnProvider`, `TraceColumn`), the `Aggregate` combinator kit, `CommonTraceColumns()`, and the SQL utilities/table bindings.
- **`pkg/telemetryai`** (domain-only): the gen_ai gate provider, the gen_ai column provider, and a thin `NewAITraceStatementBuilder` that wires the two into the generic builder.

Dependency direction is clean (`telemetryai → telemetryscopedtraces → telemetrytraces/querybuilder/telemetryresourcefilter`), and `signozquerier/provider.go` wiring doesn't change. This keeps the next category (db traces, messaging, RUM) from copy-pasting into `telemetryai`, and keeps `telemetryai` honest as "gen_ai domain knowledge only" — the part most likely to churn with semconv changes.

One coupling to fix while splitting: `genAIColumnProvider` reaches for `spanTable()+".duration_nano"` (`max_llm_latency_ns`) — a domain provider shouldn't touch physical table names. Let the generic package own qualification (e.g. a `ScopedToKeyColumn(fn, column, scopeKey)` constructor that qualifies internally).

**2. Align with the module topology (field_mapper / condition_builder).**
Every `pkg/telemetryX` module owns its mapper/builder trio — but `telemetryaudit` owns real implementations because it targets a *different table*. `telemetryai` scans the same span index as `telemetrytraces`, so don't duplicate the mapper; instead, have the scoped-traces builder **construct** `telemetrytraces.NewFieldMapper()` / `NewConditionBuilder()` internally rather than taking them as constructor params. It already hardcodes the telemetrytraces table names, so injecting the mapper for that same schema splits one unit of schema knowledge across two mechanisms. This also trims the 11-param constructor. Keep the delegate `traceStmtBuilder` injected though — it's expensive to duplicate and the provider has the canonical instance.

`SplitFilterForAggregates` is in the right place (`pkg/querybuilder`, next to the other grammar walkers) 👍

### Correctness

**3. `existsExpr` uses only `conds[0]`** (`statement_builder.go`). `ConditionFor(..., cands, ...)` returns one condition per candidate key; if a gen_ai attribute exists under multiple variants (e.g. string + number from a misbehaving SDK), the gate mask and `countIf` checks silently cover only the first variant, while the standard visitor path combines all. Either OR all `conds` or assert `len(cands) == 1`.

**4. Enrichment can disagree with the ranking pass.** `matched` ranks/paginates on window-clipped aggregates; enrichment recomputes the same aliases over the *full* trace and ORDER BYs them again — so page order can differ from the ranking that selected the page, and LIMIT/OFFSET can skip/duplicate at page boundaries when a trace has activity outside the window. If full-trace display values are intentional, consider ordering enrichment by the matched pass's values (join on `matched`), or at least document the discrepancy.

**5. `tracefield.` is inconsistent between Filter and Having.** The splitter classifies `tracefield.completion_tokens` as span-level (fails downstream as an unknown span field), but `buildHaving`'s column map accepts `tracefield.<alias>` in `query.Having`. Pick one behavior for both.

### Minor

- `embedExpr` treats every `?` byte as a placeholder — safe today since all embedded exprs are parameterized, but fragile; cheap hardening is to error if placeholder count ≠ `len(args)`.
- `total_tokens = sum(in) + sum(out)`: a trace with only output tokens yields NULL total (`NULL + n = NULL`); `coalesce` the two sums.
- `QueryRawStream` silently ignores `source=ai` instead of rejecting like `buildQueries` does (logs-only path, so negligible).
- `enrichWithGenAIKeys` surfaces gen_ai keys in trace autocomplete for every tenant, including those with no AI data — presumably intended (fresh-install queries work), but worth a line in the docs; the changelog table in the description is also unfilled.

nityanandagohain:
Thanks for the review
1. Done
2. Done
3. Done
4. Added inline comment about the behaviour. Will also document on the User facing docs.
5. Added support for `tracefield.` as well.
6. All minor concerns addressed.

docs/api/openapi.yml
```diff
@@ -8312,6 +8312,7 @@ components:
     TelemetrytypesSource:
       enum:
       - meter
+      - ai
```
therealpandey: Has this been in discussion? That ai can be source?

therealpandey (changes requested):

docs/api/openapi.yml
nityanandagohain: Wen't though with Tushar, but didn't focus the discussion on the naming here. Let me know if you have a suggestion. 

The idea was to have `signal: trace, source: ai` .

pkg/querier/querier.go
```diff
@@ -235,10 +238,18 @@ func (q *querier) buildQueries(
...
+					event.Source = telemetrytypes.SourceAI.StringValue()
+					stmtBuilder = q.aiTraceStmtBuilder
+				}
+				bq := newBuilderQuery(q.logger, q.telemetryStore, stmtBuilder, spec, timeRange, req.RequestType, tmplVars, builderConfig{})
 				queries[spec.Name] = bq
 				steps[spec.Name] = spec.StepInterval
 			case qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]:
+				if spec.Source == telemetrytypes.SourceAI {
+					return nil, nil, errors.NewInvalidInputf(errors.CodeInvalidInput, "source \"ai\" is only supported for the traces signal, not logs")
+				}
```
tushar-signoz: This should be in validation.go when validating the query range request.
Internal calls will should be assumed to be valid

pkg/querier/querier.go
```diff
@@ -249,6 +260,9 @@ func (q *querier) buildQueries(
 				queries[spec.Name] = bq
 				steps[spec.Name] = spec.StepInterval
 			case qbtypes.QueryBuilderQuery[qbtypes.MetricAggregation]:
+				if spec.Source == telemetrytypes.SourceAI {
+					return nil, nil, errors.NewInvalidInputf(errors.CodeInvalidInput, "source \"ai\" is only supported for the traces signal, not metrics")
+				}
```
tushar-signoz: same

pkg/querier/querier.go
```diff
@@ -479,6 +493,10 @@ func (q *querier) QueryRawStream(ctx context.Context, orgID valuer.UUID, req *qb
 		if query.Type == qbtypes.QueryTypeBuilder {
 			switch spec := query.Spec.(type) {
 			case qbtypes.QueryBuilderQuery[qbtypes.LogAggregation]:
+				if spec.Source == telemetrytypes.SourceAI {
+					client.Error <- errors.NewInvalidInputf(errors.CodeInvalidInput, "source \"ai\" is only supported for the traces signal, not logs")
+					return
+				}
```
tushar-signoz: same

pkg/querybuilder/filter_split_test.go
```diff
@@ -0,0 +1,174 @@
...
+			name:   "agg only trace prefix",
+			query:  "trace.completion_tokens > 1000",
+			having: "trace.completion_tokens > 1000",
+		},
+		{
+			// `tracefield.` is the explicit trace field context (Normalize parses it into
+			// FieldContextTrace), so it marks a trace-level aggregate like `trace.`.
+			name:   "agg only tracefield prefix",
+			query:  "tracefield.completion_tokens > 1000",
+			having: "tracefield.completion_tokens > 1000",
```
tushar-signoz: Do we need tracefield when trace is there ?

pkg/telemetrymetadata/metadata.go
```diff
@@ -1190,6 +1191,45 @@ func enrichWithIntrinsicMetricKeys(keys map[string][]*telemetrytypes.TelemetryFi
...
+	claims, err := authtypes.ClaimsFromContext(ctx)
+	if err != nil {
+		return false
+	}
+	orgID, err := valuer.NewUUID(claims.OrgID)
+	if err != nil {
+		return false
+	}
+	return t.fl.BooleanOrEmpty(ctx, flagger.FeatureEnableAIObservability, featuretypes.NewFlaggerEvaluationContext(orgID))
+}
```
tushar-signoz: We cannot do this, the OrgId is not guranteed to be in context. If you need it, thread OrgId from the upstream and then use the flagger.

pkg/telemetryscopedtraces/statement_builder.go
```diff
@@ -0,0 +1,817 @@
...
+		EndNs:              end,
+	})
+	if err != nil {
+		return "", nil, nil, "", err
+	}
+	if prepared.IsEmpty() {
+		return "", nil, nil, "", nil
+	}
+	sb := sqlbuilder.NewSelectBuilder()
+	sb.Select("1")
```
tushar-signoz: nit: I dont think this is needed

tushar-signoz (commented):
Hey @nityanandagohain 
Can you try to mold the the column functions in fieldMapper and conditionBuilder. 
Otherwise this module will deviate if we try to change something that is common to all statement builders.

pkg/querybuilder/filter_split_test.go
nityanandagohain: Yeah better to remove this

nityanandagohain: Explicitly disabled it on this.

pkg/telemetrymetadata/metadata.go
nityanandagohain: removed the check

docs/api/openapi.yml
tushar-signoz: @nityanandagohain Now that we're discussing on it.
`source` has a specific meaning to it that it defines a physical dataset. For eg `audit` and `meter` are physical independent data sets. `audit logs` are different from `logs` similarly for `meter metrics`. 
Here we're extending/diluting the meaning of `source` by adding `ai` to it. The `ai` traces are not physically new traces, but a a view on the existing traces. Imagine if were to add `frontend` or `db` to `source`. Either we're okay with extending the meaning of source. Or we introduce a new identifier lets say `scope` or `domain`, which means a view over existing data. 

Maybe consider extending this AI solution to a be general solution that it works for a hypothetical use case lets say`db`.

nityanandagohain: changed it to `"type": "builder_ai_query",` after discussion

tushar-signoz:
## Making the AI trace-list code fit the rest of the codebase

### The short version
The new `builder_ai_query` feature works, but two pieces of its structure don't match how the rest of the query code is organized. We can fix that in **two stages**: a safe rename/reshuffle now, and a deeper cleanup later that reuses machinery we already have. Nothing here changes what queries actually return.

### What's a little off today

**1. `telemetryai` looks like a query builder, but it isn't one.**
All the real work — building the ClickHouse SQL for an AI trace list — happens in a generic engine (`telemetryscopedtraces`). The `telemetryai` package just says "here's what counts as an AI span, and here are the columns to show," then hands that to the engine. But it's dressed up as a full builder (with a `field_mapper.go` that contains no field mapper, and a constructor that only forwards its arguments). So a reader expecting a builder finds a thin wrapper instead.

**2. The AI columns reinvent a bit of what we already have.**
Each column (token counts, cost, call counts…) is defined with a small set of Go helpers. Those helpers lean on our existing condition/field logic — good — but for the simple columns they end up re-encoding something we already have a tool for: turning an expression like `sum(input_tokens)` or `countIf(has_error = true)` into SQL. That tool is the shared aggregation rewriter every other builder uses.

We checked how far that existing rewriter could take us. Answer: it handles the plain ones perfectly (a few columns come out **byte-for-byte identical**), but it can't yet express the AI-specific shapes — "count spans where this attribute exists," a few functions it doesn't know (`any`, `argMax`, …), and one or two columns that are really arithmetic across two values. So it's a real overlap, but a partial one.

### The idea
- Treat "which spans, which columns" as a **configuration value** the engine takes — not as its own builder.
- Over time, express the columns as ordinary aggregation strings run through the **shared** rewriter, instead of a parallel set of helpers — teaching the rewriter the few tricks it's missing so it can handle them.

### Why two stages
The first change is local and risk-free. The second touches shared code that every signal (logs, traces, metrics) uses, so it deserves its own review and its own safety net. Splitting them keeps the risky part from ever blocking the feature.

### Stage 1 — tidy up now (safe, no behavior change)
- Bundle "which spans + which columns" into one `TraceScope` value the engine accepts.
- Delete the wrapper that pretended to be a builder; `telemetryai` becomes a single small file that just describes gen_ai.
- Rename the misleading files so their names match their contents.

This is purely moving furniture — **the generated SQL is unchanged.** The tests that pin the entire query, character for character, all still pass, and the build is clean. It's contained to the AI feature, so there's no wider blast radius.

_(One sub-piece — merging the "which spans" definition into a single expression — is intentionally left for Stage 2, because it's the one part that could subtly shift the SQL and deserves the golden-test check that comes with Stage 2.)_

### Stage 2 — reuse the shared rewriter (later, its own PR)
- First, teach the shared rewriter the handful of things it's missing (a few functions, an "attribute exists" form, and not mangling built-in columns). Guard it with the rewriter's existing tests.
- Then, rewrite the AI columns as plain aggregation strings that go through it, keeping the tiny closure escape-hatch only for the two arithmetic columns. Re-run the golden tests to prove the SQL is still identical.

The end result: one generic engine configured by a small `TraceScope`, columns written as one-liners resolved by the machinery we already have, and gen_ai reduced to a simple description file. No parallel abstraction to maintain.

nityanandagohain:
Have refactored for stage1 as you requested, creating a new issue for addressing  stage 2 https://github.com/SigNoz/engineering-pod/issues/5770.

Also for now keeping scope.go in this folder itself, later when doing stage 2 will move it to types.

tushar-signoz (dismissed):
LGTM ! Unblocking we will continue to iterate over this

pkg/telemetryscopedtraces/field_mapper.go
```diff
@@ -0,0 +1,85 @@
...
+	return &fieldMapper{fm: fm, cb: cb, keys: keys}
+}
+
+// FieldFor returns the column expression for key via the field mapper.
+func (r *fieldMapper) FieldFor(ctx context.Context, orgID valuer.UUID, startNs, endNs uint64, key *telemetrytypes.TelemetryFieldKey) (string, error) {
+	expr, err := r.fm.FieldFor(ctx, orgID, startNs, endNs, key)
+	if err != nil {
+		return "", err
+	}
+	return sqlbuilder.Escape(expr), nil
```
srikanthccv: Why is it escaped here?

pkg/statementbuilder/scopedtracesstatementbuilder/field_mapper.go
```diff
@@ -0,0 +1,85 @@
...
+	// One condition per candidate variant (a key can be ingested under several data
+	// types); OR them all, like the visitor does for EXISTS.
+	if len(conds) == 1 {
+		sb.Where(conds[0])
+	} else {
+		sb.Where(sb.Or(conds...))
+	}
+	expr, args := sb.BuildWithFlavor(sqlbuilder.ClickHouse)
+	expr = strings.TrimPrefix(expr, "WHERE ")
+	return sqlbuilder.Escape(expr), args, nil
```
srikanthccv: Why is this happening here in the `ConditionFor`? This file has number of implementation that should be done outside the field mapper. The `ConditionFor` should be returning the list of condition that are then used inside wherever needed instead of building the expression are returning from here.

pkg/telemetryscopedtraces/field_mapper.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// ExistsFor returns the EXISTS predicate for key.
+func (r *fieldMapper) ExistsFor(ctx context.Context, orgID valuer.UUID, startNs, endNs uint64, key *telemetrytypes.TelemetryFieldKey) (string, []any, error) {
+	return r.ConditionFor(ctx, orgID, startNs, endNs, key, qbtypes.FilterOperatorExists, nil)
+}
+
+// ValueFor returns the value expression for an attribute key. The metadata variant
+// is preferred because it carries Materialized — a provider's static definition
+// never does, so a promoted attribute would otherwise fall back to map access.
```
srikanthccv: Which model was this? I don't understand what any of this means. 3 lines of description and still not be clear

pkg/statementbuilder/scopedtracesstatementbuilder/field_mapper.go
```diff
@@ -0,0 +1,85 @@
...
+// column resolution reads like the other statement builders. keys is the fetched
+// metadata for the keys the columns reference; the gate mask is set by the builder
+// after resolveMask (Scoped* aggregates embed it). All returned expressions are
+// escaped once, ready to embed in an outer builder.
+type fieldMapper struct {
+	fm       qbtypes.FieldMapper
+	cb       qbtypes.ConditionBuilder
+	keys     map[string][]*telemetrytypes.TelemetryFieldKey
+	maskExpr string
+	maskArgs []any
```
srikanthccv: Why are these properties of the field mapper?

pkg/telemetryscopedtraces/field_mapper.go
srikanthccv: essentially, I don't see the point of this on field mapper.

pkg/querybuilder/filter_split.go
```diff
@@ -0,0 +1,154 @@
...
+// aggregates), splitting on the top-level AND.
+//
+// A key is trace-level when it carries the trace field context (`trace.completion_tokens`)
+// or, with no context, its bare name is in aggregateNames. Any other explicit context
+// (`span.`, `resource.`, …) is span-level. Trace-level and span-level keys may be
+// AND-combined (they run at different query stages) but not OR-combined; an OR that
+// mixes the two is an error.
+//
+// Syntax errors are ignored here — each part is re-parsed downstream (PrepareWhereClause
+// for the span part, the HAVING rewriter for the trace part), which surface them.
```
srikanthccv: What happens when there is an error which is ignored here, and as a result the `s.span` and `s.having` are empty, does the downstream handle this?

pkg/telemetryscopedtraces/statement_builder.go
```diff
@@ -0,0 +1,781 @@
...
+// HAVING expression (an explicit query.Having is ANDed onto the latter), then
+// validates the trace-level part against the matched-pass aggregates.
+func (b *scopedTraceStatementBuilder) splitFilter(ctx context.Context, orgID valuer.UUID, query qbtypes.QueryBuilderQuery[qbtypes.TraceAggregation], classifySet, orderableSet map[string]struct{}, start, end uint64, variables map[string]qbtypes.VariableItem) (filterParts, error) {
+	var fp filterParts
+	// The legacy tracefield. spelling parses identically to trace.; only the
+	// user-facing form is supported here.
+	if (query.Filter != nil && strings.Contains(query.Filter.Expression, "tracefield.")) ||
+		(query.Having != nil && strings.Contains(query.Having.Expression, "tracefield.")) {
+		return fp, errors.NewInvalidInputf(errors.CodeInvalidInput, "\"tracefield.\" is not supported; use the \"trace.\" prefix")
+	}
```
srikanthccv: Let's just remove this, not only is this sub string check fragile, but this is never used as context. It existed because in some metadata we use `tracefield` on collector to indicate trace context.

pkg/statementbuilder/aistatementbuilder/statement_builder.go
srikanthccv, line 1: what's the clarity around when the orgs spans carry an attribute named with these names below?

srikanthccv (commented):
I initially misunderstood that this was about the statement builder https://github.com/SigNoz/signoz/pull/12027#issuecomment-5063974300 but re-reading it couple of times made it clear that it was about the `agg_rewriter`. So my point  about this on slack can be ignored.

pkg/telemetryscopedtraces/field_mapper.go
nityanandagohain: Yeah not really needed removing.

pkg/telemetryscopedtraces/statement_builder.go
nityanandagohain: So we added this explicitly restrict someone from writing `tracefield.` in the query as it was getting accepted check point 5 here https://github.com/SigNoz/signoz/pull/12027#issuecomment-4963484332.

So we ignore and let it pass for filter query i.e `tracefield.` ?

cc @tushar-signoz

pkg/querybuilder/filter_split.go
nityanandagohain: Yeah this is bad, adding error listener and rejecting it here.

pkg/statementbuilder/aistatementbuilder/statement_builder.go
nityanandagohain: So if there is no prefix or there is `trace.` prefix these are selected. If there is an explicit `attribute.` prefix then it goes to span attribute.

Adding more test for clarity.

pkg/statementbuilder/scopedtracesstatementbuilder/field_mapper.go
nityanandagohain: Discussed with Tushar that we are keeping it a custom one for the builder right now and make it better.

pkg/telemetryscopedtraces/field_mapper.go
nityanandagohain: Updated the comment, the structure keeping it same for now.

pkg/statementbuilder/scopedtracesstatementbuilder/field_mapper.go
nityanandagohain: keeping the field mapper like this for now

pkg/telemetryscopedtraces/statement_builder.go
nityanandagohain: Ok so testing the API without this change returns this error which is good to go with right now. I have removed that piece of code.

```
{
    "status": "error",
    "error": {
        "type": "invalid-input",
        "code": "invalid_input",
        "message": "Invalid references in `Having` expression: [tracefield.max_llm_duration_nano]",
        "errors": [
            {
                "message": "Valid references are: [distinct_tool_count, estimated_total_cost, input_tokens, last_activity_time, llm_call_count, max_llm_duration_nano, output_tokens, tool_call_count, total_tokens, trace.distinct_tool_count, trace.estimated_total_cost, trace.input_tokens, trace.last_activity_time, trace.llm_call_count, trace.max_llm_duration_nano, trace.output_tokens, trace.tool_call_count, trace.total_tokens]",
                "suggestions": []
            }
        ],
        "suggestions": [
            "did you mean: `trace.max_llm_duration_nano > 100`",
            "valid references are `distinct_tool_count`, `estimated_total_cost`, `input_tokens`, `last_activity_time`, `llm_call_count`, `max_llm_duration_nano`, `output_tokens`, `tool_call_count`, `total_tokens`, `trace.distinct_tool_count`, `trace.estimated_total_cost`, `trace.input_tokens`, `trace.last_activity_time`, `trace.llm_call_count`, `trace.max_llm_duration_nano`, `trace.output_tokens`, `trace.tool_call_count`, `trace.total_tokens`"
        ]
    }
}
```

srikanthccv (dismissed):
will follow other set of PRs going to build on top of this as well.

tushar-signoz (dismissed):
lgtm

tushar-signoz (commented):
Read through the trace-list builder end to end and prototyped fixes locally to check they hold. Two correctness issues, then some structural notes. The CTE pipeline itself (`matched` → `ranked` → `buckets` → enrichment) is a nice design, and the documented trade-offs — the accepted ranking discrepancy in `buildEnrichmentSelect`, the deliberate absence of a skip-fingerprint fallback — are exactly the comments I want to find. The 20 golden SQL tests are what made it possible to verify refactors were byte-identical, so thanks for those.

## 1. Typed aggregates can bind to the wrong column

`scopedtracesstatementbuilder/field_mapper.go`, `ValueFor`:

```go
if cands := r.keys[key.Name]; len(cands) > 0 {
    key = cands[0]
}
```

This takes whatever the metadata store happened to return first, with no check that it matches the type the scope declared. `gen_ai.usage.input_tokens` is declared `float64`; if the same name also exists as a string variant and sorts first, `sum()` runs against `attributes_string` — silently wrong token totals, no error.

Worth knowing before fixing it: the obvious fix (filter candidates by the declared context and data type) is **not** sufficient on its own. I tried it, and it breaks the opposite case — when metadata knows the name only under a non-declared type (everything ingested as string), filtering leaves nothing, you fall back to the declaration, and the float64 expression reads `attributes_number`, which holds nothing. Always zero. The current `cands[0]` accidentally gets that case right.

So both directions need handling. See point 3 for where I think the fix belongs.

## 2. Args are threaded by hand and re-bound by counting `?`

The resolver builds each condition against a *throwaway* `SelectBuilder`, compiles it, strips the `WHERE ` prefix, re-escapes the result, and hands back `(expr, args)`. `embedExpr` then re-binds by walking the string and replacing every `?`.

That round trip is why a few fragile things are load-bearing:

- `embedExpr` has to raise an internal error when the placeholder count doesn't match the arg count, because a mismatch would shift every subsequent arg into the wrong slot.
- The escape/unescape asymmetry has to be exactly right — the comment in `buildMatchedCTE` noting that the `countIf` entries "hold live placeholders and must stay unescaped" is doing real work.
- `buildEnrichmentSelect` relies on select-clause args being positionally first in the built SQL.

None of this is needed. `sqlbuilder`'s `$n` markers are **absolute indices** into the builder's arg list, so if you pass the target builder down and let the condition builder bind into it, the returned expression can be embedded in SELECT, WHERE and HAVING of that builder and every occurrence resolves to the same arg. `embedExpr` disappears entirely, along with the counting, the re-escaping and the positional assumption.

One small addition to shared code makes it work: `PreparedWhereClause` needs to expose the bare predicate string (the visitor already computes it as `cond`) so the span predicate can go inside `countIf(...)` instead of only into a WHERE clause. Existing callers pass no `Builder` and are unaffected.

I did this locally and all 20 goldens passed unchanged, which is a good signal — `renderSQL` inlines args positionally and asserts count parity, so any ordering or count regression would have failed them.

## 3. The root cause of #1 is one branch in `ColumnExpressionFor`

`tracestelemetryschema/field_mapper.go`:

```go
switch _, err := m.FieldFor(ctx, orgID, startNs, endNs, field); {
case err == nil:
    candidates = []*telemetrytypes.TelemetryFieldKey{field}   // uses the key AS GIVEN
case errors.Is(err, qbtypes.ErrColumnNotFound):
    candidates = m.CandidateKeys(ctx, orgID, field, nil, keys)
```

Candidates are resolved from metadata only in the **second** branch. For attribute keys `FieldFor` succeeds, so the first branch wins and the key is used verbatim — never consulting `Materialized`, `Indexes` or `Evolutions`, and never learning the other data types the name was ingested under. That is precisely why `ValueFor` has to compensate at all, and why deleting the pre-pick breaks the materialized-column golden.

Preferring metadata candidates before falling back to the literal field fixes all of it at once:

- honours `Materialized` (materialized column instead of a map lookup)
- produces a cross-type `multiIf`, so a string-encoded token count is no longer dropped
- removes the need for any compensation in the scoped builder

I prototyped it: no test failures in any package, and the AI goldens stay green.

**But** — and this is the reason I'd keep it out of this PR — that "no failures" result is not real coverage. I checked, and the multi-variant fixtures (`multi.mat.key`, `mixed.materialization.key`) exercise `AdjustKey`, not `ColumnExpressionFor`. Nothing in the suite renders a column expression from multi-variant metadata, which is exactly why nothing moved. In production, metadata routinely holds several variants per name, so this would reshape SQL across the traces group-by/select path with no unit test able to see it. It wants its own PR gated on a dry-run SQL diff against `main` over real query shapes.

## 4. The local `fieldMapper` merges two roles the codebase keeps apart

Everywhere else, `FieldMapper` (key → column/expression) and `ConditionBuilder` (key + operator + value → predicate) are distinct collaborators — separate interfaces, separate files in each `telemetryschema` package, separate fields on each statement builder. Here one type named `fieldMapper` wraps both, plus the gate mask, and exposes `FieldFor` / `ValueFor` alongside `ConditionFor` / `ExistsFor`.

The name is the sharpest part: it reads as an implementation of `qbtypes.FieldMapper`, and it isn't — the signatures differ, so it can't satisfy the interface. Same for anything named `conditionBuilder` in this package.

A useful thing falls out of separating them: the field-mapper half returns bare literals and needs no SQL builder at all. Only the condition-builder half binds args. That makes the arg-binding boundary structural rather than a comment.

## 5. `NewScopedTraceStatementBuilder` has no injection seam

It takes `metadataStore`, `scope`, `traceStmtBuilder`, `fl` but constructs `fm`/`cb` internally. Both siblings accept them:

```go
NewTraceQueryStatementBuilder(settings, metadataStore, fieldMapper, conditionBuilder, aggExprRewriter, ...)
NewLogQueryStatementBuilder(settings, metadataStore, fieldMapper, conditionBuilder, aggExprRewriter, ...)
```

Those params are a test seam, and the siblings lean on it hard — 11 and 13 test call sites respectively, against 1 production caller each. This package has 0, which is why its tests have to go through the production factory. Moving the two `tracestelemetryschema.New*` lines up into `NewFactory` and passing them in matches the convention and costs nothing behaviourally. (The params have to be interface-typed, since those constructors return unexported types.)

## 6. A lot of single-use helpers

The package has ~14 small helpers, 4 with one- or two-line bodies, versus 2 each in `logsstatementbuilder` and `tracesstatementbuilder`. The ones I'd question are the single-call-site ones — `summaryTable`, `selectAllColumns`, `windowWhere`, `buildHaving`, `buildBucketsCTE` — plus `spanTable` (a trivial `Sprintf`) and `keysOf` (hides a plain slice literal). Naming something and using it once moves code rather than abstracting it.

Not all of them: anything holding a decision earns its name. The `[]string` → single-predicate collapse is worth keeping — the filter visitor keeps the same wrapper around that same call, at the same 7 call sites — and so is the conditional backticking in `quoteAlias`.

## 7. Smaller things

- `resolveSpanPredicate` doesn't pass `OrgID` into `FilterExprVisitorOpts`, so the field mapper and condition builder see a zero org there. Every other caller passes it. Looks unintentional.
- `TraceDuration` in `aggregate.go` hardcodes `toUnixTimestamp64Nano`, which assumes the timestamp column's physical type. The only other use of that function lives in `tracestelemetryschema` — that bit of schema knowledge arguably belongs down there too.
- The comment on the `ValueFor` pre-pick says it exists for evolutions ("a key like service.name can have evolutions"). It's actually load-bearing for `Materialized`: `service.name` renders identically with or without the pre-pick, because `FieldFor` fails for resource keys so `CandidateKeys` already runs. Worth correcting so the next reader doesn't delete the wrong half.

---

Happy to push the #1/#2 fixes as a follow-up if that's easier than folding them in — I have them working locally with the goldens green. I validated against a base a few commits behind this branch, so line numbers may drift, but I re-checked each item against the current head: `embedExpr`, the `cands[0]` pre-pick, `maskArgs` and the internally-constructed `fm`/`cb` are all still present.

nityanandagohain:
@tushar-signoz , have addressed others apart from 1 & 3. Will keep it separate. Thanks for the review.

tushar-signoz (dismissed):
LGTM ! Thanks for bearing my reviews.

therealpandey (dismissed):

pkg/querybuilder/filter_split_test.go
```diff
@@ -0,0 +1,222 @@
...
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			span, having, err := SplitFilterForAggregates(c.query, agg)
+			if c.wantErr {
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			require.Equal(t, c.span, span, "span part")
+			require.Equal(t, c.having, having, "having part")
```
tushar-signoz: Hey @nityanandagohain 
This was missed, it should be assert instead of require. lets fix these or fix them in a follow up PR your call.

tushar-signoz: Read about github stack PRs somewhere, may be good opportunity to try them out.

tushar-signoz (approved):
LGTM

therealpandey (approved):

PR #12033, opened 2026-07-08
fix(dashboard-v2): view-modal, pie chart, and panel fixes

ahrefabhi:
## Pull Request

---

### 📄 Summary

A batch of Dashboard V2 fixes plus a few small features, found while polishing the panel editor and the View-panel modal. Each concern is its own commit, so the PR reads cleanly commit-by-commit:

- **fix:** tint the chart-legend checkbox with its series color (the `@signozhq/ui` Checkbox shadows the ancestor CSS var, so the tint never applied).
- **fix:** make antd Selects/pickers usable inside the View-panel modal (they portalled outside the Radix dialog → opened but unclickable).
- **feat:** add the List columns editor + pagination loader to the View modal (parity with the full editor).
- **fix:** measure chart dimensions before first paint — kills the pie-legend "render left, then jump to center" flash.
- **feat:** per-slice color overrides for pie charts.
- **feat:** "Move to section" now lists the dashboard root and every section.
- **fix:** hide the empty panel header in the editor preview.
- **fix:** lazy-load panels per-panel instead of per-section — a board only queries the panels actually on screen.
- **fix:** hide "Extend time range" in the empty state for panels with a fixed time preference — widening the dashboard window can't add data to a panel locked to a relative span, so the CTA was a no-op (Retry still shows).
- **refactor:** read a panel's time preference through a shared `getPanelTimePreference` helper (drops a duplicated inline cast).
- **feat:** per-series legend colors for bar and histogram panels (parity with the pie per-slice colors).
- **feat:** pin the query-type tabs + Run Query to the top of the panel editor's scroll area so they stay visible while the query body scrolls.

#### Issues closed by this PR
Closes https://github.com/SigNoz/pulse-pod/issues/105

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
- **Legend checkbox tint:** the `@signozhq/ui` Checkbox declares `--checkbox-checked-background` on the element itself (via its `data-color` rule), which shadows any value inherited from an ancestor — so the per-series color set on a wrapper never took effect.
- **antd popups in the View modal:** the editor's antd popups portal to `document.body`, outside the modal Radix dialog. A modal Radix dialog sets `pointer-events: none` on the body and traps focus, so those menus opened but weren't clickable. The List columns combobox (a Radix popover) had the same problem plus a stacking one — it opened behind the dialog.
- **Pie-legend jump:** `useResizeObserver` seeded its size to `0` and only reported real dimensions after paint (debounced observer), so charts painted once at `0` then re-laid-out; the pie legend rendered left-aligned, then re-centered once it learned it fit on one row.
- **List pagination loader in the View modal:** `PreviewPane` never received `isPreviousData` from the modal, so the List panel couldn't tell a page change was in flight.
- **Empty panel header:** the header bar rendered even when it had no title/description/status and actions were suppressed (editor preview).
- **Panel lazy-loading was per-section, not per-panel:** the viewport `IntersectionObserver` lived on `Section` and covered the whole section, so scrolling any part of a section into view fetched *every* panel in it — including panels still off-screen below the fold.
- **"Extend time range" on fixed-preference panels:** a panel with a non-global `timePreference` queries a fixed relative span anchored to the dashboard end and ignores the ambient/global window. The empty-state "Extend time range" action widens that global window (via the zoom-out ladder), so on such a panel it could never change the data — the CTA looked actionable but did nothing.

#### Fix Strategy
- Feed the checkbox color through a private `--series-color` var and override the vars on the element with a selector that out-specifies the library rule.
- Wrap the modal body in an antd `ConfigProvider` whose `getPopupContainer` targets the dialog content, so every antd popup renders inside the interactive, focus-trapped layer. For the Radix columns combobox, render it inline (`withPortal={false}`) and raise its `z-index` above the query-builder popups.
- Add a `useLayoutEffect` to `useResizeObserver` that measures the element synchronously before paint.
- Thread `isPreviousData` into the modal's `PreviewPane`.
- Return nothing from the panel header when there's nothing to show.
- Move the observer down into a new `SectionGridItem` wrapper that watches its own intersection (latched) and gates its panel's fetch via `isVisible`, so a board of many panels only queries the ones actually on screen.
- Pass the panel to the shared `NoData`; when `panelHasFixedTimePreference(panel)` is true, drop the global extender so only Retry shows. The View modal's local extender (published to the store) still wins, because its override window ignores `timePreference` anyway. The read lives in a shared `getPanelTimePreference` helper, which also replaces the header pill's duplicated inline cast.

---

- **Tests added/updated:**
  - `NoData` — added cases for a fixed-preference panel (Extend hidden, Retry only), a global-time panel (Extend still shown), and the View-modal extender still winning over the (suppressed) global one.
  - `useLegendSeries` — added a pie-panel case (resolves from scalar slices, deduped) and a histogram case (resolves via the time-series path; the pie path is not invoked).
  - `usePanelActionItems` — updated for the reworked "Move to section" menu.
  - All existing legend / pie / view-modal / chart suites pass.
- **Manual verification:** legend checkbox tint; antd dropdowns + List columns editor + pagination inside the View modal; pie legend centers with no flash; pie per-slice color picker repaints slices; per-series legend colors on bar + histogram; move-to-section targets; empty header hidden in editor preview; panels below the fold stay unfetched until scrolled into view; the "Extend time range" CTA is hidden on a panel with a fixed time preference (e.g. a `6h` panel) while Retry remains; query-type tabs stay pinned while the query body scrolls.
- **Edge cases:** disabled pie kinds still show tooltips; non-supported panel kinds return no legend series; jsdom has `clientWidth === 0`, so the pre-paint measurement is a no-op in tests; a panel that follows the global window (or a `NoData` rendered without a panel) keeps the Extend action.

---

- **Blast radius:** `useResizeObserver` is shared (~28 consumers), but the change only delivers the correct width one render earlier (before paint); jsdom is unaffected (`clientWidth` is `0` there). `useLegendSeries` was refactored into `utils/legendSeries.ts`; the `LegendSeries` type moved with it and consumers were repointed. The lazy-load change is scoped to the V2 `SectionGrid`/`Section` render path — the per-section observer was replaced 1:1 by a per-panel one. The Extend-CTA change adds an optional `panel` prop to the shared `NoData`; omitting it preserves the prior behavior, so the only affected surface is the 7 panel renderers that now pass it.
- **Potential regressions:** chart initial-layout timing; any panel that relied on the empty header occupying space (mitigated by a `min-height` when it does render); lazy-load gating now fires per panel rather than per section (each panel latches on first intersection); the empty-state action set for panels that use a fixed time preference.
- **Rollback plan:** each concern is an isolated commit — revert the specific commit(s); no data or API changes.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature + Bug Fix |
| Description | Dashboard V2 polish: pie charts support per-slice legend colors and bar/histogram support per-series legend colors; the View-panel modal now supports antd dropdowns, the List columns editor, and the pagination loader; the pie legend no longer flashes on load; "Move to section" lists the dashboard root and all sections; panels lazy-load individually so a board only queries what's on screen; the empty-state "Extend time range" action is hidden for panels locked to a fixed time preference; and the panel editor's query-type tabs stay pinned while the query body scrolls. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

Please review commit-by-commit — the 12 commits are independent and each is self-contained (tests folded into their parent commit). The largest is the pie-colors commit, which also extracts `useLegendSeries`'s logic into `utils/legendSeries.ts` (a shared dedupe primitive + a pie and a time-series resolver) and reduces the hook to a kind switch that returns none for kinds without a colors control.

The four most recent commits are small and orthogonal to the rest:
- `fix: hide "Extend time range" for panels with a fixed time preference` — behavior + `NoData` tests.
- `refactor: read panel time preference via shared helper` — pure cleanup enabled by the fix's new helper.
- `feat: per-series legend colors for bar and histogram panels` — routes both kinds through the time-series legend path and enables the Legend `colors` control.
- `feat: pin the query-type tabs to the top of the editor scroll area` — CSS-only sticky nav.


frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/ViewPanelModal/ViewPanelModalContent.tsx
```diff
@@ -119,6 +131,16 @@ function ViewPanelModalContent({
 					isLoadingQueries={isFetching}
 					onStageRunQuery={runQuery}
 					onCancelQuery={cancelQuery}
+					footer={
+						isListPanel ? (
+							<ListColumnsEditor
+								spec={draft.spec}
+								onChangeSpec={setSpec}
+								signal={signal}
+								withPortal={false}
```
AshwinBhatkal: do we need this? What happens if we don't have it? I'm just trying to avoid this prop at so many places

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/hooks/useLegendSeries.ts
```diff
@@ -33,27 +22,13 @@ export function useLegendSeries(
...
-
-		return Array.from(byLabel, ([label, defaultColor]) => ({
-			label,
-			defaultColor,
-		}));
-	}, [panel.spec.queries, data.response, data.legendMap, isDarkMode]);
+		switch (panel.spec.plugin.kind) {
+			case 'signoz/PieChartPanel':
+				return resolvePieLegendSeries(data, isDarkMode);
+			case 'signoz/TimeSeriesPanel':
```
AshwinBhatkal, line 28: Only for these 2 type of charts?

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/PanelHeader/PanelHeader.tsx
```diff
@@ -74,6 +74,14 @@ function PanelHeader({
 		[warning],
 	);
 
+	/**
+	 * Hide the entire header when there's no title, description, or status to show,
```
AshwinBhatkal, line 78: With this, there will be a jank if error or warning does come right? I think we

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/hooks/useLegendSeries.ts
ahrefabhi: adding other panels as well

ahrefabhi: Added

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/ViewPanelModal/ViewPanelModalContent.tsx
ahrefabhi: removed

AshwinBhatkal (approved):

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/PanelHeader/PanelHeader.tsx
AshwinBhatkal: can you please verify this?

ahrefabhi: on the panel dashboard view page, no jank will be there because actions will always come.

PR #12041, opened 2026-07-09
fix(dashboard-v2): bug bash fixes III

ahrefabhi:
## Pull Request

---

Third batch of dashboard-v2 bug bash fixes, one commit per fix:

1. **Reveal the affected panel/section** — saving a panel, closing the editor, cloning a panel, or adding a section left the dashboard scrolled to the top. Save now resolves with the persisted panel id, a small scroll-target store hands it to the grid, and a shared `scrollIntoViewWhenReady` util polls for the optimistic DOM commit before scrolling.
2. **Metric unit seeding follows the kind's Formatting controls** — `useMetricYAxisUnit` is replaced by `useSeedMetricUnit`, driven by the same per-kind controls `buildPluginSpec` reads. Kinds with a panel-wide unit seed `formatting.unit`; Table (which has none) seeds each resolved value column's `formatting.columnUnits`, and column unit selectors now surface the metric-unit mismatch warning.
3. **Grid placement can no longer overlap tall panels** — `findFreeSlot` probes the last-row candidate against every existing item (mirroring the backend's overlap rejection) before placing there, falling back to a fresh bottom row. Move-to-section reuses the same primitive instead of always dropping to the bottom.
4. **Panel type switches carry supported config forward** — switching visualization kind kept only formatting and thresholds, dropping axes entirely and resetting legend/visualization/chartAppearance to defaults. Each section seed now carries the old spec's fields that the target kind's controls declare: axes soft min/max + log scale, time preference, stacking/fillSpans, legend position, and chart appearance. Unsupported fields (and legend `customColors`, keyed by series labels the new kind may not reproduce) are still dropped.
5. **Context link editor survives undecodable URL params** — a literal `%` in a query value (e.g. `?search=95%`) is not valid percent-encoding, so `decodeURIComponent` threw while seeding the edit dialog and unmounted the page. Keys/values that fail to decode now fall back to the raw string.
6. **Adding a query no longer auto-runs the preview** — the query-type/datasource effect re-committed the draft on any structural builder change, so adding a query refetched the preview for a still-empty query. A pure append is now skipped; the new query commits on Run Query as usual.
7. **The editor route bootstraps variables and edit context itself** — a hard refresh (or direct URL) of the full-page editor mounts without `DashboardContainer` or the variables bar, so the preview fetched with unresolved variables and the store's edit context stayed cold. Selection seeding moves into a standalone `useSeedVariableSelection` (URL → persisted store → default, pruning stale URL entries), `PanelEditorPage` mounts it alongside `useResolvedVariables` and the edit-context seed, and `?variables=` is carried across dashboard ↔ editor navigation (V1 parity).
8. **Carried table thresholds are keyed onto a real column** — switching timeseries/number → table carried thresholds with an empty `columnName`, which the save API rejects (`TablePanelSpec.Thresholds[].ColumnName` is required). The renderer's keying rule is extracted as `getAggregationColumnKey` and reused at switch time to key carried thresholds onto the first enabled query's value column; thresholds with no derivable column are dropped instead of blocking the save.

#### Issues closed by this PR
Closes https://github.com/SigNoz/pulse-pod/issues/106

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
- **Reveal:** nothing connected a mutation (save/clone/add) to the grid item it produced; the old add-section scroll helper was a one-off rAF poll and other flows had nothing.
- **Unit seeding:** seeding was hardcoded to the panel-wide `formatting.unit`, a field Table doesn't declare — so Table panels never got the metric unit and the save API would reject the field.
- **Placement:** `findFreeSlot` only checked the last row's right edge against the grid width; a tall panel from an earlier row reaching down into the last row made "right of the last row" occupied, and the backend rejects overlapping patches.
- **Type switch:** the switch-time spec seeding only carried formatting and thresholds; the axes section had no seed at all and the other sections unconditionally re-seeded defaults.
- **Context links:** the URL-param seeding assumed every param was valid percent-encoding; a user-entered literal `%` made `decodeURIComponent` throw `URIError` during render.
- **Preview auto-run:** the effect keyed on a joined datasource signature, which also changes when a query is appended — indistinguishable from a datasource switch.
- **Editor bootstrap:** seeding lived inside the variables bar's `useVariableSelection`, and the editor route never mounts the bar; the edit context was only seeded by `DashboardContainer`.
- **Table thresholds:** the threshold carry seeded `columnName: ''` because column keys were only derivable from query response data, which doesn't exist at switch time.

#### Fix Strategy
Each fix routes through the existing single source of truth instead of adding parallel state: the scroll flows share one util + store; unit seeding and switch-time seeding both read the per-kind section `controls`; placement reuses one `findFreeSlot` primitive with an explicit overlap guard; carry-forward is declared per section seed, gated by the target kind's controls; variable seeding is one hook mounted by both the bar and the editor route, with URL parse/serialize extracted to `variablesUrlState`; and table-threshold keying reuses the renderer's own `getAggregationColumnKey` + `toQueryEnvelopes` rather than duplicating either.

---

- Tests added/updated: `scrollIntoViewWhenReady`, `useScrollPanelIntoView`, `useClonePanel`, `usePanelEditorSave`, `PanelEditorContainer`, `useSeedMetricUnit`, `formattingSpec`, `FormattingSection`, `patchOps` (overlap + placement), `buildPluginSpec` (carry rules per section + threshold keying), `ContextLinksSection/utils` (undecodable params), `usePanelEditorQuerySync` (append skip), `useSeedVariableSelection`, `variablesUrlState`, `useResolvedVariables`, `useDrilldownDashboardVariables`, `getDefaultTableColumnKey`
- Manual verification: flows exercised during the bug bash on a local dashboard
- Edge cases covered: unsaved new panel on editor close (no scroll target), Table with no panel-wide unit, decimal precision 0, null soft min/max, threshold variant remaps, fields the target kind doesn't declare, tall panel reaching into the last row, literal `%` in context-link URLs, multi-aggregation column keys (`name.expression`), disabled/promql/formula queries, variables renamed since a link was shared
- `tsgo --noEmit`, `oxlint`, and `pnpm build` pass; pre-commit hooks ran the same per commit

---

- Blast radius: DashboardPageV2 only (behind the dashboards v2 flag); no V1 files touched
- Potential regressions: switch-time specs now include carried fields — save payloads change shape for switched panels; placement picks a different (earlier) slot than before for move-to-section; the editor route now writes edit context + variable selection into the dashboard store on mount
- Rollback plan: revert individual commits — each fix is self-contained and passes checks independently

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | N/A |
| Change Type | N/A |
| Description | N/A (dashboard v2 is behind a feature flag) |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

Review commit-by-commit — the eight fixes are independent. The `index.tsx` / `PanelEditorContainer.test.tsx` changes are split across the first two commits (scroll target on save/close vs. unit-seeding rewire). In the editor-bootstrap commit, `useVariableSelection`'s diff is mostly deletion — the seeding moved verbatim into `useSeedVariableSelection`.


frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/index.tsx
```diff
@@ -203,19 +204,33 @@ function PanelEditorContainer({
...
+
 	const onSave = useCallback(async (): Promise<void> => {
 		if (!isEditable) {
 			return;
 		}
 		try {
 			// Bake the live query into the spec so unstaged edits are saved too.
-			await save(buildSaveSpec(draft.spec));
+			const savedPanelId = await save(buildSaveSpec(draft.spec));
 			toast.success('Panel saved');
```
AshwinBhatkal: nit: this should come below the setScroll

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Section/SectionGrid/useScrollPanelIntoView.ts
```diff
@@ -0,0 +1,24 @@
+import { RefObject, useEffect } from 'react';
+
+import { useScrollToPanelStore } from '../../../store/useScrollToPanelStore';
+
+/**
+ * Scrolls this panel into view when the store requests it (e.g. returning from
+ * the editor or after creating a panel), then clears the request so it fires
+ * once. Runs on mount, so it catches the panel as soon as the grid renders it.
+ */
+export function useScrollPanelIntoView(
```
AshwinBhatkal: Does this not also scroll section to view and not just panel?

frontend/src/utils/scrollIntoViewWhenReady.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Polls via rAF until `resolveTarget` returns an element, then scrolls it into
+ * view — for nodes that mount a frame or two after the call (e.g. an optimistic
+ * write). Return `null`/`undefined` while not ready; bails after `attempts`.
+ */
+export function scrollIntoViewWhenReady(
```
AshwinBhatkal: instead of this, why not follow the earlier approach? 

This scrolling logic about identifying can be inside the panel or section that renders and it will automatically scroll
useScrollWidgetIntoView

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/hooks/usePanelEditorQuerySync.ts
```diff
@@ -111,17 +111,27 @@ export function usePanelEditorQuerySync({
...
 	const didMountRef = useRef(false);
 	useEffect(() => {
+		const prev = prevDataSourcesRef.current;
+		prevDataSourcesRef.current = dataSources;
 		if (!didMountRef.current) {
 			didMountRef.current = true;
 			return;
 		}
+		// An added query is still empty — don't auto-run it; it commits on Run Query.
+		const isQueryAdded =
```
AshwinBhatkal, line 129: curious. Is this the behaviour we added here - https://github.com/SigNoz/signoz/pull/11358?
Or this happens just for this?

frontend/src/pages/DashboardPageV2/DashboardContainer/queryV5/prepareScalarTables.ts
```diff
@@ -108,8 +108,23 @@ function getColName(
 }
 
 /**
- * Stable row-data key (port of V1 `getColId`). Multi-aggregation queries need
- * `queryName.expression` so value columns don't collide.
+ * Key a value column is stored and rendered under: the query name, or
```
AshwinBhatkal: Key a value column in stored means?

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/sections/ContextLinksSection/utils.ts
```diff
@@ -29,18 +29,24 @@ export function insertVariableAtCursor(
...
+		return value;
 	}
 }
 
+// Values may be double-encoded on the wire; decode a second time only when it changes
+// the string, so already-single-encoded values are left intact.
+function decodeForDisplay(value: string): string {
+	const decoded = safeDecodeURIComponent(value);
+	const doubleDecoded = safeDecodeURIComponent(decoded);
+	return doubleDecoded !== decoded ? doubleDecoded : decoded;
```
AshwinBhatkal: nit : could always be double decoded. Anyways computation is happening

frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/buildPluginSpec.ts
```diff
@@ -90,35 +91,103 @@ interface SectionSeed {
...
+		seed: (
+			controls,
+			{ oldSpec },
+		): SectionSpecMap[SectionKind.Visualization] | undefined => {
 			const c = controls as SectionControls[SectionKind.Visualization];
-			return c.timePreference
-				? { timePreference: DashboardtypesTimePreferenceDTO.global_time }
-				: undefined;
+			const old = (
+				oldSpec?.plugin.spec as {
```
AshwinBhatkal: do we need this cast?

frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/buildPluginSpec.ts
```diff
@@ -90,35 +91,103 @@ interface SectionSeed {
...
+	},
+	[SectionKind.Axes]: {
+		specKey: 'axes',
+		seed: (
+			controls,
+			{ oldSpec },
+		): SectionSpecMap[SectionKind.Axes] | undefined => {
+			const c = controls as SectionControls[SectionKind.Axes];
+			const old = (
+				oldSpec?.plugin.spec as {
```
AshwinBhatkal: same here

frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/buildPluginSpec.ts
```diff
@@ -90,35 +91,103 @@ interface SectionSeed {
...
 	[SectionKind.ChartAppearance]: {
 		specKey: 'chartAppearance',
-		seed: (controls): SectionSpecMap[SectionKind.ChartAppearance] | undefined => {
+		seed: (
+			controls,
+			{ oldSpec },
+		): SectionSpecMap[SectionKind.ChartAppearance] | undefined => {
 			const c = controls as SectionControls[SectionKind.ChartAppearance];
+			const old = (
+				oldSpec?.plugin.spec as {
```
AshwinBhatkal: same here

AshwinBhatkal: too many "?", don't you think?
Maybe evaluate how the spec is being passed

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/hooks/useMovePanelToSection.ts
```diff
@@ -20,8 +20,8 @@ interface Params {
 
 /**
  * Relocates a panel's item ref from one section to another. The panel itself
- * stays in `spec.panels`; only the grid item moves, dropped into a free row at
- * the bottom of the target section. Persisted as one atomic patch.
+ * stays in `spec.panels`; only the grid item moves, dropped into the target
+ * section's first free slot (`findFreeSlot`). Persisted as one atomic patch.
```
AshwinBhatkal: It is better off at the end only. Why, start?

AshwinBhatkal: Reason, the users might already arranged the their panels. Adding it to the start disrupts it all

AshwinBhatkal (commented):
few comments

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/hooks/usePanelEditorQuerySync.ts
ahrefabhi: no this happens just for this

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/index.tsx
ahrefabhi: Done

frontend/src/utils/scrollIntoViewWhenReady.ts
ahrefabhi: Done

frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/buildPluginSpec.ts
ahrefabhi: Removed

frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/buildPluginSpec.ts
ahrefabhi: Done

frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/buildPluginSpec.ts
ahrefabhi: Done

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/hooks/useMovePanelToSection.ts
ahrefabhi: Done

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/sections/ContextLinksSection/utils.ts
ahrefabhi: Done

frontend/src/pages/DashboardPageV2/DashboardContainer/queryV5/prepareScalarTables.ts
ahrefabhi: updated the comment

frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/buildPluginSpec.ts
```diff
@@ -90,22 +95,18 @@ function toThresholdVariant(
...
-			{ oldSpec },
+			{ oldPluginSpec },
 		): SectionSpecMap[SectionKind.Visualization] | undefined => {
 			const c = controls as SectionControls[SectionKind.Visualization];
-			const old = (
-				oldSpec?.plugin.spec as {
-					visualization?: SectionSpecMap[SectionKind.Visualization];
-				}
-			)?.visualization;
+			const old = oldPluginSpec?.visualization;
```
AshwinBhatkal, line 123: I'm concerned with "?"
Is there a way we never accept or return undefined. I saw the return type of seed right now. It should ideally never be undefined

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Section/SectionGrid/useScrollPanelIntoView.ts
AshwinBhatkal: Can we unify it it into 1 hook for section and panel both? The store will store type and id instead of just id

ahrefabhi: Done

frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/buildPluginSpec.ts
ahrefabhi: Done

AshwinBhatkal (approved):

PR #12046, opened 2026-07-09
feat(dashboard-v2): query-first DSL filter for the dashboards list

AshwinBhatkal:
## Summary

Reworks the V2 dashboards **list filter** into a query-first model: a single DSL query string is the source of truth, edited both by typing and by the Created-by / Updated dropdowns.

- **One unified query string.** The box always shows the full query, including `created_by = '…'` and `updated_at >= '…'`. The separate Tags dropdown is gone — tags are typed directly (`env = 'prod'`).
- **Stage-aware autocomplete.** Typing (or clicking into the box) suggests keys → operators (the full valid set per field type) → values (creator emails, org tag values, booleans), then chains `AND`/`OR`. Reserved fields and tag keys are distinguished with a `field` / `tag` pill.
- **CodeMirror query box** with query-builder-parity token highlighting, keyboard navigation, and a full dropdown (SUGGESTIONS header + keyboard-hint footer). Runs on ⌘/Ctrl+Enter or the Run button (a blue dot marks unrun changes).
- **Two-way dropdowns.** Created-by (antd multi-select with checkboxes, clear-all) and Updated (single-select) splice their clause into the query; editing the query text reflects back into the dropdowns for simple top-level terms. Created-by runs when the dropdown closes; Updated runs on selection.
- **Views are query strings.** Built-in (Mine / Locked / …) and saved views store and apply the query verbatim; "modified" is a trimmed string compare.
- **List polish.** Sticky header while results scroll; deleting the last row on a page steps back a page instead of showing an empty list.

## Grammar (mirrors the backend list filter)

Reserved keys `name / description / created_by / created_at / updated_at / locked`; any other key is a tag. Per-field operator sets (string/tag: `= != CONTAINS LIKE ILIKE IN …`, tags also `EXISTS`; timestamps: `= != < <= > >= BETWEEN`; `locked`: `= !=`). Strings quoted, `IN ['a','b']`, booleans unquoted, timestamps RFC3339.

## Test plan

- Type `env` → key suggestions; space → operators; then tag values; picking a value chains `AND`/`OR`.
- Pick **Created by → me**, confirm `created_by = '…'` appears; type `updated_at >= '…'`, confirm the Updated dropdown reflects it.
- Editing the box does not fetch; Run / ⌘Enter does; the dirty dot clears after running.
- Select the **Locked** built-in view → `locked = true`; save a view and re-select it (not spuriously "modified").
- Delete the last dashboard on the last page → lands on the previous page.

Unit tests: `dslTokenizer`, `dslSuggestions`, `filterQuery` (52 tests).

Closes https://github.com/SigNoz/pulse-pod/issues/108


frontend/src/pages/DashboardsListPageV2/utils/dslGrammar.ts
```diff
@@ -0,0 +1,102 @@
...
+export const RESERVED_KEYS = [
+	'name',
+	'description',
+	'created_by',
+	'created_at',
+	'updated_at',
+	'locked',
+] as const;
+
+export type DslFieldType = 'string' | 'timestamp' | 'bool' | 'tag';
```
AshwinBhatkal, line 18: remove tag

frontend/src/pages/DashboardsListPageV2/utils/dslGrammar.ts
```diff
@@ -0,0 +1,102 @@
...
+	locked: 'bool',
+};
+
+// A reserved key maps to its column type; anything else is a tag.
+export const classifyField = (key: string): DslFieldType =>
+	RESERVED_FIELD_TYPES[key.toLowerCase()] ?? 'tag';
+
+// String/tag comparison operators (REGEXP is allow-listed by the backend but
+// unimplemented, so it is deliberately omitted).
+const STRING_OPS = [
```
AshwinBhatkal, line 35: can this be reused

AshwinBhatkal: should be usages already

frontend/src/pages/DashboardsListPageV2/components/FilterZone/FilterChips.tsx
```diff
@@ -1,7 +1,5 @@
-import { useMemo } from 'react';
-import { Button } from '@signozhq/ui/button';
-import { DropdownMenuSimple, type MenuItem } from '@signozhq/ui/dropdown-menu';
-import { CalendarClock, ChevronDown, User } from '@signozhq/icons';
+// eslint-disable-next-line signoz/no-antd-components -- antd Select (no @signozhq multi-select equivalent that fits the layout)
```
AshwinBhatkal: can we try using signoz hq here?

ahrefabhi (commented):
Image

In operator is not converting the value suggestion to array

ahrefabhi (approved):

PR #12049, opened 2026-07-09
fix(dashboard-v2): carry unsaved panel edits into view mode

ahrefabhi:
## Pull Request

---

In the V2 panel editor, clicking **Switch to View Mode** with un-saved changes dropped everything except the query. The View modal re-seeded its draft from the *saved* panel spec and only overlaid the live query, so config edits — thresholds, units, table/list columns, legend, formatting, axes — were silently lost.

This PR hands the live draft spec off to the View modal so those edits carry over. The query already travels in the URL (`compositeQuery`) because the query builder hydrates from it; the rest of the spec now travels in a **tab-scoped `sessionStorage` handoff** keyed by `dashboardId + panelId`.

Why `sessionStorage` and not the URL: an earlier iteration put the whole spec in the URL, but that bloated the link badly. `sessionStorage` keeps the URL small, still **survives a page refresh** (the key requirement), and auto-clears when the tab closes. `expandedWidgetId` in the URL is the correlator, so the handoff is only ever read back for the exact panel it was written for.

N/A — behavior is state-carry across a mode switch; no visual chrome changes. Verified via unit tests + build (see Testing Strategy).

#### Issues closed by this PR
Closes https://github.com/SigNoz/engineering-pod/issues/5663

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
`useSwitchToViewMode` carried only the live query (via the `compositeQuery` URL param) and the panel type (`graphType`). On the receiving side, `useViewPanelMode` seeded its draft from the **saved** `panel.spec` via `buildViewPanelSpec`, overlaying just the query. There was no channel for the editor's un-saved config edits, so they never reached the View modal.

#### Fix Strategy
- `useSwitchToViewMode` writes the live draft spec to a `sessionStorage` handoff (`writeViewPanelHandoff`) before navigating; the query stays in the URL.
- `useViewPanelMode` uses `readViewPanelHandoff(dashboardId, panelId)` as the base spec for its seed, falling back to the saved panel when there's no handoff / it belongs to another panel / it's malformed.
- The handoff is cleared (`clearViewPanelHandoff`) on plain grid **View**, grid drilldown, and modal **close**, so a stale editor spec can never seed an unrelated view.

---

- **Tests added/updated:** `useSwitchToViewMode.test.ts` — asserts the query still goes to the URL, the spec goes to the `sessionStorage` handoff (and *not* the URL), and the handoff is scoped to the exact `dashboardId + panelId`. Ran the full PanelEditor + PanelsAndSectionsLayout suites: **332 passing**.
- **Manual verification:** `pnpm tsgo --noEmit`, `pnpm oxlint`, and `pnpm build` all clean. Not driven in a live browser session.
- **Edge cases covered:** no handoff (plain View → saved panel); handoff for a different panel (ignored); malformed stored value (falls back); refresh (sessionStorage persists, URL re-seeds the query); plain View / drilldown / close clear the handoff.

---

- **Blast radius:** V2 dashboard panel View modal + the editor's "Switch to View Mode" only. No API, schema, or shared query-builder changes.
- **Potential regressions:** low. The new base-spec source falls back to the exact prior behavior (`panel.spec`) whenever no handoff is present, so non-editor entry points into the View modal are unchanged.
- **Rollback plan:** revert this commit; no migrations or persisted server state involved. `sessionStorage` entries are tab-scoped and self-expire on tab close.

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | V2 dashboards: switching a panel from edit to view mode now keeps your un-saved changes (query and config) instead of reverting to the last saved version. |

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

- Because the handoff lives in `sessionStorage` (not the URL), a **copied/shared link won't carry un-saved edits** — it opens on the saved panel. This is intentional (you can't share someone's local unsaved edits) and mirrors the existing reverse view→edit handoff.
- Scope is deliberately editor → view. The reverse direction (view → edit, via router `state`) still has the same refresh limitation and is untouched here — happy to follow up if we want it made refresh-safe too.
- In-modal config edits made *after* switching to view are still not URL/storage-persisted (unchanged, pre-existing).

---

### DIFF
--- a/frontend/src/constants/sessionStorage.ts
+++ b/frontend/src/constants/sessionStorage.ts
@@ -1,3 +1,4 @@
 export enum SESSIONSTORAGE {
 	RETRY_LAZY_REFRESHED = 'retry-lazy-refreshed',
+	VIEW_PANEL_HANDOFF = 'view-panel-handoff',
 }
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/hooks/__tests__/useSwitchToViewMode.test.ts
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/hooks/__tests__/useSwitchToViewMode.test.ts
@@ -1,5 +1,10 @@
 import { renderHook } from '@testing-library/react';
+import type { DashboardtypesPanelSpecDTO } from 'api/generated/services/sigNoz.schemas';
 import { PANEL_TYPES } from 'constants/queryBuilder';
+import {
+	clearViewPanelHandoff,
+	readViewPanelHandoff,
+} from 'pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/ViewPanelModal/viewPanelHandoffStore';
 import type { Query } from 'types/api/queryBuilder/queryBuilderData';
 
 import { useSwitchToViewMode } from '../useSwitchToViewMode';
@@ -18,11 +23,16 @@ jest.mock('hooks/useUrlQuery', () => ({
 }));
 
 const query = { queryType: 'builder' } as unknown as Query;
+const spec = {
+	plugin: { kind: 'signoz/TimeSeriesPanel' },
+	display: { name: 'CPU' },
+} as unknown as DashboardtypesPanelSpecDTO;
 
 describe('useSwitchToViewMode', () => {
 	beforeEach(() => {
 		jest.clearAllMocks();
 		mockSearch = '';
+		clearViewPanelHandoff();
 	});
 
 	function invoke(): void {
@@ -32,6 +42,7 @@ describe('useSwitchToViewMode', () => {
 				panelId: 'panel-1',
 				panelType: PANEL_TYPES.TIME_SERIES,
 				query,
+				spec,
 			}),
 		);
 		result.current();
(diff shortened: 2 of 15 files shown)


frontend/src/constants/sessionStorage.ts
```diff
@@ -1,3 +1,4 @@
 export enum SESSIONSTORAGE {
 	RETRY_LAZY_REFRESHED = 'retry-lazy-refreshed',
+	VIEW_PANEL_HANDOFF = 'view-panel-handoff',
```
AshwinBhatkal, line 3: can we unify the handoff across view and edit?

AshwinBhatkal: instead of navigation state and storage

ahrefabhi: add events + switch to edit mode(isEditable and locked) check

AshwinBhatkal (approved):

PR #8107, opened 2025-05-30
feat(sharder): add simple and noop sharder

grandwizard28:
## 📄 Summary

add simple and noop sharder

----

> [!IMPORTANT]
> Introduces a new sharding mechanism with simple and noop sharders, integrating it into the existing system and updating configurations, modules, and tests accordingly.
> 
>   - **Sharder Implementation**:
>     - Introduces `noopsharder` and `singlesharder` in `pkg/sharder`.
>     - Adds `Sharder` interface in `sharder.go` with methods `GetMyOwnedKeyRange()` and `IsMyOwnedKey()`.
>     - Configures sharder in `config.go` and integrates with `signoz.go`.
>   - **Configuration**:
>     - Updates `example.yaml` to include sharder configuration.
>     - Adds `NewConfigFactory()` for sharder in `config.go`.
>   - **Integration**:
>     - Modifies `signoz.go` to initialize sharder and integrate with existing modules.
>     - Updates `provider.go` to include sharder provider factories.
>   - **Modules and Handlers**:
>     - Updates `module.go` and `handler.go` to use sharder in organization module.
>     - Refactors organization module to separate `Getter` and `Setter` interfaces.
>   - **SQL Migration**:
>     - Adds migration `036_add_key_organization.go` to add a key column to organizations.
>   - **Testing**:
>     - Updates integration tests to use new sharder setup in various test files.
>     - Ensures test utilities support sharder configuration.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SigNoz%2Fsignoz&utm_source=github&utm_medium=referral)<sup> for 9b1cd3a7e25ba3a68dc22b0f3ba07b15c0d3dfe7. You can [customize](https://app.ellipsis.dev/SigNoz/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #12060, opened 2026-07-09
fix(rules): prevent early return in initiate when an org has zero rules

### LINKED ISSUE #12059: [bug]: Rules manager initiate() silently drops alerts for subsequent orgs when an org has no rules
## Bug description

In a multi-org SigNoz deployment, when the rule manager initializes at startup (`initiate()`), it iterates over all organizations to load their stored alert rules. However, if any organization in the list has no rules, the function executes `return nil` immediately. This silently terminates the entire startup loop, abandoning all subsequent organizations and their rules. 

**Affected code** — `pkg/query-service/rules/manager.go`, lines 262–264:
```go
		if len(storedRules) == 0 {
			return nil // BUG: exits the entire function instead of just skipping this org
		}
```
As a result, any organization that happens to be iterated *after* an empty-rules org will have **zero alert rules loaded at startup**. All alerts for those orgs are silently lost.

## Expected behaviour

When an organisation has no stored rules (`len(storedRules) == 0`), the loop should skip that organisation and continue processing rules for all remaining organisations. No alerts from any organisation should be silently dropped at startup. 
The code should be:
```go
		if len(storedRules) == 0 {
			continue
		}
```

## How to reproduce

1. Create two organisations in SigNoz (multi-org setup).
2. In the first organisation (the one that appears first in the database), do **not** create any alert rules.
3. In the second organisation, create at least one alert rule.
4. Restart the SigNoz query-service / backend.
5. Observe: The second organisation's alert rules are not scheduled for evaluation after startup, and no error is logged.

## Version information 
* **Signoz version**: v0.132.0 (current `main` branch, introduced in commit `91cbd1727551621a863828b7280f5f7e15b39ca7`)
* **Browser version**: N/A (Backend issue)
* **Your OS and version**: Any (Linux/macOS/Docker)
* **Your CPU Architecture**(ARM/Intel): Any

## Additional context

The default sharder (`noopsharder`) returns the full key range `[0, MaxUint32]`, so `ListByOwnedKeyRange` returns **all organisations** in the database. This means the bug is active in every standard self-hosted deployment that has more than one organisation.

This was likely a copy-paste oversight from when this function was refactored from a single-org loop to a multi-org loop (PR #8107). The impact is a silent failure where the system appears healthy but monitoring coverage is dropped for subsequent organisations.

MannanSaood:
## Pull Request

---

This change fixes a critical bug where the alert rules manager (`initiate()`) aborts its entire startup loop if it encounters an organization with zero rules. Because it was using `return nil`, it silently skipped all organizations that came after the empty one in the database, completely dropping their alert coverage at startup. Changing this to `continue` ensures that an empty organization is simply skipped and the remaining organizations load their rules properly.

N/A (Backend logic fix)

Closes #12059 

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

**Refactor regression:** When the rule manager was refactored from a single-org design to support multiple organizations (in PR #8107), the early-return logic for an empty rule set (`if len(storedRules) == 0 { return nil }`) was left unchanged. In the context of the new `for _, org := range orgs` loop, this mistakenly terminated the entire function instead of just the current iteration.

Changed `return nil` to `continue` on line 268 of `pkg/query-service/rules/manager.go` inside the `initiate()` function.

---

- Tests added/updated: Explicitly not required (1-line logic fix).
- Manual verification: Verified the loop logic in `pkg/query-service/rules/manager.go`. With `continue`, the `initiate()` loop correctly moves to the next organization in the slice without halting execution.
- Edge cases covered: Multi-org deployments where early organizations have no rules configured.

---

- Blast radius: Extremely low. This purely restores intended initialization flow.
- Potential regressions: None expected.
- Rollback plan: Revert this commit if it unexpectedly impacts startup performance (unlikely).

---

| Field | Value |
|------|-------|
| Deployment Type | OSS / Enterprise |
| Change Type | Bug Fix |
| Description | Fixed a bug where alert rules would silently fail to load for organizations at startup if a preceding organization had no rules configured. |

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

This is a straightforward 1-line fix addressing a copy-paste oversight from when multi-org support was introduced. Let me know if you need me to write an explicit unit test for this edge case!

### DIFF
--- a/pkg/query-service/rules/manager.go
+++ b/pkg/query-service/rules/manager.go
@@ -260,7 +260,7 @@ func (m *Manager) initiate(ctx context.Context) error {
 			return err
 		}
 		if len(storedRules) == 0 {
-			return nil
+			continue
 		}
 
 		for _, rec := range storedRules {


eeshsaxena (commented):
Good catch - this looks correct to me. The `return nil` sits inside the `for _, org := range orgs` loop in `initiate`:

```go
for _, org := range orgs {
    if err != nil {
        return err
    }
    if len(storedRules) == 0 {
        return nil        // exits initiate() entirely
    }
    for _, rec := range storedRules { ... }
}
```

So the first org (in iteration order) that happens to have zero rules aborts the whole startup loop, and every org after it silently never gets its rules loaded. On a multi-tenant instance that's a real problem - a single ruleless org shadows all the others. Switching to `continue` correctly skips just that org and moves on, which matches the intent of the surrounding per-org loop (it already accumulates `loadErrors` and keeps going).

I couldn't spot any reason the early `return nil` would be intentional - there's nothing after the loop that a ruleless org should short-circuit. A small regression test with two orgs (first empty, second with a rule) asserting the second's rule is loaded would lock this in.

Deferring to the maintainers.

PR #12062, opened 2026-07-09
fix(dashboard-v2): validate key:value tags in the editor (inline-edit + backend rules)

AshwinBhatkal:
## Summary

Validation for the strict `key:value` tag editor (`TagKeyValueInput`).

- **Inline chip edit** now surfaces errors: editing an existing chip to an invalid or duplicate value shows the same inline error as the new-tag input and keeps the edit box open on Enter, while still reverting on blur (a click-away is never trapped).
- **Backend-rule validation**: tags are validated client-side against the backend rules in `pkg/types/tagtypes/tag.go` — key/value character allowlists (no spaces), a 32-char limit, and duplicates — via one shared `validateTag` used by both the new-tag input and inline edit. An invalid tag now surfaces inline instead of round-tripping to a 400.

## Test plan
- `TagKeyValueInput.test.tsx` (userEvent): invalid/duplicate/valid inline edits, revert-on-blur, and key/value space rejection on both the new-tag input and inline edit.

> The error-modal backend-message fix that was previously here has moved to its own PR (#12215).

### Screen Recording

https://github.com/user-attachments/assets/f8ce1763-ef67-495f-95a3-17e7b53e6594

### DIFF
--- a/frontend/src/components/TagKeyValueInput/TagKeyValueInput.test.tsx
+++ b/frontend/src/components/TagKeyValueInput/TagKeyValueInput.test.tsx
@@ -1,0 +1,112 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import TagKeyValueInput from './TagKeyValueInput';
+
+const TID = 'tag-key-value-input';
+
+type User = ReturnType<typeof userEvent.setup>;
+
+const startEditingFirstChip = async (user: User): Promise<HTMLElement> => {
+	await user.dblClick(screen.getAllByTestId(`${TID}-chip`)[0]);
+	return screen.getByTestId(`${TID}-edit`);
+};
+
+describe('TagKeyValueInput — inline chip edit', () => {
+	it('shows an error and stays in edit mode when Enter commits an invalid value', async () => {
+		const user = userEvent.setup();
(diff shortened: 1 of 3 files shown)


frontend/src/providers/ErrorModalProvider.tsx
```diff
@@ -29,8 +35,15 @@ export function ErrorModalProvider({
 	const [error, setError] = useState<APIError | null>(null);
 	const [isVisible, setIsVisible] = useState(false);
 
-	const showErrorModal = useCallback((error: APIError): void => {
-		setError(error);
+	const showErrorModal = useCallback((rawError: unknown): void => {
+		const apiError =
+			rawError instanceof APIError
+				? rawError
+				: convertToApiError(rawError as Parameters<typeof convertToApiError>[0]);
```
H4ad: nit: we could just make the convertToApiError(value: unknown) since we ignore/convert anyway any shape that is not similar to axios

H4ad (commented):
I remember a thread from design team not wanting to show any error message like in the video (blur + modal)

Confirm with design team later before landing this change for dashboards as well

H4ad (dismissed):
Ideally this should be divided in two PRs, one for provider error message and one for tag key

H4ad (approved):

PR #12065, opened 2026-07-10
fix: make span gaps accept day and week as units

namanv3:
## Pull Request

---

### 📄 Summary

Currently if someone types > 1d in span gaps then backend fails it. This is because valuer.TextDuration caps itself at hours as accepted unit.

To keep the blast radius minimal, SpanGaps struct gets it own validations instead of changing valuer.TextDuration

#### Issues closed by this PR

https://github.com/SigNoz/engineering-pod/issues/5668

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---


pkg/types/dashboardtypes/perses_signoz_plugins.go
```diff
@@ -621,8 +622,35 @@ func (fm *FillMode) UnmarshalJSON(data []byte) error {
...
+	return sg.validate()
+}
+
+// validate checks that FillLessThan, when set, is a valid positive duration.
+// It uses prometheus's parser so day/week/year units (e.g. "1d") are accepted;
+// time.ParseDuration caps at hours.
+func (sg SpanGaps) validate() error {
+	if sg.FillLessThan == "" {
+		return nil
+	}
```
srikanthccv: it can be allowed to be empty only when the `FillOnlyBelow` is false right? otherwise it should be rejected that it can't be empty.

namanv3: fixed

srikanthccv (approved):

PR #12076, opened 2026-07-10
fix(infra-monitoring-v2): ensure workloads has filters by cluster+namespace

H4ad:
## Pull Request

---

This PRs aims to fix a behavior we had since the introduction of infra monitoring, some workloads names could not be unique across clusters and namespaces.

Before:

screenshot-2026-07-09_18-19-12

After:

screenshot-2026-07-09_18-19-40

Fixes https://github.com/SigNoz/engineering-pod/issues/5398

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

If we have a jobs called `job-1` defined for namespace `a` and `b`, if we open the details:
- metrics/charts will sum metrics of both jobs
- logs will show logs of both jobs
- traces will show traces of both jobs
- events will show events of both jobs

Because the only thing we are filtering is by the name, all these tabs sum the identical jobs.

The issue is not present for:
- clusters
- nodes
- pods
- hosts

Because they usually have an unique name that don't cause any issue even across the clusters+namespaces.

The lack of filters for cluster + namespaces.

For the following workloads, I added cluster+namespace:
- jobs
- statefulsets
- daemon-sets
- deployments
- volumes

And for the following workloads, I added only cluster:
- namespace

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Infrastructure Monitoring V2
- Potential regressions: Missing a filter or wrong checks
- Rollback plan: Find and fix the issue specifically.

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | For Infrastructure Monitoring V2, workloads such as jobs, daemonsets, etc... with duplicated names will not be summarized anymore due to lack of filter by cluster+namespace.  |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

Few gotchas:
- I had to change the tanstack table to support itemKey as `object`.
  - This is needed since I now need to convert the row itemKey to `{jobName, cluster, namespace}` instead of just jobName.
  - And instead of storing the `selectedItem=<json>`, I can write something like: `selectedItem=<value>, selectedClusterName=<value>`.
- This started just to add few filters to fetch each workload, but since I identified the bugs in each tab, I pushed fix for everything.
- For events tab, I had to use `attribute.k8s.namespace.name`, otherwise, it won't return any value since namespace is stored as attribute instead of resource, like cluster.
  - And query builder works without prefix of `resource` for a reason that I don't know

### DIFF
--- a/frontend/src/container/InfraMonitoringK8sV2/Base/K8sBaseList.tsx
+++ b/frontend/src/container/InfraMonitoringK8sV2/Base/K8sBaseList.tsx
@@ -50,7 +50,10 @@ export type K8sBaseListEmptyStateContext = {
 /** Base type constraint for K8s entity data */
 export type K8sEntityData = { meta?: Record<string, string> | null };
 
-export type K8sBaseListProps<T extends K8sEntityData, TItemKey = string> = {
+export type K8sBaseListProps<
+	T extends K8sEntityData,
+	TItemKey extends string | SelectedItemParams = string,
+> = {
 	controlListPrefix?: React.ReactNode;
 	leftFilters?: React.ReactNode;
 	entity: InfraMonitoringEntity;
@@ -79,7 +82,10 @@ export type K8sBaseListProps<T extends K8sEntityData, TItemKey = string> = {
 	extraQueryKeyParts?: string[];
 };
 
-export function K8sBaseList<T extends K8sEntityData, TItemKey = string>({
+export function K8sBaseList<
+	T extends K8sEntityData,
+	TItemKey extends string | SelectedItemParams = string,
+>({
 	controlListPrefix,
 	leftFilters,
 	entity,
@@ -229,10 +235,10 @@ export function K8sBaseList<T extends K8sEntityData, TItemKey = string>({
 		(_record: T, itemKey: TItemKey): void => {
 			if (groupBy.length === 0) {
 				if (typeof itemKey === 'object' && itemKey !== null) {
-					setSelectedItemParams(itemKey as unknown as SelectedItemParams);
+					setSelectedItemParams(itemKey);
 				} else {
 					setSelectedItemParams({
-						selectedItem: itemKey as string,
+						selectedItem: itemKey,
 						clusterName: null,
 						namespaceName: null,
 					});
@@ -257,7 +263,7 @@ export function K8sBaseList<T extends K8sEntityData, TItemKey = string>({
 			// Build URL with selectedItem params
 			const url = new URL(window.location.href);
 			if (typeof itemKey === 'object' && itemKey !== null) {
-				const params = itemKey as unknown as SelectedItemParams;
+				const params = itemKey;
 				if (params.selectedItem) {
 					url.searchParams.set(
 						INFRA_MONITORING_K8S_PARAMS_KEYS.SELECTED_ITEM,
@@ -279,7 +285,7 @@ export function K8sBaseList<T extends K8sEntityData, TItemKey = string>({
 			} else {
 				url.searchParams.set(
 					INFRA_MONITORING_K8S_PARAMS_KEYS.SELECTED_ITEM,
-					itemKey as string,
+					itemKey,
 				);
 			}
 			openInNewTab(url.pathname + url.search);
--- a/frontend/src/container/InfraMonitoringK8sV2/Base/K8sBaseList.tsx
+++ b/frontend/src/container/InfraMonitoringK8sV2/Base/K8sBaseList.tsx
@@ -10,7 +10,6 @@ import TanStackTable, {
 } from 'components/TanStackTableView';
 import { InfraMonitoringEvents } from 'constants/events';
 import { useQueryBuilder } from 'hooks/queryBuilder/useQueryBuilder';
-import { parseAsString, useQueryState } from 'nuqs';
 import { useGlobalTimeStore } from 'store/globalTime';
 import { NANO_SECOND_MULTIPLIER } from 'store/globalTime/utils';
 import { Querybuildertypesv5QueryWarnDataDTO } from 'api/generated/services/sigNoz.schemas';
@@ -21,8 +20,10 @@ import {
 	InfraMonitoringEntity,
 } from '../constants';
 import {
+	SelectedItemParams,
 	useInfraMonitoringGroupBy,
 	useInfraMonitoringOrderBy,
+	useInfraMonitoringSelectedItemParams,
 	useInfraMonitoringStatusFilter,
 } from '../hooks';
 import { useInfraMonitoringLineClamp } from '../components';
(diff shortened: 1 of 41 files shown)


frontend/src/container/InfraMonitoringK8sV2/Base/K8sExpandedRow.tsx
```diff
@@ -175,10 +176,18 @@ export function K8sExpandedRow<T>({
 	const expandedData = data?.data ?? [];
 
 	const handleRowClick = useCallback(
-		(_row: T, itemKey: string): void => {
-			void setSelectedItem(itemKey);
+		(_row: T, itemKey: TItemKey): void => {
+			if (typeof itemKey === 'object' && itemKey !== null) {
+				setSelectedItemParams(itemKey as unknown as SelectedItemParams);
```
AshwinBhatkal, line 181: What will it take to remove this unknown cast?

frontend/src/container/InfraMonitoringK8sV2/Base/utils.tsx
```diff
@@ -89,3 +94,91 @@ export function buildExpressionFromGroupMeta(
 	}
 	return parent || metaExpression;
 }
+
+export interface EventsExpressionParams {
+	objectKind: string;
+	objectName: string;
+	clusterName?: string | null;
```
AshwinBhatkal, line 101: can it be one of null or undefined?

frontend/src/container/InfraMonitoringK8sV2/Base/utils.tsx
```diff
@@ -89,3 +94,91 @@ export function buildExpressionFromGroupMeta(
...
+			`attribute.${INFRA_MONITORING_ATTR_KEYS.K8S_NAMESPACE_NAME} = ${formatValueForExpression(params.namespaceName)}`,
+		);
+	}
+
+	return clauses.join(' AND ');
+}
+
+export interface LogsTracesExpressionParams {
+	mainAttributeKey: string;
+	mainAttributeValue?: string | null;
```
AshwinBhatkal, line 130: same here

frontend/src/container/InfraMonitoringK8sV2/Namespaces/table.config.tsx
```diff
@@ -29,8 +30,15 @@ export function getK8sNamespaceRowKey(
 
 export function getK8sNamespaceItemKey(
 	namespace: InframonitoringtypesNamespaceRecordDTO,
-): string {
-	return namespace.namespaceName;
+): SelectedItemParams {
+	return {
+		selectedItem:
+			namespace.namespaceName ??
+			namespace.meta?.[INFRA_MONITORING_ATTR_KEYS.K8S_NAMESPACE_NAME] ??
```
AshwinBhatkal, line 37: can we ensure this is always defined? What will it take if so?

frontend/src/container/InfraMonitoringK8sV2/Volumes/table.config.tsx
```diff
@@ -1,6 +1,5 @@
-import { TableColumnDef } from 'components/TanStackTableView';
+import TanStackTable, { TableColumnDef } from 'components/TanStackTableView';
```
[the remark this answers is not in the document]
AshwinBhatkal: No error thrown or is it used later?

AshwinBhatkal (commented):
few comments but more or less can be applied to all the changes

frontend/src/container/InfraMonitoringK8sV2/Volumes/table.config.tsx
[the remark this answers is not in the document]
H4ad: This is used later, TanStackTable.Text

frontend/src/container/InfraMonitoringK8sV2/Base/K8sBaseList.tsx
```diff
@@ -237,18 +245,43 @@ export function K8sBaseList<T extends K8sEntityData>({
...
 	const handleRowClickNewTab = useCallback(
-		(_record: T, itemKey: string): void => {
+		(_record: T, itemKey: TItemKey): void => {
 			if (groupBy.length > 0) {
 				return;
 			}
 
-			// Build URL with selectedItem param
+			// Build URL with selectedItem params
 			const url = new URL(window.location.href);
```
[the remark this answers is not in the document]
H4ad: This cannot be triggered with these params defined

frontend/src/container/InfraMonitoringK8sV2/Base/utils.tsx
H4ad: It can, almost all the times is not, but it can

frontend/src/container/InfraMonitoringK8sV2/Base/utils.tsx
H4ad: It's highly unlikely it to be null but it can happen (under conditions I'm not sure)

frontend/src/container/InfraMonitoringK8sV2/Namespaces/table.config.tsx
H4ad: You need to force types on backend for each `.meta` object, instead of `Record<string, string>`.

frontend/src/container/InfraMonitoringK8sV2/Base/K8sExpandedRow.tsx
H4ad: I thought it would take more effort but took just few line changes

frontend/src/container/InfraMonitoringK8sV2/Base/utils.tsx
AshwinBhatkal: Which case is that?

frontend/src/container/InfraMonitoringK8sV2/Base/utils.tsx
AshwinBhatkal: If we are not sure, then we should verify and accordingly make the update

frontend/src/container/InfraMonitoringK8sV2/Namespaces/table.config.tsx
AshwinBhatkal: If that is possible, then we should do it

H4ad: Created https://github.com/SigNoz/engineering-pod/issues/5710

frontend/src/container/InfraMonitoringK8sV2/Base/utils.tsx
H4ad: Created https://github.com/SigNoz/engineering-pod/issues/5710

frontend/src/container/InfraMonitoringK8sV2/Base/utils.tsx
H4ad: cluster does not include cluster, namespace does not include filter for namespace

AshwinBhatkal (approved):

PR #12084, opened 2026-07-12
fix(rules): keep loading rules for orgs after an empty-rules org

### LINKED ISSUE #12059: [bug]: Rules manager initiate() silently drops alerts for subsequent orgs when an org has no rules

Harsh23Kashyap:
## Pull Request

---

### Summary

`Manager.initiate()` iterates every org at startup and loads its stored alert rules. When the iteration hit an org with zero stored rules, the function called `return nil` instead of `continue`, which exited `initiate()` early and silently dropped every org that came after an empty-rules org in the iteration order. In a multi-org deployment that includes one org with no rules, every later org ended up with no alert rules scheduled for evaluation at startup — alarms were silently lost.

This PR replaces the early return with `continue` so the loop moves on to the next org.

### Screenshots / Screen Recordings (if applicable)

N/A — backend fix; no UI surface change.

### Issues closed by this PR

- Closes #12059

### Change Type

- [x] Bug fix

#### Root Cause

`pkg/query-service/rules/manager.go` line 263 (in `Manager.initiate`):

```go
for _, org := range orgs {
    if err != nil {
        return err
    }
    if len(storedRules) == 0 {
        return nil   // BUG: exits the entire function instead of skipping this org
    }
    ...
}
```

`return nil` inside the `for _, org := range orgs` loop terminates the whole function, not just the current iteration. Any org that sorts after an empty-rules org in the result of `m.orgGetter.ListByOwnedKeyRange(ctx)` never gets its rules loaded.

#### Fix Strategy

Replace `return nil` with `continue` so the loop skips the empty org and keeps processing the rest. Identical suggested fix is included in the issue body, so this is the maintainer-acknowledged change.

### Testing Strategy

- Tests added/updated: none — see "Notes" below.
- Manual verification:
  - `go build ./pkg/query-service/rules/...` — pass
  - `go vet ./pkg/query-service/rules/...` — pass
  - `go test ./pkg/query-service/rules/ -count=1` — pass (1.4s, all existing tests)
- Edge cases covered: the fix is a one-token control-flow change (`return nil` → `continue`) inside a function whose body is otherwise unchanged. Behavior of the existing per-org path (rule unmarshalling, alertmanager notification config, `addTask` for non-disabled rules) is identical.

### Risk & Impact Assessment

- Blast radius: rule manager startup only. No runtime rule evaluation, no API path, no DB schema change.
- Potential regressions: a `continue` here causes the loop to skip the `for _, rec := range storedRules` block (which is empty in this branch anyway) and proceed to the next org. The existing test suite passes; the behavior is provably identical for orgs with rules and now correct for orgs without rules.
- Rollback plan: revert the single-line change. No data migration, no schema, no compatibility concern.

### Changelog

| Field | Value |
|---|---|
| Deployment Type | OSS / Enterprise / Cloud |
| Change Type | Bug Fix |
| Description | Multi-org SigNoz deployments no longer silently drop alert rules for any org that sorts after an org with zero rules. |

### Checklist

- [x] Tests added or explicitly not required (see "Notes")
- [x] Manually tested
- [x] Backward compatibility considered (no behavior change for orgs with rules)

### Notes

A regression test for the multi-org skip-org behavior would require mocking `organization.Getter.ListByOwnedKeyRange` to return two orgs and `ruleStore.GetStoredRules` to return `[]` for the first and `[{...}]` for the second. The existing `NewTestManager` helper does not currently inject an `orgGetter` mock (it uses the real one from sqlstore), and adding that plumbing is a much larger change than the one-line fix it would test. Given the fix is a single-token control-flow change with a clear comment and a suggested identical fix from the issue author, the test was scoped out of this PR. A follow-up issue to add a proper `orgGetter` mock to the rules test suite is the right next step if the maintainers want regression coverage.

---

### External links

- Issue: https://github.com/SigNoz/signoz/issues/12059


Harsh23Kashyap:
Can someone take a look when you have a chance?

Harsh23Kashyap:
Can someone take a look when you have a chance?

Harsh23Kashyap:
@MannanSaood thanks for taking it — the one-line `return nil` → `error` change is exactly the fix. Let me know once #12084 merges and I'll close my version #12084 too.

Harsh23Kashyap:
CI still pending on the merge queue. No new reviewer comments since the last update. (rules manager continue after empty)

Harsh23Kashyap:
MannanSaood's review note is in (the one-line `return nil` → `error` change). CI still pending. Standing by for a maintainer pass.

Harsh23Kashyap:
Synced with upstream. Rules manager continue after empty-rules org. Synced (8→0).

Harsh23Kashyap:
Closing this in favor of #12060 — same one-line `return nil` → `continue` fix in `pkg/query-service/rules/manager.go`. My version added a 3-line explanatory comment but otherwise duplicates the work. Will keep the branch and rebase onto a fresh upstream/main for the next thing.

PR #12089, opened 2026-07-13
fix: has function family revamp

tushar-signoz:
## Pull Request

---

### 📄 Summary

`has`, `hasAny`, `hasAll`, `hasToken` search arrays and tokens inside the log `body`. This PR fixes their behaviour in **both** storage modes and makes the two consistent.

- **Legacy** (`use_json_body` OFF) — body is a raw JSON string.
- **JSON** (`use_json_body` ON) — body is native ClickHouse JSON (`body_v2`).

#### Legacy mode

No schema, so the element type is inferred from the **needle** and the array is extracted as `Array(Nullable(<type>))`:

| needle | extracted as |
|---|---|
| `'text'` | `String` |
| `"200"` (quoted integer) | `Int64` — exact past 2^53 |
| `200` (unquoted) | `Float64` |
| `1.5` | `Float64` |
| `true` / `"true"` | `String` (rendered `'true'`/`'false'`) |

**`hasAny` / `hasAll` on a body array**
- **Ingested** `{"tags": ["staging", "api", "test"]}`
- **Query** `hasAny(body.tags, ['critical', 'test'])`
- **Before** → **500**, ClickHouse "Argument 0 must be an array" (the path extracted a single string, not an array)
- **After** → **200**, matches (shares `test`)

**`has` on a numeric array**
- **Ingested** `{"ids": [100, 200, 300]}`
- **Query** `has(body.ids, "200")`
- **Before** → **500** (same array-extraction bug)
- **After** → **200**, matches — the quoted `"200"` is compared numerically as `Int64`

**`has` on a large-integer array — precision (2^53)**
- **Ingested** `{"id": [1234567890123456788]}` (a neighbour of the searched id)
- `has(body.id, "1234567890123456789")` (quoted → `Int64`) → **no match** — exact
- `has(body.id, 1234567890123456789)` (unquoted → `Float64`) → **false-matches** the `…788` neighbour, because an unquoted literal is parsed as `float64` upstream. Quote large numeric ids to keep them exact.

**`has*` on a scalar (non-array) path**
- **Ingested** `{"level": "info"}`
- **Query** `has(body.level, 'info')`
- **Before** → **500** (ClickHouse rejected an array function on a scalar)
- **After** → **200**, matches — a scalar is treated as a single-element array (an OR-ed scalar comparison alongside the array form)

> ⚙️ **Safe by construction.** No ClickHouse "no supertype" (code 386) error ever — the element type is chosen (by scanning *every* needle value) so the array type and all coerced needles agree; a type-mismatched needle (e.g. numeric vs a string array) simply matches nothing (`Nullable` → off-type elements become `NULL`, never a corrupt `0`). Mixed-type value lists (`[200, "abc"]`) and mixed-type body arrays are handled safely (string fallback / per-type view), never rejected. Validated against real ClickHouse.

**Type isolation** — a needle matches only its own type, so no false matches:
- `has(body.ids, true)` on `[100, 200, 300]` → **no match** (a bool needle is compared as the string `'true'`/`'false'`, so it matches only genuine booleans, not truthy numbers)
- `has(body.tags, '')` / `has(body.ids, 0)` on any array → **no match** (the scalar fallback is guarded by `JSONType` + `Nullable`, so it fires only for a genuinely scalar body)
- genuine scalars still match: `has(body.flag, true)` on `{"flag": true}`, `has(body.code, 0)` on `{"code": 0}` → **match**

#### JSON-body mode

**`has` on a scalar leaf reached through an array** (`SigNoz/engineering-pod#4620`)
- **Ingested** `{"education": [{"name": "IIT"}, {"name": "MIT"}]}`
- **Query** `has(body.education[].name, 'IIT')`
- **Before** → **400** "no array fields found" (nested scalar leaves weren't supported)
- **After** → **200**, matches (checks each element)
- Note: `hasAll` is contains-all here too — `hasAll(body.education[].name, ['IIT','MIT'])` = *some* element is IIT **and** *some* element is MIT (extra elements are fine).

**`has` on a numeric array — type collision** (`SigNoz/engineering-pod#4678`)
- **Ingested** `{"education": [{"scores": [90, 80]}]}`
- **Query** `has(body.education[].scores, 90)`
- **Before** → could silently **miss** when the literal's type didn't match the array's element type
- **After** → **200**, matches (same type-collision handling every other operator uses)

**`hasToken` on a JSON string field** (`SigNoz/engineering-pod#3793`)
- **Ingested** `{"education": [{"name": "harvard university"}]}`
- **Query** `hasToken(body.education[].name, 'harvard')`
- **Before** → **400** "only supports body/body.message field"
- **After** → **200**, matches (works on nested strings and string arrays too)

**`has*` on a scalar (non-array) path**
- **Ingested** `{"level": "info"}`
- **Query** `has(body.level, 'info')`
- **Before** → rejected (scalars unsupported)
- **After** → **200**, matches — the scalar leaf is compared directly, as a single-element set

#### Argument validation — both modes (`SigNoz/engineering-pod#4560`)
- `has(body.tags, "a", "b")` → **Before**: `"b"` silently dropped · **After**: **400** "expects exactly one value argument"
- `has(body.tags, ["a", "b"])` → **Before**: mis-handled · **After**: **400** "expects a single scalar value, not an array"
- `hasAny(body.tags, ["a"], "b")` → **After**: **400** "not a mix of the two"
- Mixed *element* types in a value list (e.g. `[200, "abc"]`) are **not** rejected — they're handled safely (see the legacy note above / JSON coerces per value).

> 💡 **`hasAll` = contains-all, uniformly (both modes, every leaf shape).** `hasAll(path, [v…])` means "the path HAS every listed value" — i.e. `has(v₁) AND has(v₂) AND …` (and `hasAny` = OR-of-`has`). It is false only when a listed value is *missing*, never because of an *extra* value. So `hasAll(body.tags, ['a','b'])` matches `["a","b","c"]`; `hasAll(body.education[].name, ['a','b'])` needs one element `a` and one element `b`; and for a plain scalar (a one-element set) a multi-value `hasAll` can't match while a single-value `hasAll` collapses to `has`. Validated against real ClickHouse.

> 🔁 **`NOT has*` = uniform complement (both modes, every leaf shape).** Matches every row that doesn't contain the value — **including missing-key and empty-array rows**. Scalar-leaf comparisons coalesce to false (`ifNull`), so an absent key is a non-match, not a NULL that drops the row. Validated against real ClickHouse.

#### Issues closed by this PR

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
- **Legacy (string body):** a body-array path was extracted as a single scalar with a loosely-guessed type, so array functions hit ClickHouse errors — `hasAny`/`hasAll` → 500 "Argument 0 must be an array", `has` on a scalar path → 500, and numeric/bool/large-int needles mismatched or lost precision. Extra / mis-shaped function arguments were silently dropped.
- **JSON body:** scalar leaves reached through an array (`body.education[].name`) were rejected (400 "no array fields found"); `hasToken` accepted only `body`/`body.message`; numeric arrays could silently miss when the literal's type didn't match the element type.

#### Fix Strategy
- **Legacy:** infer the element type from the needle, extract as `Array(Nullable(<type>))`, and OR it with a scalar comparison guarded by `JSONType`/`Nullable` so genuinely-scalar bodies match too. Every needle is coerced to the chosen type, and the type is picked by scanning *all* values so the array type and needles always agree → never a "no supertype" (386) error; off-type needles become `NULL` and match nothing.
- **JSON:** support nested scalar leaves, apply the same type-collision handling everywhere, extend `hasToken` to nested JSON string fields/arrays, and make `hasAll` uniformly contains-all (`has(v₁) AND has(v₂) …`).
- **Both:** validate arguments — reject extra values, array-as-scalar, and mixed arg forms with a 400 instead of silently dropping them. Coalesce scalar-leaf comparisons to false (`ifNull`) so `NOT has*` uniformly matches missing-key rows.

---

### 🧪 Testing Strategy

- **Tests added/updated:** Go golden-SQL unit tests (legacy + JSON) and integration suites — see file list under *Notes for Reviewers*.
- **Manual verification:** validated against real ClickHouse.
- **Edge cases covered:** type-match, large-int precision (2^53), no-supertype safety, mixed types, nested paths, type isolation, negation over missing keys / empty arrays, and argument validation.

---

- **Blast radius:** `has` / `hasAny` / `hasAll` / `hasToken` over the log `body` (both legacy and JSON body modes). No other operators are affected.
- **Potential regressions:** the generated ClickHouse SQL for these functions changed (golden SQL updated to match); behaviour is now stricter on malformed arguments (400 instead of silently dropping).
- **Rollback plan:** revert the PR; the JSON-body path is additionally gated behind the `use_json_body` flag.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | `has`/`hasAny`/`hasAll`/`hasToken` now work correctly and consistently on log-body arrays, nested JSON paths, and scalar values in both legacy and JSON body modes. |

---

### 📋 Checklist
- [x] Manually tested

---


tushar-signoz:
## Code Review

### Overview

Fixes `has`/`hasAny`/`hasAll`/`hasToken` over the log body in both storage modes and makes them consistent:

- **Shared visitor**: `normalizeFunctionValue` shape-validates function arguments (400 instead of silently dropping extras) and folds `hasAny(k, 'a', 'b')` into one value set.
- **Legacy mode**: needle-driven type inference (`legacyElemType`), extraction as `Array(Nullable(<type>))`, and an OR-ed scalar fallback guarded by `JSONType` — fixes the 500s ("Argument 0 must be an array") and precision/coercion quirks.
- **JSON mode**: `buildArrayFunctionCondition` supports scalar leaves reached through arrays (`body.education[].name`), applies collision handling to array elements, extends `hasToken` to nested JSON string fields/arrays, and makes `hasAll` uniformly contains-all (AND of per-value `has`).
- **Negation**: `NOT has*` is a uniform complement across every leaf shape — scalar-leaf comparisons coalesce to false (`ifNull`) at three sites (JSON scalar leaf, JSON `hasToken` string leaf, legacy scalar fallback), so missing-key and empty-array rows match the negation instead of being NULL-dropped. Positive semantics unchanged.
- Removes the "no array fields found" pre-filter, since scalars are now first-class.

### Correctness — verified

- `normalizeFunctionValue` can't see an empty list (`len(params) < 2` guard); the `[]any{values}` wrapper matches the condition builders' unwrap.
- JSON path no longer calls `fm.FieldFor` — safe: `key.JSONPlan` is populated during key resolution, same as the existing primitive path; both builders guard on empty `JSONPlan`.
- Blast radius is logs-only: `conditionForArrayFunction` rejects non-body keys; no traces/metrics builder handles has-family operators.
- `legacyElemType` picks `Int64`/`Float64` only if *every* value coerces, so `legacyCoerceNeedle` never hits an unparseable string.
- Zero-value guards: `Nullable(...)` extraction + `JSONType(...) NOT IN ('Array','Object','Null')` prevent `has(x, 0)` / `has(x, '')` false matches; genuine `0`/`''` elements still match (tested).

### Non-blocking notes

1. **Legacy per-row cost** — each legacy `has` now does ~3 JSON parses per row instead of 1 (array extraction + `JSON_VALUE` + `JSONType`). Acceptable on a deprecated path; where to look if a tenant runs hot on flag-off body filters.
2. **JSON-mode `hasAll` fan-out** — N values × plan roots × branches each emit an `arrayExists` chain; large WHERE clauses for long value lists over deep promoted paths. The switch away from `arrayFlatten(arrayConcat(...))` likely wins on deep paths via short-circuiting.
3. **Key interpolation** — key parts go unescaped into `JSONType(body, 'k')`, mirroring the pre-existing `getBodyJSONPath` pattern; two places to fix if key quoting is ever hardened.
4. **Dropped `assumeNotNull`** in `branchArrayExpr` vs the old nested-hasAny SQL — safe given the `dynamicType(x) = 'JSON'` pre-filter.

### Tests

Golden SQL for every new shape, plus integration coverage of the exact failure modes from the linked issues: type isolation, 2^53 precision, no-supertype safety, mixed types, negation over missing keys / empty arrays (incl. a legacy numeric-needle regression pair), unicode/escaping, and the argument-shape 400s. Verified against real ClickHouse (flag-on suite on CH 25.12.5, legacy suite on 25.5.6); `go build` + full `go test ./pkg/...` green; CI green.

### Verdict

Approve. Design is coherent (contains-all and uniform-complement semantics across both modes and every leaf shape), the risky removals (`FieldFor` call, array-key pre-filter) are safe on inspection, and behavior is locked in by golden + integration tests.

pkg/telemetrylogs/filter_expr_logs_body_json_test.go
```diff
@@ -44,34 +44,66 @@ func TestFilterExprLogsBodyJSON(t *testing.T) {
...
+			shouldPass:            true,
+			expectedQuery:         `WHERE (hasAll(JSONExtract(JSON_QUERY(body, '$."tags"[*]'), 'Array(Nullable(String))'), ?) OR ifNull(((JSON_VALUE(body, '$."tags"') = ? AND JSON_VALUE(body, '$."tags"') = ?) AND JSONType(body, 'tags') NOT IN ('Array', 'Object', 'Null')), false))`,
+			expectedArgs:          []any{[]any{"production", "web"}, "production", "web"},
+			expectedErrorContains: "",
+		},
+		{
+			category:              "json",
+			query:                 "has(body.ids, \"200\")",
+			shouldPass:            true,
+			expectedQuery:         `WHERE (has(JSONExtract(JSON_QUERY(body, '$."ids"[*]'), 'Array(Nullable(Int64))'), ?) OR ifNull((JSONExtract(JSON_VALUE(body, '$."ids"'), 'Nullable(Int64)') = ? AND JSONType(body, 'ids') NOT IN ('Array', 'Object', 'Null')), false))`,
```
krsoninikhil, line 103: Will this match if the stored value is string `"200"` as well?

pkg/telemetrylogs/filter_expr_logs_body_json_test.go
```diff
@@ -44,34 +44,66 @@ func TestFilterExprLogsBodyJSON(t *testing.T) {
...
-			expectedQuery:         `WHERE NOT (has(JSONExtract(JSON_QUERY(body, '$."nested_num"[*]."float_nums"[*]'), 'Array(Float64)'), ?))`,
+			expectedQuery:         `WHERE NOT (has(JSONExtract(JSON_QUERY(body, '$."nested_num"[*]."float_nums"[*]'), 'Array(Nullable(Float64))'), ?))`,
 			expectedArgs:          []any{float64(2.2)},
 			expectedErrorContains: "",
 		},
+		{
+			category:              "json",
+			query:                 "has(body.tags, 'production')",
+			shouldPass:            true,
+			expectedQuery:         `WHERE (has(JSONExtract(JSON_QUERY(body, '$."tags"[*]'), 'Array(Nullable(String))'), ?) OR ifNull((JSON_VALUE(body, '$."tags"') = ? AND JSONType(body, 'tags') NOT IN ('Array', 'Object', 'Null')), false))`,
```
krsoninikhil, line 79: We were doing `lower()` on value ([ref](https://github.com/SigNoz/signoz/pull/12089/changes#diff-4acd142f387e66b74e69df26ed902bc54f98469c96a6b13f8e59d2e9b1eca50aL1552)) in log body query earlier, is that behavior going to change now?

pkg/telemetrylogs/filter_expr_logs_body_json_test.go
tushar-signoz: For legacy yes

pkg/telemetrylogs/filter_expr_logs_body_json_test.go
tushar-signoz: HasToken uses `lower()`.
Has* don't

pkg/telemetrylogs/filter_expr_logs_body_json_test.go
krsoninikhil: What about json? Would `has(body.ids, "200")` not match the exact string?

pkg/telemetrylogs/filter_expr_logs_body_json_test.go
krsoninikhil: Can we make them user `lower()` as well for consistent behavior?

tushar-signoz: No we cannot. 
If we make has* case-insensitive. then we restict user from doing any case-sensitive work.
With the current behaviour user can always do`HasAny(body..., "ABC", "abc")`

tushar-signoz: If you're taking about consistent with main, legacy has* was not using lower, only hasToken was using lower

pkg/telemetrylogs/json_string.go
```diff
@@ -130,3 +130,121 @@ func GetBodyJSONKey(_ context.Context, key *telemetrytypes.TelemetryFieldKey, op
...
+		// Nullable so a scalar of a different type (e.g. a bool/string where a number is
+		// searched) extracts to NULL rather than the type's default (0/false), which would
+		// otherwise zero-value match has(x, 0).
+		expr = fmt.Sprintf("JSONExtract(JSON_VALUE(body, '$.%s'), 'Nullable(%s)')", path, dt.CHDataType())
+	}
+	keys := strings.Split(name, ".")
+	for i, k := range keys {
+		keys[i] = "'" + k + "'"
+	}
+	guard = fmt.Sprintf("JSONType(body, %s) NOT IN ('Array', 'Object', 'Null')", strings.Join(keys, ", "))
```
krsoninikhil, line 233: Is this list of `JSONType` exhaustive? Wondering if there is better way to handle the zero value match case for non-scaler values. 

How about this `function_json_value_return_type_allow_complex = true` setting? ([reference](https://clickhouse.com/docs/sql-reference/functions/json-functions#JSON_VALUE))

pkg/telemetrylogs/json_condition_builder.go
```diff
@@ -309,6 +305,174 @@ func (c *jsonConditionBuilder) buildArrayMembershipCondition(node *telemetrytype
...
+			return conditions[0], nil
+		}
+		return sb.And(conditions...), nil
+	}
+	return "", qbtypes.ErrUnsupportedOperator
+}
+
+// buildOredRootChains applies leafFn down every JSONPlan root (base + promoted), wrapping each
+// in its arrayExists chain, and ORs the per-root results.
+func (c *jsonConditionBuilder) buildOredRootChains(leafFn func(*telemetrytypes.JSONAccessNode) (string, error), sb *sqlbuilder.SelectBuilder) (string, error) {
```
krsoninikhil, line 357: `buildORedRootChains`?

pkg/telemetrylogs/condition_builder.go
```diff
@@ -92,28 +114,37 @@ func (c *conditionBuilder) conditionForHasToken(
...
 
-	columnName := LogsV2BodyColumn
-	if bodyJSONEnabled {
-		if key.Name != LogsV2BodyColumn && key.Name != bodyMessageField {
+	if !bodyJSONEnabled {
+		// legacy: token search over the plain body string column only.
+		if key.Name != LogsV2BodyColumn {
 			return "", errors.NewInvalidInputf(errors.CodeInvalidInput,
-				"function `hasToken` only supports body/body.message field as first parameter").WithUrl(hasTokenFunctionDocURL)
+				"function `hasToken` only supports body field as first parameter").WithUrl(hasTokenFunctionDocURL)
```
krsoninikhil, line 131: Is there any limitation to not support nested path for `hasToken` in legacy body if we are supporting that in json body?

krsoninikhil (commented):
Requested a few clarification, rest LGTM.

pkg/telemetrylogs/condition_builder.go
tushar-signoz: Not investing a whole lot in legacy body since we're moving to json body

pkg/telemetrylogs/filter_expr_logs_body_json_test.go
tushar-signoz: Json does exact type matches on whatever user provided

pkg/telemetrylogs/json_string.go
tushar-signoz: Let me create a ticket for this to look further into this.

pkg/telemetrylogs/json_condition_builder.go
tushar-signoz: For a json path a.b[].c[].d[], we build chains for each diverged path. We OR these chains from the root hence the name

tushar-signoz:
## Code review

**Verdict: no correctness bugs found.** The logic is sound across both storage modes. I traced the risky paths against real ClickHouse semantics and the code holds up.

### What I verified (the parts most likely to hide a bug)

- **Operator routing** — `has`/`hasAny`/`hasAll` are in `IsArrayFunctionOperator()` but **not** `IsArrayOperator()`. This matters: `InferDataType` only list-wraps a scalar when `IsArrayOperator()`, so a scalar `has(body.x, 'v')` needle is *not* spuriously wrapped into `['v']` (which would have produced `x = ['v']`). Correct by construction. `arrayFunctionLeaf`/`arrayLeafMembership` only ever receive `Has`/`HasAny`; `hasAll` composes from `has` outside the array hops.
- **Negation over missing keys** — traced `NOT has(body.level,'info')` in both modes. Legacy: the `JSONType(...) NOT IN ('Array','Object','Null')` guard → false for a missing key, whole thing `ifNull(...,false)` → `NOT` matches. JSON: `arrayFuncScalarLeaf`/`tokenLeaf` wrap scalar comparisons in `ifNull(cond,false)`; array leaves rely on `arrayExists([])=0`. Both coalesce to false, never NULL. Matches the stated semantics.
- **Empty-needle edge** — `hasAny(k, [])` can't reach `sb.And()`/`sb.In()` with zero args and emit malformed SQL: `[]` is a parse-level 400 (grammar rejects empty array literals), pinned by `test_logs_json_body_empty_needle_array`. `len(params) < 2` also guards the scalar-folding paths, so the value list is never empty.
- **Legacy type agreement** — `legacyElemType` scans *every* value, and both `getBodyJSONArrayKey` (`Array(Nullable(dt))`) and `getBodyJSONScalarKey` (`Nullable(dt)`) plus every coerced needle use that same `dt` → array type and bound args always agree, so no "no supertype" (386). `Nullable` maps off-type values to NULL instead of a corrupt `0`/`false`. The scalar guard correctly excludes `Array`/`Object`/`Null` bodies.
- **Argument validation** — `normalizeFunctionValue`: `has`/`hasToken` reject count≠1 and array args; `hasAny`/`hasAll` accept one array literal *or* fold positional scalars, rejecting a mix. All four branches check out.
- **Refactor equivalence** — `BranchesInOrder()` sorts descending → `["json","dynamic"]`, exactly the old manual JSON-then-Dynamic ordering, so extracting `buildAccessNodeBranches` into `branchArrayExpr` is behavior-preserving.

### Minor / optional (non-blocking)

- `json_condition_builder.go` — `v := v` in the `hasAll` loop is redundant (Go 1.22+ per-iteration vars, and the closure runs synchronously). Harmless.
- The `valueType` passed to `NewJSONConditionBuilder` on the array-function path is never read (`c.valueType` has zero uses in that file) — the `InferDataType` call there is used only for its transformed-needle return. Not wrong, just slightly non-obvious.

### Coverage & gates

- `06_json_body.py`'s removed `array_membership` test wasn't lost — it moved to `09_json_body_functions.py`, which grew to 26 functions covering type-collision, 2^53 precision, no-supertype safety, mixed types, nested paths, type isolation, negation/missing, and arg-shape errors.
- Verified locally on the rebased branch: `go build ./...` ✅, `go test ./pkg/telemetrylogs/...` ✅ (golden SQL matches actual output), `golangci-lint` (primus v2.6.0) ✅ `0 issues`.

Solid, well-commented, well-tested change — good to merge from a correctness standpoint.

therealpandey (approved):

pkg/telemetrylogs/json_condition_builder.go
krsoninikhil: `Or` => `OR`

tushar-signoz: Ahh, sorry mis-read it. Yes valid will change it

PR #12091, opened 2026-07-13
fix: handling missing keys metadata

tushar-signoz:
## Pull Request

---

### 📄 Summary

Traces queries referencing a key that is absent from the field-metadata store (metadata lag, TTL eviction, or a key that was never indexed) currently fail hard with `key not found` — in filters, select fields, group-bys, and aggregations. Existence is a property of the data, not of metadata, so this PR makes traces **synthesize the candidate column(s) and query anyway**, surfacing a warning instead of an error.

**How it works**

- `querybuilder.SynthesizeKeys`: a qualified key (`resource.x`, `x:string`) is honored as-is; a bare key defaults to attribute context with its data type inferred from the filter operand (`x = 'a'` → string, `x > 5` → number), or fanned out across the `attributes_string`/`attributes_number`/`attributes_bool` maps when there is no operand (exists / group by / select).
- The traces field mapper's `CandidateKeys` owns resolution: metadata matches for the name (or `{context}.{name}`) first, else synthesized type-variant keys.
- All four not-found surfaces go through it: WHERE conditions (condition builder, with a `key not found in metadata; querying the underlying data directly` warning), select fields and order-by (`ColumnExpressionFor`), group-by (statement builders), and aggregation arguments (shared `CollisionHandledFinalExpr`).
- Columns are wrapped with existence guards (`multiIf(mapContains(...), col, NULL)`) so absent keys yield `NULL` rather than the map default (`''`/`0`), and select/group-by columns get positional sentinel aliases (`__SELECT_KEY_<i>_<name>` / `__GROUP_BY_KEY_<i>_<name>`) that the querier strips in `consume.go` — response field names are unchanged.

**Supporting refactors**

- `ResolvingConditionBuilder`: traces (and the resource-filter builder) own key resolution end-to-end from the raw key + full metadata map, instead of the where-clause visitor pre-resolving — all traces key resolution now lives in one place.
- `orgID` threaded through `StatementBuilder.Build`, `ConditionBuilder`, `FieldMapper.ColumnExpressionFor`/`CandidateKeys`, and `AggExprRewriter`, so builders can make per-org decisions.
- The traces `ColumnExpressionFor` returns a **bare** (unaliased) expression — callers append the alias; other signals' mappers still return `expr AS name`.
- Other signals (logs, audit, metrics, metadata, resource filter, rule-state history) implement `CandidateKeys` as a no-op — their behavior is unchanged.

#### Issues closed by this PR

Part of the missing-keys-metadata epic (engineering-pod).

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause

The query builder resolves every referenced key against the field-metadata store and errors on a miss. Metadata can legitimately lag or age out while the underlying span data still has the key, so valid queries fail with `key not found`.

#### Fix Strategy

On a metadata miss, synthesize the type-variant keys the user's input implies and build the query against the actual attribute-map columns, attaching a warning so typos still surface (filters get the explicit key-not-found warning; genuinely nonexistent keys simply match no rows).

---

### 🧪 Testing Strategy

- **Tests added/updated**: unit tests for synthesized filters (`TestConditionForSynthesizedKeys`), group-by/aggregation on unseen keys (`TestStatementBuilderGroupByUnseenKey`, `TestStatementBuilderAggregationUnseenKey`), and list-query synthesis; trace-operator expected SQL regenerated for the new rendering. Integration test `queriertraces/07_filter_errors.py`: unknown key → 200 + key-not-found warning; has/hasToken remain 400.
- **Manual verification**: 25-case black-box probe of `POST /api/v5/query_range/preview?verbose=true` (cases derived from the queriertraces integration tests: 8 list, 7 aggregation, 10 trace-operator), against a live ClickHouse, comparing normalized statements per case between this branch and `main`. Every difference is one of the designed classes — sentinel aliases, exists-guard wraps, synthesized fan-out, unknown-select-key `valid:false→true` — with filter WHERE clauses, resource-fingerprint CTEs, and all trace-operator join/CTE shapes unchanged; aggregation-only queries are byte-identical.
- **Edge cases covered**: key collisions (multiple contexts/types), inline-context keys (`resource.service.name`, `span.name:string`), deprecated/intrinsic keys, resource JSON evolution windows, order-by on group-by keys, trace-operator `returnSpansFrom` variants.

---

- **Blast radius**: traces query building (list / time-series / scalar / trace / trace-operator). Other signals only see no-op interface additions.
- **Potential regressions**: queries on genuinely nonexistent keys now return empty results (with a warning on the filter path) instead of a 4xx; group-by/select rendering changed (exists-guard wraps make absent keys bucket as `null` instead of `''`; sentinel aliases rely on `consume.go` stripping, covered by querier tests).
- **Rollback plan**: revert the PR — no schema or data migrations involved.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Trace queries referencing a key missing from metadata run against the underlying data with a warning instead of failing with `key not found` |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- The traces `ColumnExpressionFor` now returns a **bare** (unaliased) expression — callers append the (sentinel) alias — while other signals' mappers still return `expr AS name`. A follow-up could migrate the other signals for consistency.
- `FieldMapper` interface: no new methods — `ColumnExpressionFor` gained `orgID` and `CandidateKeys` was added (no-op for non-traces signals).
- The `__SELECT_KEY_`/`__GROUP_BY_KEY_` prefixes never reach API responses: `stripKeyAlias` in `pkg/querier/consume.go` removes them when reading rows.
- Group-by result semantics: a span without the grouped key now lands in a `null` bucket (exists-guard) instead of the `''` bucket.


tushar-signoz:
## Live verification against real data (unknown-key matrix)

Black-box probe of this PR (head `0c10ed40f`) against a live tenant ClickHouse with **~90k freshly-ingested spans** (~75k in the probe window) across `checkout/inventory/notification/user/payment-service` plus deliberate torture cases: numeric service names, attributes shadowing intrinsic columns (`duration_nano`, `trace_id`, `service.name`, `trace_state` as *attributes*), and 3–4-way context/type collisions on most keys.

Every case hit both `POST /api/v5/query_range/preview?verbose=true` (SQL renders, `valid=true`) and `POST /api/v5/query_range` (executed against the data). `totally.absent.key` / `absent.res.key` / `absent.numeric.key` / `absent.metric` are absent from both metadata **and** data.

**Result: 17/17 cases — preview valid, execution 200, no ClickHouse errors.**

### List queries (raw)

| Case | Query | Result |
|---|---|---|
| select known + unknown | select `service.name`, `name` + unknown attr + unknown resource key | 5 rows; known fields populated, unknown fields present as `null` |
| order by unknown attr | order by unknown attribute desc | 5 rows (NULL order key, no error) |
| filter known AND unknown | `service.name = "checkout-service" AND totally.absent.key = "x"` | **0 rows** — unknown term filters, doesn't vanish |
| filter unknown **resource** key | `resource.absent.res.key = "staging"` | **0 rows** — WHERE contains `resources_string['absent.res.key'] = ?` (before the review fix this returned *all* spans) |
| filter unknown resource, numeric operand | `resource.absent.numeric.key = 3` | **0 rows**, no CH type error — `toFloat64OrNull(...) = ?` coercion in SQL |
| filter known resource (control) | `deployment.environment = "staging"` (metadata-backed, CTE path) | 5 rows — resource-CTE path unaffected |
| filter unknown exists | `totally.absent.key exists` | **0 rows** (string/number/bool fan-out) |
| filter unknown numeric | `absent.metric > 5` | **0 rows** (operand-inferred `attributes_number` synth) |

### Group-by

| Case | Query | Result |
|---|---|---|
| ts by unknown | `count()` by unknown key | 1 series (NULL bucket) carrying full span counts |
| ts by known + unknown, **limit 5** | | ⚠️ 0 series — pre-existing edge, see below |
| — same, no limit | | 766 series (every service × NULL bucket), works |
| scalar p99 by unknown | `p99(duration_nano)` by unknown | `[null, 879000000]` |
| ts by collision key, limit 5 | `count()` by `user.id` (resource/string + attribute/string + attribute/number) | 4 series with real values — multiIf picks the existing variant |
| scalar by unknown, order by it | | `[null, 90094]` |

### Aggregations

| Case | Query | Result |
|---|---|---|
| `count_distinct(totally.absent.key)` | | `0` |
| `countIf(totally.absent.key = "x")` | | `0` |
| `avg(deployment.version)` on ghost-service | attribute number colliding with resource string | `265.277` — real value, collision handled |

### Trace operator

| Case | Query | Result |
|---|---|---|
| `A => B`, returnSpansFrom=A, select + order by unknown | | 10 rows; known fields populated, unknown = `null` |

### Pre-existing edge discovered (not introduced by this PR)

**NULL group-by dimensions are dropped when `limit` is set on a time-series query.** The top-N `__limit_cte` constrains the main query with a tuple `GLOBAL IN`; a NULL dimension (`(service, NULL) IN (...)`) never matches in ClickHouse, so groups containing a NULL dimension are excluded. Repro: `groupBy(service.name, totally.absent.key), limit 5` → 0 series; identical query without limit → 766 series. `main` renders the same `multiIf(..., NULL)` for spans missing a *collision* key, so partial-NULL groups were already dropped under limit — an unknown key just makes the whole result the NULL group. Fixing it means deciding how NULL dimensions compare in the tuple (e.g. `ifNull(x, '')` on both sides), which changes group-bucketing semantics — flagged as a follow-up.

therealpandey (approved):

PR #12093, opened 2026-07-13
feat(prometheus): add the Prometheus query API under a /prometheus prefix

srikanthccv:
#### Description

- Adds `GET|POST /prometheus/api/v1/query_range` and `/prometheus/api/v1/query` (`pkg/prometheus/promapi`), following the Prometheus HTTP API contract: float-unix or RFC3339 times, float-seconds or duration-string durations, the `{status, data, errorType, error, warnings, infos}` envelope with Prometheus' status codes, and the 11,000-point cap.
- The `/prometheus` prefix works as a drop-in Prometheus base URL: Grafana's Prometheus data source, promtool, and the PromQL compliance tester append `/api/v1/*` to a base URL, so they can point at SigNoz unmodified. Same layout as Mimir/Cortex.
- Wired through `signoz.Handlers` (`prometheus.Handler` interface, constructed in `NewHandlers`) like the other domain handlers.
- Range queries serve through the `RangeExecutor` capability when the provider has it, so a clickhousev2-serving deployment transpiles through these endpoints too.
- New `promapiconformance` integration suite: the frozen promqltest corpus replayed against these endpoints with `prometheus::provider: clickhousev2` — the two paths nothing else exercises (v2 as serving provider, and this API surface). Instant cases go through `/query` with a real `time` parameter. The `instant-coarse` corpus variants are skipped — they exist only to encode instant evals as coarse ranges for the v5 API, and their transpiled coarse-step serving is already covered and ledgered by promqlconformance's clickhousev2 leg — so this suite asserts zero divergences with no ledger of its own.
- Purely additive: the existing `GET /api/v1/query_range` and `GET /api/v1/query` handlers are untouched. `openapi.yml` is generated and these mux-registered routes are outside the generator, so their documentation is the upstream Prometheus API contract they follow.

#### Additional Information

Final slice of the clickhouseprometheusv2 stack (#12323, #12324, #12325 — merged). Legacy endpoint removal, if ever, is a separate change after usage drains.


pkg/prometheus/promapi/handler.go
```diff
@@ -0,0 +1,259 @@
+// Package promapi serves the Prometheus HTTP query API over a
```
therealpandey: You can keep this in promethues/ only along with the handler interface as it is just dependent on the prometheus package itself and should remain consistent across different implementations of prometheus such as v1 and v2

pkg/prometheus/promapi/handler.go
```diff
@@ -0,0 +1,259 @@
...
+type handler struct {
+	logger *slog.Logger
+	prom   prometheus.Prometheus
+}
+
+func NewHandler(logger *slog.Logger, prom prometheus.Prometheus) prometheus.Handler {
+	return &handler{logger: logger, prom: prom}
+}
+
+type errorType string
```
therealpandey: Can extract all of this out in a render.go.

Then you can use this to create the openapi spec for it.

therealpandey (changes requested):

therealpandey:
We can override the default openapi with our custom implementation also. Can see how the health endpoints have been registered.

therealpandey (approved):

PR #12094, opened 2026-07-13
fix: set correct openapi response struct for span mapper list

nityanandagohain:
## Pull Request

---

### 📄 Summary
The defination had a incorrect response struct set becuase of which the the response was not matching the struct in openapi spec .

While at it also found that `goupId` was set to `group_id` in the response struct, updated it as well 

#### Issues closed by this PR

Hotfix for  https://github.com/SigNoz/signoz/pull/11779#discussion_r3568074992

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context
Incorrect response struct in api spec.

---

- Blast radius: None
- Potential regressions:
- Rollback plan:

---


pkg/types/spantypes/mapper.go
```diff
@@ -54,7 +54,7 @@ type SpanMapper struct {
 	types.UserAuditable
 
 	ID           valuer.UUID      `json:"id"            required:"true"`
-	GroupID      valuer.UUID      `json:"group_id"      required:"true"`
+	GroupID      valuer.UUID      `json:"groupId"      required:"true"`
```
therealpandey: @nityanandagohain this file is not correctly formatted. Can you run Ctrl + Save in this file?

therealpandey (changes requested):

pkg/types/spantypes/mapper.go
nityanandagohain: thanks, something is wrong with my dev env. Let me fix.

nityanandagohain: fixed the spaces between tags, couldn't make the formatter do it for some reason.

therealpandey (dismissed):

therealpandey (approved):

PR #12100, opened 2026-07-13
feat(ruletypes): publish v2alpha1-only rule schema with required fields

srikanthccv:
### 📄 Summary

Restricts the **published** OpenAPI contract for `/api/v2/rules` to `schemaVersion: v2alpha1` so generated clients (terraform-provider-signoz#127 `signoz_rule`, frontend orval client) stop seeing legacy v1 alert fields, and marks the fields the server actually requires for v2alpha1 as `required`. Two small server-side behavior fixes ride along so the tightened contract is honest.

Part of SigNoz/engineering-pod#5671. Unblocks SigNoz/terraform-provider-signoz#127 removing all v1 alert fields from the generated resource.

### 📜 Contract changes (docs/api/openapi.yml)

`RuletypesPostableRule` / `RuletypesRule`:
- removed v1-only fields: `evalWindow`, `frequency`, `preferredChannels`, `source`, and `version` (always "v5", now defaulted by the server)
- `schemaVersion` is now `enum: [v2alpha1]` and required
- `evaluation` and `notificationSettings` are now required

`RuletypesRuleCondition`:
- removed v1-only fields: `target`, `op`, `matchType`, `targetUnit` (expressed per threshold entry in v2alpha1)
- `thresholds` is now required

Implemented via `jsonschema.Preparer` on `PostableRule`/`RuleCondition` (`pkg/types/ruletypes/jsonschema.go`) — the published schema is shaped at reflection time; **runtime JSON handling is unchanged** and stored v1 rules keep loading and serving exactly as before.

### 🔧 Server-side behavior changes

1. **`version` defaults to `"v5"`** when omitted/empty (`processRuleDefaults`). The field is validation-only — nothing in the evaluation engine reads it — so clients no longer need to send it. Explicit non-v5 values are still rejected.
2. **`notificationSettings.renotify` now round-trips.** It was a value struct with `omitzero`, so an explicitly disabled renotify (`{"enabled": false}`) vanished from every read response (create response included) — surfaced by the Terraform round-trip testing. It is now a pointer with `omitempty`: what you POST is what you GET.

### ⚠️ Sequencing — why this is a draft

- The v2 read models serialize **stored v1-schema rules** without `thresholds`/`evaluation`/`notificationSettings`/`schemaVersion` (schema-version-specific `MarshalJSON`). Until the v1→v2alpha1 storage migration lands, list/get responses containing legacy rules won't satisfy the required marks — strict generated readers (e.g. `terraform import` of a legacy rule) will see missing fields. **Merge after the storage migration.**
- The server still **accepts** v1-schema writes on `/api/v2/rules` (frontend v1 editor + clone flows depend on this). Rejecting v1 writes is a follow-up gated on the migration + v2 anomaly-rule authoring in the UI.
- `PATCH /api/v2/rules/{id}` has JSON-merge (partial) semantics but reuses the full `PostableRule` body schema, so the required marks overstate what a patch body needs. Follow-up: dedicated patch schema.

The frontend keeps sending the legacy fields on the wire for v1 editor flows (boundary casts in `src/types/api/alerts/convert.ts`), so no user-facing behavior changes.

### 🧪 Testing Strategy

- New unit tests: renotify round-trip through the v2 read path (absent stays absent, explicit-disable echoes, enabled config echoes verbatim) and `version` defaulting; updated version-validation cases.
- `go test ./pkg/types/ruletypes/... ./pkg/query-service/rules/... ./pkg/alertmanager/... ./pkg/ruler/...` — green.
- `go run cmd/enterprise/*.go generate openapi` and `pnpm generate:api` re-run to verify the committed spec/clients are in sync.
- Frontend: `tsgo --noEmit` clean, `pnpm build` passes, ListAlertRules (59 tests) and CreateAlertV2 Footer (43 tests) suites green.

### DIFF
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -7551,35 +7551,28 @@ components:
           type: string
         disabled:
           type: boolean
-        evalWindow:
-          type: string
         evaluation:
           $ref: '#/components/schemas/RuletypesEvaluationEnvelope'
-        frequency:
-          type: string
         labels:
           additionalProperties:
             type: string
           type: object
         notificationSettings:
           $ref: '#/components/schemas/RuletypesNotificationSettings'
-        preferredChannels:
-          items:
-            type: string
-          type: array
         ruleType:
           $ref: '#/components/schemas/RuletypesRuleType'
         schemaVersion:
-          type: string
-        source:
-          type: string
-        version:
+          enum:
+          - v2alpha1
           type: string
       required:
       - alert
       - alertType
       - ruleType
       - condition
+      - schemaVersion
+      - evaluation
+      - notificationSettings
       type: object
     RuletypesQueryType:
       enum:
@@ -7629,12 +7622,8 @@ components:
           type: string
         disabled:
           type: boolean
-        evalWindow:
-          type: string
         evaluation:
           $ref: '#/components/schemas/RuletypesEvaluationEnvelope'
-        frequency:
-          type: string
         id:
           type: string
         labels:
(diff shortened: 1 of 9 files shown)


srikanthccv:
Superseded by https://github.com/SigNoz/signoz/pull/12112

PR #11596, opened 2026-06-05
fix(frontend): always show SigNoz version in sidebar header

pandareen:
## Pull Request

---

The SigNoz version badge in the left sidebar header was intermittently missing when users expanded the sidebar. It only rendered when the license type (`licenseTag`) had loaded, so the version often did not appear even after hover or pin.

This PR keeps the version in the header next to the logo and shows it when the sidebar is expanded (hovered, pinned, or dropdown open). When the sidebar is collapsed, only the SigNoz logo is shown. The version is decoupled from `licenseTag`, so it appears as soon as `currentVersion` is available. This is a minimal, UI-only fix that restores discoverability without adding new navigation surfaces.

**Before:** Sidebar expanded but version missing when license tag had not loaded.

**After (collapsed):** Only SigNoz logo visible.

**After (expanded):** License tag and version visible next to logo.

_Add screenshots or a short screen recording showing collapsed vs expanded sidebar states._

Closes #11454

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

The header version was wrapped in `{licenseTag && (...)}`, so it did not render while license info was loading or unavailable. Users who expanded the sidebar still could not see the version in those cases.

- Show the header version when `currentVersion` is available, independent of `licenseTag`
- Keep version hidden when sidebar is collapsed (logo only)
- Show version in `.brand-title-section` when sidebar is expanded (pinned, hovered, or dropdown open)
- Add `.version-container-standalone` styling when only the version badge is shown (no license tag)

---

- Tests added/updated: N/A
- Manual verification:
  - Sidebar collapsed — only SigNoz logo visible, no version badge
  - Sidebar expanded (hover/pin) — version visible next to logo
  - Version visible before `licenseTag` loads on expansion
  - Confirmed no version in bottom nav or Help & Support menu
- Edge cases covered:
  - Version-only display when license tag is absent
  - Collapsed vs expanded sidebar states
  - Changelog update notification dot on expanded sidebar

---

- Blast radius: `SideNav.tsx`, `SideNav.styles.scss` only
- Potential regressions: Header layout/spacing when sidebar expands; version badge truncation with long version strings
- Rollback plan: Revert the two SideNav files

---

| Field | Value |
|------|-------|
| Deployment Type | Cloud / OSS / Enterprise |
| Change Type | Bug Fix |
[...]

PR #12109, opened 2026-07-14
fix(sidenav): expose running version in "Help & Support" dropdown (#11602)

### LINKED ISSUE #11602: Consider moving/adding version number display back to where help information is
Currently the SigNoz version number, [when it works][1], is displayed at the top of the sidebar, above the "Home" menu item:
Image
(It would be displayed to the right of the orange icon; #11596 is supposed to fix this.)

In older versions of Signoz, it was displayed at the bottom of the sidebar, along with help and support information:
Image

That bottom part of the sidebar was a mess in older version, and has correctly been cleaned up to match, more or less, how standard GUI applications have worked for many decades: with a "Help" menu after all the other menus.
Image

However, traditionally that Help menu includes an "About" item that gives version information and so on. I believe that by now all users will instinctively, if asked to look up a version number in a GUI-like application, start with the Help menu. (I have never seen a GUI application that, if it offered the ability to see a version number, put it anywhere other than the Help menu.)

I suggest that, regardless of whether or not we keep the version number display at the top of the sidebar, we add either:
- a) an "About" item under the "Help & Support" menu that, when clicked will in some way display the version number (plus any other information felt appropriate); or
- b) when the "Help & Support" menu is clicked, include a (possibly clickable, possibly not) item directly in that menu that displays the version number, much as the old version did.

(FWIW, the old version gave you the following when you clicked on the version number:
Image)

The primary purpose of this suggestion is to make it as easy as possible for people to find the version number to provide when submitting bug reports, and the like. Users writing a bug report are not particularly motivated to do any work to find the version, so we we should ensure that the traditional "instinctive" actions that let users see the version number do not fail.

[1]: https://github.com/SigNoz/signoz/pull/11596

Taranum01:
## Summary

Fixes #11602.

The current SigNoz version is rendered at the top of the sidenav next to the brand tag — but only when it works (see #11596). When users file a bug report, the instinctive place to look for the version is the "Help" menu: that is where nearly every native GUI application puts the "About" item. Today nothing in the Help & Support dropdown surfaces the version.

This change appends an informational "SigNoz <version>" entry to the bottom of the dropdown so users have a single, predictable click path to copy the version into a bug report. The entry is non-interactive (no click handler, no router navigation) and lives below a divider so the existing docs/GitHub/Slack/Chat support items stay in their original order.

## Changes

- `frontend/src/container/SideNav/menuItems.tsx` — refactor `helpSupportDropdownMenuItems` from a constant into a builder `buildHelpSupportDropdownMenuItems(currentVersion?)`. Returns the union `(SidebarItem | DropdownSeparator)[]` that `SideNav` state already accepts, and keep a stable fallback export of the original seed for tests.
- `frontend/src/container/SideNav/SideNav.tsx` — seed the dropdown from the new builder using the `currentVersion` it already pulls from `state.app`. Add a small effect that keeps the version block in sync as `/version` resolves after mount, using a strip-old + re-append strategy that preserves the existing changelog/license entries added by other effects.
- `frontend/src/container/SideNav/SideNav.styles.scss` — add rules for `.help-support-dropdown-version` (uppercase label + tabular-numerals value), matching the existing `.user-settings-dropdown-logged-in-section` visual language.
- `frontend/src/container/SideNav/__tests__/menuItems.test.tsx` — five new unit tests pinning the new contract.

## Tests

```
$ cd frontend && npx jest src/container/SideNav/__tests__/menuItems.test.tsx
PASS src/container/SideNav/__tests__/menuItems.test.tsx
Test Suites: 1 passed, 1 total
Tests:       10 passed, 10 total
```

All 10 tests pass (5 existing + 5 new). No new lint warnings introduced.

## Risk

Low. Only adds menu entries at the bottom of the dropdown; never removes any existing item. The two small behavior changes (sync read on mount + stale-strip / re-append effect) are encapsulated inside `SideNav` and do not touch the changelog effect that already mutates the same state.

## Out of scope

- No backend, route, or i18n changes — frontend-only display tweak.
- The version-rendering issue tracked in #11596 (where the version at the top of the sidebar sometimes does not appear) is left untouched; that bug belongs to the brand container, not the dropdown.

## UX

The new entry sits at the bottom of the Help & Support dropdown, below a thin divider:

```
  Documentation           ↗
  GitHub                  ↗
  Community Slack         ↗
  Chat with Support
  Invite a Team Member
  ─────────────────────
  SIGNOZ          v0.99.0-rc.1
```

The label is uppercase and secondary, the value is bold and tabular, so the running version reads at a glance but does not compete with the interactive items above it.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/frontend/src/container/SideNav/SideNav.styles.scss
+++ b/frontend/src/container/SideNav/SideNav.styles.scss
@@ -1141,6 +1141,43 @@
 	}
 }
 
+// Informational "SigNoz v0.x.y" entry inside the Help & Support dropdown.
+// Keeps the version visible at the standard "Help" menu like virtually
+// every native GUI application, so users have a one-click path to copy
+// the version when filing bug reports. See issue #11602.
+.help-support-dropdown {
+	.ant-dropdown-menu-item {
+		.help-support-dropdown-version {
+			display: flex;
+			align-items: center;
+			justify-content: space-between;
+			gap: 8px;
+			padding: 4px 0;
+
+			.help-support-dropdown-version-label {
+				color: var(--l3-foreground);
+				font-family: Inter;
+				font-size: var(--font-size-xs);
+				font-style: normal;
+				line-height: normal;
+				letter-spacing: 0.14px;
+				text-transform: uppercase;
+			}
+
+			.help-support-dropdown-version-value {
+				color: var(--l1-foreground);
+				font-family: Inter;
+				font-size: var(--font-size-xs);
+				font-style: normal;
+				font-weight: 600;
+				line-height: normal;
+				letter-spacing: 0.14px;
+				font-variant-numeric: tabular-nums;
+			}
+		}
+	}
+}
+
 .secondary-nav-items {
 	.nav-item {
 		position: relative;
--- a/frontend/src/container/SideNav/SideNav.tsx
+++ b/frontend/src/container/SideNav/SideNav.tsx
@@ -84,7 +84,7 @@ import { buildNavUrl, getQueryString } from './helper';
 import {
 	defaultMoreMenuItems,
 	getUserSettingsDropdownMenuItems,
-	helpSupportDropdownMenuItems as DefaultHelpSupportDropdownMenuItems,
+	buildHelpSupportDropdownMenuItems,
 	helpSupportMenuItem,
 	primaryMenuItems,
 	aiAssistantMenuItem,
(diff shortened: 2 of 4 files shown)


Taranum01:
Friendly bump — this PR has been sitting without review attention. Small sidenav addition, ready for review.

If anything needs to be addressed (additional tests, naming, scope, etc.), happy to iterate.

PR #12110, opened 2026-07-14
fix(ruletypes): address rule read round-trip gaps

srikanthccv:
## Pull Request

Part of https://github.com/SigNoz/engineering-pod/issues/5671

- `notificationSettings.renotify` It was a value struct with `omitzero`, so an explicitly disabled renotify (`{"enabled": false}`) was dropped from every read response, create response included. It is now `*Renotify` with `omitempty`: what you POST is what you GET.
- `version` defaults to `"v5"` when omitted or empty. The field is validation-only (so we wouldn't allow users to override the migration from v4 to v5) nothing in the evaluation engine reads it so clients no longer need to send it. Explicit non-v5 values are still rejected.

### DIFF
--- a/pkg/types/ruletypes/roundtrip_test.go
+++ b/pkg/types/ruletypes/roundtrip_test.go
@@ -1,0 +1,258 @@
+package ruletypes
+
+import (
+	"encoding/json"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func readBody(t *testing.T, stored string) map[string]json.RawMessage {
+	t.Helper()
+	g := GettableRule{}
+	require.NoError(t, json.Unmarshal([]byte(stored), &g))
+	out, err := json.Marshal(NewRule(&g))
+	require.NoError(t, err)
+	var body map[string]json.RawMessage
+	require.NoError(t, json.Unmarshal(out, &body))
+	return body
+}
+
+func TestV2RoundTripRolling(t *testing.T) {
+	thresholds := `{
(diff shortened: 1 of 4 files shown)


pkg/types/ruletypes/roundtrip_test.go
```diff
@@ -0,0 +1,258 @@
...
+		"spec": [
+			{"name": "critical", "target": 90.5, "targetUnit": "%", "recoveryTarget": 80.5, "matchType": "at_least_once", "op": "above", "channels": ["slack-critical"]},
+			{"name": "warning", "target": 75, "targetUnit": "%", "matchType": "at_least_once", "op": "above", "channels": ["slack-warnings", "email-oncall"]}
+		]
+	}`
+	thresholdsEchoed := `{
+		"kind": "basic",
+		"spec": [
+			{"name": "critical", "target": 90.5, "targetUnit": "%", "recoveryTarget": 80.5, "matchType": "at_least_once", "op": "above", "channels": ["slack-critical"]},
+			{"name": "warning", "target": 75, "targetUnit": "%", "recoveryTarget": null, "matchType": "at_least_once", "op": "above", "channels": ["slack-warnings", "email-oncall"]}
```
srikanthccv, line 35: we have these fields `BasicRuleThreshold.{targetUnit,recoveryTarget,channels}` that are not marked as `omitempty`, so they come back as as `""`/`null` on read. but it seems like they are handled as semantically equivalent items so no diff? I could see `recoveryTarget` and `channels` being nullable so that answers it for them.

therealpandey (approved):

PR #12115, opened 2026-07-14
feat(trace-details): download full trace from trace details page

aks07:
## Pull Request

---

### 📄 Summary

This PR adds Download Trace option to the trace details page which will help users Download all the spans of a trace in CSV or JSONL format.

**Implementation**
- Use the existing exportsData api we have in the Explorer pages for this as well.
- We fetch at max 50k rows per API call, so we are making multiple calls for traces with spans greater than 50k (at max 3 API calls in parallel). Getting the `Blob` and stitching it to create a single csv or jsonl.
- On **Download (CSV/JSONL)** click, `runChunkedExport` decides how many workers to spawn, allowing API calls to run in parallel.
- The response from each API call is used as-is (`Blob`). The existing export fetch logic has been extracted into the `fetchExportData` utility.
- The `stitchParts` utility stitches the blobs together into a single `Blob`.

The final output is intentionally kept as a `Blob` because, as long as it remains a `Blob`, it does not consume JavaScript heap memory, which would otherwise become an issue for large exports. Because of this, no Web Worker is needed.

`stitchParts` logic:

- if jsonl
  - add a simple `\n` if missing at the end of each part and join
- if csv
  - use header of the first csv and remove from subsequent api blobs

Finally, trigger the download using the `downloadFile` utility created in the Export Timeseries PR (extended here to accept a `Blob`).

Any page failure or a user cancel aborts all in-flight requests and nothing partial is saved.

https://github.com/user-attachments/assets/eb050f0f-f3c9-4be0-b3ea-ab5e0d783085

https://github.com/SigNoz/engineering-pod/issues/5376
---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | From the trace details page, you can now download all spans of a trace as a single CSV or JSONL file, with download progress and the ability to cancel. Available for traces up to 500k spans. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

---


YounixM:
Experience - We should show explicit cancel and messaging for cases that could stop the download.

X - close the modal 

 We should show info message if the download is time taking. (In this scenario, mostly yes) - download would depend upon network speed, user system.

frontend/src/pages/TraceDetailsV3/TraceDetailsHeader/TraceDownloadPanel.module.scss
```diff
@@ -0,0 +1,42 @@
...
+}
+
+.percent {
+	font-size: var(--paragraph-base-400-font-size);
+	font-weight: var(--font-weight-semibold);
+	color: var(--l2-foreground);
+	flex-shrink: 0;
+}
+
+.cancelBtn {
```
YounixM, line 37: Any specific reason we want to override the height and width of the button ?

YounixM (approved):
Minor query . We can merge this.

frontend/src/pages/TraceDetailsV3/TraceDetailsHeader/TraceDownloadPanel.module.scss
aks07: Will raise a follow up to replace the cross with cancel

PR #12129, opened 2026-07-15
feat(qb): parse search() in the filter grammar (not yet implemented)

tushar-signoz:
## What this does

This adds a new `search('needle')` function to the filter query language — but only teaches the parser to *recognize* it. The actual search isn't implemented yet.

The idea behind `search()` is a full-text search that looks across **every field** of a log or trace. That's different from plain quoted text in a filter today, which only searches the body.

## Why land it in two steps

Rather than ship everything at once, this PR puts the plumbing in place first — the grammar and the code that walks filter expressions — so the real implementation can follow in a smaller, easier-to-review PR.

## What happens if you use it today

If someone writes `search(...)` in a filter right now, they get a clear message — **"search is not yet supported"** — instead of a confusing syntax error. So it fails gracefully.

## The changes

- **Grammar**: added a `SEARCH` token and a `search(...)` rule. It's built to allow scoped versions later (like `search(body, 'abc')`) without having to change the grammar again.
- **Generated parser code**: most of the line count in this PR is auto-generated ANTLR output — not hand-written.
- **The filter visitors**: several places in the code read filter expressions. The query builder rejects `search()` with the friendly "not yet supported" error; alert links, dashboard variable replacement, and the always-false-filter check just pass it through untouched, since there's nothing to act on yet.
- Some small, incidental cleanups in the files that were already being touched (no behavior change).

## One thing to note

`search` is now a **reserved word** in the filter language, just like `has`, `hasAny`, `hasAll`, and `hasToken` already are. So an unquoted, bare `search` used as free text no longer parses the way it used to — quoting it (`'search'`) still works fine.

## Testing

- Updated the existing parser tests so the expected error messages include the new `search()` option.
- Manually confirmed `search('abc')` parses and returns the "not yet supported" error, and that existing filters are unaffected.

## Rollback

Safe to revert — no database, API, or stored-data changes.


grammar/FilterQuery.g4
```diff
@@ -184,6 +197,7 @@ HASTOKEN    : [Hh][Aa][Ss][Tt][Oo][Kk][Ee][Nn];
 HAS         : [Hh][Aa][Ss] ;
 HASANY      : [Hh][Aa][Ss][Aa][Nn][Yy] ;
 HASALL      : [Hh][Aa][Ss][Aa][Ll][Ll] ;
+SEARCH      : [Ss][Ee][Aa][Rr][Cc][Hh] ;
```
tushar-signoz, line 200: **Frontend parser not regenerated — FE/BE grammar divergence for `search`.**

The frontend ships its own generated copy of this grammar (`frontend/src/parser/FilterQueryLexer.ts` etc.), and it has no `SEARCH` token after this change. Consequences until it is regenerated:

- Bare `search` as a free-text term: the query editor will tokenize it as a key/free text and consider the query valid, but the backend now returns a syntax error (`expecting one of {…} but got …`).
- `search('x')`: the frontend parses it as `search` + parenthesized free text, so editor validation/highlighting won't match what the backend actually does.

Worth either regenerating the frontend parser in this PR or noting the deferral explicitly in the description so the FE change is tracked.

tushar-signoz: This we will do, once the feature is working end-to-end

therealpandey (approved):

PR #12130, opened 2026-07-15
fix(infra-monitoring): use proper metric names corresponding to entities to be counted

nikhilmantri0902:
## Pull Request

---

### 📄 Summary
Per-cluster and per-namespace resource counts (nodes, namespaces, deployments, daemonsets, jobs, statefulsets) were queried against the wrong metric universes — clusters unioned the pod + cluster metric lists, namespaces reused the display-table list. Those metrics don't reliably carry each entity's name label, so distinct counts were inaccurate.

This replaces them with curated metric-name lists (`clusterMetricNamesListForCounts`, `namespacesMetricNamesListForCounts`) where each entity is counted from the metric that actually emits its name label (e.g. daemonsets from `k8s.daemonset.*`, deployments from `k8s.deployment.*`), giving accurate counts.

#### Issues closed by this PR
N/A

---

### ✅ Change Type
- [x] 🐛 Bug fix

---

#### Root Cause
Count queries used generic/unioned metric lists that don't carry the target entity's name label, so distinct-value counts were wrong.

#### Fix Strategy
Introduce dedicated per-entity metric-name lists for counts, mapping each counted entity to metrics that emit its name label.

---

### 🧪 Testing Strategy
- Tests added/updated: No
- Manual verification: Verified cluster/namespace counts on staging match actual entity coun
- Edge cases covered: Entities with no active pods still counted via their own metrics

---

- Blast radius: infra-monitoring clusters & namespaces list count fields only
- Potential regressions: Counts shift if a curated metric isn't emitted in an env
- Rollback plan: Revert commit

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Accurate node/namespace/workload counts on infra-monitoring clusters & names

---

### 📋 Checklist
- [x] Manually tested
- [x] Breaking changes considered (none)


pkg/modules/inframonitoring/implinframonitoring/clusters_constants.go
```diff
@@ -51,11 +49,45 @@ var clusterCountAttrKeys = []string{
 }
 
 // clusterMetricNamesListForCounts is the metric universe for per-cluster distinct
-// counts. It unions the pod universe (carries namespace + workload owner labels)
-// with the cluster/node universe (carries k8s.node.name), so a single query can
-// count nodes, namespaces, and workloads per cluster. Overlapping pod
-// phase/status metrics are left in — harmless in a metric_name IN (...) list.
-var clusterMetricNamesListForCounts = slices.Concat(podsTableMetricNamesList, clustersTableMetricNamesList)
+// counts.
+var clusterMetricNamesListForCounts = []string{
```
therealpandey, line 53: Catch this in an integration test please. There needs to be one test whoch asserts the correct behaviour.

therealpandey (changes requested):

pkg/modules/inframonitoring/implinframonitoring/clusters_constants.go
nikhilmantri0902: Hi @therealpandey, I have now updated integration tests to assert the correct behaviour for this scenario of the fix.

therealpandey (approved):

namanv3 (approved):

PR #12132, opened 2026-07-15
chore(alert-history): use v2 apis

H4ad:
## Pull Request

---

This PR drops usage of V1 in favor of V2 APIs.

The APIs is almost the same with a very problematic change that was around the filter for the history of alerts, the previous implementation was using V3 with ClientQBSearch and we had to move to QuerySearch + V5.

Because the keys + values suggested in this route is not baked by logs/traces/metrics directly, instead it uses their own set of APIs, `field/keys and field/values`, I had to add another prop to customize the value fetcher for each key. We already had a way to override the keys, it was just missing a way to override the values.

This PR also includes a fix on backend, because keys like `severity` came from labels, they don't exist in the normal datasource (like any metric/log/trace), and without the change, when filtering via backend it throws `key not found`. The fix is just to treat them like they exist in case we don't find them using `GetKeysMulti`.

Before / After:

https://github.com/user-attachments/assets/ad11c42d-7a11-428a-8fb3-71efe3d03300

On search key invalid:

image

---

About the fix on query search css, I created that because of:

image

After:

https://github.com/user-attachments/assets/787fabe3-3c9a-4302-b5de-56cb74f86d87

Closes https://github.com/SigNoz/engineering-pod/issues/5722

Closes https://github.com/SigNoz/engineering-pod/issues/5802

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: No
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius:  Alert - History
- Potential regressions: Many, it's a refactor to use V2 API.
- Rollback plan: Find and fix the issue instead of reverting the PR.

---

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | The Alert History page is now baked by V2 APIs, this includes the usage of the new Query Builder in the history search. |

---

### 📋 Checklist
- [x] Manually tested


frontend/src/container/AlertHistory/Timeline/Table/Table.tsx
```diff
@@ -91,10 +91,14 @@ function TimelineTableContent(): JSX.Element {
 
 	const handleRunQuery = useCallback(
 		(updatedExpression?: string): void => {
-			querySearchOnRun(updatedExpression ?? inputExpression);
-			refetch();
+			const nextExpression = updatedExpression ?? inputExpression;
+			querySearchOnRun(nextExpression);
+
+			if (nextExpression === expression) {
```
AshwinBhatkal, line 101: Should this be !==?

frontend/src/container/AlertHistory/Timeline/Table/useAlertHistoryFilterSuggestions.ts
```diff
@@ -81,7 +81,7 @@ export function useAlertHistoryFilterSuggestions(
 			{ startUnixMilli: startTime, endUnixMilli: endTime },
 			{
 				query: {
-					enabled: !!ruleId,
+					enabled: !!ruleId && startTime !== null && endTime !== null,
```
AshwinBhatkal: can you control it's value being not null from the hook itself?

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
```diff
@@ -466,28 +472,47 @@ function QuerySearch({
...
+						lastKeyRef.current !== key ||
+						lastValueRef.current !== sanitizedSearchText
+					) {
+						return; // Skip updating if key has changed or component unmounted
+					}
+
+					// Process the response data
+					const responseData = response.data as any;
+					const values = responseData.data?.values || {};
+					stringValues = values.stringValues || [];
```
AshwinBhatkal: why "||" here and ?? above

frontend/src/types/api/alerts/def.ts
```diff
@@ -99,8 +100,6 @@ export interface AlertRuleTimelineTableResponse {
 	labels: Labels;
 	fingerprint: number;
 	value: number;
-	relatedTracesLink: string;
```
AshwinBhatkal, line 102: is this being handled somewhere else?

frontend/src/container/AlertHistory/Timeline/Table/useAlertHistoryFilterSuggestions.ts
```diff
@@ -0,0 +1,141 @@
...
+	) {
+		return QUERY_BUILDER_KEY_TYPES.NUMBER;
+	}
+	if (dt === TelemetrytypesFieldDataTypeDTO.bool) {
+		return QUERY_BUILDER_KEY_TYPES.BOOLEAN;
+	}
+	return QUERY_BUILDER_KEY_TYPES.STRING;
+}
+
+function fieldContextToSuggestionContext(
```
AshwinBhatkal: can these functions be moved to helpers?

frontend/src/container/AlertHistory/Timeline/Table/Table.tsx
```diff
@@ -55,23 +133,99 @@ function TimelineTable(): JSX.Element {
...
+				locale={{
+					emptyText:
+						isError && apiError ? (
+							<div className="timeline-table__error">
+								<ErrorContent error={apiError} />
+							</div>
+						) : undefined,
+				}}
+				footer={(): JSX.Element => (
+					<div className="timeline-table__pagination">
```
AshwinBhatkal, line 181: Can we have this on the right of the table?

frontend/src/container/AlertHistory/Timeline/Table/Table.tsx
```diff
@@ -55,23 +133,99 @@ function TimelineTable(): JSX.Element {
...
+						<div className="timeline-table__pagination-info">
+							{paginationConfig.showTotal?.(totalItems, [
+								((paginationConfig.current ?? 1) - 1) *
```
AshwinBhatkal: In the other file, we have a check for totalItems === 0

frontend/src/container/AlertHistory/Statistics/TopContributorsRenderer/TopContributorsRenderer.tsx
```diff
@@ -13,15 +15,27 @@ function TopContributorsRenderer({
 }: TopContributorsRendererProps): JSX.Element {
 	const { isLoading, isRefetching, isError, data, isValidRuleId, ruleId } =
 		useGetAlertRuleDetailsTopContributors();
-	const response = data?.payload?.data;
 
 	// TODO(shaheer): render the DataStateRenderer inside the TopContributorsCard, it should display the title and view all
```
AshwinBhatkal, line 19: is this something we can take care of now?

frontend/src/container/AlertHistory/Statistics/StatsCardsRenderer/StatsCardsRenderer.tsx
```diff
@@ -25,24 +26,59 @@ type StatsCardsRendererProps = {
 };
 
 // TODO(shaheer): render the DataStateRenderer inside the TotalTriggeredCard/AverageResolutionCard, it should display the title
```
AshwinBhatkal, line 28: remove comment?

frontend/src/container/AlertHistory/Timeline/Table/Table.tsx
H4ad: The `querySearchOnRun` commits the query, but this only re-run if the query is different, when it's the same, it does not re-run, that's why we have this refetch when is equal to force it to refetch

frontend/src/container/AlertHistory/Statistics/StatsCardsRenderer/StatsCardsRenderer.tsx
H4ad: I prefer to keep it since they wanted to improve this UI by moving the error feedback for each card instead of all cards.

We can decide to remove this when rewriting this page or during a cleanup task.

frontend/src/container/AlertHistory/Statistics/TopContributorsRenderer/TopContributorsRenderer.tsx
H4ad: Yes, but don't think we should, I wanted to keep this PR minimal as possible

frontend/src/container/AlertHistory/Timeline/Table/Table.tsx
H4ad: Done

Image

frontend/src/container/AlertHistory/Timeline/Table/useAlertHistoryFilterSuggestions.ts
H4ad: It was just a mistake, start/end cannot be null

Probably something I left here while thinking to rewrite this to globalTimeAdapter (or extract the time from URL with nuqs)

frontend/src/types/api/alerts/def.ts
H4ad: It's in the PR description, I removed the support to see related traces since we don't get this from backend

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
H4ad: Because I kept the exact code as previously (using ||) and new uses ??

Both will have same behavior since || on `[]` will return true since `[]` is truthy value.

frontend/src/container/AlertHistory/Timeline/Table/Table.tsx
H4ad: Yeah, I added in another commit.

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
AshwinBhatkal: This will require a migration from the BE right to be completely removed?

frontend/src/container/AlertHistory/Statistics/StatsCardsRenderer/StatsCardsRenderer.tsx
AshwinBhatkal: This is at the wrong place it felt. The other comment that is similar seemed appropriate

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
H4ad: What do you mean by completely removed?

frontend/src/container/AlertHistory/Statistics/StatsCardsRenderer/StatsCardsRenderer.tsx
H4ad: Maybe, but again, I don't want to change too much code that is not related to migration to v2

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
AshwinBhatkal: To not have the override anymore and it to be just a pure function

AshwinBhatkal (dismissed):

frontend/src/container/AlertHistory/Timeline/Table/Table.tsx
```diff
@@ -55,23 +125,100 @@ function TimelineTable(): JSX.Element {
...
+			{!isLoadingKeys || hardcodedAttributeKeys ? (
+				<div className="timeline-table__filter">
+					<div className="timeline-table__filter-row">
+						<div className="timeline-table__filter-search">
+							<QuerySearch
+								onChange={querySearchOnChange}
+								queryData={queryData}
+								dataSource={DataSource.LOGS}
+								onRun={handleRunQuery}
+								hardcodedAttributeKeys={hardcodedAttributeKeys}
```
srikanthccv: If the request for keys fails (it shouldn't ideally, but assume, for some reason, it fails), then does it show logs keys?

frontend/src/container/AlertHistory/Timeline/Table/Table.tsx
```diff
@@ -1,52 +1,122 @@
...
+			}
+		},
+		[querySearchOnRun, refetch, inputExpression, expression],
+	);
+
+	const queryData = useMemo(
+		() => ({
+			...initialQueryBuilderFormValuesMap.logs,
+			queryName: 'A',
+			dataSource: DataSource.LOGS,
```
srikanthccv, line 112: Why does it have datasource logs here?

frontend/src/container/AlertHistory/Timeline/Table/useAlertHistoryFilterSuggestions.ts
```diff
@@ -0,0 +1,91 @@
...
+		): Promise<{ stringValues?: string[]; numberValues?: number[] } | null> => {
+			if (!ruleId) {
+				return null;
+			}
+			const response = await getRuleHistoryFilterValues(
+				{ id: ruleId },
+				{
+					name: key,
+					searchText,
+					limit: 10,
```
srikanthccv: any reason why this is 10?

frontend/src/container/AlertHistory/Timeline/Table/useAlertHistoryFilterSuggestions.ts
```diff
@@ -0,0 +1,91 @@
...
+					startUnixMilli: startTime,
+					endUnixMilli: endTime,
+				},
+			);
+			const values = response.data?.values;
```
srikanthccv, line 89: Good to have: The response contains a `complete` flag to indicate whether the values list is complete or are there more values to fetch to avoid the unnecessary round trips. I see that is not implemented here (it is implemented for QB). It would be nice to have it supported.

frontend/src/container/AlertHistory/Timeline/Table/useTimelineTable.tsx
```diff
@@ -115,19 +43,4 @@ export const timelineTableColumns = ({
...
-		align: 'right',
-		render: (record): JSX.Element => (
-			<ConditionalAlertPopover
-				relatedTracesLink={record.relatedTracesLink}
-				relatedLogsLink={record.relatedLogsLink}
-			>
-				<Button type="text" ghost>
-					<Ellipsis className="dropdown-icon" size="md" />
-				</Button>
-			</ConditionalAlertPopover>
```
srikanthccv, line 130: Can you help me understand why the actions is removed? It is legitimately useful because it takes user from history page to the explorer with the relevant context.

frontend/src/pages/AlertDetails/hooks.tsx
```diff
@@ -357,38 +382,52 @@ export const useTimelineTable = ({
...
+			return;
+		}
+		void setPage(page - 1);
+	}, [page, setPage]);
 
 	const paginationConfig: TablePaginationConfig = {
-		pageSize: pageSizeInt,
-		showTotal: PaginationInfoText,
-		current: offsetInt / TIMELINE_TABLE_PAGE_SIZE + 1,
+		pageSize: TIMELINE_TABLE_PAGE_SIZE,
```
srikanthccv, line 407: existing bug: Pagination isn’t reset when the time range changes. We should fix it.

frontend/src/container/AlertHistory/Timeline/Table/useTimelineTable.tsx
H4ad: We don't receive the links in the response like in the V1, I decided to just remove the support for now until we decide on what we do with those links (construct on UI or move to backend).

frontend/src/container/AlertHistory/Timeline/Table/useAlertHistoryFilterSuggestions.ts
H4ad: Arbitrary value, the original fields/value has no limit

frontend/src/container/AlertHistory/Timeline/Table/Table.tsx
H4ad: Can be any dataSource, this is being used only to generate an expression that we use for the filter

These dataSource usually help fetching different key/values but since we are overriding all of them, this has no effect.

frontend/src/container/AlertHistory/Timeline/Table/Table.tsx
H4ad: Hm... we can show a hardcoded attribute list to prevent the log keys to be shown

frontend/src/container/AlertHistory/Timeline/Table/useAlertHistoryFilterSuggestions.ts
H4ad: The query builder don't use complete flag, but I added anyways

frontend/src/pages/AlertDetails/hooks.tsx
H4ad: Good catch

frontend/src/components/QueryBuilderV2/QueryV2/QuerySearch/QuerySearch.tsx
H4ad: Cleaned this code path a little bit

frontend/src/container/AlertHistory/Timeline/Table/useTimelineTable.tsx
srikanthccv: please check the logs/traces based alerts because the links should be present in the response for them. Metric based alerts currently don't have but they will be eventually added. In summary, they are present in the response, and are constructed on the backend.

frontend/src/container/AlertHistory/Timeline/Table/useAlertHistoryFilterSuggestions.ts
srikanthccv: When it is left unset, backend sets to max of 50 and sends `complete` flag to let the client know if there are more values or does the list client is complete list. Given the backend has, we can remove this limit.

frontend/src/container/AlertHistory/Timeline/Table/Table.tsx
srikanthccv: Ok, it was strange to see logs datasource so that's why asked, although I wonder if this is going to bite us in the future.

frontend/src/container/AlertHistory/Timeline/Table/Table.tsx
srikanthccv: Looks like with the last commit changes, when the request is successful, the success response is not shown but only the fixed tow keys are shown. Instead of fixed keys, you can show error and ask user to reload to fetch again or some graceful action.

frontend/src/container/AlertHistory/Timeline/Table/useAlertHistoryFilterSuggestions.ts
H4ad: Changed to unset.

frontend/src/container/AlertHistory/Timeline/Table/useTimelineTable.tsx
H4ad: The response does not return it, pushed the changes in the UI to added it back, waiting the changes on backend.

frontend/src/container/AlertHistory/Timeline/Table/Table.tsx
srikanthccv: This is still valid and broken at the moment.

frontend/src/container/AlertHistory/Timeline/Table/useTimelineTable.tsx
srikanthccv: They are present now, please check.

frontend/src/container/AlertHistory/Timeline/Table/Table.tsx
H4ad: Will keep as empty instead of hardcoded attributes on frontend because this is the way we are doing for other pages (it always render empty keys on failure)

AshwinBhatkal (approved):

srikanthccv (approved):

PR #12134, opened 2026-07-16
fix: convert key not found to warnings for logs

tushar-signoz:
### 📄 Summary

A key that isn't in field metadata is no longer a hard error. The query runs, falls back to the actual data (attribute columns + the body JSON path), and returns a `key not found in metadata` warning so typos still surface. Applies to logs and traces.

### 🔍 What changes

- **Filters** on an unknown key match against synthesized attribute/body columns and return a warning, instead of a `400`. Resource-qualified unknown keys still filter to nothing — they aren't silently dropped.
- **Group by / select / order by** on an unknown key synthesize a null column / null bucket instead of failing.
- Absent map attributes now surface as `NULL` in selected columns (matching how resource attributes already behave) rather than `''`.
- Key resolution — metadata lookup, collision handling, synthesis — is unified into a single condition-builder path shared across signals.

### ✅ Change Type
- [x] ✨ Feature
- [x] ♻️ Refactor

### 🧪 Testing
- Unit suites across logs / traces / audit / querybuilder.
- Integration coverage for unknown keys and key collisions in both body modes (legacy string body + `use_json_body`), plus metrics/scalar key-resolution.
- Black-box replay of real SaaS query shapes: with metadata seeded the generated SQL matches prod; with metadata purged, no query loses a user filter condition.

### ⚠️ Notable behavior change
Selecting or ordering by an absent map attribute now yields `NULL` instead of `''` (and `ORDER BY` sorts absent rows as NULL). Exists guards render their constant as a literal (`mapContains(...)`), so those terms carry no bind arg.

### 📝 Changelog
| Field | Value |
|------|-------|
| Deployment Type | Cloud / OSS |
| Change Type | Feature |
| Description | Queries referencing keys missing from metadata now run with a warning instead of failing; unknown keys fall back to attributes and the body JSON path. |


pkg/types/querybuildertypes/querybuildertypesv5/qb.go
```diff
@@ -18,38 +18,27 @@ var (
...
-// ConditionBuilder builds the conditions for the filter.
+// ConditionBuilder builds the conditions for a filter term. The builder owns key resolution:
+// the visitor hands it the raw key + full metadata map + options, not a pre-matched key list.
 type ConditionBuilder interface {
-	// ConditionFor returns the conditions and any advisory warnings for a filter
-	// term. key is the field key as parsed from the query text; fieldKeysForName is
-	// the set of known field keys matching it (may be empty). The builder owns the
-	// decision of what to do — resolve ambiguity, fall back to a body JSON search,
-	// emit a "not found" error, or skip — and which errors/warnings are apt.
-	ConditionFor(ctx context.Context, orgID valuer.UUID, startNs uint64, endNs uint64, key *telemetrytypes.TelemetryFieldKey, fieldKeysForName []*telemetrytypes.TelemetryFieldKey, operator FilterOperator, value any, sb *sqlbuilder.SelectBuilder) ([]string, []string, error)
```
therealpandey, line 46: We can keep the same ConditionFor function name instead of ConditionForKeys.

Also lets remove options from this interface.

The config needs to go as part of NewConditionBuilder wherever required. Injected instead of passed.

tests/integration/tests/queriertraces/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+# varies the key names.
+CORRUPT_RESOURCES: dict[str, Any] = {
+    "timestamp": "corrupt_data",
+    "duration_nano": "corrupt_data",
+    "http_method": "corrupt_data",
+}
+
+
+@pytest.fixture(name="trace_noise", params=["clean", "corrupt"])
+def trace_noise(request: pytest.FixtureRequest) -> tuple[dict[str, Any], dict[str, Any]]:
```
therealpandey: Move this to fixtures please. Why is it in conftest?

pkg/telemetrytraces/field_mapper.go
```diff
@@ -411,41 +412,112 @@ func (m *defaultFieldMapper) ColumnExpressionFor(
...
+// metadata match for the stripped name when present; strict attribute/resource contexts
+// are honored as-is. Falls back to synthesized type-variant keys.
+func (m *defaultFieldMapper) CandidateKeys(ctx context.Context, _ valuer.UUID, field *telemetrytypes.TelemetryFieldKey, value any, keys map[string][]*telemetrytypes.TelemetryFieldKey) []*telemetrytypes.TelemetryFieldKey {
+	// A real column is considered before metadata for bare and forgiving contexts, so a
+	// same-named corrupt attribute can't shadow the intrinsic/calculated column.
+	switch field.FieldContext {
+	case telemetrytypes.FieldContextUnspecified:
+		// bare key: the column comes first, alongside same-named metadata keys under other
+		// contexts whose type is consistent with the column (corrupt entries dropped).
+		probe := *field
```
therealpandey: If we are deferencing it, we need to have a check for nil somewhere

therealpandey (changes requested):

pkg/types/querybuildertypes/querybuildertypesv5/qb.go
tushar-signoz: Yes, makes sense we can retain the name.
The options is not a setup time config, its execution time parameter (changes with v5 requests).
Currently only `skipResource` bool flag is part of the options. The name options might be misleading

tests/integration/tests/queriertraces/conftest.py
tushar-signoz: So this fixture only create a certain type of data corruption which doesn't necessarily be usable for everybody, to avoid noice in the actual fixtures chose to put in in the local conftest file.

therealpandey (approved):

PR #12137, opened 2026-07-16
feat(dashboards-v2): pre-populate alert condition from panel Reduce To + thresholds (#5291)

### LINKED ISSUE #5291: Pre-populate the alert condition based on the selected `Reduce To` and `Thresholds`

ahrefabhi:
## Pull Request

---

Creating an alert from a V2 dashboard panel used to land on `/alerts/new` with a generic default condition (`A above <threshold> at least once in the last 5 minutes`), discarding the intent the panel already encodes via **Reduce To** and **Thresholds**. Users had to re-enter what the panel already knew.

This PR pre-populates the alert condition from the panel (issue #5291), so the most-likely alert is already filled in.

The mapping is driven **producer → URL → consumer**: the dashboard derives an "alert seed" and encodes it into the `/alerts/new` URL; the CreateAlertV2 page applies whatever the URL declares. To make that seam robust, the consumer was refactored to be **origin-agnostic** — each producer fully declares its own prefill, so adding a producer needs no consumer change. The existing ingestion/metering producer was migrated onto the same declarative contract (it now states `matchType=in_total` + `evaluationWindowPreset=meter` explicitly, instead of the consumer inferring "metering" from the absence of a source).

Mapping rules (per the issue):
- **Reduce To → occurrence type** (Value panel only, where Reduce To is user-selected): `sum → in total`, `avg → on average`. Max/Min/Latest have no equivalent → occurrence stays at default.
- **Thresholds → operator + value + unit**: the highest-danger threshold wins (Red > Orange > Green > Blue).
- **Formula panels**: apply the reduce only when every constituent query agrees; otherwise leave the occurrence at default.
- The default baseline is preserved whenever a rule doesn't override it.

_To add:_ recording of **Create Alert** from a Value panel (Reduce To = Sum, threshold `> 90` Red) → alert opens pre-filled with occurrence **In total**, operator **Is above**, target **90**. Before this change the same click opened the generic default (`above / at least once / 0`).

Closes #5291

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

N/A — not a bug fix.

---

- **Tests added/updated:**
  - `deriveAlertPrefill.test.ts` (new) — reduce→occurrence (incl. Max/Min/Latest → no change), formula uniform vs mixed, color danger ranking incl. custom colors, operator mapping incl. `*_or_equal`, Value-panel scope gate, empty-string aggregation fallback, threshold shape.
  - `resolveUrlAlertPrefill.test.ts` (new) — URL → declarative plan: threshold parse/malformed/non-array, matchType/operator normalization, meter preset recognition.
  - `CreateAlertProvider.prefill.test.tsx` (new) — dashboard-style prefill applied without the meter window; occurrence-only prefill; meter preset applies the cumulative window.
  - Updated `buildCreateAlertUrl.test.ts` (asserts params, and reads `thresholds` the same way the consumer does — guards against double-encoding) and `useCreateAlertFromPanel.test.ts` (prefill threaded through both the sync and `/substitute_vars` paths).
- **Manual verification:** Create Alert from Value / TimeSeries / formula panels; the dashboard-variable `/substitute_vars` path; and the ingestion "create alert" flow (regression — still in total + cumulative window).
- **Edge cases covered:** panel with neither reduce nor thresholds; empty-string aggregation `reduceTo` shadowing the legacy field; unknown threshold colors; formula with mixed reduces; label-variant thresholds with no operator.
- `pnpm tsgo --noEmit`, `pnpm lint:js`, and `pnpm build` all pass.

---

- **Blast radius:** the `/alerts/new` prefill path. Two producers touch it — the V2 dashboard panel (new) and ingestion settings (migrated to the declarative contract). Alerts created any other way are unaffected.
- **Potential regressions:** the ingestion "create alert" flow, since its producer now declares `matchType`/`evaluationWindowPreset` explicitly instead of relying on source-absence. Behavior is intended to be identical and is covered by the provider test; worth a manual sanity check.
- **Rollback plan:** revert the PR — the change is purely additive URL params + a self-contained consumer; no schema, API, or persisted-state changes.

---

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Creating an alert from a dashboard panel now pre-fills the alert condition from the panel's Reduce To and Thresholds. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- Structured as three commits for easier review: (1) extract the condition normalizers to a leaf module (breaks an `index↔utils` import cycle), (2) make the CreateAlertV2 consumer apply a **declarative** URL prefill + migrate the ingestion producer onto it, (3) the dashboard panel producer (the #5291 feature).
- The consumer no longer branches on request source — it applies whatever the URL declares (`resolveUrlAlertPrefill`). New producers need zero consumer change.
- Scope decisions (confirmed against the issue): CreateAlertV2 only (not the classic `FormAlertRules`), a single highest-danger threshold, `sum`/`avg` occurrence mapping only, and reduce→occurrence restricted to the Value panel (where Reduce To is user-facing; elsewhere it's a vestigial default).
- No changes to V1 dashboard code.

### DIFF
--- a/frontend/src/container/CreateAlertV2/context/conditionNormalizers.ts
+++ b/frontend/src/container/CreateAlertV2/context/conditionNormalizers.ts
@@ -1,0 +1,63 @@
+import { AlertThresholdMatchType, AlertThresholdOperator } from './types';
+
+// Mirrors the backend's CompareOperator.Normalize() in
+// pkg/types/ruletypes/compare.go. Maps any accepted alias to the enum value
+// the dropdown understands. Returns undefined for aliases the UI does not
+// expose (e.g. above_or_equal, below_or_equal) so callers can keep the raw
+// value on screen instead of silently rewriting it.
+export function normalizeOperator(
+	raw: string | undefined,
+): AlertThresholdOperator | undefined {
+	switch (raw) {
+		case '1':
+		case 'above':
+		case '>':
+			return AlertThresholdOperator.IS_ABOVE;
+		case '2':
+		case 'below':
+		case '<':
+			return AlertThresholdOperator.IS_BELOW;
+		case '3':
+		case 'equal':
+		case 'eq':
+		case '=':
+			return AlertThresholdOperator.IS_EQUAL_TO;
+		case '4':
+		case 'not_equal':
+		case 'not_eq':
+		case '!=':
+			return AlertThresholdOperator.IS_NOT_EQUAL_TO;
+		case '7':
+		case 'outside_bounds':
+			return AlertThresholdOperator.ABOVE_BELOW;
+		default:
+			return undefined;
+	}
+}
+
+// Mirrors the backend's MatchType.Normalize() in pkg/types/ruletypes/match.go.
+export function normalizeMatchType(
+	raw: string | undefined,
+): AlertThresholdMatchType | undefined {
+	switch (raw) {
+		case '1':
+		case 'at_least_once':
+			return AlertThresholdMatchType.AT_LEAST_ONCE;
+		case '2':
+		case 'all_the_times':
+			return AlertThresholdMatchType.ALL_THE_TIME;
+		case '3':
+		case 'on_average':
+		case 'avg':
+			return AlertThresholdMatchType.ON_AVERAGE;
+		case '4':
+		case 'in_total':
+		case 'sum':
+			return AlertThresholdMatchType.IN_TOTAL;
+		case '5':
+		case 'last':
+			return AlertThresholdMatchType.LAST;
+		default:
+			return undefined;
+	}
+}
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/utils/deriveAlertPrefill.ts
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/utils/deriveAlertPrefill.ts
@@ -1,0 +1,158 @@
+/** Derives an alert-condition seed from a panel's Reduce To + Thresholds (issue #5291). */
+
+import type {
+	DashboardtypesComparisonOperatorDTO,
+	DashboardtypesPanelDTO,
+	DashboardtypesPanelPluginDTO,
+} from 'api/generated/services/sigNoz.schemas';
+import {
+	AlertThresholdMatchType,
+	AlertThresholdOperator,
+	Threshold,
+} from 'container/CreateAlertV2/context/types';
+import type { MetricAggregation } from 'types/api/v5/queryRange';
+import type { Query } from 'types/api/queryBuilder/queryBuilderData';
+import { ReduceOperators } from 'types/common/queryBuilder';
+import { v4 as uuid } from 'uuid';
+
+export interface PanelAlertPrefill {
+	matchType?: AlertThresholdMatchType;
+	operator?: AlertThresholdOperator;
+	threshold?: Threshold;
+}
+
+// Most-dangerous first, matching the panel editor palette; unknown colors sort last.
+const THRESHOLD_COLOR_DANGER_ORDER = [
+	'#f1575f',
+	'#f5b225',
+	'#2bb673',
+	'#4e74f8',
+];
+
+interface NormalizedPanelThreshold {
+	color: string;
+	value: number;
+	unit?: string;
+	operator?: DashboardtypesComparisonOperatorDTO;
+}
+
+export function reduceToMatchType(
+	reduceTo: ReduceOperators | undefined,
+): AlertThresholdMatchType | undefined {
+	switch (reduceTo) {
+		case ReduceOperators.SUM:
+			return AlertThresholdMatchType.IN_TOTAL;
+		case ReduceOperators.AVG:
+			return AlertThresholdMatchType.ON_AVERAGE;
+		default:
+			return undefined;
+	}
+}
+
+function readReduceTo(
+	queryData: Query['builder']['queryData'][number],
+): ReduceOperators | undefined {
+	const aggregationReduceTo = (
+		queryData.aggregations?.[0] as MetricAggregation | undefined
+	)?.reduceTo;
+	// `||` not `??`: the aggregation often holds an empty-string placeholder.
+	return aggregationReduceTo || queryData.reduceTo || undefined;
+}
+
+export function uniformReduceTo(query: Query): ReduceOperators | undefined {
+	const { queryData } = query.builder;
+	const first = queryData[0] && readReduceTo(queryData[0]);
+	if (!first) {
+		return undefined;
+	}
+	return queryData.every((data) => readReduceTo(data) === first)
+		? first
+		: undefined;
+}
+
+function readPanelThresholds(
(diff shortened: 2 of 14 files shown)


frontend/src/container/CreateAlertV2/context/conditionNormalizers.ts
```diff
@@ -0,0 +1,63 @@
+import { AlertThresholdMatchType, AlertThresholdOperator } from './types';
+
+// Mirrors the backend's CompareOperator.Normalize() in
+// pkg/types/ruletypes/compare.go. Maps any accepted alias to the enum value
+// the dropdown understands. Returns undefined for aliases the UI does not
+// expose (e.g. above_or_equal, below_or_equal) so callers can keep the raw
+// value on screen instead of silently rewriting it.
+export function normalizeOperator(
```
AshwinBhatkal, line 8: there is already a function with the same name
frontend/src/container/CreateAlertV2/utils.tsx

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/utils/deriveAlertPrefill.ts
```diff
@@ -0,0 +1,158 @@
...
+
+function pickHighestDanger(
+	thresholds: NormalizedPanelThreshold[],
+): NormalizedPanelThreshold | undefined {
+	return [...thresholds].sort(
+		(a, b) => colorRank(a.color) - colorRank(b.color),
+	)[0];
+}
+
+function panelOperatorToAlertOperator(
```
AshwinBhatkal, line 109: same normalizer can be used?

AshwinBhatkal (approved):

frontend/src/container/CreateAlertV2/context/conditionNormalizers.ts
AshwinBhatkal: I see this was moved

PR #12139, opened 2026-07-16
fix(query-search): do not override token classes

H4ad:
## Pull Request

---

This removes the override for token expression causing the syntax highlight to break, introduced at https://github.com/SigNoz/signoz/pull/11992

Before:

image

image

After:

image

image

Related to https://github.com/SigNoz/platform-pod/issues/2739

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

We override a specific token that should not override by mistake.

We should not override that specific CSS class, and we didn't notice in dev because it was not taking precedence over the default css.

After some change in the chunk order, it started getting precedence, causing this issue.

Just remove the custom css.

---

- Tests added/updated: No
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Query Builder
- Potential regressions: Another visual bug.
- Rollback plan: Find and fix the issue.

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | This PR fixes a little bug introduced in the Query Search causing the values to not be correctly syntax highlighted on dark mode. |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/tests/e2e/helpers/query-builder.ts
+++ b/tests/e2e/helpers/query-builder.ts
@@ -1,5 +1,4 @@
 import { Page } from '@playwright/test';
-import { expect, test } from '../../fixtures/auth';
 
 const EDITOR_CONTENT = '.query-where-clause-editor .cm-content';
 const SPANS = '.query-where-clause-editor .cm-line span';
--- a/tests/e2e/tests/query-builder/token-highlight.spec.ts
+++ b/tests/e2e/tests/query-builder/token-highlight.spec.ts
@@ -1,11 +1,6 @@
 import { expect, test } from '../../fixtures/auth';
-import {
-	setupTheme,
-	tokenColor,
-	typeExpression,
-} from '@tests/query-builder/utils';
+import { setupTheme, tokenColor, typeExpression } from '@helpers/query-builder';
 
-const SPANS = '.query-where-clause-editor .cm-line span';
 const LOGS_EXPLORER = '/logs/logs-explorer';
 
 type ThemeConfig = {
@@ -57,25 +52,15 @@ for (const { name, theme, colors } of THEMES) {
 			expect(await tokenColor(page, 'name')).toBe(colors.property);
 		});
 
-		test(`['value', 'value'] — string tokens have native theme color`, async ({
+		test(`['value', 'value2'] — string tokens have native theme color`, async ({
 			authedPage: page,
 		}) => {
 			await setupTheme(page, theme);
 			await page.goto(LOGS_EXPLORER);
-			await typeExpression(page, "['value', 'value']");
+			await typeExpression(page, "['value', 'value2']");
 
-			const allStringColors: string[] = await page.evaluate(
-				({ selector, text }) =>
-					Array.from(document.querySelectorAll<HTMLSpanElement>(selector))
-						.filter((el) => el.textContent === text)
-						.map((el) => window.getComputedStyle(el).color),
-				{ selector: SPANS, text: "'value'" },
-			);
-
-			expect(allStringColors.length).toBe(2);
-			for (const color of allStringColors) {
-				expect(color).toBe(colors.string);
-			}
+			expect(await tokenColor(page, "'value'")).toBe(colors.string);
+			expect(await tokenColor(page, "'value2'")).toBe(colors.string);
 		});
 	});
 }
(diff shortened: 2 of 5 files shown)


ahrefabhi (dismissed):

tests/e2e/tests/query-builder/token-highlight.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+
+		expect(allStringColors.length).toBe(2);
+		for (const color of allStringColors) {
+			expect(color).toBe(DARK.string);
+		}
+	});
+});
+
+test.describe('QueryBuilder token highlighting — light mode (githubLight theme)', () => {
+	test.use({
```
AshwinBhatkal: do we need this if initScript is handling localStorage state?

tests/e2e/tests/query-builder/token-highlight.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+		},
+		{ selector: SPANS, target: text },
+	);
+}
+
+async function typeExpression(page: Page, expr: string): Promise<void> {
+	const editor = page.locator(EDITOR_CONTENT).first();
+	await editor.waitFor({ state: 'visible', timeout: 15_000 });
+	await editor.click();
+	await page.keyboard.press('Control+a');
```
AshwinBhatkal: ControlOrMeta

tests/e2e/tests/query-builder/token-highlight.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+			const el = Array.from(
+				document.querySelectorAll<HTMLSpanElement>(selector),
+			).find((s) => s.textContent === target);
+			return el ? window.getComputedStyle(el).color : null;
+		},
+		{ selector: SPANS, target: text },
+	);
+}
+
+async function typeExpression(page: Page, expr: string): Promise<void> {
```
AshwinBhatkal: move this into helpers for query builder

tests/e2e/tests/query-builder/token-highlight.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+
+	test('TC-03 k8s.namespace.name — each segment gets its own color', async ({
+		authedPage: page,
+	}) => {
+		await page.addInitScript(() => localStorage.setItem('THEME', 'light'));
+		await page.goto(LOGS_EXPLORER);
+		await typeExpression(page, 'k8s.namespace.name');
+
+		// In githubLight, the dot (operator) hashes to the same class as identifier.
+		expect(await tokenColor(page, 'k8s')).toBe(LIGHT.identifier);
```
AshwinBhatkal: We should have a loop with the tests for light and dark. It will reduce the duplicates

const THEMES = [
	{
		name: 'dark',
		force: null,
		segments: [ {token: '', color: ''}],
		stringColor: 'x',
	},
	{
... same for light
}
] as const;

tests/e2e/tests/query-builder/token-highlight.spec.ts
H4ad: Actually, we don't

tests/e2e/helpers/query-builder.ts
```diff
@@ -0,0 +1,37 @@
+import { Page } from '@playwright/test';
```
AshwinBhatkal, line 1: Should be in helpers folder

tests/e2e/tests/query-builder/token-highlight.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+			await page.goto(LOGS_EXPLORER);
+			await typeExpression(page, 'k8s.namespace.name');
+
+			expect(await tokenColor(page, 'k8s')).toBe(colors.identifier);
+			expect(await tokenColor(page, '.')).toBe(colors.operator);
+			expect(await tokenColor(page, 'namespace')).toBe(colors.property);
+			expect(await tokenColor(page, 'name')).toBe(colors.property);
+		});
+
+		test(`['value', 'value'] — string tokens have native theme color`, async ({
```
AshwinBhatkal: can this be 2 different values? 🤔

AshwinBhatkal: Just for the test

AshwinBhatkal (approved):

AshwinBhatkal (approved):

PR #12140, opened 2026-07-16
feat[ai-011y]: fields API for ai query builder

nityanandagohain:
## Pull Request

---

### 📄 Summary
  - Add a `type` param to `/api/v1/fields/keys`; for type=builder_ai_query (flag-gated) the metadata store returns the per-trace aggregate columns (llm_call_count, input_tokens, …) as
  trace-context keys — they're computed at query time, never ingested, so the attribute scan can't serve them.
  - Split `TraceColumn.Orderable` into `Orderable + Filterable`: ORDER BY uses orderable, the trace-level filter validates against filterable, and the API only returns keys that are both. `last_activity_time` is order-only and now rejected in filters with a targeted error.**
  - UI note: last_activity_time  should be added to client-side list (it's the default sort).

#### Issues closed by this PR
Part of https://github.com/SigNoz/engineering-pod/issues/5714

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: ✅ 
- Manual verification: ✅ 
- Edge cases covered: ✅ 

---

- Blast radius: None
- Potential regressions:
- Rollback plan:


tests/integration/tests/querierai/03_ai_metadata.py
```diff
@@ -0,0 +1,79 @@
+"""Fields metadata API with type="builder_ai_query": gen_ai attributes and per-trace
```
tushar-signoz: Do these tests conform the rules that have been recently added under `.claude/`

pkg/types/telemetrytypes/field.go
```diff
@@ -238,6 +238,7 @@ type FieldKeySelector struct {
 	EndUnixMilli      int64                  `json:"endUnixMilli"`
 	Signal            Signal                 `json:"signal"`
 	Source            Source                 `json:"source"`
+	QueryType         string                 `json:"queryType"`
```
tushar-signoz: If it is QueryType then lets use the enum, if we don't want to couple it be QueryType then lets keep this as type as well.

pkg/telemetrymetadata/metadata.go
```diff
@@ -159,6 +161,13 @@ func (t *telemetryMetaStore) getTracesKeys(ctx context.Context, fieldKeySelector
 		instrumentationtypes.CodeFunctionName: "getTracesKeys",
 	})
 
+	// The trace field context never matches ingested keys — it names the computed
+	// per-trace aggregates, which enrichWithAITraceAggregateKeys serves. Without this
+	// the tagType condition below has no branch for it and the scan returns every key.
+	fieldKeySelectors = slices.DeleteFunc(slices.Clone(fieldKeySelectors), func(s *telemetrytypes.FieldKeySelector) bool {
+		return s.FieldContext == telemetrytypes.FieldContextTrace
+	})
```
tushar-signoz: Not needed, let it return everything. Also Frontend shouldn't be sending fieldContext at all in the request.

tests/integration/tests/querierai/03_ai_metadata.py
nityanandagohain: have updated

srikanthccv (commented):
This doesn't look good to me. There are many problems with this change and the import cycle hack should have been the trigger to re-think about the change.

- IIUC, `telemetrytypes.QueryType` exists just because `querybuildertypesv5.QueryType` can't be imported so we are re-creating it here.

- The param means nothing for the `/fields/values`, `PostableFieldValueParams` embeds the keys params, so `type` now shows up on `/fields/values`. These are computed columns with no stored values to autocomplete. It's hard to reason about this for anyone who just looks at the params as a way to query.
- `telemetrymetadata` now imports `statementbuilder/aistatementbuilder`, which is another smell that shouldn't appear, bringing the product specific statement builder into a generic metadata service (the current objective).

More importantly, the rest of the query params describe the fields but the new addition, in the best interpretation describes the person requesting it.

The current selectors, `signal`, `fieldContext`, `fieldDataType`, `metricName` is a property of the field domain but the new type is about the caller than it is about the fields. 

Here is what I suggest instead,

## Option 1
`FieldContextTrace` has always been a first class field context since the introduction of the field context as thing in the system.

Un-comment `FieldContextTrace` from the request enum and treat the trace context virtual keys to the true meaning, i.e fields in the trace context of the traces signal.

## Option 2

don't put it in fields/keys at all. The list is fixed, and hand defined, and identical unless regardless of the org. The UI client needs the scope contract anyway (the order able and filterable info). So a separate end point serving and being the source of truth for the info needed to clients, and leave `/fields/keys` to actual telemetry fields.


if we find ourselves seeing another second scope sort of thing, then it's legitimately time to introduce another selector like `scope`/`domain` or something along the lines than to introduce the caller specific behaviour into API.

pkg/modules/fields/implfields/handler.go
```diff
@@ -51,6 +55,86 @@ func (handler *handler) GetFieldsKeys(rw http.ResponseWriter, req *http.Request)
 	})
 }
 
+func (handler *handler) GetAIObservabilityFieldsKeys(rw http.ResponseWriter, req *http.Request) {
```
srikanthccv: how about creating a separate handler for ai_observability fields

srikanthccv: the current shape has AI policy leaking into the generic module, `implfields` now imports `aitelemetryschema` and encodes AI-specific decisions. The generic fields module has no reason to know the AI schema exists.

pkg/statementbuilder/aistatementbuilder/statement_builder.go
```diff
@@ -25,7 +26,7 @@ func NewFactory(
 // Scope describes gen_ai for the scoped trace builder: an AI trace has >=1 gen_ai
 // LLM, tool, or agent span, and its list adds AI/LLM per-trace metrics.
 func Scope() scopedtraces.TraceScope {
```
srikanthccv, line 29: `TraceAggregateFields` now duplicates the `Filterable: true` aliases here, checked only by a sync test. `aistatementbuilder` already imports `aitelemetryschema`, so `Scope()` can derive `Filterable` from map membership

pkg/types/telemetrytypes/field.go
```diff
@@ -282,6 +303,18 @@ type PostableFieldValueParams struct {
 	ExistingQuery string `query:"existingQuery"`
 }
 
+// existingQuery is unsupported until the computed per-trace aggregates it may
```
srikanthccv: and have them moved to `pkg/types/aiobservabilitytypes`

srikanthccv: while we are doing this, let's define the narrowed `PostableFieldKeysParams` to remove unnecessary params `signal`, `source`, `metricName`, `metricNamespace`

pkg/types/telemetrytypes/field.go
```diff
@@ -362,6 +395,21 @@ func NewFieldValueSelectorFromPostableFieldValueParams(params PostableFieldValue
 	return fieldValueSelector
 }
 
+func NewFieldValueSelectorFromPostableAIObservabilityFieldValueParams(params PostableAIObservabilityFieldValueParams) *FieldValueSelector {
```
srikanthccv: same this will be moved and you can remove the `AIObservability` from the function name

pkg/types/telemetrytypes/field.go
srikanthccv: difference b/w the `pkg/telemetryschema/aitelemetryschema` vs `pkg/types/aiobservabilitytypes`

`pkg/telemetryschema/*` describes the telemetry data itself i.e what fields exist, their contexts and types. They are consumed by statement builders and the metadata layer. `pkg/types/*` is where HTTP contracts live.

 A `Postable*Params` struct inside a package named `telemetryschema` would have the package name lying about its contents, and `aistatementbuilder` that has no business seeing params.

pkg/statementbuilder/aistatementbuilder/statement_builder_test.go
```diff
@@ -1056,3 +1058,16 @@ func TestBuild_TraceList_VariableInAggregateFilter(t *testing.T) {
...
+// that computes them; either half alone is unusable.
+func TestScope_FilterableColumnsMatchSchemaAggregates(t *testing.T) {
+	var filterable []string
+	for _, c := range Scope().Columns {
+		if c.Filterable {
+			filterable = append(filterable, c.Alias)
+		}
+	}
+
+	assert.ElementsMatch(t, slices.Collect(maps.Keys(aitelemetryschema.TraceAggregateFields)), filterable)
```
srikanthccv: you can remove this test altogether if we can maintain one source of truth

pkg/modules/fields/implfields/handler.go
```diff
@@ -51,6 +55,86 @@ func (handler *handler) GetFieldsKeys(rw http.ResponseWriter, req *http.Request)
...
+		return
+	}
+	orgID := valuer.MustNewUUID(claims.OrgID)
+
+	fieldKeySelector := telemetrytypes.NewFieldKeySelectorFromPostableFieldKeysParams(params)
+
+	keys := make(map[string][]*telemetrytypes.TelemetryFieldKey)
+	complete := true
+	// the trace context names the computed per-trace aggregates, which no scan can serve
+	if fieldKeySelector.FieldContext != telemetrytypes.FieldContextTrace {
```
srikanthccv: `FieldContext` this checks the context trace, but the generated don't make this a enum. `FieldContextTrace` is still commented out of `FieldContext.Enum()`, please update.

srikanthccv (commented):
Overall LGTM, some comments

srikanthccv (dismissed):
thanks

therealpandey (approved):

srikanthccv (approved):

PR #12145, opened 2026-07-17
fix: css fixes for llm o11y model pricing and overview 

tewarig:
## Summary

CSS-only fixes for AI observability model pricing and overview, split out of #12123 per review feedback:

- **Overview**: remove the negative `margin-top` compensation for the hidden dashboard page header

prev- 
 
image

Now - 
image

- **SourceSelector (ModelCostDrawer)**: drop the per-variant (`auto`/`override`) checked-state styling in favor of a single accent style, and remove the card background

Prev - 
image

Now - 
image

- **ExtraPricingBuckets**: remove the `--l2-background` on the bucket card

Prev -  
image

Now - 
image

- **UnpricedModelsTab**: add `margin-top` spacing to the banner

Prev - 
image

Now  - 

image

- **PatternEditor**: tighten spacing between pattern chips — drop the row gap, keep only the column gap (`var(--spacing-0) var(--spacing-3)`), and add `margin-bottom` below the chip row.

- **ModelCostDrawer**: constrain the drawer container to `overflow: hidden` and move scrolling into the inner content (`min-height: 0; overflow-y: auto`) so the  body scrolls on its own instead of the whole drawer.

No functional changes — only styling and the classNames that feed it.

Scroll on drawer body - 

https://github.com/user-attachments/assets/bb28a795-f260-4d6d-9f90-2875f6409efb

### DIFF
--- a/frontend/src/container/LLMObservability/Overview/Overview.module.scss
+++ b/frontend/src/container/LLMObservability/Overview/Overview.module.scss
@@ -9,6 +9,4 @@
 	:global([class*='dashboardPageHeader']) {
 		display: none;
 	}
-	// remove margin added by the hidden header to avoid extra whitespace at the top of the page
-	margin-top: calc(var(--spacing-8) * -1);
 }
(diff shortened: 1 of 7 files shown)


aks07:
Screenshot 2026-07-20 at 12 50 59 PM

Why this space left at the top?

tewarig:
> Screenshot 2026-07-20 at 12 50 59 PM
> Why this space left at the top?

fixed it 
image

aks07 (approved):

PR #12146, opened 2026-07-17
fix(dashboards-v2): panel editor Save/dirty, refresh retention, dropdown clipping & time-window fixes

ahrefabhi:
## Pull Request

---

A batch of correctness/UX fixes for the **dashboards V2 panel editor**, all around query state, navigation, and the Save (dirty) flow. Each was a distinct bug that made the editor feel unreliable:

1. **Dirty check was anchored to the wrong baseline, and query edits were lost on refresh.** The editor re-derived its "has unsaved changes" baseline from the incoming `panel` prop, which is seeded from *transient* state — a stale URL `compositeQuery` after a refresh, or the View‑mode handoff spec — instead of the persisted panel. Result: after a refresh the panel showed the saved query yet the Save button read *dirty* (inverted), and a **View → Edit** query change read *clean* with Save disabled. Separately, a refresh discarded in‑progress query edits entirely.
2. **Query‑builder dropdowns were clipped** in the editor (group‑by, order‑by, having, metric name, aggregator, units) because the builder sits in an `overflow:hidden` resizable pane.
3. **The preview didn't follow the staged query on browser Back/Forward** — the builder reverted but the preview kept the last Run's result.
4. **A custom time range picked in the editor was reset** when navigating between the dashboard and the editor.

The unifying approach for the Save/query issues is to stop treating transient URL/handoff state as the source of truth: the dirty check is anchored to the **persisted panel**, compared at the canonical **V5 query‑envelope** level, and the builder is allowed to **hydrate from the URL** so edits survive a refresh.

_To be added — recommend a short clip of: edit query → Run → refresh (edit retained, Save enabled) and View → Edit with a query change (Save enabled)._

Closes https://github.com/SigNoz/pulse-pod/issues/121

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [x] 🧪 Test-only

---

#### Root Cause
- **Dirty / refresh:** the dirty baseline (both spec and query) was derived from the `panel` prop on every editor mount. That prop is the handoff‑edited spec on **View → Edit**, and after a refresh the URL still carries the last‑run (edited) `compositeQuery`. The query baseline was additionally captured from the first `stagedQuery`, which *raced* the URL restore. So the editor compared the draft against an already‑edited baseline. A mount‑time `forceReset` also deliberately discarded the URL query, throwing away in‑progress edits on refresh.
- **Dropdowns clipped:** the shared query‑builder `Select` popups rendered inside the editor's `overflow:hidden` resizable pane.
- **Preview stale on Back/Forward:** the preview reads `draft.spec.queries`, which was only committed on Run; browser navigation reverted the builder but not the draft.
- **Time range reset:** editor navigation didn't carry the active time window in the URL.

#### Fix Strategy
- Thread a **`savedPanel`** baseline (the persisted `existingPanel`) through `PanelEditorPage → PanelEditorContainer → usePanelEditSession`, distinct from the seed `panel`. `usePanelEditorDraft` compares `isSpecDirty` against `savedPanel`; `usePanelEditorQuerySync` compares the live query against `savedPanel.spec.queries`.
- Compute `isQueryDirty` as a **V5‑envelope** comparison, routing *both* sides through the same `fromPerses → toPerses` round‑trip so builder‑added default fields (absent from an older stored query) never read an untouched panel as modified. This replaces the racy captured‑`stagedQuery` baseline.
- Remove the mount `forceReset` on `useShareBuilderUrl` (both its original reasons are now moot) so a refresh / Back‑Forward **hydrates the edited query from the URL** and the existing staged‑query effect syncs it into the draft → **query edits survive a refresh**.
- Portal the builder's `Select` popups to an ancestor `ConfigProvider` container (falling back to `trigger.parentNode`) via `useSelectPopupContainer`, and wrap the editor's builder in a `ConfigProvider` that renders popups to `document.body`.
- Carry the active time window in the editor URL via `useTimeSearchParams` / `timeUrlParams` helpers used by `useOpenPanelEditor` / `useCreatePanel`.

---

- **Tests added/updated:**
  - New real‑`QueryBuilderProvider` integration tests (`usePanelEditorQuerySync.integration.test.tsx`) exercising the actual builder normalization: untouched panel → not dirty (incl. an older/minimal stored query), refresh retention → edit kept + dirty, View → Edit handoff → dirty.
  - Updated `usePanelEditorQuerySync.test.ts` and `usePanelEditorDraft.test.ts` for the saved‑panel baseline and envelope comparison.
  - New tests for `useOpenPanelEditor`, `useCreatePanel`, `useTimeSearchParams`, and `timeUrlParams`.
- **Manual verification:** edit query → Run → refresh (edit retained, Save enabled); re‑apply edits after refresh (dirty toggles correctly); open editor → View → change query → back to Edit (Save enabled); open group‑by/order‑by/having/units dropdowns in the editor (not clipped); pick a custom range in the editor and navigate (range preserved).
- **Edge cases covered:** older/minimal stored query shapes (drift), new panels (no saved baseline → seed fallback), List panels (bare `BuilderQuery` envelope), View modal drilldown (no `savedPanel` → seed baseline unchanged).
- Full `DashboardPageV2` jest suite green; `tsgo --noEmit` + `oxlint` clean.

---

- **Blast radius:** almost entirely the V2 dashboards panel editor. The one shared touch‑point is the query‑builder popup container: `components/QueryBuilderV2/.../filters/*` now defer to an ancestor `ConfigProvider`'s popup container via `useSelectPopupContainer`, falling back to the previous `trigger.parentNode` behavior when no ancestor container is set — so other surfaces (explorers, alerts) are unchanged.
- **Potential regressions:** query‑builder dropdown positioning on non‑editor surfaces (mitigated by the `parentNode` fallback); the dirty flag / Save enablement in the editor (covered by unit + real‑provider integration tests). Config‑spec edits (title, thresholds, units) are still *not* retained across a hard refresh — only the query rides the URL; called out below.
- **Rollback plan:** revert this PR; the changes are self‑contained to the panel editor and the shared popup‑container helper.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Dashboards (v2) panel editor: the Save button now reflects unsaved changes correctly after a refresh and when switching between View and Edit; in‑progress query edits are kept across a refresh; query‑builder dropdowns no longer get clipped; the preview follows the query on browser Back/Forward; and a custom time range picked in the editor is preserved while navigating. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- The core of the dirty fix is comparing at the **V5 envelope level with both sides round‑tripped through `fromPerses → toPerses`**. Comparing the raw stored `savedQueries` reads *always‑dirty*, because an older stored query lacks builder‑default fields (`source`, `stepInterval`, `filter`, `spaceAggregation`, …) that `toPerses(currentQuery)` re‑emits — the round‑trip normalizes both sides. `usePanelEditorQuerySync.integration.test.tsx` guards this against the real builder provider.
- Removing `forceReset` means the editor now hydrates from the URL on mount; the pre‑existing staged‑query commit effect keeps the preview in step, and the dirty check is anchored to `savedPanel`, so a retained URL edit reads as dirty (not stale).
- **Known limitation (intentional, out of scope):** only the *query* rides in the URL, so a hard refresh does not retain in‑progress *config* edits (title, thresholds, units). Retaining those would require persisting the draft spec.
- Best reviewed commit‑by‑commit — each of the four fixes is an isolated commit.

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/hooks/usePanelEditorQuerySync.ts
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/hooks/usePanelEditorQuerySync.ts
@@ -1,7 +1,8 @@
-import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef } from 'react';
 import type {
 	DashboardtypesPanelDTO,
 	DashboardtypesPanelSpecDTO,
+	DashboardtypesQueryDTO,
 	TelemetrytypesSignalDTO,
 } from 'api/generated/services/sigNoz.schemas';
 import { initialQueriesMap, PANEL_TYPES } from 'constants/queryBuilder';
@@ -27,6 +28,12 @@ interface UsePanelEditorQuerySyncArgs {
 	alwaysSerializeQuery?: boolean;
 	/** Signal to seed a new panel's builder with — the kind's first supported signal. */
 	signal?: TelemetrytypesSignalDTO;
+	/**
+	 * The persisted panel's queries — the dirty baseline. Distinct from `draft.spec.queries`,
+	 * which the builder seeds from and may carry unsaved edits handed off from View mode. Omit
+	 * for a new panel, where the seed query is the baseline.
+	 */
+	savedQueries?: DashboardtypesQueryDTO[];
 }
 
 interface UsePanelEditorQuerySyncApi {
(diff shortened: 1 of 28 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/hooks/usePanelEditorQuerySync.ts
```diff
@@ -136,6 +131,15 @@ export function usePanelEditorQuerySync({
 		// eslint-disable-next-line react-hooks/exhaustive-deps -- structural change only
 	}, [currentQuery.queryType, dataSourceSignature]);
 
+	// Follow the staged (executed) query into the draft on a URL re-stage (mount
+	// hydration, browser Back/Forward) so the preview matches. Live edits touch only
+	// currentQuery, so they still wait for Run; commitQuery no-ops when unchanged.
+	useEffect(() => {
+		if (stagedQuery) {
```
AshwinBhatkal, line 138: could this check also include a ref that includes a value. This will run on any stage right?

mahenoor2106-pixel: Yes, it can include a ref that contains a value, depending on how the check is implemented.
And yes, if the check is configured at a level that applies globally (rather than to a specific stage), it will run on every stage. If it's attached only to a particular stage, then it will run only for that stage.

AshwinBhatkal (approved):

PR #12152, opened 2026-07-18
fix(clickhouseprometheus): merge fingerprints sharing a labelset instead of injecting a fingerprint label

srikanthccv:
Stacked on #12151.

Problem: When a label's value becomes empty, the series gets a new fingerprint in storage, but the empty label is dropped when we read it back so two different fingerprints end up with the same labels. #8563 worked around this by adding a fake `fingerprint` label to every series and removing it from the final results. The problem is that the fake label is still there while PromQL evaluates the query: `without()` split series into one group per fingerprint and then produced duplicate output series with identical labels, and plain vector matching (like `a / b`) never found a match.

Fix handle series identity in the adapter instead:
- Drop labels with empty values when unmarshaling.
- If two fingerprints end up with the same labels, merge them into one series (samples combined in timestamp order; if two samples share a timestamp, the higher fingerprint wins).
- Remove the fake label injection and every `RemoveExtraLabels` call.
- Queries that are genuinely ambiguous (e.g. `rate({__name__=~"a|b"}[5m]))` still return an error, same as upstream Prometheus.

Why the adapter is the right place:
- In Prometheus, a series is its labelset. Whenever separate physical copies of a series meet during a read, Prometheus merges the ones with equal labels `storage/merge.go` does this, and it's used for blocks (`tsdb/db.go`) and local+remote reads (`storage/fanout.go`).
- We already use that machinery when combining multiple queriers. But inside a single Read, the engine gets one already-built SeriesSet, so the per-fingerprint copies are never compared anywhere upstream. The only place left to merge them is the adapter.

OTel note: dropping empty-valued labels is a deliberate translation between two models:
- The OTel spec says empty values are meaningful and must be kept so `{a=""}` and `{}` are two different streams in OTel, and they stay separate in storage.
- Prometheus says an empty value is the same as no label at all. So at the PromQL boundary, both become one series. The merge is what translates one identity model into the other.

Cost:
- One extra hash and map insert per fetched series — about 0.6 µs and 100 bytes each. At 500k series in one selector that's ~310 ms and 50 MB of temporary memory; past that scale the query service runs out of memory on the fetch itself anyway.
- The sample-merging code only runs when a collision actually happens.

Testing:
- Fixes 176 of the 192 differences we had against Prometheus's own promqltest corpus (the test suite comes in a follow-up PR in this stack).
- Regression test: TestClient_QuerySamplesMergesIdenticalLabelSets. An end-to-end test covering the exact #8563 scenario ships with the suite PR.


srikanthccv:
Example

```
http_requests_total{service="checkout", instance="pod-1"}  = 10
http_requests_total{service="checkout", instance="pod-2"}  = 20
```

the injected label makes them look like this to the engine:

```
http_requests_total{service="checkout", instance="pod-1", fingerprint="9f3a"}  = 10
http_requests_total{service="checkout", instance="pod-2", fingerprint="c41b"}  = 20
```

Query:

```
sum without(instance) (http_requests_total)
```

What should happen: without(instance) removes instance, both series land in the same group, and you get one series:

```
{service="checkout"} = 30
```

What actually happened: `without(instance)` only removes instance and the injected fingerprint label stays in the grouping key. The two series go into two different groups, so nothing is summed:

```
{service="checkout", fingerprint="9f3a"} = 10
{service="checkout", fingerprint="c41b"} = 20
```

Then the strip removes fingerprint from the results, leaving two output series with identical labels:

```
{service="checkout"} = 10
{service="checkout"} = 20
```

therealpandey (approved):

PR #12155, opened 2026-07-18
fix(querier): strip only known internal label keys from promql results

srikanthccv:
Stacked on #12154.

The v5 output filter dropped every `__`-prefixed label from promql results. Label names are user data and may legitimately start with `__` (e.g. `__address__`), so the blanket strip mangled labelsets. Hide only known SigNoz storage keys: `__temporality__` and the `__scope.`/`__resource.` prefixes, which cover every exporter version's keys.


srikanthccv:
Getting incorporated into #12156 directly.

PR #12158, opened 2026-07-18
fix(dashboards-v2): round-trip zero-valued threshold + query fields

therealpandey:
## What & why

Two server-side bugs in the **v2 dashboards API** (`/api/v2/dashboards`) broke a create → read round-trip: a valid, UI-authored dashboard failed to apply cleanly through any typed client that echoes back what it sent (Terraform, SDKs, `PUT`-after-`GET`). Found while building Terraform acceptance tests (`terraform-provider-signoz`) against real dashboards from `SigNoz/dashboards`.

### Bug 1 — threshold `value: 0` rejected on create
A NumberPanel/TimeSeries/Table threshold with `value: 0` (a default the UI emits) returned `400 dashboard_invalid_input` — `Field validation for 'Value' failed on the 'required' tag`.

go-playground/validator's `required` treats a numeric field equal to its zero value as "missing", so `validate:"required"` on the float `Value` wrongly rejected `0`.

**Fix:** drop `validate:"required"` from `Value` on `ThresholdWithLabel` and `ComparisonThreshold`. `required:"true"` is kept — the field is always present in the schema (`0` is a valid *present* value, not an absent one), so the **OpenAPI spec and generated frontend client are unaffected** and `value` stays a required `number`. `Color` keeps both tags.

### Bug 2 — `omitempty` dropped zero-valued query fields on read-back
A query that set `disabled: false`, `legend: ""`, or an explicit empty `groupBy`/`order`/`selectFields`/etc. created fine, but the `GET` response **omitted** the field, so a typed client read back `null`/absent and reported drift.

**Fix:** make the create → GET path echo what was stored.
- **Slice fields → `,omitzero`** (not a bare `omitempty` removal): `omitzero` omits a `nil` slice (never-set stays absent) but keeps an explicit non-nil `[]`. So an empty array round-trips as `[]` **and there is no `null` regression**. Applied to `groupBy`, `order`, `selectFields`, `aggregations`, `functions`, `secondaryAggregations`, and function `args` across the builder / formula / trace-operator / join specs, plus `ListPanelSpec.selectFields`.
- **Scalars `disabled` (bool) / `legend` (string) → tag removed:** `omitzero`/`omitempty` both suppress `false`/`""`, so the only way to round-trip them is to always serialize.
- `resp.go` result types keep `,omitempty` — they are server-computed and never round-tripped.

## Scope & consequences (reviewer note)

- These query structs are **shared** across query-range v5, so the round-trip fix also applies to **alert rules** (their stored blobs and GET responses now include `disabled: false` / `legend: ""` and echo explicit empty slices). Consistent, not breaking.
- The `omitzero` slices are now `nullable: true` in the OpenAPI schema, so the generated TS types gain `| null`. The wire **never** carries `null` (omitzero omits `nil`); a frontend audit confirmed every consumer already handles this via `?? []` / `.length` / truthy checks, and the dashboard dirty/stale-check pipeline compares server-shaped values, so no spurious "unsaved changes".

## Verification

- `go test` for the changed packages and every consumer (querybuildertypesv5, dashboardtypes, ruletypes, querier, dashboard module, apiserver) — pass
- Go lint — 0 issues
- Regenerated `docs/api/openapi.yml` and the frontend client; both match their generators (CI `git diff --exit-code` gates green); frontend type-check passes


srikanthccv:
The addl candidates I think we should address

- `TelemetryFieldKey` https://github.com/SigNoz/signoz/blob/dd8cbe084443642dd98bcab6c7ebd5cec4f8fe2a/pkg/types/telemetrytypes/field.go#L34-L38. All of these can be deliberately empty, for instance, you can leave the context or data type empty to work across multiple context. This type is used in quite a number of places, selected fields, group by, order by etc. Can we check this field once?

- type `StepInterval`, right now, it is `omitempty`, and given `omitempty` on a struct is a nop, I guess it responds with 0 instead of null when user doesn't send it.

- similarly for the `source` same problem as above?

And on different dashboard types we have `ValueOrDefault` in couple of places, last time when I asked about this being a problem, I was told it's not a problem as that could be made computed field, can we check that once? example, `TimePreference`, `ComparisonOperator` or whoever there is `ValueOrDefault`

- `reduceTo` on `MetricAggregation` has `omitzero` that might be problem if they send `""` value for it.

- Function argument names are not mandatory, and the name can be absent when empty string

- `PrecisionOption` was made to work with string because it can be speicied as "full", does this type change become problem (also talked about this)

- for the custom enum types defined in dashboard types, by the time the type see it, the value is lowercased, so I think there a chance of accepting `ASC` for sort and get back `asc`.


Maybe not be deliberatly set empty but still worth considering

on the `QueryBuilderQuery` type https://github.com/SigNoz/signoz/blob/dd8cbe084443642dd98bcab6c7ebd5cec4f8fe2a/pkg/types/querybuildertypes/querybuildertypesv5/builder_query.go#L14
- offset
- cursor
- alias of Log/Trace/MetricAggregation

srikanthccv (approved):
to be followed up in another pr for remaining gaps

srikanthccv:
Some more

- Description field of Display, we use this in several places https://github.com/SigNoz/signoz/blob/22e55340d7dbfca6bdc6f2cc41cf607ddcfdaad2/pkg/types/dashboardtypes/perses_replicas.go#L25
- Constant of the variable spec https://github.com/SigNoz/signoz/blob/22e55340d7dbfca6bdc6f2cc41cf607ddcfdaad2/pkg/types/dashboardtypes/perses_replicas.go#L281
- List variable, empty string for no sort, and other two https://github.com/SigNoz/signoz/blob/22e55340d7dbfca6bdc6f2cc41cf607ddcfdaad2/pkg/types/dashboardtypes/perses_replicas.go#L164-L166
- `omitzero` for panel links https://github.com/SigNoz/signoz/blob/22e55340d7dbfca6bdc6f2cc41cf607ddcfdaad2/pkg/types/dashboardtypes/perses_replicas.go#L79 and dashboard links https://github.com/SigNoz/signoz/blob/22e55340d7dbfca6bdc6f2cc41cf607ddcfdaad2/pkg/types/dashboardtypes/perses_dashboard_data.go#L29
  - we may have to replicate the `dashboard.Link`
- QuerySpec name https://github.com/SigNoz/signoz/blob/22e55340d7dbfca6bdc6f2cc41cf607ddcfdaad2/pkg/types/dashboardtypes/perses_replicas.go#L92


- For Thresholds on panel spec, if user doesn't set anything, and response includes, `null`, is that a drift problem?
- Same for `CustomColors` and `ColumnUnits`, they are not marked nullable today

PR #12162, opened 2026-07-19
fix(dashboards-v2): round-trip all zero-valued spec fields

therealpandey:
Completes the dashboards-v2 create → GET round-trip: every zero value a typed client (Terraform / SDK / PUT-after-GET) can set now echoes back verbatim instead of being dropped.

## Fixed
- `Display.Description`, `TextVariableSpec.Constant`, `ListVariableSpec.CustomAllValue` / `CapturingRegexp` — drop `omitempty` so `""` / `false` round-trip.
- `DashboardV2.Image`, `DashboardSpec.Duration` / `RefreshInterval` — drop `omitempty` so `""` round-trips (server already accepts `""`: `DurationString` allows empty, image has no validation).
- `PanelSpec.Links` / `DashboardSpec.Links` — `omitzero` so an explicit `[]` round-trips, **and** replace the imported perses `dashboard.Link` with a SigNoz-owned `Link` replica whose `name` / `tooltip` / `renderVariables` / `targetBlank` always serialize (perses tagged them `omitempty`, dropping their zero values).
- `DashboardSpec.Datasources` — `omitzero` so an explicit `{}` round-trips.
- Extended the v2 dashboards integration round-trip test to cover all of the above.
- Regenerated `docs/api/openapi.yml` + frontend client (`links` and `datasources` are now `nullable`).

## Not changed (intentional)
- `Thresholds`, `Legend.CustomColors`, `TableFormatting.ColumnUnits`, grid `GridItem` positions — no `omitempty`, so explicit `[]` / `{}` / `0` already round-trip; unset → `null` is consistent for a typed client (not a drift) and already `nullable` in the schema.
- `ListVariableSpec.Sort` — `omitzero`; "no sort" is the value `"none"`, so only the invalid `""` is omitted.
- `QuerySpec.Name` — the query-envelope name isn't a meaningful zero value (it's an identifier; the real name lives in `plugin.spec`).
- Enum defaulting (`decimalPrecision`, legend `position` / `mode`, `lineInterpolation`, …) — these emit their default when unset by design; explicit values round-trip.

## Deferred
- **Deferred** (need replicating a perses type, low-impact): `DatasourceSpec.Display` (perses `common.Display` drops inner `""`) and `GridLayoutSpec.RepeatVariable`.

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/hooks/useDrilldown.tsx
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/hooks/useDrilldown.tsx
@@ -218,7 +218,7 @@ export function useDrilldown(
 				context={context}
 				query={v1Query}
 				isResolving={isResolving}
-				links={panel.spec.links}
+				links={panel.spec.links ?? undefined}
 				canSetDashboardVariables={dashboardVariables.hasFieldVariables}
 				onViewLogs={(): void => navigate('view_logs')}
 				onViewTraces={(): void => navigate('view_traces')}
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -2734,6 +2734,7 @@ components:
         datasources:
           additionalProperties:
             $ref: '#/components/schemas/DashboardtypesDatasourceSpec'
+          nullable: true
           type: object
         display:
           $ref: '#/components/schemas/DashboardtypesDisplay'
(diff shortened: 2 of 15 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/hooks/useDrilldown.tsx
```diff
@@ -218,7 +218,7 @@ export function useDrilldown(
 				context={context}
 				query={v1Query}
 				isResolving={isResolving}
-				links={panel.spec.links}
+				links={panel.spec.links ?? undefined}
```
srikanthccv, line 221: @ahrefabhi @AshwinBhatkal see this once before rollout

srikanthccv (approved):

PR #12165, opened 2026-07-19
refactor(clickhouseprometheus): build lookup and samples SQL with sqlbuilder

srikanthccv:
Stacked on #12154.

Replaces hand-numbered `$N` placeholders and string concatenation with `huandu/go-sqlbuilder` (ClickHouse flavor), matching telemetrymetrics and the v2 transpiler. The samples query embeds the series lookup as a nested builder, so argument order is merged by the builder in render position — the manual `argOffset` bookkeeping is gone. Rendered SQL shape and argument order are unchanged (lookup time bounds stay integer literals to keep the query byte-compatible); unit goldens pin the `?`-flavor rendering including the nested `GLOBAL IN (SELECT ...)`.

No behavior change: conformance suite green at this tip, 0 divergences across all 755 cases.

### DIFF
--- a/pkg/prometheus/clickhouseprometheus/capture.go
+++ b/pkg/prometheus/clickhouseprometheus/capture.go
@@ -66,11 +66,11 @@ func (c *captureClient) Read(ctx context.Context, query *prompb.Query, _ bool) (
 	}
 
 	// Build the executing path's queries, but only record them.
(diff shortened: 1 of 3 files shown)


srikanthccv:
Getting incorporated into #12154 directly.

PR #12171, opened 2026-07-20
fix(querybuildertypesv5): round-trip unset stepInterval and empty field-key values

therealpandey:
Some query fields didn't survive a create → GET round-trip. If you left them empty/unset, the API either echoed back a value you never sent or dropped your empty value — so a typed client (Terraform / SDK) that read the query back saw something different from what it sent. Came out of the #12158 review.

### What changed
- **`stepInterval`** now disappears when you don't set it. It's a struct, so the old `omitempty` did nothing and an unset step was coming back as `0`. Switched to `omitzero`.
- **A field key's `signal` / `fieldContext` / `fieldDataType`** can be deliberately left empty to mean "any", and that empty value now round-trips instead of being dropped. To make an empty value legal on the wire, it's now a member of each enum — the same fix #12164 used for `source`.
- **Frontend:** the `signal` enum is shared with the variable/panel signal pickers, where empty isn't a real choice, so those were narrowed back to `logs`/`traces`/`metrics` (keeping them exhaustive) and an empty drilldown signal is treated as "All".
- Regenerated the OpenAPI spec + frontend client, and updated a marshal test plus the querierlogs aggregation test whose labels now carry the empty fields.

### Next Fields to Review
 `reduceTo`, function-arg names, aggregation aliases, `offset`, `cursor`

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/PanelTypeSwitcher/utils.ts
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/PanelTypeSwitcher/utils.ts
@@ -17,6 +17,9 @@ const SIGNAL_LABEL: Record<TelemetrytypesSignalDTO, string> = {
 	[TelemetrytypesSignalDTO.logs]: 'logs',
 	[TelemetrytypesSignalDTO.traces]: 'traces',
 	[TelemetrytypesSignalDTO.metrics]: 'metrics',
+	// The empty "any" signal only appears on field keys, never on a panel query;
+	// mapped for exhaustiveness.
+	[TelemetrytypesSignalDTO['']]: '',
 };
 
 /**
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -8599,6 +8599,7 @@ components:
       - traces
       - logs
       - metrics
+      - ""
       type: string
     TelemetrytypesSource:
       enum:
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -8558,6 +8558,7 @@ components:
       - resource
       - attribute
       - body
+      - ""
       type: string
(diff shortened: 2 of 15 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/PanelTypeSwitcher/utils.ts
```diff
@@ -17,6 +17,9 @@ const SIGNAL_LABEL: Record<TelemetrytypesSignalDTO, string> = {
 	[TelemetrytypesSignalDTO.logs]: 'logs',
 	[TelemetrytypesSignalDTO.traces]: 'traces',
 	[TelemetrytypesSignalDTO.metrics]: 'metrics',
+	// The empty "any" signal only appears on field keys, never on a panel query;
+	// mapped for exhaustiveness.
+	[TelemetrytypesSignalDTO['']]: '',
```
tushar-signoz, line 22: Can we double check this ? 
Otherwise LGTM

tushar-signoz (approved):

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/PanelTypeSwitcher/utils.ts
therealpandey: The dashboard folks will take a look. It's V2 dashboards.

srikanthccv: @AshwinBhatkal @ahrefabhi this one and other one here https://github.com/SigNoz/signoz/pull/12162#discussion_r3611315417

srikanthccv (approved):

AshwinBhatkal (approved):

PR #12178, opened 2026-07-20
fix(dashboard-v2): variable, panel-layout, and dashboards-list bug-bash fixes

AshwinBhatkal:
## Pull Request

---

### 📄 Summary

A batch of V2 dashboard fixes across variables, panel layout, and the dashboards list, from the latest bug-bash. Each concern is its own commit so they can be reviewed (or reverted) independently.

Variables & form
- **Default reset** now fires only on a real option-set change (Test Run for query, attribute/signal change for dynamic, options edit for custom) — never on sort-only changes or on edit-open.
- **Multiselect** buffers edits while open and commits once on close (no per-toggle cascade); closing with nothing selected falls back to the default, so a variable is never left empty (which stranded dependents on a spinner). Tags now truncate with a bare "+N" overflow and the control is 180px wide.
- **Single-select** keeps a user-typed value across a refetch (e.g. a time-range change no longer wipes it).
- **Save** waits for the backend before closing the form; on failure the form stays open (button disabled) and the exact backend error is surfaced, instead of silently "saving".
- Apply-to-panels tags show panel **names** (not ids) when closed, and the dropdown anchors to the body so it no longer shifts when opened near the bottom of the form.

Panel layout
- Panel geometry is **compacted to a non-overlapping layout** before persisting — on drag/resize and on JSON-editor save — so the backend's no-overlap validation can't reject a valid-looking edit.
- Section creation guards against a double submit (double Enter no longer creates two sections).

Dashboards list
- Locking a dashboard from the list page now updates the detail-page cache, so returning to it reflects the locked state.
- The "+N" tags indicator opens a popover of the remaining tags instead of navigating to the dashboard; the name tooltip is width-capped (480px); non-interactive tag badges no longer show a clickable hover state.

#### Issues closed by this PR
Closes https://github.com/SigNoz/pulse-pod/issues/117

---

### ✅ Change Type

- [x] 🐛 Bug fix
- [x] 🧪 Test-only (unit tests added for the pure logic)

---

#### Root Cause
Several independent issues: default-reset keyed on the wrong signal; the multiselect cascaded per toggle and could commit an empty selection; single-select values were dropped by post-fetch reconciliation; the variable form closed before the save resolved; `onDragStop`/JSON-save persisted a not-yet-compacted (overlapping) layout; the list-page lock never invalidated the detail-query cache; and the tags "+N" fell through to the row's navigation handler.

#### Fix Strategy
Each is addressed at its source and committed separately — see the per-commit messages and the Summary above.

---

### 🧪 Testing Strategy

- Tests added/updated: `resolveVariableSelection` (typed-value persistence), `layoutCompaction` (compaction + panel-ref preservation), `useVariableForm` (default-reset semantics), plus a hoist-order regression check.
- Manual verification: pending in-app validation of the interactive flows (multiselect commit-on-close, drag/resize, share-agnostic surfaces).
- Edge cases covered: ALL selection, empty-close → default, re-run with identical values, non-Grid layouts.

---

- Blast radius: V2 dashboard variables bar/form, panel grid persistence, and the dashboards list row. `TagBadge` is shared (hover-only change).
- Potential regressions: multiselect selection timing; layout compaction on drag/JSON save.
- Rollback plan: revert the specific concern's commit.

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/useJsonEditor.ts
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/useJsonEditor.ts
@@ -12,6 +12,7 @@ import { toAPIError } from 'utils/errorUtils';
 
 import { dashboardToUpdatable } from './dashboardToUpdatable';
 import { findPanelLayoutIssues } from './danglingPanels';
+import { compactSpecLayouts } from '../../layoutCompaction';
 import { useDashboardStore } from '../../store/useDashboardStore';
 
 export interface JsonValidity {
@@ -46,15 +47,16 @@ interface Result {
 }
 
 /**
- * The editable, user-facing view: only `tags` and `spec`. Everything else
- * (id, orgId, name, timestamps, locked, schemaVersion, image, …) is redacted so it
- * can't be seen, copied, exported or edited; those keys are preserved on save (see `apply`).
+ * The editable, user-facing view: `spec`, `tags` and `image`, in that key order.
+ * Everything else (id, orgId, name, timestamps, locked, schemaVersion, …) is redacted
+ * so it can't be seen, copied, exported or edited; those keys are preserved on save.
  */
 const redact = (
 	dashboard: DashboardtypesGettableDashboardV2DTO,
-): Pick<DashboardtypesGettableDashboardV2DTO, 'tags' | 'spec'> => ({
-	tags: dashboard.tags,
+): Pick<DashboardtypesGettableDashboardV2DTO, 'spec' | 'tags' | 'image'> => ({
 	spec: dashboard.spec,
+	tags: dashboard.tags,
+	image: dashboard.image,
 });
 
 const serialize = (dashboard: DashboardtypesGettableDashboardV2DTO): string =>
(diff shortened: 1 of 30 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/useJsonEditor.ts
```diff
@@ -167,7 +169,18 @@ export function useJsonEditor({
...
+				'spec' | 'tags' | 'image'
+			>;
+			// Snap hand-edited panel geometry to a non-overlapping layout so a JSON edit
+			// can't be rejected by the backend's no-overlap check (matches drag/resize).
+			if (edited.spec?.layouts) {
+				edited.spec = {
+					...edited.spec,
+					layouts: compactSpecLayouts(edited.spec.layouts),
+				};
+			}
```
H4ad, line 183: If I write the json and spec.layouts in a format we are not expecting, it will throw and show like api error?

H4ad (approved):

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardPageToolbar/JsonEditorDrawer/useJsonEditor.ts
AshwinBhatkal: yes. For now, it will show API error. It will try to compact how rgl how try to. If that is not working, then it will

Schema integration we need to do for this

PR #12188, opened 2026-07-20
fix(metrics): resolve any filter key context instead of erroring

tushar-signoz:
## Summary

Metric query filters returned **HTTP 400 (`key <name> not found`)** for any filter term that isn't a metadata-registered key:

- a bare / quoted **free-text** term, e.g. `abc` or `"k8s"` (the partial token the UI sends while a filter is still being typed), and
- an **explicit label context that isn't registered**, e.g. `resource.region` / `scope.region`.

This broke the **Metrics Explorer** summary/treemap (`/api/v2/metrics/treemap`, `/stats`) and metric query-builder filters of these shapes.

Metrics has **no per-context storage** — every label lives in the `labels` JSON on `time_series_v4`, and the field mapper already maps `resource.` / `scope.` / `attribute.` / bare all to `JSONExtractString(labels, '<key>')`. The fix uses the key directly in the `len(keys)==0` branch and lets `FieldFor` map it (intrinsics incl. the full-text column `metric_name` → their column; every label context → a labels extract), so the query runs (matching nothing) instead of 400ing. This restores the pre-#11911 behaviour and makes filter context-handling consistent with group-by, which already collapses all label contexts to `labels`.

Regressed by #11911, which rewrote the metrics condition builder to resolve against metadata and hard-error on a miss. The equivalent key-not-found path was made graceful for traces (#12091) and logs (#12134), but the metrics condition builder was left as the outlier.

Fixes https://github.com/SigNoz/engineering-pod/issues/5756

## Changes

- `pkg/telemetrymetrics/condition_builder.go` — on `len(keys)==0`, use the key directly instead of returning `NewKeyNotFoundError`.
- `pkg/telemetrymetrics/condition_builder_test.go` — unit coverage for the branch: intrinsic `metric_name` full-text, context-less label, and explicit `resource.` context all resolve (to their column / a labels extract), no error, no warning.
- `tests/integration/tests/queriermetrics/10_key_resolution.py` — `test_metrics_filter_label_context` updated to assert all label contexts collapse to the same labels extract; new `test_metrics_full_text_filter_does_not_error`.

## Testing

- `go test ./pkg/telemetrymetrics/...` ✓
- golangci-lint (primus) ✓ — 0 issues
- `make py-fmt` (unchanged) / `make py-lint` ✓
- Verified against a local enterprise server on the quick-bee ClickHouse: treemap free-text, query_range free-text, and all label contexts (`bare` / `attribute.` / `resource.` / `scope.`) now return 200 (were 400).

### DIFF
--- a/pkg/telemetrymetrics/condition_builder.go
+++ b/pkg/telemetrymetrics/condition_builder.go
@@ -168,7 +168,12 @@ func (c *conditionBuilder) ConditionFor(
 		warnings = append(warnings, warning)
 	}
 	if len(keys) == 0 {
-		return nil, warnings, querybuilder.NewKeyNotFoundError(key.Name)
+		// Metrics keeps every label in the `labels` JSON with no per-context storage, so
+		// FieldFor maps any key on its own: intrinsics (incl. the full-text column
+		// metric_name) to their column, every label context to a labels JSON extract.
+		// Use the key directly so a full-text term or an unregistered context resolves
+		// and the query runs instead of 400ing.
+		keys = []*telemetrytypes.TelemetryFieldKey{key}
 	}
 
 	conds := make([]string, 0, len(keys))
--- a/pkg/telemetrymetrics/condition_builder.go
+++ b/pkg/telemetrymetrics/condition_builder.go
@@ -162,40 +162,32 @@ func (c *conditionBuilder) ConditionFor(
 		return nil, nil, err
 	}
 
-	keys, warning := querybuilder.ResolveKeys(key, querybuilder.MatchingFieldKeys(key, fieldKeys))
+	keys := querybuilder.MatchingFieldKeys(key, fieldKeys)
 	var warnings []string
-	if warning != "" {
-		warnings = append(warnings, warning)
-	}
 	if len(keys) == 0 {
-		return nil, warnings, querybuilder.NewKeyNotFoundError(key.Name)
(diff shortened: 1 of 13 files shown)


pkg/telemetrymetrics/condition_builder.go
```diff
@@ -162,40 +162,32 @@ func (c *conditionBuilder) ConditionFor(
...
 
-	keys, warning := querybuilder.ResolveKeys(key, querybuilder.MatchingFieldKeys(key, fieldKeys))
+	keys := querybuilder.MatchingFieldKeys(key, fieldKeys)
 	var warnings []string
-	if warning != "" {
-		warnings = append(warnings, warning)
-	}
 	if len(keys) == 0 {
-		return nil, warnings, querybuilder.NewKeyNotFoundError(key.Name)
+		if _, isColumn := timeSeriesV4Columns[key.Name]; isColumn {
```
srikanthccv, line 168: any reason this doesn't use the `ResolveKeys` now? Is it because the `MatchingFieldKeys` doesn't include the intrinsic values? How are intrinsic fields in logs/traces working then?

tushar-signoz: ResolveKeys is only needed priortizing resource context fields in case of conflicts which doesn't matter for metrics.

therealpandey (approved):

PR #12189, opened 2026-07-20
docs: explain Azure config narrowing on resource_groups

tewarig:
## Why

Reviewers on the GCP branch asked why the Azure `AccountSettingsModal` / `useAccountSettingsModal` were touched. This adds a code comment so the change is self-explanatory.

## Context

`CloudAccount.config` is a shared per-provider union (`Azure | AWS | GCP`). The Azure code narrows the union with an `in` check. It used `deployment_region`, but GCP's new config **also** has `deployment_region`, so that field no longer uniquely identifies Azure — narrowing widened to `Azure | GCP` and `.resource_groups` access failed to compile (5× TS2339). The fix (already on the GCP branch) switched the discriminator to the Azure-only `resource_groups`; this PR just documents that at both call sites.

Comment-only change. No behavior change. `tsgo --noEmit` clean.


tewarig:
Superseded — commits landed directly on feat/add-gcp-integration instead of via PR.

PR #12201, opened 2026-07-21
fix(dashboards-v2): dedicated "all" dynamic-variable signal + required non-nullable links

AshwinBhatkal:
Two dashboards-v2 spec-fidelity fixes so query fields survive a create → GET round-trip for typed clients (Terraform / SDK).

## 1. Dynamic-variable signal: a dedicated `all` value

A dynamic variable's signal is now its own enum — `traces | logs | metrics | all` — with `all` (search across every signal) as the default. This is intentionally separate from a panel query's signal (which uses an empty value to mean "unpinned"): a variable's "search everywhere" is a real, named `all` value that round-trips verbatim instead of relying on an empty string.

- **Backend:** the dynamic-variable `signal` is a required, non-nullable field (`traces`/`logs`/`metrics`/`all`). An empty or otherwise-invalid value is rejected; an omitted field defaults to `all`.
- **Frontend:** the editor, save, read and drilldown paths all use the `all` value directly — no empty-string sentinel. The field-keys/values fetch (which only accepts a concrete signal) maps `all` → omitted.

## 2. Panel/dashboard `links`: required and non-nullable

`links` was nullable, so an unset value round-tripped as `null`. Since v2 dashboards haven't shipped, the contract is tightened rather than left nullable:

- `PanelSpec.Links` and `DashboardSpec.Links` are now **required and non-nullable**; a missing value is rejected (validated on both write and read) instead of being silently defaulted.
- The generated client makes `links` a non-optional `Link[]`, so every spec builder sends `links: []` and the null guards were removed.

## Testing

- Go unit tests (`pkg/types/dashboardtypes/...`, `pkg/modules/dashboard/...`) updated and passing.
- Frontend `tsgo` / `oxlint` clean; dashboard-v2 Variables + drilldown Jest suites passing.
- v2 dashboard integration tests updated to send the required fields; these need the Docker stack, so CI is the check.

#### Tickets closed by this PR

Closes https://github.com/SigNoz/pulse-pod/issues/135
Closes https://github.com/SigNoz/pulse-pod/issues/136

### DIFF
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -2858,6 +2858,7 @@ components:
           $ref: '#/components/schemas/DashboardtypesDynamicVariableSignal'
       required:
       - name
+      - signal
       type: object
     DashboardtypesFillMode:
       enum:
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -2843,12 +2843,19 @@ components:
       required:
       - name
       type: object
+    DashboardtypesDynamicVariableSignal:
+      enum:
+      - traces
+      - logs
+      - metrics
+      - all
+      type: string
     DashboardtypesDynamicVariableSpec:
       properties:
         name:
           type: string
         signal:
-          $ref: '#/components/schemas/TelemetrytypesSignal'
+          $ref: '#/components/schemas/DashboardtypesDynamicVariableSignal'
       required:
       - name
       type: object
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -2748,7 +2748,6 @@ components:
         links:
           items:
             $ref: '#/components/schemas/DashboardtypesLink'
-          nullable: true
           type: array
         panels:
           additionalProperties:
@@ -2765,6 +2764,7 @@ components:
       - variables
       - panels
       - layouts
+      - links
       type: object
     DashboardtypesDashboardView:
       properties:
@@ -3393,7 +3393,6 @@ components:
         links:
           items:
             $ref: '#/components/schemas/DashboardtypesLink'
-          nullable: true
           type: array
         plugin:
           $ref: '#/components/schemas/DashboardtypesPanelPlugin'
@@ -3405,6 +3404,7 @@ components:
       - display
       - plugin
       - queries
+      - links
       type: object
     DashboardtypesPatchOp:
       enum:
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableAdapters.ts
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableAdapters.ts
@@ -13,11 +13,7 @@ import type {
 } from 'api/generated/services/sigNoz.schemas';
 
 import {
-	DYNAMIC_SIGNAL_ALL,
-	DYNAMIC_SIGNALS,
-	type DynamicSignalOption,
 	emptyVariableFormModel,
-	signalForApi,
 	VARIABLE_SORT_DISABLED,
 	type VariableFormModel,
 } from './variableFormModel';
@@ -69,12 +65,8 @@ export function dtoToFormModel(
 			...listCommon,
 			type: 'DYNAMIC',
 			dynamicAttribute: plugin.spec.name ?? '',
-			// Unrecognized/empty signal → "all telemetry", so the source always shows.
-			dynamicSignal: DYNAMIC_SIGNALS.includes(
-				plugin.spec.signal as DynamicSignalOption,
-			)
-				? (plugin.spec.signal as DynamicSignalOption)
-				: DYNAMIC_SIGNAL_ALL,
+			// signal is a required wire field (`all` = every telemetry signal), used as-is.
+			dynamicSignal: plugin.spec.signal,
 		};
 	}
 	// Default to Query (also covers a query plugin or a missing/unknown plugin).
@@ -102,7 +94,7 @@ function buildPlugin(
 				kind: DynamicPluginKind['signoz/DynamicVariable'],
 				spec: {
 					name: model.dynamicAttribute,
-					signal: signalForApi(model.dynamicSignal),
+					signal: model.dynamicSignal,
 				},
 			};
 		case 'QUERY':
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableFormModel.ts
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableFormModel.ts
@@ -1,6 +1,6 @@
 import {
+	DashboardtypesDynamicVariableSignalDTO,
 	DashboardtypesListVariableSpecSortDTO,
-	TelemetrytypesSignalDTO,
 } from 'api/generated/services/sigNoz.schemas';
 import type { DashboardtypesVariableDefaultValueDTO } from 'api/generated/services/sigNoz.schemas';
 import { sortBy } from 'lodash-es';
@@ -22,23 +22,6 @@ export const VARIABLE_TYPE_EVENT_LABEL: Record<VariableType, string> = {
 	TEXT: 'textbox',
 	DYNAMIC: 'dynamic',
 };
-
-/** Telemetry signal — the generated enum (traces / logs / metrics). */
-// A query/variable signal is only logs/traces/metrics. TelemetrytypesSignalDTO
-// also carries the empty "any" value used on field keys, which is not a valid
-// query/variable signal, so exclude it here.
-export type TelemetrySignal =
-	| TelemetrytypesSignalDTO.logs
-	| TelemetrytypesSignalDTO.traces
-	| TelemetrytypesSignalDTO.metrics;
-
-/**
- * Signal selected in the dynamic-variable editor. `'all'` is UI-only (the
- * generated `TelemetrytypesSignalDTO` has no "all") — it searches across every
- * signal and maps to an omitted `signal` on the wire (see {@link signalForApi}).
- */
-export const DYNAMIC_SIGNAL_ALL = 'all' as const;
-export type DynamicSignalOption = TelemetrySignal | typeof DYNAMIC_SIGNAL_ALL;
 
 /**
  * Sort order for list-variable values, keyed by the generated wire enum so the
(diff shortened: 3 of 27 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/hooks/useDrilldownDashboardVariables.tsx
```diff
@@ -105,8 +105,7 @@ export function useDrilldownDashboardVariables({
 				type: 'DYNAMIC',
 				multiSelect: true,
 				dynamicAttribute: fieldName,
-				// `||` (not `??`): an empty "any" signal maps to All, same as unset.
-				dynamicSignal: signal || DYNAMIC_SIGNAL_ALL,
+				dynamicSignal: signal ?? DYNAMIC_SIGNAL_ALL,
```
AshwinBhatkal: why not remove DYNAMIC_SIGNAL_ALL

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableFormModel.ts
```diff
@@ -73,24 +65,31 @@ export const VARIABLE_SORT_LABEL: Record<VariableSort, string> = {
...
-/** Maps the editor's signal selection to the wire value (`'all'` → omitted). */
+/**
+ * Field-keys/values API param. The empty "any" signal is omitted (that endpoint
+ * only accepts a concrete signal), everything else passes through.
+ */
 export function signalForApi(
-	signal: DynamicSignalOption,
-): TelemetrySignal | undefined {
+	signal: TelemetrySignal,
+):
```
AshwinBhatkal, line 90: I want this to be only TelemetrytypesSignalDTO 
Handle for signal === '' instead of ALL

AshwinBhatkal: Okay ALL is already '' but I hope you get the simplification. You may have to fix the consumption chain

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableAdapters.ts
```diff
@@ -69,11 +68,11 @@ export function dtoToFormModel(
 			...listCommon,
 			type: 'DYNAMIC',
 			dynamicAttribute: plugin.spec.name ?? '',
-			// Unrecognized/empty signal → "all telemetry", so the source always shows.
+			// Unrecognized/missing signal → "all telemetry", so the source always shows.
 			dynamicSignal: DYNAMIC_SIGNALS.includes(
-				plugin.spec.signal as DynamicSignalOption,
+				plugin.spec.signal as TelemetrySignal,
```
AshwinBhatkal: avoid this case if possible

docs/api/openapi.yml
```diff
@@ -2747,7 +2747,6 @@ components:
         links:
           items:
             $ref: '#/components/schemas/DashboardtypesLink'
-          nullable: true
```
AshwinBhatkal, line 2751: should you not be having required true as well?

AshwinBhatkal (commented):
Self-review notes on the non-obvious decisions, to save reviewers a round-trip. Both changes are about typed clients (Terraform/SDK) reading back exactly what they sent.

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableFormModel.ts
```diff
@@ -15,22 +15,14 @@ import { sortBy } from 'lodash-es';
...
 /**
- * Signal selected in the dynamic-variable editor. `'all'` is UI-only (the
- * generated `TelemetrytypesSignalDTO` has no "all") — it searches across every
- * signal and maps to an omitted `signal` on the wire (see {@link signalForApi}).
+ * The empty signal (`''`) is the "all telemetry" value — it searches across every
+ * signal and round-trips to the wire as-is (no longer omitted).
  */
-export const DYNAMIC_SIGNAL_ALL = 'all' as const;
-export type DynamicSignalOption = TelemetrySignal | typeof DYNAMIC_SIGNAL_ALL;
+export const DYNAMIC_SIGNAL_ALL = TelemetrytypesSignalDTO[''];
```
AshwinBhatkal: "All telemetry" is now the generated enum's empty member (`''`), not a separate UI-only `'all'` sentinel. `''` is the real wire value that round-trips, so the editor, save, read and drilldown all use it directly — no more `'all'` → `undefined` mapping. This is the alternative to narrowing `TelemetrySignal` into a union.

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableFormModel.ts
```diff
@@ -73,24 +65,31 @@ export const VARIABLE_SORT_LABEL: Record<VariableSort, string> = {
...
 	[TelemetrytypesSignalDTO.logs]: 'Logs',
 	[TelemetrytypesSignalDTO.metrics]: 'Metrics',
 };
 
-/** Maps the editor's signal selection to the wire value (`'all'` → omitted). */
+/**
+ * Field-keys/values API param. The empty "any" signal is omitted (that endpoint
+ * only accepts a concrete signal), everything else passes through.
+ */
 export function signalForApi(
```
AshwinBhatkal, line 88: `signalForApi` is now purely the field-keys/values *transport* mapper. That endpoint only accepts a concrete signal (`traces|logs|metrics`), so the empty "any" signal is omitted here. This is separate from the variable spec, which persists `''` verbatim (see the save path in variableAdapters.ts).

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableAdapters.ts
```diff
@@ -102,7 +101,7 @@ function buildPlugin(
 				kind: DynamicPluginKind['signoz/DynamicVariable'],
 				spec: {
 					name: model.dynamicAttribute,
-					signal: signalForApi(model.dynamicSignal),
+					signal: model.dynamicSignal,
```
AshwinBhatkal, line 97: Save path now persists the actual signal value (including `''` for "all") instead of `undefined`, so create → GET round-trips faithfully.

pkg/types/dashboardtypes/perses_dashboard_data.go
```diff
@@ -45,6 +45,24 @@ func (d *DashboardSpec) UnmarshalJSON(data []byte) error {
 	return d.Validate()
 }
 
+// validateLinksPresent enforces the required, non-nullable links field on the
+// create/update request paths: a missing/null links value (spec- or panel-level)
+// is rejected rather than silently defaulted, so a typed client's value round-trips
+// faithfully. Deliberately not part of Validate() — reads, clones and patches of
+// already-stored specs must not re-reject, and Validate() covers only cross-field
+// semantic checks (never required-field presence, matching its sibling slices).
+func (d *DashboardSpec) validateLinksPresent() error {
```
AshwinBhatkal: Deliberately enforced here and called only from `Postable`/`UpdatableDashboardV2.Validate` — NOT from the shared `DashboardSpec.Validate`. Reads, clones and the semantic-validation tests must not re-reject an already-stored spec, and `Validate` covers only cross-field checks (no sibling required slice is presence-enforced there). Rejecting a missing value (rather than silently defaulting it to `[]`) is intentional: silent defaulting is the same class of issue this round-trip work is fixing.

pkg/types/dashboardtypes/perses_dashboard.go
```diff
@@ -239,6 +239,9 @@ func (p *PostableDashboardV2) Validate() error {
 	if err := validateDashboardTags(p.Tags); err != nil {
 		return err
 	}
+	if err := p.Spec.validateLinksPresent(); err != nil {
```
AshwinBhatkal: Request-boundary enforcement of the required `links` field on create. The same call is added to `UpdatableDashboardV2.Validate` so update/patch are covered too.

tests/integration/tests/dashboard/03_v2_dashboard.py
```diff
@@ -1795,11 +1812,9 @@ def threshold(panel_id: str) -> dict:
...
             assert key not in spec, description
 
-        # A panel with no links comes back with no links value (null or absent);
-        # either is fine for a typed client (an unset optional attribute stays
-        # unset), so this is not a drift. The explicit [] case above is what the
-        # fix guarantees round-trips.
-        assert panels["timeseries"]["spec"].get("links") is None, "unset panel links stays unset"
+        # links is a required, non-nullable field: an explicit [] round-trips as [],
+        # so a typed client always reads a concrete array (never null or absent).
+        assert panels["timeseries"]["spec"]["links"] == [], "panel links round-trip as []"
```
AshwinBhatkal, line 1924: Heads-up for review: these v2 integration tests need the Docker stack, so I couldn't run them locally — CI is the real check. Every success-path v6 create/update in `03` and `05` now sends `links: []`; rejection tests are untouched because they fail on their intended reason (name/tags/decode/semantic/id) before the request-boundary links check.

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelsAndSectionsLayout/Panel/hooks/useDrilldownDashboardVariables.tsx
AshwinBhatkal: Kept `DYNAMIC_SIGNAL_ALL`. `signal` is optional here only because this hook is called even when there is no active drilldown (no variable gets created in that case) — at create-time it is always set. So `signal ?? DYNAMIC_SIGNAL_ALL` just defaults the never-hit empty case; it is not an "is it all?" branch. Left as-is.

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableFormModel.ts
AshwinBhatkal: `DYNAMIC_SIGNAL_ALL` is `TelemetrytypesSignalDTO['']`, so `=== DYNAMIC_SIGNAL_ALL` already is `=== ""`. `signalForApi`'s return stays narrowed to `traces|logs|metrics|undefined` because the shared `useGetFieldKeys`/`useGetFieldValues` hooks only accept a concrete signal — widening them touches V1 usage, so left out of scope.

AshwinBhatkal: Done in 1581bfd — made the dynamic-variable `signal` a required, non-nullable field (BE tag + regenerated OpenAPI/client), so the consumption chain simplifies: the read is now `dynamicSignal: plugin.spec.signal` and `""` flows through by default. Kept `DYNAMIC_SIGNAL_ALL` for the dropdown list/label + the V1 "all sources" mapping.

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableAdapters.ts
AshwinBhatkal: Done in 1581bfd — dropped the `DYNAMIC_SIGNALS.includes(...)` fallback. `signal` is now a required wire field, so it is read as-is: `dynamicSignal: plugin.spec.signal`.

docs/api/openapi.yml
AshwinBhatkal: It is required — OpenAPI 3 declares it in the schema-level `required:` array (`DashboardtypesDashboardSpec.required` and `DashboardtypesPanelSpec.required` both now list `links`), not as a per-property `required: true`.

pkg/types/dashboardtypes/perses_dashboard_data.go
namanv3: not doing this

frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/PanelTypeSwitcher/utils.ts
```diff
@@ -17,8 +17,6 @@ const SIGNAL_LABEL: Record<TelemetrytypesSignalDTO, string> = {
 	[TelemetrytypesSignalDTO.logs]: 'logs',
 	[TelemetrytypesSignalDTO.traces]: 'traces',
 	[TelemetrytypesSignalDTO.metrics]: 'metrics',
-	// The empty "any" signal only appears on field keys, never on a panel query;
-	// mapped for exhaustiveness.
 	[TelemetrytypesSignalDTO['']]: '',
```
AshwinBhatkal, line 20: why not all here?

frontend/src/pages/DashboardPageV2/DashboardContainer/DashboardSettings/Variables/variableFormModel.ts
```diff
@@ -81,27 +74,53 @@ export const VARIABLE_SORT_LABEL: Record<VariableSort, string> = {
...
-	signal: DynamicSignalOption,
-): TelemetrySignal | undefined {
+	signal: TelemetrySignal,
+):
+	| DashboardtypesDynamicVariableSignalDTO.traces
+	| DashboardtypesDynamicVariableSignalDTO.logs
+	| DashboardtypesDynamicVariableSignalDTO.metrics
+	| undefined {
 	return signal === DYNAMIC_SIGNAL_ALL ? undefined : signal;
```
AshwinBhatkal: the consumer of this function can be changed to read all instead of undefined

ahrefabhi (dismissed):

ahrefabhi (approved):

therealpandey (approved):

PR #12205, opened 2026-07-21
fix: fetch latest monotonicty for metrics instead of any random

namanv3:
## Pull Request

---

### 📄 Summary

Latest monotonicity of any metric is a better look at this field that picking any random one. Changes to this field are rare, and if a change is there, we should pick up the latest value to make such changes visible.

Currently, GCP metrics are not setting this field as true where they should, and we'll be putting up an issue upstream to fix same. SigNoz needs to be ready to accept the changed value of the flag once the fix is there upstream.

#### Issues closed by this PR

Closes https://github.com/SigNoz/pulse-pod/issues/115

---

### DIFF
--- a/pkg/modules/metricsexplorer/implmetricsexplorer/module.go
+++ b/pkg/modules/metricsexplorer/implmetricsexplorer/module.go
@@ -87,7 +87,7 @@ func (m *module) listMeterMetrics(ctx context.Context, params *metricsexplorerty
 		"any(type) AS metric_type",
 		"any(unit) AS metric_unit",
 		"argMax(temporality, unix_milli) AS temporality",
-		"any(is_monotonic) AS is_monotonic",
+		"argMax(is_monotonic, unix_milli) AS is_monotonic",
 	)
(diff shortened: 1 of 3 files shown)


pkg/modules/metricsexplorer/implmetricsexplorer/module.go
```diff
@@ -87,7 +87,7 @@ func (m *module) listMeterMetrics(ctx context.Context, params *metricsexplorerty
 		"any(type) AS metric_type",
 		"any(unit) AS metric_unit",
 		"argMax(temporality, unix_milli) AS temporality",
-		"any(is_monotonic) AS is_monotonic",
+		"argMax(is_monotonic, unix_milli) AS is_monotonic",
```
therealpandey, line 90: @namanv3 one integration test for this exact behaviour please.

therealpandey (changes requested):

pkg/modules/metricsexplorer/implmetricsexplorer/module.go
namanv3: done

srikanthccv (approved):

therealpandey (approved):

PR #12207, opened 2026-07-21
fix: show the backend error message in the error modal instead of a generic axios error

AshwinBhatkal:
## Summary

The error modal was showing a generic `Request failed with status code 400` / network error instead of the backend's actual message (e.g. `tag key "…" contains disallowed characters`) — across the V2 dashboards and anywhere else the modal is used.

**Root cause:** the generated API client's reject interceptor (`src/api/index.ts` `interceptorRejected`) rejects with the raw `AxiosError` — it never converts to `APIError`. But nearly every caller does `showErrorModal(error as APIError)` (a false cast), so `ErrorModal` couldn't read `error.getErrorMessage()` and fell back to the generic message. Only `useJsonEditor` did it right (`toAPIError(error)`).

**Fix:** normalize once, centrally, in `ErrorModalProvider.showErrorModal` — accept `unknown`, pass an `APIError` through untouched, and run anything else through the existing `convertToApiError` (which extracts `response.data.error.{code,message,errors}`). Every existing `showErrorModal(error as APIError)` call site is now correct without change; the casts are simply redundant.

This is intentionally a **central** fix, so it corrects every `showErrorModal` caller app-wide, not just the dashboards where it was reported.

---

### ✅ Change Type

- [x] 🐛 Bug fix

---

#### Root Cause
`interceptorRejected` = `Promise.reject(value)` → callers receive a raw `AxiosError`; `showErrorModal(error as APIError)` then hands the modal a non-`APIError`, so the structured `error.message` is lost.

#### Fix Strategy
Normalize `unknown → APIError` inside `showErrorModal` via `convertToApiError`, so the modal always renders the backend message.

---

### 🧪 Testing Strategy

- New `ErrorModalProvider.test.tsx`: a raw AxiosError surfaces its `response.data.error.message`; an already-built `APIError` passes through unchanged. Both pass.
- tsgo clean app-wide; oxlint/oxfmt clean.

---

- Blast radius: `ErrorModalProvider` is shared app-wide. The change only *improves* the displayed message; `APIError` inputs are unchanged (instanceof passthrough), and non-Axios/Error inputs fall back gracefully via `convertToApiError`.
- Potential regressions: none expected — no caller changes required.
- Rollback plan: revert the single commit.

### DIFF
--- a/frontend/src/providers/ErrorModalProvider.test.tsx
+++ b/frontend/src/providers/ErrorModalProvider.test.tsx
@@ -1,0 +1,62 @@
+import { render, screen } from '@testing-library/react';
+import { useEffect } from 'react';
+import APIError from 'types/api/error';
+
+import { ErrorModalProvider, useErrorModal } from './ErrorModalProvider';
+
+// Mock the heavy modal so the test only asserts which error it receives.
+jest.mock('components/ErrorModal/ErrorModal', () => ({
+	__esModule: true,
+	default: ({ error }: { error: APIError }): JSX.Element => (
+		<div data-testid="error-message">{error.getErrorMessage()}</div>
+	),
+}));
+
+function Trigger({ error }: { error: unknown }): null {
+	const { showErrorModal } = useErrorModal();
+	useEffect(() => {
+		showErrorModal(error);
+	}, [error, showErrorModal]);
+	return null;
+}
+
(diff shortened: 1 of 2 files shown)


AshwinBhatkal:
Superseded — the central error-modal fix moved into #12062 (alongside the tag inline-edit fix).

PR #10941, opened 2026-04-15
refactor(alertmanager): move API handlers to signozapiserver

therealpandey:
## Summary

- Extract `Handler` interface in `pkg/alertmanager/handler.go` and move implementation to `pkg/alertmanager/signozalertmanager/handler.go`
- Register all alertmanager routes (channels, route policies, alerts) in `signozapiserver` via `handler.New()` with `OpenAPIDef`
- Remove `AlertmanagerAPI` injection from `http_handler.go` and `server.go`
- Add `required`/`nullable`/`Enum()` tags to alertmanager types for accurate OpenAPI schemas
- Add `/api/v1/channels/test` endpoint, mark `/api/v1/testChannel` as deprecated
- Regenerate OpenAPI spec and frontend API clients

## Why?

- **Deprecate legacy API docs**: moves alertmanager routes out of `http_handler.go` into the typed `signozapiserver` with full `OpenAPIDef`, bringing us closer to fully deprecating https://signoz.io/api-reference/#/ — only rules and dashboards remain on the legacy path
- **Unblock Terraform provider**: with typed OpenAPI definitions and proper schemas (`required`, `nullable`, `Enum()`), the Terraform provider can start generating resources for channels and route policies from the spec
- **Unblock audit logs epic**: routes registered via `handler.New()` support `WithAuditDef()` injection — this is a prerequisite for instrumenting notification channel and route policy mutations with audit events (SigNoz/platform-pod#1938)

## Test plan

- [ ] `go build ./cmd/... ./pkg/... ./ee/...` passes
- [ ] All alertmanager routes respond correctly (channels CRUD, route policies CRUD, alerts GET, channel test)
- [ ] Auth levels preserved (ViewAccess/EditAccess/AdminAccess per route)
- [ ] `/api/v1/channels/test` works, `/api/v1/testChannel` still works (deprecated)
- [ ] OpenAPI spec includes alertmanager route definitions with accurate schemas

closes SigNoz/platform-pod#2087

PR #12208, opened 2026-07-21
feat(savedview): register v2 routes, migrate to v5-only query type

krsoninikhil:
## Summary

- Registers saved-view CRUD (`Create`/`Update`/`Delete`/`Get`/`List`) at `/api/v2/saved_views(/{viewId})` via `handler.New()` + `OpenAPIDef`, mirroring the alertmanager migration (#10941). Closes SigNoz/engineering-pod#4651.
- `/api/v1/explorer/views` is unchanged and shares the same handler — verified backward compatible with the existing frontend.
- Saved views now persist v5 query-builder types (`compositeQuery.queries`) instead of the legacy v3 `builderQueries`/`chQueries`/`promQueries` shape.
- Split the shared type into `PostableSavedView` (request), `GettableSavedView` (response), `StorableSavedView` (DB row) — clients no longer need to send server-only fields (id, timestamps).
- Added local `SourcePage`/`PanelType`/`QueryType` enums with validation, replacing the `pkg/query-service/model/v3` dependency.
- Dropped unused `category`/`tags` from the request/response (never meaningfully set or read by the frontend); DB columns kept as-is (marked deprecated) to avoid a migration in this PR.

## Test plan

- [x] `go build ./...`, `golangci-lint` (0 issues)
- [x] Live smoke test of all `/api/v2/saved_views` and `/api/v1/explorer/views` endpoints (create/get/update/list/delete)
- [x] OpenAPI spec + frontend generated client regenerated and committed


pkg/modules/savedview/implsavedview/handler.go
```diff
@@ -112,7 +112,13 @@ func (handler *handler) Update(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	render.Success(w, http.StatusOK, view)
+	updated, err := handler.module.GetView(ctx, claims.OrgID, viewUUID)
+	if err != nil {
+		render.Error(w, err)
+		return
+	}
```
krsoninikhil: Remove this.

pkg/types/savedviewtypes/savedview.go
```diff
@@ -2,12 +2,80 @@ package savedviewtypes
...
+// clients never supply those, the server assigns them.
+type PostableSavedView struct {
+	Name           string          `json:"name" required:"true"`
+	Category       string          `json:"category" required:"true"`
+	SourcePage     string          `json:"sourcePage" required:"true"`
+	Tags           []string        `json:"tags" required:"true" nullable:"true"`
+	CompositeQuery *CompositeQuery `json:"compositeQuery" required:"true"`
+	// ExtraData is JSON encoded data used by frontend to store additional data
+	ExtraData string `json:"extraData" required:"true"`
+}
```
krsoninikhil, line 67: Had to separate these out instead of using same `SavedView` otherwise openapi specs would have shown ID and other internal fields in the create request payload.

PR #12209, opened 2026-07-21
fix(api): guard ErrorResponseHandlerV2 against non-envelope error bodies

AshwinBhatkal:
### 📄 Summary

`ErrorResponseHandlerV2` assumed every non-2xx response body was the SigNoz V2 error envelope (`{ error: { code, message, url, errors } }`) and read `response.data.error.code` directly. That assumption holds for errors returned by the app server, but not for errors returned by the infrastructure in front of it. During a deployment the gateway (nginx / load balancer) terminates the request and returns a 5xx with an HTML or empty body — `response.data.error` is then `undefined`, and reading `.code` threw a `TypeError` *from inside the error handler itself*, masking the real (transient) failure and crashing the caller.

This fixes it at the boundary: the inbound body is typed as `unknown` and narrowed with an `isErrorV2Resp` type guard. When the body is a valid envelope we surface it exactly as before; when it isn't, we synthesize an `APIError` from the HTTP status instead of throwing. The handler can no longer throw on an unexpected body shape, so this covers 502/503/504/empty/HTML bodies uniformly — not just the two statuses seen today.

Fixing it centrally covers every one of the ~70 call sites that route through this handler.

### Screenshots
image

#### Issues closed by this PR

Closes https://github.com/SigNoz/engineering-pod/issues/5760
Closes https://github.com/SigNoz/platform-pod/issues/2502

### DIFF
--- a/frontend/src/api/ErrorResponseHandlerForGeneratedAPIs.ts
+++ b/frontend/src/api/ErrorResponseHandlerForGeneratedAPIs.ts
@@ -62,9 +62,7 @@ export function convertToApiError(
 	return new APIError({
 		httpStatusCode: response?.status || error.status || 500,
 		error: {
-			code:
-				errorData?.code ||
-				String(response?.status || error.code || 'unknown_error'),
+			code: errorData?.code || 'UPSTREAM_UNAVAILABLE',
 			message:
 				errorData?.message ||
 				response?.statusText ||
--- a/frontend/src/api/ErrorResponseHandlerForGeneratedAPIs.ts
+++ b/frontend/src/api/ErrorResponseHandlerForGeneratedAPIs.ts
@@ -2,6 +2,29 @@ import { RenderErrorResponseDTO } from 'api/generated/services/sigNoz.schemas';
 import { AxiosError } from 'axios';
 import APIError from 'types/api/error';
 
+// The wire shape these handlers can actually rely on. The generated
+// RenderErrorResponseDTO marks code/message/url/errors as required, but the
+// server omits any of them even on valid errors (e.g. a 400 with just a
+// message), so a present `error` object is all the guard can guarantee.
+type ErrorEnvelope = {
+	error: {
+		code?: string;
+		message?: string;
+		url?: string;
+		errors?: { message?: string }[];
+	};
+};
+
+function isErrorEnvelope(data: unknown): data is ErrorEnvelope {
+	return (
+		typeof data === 'object' &&
+		data !== null &&
+		'error' in data &&
+		typeof (data as ErrorEnvelope).error === 'object' &&
+		(data as ErrorEnvelope).error !== null
+	);
+}
+
 // @deprecated Use convertToApiError instead
 export function ErrorResponseHandlerForGeneratedAPIs(
 	error: AxiosError<RenderErrorResponseDTO>,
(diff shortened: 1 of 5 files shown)


frontend/src/api/ErrorResponseHandlerV2.ts
```diff
@@ -2,19 +2,39 @@ import { AxiosError } from 'axios';
...
+			});
+		}
 		throw new APIError({
 			httpStatusCode: response.status || 500,
 			error: {
-				code: response.data.error.code,
-				message: response.data.error.message,
-				url: response.data.error.url,
-				errors: response.data.error.errors,
+				code: String(response.status || 500),
```
therealpandey: When we can't find the code, can we throw a `UPSTREAM_UNAVAILABLE` code?

therealpandey (changes requested):

therealpandey:
Fixes https://github.com/SigNoz/platform-pod/issues/2502

therealpandey (dismissed):

frontend/src/api/ErrorResponseHandlerForGeneratedAPIs.ts
H4ad, line 61: this should check for v2 format as well

frontend/src/api/__tests__/ErrorResponseHandlerV2.test.ts
```diff
@@ -0,0 +1,98 @@
...
+						},
+					},
+				} as AxiosError['response'],
+			}),
+		);
+
+		expect(apiError).toBeInstanceOf(APIError);
+		expect(apiError.getHttpStatusCode()).toBe(400);
+		expect(apiError.getErrorCode()).toBe('bad_request');
+		expect(apiError.getErrorMessage()).toBe('Invalid dashboard payload');
```
H4ad: add asserts to error messages under errors, these are important for many parts of the UI

frontend/src/api/__tests__/ErrorResponseHandlerV2.test.ts
```diff
@@ -0,0 +1,98 @@
...
+	try {
+		ErrorResponseHandlerV2(error);
+	} catch (thrown) {
+		return thrown as APIError;
+	}
+	throw new Error('expected ErrorResponseHandlerV2 to throw');
+}
+
+describe('ErrorResponseHandlerV2', () => {
+	it('surfaces the V2 error envelope when the body is well-formed', () => {
```
H4ad: similar to the suggestion in the other pr for light/dark, these can be moved to be an array or "error" and "expected" object, and we can keep adding more error objects in case we find more of them (with different shapes, unlikely)

AshwinBhatkal: https://media0.giphy.com/media/v1.Y2lkPTc5MGI3NjExaDR6bGphcjI4cW5ndnVqYjhjcmVxbDhlaXBoZWhrdnFrZzZsM3NpZCZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/Fg43ZBsPdNnROf99Ys/giphy.gif

H4ad (dismissed):

H4ad (dismissed):

H4ad (approved):

PR #12227, opened 2026-07-22
test(e2e): cover alert edit-page threshold persistence

ahrefabhi:
## Pull Request

---

Adds E2E coverage for the alert **overview/edit** page: opening a saved threshold alert must show the persisted threshold value in the condition editor. This exercises the load path where the query builder serializes the loaded query back into the URL after mount — a spot that has been fragile — so a future change that drops the loaded threshold on the edit page is caught automatically.

The test follows the repo's existing pattern: seed real data via API, then drive the real UI.

- **`tests/e2e/helpers/alerts.ts`** (new) — seeding/navigation helpers mirroring `helpers/dashboards.ts`: create an email notification channel, create a v2 threshold alert (`POST /api/v2/rules`) on `signoz_calls_total`, delete both for cleanup, and `gotoAlertOverview` (open the edit page and wait for it to settle).
- **`tests/e2e/tests/alerts/alerts.spec.ts`** — adds `TC-02 edit page shows the saved threshold value` alongside the existing tabs test. It seeds an alert with target `245`, opens the overview page, and asserts `threshold-value-input` holds `245`.

N/A — test-only.

N/A.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [x] 🧪 Test-only

---

N/A — not a bug fix.

---

- **Tests added:** `TC-02 edit page shows the saved threshold value` in `tests/e2e/tests/alerts/alerts.spec.ts`, plus seeding helpers in `tests/e2e/helpers/alerts.ts`.
- **Manual verification:** Ran against a local `--with-web` stack (pytest bootstrap). The full `alerts.spec.ts` (TC-01 + TC-02) passes. Cross-checked the assertion is meaningful by serving a build where the edit page resets the value — TC-02 then fails with `Expected: "245", Received: "0"`.
- **Static checks:** `oxfmt --check`, `tsc --noEmit`, and `oxlint` all clean for the two files.

---

- **Blast radius:** E2E suite only — no product code touched.
- **Potential regressions:** None to the app. The new test depends on `signoz_calls_total` being present (seeded by the golden-dataset bootstrap) and on the `POST /api/v1/channels` + `POST /api/v2/rules` APIs.
- **Rollback plan:** Revert the PR — it only adds a test and a helper.

---

N/A — internal test-only change.

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- The seed leaves `channels: []` invalid per the API (rule creation requires at least one channel), so the helper seeds an email channel first and references it by name; both are cleaned up in `afterAll`.
- `gotoAlertOverview` waits for the query builder to serialize the query into the URL (`compositeQuery=…`) and then briefly settles so the assertion reads the final value; that single `waitForTimeout` carries an inline lint-suppression with a rationale.
- Behavior verified is exactly what #12226 fixed on the edit page; this PR just locks it in at the E2E level.

### DIFF
--- a/tests/e2e/tests/alerts/alerts.spec.ts
+++ b/tests/e2e/tests/alerts/alerts.spec.ts
@@ -40,9 +40,6 @@ test.describe('alerts — threshold persists on edit-page load', () => {
 	});
 
 	test.afterAll(async ({ browser }) => {
-		if (!ruleId && !channelId) {
-			return;
-		}
 		const ctx = await newAdminContext(browser);
 		const page = await ctx.newPage();
 		try {
--- a/tests/e2e/tests/alerts/alerts.spec.ts
+++ b/tests/e2e/tests/alerts/alerts.spec.ts
@@ -1,7 +1,70 @@
-import { test, expect } from '../../fixtures/auth';
+import { expect, test } from '../../fixtures/auth';
+import {
+	createEmailChannelViaApi,
+	createThresholdAlertViaApi,
+	deleteAlertViaApi,
+	deleteChannelViaApi,
+	gotoAlertOverview,
+} from '../../helpers/alerts';
+import { newAdminContext } from '../../helpers/auth';
 
 test('TC-01 alerts page — tabs render', async ({ authedPage: page }) => {
 	await page.goto('/alerts');
 	await expect(page.getByRole('tab', { name: /alert rules/i })).toBeVisible();
 	await expect(page.getByRole('tab', { name: /configuration/i })).toBeVisible();
+});
+
+test.describe('alerts — threshold persists on edit-page load', () => {
+	const TARGET = 245;
+	let ruleId: string;
+	let channelId: string;
+
+	test.beforeAll(async ({ browser }) => {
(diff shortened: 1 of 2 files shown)


tests/e2e/tests/alerts/alerts.spec.ts
```diff
@@ -1,7 +1,70 @@
...
+				target: TARGET,
+				channels: [channel.name],
+			});
+		} finally {
+			await ctx.close();
+		}
+	});
+
+	test.afterAll(async ({ browser }) => {
+		if (!ruleId && !channelId) {
```
AshwinBhatkal: this if is not required

AshwinBhatkal (approved):

tests/e2e/tests/alerts/alerts.spec.ts
ahrefabhi: Done

PR #12236, opened 2026-07-22
fix: address issue #12186 - [Bug]: Live-log streaming leaks goroutines after clients disconnect

### LINKED ISSUE #12186: [Bug]: Live-log streaming leaks goroutines after clients disconnect
## Bug Description

  The live-log streaming implementation in pkg/querier/querier.go can leak goroutines after an SSE client disconnects.

  QueryRawStream creates a ticker and then starts a second goroutine to forward ticks into a custom channel:

  ticker := time.NewTicker(q.liveDataRefresh)
  defer ticker.Stop()

  tick := make(chan time.Time, 1)
  tick <- time.Now()

  go func() {
        for timestamp := range ticker.C {
                tick <- timestamp
        }
  }()

  The forwarding goroutine ranges over ticker.C. Calling ticker.Stop() stops future ticks, but it does not close ticker.C.

  Therefore, after QueryRawStream returns, the forwarding goroutine remains blocked while waiting for another value from
  ticker.C. Every completed live-tail session can leave one such goroutine behind.

  There are additional blocking shutdown paths in the same stream lifecycle.

  When the request context is cancelled, QueryRawStream sends to the unbuffered client.Done channel:

  case <-ctx.Done():
        done := true
        client.Done <- &done
        return

  If the HTTP handler has already returned—for example, because writing to the SSE connection failed—there may be no receiver
  for client.Done. The producer goroutine then blocks permanently instead of returning.

  Log delivery also uses a direct channel send:

  client.Logs <- data.Rows[i]

  client.Logs has a buffer of 1,000 entries. If the client disconnects and the handler stops consuming logs, the producer can
  fill the buffer and block indefinitely. This send does not select on ctx.Done().

  The error paths have a similar problem because client.Error is unbuffered:

  client.Error <- qbErr

  These behaviors share the same root cause: the live-stream producer and its ticker-forwarding goroutine do not have a
  cancellation-safe lifecycle.

  ## Steps to Reproduce

  ### Reproduce through the live-tail endpoint

  1. Start SigNoz with the backend and ClickHouse running.
  2. Ensure logs are being ingested continuously.
  3. Record the current backend goroutine count using pprof or another runtime profiling mechanism.
  4. Open an authenticated live-tail request:

     curl -N \
       -H "Authorization: Bearer <token>" \
       "http://localhost:8080/api/v3/logs/livetail?start=0&filter="

  5. Allow the stream to remain connected for a short period.
  6. Terminate the client with Ctrl+C.
  7. Repeat the connection and disconnection cycle many times.
  8. Capture another goroutine profile.
  9. Observe goroutines remaining blocked around the ticker-forwarding loop in QueryRawStream.

  A retained goroutine is expected around this code:

  for timestamp := range ticker.C {
        tick <- timestamp
  }

  Because ticker.Stop() does not close ticker.C, the range never terminates.

  ### Reproduce the blocked shutdown send

  1. Start a live-tail request against:

     GET /api/v3/logs/livetail

  2. Disconnect the client while the backend is processing or delivering logs.
  3. Cause the SSE handler to return before the stream producer finishes.
  4. Cancel the request context.
  5. Observe the producer attempt to send on the unbuffered client.Done channel:

     client.Done <- &done

  6. Because the HTTP handler is no longer receiving from that channel, the producer remains blocked.

  ### Reproduce the blocked log-delivery path

  1. Start a live-tail stream with a high volume of incoming logs.
  2. Stop or disconnect the SSE consumer.
  3. Allow the backend producer to continue reading query results.
  4. Once the client.Logs buffer reaches its 1,000-entry capacity, observe the producer block at:

     client.Logs <- data.Rows[i]

  5. Cancelling the request does not unblock this send because it does not select on ctx.Done().

  ## Expected Behavior

  When a live-log client disconnects:

  - The HTTP request context should be cancelled.
  - QueryRawStream should stop querying for new logs.
  - Any pending log or error delivery should unblock.
  - The ticker should be stopped.
  - Every goroutine created for the stream should exit.
  - No stream-specific query state should remain retained.
  - Repeated connect/disconnect cycles should not continuously increase the backend goroutine count.

  All producer-to-handler channel sends should be cancellation-aware.

  A possible delivery pattern is:

  select {
  case client.Logs <- row:
  case <-ctx.Done():
        return
  }

  The implementation should not depend on ticker.C being closed, because time.Ticker.Stop does not close that channel.

  ## Actual Behavior

  Each live-tail stream creates a forwarding goroutine that ranges over ticker.C.

  When the stream ends:

  - ticker.Stop() stops future ticks.
  - ticker.C remains open.
  - The forwarding goroutine remains blocked waiting for a value that will never arrive.

  In addition:

  - Cancellation sends to an unbuffered client.Done channel.
  - Query errors are sent to an unbuffered client.Error channel.
  - Log rows are sent to client.Logs without selecting on cancellation.
  - If the SSE handler has returned, these sends may have no active receiver.
  - A full client.Logs buffer can block the producer even after request cancellation.

  Repeated live-tail sessions can therefore accumulate blocked goroutines and retain associated stream state.

  ## Environment

  - SigNoz edition: Community and Enterprise shared querier path
  - Branch: main
  - Go version specified by the repository: 1.25.7
  - Component: pkg/querier
  - Endpoint: GET /api/v3/logs/livetail
  - Transport: Server-Sent Events
  - Database: ClickHouse
  - OS: Any

  ## Logs/Error Messages

  No explicit error is necessarily emitted when the leak occurs.

  The issue is visible through a goroutine profile or increasing runtime.NumGoroutine() values after repeated live-tail
  connections and disconnections.

  A goroutine profile may show blocked goroutines around:

  github.com/SigNoz/signoz/pkg/querier.(*querier).QueryRawStream.func1

  with the goroutine waiting on:

  for timestamp := range ticker.C

  Other leaked goroutines may be blocked while sending to:

  client.Done
  client.Error
  client.Logs

  ## Additional Context

  Relevant code locations:

  - pkg/querier/querier.go
      - QueryRawStream
      - Ticker creation and forwarding goroutine around line 515
      - Context-cancellation handling around line 529
      - Query-error delivery around line 549
      - Log delivery around line 558

  - pkg/querier/api.go
      - QueryRawStream HTTP handler
      - SSE channel construction around line 206
      - Producer goroutine startup around line 207
      - SSE consumer loop around line 209

  - pkg/query-service/app/http_handler.go
      - Live-tail route registration:

        GET /api/v3/logs/livetail

  No focused tests for QueryRawStream were found.

  A possible fix would be:

  1. Remove the ticker-forwarding goroutine.
  2. Perform the first query directly before entering the ticker loop, or use a one-shot timer for the initial execution.
  3. Select directly on:

     ticker.C
     ctx.Done()

  4. Make all log, error, and completion delivery cancellation-aware.
  5. Avoid requiring a completion-channel send when the request context itself already communicates stream termination.
  6. Add tests that:
      - Start and cancel a raw stream.
      - Verify QueryRawStream returns promptly.
      - Fill the log channel and then cancel the context.
      - Verify cancellation unblocks the producer.
      - Simulate an absent error-channel receiver.
      - Repeat stream startup and cancellation and confirm no goroutines accumulate.
      - Run under the Go race detector.

kunalrawat425:

Closes #12186


therealpandey:
Hi @kunalrawat425,
Only start working on pull requests when a maintainer has triaged and assigned the corresponding issue to you.

kunalrawat425:

PR #12244, opened 2026-07-22
feat(logs): implement search()

tushar-signoz:
Adds a `search('term')` function to the logs filter grammar — a keyless, case-insensitive full-text search that matches the term across every log field at once, so you don't have to name a column.

### What it does
- Fans the match across all searchable columns: log fields (severity, trace/span id), the body (`body` / `body_v2`), and the attribute + resource maps (both keys and values).
- Scoped forms narrow the fan-out: `search('term', body, resource)` searches only the named field contexts (`body` / `attribute` / `resource` / `log`), bare or quoted. An unknown scope, or a qualified field like `body.message`, is rejected with a 400.
- Case-insensitive literal substring match — regex metacharacters in the term are matched literally, not interpreted. The term comes from the raw token, so `search(1000000)` scans for the digits rather than `1e+06`.

### Guardrails
`search()` always runs (no enable flag) but is scan-heavy, so the statement carries a `CostGuard` — an advisory warning plus a per-shard scan-row budget. Before running, the querier estimates the query via `EXPLAIN ESTIMATE` and rejects it if the estimate exceeds the budget, failing closed if the estimate itself can't be produced. On the window-list pagination path the estimate accumulates across the buckets actually visited, so a broad search can't slip through one bucket at a time. `body_v2` gets its own, lower budget (`querier.search_max_scan_rows_json_body`): `toString()` rebuilds every document and no skip index can prune. A rejection leads with the `search()` advisory, then says how to get back under budget — narrow the range, or add a more selective filter.

### Notes
- Logs-only: other signals reject `search()`.
- Builds on the grammar/parse layer (#12129) and the querier config squash (#12385), both already on main.

### Testing
- `filter_expr_search_test.go` — the per-field-context fan-out SQL, scope parsing and rejection, and the scan-heavy flag.
- `search_cost_guard_test.go` — the budget and advisory the statement carries, including the separate `body_v2` budget.
- `querierlogs/15_search.py` and `querier_json_body/05_search.py` — the same matrix end to end over the legacy body and over `body_v2`, plus both cost-guard recoveries (a selective filter, a narrower time range).


pkg/telemetryschema/logstelemetryschema/filter_expr_search_test.go
```diff
@@ -0,0 +1,288 @@
...
+
+// TestFilterExprSearch covers the search('needle') function, which fans out
+// across every searchable column via FilterOperatorSearch.
+func TestFilterExprSearch(t *testing.T) {
+	releaseTime := time.Date(2024, 1, 15, 10, 0, 0, 0, time.UTC)
+	inWindowStart := uint64(releaseTime.Add(-5 * time.Minute).UnixNano())
+	inWindowEnd := uint64(releaseTime.Add(5 * time.Minute).UnixNano())
+
+	legacyBody := "match(LOWER(body), LOWER(?))"
+	jsonBody := "match(LOWER(toString(body_v2)), LOWER(?))"
```
krsoninikhil, line 47: Is the expectation with search to support the regex characters like `.`/`?`/`.*` as well? 

If not, then why can't we replace the match with `LIKE` operator? See the performance issues we get due to all search using the `match` in our QB https://github.com/SigNoz/engineering-pod/issues/5408

tushar-signoz: Design choices were already finalised by the previous champion, Im just taking it to the finish line. 
But indeed a good suggestion, will allocate time for this.

tushar-signoz:
## Review

Reviewed the full diff at `46ad34f`. Verified locally: `go build ./...` clean, and `go test` green across `telemetryschema/logstelemetryschema`, `statementbuilder/...`, `querier`, `variables`, `querybuildertypesv5`, `querybuilder`.

The fan-out shape is solid and the reasoning behind `QuoteMeta` + `LOWER` on both sides checks out — `lower(body)` (ngrambf_v1 + tokenbf_v1) and `lower(toString(body_v2))` (`JSONFullTextIndexExpr`) both exist as skip indexes in the schema migrator, so keeping the match literal instead of `(?i)` is well-founded. Extracting the raw token text so `search(1000000)` doesn't become `1e+06`, the separate lower `body_v2` budget, failing closed when the estimate can't be produced, and dropping `search` from the resource-fingerprint subquery (safe superset) are all right. The integration tests earning both recoveries the advisory suggests are a nice touch.

Findings below, most significant first.

---

### 1. Non-logs scope words silently widen to the full fan-out instead of being rejected

`FieldContextFromText` accepts every key in `fieldContexts` — which includes `scope`, `span`, `spanfield`, `event`, `metric`, `tracefield` — but `searchColumns` only handles log/body/attribute/resource. Everything else falls through `default:`, which is the **full keyless fan-out**.

Confirmed by rendering the SQL: `search('error', scope)`, `search('error', span)`, `search('error', event)`, `search('error', metric)` and `search('error', tracefield)` each produce byte-identical SQL to bare `search('error')`.

So a user *narrowing* the scope gets the widest and most expensive query available, and in the `scope` case it doesn't even search the scope columns. `scope` is a legitimate logs context with real columns behind it. The invalid-scope test only covers `bogus` and `body.message`, so this slips through.

Suggest whitelisting the four supported contexts in `VisitSearchCall` and rejecting the rest with the existing `invalid search scope` error, and making `searchColumns`'s `default` mean "unspecified only" rather than a catch-all.

### 2. The keyless fan-out isn't actually "every log field"

Not covered: `scope_name`, `scope_version`, `scope_string`, `severity_number`, `id`. `scope_name` even carries its own `tokenbf_v1(10240, 3, 0)` index, so it's cheap to include. The PR body says search "matches the needle across every log field at once" and `SearchWarning` says "runs across all fields" — either add them or state the exclusions.

### 3. Whether the guard trips depends on the ORDER BY shape

`isWindowList()` requires `RequestTypeRaw` plus `timestamp,id` in the same direction for logs. That splits `search()` into two very different cost regimes:

- `ORDER BY timestamp DESC, id DESC` → windowed path, estimates bucket-by-bucket and stops once the page fills → usually passes.
- `ORDER BY timestamp DESC` alone → single-shot `enforceEstimate` over the whole range → usually rejected.

Same needle, same window, verdict flipped by a tie-breaker key. Worth deciding whether that's intended; at minimum the budget's meaning ("per-shard scan rows for this query") isn't stable across equivalent queries.

Related: `RequestTypeRawStream` never takes the windowed path, so `search()` in live tail estimates the whole `[start, end]` in one shot.

### 4. The windowed cumulative-reject branch has no test

Both cost-guard integration tests pass `order=[build_order_by("timestamp", "desc")]` — a single key — so for logs `isWindowList()` is false and they only exercise `enforceEstimate`. The accumulate-and-reject branch in `executeWindowList` (`builder_query.go:576-587`) is the trickier of the two and the one the logs explorer actually hits (it sends timestamp+id), and no test rejects through it.

Also worth thinking about: because `estimatedScan` is per-request, a cursor-paginated scroll can serve several pages fine and then 400 mid-scroll once the cursor lands in a sparse region that needs more buckets to fill a page.

### 5. The JSON-body budget applies even when no body column is scanned

`costGuardFor` keys only off the `use_json_body` flag, not off the columns the fan-out actually produced. On a JSON-body tenant, `search('x', resource)` and `search('x', log)` get the 6M budget instead of 60M even though `toString(body_v2)` never appears in the SQL. Deriving the budget from the emitted columns (or from the resolved scopes) would match the stated rationale — "each row costs far more" is only true when `body_v2` is in the fan-out.

### 6. A transient ClickHouse failure is reported as a 400

`estimateRows` fails closed with `CodeInvalidInput` for *any* `Estimate` error, so a momentary CH blip tells the user "Could not estimate this query's scan cost" as a client error. Failing closed is the right call; the code class arguably isn't.

Minor detail in the same block: the `DeadlineExceeded` branch is the one that *doesn't* log, and a 5s planning timeout seems like the more interesting signal of the two.

### 7. The guard's own cost

One extra `EXPLAIN ESTIMATE` round trip per bucket on the windowed path, each with its own 5s timeout — up to 2x the round trips for a wide window. Also `estimateTimeout` is hardcoded while the row budgets are configurable; if the budgets are tunable per deployment, the timeout probably should be too.

### 8. Commit message references a flag that doesn't exist

The commit body says search is "gated by the `allow_logs_search` feature flag". There is no such flag anywhere in the tree — `pkg/flagger/registry.go` has no entry and nothing references it. The PR body is the accurate one ("always runs, no enable flag"). Worth fixing before merge, since the commit message is what lands.

### 9. Moving `statementbuilder` under `querier` is a breaking config change

`SIGNOZ_STATEMENTBUILDER_SKIP__RESOURCE__FINGERPRINT_{ENABLED,THRESHOLD}` silently stops being read and becomes `SIGNOZ_QUERIER_SKIP__RESOURCE__FINGERPRINT_*`. That knob ships on main. Nothing inside the repo still references the old namespace (checked), but charts and deployment configs outside it will — this needs a release note, and the removal of the top-level `statementbuilder:` YAML section deserves a mention too.

Separately: `yaml:",squash"` isn't a yaml.v3 flag — the equivalent is `,inline`. Harmless while nothing yaml-marshals `Config`, but it reads as if it works.

### 10. Nits

- `NewLogQueryStatementBuilder`: `b := &logQueryStatementBuilder{...}` followed immediately by `return b` — the intermediate can go.
- `ConditionFor`: the search early-return sits *after* `matches := querybuilder.MatchingFieldKeys(...)`, so that work is done and thrown away. The comment says "handle it before key resolution" — moving it above `matches` would make that true.
- `filter_expr_search_test.go`: the `expectWarning` field actually asserts `clause.RequiresCostGuard`, not a warning. Rename to match.
- `search('')` renders `match(LOWER(col), LOWER(''))`, which matches every row and then just burns the budget. Probably worth rejecting an empty needle up front.
- A multi-value dashboard variable in `search($var)` expands to `search(['a', 'b'])`, which the new `valueList` grammar can't parse (the old `functionParamList` accepted arrays). Not a behavior regression since `search()` never worked before, but the UI shouldn't offer multi-select variables there.
- `trimQuotes` doesn't unescape `\'`, so `search('it\'s')` searches for a literal backslash. Pre-existing and consistent with the rest of the codebase, but it now composes with `regexp.QuoteMeta`, which escapes that backslash again.

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -776,11 +779,77 @@ func normalizeFunctionValue(operator qbtypes.FilterOperator, functionName string
...
+	}
+	params := valueList.AllValue()
+	if len(params) == 0 {
+		v.errors = append(v.errors, "function `search` expects a needle, e.g. search('error')")
+		return ErrorConditionLiteral
+	}
+
+	searchText, ok := searchParamText(params[0])
+	if !ok {
+		v.errors = append(v.errors, "function `search` expects a needle as its first argument, e.g. search('error')")
```
therealpandey: What's a needle?

therealpandey (changes requested):

pkg/querybuilder/where_clause_visitor.go
tushar-signoz: Its needle (searched term)  as in needle and haystack, let me remove it if its creating confusion.

therealpandey (dismissed):

therealpandey (approved):

PR #12246, opened 2026-07-23
fix(dashboard): filter-quote escaping, panel export encoding, right-legend width, number-panel query warning

### LINKED ISSUE #9512: Show warning when number panel has more than one query enabled
It has been observed that users create a formula query for the number panel but forget to disable the other queries. This results in showing the value of A, which confuses.

ahrefabhi:
## Pull Request

---

### 📄 Summary

Four small, independent changes in the V2 dashboard / query-builder area, one per commit:

1. **`fix(qb)`** — Filter values with an escaped single quote gained an extra backslash on every `expression ⇄ filters` round-trip, breaking **Create alert from panel**.
2. **`fix(dashboard)`** — The explorer→V2-dashboard panel export link single-encoded `compositeQuery`, so a `%`/`+` literal in a filter dropped the whole query on read.
3. **`fix(dashboard)`** — A right-positioned pie/donut legend was capped too narrowly and ellipsized long series names.
4. **`feat(dashboard)`** — A Number panel with more than one enabled query silently shows only the first query's value; now it surfaces a warning (in the editor and on the rendered panel) prompting the user to disable the extras.

Each is committed separately so they can be reviewed in isolation.

Fix 1 (before): `name = 'it\'s a ribbon'` in a panel became `name = 'it\\'s a ribbon'` on the alert page and failed with `token recognition error at: '''`.

#### Issues closed by this PR

Closes https://github.com/SigNoz/pulse-pod/issues/137
Closes https://github.com/SigNoz/signoz/issues/9512

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause

**1. Escaped-quote doubling.** `formatSingleValue` escapes `'` → `\'` when it serializes a filter value into an expression, but its inverse (`unquote`, used by `formatSingleValueForFilter`) only stripped the surrounding quotes and never unescaped. So each `expression → filters → expression` pass added one backslash. Creating an alert from a panel re-reads the query through the URL twice (`useGetCompositeQueryParam` populates `filters.items` with the lossy value on the first read; a subsequent `redirectWithQueryBuilderData` writes it back and the second read re-escapes), turning `name = 'it\'s a ribbon'` into `name = 'it\\'s a ribbon'`.

**2. Single-encoded compositeQuery.** `useGetCompositeQueryParam` decodes the `compositeQuery` param twice (once via `URLSearchParams.get()`, once via `decodeURIComponent`). `buildExportPanelLink` encoded only once, so a bare `%`/`+` from a filter expression (e.g. `ILIKE 'Inf%'`) made the reader's second decode throw and drop the query.

**3. Right-legend width.** The right legend reused the bottom-legend width cap (`min(MAX_LEGEND_WIDTH, 30% width)`), which is tuned for row-packing and left a vertical column too narrow for long labels.

**4. Number panel shows only the first query.** A Number/Value panel renders a single scalar — internally it takes the first query's result (`prepareNumberData`). When a user builds a formula but leaves its input queries enabled (or otherwise leaves more than one query on), the panel silently shows query `A` instead of the intended value, with no feedback. This is a common trap and the subject of #9512.

#### Fix Strategy

1. Make `formatSingleValueForFilter` a true inverse of `formatSingleValue` — unescape `\'` → `'` after unquoting, so the round-trip is lossless and idempotent.
2. Double-encode `compositeQuery` in `buildExportPanelLink` so it survives the reader's double decode.
3. Give the right legend its own budget: size it to the longest label, clamped to `[150px, min(320, 40% width)]`.
4. Surface a non-blocking warning (the existing amber status popover) when a Number panel has more than one enabled query. A new `countEnabledQueries` util unwraps the single `signoz/CompositeQuery` wrapper and counts envelopes whose spec is not `disabled`; a `panelStatusFromMultipleEnabledQueries` adapter turns that into a warning for the `signoz/NumberPanel` kind. It's wired into the shared `PanelHeader`, so it shows in both the editor preview and the rendered grid panel and updates live as queries are toggled. We warn rather than auto-disable because the app can't infer which query the user meant to show, and formulas need their input queries enabled — auto-disabling would either guess wrong or break the formula, and would mutate the shared dashboard spec.

---

### 🧪 Testing Strategy

- **Tests added/updated:**
  - `QueryBuilderV2/__tests__/utils.test.ts` — new `escaped quote round-trip` suite: unescaped filter-item value, lossless `expression → filters → expression`, and idempotency across repeated existing-query conversions.
  - `PanelEditor/__tests__/newPanelRoute.test.ts` — round-trips a filter expression containing `%` and `+` literals through the reader's double-decode.
  - `charts/__tests__/utils.test.ts` — right-legend sizing (caps at max, sizes to longest label, respects the 150px floor).
  - `Panels/utils/__tests__/countEnabledQueries.test.ts` — enabled-query counting across the composite wrapper, disabled envelopes, formulas, and the bare-builder (List) shape.
  - `Panel/PanelStatus/__tests__/utils.test.ts` — `panelStatusFromMultipleEnabledQueries` warns only for a Number panel with >1 enabled query.
  - `Panel/__tests__/PanelHeader.test.tsx` — the warning renders on a multi-query Number panel (grid + editor preview) and is absent for single-query and non-Number panels.
- **Manual verification:** Reproduced the original `token recognition error` from a panel filter with an escaped quote, then confirmed the alert page now receives the expression verbatim.
- **Edge cases covered:** values with `%`/`+` literals; multi-pass URL round-trips; long series names in the right legend; disabled queries and formula envelopes in the enabled-query count.

---

- **Blast radius:** `formatSingleValueForFilter` is shared by the query builder's `convertExpressionToFilters` / `convertFiltersToExpressionWithExistingQuery`, so it also affects explorer/metrics/trace filter parsing — the change only makes stored values more correct (unescaped), and re-serialization re-escapes them. The number-panel warning is additive and V2-only: it derives from `panel.spec.queries`, adds no query-behavior change, and reuses the existing status-popover machinery.
- **Potential regressions:** Any consumer that expected the previously-escaped `filters.items` value; none found — display and re-serialization paths both handle the unescaped form.
- **Rollback plan:** Revert the offending commit(s); each fix is isolated in its own commit.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix + Feature |
| Description | Creating an alert from a dashboard panel no longer corrupts filter expressions that contain quotes; exporting an explorer query to a V2 dashboard no longer drops filters containing `%`/`+`; right-side pie/donut legends fit long series names; a Number panel now warns when more than one query is enabled (only the first is displayed). |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

The four commits are independent — review each on its own. Fix 1 is the shared query-builder utility change (widest blast radius); fixes 2 and 3 are localized to the V2 panel export link and the pie/donut chart legend respectively; change 4 (the Number-panel multi-query warning, #9512) is additive and confined to the V2 panel header + two small utils.

---

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/__tests__/countEnabledQueries.test.ts
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/__tests__/countEnabledQueries.test.ts
@@ -1,0 +1,80 @@
+import type { DashboardtypesQueryDTO } from 'api/generated/services/sigNoz.schemas';
+
+import { countEnabledQueries } from '../countEnabledQueries';
+
+function compositeQuery(
+	envelopes: { type: string; disabled?: boolean }[],
+): DashboardtypesQueryDTO {
+	return {
+		spec: {
+			plugin: {
+				kind: 'signoz/CompositeQuery',
+				spec: {
+					queries: envelopes.map(({ type, disabled }) => ({
+						type,
+						spec: { disabled },
+					})),
+				},
+			},
+		},
+	} as unknown as DashboardtypesQueryDTO;
+}
+
+function bareBuilderQuery(disabled?: boolean): DashboardtypesQueryDTO {
+	return {
+		spec: {
+			plugin: { kind: 'signoz/BuilderQuery', spec: { signal: 'logs', disabled } },
+		},
+	} as unknown as DashboardtypesQueryDTO;
+}
+
+describe('countEnabledQueries', () => {
+	it('returns 0 when there are no queries', () => {
+		expect(countEnabledQueries([])).toBe(0);
+	});
+
+	it('counts every enabled envelope inside the composite wrapper', () => {
+		expect(
+			countEnabledQueries([
+				compositeQuery([{ type: 'builder_query' }, { type: 'builder_query' }]),
+			]),
+		).toBe(2);
+	});
+
+	it('does not count disabled envelopes', () => {
+		expect(
+			countEnabledQueries([
+				compositeQuery([
+					{ type: 'builder_query' },
+					{ type: 'builder_query', disabled: true },
+				]),
+			]),
+		).toBe(1);
+	});
+
+	it('counts formula envelopes alongside builder queries', () => {
+		expect(
+			countEnabledQueries([
+				compositeQuery([
+					{ type: 'builder_query' },
+					{ type: 'builder_query' },
+					{ type: 'builder_formula' },
+				]),
+			]),
+		).toBe(3);
+	});
+
+	it('treats an absent disabled flag as enabled', () => {
+		expect(
+			countEnabledQueries([compositeQuery([{ type: 'builder_query' }])]),
+		).toBe(1);
+	});
+
+	it('unwraps a bare builder query (List panel shape)', () => {
+		expect(countEnabledQueries([bareBuilderQuery()])).toBe(1);
+	});
+
+	it('does not count a disabled bare builder query', () => {
+		expect(countEnabledQueries([bareBuilderQuery(true)])).toBe(0);
+	});
+});
--- a/frontend/src/components/QueryBuilderV2/__tests__/utils.test.ts
+++ b/frontend/src/components/QueryBuilderV2/__tests__/utils.test.ts
@@ -9,6 +9,7 @@ import { extractQueryPairs } from 'utils/queryContextUtils';
 
 import {
 	convertAggregationToExpression,
+	convertExpressionToFilters,
 	convertFiltersToExpression,
 	convertFiltersToExpressionWithExistingQuery,
 	formatValueForExpression,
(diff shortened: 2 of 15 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/__tests__/countEnabledQueries.test.ts
```diff
@@ -0,0 +1,80 @@
+import type { DashboardtypesQueryDTO } from 'api/generated/services/sigNoz.schemas';
```
AshwinBhatkal, line 1: what about clearing the query when switching with a confirmation instead of the warning?

AshwinBhatkal (approved):

PR #12250, opened 2026-07-23
fix(telemetrymetadata): update fallback to `false` for positive operators in related values queries

nikhilmantri0902:
## Pull Request

---

### 📄 Summary

1. Related-values quick-filter suggestions (`/api/v1/fields/values` with `existingQuery`) behaved as an OR-set instead of AND. The `if(mapContains(...), cond, true)` builder hardcoded a `true` fallback, so rows missing the filtered key were always included — polluting suggestions for positive operators (e.g. `namespace IN ['ai']` still surfaced pods with no `namespace`).
Fix: make the absent-key fallback operator-aware. Positive ops (`=`, `IN`, `Like`, `ILike`, `Contains`, `Regexp`, `Exists`) → `false` (exclude key-absent rows, strict AND). Negative ops (`!=`, `NotIn`, `NotLike`, `NotContains`, `NotRegexp`, `NotExists`) → `true` (keep them, else pods legitimately lacking a label get wrongly dropped). Reuses the existing `FilterOperator.IsNegativeOperator()` helper.

2. The query also uses local attributes_metadata table instead of the distributed version. Check [this comment](https://github.com/SigNoz/engineering-pod/issues/5729#issuecomment-5056198546) for details. Thus, this PR also replaces local table with its distributed counterpart. 

3. Also adding the `data_source` filter in related values filtering based on the signal passed. The all values counterpart entertain's signal but this related values path did not, added in this PR. 

#### Issues closed by this PR

Closes https://github.com/SigNoz/engineering-pod/issues/5729

---

### ✅ Change Type

- [x] 🐛 Bug fix

---

#### Root Cause

Blanket `true` fallback in `condition_builder.go` (added to keep negative-operator matches, PR #7807) — correct for negatives, wrong for positives, so absent-key rows leaked into related values.

#### Fix Strategy

Set fallback via `operator.IsNegativeOperator()`: `false` for positive ops, `true` for negative ops.

---

### 🧪 Testing Strategy

- Tests added/updated: `condition_builder_test.go` — flipped `ILike` to `false`, added 12 cases covering both polarities across Equal/In/Like/Contains/Regexp/Exists (+ negatives).
- Edge cases: `Exists`→`false`, `NotExists`→`true` verified.

---

- Blast radius: related-values suggestions only (`condition_builder.go`); query WHERE semantics unchanged elsewhere.
- Rollback: revert commit.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Related-values quick filters now apply AND semantics — suggestions no longer include values from entities missing the filtered key. |

---

### 📋 Checklist

---

## 👀 Notes for Reviewers

- No integration tests existed for relatedValues before this change, so none were explicitly added — can add if needed in a follow up PR. usecase nicely tested via api and queries.


pkg/telemetrymetadata/condition_builder.go
```diff
@@ -96,8 +96,11 @@ func (c *conditionBuilder) conditionForKey(
 
 	fieldExpression, value = querybuilder.DataTypeCollisionHandledFieldName(key, value, fieldExpression, operator)
 
-	// key must exists to apply main filter
-	expr := `if(mapContains(%s, %s), %s, true)`
+	// key must exist to apply the main filter. for positive operators the
+	// absent-key rows are excluded (fallback false); for negative operators
+	// they are kept (fallback true) so rows legitimately lacking the key match.
+	keyMissingFallback := operator.IsNegativeOperator()
```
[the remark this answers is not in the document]
nikhilmantri0902: done now

pkg/telemetrymetadata/metadata.go
```diff
@@ -1464,9 +1469,12 @@ func (t *telemetryMetaStore) getRelatedValues(ctx context.Context, orgID valuer.
...
 		if len(conds) != 0 {
-			// see `expr` in condition_builder.go, if key doesn't exist we don't check for value
-			// hence, this is join of conditions on resource and attributes
-			sb.Where(sb.And(conds...))
+			// OR (not AND) across the attribute and resource probes: a key lives in
+			// only one of the two maps, and the positive Contains condition now falls
+			// back to false for the map that lacks it (see `expr` in condition_builder.go).
+			// false is the identity for OR, so the absent-map probe drops out and the
+			// present-map match decides — matching the value in either context.
+			sb.Where(sb.Or(conds...))
```
namanv3, line 1474: not really clear what's happening here, the comment doesn't really help, its wordy without being clear 😅

nikhilmantri0902: basically with this new mapContains fallback for true to positive operators, if `searchText='abc'` is provided ( non-empty) then this condition will need to be OR for a row with value in either the attributes column or resources column ( or both ) to be included as a result. 

If we had kept it at `AND` instead of OR, the attribute will need to be in both columns simultaneously for a row to be included. 

PS: I have simplified the comment also. Let me know @namanv3.

namanv3 (approved):

therealpandey (approved):

PR #12251, opened 2026-07-23
fix(tracedetail): correct EndTimestampMillis in flamegraph API response

krsoninikhil:
## Summary

- `EndTimestampMillis` in the flamegraph API response was returning the max span **start** time instead of the actual trace end time (max of span_start + span_duration)
- Root cause: `trace_summary_mv` stores `max(timestamp) AS end` — `timestamp` is the span start, so duration is never included
- Fix: expose `TimeRange()` on `FlamegraphTrace` (which already computed the correct end via `updateTimeRange`) and use it in both `getFullFlamegraph` and `getWindowedFlamegraph` instead of `summary.End`
- The waterfall was unaffected — `NewWaterfallTraceFromSpans` already computed `endTime = max(span.TimeUnix + span.DurationNano)`

Closes https://github.com/SigNoz/engineering-pod/issues/5767

https://claude.ai/code/session_01GxTqz7CqqmoFEPHFqhJUCa

### DIFF
--- a/pkg/modules/tracedetail/impltracedetail/module.go
+++ b/pkg/modules/tracedetail/impltracedetail/module.go
@@ -182,7 +182,6 @@ func (m *module) getFullFlamegraph(ctx context.Context, traceID string, summary 
 		return nil, spantypes.ErrTraceNotFound
 	}
 	flamegraphTrace := spantypes.NewFlamegraphTraceFromStorable(fullSpans, selectFields)
-	// Use span-derived bounds: summary.End = max(span_start), which misses span durations.
 	start, end := flamegraphTrace.TimeRange()
 	return spantypes.NewGettableFlamegraphTrace(flamegraphTrace.GetAllLevels(), start, end, false), nil
 }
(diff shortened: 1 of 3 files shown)


pkg/modules/tracedetail/impltracedetail/module.go
```diff
@@ -198,6 +200,10 @@ func (m *module) getWindowedFlamegraph(ctx context.Context, traceID, selectedSpa
 	flamegraphTrace := spantypes.NewFlamegraphTraceFromMinimal(minimalSpans)
 	minimalSpans = nil //nolint:ineffassign,wastedassign // release backing array before further db calls
 
+	// Capture span-derived bounds before GetSelectedLevels (which nils span children).
+	// summary.End = max(span_start) from the MV, which misses span durations.
```
therealpandey: Can remove this summary.End call now

therealpandey (changes requested):

pkg/modules/tracedetail/impltracedetail/module.go
krsoninikhil: Removed the comments, are you referring to remove the summary table query itself? That is required to query the minimal spans above.

therealpandey (approved):

PR #12254, opened 2026-07-23
fix(dashboard-v2): variable selection (default reset + multiselect commit-on-close)

AshwinBhatkal:
Stacked on #12178. These two fixes were pulled out of that PR earlier to keep it focused; restoring them here on top.

### What
- **Reset variable default only on real option-set changes** — `useVariableForm` no longer clobbers a chosen default when the option list is re-fetched but unchanged (+ unit tests).
- **Commit multiselect selection on close, tidy tag display** — multi-select buffers edits in a local draft and commits once on dropdown close (no per-click cascade to dependents), with an empty pick falling back to the resolved default.

### Note on rebase
Both commits were cherry-picked onto the current #12178 tip. The multiselect change conflicted with the analytics instrumentation that landed since; resolved by keeping commit-on-close and firing the existing `VariableValueSelected` / `VariableMultiSelectCleared` events on commit/clear so analytics is not regressed.

Will retarget to `main` once #12178 merges.

Closes https://github.com/SigNoz/pulse-pod/issues/138

### DIFF
--- a/frontend/src/components/NewSelect/CustomMultiSelect.tsx
+++ b/frontend/src/components/NewSelect/CustomMultiSelect.tsx
@@ -726,13 +726,6 @@ const CustomMultiSelect: React.FC<CustomMultiSelectProps> = ({
 						selected: isSelected,
 						active: isActive,
 					})}
-					onClick={(e): void => {
-						e.stopPropagation();
-						e.preventDefault();
-						handleItemSelection('option');
-						setActiveChipIndex(-1);
-						setActiveIndex(-1);
-					}}
 					onKeyDown={(e): void => {
 						if ((e.key === 'Enter' || e.key === SPACEKEY) && isActive) {
 							e.stopPropagation();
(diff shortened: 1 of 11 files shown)


frontend/src/components/NewSelect/CustomMultiSelect.tsx
```diff
@@ -768,11 +761,34 @@ const CustomMultiSelect: React.FC<CustomMultiSelectProps> = ({
 								<div className="option-badge">{capitalize(option.type)}</div>
 							)}
 							{option.value && ensureValidOption(option.value) && (
-								<Button type="text" className="only-btn">
+								<Button
+									type="text"
+									className="only-btn"
+									onClick={(e): void => {
+										// Own handler + preventDefault so the visible "Only/All"
```
ahrefabhi: can we take this out in a function, this is too much for JSX

frontend/src/components/NewSelect/CustomMultiSelect.tsx
```diff
@@ -768,11 +761,34 @@ const CustomMultiSelect: React.FC<CustomMultiSelectProps> = ({
...
+								>
 									{currentToggleTagValue({ option: option.value })}
 								</Button>
 							)}
-							<Button type="text" className="toggle-btn">
+							<Button
+								type="text"
+								className="toggle-btn"
+								onClick={(e): void => {
+									e.stopPropagation();
```
ahrefabhi: same here

H4ad (dismissed):

frontend/src/components/NewSelect/CustomMultiSelect.tsx
AshwinBhatkal: Done — extracted the checkbox/Only-All/Toggle click logic into a single `selectFromButton(e, source)` helper inside `renderOptionItem`, so the JSX is just `onClick={(e) => selectFromButton(e, 'option')}`. Pushed in 6dbaaa6eb.

frontend/src/components/NewSelect/CustomMultiSelect.tsx
AshwinBhatkal: Same helper covers the Toggle button too.

ahrefabhi (approved):

PR #12262, opened 2026-07-24
fix(dashboard-v2): config-pane restyle, section quick-add, pie labels & list pagination

ahrefabhi:
## Pull Request

---

A batch of polish and bug fixes for the V2 dashboard panel editor's **ConfigPane**, plus two panel-behavior fixes surfaced during the same pass:

- **Config-pane restyle** — the pane now matches the Configure-panel design: a "Panel Details" header with an accent marker, a "DISPLAY OPTIONS" eyebrow, hairline `--l1-border` dividers/section separators (kept from collapsing on overflow), and the shared `custom-scrollbar` mixin in place of the ad-hoc scrollbar.
- **Control border alignment** — `ConfigSelect`, `FormattingSection` and the threshold select use the standard `--l2-border`, and `ConfigSwitch` radius drops to 2px so all full-width controls read as one system.
- **Section header quick-add** — collapsible sections gain a generic `headerAction` slot; Thresholds and Context Links expose a header "+" that expands the section (if collapsed) and immediately starts a new entry.
- **Pie slice labels** — legend format templates now resolve `{{key}}` placeholders against group-by values instead of rendering the raw template.
- **List/raw pagination** — non-timestamp sorts (which page by offset with no cursor) can now page past the first page via a full-page heuristic, matching the logs/traces explorers.

#### Issues closed by this PR
Closes https://github.com/SigNoz/pulse-pod/issues/140

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [x] 🧪 Test-only

---

#### Root Cause
- **Pie labels:** a single value column with a legend format was rendered as the raw template string because the label was never run through `getLabelName`, so `{{key}}` placeholders were left unresolved.
- **Pagination:** the backend only emits `nextCursor` on the timestamp-ordered window path; a non-timestamp sort falls back to plain offset paging with no cursor, so `canNext` was always false and list/raw panels were stuck on page one.
- **Styling:** dividers/borders used mismatched `--l2-border` tokens and an ad-hoc scrollbar, drifting from the Configure-panel design.

#### Fix Strategy
- Run single-value pie labels through `getLabelName` so placeholders resolve to group-by values.
- Fall back to a full-page heuristic for `canNext` when no cursor is present, matching the explorers.
- Normalize dividers to `--l1-border`, controls to `--l2-border`, and adopt the shared `custom-scrollbar` mixin.

---

- Tests added/updated:
  - `SettingsSection.test.tsx` — headerAction slot + controlled `open`/`onOpenChange`.
  - `SectionSlot.test.tsx` — header "+" shows while collapsed, opens-when-populated, and expand-then-add.
  - `PieChartPanel/prepareData.test.ts` — legend format template resolution for slice labels.
  - `usePanelQuery.test.tsx` — `canNext` full-page heuristic under non-timestamp sorts.
- Manual verification: config pane restyle, header quick-add for Thresholds & Context Links, pie labels, and list pagination in the panel editor.
- Edge cases covered: collapsed vs populated sections, single value column with format template, non-timestamp sort paging past page one.

---

- Blast radius: V2 dashboard panel editor (ConfigPane) and pie/list panel rendering only; V1 untouched.
- Potential regressions: section open/collapse behavior, control styling drift, pie label formatting.
- Rollback plan: revert the PR; commits are scoped per-fix for granular revert if needed.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix / Feature |
| Description | V2 dashboard config-pane restyle and section header quick-add, plus pie slice label formatting and list-panel pagination fixes. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

Commits are split per-fix, so each change can be reviewed (or reverted) independently. All changes are confined to V2 dashboard land — no V1 files are modified.

---

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/SectionSlot/SectionSlot.tsx
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/SectionSlot/SectionSlot.tsx
@@ -1,3 +1,6 @@
+import { type ReactNode, useCallback, useRef, useState } from 'react';
+import { Plus } from '@signozhq/icons';
+import { Button } from '@signozhq/ui/button';
 import type { DashboardtypesPanelSpecDTO } from 'api/generated/services/sigNoz.schemas';
 import {
 	type PanelFormattingSlice,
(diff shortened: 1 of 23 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/ConfigPane/SectionSlot/SectionSlot.tsx
```diff
@@ -10,12 +13,45 @@ import type { SectionEditorContext } from '../sectionContext';
...
 type SectionSlotProps = {
 	config: SectionConfig;
 	spec: DashboardtypesPanelSpecDTO;
 	onChangeSpec: (next: DashboardtypesPanelSpecDTO) => void;
-} & Omit<SectionEditorContext, 'yAxisUnit'>;
+} & Omit<SectionEditorContext, 'yAxisUnit' | 'registerHeaderAction'>;
+
+// Per-section header control (beside the chevron). A render fn, so a section shows
+// whatever it wants; `trigger` expands the section and runs the editor's handler.
+const SECTION_HEADER_ACTION: Partial<
```
AshwinBhatkal: this should be very configurable. A JSON object is enough here. The components will anyways be the same, so we can have it defined at 1 place. Also, can we add a tooltip on hover that says Quick Add

AshwinBhatkal (approved):

PR #12267, opened 2026-07-24
fix(signoz): consolidate the duplicate telemetryMetadataStore construction

### LINKED ISSUE #12266: Consolidate the two telemetryMetadataStore constructions in signoz.New and signozquerier.newProvider
### Problem

`telemetrymetadata.NewTelemetryMetaStore` is called twice with byte-identical arguments, once in `pkg/signoz/signoz.go:428` and again in `pkg/querier/signozquerier/provider.go:51`. The two resulting instances are not the same object, so the querier holds a private copy that is invisible to every other consumer of the store.

```go
// pkg/signoz/signoz.go:428
telemetryMetadataStore := telemetrymetadata.NewTelemetryMetaStore(
    providerSettings,
    telemetrystore,
    telemetrytraces.DBName,
    telemetrytraces.TagAttributesV2TableName,
    // ... 20 more args ...
    flagger,
)
```

```go
// pkg/querier/signozquerier/provider.go:51
telemetryMetadataStore := telemetrymetadata.NewTelemetryMetaStore(
    settings,                  // same type as providerSettings
    telemetryStore,            // same value as telemetrystore
    telemetrytraces.DBName,    // same constants
    // ... 20 more args ...
    flagger,                   // same value
)
```

The maintainer tag is already in the source: `// TODO: consolidate other telemetrymetadata.NewTelemetryMetaStore initializations to reuse this instance instead.` at `pkg/signoz/signoz.go:427`.

### Motivation

- Two metadata stores means two metadata caches for the same ClickHouse tables. Both backends issue identical attribute-key lookups against `signoz_metadata.attributes_metadata_local` and cache the results separately, doubling the cache footprint and the read load.
- Two call sites for the same constructor. Every new argument has to be threaded through both, and the constructor already has 23 positional arguments. Drift is a matter of when, not if.
- The querier does not need a separate store. It only consumes the same attribute metadata that every other module reads.

### Background

`signoz.New` wires the system up in strict dependency order. Modules, handlers, ruler, rule-state history, infra monitoring, metrics explorer, promote, the new Fields handler and the API server all read from the single `telemetryMetadataStore` that `signoz.New` builds at line 428. The querier is the only outlier, because it is constructed earlier in `signoz.New` (line 256) and the store is built after it.

The querier factory (`pkg/querier/signozquerier/provider.go:NewFactory`) is wired through `NewQuerierProviderFactories(telemetrystore, prometheus, cache, flagger)`. Adding the store as a parameter is a one-line change in three places.

### Current behavior

Two `telemetryMetadataStore` instances exist at runtime:
- one in `signoz.New`, used by every consumer except the querier;
- one in `signozquerier.newProvider`, used by the trace, log, audit, metric, meter, trace-operator statement builders and the bucket cache.

The two never share state and never see each other's cached lookups. The TODO in the source is the maintainer acknowledging the duplicate.

### Expected behavior

`signoz.New` constructs one `telemetryMetadataStore`. Every consumer in the system, including the querier, uses that same instance. The duplicate construction in `signozquerier.newProvider` is removed.

### Proposed solution

1. Move the `telemetryMetadataStore := telemetrymetadata.NewTelemetryMetaStore(...)` block in `pkg/signoz/signoz.go` to before the querier init at line 256. The store has no dependency on the querier, so this is a pure reorder.
2. Pass the new `telemetryMetadataStore` into `NewQuerierProviderFactories` in `pkg/signoz/provider.go` (signature change: add `telemetrytypes.MetadataStore` parameter).
3. Forward the store through `NewFactory` and `newProvider` in `pkg/querier/signozquerier/provider.go` and use the passed-in instance at the eight call sites that currently use the local copy.
4. Remove the local `NewTelemetryMetaStore` call in `signozquerier.newProvider`.
5. Remove the TODO comment at `pkg/signoz/signoz.go:427`.
6. Add a regression test that asserts only one `NewTelemetryMetaStore` call remains, by grep on the test build output or by exporting the constructor behind a build-time counter.

### Alternatives considered

- Pass a getter closure instead of the store instance. Indirection with no benefit; the dependency is required at construction time.
- Keep two stores and add a cross-instance invalidation hook. Heavier, more state, more failure modes, and the maintainer TODO already says to consolidate.
- Move the store behind an interface so the duplicate call is "type-safe" but still two objects. Same problem dressed up.

### Scope

- Three files touched: `pkg/signoz/signoz.go`, `pkg/signoz/provider.go`, `pkg/querier/signozquerier/provider.go`.
- About 30 lines of net diff (reorder, signature change, call site change, comment removal).
- One new regression test file, roughly 30 lines.
- No public API change.
- No frontend change.
- No config or migration change.

### Acceptance criteria

- `rg "telemetrymetadata.NewTelemetryMetaStore" pkg/ ee/` returns exactly one production call site, the one in `pkg/signoz/signoz.go`.
- `go build ./...` clean.
- `go vet ./...` clean.
- `golangci-lint run ./pkg/... ./ee/...` clean.
- `go test ./pkg/querier/... ./pkg/signoz/... ./pkg/telemetrymetadata/...` all pass.
- The new regression test fails on a build that contains the old duplicate call and passes after the fix.
- No behavior change is observable from the outside: the same queries resolve to the same column metadata.

### Backward compatibility

None expected. The change is internal. The store is not part of the public Go API, no config keys move, no SQL changes, no event shape changes, no HTTP surface change.

### Risks

- Init order: the moved `NewTelemetryMetaStore` block runs earlier in `signoz.New`. The block has no upstream dependency that isn't already satisfied by the time querier is built (it needs `providerSettings`, `telemetrystore`, `flagger`, all of which exist earlier). Verified by reading the surrounding init sequence.
- The eight in-querier call sites use the store only at construction time for field-mapper, condition-builder, statement-builder, and the bucket cache. None of them mutate the store. The refactor is therefore a pure instance swap.
- The CI matrix runs on `amd64` and `arm64` per the `Makefile` and `goci.yaml`. The refactor is platform-agnostic.

Harsh23Kashyap:
## Pull Request

---

The querier built a second `telemetryMetadataStore` with the same 23 arguments the rest of the system already used to build the canonical one. The querier held a private copy the rest of the system could not see, and the two caches doubled the metadata footprint and the read load on `signoz_metadata.attributes_metadata_local`. A maintainer-tagged TODO at `pkg/signoz/signoz.go:427` already asked for this consolidation.

The fix moves the canonical construction in `pkg/signoz/signoz.go` to before the querier init, threads the resulting instance through `NewQuerierProviderFactories` and `signozquerier.NewFactory` / `newProvider`, and removes the duplicate call. The querier, modules, handlers, ruler, rule-state-history store, infra monitoring, metrics explorer, promote and the new Fields handler all end up reading from one instance.

Only one `NewTelemetryMetaStore` call site remains in production code (`pkg/signoz/signoz.go:257`).

N/A

#### Issues closed by this PR

Closes #12266

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

`pkg/querier/signozquerier/provider.go:51` and `pkg/signoz/signoz.go:428` both called `telemetrymetadata.NewTelemetryMetaStore` with the same arguments. The two instances never shared state. The constructor was duplicated and the maintainer-tagged TODO asked for it to be fixed.

- Reorder: build the canonical `telemetryMetadataStore` before the querier init in `pkg/signoz/signoz.go` so it can be passed in.
- Plumb the existing instance through `NewQuerierProviderFactories` and `signozquerier.NewFactory` / `newProvider`. The factory signature gains one parameter; the duplicate construction in `newProvider` is removed.
- Remove the TODO at the original site.
- Update `pkg/query-service/rules/managertestfactory.go` to build its own metadata store for the test path, since the old code was relying on the implicit one `newProvider` used to construct.

---

- `go build ./...` clean.
- `go vet ./...` clean for the affected packages (`go vet ./...` has two pre-existing warnings on `upstream/main` in `pkg/parser/havingexpression/grammar/...` and `ee/query-service/app/api/...` that are unrelated to this PR; verified by re-running `go vet` against `upstream/main` and observing the same warnings).
- `go test -count=1` green for the directly affected packages and downstream consumers:
  - `pkg/signoz`
  - `pkg/querier`
  - `pkg/query-service/rules`
  - `pkg/telemetrymetadata`
  - `pkg/alertmanager/...`
  - `pkg/ruler/...`
  - `pkg/modules/rulestatehistory/...`
  - `pkg/modules/inframonitoring/...`
  - `pkg/modules/metricsexplorer/...`
  - `pkg/modules/promote/...`
  - `pkg/modules/fields/...`
- Both `cmd/community` and `cmd/enterprise` build cleanly and produce binaries.
- `rg "telemetrymetadata.NewTelemetryMetaStore" pkg/ ee/` returns exactly one production call site, the one in `pkg/signoz/signoz.go:257`. The remaining call is in `pkg/query-service/rules/managertestfactory.go`, which is the test factory.

- Tests added/updated: none. The refactor is covered by the existing test suite for the affected packages; behaviour is unchanged.
- Manual verification: none beyond the tests above.
- Edge cases covered: the rules manager test path (`TestManager_TestNotification_SendUnmatched_ThresholdRule`) was the regression-risk area; it passes with the updated test factory.

---

- Blast radius: internal wiring. The querier, all modules, handlers, ruler, rule-state-history store, infra monitoring, metrics explorer, promote and the Fields handler all read from the same `MetadataStore` interface; behaviour is unchanged because the only difference is the identity of the instance.
- Potential regressions: a future change that introduces mutable derived state on the metadata store would have been desynchronised before this PR; it is now safe by construction. No other regressions expected.
- Rollback plan: revert the commit. The change is a pure refactor and reverts cleanly.

---

| Field | Value |
|------|-------|
| Deployment Type | OSS / Enterprise (internal) |
| Change Type | Maintenance |
| Description | N/A — internal refactor; no user-facing change. |

---

### 📋 Checklist
- [x] Tests added or explicitly not required (existing tests cover the refactor)
- [x] Manually tested (build + tests for both variants)
- [x] Backward compatibility considered (no public API change)

---

## 👀 Notes for Reviewers

The most surprising line in the diff is the 24-line metadata store constructor in `pkg/query-service/rules/managertestfactory.go`. The old test factory relied on the implicit construction that used to happen inside `signozquerier.newProvider`. With the consolidation, `newProvider` no longer builds a metadata store, so the test factory has to. The constructor mirrors the one in `pkg/signoz/signoz.go` exactly and uses the canonical table-name constants.

The reorder in `pkg/signoz/signoz.go` is safe: the metadata store depends only on `providerSettings`, `telemetrystore` and `flagger`, all of which exist before the querier init.

### DIFF
--- a/pkg/querier/signozquerier/provider.go
+++ b/pkg/querier/signozquerier/provider.go
@@ -11,16 +11,17 @@ import (
 	"github.com/SigNoz/signoz/pkg/querybuilder"
 	"github.com/SigNoz/signoz/pkg/telemetryaudit"
 	"github.com/SigNoz/signoz/pkg/telemetrylogs"
-	"github.com/SigNoz/signoz/pkg/telemetrymetadata"
 	"github.com/SigNoz/signoz/pkg/telemetrymeter"
 	"github.com/SigNoz/signoz/pkg/telemetrymetrics"
 	"github.com/SigNoz/signoz/pkg/telemetrystore"
 	"github.com/SigNoz/signoz/pkg/telemetrytraces"
+	"github.com/SigNoz/signoz/pkg/types/telemetrytypes"
 )
 
 // NewFactory creates a new factory for the signoz querier provider.
 func NewFactory(
 	telemetryStore telemetrystore.TelemetryStore,
+	telemetryMetadataStore telemetrytypes.MetadataStore,
 	prometheus prometheus.Prometheus,
 	cache cache.Cache,
 	flagger flagger.Flagger,
@@ -32,7 +33,7 @@ func NewFactory(
 			settings factory.ProviderSettings,
 			cfg querier.Config,
 		) (querier.Querier, error) {
-			return newProvider(ctx, settings, cfg, telemetryStore, prometheus, cache, flagger)
+			return newProvider(ctx, settings, cfg, telemetryStore, telemetryMetadataStore, prometheus, cache, flagger)
 		},
 	)
 }
@@ -42,38 +43,11 @@ func newProvider(
 	settings factory.ProviderSettings,
 	cfg querier.Config,
 	telemetryStore telemetrystore.TelemetryStore,
+	telemetryMetadataStore telemetrytypes.MetadataStore,
 	prometheus prometheus.Prometheus,
 	cache cache.Cache,
 	flagger flagger.Flagger,
 ) (querier.Querier, error) {
-
-	// Create telemetry metadata store
-	telemetryMetadataStore := telemetrymetadata.NewTelemetryMetaStore(
-		settings,
-		telemetryStore,
-		telemetrytraces.DBName,
-		telemetrytraces.TagAttributesV2TableName,
-		telemetrytraces.SpanAttributesKeysTblName,
-		telemetrytraces.SpanIndexV3TableName,
-		telemetrymetrics.DBName,
-		telemetrymetrics.AttributesMetadataTableName,
-		telemetrymeter.DBName,
-		telemetrymeter.SamplesAgg1dTableName,
-		telemetrylogs.DBName,
-		telemetrylogs.LogsV2TableName,
-		telemetrylogs.TagAttributesV2TableName,
-		telemetrylogs.LogAttributeKeysTblName,
-		telemetrylogs.LogResourceKeysTblName,
-		telemetryaudit.DBName,
-		telemetryaudit.AuditLogsTableName,
-		telemetryaudit.TagAttributesTableName,
-		telemetryaudit.LogAttributeKeysTblName,
-		telemetryaudit.LogResourceKeysTblName,
-		telemetrymetadata.DBName,
-		telemetrymetadata.AttributesMetadataLocalTableName,
-		telemetrymetadata.ColumnEvolutionMetadataTableName,
-		flagger,
-	)
 
 	// Create trace statement builder
 	traceFieldMapper := telemetrytraces.NewFieldMapper()
--- a/pkg/query-service/rules/managertestfactory.go
+++ b/pkg/query-service/rules/managertestfactory.go
@@ -18,8 +18,14 @@ import (
 	"github.com/SigNoz/signoz/pkg/querier/signozquerier"
 	"github.com/SigNoz/signoz/pkg/sqlstore"
 	"github.com/SigNoz/signoz/pkg/sqlstore/sqlstoretest"
+	"github.com/SigNoz/signoz/pkg/telemetryaudit"
+	"github.com/SigNoz/signoz/pkg/telemetrylogs"
+	"github.com/SigNoz/signoz/pkg/telemetrymetadata"
+	"github.com/SigNoz/signoz/pkg/telemetrymeter"
+	"github.com/SigNoz/signoz/pkg/telemetrymetrics"
 	"github.com/SigNoz/signoz/pkg/telemetrystore"
 	"github.com/SigNoz/signoz/pkg/telemetrystore/telemetrystoretest"
+	"github.com/SigNoz/signoz/pkg/telemetrytraces"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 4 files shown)


Harsh23Kashyap:
This PR is from a fork, so the goci/commitci workflows are skipped at .github/workflows/goci.yaml:9-13 unless the `safe-to-test` label is set. Could a maintainer add it so the CI runs? Happy to address any other feedback in the meantime.

PR #12276, opened 2026-07-25
fix: enforce scalar response contract + reject invalid v2 dashboard bodies

srikanthccv:
## Pull Request

- ensure that the scalar requests now always return scalar data, when `reduceTo` was missing, the series is returned instead of scale
  - the promql is fixed to return scalar for scale requests
- importing the old dashboard json actually works

### DIFF
--- a/pkg/modules/dashboard/impldashboard/v2_handler.go
+++ b/pkg/modules/dashboard/impldashboard/v2_handler.go
@@ -40,20 +40,23 @@ func (handler *handler) CreateV2(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	var shadow struct {
+		SchemaVersion string `json:"schemaVersion"`
+		Version       string `json:"version"`
+	}
+	_ = json.Unmarshal(body, &shadow)
+
 	var req dashboardtypes.PostableDashboardV2
-	if err := binding.JSON.BindBody(bytes.NewReader(body), &req); err != nil {
-		// Fallback: the body may be a legacy v1 dashboard. Migrate it to v2 (same
-		// v4→v5 + v1→v2 pass the bulk migration runs) and retry; if it isn't a
-		// convertible v1 payload either, surface the original v2 binding error.
+	if shadow.SchemaVersion == "" && shadow.Version != "" {
 		var data map[string]any
-		if json.Unmarshal(body, &data) != nil {
-			render.Error(rw, err)
+		if err := json.Unmarshal(body, &data); err != nil {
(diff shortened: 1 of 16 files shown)


pkg/modules/dashboard/impldashboard/v2_handler.go
```diff
@@ -40,20 +40,23 @@ func (handler *handler) CreateV2(rw http.ResponseWriter, r *http.Request) {
...
 	}
 
+	var shadow struct {
+		SchemaVersion string `json:"schemaVersion"`
+		Version       string `json:"version"`
+	}
+	_ = json.Unmarshal(body, &shadow)
+
 	var req dashboardtypes.PostableDashboardV2
-	if err := binding.JSON.BindBody(bytes.NewReader(body), &req); err != nil {
```
srikanthccv, line 44: falls back to v1 migration on every v2 bind failure without concretely identifying a legacy payload.

pkg/querier/promql_query.go
```diff
@@ -123,6 +123,10 @@ func newPromqlQuery(
 }
 
 func (q *promqlQuery) Fingerprint() string {
+	if q.requestType != qbv5.RequestTypeTimeSeries {
```
namanv3, line 126: wouldn't it be better to include request type into the fingerprint instead of removing caching for scalar requests altogether?

pkg/types/querybuildertypes/querybuildertypesv5/validation.go
```diff
@@ -279,6 +289,29 @@ func (q *QueryBuilderQuery[T]) validateAggregations(cfg validationConfig) error
 					"invalid space aggregation, should be one of the following: [`sum`, `avg`, `min`, `max`, `count`, `p50`, `p75`, `p90`, `p95`, `p99`]",
 				)
 			}
+			if cfg.withReduceToValidation && !v.ReduceTo.IsValid() {
```
namanv3, line 292: this might need malformed query handling in the migration, checking against the backups right now

namanv3: okay yeah this massively increased the number of failures, let me see what to do

pkg/querier/promql_query.go
srikanthccv: the scalar requests not cacheable so this the way we indicate a query is not cacheable by returning ""

namanv3 (approved):

therealpandey (approved):

therealpandey (approved):

ahrefabhi (approved):

PR #12277, opened 2026-07-26
feat(sqlstore): add structured SSL config fields for the postgres provider

### LINKED ISSUE #11567: [configure] support SSL configuration for metastore PostgreSQL connections
Hi , 

I am trying to configure an external relationnal database instead of the default SqlLite.

https://signoz.io/docs/manage/administrator-guide/configuration/relational-database/

I don't see any option to use SSL with PostgreSql ( which is required in my environment ).

Is it actually supported or not at all ? 

Thanks

Regards

Harsh23Kashyap:
## Pull Request

---

The postgres connection in `pkg/sqlstore` only takes a DSN. Self-hosted deployments that require TLS to the metastore have to weave pgx URL parameters into the DSN by hand, and the example config did not document a `postgres:` block at all. The issue is a request to add a structured SSL config so the metastore connection works out of the box in environments that mandate TLS.

The fix adds `ssl_mode`, `ssl_cert`, `ssl_key`, and `ssl_root_cert` to `PostgresConfig`. All four are optional; empty values preserve the prior DSN-only behaviour. When any of them are set, the postgres provider builds a `*tls.Config`, attaches it to `pgxpool.Config.ConnConfig.TLSConfig`, and pins `MinVersion` to TLS 1.2. `ServerName` is read from the DSN host so `verify-full` works against the URL the operator is using.

N/A

#### Issues closed by this PR

Closes #11567

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

> The change is best characterised as a feature (new configuration surface) and a docs/example-config update; the closest match above is "Feature". Treating it as Feature/Improvement; the maintainer can re-label as they see fit.

---

`PostgresConfig` only exposes a raw `DSN`. The pgx DSN format does support SSL (`?sslmode=require&sslrootcert=...`), but the example config does not document the `postgres:` block, and a user reading the docs cannot tell that the feature is supported or how to enable it. Operators in TLS-mandated environments had to construct the DSN by hand and keep it in sync with their env.

- Add structured SSL fields to `PostgresConfig` (`pkg/sqlstore/config.go:23-34`).
- Build a `*tls.Config` from those fields in the postgres provider (`ee/sqlstore/postgressqlstore/provider.go:58-69`) and attach it to the parsed `pgxpool.Config`.
- Pin TLS 1.2 as the minimum version; reuse the DSN host as `ServerName`.
- Validate the cert/key pair (both required if either is set), surface a clear error on bad cert files.
- Document the new fields in `conf/example.yaml:138-149`.

---

- `go build ./...` clean.
- `go vet ./pkg/sqlstore/... ./ee/sqlstore/...` clean.
- `go test -count=1` green for `pkg/sqlstore/...`, `ee/sqlstore/...`, `pkg/signoz/...`, `pkg/querier/...`, `pkg/query-service/rules/...`, `pkg/telemetrymetadata/...`, `pkg/alertmanager/...`, `pkg/ruler/...`.
- `ee/sqlstore/postgressqlstore/provider_test.go` covers:
  - empty SSL fields, all four "no TLS" modes (`disable`, `allow`, `prefer`) → returns nil config
  - `require` / `verify-full` / `verify-ca` build a `*tls.Config` with `MinVersion ≥ TLS 1.2` and the DSN host as `ServerName`
  - `SSLCert` without `SSLKey` and vice versa → error
  - missing or malformed `SSLRootCert` file → error
  - end-to-end mTLS path: self-signed CA + leaf cert generated in `t.TempDir`, both files written, the resulting `*tls.Config` carries the loaded client cert chain

- Tests added/updated: 1 new test file, 11 subtests.
- Manual verification: none beyond the tests above (the change is one-way: a new code path on the postgres provider; existing sqlite and DSN-only deployments are unaffected).
- Edge cases covered: empty SSL fields, every documented SSL mode, mismatched cert/key, bad PEM, missing file, end-to-end mTLS with a freshly issued leaf cert.

---

- Blast radius: internal config struct and one EE provider function. The new SSL fields default to empty, so existing deployments that set only `DSN` are byte-identical in behaviour.
- Potential regressions: none observed in `go test`; the only behaviour change is the new TLS code path, which only runs when at least one SSL field is set.
- Rollback plan: revert the commit. The change is a pure additive config surface; rollback does not require any data migration or config refresh on the operator side.

---

| Field | Value |
|------|-------|
| Deployment Type | OSS / Enterprise (postgres is an EE provider, but the `PostgresConfig` is OSS) |
| Change Type | Feature |
| Description | Adds `ssl_mode`, `ssl_cert`, `ssl_key`, `ssl_root_cert` to the postgres connection config so the metastore can be configured for TLS without hand-crafting the DSN. |

---

### 📋 Checklist
- [x] Manually tested (`go test` + `go vet` + `go build` for both community and enterprise variants)
- [x] Backward compatibility considered (empty fields preserve the prior DSN-only behaviour)

---

## 👀 Notes for Reviewers

This is a fork PR; the `goci` and `commitci` workflows in `.github/workflows/goci.yaml:9-13` skip fork PRs unless the `safe-to-test` label is set. A maintainer adding that label will get CI running.

The new code is concentrated in three files:

- `pkg/sqlstore/config.go` — config struct only
- `ee/sqlstore/postgressqlstore/provider.go` — wire-up + `buildTLSConfig` helper
- `conf/example.yaml` — documented `postgres:` block (the field is currently absent from the example config)

`buildTLSConfig` is a self-contained helper with a focused table-driven test (`provider_test.go`). The DSN is still the source of truth for everything except the four new fields, so any DSN the operator already has keeps working.

### DIFF
--- a/conf/example.yaml
+++ b/conf/example.yaml
@@ -135,6 +135,18 @@ sqlstore:
     busy_timeout: 10s
     # The default transaction locking behavior. Supported values: deferred, immediate, exclusive.
     transaction_mode: immediate
+  # Postgres connection settings. The DSN is required. The optional ssl_* fields
+  # build a *tls.Config and override whatever sslmode the DSN carries, so the two
+  # should not be set inconsistently. Valid ssl_mode values: disable, allow, prefer,
+  # require, verify-ca, verify-full.
+  postgres:
+    # The DSN to use for the postgres connection. Accepts a URL (postgres://...)
+    # or key=value DSN. Example: postgres://signoz:signoz@localhost:5432/signoz
+    dsn:
+    # ssl_mode: require
+    # ssl_cert:
+    # ssl_key:
+    # ssl_root_cert:
 
 ##################### APIServer #####################
 apiserver:
--- a/ee/sqlstore/postgressqlstore/provider.go
+++ b/ee/sqlstore/postgressqlstore/provider.go
@@ -2,7 +2,11 @@ package postgressqlstore
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"database/sql"
+	"net/url"
+	"os"
 
 	"github.com/SigNoz/signoz/pkg/errors"
 	"github.com/SigNoz/signoz/pkg/factory"
@@ -50,6 +54,16 @@ func New(ctx context.Context, providerSettings factory.ProviderSettings, config 
 	pgConfig, err := pgxpool.ParseConfig(config.Postgres.DSN)
 	if err != nil {
 		return nil, err
+	}
+
+	// Build TLS config from structured SSL fields if any are set. Empty SSL fields
+	// leave TLS handling to the DSN, preserving the prior behaviour.
+	if config.Postgres.SSLMode != "" || config.Postgres.SSLCert != "" || config.Postgres.SSLKey != "" || config.Postgres.SSLRootCert != "" {
+		tlsConfig, err := buildTLSConfig(config.Postgres)
+		if err != nil {
+			return nil, err
+		}
+		pgConfig.ConnConfig.TLSConfig = tlsConfig
 	}
 
 	// Set the maximum number of open connections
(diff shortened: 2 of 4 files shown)


Harsh23Kashyap:
Heads up: this is from a fork, so the `goci` and `commitci` workflows at `.github/workflows/goci.yaml:9-13` skip until `safe-to-test` is set. Could a maintainer add that label so the CI runs? Happy to address feedback in the meantime.

therealpandey:
Hi @Harsh23Kashyap,
Please get an issue assigned by a maintainer before raising a PR.

PR #12278, opened 2026-07-26
feat(infra-monitoring): filter by pod status

nikhilmantri0902:
## Pull Request

---

### 📄 Summary

Adds a `filterByPodStatus` secondary filter to the v2 infra-monitoring list APIs (pods, nodes, namespaces, clusters, deployments, statefulsets, jobs, daemonsets).

Pod status is a derived kubectl-style value (`k8s.pod.phase` + status reasons, resolved via `argMax`), not a real label, so it can't go through the normal query-builder filter. This PR resolves the full-scope status keyset up-front and intersects it with the metadata + ranked groups, keeping `total` and pagination correct.

- Multi-select: the field is an array, pushed down as `WHERE lower(display_status) IN (...)` (OR within status, AND with the attribute filter).
- When the optional status metrics were never ingested, the endpoint returns a non-blocking warning + empty page instead of silently filtering everything out.

N/A — backend + generated FE API types only; the UI is a separate change.

#### Issues closed by this PR

Part of SigNoz/engineering-pod#5778.

---

### ✅ Change Type

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context

N/A — not a bug fix.

---

### 🧪 Testing Strategy

- Tests added/updated:
  - Unit test for the status push-down (`applyPodStatusFilter`, built with go-sqlbuilder).
  - Integration tests across all 8 entity APIs: list mode, grouped mode, validation, missing-metric warning, and multi-select union.
- Manual verification: smoke-tested against staging data (single, multi, and grouped filters).
- Edge cases covered: missing status metric → warning + empty; grouped mode keeps a group if ≥1 pod matches; multi-select returns the union of the selected statuses.

---

- Blast radius: v2 infra-monitoring list endpoints only.
- Potential regressions: none when the filter is unset (empty = off, fully additive). When set, an extra status query runs; it is gated behind the filter being present.
- Rollback plan: revert the PR — no schema or data migrations involved.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | OSS, Cloud, Enterprise |
| Change Type | Feature |
| Description | v2 infra-monitoring lists can now be filtered by pod status (multi-select). |

---

### 📋 Checklist

- [x] Manually tested

---

## 👀 Notes for Reviewers

- `filterByPodStatus` is optional and additive — no change to existing responses when omitted.
- The status keyset is resolved once at full scope, then intersected — this is what keeps `total`/pagination correct despite status being a post-aggregation value.
- OpenAPI spec + FE API types are regenerated (scalar → array); no hand-written FE.


nikhilmantri0902:
@namanv3 can you review this ?

pkg/types/inframonitoringtypes/containers_constants.go
```diff
@@ -31,6 +35,12 @@ var (
 	ContainerStatusNoData = ContainerStatus{valuer.NewString("no_data")}
 )
 
+// IsFilterableContainerStatus reports whether c is a concrete, user-filterable
+// container status: any Enum() member except the no_data sentinel.
+func IsFilterableContainerStatus(c ContainerStatus) bool {
```
namanv3: can flip this to be 
func (c ContainerStatus) IsFilterable() bool

pkg/types/inframonitoringtypes/nodes_constants.go
```diff
@@ -20,6 +24,12 @@ func (NodeCondition) Enum() []any {
 	}
 }
 
+// IsFilterableNodeCondition reports whether c is a concrete, user-filterable
+// node readiness: any Enum() member except the no_data sentinel.
+func IsFilterableNodeCondition(c NodeCondition) bool {
```
namanv3: same as the container filterable comment

pkg/types/inframonitoringtypes/pods_constants.go
```diff
@@ -66,6 +70,12 @@ func (PodStatus) Enum() []any {
 	}
 }
 
+// IsFilterablePodStatus reports whether s is a concrete, user-filterable pod
+// status: any Enum() member except the no_data sentinel.
+func IsFilterablePodStatus(s PodStatus) bool {
+	return s != PodStatusNoData && slices.Contains((PodStatus{}).Enum(), any(s))
```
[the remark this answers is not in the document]
namanv3: not the biggest concern

pkg/modules/inframonitoring/implinframonitoring/nodes.go
```diff
@@ -192,15 +251,36 @@ func (m *module) getNodesTableMetadata(ctx context.Context, orgID valuer.UUID, r
...
+
 func (m *module) getPerGroupNodeConditionCounts(
 	ctx context.Context,
 	orgID valuer.UUID,
 	start, end int64,
 	filter *qbtypes.Filter,
 	groupBy []qbtypes.GroupByKey,
 	pageGroups []map[string]string,
+	filterByNodeReadiness []inframonitoringtypes.NodeCondition,
 ) (map[string]nodeConditionCounts, error) {
```
namanv3, line 280: will we need a check for existence of `k8s.node.condition_ready` here? or is it covered earlier?

pkg/types/inframonitoringtypes/clusters.go
```diff
@@ -88,6 +97,22 @@ func (req *PostableClusters) Validate() error {
 		return errors.NewInvalidInputf(errors.CodeInvalidInput, "offset cannot be negative")
 	}
 
+	if req.Filter != nil {
```
namanv3, line 100: the two req.filter != nil conditions can be combined, same for all other types files

pkg/modules/inframonitoring/implinframonitoring/containers_test.go
```diff
@@ -0,0 +1,63 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			cb := sqlbuilder.NewSelectBuilder()
+			cb.Select("pod_uid")
+			cb.From("container_status")
+			applyContainerStatusFilter(cb, tt.statuses)
+			sql, args := cb.BuildWithFlavor(sqlbuilder.ClickHouse)
+
+			hasWhere := strings.Contains(sql, "lower(display_status) IN (")
+			if hasWhere != tt.wantWhere {
+				t.Errorf("applyContainerStatusFilter(%v) sql = %q, wantWhere = %v", tt.statuses, sql, tt.wantWhere)
```
namanv3: use assert package instead of t.error

pkg/modules/inframonitoring/implinframonitoring/pods.go
```diff
@@ -303,13 +356,21 @@ func (m *module) getPerGroupPodStatusCounts(
...
 
+	var (
+		filterClause   *sqlbuilder.WhereClause
+		err            error
+		userFilterExpr string
+	)
+
 	// Merge user filter with page-groups IN clauses.
-	userFilterExpr := ""
+	userFilterExpr = ""
```
namanv3: noop line, can remove

pkg/modules/inframonitoring/implinframonitoring/nodes.go
```diff
@@ -114,12 +128,26 @@ func (m *module) getTopNodeGroupsAndMetadata(
 		return err
 	})
 
+	if len(filterByPodStatus) != 0 {
+		g.Go(func() error {
+			var err error
+			statusCounts, statusWarning, err = m.getPerGroupPodStatusCountsWithReqMetricChecks(gCtx, orgID, req.Start, req.End, filter, req.GroupBy, nil, filterByPodStatus)
```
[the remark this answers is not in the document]
nikhilmantri0902: This is expected behavior. The full-scope pod-status scan runs only when `filterByPodStatus` is set, and in that case the request's filter expression is applied consistently to both the entity (node) scan and the pod-status scan — the feature is "entities matching the filter that also have a pod in the requested status", so the attribute filter and the status filter are ANDed.

A consequence is that if the filter references a label not present on the pod series (e.g. `k8s.node.uid`), the pod-status scan matches nothing and the entity is excluded under that AND. The attributes actually used to filter these tabs — `k8s.node.name`, `k8s.namespace.name`, `k8s.cluster.name` — are present on the pod series and behave as intended, so we're accepting the stricter behavior for entity-only labels rather than special-casing the scan.

pkg/types/inframonitoringtypes/pods_constants.go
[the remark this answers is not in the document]
nikhilmantri0902: deferring this, will be under frontend control

pkg/types/inframonitoringtypes/containers_constants.go
nikhilmantri0902: done

pkg/types/inframonitoringtypes/nodes_constants.go
nikhilmantri0902: done

pkg/modules/inframonitoring/implinframonitoring/nodes.go
nikhilmantri0902: nope not needed, this kind of a warning output like pod status is only needed for queries depending upon multiple different metrics at once, because u need all to compute the right status, one or 2 can't do. 

But since here there is only one metric, does not need a warning. 

BTW still, this is indicated to the user in instrumentation checks API on the UI.

pkg/types/inframonitoringtypes/clusters.go
nikhilmantri0902: done

pkg/modules/inframonitoring/implinframonitoring/containers_test.go
nikhilmantri0902: done

pkg/modules/inframonitoring/implinframonitoring/pods.go
nikhilmantri0902: done

pkg/modules/inframonitoring/implinframonitoring/nodes.go
namanv3: then why do we have `getPerGroupPodStatusCountsWithReqMetricChecks`? Why does that need a check?

nikhilmantri0902: Becase for computing pod status you need 3 metrics at once - pod.phase, pod.status.reason, container.status.reason and the last 2 need to be explicitly enabled in the receivers. So a warning helps there. node.condition_ready is enabled by default.

namanv3 (approved):

therealpandey (approved):

PR #12281, opened 2026-07-26
fix(alertmanager): resolve email SMTP settings from env

srikanthccv:
## Pull Request

Follow up to https://github.com/SigNoz/platform-pod/issues/2135, to read from the updated env instead of the previously persisted data

Added data migration to correct already existing data.

Fixes https://github.com/SigNoz/engineering-pod/issues/4660


pkg/types/alertmanagertypes/config.go
```diff
@@ -279,16 +348,21 @@ func (c *Config) CreateReceiver(receiver *Receiver) error {
...
 		return err
 	}
 
 	c.alertmanagerConfig.Route.Routes = append(c.alertmanagerConfig.Route.Routes, route)
-	c.alertmanagerConfig.Receivers = append(c.alertmanagerConfig.Receivers, *receiver.Receiver)
-	c.setCustomConfigs(receiver)
+	c.alertmanagerConfig.Receivers = append(c.alertmanagerConfig.Receivers, *owned.Receiver)
+	c.setCustomConfigs(owned)
 
-	if err := c.alertmanagerConfig.UnmarshalYAML(func(i interface{}) error { return nil }); err != nil {
```
srikanthccv, line 291: This noop unmarshal is for skip decoding, run the post decoding body against whatever the fields already hold. `Config.UnmarshalYAML`'s body is the global fill-if-empty and it's a method on `c.alertmanagerConfig`, this in next step gets flushed which in turn gets persisted.

It iterates `c.Receivers` and writes `Global.SMTPSmarthost`, `SMTPAuthPassword`, etc... directly into every empty email config in the `receiver.Receiver` channels slice.  Was the intent just the "validate the config I just modified"? Or was it also "validate and mutate by filling empty values", regardless, this function is validation and resolution in one step, so you can't get the validation without also getting the materialization.

srikanthccv: the `append(..., *receiver.Receiver)` dereferences, it's a shallow struct copy, and the `Receiver.EmailConfigs` is `[]*EmailConfig`, pointing to the same underlying, so caller receiver is updated too.

pkg/alertmanager/alertmanagerserver/server.go
```diff
@@ -241,9 +241,12 @@ func (server *Server) PutAlerts(ctx context.Context, postableAlerts alertmanager
 }
 
 func (server *Server) SetConfig(ctx context.Context, alertmanagerConfig *alertmanagertypes.Config) error {
-	config := alertmanagerConfig.AlertmanagerConfig()
+	resolved, err := alertmanagerConfig.Resolved()
```
srikanthccv, line 244: the resolved config has the globals populated, since the global is from the config, there is nothing specifically about env or config file based setup

pkg/types/alertmanagertypes/receiver.go
```diff
@@ -37,6 +37,7 @@ func NewReceiver(input string) (*Receiver, error) {
 	if err != nil {
 		return nil, err
 	}
+	stripEmailTransport(withDefaults)
```
srikanthccv, line 40: we will not support per-receiver smtp for now, as and when it is needed we will expose those fields as part of the channel itself so people can configure and be the responsible to update if they want to rotate.

pkg/sqlmigration/106_scrub_email_channel_transport.go
```diff
@@ -0,0 +1,263 @@
...
+		if globalRaw, ok := cfg["global"]; ok && string(globalRaw) != "null" {
+			global := make(map[string]json.RawMessage)
+			if err := json.Unmarshal(globalRaw, &global); err != nil {
+				migration.logger.WarnContext(ctx, "skipping alertmanager config with unreadable global", slog.String("config_id", row.ID), errors.Attr(err))
+				continue
+			}
+			if deleteKeys(global, globalSMTPKeys) {
+				newGlobal, err := json.Marshal(global)
+				if err != nil {
+					return err
```
therealpandey: Do we want to hard fail here or just log and move on?

pkg/sqlmigration/106_scrub_email_channel_transport.go
```diff
@@ -0,0 +1,263 @@
...
+			if err := json.Unmarshal(receiversRaw, &receivers); err != nil {
+				migration.logger.WarnContext(ctx, "skipping alertmanager config with unreadable receivers", slog.String("config_id", row.ID), errors.Attr(err))
+				continue
+			}
+
+			receiversChanged := false
+			for _, receiver := range receivers {
+				scrubbed, err := scrubEmailConfigs(receiver)
+				if err != nil {
+					return err
```
therealpandey: Do we want to hard fail here or just log and move on?

pkg/sqlmigration/106_scrub_email_channel_transport.go
```diff
@@ -0,0 +1,263 @@
...
+				}
+				receiversChanged = receiversChanged || scrubbed
+			}
+
+			if receiversChanged {
+				newReceivers, err := json.Marshal(receivers)
+				if err != nil {
+					return err
```
therealpandey: Do we want to hard fail here or just log and move on?

pkg/types/alertmanagertypes/config.go
```diff
@@ -194,6 +201,28 @@ func newRawFromConfig(c *config.Config, customConfigs map[string]customReceiverC
...
+	stripped.SMTPFrom = ""
+	stripped.SMTPHello = ""
+	stripped.SMTPSmarthost = config.HostPort{}
+	stripped.SMTPAuthUsername = ""
+	stripped.SMTPAuthPassword = ""
+	stripped.SMTPAuthPasswordFile = ""
+	stripped.SMTPAuthSecret = ""
+	stripped.SMTPAuthSecretFile = ""
+	stripped.SMTPAuthIdentity = ""
+	stripped.SMTPRequireTLS = false
```
therealpandey: Do this cause issues? Especially with merging true/false here? The default has been false and hopefully setting is to true would happen transparently?

tests/integration/tests/alertmanager/01_channels.py
```diff
@@ -0,0 +1,268 @@
+import time
```
therealpandey, line 1: Can alertmanager be combined with the alerts test suite? Or is the alerts test suite really the rules test suite?

therealpandey (changes requested):

therealpandey:
We have supported all alertmanager global configs https://github.com/prometheus/alertmanager/blob/846bdbffb8d716d7014f2aed7c788b2e1bbdc8cb/config/config.go#L830.

What should we do about the other configs apart from SMTP?

pkg/sqlmigration/106_scrub_email_channel_transport.go
srikanthccv: this is marshalling the data that was just previously unmarshalled with some keys deleted, so theoretically it should never be possible to have an error here.

srikanthccv: will change the rest of the errors to log and move instead

tests/integration/tests/alertmanager/01_channels.py
srikanthccv: > is the alerts test suite really the rules test suite

this, will rename in separate follow up

pkg/types/alertmanagertypes/config.go
srikanthccv: yes no issue, the false value drops field during marshalling and the value of this is always related on `SetGlobalConfig`

pkg/types/alertmanagertypes/receiver.go
```diff
@@ -53,6 +54,23 @@ func NewReceiver(input string) (*Receiver, error) {
 	return receiver, nil
 }
 
+func stripEmailTransport(base *config.Receiver) {
```
therealpandey, line 57: @srikanthccv should we think of making this stripGlobalConfig where we reflect on the global struct and strip anything global not just smtp?

It also guards us from future upgrades to the alertmanager when they add more fields to SMTP?

therealpandey (changes requested):

pkg/types/alertmanagertypes/receiver.go
srikanthccv: let's do that in follow up PR (along with the naming change mentioned here https://github.com/SigNoz/signoz/pull/12281#discussion_r3699834484)?

and for SMTP specifically, there is a dedicated integration test (smtp_rotation) that rotates the env and asserts the emails are still delivered.

therealpandey (approved):

PR #12286, opened 2026-07-26
fix(dashboard): variable value hover reveals, readable collapsed-bar tooltips, and Noz in the panel editor

AshwinBhatkal:
## Pull Request

---

### 📄 Summary

Four dashboard-V2 fixes around variables and hover reveals, one commit per concern.

1. **A variable's options no longer blink empty mid-refetch.** A sibling dynamic's change bumps the variable's `cycleId` (part of its react-query key), so the new key started with no data — and a dynamic `ALL` selection is rendered *from* the option list (its value stays `null` since ALL travels as the `__all__` sentinel), so it fell back to the `Select value` placeholder. Both fetched-variable queries now keep their previous data across the cycle bump; V1 got the same stickiness from local state.
2. **A truncated multi-select tag reveals its full value on hover.** rc-select sets a `title` on its own tags but not once a custom `tagRender` is in play — which `CustomMultiSelect` always supplies, so a tag cut by `maxTagTextLength` was unreadable. Now every tag carries the reveal.
3. **Hover reveals read as data, not strings.** The variable pill's `+N` lists every hidden value (bullets); the collapsed variables bar's `+N` puts `$name` muted above its value; the dashboard tags `+N` renders the same chips used inline, one per line. Nothing is capped — each body scrolls at 480px with the scrollbar pinned visible, so a long list stays reachable and obviously scrollable.
4. **Noz is reachable from the panel editor**, which is a chromeless route and so loses the side nav's entry point.

_To add: variable pill tag hover, pill `+N` hover, collapsed bar `+N`, tags `+N`, panel editor header._

#### Issues closed by this PR

Closes https://github.com/SigNoz/pulse-pod/issues/143

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause

- **Placeholder flash:** `cycleId` is keyed into each variable's react-query key so a bump cancels stale requests. Without `keepPreviousData` the new key renders with `data: undefined`, and a `DYNAMIC` ALL selection derives its display purely from the options — `materializeAll` deliberately leaves its value `null`.
- **No hover reveal:** rc-select only sets `title` in `defaultRenderSelector`; `CustomMultiSelect` always supplies a custom `tagRender`, and each caller's `+N` placeholder was a plain string here.
- **Unreadable tooltips:** the bodies were built as `name: value` strings / `join(', ')`.
- **Missing Noz:** `AppLayout` marks `DASHBOARD_PANEL_EDITOR` as `renderFullScreen`, which hides `SideNav` — the surface that carries `aiAssistantMenuItem`.

#### Fix Strategy

Root-cause fixes at the layer that owns each behaviour: query options for the flash, the shared select for the per-tag reveal, per-caller components for the `+N` bodies, and the editor header for Noz.

The `+N` overflow is deliberately **not** handled inside `CustomMultiSelect`: V1 variables (`renderMaxTagPlaceholder`) and Messaging Queues / Celery (`SelectMaxTagPlaceholder`) already wrap their own placeholder in an antd tooltip, and a shared one would stack a second tooltip on top of theirs.

---

### 🧪 Testing Strategy

- **Tests added:** `useFetchedVariableOptions` (options survive a `cycleId` bump — verified failing without the fix), `CustomMultiSelect` tag tooltip (reveal on hover; `+N` left to the caller), `ValueSelector` (tag / `+N` / nothing elsewhere), `HiddenVariablesTooltip` + `describeSelection`, `TagsOverflowTooltip`, `PanelEditor/Header` (Noz present when enabled, absent when not).
- **Suites run:** the full frontend suite (~7,240 tests) plus focused runs of `DashboardPageV2`, `DashboardsListPageV2`, `components/NewSelect` and V1 `DashboardContainer` (1,454 tests, all green). `tsgo --noEmit`, `oxlint`, `oxfmt --check`, `stylelint` clean.
- **Baseline comparison:** the full suite is flaky under parallel load — it fails *more* suites on `main` without this branch (5) than with it (2–3, incl. SIGSEGV worker crashes). The one suite that failed in a full run (`QuerySearch.contextSuggestions`) passes in isolation both with and without these changes.
- **Edge cases:** ALL selections (dynamic vs query), all-options-selected rendering as ALL, empty/unset selections, freeform values absent from the options, uncapped value/variable/tag lists.
- **Pending:** manual browser pass — hence `safe-to-e2e` on this PR.

---

- **Blast radius:** commits 3–7 are inside dashboard V2 (behind `use_dashboard_v2`). The one shared-code change is the tag tooltip in `CustomMultiSelect` (commit 2), which **every** consumer now gets — V1 variables, the query builder, alerts, routing policies, Messaging Queues, Celery.
- **Potential regressions:** the tag tooltip is additive (hover-only; no layout, value or callback change), and the component supplies its **own** `TooltipProvider` so it cannot throw `Tooltip must be used within TooltipProvider` in a consumer rendered outside the app-level provider — that failure mode was caught by the existing `NewSelect` suites and fixed before this landed. The `+N` overflow is untouched for every existing caller, so no double tooltips. Commit 1 skips the V1 dashboards e2e specs, which fail against V2 until ported.
- **Rollback plan:** commits are independent by concern (only the pill `+N` reveal pairs naturally with the tag reveal), so any one can be reverted alone. `use_dashboard_v2` remains the kill switch for the V2 surface.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Dashboard variables no longer flash "Select value" while a related variable updates, hovering a truncated value or `+N` now reveals the full selection, and the collapsed variables / dashboard tags tooltips are easier to read. |

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

- **Release-day call:** `feat(NewSelect): reveal a tag's full value on hover` is the only commit touching code V1 executes. Everything else is V2-only. Dropping it (and optionally the pill `+N` commit that pairs with it) leaves the rest intact.
- The `+N` tooltips carry an inner `title=""`: rc-select copies the placeholder node into the overflow wrapper's `title`, which would otherwise surface a native `[object Object]` tooltip next to the real one.
- Hover reveals are tooltips, so reaching the scrollbar means crossing from the trigger into the content — radix's hoverable content bridges that gap, but a `Popover` would be the sturdier home if scrolling proves fiddly in review.

### DIFF
--- a/frontend/src/components/TooltipScrollArea/TooltipScrollArea.module.scss
+++ b/frontend/src/components/TooltipScrollArea/TooltipScrollArea.module.scss
@@ -1,0 +1,17 @@
+@use '../../styles/scrollbar' as *;
+
+// How tall a hover reveal grows before its content starts scrolling.
+.scrollArea {
+	max-height: 480px;
+	overflow-y: auto;
+	// Keep the page behind the tooltip still once the list hits its end.
+	overscroll-behavior: contain;
+
+	// Always-visible thin scrollbar: a tooltip is transient, so an auto-hiding one
+	// would leave no hint that there is more below. Firefox has no ::-webkit rules,
+	// hence the standard properties alongside the mixin.
+	scrollbar-width: thin;
+	scrollbar-color: var(--l3-background) transparent;
+
+	@include custom-scrollbar;
+}
--- a/frontend/src/components/TooltipScrollArea/TooltipScrollArea.module.scss
+++ b/frontend/src/components/TooltipScrollArea/TooltipScrollArea.module.scss
@@ -1,17 +1,20 @@
 @use '../../styles/scrollbar' as *;
 
+// Padding for the tooltip body that hosts a scroll area: the right side is given
+// up so the scrollbar can sit flush against the tooltip's edge instead of floating
+// inset from it. `.scrollArea` puts that spacing back between the text and the bar.
+.tooltipContent {
+	--tooltip-padding: var(--spacing-2) 0 var(--spacing-2) var(--spacing-4);
+}
+
 // How tall a hover reveal grows before its content starts scrolling.
 .scrollArea {
 	max-height: 480px;
 	overflow-y: auto;
 	// Keep the page behind the tooltip still once the list hits its end.
 	overscroll-behavior: contain;
-
-	// Always-visible thin scrollbar: a tooltip is transient, so an auto-hiding one
-	// would leave no hint that there is more below. Firefox has no ::-webkit rules,
-	// hence the standard properties alongside the mixin.
-	scrollbar-width: thin;
-	scrollbar-color: var(--l3-background) transparent;
+	// Gap between the content and the scrollbar (or the tooltip edge when short).
+	padding-right: var(--spacing-4);
 
 	@include custom-scrollbar;
 }
--- a/frontend/src/components/NewSelect/CustomMultiSelect.tsx
+++ b/frontend/src/components/NewSelect/CustomMultiSelect.tsx
@@ -20,6 +20,7 @@ import {
 import { Color } from '@signozhq/design-tokens';
 import { Button, Select } from 'antd';
 import { Checkbox } from '@signozhq/ui/checkbox';
+import { TooltipProvider, TooltipSimple } from '@signozhq/ui/tooltip';
 import { Typography } from '@signozhq/ui/typography';
 import cx from 'classnames';
 import TextToolTip from 'components/TextToolTip/TextToolTip';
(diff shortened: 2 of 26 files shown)


frontend/src/components/TooltipScrollArea/TooltipScrollArea.module.scss
```diff
@@ -0,0 +1,28 @@
...
+.scrollArea {
+	max-height: 480px;
+	overflow-y: auto;
+	// Keep the page behind the tooltip still once the list hits its end.
+	overscroll-behavior: contain;
+
+	// Thin scrollbar, always rendered so a capped list reads as scrollable.
+	$thumb: color-mix(in srgb, var(--bg-vanilla-100) 32%, transparent);
+
+	scrollbar-width: thin;
```
ahrefabhi: can we use custom-scrollbar mixin here?
frontend/src/styles/_scrollbar.scss

ahrefabhi (dismissed):

ahrefabhi (approved):

AshwinBhatkal:
@therealpandey e2e green

PR #12095, opened 2026-07-13
feat(authz): enable FGA for telemetry resources

vikrantgupta25:
## Summary

- enable FGA for telemetry resources

closes: https://github.com/SigNoz/platform-pod/issues/2682
closes: https://github.com/SigNoz/platform-pod/issues/2606

PR #12294, opened 2026-07-27
fix(sqlmigration): write integer operation in migration 101 PG branch

### LINKED ISSUE #12293: [authz] migration 101_add_telemetry_tuples inserts "TUPLE_OPERATION_WRITE" into an INTEGER column, crash-looping v0.134.0 on PostgreSQL
## Bug description

Migration `101_add_telemetry_tuples` (new in v0.134.0) fails on PostgreSQL because its Postgres branch inserts the Go string literal `"TUPLE_OPERATION_WRITE"` into `changelog.operation`, which is an `INTEGER NOT NULL` column:

https://github.com/SigNoz/signoz/blob/v0.134.0/pkg/sqlmigration/101_add_telemetry_tuples.go#L106

```go
storeID, tuple.objectType, objectID, tuple.relation, user, "TUPLE_OPERATION_WRITE", tupleID, now,
```

The SQLite branch of the same function passes the correct integer 26 lines below:

https://github.com/SigNoz/signoz/blob/v0.134.0/pkg/sqlmigration/101_add_telemetry_tuples.go#L132

```go
storeID, tuple.objectType, objectID, tuple.relation, "role", roleSubject, "assignee", 0, tupleID, now,
```

`changelog.operation` is deliberately `INTEGER` — migration `090_fix_changelog_operation_type` changed it to that on purpose, with the comment "changelog.operation is TEXT, should be INTEGER (OpenFGA v1.14.0 passes int32)".

Startup fails with:

```json
{"level":"ERROR","msg":"starting sqlstore migrations","dialect":"pg"}
{"level":"ERROR","msg":"failed to create signoz","exception.type":"internal","exception.message":"ERROR: invalid input syntax for type integer: \"TUPLE_OPERATION_WRITE\" (SQLSTATE 22P02)"}
```

**This prevents SigNoz from coming up at all on Kubernetes.** The migration runs inside a transaction, so it rolls back and re-fails identically on every restart — the pod goes into `CrashLoopBackOff` permanently rather than eventually recovering. The Deployment therefore never becomes Ready, and the rollout fails:

```
Error: UPGRADE FAILED: resource Deployment/<ns>/<release>-signoz not ready. status: Failed, message: Progress deadline exceeded
```

Any `helm upgrade --wait` or `kubectl rollout status` gated on SigNoz becoming healthy will time out and fail the deployment. There is no way to complete an upgrade to v0.134.0 short of pinning back to v0.133.0 or hand-editing the database.

Scope of who is affected:

- **PostgreSQL metadata stores only** (`SIGNOZ_SQLSTORE_PROVIDER=postgres`). SQLite is unaffected: its branch passes `0`, and SQLite's dynamic typing would accept the string in an `INTEGER` column regardless. This is likely why it wasn't caught.
- **Existing installations only.** On a fresh database the `organizations` table is empty when migrations run, so the loop body never executes. Triggering it requires at least one organization *and* at least one of the 13 telemetry tuples to be missing, since the changelog insert is only reached when the preceding `tuple` insert returns `rowsAffected == 1`.
- **v0.134.0 is the first affected release.** Migration 101 does not exist in v0.133.0.

## Expected behavior

The Postgres branch of migration 101 should write the integer OpenFGA `TupleOperation` value — `0`, or explicitly `int32(openfgav1.TupleOperation_TUPLE_OPERATION_WRITE)` — exactly as the SQLite branch already does, so the migration commits and SigNoz starts normally after the upgrade.

## How to reproduce

1. Run SigNoz v0.132.x or v0.133.0 with `SIGNOZ_SQLSTORE_PROVIDER=postgres`. Let it start and provision an organization, so `organizations` has a row and the `tuple` table holds the older `telemetryresource` set — i.e. at least one of the 13 tuples migration 101 inserts is still missing. (In our case the `signoz-editor` and `signoz-viewer` `meter-metrics` tuples were the two absent ones.)
2. Upgrade to v0.134.0.
3. On startup, sqlstore migrations run, 099 and 100 commit, and 101 fails with `ERROR: invalid input syntax for type integer: "TUPLE_OPERATION_WRITE" (SQLSTATE 22P02)`. The process exits and repeats on every restart; on Kubernetes the pod enters `CrashLoopBackOff` and the Deployment never becomes Ready.

The failing statement can be confirmed in isolation against the same database. Running migration 101's Postgres insert verbatim reproduces the error, while the identical statement with `0` in the `operation` position succeeds:

```
BEGIN
ERROR:  invalid input syntax for type integer: "TUPLE_OPERATION_WRITE"

BEGIN
INSERT 0 1
```

This is not a PostgreSQL version quirk — the same rejection occurs on PostgreSQL 16.14 and 18.4 alike.

## Version information

* **Signoz version**: v0.134.0 (`signoz/signoz:v0.134.0`, enterprise variant), upgraded from v0.132.2. Metadata store: PostgreSQL 18.4 via `SIGNOZ_SQLSTORE_PROVIDER=postgres`.
* **Browser version**: N/A — backend crash on startup, the UI never becomes reachable.
* **Your OS and version**: Kubernetes (GKE), stock `signoz/signoz` container image.
* **Your CPU Architecture**(ARM/Intel): x86_64 / amd64 (Intel)

## Additional context

Related but distinct, and already fixed: openfga/openfga#3011 and openfga/openfga#3012 covered the same enum-serialized-as-string problem in OpenFGA's own changelog insert path (`pkg/storage/sqlcommon/sqlcommon.go`), resolved by casting to `int32`. That fix is present in openfga v1.14.1, which SigNoz v0.134.0 pins, so the runtime write path is correct. This report is a separate, hardcoded string literal in SigNoz's own migration and is not covered by those fixes.

The line is unchanged on `main` as of this report.

jesse-schein:
## Pull Request

---

### 📄 Summary

Migration `101_add_telemetry_tuples` (new in v0.134.0) inserts the Go string literal `"TUPLE_OPERATION_WRITE"` into `changelog.operation` in its PostgreSQL branch. That column is `INTEGER NOT NULL`, so PostgreSQL rejects the statement:

```
ERROR: invalid input syntax for type integer: "TUPLE_OPERATION_WRITE" (SQLSTATE 22P02)
```

The column type is deliberate — migration `090_fix_changelog_operation_type` recreated it as `INTEGER` on purpose, with the comment *"changelog.operation is TEXT, should be INTEGER (OpenFGA v1.14.0 passes int32)"*.

Because the migration runs inside a transaction, it rolls back and re-fails identically on every restart, so SigNoz never starts:

```json
{"level":"ERROR","msg":"starting sqlstore migrations","dialect":"pg"}
{"level":"ERROR","msg":"failed to create signoz","exception.type":"internal","exception.message":"ERROR: invalid input syntax for type integer: \"TUPLE_OPERATION_WRITE\" (SQLSTATE 22P02)"}
```

On Kubernetes the pod lands in permanent `CrashLoopBackOff`, the Deployment never becomes Ready, and any `helm upgrade --wait` or `kubectl rollout status` gated on it fails with `Progress deadline exceeded`. The only workarounds are pinning back to v0.133.0 or hand-editing the database.

The fix passes the integer `TupleOperation` value `0`, exactly as the non-PG branch of the same function already does 26 lines below.

#### Issues closed by this PR

Closes #12293

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause

A hardcoded string literal where an integer enum value was required, introduced with migration 101 in #12095 and unchanged on `main`.

`changelog.operation` holds OpenFGA's `TupleOperation` enum, which the driver passes as `int32`. The two branches of migration 101's insert loop disagreed about how to express it:

```go
// PG branch — line 106, string literal:
storeID, tuple.objectType, objectID, tuple.relation, user, "TUPLE_OPERATION_WRITE", tupleID, now,

// non-PG branch — line 132, correct integer:
storeID, tuple.objectType, objectID, tuple.relation, "role", roleSubject, "assignee", 0, tupleID, now,
```

Why this wasn't caught earlier:

- **SQLite is unaffected.** Its branch already passes `0`, and SQLite's dynamic typing would accept the string in an `INTEGER` column anyway.
- **Fresh databases are unaffected.** `organizations` is empty when migrations run, so the loop body never executes.
- **Only existing installations trip it** — those with at least one organization *and* at least one of the 13 telemetry tuples missing, since the `changelog` insert is only reached when the preceding `tuple` insert returns `rowsAffected == 1`.
- Migration 101 is the first migration ordered *after* 090 that writes `changelog.operation`, so it is the first to meet the `INTEGER` column.

#### Fix Strategy

Pass `0` instead of the string, matching the non-PG branch of the same function and every sibling migration's non-PG branch. One line, no behavior change beyond the corrected value — `0` is `TUPLE_OPERATION_WRITE`, so the row written is what the migration always intended.

Happy to switch to an explicit `int32(openfgav1.TupleOperation_TUPLE_OPERATION_WRITE)` if reviewers prefer self-documenting over consistency with the surrounding code; it would add an import to this file for a single literal.

---

### 🧪 Testing Strategy

- **Tests added/updated:** none. Migration 101 has no existing test coverage, and the repo has no PostgreSQL-backed sqlmigration test harness to hook into — the defect is a wrong literal in a SQL parameter list, invisible to Go-level unit tests and only observable against a real PG instance. Guidance welcome if maintainers want a harness added here.
- **Manual verification:** `go build` and `go vet` pass on `./pkg/sqlmigration/`. Migration 101's PG insert was run verbatim against affected PostgreSQL 16.14 and 18.4 instances and reproduces `22P02`; the identical statement with `0` in the `operation` position returns `INSERT 0 1`. A full v0.133.0 → patched-v0.134.0 upgrade against a real affected database is being validated now, and I'll report back here.
- **Edge cases covered:** re-running the migration is still a no-op — the `ON CONFLICT (store, ulid, object_type) DO NOTHING` clause and the `rowsAffected == 0 → continue` guard are untouched.

---

- **Blast radius:** one line in one migration, on the PostgreSQL path only. The migration currently cannot succeed on PostgreSQL, so there is no working behavior to regress. SQLite is untouched.
- **Potential regressions:** none identified. Installations that already applied 101 successfully are by definition SQLite or had nothing to insert; neither is affected by this change.
- **Rollback plan:** revert the commit. Note that anyone who already hit this bug has no committed rows from 101 (the transaction rolled back), so the fixed migration applies cleanly with no manual cleanup.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | OSS / Enterprise (self-hosted, PostgreSQL metadata store) |
| Change Type | Bug Fix |
| Description | Fixed a migration that prevented v0.134.0 from starting on PostgreSQL, crash-looping upgrades from v0.133.0 and earlier. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

**Seven sibling migrations carry the same string literal, and I deliberately left them alone** — 061, 062, 063, 075, 078, 081, 083. They are currently inert, because all of them are ordered *before* 090, and 090 drops and recreates `changelog` outright rather than casting the column, so any string rows they wrote are discarded rather than failing a type conversion. Migration 101 is the only one that runs after 090 and therefore the only one that actually breaks.

They are still a latent footgun: the next migration written by copy-pasting one of them will reintroduce this bug. Worth a follow-up sweep, or a shared constant for the operation value — happy to fold either into this PR if you'd rather fix it all at once.

For context on the related-but-distinct upstream issue: openfga/openfga#3011 and openfga/openfga#3012 fixed the same enum-as-string problem in OpenFGA's own changelog insert path by casting to `int32`. That fix ships in openfga v1.14.1, which v0.134.0 already pins, so the *runtime* write path is correct. This is purely SigNoz's own hardcoded literal in a migration and is not covered by those fixes.

### DIFF
--- a/pkg/sqlmigration/101_add_telemetry_tuples.go
+++ b/pkg/sqlmigration/101_add_telemetry_tuples.go
@@ -103,7 +103,7 @@ func (migration *addTelemetryTuples) Up(ctx context.Context, db *bun.DB) error {
 					INSERT INTO changelog (store, object_type, object_id, relation, _user, operation, ulid, inserted_at)
 					VALUES (?, ?, ?, ?, ?, ?, ?, ?)
 					ON CONFLICT (store, ulid, object_type) DO NOTHING`,
-					storeID, tuple.objectType, objectID, tuple.relation, user, "TUPLE_OPERATION_WRITE", tupleID, now,
+					storeID, tuple.objectType, objectID, tuple.relation, user, 0, tupleID, now,
 				)
 				if err != nil {
 					return err


therealpandey (approved):

therealpandey:
Hi @jesse-schein,
It looks like Github has changed their policy and it's breaking the CI.

We're going to look into what has changed here.

We appreciate the contribution. However, at this point, we'll raise one internally and merge this.

jesse-schein:
> Hi @jesse-schein, It looks like Github has changed their policy and it's breaking the CI.
> 
> We're going to look into what has changed here.
> 
> We appreciate the contribution. However, at this point, we'll raise one internally and merge this.

Yes I see that. It seems like github just changed this a afew days ago.

https://github.blog/changelog/2026-06-18-safer-pull_request_target-defaults-for-github-actions-checkout/

They noted this:


Editor’s note (July 15, 2026): We updated this post to reflect a revised backport enforcement date. Enforcement for backported versions of actions/checkout has been moved from July 16, 2026 to Monday, July 20, 2026. We also clarified the scope of the backport. V1 of actions/checkout will not receive this change. The security update will be backported to all other supported versions of actions/checkout.

PR #12301, opened 2026-07-27
feat(querier): log user-authored clickhouse sql after running them through the parser

therealpandey:
User-authored ClickHouse SQL reaches the telemetry store unchecked. That connection is shared with the write paths and, by default, is a privileged account.

1. `querybuilder.ErrIfStatementIsNotValid` — rejects unless the statement is a single SELECT with no table functions, no `system` / `information_schema` reads, and no `readonly` override. Runs on the rendered query, because substituted variable values are user input too.

2. `querybuilder.LogIfStatementIsNotValid` — wraps it, logs a warning with the query, does not fail the request.

## Why it logs rather than rejects

The parser has gaps against SQL ClickHouse accepts. All gaps are pinned in `TestErrIfStatementIsNotValid_ShouldPassButFails` with the construct each stops at and its accepted rewrite, so an upstream fix fails that test.

## Not covered

- **v3 and v4**: `POST /api/v3|v4/query_range` still take `chQueries` and execute them via `runClickHouseQueries` (`pkg/query-service/app/querier/querier.go:269`, `pkg/query-service/app/querier/v2/querier.go:274`) with no check. Own advisories. Both queriers already hold a `*slog.Logger`, so the same call is one line at each site.
- **Enforcement**: nothing is blocked. `TODO(@therealpandey)` in `LogIfStatementIsNotValid`.

closes https://github.com/SigNoz/engineering-pod/issues/5716


pkg/querybuilder/clickhouse_sql_test.go
```diff
@@ -0,0 +1,154 @@
...
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			err := ErrIfStatementIsNotValid(testCase.query)
+			assert.Error(t, err)
+		})
+	}
+}
+
+// Queries the parser cannot read. ClickHouse runs all of them.
+func TestErrIfStatementIsNotValid_ShouldPassButFails(t *testing.T) {
```
therealpandey, line 98: @srikanthccv here is the list right now.

tushar-signoz (approved):

srikanthccv (approved):

PR #12304, opened 2026-07-27
refactor(querier): 3-layer per-signal query architecture

tushar-signoz:
Splits the five `telemetry<signal>` packages — which each mixed schema primitives, SQL generation, and key/value resolution — into three per-signal layers, and centralizes query-stack assembly in `signoz.go`. Behavior-preserving: the relocated SQL-fixture tests pass unchanged.

### What
- **Primitives** → `pkg/telemetryschema/<signal>telemetryschema/` (const + table-selection helpers, field_mapper, condition_builder, trace helpers). Leaf layer; ~22 consumers point here.
- **SQL generation** → `pkg/statementbuilder/<signal>statementbuilder/`, each with a cohesive `New()` that internalizes FieldMapper → ConditionBuilder → AggExprRewriter. `telemetryresourcefilter` moves here as `statementbuilder/resourcefilter` (it's a subquery strategy, not schema).
- **Resolution** → `telemetrymetadata` keeps its role; `NewTelemetryMetaStore` collapses from 24 args to `(settings, telemetrystore, flagger)`, sourcing table names from the schema constants.
- `signoz.go::newQueryStack` builds **one** metadata store + a `statementbuilder.Builders` bundle + the bucket cache once, then injects them — removing the duplicate metadata store `signozquerier` used to build (resolves the standing `// TODO: consolidate`). `signozquerier` is now a thin `querier.New(*statementbuilder.Builders)` adapter.
- `SkipResourceFingerprint` moves from `querier.Config` to a new `statementbuilder.Config`. `Querier` interface moves into `querier.go` (`interfaces.go` removed); adds `pkg/querier/queriertest.MockQuerier`.

### Guardrails
- Dependency direction is one-way and cycle-free: `telemetryschema` is a leaf; both `statementbuilder` and `telemetrymetadata` depend on it, never the reverse.
- Query-range v5 red lines preserved: FieldMapper → ConditionBuilder → AggExprRewriter → StatementBuilder stay distinct composable layers; only instantiation sites moved.

### Notes
- ⚠️ Config key moves `querier.skip_resource_fingerprint` → `statementbuilder.skip_resource_fingerprint`. Not present in shipped config; flag if any deployment sets it in the field (a one-release mapstructure alias can be added if needed).

### Testing
- `go build ./...` and `go vet ./...` clean; `make go-build-community` builds both binaries.
- `make go-lint` (golangci-lint v2.6.0): 0 issues.
- All relocated schema + statement-builder SQL-assertion tests pass, plus `querier`, `query-service/rules`, `telemetrymetadata`, `transition`.

Fixes https://github.com/SigNoz/engineering-pod/issues/5813


pkg/statementbuilder/statementbuilder.go
```diff
@@ -0,0 +1,59 @@
...
+	Trace         qbtypes.StatementBuilder[qbtypes.TraceAggregation]
+	TraceOperator qbtypes.TraceOperatorStatementBuilder
+	Log           qbtypes.StatementBuilder[qbtypes.LogAggregation]
+	Audit         qbtypes.StatementBuilder[qbtypes.LogAggregation]
+	Metric        qbtypes.StatementBuilder[qbtypes.MetricAggregation]
+	Meter         qbtypes.StatementBuilder[qbtypes.MetricAggregation]
+}
+
+// NewFactory returns a provider factory that builds the per-signal statement builders once.
+func NewFactory(
```
therealpandey: The instantiation is never here to prevent circular dependencies. This is only for interface declaration

therealpandey (changes requested):

pkg/statementbuilder/logsstatementbuilder/new.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/SigNoz/signoz/pkg/querybuilder"
+	"github.com/SigNoz/signoz/pkg/telemetryschema/logstelemetryschema"
+	"github.com/SigNoz/signoz/pkg/telemetrystore"
+	qbtypes "github.com/SigNoz/signoz/pkg/types/querybuildertypes/querybuildertypesv5"
+	"github.com/SigNoz/signoz/pkg/types/telemetrytypes"
+)
+
+// New builds the logs statement builder, internalizing the FieldMapper,
+// ConditionBuilder, and AggExprRewriter.
+func New(
```
therealpandey: This needs to be returning a factory of statementbuider, config.

therealpandey (changes requested):

therealpandey (approved):

PR #12307, opened 2026-07-27
fix(authz): app routing and sidebar fixes when no built-in role

H4ad:
## Pull Request

---

This PR fixes two issues when the user does not have built-in role:

- Sidebar does not load any item due to userPreference throw no permission error (in current revision)
- App routing rendering unauthorized since the current routing system works by checking the `user.role`

The fix in this PR will:

- show sidebar even if user preference fails, do avoid having empty sidebar
- render page even when user don't have permission (not viewer/editor/admin) when authz feature flag is enabled

The main side-effect of this change is to not show un-authorized pages anymore because we will now handle auth checks via authz directly.

Before:

https://github.com/user-attachments/assets/c6afb395-2ed6-4c04-b5c0-2b2bd2ddeb96

After:

https://github.com/user-attachments/assets/24cafc12-4a95-4129-8314-787ab66a858f

Contributes to https://github.com/SigNoz/platform-pod/issues/2734

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Sidebar & App Routing
- Potential regressions: -
- Rollback plan: Find and fix the specific issue

---

### 📋 Checklist
- [x] Manually tested


frontend/src/utils/permission/index.ts
```diff
@@ -23,6 +23,11 @@ export type ComponentTypes =
 	| 'add_panel_locked_dashboard'
 	| 'manage_llm_pricing';
 
+/**
+ * @deprecated Before adding a new value here, check if what you want to add permission is supported by authz.
+ * If so, read AuthZ Guidelines on how to add permission check via AuthZ.
```
AshwinBhatkal, line 28: where is this guidelines?

frontend/src/utils/permission/index.ts
```diff
@@ -141,3 +153,34 @@ export const routePermission: Record<keyof typeof ROUTES, ROLES[]> = {
...
 };
+
+/**
+ * Any route that will start be supported under AuthZ should be added here.
+ * This will help correctly identify when fallback to show the page
+ * or just return unauthorized.
+ *
+ * This prevents us from adding `ANONYMOUS` on the `routePermission`
+ */
+export const routeWithInitialAuthZSupport = {
```
AshwinBhatkal, line 164: should we building this as we go?

frontend/src/utils/permission/index.ts
H4ad: I didn't create yet, part of https://github.com/SigNoz/platform-pod/issues/2815

frontend/src/utils/permission/index.ts
H4ad: Yes, because this requires backend checks to be updated, and the UI to be handled correctly, I prefer to make this gradually

At the same time, we can just remove this and disable completely the Unauthorized page but I'm not sure if this is a good idea

AshwinBhatkal: So, all these routes have authz support right now?

AshwinBhatkal (approved):

PR #12311, opened 2026-07-28
fix(dashboard-v2): keep new-panel Save enabled after running the query

ahrefabhi:
## Pull Request

---

### 📄 Summary
Fixes a P0 in the **new-panel** flow and hardens the panel-editor save experience:

1. **New-panel Save re-disabled after running the query.** Create a panel → edit the query → Stage & Run, and the Save button went back to disabled, leaving you unable to save.
2. **Save button gated on "dirtiness" at all.** Per product direction, Save should always be available (subject to permissions) — the dirty check should keep working for its *other* job (the discard-on-close confirmation), not gate Save.
3. **Opaque save errors.** A failed save only ever showed a generic "Failed to save panel", hiding the server's actual reason.

image

#### Issues closed by this PR
Closes https://github.com/SigNoz/pulse-pod/issues/157

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
- **(1) Dirty baseline drift.** For a new panel the query-dirty baseline was derived from `seedQuery`, which reads the **live** `draft.spec.queries`. Stage & Run's `commitQuery` writes the current query *into* `draft.spec.queries`, so the baseline recomputed onto the just-run query and `isQueryDirty` flipped back to `false`. Since a new non-List panel's seed is query-less, the Save `isDirty` collapsed to that value → Save re-disabled.
- **(3)** The catch block hard-coded a generic toast string and discarded the caught error.

#### Fix Strategy
- **(1)** Freeze the seed used for the new-panel dirty baseline (captured once at mount) so a run can't drift it. Existing panels are unaffected — they anchor the baseline to `savedQueries`.
- **(2)** Drop `isDirty` from the Save button's `disabled` (now `readOnly || isSaving`). `isDirty` is retained and still drives the discard-on-close confirmation — which is exactly why keeping the dirty logic *correct* (fix 1) still matters.
- **(3)** Run the caught error through the sanctioned `toAPIError` converter and show the server's `error.message` in the toast **description**, keeping "Failed to save panel" as the stable headline; falls back to a generic description when the error is unparseable.

---

- **Tests added/updated:**
  - Integration test (real query builder) reproducing new panel → edit → Run stays query-dirty. Verified it **fails** on pre-fix code (`Received: false`) and passes after.
  - Header tests: Save stays enabled with no unsaved edits; disabled only while read-only/saving.
  - Container tests: failure toast surfaces the server message in its description; generic fallback for an unparseable error. Existing dirty-flag assertions retained (now framed as the discard-guard signal).
- **Manual verification:** `pnpm tsgo --noEmit`, `pnpm oxlint`, and `pnpm build` all clean. Affected PanelEditor suites green.
- **Edge cases covered:** untouched new panel stays clean (no discard prompt); existing-panel dirty check unaffected; single-fetch-on-Run guarantee preserved (V5-envelope no-op guard, not the seed); View/drilldown modal (the other consumer of the shared hook) doesn't read `isQueryDirty`.

---

- **Blast radius:** `usePanelEditorQuerySync` (new-panel baseline only), the editor Header's Save `disabled` condition, and the editor's save-error toast. All within the V2 panel editor.
- **Potential regressions:** Low. Save being always-enabled means a no-op save is now possible (re-PATCHes the same spec); acceptable per product direction. Dirty logic itself is unchanged in behavior except being made drift-free.
- **Rollback plan:** Revert the two commits on this branch — isolated, no schema/API/migration changes.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Dashboards V2: new panels can be saved after editing/running their query; the Save button is no longer disabled by the dirty check; failed saves now show the server's error message. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

Behind the V2 dashboards flag. The dirty-baseline fix deliberately does **not** touch `seedQuery` (still used by `commitQuery` and `useShareBuilderUrl`) — only the baseline is frozen — to avoid regressing the preview-commit path. Keeping `isDirty` correct still matters after decoupling Save, because it now solely powers the discard-on-close confirmation.

---


frontend/src/pages/DashboardPageV2/DashboardContainer/PanelEditor/index.tsx
```diff
@@ -236,10 +238,18 @@ function PanelEditorContainer({
...
+			toast.success('Panel saved', {
+				position: 'top-center',
+			});
 			onSaved();
-		} catch {
-			toast.error('Failed to save panel');
+		} catch (err) {
+			// Surface the server's failure message.
+			const apiError = toAPIError(err as Parameters<typeof toAPIError>[0]);
+			toast.error('Failed to save panel', {
```
AshwinBhatkal: showErrorModal

AshwinBhatkal: use that. It handles the conversion

ahrefabhi: updated

AshwinBhatkal (approved):

PR #12313, opened 2026-07-28
chore: delta temporality metrics should always be considered as a sum metric

namanv3:
## Pull Request

---

### 📄 Summary

If a Sum metric has delta temporality, currently we still check if isMonotonic is true or not. However, the monotonicity flag only makes sense for cumulative metrics.

Regardless of the monotonicity, the delta sum metrics will be considered as counters. The reason is the monotonic as a property is important for cumulative and doesn't affect any semantics of delta.

This PR ensures that the type is considered as Sum for delta metrics at these places:
1. In the query range API under the `fetchTemporalityTypeForTable` method. This ensures that reduced metrics choose the correct table. For non reduced metrics, the statement builder uses temporality currently for differentiation, but this change ensures that if it later needs to use type, it will be able to.
2. `toAttributeType` method in `queryBuilder.ts`: the temporality check here ensures that the metrics dropdown in metrics explorer shows the correct type, and subsequently the aggregation options shown are correct.
3. `getMetricDetailsQuery` in `frontend/src/container/MetricsExplorer/MetricDetails/utils.tsx`: the temporality check ensures that when a delta metric's "Open in Explorer" button is clicked, then the correct aggregations are chosen.

#### Issues closed by this PR

https://github.com/SigNoz/pulse-pod/issues/150

#### Testing strategy
* Created a local setup where a dummy app emits a delta metric with monotonicity false to confirm that UI treats it as sum.
* Integration tests added for non-reduced scenario and reduced scenario


pkg/query-service/app/clickhouseReader/reader.go
```diff
@@ -3016,6 +3016,7 @@ func (r *ClickHouseReader) GetMetricAggregateAttributes(ctx context.Context, org
 		temporality := string(metadata.Temporality)
 		isMonotonic := metadata.IsMonotonic
 
+		// Cumulative gate already present here; metadata.go was aligned to match.
```
srikanthccv: can you check if `GetMetricAggregateAttributes` is really used anywhere so we can get rid of it, because the List metrics API is already does everything this does.

srikanthccv: not in this PR but there are bunch of functions that can be removed from `reader.go`

pkg/modules/metricsexplorer/implmetricsexplorer/module.go
```diff
@@ -825,6 +825,7 @@ func (m *module) validateAndNormalizeMetricType(req *metricsexplorertypes.Update
 		if req.Temporality != metrictypes.Delta && req.Temporality != metrictypes.Cumulative {
 			return errors.NewInvalidInputf(errors.CodeInvalidInput, "invalid value for temporality")
 		}
+		// Cumulative gate already present here; metadata.go was aligned to match.
```
srikanthccv: you can remove this added line, it's local context for this PR (like an agent telling itself) rather than converting anything for the future (agent) reader.

tests/integration/tests/metricreduction/04_reduced_delta_nonmonotonic_sum.py
```diff
@@ -0,0 +1,90 @@
...
+                min_value=999.0,
+                max_value=999.0,
+                sum_values=999.0,
+                count_series=2,
+                count_samples=2,
+                temporality="Delta",
+            )
+            for ts in time_series
+            for minute in range(20)
+        ],
```
srikanthccv, line 87: Couldn't make sense of these values, what are they doing?

srikanthccv (commented):
Overall LGTM, one question about integration test data

pkg/query-service/app/clickhouseReader/reader.go
namanv3: It is eventually used in this API:
```
subRouter := router.PathPrefix("/api/v3").Subrouter()
	subRouter.HandleFunc("/autocomplete/aggregate_attributes", am.ViewAccess(
		withCacheControl(AutoCompleteCacheControlAge, aH.autocompleteAggregateAttributes))).Methods(http.MethodGet)
```

pkg/modules/metricsexplorer/implmetricsexplorer/module.go
namanv3: done

tests/integration/tests/metricreduction/04_reduced_delta_nonmonotonic_sum.py
namanv3: its a decoy, its not meant to be read. If it is read then the test should fail, that's why the values are bogus. Added code comment for explanation

pkg/query-service/app/clickhouseReader/reader.go
srikanthccv: Please create a deprecation ticket to sunset and remove the `autocomplete/aggregate_attributes` in favour of list metrics API, for both in-product and API users.

srikanthccv (approved):

tests/integration/tests/metricreduction/04_reduced_delta_nonmonotonic_sum.py
srikanthccv: It would be great, if you push these test in one commit , run the integration tests that show the failure , then add the change to prove that it is indeed working.

pkg/query-service/app/clickhouseReader/reader.go
namanv3: https://github.com/SigNoz/pulse-pod/issues/182

tests/integration/tests/metricreduction/04_reduced_delta_nonmonotonic_sum.py
namanv3: here's the PR with just the integration test and a failed run: https://github.com/SigNoz/signoz/pull/12386

therealpandey (approved):

PR #12314, opened 2026-07-28
feat(alert-channel-integrations): google chat backend

nikhilmantri0902:
## Pull Request

---

### 📄 Summary

Adds **Google Chat** as an alert notification channel (Alertmanager receiver). Google Chat isn't in the upstream Prometheus Alertmanager fork, so this ships a SigNoz custom notifier that posts a **cardsV2 rich card** to an incoming webhook: a short text summary plus a card with a colored FIRING/RESOLVED banner, the alert body with deep-link buttons to logs/traces, and a shared Open-in-SigNoz button, with per-rule threading. Grouped alerts render as a single combined card section.

The backend *type* + OpenAPI scaffold already landed on `main` (channel was API-storable but never delivered). This PR wires the actual notifier + registration + webhook-URL validation, so the channel now sends.

Sample for google chat alerts - 

Screenshot 2026-07-30 at 4 40 01 PM

#### Issues closed by this PR
Closes https://github.com/SigNoz/pulse-pod/issues/163

---

### ✅ Change Type

- [x] ✨ Feature

---

### 🧪 Testing Strategy

**Tests added (notifier)**

- **Delivery:** the notifier sends the message, renders the title as the summary and the body into the card, shows the correct firing or resolved status banner, and includes link buttons only when the alert carries them.
- **Rendering:** a custom template's markdown is converted to card HTML.
- **Grouping:** a custom template produces one section per alert, the default template produces a single combined section, and grouping is capped at 30 sections with a "+N more" note beyond that.
- **Threading:** alerts from the same rule share a thread key.
- **Reliability:** retryable status codes such as 429 are retried, oversized payloads are trimmed by both the byte and serialized-size guards, an empty render returns a non-retryable error, and a nil `HTTPConfig` is guarded.
- **Security:** the webhook URL is redacted from error messages.
- **Validation:** the webhook URL rules are enforced, and the receiver rejects an invalid URL.

**Manual verification**

- We delivered alerts live to a real Google Chat space, and the card rendered with its status banner, bold body, and clickable deep-link buttons.
- Both firing and resolved notifications returned HTTP 200, and the resolved message threaded under the firing one (same `thread.name`).
- We confirmed that Google Chat renders goldmark's output (bold, italic, lists, links, and code) and that the link buttons click through.

**Edge cases covered**

- A payload over 32 KB is trimmed by truncating the body first and then the title.
- Invalid UTF-8 is handled by `encoding/json`, which substitutes U+FFFD.
- An empty rendered title and body produces a clean, non-retryable error.
- A nil `HTTPConfig` is caught in `New` with a clear error instead of a panic.
- Distinct rules produce distinct thread keys.

---

- **Blast radius:** new isolated receiver type; no change to existing channels' send paths.
- **Potential regressions:** none expected — new code only runs when a `googlechat_configs` channel exists.
- **Rollback plan:** revert the PR; stored channels become inert (as on current `main`).

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Send alert notifications to Google Chat via an incoming webhook, as rich cards with deep-link buttons. |

---

### 📋 Checklist
- [x] Manually tested (live Google Chat space)

---

## 👀 Notes for Reviewers

- **Rich [cardsV2](https://developers.google.com/workspace/chat/api/reference/rest/v1/cards) message, not plain `{"text"}`:** We send a card (FIRING/RESOLVED banner, body, deep-link buttons) because plain text can't render buttons. Modeled on `msteamsv2`; the body reuses the existing `RenderHTML`, no new renderer.
  - **Verified live:** card rendering, threading, renotify, and the deep-link buttons all work against a real space.
  - **Caveat:** in fully-custom templates, markdown tables flatten (grid lost); bold/italic/links/lists/code render fine. Rare in alert bodies — noted for docs in [engineering-pod#5801](https://github.com/SigNoz/pulse-pod/issues/166).
  - **Grouped alerts:** today a group renders as one combined section; per-alert sections are latent shared infra (like Slack/MS Teams), triggered by a `_body_template` annotation nothing sets via UI or API yet. They activate once a per-rule notification-template backend and UI land — no notifier changes.

- **429 is retryable:** Google Chat's [1 write/sec per-space limit](https://developers.google.com/workspace/chat/limits#per-space_quotas) returns `429` on a transient rate hit and recommends backoff, so retrying the same payload succeeds once the window resets. This follows the PagerDuty/OpsGenie precedent (both add `429` to retry codes), unlike Slack/MS Teams.

- **Empty render → non-retryable error:** Google Chat rejects empty messages, and a render is only empty when fully custom title *and* body both produce nothing (a misconfig — the default always renders `[FIRING:n] alertname`). We fail fast and loud rather than send a placeholder or POST an empty payload for an opaque `400`.

- **Webhook-URL validation:** enforce https + host `chat.googleapis.com`, strict on create/update. We can relax it later if needed.

- **Threading** = `threadKey = sha256(group key)`, `messageReplyOption=REPLY_MESSAGE_FALLBACK_TO_NEW_THREAD`. No-op in non-threaded spaces (Google ignores the field, not an error).
Caveat: editing a rule's grouping starts a new thread.

- **Size guard** = a hard [32 KB](https://developers.google.com/workspace/chat/create-messages#:~:text=The%20maximum%20message%20size%20(including%20any%20text%20or%20cards)%20is%2032%2C000%20bytes.%20To%20send%20a%20message%20that%20exceeds%20this%20size%2C%20your%20Chat%20app%20must%20send%20multiple%20messages%20instead.) guard measured on the *serialized* payload; it trims the longest alert body first, then the title (which appears in both the summary and the card header), so the payload is always within Google's limit (a 400 would silently drop the alert).

- **Label-value markup escaping** = resolved by the card path: body text is HTML-escaped via `RenderHTML` (goldmark), so label values can't break the card or inject markup. (This was a known limitation of the earlier plain-text approach.)

- **`send_resolved` default = `false`.** Matches Slack and Email, which also default to `false` upstream. Users can toggle it per channel on the UI that would be built.

- **Follow-up (sibling PRs):** frontend (channel form + generated hooks) and user docs are tracked separately and land after this backend PR.


pkg/types/alertmanagertypes/googlechat.go
```diff
@@ -32,3 +36,18 @@ func (c *GoogleChatReceiverConfig) UnmarshalYAML(unmarshal func(any) error) erro
 	type plain GoogleChatReceiverConfig
 	return unmarshal((*plain)(c))
 }
+
+// ValidateGoogleChatWebhookURL validates the Google Chat webhook URL format.
+func ValidateGoogleChatWebhookURL(rawURL string) error {
```
therealpandey: @nikhilmantri0902 can you implement UnmarshalYAML here? If you see how the other receivers work, they all validate in UnmarshalYaml. See the upstream code once if unclear.

pkg/alertmanager/alertmanagernotify/googlechat/types.go
```diff
@@ -0,0 +1,97 @@
+package googlechat
```
therealpandey: Do we need this explicit file `types.go`? If it's following convention for the rest of the receiver, its okay. Otherwise can put them in googlechat.go.

therealpandey (changes requested):

pkg/alertmanager/alertmanagernotify/googlechat/types.go
nikhilmantri0902: I thought of sort of keeping seperate for readability but combining them now to match rest of the receivers.

pkg/types/alertmanagertypes/googlechat.go
nikhilmantri0902: Moved the webhook-URL validation into GoogleChatReceiverConfig.UnmarshalYAML (defaults → parse → validate), matching the upstream receiver convention, and dropped the separate validateGoogleChatConfig pass in receiver.go

pkg/alertmanager/alertmanagernotify/googlechat/googlechat.go
```diff
@@ -0,0 +1,307 @@
...
+		return false, err
+	}
+	for buf.Len() > maxMessageBytes {
+		over := buf.Len() - maxMessageBytes
+		if i := longestBodyIndex(bodies); i >= 0 {
+			bodies[i] = truncateToByteLimit(bodies[i], max(len(bodies[i])-over, 0))
+		} else if title != "" {
+			title = truncateToByteLimit(title, max(len(title)-over, 0))
+		} else {
+			break
```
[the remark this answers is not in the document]
nikhilmantri0902: Valid catch. The size loop trims bodies then title but not button URLs, so a card with many long-URL buttons could still overflow and get rejected.

It is not reachable on the shipping path though: the default template collapses a group into a single section with at most ~3 short buttons (first alert's logs/traces + the SigNoz footer), so button bytes are negligible and the body — which is trimmable — carries the bulk. The many-section, many-button case only arises with per-alert templating (the `_body_template` annotation), which nothing sets via UI or API yet.

Rather than error out (that still drops the alert), the right fix is to drop trailing sections and bump the "+N more" note until it fits, guaranteeing delivery down to banner + footer + note. Since it cannot fire until per-alert templating lands, tracked as a TODO at that spot for now (36f355dc19).

pkg/types/alertmanagertypes/googlechat.go
nikhilmantri0902: inlined into UnmarshalYAML

therealpandey (approved):

PR #12315, opened 2026-07-28
chore(claude): add authz skill

vikrantgupta25:
## Summary

Adds a Claude Code skill at `.claude/skills/authz/` that automates the workflow documented in `docs/contributing/go/authz.md` — so anyone (or any agent) adding authorization to a feature follows the full checklist instead of rediscovering it.

### What the skill covers

- **Registries**: kind + `Kind.Enum()`, resource, managed-role transactions (admin/editor/viewer/anonymous conventions)
- **CI-checked generated artifacts**: `generate openapi`, `generate authz` (including the `allowedResources` allowlist in `cmd/authz.go` for UI gating), `generate config transaction-groups`
- **Golden testdata**: `tests/integration/testdata/role/managed_role_transactions.json`
- **Route wiring** (`references/route-wiring.md`): `CheckResources` + `ResourceDef`s, extractors, selectors, sibling and parent-child linking patterns, custom `SelectorFunc`s — all examples taken from `serviceaccount.go`
- **Backfill migrations**: two-part pattern — FGA tuples (083-style, idempotent, PG + SQLite layouts) plus the managed-role `transaction_groups` JSON refresh (099-style); includes the 31-char factory-name limit gotcha
- **Integration tests** (`references/integration-tests.md`): suite layout, shared fixtures, managed-role matrix and per-object FGA test shapes, run commands

### Validation

Benchmarked with-skill vs no-skill agents on three tasks (register a resource, wire a linking route, write the authz test suite): equal-or-better correctness with roughly half the wall-clock time and 20% fewer tokens; the register-resource re-run after fixes scores 7/7 including the CI generated-artifacts check that agents without the skill only find by exploration.

Note: `.claude/` is only ignored via local `.git/info/exclude`, so these files track normally once force-added (same as the existing `.claude/agents/playwright-*.md`).

### DIFF
--- a/.claude/skills/authz/SKILL.md
+++ b/.claude/skills/authz/SKILL.md
@@ -1,0 +1,142 @@
+---
+name: authz
+description: Add authorization (authz) to a SigNoz backend feature end-to-end — register kinds/resources, grant managed-role permissions, wire routes with ResourceDefs, backfill tuples for existing orgs, and write the authz integration tests. Use this whenever the user mentions authz, authorization, permissions, access control, managed roles (admin/editor/viewer), OpenFGA, ResourceDef, CheckResources, 403/forbidden behavior on an endpoint, wiring or protecting a new route, registering a new resource/kind, or writing integration tests that assert role-based allow/deny. Also use it when a new API endpoint is being added and its access control hasn't been decided yet — every route needs an authz decision even if the user didn't say the word.
+---
+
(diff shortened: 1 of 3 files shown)


vikrantgupta25:
Moving this to SigNoz/platform-pod#2823 to incubate there first — we'll bring it back to this repo once the team is comfortable with the skill.

PR #12317, opened 2026-07-28
chore(docs): add authz guide

H4ad:
## Pull Request

---

This is the initial guide to help migrate the UI to use the new authz system.

This documents the decision made at https://github.com/SigNoz/platform-pod/issues/1525 and also add more instructions on how to handle the error feedback on UI for few scenarios.

Example on how it should look when you follow the guidelines:

https://github.com/user-attachments/assets/6add0e24-51de-4219-804b-07dec36bb393

Closes https://github.com/SigNoz/platform-pod/issues/2815

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Blast radius: None
- Potential regressions: None

---

| Field | Value |
|------|-------|
| Change Type | Maintenance |
| Description | N/A |

---

### 📋 Checklist
- [ ] Manually tested


frontend/docs/authz-guide.md
```diff
@@ -0,0 +1,136 @@
+# AuthZ Guide
+
+How to structure a page so it works with the permission system.
```
vikrantgupta25, line 3: Can we have it in the format as in what steps you need to follow to enable authz for a particular resource ? same as the backend contributing guidelines ?

vikrantgupta25: a workflow which a human/agent can execute directly

vikrantgupta25 (changes requested):

frontend/docs/authz-guide.md
H4ad: The agent can follow this already, example of me running at roles pages:

```
Blocking

CreateEditRolePage.tsx:36-48 — 🔴 fail-open when ?name= absent. roleName comes only from query param. No name → checks returns [] → useAuthZ permissions.every(...) on empty array = true → AuthZGuard renders children. So /settings/roles/<id>/edit (no ?name=) mounts the full editor with zero checks. Server still guards GET/PUT, but UI gate is bypassed. Fix: gate on data, not URL — mount unguarded shell, fetch role by roleId, then AuthZGuardContent/withAuthZContent with buildRoleReadPermission(data.roleName). Or minimum: when name missing, render denied instead of []. Same hole at ViewRolePage.tsx:338 (roleName ? [...] : []) — same family, fix together.

CreateEditRolePage.tsx:320,43-46 — 🔴 guide violation: edit page uses withAuthZPage with [read, update]. Guide (lines 36-39): edit page → prefer withAuthZContent to block content; header/back/delete stay visible; update button blocked separately. Today read-granted/update-denied user gets PermissionDeniedFullPage — and EditRolePage.authz.test.tsx:70 codifies that as expected. Guide wants: content visible (read), save button disabled with tooltip (update). Create mode withAuthZPage([RoleCreatePermission]) is correct per guide line 44 — keep it, split edit path off.

Should fix

CreateEditRolePage.tsx:220-225 — 🟠 save button checks [update] only. Guide: "only allow when user has permission to read and update" → [buildRoleReadPermission(roleName), buildRoleUpdatePermission(roleName)]. Also today it's dead code: page gate already demands update, so the denied/tooltip branch can never render. Two gates, one redundant — collapse per finding above.

CreateEditRolePage.tsx:34,55 — 🟠 wrong route pattern. Matches ROUTES.ROLE_DETAILS (/settings/roles/:roleId) while real routes are ROLE_CREATE + ROLE_EDIT (/settings/roles/:roleId/edit). Works only because matchPath is non-exact, and create mode depends on roleId === 'new' string literal matching the ROLE_CREATE URL tail. Match ROLE_EDIT explicitly; derive isCreateMode from ROLE_CREATE match, not 'new' (literal duplicated 3×: :35, :58, useCreateEditRolePageActions.ts:52).

__tests__/EditRolePage.authz.test.tsx:49-55 — 🟠 mounts edit page at the view route. <Route path={ROUTES.ROLE_DETAILS}> + initialRoute: /settings/roles/<id>?name= — real edit URL is /…/edit (featureGate.test.tsx:48 gets it right). Tests pass only via loose matching → they don't cover real routing.
```

vikrantgupta25 (dismissed):

frontend/docs/assets/drawer-example.svg
```diff
@@ -0,0 +1,372 @@
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 4810.804337955204 719.1627574653057"
```
AshwinBhatkal: nit: when no permission to read, we are showing red strokes. Ideally nothing is to be seen there at all. 
1. How would they reach that point if not read permission?
2. We should have a different color than disabled state because it is different

frontend/docs/assets/list-example.svg
```diff
@@ -0,0 +1,2 @@
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 4761.009446913065 741.8330576198573" width="4761.009446913065" height="741.8330576198573"><!-- svg-source:excalidraw --><metadata></metadata><defs><style class="style-fonts">
```
AshwinBhatkal, line 1: this ties to my first comment
List can return only those resources that user has min read access for

AshwinBhatkal: so create is not a property of the list, but the property of the resource being created

frontend/docs/assets/drawer-example.svg
H4ad: > we are showing red strokes.

This is what we show to the user:

image

1. They might have permission to list but not to read

2. I don't think so, the button is disabled and they will be able to know why if they hover the button, this is a pattern we already do in other parts of the UI (disabled + tooltip)

frontend/docs/assets/list-example.svg
H4ad: > List can return only those resources that user has min read access for

This is not true, list is different than read, and you can list things you don't have access to read

frontend/docs/assets/drawer-example.svg
AshwinBhatkal: We should have a different color than disabled state because it is different

In the guide I meant. Not the UI

AshwinBhatkal (dismissed):

frontend/docs/assets/drawer-example.svg
H4ad: Updated the text to clarify that part will be restricted.

AshwinBhatkal (approved):

PR #12322, opened 2026-07-29
fix: add meter source where needed during dashboard migration

namanv3:
## Pull Request

---

### 📄 Summary

`qb.WrapInV5Envelope` did not handle the source field, which is causing query range API in meter panels to break. Need to add handling for this field in migration code.

#### Testing
Ran the dashbaord repo migration with this fix, and the `meter.json` works fine after this fix.

### DIFF
--- a/pkg/types/dashboardtypes/perses_v1_to_v2_queries.go
+++ b/pkg/types/dashboardtypes/perses_v1_to_v2_queries.go
@@ -189,9 +189,7 @@ func (d *v1Decoder) collectV1QueryEnvelopes(widget map[string]any, panelKind Pan
 		queries = dropUnrenderableQueries(queries)
 		for _, q := range queries {
 			name := d.readString(q, "queryName")
-			env := qb.WrapInV5Envelope(name, q, string(qb.QueryTypeBuilder.StringValue()))
-			carryQuerySource(env, q)
-			out = append(out, env)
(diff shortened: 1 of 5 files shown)


pkg/types/dashboardtypes/perses_v1_to_v2_queries.go
```diff
@@ -446,3 +448,16 @@ func signalFromDataSource(raw any) telemetrytypes.Signal {
 	}
 	return telemetrytypes.Signal{}
 }
+
+// carryQuerySource copies a v1 builder query's meter source onto the v5 spec.
+// WrapInV5Envelope maps dataSource→signal but ignores the separate `source` field, so a
+// meter query (dataSource metrics + source "meter") would otherwise migrate as plain
+// metrics and read the wrong data. Only "meter" is carried through — the sole non-empty
+// source the v1 UI produces and v5 accepts; anything else stays unspecified.
+func carryQuerySource(env, query map[string]any) {
```
srikanthccv: please move this to `WrapInV5Envelope` only, it's not something specific to here so it should be part of `WrapInV5Envelope` so that every consumer of `WrapInV5Envelope` gets the same fix applied.

namanv3: done

srikanthccv (approved):

therealpandey (approved):

PR #12323, opened 2026-07-29
feat(prometheus): add clickhouseprometheusv2 native read path

srikanthccv:
> **Stack** (review in order; each PR's diff is against its predecessor):
> 1. #12323 `v2-read-path` — v2 native read path (leaf package)
> 2. #12324 `v2-wiring` — wiring, shadow/pin rollout machinery, dual-leg conformance
> 3. #12325 `v2-transpiler` — PromQL→ClickHouse transpiler + classification golden
> 4. #12093 `issue-4293` — the /prometheus API move (breaking slice, last)

### What

A second-generation ClickHouse-backed Prometheus provider as a leaf package —
compiled and tested, wired to nothing. The stock promql engine evaluates over a
native `storage.Querier` instead of v1's remote-read protobuf adapter:

- **Per-selector fetch windows** from the engine's `SelectHints`, not the
  query-wide union window: `foo / foo offset 1d` reads two narrow windows
  instead of the widest one twice.
- **Last-sample-per-step reduction** for instant selectors of subquery-free
  queries: buckets anchored at the selector's first evaluation timestamp so a
  non-final sample of a bucket can never be the latest in any grid lookback
  window. The proof travels as `prometheus.QueryTraits` in the context; call
  sites that don't attach traits get the conservative raw fetch.
- **Series identity done right at the boundary**: empty-valued labels dropped
  at read (Prometheus absent semantics), no synthetic `fingerprint` label,
  identical-labelset series merged (`sortAndMerge`) because the engine assumes
  storages never emit duplicates.
- **The v1 parity fixes baked in from birth** (they were retrofitted to v1 in
  #12151–#12157): `__name__` matchers of all four types resolve against the
  `metric_name` column, matcher regexes anchored before ClickHouse `match()`
  (substring semantics), series-lookup upper bound inclusive because
  registration rows are hour/6h/1d/1w-floored by the exporter.
- **Statement capture** (`CapturingStorage`) for the preview path, and
  `log_comment` attribution on every statement.

### Alternatives considered and discarded

- **Fixing v1 instead.** The correctness gaps *were* fixed in v1
  (#12151–#12157) — but the cost model cannot be: remote-read serializes every
  raw sample of the union window through protobuf regardless of the question
  asked. That layer is the floor. A native querier removes the layer instead
  of optimizing under it.
- **Exporting the concrete `Provider`.** Debated in review and settled the
  other way: the concrete type stays unexported and `New` returns
  `prometheus.Prometheus` — the interface is the boundary between the two
  provider implementations. The transpiler entry point (PR 3) is reached
  through a small `prometheus.RangeExecutor` capability instead of the
  concrete type; it folds into the main interface once v1 is removed.
- **Forking the engine to thread the subquery-free proof.** The engine's
  interfaces offer no channel to `Querier.Select` other than the context;
  `SelectHints.Step` always carries the top-level step, so the traits value is
  the only fork-free way to make step-aligned reductions safe.
- **Keeping empty-valued labels (OTel says empty is meaningful).** Prometheus
  says absent; serving PromQL means Prometheus semantics win at this boundary.
  Same deliberate translation v1 now performs, documented in docs/contributing/prometheus.md.

**Deliberately left out for now** (per review): the fetch budgets
(`MaxFetchedSeries`/`MaxFetchedSamples` config and their typed limit errors).
They return as their own change.

Per review: the subsystem context lives in `docs/contributing/prometheus.md`
(next to query-range.md) instead of a doc.go, and the table names +
window-to-table selection delegate to the shared
`pkg/telemetryschema/metricstelemetryschema` package.

### Test plan

`go test ./pkg/prometheus/...` — unit tests for SQL generation, querier
assembly, the merge, and the last-sample-per-step argument. The package is
inert in production until PR 2 wires it.

### DIFF
--- a/pkg/prometheus/clickhouseprometheusv2/provider.go
+++ b/pkg/prometheus/clickhouseprometheusv2/provider.go
@@ -1,0 +1,71 @@
+package clickhouseprometheusv2
+
+import (
+	"context"
+
+	"github.com/SigNoz/signoz/pkg/factory"
+	"github.com/SigNoz/signoz/pkg/prometheus"
+	"github.com/SigNoz/signoz/pkg/telemetrystore"
+	"github.com/prometheus/prometheus/storage"
+)
+
+// provider ties the package together: its own engine and parser, and the
+// ClickHouse client behind the native storage.Querier. It stays unexported:
+// callers hold the prometheus.Prometheus interface, which is the boundary
+// between the two provider implementations.
+type provider struct {
+	settings factory.ScopedProviderSettings
+	engine   *prometheus.Engine
+	parser   prometheus.Parser
+	client   *client
+}
+
+var (
+	_ prometheus.Prometheus        = (*provider)(nil)
+	_ prometheus.StatementCapturer = (*provider)(nil)
+)
+
+func NewFactory(telemetryStore telemetrystore.TelemetryStore) factory.ProviderFactory[prometheus.Prometheus, prometheus.Config] {
+	return factory.NewProviderFactory(factory.MustNewName("clickhousev2"), func(ctx context.Context, providerSettings factory.ProviderSettings, config prometheus.Config) (prometheus.Prometheus, error) {
+		return New(ctx, providerSettings, config, telemetryStore)
+	})
+}
+
+func New(_ context.Context, providerSettings factory.ProviderSettings, config prometheus.Config, telemetryStore telemetrystore.TelemetryStore) (prometheus.Prometheus, error) {
+	settings := factory.NewScopedProviderSettings(providerSettings, "github.com/SigNoz/signoz/pkg/prometheus/clickhouseprometheusv2")
+
+	engine := prometheus.NewEngine(settings.Logger(), config)
+	parser := prometheus.NewParser()
+	client := newClient(settings, telemetryStore, config)
+
+	return &provider{
+		settings: settings,
+		engine:   engine,
+		parser:   parser,
+		client:   client,
+	}, nil
+}
+
+func (p *provider) Engine() *prometheus.Engine {
+	return p.engine
+}
+
+func (p *provider) Parser() prometheus.Parser {
+	return p.parser
+}
+
+func (p *provider) Storage() storage.Queryable {
+	return p
+}
+
+func (p *provider) Querier(mint, maxt int64) (storage.Querier, error) {
+	return &querier{mint: mint, maxt: maxt, client: p.client}, nil
+}
+
+// CapturingStorage implements prometheus.StatementCapturer: a storage that
+// records each selector's SQL without executing it, for the preview path.
+// A fresh recorder per call keeps concurrent dry-runs isolated.
+func (p *provider) CapturingStorage() (storage.Queryable, prometheus.StatementRecorder) {
+	recorder := &statementRecorder{}
+	return &captureQueryable{client: p.client, recorder: recorder}, recorder
+}
--- a/docs/contributing/prometheus.md
+++ b/docs/contributing/prometheus.md
@@ -1,0 +1,123 @@
+# PromQL Serving — clickhouseprometheusv2
+
+This document is the subsystem context for `pkg/prometheus/clickhouseprometheusv2`,
+the second-generation ClickHouse-backed Prometheus provider. It explains why the
+package exists, the correctness constraints that shaped it, and how each fetch
+reduction is proven not to change results. Any change to the provider must keep
+these invariants; if a change would violate one, it must be flagged and
+discussed.
+
+---
+
+## Why a second provider
+
+The v1 provider (`pkg/prometheus/clickhouseprometheus`) serves the promql engine
+through the remote-read protobuf adapter: every raw sample of a query's union
+window is fetched, serialized, and handed to the engine. The cost is a function
+of ingested data, not of the question asked — which is how a dashboard of PromQL
+panels can take an instance down.
+
+In v2 the stock promql engine evaluates over a native `storage.Querier`: no
+translation layer, per-selector fetch windows, and fetch reductions that are
+provably invisible to the engine.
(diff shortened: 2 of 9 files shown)


pkg/prometheus/clickhouseprometheusv2/doc.go
```diff
@@ -0,0 +1,49 @@
+// Package clickhouseprometheusv2 is the second-generation ClickHouse-backed
```
therealpandey: This doc.go is a design document, not godoc. We should move the content todocs/contributing/prometheus.md, next to query-range.md, which is where we keep subsystem context. The existing doc.go files (pkg/errors, pkg/http, pkg/instrumentation) are stale and should be removed. I will raise a PR to remove them

pkg/prometheus/clickhouseprometheusv2/provider.go
```diff
@@ -0,0 +1,72 @@
...
+	_ prometheus.StatementCapturer = (*Provider)(nil)
+)
+
+func NewFactory(telemetryStore telemetrystore.TelemetryStore) factory.ProviderFactory[prometheus.Prometheus, prometheus.Config] {
+	return factory.NewProviderFactory(factory.MustNewName("clickhousev2"), func(ctx context.Context, providerSettings factory.ProviderSettings, config prometheus.Config) (prometheus.Prometheus, error) {
+		return New(ctx, providerSettings, config, telemetryStore)
+	})
+}
+
+func New(_ context.Context, providerSettings factory.ProviderSettings, config prometheus.Config, telemetryStore telemetrystore.TelemetryStore) (*Provider, error) {
```
therealpandey: New should return prometheus.Prometheus like the factory does, and the concrete Provider should stay unexported. Two implementations of the interface already exist (v1 and this), so the interface is the boundary.

Why?
- PR 2 currently threads `*clickhouseprometheusv2.Provider` through querier and signoz wiring, which couples every caller to this leaf package and is how import cycles start.
- Wiring needs to happen only in the signoz/ package.

pkg/prometheus/clickhouseprometheusv2/tables.go
```diff
@@ -0,0 +1,70 @@
+package clickhouseprometheusv2
```
therealpandey: The table names and window-to-table selection here should live in a shared schema package (which now exists after tushar's 3 layer refactor), `pkg/telemetryschema/metricstelemetryschema`, not in the provider.

We can scope it to v2 only.

therealpandey (changes requested):

pkg/prometheus/clickhouseprometheusv2/provider.go
therealpandey: Briefly looked at PR-3, here's the function that can be added on the prometheus interface:
```
TryExecuteRange(ctx context.Context, query string, start, end time.Time, step time.Duration) (promql.Matrix, bool, error)
```

v1 can return `nil, false, nil` which is also the correct answer for v1 as there is nothing to transpile.

pkg/prometheus/clickhouseprometheusv2/doc.go
therealpandey: Here's the PR for the removal: https://github.com/SigNoz/signoz/pull/12365

pkg/prometheus/clickhouseprometheusv2/provider.go
srikanthccv: There are not going to be two implementations, the plan was not to touch v1 at all while we do this, and when the v2 lands fully and made the main implementation the entire v1 is going to go away. So trying to add a method `TryExecuteRange` for v1 is just trying to come up with an interface purely for the rollout than it actually being an interface for its true reason from original problem. Let me know if you feel strongly about this.

pkg/prometheus/clickhouseprometheusv2/tables.go
srikanthccv: I had a todo on this already, let me integrate it now.

pkg/prometheus/clickhouseprometheusv2/provider.go
therealpandey: Post the v1 deletion, it will always be a part of the interface. Its a resonable extension to the interface as you will directly support promql executes in the promql engine.

Or are you saying that the function itself will also be deleted from clickhouseprometheusv2? I might have misunderstood in that case

pkg/prometheus/clickhouseprometheusv2/tables.go
therealpandey: Yes the TODO was before the 3 layer refactor and it wasn't advisable to it before that because you would have gotten statementbuilder imports in prometheus.

pkg/prometheus/clickhouseprometheusv2/provider.go
srikanthccv: it's going to be removed altogether, the only way the promql_query run the query is going to be the client

pkg/prometheus/clickhouseprometheusv2/doc.go
srikanthccv: done

pkg/prometheus/clickhouseprometheusv2/tables.go
srikanthccv: done

pkg/prometheus/clickhouseprometheusv2/doc.go
therealpandey: Can you please approve my PR also?

pkg/prometheus/clickhouseprometheusv2/provider.go
srikanthccv: done

therealpandey (approved):

PR #12326, opened 2026-07-29
fix(logs): highlight row when navigating to a shared log link

okxint:
## What

Fixes shared log links not highlighting the target row in the logs explorer.

## Root cause

`useCopyLogLink.ts` — the `useEffect` that runs when `isActiveLog` becomes `true` only set up the fade-out timer. It never called `setIsHighlighted(true)` first, so the highlighted state was never turned on before being scheduled to turn off.

## Fix

Add `setIsHighlighted(true)` immediately after the early-return guard, before the `setTimeout`. The effect now correctly transitions: `false → true → (after delay) → false`.

### DIFF
--- a/frontend/src/container/InfraMonitoringHosts/HostsListTable.tsx
+++ b/frontend/src/container/InfraMonitoringHosts/HostsListTable.tsx
@@ -227,7 +227,7 @@ export default function HostsListTable({
 			}}
 			scroll={{ x: true }}
 			loading={{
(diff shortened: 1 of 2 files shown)


tewarig:
Hi @okxint,

Thanks for raising this PR! It would be really helpful if you could share steps to reproduce the bug — a screenshot or short video would be great. Also, you might need to sync this branch with main.

PR #12327, opened 2026-07-29
feat(dashboard-v2): format panel values with thousand separators

### LINKED ISSUE #7669: Number/Metric formatting options in dashboard panels
Formatting numbers like this 1,234,567 with the commas.

ahrefabhi:
## Pull Request

---

### 📄 Summary

Large numbers in dashboard panels rendered as an undelimited digit run — `1234567` instead of `1,234,567` — which is hard to read at a glance and hard to compare between panels.

Unitless values were the visible gap. They format through the `'none'` unit, which routes to `formatDecimalWithLeadingZeros`, whose `Intl.NumberFormat` is constructed with `useGrouping: false`. Units that scale their own value (`bytes` → `1.18 MiB`, `short` → `1.23 Mil`) never reach four integer digits, so they never showed the problem.

The grouping is applied inside `formatPanelValue` — the single seam through which V2 panels reach `getYAxisFormattedValue` — rather than at one call site. That is deliberate: readability of a large scalar is not specific to one panel kind, so the Number panel, Table value cells and the threshold-row previews all pick it up from one place instead of each opting in.

`groupThousands` itself is conservative. It touches only the first numeric token's integer digits, so fractions, unit labels and formatter-scaled values pass through untouched, and exponent notation is skipped (grouping a mantissa reads as noise).

No capture attached. The visible delta is purely the separators:

| Panel | Unit | Before | After |
|---|---|---|---|
| Number | — | `1234567` | `1,234,567` |
| Number | `percent` | `1234567%` | `1,234,567%` |
| Number | `bytes` | `1.18 MiB` | `1.18 MiB` (unchanged) |
| Table cell | — | `1234567` | `1,234,567` |

#### Issues closed by this PR

Closes #7669

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context

**N/A** — this is an enhancement, not a regression. Grouping was never implemented; `useGrouping: false` in `formatDecimalWithLeadingZeros` is longstanding and intentional for the axis-tick path it was written for.

---

### 🧪 Testing Strategy

- **Tests added/updated:**
  - `groupThousands.test.ts` (new) — the transform in isolation.
  - `parseFormattedValue.test.ts` (new) — this util had no suite; covers the unit split, including grouped input.
  - `formatPanelValue.test.ts` — asserts grouping at the seam, and that unit-scaled values stay ungrouped.
  - `NumberPanel/__tests__/Renderer.test.tsx` — grouped value, grouped value + separate unit, and unit-scaled value left alone.
  - `TablePanel/__tests__/Renderer.test.tsx` — pins the grouping that Table cells inherit. Worth noting for reviewers: `tableColumns.test.ts` and `tableCsv.test.ts` both stub `formatPanelValue`, so they are blind to this change by construction — the renderer test is what actually covers the Table path.

- **Manual verification:** not a click-through of a live dashboard. Instead: the full frontend jest suite (7,312 passing; the 3 failures are in unrelated suites — `QuerySearch`, `AuthDomain` — and were confirmed flaky/pre-existing by re-running them alone and against a stashed pristine tree), plus `tsgo --noEmit`, `oxlint`, `oxfmt --check` and `vite build` all clean. The real `getYAxisFormattedValue` output was probed directly for 11 value/unit combinations before writing the transform, and `papaparse.unparse` was run directly to confirm exactly how a grouped cell serializes.

- **Edge cases covered:** negative values (sign stays outside the first group), fractions (never grouped), exponent notation (skipped), `∞` / `-∞` / `NaN` (untouched), prefix and suffix unit decoration (`$ 1,234,567`, `1,234,567%`, `1,234,567 ms`), formatter-scaled units, values below 1000, zero, and idempotency on already-grouped input.

---

- **Blast radius:** every `formatPanelValue` consumer — the Number panel, Table panel value cells, the three threshold-row previews in the config pane, and the Table CSV export. Display-only in all cases; no spec/DTO or API change, nothing persisted.

- **Potential regressions:**
  - **The CSV export is the one behavior change worth a reviewer's attention.** `formatTableCellText` is shared by the Table renderer and the export, so a unitless numeric column now serializes as `"1,234,567"` (papaparse quotes any field containing the delimiter) instead of `1234567` — spreadsheet `SUM`/`AVG` and downstream `parseFloat` would read it as text. Columns with a unit were *already* display-text in the export (`295.43 ms`, `1.18 MiB`) by design ("reusing the on-screen cell formatting", V1 parity), so only unitless numeric columns change in kind. Called out explicitly because it is an accepted trade-off, not an oversight — if we would rather keep the export numeric, the contained fix is a flag threaded through `formatTableCellText` from `tableCsv.ts` only, leaving the render path grouped.
  - Table **sorting and threshold evaluation are unaffected** — both read `toCellNumber(raw)`, never the formatted string.
  - `parseFormattedValue` had to learn to accept `,`, otherwise a grouped value would fall through to the whole-string fallback and lose its unit split. Covered by its new suite.

- **Rollback plan:** revert the PR. Display-only with no migration or persisted state, so a revert is immediate and total.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Large numbers in dashboard panels are now formatted with thousand separators (`1,234,567`), in the Number panel, Table panel value cells and threshold labels. |

---

### 📋 Checklist
- [ ] Manually tested

---

## 👀 Notes for Reviewers

Two things I would want a second pair of eyes on:

1. **"Manually tested" is deliberately unchecked.** This was validated through tests and direct probes of the real formatter, not by clicking through a running dashboard. A quick look at a Number panel and a Table panel — plus a screenshot for this PR — is worth doing before merge.
2. **The CSV trade-off** under Risk & Impact. Grouping at the seam is what makes the change one line instead of four call sites, but the export rides the same path. The narrower alternative is described there if you would rather not accept it.

The three commits are independently reviewable: the transform, the parser tolerance it requires, then the seam that turns it on.

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/formatPanelValue.ts
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/formatPanelValue.ts
@@ -1,16 +1,21 @@
 import type { PrecisionOption } from 'components/Graph/types';
 import { getYAxisFormattedValue } from 'components/Graph/yAxisConfig';
 
+import { groupThousands } from './groupThousands';
+
 /**
- * Formats a scalar for display in a V2 panel, honoring decimal precision. The
- * single seam through which panels touch `getYAxisFormattedValue`. Unitless
- * values format through the `'none'` unit, which still respects precision — so
- * precision isn't silently dropped when no unit is set.
+ * Formats a scalar for display in a V2 panel, honoring decimal precision and
+ * grouping the integer part into thousands. The single seam through which panels
+ * touch `getYAxisFormattedValue`. Unitless values format through the `'none'`
+ * unit, which still respects precision — so precision isn't silently dropped
+ * when no unit is set.
  */
 export function formatPanelValue(
 	value: number,
 	unit?: string,
 	precision?: PrecisionOption,
 ): string {
-	return getYAxisFormattedValue(String(value), unit || 'none', precision);
+	return groupThousands(
+		getYAxisFormattedValue(String(value), unit || 'none', precision),
+	);
 }
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/kinds/NumberPanel/__tests__/Renderer.test.tsx
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/kinds/NumberPanel/__tests__/Renderer.test.tsx
@@ -125,6 +125,37 @@ describe('NumberPanelRenderer', () => {
 		expect(queryByText('3.14159')).not.toBeInTheDocument();
 	});
 
+	// #7669: large scalars are unreadable as an undelimited digit run.
+	it('groups large values into thousands', () => {
+		const { getByText, queryByText } = renderPanel({
+			panel: panelWith({}),
+			data: dataWith('1234567'),
+		});
+
+		expect(getByText('1,234,567')).toBeInTheDocument();
+		expect(queryByText('1234567')).not.toBeInTheDocument();
+	});
+
+	it('groups the value while keeping its unit separate', () => {
+		const { getByText } = renderPanel({
+			panel: panelWith({ formatting: { unit: 'percent' } }),
+			data: dataWith('1234567'),
+		});
+
+		expect(getByText('1,234,567')).toBeInTheDocument();
+		expect(getByText('%')).toBeInTheDocument();
+	});
+
+	it('leaves a unit-scaled value ungrouped', () => {
+		const { getByText } = renderPanel({
+			panel: panelWith({ formatting: { unit: 'bytes' } }),
+			data: dataWith('1234567'),
+		});
+
+		expect(getByText('1.18')).toBeInTheDocument();
+		expect(getByText('MiB')).toBeInTheDocument();
+	});
+
 	it('renders No Data when the response has no scalar results', () => {
 		const { getByTestId } = renderPanel({ data: emptyData });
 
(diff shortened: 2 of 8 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/formatPanelValue.ts
```diff
@@ -1,16 +1,21 @@
...
+ * unit, which still respects precision — so precision isn't silently dropped
+ * when no unit is set.
  */
 export function formatPanelValue(
 	value: number,
 	unit?: string,
 	precision?: PrecisionOption,
 ): string {
-	return getYAxisFormattedValue(String(value), unit || 'none', precision);
+	return groupThousands(
```
AshwinBhatkal, line 18: why not have groupThousands inside getYAxisFormatted value?

AshwinBhatkal (approved):
Approved with a comment

frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/utils/formatPanelValue.ts
ahrefabhi: getYAxisFormattedValue is being used in other places as well where we don't need groupThousands

PR #12329, opened 2026-07-29
fix(query-builder): guard the operator lookup against an unmapped dataType

tewarig:
## Pull Request

---

### 📄 Summary

Picking a filter key whose `dataType` isn't one of the four scalars crashes the query builder. Reported from `/logs/pipelines` (Sentry `SIGNOZ-UI-5GB`, v0.131.1), but `QueryBuilderSearchV2` is also rendered by `ResourceAttributeFilterV2`, ApiMonitoring's endpoint views and `TracesFunnelDetails/FunnelStep`, so it isn't pipelines-only. The new query builder (`QueryBuilderV2/QueryV2/QuerySearch`) is unaffected — it keys a different map off literal constants.

Falls back to the `universal` operator list when the lookup misses — the same list the unresolved-key branch right below already uses.

N/A — reproducing needs a tenant with a non-canonical `tag_data_type` row. Sentry replay: `e892aa35ea3d421dbf3f5dfad1978bc0`.

#### Issues closed by this PR

Sentry `SIGNOZ-UI-5GB`

closes - https://github.com/orgs/SigNoz/projects/39/views/11?filterQuery=assignee%3Atewarig&pane=issue&itemId=220221929&issue=SigNoz%7Cengineering-pod%7C5797

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause

`QUERY_BUILDER_OPERATORS_BY_TYPES` has entries for `string`, `int64`, `float64`, `bool` and `universal`. The branch reading it only checks that `dataType` is *truthy*, not that it's a key in the map:

```ts
if (currentFilterItem?.key?.dataType) {
  operatorOptions = QUERY_BUILDER_OPERATORS_BY_TYPES[
    currentFilterItem.key.dataType as keyof typeof QUERY_BUILDER_OPERATORS_BY_TYPES
  ].map(...)   // undefined.map for anything outside the four
```

Not a regression — a faulty assumption from the start. Plenty of values legitimately reach it: `DataTypes` itself declares four `array(...)` types, and the v5 metadata API spells types `number` / `[]string`. The `as keyof typeof` cast is why tsc never flagged it.

Where the value comes from: `GetLogAttributeKeys` casts `tag_data_type` straight out of ClickHouse with no normalisation (`clickhouseReader/reader.go:3639`), so whatever the collector wrote reaches the UI verbatim.

#### Fix Strategy

`?? QUERY_BUILDER_OPERATORS_BY_TYPES.universal` on the lookup. Known types behave exactly as before; an unmapped one gets a wider operator list instead of a blank page.

---

### 🧪 Testing Strategy

- **Tests added/updated:** none. The right shape is a unit test over the lookup rather than a full component render — doing that separately instead of bolting it on here.
- **Manual verification:** `tsgo --noEmit` clean; existing `QueryBuilderSearchV2.test.tsx` passes (4/4). Not reproduced in a browser — that needs a tenant carrying an offending row.
- **Edge cases covered:** four scalars unchanged; array and `[]string`-style types now fall back; empty/`undefined` `dataType` unchanged (still routed to the `body.…[*]` and `universal` branches below).

---

- **Blast radius:** one lookup in the operator dropdown. No change for any `dataType` that already worked.
- **Potential regressions:** effectively none — the only new path is one that previously threw.
- **Rollback plan:** revert; restores the crash, nothing else.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Fixed a crash when selecting a query-builder filter key with an unrecognised data type. |

---

### 📋 Checklist
- [x] Tests added or explicitly not required — not added, see Testing Strategy
- [ ] Manually tested — not reproduced in a browser
- [x] Breaking changes documented — none

---

## 👀 Notes for Reviewers

**This PR only fixes one of three copies.** The identical unguarded lookup also lives in `ClientSideQBSearch.tsx:403` (AlertHistory timeline) and `useOperators.ts:25` (InfraMonitoringK8s, via the v1 search). They crash the same way on the same data — worth a follow-up, and arguably worth folding in here if you'd rather close it in one go.

Related: #12298 fixed this exact class in the table drilldown two days ago (a `number` dataType missing the same map), with a resolution table in `drilldownUtils.tsx`. Four call sites have now drifted from this one map twice, which suggests the real fix is typing it `Record<DataTypes, string[]>` and dropping the `as keyof typeof` casts so tsc enforces coverage — plus possibly one shared accessor instead of four independent lookups.

Two smaller follow-ups left out on purpose:

1. **Array types still get the wrong operators.** `universal` offers `>`, `<`, `CONTAINS`, which mean nothing against an array — the right list is `HAS`/`NHAS`, matching the `body.…[*]` branch. This PR stops the crash; it doesn't make arrays behave.
2. **The backend shouldn't emit these values.** `field_datatype.go` already has a canonicaliser that `GetLogAttributeKeys` doesn't use. Fixing it there stops the bad value at the source.

### DIFF
--- a/frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/QueryBuilderSearchV2.tsx
+++ b/frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/QueryBuilderSearchV2.tsx
@@ -756,10 +756,15 @@ function QueryBuilderSearchV2(
 
 			let operatorOptions;
 			if (currentFilterItem?.key?.dataType) {
-				operatorOptions = QUERY_BUILDER_OPERATORS_BY_TYPES[
-					currentFilterItem.key
-						.dataType as keyof typeof QUERY_BUILDER_OPERATORS_BY_TYPES
-				].map((operator) => ({
+				// the lookup only covers the four scalars, so an array type
+				// or a `[]string` spelling finds nothing — fall back rather than throw
+				const operatorsForDataType =
+					QUERY_BUILDER_OPERATORS_BY_TYPES[
+						currentFilterItem.key
+							.dataType as keyof typeof QUERY_BUILDER_OPERATORS_BY_TYPES
+					] ?? QUERY_BUILDER_OPERATORS_BY_TYPES.universal;
+
+				operatorOptions = operatorsForDataType.map((operator) => ({
 					label: operator,
 					value: operator,
 				}));
--- a/frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/QueryBuilderSearchV2.tsx
+++ b/frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/QueryBuilderSearchV2.tsx
@@ -756,8 +756,7 @@ function QueryBuilderSearchV2(
 
 			let operatorOptions;
 			if (currentFilterItem?.key?.dataType) {
-				// the lookup only covers the four scalars, so an array type
-				// or a `[]string` spelling finds nothing — fall back rather than throw
+				// Fallback to universal suggestions if no match found for currentFilter dataType
 				const operatorsForDataType =
 					QUERY_BUILDER_OPERATORS_BY_TYPES[
 						currentFilterItem.key
--- a/frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/__test__/QueryBuilderSearchV2.test.tsx
+++ b/frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/__test__/QueryBuilderSearchV2.test.tsx
@@ -113,6 +113,14 @@ const mockAggregateKeysData = {
 				isJSON: IS_JSON_FALSE,
 				id: 'service.name--string--tag--false',
 			},
+			{
+				key: 'unmapped.attribute',
+				dataType: 'String' as unknown as DataTypes,
+				type: TYPE_TAG,
+				isColumn: IS_COLUMN_FALSE,
+				isJSON: IS_JSON_FALSE,
+				id: 'unmapped.attribute--String--tag--false',
+			},
 		],
 	},
 };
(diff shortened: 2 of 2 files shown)


frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/QueryBuilderSearchV2.tsx
```diff
@@ -756,10 +756,15 @@ function QueryBuilderSearchV2(
 
 			let operatorOptions;
 			if (currentFilterItem?.key?.dataType) {
-				operatorOptions = QUERY_BUILDER_OPERATORS_BY_TYPES[
-					currentFilterItem.key
-						.dataType as keyof typeof QUERY_BUILDER_OPERATORS_BY_TYPES
-				].map((operator) => ({
+				// the lookup only covers the four scalars, so an array type
```
aks07: Please write a simpler comment.
Eg: Fallback to universal suggestions if no match found for currentFilter dataType

aks07:
A test file for this already exists. Lets add a test for this as well

tewarig:
> A test file for this already exists. Lets add a test for this as well

added

frontend/src/container/QueryBuilder/filters/QueryBuilderSearchV2/__test__/QueryBuilderSearchV2.test.tsx
```diff
@@ -165,52 +171,36 @@ jest.mock('hooks/dashboard/useDashboardVariables', () => ({
 }));
 
 describe('Suggestion Key -> Operator -> Value Flow', () => {
-	beforeEach(() => {
-		jest.clearAllMocks();
-	});
```
tewarig, line 170: Need to remove "before each" from a lot of places since we have already enabled , clearMocks .

aks07 (approved):

PR #12335, opened 2026-07-29
feat(alert-channel-integrations): google chat final leg

AshwinBhatkal:
## Pull Request

---

### 📄 Summary

Adds **Google Chat** as a notification channel type, so it can be created, edited and tested from the "New Notification Channel" screen like Slack, Email or Microsoft Teams. Once configured, an alert firing (or resolving) posts a message to the Google Chat space the webhook points at.

Depends on the backend support in https://github.com/SigNoz/signoz/pull/12314 — the type is inert until that lands.

**Form fields.** Selecting **Type → Google Chat** shows:

| Field | Required | Default |
| --- | --- | --- |
| Name | Yes | — |
| Send resolved alerts | No | ON (shared form default) |
| Webhook URL | Yes | — |
| Title | No | prefilled template |
| Description | No | prefilled template |

Title and Description are Go `text/template` snippets, prefilled verbatim from the backend's `DefaultGoogleChatReceiverConfig` (`pkg/types/alertmanagertypes/googlechat.go`), so the prefill and the server-side fallback stay identical and clearing a box is still safe.

**Validation.** The webhook URL is checked client-side before submitting — must be `https` and on `chat.googleapis.com` — surfaced both as an inline field error and as an error notification if Save/Test is pressed with a bad URL. The backend enforces the same two rules; this only avoids a round-trip to a 400.

**API.** Create / Update / Test all send the same body via the generated hooks (`useCreateChannel`, `useUpdateChannelByID`, `useTestChannel`) rather than new hand-written `api/channels/*` files. `http_config` is backend-managed and is not sent.

```json
{
  "name": "my-gchat-channel",
  "googlechat_configs": [
    {
      "webhook_url": "https://chat.googleapis.com/v1/spaces/AAAA/messages?key=...&token=...",
      "title": "...",
      "text": "...",
      "send_resolved": true
    }
  ]
}
```

**Prefilled templates.** Slack's title/description templates were inline in `CreateAlertChannels` state. They move to a `SlackInitialConfig` constant (byte-identical), and every channel type's prefills now sit in one `ChannelInitialConfig` map keyed by `ChannelType`. Changing the Type dropdown applies exactly that type's entry to both `selectedConfig` and the antd form. The `setFieldsValue` call is the part that matters: `title`, `text` and `description` are reused across types, and antd keeps a field's value after it unmounts, so without it the new type inherits the previous type's text — picking Google Chat and switching back to Slack would otherwise leave Google Chat's templates on a Slack channel. This replaces the per-type `if` branches in `onTypeChangeHandler` and an `initialValue` that spread three unrelated channels' defaults on every render (see Bug Context).

https://github.com/user-attachments/assets/a315b573-0423-4c1d-bbcf-8727ad994bb8

#### Issues closed by this PR

Closes https://github.com/SigNoz/pulse-pod/issues/167

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause

`initialValue` passed to `FormAlertChannels` unconditionally spread the PagerDuty, Opsgenie and Email defaults regardless of the selected type. PagerDuty and Opsgenie both use a field named `description`, Opsgenie's spread came last and won, so **PagerDuty's Description box was prefilled with the Opsgenie template**. It went unnoticed because the value the test asserted against (`pagerDutyDescriptionDefaultVaule`) was itself a copy of the Opsgenie template — the test encoded the bug.

#### Fix Strategy

Prefills come from a single per-type map, and only the selected type's entry is applied — to `selectedConfig` and, via `formInstance.setFieldsValue`, to the form store that holds the values of unmounted fields. The mock is corrected to PagerDuty's actual template (and its typo'd name fixed), so the existing PagerDuty assertion now fails against the old code.

---

### 🧪 Testing Strategy

- Tests added/updated: `src/container/AllAlertChannels` — 124 passing. New cases for the Google Chat dropdown option, both prefilled templates, the URL validation error notification, and the exact `googlechat_configs` request body; plus two cases that drive the Type dropdown (Slack ⇄ Google Chat swaps both templates; Opsgenie → PagerDuty prefills PagerDuty's description). Corrected the PagerDuty description mock.
- Manual verification: pending — needs a real Google Chat space webhook to confirm Save, Test, edit-and-persist, and that a firing/resolving alert posts with the resolved message respecting the toggle.
- Edge cases covered: non-`chat.googleapis.com` and `http` URLs rejected client-side; empty Title/Description fall back to the backend defaults; switching the Type dropdown in and out of Google Chat restores the right templates; Opsgenie → PagerDuty no longer leaks Opsgenie's description.

Reverting only `CreateAlertChannels/index.tsx` makes the Opsgenie → PagerDuty case fail, along with the pre-existing PagerDuty description case — i.e. both new tests are load-bearing.

---

- Blast radius: the New / Edit Notification Channel form only. No change to alert evaluation, routing, or any existing channel's request payload.
- Potential regressions: the prefill rewrite touches every channel type, not just Google Chat — the risk is a type showing the wrong prefilled template after switching the dropdown, which the two new tests cover. Switching to Email re-applies `send_resolved: true`; that was already true of the stored config, and the switch now visibly matches what gets sent.
- Rollback plan: revert the PR. Nothing is persisted by the frontend alone, and channels of type `googlechat` created against the backend PR remain valid server-side.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Feature |
| Description | Google Chat can now be selected as a notification channel — paste a space's incoming webhook URL, optionally customise the title and description templates, and alerts (and their resolutions) post to that space. |

---

### 📋 Checklist
- [ ] Manually tested
- [x] Breaking changes documented — none; this only adds a channel type

---

## 👀 Notes for Reviewers

The diff is larger than "add one channel type" because the per-type prefill handling was consolidated rather than extended with a seventh special case — see Bug Context for the defect that surfaced. Commits are ordered so each stands on its own: type/defaults/validation, then form fields, then the create/edit/test wiring, then tests.

---

### DIFF
--- a/frontend/src/container/AllAlertChannels/__tests__/CreateAlertChannel.test.tsx
+++ b/frontend/src/container/AllAlertChannels/__tests__/CreateAlertChannel.test.tsx
@@ -1,18 +1,28 @@
 import CreateAlertChannels from 'container/CreateAlertChannels';
 import { ChannelType } from 'container/CreateAlertChannels/config';
+import { GoogleChatInitialConfig } from 'container/CreateAlertChannels/defaults';
 import {
+	googleChatDescriptionDefaultValue,
+	googleChatTitleDefaultValue,
 	opsGenieDescriptionDefaultValue,
 	opsGenieMessageDefaultValue,
 	opsGeniePriorityDefaultValue,
 	pagerDutyAdditionalDetailsDefaultValue,
-	pagerDutyDescriptionDefaultVaule,
+	pagerDutyDescriptionDefaultValue,
 	pagerDutySeverityTextDefaultValue,
 	slackDescriptionDefaultValue,
 	slackTitleDefaultValue,
 } from 'mocks-server/__mockdata__/alerts';
 import { server } from 'mocks-server/server';
 import { rest } from 'msw';
-import { act, fireEvent, render, screen, waitFor } from 'tests/test-utils';
+import {
+	act,
+	fireEvent,
+	render,
+	screen,
+	userEvent,
+	waitFor,
+} from 'tests/test-utils';
 
 import { testLabelInputAndHelpValue } from './testUtils';
 
@@ -225,7 +235,7 @@ describe('Create Alert Channel', () => {
 				);
 
 				expect(descriptionTextArea).toHaveTextContent(
-					pagerDutyDescriptionDefaultVaule,
+					pagerDutyDescriptionDefaultValue,
 				);
 			});
 			it('Should check if Severity label, info (help_pager_severity), and textbox are displayed properly', () => {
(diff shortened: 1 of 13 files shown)


frontend/src/container/AllAlertChannels/__tests__/CreateAlertChannel.test.tsx
```diff
@@ -419,5 +422,99 @@ describe('Create Alert Channel', () => {
...
+				);
+
+				fireEvent.change(screen.getByTestId('channel-name-textbox'), {
+					target: { value: 'gchat-channel' },
+				});
+				fireEvent.change(screen.getByTestId('webhook-url-textbox'), {
+					target: { value: validWebhookUrl },
+				});
+
+				fireEvent.click(
```
AshwinBhatkal: use userEvent

frontend/src/container/CreateAlertChannels/index.tsx
```diff
@@ -60,38 +74,22 @@ function CreateAlertChannels({
...
-       {{- with .CommonLabels.Remove .GroupLabels.Names }}
-         {{- range $index, $label := .SortedPairs -}}
-           {{ if $index }}, {{ end }}
-           {{- $label.Name }}="{{ $label.Value -}}"
-         {{- end }}
-       {{- end -}}
-       )
-     {{- end }}`,
-	});
+		...(preType === ChannelType.GoogleChat
```
AshwinBhatkal: why separate handling for GoogleChat. Are the other channels correct?

ahrefabhi (approved):

PR #12345, opened 2026-07-30
fix(dashboard-v2): open the View modal and editor on the panel's own query

ahrefabhi:
## Pull Request

---

### 📄 Summary

Two independent V2 dashboard bug fixes.

**1. The View modal and the panel editor showed the previously opened panel's query.** Open one panel's View modal, close it, open another — the builder still showed the first panel's query. Same one-step-behind symptom on the editor route (most visible on PromQL panels).

**2. Bar and time-series panels reordered their series between identical refreshes.** On a stacked bar this recomposed every bar; on a time series it reshuffled draw and legend order.

_To be added — both fixes reproduced and verified via the tests below; reviewer-facing recording pending._

#### Issues closed by this PR

Closes https://github.com/SigNoz/pulse-pod/issues/156

---

### ✅ Change Type

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### 1. Stale panel query

**Root cause.** `QueryBuilderProvider` is mounted above the router, so neither closing the modal nor navigating to the editor re-reads the URL — `currentQuery`/`stagedQuery` keep pointing at the panel you just left. Both surfaces only re-seeded that context in a mount effect, one render too late for anything that reads the query once on mount: PromQL inputs are uncontrolled (`defaultValue`) and keyed by `q.name` (always `A`), so they neither re-sync nor remount; `QueryAddOns` picks which add-on rows are visible once (`initializedRef`). The previous panel's `stagedQuery` also survived the close, and `usePanelEditorQuerySync` commits any staged query on mount. V1's editor is exposed identically — its View modal just never renders a builder for PromQL.

**Fix.** `useViewPanel.openView`/`openViewWithQuery` and `useOpenPanelEditor` write the panel's query to `compositeQuery` (as V1's `WidgetHeader.onEditHandler` does, so a View/editor URL survives refresh and sharing) **and** seed the shared builder with it before navigating — the URL alone isn't enough, since the provider applies it in an effect. One shared `Panels/utils/getPanelBuilderQuery.ts` produces both. Two paths opt out deliberately: the View → editor handoff (its builder already holds that query) and new-panel creation (`PanelEditorPage` reads a `compositeQuery` at mount as an explorer export).

Seeding uses `resetQuery`, **not** `initQueryBuilderData`: the latter sets a *staged* query, and `useSyncTimeOnStagedQueryChange` dispatches `UpdateTimeInterval` when one non-null staged id replaces another — refetching the whole grid on modal open. `resetQuery` clears it instead (a `null` hop never triggers that), which also fixes the stale-draft cause and keeps the provider's URL hydration — and with it the legacy `filters.items` normalisation drilldowns rely on.

#### 2. Series order

**Root cause.** The response order isn't stable, at two levels: `pkg/querier/querier.go` builds `results` with `maps.Values(...)`, and `pkg/querier/consume.go` collects one query's series out of a map (`for k, s := range seriesMap`) — so even a single-query panel flips. Pre-V2 was insensitive to this because legends and colours key off labels, not position; bar panels were the exception and the one place it explicitly sorted (`utils/getSortedSeriesData`, descending mean, five call sites). V2 had no equivalent. It affects drilldown too: `enrichChartClick` resolves `series[seriesIndex - 1]`.

**Fix.** A port of the V1 sort (V2 duplicates rather than reaching across the split), in two pieces so it isn't a bar-panel local:

- `visualization/charts/utils/sortByMeanDesc.ts` — the mean/tiebreak rules, generic over `getValues`/`getKey`, in the shared chart layer so other bar consumers can adopt it in one line.
- `Panels/utils/sortSeriesByMean.ts` — an 18-line `PanelSeries` adapter, called by both the bar and time-series renderers inside their existing `flatSeries` memo (the config, the aligned data and click attribution all index that one list).

Three deviations from V1, each fixing something that would undercut the change: it doesn't mutate (V1 sorts in place, into the react-query cache); `NaN` no longer poisons the mean (V1's comparator returns `NaN`, which `sort` treats as equal); and equal means tiebreak on `queryName|aggIndex|kind|labels` rather than falling back to the wire order.

It stops one level above `BarChart`, which was the review's first suggestion — both alternatives were measured. Ordering only the stacking sequence renders wrong: uPlot draws a stacked bar's segments in series-index order and needs the total at index 1 so shorter segments paint over it. A real reorder inside the component renders fine but desyncs what indexes the *caller's* arrays — `ChartManager` zips `config.getConfig().series[i]` with the `alignedData` prop, and `enrichChartClick` indexes the renderer's list — and would resort the curated billing and status-code charts.

**Scope:** bar and time series. Pie slice order and the Value panel picking a random query's number stay wire-order dependent, exactly as pre-V2.

---

### 🧪 Testing Strategy

**1. Stale panel query** — `useViewPanel.test.tsx` (correct context on first render; query in `compositeQuery` with no `graphType`; no stale draft; ×2 that one staged query is never swapped for another, the exact grid-refetch condition), `useViewPanel.promql.test.tsx` and `panelEditorUrlQuery.promql.test.tsx` (real builder, asserting the PromQL input across a panel switch — both reproduce the report verbatim and were verified to fail without the fix), `useViewPanel.addons.test.tsx`, `useOpenPanelEditor.test.tsx`, plus `usePanelActionItems.test.tsx` for the new signatures.

**2. Series order** — `sortByMeanDesc.test.ts` (8 cases on the shared util) and `sortSeriesByMean.test.ts` (7 cases on the adapter): descending order, no-finite-points sinking, `NaN`/`null` not poisoning the mean, identical output across reversed inputs, tiebreak on equal means, non-mutation, empty input.

**Verification:** `tsgo --noEmit` clean, `pnpm lint:js --quiet` clean, `pnpm build` succeeded, 1184 tests across 153 suites green (`DashboardPageV2` + `visualization`).

---

- **Blast radius:** V2 dashboards only; no V1 file modified. `openView` has one caller and `openPanelEditor` three; the new shared sort util has no other callers yet.
- **By design:** a plain View now hydrates from `compositeQuery`, so its draft queries are `toPerses(fromPerses(saved))` — the same round trip the dirty check already compares through. Series order on bar and time-series panels changes (deterministic instead of arbitrary); colours and labels don't move, since both derive from the label string. On anomaly panels the companion series join the same sort; nothing in V2 pairs them by index.
- **Known gap:** `resetQuery` doesn't set `queryType`, so a PromQL panel opened right after a builder panel shows the wrong query-type tab for one render before hydration corrects it. Pre-existing, not introduced here.
- **Rollback:** each commit reverts independently; no migrations, persisted state, or API surface touched.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | Dashboards (V2): opening a panel's View modal or editor no longer shows the previously opened panel's query. |
| Description | Dashboards (V2): bar and time-series panel series are ordered consistently across refreshes instead of following the server's response order. |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

- **Five commits, two fixes.** Stale query: `f9e8a6c` + `f5426d8`, with `d66db74` speeding up its suites. Series order: `66e274a` (bar) + `64a8b85` (shares the sort, extends it to time series) — reviewable or splittable on their own.
- The `resetQuery` vs `initQueryBuilderData` choice is load-bearing and non-obvious: `initQueryBuilderData` looks like the better fit (it also sets `queryType`) but refetches every panel on the grid. There's a comment at the call site and a test for it — please keep both.
- **Review comment addressed in `64a8b85`:** the sort is no longer bar-local. It stops above `BarChart` for the reasons under Fix Strategy 2.
- **Follow-up, not in this PR:** the PromQL query/legend inputs are uncontrolled, so they also ignore query changes that land while mounted (Reset, browser Back/Forward). The fix touches shared V1 code used by the V1 editor and `FormAlertRules`, and is a V1 bug fix too, so it ships separately. This PR holds either way.
- **Follow-up, suggested:** ordering `results` by the request's query order once at `queryV5/v5ResponseData.ts` would make pie, value and list order-independent too. Needs `requestPayload` threaded into three getters, so it's its own PR.
- Screenshots/recording still to be attached.


AshwinBhatkal (approved):

AshwinBhatkal (dismissed):

frontend/src/pages/DashboardPageV2/DashboardContainer/Panels/kinds/BarChartPanel/utils/sortSeriesByMean.ts
```diff
@@ -0,0 +1,45 @@
...
+	return `${series.queryName}|${series.aggregation.index}|${series.kind}|${labels}`;
+}
+
+/**
+ * Orders series by descending mean so stacked bars keep the tallest segments at the base instead
+ * of burying short series (V1 parity: `utils/getSortedSeriesData`). Series with no finite points
+ * sink to the bottom; equal means tiebreak on `seriesKey`, since a stable sort would otherwise
+ * fall back to the response order and the backend doesn't guarantee one.
+ */
+export function sortSeriesByMeanDesc(series: PanelSeries[]): PanelSeries[] {
```
AshwinBhatkal: suggestion: what if this was inside UPlot render itself

AshwinBhatkal: I was expecting it's usage to be at a higher level before panelwrapper or at the root level uplot

So that any other series is always sorted before rendering. Similar to the issue we found in the customer instance of flipping time series

ahrefabhi: can't move it closer to uplot @AshwinBhatkal, because sorting needs to happen before we create the config and add series in the config, have made the util generic and moved it to charts folder though.

AshwinBhatkal (approved):

PR #12351, opened 2026-07-30
fix(clickhouseReader): bind user-supplied values in IN() lists

kah-ja:
### What

`GetLogFieldsFromNames`, `GetSpanAttributeKeysByNames`, `GetMinAndMaxTimestampForTraceID`, `GetUpdatedMetricsMetadata` and `GetNormalizedStatus` build their `IN (...)` list by joining the incoming values with `','` and pasting the result into the statement:

```go
query := fmt.Sprintf("... where name in ('%s') ...", r.logsDB, r.logsAttributeKeys, strings.Join(fieldNames, "','"))
err := r.db.Select(ctx, &attributes, query)
```

Those values come from the request payload, not from configuration:

- `fieldNames` is `logsv3.GetFieldNames(...)`, i.e. `filters.items[].key.key`, `groupBy[].key`, `orderBy[].columnName` and `aggregateAttribute.key` of a `/api/v3/query_range` request (`http_handler.go` L3617 and L3971), which is behind `ViewAccess`
- `names` in `GetSpanAttributeKeysByNames` comes from the same helper (L3561)
- `traceID` is the trace-id filter value of the same request
- the metric names come from the builder query / PromQL selector on the metrics paths

A value that contains a single quote therefore does not end up as a literal, it changes the shape of the statement. In other words these are SQL injection sinks reachable by any authenticated user, including the lowest (viewer) role. I am not posting a payload here.

### How

A small `inClause()` helper returns the `?, ?, ...` placeholder list plus the argument slice, and the values are passed to `db.Select` / `db.Query` as bind parameters. This is the pattern `validateResourceKeys` in the same file already uses.

Notes:

- an empty input list now skips the lookup entirely, which is equivalent to the previous `in ('')` (it matched nothing)
- the two queries that reference the list twice (the `UNION ALL` variants behind the metrics-reduction flag) bind the argument slice twice, once per arm
- in `GetSpanAttributeKeysByNames` the `SHOW CREATE TABLE` lookup moved above the key lookup so the row scan can live inside the guarded block; no behaviour change
- the `IN` lists in `app/inframetrics` are built from package constants and are left untouched

### Testing

- `go build`, `go vet` and `gofmt` clean on the touched package
- built the enterprise binary from this branch, ran it inside the `signoz/signoz:v0.135.0` image against ClickHouse 25.5.6 next to an unpatched v0.135.0 container sharing the same ClickHouse
- `/api/v3/query_range` with an ordinary attribute key returns the identical response on both containers (HTTP 200, same body, same latency)
- with a key containing a quote, `system.query_log` on the patched container shows the value arriving as one escaped string literal and the statement finishing in single-digit ms, while the unpatched container parses it as SQL

Happy to split this per function or add a unit test for `inClause` if you prefer.

### DIFF
--- a/pkg/query-service/app/clickhouseReader/reader.go
+++ b/pkg/query-service/app/clickhouseReader/reader.go
@@ -2626,6 +2626,20 @@ func (r *ClickHouseReader) GetLogFields(ctx context.Context) (*model.GetFieldsRe
 	return &response, nil
 }
 
+// inClause builds a bind-parameter list ("?, ?, ...") for values together with
+// the matching argument slice. Values that originate from the API must be bound
+// this way instead of being concatenated into the statement, so that ClickHouse
+// escapes them.
+func inClause(values []string) (string, []any) {
+	placeholders := make([]string, len(values))
+	args := make([]any, len(values))
+	for i, value := range values {
+		placeholders[i] = "?"
+		args[i] = value
+	}
+	return strings.Join(placeholders, ", "), args
+}
+
 func (r *ClickHouseReader) GetLogFieldsFromNames(ctx context.Context, fieldNames []string) (*model.GetFieldsResponse, *model.ApiError) {
 	ctx = ctxtypes.NewContextWithCommentVals(ctx, map[string]string{
 		instrumentationtypes.TelemetrySignal:  telemetrytypes.SignalLogs.StringValue(),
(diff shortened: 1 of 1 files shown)


kah-ja:
CLA signed and rebased onto main. The conflict was GetNormalizedStatus being removed in a850f548f, so that hunk is dropped; the change now covers GetLogFieldsFromNames, GetSpanAttributeKeysByNames, GetMinAndMaxTimestampForTraceID and GetUpdatedMetricsMetadata. go build, go vet and gofmt are clean.

kah-ja:
Any updates?

kah-ja:
ping

PR #12352, opened 2026-07-30
fix(alertmanager): reject file references in channels sent over the API

kah-ja:
### What

Notifier configs can carry a secret indirectly instead of inline, by naming a file: `basic_auth.password_file`, `authorization.credentials_file`, `oauth2.client_secret_file`, `tls_config.ca_file` / `cert_file` / `key_file`, the opsgenie `api_key_file`, the `files` list of an http header, and the equivalents on the other notifier types. Alertmanager reads those paths from the local filesystem when the notification is sent. For a config file that an operator owns and edits on disk, that is exactly the intended behaviour.

Channels, however, are submitted as JSON over the API and decoded straight into `config.Receiver`:

```go
receiver, err := alertmanagertypes.NewReceiver(string(body))   // handler.go, test/create/update channel
```

`NewReceiver` only applies per-notifier defaults, it does not restrict which fields a caller may set. So a channel can name any path that the SigNoz process can read, and the content is then sent to the webhook target named in the same channel. SigNoz runs as root in the published image, and `/api/v1/channels/test` (plus the deprecated `/api/v1/testChannel`) only requires the editor role, so this does not even need an admin.

### How

`Receiver.ValidateFileReferences` walks the decoded receiver with reflection and rejects any populated field whose serialized name ends in `_file`, plus the `files` list of an http header. Reflection rather than an explicit list because upstream names these fields consistently across notifier types, and a new notifier (or a new upstream field) is then covered without anyone having to remember this check.

It is called from the three handlers that build a receiver from a request body: test channel, create channel, update channel.

Deliberately **not** put inside `NewReceiver`: the same constructor also parses receivers that are already stored (`channel.go`, `config.go`). Validating there would make an existing stored channel unloadable rather than just rejecting new input, so the check sits at the API boundary.

Everything a channel needs can still be provided inline, that path is untouched.

### Testing

- new table test in `receiver_file_reference_test.go`: inline password and inline slack api_url pass, `password_file`, `credentials_file`, `tls_config.ca_file` and opsgenie `api_key_file` are rejected
- `go vet` and `gofmt` clean, `go test ./pkg/types/alertmanagertypes/... ./pkg/alertmanager/...` passes except `TestEmailRejected` in `alertmanagernotify/email`, which fails the same way on unmodified `main` here ("smtp: server already closed"), so it is unrelated to this change
- built the enterprise binary from this branch, ran it in the `signoz/signoz:v0.135.0` image next to an unpatched v0.135.0 container, both pointed at a webhook receiver:
  - `POST /api/v1/channels/test` with `basic_auth.password_file` set: unpatched returns 204 and the receiving endpoint gets the file content as the basic-auth password; patched returns 400 with `"webhook_configs.http_config.basic_auth.password_file" is not supported in a channel ...` and sends nothing
  - same call with an ordinary inline `password`: 204 on the patched build and the webhook target receives the notification as before

Happy to adjust the wording of the error or to turn the reflection walk into an explicit field list if you would rather have that.

### DIFF
--- a/pkg/alertmanager/signozalertmanager/handler.go
+++ b/pkg/alertmanager/signozalertmanager/handler.go
@@ -72,6 +72,11 @@ func (handler *handler) TestReceiver(rw http.ResponseWriter, req *http.Request) 
 		return
 	}
 
+	if err := receiver.ValidateFileReferences(); err != nil {
+		render.Error(rw, err)
+		return
+	}
+
 	err = handler.alertmanager.TestReceiver(ctx, claims.OrgID, receiver)
 	if err != nil {
 		render.Error(rw, err)
@@ -196,6 +201,11 @@ func (handler *handler) UpdateChannelByID(rw http.ResponseWriter, req *http.Requ
 		return
 	}
 
+	if err := receiver.ValidateFileReferences(); err != nil {
+		render.Error(rw, err)
+		return
+	}
+
 	err = handler.alertmanager.UpdateChannelByReceiverAndID(ctx, claims.OrgID, receiver, id)
 	if err != nil {
 		render.Error(rw, err)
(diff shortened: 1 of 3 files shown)


kah-ja:
Any updates?

kah-ja:
ping

therealpandey:
Hi @kah-ja,
V1 channel apis are being deprecated in favor of v2 notification channel apis.

Appreciate the effort that was put here, but we're going to be deleting this endpoints soon.

PR #12367, opened 2026-08-01
feat(tests): fixed network alias for zeus wiremock container

therealpandey:
## Summary

- Adds an opt-in `--zeus-network-alias <alias>` pytest option; default behavior is unchanged when unset.
- When set, the zeus WireMock container gets the given docker network alias and its `container_configs` address uses that alias instead of the random container name, making the `signoz:integration` build-arg `ZEUSURL` deterministic (e.g. `http://zeus:8080`).
- This is a prerequisite for CI to prebuild the integration image once and share it across jobs. The gateway needs no alias: `SIGNOZ_GATEWAY_URL` is runtime container env, not a build-arg.

Usage: `uv run pytest --basetemp=./tmp/ -vv --zeus-network-alias zeus integration/bootstrap/setup.py::test_setup`

### DIFF
--- a/tests/fixtures/http.py
+++ b/tests/fixtures/http.py
@@ -31,12 +31,13 @@ def zeus(
     def create() -> types.TestContainerDocker:
         container = WireMockContainer(image="wiremock/wiremock:2.35.1-1", secure=False)
         container.with_network(network)
-        if pytestconfig.getoption("--zeus-network-aliases"):
-            container.with_network_aliases("zeus")
+        network_alias = pytestconfig.getoption("--zeus-network-alias")
+        if network_alias:
+            container.with_network_aliases(network_alias)
         container.start()
 
-        if pytestconfig.getoption("--zeus-network-aliases"):
-            container_address = "zeus"
+        if network_alias:
+            container_address = network_alias
         else:
             container_address = container.get_wrapped_container().name
 
--- a/tests/fixtures/http.py
+++ b/tests/fixtures/http.py
@@ -31,7 +31,14 @@ def zeus(
     def create() -> types.TestContainerDocker:
         container = WireMockContainer(image="wiremock/wiremock:2.35.1-1", secure=False)
         container.with_network(network)
+        if pytestconfig.getoption("--network-aliases"):
+            container.with_network_aliases("zeus")
         container.start()
+
+        if pytestconfig.getoption("--network-aliases"):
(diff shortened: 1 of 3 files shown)


tests/fixtures/http.py
```diff
@@ -31,8 +31,15 @@ def zeus(
     def create() -> types.TestContainerDocker:
         container = WireMockContainer(image="wiremock/wiremock:2.35.1-1", secure=False)
         container.with_network(network)
+        if pytestconfig.getoption("--network-aliases"):
+            container.with_network_aliases("zeus")
         container.start()
 
+        if pytestconfig.getoption("--network-aliases"):
```
therealpandey:
```suggestion
        if pytestconfig.getoption("--zeus-network-aliases"):
```

If not passed, default to existing behaviour.

tests/fixtures/http.py
```diff
@@ -84,8 +91,15 @@ def create() -> types.TestContainerDocker:
         container = WireMockContainer(image="wiremock/wiremock:2.35.1-1", secure=False)
         container.with_exposed_ports(8080)
         container.with_network(network)
+        if pytestconfig.getoption("--network-aliases"):
+            container.with_network_aliases("gateway")
         container.start()
 
+        if pytestconfig.getoption("--network-aliases"):
```
therealpandey: Why do we need a network alias for gateway? It's passed via the env and has no contribution in the build step.

tests/fixtures/http.py
therealpandey: Renamed to `--zeus-network-aliases` in 42f19dd. Default behaviour unchanged when not passed.

tests/fixtures/http.py
therealpandey: Right — `SIGNOZ_GATEWAY_URL` is runtime env, not a build-arg. Dropped the gateway alias and reverted the gateway fixture to its original state in 42f19dd; only zeus gets the alias now.

therealpandey:
Follow-up in 77286f5: renamed to `--zeus-network-alias` (singular) and it now takes the alias as its value (`action="store"`, default `""`) — the fixture uses the passed value, nothing hardcoded. Unset keeps existing behavior; a bare `--zeus-network-alias` errors via argparse ("expected one argument"). Docs/help updated with example `--zeus-network-alias zeus`.

PR #11245, opened 2026-05-09
feat(identn): add trusted-header IdentN provider

CaffeineDuck:
## Summary

This PR adds a new `IdentN` provider, `trustedheaderidentn`, that authenticates requests based on identity headers injected by an upstream reverse proxy. It is the implementation of `IdentN` invited in #6197.

The motivating use case is operators running SigNoz behind a forward-auth-style identity proxy (Authentik forward-auth, `oauth2-proxy`, Cloudflare Access, etc.). The proxy is the source of truth for the user's identity; SigNoz simply trusts the headers it receives and creates a session — no local password/SSO flow needed.

The provider is disabled by default and only activates when explicitly enabled in config. When enabled, SigNoz logs a startup warning, since the design assumes the proxy strips client-supplied identity headers — without that, any client could forge identity.

Closes #6197

## Behaviour

- `Test()` returns `true` when the configured email header is non-empty.
- `GetIdentity()`:
  - Reads the configured email header and validates it.
  - Looks up the user by email across all instance-owned organizations (mirrors `pkg/modules/session/implsession`).
  - If found and not the root user → returns the identity.
  - If not found and `auto_provision` is `true` AND the instance has exactly one org → creates a Viewer-role user via `user.Setter.GetOrCreateUser` and returns the identity.
  - If not found and `auto_provision` is `false` → 401.
  - Refuses to auto-provision when multiple orgs exist (operator must disable auto-provision or add per-org pinning).
  - Refuses root-user auth via headers (matches the SAML/OIDC posture in `implsession.CreateCallbackAuthNSession`).

## Config example

```yaml
identn:
  trusted_header:
    enabled: true
    email_header: X-Authentik-Email   # or X-Forwarded-Email (default), etc.
    name_header: X-Authentik-Name     # optional
    auto_provision: false             # opt-in
```

## Files

- `pkg/types/authtypes/identn.go` — `IdentNProviderTrustedHeader`
- `pkg/types/authtypes/authn.go` — `AuthNProviderTrustedHeader` (added to `Enum()` for completeness)
- `pkg/identn/config.go` — `TrustedHeaderConfig`, default, validation
- `pkg/identn/resolver.go` — wiring
- `pkg/identn/trustedheaderidentn/provider.go` — the IdentN itself
- `pkg/identn/trustedheaderidentn/provider_test.go` — unit tests covering Test/GetIdentity, auto-provision on/off, multi-org refusal, root-user refusal, missing/invalid email
[...]

PR #12379, opened 2026-08-02
feat(identn): add trusted-header auth with verified proxy provenance

### LINKED ISSUE #6197: [authentication] trusted header authentication

gamunu:
## Summary

We run SigNoz behind Teleport, which fronts internal apps and mints a signed JWT assertion per request. Reaching it meant a second set of SigNoz passwords, with no supported way to hand the already-authenticated user through.

Builds on #11245, which trusted the identity headers on sight. Any client that could reach the pod could set `X-Forwarded-Email` and become that user. In Kubernetes every pod shares the pod CIDR, so a peer address cannot tell the proxy from an attacker.

This PR adds the proof that a request came from the proxy. `trust.mode` is required when the resolver is enabled, and no mode trusts network placement alone.

Closes #6197. Supersedes #11245.

## Trust modes

- `jwt`: a signed assertion verified against a JWKS endpoint, then issuer, audience, expiry and not-before. Identity comes from a verified claim, so the email headers are never read and there is no shared secret to leak. Recommended, and the only mode an attacker with network access cannot forge.
- `secret`: a shared secret header, compared in constant time over SHA-256 digests, so neither the secret's contents nor its length leaks through timing. Identity comes from the email headers. Weaker, because anything that can read the pod spec can read the secret.

## Config example

Teleport application access mints the assertion and publishes the keys at `/.well-known/jwks.json`. `issuer` is the cluster name, `audience` is the application's URI:

```yaml
identn:
  trusted_header:
    enabled: true
    trust:
      mode: jwt
      jwt:
        assertion_header: Teleport-Jwt-Assertion
        jwks_url: https://teleport.example.com/.well-known/jwks.json
        issuer: teleport.example.com
        audience: https://signoz.example.com
        identity_claim: sub
    auto_provision: true
```

oauth2-proxy and Authentik forward-auth inject an email header instead of an assertion, so the secret is what proves provenance:

```yaml
identn:
  trusted_header:
    enabled: true
    trust:
      mode: secret
      secret:
        header: X-Signoz-Trusted-Secret
        value: ${SIGNOZ_TRUSTED_HEADER_SECRET}
    email_headers:
      - X-Forwarded-Email
    name_headers:
      - X-Forwarded-User
    auto_provision: false
    logout_redirect_url: https://proxy.example.com/oauth2/sign_out
```

The proxy must overwrite or strip every header named here on inbound requests. Teleport passes through any header not listed in that application's `rewrite.headers`, so an unlisted assertion or email header lets a client set its own. `conf/example.yaml` carries the full annotated block.

## Behaviour

- Disabled by default. Refuses to boot without a `trust.mode`, on a header name containing an underscore (nginx and similar drop those silently), on an empty `assertion_header`, and on a `trusted_proxies` entry with host bits set.
- 401 for a pending invite, an ambiguous match, an email resolving to a user who cannot authenticate this way, a header sent more than once (a proxy that appends instead of replacing), and any attempt to resolve to a root user.
- Auto-provisioning creates a Viewer, refuses to guess when several orgs exist, and uses `CreateUser` rather than `GetOrCreateUser`, since adoption would silently claim a pending-invite row.
- `trusted_proxies` is an optional peer CIDR allowlist. It is an extra layer on top of the trust mode, never a control to rely on by itself.
- The public config gains `identN.trustedHeader.{enabled,logoutRedirectUrl}`, carrying neither the trust mode nor any secret. The UI uses it to skip the login form and to send logout to the proxy.

## Test plan

- [x] `go test -race ./pkg/identn/... ./pkg/http/middleware/... ./pkg/global/...`
- [x] `golangci-lint run` clean on the changed packages
- [x] `pnpm jest src/api src/providers/App src/utils`, 206 pass; `pnpm tsgo --noEmit`; `pnpm build`
- [x] `pytest tests/integration/tests/trustedheaderauthn/`
- [x] Manual, both modes, behind oauth2-proxy with Keycloak as the IdP

The manual pass is the box #11245 left unchecked. In `secret` mode, login through Keycloak lands in the UI with no login form, an unknown email provisions as `signoz-viewer`, and logout reaches the proxy's sign-out endpoint. Eight forgery variants all 401, including a correct secret naming the root user, and duplicated headers.

In `jwt` mode I tested against real Keycloak `id_token`s. A valid assertion for one user plus `X-Forwarded-Email` naming root resolves to the user in the assertion, which confirms the email headers are ignored once the proof carries identity. Rejected: tampered payload, tampered signature, `alg=none` forging a root claim, malformed input, absent and duplicated assertion headers, wrong audience, and an unreachable `jwks_url`. The last two are distinguishable in the log (`get keys failed: 404` against `failed to verify id token signature`) while the caller gets the same opaque code.

That pass found one real bug, fixed in the last commit. The decision on whether a 401 had a session worth rotating read `getIsProxyAuthMode()`. That is a module singleton, and the preflight effect only sets it once the global config lands, while the protected queries fire earlier off `isLoggedIn` in localStorage. With a deliberately wrong secret this produced two rotate attempts and three hits on the proxy's sign-out endpoint, two of them inside the same second. That is the logout loop the preceding commit was meant to prevent. The interceptor now awaits preflight before reading either mode flag, and only for 401s, so password sessions keep rotating.

### DIFF
--- a/.github/workflows/integrationci.yaml
+++ b/.github/workflows/integrationci.yaml
@@ -62,6 +62,7 @@ jobs:
           - spanmapper
           - querier_json_body
           - querier_skip_resource_fingerprint
+          - trustedheaderauthn
           - ttl
           - clickhousecluster
           - metricreduction
--- a/conf/example.yaml
+++ b/conf/example.yaml
@@ -407,6 +407,40 @@ identn:
   impersonation:
     # toggle impersonation identN, when enabled, all requests will impersonate the root user
     enabled: false
+  trusted_header:
+    # toggle trusted-header identN. When enabled, identity comes from a fronting
+    # identity-aware proxy. Cannot be enabled together with impersonation.
+    enabled: false
+    trust:
+      # how proxy provenance is established. Required when enabled.
+      # "jwt": verify a signed assertion against a JWKS endpoint. Recommended:
+      #   it is the only mode an attacker with network access cannot forge.
+      # "secret": require a secret header, compared in constant time. Weaker,
+      #   because anything that can read the pod spec can read the secret.
+      mode: ""
+      jwt:
+        assertion_header: Teleport-Jwt-Assertion
+        jwks_url: ""
+        issuer: ""
+        audience: ""
+        identity_claim: sub
+      secret:
+        header: ""
+        value: ""
+    # headers carrying the authenticated user's email. Ignored in jwt mode.
+    email_headers:
+      - X-Forwarded-Email
+    # optional headers setting a display name when provisioning
+    name_headers: []
+    # create a Viewer for an unknown email. Note that Viewer can run raw
+    # ClickHouse SQL, so this is not a small grant.
+    auto_provision: false
+    # optional peer CIDR allowlist. Defence in depth only: in Kubernetes every
+    # pod shares the pod CIDR, so this cannot distinguish the proxy from any
+    # other pod. Never rely on it as the only control.
+    trusted_proxies: []
+    # where the UI sends a user on logout, so the proxy session ends too
+    logout_redirect_url: ""
 
 ##################### Service Account #####################
 serviceaccount:
--- a/conf/example.yaml
+++ b/conf/example.yaml
@@ -410,6 +410,9 @@ identn:
   trusted_header:
     # toggle trusted-header identN. When enabled, identity comes from a fronting
     # identity-aware proxy. Cannot be enabled together with impersonation.
+    # The API port must not be reachable except through that proxy: the
+    # provider trusts the header on any request that passes the trust check
+    # below, direct access bypasses the proxy's authentication entirely.
     enabled: false
     trust:
       # how proxy provenance is established. Required when enabled.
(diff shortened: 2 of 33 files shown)


mallaghi (approved):

simonepri (approved):

simonepri:
@gamunu any chance in splitting this change into smaller chunks? I suspect that would help maintainers review and submit this

PR #12382, opened 2026-08-03
fix(querybuilder): keep DateTime64 columns native in aggregations

tushar-signoz:
A numeric coercion of a `DateTime64` yields **seconds** since epoch, so `max(timestamp)` returned `1758113657.04` instead of an instant and the third-party APIs "Last Seen" column rendered as January 1970.

### What

Two rules, both keyed on the **physical column type** — no `FieldDataType` denotes a timestamp, which is how the declaration drifted from reality in the first place (`const.go` says `Number`, the column is `DateTime64(9)`):

- a time column is never coerced, so it reaches the aggregate natively and the driver hands back a `time.Time`
- a bare `timestamp` resolves to the intrinsic column alone — a same-named numeric attribute no longer joins the candidate union

```sql
-- before
max(multiIf(timestamp <> '', accurateCastOrNull(timestamp,'Float64'), NULL))  -- 1758113657.04
-- after
max(multiIf(timestamp <> toDateTime64(0, 9), timestamp, NULL))               -- 2025-09-17T12:54:17.040Z
```

The exists guard for a time column also moves out of the String bucket: `timestamp <> ''` becomes a typed epoch-zero comparison.

No change to `thirdpartyapi/translator.go` or the frontend — `ApiMonitoring/utils.tsx` already does `new Date(value)` over `number | string`, so restoring the instant is enough.

### Guardrails

Rule 2 is not a nicety. Without it a tenant with a numeric `timestamp` attribute gets a **hard failure**, not bad data:

```
DB::Exception: There is no supertype for types DateTime64(9, 'UTC'), Float64
```

Every pre-existing golden-SQL expectation is unchanged, so nothing outside `timestamp` moved — no count, group-by or guard semantics shift.

### Notes

`max` / `min` / `quantile` / `count` / `rate` keep working. `sum`, `avg` and the `rate_*` variants now **fail at the database** rather than returning a rescaled number — a 500 where there used to be a wrong value. Deliberate for now and pinned by tests; the follow-up is to reject them up front with a message naming the field.

Two unrelated bugs surfaced while writing the tests, neither addressed here:

1. `ThirdPartyApiRequest.Domain` / `.Endpoint` are decoded but never read anywhere in the module — scoping only works via `filter`.
2. On `/overview/domain`, only the endpoints query groups by `http_url`; `p99`/`error_rate`/`lastseen` pass `req.GroupBy` through unchanged, so with no caller groupBy they land on their own row and the UI cannot show them per endpoint. Pinned as `test_domain_info_metrics_are_not_per_endpoint`.

### Testing

Verified against a real ClickHouse 25.5 rather than inferred — seconds root cause, `toInt64` giving whole seconds, guard equivalence (0 disagreements incl. epoch-zero and a non-UTC server tz), the `Nullable(DateTime64)` result type that maps to `time.Time`, the `no supertype` failure, and which aggregates accept `DateTime64` natively.

- unit: temporal-column cases + the attribute-collision regression in `tracestelemetryschema`
- integration: new `thirdpartyapi` suite (19 cases across both routes — every response column asserted by value: `endpoints`, `lastseen`, `p99`, `rps`, `error_rate` at 0/25/100%, plus base filter, caller filter, caller groupBy, IPv4 + IPv6 masking, empty window, 401, 400) and `queriertraces/12_timestamp_aggregations.py` (11 cases pinning the supported and unsupported aggregations)
- 238 passed across the full traces + thirdpartyapi suites; `go test ./pkg/... ./ee/...` green; primus `go-lint` 0 issues; `py-fmt`/`py-lint` clean

### Fixture fix

`tests/fixtures/traces.py` wrote `kind_string` as the enum member name (`SPAN_KIND_CLIENT`), so **any** feature filtering on that column matched nothing — the third-party APIs base filter requires `kind_string = 'Client'`, and the first run of the new suite returned zero rows for every test because of it. The six kinds are now mapped explicitly, mirroring what the collector writes:

- `clickhouse_exporter_v3.go:328` → `SpanKind: otelSpan.Kind().String()`
- `pdata/ptrace/span_kind.go:38-54` → `Unspecified` / `Internal` / `Server` / `Client` / `Producer` / `Consumer`
- binds to `kind_string`: `writer.go:180` position 9 = the INSERT's column 9 (`clickhouse_exporter.go:51`), with `kind` before and `duration_nano` after
- upstream collector `main` is identical to the pinned v0.144.6, and the exporter's own test asserts the literal `"Server"`

An earlier version derived the string (`name` minus `SPAN_KIND_`, title-cased). That matched only because every OTel kind is a single word, so it was a coincidence rather than a correspondence — replaced with the explicit map.

Fixes https://github.com/SigNoz/engineering-pod/issues/5824

### DIFF
--- a/tests/integration/tests/thirdpartyapi/01_domain_list.py
+++ b/tests/integration/tests/thirdpartyapi/01_domain_list.py
@@ -1,0 +1,551 @@
+"""Integration tests for the third-party (external) API monitoring domain list.
+
+A translator over v5 builder queries, so these cover the contract it exposes (columns,
+grouping, base filter, IP masking) plus the `max(timestamp)` Last Seen from #5824.
+"""
+
+from collections.abc import Callable
+from datetime import UTC, datetime, timedelta
+from http import HTTPStatus
+
+import pytest
+
+from fixtures import types
+from fixtures.auth import USER_ADMIN_EMAIL, USER_ADMIN_PASSWORD
+from fixtures.thirdpartyapi import (
+    domain_column_index,
+    index_columns_by_query,
+    make_third_party_apis_request,
+    rows_by_domain,
+    scalar_result,
+)
+from fixtures.traces import TraceIdGenerator, Traces, TracesKind, TracesStatusCode
+
+# the placeholder the response layer writes where a query has no value for a row
+NO_VALUE = "n/a"
+
+
+# timestamp_attr seeds a numeric attribute named `timestamp`: resolution must keep the
+# intrinsic column, else the multiIf union fails with "no supertype for DateTime64, Float64".
+@pytest.mark.parametrize("noise", ["clean", "timestamp_attr"])
+def test_domain_list_last_seen_is_the_span_instant(
+    signoz: types.SigNoz,
+    create_user_admin: None,  # pylint: disable=unused-argument
+    get_token: Callable[[str, str], str],
+    insert_traces: Callable[[list[Traces]], None],
+    noise: str,
+) -> None:
+    """
+    Setup:
+    Two client spans to api.stripe.com, 3 and 1 minutes ago.
+
+    Tests:
+    Last Seen resolves to the newer span's instant. A numeric coercion of the
+    DateTime64 column would yield seconds since epoch (~1.7e9), which lands in
+    January 1970 once read as milliseconds.
+    """
+    extra_attrs = {"timestamp": 1.0} if noise == "timestamp_attr" else {}
+    now = datetime.now(tz=UTC).replace(microsecond=0)
+    newest = now - timedelta(minutes=1)
+    domain = f"api.stripe-{noise}.com"
+
+    insert_traces(
+        [
+            Traces(
+                timestamp=ts,
+                duration=timedelta(milliseconds=120),
+                trace_id=TraceIdGenerator.trace_id(),
+                span_id=TraceIdGenerator.span_id(),
+                name="POST /v1/charges",
+                kind=TracesKind.SPAN_KIND_CLIENT,
+                status_code=TracesStatusCode.STATUS_CODE_OK,
+                resources={"service.name": "checkout"},
+                attributes={
+                    "http.url": f"https://{domain}/v1/charges",
+                    "http.host": domain,
+                    **extra_attrs,
+                },
+            )
+            for ts in (now - timedelta(minutes=3), newest)
+        ]
+    )
+
+    response = make_third_party_apis_request(
+        signoz,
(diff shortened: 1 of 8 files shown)


tushar-signoz:
The changes are kept only for the traces module to limit the surface area this change touches.
Once i fix the bug where attributes have colliding names with intrinsic columns, I'll circle back to this

therealpandey (dismissed):

tests/integration/tests/thirdpartyapi/02_domain_info.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def request_domain_info(signoz: types.SigNoz, token: str, now: datetime):
+    return make_third_party_apis_request(
+        signoz,
+        token,
+        start_ms=int((now - timedelta(minutes=10)).timestamp() * 1000),
+        end_ms=int((now + timedelta(minutes=1)).timestamp() * 1000),
+        endpoint=DOMAIN_INFO_ENDPOINT,
+    )
```
krsoninikhil: This endpoint is not being used in the domain info page, all data in domain / endpoint page uses query_range API to get the details.

It'll be better to use either add tests for that or just remove this.

tests/integration/tests/thirdpartyapi/01_domain_list.py
```diff
@@ -0,0 +1,552 @@
...
+    result = scalar_result(response)
+    rows = rows_by_domain(result)
+    assert domain in rows, result["data"]
+
+    raw_last_seen = rows[domain][index_columns_by_query(result)["lastseen"]]
+    # accept an RFC 3339 instant or epoch millis: this pins the instant, not the encoding
+    if isinstance(raw_last_seen, str):
+        last_seen = datetime.fromisoformat(raw_last_seen.replace("Z", "+00:00"))
+    else:
+        last_seen = datetime.fromtimestamp(float(raw_last_seen) / 1000, tz=UTC)
```
krsoninikhil: Is the response type not deterministic here?

krsoninikhil (changes requested):

tests/integration/tests/thirdpartyapi/02_domain_info.py
tushar-signoz: Removing it

krsoninikhil (approved):

krsoninikhil:
Let's keep the new changes as separate commit after every review, it's easier to check what changed after last review.

therealpandey (approved):

PR #12396, opened 2026-08-04
fix(dashboard): count panel stats from the v2 spec

AshwinBhatkal:
## Pull Request

---

### 📄 Summary

Every `dashboard.panels.*` stat we report has been zero since v2 became the stored
dashboard shape. `NewStatsFromStorableDashboards` was never updated for the v2 schema,
so panel and per-signal counts are unusable in product analytics.

This teaches the counter to read the v2 spec. It reads **v2 only** — v1 rows contribute
to `dashboard.count` and nothing else — since migration 103 already rewrote every stored
dashboard to v2.

N/A — telemetry only, no user-facing surface.

#### Issues closed by this PR

Closes https://github.com/SigNoz/pulse-pod/issues/191

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause

`addStatsFromStorableDashboard` walked a top-level `widgets` array in the stored
dashboard JSON and returned early when the key was missing:

```go
if dashboard.Data["widgets"] == nil {
    return
}
```

A v2 dashboard's stored `data` blob has exactly two top-level keys, `metadata` and
`spec`; panels live as a map under `spec.panels`. There is no `widgets` key, so every v2
row hit that early return and `dashboard.panels.count`, `dashboard.panels.traces.count`,
`dashboard.panels.metrics.count` and `dashboard.panels.logs.count` all stayed at their
initialized zeros. `dashboard.count` kept working because it is `len(dashboards)` — a row
count that never touches the JSON.

The per-signal breakdown was broken twice over: the counter read
`query.builder.queryData[].dataSource`, while a v2 query carries its signal at
`spec.panels.<key>.spec.queries[].spec.plugin.spec.signal`.

`pkg/sqlmigration/103_migrate_dashboards_v1_to_v2.go` rewrites every existing dashboard
to v2 in place and first shipped in v0.135.0 — that is when the counters went to zero for
everyone at once. The v2 create API landed earlier (v0.126.1), so dashboards created
through it were already uncounted before the migration.

#### Fix Strategy

`addStatsFromStorableDashboard` now decodes the row through the existing typed v2 read
path (`StorableDashboard.ToDashboardV2`), counts each entry in `spec.panels`, and derives
the signal per query by normalizing the query plugin into a v5 composite query and
reading `QueryEnvelope.GetSignal()` — so all six query plugin kinds are handled through
one path instead of a hand-rolled map walk.

Counting semantics:

- A panel carries exactly one query envelope, so multi-signal panels arrive as a
  composite; each builder sub-query contributes its own signal.
- Signal-less queries (promql, clickhouse sql, formulas) count towards the panel total
  only — matching the old behaviour, which only counted `queryType == "builder"`.
- Rows that do not decode as v2 are skipped for panel stats and still count towards
  `dashboard.count`.

The magic stat keys are now named constants, and the repeated
`stats[k] = stats[k].(int64) + 1` assertions are behind one `incrementStat` helper.

---

### 🧪 Testing Strategy

- Tests added/updated: new `pkg/types/dashboardtypes/dashboard_stats_test.go` — panel and
  per-signal counting for traces/metrics/logs, composite sub-query counting, signal-less
  queries counting towards the panel total only, aggregation across dashboards, v1 and
  empty rows contributing to `dashboard.count` alone, and the empty input case. Fixtures
  go through the untyped `data` blob so they exercise the same decode a DB row does.
- Manual verification: `go test ./pkg/types/dashboardtypes/ ./pkg/modules/dashboard/...`,
  `go build ./pkg/... ./ee/...`, `go vet`, `gofmt`.
- Edge cases covered: missing/nil `data`, non-v2 schema versions, nil panel entries,
  queries whose plugin cannot be normalized (skipped, panel still counted).

---

- Blast radius: analytics only. Four call sites share the function — `Dashboard Created`
  on the v1 and v2 create paths, and the OSS and EE periodic stats collectors. No API,
  schema, storage or UI change.
- Potential regressions: v1 dashboards no longer report panel stats. That is intended
  here, and after migration 103 there should be no v1 rows left to report on. Decoding
  now goes through full v2 validation, so a stored row that fails validation reports no
  panel stats rather than partial ones — it still counts as a dashboard.
- Rollback plan: revert the commit; the counters return to reporting zero.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | N/A — internal telemetry, not user-facing |

---

### 📋 Checklist
- [x] Manually tested

---

## 👀 Notes for Reviewers

Two decisions worth a look:

1. **v2-only, by request.** v1 rows are skipped rather than dual-parsed. If any v1 rows
   are expected to survive migration 103, their panel stats go dark.
2. **Typed decode vs. raw map walk.** Using `ToDashboardV2` buys reuse and type safety at
   the cost of a marshal/unmarshal plus full spec validation per dashboard on the periodic
   collector path. Happy to switch to a lightweight map walk if that cost matters at
   collector scale.


pkg/types/dashboardtypes/dashboard.go
```diff
@@ -176,69 +177,75 @@ func NewGettableDashboardFromDashboard(dashboard *Dashboard) (*GettableDashboard
...
+// panelSignalStatKeys maps a builder query's signal to the stat it contributes
+// to. Signal-less queries (promql, clickhouse sql, formulas) count towards the
+// panel total only.
+var panelSignalStatKeys = map[telemetrytypes.Signal]string{
+	telemetrytypes.SignalTraces:  statKeyPanelTracesCount,
+	telemetrytypes.SignalMetrics: statKeyPanelMetricsCount,
+	telemetrytypes.SignalLogs:    statKeyPanelLogsCount,
+}
+
 func NewStatsFromStorableDashboards(dashboards []*StorableDashboard) map[string]any {
```
namanv3: as such this is correct but can you move this method (and everything it needs) to a new file called perses_dashbaord_stats. We should keep all v2 related things in `perses_` prefixed files until we remove all the v1 code.

pkg/types/dashboardtypes/dashboard.go
```diff
@@ -176,69 +177,75 @@ func NewGettableDashboardFromDashboard(dashboard *Dashboard) (*GettableDashboard
...
 		return
 	}
 
-	if dashboard.Data["widgets"] == nil {
-		return
-	}
-
-	widgets, ok := dashboard.Data["widgets"]
-	if !ok {
+	dashboardV2, err := dashboard.ToDashboardV2(nil)
```
namanv3: storable -> dashboard v2 is happening, this is fine for Collect() module method, but for create v2 this is yet another type conversion. create v2 already has postable v2, and maybe we can create another method like NewStatsFromStorableDashboards but for postable v2 so that the create API does not deal with an extra type conversion.

namanv3 (approved):

srikanthccv (approved):

PR #12397, opened 2026-08-04
fix(version): detect docker under private cgroup namespaces

Nageshbansal:
### Summary

- Docker and compose installs show up as `deployment.mode=unknown` in stats. On cgroup v2 hosts, docker runs containers in a private cgroup namespace by default, so `/proc/self/cgroup` just reads `0::/` and the runtime check finds nothing to match.
- This change looks for the container manager's own markers first: the `container` env var, then `/run/.containerenv` (podman) and `/.dockerenv` (docker). If those are missing and the cgroup path is namespaced away, it checks mount points as a last resort. This is the same order systemd uses in `detect_container()`.
- Tested in real containers: plain `docker run`, docker compose, and the shipped `signoz/signoz:latest` image now all report `docker`.

Related to SigNoz/platform-pod#2794

### DIFF
--- a/pkg/version/deployment.go
+++ b/pkg/version/deployment.go
@@ -62,12 +62,9 @@ func detectMode() string {
 		return "kubernetes"
 	}
 
-	// Check the container environment variable, set by some container managers (podman does, docker does not)
-	switch os.Getenv("container") {
-	case "podman":
+	// Check if running in podman, which sets the container environment variable
+	if os.Getenv("container") == "podman" {
 		return "podman"
-	case "docker":
-		return "docker"
 	}
 
 	// Check for container manager marker files, Podman must come before Docker — mirrors systemd's detect_container to avoid mistaking another container manager for docker
--- a/pkg/version/deployment.go
+++ b/pkg/version/deployment.go
@@ -62,9 +62,26 @@ func detectMode() string {
 		return "kubernetes"
 	}
 
+	// Check the container environment variable, set by some container managers (podman does, docker does not)
+	switch os.Getenv("container") {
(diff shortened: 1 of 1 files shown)


pkg/version/deployment.go
```diff
@@ -62,9 +62,26 @@ func detectMode() string {
...
+		return "docker"
+	}
+
+	// Check for container manager marker files, Podman must come before Docker — mirrors systemd's detect_container to avoid mistaking another container manager for docker
+	if _, err := os.Stat("/run/.containerenv"); err == nil {
+		return "podman"
+	}
+	if _, err := os.Stat("/.dockerenv"); err == nil {
+		return "docker"
+	}
```
therealpandey: Why is the first and not the last?

And if docker does not why are we switching on container for docker?

therealpandey (changes requested):

pkg/version/deployment.go
Nageshbansal: - Followed it the same way systemd detects, for example. [virt.c#L704-L720](https://github.com/systemd/systemd/blob/2dd07068a388c31610fca8c8e181dc6b501be9e4/src/basic/virt.c#L704-L720),  [virt.c#L741-L746](https://github.com/systemd/systemd/blob/2dd07068a388c31610fca8c8e181dc6b501be9e4/src/basic/virt.c#L741-L746). 
so to read the container environment variable before anything else because the manager writes it, and it would be a direct answer.

- Dont need the docker env detection, removing it

therealpandey (approved):

PR #12413, opened 2026-08-05
fix(infra-monitoring): use local sample tables in custom queries in `metric-reduction-feature` path

nikhilmantri0902:
### 📄 Summary

Infra-monitoring v2 APIs fail with ClickHouse error 288 (`Double-distributed IN/JOIN subqueries is denied`) for tenants with metrics reduction enabled. The reduced path's `fingerprint IN (...)` subquery read the distributed reduced-samples tables inside a query that is itself on a distributed table. This PR switches the subquery to the local tables, mirroring what the raw path already does with `samples_v4`.

Closes https://github.com/SigNoz/pulse-pod/issues/205

### ✅ Change Type

- [x] 🐛 Bug fix

### 🐛 Bug Context

**Root Cause**: `buildReducedSamplesTblFingerprintSubQuery` built its subquery on `distributed_samples_v4_reduced_{last,sum}_60s` while the outer query reads `distributed_time_series_v4_reduced` — rejected under ClickHouse's default `distributed_product_mode = 'deny'`. All three callers (`getMetadata`, `getPerGroupDistinctCounts`, `getPerGroupHostStatusCounts`) share this helper, so every entity page broke for reduction-enabled tenants.

**Fix Strategy**: The subquery now reads the local tables, so each shard resolves fingerprints against its own data. This is exact because the reduced time-series and reduced-samples tables share the same sharding key (`cityHash64(env, temporality, metric_name, <reduced_>fingerprint)`), keeping a series row and its samples on the same shard — verified on the affected tenant's cluster. Local tables were preferred over `GLOBAL IN` to avoid broadcasting the subquery result and to stay consistent with the raw path and the QB v5 statement builder.

### 🧪 Testing Strategy

- Added `tests/integration/tests/metricreduction/05_inframonitoring.py`: hosts, nodes, pods, clusters, and deployments list endpoints are tested on a 2-shard cluster with reduction enabled, seeding raw-only and reduced-only entities and asserting the response returns the union. These tests fail with error 288 without the fix. A hosts groupBy test covers the per-group status counts path. Full metricreduction suite passes (18/18).
- Manually reproduced against the affected tenant's ClickHouse: infra pages failed with error 288 before the fix, and the corrected query returns results.
- Non-reduced tenants are unaffected; the helper only runs on the reduction-enabled path.

- Blast radius is limited to infra-monitoring v2 queries for tenants with `enable_metrics_reduction` on.
- Rollback: revert the commit, or disable the flag per tenant to bypass the reduced path.

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | Cloud |
| Change Type | Bug Fix |
| Description | Fixed infra-monitoring pages failing with a ClickHouse error for tenants with metrics reduction enabled. |

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/pkg/modules/inframonitoring/implinframonitoring/helpers.go
+++ b/pkg/modules/inframonitoring/implinframonitoring/helpers.go
@@ -372,9 +372,7 @@ func (m *module) buildSamplesTblFingerprintSubQuery(metricNames []string, sample
 }
 
 // buildReducedSamplesTblFingerprintSubQuery is like buildSamplesTblFingerprintSubQuery
-// but for the reduced tables. Uses local tables: a distributed subquery inside the
-// distributed outer query is rejected by ClickHouse (distributed_product_mode = 'deny'),
-// and reduced series/samples share the same sharding key so per-shard lookup is exact.
+// but for the reduced tables.
 func (m *module) buildReducedSamplesTblFingerprintSubQuery(metricNames []string, flooredStart, flooredEnd uint64) *sqlbuilder.SelectBuilder {
 	lastSB := sqlbuilder.NewSelectBuilder()
 	lastSB.Select("reduced_fingerprint")
--- a/pkg/modules/inframonitoring/implinframonitoring/helpers.go
+++ b/pkg/modules/inframonitoring/implinframonitoring/helpers.go
@@ -372,11 +372,13 @@ func (m *module) buildSamplesTblFingerprintSubQuery(metricNames []string, sample
 }
 
 // buildReducedSamplesTblFingerprintSubQuery is like buildSamplesTblFingerprintSubQuery
-// but for the reduced tables.
+// but for the reduced tables. Uses local tables: a distributed subquery inside the
+// distributed outer query is rejected by ClickHouse (distributed_product_mode = 'deny'),
+// and reduced series/samples share the same sharding key so per-shard lookup is exact.
 func (m *module) buildReducedSamplesTblFingerprintSubQuery(metricNames []string, flooredStart, flooredEnd uint64) *sqlbuilder.SelectBuilder {
 	lastSB := sqlbuilder.NewSelectBuilder()
 	lastSB.Select("reduced_fingerprint")
-	lastSB.From(fmt.Sprintf("%s.%s", metricstelemetryschema.DBName, metricstelemetryschema.SamplesV4ReducedLastTableName))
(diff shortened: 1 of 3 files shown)


pkg/modules/inframonitoring/implinframonitoring/helpers.go
```diff
@@ -372,11 +372,13 @@ func (m *module) buildSamplesTblFingerprintSubQuery(metricNames []string, sample
 }
 
 // buildReducedSamplesTblFingerprintSubQuery is like buildSamplesTblFingerprintSubQuery
-// but for the reduced tables.
+// but for the reduced tables. Uses local tables: a distributed subquery inside the
+// distributed outer query is rejected by ClickHouse (distributed_product_mode = 'deny'),
+// and reduced series/samples share the same sharding key so per-shard lookup is exact.
```
namanv3: NIT: 
> Uses local tables: a distributed subquery inside the
// distributed outer query is rejected by ClickHouse (distributed_product_mode = 'deny'),
// and reduced series/samples share the same sharding key so per-shard lookup is exact.

No need for this cuz this comment only makes sense if someone knows that an issue was seen

tests/integration/tests/metricreduction/05_inframonitoring.py
```diff
@@ -0,0 +1,258 @@
...
+    "k8s.node.name": "node-mr",
+    "k8s.cluster.name": "cluster-mr",
+    "k8s.statefulset.name": "",
+    "k8s.daemonset.name": "",
+    "k8s.job.name": "",
+    "k8s.cronjob.name": "",
+    "k8s.pod.start_time": "2025-01-01T00:00:00Z",
+}
+
+ENTITY_CASES = [
```
namanv3: NIT: no need to define this extra variable right? Can just directly add it in the annotation?

pkg/modules/inframonitoring/implinframonitoring/helpers.go
nikhilmantri0902: done

tests/integration/tests/metricreduction/05_inframonitoring.py
nikhilmantri0902: done

namanv3 (approved):

therealpandey (approved):

PR #12416, opened 2026-08-05
fix(dashboard-v2): stop a time-range change resetting a dynamic variable's selection to ALL

AshwinBhatkal:
## Summary

On a V2 dashboard, picking values in a multi-select variable and then changing the time range could silently switch the variable to **ALL** — widening every panel to all values without the user touching the variable. A value *typed into* a variable was dropped on any refetch for the same underlying reason.

The post-fetch reconcile compares a selection against freshly-fetched options and could not tell *why* those options changed: "the user has nothing selected yet" and "the user's selection was just invalidated by a refetch" arrived as the same input, and both resolved to the variable's default — ALL for an ALL-enabled multi-select. A time-range change refetches every variable, so any window without data for the selected value hit that path.

Two guarantees now, each with its own mechanism:

| Guarantee | Mechanism |
| --- | --- |
| A refetch nothing else caused (time range, reload) never re-defaults a selection | The fetch engine tags each cycle with why it was enqueued; only a value cascade may re-default |
| A typed-in value survives every refetch, whatever caused it | The selection records which entries were typed, judged at pick time against the options then offered |

A parent variable's value changing still re-scopes its children — that behaviour is unchanged and intended. Single-select variables have preserved a non-empty value since #12178; this brings multi-select in line, which is the half that was left untouched then.

Also fixed, same family: opening and closing the dropdown without touching it promoted an explicit pick into a standing ALL whenever the current window offered only the selected values, and rewrote a dynamic ALL's `__all__` into concrete values.

Closes https://github.com/SigNoz/pulse-pod/issues/207

## Commits

1. `refactor` — record why each variable fetch cycle was enqueued (full cycle vs value cascade)
2. `fix` — keep a variable's selection across a time-range refetch
3. `fix` — keep typed-in variable values through every refetch; ALL now means exactly the option set
4. `fix` — a no-op close of the variable list commits nothing; commit rule extracted out of the component

Each commit typechecks on its own.

## Test plan

- [x] `jest src/pages/DashboardPageV2` — 136 suites / 1073 tests pass, including 20 added: the reconcile split by cycle reason, the cycle tagging in the store, a time-range change tagging every variable as a full cycle, the typed-value rules, and the commit resolver
- [x] `tsgo --noEmit` clean, at every commit
- [x] `oxlint` and `oxfmt --check` clean on the changed files
- [x] Manual: multi-select variable, pick one value, switch to a window with no data for it → selection holds, panels show no data rather than everything
- [x] Manual: type a custom value into a variable, change the time range and switch a sibling variable → the typed value stays selected
- [x] Manual: namespace → pod pair, change namespace → pod values still re-scope

## Notes for reviewers

- `customValues` is new on the runtime selection and is persisted with it. It never reaches the wire or a shared link: `buildVariablesPayload` and the share-URL builder both project `value` / the `__all__` sentinel explicitly.
- A selection seeded from a `?variables=` share link carries no typed-value marker — that URL format stores `name → value` only, so a typed value from a link is indistinguishable from a fetched one and a cascade can still drop it.
- The pill can still *read* ALL when the current option list happens to be a subset of the selection: `CustomMultiSelect` infers that from `options ⊆ value`, and V1 depends on the inference. It is display-only now and self-corrects as the window widens; making it exact needs an explicit prop on the shared control.

### DIFF
--- a/frontend/src/pages/DashboardPageV2/DashboardContainer/VariablesBar/components/selectors/ValueSelector.tsx
+++ b/frontend/src/pages/DashboardPageV2/DashboardContainer/VariablesBar/components/selectors/ValueSelector.tsx
@@ -6,6 +6,7 @@ import { DashboardDetailEvents } from 'pages/DashboardPageV2/constants/events';
 
 import type { VariableSelection } from '../../selectionTypes';
 import { areSelectionsEqual } from '../../utils/resolveVariableSelection';
+import { selectionFromCommittedValues } from '../../utils/selectionUtils';
 import OverflowValuesTooltip from './OverflowValuesTooltip';
 import styles from '../../VariablesBar.module.scss';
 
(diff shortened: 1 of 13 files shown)


frontend/src/pages/DashboardPageV2/DashboardContainer/VariablesBar/components/selectors/ValueSelector.tsx
```diff
@@ -75,13 +76,23 @@ function ValueSelector({
 		options.every((option) => draft.includes(option));
 
 	const commit = (values: string[]): void => {
-		// CustomMultiSelect emits the full value set when ALL is picked.
-		const isAll =
```
AshwinBhatkal, line 79: just moved to functions

ahrefabhi (approved):

PR #12420, opened 2026-08-05
test(e2e): rewrite the variables-bar spec for V2

AshwinBhatkal:
## Summary

Stacked on #12419 (`test/e2e-dashboards-foundation`) — review that one first; this PR's diff is the variables spec alone.

`67-variables.spec.ts` was written against the V1 variables bar and has been parked since the V2 migration. This rewrites it against the V2 runtime bar, and deletes its `parked-specs.json` entry — so from here the guard added in #12419 forbids a skipped test in this file.

**10 tests, no skips.** The four `test.skip` placeholders the V1 file carried are gone: three are now covered for real (custom variable without a default, pre-seeded selection rendering, panel refetch on change is implicit in the commit path), and the fourth — drag-reordering variables — belongs to the settings spec, not the bar.

| Test | Behaviour |
| --- | --- |
| TC-01 | Every seeded variable renders as a pill labelled `$name` |
| TC-02 | An ALL-enabled multi-select with no default resolves to ALL |
| TC-03 | A single-select renders its configured default |
| TC-04 | "Only" on a row collapses an ALL selection to that one value |
| TC-05 | Checking a second value adds it to the selection |
| TC-06 | A multi-select edit is committed by closing the list, not per toggle |
| TC-07 | A selection survives a reload |
| TC-08 | A text variable keeps a typed value across a reload |
| TC-09 | Switching a single-select replaces its value |
| TC-10 | Variables that do not fit collapse into a "+N" overflow, with a tooltip listing them |

Every assertion is driven from text + custom variables, whose option lists come from the definition — so nothing here can go red because of what telemetry the stack happens to hold.

## What this caught while being written

TC-05 initially asserted only that the *first* picked value was shown, and passed — while the second value was never committed at all, because clicking an option's row does not toggle its checkbox. Reading the closed control cannot tell the two cases apart (it shows one tag plus a "+N"), so the assertion now reads the open list. The helper fix and that lesson are in #12419.

## Deliberately not here

Fetched variables (QUERY / DYNAMIC). Their option lists come from telemetry, and the behaviours only they can show — a selection surviving a time-range refetch, a typed value surviving a dependency cascade — need seeded traces **and** the fix in #12416. They follow in their own spec, based on that branch, rather than sitting here as skips.

## Test plan

- [x] `67-variables.spec.ts` — **10/10 pass** (chromium, 2 workers, local bootstrap stack)
- [x] Run together with `01-smoke.spec.ts` — 14/14, no cross-file interference
- [x] Repeat runs green (no flakes observed across 4 runs)
- [x] `pnpm guard:specs` — ok, 10/21 running, 11 parked (this file no longer parked)
- [x] `pnpm typecheck` 0 errors · `pnpm lint` 0 errors · `pnpm fmt:check` clean

### DIFF
--- a/tests/e2e/parked-specs.json
+++ b/tests/e2e/parked-specs.json
@@ -8,7 +8,6 @@
 		"**/tests/dashboards/details/35-add-panel.spec.ts",
 		"**/tests/dashboards/details/44-edit-panel.spec.ts",
 		"**/tests/dashboards/details/56-time-range.spec.ts",
-		"**/tests/dashboards/details/67-variables.spec.ts",
 		"**/tests/dashboards/details/78-edit-mode.spec.ts",
 		"**/tests/dashboards/details/87-configure.spec.ts",
 		"**/tests/dashboards/details/95-edge-cases.spec.ts",
--- a/tests/e2e/tests/dashboards/details/67-variables.spec.ts
+++ b/tests/e2e/tests/dashboards/details/67-variables.spec.ts
@@ -2,42 +2,59 @@ import type { Page } from '@playwright/test';
 
 import { expect, test } from '../../../fixtures/auth';
 import { newAdminContext } from '../../../helpers/auth';
+import { authToken } from '../../../helpers/dashboards';
 import {
-	authToken,
-	awaitVariablesResolved,
-	createVariablesDashboardViaApi,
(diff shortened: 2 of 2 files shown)


AshwinBhatkal:

PR #12421, opened 2026-08-05
test(e2e): pin the v1 dashboard write API as deprecated

AshwinBhatkal:
## Summary

Stacked on #12419.

`POST /api/v1/dashboards` now answers **501 `dashboard_deprecated`**, pointing callers at `/api/v2/dashboards`. Two tests hold that contract, and record why every dashboards spec seeds through v2.

This also settles something the E2E plan had open. A V1-schema dashboard can no longer be created through the API, so the v1 → v2 migration (`pkg/transition/migrate_dashboard.go`) **cannot be exercised end to end from a browser test** — it only ever runs on rows that predate the deprecation. That coverage stays where it already is, in `pkg/types/dashboardtypes/perses_v1_to_v2_test.go`; a spec claiming it here would assert nothing.

Knock-on for the rest of the series: the parked specs seed through the v1 API, so they cannot be *revived* as written — each area is a rewrite against v2 seeding. That is what #12420 does for the variables bar.

## Test plan

- [x] 2/2 pass (chromium, 2 workers, local bootstrap stack)
- [x] `pnpm guard:specs` / `typecheck` / `lint` / `fmt:check` clean

### DIFF
--- a/tests/e2e/tests/dashboards/details/05-v1-api-deprecated.spec.ts
+++ b/tests/e2e/tests/dashboards/details/05-v1-api-deprecated.spec.ts
@@ -1,0 +1,46 @@
+import { expect, test } from '../../../fixtures/auth';
+import { authToken } from '../../../helpers/dashboards';
+import { createDashboardV2ViaApi } from '../../../helpers/dashboards-v2';
+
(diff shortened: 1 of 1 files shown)


AshwinBhatkal:

PR #12422, opened 2026-08-05
test(e2e): rewrite the variable-settings spec for V2

AshwinBhatkal:
## Summary

Stacked on #12419.

`87-configure.spec.ts` was the worst of the parked specs: **30 tests, 15 of them `test.skip` placeholders**, all written against the V1 settings UI. Replaced with eight tests against the V2 form and list, and its `parked-specs.json` entry is deleted — so the guard from #12419 now forbids a skipped test in this file.

| Test | Behaviour |
| --- | --- |
| TC-01 | The list shows every variable the dashboard defines |
| TC-02 | A new custom variable is persisted **and** reaches the runtime bar |
| TC-03 | A duplicate name cannot create a second variable of that name |
| TC-04 | An empty name cannot be saved |
| TC-05 | Editing a custom variable changes the options it offers |
| TC-06 | Deleting takes it off the list and out of the bar, and stays gone across a reload |
| TC-07 | A variable saved as single-select does not render as ALL |
| TC-08 | Discarding the form leaves the dashboard byte-identical |

Each test seeds its own dashboard, so a create or delete cannot leak into another and the file runs in parallel.

## What running them taught

- The variables list is reached through the form's **"All variables"** back-link. My first attempt went through the settings drawer and failed — this path also doesn't depend on the drawer's tab layout.
- A delete removes the row **optimistically**, so a reload can outrun the write and legitimately still show the variable. TC-06 waits for the persisted spec first. Worth noting the delete *is* persisted — I confirmed that against the API rather than assuming, since the alternative reading was a real bug.

## Test plan

- [x] **8/8 pass** (chromium, 2 workers, local bootstrap stack)
- [x] `pnpm guard:specs` — ok, 13/23 running, 10 parked
- [x] `pnpm typecheck` 0 errors · `pnpm lint` 0 errors · `pnpm fmt:check` clean

### DIFF
--- a/tests/e2e/parked-specs.json
+++ b/tests/e2e/parked-specs.json
@@ -10,7 +10,6 @@
 		"**/tests/dashboards/details/56-time-range.spec.ts",
 		"**/tests/dashboards/details/67-variables.spec.ts",
 		"**/tests/dashboards/details/78-edit-mode.spec.ts",
-		"**/tests/dashboards/details/87-configure.spec.ts",
 		"**/tests/dashboards/details/95-edge-cases.spec.ts",
 		"**/tests/trace-details/preview-fields.spec.ts"
 	],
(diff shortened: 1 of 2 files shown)


AshwinBhatkal:

PR #12423, opened 2026-08-05
test(e2e): rewrite the dashboards-list spec for V2

AshwinBhatkal:
## Summary

Stacked on #12419.

The V1 list spec seeded through `POST /api/v1/dashboards` — which now returns 501 (see #12421) — and asserted against a list that has since become views-first. Ten tests against the V2 list, and its `parked-specs.json` entry is deleted.

| Test | Behaviour |
| --- | --- |
| TC-01 | Chrome and core controls render (Views rail, CTA, search, sort) |
| TC-02 | Every view in the rail is reachable and keeps the list frame |
| TC-03 | A newly created dashboard is listed |
| TC-04 | Opening one from the list lands on its detail page |
| TC-05 | Search narrows the list to a matching dashboard |
| TC-06 | A search matching nothing leaves none of our rows |
| TC-07 | Pinning puts a dashboard in the Pinned view |
| TC-08 | The create CTA opens the new-dashboard modal |
| TC-09 | Creating through the modal lands on the new dashboard |
| TC-10 | A deleted dashboard leaves the list |

Titles carry a per-run, per-worker suffix and every assertion is scoped to this suite's own dashboards — the workspace is shared, so counting rows or asserting on "the first row" would depend on what else exists.

## One view deliberately not covered

**Recently viewed** is client-side application state, so a fresh browser context — which is what every test gets — has nothing to show, and a full page load drops it too. There is no server state to seed and nothing meaningful for an E2E assertion to hold onto, so it is left out here; a unit test over the store slice is the right home for it.

## Test plan

- [x] **10/10 pass** (chromium, 2 workers, local bootstrap stack)
- [x] `pnpm guard:specs` / `typecheck` / `lint` / `fmt:check` clean

### DIFF
--- a/tests/e2e/parked-specs.json
+++ b/tests/e2e/parked-specs.json
@@ -1,7 +1,6 @@
 {
 	"$comment": "Specs not currently running, and why. This is the ONLY place a spec may be excluded from the suite: `playwright.config.ts` feeds `specs` to `testIgnore`, and `pnpm guard:specs` fails if any spec that is NOT listed here contains a skipped, fixme'd or .only test. So a spec is either running and complete, or parked here with a reason — nothing rots quietly in between. Every entry is removed by the PR that migrates it; the list only shrinks.",
 	"specs": [
-		"**/tests/dashboards/list.spec.ts",
 		"**/tests/dashboards/details/03-viewing.spec.ts",
 		"**/tests/dashboards/details/12-sections.spec.ts",
 		"**/tests/dashboards/details/21-panel-actions.spec.ts",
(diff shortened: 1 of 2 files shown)


AshwinBhatkal:

AshwinBhatkal:
Correction to an earlier note in this description: I had flagged Recently viewed as looking like a defect. It isn't — it's client-side application state, so a fresh browser context has nothing to show and a full page load drops it. Nothing to chase; a unit test over the store slice is the right home. Description updated.

PR #12426, opened 2026-08-05
feat(logs): render DataViewer with filter/group-by actions in log details overview

aks07:
## Pull Request

---

Renders the new `DataViewer` (Pretty tree + JSON) inside the V2 log-details Overview tab
Wires up filter / group-by on each attribute similar to pretty view in trace details

**Change points**

**Rendering the View**
- Data Viewer renders using `aggregateAttributesResourcesToObject` which is not written fresh, just extracted from an existing logic `aggregateAttributesResourcesToString`.
- jsonData is separately sent to the DataViewer as rendering logic for pretty and json view is different in this case. unlike trace details where both views had same source data.

**Group by / FIlter and other logic**

- All the related logic resides in 2 major files: `useLogAttributeActions.tsx` and `logAttributeActions.utils.ts`.
- we build the fieldKey ourselves (`buildLogFilterTarget`) as this is now different from old representation.
- filter / group by / replace build the query locally now...we do not make the `getAggregateKeys` call at all. we fabricate the telemetry field key ourselves with just the name and dataType filled and rest kept empty. so no prefetch, no resolver, no loader.
- the query building is extracted into 3 utils: `getFilterQueryData`, `getGroupByQueryData`, `getReplaceFilterQueryData`. the hook just calls these over `updateQueriesData`.
- restricted fields apply for body as well: `timestamp` / `id` / `date` inside body no longer show filter / group by. reuses `RESTRICTED_SELECTED_FIELDS`.

**Pretty View**
- renderLeafValue: introduce to render custom leaf value. we are using this here for body. This is extensible for other usecases as well...like showing md format leaf for LLMs in the future.
- fixed leaf vs nested row indentation so keys line up at every depth.

**Other changes**
- Filter value keeps its data type: dataType is threaded through so a numeric/bool filter value stays unquoted.
- Removed the redundant outer JSON tab in V2.

https://github.com/user-attachments/assets/ec1f0842-0a3b-407b-807c-0c3b0f9ed86a

Closes: https://github.com/SigNoz/engineering-pod/issues/4618
Closes: https://github.com/SigNoz/engineering-pod/issues/4630
Closes: https://github.com/SigNoz/engineering-pod/issues/5781
---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---


frontend/src/container/LogDetailedView/utils.tsx
```diff
@@ -291,12 +293,39 @@ export const aggregateAttributesResourcesToString = (logData: ILog): string => {
...
 			outputJson[key] = logData[key as keyof ILog];
 		}
 	});
 
-	return JSON.stringify(outputJson, null, 2);
+	return outputJson;
+};
+
+export const aggregateAttributesResourcesToString = (logData: ILog): string =>
+	JSON.stringify(aggregateAttributesResourcesToObject(logData), null, 2);
```
YounixM: Use utils to handle json operations safely

frontend/src/periscope/components/DataViewer/DataViewer.tsx
```diff
@@ -55,7 +62,14 @@ function DataViewer({
 	};
 
 	return (
-		<div className="data-viewer">
+		<div
+			className="data-viewer"
+			style={
```
YounixM, line 67: Inline styles ?

frontend/src/container/LogDetailedView/logAttributeActions.utils.ts
```diff
@@ -0,0 +1,228 @@
...
+			out += arrayMarker;
+		} else {
+			out += out ? `.${seg}` : seg;
+		}
+	});
+	return out;
+};
+
+const metricsTypeForRoot = (root: string | number): MetricsType | undefined => {
+	if (root === 'attributes') {
```
YounixM: use types

frontend/src/container/LogDetailedView/hooks/useLogAttributeActions.tsx
```diff
@@ -0,0 +1,224 @@
...
+		replaceFilter,
+		isBodyJsonQueryEnabled,
+		isOldExplorerOrLive,
+	]);
+
+	const visibleActions = useMemo<VisibleActionsConfig>(
+		() => ({
+			leaf: isListViewPanel
+				? ['copy']
+				: ['copy', 'filter-in', 'filter-out', 'group-by', 'replace-filter'],
```
YounixM: NIT - can move to constant

YounixM (dismissed):

frontend/src/container/LogDetailedView/utils.tsx
aks07: This is existing logic, just extracted out...not new here. Also the arg is typed, so it is always a valid plain object that follows the type. and if stringify does fail, we would want that to surface rather than swallow it with a try catch.

frontend/src/periscope/components/DataViewer/DataViewer.tsx
aks07: fontSize is a prop which lets the consumer set the font size. its not inline styling...we just set the -`-data-viewer-font-size` var to the fontSize passed and the scss file reads that var.

frontend/src/container/LogDetailedView/logAttributeActions.utils.ts
aks07: done

frontend/src/container/LogDetailedView/hooks/useLogAttributeActions.tsx
aks07: done

frontend/src/container/LogDetailedView/utils.tsx
YounixM: https://docs.sentry.io/platforms/javascript/usage/

aks07: Done

frontend/src/periscope/components/DataViewer/DataViewer.tsx
aks07: fontSize prop is needed since monaco editor requires the fontSize to be passed as prop. We cannot completely handle this with classes

ankitnayan:
Do we have some default for showing what level of JSON depth should be uncollapsed? Having everything uncollapsed in a big nested json can be a pain again to browse the attributes

aks07:
> Do we have some default for showing what level of JSON depth should be uncollapsed? Having everything uncollapsed in a big nested json can be a pain again to browse the attributes

1. Wide collections (>50 children) - chunked into collapsible ranges.
2. Deep nesting: anything beyond level 5 is collapsed.


https://github.com/user-attachments/assets/784bd7e1-2a49-4872-8950-20c1ccf16d7e

YounixM (approved):

PR #12435, opened 2026-08-06
feat(alert-channel-integrations): integration tests using wiremock for google chat, incident.io, jira and jsm ops

nikhilmantri0902:
#### Description

- Adds WireMock-backed integration tests for four alert channels: Google Chat, incident.io, Jira, and JSM Ops.
- Each channel gets two suites: the synchronous channel-test API (`tests/alerts/`) and the real firing-rule pipeline (`tests/alertmanager/`).
- The shared WireMock container serves real TLS (CA from #12486) under network aliases for `chat.googleapis.com`, `api.incident.io`, `signoz-test.atlassian.net`, and `api.atlassian.com`. A `tls-hosts` container label recreates reused containers when the host list changes.
- Google Chat covers the cardsV2 payload with deep links, verbatim webhook url, permanent 400/403, and a 429→200 retry.
- incident.io covers the alert event shape, a single `Bearer` prefix on the token, permanent 401/422, and a 429→202 retry.
- Jira covers issue create (JQL + basic auth + ADF panel), the wont-fix JQL clause, update of an existing open issue, reopen of a done issue, permanent 400/401, service-account routing via the `api.atlassian.com` gateway, and a search 429 retry.
- JSM Ops covers alert create plus a timeline note (`GenieKey` auth, HTML body), the drop-failed-note tolerance, permanent 422/401, and a create 429 retry.
- The shared webhook-verify helper gains `urlPath`/`count`/`min_count` matching and a `path_pattern` regex variant for paths embedding the group-hash alias.
- Fixes a harness bug: ClickHouse mounted its cluster config from pytest basetemp, which is wiped every session, so reused containers hot-reloaded away their cluster definition. The configs now live in the cross-session cache.

#### Issues closed by this PR

Closes https://github.com/SigNoz/pulse-pod/issues/165

#### Additional Information

- Channel-test readiness is polled (`channels/test` returns 404 until the org's alertmanager registers) instead of a fixed sleep.
- Resolved-path delivery (`send_resolved` close/transition flows) is deliberately out of scope: it needs a real alert to stop firing, which is timing-heavy; the status mapping is unit-tested.


tests/fixtures/alerts.py
```diff
@@ -455,4 +497,11 @@ def update_raw_channel_config(
                     path = urlparse(original_url).path
                     entry[url_field] = notification_channel.container_configs["8080"].get(path)
 
+    # Google Chat validates the webhook host, so route via the https alias config
+    # (chat.googleapis.com:8443) keeping the path, and skip tls for wiremock's cert.
+    for entry in config.get("googlechat_configs", []):
+        path = urlparse(entry["webhook_url"]).path
+        entry["webhook_url"] = notification_channel.container_configs["8443"].get(path)
```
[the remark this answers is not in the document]
nikhilmantri0902: Fixed in 7774544809. The `notification_channel` fixture now detects a stale `--reuse` cache (one whose `container_configs` lacks `"8443"`, i.e. created before the https alias was added), removes that container, and clears the cache so a correctly-aliased https:8443 container is recreated — instead of raising `KeyError` on `container_configs["8443"]`. Verified by seeding an 8080-only cache: the fixture logs `Recreating stale notification_channel (cache missing https:8443)` and all Google Chat suites pass.

tests/fixtures/notification_channel.py
```diff
@@ -135,9 +212,25 @@ def notification_channel(
...
+
         container = WireMockContainer(image="wiremock/wiremock:2.35.1-1", secure=False)
+        container.with_volume_mapping(str(keystore_path.parent), "/certs", "ro")
         container.with_network(network)
-        container.start()
+        container.with_network_aliases(GOOGLE_CHAT_HOST)
+        container.with_kwargs(labels={CA_ID_LABEL: ca_id(tls)})
+
+        try:
+            container.start(f"--port 8080 --https-port 443 --https-keystore /certs/keystore.p12 --keystore-type PKCS12 --keystore-password {KEYSTORE_PASSWORD}")
```
namanv3, line 527: PR description mentions some other port, I am guessing the description is outdated, but confirming once

tests/fixtures/notification_channel.py
```diff
@@ -135,9 +212,25 @@ def notification_channel(
...
+        # chat.googleapis.com with a CA-issued cert so Google Chat's validated
+        # webhook host routes here over real TLS (signoz trusts the integration CA).
+        keystore_path = issue_server_keystore(tls, tmpfs("notification-channel-certs"), GOOGLE_CHAT_HOST)
+
         container = WireMockContainer(image="wiremock/wiremock:2.35.1-1", secure=False)
+        container.with_volume_mapping(str(keystore_path.parent), "/certs", "ro")
         container.with_network(network)
-        container.start()
+        container.with_network_aliases(GOOGLE_CHAT_HOST)
+        container.with_kwargs(labels={CA_ID_LABEL: ca_id(tls)})
```
namanv3: description says `channels use insecure_skip_verify for its self-signed cert` but here a cert is being used right?

tests/integration/tests/alertmanager/04_googlechat.py
```diff
@@ -0,0 +1,182 @@
...
+    channel_name = str(uuid.uuid4())
+    path = gc_test_case.notification_expectation.notification_validations[0].validation_data["path"]
+
+    channel_config = update_raw_channel_config(gc_test_case.channel_config, channel_name, notification_channel)
+
+    stub_factory = CASE_STUBS.get(gc_test_case.name, googlechat_ok_mappings)
+    make_http_mocks(notification_channel, stub_factory(path))
+
+    create_notification_channel(channel_config)
+    time.sleep(12)  # org registration in alertmanager
```
namanv3: shouldn't we poll instead? I see other tests do that, test_googlechat_test_channel for eg.

tests/integration/tests/alertmanager/04_googlechat.py
```diff
@@ -0,0 +1,182 @@
...
+    insert_alert_data(gc_test_case.alert_data, base_time=datetime.now(tz=UTC) - timedelta(minutes=5))
+
+    with open(get_testdata_file_path(gc_test_case.rule_path), encoding="utf-8") as f:
+        rule_data = json.loads(f.read())
+    update_rule_channel_name(rule_data, channel_name)
+    create_alert_rule(rule_data)
+
+    verify_notification_expectation(notification_channel, maildev, gc_test_case.notification_expectation)
+
+    if gc_test_case.name == "googlechat_retry_429_then_200":
```
namanv3: this looks awkward, if we need to add if conditions on test case names might as well just add separate tests?

tests/integration/tests/alertmanager/04_googlechat.py
```diff
@@ -0,0 +1,182 @@
...
+
+GOOGLECHAT_CASES = [
+    types.AlertManagerNotificationTestCase(
+        name="googlechat_default_metrics_firing",
+        rule_path=METRICS_RULE,
+        alert_data=[types.AlertData(type="metrics", data_path=METRICS_DATA)],
+        channel_config=googlechat_config("gc-metrics"),
+        notification_expectation=types.AMNotificationExpectation(
+            should_notify=True,
+            wait_time_seconds=150,
```
namanv3: is 2.5 minutes too long? That too having 4 such tests so this test can take 10 minutes. PR checks can be a pain

tests/fixtures/notification_channel.py
nikhilmantri0902: Yes, updated the Pr description. Will keep every PR's description updated to latest state from now on.

tests/fixtures/notification_channel.py
nikhilmantri0902: Yes sir. There is a cert being used.

tests/integration/tests/alertmanager/04_googlechat.py
nikhilmantri0902: It's a poll ceiling, not a sleep - function returns as soon as the notification lands (~10–25s per test in practice; the last CI run finished all 4 in ~90s combined). The full wait is only paid when a test is already failing. That said, 150 was arbitrary, can u suggest a smaller cap ?

tests/integration/tests/alertmanager/04_googlechat.py
nikhilmantri0902: done moved to a seperate test.

tests/integration/tests/alertmanager/04_googlechat.py
nikhilmantri0902: yes changed to pollling

tests/integration/tests/alertmanager/04_googlechat.py
namanv3: yeah I get that its a ceiling, but we should fail much faster like 60s, cuz if something is down, I doubt waiting for 1.5 minutes extra will help with anything

nikhilmantri0902: done, fixed to 60s.

namanv3 (dismissed):

PR #12438, opened 2026-08-06
feat(dashboards): show disabled dashboard actions with a reason

AshwinBhatkal:
#### Description

Moves the dashboard detail page off the deprecated role table and onto per-action authz checks. The change reviewers will notice first: **a control you can't use is now visible and disabled with the reason, instead of missing.**

- **Every control goes through `AuthZButton` or `AuthZTooltip`** — toolbar, Actions menu, section and panel menus, the variables `+`, settings tabs, publish, the JSON drawer, the panel editor's Save. Denial wording comes from the components, so there is one tooltip implementation in the tree.
- **Dropdown rows are real buttons.** Each row renders an `AuthZButton`, the same as the dashboards list, so the authz component disables it like any other control. The dropdown *item* keeps its own `disabled`, `onClick` and `danger` — the menu needs the first two for keyboard navigation and the last for the destructive colour — and the row carries only its icon, label and reason.
- **The row's padding belongs to the button.** The menu hands its item padding to the rows, as the list page's popover already does, so a row's box reaches the panel edge. Left on the item, the button the tooltip anchors to was a box inset 16px inside the panel and the reason opened underneath it.
- **The edit context carries no permission copy.** It hands out the checks a control needs plus the non-permission obstacle for `disabledTooltip`, left empty when a permission is what's missing. That produces `update` → lock → `read`: someone without access hears about the access, not the lock, which would send them asking for an unlock.
- **The page gates `read` alone** — the only verb the dashboard request needs — and preloads `update`/`delete` in the same batch, so the controls below resolve from cache in one round trip instead of holding up the page.
- **Publishing is no longer admin-only.** It gates on `update`, so an editor can manage a public link (guide rule 2).
- **Fixes the settings drawer** rendering tabs from `Object.values(TabKeys)` instead of `items`, which ignored both `disabled` and the `enablePublicDashboard` flag — on OSS the Publish trigger rendered and landed on an empty body.
- **Fixes a clone gate**: `canReadDashboard` was aliased from `canEditDashboard`, which is read *and* update, so read + create without update was wrongly refused.

#### Screenshots
Screenshot 2026-09-10 at 11 24 23 PM
Screenshot 2026-09-10 at 11 24 18 PM

Screenshot 2026-09-10 at 11 24 00 PM
Screenshot 2026-09-10 at 11 25 11 PM
Screenshot 2026-09-10 at 11 24 50 PM

Screenshot 2026-09-10 at 11 24 30 PM

#### Additional Information

- Delete is unreachable without `read` here, because `withAuthZPage` replaces the whole page. Agreed as a dashboards exception — delete stays available from the list's row actions.
- **Configure stays disabled without `update`.** Opening the drawer read-only needs every field inside it to go read-only too — the Overview form, the tag inputs, and a view mode for the variables table — which is a larger change than this PR should carry. Planned as a follow-up once the three merge.
- Worth a visual check on the three dropdowns: their rows changed from library-rendered items to full-width buttons, which tests can't see.
- Requires #12448. Rebased onto current `main`; disjoint from #12449.


frontend/src/pages/DashboardPage/DashboardContainer/hooks/useDashboardEditContext.ts
H4ad, line 1: Review this and `useDashboardEditGuard`, maybe combine into single hook (and maybe drop `dashboardEditContext`), these three places feels we are overdoing something that could be simple

frontend/src/pages/DashboardPage/DashboardContainer/hooks/useDashboardReadDenied.ts
H4ad, line 1: Drop this, use `withAuthZPage` to guard the pages or `withAuthZContent.

You should not handle errors as failed to read permission, they are different types of error.

frontend/src/pages/DashboardPage/DashboardPage.tsx
```diff
@@ -33,10 +50,23 @@ function DashboardPage(): JSX.Element {
...
+	// normal outcome of the backend's collection-scoped list.
+	if (isReadDenied) {
+		return (
+			<div className={styles.errorState}>
+				<PermissionDeniedCallout
+					deniedPermissions={[buildDashboardReadPermission(dashboardId)]}
+				/>
+			</div>
+		);
+	}
```
H4ad: Drop this, this is not the correct way of showing permission denied to a page, use `withAuthZPage`.

frontend/src/pages/DashboardPage/PanelEditorPage/PanelEditorPage.tsx
```diff
@@ -113,10 +130,20 @@ function PanelEditorPage(): JSX.Element {
...
 
+	if (isReadDenied) {
+		return (
+			<div className={styles.errorState}>
+				<PermissionDeniedCallout
+					deniedPermissions={[buildDashboardReadPermission(dashboardId)]}
+				/>
+			</div>
+		);
+	}
```
H4ad: Same here, this is not the correct way of showing error to a page, drop in favor of `withAuthZPage`.

H4ad (changes requested):

frontend/src/pages/DashboardPage/PanelEditorPage/PanelEditorPage.tsx
AshwinBhatkal: done

frontend/src/pages/DashboardPage/DashboardContainer/DashboardPageToolbar/DashboardActions/DashboardActions.tsx
```diff
@@ -273,38 +303,31 @@ function DashboardActions({
...
-						onClose={(): void => {
-							setIsSettingsDrawerOpen(false);
-							clearSettingsRequest();
-						}}
-					>
-						<DashboardSettings dashboard={dashboard} />
-					</SettingsDrawer>
-				</>
-			)}
+			<DisabledReasonTooltip reason={editDisabledReason} kind={editDisabledKind}>
```
H4ad: We can bake in the AuthZButton an `disabledTooltip` to address this issue where the button can be disabled by either locked or permission

In the new Button component, I already baked this into the component, so by keeping only `AuthZButton`, this will make our life way easier during migration, that's another reason I want to avoid components other then the ones we already have for AuthZ

frontend/src/pages/DashboardPage/PanelEditorPage/PanelEditorPage.tsx
```diff
@@ -47,22 +49,16 @@ function PanelEditorPage(): JSX.Element {
...
 		useDashboardFetch(dashboardId);
-	// Derived here (not from the store) because the editor route doesn't mount
-	// DashboardContainer, so the store's edit context may be cold on a direct URL.
-	const { isEditable, isLocked, canEditDashboard, editDisabledReason } =
-		useDashboardEditGuard(dashboard);
-
+	// `read` is resolved by the page guard; this resolves `update` and `delete` so
+	// the editor mounts with every permission known.
+	const { isLoading: isPermissionLoading } =
+		useDashboardPermissions(dashboardId);
```
H4ad: nit: I think we should not reload the entire page due to change on create/update/delete, this should only gate the read permission since it's required to perform the read request

frontend/src/pages/DashboardPage/DashboardPage.tsx
```diff
@@ -4,17 +4,25 @@ import { useParams } from 'react-router-dom';
...
 function DashboardPage(): JSX.Element {
 	const { dashboardId } = useParams<{ dashboardId: string }>();
 
 	const { dashboard, isLoading, isError, error, refetch } =
 		useDashboardFetch(dashboardId);
+	// `read` is already resolved by the page guard; this resolves `update` and
+	// `delete` so the subtree mounts once with every permission known, instead of
+	// rendering controls enabled and flipping them.
+	const { isLoading: isPermissionLoading } =
+		useDashboardPermissions(dashboardId);
```
H4ad: nit: Same as the other comment, we should gate the entire page to wait update/delete, this can be performed/waited in the specific UI component

If we really want to optimize these cases, we can introduce a `preloadChecks` in the `withAuthZPage` to avoid this two step loading

H4ad (commented):
Same issue as the list, let's keep only standard messages and avoid custom messages:

Image

frontend/src/pages/DashboardPage/DashboardPage.tsx
AshwinBhatkal: dropped

frontend/src/pages/DashboardPage/DashboardContainer/DashboardPageToolbar/DashboardActions/DashboardActions.tsx
AshwinBhatkal: done

frontend/src/pages/DashboardPage/PanelEditorPage/PanelEditorPage.tsx
AshwinBhatkal: done

frontend/src/pages/DashboardPage/DashboardPage.tsx
AshwinBhatkal: done

H4ad (dismissed):

H4ad (dismissed):

H4ad (approved):

H4ad (approved):

PR #12440, opened 2026-08-06
fix(sentry): stop reporting self-healed chunk load failures

therealpandey:
A tab that outlives a deploy requests hashed assets the new build no longer has. `lazyRetry` already recovers from this by reloading once, so the resulting errors are noise — they spike on every deploy and each one burns a Session Replay (`replaysOnErrorSampleRate: 1.0`).

### Sentry `ignoreErrors`

Filters the whole class. Four patterns because the same failure is worded differently per source:

| Pattern | Source |
|---|---|
| `Unable to preload CSS for` | Vite's own thrown `Error`, identical everywhere |
| `Failed to fetch dynamically imported module` | Chromium |
| `error loading dynamically imported module` | Firefox |
| `Importing a module script failed` | Safari |

`ignoreErrors` is applied as an event processor (`@sentry/core` `eventFilters.js`), so the event is dropped before transport. Replay's error flush hooks `afterSendEvent`, which never fires for a dropped event — so this stops the replay burn too, not just the issue count.

Trade-off, stated plainly: stale-asset failures now produce no Sentry signal at all, including the case where the reload doesn't fix it. A genuinely broken deploy has to be caught from asset 404 rates rather than from Sentry.

### `lazyRetry`

Behaviour is unchanged. One guard added: `setSessionStorageApi` returns `false` when sessionStorage is blocked (iframe, storage disabled), and the retry flag can't persist. The reload was previously issued anyway, so every failed import reloaded forever with no way out. It now reloads only when the flag was actually written.


frontend/src/AppRoutes/index.tsx
```diff
@@ -354,6 +354,10 @@ function App(): JSX.Element {
 					tunnel: window.signozBootData.settings.sentry.tunnel,
 					environment: process.env.ENVIRONMENT,
 					release: process.env.VERSION,
+					// Vite reports a CSS preload failure when a tab's index.html points at a
+					// stylesheet the new build replaced. It is the same deploy race as a stale
+					// JS chunk and recovers through the reload in `lazyRetry`.
+					ignoreErrors: [/Unable to preload CSS for/],
```
H4ad:
```suggestion
					ignoreErrors: [/Unable to preload CSS for/, /Failed to fetch dynamically imported module/],
```

H4ad (approved):
Follow-up: mark all those issues on Sentry as "resolve on next release"

Issue #6143, opened 2024-10-09
Support both `deployment.environment.name` and `deployment.environment`


srikanthccv:
The latest semconv recommends `deployment.environment.name` instead of `deployment.environment`. Make it backwards compatible.

Alexandr-Dolgov:
I Just add some info. It was released in [semantic-conventions v1.27.0](https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.27.0)
> deployment.environment has been deprecated and moved to deployment.environment.name

PR #12441, opened 2026-08-06
feat: generate semantic convention families

srikanthccv:
## Summary

Adds the semantic-convention evolution foundation:

- vendors the OpenTelemetry schema and SigNoz overlay
- generates Go and TypeScript family tables deterministically
- exposes the Go resolver API for family members, current names, and historical names
- adds generation checks and unit tests

Related to #6143.

## Stack

1. #12441 — Foundations (base: main)
2. #12442 — Phase 1 query (base: #12441)
3. #12443 — Phase 1 services (base: #12442)
4. #12444 — Phase 1 closure gate (base: #12443)
5. #12445 — Phase 2 signals (base: #12444)
6. #12446 — Phase 3 migration UX (base: #12445)
7. #12447 — Phase 4 rollout (base: #12446)

**Current layer:** #12441

## Testing

- `go test ./scripts/semconv`
- `go test ./pkg/types/telemetrytypes/semconv`
- `make semconv-check`

## Risk and rollback

This layer is additive apart from CI generation checks. Roll back by reverting this PR; no stored telemetry is changed.


tushar-signoz (commented):
Claude assisted review

Verified locally first: `go run ./scripts/semconv` reproduces both generated files byte-for-byte, `go vet` and both test packages are green, and `scripts/semconv/schema-1.42.0.yaml` is byte-identical to `https://opentelemetry.io/schemas/1.42.0` — nice that the vendored copy is unmodified. The generated TS also passes `oxfmt --check` and `oxlint` against the frontend config, so the new `jsci` jobs won't trip on it.

The two things I'd want resolved before more families get enabled are the rename fan-out collapse and the `apply_to_metrics` union. Everything below that is a design note or a test gap.

---

## 1. Rename fan-out collapses — `scripts/semconv/generate.go:336`

The rename graph stores one successor per `(kind, old)`, so an old name that OTel renames into several different families collapses to whichever edge is processed last. The vendored 1.42.0 schema has 8 such names, and they're the common generic metric attributes: `direction`, `state`, `type`, `pool`, `system.processes.status`, `feature_flag.evaluation.error.message`, plus two k8s metric names.

Generating with `default_enabled: true` and probing the resolver:

```
Current("state",     metric "system.memory.usage")     = db.client.connection.state   (schema: system.memory.state)
Current("state",     metric "system.cpu.time")         = db.client.connection.state   (schema: cpu.mode)
Current("direction", metric "system.disk.io")          = network.io.direction         (schema: disk.io.direction)
Current("pool",      metric "jvm.memory.usage")        = jvm.buffer.pool.name         (schema: jvm.memory.pool.name)
Current("type",      metric "system.filesystem.usage") = system.paging.fault.type     (schema: system.filesystem.type)
```

The winner also absorbs the losers' `apply_to_metrics`, so `db.client.connection.state` ends up claiming `system.cpu.time` / `system.memory.usage` / `system.network.io`, and `network.io.direction` claims `system.disk.io` and `system.disk.merged`.

`rejectSameVersionChains` doesn't catch it — it detects a *chain* (an old that is also a current), not fan-out — so generation succeeds silently. Which family wins is deterministic but arbitrary: it's decided by YAML change order.

Latent right now because `default_enabled: false` and only two families are on, but these are precisely the families the later layers turn on.

## 2. `apply_to_metrics` is unioned, but empty means "all metrics" — `scripts/semconv/generate.go:380`

`apply_to_metrics` is unioned across a family's edges, but in the schema an *absent* `apply_to_metrics` means "applies to every metric". Unioning an unscoped edge with a scoped one turns an unbounded rename into a bounded list — empty should absorb, not be absorbed.

Concretely, family `network.io.direction` merges the 1.24.0 metrics rename `system.network.io.direction → network.io.direction` (no `apply_to_metrics`, so: all metrics) with the 1.25.0 `direction → network.io.direction` scoped to 14 metric names. The generated family carries only those 14, so:

```
Members(KindAttribute, {Name: "system.network.io.direction", Signal: metrics,
                        MetricContext: {MetricName: <any metric outside the 14>}})
  → ["system.network.io.direction"]     // rename never resolves
```

## 3. Missing selector fields widen on one axis and kill on another — `pkg/semconv/semconv.go:107-124`

`FieldKeySelector` is built from user input and often won't have `FieldContext` or `MetricContext` populated — only the strict `resource.` / `attribute.` prefixes fill in a context at all. `matchesSelector` reacts to that missing information in opposite directions on the two axes:

- absent `FieldContext` → the check is skipped → the family matches **more** widely
- absent `MetricContext` with `Signal == metrics` → returns false → the family matches **nothing**

Across the 209 families generated with `default_enabled: true`, 15 carry `apply_to_metrics` and so never resolve at all unless the caller supplies a metric name. On the context axis the widening is small today — 65 families have `Contexts: nil` (they come from `all:`) so the check is a no-op, 142 are still gated by `Signals`, and only 2 are gated by context alone (`telemetry.distro.version`, `user_agent.original`).

It's small *today* for an accidental reason though: no name currently belongs to more than one family, so matching too widely can only ever return the right one. That stops holding the moment the fan-out in (1) is fixed, since `lookupIndex` returns the first match. Worth settling on one policy for "the caller didn't tell us" across all three axes before more families are enabled.

## 4. Unexported fields / accessors — `pkg/semconv/semconv.go:19-27` and `:74`

Could these fields be unexported with accessors? `families_gen.go` is generated into this same package, so it's only a rename in `renderGo`.

The concrete leak is aliasing rather than the field names: `Lookup` returns `Family` by value, but `Old`, `Contexts`, `Signals`, `ApplyToMetrics` are slice headers and `ValueMap` is a map, all pointing at the single package-level table — so `// must not be modified` is unenforceable as written. `All()` is the sharpest edge, since it hands back the table itself and `semconv.All()[0] = Family{}` corrupts it process-wide.

Across the whole stack the only fields read outside this package are `Current`, `Kind` and `len(Old)`; `Contexts`, `Signals` and `ApplyToMetrics` are used only by `matchesSelector`, and `ValueMap` has no reader anywhere yet. So those four could be unexported now at zero cost, and `All()` could return `iter.Seq[Family]`.

I'd leave `Members()` returning the shared precomputed slice — it's on the per-field query path in the later layers and a cloning accessor there is a real regression; a test pinning that contract is cheaper.

## 5. Test gaps — `scripts/semconv/generate_test.go:82`

`TestBuildFamiliesMapsSchemaSectionsToScopes` covers `resources`, `logs` and `metrics` but not `all` — across the whole file the counts are `spans:` ×9, `metrics:` ×4, `resources:` ×1, `logs:` ×1, `all:` ×0.

That's the costly one to skip: `all` is the only section that sets both `allContexts` and `allSignals`, which collapse to `Contexts: nil, Signals: nil` — and nil means *matches every context and every signal* in `matchesSelector`. It's also the path both shipped families take (`db.system` under `1.30.0 all:`, `deployment.environment` under `1.27.0 all:`), so every family in the committed `families_gen.go` goes through the one section no test exercises.

Worth adding alongside it: a family fed by both an `all` edge and a scoped `spans`/`metrics` edge, asserting `allContexts` wins and the accumulated list is discarded rather than unioned.

Other gaps I noticed:

- same-version fan-out and the bare-vs-scoped `apply_to_metrics` merge (1 and 2 above) have no test
- `renderGo` / `renderTypeScript` are only asserted for determinism plus two `Contains` checks, so nothing pins the emitted text and a TS formatting regression surfaces in `jsci` rather than `go test`
- the `unsupported field context` / `unsupported signal` branches are reachable from a hand-written overlay but untested
- `parseSchemaVersion` covers only the wrong-arity case, not `1.2.x`

Separately — `rootFor`'s `rename cycle for ...` branch looks unreachable, since every insertion does `delete(next, current)` first and the dedup `continue` is a no-op. Either drop it or note the invariant.

## 6. Scope belongs on the edge, not the family — `pkg/semconv/semconv.go:19-27`

Flattening scope onto the family loses cross-context renames. Schema 1.19.0 renames into `user_agent.original` from two different contexts in the same version:

```
spans:      http.user_agent:    user_agent.original
resources:  browser.user_agent: user_agent.original
```

which generates `Old: ["browser.user_agent", "http.user_agent"]` with `Contexts: [Attribute, Resource]`. The table can no longer say that `browser.user_agent` only existed at resource level and `http.user_agent` only at span level, so `Members(KindAttribute, {Name: "http.user_agent", FieldContext: FieldContextResource})` returns all three — telling a strict `resource.` query to look for `http.user_agent` where it never existed. The overlay can't correct it either, since `Contexts` / `AddContexts` are family-granular.

This is benign on the forgiving bare-key path, which already ORs col → resource → scope → attribute → body; it bites on the strict prefixed path, where the user has explicitly given the context.

One family out of 209 today, so a note rather than a blocker — but it's the same shape as the two `apply_to_metrics` issues above. Moving `Old` from `[]string` to per-member records carrying their own contexts and metric scope would close all three at once.

Also worth checking: a *pure* move with no rename is invisible to this design entirely. The schema format only has `rename_attributes` / `rename_metrics` (confirmed — those are the only two change types across all 41 versions), so an attribute that keeps its name and changes context produces no edge and gets no fallback.

## 7. Nit — PR description

Tests are at `pkg/semconv`, not `pkg/types/telemetrytypes/semconv`, and there's no `make semconv-check` — the Makefile adds `semconv-generate` and CI calls `go run ./scripts/semconv -check` directly. Might be worth a `semconv-check` target so the CI command has one home.

srikanthccv:
@tushar-signoz 

1, 2, 3, and 6 (point basis on the earlier points IIUC), are all metrics. So the way I am looking about getting changes merged it iterative (would like to keep the plan that way), with initial focus on

- `deployment.environment.name`

For traces and across the traces surface area

Then come to logs, and then come to merics

When each signal is covered, I will have an integration tests suite gets merged to main before the signal coverage that ensures the sem conv changes, when merged, won't break any behaviour, and along with the change for the family behaviour.

Additionally the `default_enabled` is false by default to ensure the surface of the attributes and metrics where this is applied is controlled and provided by us till the point it's fully functional and has test suite to cover the cases.

-

regarding the point 4, throught our codebase we pass around the mutable slices or pointers, such field keys example, and rely on the consumers to not mutate where all the consumers of the args systemically have no need to mutate them.  So following the same reasoning, they mutable are returned. 

The test gaps, I will add additional tests.

srikanthccv:
@tushar-signoz , added the tests, the rest are heavily metrics, I will come there when I come there, because we also discussed about having ability to indicate two metrics mean the same thing or not with the engineer confirmation instead of taking renames as literal renames but can also mean the semantic changes. PTAL

tushar-signoz (approved):
LGTM

therealpandey (approved):

PR #12442, opened 2026-08-06
feat: resolve semantic convention names in trace queries

srikanthccv:
#### Description

- A referenced name in a trace query now resolves to a `LogicalField` (#12499): one field, addressed by the requested spelling, backed by its physical member keys. A semantic-convention family (`deployment.environment.name` / `deployment.environment`) merges into one expression with current-wins precedence; the response keeps the requested spelling.
- `FieldMapper` gets one new method, `ExistsFor` (the per-key presence primitive). `LogicalValueExpr` and `LogicalExistsExpr` build all family SQL in one place from `FieldFor` and `ExistsFor`; no signal implements family logic.
- Statement builders prefetch sibling spellings; the metadata store stays family-blind and autocomplete stays literal. Traces and the resource filter compile per logical field; logs, metrics, and the other signals keep their SQL unchanged.
- The `resolve_semconv_families` feature flag (default: disabled) gates all family behavior. With the flag off, the generated SQL is the same as main; tests pin this. Part of #6143.

#### Additional Information

- Stack: #12441 (merged) → **#12442** → #12443 → #12444 → #12445 → #12446 → #12447. This layer bases on main.
- Rollback: turn the flag off; stored telemetry is untouched.


pkg/modules/rulestatehistory/implrulestatehistory/field_mapper.go
```diff
@@ -64,6 +65,23 @@ func (m *fieldMapper) ColumnFor(ctx context.Context, _ valuer.UUID, _, _ uint64,
...
+// Real columns always exist; labels are checked for key membership.
+func (m *fieldMapper) ExistsFor(ctx context.Context, _ valuer.UUID, _, _ uint64, key *telemetrytypes.TelemetryFieldKey, exists bool) (string, error) {
+	col, err := m.getColumn(ctx, key)
+	if err != nil {
+		return "", err
+	}
+	if col.Name != "labels" || key.Name == "labels" {
+		return "true", nil
+	}
+	pred := fmt.Sprintf("has(JSONExtractKeys(labels), %s)", querybuilder.ClickHouseStringLiteral(key.Name))
```
tushar-signoz: `JSONHas(labels, key)` ?

pkg/modules/rulestatehistory/implrulestatehistory/field_mapper.go
```diff
@@ -64,6 +65,23 @@ func (m *fieldMapper) ColumnFor(ctx context.Context, _ valuer.UUID, _, _ uint64,
...
+// ExistsFor implements the per-key existence primitive of qbtypes.FieldMapper.
+// Real columns always exist; labels are checked for key membership.
+func (m *fieldMapper) ExistsFor(ctx context.Context, _ valuer.UUID, _, _ uint64, key *telemetrytypes.TelemetryFieldKey, exists bool) (string, error) {
+	col, err := m.getColumn(ctx, key)
+	if err != nil {
+		return "", err
+	}
+	if col.Name != "labels" || key.Name == "labels" {
+		return "true", nil
+	}
```
tushar-signoz: If key.name is "labels" then col.Name would always be "labels", condition appears to be weird ?

pkg/querybuilder/family_selectors.go
```diff
@@ -0,0 +1,64 @@
...
+// ExpandKeySelectorsForFamilies adds selectors for the other members of each
+// semantic-convention family that a selector names. The metadata fetched for
+// a query then contains each spelling that MatchingLogicalFields can group.
+// This function is the prefetch of the resolution layer: statement builders
+// call it after they derive the selectors, and the metadata store stays
+// family-blind (autocomplete responses keep the literal spelling that the
+// user typed). It does nothing when the resolve_semconv_families flag is off
+// for the org. Only trace selectors expand today, because that matches the
+// family support. Fuzzy (search-style) selectors never expand.
+func ExpandKeySelectorsForFamilies(ctx context.Context, orgID valuer.UUID, fl flagger.Flagger, selectors []*telemetrytypes.FieldKeySelector) []*telemetrytypes.FieldKeySelector {
```
tushar-signoz, line 32: nit: does it really need flagger and orgId ? we could just pass a bool flag as method is completely mechanical

pkg/querybuilder/family_selectors.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/SigNoz/signoz/pkg/valuer"
+)
+
+// semconvFamiliesEnabled evaluates the resolve_semconv_families flag for the
+// org. A nil flagger means off, so a caller without family support stays
+// literal by default.
+func semconvFamiliesEnabled(ctx context.Context, orgID valuer.UUID, fl flagger.Flagger) bool {
+	if fl == nil {
+		return false
+	}
```
tushar-signoz, line 19: lets not do over loading flagger
edit: over loading on flagger

pkg/querybuilder/family_selectors.go
```diff
@@ -0,0 +1,64 @@
...
+
+	out := selectors
+	seen := make(map[string]bool, len(selectors))
+	for _, selector := range selectors {
+		seen[selector.Name] = true
+	}
+
+	for _, selector := range selectors {
+		if selector.Signal != telemetrytypes.SignalTraces ||
+			selector.SelectorMatchType == telemetrytypes.FieldSelectorMatchTypeFuzzy {
```
tushar-signoz, line 45: would need some clarity on the selectorMatchType, may be in second pass I would understand

pkg/querybuilder/logical_expr.go
```diff
@@ -0,0 +1,94 @@
...
+	if logical.FieldDataType == telemetrytypes.FieldDataTypeString {
+		// The trailing '' keeps single-key semantics for rows without any
+		// member: string maps read '' for an absent key, and negative
+		// operators must keep including such rows (see AddDefaultExistsFilter).
+		values := make([]string, 0, len(memberExprs))
+		for _, expr := range memberExprs {
+			values = append(values, fmt.Sprintf("NULLIF(%s, '')", expr))
+		}
+		return "COALESCE(" + strings.Join(values, ", ") + ", '')", nil
+	}
```
tushar-signoz, line 53: Didn't get this part, why not return NULL ?

pkg/querybuilder/logical_expr.go
```diff
@@ -0,0 +1,94 @@
...
+		if err != nil {
+			return "", err
+		}
+		guards = append(guards, guard)
+	}
+	combined := "(" + strings.Join(guards, " OR ") + ")"
+	if exists {
+		return combined, nil
+	}
+	return "NOT " + combined, nil
```
tushar-signoz, line 95: Should it not be ANDed for negative existence ?

pkg/querybuilder/where_clause_visitor.go
```diff
@@ -48,8 +51,11 @@ type filterExpressionVisitor struct {
 }
 
 type FilterExprVisitorOpts struct {
-	Context            context.Context
-	OrgID              valuer.UUID
+	Context context.Context
+	OrgID   valuer.UUID
+	// Flagger evaluates the resolve_semconv_families flag during resolution.
+	// A nil Flagger keeps resolution literal.
```
tushar-signoz, line 57: nit: usecase specific comment

pkg/statementbuilder/resourcefilter/condition_builder.go
```diff
@@ -14,12 +15,15 @@ import (
 
 type defaultConditionBuilder struct {
 	fm qbtypes.FieldMapper
+	// fl evaluates the resolve_semconv_families flag during resolution.
+	// A nil flagger keeps resolution literal.
```
tushar-signoz, line 19: 👎

pkg/statementbuilder/resourcefilter/condition_builder.go
```diff
@@ -44,10 +48,74 @@ func keyIndexFilter(key *telemetrytypes.TelemetryFieldKey) any {
...
+	if len(conditions) == 1 {
+		return conditions[0]
+	}
+	return sb.Or(conditions...)
+}
+
+func memberPresenceCondition(sb *sqlbuilder.SelectBuilder, column string, members []*telemetrytypes.TelemetryFieldKey, exists bool) string {
+	conditions := make([]string, 0, len(members))
+	for _, member := range members {
+		field := fmt.Sprintf("simpleJSONHas(%s, '%s')", column, member.Name)
```
tushar-signoz, line 99: JSONHas ?

pkg/querybuilder/family_selectors.go
srikanthccv: Would you say the same thing about everywhere where flagger is used? i.e Why pass the orgID and falgger, instead pass bool to check feature?

I would keep the feature existing same regardless of the function, it's better that way instead of passing a boolean for feature flag overall.

pkg/querybuilder/family_selectors.go
srikanthccv: I didn't get this, what is "over loading flagger"

pkg/querybuilder/family_selectors.go
srikanthccv: The family selector function adds the selector for the all member of the family but only when the selector type is exact, otherwise we can't add them as that would result in unrelated keys.

pkg/querybuilder/logical_expr.go
srikanthccv: returning NULL is incorrect, would a prepared query help you understand this better? there are tests in associated, let me know if this is still not clear.

pkg/querybuilder/logical_expr.go
srikanthccv: `deployment.environment`, `deployment.environment.name`

and the two examples,

```
deployment.environment exists
```

condition
```
(deployment.environment exists OR deployment.environment.name exists)
```

```
deployment.environment NOT exists
```

condition

```
NOT (deployment.environment exists OR deployment.environment.name exists)
```

It is `AND`, when we apply de-morgan, but either way it's correct,

it could be written also as

```
(deployment.environment NOT exists AND deployment.environment.name NOT exists)

```

but given the `NOT` as negation is better represents the intent.

pkg/statementbuilder/resourcefilter/condition_builder.go
srikanthccv: is it about the comment?

pkg/statementbuilder/resourcefilter/condition_builder.go
srikanthccv: everywhere in the `resource` builder, we deliberately use the `simpleJSONX` so will keep it that way

pkg/modules/rulestatehistory/implrulestatehistory/field_mapper.go
srikanthccv: `getColumn` maps a real column name to its column and everything else to `labels`. So the _always exists_ cases are two, the key is a real column, such as `rule_id`, or the key is the `labels` column itself.

There are two ways to express the same, either above one, oe

```
if col.Name == "labels" && key.Name != "labels" {
    return JSONHas(labels, key.Name)
    ...
}
return "true", nil
```

The code in opposite of this, still the same rule, does this clarify and look more readable?

pkg/modules/rulestatehistory/implrulestatehistory/field_mapper.go
srikanthccv: done

pkg/modules/rulestatehistory/implrulestatehistory/field_mapper.go
srikanthccv: Done

pkg/querybuilder/family_selectors.go
tushar-signoz: I meant lets not use flagger being nil as a signal.

srikanthccv: So there two options

- pass the bool as discussed in other thread
- pass the nil falgger, this not as a pattern to follow, but because we have stacked PRs and the support for logs and metrics is in draft.

for the passing the bool, did you mean, pass it just to this function so we can avoid nil check? Or did you mean pass bool in other instances also?

pkg/statementbuilder/resourcefilter/condition_builder.go
```diff
@@ -44,10 +48,74 @@ func keyIndexFilter(key *telemetrytypes.TelemetryFieldKey) any {
...
+	caseInsensitive bool,
+) string {
+	conditions := make([]string, 0, len(members))
+	for _, member := range members {
+		patterns := valueForIndexFilter(op, member, value)
+		switch values := patterns.(type) {
+		case []string:
+			for _, pattern := range values {
+				conditions = append(conditions, sb.Like(column, pattern))
+			}
```
tushar-signoz, line 81: Should this also consider caseInsensitive flag ?

pkg/statementbuilder/resourcefilter/condition_builder.go
```diff
@@ -44,10 +48,74 @@ func keyIndexFilter(key *telemetrytypes.TelemetryFieldKey) any {
...
+		default:
+			if caseInsensitive {
+				conditions = append(conditions, sb.ILike(column, values))
+			} else {
+				conditions = append(conditions, sb.Like(column, values))
+			}
```
tushar-signoz, line 87: Currently we have the indexes using lower() (correct me if this is not the case), should we skip adding condition in this case ?

pkg/statementbuilder/resourcefilter/condition_builder.go
tushar-signoz: Yes, I mis understood something. So this part is just for the index help ? should we use the keyIndexHelper that is already there ?

tushar-signoz (approved):
LGTM

pkg/statementbuilder/resourcefilter/condition_builder.go
srikanthccv: `memberPresenceCondition` is used when user writes expression `deployment.environment EXISTS`


it creates condition, `deployment.environment EXISTS OR deployment.environment.name EXISTS`

and the key index filter is already there. so the end expression will be

`(deployment.environment EXISTS OR deployment.environment.name EXISTS) AND (labels LIKE '% deployment.environment%' OR labels LIKE '% deployment.environment.name%')`

pkg/statementbuilder/resourcefilter/condition_builder.go
srikanthccv: Discussed, this should not be required in practice i.e the values is list and the list can't be input for `CONTAINS` or `LIKE`

pkg/statementbuilder/resourcefilter/condition_builder.go
srikanthccv: both indexes exist, `labels` and `lower(labels)`

pkg/querybuilder/family_selectors.go
srikanthccv: discussed; keeping as is, and it will be removed soon when the logs and metrics implementation lands.

therealpandey (approved):

PR #12445, opened 2026-08-06
feat: resolve semconv families across logs and metrics

srikanthccv:
#### Description

Phase 2 of #6143: family resolution for logs and metrics, behind the `resolve_semconv_families` flag (default off). One commit, rebuilt on the storage contract of #12802.

- **Registry.** Each member carries the scope of its own rename edges. A fan-out keeps one membership per target, an unscoped rename stays unbounded, and an ambiguous name stays literal. Family fields are unexported, `All()` iterates, and `make semconv-check` backs CI.
- **Gate.** The family lookup applies to every signal. A string entry of a family signal under the resource or attribute context groups into the family. `LogicalRead` merges the members through each storage's `Read`, so logs and metrics need no family code of their own. The traces storage keeps its behaviour.
- **Logs.** Filters, select fields, group by, order by, and aggregations resolve the family. The mid-migration state, where metadata holds one spelling and the query names the other, reads the one stored spelling in the filter and in the group by.
- **Metrics.** Label families expand into the stored spellings: dotted, normalized, and `resource_`-prefixed. Every `metric_name` filter unions the storage names of a metric-name family. The querier resolves type and temporality through the same union, so a dashboard on the old name keeps working. A family of labels keeps the metrics keyless contract: no guard, and no NULL group.
- **Values.** Suggestions union the family spellings on traces, logs, metric labels, and metric names. Related values resolve the family in the selected column, the search narrowing, and the existing-query filter.
- **Dead code.** `pkg/query-service/metrics/transition.go` is removed. The v3 and v4 readers resolve the three CPU metric renames through the registry, with the same one-directional exact-name semantics. The legacy metric storage spellings live in a deletable shim, `pkg/querybuilder/legacy_metric_spellings.go`.
- **Enablement.** `deployment.environment.name` resolves on traces, logs, and metrics. `db.system.name` is disabled: its value domain also renamed, and no value mapping is read yet.
- **Integration.** The `semconvfamilies` suite covers the logs matrix, the metric label and metric-name families in every storage layout, the dual-emission double count as the documented trade-off, and the `fields/keys` and `fields/values` pins.

#### Additional Information

- Verified: full Go suite, lint clean, `semconvfamilies` 131 cases with the flag on and off, and the common, logs, traces, and metrics querier suites (792 cases) with the flag off.
- A family whose members all read their sentinel as a value (metrics labels) reads that way as a family. See `familyAbsence` in `pkg/querybuilder/logical_expr.go`. Without it, a metrics group by on a family key put keyless series in a NULL group instead of the empty string group that a single label gives.
- Rollout caveat, pinned by a test: a target that emits both storage names of a metric-name family double-counts in `sum()` during the overlap window. Reading both names is the feature.
- Perf validation before broad enablement stays tracked in the design ledger: a family condition over a materialized member bypasses that column's skip index, and metric family terms read several label extractions per row in the time-series CTE.


pkg/querybuilder/family_selectors.go
```diff
@@ -4,57 +4,66 @@ import (
...
+	if !SemconvFamiliesEnabled(ctx, orgID, fl) {
 		return selectors
 	}
 
+	// Two selectors may share a name under different contexts, signals, or
+	// data types, and each needs its own sibling selectors, so the dedupe key
+	// is the full identity.
+	identity := func(selector *telemetrytypes.FieldKeySelector, name string) string {
+		return selector.Signal.StringValue() + ";" + selector.FieldContext.StringValue() + ";" + selector.FieldDataType.StringValue() + ";" + name
+	}
```
tushar-signoz, line 45: nit: can we have this method on the FieldKeySelector struct itself ?

pkg/telemetryschema/tracestelemetryschema/condition_builder.go
```diff
@@ -20,15 +19,12 @@ import (
 
 type conditionBuilder struct {
 	fm qbtypes.FieldMapper
-	// fl evaluates the resolve_semconv_families flag during resolution.
-	// A nil flagger keeps resolution literal.
-	fl flagger.Flagger
```
tushar-signoz, line 25: nit: lets keep fl if we have already wired it in

tushar-signoz (approved):

srikanthccv:
Superseded by https://github.com/SigNoz/signoz/pull/12802, and will have another PR to support logs and metrics instead.

PR #12448, opened 2026-08-06
feat(authz): add dashboard permissions

AshwinBhatkal:
#### Description

Registers the `dashboard` authz kind and adds the shared pieces the two dashboard PRs stack on. No dashboard UI changes here.

- **Registers `dashboard`** in `cmd/authz.go` and regenerates `permissions.type.ts`. `public-dashboard` is deliberately *not* registered: those routes are open access, the only verb any of them checks is an anonymous read seeded by migration, and nothing in the frontend builds a permission against the kind.
- **Adds a Dashboards panel** to the custom-role editor. The map stays a total `Record`, so a newly registered resource still has to be given a panel — its length assertions are that guard.
- **`disabledTooltip` on `AuthZTooltip` / `AuthZButton`**, for a block the consumer already knows about that is *not* a permission — a lock, an integration dashboard, a deliberately read-only mount. It takes precedence over `checks`, which are then **skipped entirely** rather than fetched for a control that is unavailable either way. Callers set it only when that block is the real obstacle, so a missing permission still surfaces its own standard wording. That is how `update` ends up outranking the lock.
- **`preloadChecks` on the HOCs**, fetched in the same batch as `checks` and never gated on, so a control below the guard resolves from cache instead of firing a second round trip and flipping once it lands. `AuthZGuard` is untouched.
- `AuthZTooltip` also takes `side` and `tooltipMessage`, and keeps its reason up when the dead control is clicked — Radix closes on both pointerdown and click, and merges its own handlers after the trigger's regardless of `preventDefault`, so the close is filtered in the component.
- **Dashboard permission hooks**: `canEdit` is `read && update` per guide rule 2, `read` reports loading separately from the rest, and the lock hook checks `dashboard:update` plus the one non-permission rule the handler adds — an integration dashboard can never be locked.

Screenshot 2026-09-10 at 11 32 56 PM

Screenshot 2026-09-10 at 11 32 41 PM

#### Additional Information

- **Denial copy comes from the components.** There is no dashboard permission-message module any more; the non-permission strings live in the i18n bundle. `useIsOrgAdmin` goes with it — the lock was its only caller.
- The lock hook no longer considers who created the dashboard — access is decided by role. **The handler still enforces creator-or-admin**, so that check wants removing backend-side too, or a non-creator with `update` gets an enabled Unlock and a 403.
- The grant-all MSW default is gone: each suite states the permissions it needs, so an authz test can't pass for the wrong reason.
- Stacked: #12438 then #12449, and rebased onto current `main`. The three squash-merge in order with no conflicts — verified by replaying the whole sequence locally, including the two files this PR shares with #12449 (`utils/permission/index.ts` and `AppRoutes/__tests__/Private.test.tsx`, additions ~24 and ~68 lines apart). Each branch above needs a rebase after the one below merges.


frontend/src/hooks/dashboards/useDashboardPermissions.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Resource-level dashboard permissions. Pass `enabled: false` while the id is
+ * unknown so no check is fired against an empty selector.
+ */
+export function useDashboardPermissions(
+	dashboardId: string,
+	options?: { enabled?: boolean },
+): DashboardPermissions {
+	const enabled = (options?.enabled ?? true) && !!dashboardId;
```
AshwinBhatkal: dashboardID check should not required here

frontend/src/hooks/dashboards/useDashboardPermissions.ts
```diff
@@ -0,0 +1,70 @@
...
+	);
+
+	const checks = useMemo(
+		() => [readPermission, updatePermission, deletePermission],
+		[readPermission, updatePermission, deletePermission],
+	);
+
+	const { permissions, isLoading } = useAuthZ(checks, { enabled });
+
+	const canRead = permissions?.[readPermission]?.isGranted ?? false;
```
AshwinBhatkal: can this be more reliable? Don't want this much optional chaining throughout the chain. Maybe useAuthz can change differently

frontend/src/hooks/dashboards/useDashboardLockPermission.ts
```diff
@@ -0,0 +1,63 @@
...
+	dashboardId: string;
+	createdBy: string | undefined;
+	enabled?: boolean;
+}): DashboardLockPermission {
+	const { user } = useAppContext();
+	const { canEdit, isLoading: isEditLoading } = useDashboardPermissions(
+		dashboardId,
+		{ enabled },
+	);
+	const { permissions, isLoading: isAdminLoading } = useAuthZ(ADMIN_CHECKS, {
```
AshwinBhatkal: this is called by useDashboardPermissions internally anyways, so why another one here? Reuse it?

frontend/src/lib/authz/hooks/useAuthZ/types.ts
```diff
@@ -97,5 +97,11 @@ export type UseAuthZResult = {
 	 * Checks that resolved as not granted (empty while loading/error).
 	 */
 	deniedPermissions: BrandedPermission[];
+	/**
+	 * Whether a single check resolved as granted. False while loading, on error,
+	 * and for any permission that wasn't part of `checks` — so callers get a plain
+	 * boolean instead of walking a possibly-null map.
```
H4ad:
```suggestion
     * Use this to check if a specific permission is granted, false while loading or on error.
```

frontend/src/lib/authz/components/DisabledReasonTooltip/DisabledReasonTooltip.tsx
H4ad, line 1: Drop this component, you can move the prop `side` to AuthZTooltip, any permission error should be shown in the same way (standardized by AuthZTooltip)

frontend/src/lib/authz/hooks/useIsOrgAdmin.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const CHECKS = [IsAdminPermission];
+
+/**
+ * Whether the caller holds the managed org-admin role, resolved through authz
+ * rather than `user.role` so "admin" means what the authz store means.
+ *
+ * Named once here so the handful of backend rules that still fall back to the
+ * admin role have a single place to change when `legacy.ts` goes away.
+ */
```
H4ad:
````suggestion
````

File: frontend/src/mocks-server/handlers.ts [hunk 6]
```diff
@@ -258,4 +259,7 @@ export const handlers = [
 	rest.get('http://localhost/locales/en-US/common.json', (_, res, ctx) =>
 		res(ctx.status(200), ctx.json(commonEnTranslation)),
 	),
+	// Grant-all default so suites that don't care about permissions behave as they
+	// did before their page was authz-gated. Override per test with server.use().
+	setupAuthzAdmin(),
```
H4ad:
```suggestion
```

We should not do this, by default, we should update/fix each test while adopting authz

This can introduce false-positives on tests specifically for authz by mistake because it will fallback to admin

frontend/src/container/RolesSettings/permissions.config.ts
```diff
@@ -37,9 +38,19 @@ export interface ResourcePanelConfig {
 /**
  * Do not use CoretypesTypeDTO to represent this,
  * we want to add resource panel configs for only types we actually are using,
- * not all of them
+ * not all of them. Unmapped resources fall back in `getResourcePanel`.
  */
-export const RESOURCE_PANELS: Record<AuthZResource, ResourcePanelConfig> = {
+export const RESOURCE_PANELS: Partial<
+	Record<AuthZResource, ResourcePanelConfig>
+> = {
```
H4ad: Do not convert to partial, it's required to avoid drift when adding a new resource.

Also, it's missing public-dashboard

H4ad (changes requested):

cmd/authz.go
```diff
@@ -96,6 +96,8 @@ func runGenerateAuthz(_ context.Context) error {
 		coretypes.NewResourceRef(coretypes.ResourceServiceAccount).String():                true,
 		coretypes.NewResourceRef(coretypes.ResourceRole).String():                          true,
 		coretypes.NewResourceRef(coretypes.ResourceMetaResourceFactorAPIKey).String():      true,
+		coretypes.NewResourceRef(coretypes.ResourceMetaResourceDashboard).String():         true,
+		coretypes.NewResourceRef(coretypes.ResourceMetaResourcePublicDashboard).String():   true,
```
vikrantgupta25: this shouldn't be added

vikrantgupta25 (changes requested):

cmd/authz.go
vikrantgupta25: or is this needed to create a permission check on public page ?

vikrantgupta25: if yes then we shouldn't be checking any permission there as its `open access` technically. LEt the backend API fail

frontend/src/lib/authz/components/DisabledReasonTooltip/DisabledReasonTooltip.tsx
AshwinBhatkal: This component is not just for that. I can pass in side to AuthZTooltip, yes

But this component is for showing disabled buttons that are not denied because of permission. Like locked dashboards.

Reusing the styles, so nothing should deviate in styles as such

frontend/src/lib/authz/hooks/useAuthZ/types.ts
AshwinBhatkal: done

frontend/src/lib/authz/hooks/useIsOrgAdmin.ts
AshwinBhatkal: done

frontend/src/mocks-server/handlers.ts
AshwinBhatkal: removed

frontend/src/container/RolesSettings/permissions.config.ts
AshwinBhatkal: missed it. reverted

cmd/authz.go
namanv3: okay I don't remember what the discussion was but you had told me that this is the base change needed to start the UI work for FGA 😛

vikrantgupta25: only the `dashboard` 😅  not `public-dashboard`

AshwinBhatkal: resolved

frontend/src/lib/authz/components/DisabledReasonTooltip/DisabledReasonTooltip.tsx
H4ad: Then move out of the authz lib, this is not related to authz

frontend/src/lib/authz/components/DisabledReasonTooltip/DisabledReasonTooltip.tsx
```diff
@@ -0,0 +1,137 @@
...
+	asChild?: boolean;
+	/**
+	 * `denied` — an access problem: the user lacks the permission and can only get
+	 * it from someone else. Styled as an error, matching `AuthZTooltip`.
+	 *
+	 * `blocked` — a state problem the user can act on themselves: a locked
+	 * dashboard, or an integration-owned or legacy one. Styled neutrally, since
+	 * it is not an error.
+	 */
+	kind?: 'denied' | 'blocked';
```
H4ad: The style is not the only thing we should match, the error messages as well, they need to be the standard error messages

So you can remove any reference for authz from this component and move out of the authz.

frontend/src/lib/authz/components/DisabledReasonTooltip/DisabledReasonTooltip.tsx
H4ad: After reading the usage of this component in the other PRs, let's delete in favor of introducing `disabledTooltip` to the AuthZ, in this way, we can show the `disabledTooltip` when the button is disabled by the consumer, and show always the consumer disabled reason first instead of showing the permission (in case both are bad)

frontend/src/hooks/dashboards/dashboardPermissionReasons.ts
```diff
@@ -0,0 +1,33 @@
...
+	'You need delete access on this dashboard to delete it';
+
+export const DASHBOARD_NO_CREATE_PERMISSION_REASON =
+	'You need create access on dashboards';
+
+export const DASHBOARD_NO_LIST_PERMISSION_REASON =
+	'You need list access on dashboards to browse them';
+
+export const DASHBOARD_CLONE_DENIED_REASON =
+	'Cloning needs read access on this dashboard and create access on dashboards';
```
H4ad: No custom message for standard permission checks, let's use what the component already provide to us

frontend/src/lib/authz/components/DisabledReasonTooltip/DisabledReasonTooltip.tsx
AshwinBhatkal: this sounds good

frontend/src/lib/authz/components/DisabledReasonTooltip/DisabledReasonTooltip.tsx
AshwinBhatkal: removed component

frontend/src/hooks/dashboards/dashboardPermissionReasons.ts
AshwinBhatkal: removed

frontend/src/lib/authz/components/withAuthZ/withAuthZ.utils.tsx
```diff
@@ -86,15 +98,23 @@ export function createAuthZHOC<P extends object>(
...
+				? preloadChecks(props, router)
+				: preloadChecks;
 
 		return (
-			<Guard checks={resolvedChecks} {...guardProps}>
+			<Guard
+				checks={resolvedChecks}
+				preloadChecks={resolvedPreload}
+				{...guardProps}
+			>
```
H4ad: The implementation for this, instead of spanning across multiple components, could be just:

```suggestion
		const resolvedPreload =
			typeof preloadChecks === 'function'
				? preloadChecks(props, router)
				: preloadChecks;
		const _ = useAuthZ(resolvedPreload);
	

        return (
			<Guard checks={resolvedChecks}  {...guardProps}>
```

AshwinBhatkal: done

frontend/src/lib/authz/components/AuthZTooltip/AuthZTooltip.tsx
```diff
@@ -61,6 +94,20 @@ function AuthZTooltip({
...
+	 * filtered here. Clicking a dead control does nothing, which is exactly when
+	 * its reason is still wanted, so a close is ignored while the pointer remains
+	 * on it. Everything else stays Radix's to decide.
+	 */
+	const handleOpenChange = useCallback((next: boolean): void => {
+		if (!next && isPointerOverRef.current) {
+			return;
+		}
+		setIsOpen(next);
+	}, []);
```
H4ad: nit: can be prevented with: https://github.com/SigNoz/signoz/blob/30dcb944eb4f788cff2c67366d983cdccce47211/frontend/src/components/TanStackTableView/TanStackHoverTooltip.tsx#L29-L38

H4ad, line 109: Note to myself: I can likely replace this with tooltipContentProps (similar pattern from TanstackHoverTooltip)

H4ad (dismissed):

H4ad (approved):

therealpandey (approved):

vikrantgupta25 (approved):

PR #12449, opened 2026-08-06
feat(dashboards): gate the dashboards list on fine-grained authz

AshwinBhatkal:
#### Description

Applies the authz list-page pattern to the dashboards list and gates the row actions per dashboard.

- **The table is blocked** without `list`, via an inline callout. **New dashboard stays enabled** if the caller holds `create` — it is independent of `list`.
- **Everything that only feeds the table goes with it.** The search box and the filter chips edit a query that can only be run through the list request, so they are disabled rather than hidden, and each carries the standard denial on hover. Two tooltip zones cover them — the search field, and the chips with Clear — instead of one tooltip per control.
- **The views rail is one guarded block**, not a column of dead controls. Every view leads to a table the caller cannot see, and the saved views behind it need the same grant, so the rail carries a single callout. That also drops an empty state that was asserting something it could not know ("No saved views yet" when the request was never made).
- **Row actions run their own checks** through `AuthZButton`, so the menu opens immediately. View, Open in New Tab and Copy Link need no permission and no longer wait behind the gated rows; all nine rows now render through the same component rather than three of them being plain buttons.
- **Delete gates on `delete` alone** (guide rule 3), so it survives without `read`. **Rows are never gated** (rule 4): the list is collection-scoped and deliberately returns rows the caller cannot `read`, so every row stays clickable and the denial is explained on arrival.
- **Per-row checks fire lazily**, when a menu opens rather than for all 20 rows, and the menu itself renders lazily — it previously evaluated its whole tree and instantiated two mutations per row on every render.
- **Viewers can manage saved views again.** Save, rename and delete were behind `edit_dashboard`; the backend only requires `list`, and the views request now follows the same grant.
- **The row menu's lock gates on the dashboard's `source`**, not on who created it — an integration dashboard can never be locked, and everything else is decided by `update`.
- **Route gating**: the three dashboard routes bypass the legacy role check, so an FGA user with no managed role isn't bounced to `/un-authorized`.
- **Guide updated to match.** The list-page rule previously said to leave the search, filters and views interactive. It now says to disable what only shapes a blocked request, and to gate a region once rather than repeat a tooltip per control.

#### Screenshots
Screenshot 2026-09-10 at 11 13 20 PM

Screenshot 2026-09-10 at 11 14 09 PM

Screenshot 2026-09-10 at 11 22 32 PM
Screenshot 2026-09-10 at 11 16 41 PM
Screenshot 2026-09-10 at 11 17 00 PM
Screenshot 2026-09-10 at 11 18 45 PM
Screenshot 2026-09-10 at 11 18 52 PM
Screenshot 2026-09-10 at 11 16 31 PM

 

#### Additional Information

- `DASHBOARD` and `DASHBOARD_PANEL_EDITOR` are registered for authz here, though their pages ship in #12438. Harmless in this order — until this PR merges those routes keep their legacy role checks, which is the fail-safe direction.
- Two callouts show when `list` is denied, one on the rail and one where the table would be. They are separate blocks with separate reasons to exist, but it is more denial than the guide's "state it once".
- A 403 from the list request still renders as "Invalid query", because `ErrorState` treats every 4xx as a client error. Pre-existing and not authz-specific, so left alone.
- Saved views are org-shared with no ownership rule, so anyone holding `list` can delete anyone's view. Unchanged here — flagging it as a product question.
- Requires #12438. Rebased onto current `main`; shares `utils/permission/index.ts` and `AppRoutes/__tests__/Private.test.tsx` with #12448, which merges without conflict.

### DIFF
--- a/frontend/docs/authz-guide.md
+++ b/frontend/docs/authz-guide.md
@@ -107,6 +107,32 @@ denial message and the loading state.
 Prefer the HOC (`withAuthZ*`); reach for the JSX guard (`AuthZGuard*`) when the gate depends on conditional rendering
 and an HOC cannot express it. The components README has the full decision tree and how to build the `checks` array.
 
+## Worked example
+
+Dashboards are fully migrated and exercise most of this guide:
+
+| Piece | Where |
+| --- | --- |
+| Permission constants and builders | `src/lib/authz/hooks/useAuthZ/permissions/dashboard.permissions.ts` |
+| Resource-level hook (`canEdit` = read AND update, per rule 2) | `src/hooks/dashboards/useDashboardPermissions.ts` |
+| Collection-level hook (`list` / `create`) | `src/hooks/dashboards/useDashboardCollectionPermissions.ts` |
+| A compound rule the backend also enforces handler-side | `src/hooks/dashboards/useDashboardLockPermission.ts` |
+| Tooltip copy for disabled controls | `src/hooks/dashboards/dashboardPermissionReasons.ts` |
+| List-page pattern (table guarded, chrome kept, create independent of list) | `src/pages/DashboardsListPage/components/DashboardsList/DashboardsList.tsx` |
+| Per-row checks fired lazily, on menu open (rule 4) | `src/pages/DashboardsListPage/components/ActionsPopover/` |
+| Denial explained in place rather than as a load failure | `src/pages/DashboardPage/DashboardPage.tsx` |
+
+Two things it settled that aren't obvious from the rules alone:
+
+- **Disabled vs denied, while loading.** Permission checks are async, so a
+  control is disabled both in flight and when denied. Show the reason only once
+  the check resolves — an empty tooltip, or one that briefly claims a denial
+  that doesn't hold, is worse than none. Tests must wait on a post-resolution
+  signal, since a `toBeDisabled` assertion also passes during loading.
+- **Lock (or any non-authz block) wins over permission.** When a control is
+  blocked for a reason the user can act on, say that one; naming the missing
+  permission first sends them to an admin when they only needed to unlock.
+
 ## Migration checklist
 
 Use the existing components. Create a new one only if none fit.
--- a/frontend/docs/authz-guide.md
+++ b/frontend/docs/authz-guide.md
@@ -34,6 +34,9 @@ These hold for every page. The per-pattern sections below only add to them.
 4. **Never gate per row.** If a user can `list`, render every row. Check `read` only when the row is opened (drawer or
    detail route).
 5. **Gate the narrowest thing that works**, a button over a section, a section over a page.
+   - **Gate a page on `read` alone.** It is the only verb the page's own request needs. `update` and `delete` gate
+     individual controls, so waiting for them holds up the whole page for nothing — pass them as `preloadChecks` and
+     they resolve in the same request, leaving the controls to read from cache.
 6. **A resource may be gated while a sub-resource is not.** A user without `read` on Service Accounts can still hold
    `create` on API Keys, so blocking the outer container would hide work they are allowed to do.
 7. **Verbs not covered here** (`attach`, `detach`, `assignee`) behave like `delete`: gate the control that triggers
@@ -95,6 +98,11 @@ Gate the control with `AuthZButton`, or `AuthZTooltip` for a non-button trigger 
 Blocking is always a visible denial, never a silent removal. Use the components in
 [`lib/authz/components`](../src/lib/authz/components/README.md) rather than hand-rolling a check, they carry the
 denial message and the loading state.
+
+**Denial copy comes from the components.** Never write a custom message for a permission check. `disabledTooltip` is
+for a block that is *not* a permission — a lock, an immutable resource, a mount that is deliberately read-only. It
+takes precedence over the checks, which are then skipped, so set it only when that block is the real obstacle: a
+missing permission must still surface its own wording.
 
 | Scope           | Component                                 | Denied state                                |
 | --------------- | ----------------------------------------- | ------------------------------------------- |
--- a/frontend/docs/authz-guide.md
+++ b/frontend/docs/authz-guide.md
@@ -107,32 +107,6 @@ denial message and the loading state.
 Prefer the HOC (`withAuthZ*`); reach for the JSX guard (`AuthZGuard*`) when the gate depends on conditional rendering
 and an HOC cannot express it. The components README has the full decision tree and how to build the `checks` array.
 
-## Worked example
-
-Dashboards are fully migrated and exercise most of this guide:
-
-| Piece | Where |
-| --- | --- |
-| Permission constants and builders | `src/lib/authz/hooks/useAuthZ/permissions/dashboard.permissions.ts` |
-| Resource-level hook (`canEdit` = read AND update, per rule 2) | `src/hooks/dashboards/useDashboardPermissions.ts` |
-| Collection-level hook (`list` / `create`) | `src/hooks/dashboards/useDashboardCollectionPermissions.ts` |
-| A compound rule the backend also enforces handler-side | `src/hooks/dashboards/useDashboardLockPermission.ts` |
-| Tooltip copy for disabled controls | `src/hooks/dashboards/dashboardPermissionReasons.ts` |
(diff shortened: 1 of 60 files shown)


frontend/docs/authz-guide.md
```diff
@@ -107,6 +107,32 @@ denial message and the loading state.
 Prefer the HOC (`withAuthZ*`); reach for the JSX guard (`AuthZGuard*`) when the gate depends on conditional rendering
 and an HOC cannot express it. The components README has the full decision tree and how to build the `checks` array.
 
+## Worked example
```
AshwinBhatkal: remove this?

frontend/src/pages/DashboardsListPage/components/DashboardsList/DashboardsList.tsx
H4ad, line 1: You should block the interaction with search input when no permission to list, similar to what you did for filters.

frontend/src/pages/DashboardsListPage/components/DashboardsList/DashboardsList.tsx
```diff
@@ -43,16 +46,19 @@ const PAGE_SIZE = 20;
...
 	const { user } = useAppContext();
-	const [editDashboard, canCreateNewDashboard] = useComponentPermission(
-		['edit_dashboard', 'create_new_dashboards'],
-		user.role,
-	);
-	const canEdit = !!editDashboard;
+	// `list` also authorizes pinning and saved views, so it gates the table only.
+	// Already resolved — the page gates on it before mounting this tree. A failed
+	// check falls open: fetch anyway and let the API decide, so an authz outage
+	// doesn't leave a permanently empty list.
```
H4ad: Avoid this and other cases of comments that does not add value to the code.

frontend/docs/authz-guide.md
```diff
@@ -107,6 +107,32 @@ denial message and the loading state.
...
+| Piece | Where |
+| --- | --- |
+| Permission constants and builders | `src/lib/authz/hooks/useAuthZ/permissions/dashboard.permissions.ts` |
+| Resource-level hook (`canEdit` = read AND update, per rule 2) | `src/hooks/dashboards/useDashboardPermissions.ts` |
+| Collection-level hook (`list` / `create`) | `src/hooks/dashboards/useDashboardCollectionPermissions.ts` |
+| A compound rule the backend also enforces handler-side | `src/hooks/dashboards/useDashboardLockPermission.ts` |
+| Tooltip copy for disabled controls | `src/hooks/dashboards/dashboardPermissionReasons.ts` |
+| List-page pattern (table guarded, chrome kept, create independent of list) | `src/pages/DashboardsListPage/components/DashboardsList/DashboardsList.tsx` |
+| Per-row checks fired lazily, on menu open (rule 4) | `src/pages/DashboardsListPage/components/ActionsPopover/` |
+| Denial explained in place rather than as a load failure | `src/pages/DashboardPage/DashboardPage.tsx` |
```
H4ad: You can drop this part, no need to have this.

frontend/src/lib/authz/components/README.md
```diff
@@ -43,6 +43,10 @@ const checks = [buildPermission('read', buildObjectString('dashboard', dashboard
 
 When adding authz to a new resource, create a permissions file under `lib/authz/hooks/useAuthZ/permissions/`.
 
+`dashboard.permissions.ts` (below) is a real file — see it, and the hooks that
+consume it in `hooks/dashboards/`, for a worked example of a fully migrated
+resource.
+
```
H4ad:
```suggestion
```

frontend/src/pages/DashboardsListPage/components/ActionsPopover/ActionsPopoverContent.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+	// The row's checks only fire when the menu opens, so hold the body until they
+	// resolve — items must never render enabled and then flip to disabled.
+	if (isPermissionLoading || isLockPermissionLoading) {
+		return (
+			<div className={styles.content}>
+				<Spinner size="small" height="64px" />
+			</div>
+		);
+	}
```
H4ad: We don't have guidelines for action items but we should avoid doing loading like this.

This feels way slower than it actually is, move the checks by button with AuthZButton instead, we should only disable the buttons that actually need the permission checks and show the other buttons without loading.

We can raise a TDD later to change the AuthZButton to show loading instead of disable state if you think it's missing more clarity for the user about the authz checks.

H4ad (changes requested):

frontend/docs/authz-guide.md
AshwinBhatkal, line 33: @H4ad there is not state today we cannot access a resource but allow delete correct? withAuthzPage will check read permissions and block the whole page. Just wanted to call this out. Can be purely enforced on something like a the list page

H4ad: We can have a combination of permissions that allow to list + delete

Because the layout of dashboards has the delete button very different (where it's located) than the guidelines, we can make it a exception as long as we have a option to delete somewhere else

In case of dashboards, we can delete via list actions, so it's fine

After these PRs, I will update the guidelines to address this case as well

frontend/src/pages/DashboardsListPage/components/DashboardsList/DashboardsList.tsx
AshwinBhatkal: done

frontend/src/pages/DashboardsListPage/components/DashboardsList/DashboardsList.tsx
AshwinBhatkal: removed

frontend/docs/authz-guide.md
AshwinBhatkal: removed

frontend/src/lib/authz/components/README.md
AshwinBhatkal: removed

frontend/src/pages/DashboardsListPage/components/ActionsPopover/ActionsPopoverContent.tsx
AshwinBhatkal: I will discuss this. But need not be part of this PR

frontend/src/pages/DashboardsListPage/components/DashboardsList/DashboardsList.tsx
H4ad: If you click, the suggestions still available (and clickable)

frontend/src/pages/DashboardsListPage/components/ActionsPopover/ActionsPopoverContent.tsx
H4ad: Let's fix and ship in this PR instead of inside a follow-up, we discussed the alternatives for this we can go ahead with them:

- remove the loading
- block any delete/edit action with AuthZButton

H4ad (commented):
We should use the standard message from the authz components, update this and any other place that might have these custom messages

Image

frontend/src/pages/DashboardsListPage/components/DashboardsList/DashboardsList.tsx
AshwinBhatkal: handled

frontend/src/pages/DashboardsListPage/components/ActionsPopover/ActionsPopoverContent.tsx
AshwinBhatkal: fixed

frontend/src/pages/DashboardsListPage/components/DashboardsList/DashboardsList.tsx
AshwinBhatkal: done

frontend/src/pages/DashboardsListPage/components/ActionsPopover/ActionsPopoverContent.tsx
AshwinBhatkal: done

H4ad (dismissed):

H4ad (dismissed):

H4ad (approved):

PR #12450, opened 2026-08-06
fix(infrastructure-monitoring): do not overflow outside table for large error message

H4ad:
## Pull Request

---

Prevent the error message to overflow outside the table.

Before:

https://github.com/user-attachments/assets/d53845ed-7db9-4451-a2ca-31bc6127ec31

After:

https://github.com/user-attachments/assets/78e1e2d4-a5e1-46d7-9f66-117d1868e9f5

Closes https://github.com/SigNoz/pulse-pod/issues/222

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

- Tests added/updated: No
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Infrastructure Monitoring
- Potential regressions: -

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | We updated the error layout to ensure it won't overflow the table in case the APIs fail with a large message. |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/container/InfraMonitoringK8sV2/Base/K8sEmptyState.module.scss
+++ b/frontend/src/container/InfraMonitoringK8sV2/Base/K8sEmptyState.module.scss
@@ -51,4 +51,5 @@
 	gap: var(--spacing-3);
 	max-width: 500px;
 	padding: 24px;
+	max-height: 100%;
 }


AshwinBhatkal (approved):

AshwinBhatkal:
suggestion: what if we truncate it. show on hover or in details view

H4ad:
@AshwinBhatkal we can create a new issue for that, this page reuses the shared component `ErrorContent`, so changing the behavior will affect multiple pages. Would be better to keep this contained and then re-think the error messages in general if needed in a separated task, but I agree that truncating would be better than having large page to scroll.

PR #12453, opened 2026-08-06
fix(infrastructure-monitoring): page reset on switch category & page outside total

H4ad:
## Pull Request

---

This PR fixes the following issues:

- page not resetting to 1 when switch
  - bug was only detected/present when coming from deep link
- page not resetting to 1 when page produces a offset higher than total
  - you had to switch to hosts to be able to see data again

Before:

Issue with page not reseting to 1 when changing category (after refresh):

https://github.com/user-attachments/assets/00872b38-1263-43c1-8322-64d31ee1ee6a

Issue with page outside the offset:

https://github.com/user-attachments/assets/5194fb2e-5af3-491b-baf7-b4aa3a330c83

---

After:

Issue with page not reseting to 1 when changing category (after refresh):

https://github.com/user-attachments/assets/545e5914-c26f-4189-b15a-dc399bdee28b

Issue with page outside the offset:

https://github.com/user-attachments/assets/1b93d162-22a3-41c8-802e-aa2aa6012db9

Closes https://github.com/SigNoz/pulse-pod/issues/208

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

Both issues are caused after the refactor to the new table component and after joining the categories into single component (without unmount/mount when switching categories).

Lack of reset the page to 1, and no proper way to detect and reset page to 1 when outside the boundaries.

Reset to page 1 after switch category and also include hook on tanstack to ensure we reset page to last when outside the params.

---

- Tests added/updated: Yes
- Manual verification: Yes
- Edge cases covered: -

---

- Blast radius: Infrastructure Monitoring
- Potential regressions: -
- Rollback plan: Open a new PR to fix the issue

---

| Field | Value |
|------|-------|
| Change Type | Bug Fix |
| Description | We fixed two issues around pagination inside Infrastructure Monitoring causing the page not resetting to 1 after switch category or when offset is higher than total amount of items. |

---

### 📋 Checklist
- [x] Manually tested

### DIFF
--- a/frontend/src/components/TanStackTableView/useRecoverFromEmptyPage.ts
+++ b/frontend/src/components/TanStackTableView/useRecoverFromEmptyPage.ts
@@ -1,9 +1,23 @@
-import { useEffect } from 'react';
+import { useEffect, useRef } from 'react';
 
 import { SetPageOptions } from './useTableParams';
 
 const FIRST_PAGE = 1;
 const REPLACE_HISTORY: SetPageOptions = { history: 'replace' };
+
+/**
+ * How many single-page step-backs to attempt before giving up and going to page 1.
+ *
+ * A step-back only happens when `total` claims the current page should hold data but the
+ * response came back empty. Each hop costs a request, so an inflated `total` on a high page
+ * number would otherwise walk the user down one page at a time behind a spinner.
+ */
+const MAX_STEP_BACKS = 2;
+
+type Correction = {
+	from: number;
+	to: number;
+};
 
 export type UseRecoverFromEmptyPageParams = {
 	page: number;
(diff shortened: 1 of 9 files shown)


frontend/src/components/TanStackTableView/useRecoverFromEmptyPage.ts
```diff
@@ -0,0 +1,107 @@
...
+	page: number;
+	pageSize: number;
+	rowCount: number;
+	total: number;
+	isFetching: boolean;
+	isDisabled?: boolean;
+	setPage: (page: number, options?: SetPageOptions) => void;
+};
+
+export function useRecoverFromEmptyPage({
```
AshwinBhatkal, line 32: this seems a lot more complex than the problem itself. What about just resetting states the right way? 
Also for coming from another page, if there are no results, then give CTA to clear/reset filters. Maybe even keeping the pagination buttons when total is greater than 0 is good. So that user can go back

H4ad: The complexity is intentional since this is meant to be used as public API from `TanStackTableView`, while this fixes a bug on Infrastructure Monitoring, this could be used (and misused) by any part of the app using tables.

And since TanStack supports infinite mode as well, I prefer to keep this as simple hook that can be opt-in when needed.

> if there are no results, then give CTA to clear/reset filters. Maybe even keeping the pagination buttons when total is greater than 0 is good. So that user can go back

All these suggestions are improvements that can be implemented but for the current changes I preferred to just focus on the fix instead of possible changes in the behavior itself.

Eg:

- not hiding pagination:
  - needs https://github.com/SigNoz/components/pull/317
  - not released yet, part of 0.1.0, https://github.com/SigNoz/components/pull/264
- give CTA to clear/reset filters:
  - change the UI of no results
  - requires change on tanstack to allow render fallback on `empty data` inside the table body itself
  
I agree with the proposal changes but since I had limited time, I put effort to change/fix with less outside changes possible, introducing a hook that does not affect table directly and can be opt-in by places that needs it.

AshwinBhatkal (approved):

PR #12456, opened 2026-08-07
test(querier): pin keyless-row semantics for filter operators

srikanthccv:
## Summary

Adds an integration-test matrix that pins the deliberate keyless-row contract for filter operators, independent of any feature work:

- **Negative operators are a set complement over all rows.** A row that does not carry the key at all must match `!=`, `NOT IN`, `NOT LIKE`, and `NOT CONTAINS`. Users opt into presence explicitly with `AND key EXISTS`.
- **Positive operators carry an implicit existence guard** (`FilterOperator.AddDefaultExistsFilter`), so keyless rows never false-positive against sentinel defaults.
- **`EXISTS` / `NOT EXISTS` partition rows exactly** by key presence, and `!= x AND EXISTS` is the documented composition for "present and not x".
- **Numeric attributes inherit the map-default sentinel**: a missing key reads as `0`, so `num != 5` includes keyless rows while `num != 0` excludes them. This conflation is deliberate and pinned by name (`numeric_neq_zero_sentinel_conflation`) as the reference point for any value-expression change.

Coverage: 46 cases — one shared matrix over traces and logs (resource and attribute contexts), metric labels (series without the label), the numeric sentinel, and the EXISTS composition. The contract, matrix, seed data, and assertions live together in one file so the contract reads top to bottom.

## Why

These semantics were enforced only implicitly by the operator list in `AddDefaultExistsFilter`, with no test naming the intent. That gap allows an implementation change to alter negative-filter results silently and lets new tests calibrate expectations against the implementation instead of the contract. The attribute names used here are deliberately outside every semantic-convention family, so this file pins the base contract regardless of the semconv overlay state and serves as the oracle that family-field behavior (`queriertraces/13_semconv_evolution.py` in the semconv stack) must mirror.

## Testing

- `uv run pytest --basetemp=./tmp/ --reuse integration/tests/queriercommon/06_keyless_semantics.py` — 46/46 passed against a stack built from main-based sources, and 2×46 against a long-lived shared stack, confirming the set-based assertions stay stable under environment reuse.

### DIFF
--- a/tests/integration/tests/queriercommon/06_keyless_semantics.py
+++ b/tests/integration/tests/queriercommon/06_keyless_semantics.py
@@ -1,0 +1,314 @@
+"""Pins the keyless-row contract for filter operators, per signal.
+
+The contract (deliberate product semantics, enforced by
+`FilterOperator.AddDefaultExistsFilter` in
+pkg/types/querybuildertypes/querybuildertypesv5/builder_elements.go):
+
+  - Negative operators (!=, NOT IN, NOT LIKE, NOT CONTAINS, ...) are a set
+    complement over ALL rows: a row that does not carry the key at all MUST
+    match. Users opt into presence explicitly with `AND key EXISTS`.
+  - Positive operators carry an implicit existence guard: a keyless row must
+    NOT match `key = ''`-style comparisons against sentinel defaults.
+  - EXISTS / NOT EXISTS partition rows exactly by key presence.
+  - Numeric attributes inherit the map-default sentinel: a missing key reads
+    as 0, so `num != 0` excludes keyless rows while `num != 5` includes them.
+    This conflation is deliberate and pinned here as the reference for any
+    value-expression change (for example coalesce tails in semconv families).
+
+Any implementation change that makes these assertions fail is a behavior
+break, not a cleanup. Family-field behavior must mirror this matrix; see
+queriertraces/13_semconv_evolution.py.
+
+The attribute names used here are deliberately outside every semantic
+convention family so this file pins the base contract regardless of the
+semconv overlay state.
+"""
+
+from collections.abc import Callable, Generator
+from datetime import UTC, datetime, timedelta
+from http import HTTPStatus
+
+import pytest
+
+from fixtures import types
+from fixtures.auth import USER_ADMIN_EMAIL, USER_ADMIN_PASSWORD
+from fixtures.logs import Logs
+from fixtures.metrics import Metrics
+from fixtures.querier import (
+    BuilderQuery,
+    OrderBy,
+    RequestType,
+    TelemetryFieldKey,
+    aligned_epoch,
+    build_builder_query,
+    get_all_series,
+    get_column_data_from_response,
+    make_query_request,
+)
+from fixtures.traces import TraceIdGenerator, Traces, TracesKind, TracesStatusCode
+
+PREFIX = "keyless-sem"
+STRING_KEY = "tenant.tier"
+NUMBER_KEY = "retry.count"
+METRIC_NAME = "keyless_semantics_gauge"
+METRIC_LABEL = "tenant_tier"
+
+# Row identities, keyed by which value of the string key they carry.
+GOLD = f"{PREFIX}-gold"
+SILVER = f"{PREFIX}-silver"
+NONE = f"{PREFIX}-none"  # carries neither the string nor the number key
+
+# One matrix, three signals. Each case: (filter over the string key, expected
+# row identities). The keyless row's membership is the point of every case.
+STRING_MATRIX = [
+    pytest.param("{key} = 'gold'", {GOLD}, id="eq_excludes_keyless"),
+    pytest.param("{key} != 'gold'", {SILVER, NONE}, id="neq_includes_keyless"),
+    pytest.param("{key} NOT IN ['gold', 'silver']", {NONE}, id="not_in_includes_keyless"),
+    pytest.param("NOT {key} LIKE '%gold%'", {SILVER, NONE}, id="not_like_includes_keyless"),
+    pytest.param("{key} NOT CONTAINS 'gol'", {SILVER, NONE}, id="not_contains_includes_keyless"),
+    pytest.param("{key} EXISTS", {GOLD, SILVER}, id="exists_partitions"),
+    pytest.param("{key} NOT EXISTS", {NONE}, id="not_exists_partitions"),
+    # The documented idiom for "present and not X": composition, not a new
+    # operator semantic.
+    pytest.param("{key} != 'gold' AND {key} EXISTS", {SILVER}, id="neq_composed_with_exists"),
+]
+
+# Numeric attributes read the map default (0) for missing keys. `!= 0` is the
+# deliberate blind spot: a keyless row is indistinguishable from a stored 0.
+NUMBER_MATRIX = [
+    pytest.param("{key} = 0", {GOLD}, id="numeric_eq_zero_excludes_keyless"),
+    pytest.param("{key} != 5", {GOLD, NONE}, id="numeric_neq_includes_keyless"),
+    pytest.param("{key} != 0", {SILVER}, id="numeric_neq_zero_sentinel_conflation"),
+]
+
+
+@pytest.fixture(name="keyless_rows")
+def keyless_rows(
+    insert_logs: Callable[[list[Logs]], None],
+    insert_traces: Callable[[list[Traces]], None],
+) -> Generator[datetime]:
+    now = datetime.now(tz=UTC).replace(microsecond=0) - timedelta(minutes=1)
+
+    def resources(identity: str, tier: str | None) -> dict:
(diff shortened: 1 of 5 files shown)


tests/integration/tests/queriercommon/06_keyless_semantics.py
```diff
@@ -0,0 +1,314 @@
...
+    pytest.param("{key} = 'gold'", {GOLD}, id="eq_excludes_keyless"),
+    pytest.param("{key} != 'gold'", {SILVER, NONE}, id="neq_includes_keyless"),
+    pytest.param("{key} NOT IN ['gold', 'silver']", {NONE}, id="not_in_includes_keyless"),
+    pytest.param("NOT {key} LIKE '%gold%'", {SILVER, NONE}, id="not_like_includes_keyless"),
+    pytest.param("{key} NOT CONTAINS 'gol'", {SILVER, NONE}, id="not_contains_includes_keyless"),
+    pytest.param("{key} EXISTS", {GOLD, SILVER}, id="exists_partitions"),
+    pytest.param("{key} NOT EXISTS", {NONE}, id="not_exists_partitions"),
+    # The documented idiom for "present and not X": composition, not a new
+    # operator semantic.
+    pytest.param("{key} != 'gold' AND {key} EXISTS", {SILVER}, id="neq_composed_with_exists"),
```
tushar-signoz, line 65: Should we add {key} EXISTS for all other NOT variants ?

tests/integration/tests/queriercommon/06_keyless_semantics.py
```diff
@@ -0,0 +1,314 @@
...
+# deliberate blind spot: a keyless row is indistinguishable from a stored 0.
+NUMBER_MATRIX = [
+    pytest.param("{key} = 0", {GOLD}, id="numeric_eq_zero_excludes_keyless"),
+    pytest.param("{key} != 5", {GOLD, NONE}, id="numeric_neq_includes_keyless"),
+    pytest.param("{key} != 0", {SILVER}, id="numeric_neq_zero_sentinel_conflation"),
+]
+
+
+@pytest.fixture(name="keyless_rows")
+def keyless_rows(
```
tushar-signoz: Lets move it to the fixture files with a proper name

tests/integration/tests/queriercommon/06_keyless_semantics.py
```diff
@@ -0,0 +1,314 @@
...
+                attributes=attributes(tier, retries),
+            )
+            for identity, tier, retries, offset in rows
+        ]
+    )
+    yield now
+
+
+@pytest.fixture(name="keyless_series")
+def keyless_series(insert_metrics: Callable[[list[Metrics]], None]) -> Generator[tuple[int, int]]:
```
tushar-signoz: Same, name and move to fixtures

tests/integration/tests/queriercommon/06_keyless_semantics.py
```diff
@@ -0,0 +1,314 @@
...
+
+def _matching_rows(
+    signoz: types.SigNoz,
+    token: str,
+    now: datetime,
+    signal: str,
+    identity_field: str,
+    identity_column: str,
+    expression: str,
+) -> set[str]:
```
tushar-signoz: This is making it difficult to understand what it is asserting, can we not have this function and do simple assertions in the test themselves ?

tests/integration/tests/queriercommon/06_keyless_semantics.py
```diff
@@ -0,0 +1,314 @@
...
+    # operator semantic.
+    pytest.param("{key} != 'gold' AND {key} EXISTS", {SILVER}, id="neq_composed_with_exists"),
+]
+
+# Numeric attributes read the map default (0) for missing keys. `!= 0` is the
+# deliberate blind spot: a keyless row is indistinguishable from a stored 0.
+NUMBER_MATRIX = [
+    pytest.param("{key} = 0", {GOLD}, id="numeric_eq_zero_excludes_keyless"),
+    pytest.param("{key} != 5", {GOLD, NONE}, id="numeric_neq_includes_keyless"),
+    pytest.param("{key} != 0", {SILVER}, id="numeric_neq_zero_sentinel_conflation"),
```
tushar-signoz, line 71: This is unexpected, this shouldn't be the case.

tushar-signoz: numeric_neq_zero_sentinel_conflation case

tushar-signoz: We do `toFloat64(attributes_number['code.line.number'])` for numbers. ignore the comment

srikanthccv: I know, and I called out this explicitly

>to ensure none of the sem conv changes anything behaviourally or semantically compared to existing version, even if there are some gotchas with the current behaviour

my goal was to ensure, nothing about the sem conv changes breaks the existing behaviour, and neither should it attempt to fix any exiting behaviour. The test here is purely a documented behaviour. And when that behaviour is modified this gets corrected but that's not in the scope of the sem conv changes.

tests/integration/tests/queriercommon/06_keyless_semantics.py
srikanthccv: ok, so I moved but no one really uses other than the querier common sem conv tests so it didn't make sense to keep it in fixtures yet so I undid the changes (in the last push), and I believe it would be premature to move it to fixtures, and let it earn it's place in fixtures when there is need. Let me know what do we think about it so we can update docs, otherwise same discussion will come up.

tests/integration/tests/queriercommon/06_keyless_semantics.py
tushar-signoz: @srikanthccv If semconv changes will be applied to body as well then we should include body test cases as well.

tests/integration/tests/queriercommon/06_keyless_semantics.py
srikanthccv: Don't see a value in it because `key EXISTS` is the explicit part of the expression and doesn't change behaviour because of the prior condition i.e `x AND key EXISTS` is same regardless of the which variant of `x` is. The test matrix main target is without explicit `key EXISTS` returns rows without key as well. And one case with `key EXISTS` is enough to show the when explicit `key EXISTS` is added it doesn't include the non key present rows.

tests/integration/tests/queriercommon/06_keyless_semantics.py
srikanthccv: it returns the list of `identity_field` values and that list is asserted against.

srikanthccv: would we rather repeat the code of the function block or have the function document what it is asserting? let me know and let's document.

either way, one has to go through the test cases to understand what's going on.

srikanthccv: if you noticed, every coding assistant does this, where it prefer the function over the duplicated code, despite having the personal rules discouraging this, it ends up going this way, so trying to correct each such instance is misplaced effort and time waste IMO.

tushar-signoz: Yes claude prefers _ helper functions a lot, but we are following no _ underscore named functions, If its fixture keep it in the fixtures folder, if it doesn't fit there then its probably a helper and not a fixture, integration tests should be understandable to us humans then worrying about code duplicity policy and lets invest some extra effort to maintain it that way.
We should invest in developing skills and rule for claude in the repo itself to avoid revisiting this discussion again and again.

tests/integration/tests/queriercommon/06_keyless_semantics.py
```diff
@@ -0,0 +1,209 @@
...
+]
+
+SIGNALS = [
+    pytest.param("traces", "span.name", "name", id="traces"),
+    pytest.param("logs", "body", "body", id="logs"),
+]
+
+
+def _raw_query(signal: str, identity_field: str, expression: str) -> dict:
```
tushar-signoz: I think we have an helper already for this, build_something_something

tests/integration/tests/queriercommon/06_keyless_semantics.py
srikanthccv: done

tests/integration/tests/queriercommon/06_keyless_semantics.py
srikanthccv: yes, with each signal, I will have the integration tests to ensure the behaviour remain unchanged and handled.

tests/integration/tests/queriercommon/06_keyless_semantics.py
srikanthccv: correct

tushar-signoz (approved):
LGTM

therealpandey (approved):

PR #12458, opened 2026-08-07
feat(user): enable FGA for users and their role assignments

vikrantgupta25:
#### Description

- Enables per-resource FGA on the v2 user routes so custom roles can be scoped to individual users; user↔role link routes check attach/detach on both sides.
- Creating a user with roles now requires `user:attach` + `role:attach` (per role) alongside `user:create`, so `user:create` alone can no longer mint a `signoz-admin` via the invite body.
- Creating a reset password token requires `user:attach` on the target user plus the token scope; the reset-token routes now sit under a `reset-password-token` resource (renamed from `factor-password`).
- Self mutation (roles, profile, delete) is allowed for non-root users — grants are the only boundary; the root user stays fully protected.
- Migration `add_user_authz_tuples` backfills the user and reset-password-token tuples for existing orgs and refreshes the managed-role transaction_groups JSON.

#### Additional Information

- Community behaviour is unchanged: the fallback role gate remains admin-only.
- Integration coverage lives in `tests/integration/tests/passwordauthn/`: `09_fga.py` for per-object scoping and create-with-roles escalation, `08_user.py` for the managed-role matrix and self-delete.


vikrantgupta25:
will open this again. stale now

PR #12460, opened 2026-08-07
chore: add pytest conventions rule and apply it across integration tests

therealpandey:
#### Description

- Add `.claude/rules/pytest.md` with conventions for the Python integration suite. The lead rule: **no `_`-prefixed helper functions in test modules** — inline the logic; repetition across tests is cheaper than indirection; genuinely shared machinery becomes a fixture. Fixtures live in `tests/fixtures/` only, never under `integration/tests/` — with one exception: SigNoz-level fixtures (a suite spinning up SigNoz with different envs via `create_signoz`/`create_migrator`) always belong in that suite's `conftest.py`. Plus: fixture-factory over indirect parametrization, skip at collection, config via explicit `--flags`, snake_case parametrize ids, and collection gotchas (`python_files` prefix matching, `--import-mode=importlib`).
- Apply the no-`_helper` rule across `tests/integration`: all 40 module-level `_` helpers eliminated in dashboard, inframonitoring, promqlconformance, querier_json_body, querierlogs, queriermetrics, and queriertraces. Pure transforms and request wrappers were inlined at their call sites; case-table verifier callables became data flags with inline branches; the two 115-line resource-evolution mega-helpers folded into parametrized tests; shared machinery moved to `tests/fixtures/` as fixture-factories (`wipe_all_dashboards`, `load_pods_metrics`, `run_query_case`) registered via `pytest_plugins`.
- Apply the py-comments rule across `tests/`: drop module docstrings that restate the filename, relocate the ones carrying real constraints next to the code they constrain, delete function/class docstrings that restate the identifier, trim narrated steps and Args/Returns boilerplate.
- Fix camelCase parametrize ids in `queriermetrics/01_fill.py` (`fillGaps`/`fillZero` → `fill_gaps`/`fill_zero`).

#### Additional Information

- All 629 tests in the touched suites collect cleanly; `py-fmt`/`py-lint`/compileall pass. Runtime verification against the docker stack was not run.


tests/fixtures/dashboards.py
```diff
@@ -0,0 +1,36 @@
...
+# MaxListLimit caps a single list page, so wiping a shared DB has to drain pages
+# until the list comes back empty.
+MAX_LIST_LIMIT = 200
+
+
+@pytest.fixture(name="wipe_all_dashboards", scope="function")
+def wipe_all_dashboards(signoz: types.SigNoz) -> Callable[[str], None]:
+    def _wipe_all_dashboards(token: str) -> None:
+        while True:
+            response = requests.get(
```
tushar-signoz, line 19: Should it create method like make_get_dashboard_request same as we have for query_range, make_query_request ?

tests/fixtures/querier.py
```diff
@@ -1109,3 +1111,51 @@ def make_scalar_query_request(
             "formatOptions": {"formatTableResultForUI": True, "fillGaps": False},
         },
     )
+
+
+@pytest.fixture(name="run_query_case", scope="function")
+def run_query_case(signoz: types.SigNoz) -> Callable[[str, datetime, dict[str, Any]], None]:
```
tushar-signoz, line 1117: Generalise it a bit more and it should get more purpose. 
Add a param for requestType eg:  RequestType.TIME_SERIES,  etc etc.

tests/fixtures/dashboards.py
therealpandey: I will fix what is a fixture and what is a function in a follow up

krsoninikhil (approved):

tests/fixtures/querier.py
therealpandey: This is just being used in `querier_json_logs`. I will get rid of this when I put in a rule for what is a fixture vs what is a function

tushar-signoz (approved):

tushar-signoz:
@therealpandey Run the claude to check the semconv integration file in queriercommon

therealpandey:
> @therealpandey Run the claude to check the semconv integration file in queriercommon

Already in my TODOs.

